PR #2349, opened 2020-04-06
[GL backend] ensure that Handle<> are synchronously constructed

pixelflinger:
Until now, only the memory of Hw handes was allocated synchronously,
handles were constructed later asynchronously. This was error prone
with the few (but growing) synchronous calls we have.

Now we have an init<> call that allocates and constructs handles synchronously,
the asynchronous construction if needed, is still performed with construct<>,
which is now implemented as "destroy+construct", this is okay because
most of our handles have trivial dtors.

This fixes a race with TimerQuery and Fence. Fences are still not fully 
working though, but at least won't access uninitialized memory.
Currently backend fences MUST be used trough FFence.

### DIFF
--- a/filament/backend/src/DriverBase.h
+++ b/filament/backend/src/DriverBase.h
@@ -51,12 +51,13 @@ struct HwBase {
 };
 
 struct HwVertexBuffer : public HwBase {
-    AttributeArray attributes;            // 8 * MAX_VERTEX_ATTRIBUTE_COUNT
-    uint32_t vertexCount;                 //   4
-    uint8_t bufferCount;                  //   1
-    uint8_t attributeCount;               //   1
+    AttributeArray attributes{};          // 8 * MAX_VERTEX_ATTRIBUTE_COUNT
+    uint32_t vertexCount{};               //   4
+    uint8_t bufferCount{};                //   1
+    uint8_t attributeCount{};             //   1
     uint8_t padding[2]{};                 //   2 -> total struct is 136 bytes
 
+    HwVertexBuffer() noexcept = default;
     HwVertexBuffer(uint8_t bufferCount, uint8_t attributeCount, uint32_t elementCount,
             AttributeArray const& attributes) noexcept
             : attributes(attributes),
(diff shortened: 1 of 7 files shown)


filament/backend/src/DriverBase.h
```diff
@@ -67,61 +68,67 @@ struct HwVertexBuffer : public HwBase {
...
 
 struct HwUniformBuffer : public HwBase {
 };
 
 struct HwTexture : public HwBase {
+    uint32_t width{};
+    uint32_t height{};
+    uint32_t depth{};
+    SamplerType target{};
+    uint8_t levels : 4;  // This allows up to 15 levels (max texture size of 32768 x 32768)
```
romainguy, line 114: These two don't have the trailing initializer call, could that cause an issue?

filament/backend/src/opengl/OpenGLDriver.cpp
```diff
@@ -358,12 +358,25 @@ HandleBase::HandleId OpenGLDriver::allocateHandle(size_t size) noexcept {
     return HandleBase::HandleId(offset >> HandleAllocator::MIN_ALIGNMENT_SHIFT);
 }
 
+template<typename D, typename ... ARGS>
+backend::Handle<D> OpenGLDriver::init(ARGS&& ... args) noexcept {
```
romainguy, line 362: `initHandle()`?

filament/backend/src/opengl/OpenGLDriver.cpp
```diff
@@ -358,12 +358,25 @@ HandleBase::HandleId OpenGLDriver::allocateHandle(size_t size) noexcept {
...
+}
+
+
 template<typename D, typename B, typename ... ARGS>
 typename std::enable_if<std::is_base_of<B, D>::value, D>::type*
 OpenGLDriver::construct(Handle<B> const& handle, ARGS&& ... args) noexcept {
     assert(handle);
-    static_assert(sizeof(D) <= 208, "Handle<> too large");
     D* addr = handle_cast<D *>(const_cast<Handle<B>&>(handle));
+    addr->~D();
```
romainguy, line 379: This could be worth a comment.

bejado:
Seems like Metal and Vulkan will need to do something similar?

pixelflinger:
@bejado about Metal/Vulkan backends, it depends :-) but for instance for TimerQuery, we have a synchronous call (getTimerQueryValue) that takes the handle as parameter, and it cannot rely on the asynchronous part of `createTimerQuery` to have executed. So, you probably need something similar, at least make sure HwTimerQuery is zero-initialized, so `elapsed` doesn't contain garbage.

PR #2353, opened 2020-04-07
better GPU timings on Android

pixelflinger:
On drivers that don't implement timer queries properly, we use
fences instead, however we can't use a fence to determine the start
time of the GPU work (only the end time). To workaround this we
use the last call to glFlush() as a proxy.
At least on Pixel it works very well.

We also remove all mid-frame calls to flush(), we did this to kick
cpu work early, but it's not as efficient as overlapping the work
with the previous frame, it increases latency a bit though.

### DIFF
--- a/filament/backend/src/opengl/TimerQuery.cpp
+++ b/filament/backend/src/opengl/TimerQuery.cpp
@@ -39,6 +39,9 @@ TimerQueryNative::TimerQueryNative(OpenGLContext& context)
 
 TimerQueryNative::~TimerQueryNative() = default;
 
+void TimerQueryNative::flush() {
+}
+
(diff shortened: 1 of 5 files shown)


filament/backend/src/opengl/TimerQuery.cpp
```diff
@@ -61,6 +64,7 @@ uint64_t TimerQueryNative::queryResult(GLTimerQuery* query) {
     // IOS doesn't have glGetQueryObjectui64v, we'll never end-up here on ios anyways
 #ifndef IOS
     glGetQueryObjectui64v(query->gl.query, GL_QUERY_RESULT, &elapsedTime);
+    utils::slog.d << double(elapsedTime) / 1e6 << utils::io::endl;
```
romainguy, line 67: Remove?

Issue #2356, opened 2020-04-07
Should Gltfio.init() call Filament.init() like Utils.init() does?


alexey-pelykh:
**Describe the bug**
It seems that after recent changes between v1.4.5 and v1.5.2 due to `registerCallbackUtils(env);` is not called from Gltfio.cpp

**To Reproduce**
Start Sceneform demo with Filament 1.5.2

**Expected behavior**
It does not crash

**Screenshots**
It crashes with 
```
A/ot.arsdkdemoap: java_vm_ext.cc:570] JNI DETECTED ERROR IN APPLICATION: CallStaticIntMethodV received NULL jclass
    java_vm_ext.cc:570]     in call to CallStaticIntMethodV
    java_vm_ext.cc:570]     from long com.google.android.filament.Material.nBuilderBuild(long, java.nio.Buffer, int)
```

alexey-pelykh:
Or,
https://github.com/google/filament/blob/887f8e8a1c8d6dcd85188d1b77bfb4b4d4465557/android/filament-utils-android/src/main/java/com/google/android/filament/utils/Utils.kt#L27
should be in
https://github.com/google/filament/blob/887f8e8a1c8d6dcd85188d1b77bfb4b4d4465557/android/gltfio-android/src/main/java/com/google/android/filament/gltfio/Gltfio.java#L23
as well

prideout:
Our demos use `Utils` rather than `gltfio` so I think this oversight had gone unnoticed.  Thanks for the PR :)

PR #2357, opened 2020-04-07
Fix Gltfio.init(): call Filament.init() as well

alexey-pelykh:
Addresses #2356


alexey-pelykh:
@googlebot I signed it!

PR #2358, opened 2020-04-07
Do not clobber VK_ICD_FILENAMES.

### LINKED ISSUE #2343: Stop overriding VK_ICD_FILENAMES on macOS
**Describe the bug**
```
    const Path dylibPath = executableFolder.concat(VKLIBRARY_PATH);
    const Path jsonPath = executableFolder.concat("MoltenVK_icd.json");
    setenv("VK_ICD_FILENAMES", jsonPath.c_str(), 1);
```

**To Reproduce**
  1. set `VK_ICD_FILENAMES` to something
  2. Run on macOS

**Expected behavior**
Vulkan Loader interface is respected, and the user-chosen ICD is loaded.

 - OS: macOS 10.15
 - Backend: Vulkan

**Additional context**
Running with [gfx-portability](https://github.com/gfx-rs/portability).

prideout:
Fixes #2343


prideout:
That's a good idea, I forgot that we provided an exists method.

prideout:
On second thought this variable can be a list of colon-separated paths, I don't think we should be expected to parse and validate it.  Let's simply avoid clobbering it.

PR #2362, opened 2020-04-08
fix shader overflows leading to black pixels

pixelflinger:
we clamp the output of post_process materials to 0,FLT16_MAX

### DIFF
--- a/shaders/src/post_process.fs
+++ b/shaders/src/post_process.fs
@@ -5,5 +5,9 @@ void main() {
     // Invoke user code
     postProcess(inputs);
 
+#if defined(TARGET_MOBILE)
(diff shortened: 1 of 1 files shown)


shaders/src/post_process.fs
```diff
@@ -5,5 +5,9 @@ void main() {
     // Invoke user code
     postProcess(inputs);
 
+#if defined(TARGET_MOBILE)
+    inputs.color = clamp(inputs.color, 0.0, MEDIUMP_FLT_MAX);
```
romainguy, line 9: We have a function called saturateMediump() for this you should add a vec3 variant. It's in common_math.fs I think.

PR #2370, opened 2020-04-09
Only build x86 and arm64-v8a in presubmit

romainguy:


romainguy:
Questions for @bejado and @pixelflinger:
- Should we build only x86 and arm4-v8a in continuous as well?
- Should continuous builds build debug *and* release as they do today?

bejado:
> Should we build only x86 and arm4-v8a in continuous as well?

I don't have much of an opinion on this. Have we seen build issues that only manifest in a specific architecture? If not, it's probably fine to only build a subset of architectures. What I'd hate to see happen though is come release time we realize there's an issue in a non-x86 / arm4-v8a arch.

> Should continuous builds build debug and release as they do today?

I think it's worth building both debug and release, because we've seen issues before that are unique to debug builds.

PR #2374, opened 2020-04-09
Implement headless swapchains on Windows

### LINKED ISSUE #2365: headless swapchain/readPixels test doesn't work on windows + gl
**Describe the bug**
I compiled a very small program based on filament_rendering_test.cpp to check why readPixels isn't working as expected in my code. The assert fails as the buffer is filled with 0.
I was having the issue with 1.4.* and I can reproduce it with 1.5.2

**To Reproduce**
1. here's the code I compiled and tested in debug: 
edited: created a gist instead because the code formatting didn't seem to work
https://gist.github.com/teub/0b03062c51332d7772572ba02ee42e40

libraries linked:
`
lib/x86_64/mdd/filament.lib
lib/x86_64/mdd/utils.lib
lib/x86_64/mdd/backend.lib
lib/x86_64/mdd/filamat.lib
lib/x86_64/mdd/filaflat.lib
lib/x86_64/mdd/filabridge.lib
lib/x86_64/mdd/geometry.lib
lib/x86_64/mdd/ibl.lib
lib/x86_64/mdd/bluegl.lib
opengl32.lib
`

**Expected behavior**
Asserts should pass, they correspond to the tests in filament_rendering_test.cpp

**Desktop :**
 - OS: Windows
 - GPU: NVIDIA GTX 1080
 - Backend: OpenGL

bejado:
Fixes #2365

### DIFF
--- a/filament/backend/src/opengl/PlatformWGL.cpp
+++ b/filament/backend/src/opengl/PlatformWGL.cpp
@@ -70,6 +70,12 @@ namespace filament {
 
 using namespace backend;
 
+struct WGLSwapChain {
+    HDC hDc = NULL;
+    HWND hWnd = NULL;
+    bool isHeadless = false;
+};
+
 Driver* PlatformWGL::createDriver(void* const sharedGLContext) noexcept {
     mPfd = {
         sizeof(PIXELFORMATDESCRIPTOR),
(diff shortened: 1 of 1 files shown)


filament/backend/src/opengl/PlatformWGL.cpp
```diff
@@ -183,8 +183,22 @@ Platform::SwapChain* PlatformWGL::createSwapChain(void* nativeWindow, uint64_t&
...
+    // WS_POPUP was chosen for the window style here after some experimentation.
+    // For some reason, using other window styles resulted in corrupted pixel buffers when using
+    // readPixels.
+    RECT rect = {0, 0, width, height};
+    AdjustWindowRect(&rect, WS_POPUP, FALSE);
+    width = rect.right - rect.left;
+    height = rect.bottom - rect.top;
+
+    // TODO: this window never gets destroyed.
+    HWND hWnd = CreateWindowA("STATIC", "headless", WS_POPUP, 0, 0,
```
romainguy, line 195: Could we make this a member of the SwapChain so we can destroy it when the SwapChain is destroy for instance? This window could be large, we really need to clean it up.

pixelflinger: I agree it's important because typically we will be running unit tests headless, could be many many in a row.

Issue #2369, opened 2020-04-08
Clearing multiple views is confusing


romainguy:
Here is what happens when rendering 4 views called `view0` to `view3`. `view1`, `view2` and `view3` have the clear flags disabled and their render target buffer flags is set to `NONE`:

```
        view1.setRenderTarget(null, View.TargetBufferFlags.NONE)
        view2.setRenderTarget(null, View.TargetBufferFlags.NONE)
        view3.setRenderTarget(null, View.TargetBufferFlags.NONE)

        // Clears are not restricted by the scissor so we want to disable clear targets
        // except for the first view we render
        view1.setClearTargets(false, false, false)
        view2.setClearTargets(false, false, false)
        view3.setClearTargets(false, false, false)

        view0.setClearColor(0.035f, 0.035f, 0.035f, 1.0f)
```

The result shows that the background surface wasn't cleared entirely by `view0`, or that its content was not loaded when the subsequent views started rendering (I don't believe we ever preserve tile content?).

<img width="878" alt="Screen Shot 2020-04-08 at 3 55 22 PM" src="https://user-images.githubusercontent.com/869684/78841633-bc638600-79b2-11ea-8b15-1a60b9a1c103.png">

This means the only way to show multiple views with a color background is to not rely on clear color but instead to use a skybox or equivalent (clip-space triangle/quad). This API is very confusing and requires a lot of work on the user's part to get it right. We need to fix this.

romainguy:
CC @rawnsley

romainguy:
@pixelflinger BTW the screenshot above is from the emulator. I didn't test on an actual device so it may work there.

romainguy:
PR #2371 fixes some of the problem.

Issue #2372, opened 2020-04-09
How to render multi overlay view in same viewport? (Win10 OpenGL )


BillyUSA:
I want to render two view with differen scene/camera in same viewport ( win32 window ),
if both view without PostProcessing, can see the overlay result, but not smoothed ( no AntiAliasing );
if the last rendered view with PostProcessing(MSAA /FXAA) enabled，the reult is smoothed，but can't see the first view .

How to render multi overlay view with  PostProcessing ( MSAA /FXAA ) Enabled ?

romainguy:
Are the two views side by side or on top of each other?

BillyUSA:
two views are on top of each other

romainguy:
What clear flags and target buffer flags do you use?

@pixelflinger

BillyUSA:
view 0 :  setClearTargets(true, true, true)
view 1 :  setClearTargets(false, true, true)

romainguy:
Thanks you could also try to call setRenderTarget(nullptr, View:: TargetBufferFlags::NONE) on the the top view?

BillyUSA:
nothing changed after setRenderTarget(nullptr, View:: TargetBufferFlags::NONE)   on the the top view.

romainguy:
This is fixed by https://github.com/google/filament/pull/2376. See the PR description to understand what APIs to use.

BillyUSA:
#2376 when setPostProcessingEnabled(true) and BlendMode with TRANSLUCENT , the 2nd view cover and erase the 1st view .
Is there example about  one view  drawn on top of another ?

romainguy:
How do you setup the background of the top view? @pixelflinger

Issue #2364, opened 2020-04-08
Multiple views no longer working on Android


rawnsley:
Rendering multiple views in a single frame hasn't worked for a while. This is [already documented](https://github.com/google/filament/pull/1916), but I wanted to raise it's profile a little and provide an easy reproduce.

It we modify the lit-cube sample to have four views arranged in quadrants ([patch file](https://github.com/google/filament/files/4451048/patch.diff.zip)) we get:

![before](https://user-images.githubusercontent.com/303516/78796606-73113780-79ae-11ea-9d52-fe217610c551.png)

before [this change](https://github.com/google/filament/pull/1916/commits/7d0f2fff343f2e1aca18cee47f12a9cbbab4619b). After the change we get the following output:

![after](https://user-images.githubusercontent.com/303516/78797203-4b6e9f00-79af-11ea-88c5-748ae026a0ab.png)

I appreciate that multiple views is a minority use case, but it might be worth having at least one sample that demonstrates it to avoid retros.

romainguy:
Have you tried wit the latest version of Filament. There were several fixes related to multi-view rendering recently.

romainguy:
@pixelflinger Looks like with the patch linked above we always invalidate a framebuffer if we use several times (for several views). It used to work I believe because of the test `currentFbo != newFbo`.

rawnsley:
@romainguy Yes - this is straight from today's master branch at [68a78a77](https://github.com/google/filament/commit/68a78a77a85f2fbe5dce4d9c5fbe859ca6fd014f).

pixelflinger:
@rawnsley are you setting the discard flags on the supplemental views? i.e. like this:

`view.setRenderTarget(TargetBufferFlags::NONE);`

You need to do that for all views, except the first one.

rawnsley:
@pixelflinger Is this possible through the JNI interface?

romainguy:
Looks like the Java API is missing the TargetBufferFlags parameter and overload.

romainguy:
We should also add a sample app to explain this behavior.

romainguy:
I just checked in the missing API and a sample app.

Note it does *not* fix the issue entirely. You have to disable color clears after the 1st view and the only way to show a background is to use a skybox of equivalent. I'm going to file a new bug so we can fix how clear/clear color works in multi view mode.

rawnsley:
Can confirm fixed for me. Thank you so much!

PR #2376, opened 2020-04-10
API BREAKAGE: this change aims to fix multi-view support

pixelflinger:
What has changed:
- View doesn't have a notion of clear color anymore
- View doesn't have a notion of discard flags anymore
- The clear color and color-buffer discard/clear flags are moved to Renderer
- Skybox can now be set to a constant color
- View have a blend-mode

What does is all mean:
"Clearing" (i.e.) setting its background is now handled by Skybox, by
setting a constant color to the Skybox. This should take care of
drawing views side by side.

When a view needs to be drawn on top of another, it's BlendMode needs to
be set to TRANSLUCENT and of course and, generally, it wither won't have
a skybox, or will have one that sets some translucent pixels.

As an optimization, a View with BlendMode::OPAQUE will have its
background cleared with the color specified in Renderer.

If the SwapChain already has some content, it's now possible to set
the Renderer to not discard the content, together with TRANSLUCENT views,
it's possible to draw on top of that content.

It is NOT possible to share depth/stencil buffers between views.

Fixes: #2369, #2372, #2364


filament/src/materials/skybox.mat
```diff
@@ -23,8 +31,13 @@ material {
...
         prepareMaterial(material);
-        vec3 sky = texture(materialParams_skybox, variable_eyeDirection.xyz).rgb;
-        sky *= frameUniforms.iblLuminance;
+        vec4 sky;
+        if (materialParams.constantColor) {
+            sky = materialParams.color;
+        } else {
+            sky = vec4(texture(materialParams_skybox, variable_eyeDirection.xyz).rgb, 1.0);
+        }
+        sky.rgb *= frameUniforms.iblLuminance;
```
romainguy, line 40: Should we ignore iblLuminance in constant color mode?

samples/frame_generator.cpp
```diff
@@ -329,14 +334,15 @@ static void setup(Engine* engine, View* view, Scene* scene) {
     if (!g_skyboxOn) {
         auto ibl = FilamentApp::get().getIBL();
         if (ibl) ibl->getSkybox()->setLayerMask(0xff, 0x00);
+    } else {
+        g_skybox = Skybox::Builder().color({
+                ((g_clearColor >> 16) & 0xFF) / 255.0f,
+                ((g_clearColor >>  8) & 0xFF) / 255.0f,
+                ((g_clearColor      ) & 0xFF) / 255.0f,
+                1.0f
+        }).build(*engine);
```
romainguy, line 343: Shouldn't iblLuminance be set to 1 in this case?

filament/src/materials/skybox.mat
pixelflinger: I was wondering about that? Should we ignore, or just let the caller set it to 1?
I thought it would be nice to have the background color behave just like a contant skybox, especially we can draw the sun?

I was also wondering if we need a dedicated material, to avoid the test?

samples/frame_generator.cpp
pixelflinger: probably? :)

android/samples/sample-multi-view/src/main/java/com/google/android/filament/multiview/MainActivity.kt
```diff
@@ -406,10 +406,18 @@ class MainActivity : Activity() {
                 // If beginFrame() returns false you should skip the frame
                 // This means you are sending frames too quickly to the GPU
                 if (renderer.beginFrame(swapChain!!, frameTimeNanos)) {
+                    scene.skybox = skybox0
                     renderer.render(view0)
+
+                    scene.skybox = skybox1
```
prideout, line 412: As an alternative to maintaining multiple skybox objects, can multi-view apps simply have 1 skybox that they mutate between calls to  `renderer.render(view)`?  In this particular demo, the background color of the skybox could be adjusted before each call to render().

If this is allowed, then I suggest that we add a default Skybox to Scene.  This would greatly reduce the amount of boilerplate that apps need to write.

bejado:
A couple iOS samples use the old clear API and are failing to compile:

```
ios/samples/transparent-rendering/transparent-rendering/FilamentView.mm
167:    filaView->setClearColor({0.0f, 0.0f, 0.0f, 0.0f});

ios/samples/hello-triangle/hello-triangle/FilamentView.mm
156:    filaView->setClearColor({0.1, 0.125, 0.25, 1.0});
```

filament/src/Renderer.cpp
```diff
@@ -258,43 +258,44 @@ void FRenderer::renderJob(ArenaScope& arena, FView& view) {
 
     FrameGraph fg(engine.getResourceAllocator());
 
-    const TargetBufferFlags discardedFlags = view.getDiscardedTargetBuffers();
+    const TargetBufferFlags discardedFlags = mDiscardedFlags;
+    const TargetBufferFlags clearFlags = mClearFlags;
+    const float4 clearColor = mClearOptions.clearColor;
 
-    TargetBufferFlags clearFlags = (view.getClearFlags() & TargetBufferFlags::COLOR);
+    // once we've done the first View, we don't discard or clear the color buffer for the next ones
```
bejado, line 265: This comment is confusing to me.. does this mean we only clear color for the first view (if clear is enabled)?

filament/src/materials/skybox.mat
pixelflinger: okay I have decided to not multiply by iblLuminance.

android/samples/sample-multi-view/src/main/java/com/google/android/filament/multiview/MainActivity.kt
pixelflinger: yes. will do.

filament/src/Renderer.cpp
pixelflinger: okay will update comment (then you can tell me if it's clearer :))

PR #2380, opened 2020-04-10
Revamp MetalFence and implement Metal sync objects

bejado:
This implements `createSync` and `getSyncStatus` in Metal. It also fixes a race condition with fences in Metal.

### DIFF
--- a/filament/backend/src/metal/MetalDriver.mm
+++ b/filament/backend/src/metal/MetalDriver.mm
@@ -239,11 +239,13 @@ void MetalDriver::createRenderTargetR(Handle<HwRenderTarget> rth,
 }
 
 void MetalDriver::createFenceR(Handle<HwFence> fh, int dummy) {
-    construct_handle<MetalFence>(mHandleMap, fh, *mContext);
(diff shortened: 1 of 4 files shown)


filament/backend/src/metal/MetalDriver.mm
```diff
@@ -612,8 +619,14 @@
 }
 
 SyncStatus MetalDriver::getSyncStatus(Handle<HwSync> sh) {
-    // TODO: implement Sync objects
-    return SyncStatus::SIGNALED;
+    auto* fence = handle_cast<MetalFence>(mHandleMap, sh);
+    FenceStatus status = fence->wait(0);
```
pixelflinger, line 623: does the correct thing happen here if we get to this point before `encode()` has been called? (i.e. it should return `NOT_SIGNALED`)

bejado: Yes, `fence->wait(0)` will return `FenceStatus::TIMEOUT_EXPIRED` (because `state->status` has already been initialized to `TIMEOUT_EXPIRED`) and this function will then return `SyncStatus::NOT_SIGNALED`.

PR #2393, opened 2020-04-14
Implement texture swizzling

pixelflinger:


RELEASE_NOTES.md
```diff
@@ -26,6 +26,7 @@ A new header is inserted each time a *tag* is created.
 - Fix several issues related to multi-view support. (⚠️ **API breakage**) removed
   `View::setClearColor()`, a similar functionality is now handled by `Renderer::setClearOptions()`
   and `Skybox`, the later now can be set to a constant color.
+- Textures can now be swizzled.
```
romainguy, line 29: Mention API?

filament/backend/src/vulkan/VulkanDriver.cpp
```diff
@@ -691,6 +691,11 @@ void VulkanDriver::updateIndexBuffer(Handle<HwIndexBuffer> ibh, BufferDescriptor
     scheduleDestroy(std::move(p));
 }
 
+void VulkanDriver::setTextureSwizzle(Handle<HwTexture> th,
```
romainguy, line 694: Can you add a stub for the Metal driver as well?

pixelflinger: there is one already in this PR

PR #2399, opened 2020-04-15
fix spotlight documentation and clamping

pixelflinger:
filament spotlights use half-angles, which is the same as gltf.

### DIFF
--- a/filament/src/components/LightManager.cpp
+++ b/filament/src/components/LightManager.cpp
@@ -253,13 +253,11 @@ void FLightManager::setIntensity(Instance i, float intensity) noexcept {
                 break;
 
             case Type::FOCUSED_SPOT: {
-                // li = lp / (2*pi*(1-cos(outer/2)))
+                // li = lp / (2 * pi * (1 - cos(cone_outer / 2)))
                 SpotParams& spotParams = manager[i].spotParams;
-                float2 scaleOffset = spotParams.scaleOffset;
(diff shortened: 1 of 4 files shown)


filament/src/components/LightManager.cpp
```diff
@@ -253,13 +253,11 @@ void FLightManager::setIntensity(Instance i, float intensity) noexcept {
                 break;
 
             case Type::FOCUSED_SPOT: {
-                // li = lp / (2*pi*(1-cos(outer/2)))
+                // li = lp / (2 * pi * (1 - cos(outer / 2)))
```
romainguy, line 256: If we pass the half angle this comment should be updated, or at least augmented explaining that "outer" is the full angle of the spot

filament/src/components/LightManager.cpp
```diff
@@ -306,9 +304,9 @@ void FLightManager::setSpotLightCone(Instance i, float inner, float outer) noexc
         // we need to recompute the luminous intensity
         Type type = getLightType(i).type;
         if (type == Type::FOCUSED_SPOT) {
+            // li = lp / (2 * pi * (1 - cos(outer / 2)))
```
romainguy, line 307: Same here

filament/src/components/LightManager.cpp
pixelflinger: oh yeah I see. I took this comment as the "real formula"

PR #2404, opened 2020-04-16
texture swizzling is immutable

pixelflinger:
texture swizzling is now set at creation time to accommodate 
vulkan/metal texture immutability

### DIFF
--- a/filament/src/Texture.cpp
+++ b/filament/src/Texture.cpp
@@ -114,6 +114,17 @@ Texture* Texture::Builder::build(Engine& engine) {
             "Texture format %u not supported on this platform", mImpl->mFormat)) {
         return nullptr;
     }
+
+    const bool sampleable = bool(mImpl->mUsage & TextureUsage::SAMPLEABLE);
+    const bool swizzled = mImpl->mTextureIsSwizzled;
+    const bool imported = mImpl->mImportedId;
+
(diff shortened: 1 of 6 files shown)


filament/src/Texture.cpp
```diff
@@ -132,18 +143,19 @@ FTexture::FTexture(FEngine& engine, const Builder& builder) {
 
     FEngine::DriverApi& driver = engine.getDriverApi();
     if (UTILS_LIKELY(builder->mImportedId == 0)) {
-        mHandle = driver.createTexture(
-                mTarget, mLevelCount, mFormat, mSampleCount, mWidth, mHeight, mDepth, mUsage);
+        if (UTILS_LIKELY(builder->mTextureIsSwizzled)) {
+            mHandle = driver.createTexture(
```
bejado, line 147: Do we need two driver methods for swizzled textures? What if we added a `TextureSwizzle` argument to `createTexture`, with defaults set for no swizzling.

pixelflinger: I didn't want to do that because it's a very uncommon case, and there are other places like the framegraph that create textures directly by calling the driver, and I didn't want to have to track and change all these places -- also didn't want to make those GL calls that won't be needed 99.99% of the time.

PR #2406, opened 2020-04-16
Metal backend: get rid of non-const global variables

bejado:


filament/backend/src/metal/MetalBlitter.mm
```diff
@@ -31,7 +31,7 @@
 namespace backend {
 namespace metal {
 
-static std::string functionLibrary (R"(
+const static std::string functionLibrary (R"(
```
pixelflinger, line 34: could you just use a c-string for this?

PR #2408, opened 2020-04-17
Add new Android demo: sample-page-curl.

prideout:
Pure Java app that demonstrates custom vertex shader animation and two-sided texturing.  Generates vertex positions as described in "Deforming Pages of Electronic Books" by Hong et al.  Users can drag horizontally to turn the page.

sample_page_curl


prideout:
> LGTM but what's the license on the two .jpg files?

They are CC0, I'll add a txt file.

PR #2412, opened 2020-04-17
Vulkan: use more appropriate constness.

prideout:


filament/backend/src/vulkan/PlatformVkLinux.cpp
```diff
@@ -32,7 +32,7 @@ using namespace backend;
 // passing in a null pointer, and we highlight the argument by using the VKALLOC constant.
 constexpr VkAllocationCallbacks* VKALLOC = nullptr;
 
-static const char* LIBRARY_X11 = "libX11.so.6";
+static constexpr char* LIBRARY_X11 = "libX11.so.6";
```
pixelflinger, line 35: I know c++ is stupid,  but technically this should be `constexpr const`

filament/backend/src/vulkan/VulkanContext.cpp
```diff
@@ -150,7 +150,7 @@ void selectPhysicalDevice(VulkanContext& context) {
 
 void createVirtualDevice(VulkanContext& context) {
     VkDeviceQueueCreateInfo deviceQueueCreateInfo[1] = {};
-    static const float queuePriority[] = {1.0f};
+    static constexpr float queuePriority[] = {1.0f};
```
pixelflinger, line 153: I wouldn't make this static, I would just do `float queuePriority = 1.0;` and use `&queuePriority`.

filament/backend/src/vulkan/VulkanDriver.cpp
```diff
@@ -154,7 +154,7 @@ VulkanDriver::VulkanDriver(VulkanPlatform* platform,
 
 #if ENABLE_VALIDATION
     if (createDebugReportCallback) {
-        static VkDebugReportCallbackCreateInfoEXT cbinfo = {
+        static constexpr VkDebugReportCallbackCreateInfoEXT cbinfo = {
```
pixelflinger, line 157: I would just use `const blah = ` and not `static`. I have no idea what the compiler will do, but it could run extra code to initialize this on first run.

filament/backend/src/vulkan/PlatformVkLinux.cpp
```diff
@@ -48,7 +48,7 @@ struct X11Functions {
 
 Driver* PlatformVkLinux::createDriver(void* const sharedContext) noexcept {
     ASSERT_PRECONDITION(sharedContext == nullptr, "Vulkan does not support shared contexts.");
-    static const char* requestedExtensions[] = {
+    static constexpr char* requestedExtensions[] = {
```
pixelflinger, line 51: I wouldn't use `static` inside a function, just to not run the risk that the compiler generates extra code for initialization on first call.

filament/backend/src/vulkan/PlatformVkWindows.cpp
```diff
@@ -28,7 +28,7 @@ using namespace backend;
 
 Driver* PlatformVkWindows::createDriver(void* const sharedContext) noexcept {
     ASSERT_PRECONDITION(sharedContext == nullptr, "Vulkan does not support shared contexts.");
-    static const char* requestedExtensions[] = {
+    static constexpr char* requestedExtensions[] = {
```
pixelflinger, line 31: same

PR #2414, opened 2020-04-17
improve contact shadows support

pixelflinger:
- contact shadows are now supported for point and spot lights
- and are now independent from regular shadows, that is they can
  be enabled without enabling regular shadows

The only limitation currently is that the distance and step count for
ALL contact shadow are taken from the directional light options.

### DIFF
--- a/filament/include/filament/LightManager.h
+++ b/filament/include/filament/LightManager.h
@@ -21,6 +21,7 @@
 #include <filament/Color.h>
 
 #include <utils/compiler.h>
+#include <utils/Entity.h>
 #include <utils/EntityInstance.h>
 
 #include <math/mathfwd.h>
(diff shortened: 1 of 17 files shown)


filament/include/filament/LightManager.h
```diff
@@ -236,22 +252,28 @@ class UTILS_PUBLIC LightManager : public FilamentAPI {
...
          * (off by default)
          */
         bool screenSpaceContactShadows = false;
 
         /**
-         * Number of ray-marching steps for screen-space contact shadows.
-         * (8 by default)
+         * Number of ray-marching steps for screen-space contact shadows (8 by default).
+         *
+         * CAUTION: this parameter is ignored for all lights except the directional/sun light,
```
romainguy, line 264: Do you think we should change that eventually?

pixelflinger: Yes I think so. That's why I didn't change the API. I mean if we had an editor, I could imagine the artist tweaking this per light ?

romainguy: Yeah I think we definitely want to tweak this per light.

PR #2428, opened 2020-04-20
Remove monkey mesh dependency from samples.

prideout:
This change allows us to modify + test Filament without re-invoking resgen.

The only apps that need the monkey filamesh are `suzanne` and `hellopbr`, so they are now the only apps that have it as a dependency.

### DIFF
--- a/samples/CMakeLists.txt
+++ b/samples/CMakeLists.txt
@@ -13,23 +13,6 @@ set(TEXTURE_DIR "${GENERATION_ROOT}/generated/texture")
 set(RESOURCE_BINS)
 
 # ==================================================================================================
-# Convert OBJ files into filamesh files.
-# ==================================================================================================
-
-function(add_mesh SOURCE TARGET)
-    set(source_mesh "${ROOT_DIR}/${SOURCE}")
-    set(target_mesh "${RESOURCE_DIR}/${TARGET}")
-    set(RESOURCE_BINS ${RESOURCE_BINS} ${target_mesh} PARENT_SCOPE)
-    add_custom_command(
-        OUTPUT ${target_mesh}
-        COMMAND filamesh --compress ${source_mesh} ${target_mesh}
-        MAIN_DEPENDENCY ${source_mesh}
-        DEPENDS filamesh)
-endfunction()
-
-add_mesh("assets/models/monkey/monkey.obj" "suzanne.filamesh")
-
-# ==================================================================================================
 # Build materials
 # ==================================================================================================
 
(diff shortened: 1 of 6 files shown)


samples/CMakeLists.txt
```diff
@@ -129,7 +112,7 @@ add_envmap("third_party/environments/venetian_crossroads_2k.hdr" "venetian_cross
 function(add_ktxfiles SOURCE TARGET EXTRA_ARGS)
     set(source_path "${ROOT_DIR}/${SOURCE}")
     set(target_path "${TEXTURE_DIR}/${TARGET}")
-    set(TEXTURE_FILES ${TEXTURE_FILES} ${target_path} PARENT_SCOPE)
+    set(MONKEY_FILES ${MONKEY_FILES} ${target_path} PARENT_SCOPE)
```
romainguy, line 115: That doesnt seem to be the rename you wanted to do

samples/CMakeLists.txt
```diff
@@ -140,27 +123,40 @@ endfunction()
 function(add_pngfile SOURCE TARGET)
     set(source_path "${ROOT_DIR}/${SOURCE}")
     set(target_path "${TEXTURE_DIR}/${TARGET}")
-    set(TEXTURE_FILES ${TEXTURE_FILES} ${target_path} PARENT_SCOPE)
+    set(MONKEY_FILES ${MONKEY_FILES} ${target_path} PARENT_SCOPE)
```
romainguy, line 126: That doesn't seem to be the rename you wanted to do

samples/CMakeLists.txt
```diff
@@ -140,27 +123,40 @@ endfunction()
...
+
 add_ktxfiles("assets/models/monkey/albedo.png" "albedo_s3tc.ktx" "--compression=s3tc_rgb_dxt1")
 add_ktxfiles("assets/models/monkey/roughness.png" "roughness.ktx" "--grayscale")
 add_ktxfiles("assets/models/monkey/metallic.png" "metallic.ktx" "--grayscale")
 add_ktxfiles("assets/models/monkey/ao.png" "ao.ktx" "--grayscale")
 
 add_pngfile("assets/models/monkey/normal.png" "normal.png")
 
-get_resgen_vars(${RESOURCE_DIR} textures)
+get_resgen_vars(${RESOURCE_DIR} monkey)
```
romainguy, line 153: Don't you need the textures still?

prideout: The mesh file has been moved from the common resource archive into the suzanne-specific archive.  I renamed the suzanne-specific archive because it now includes textures AND a mesh.

PR #2436, opened 2020-04-22
Add a new free flight camera manipulator

bejado:
This adds a new camera manipulator that allows WASD movement with nodal camera panning. This PR only adds the new manipulator type, it doesn't enable it yet inside of `FilamentApp` (we'll need to decide how we want to expose it, possibly through a command-line flag).

free_flight

### DIFF
--- a/libs/camutils/include/camutils/Manipulator.h
+++ b/libs/camutils/include/camutils/Manipulator.h
@@ -34,8 +34,9 @@ enum class Fov { VERTICAL, HORIZONTAL };
  * Helper that enables camera interaction similar to sketchfab or Google Maps.
  *
  * Clients notify the camera manipulator of various mouse or touch events, then periodically call
- * its getLookAt() method so that they can adjust their camera(s). Two modes are supported: ORBIT
- * and MAP. To construct a manipulator instance, the desired mode is passed into the create method.
+ * its getLookAt() method so that they can adjust their camera(s). Three modes are supported: ORBIT,
+ * MAP, and FREE_FLIGHT. To construct a manipulator instance, the desired mode is passed into the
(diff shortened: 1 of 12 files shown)


prideout:
Nice!  Reviewing this now.

libs/camutils/include/camutils/Manipulator.h
```diff
@@ -186,14 +206,53 @@ class Manipulator {
...
+
+    /**
+     * In MAP and ORBIT modes, dollys the camera along the viewing direction.
+     * In FREE_FLIGHT mode, adjusts the move speed of the camera.
      *
      * @param x X-coordinate for point of interest in viewport space
      * @param y Y-coordinate for point of interest in viewport space
      * @param scrolldelta Negative means "zoom in", positive means "zoom out"
      */
     virtual void zoom(int x, int y, FLOAT scrolldelta) = 0;
```
prideout, line 245: Nice, this feature surprised me but then I saw that sketchfab has it too!  Maybe we should rename this method to "scroll" everywhere, since it's not really zooming in free flight mode.  Also add a note that `x` and `y` are ignored in free flight mode.

bejado: Yup, it seems to be a common pattern. I got the idea from Unreal.

bejado:
@prideout I just noticed we have Android JNI bindings for Manipulator as well.. any concerns about me propagating these changes on the Java side?

prideout:
Yep, we use the Java bindings in the Android glTF sample.  For completeness, I do think we should go ahead and add this feature to the Java bindings.

pixelflinger:
I can't wait for this!

PR #2440, opened 2020-04-22
material_sandbox: set a clear color when no IBL is present

bejado:


bejado:
For some reason, didn't merge this before..

PR #2445, opened 2020-04-23
Vulkan: add support for 2D texture arrays.

prideout:
In Vulkan, the arrayness of a texture is an aspect of the "image view", not the image itself.

In my earlier attempt I used `VK_IMAGE_CREATE_2D_ARRAY_COMPATIBLE_BIT` on the image, which MoltenVK does not support. Confusingly, that flag is actually meant for rendering into a slice of a 3D texture, which we are not doing.

Vulkan is hard.

### DIFF
--- a/filament/backend/src/vulkan/VulkanHandles.cpp
+++ b/filament/backend/src/vulkan/VulkanHandles.cpp
@@ -95,9 +95,9 @@ static VulkanAttachment createOffscreenAttachment(VulkanTexture* tex) {
 }
 
 VulkanRenderTarget::VulkanRenderTarget(VulkanContext& context, uint32_t width, uint32_t height,
-        uint32_t colorLevel, VulkanTexture* color, uint32_t depthLevel, VulkanTexture* depth) :
-        HwRenderTarget(width, height), mContext(context), mOffscreen(true), mColorLevel(colorLevel),
-        mDepthLevel(depthLevel) {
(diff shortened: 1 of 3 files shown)


filament/backend/src/vulkan/VulkanHandles.cpp
```diff
@@ -116,6 +116,9 @@ VulkanRenderTarget::VulkanRenderTarget(VulkanContext& context, uint32_t width, u
         if (color->target == SamplerType::SAMPLER_CUBEMAP) {
             viewInfo.viewType = VK_IMAGE_VIEW_TYPE_CUBE;
             viewInfo.subresourceRange.layerCount = 6;
+        } else if (color->target == SamplerType::SAMPLER_2D_ARRAY) {
+            viewInfo.viewType = VK_IMAGE_VIEW_TYPE_2D_ARRAY;
+            viewInfo.subresourceRange.layerCount = color->depth;
```
bejado, line 121: I think we'll also need to set `baseArrayLayer` here, analogous to how we use `colorLevel` to set `baseMipLevel` above.

prideout: Wow excellent catch!

prideout: Fixed.

PR #2450, opened 2020-04-24
A new depth of field (DoF) post-process effect

pixelflinger:
dof

### DIFF
--- a/android/filament-android/src/main/java/com/google/android/filament/View.java
+++ b/android/filament-android/src/main/java/com/google/android/filament/View.java
@@ -69,6 +69,7 @@ public class View {
     private FogOptions mFogOptions;
     private RenderTarget mRenderTarget;
     private BlendMode mBlendMode;
+    private DepthOfFieldOptions mDepthOfFieldOptions;
 
     /**
      * Generic Quality Level
@@ -309,6 +310,23 @@ public class View {
          */
         public boolean enabled = false;
     }
+
+    /**
+     * Options to control Depth of Field (DoF) effect in the scene
+     *
+     * @see View#setDepthOfFieldOptions
+     */
+    public static class DepthOfFieldOptions {
+
+        /** focus distance in world units */
+        public float focusDistance = 10.0f;
+
+        /** scale factor controlling the amount of blur (values other than 1.0 are not physically correct)*/
+        public float blurScale = 1.0f;
+
+        /** enable or disable Depth of field effect */
+        public boolean enabled = false;
+    };
 
     /**
      * Structure used to set the color precision for the rendering of a <code>View</code>.
@@ -977,6 +995,32 @@ public class View {
     }
 
 
+    /**
+     * Sets Depth of Field options.
+     *
+     * @param options Options for depth of field effect.
+     * @see #getDepthOfFieldOptions
+     */
+    public void setDepthOfFieldOptions(@NonNull DepthOfFieldOptions options) {
+        mDepthOfFieldOptions = options;
+        nSetDepthOfFieldOptions(getNativeObject(), options.focusDistance, options.blurScale, options.enabled);
+    }
+
+    /**
+     * Gets the Depth of Field options
+     *
+     * @return Depth of Field options currently set.
+     * @see #setDepthOfFieldOptions
+     */
+    @NonNull
+    public DepthOfFieldOptions getDepthOfFieldOptions() {
+        if (mDepthOfFieldOptions == null) {
+            mDepthOfFieldOptions = new DepthOfFieldOptions();
+        }
+        return mDepthOfFieldOptions;
+    }
+
+
     public long getNativeObject() {
         if (mNativeObject == 0) {
             throw new IllegalStateException("Calling method on destroyed View");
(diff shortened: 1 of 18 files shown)


android/filament-android/src/main/java/com/google/android/filament/View.java
```diff
@@ -310,6 +311,20 @@
...
+    /**
+     * Options to control Depth of Field (DoF) effect in the scene
+     *
+     * @see View#setDepthOfFieldOptions
+     */
+    public static class DepthOfFieldOptions {
+
+        /** focus distance in world units */
+        public float focusDistance = 10.0f;
+
```
romainguy, line 323: Should we add a blurScale or for later?

filament/src/Camera.cpp
```diff
@@ -234,6 +235,34 @@ Frustum FCamera::getFrustum(mat4 const& projection, mat4f const& viewMatrix) noe
...
+CameraInfo::CameraInfo(FCamera const& camera, const math::mat4f& worldOriginCamera) noexcept {
+    const mat4f modelMatrix{ worldOriginCamera * camera.getModelMatrix() };
+    projection         = mat4f{ camera.getProjectionMatrix() };
+    cullingProjection  = mat4f{ camera.getCullingProjectionMatrix() };
+    model              = modelMatrix;
+    view               = FCamera::getViewMatrix(model);
+    zn                 = camera.getNear();
+    zf                 = camera.getCullingFar();
+    ev100              = Exposure::ev100(camera);
+    f                  = (FCamera::SENSOR_SIZE * (float)projection[1][1]) * 0.5f;
```
romainguy, line 261: Should we add a `getFocalLength()` to the public `Camera` API?

filament/src/PostProcessManager.cpp
```diff
@@ -316,6 +320,148 @@ FrameGraphId<FrameGraphTexture> PostProcessManager::fxaa(FrameGraph& fg,
...
+        const View::DepthOfFieldOptions& dofOptions,
+        const CameraInfo& cameraInfo) noexcept {
+
+    FEngine& engine = mEngine;
+    Handle<HwRenderPrimitive> const& fullScreenRenderPrimitive = engine.getFullScreenRenderPrimitive();
+
+    FrameGraphId<FrameGraphTexture> depth = fg.getBlackboard().get<FrameGraphTexture>("structure");
+    assert(depth.isValid());
+
+    const size_t sampleCount = 25;  // (keep in sync with dofUtils.fs)
```
romainguy, line 334: Should we control this with a quality setting?

filament/src/materials/dof.mat
```diff
@@ -0,0 +1,65 @@
...
+
+vertex {
+    void postProcessVertex(inout PostProcessVertexInputs postProcess) {
+        postProcess.vertex.xy = postProcess.normalizedUV;
+    }
+}
+
+fragment {
+
+void foo() {}  // twart the IDE
```
romainguy, line 45: twart -> thwart

Is this required?

filament/src/materials/dofBlur.mat
```diff
@@ -0,0 +1,77 @@
...
+    ],
+    domain : postprocess,
+    depthWrite : false,
+    depthCulling : false
+}
+
+vertex {
+    void postProcessVertex(inout PostProcessVertexInputs postProcess) {
+    postProcess.vertex.xy = postProcess.normalizedUV;
+}
```
romainguy, line 35: Incorrect indentation on } and the previous line

filament/src/materials/dofBlur.mat
```diff
@@ -0,0 +1,77 @@
...
+ *
+ * This uses a lot of ideas from
+ *   "Hexagonal Bokeh Blur Revisited" by Colin Barré-Brisebois
+ *   (https://colinbarrebrisebois.com/2017/04/18/hexagonal-bokeh-blur-revisited-part-1-basic-3-pass-version/)
+ *
+ *   "Efficiently Simulating the Bokeh of Polygonal Apertures in a Post-Process Depth of Field Shader"
+ *   by L. McIntosh, B. E. Riecke and S. DiPaola
+ */
+
+void foo() {}  // twart the IDE
```
romainguy, line 51: thwart

filament/src/materials/dofUtils.fs
```diff
@@ -0,0 +1,71 @@
...
+// Filter sample count, prefer odd values
+// (keep in sync with PostProcessManager.cpp:dof())
+const float SAMPLE_COUNT = 25.0;
+
+// A value > 1.0 allows a larger blur w/ dithering
+const float BLUR_SCALE = 1.0;
+
+#define unitvec(angle) vec2(cos(angle), sin(angle))
+
+float random(vec2 w) {
```
romainguy, line 24: Add a comment explaining this is interleaved gradient noise
Even better, use our function `interleavedGradientNoise()` from `dither.fs`.

Also `vec2 w` needs to be `highp`.

android/filament-android/src/main/java/com/google/android/filament/View.java
pixelflinger: I'm not sure where the blur scale would be applied, if we just apply it to the CoC scale and bias, would it be useful? for e.g.: I mean that a scale of 0.5 wouldn't always give the same amount of blur, it would still depend on the camera parameters.  It would be useful for the relative mode though.

filament/src/Camera.cpp
pixelflinger: Yes. It would make sense. I'll do that in a separate PR.

filament/src/PostProcessManager.cpp
pixelflinger: Maybe we should just make it a value the user can set, after all.

filament/src/materials/dof.mat
pixelflinger: ah no I forgot to get rid of it.

android/filament-android/src/main/java/com/google/android/filament/View.java
romainguy: I was thinking it would be relative to the camera settings. At least you would be able to set your exposure and if you don't like the DoF, the blur scale would let you tweak just the DoF without changing anything else. It's the "break the physically based camera" slider basically.

pixelflinger: Sounds good.

PR #2465, opened 2020-04-29
Add intensityCandela API to LightManager

bejado:
This adds a new `intensityCandela` builder method and `setIntensityCandela` API method for updating a punctual light's intensity in candela, as opposed to lumens or watts.


romainguy:
Well that's not helpful here. Our existing APIs take lumens but the shaders work in candelas. For glTF we want a way to directly set the candelas intensity. That or a `lumensFromCandela()` function but it does require passing the type of light. I still don't understand why glTF went with candelas…

bejado:
A `lumensFromCandela` function would require passing the type of light, and in the case of spot lights, the cone angles. But then we're back to the case where we're doing two conversions just to cancel each other out.

PR #2467, opened 2020-04-30
Vulkan: add support for timer queries.

prideout:


filament/backend/src/vulkan/VulkanDriver.cpp
```diff
@@ -726,7 +734,30 @@ void VulkanDriver::cancelExternalImage(void* image) {
 }
 
 bool VulkanDriver::getTimerQueryValue(Handle<HwTimerQuery> tqh, uint64_t* elapsedTime) {
-    return false;
+    VulkanTimerQuery* vtq = handle_cast<VulkanTimerQuery>(mHandleMap, tqh);
```
bejado, line 737: I think there might be a race condition here. Because `getTimerQueryValue` is synchronous, it could be called on the main thread before `createTimerQueryR` has had a chance to construct the `VulkanTimerQuery` object. The Metal driver had to take scenarios like this into account.

At the very least I don't think we can trust that `vtq->startingQueryIndex` is necessarily valid here.

prideout: Oh right, let me fix this up.

prideout: I've added a commit, I took an approach similar to what you did.

filament/backend/src/vulkan/VulkanDriver.cpp
```diff
@@ -1237,11 +1272,26 @@ void VulkanDriver::draw(PipelineState pipelineState, Handle<HwRenderPrimitive> r
...
 }
 
-
 void VulkanDriver::beginTimerQuery(Handle<HwTimerQuery> tqh) {
+    VulkanCommandBuffer* commands = mContext.currentCommands;
+    ASSERT_POSTCONDITION(commands, "Timer queries can occur only within a beginFrame / endFrame.");
+
+    VulkanTimerQuery* vtq = handle_cast<VulkanTimerQuery>(mHandleMap, tqh);
+    const uint32_t index = vtq->startingQueryIndex;
+    const VkPipelineStageFlagBits stage = VK_PIPELINE_STAGE_TOP_OF_PIPE_BIT;
```
pixelflinger, line 1281: I think you might need `VK_PIPELINE_STAGE_BOTTOM_OF_PIPE_BIT` here as well. 
i.e. you want to take a timestamp after everything before this call has been executed on the GPU.

It seems to me that using `TOP` here would means that you measure from when the CPU gets to this command.

prideout: You're right, both should be BOTTOM.  Fixed.  (Btw I think `TOP` has nothing to do with CPU, it's more about the when the command is initially consumed by the GPU.)

PR #2468, opened 2020-04-30
Fix incorrect lightspace 2D bounds calculation for stable shadows

bejado:
This fixes an issue when using stable directional shadows. See how the shadows disappear when the light's direction becomes perfectly orthogonal to the camera:

https://drive.google.com/a/google.com/file/d/1qWTavtkbL2PO-oEVaTZ9bJwZGbMw7ZMt/view?usp=sharing

### DIFF
--- a/filament/src/ShadowMap.cpp
+++ b/filament/src/ShadowMap.cpp
@@ -670,7 +670,7 @@ Aabb ShadowMap::compute2DBounds(const mat4f& lightView,
 }
 
 Aabb ShadowMap::compute2DBounds(const mat4f& lightView, float4 const& sphere) noexcept {
-    // this assumes a rotation + translate only
+    // this assumes a rigid body transform
     float4 s;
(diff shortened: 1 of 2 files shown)


bejado:
This fixes the bug, but breaks the stability of shadows.. will re-investigate.

filament/src/ShadowMap.cpp
```diff
@@ -670,11 +670,10 @@ Aabb ShadowMap::compute2DBounds(const mat4f& lightView,
 }
 
 Aabb ShadowMap::compute2DBounds(const mat4f& lightView, float4 const& sphere) noexcept {
-    // this assumes a uniform scale + rotation + translate only
-    // TODO: we need to handle perspective projections
+    // this assumes a rotation + translate only
```
pixelflinger, line 673: You could rephrase as "rigid body transform".

PR #2471, opened 2020-05-01
compute bitangeant in fragment shader

### LINKED ISSUE #2092: Consider moving bitangent construction to fragment shader.
We currently reconstruct the bitangent in `main.vs` but popular normal map baking tools expect this to occur in the fragment shader.  More info in [this](https://medium.com/@bgolus/generating-perfect-normal-maps-for-unity-f929e673fc57) article. See especially "comments from the original mikktspace.h".

To verify the correctness of normal mapping, it might be worth crafting a test model.

On a related note, we should consider adding a mikktspace option to `SurfaceOrientation`, since this is the industry standard; however this would be slow since it produces an unindexed mesh.

pixelflinger:
This saves 4 varying components.
Fixes #2092

### DIFF
--- a/shaders/src/main.vs
+++ b/shaders/src/main.vs
@@ -9,7 +9,7 @@ void main() {
     #if defined(MATERIAL_NEEDS_TBN)
         // Extract the normal and tangent in world space from the input quaternion
         // We encode the orthonormal basis as a quaternion to save space in the attributes
-        toTangentFrame(mesh_tangents, material.worldNormal, vertex_worldTangent);
+        toTangentFrame(mesh_tangents, material.worldNormal, vertex_worldTangent.xyz);
 
         #if defined(HAS_SKINNING_OR_MORPHING)
         if (objectUniforms.morphingEnabled == 1) {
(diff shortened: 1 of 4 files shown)


shaders/src/main.vs
```diff
@@ -27,21 +27,17 @@ void main() {
...
         }
         #endif
 
         // We don't need to normalize here, even if there's a scale in the matrix
         // because we ensure the worldFromModelNormalMatrix pre-scales the normal such that
         // all its components are < 1.0. This precents the bitangent to exceed the range of fp16
         // in the fragment shader, where we renormalize after interpolation
-        vertex_worldTangent = objectUniforms.worldFromModelNormalMatrix * vertex_worldTangent;
+        vertex_worldTangent.xyz = objectUniforms.worldFromModelNormalMatrix * vertex_worldTangent.xyz;
+        vertex_worldTangent.w = mesh_tangents.w;
```
romainguy, line 39: Can you just store sign(w) here so we don't need sign() in the fragment shader?

shaders/src/shading_parameters.fs
```diff
@@ -10,8 +10,8 @@ void computeShadingParams() {
 #if defined(HAS_ATTRIBUTE_TANGENTS)
     highp vec3 n = vertex_worldNormal;
 #if defined(MATERIAL_NEEDS_TBN)
-    highp vec3 t = vertex_worldTangent;
-    highp vec3 b = vertex_worldBitangent;
+    highp vec3 t = vertex_worldTangent.xyz;
+    highp vec3 b = cross(n, t) * sign(vertex_worldTangent.w);
```
romainguy, line 14: See previous comment

PR #2472, opened 2020-05-01
Don't use gl_FragCoord anywhere as it breaks with custom viewports

pixelflinger:
This fixes dynamic lighting and SSAO when a viewport is not in 0,0.
In practice this currently happens only when all post-processing is
disabled.

Instead of using gl_FragCoord we introduce a new API, 
getNormalizedViewportCoord(), which as the name implies returns
normalized [0, 1] viewport coordinates with origin at the bottom-left,
on all platforms.

This is implemented in this PR by interpolating gl_Position.

### DIFF
--- a/shaders/src/inputs.fs
+++ b/shaders/src/inputs.fs
@@ -13,6 +13,8 @@ LAYOUT_LOCATION(6) SHADING_INTERPOLATION in mediump vec4 vertex_worldTangent;
 #endif
 #endif
 
+LAYOUT_LOCATION(7) in highp vec4 vertex_position;
+
 #if defined(HAS_ATTRIBUTE_COLOR)
 LAYOUT_LOCATION(9) in mediump vec4 vertex_color;
 #endif
--- a/shaders/src/inputs.vs
+++ b/shaders/src/inputs.vs
@@ -64,6 +64,8 @@ LAYOUT_LOCATION(6) SHADING_INTERPOLATION out mediump vec4 vertex_worldTangent;
 #endif
 #endif
 
+LAYOUT_LOCATION(7) out highp vec4 vertex_position;
+
 #if defined(HAS_ATTRIBUTE_COLOR)
 LAYOUT_LOCATION(9) out mediump vec4 vertex_color;
 #endif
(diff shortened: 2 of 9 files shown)


shaders/src/inputs.fs
```diff
@@ -13,6 +13,8 @@ LAYOUT_LOCATION(6) SHADING_INTERPOLATION in mediump vec4 vertex_worldTangent;
 #endif
 #endif
 
+LAYOUT_LOCATION(7) in highp vec4 vertex_Position;
```
romainguy, line 16: `vertex_position` (no upper case) to match all the other variables

shaders/src/inputs.vs
```diff
@@ -64,6 +64,8 @@ LAYOUT_LOCATION(6) SHADING_INTERPOLATION out mediump vec4 vertex_worldTangent;
 #endif
 #endif
 
+LAYOUT_LOCATION(7) out highp vec4 vertex_Position;
```
romainguy, line 67: `vertex_position` (no upper case) to match all the other variables

shaders/src/shading_parameters.fs
```diff
@@ -34,6 +34,11 @@ void computeShadingParams() {
 
     shading_position = vertex_worldPosition;
     shading_view = normalize(frameUniforms.cameraPosition - shading_position);
+
+    // we do this so we avoid doing (matrix multiply), but we burn 4 varyings:
+    //    p = clipFromWorldMatrix * shading_position;
+    //    shading_normalizedViewportCoord = p.xy * 0.5 / p.w + 0.5
+    shading_normalizedViewportCoord = vertex_Position.xy * (0.5 / vertex_Position.w) + 0.5;
```
romainguy, line 41: Could we store 0.5/w in the varying instead of doing the division here? Or at least 1/w.

pixelflinger: I don't think you can interpolate 1/w -- I tried to interpolate `.xy/w` (which would be the same) initially and that didn't work.

romainguy: 💩

Yaski:
typo in documentation - getNormalizedViewportCoord doesn't have float2 input

PR #2475, opened 2020-05-02
Reduce DoF sample count from 25 to 11

pixelflinger:
We keep the same (or very close) quality by exploiting h/w
filtering. On the flip side, this is not compatible with dithering,
which is removed here.

### DIFF
--- a/filament/src/materials/dofUtils.fs
+++ b/filament/src/materials/dofUtils.fs
@@ -14,21 +14,12 @@
 
 // Filter sample count, prefer odd values
 // (keep in sync with PostProcessManager.cpp:dof())
-const float SAMPLE_COUNT = 25.0;
-
-// A value > 1.0 allows a larger blur w/ dithering
-const float BLUR_SCALE = 1.0;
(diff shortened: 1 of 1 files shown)


filament/src/materials/dofUtils.fs
```diff
@@ -56,20 +53,14 @@ vec4 blurTexture(sampler2D colorBuffer, highp vec2 uv, highp vec2 direction, flo
     vec4 finalColor = textureLod(colorBuffer, uv, 0.0);
 
     highp vec2 unit = materialParams.resolution.zw;
-    direction *= unit * BLUR_SCALE;
-
-    float noise = 0.0;
-    if (BLUR_SCALE != 1.0) {
-        // we span 2 samples because the first sample is always fixed (no noise added)
-        noise = (random(gl_FragCoord.xy) - 0.5) * 2.0;
```
romainguy, line 64: Did you remove the random() function? Since you don't use it anymore

pixelflinger: good catch, thanks!

PR #2477, opened 2020-05-03
Adding support for openjdk 13

nakshathru:
Updated gradle version to 6.3 for openjdk 13 support
`./build.sh -p android release` was getting error `FilamentPlugin.groovy' Unsupported class file major version 57` when java_home is set to openjdk 13


nakshathru:
@googlebot I signed it!

romainguy:
I may have to revert this if it causes incompatibilities.

PR #2482, opened 2020-05-04
Add rudimentary cascading shadow map support

bejado:
CSM

This adds basic CSM support with a uniform split scheme, and no blending between cascades.

### DIFF
--- a/filament/include/filament/LightManager.h
+++ b/filament/include/filament/LightManager.h
@@ -196,6 +196,22 @@ public:
         /** Size of the shadow map in texels. Must be a power-of-two. */
         uint32_t mapSize = 1024;
 
+        /**
+         * Number of shadow cascades to use for this light. Must be between 1 and 4 (inclusive).
+         * A value greater than 1 turns on cascaded shadow mapping (CSM).
+         * Only applicable to Type.SUN or Type.DIRECTIONAL lights.
+         *
+         * @warning This API is still experimental and subject to change.
+         */
+        uint8_t shadowCascades = 1;
+
+        /**
+         * Visualize the shadow cascades through different colors.
+         *
+         * @warning This API is still experimental and subject to change.
+         */
+        bool debugVisualizeCascades = false;
+
         /** Constant bias in world units (e.g. meters) by which shadows are moved away from the
          * light. 1mm by default.
          */
--- a/filament/include/filament/LightManager.h
+++ b/filament/include/filament/LightManager.h
@@ -205,13 +205,6 @@ public:
          */
         uint8_t shadowCascades = 1;
 
-        /**
-         * Visualize the shadow cascades through different colors.
-         *
-         * @warning This API is still experimental and subject to change.
-         */
-        bool debugVisualizeCascades = false;
-
         /** Constant bias in world units (e.g. meters) by which shadows are moved away from the
          * light. 1mm by default.
          */
--- a/filament/src/ShadowMap.cpp
+++ b/filament/src/ShadowMap.cpp
@@ -98,7 +98,8 @@ void ShadowMap::render(DriverApi& driver, Handle<HwRenderTarget> rt,
 }
 
 void ShadowMap::update(const FScene::LightSoa& lightData, size_t index, FScene const* scene,
-        details::CameraInfo const& camera, uint8_t visibleLayers, ShadowMapLayout layout) noexcept {
+        details::CameraInfo const& camera, uint8_t visibleLayers, ShadowMapLayout layout,
+        CascadeParameters cascadeParams) noexcept {
     // this is the hard part here, find a good frustum for our camera
 
     auto& lcm = mEngine.getLightManager();
@@ -144,15 +145,13 @@ void ShadowMap::update(const FScene::LightSoa& lightData, size_t index, FScene c
     else            params.options.shadowNearHint = dzn * dz - camera.zn;
     if (dzf > 0)    dzf =-std::max(0.0f, camera.zf - params.options.shadowFarHint) / dz;
     else            params.options.shadowFarHint = dzf * dz + camera.zf;
-
-
     using Type = FLightManager::Type;
     switch (lcm.getType(li)) {
         case Type::SUN:
         case Type::DIRECTIONAL:
             computeShadowCameraDirectional(
                     lightData.elementAt<FScene::DIRECTION>(index), scene, cameraInfo, params,
-                    visibleLayers);
+                    visibleLayers, cascadeParams);
             break;
         case Type::FOCUSED_SPOT:
         case Type::SPOT:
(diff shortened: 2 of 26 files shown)


libs/filabridge/src/UibGenerator.cpp
```diff
@@ -43,7 +43,8 @@ UniformInterfaceBlock const& UibGenerator::getPerViewUib() noexcept  {
             .add("viewFromClipMatrix",      1, UniformInterfaceBlock::Type::MAT4, Precision::HIGH)
             .add("clipFromWorldMatrix",     1, UniformInterfaceBlock::Type::MAT4, Precision::HIGH)
             .add("worldFromClipMatrix",     1, UniformInterfaceBlock::Type::MAT4, Precision::HIGH)
-            .add("lightFromWorldMatrix",    1, UniformInterfaceBlock::Type::MAT4, Precision::HIGH)
+            .add("lightFromWorldMatrix",    CONFIG_MAX_SHADOW_CASCADES, UniformInterfaceBlock::Type::MAT4, Precision::HIGH)
```
pixelflinger, line 46: I would hardcode 4 here, and assert that `CONFIG_MAX_SHADOW_CASCADES` is 4 because changing it would break materials.

shaders/src/common_graphics.fs
```diff
@@ -97,3 +97,19 @@ vec3 heatmap(float v) {
     vec3 r = v * 2.1 - vec3(1.8, 1.14, 0.3);
     return 1.0 - r * r;
 }
+
+vec3 uintToColorDebug(uint v) {
```
pixelflinger, line 101: there is already some code that does that in `tone_mapping.fs` maybe this could be moved here.

shaders/src/getters.fs
```diff
@@ -108,3 +108,31 @@ bool isDoubleSided() {
     return materialParams._doubleSided;
 }
 #endif
+
+/**
+ * Returns the cascade index for this fragment (between 0 and CONFIG_MAX_SHADOW_CASCADES - 1).
+ */
+uint getShadowCascade() {
```
pixelflinger, line 115: shouldn't this be inside the `defined(HAS_SHADOWING)`?

libs/filabridge/include/private/filament/UibGenerator.h
```diff
@@ -51,7 +51,9 @@ struct PerViewUib { // NOLINT(cppcoreguidelines-pro-type-member-init)
     filament::math::mat4f viewFromClipMatrix;
     filament::math::mat4f clipFromWorldMatrix;
     filament::math::mat4f worldFromClipMatrix;
-    filament::math::mat4f lightFromWorldMatrix;
+    filament::math::mat4f lightFromWorldMatrix[CONFIG_MAX_SHADOW_CASCADES];
```
pixelflinger, line 54: this is definitely a material breakage, so need to update the material version

filament/src/ShadowMap.cpp
```diff
@@ -744,7 +744,7 @@ void ShadowMap::snapLightFrustum(float2& s, float2& o,
 
 size_t ShadowMap::intersectFrustumWithBox(
         FrustumBoxIntersection& UTILS_RESTRICT outVertices,
-        const Frustum& UTILS_RESTRICT frustum, const float3* UTILS_RESTRICT wsFrustumCorners,
```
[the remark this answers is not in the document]
bejado: Looks like it wasn't :)

bejado:
@pixelflinger I was going to wait to update the release notes until we have a more stable API, but maybe it's still worth mentioning it now.. what do you think?

filament/include/filament/LightManager.h
```diff
@@ -196,6 +196,22 @@ class UTILS_PUBLIC LightManager : public FilamentAPI {
...
+         * @warning This API is still experimental and subject to change.
+         */
+        uint8_t shadowCascades = 1;
+
+        /**
+         * Visualize the shadow cascades through different colors.
+         *
+         * @warning This API is still experimental and subject to change.
+         */
+        bool debugVisualizeCascades = false;
```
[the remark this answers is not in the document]
bejado: Good idea

shaders/src/getters.fs
bejado: `getShadowCascade` is also used for visualizing the different cascades. For this to work correctly, all fragments (even those not affected by shadowing) should be colored with their respective cascade.

pixelflinger:
> @pixelflinger I was going to wait to update the release notes until we have a more stable API, but maybe it's still worth mentioning it now.. what do you think?

Sure.

libs/filabridge/include/private/filament/UibGenerator.h
bejado: Whoops, meant to update it!

shaders/src/common_graphics.fs
bejado: `Tonemap_DisplayRange`? Hmm, it seems like it's written specifically for tone mapping, so I'm not sure it makes sense to move it into common_graphics.

romainguy: It's `log2` based but the only line that's display range specific is: `float v = log2(luminance(x) / 0.18);`. We could make a more generic variant. I used to use it for other things as well. Now I often just use `heatmap()`.

filament/src/ShadowMap.cpp
```diff
@@ -780,6 +780,7 @@ size_t ShadowMap::intersectFrustumWithBox(
 
     // at this point if we have 8 vertices, we can skip the rest
     if (vertexCount < 8) {
+        Frustum frustum(wsFrustumCorners);
```
pixelflinger, line 783: Just a comment here (don't change it for now). I think it's a little strange to have to recompute the frustum from the corners, because we started from coordinates in clip-space. And it should be trivial to go from that to a Frustum in light space.
 The frustum planes will be:
left {-1,0,0,-1}
right {1,0,0,1}
bottom {0,-1,0,-1}
top {0,1,0,1}
near {0,0,1,near}
far {0,0,-1,far}

then you can just transform them by the transpose(projection()) and that will give you the planes in light space. And as you can see, there are so many zeros that the transform is actually trivial:
```
            float4 p = transpose(projection} * float4{ -1, 0, 0, x };
            plane = float4{ normalize(p.xyz), 0 };
```

which will be more efficient and accurate than doing all the `cross()`

bejado: Why does the plane have a 0 for the w component? Wouldn't that affect the "position" of the plane?

bejado: I guess to rephrase this, wouldn't the near and far planes need special treatment? (because their w component cannot be 0, as they don't cross the origin)

shaders/src/common_graphics.fs
bejado: I'm all for refactoring it into a generic function, but I don't think cascade visualization is a good use case. Consecutive colors inside `Tonemap_DisplayRange` are too visually similar; we want cascade steps to be very discrete.

filament/src/ShadowMap.cpp
pixelflinger: yes they would, I typed too fast

PR #2487, opened 2020-05-06
Vulkan: add support for resizing the window.

prideout:


filament/backend/src/vulkan/VulkanDriver.cpp
```diff
@@ -254,7 +256,15 @@ void VulkanDriver::beginFrame(int64_t monotonic_clock_ns, uint32_t frameId,
     acquireWorkCommandBuffer(mContext);
     mDisposer.release(mContext.work.resources);
 
-    acquireSwapCommandBuffer(mContext);
+    // It might take several attempts to acquire a swap chain that is not marked as "out of date".
+    int attempts = 0;
+    while (!acquireSwapCommandBuffer(mContext)) {
```
pixelflinger, line 261: do you need to sleep in between attempts? is this happening because there are no buffers ready?

prideout: Ok here's the story:

*Usually* when we see VK_ERROR_OUT_OF_DATE_KHR returned by `vkAcquireNextImageKHR`, it works to simply refresh the swap chain once, and the subsequent call acquire is successful.  This is consistent with my expectations.

However, *sometimes* the subsequent acquire call is still out of date!  I don't really understand why this happens, but I have a theory that it is because we are not doing the acquisition from the UI thread, and the UI thread is what's doing the resize.

It seems to work to simply try again without sleeping, but yes maybe it'll be more robust to add a small sleep here.  Very gross stuff.  :(

For now maybe I should add a sleep(1) and augment the comment with a TODO that this needs to be investigated further.  WDYT?

prideout: I didn't answer your question: "is this happening because there are no buffers ready".  The answer is no; the acquire call has a timeout (which we set to infinity) that waits until a buffer is ready.

In other words, the horrible loop you see here is *not* for NOT_READY, instead it is for OUT_OF_DATE, which only happens after a resize, and only on some platforms.

PR #2491, opened 2020-05-06
MaterialInstance now has an optional name.

### LINKED ISSUE #2485: add names to material instances, expose them to Java

prideout:
In a subsequent PR, this feature will be exercised and tested via the
gltfio AssetLoader.

Fixes #2485.

### DIFF
--- a/filament/include/filament/MaterialInstance.h
+++ b/filament/include/filament/MaterialInstance.h
@@ -66,7 +66,7 @@ public:
     Material const* getMaterial() const noexcept;
 
     /**
-     * @return the name associated with this instance if one exists, otherwise null
+     * @return the name associated with this instance
      */
     const char* getName() const noexcept;
 
--- a/filament/include/filament/MaterialInstance.h
+++ b/filament/include/filament/MaterialInstance.h
@@ -66,6 +66,11 @@ public:
     Material const* getMaterial() const noexcept;
 
     /**
+     * @return the name associated with this instance if one exists, otherwise null
+     */
+    const char* getName() const noexcept;
+
+    /**
      * Set a uniform by name
      *
      * @param name      Name of the parameter as defined by Material. Cannot be nullptr.
--- a/filament/src/MaterialInstance.cpp
+++ b/filament/src/MaterialInstance.cpp
@@ -181,6 +181,16 @@ void FMaterialInstance::setDepthCulling(bool enable) noexcept {
     mDepthFunc = enable ? RasterState::DepthFunc::LE : RasterState::DepthFunc::A;
 }
 
+const char* FMaterialInstance::getName() const noexcept {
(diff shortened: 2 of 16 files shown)


filament/src/MaterialInstance.cpp
```diff
@@ -226,6 +227,10 @@ Material const* MaterialInstance::getMaterial() const noexcept {
     return upcast(this)->mMaterial;
 }
 
+const char* MaterialInstance::getName() const noexcept {
+    return upcast(this)->mName.c_str();
```
pixelflinger, line 231: this should call `FMaterialInstance::getName()` instead of using `mName `directly

filament/src/details/MaterialInstance.h
```diff
@@ -156,6 +158,7 @@ class FMaterialInstance : public MaterialInstance {
     backend::RasterState::DepthFunc mDepthFunc;
 
     uint64_t mMaterialSortingKey = 0;
+    utils::CString mName;
```
pixelflinger, line 161: I would put this last (because it's rarely used) and also make it `const`.

filament/include/filament/MaterialInstance.h
```diff
@@ -65,6 +65,11 @@ class UTILS_PUBLIC MaterialInstance : public FilamentAPI {
      */
     Material const* getMaterial() const noexcept;
 
+    /**
+     * @return the name associated with this instance if one exists, otherwise null
```
pixelflinger, line 69: I wonder if we shouldn't have it return `getMaterial().getName()` instead of `null`, this way users wouldn't have to check for null before printing, etc...

prideout: I thought of this, but I was worried that a user might need the ability to determine whether or not the instance has an explicitly assigned name.

filament/src/details/MaterialInstance.h
prideout: I didn't make it const because it causes C++ to complain about the default constructor, which might be why Material's name isn't const either.

filament/src/MaterialInstance.cpp
prideout: Done.

filament/src/details/MaterialInstance.h
prideout: Moved it to the end.

filament/include/filament/MaterialInstance.h
pixelflinger: I think it would be relevant only if you could set the name? I mean otherwise, there is not much they can do about it.

prideout: I've added a new commit to the PR that adds a fallback to `getName()`.

PR #2493, opened 2020-05-06
Add CMAKE_EXPORT_COMPILE_COMMANDS.

prideout:


prideout:
> If the file is generated at the root make sure .gitignore lists that file

It's generated in `out/cmake-foo/`, which is interesting when you've done builds for several platforms, because VSCode autodetects one arbitrarily.  Thankfully it lets you specify which one to use by going to your language settings file.

PR #2497, opened 2020-05-06
fix JNI objects allocation and memory corruption

pixelflinger:
- we were allocating objects with a destructor in the command stream
  which is always invalid because there is no guarantee that when 
  the callback is called, the underlaying memory is still valid
  (and it wasn't).

- AutoBuffer move-ctor wasn't moving some of its state, which would
  lead to destroying the same ref several times.

### DIFF
--- a/android/common/CallbackUtils.cpp
+++ b/android/common/CallbackUtils.cpp
@@ -34,8 +34,7 @@ static void initCallbackJni(JNIEnv* env, CallbackJni& callbackUtils) {
 
 JniBufferCallback* JniBufferCallback::make(filament::Engine* engine,
         JNIEnv* env, jobject handler, jobject callback, AutoBuffer&& buffer) {
-    void* that = engine->streamAlloc(sizeof(JniBufferCallback), alignof(JniBufferCallback));
-    return new (that) JniBufferCallback(env, handler, callback, std::move(buffer));
+    return new JniBufferCallback(env, handler, callback, std::move(buffer));
 }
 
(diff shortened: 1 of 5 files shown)


pixelflinger:
> Looks like Texture is the only remaining use of `streamAlloc`...

What do you mean?

prideout:
> > Looks like Texture is the only remaining use of `streamAlloc`...
> 
> What do you mean?

Not a critique, just an observation that after this PR we will have only one use of the public streamAlloc API, which is for the Android texture wrapper.

pixelflinger:
> > > Looks like Texture is the only remaining use of `streamAlloc`...
> > 
> > 
> > What do you mean?
> 
> Not a critique, just an observation that after this PR we will have only one use of the public streamAlloc API, which is for the Android texture wrapper.

but I looking through the whole source and I didn't find any other usage? In which file is this?

prideout:
> > > > Looks like Texture is the only remaining use of `streamAlloc`...
> > > 
> > > 
> > > What do you mean?
> > 
> > 
> > Not a critique, just an observation that after this PR we will have only one use of the public streamAlloc API, which is for the Android texture wrapper.
> 
> but I looking through the whole source and I didn't find any other usage? In which file is this?

https://github.com/google/filament/blob/master/android/filament-android/src/main/cpp/Texture.cpp#L453

prideout:
Woops you've removed it from there too!  Sorry about that.  Please ignore me :)

PR #2502, opened 2020-05-08
several fixes and improvements around objects lifetime management

pixelflinger:


android/filament-android/src/main/java/com/google/android/filament/Engine.java
```diff
@@ -634,33 +634,39 @@ private void clearNativeObject() {
         mNativeObject = 0;
     }
 
+    private void assertDestroy(boolean success) {
```
romainguy, line 637: This can be static

filament/backend/src/metal/MetalDriver.h
```diff
@@ -112,8 +112,9 @@ class MetalDriver final : public DriverBase {
 
     template<typename Dp, typename B>
     Dp* handle_cast(HandleMap& handleMap, Handle<B> handle) noexcept {
-        std::lock_guard<std::mutex> lock(mHandleMapMutex);
         assert(handle);
+        if (!handle) return nullptr; // better to get a NPE than random behavior/corruption
+        std::lock_guard<std::mutex> lock(mHandleMapMutex);
```
bejado, line 117: Probably worth adding this to `handle_const_cast` and `destruct_handle` as well (same for Vulkan driver).

PR #2506, opened 2020-05-08
gltfio: expose getMaterialInstances to Java.

prideout:
Tested locally by hacking gltf-viewer on Android, added the following
lines to MainActivity:

```kotlin
for (mat in modelViewer.asset!!.materialInstances) {
    Log.d("gltf-viewer", mat.name)
}
```

### DIFF
--- a/android/gltfio-android/src/main/cpp/FilamentAsset.cpp
+++ b/android/gltfio-android/src/main/cpp/FilamentAsset.cpp
@@ -65,6 +65,27 @@ Java_com_google_android_filament_gltfio_FilamentAsset_nGetEntities(JNIEnv* env, 
     env->ReleaseIntArrayElements(result, (jint*) entities, 0);
 }
 
+extern "C" JNIEXPORT jint JNICALL
+Java_com_google_android_filament_gltfio_FilamentAsset_nGetMaterialInstanceCount(JNIEnv*, jclass,
+        jlong nativeAsset) {
+    FilamentAsset* asset = (FilamentAsset*) nativeAsset;
+    return asset->getMaterialInstanceCount();
+}
+
+extern "C" JNIEXPORT void JNICALL
+Java_com_google_android_filament_gltfio_FilamentAsset_nGetMaterialInstances(JNIEnv* env, jclass,
+        jlong nativeAsset, jlongArray result) {
+    FilamentAsset* asset = (FilamentAsset*) nativeAsset;
+    jsize available = env->GetArrayLength(result);
+    jsize count = std::min(available, (jsize) asset->getMaterialInstanceCount());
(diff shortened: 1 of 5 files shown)


android/gltfio-android/src/main/cpp/FilamentAsset.cpp
```diff
@@ -65,6 +65,24 @@ Java_com_google_android_filament_gltfio_FilamentAsset_nGetEntities(JNIEnv* env,
...
+}
+
+extern "C" JNIEXPORT void JNICALL
+Java_com_google_android_filament_gltfio_FilamentAsset_nGetMaterialInstances(JNIEnv* env, jclass,
+        jlong nativeAsset, jlongArray result) {
+    FilamentAsset* asset = (FilamentAsset*) nativeAsset;
+    jsize available = env->GetArrayLength(result);
+    jsize count = std::min(available, (jsize) asset->getMaterialInstanceCount());
+    MaterialInstance** mats = (MaterialInstance**) env->GetLongArrayElements(result, nullptr);
+    std::copy_n(asset->getMaterialInstances(), count, mats);
```
romainguy, line 82: Will this work correctly on a 32 bit machine?

android/gltfio-android/src/main/java/com/google/android/filament/gltfio/FilamentAsset.java
```diff
@@ -99,6 +100,17 @@ public int popRenderables(@Nullable @Entity int[] entities) {
         return result;
     }
 
+    public @NonNull MaterialInstance[] getMaterialInstances() {
+        final int count = nGetMaterialInstanceCount(mNativeObject);
+        MaterialInstance[] result = new MaterialInstance[count];
+        long[] natives = new long[count];
+        nGetMaterialInstances(mNativeObject, natives);
+        for (int i = 0; i < count; i++) {
```
romainguy, line 108: Something you might want to double-check, when I tried hacking this myself for my demo I ran into a weird issue where some of the material instances where `nullptr`, which would lead to crashes in `->getMaterial()` later on. My fix was to first count the number of non nullptr instances and use that as the real count. I'll share the glTF files with you that were causing the crashes.

android/gltfio-android/src/main/cpp/FilamentAsset.cpp
prideout: Wow excellent catch, this code was wrong (fine on device, crash on emulator).  Fixed.

android/gltfio-android/src/main/java/com/google/android/filament/gltfio/FilamentAsset.java
prideout: I haven't been able to repro with that asset (on native anyway).

PR #2519, opened 2020-05-11
filament-utils: Improve ModelViewer flexibility.

prideout:


android/filament-utils-android/src/main/java/com/google/android/filament/utils/ModelViewer.kt
```diff
@@ -278,7 +286,8 @@ class ModelViewer : android.view.View.OnTouchListener {
         override fun onNativeWindowChanged(surface: Surface) {
             swapChain?.let { engine.destroySwapChain(it) }
             swapChain = engine.createSwapChain(surface)
-            displayHelper.attach(renderer, surfaceView.display)
+            surfaceView?.let { displayHelper.attach(renderer, it.display) }
+            textureView?.let { displayHelper.attach(renderer, it.display) }
```
romainguy, line 290: Excellent catch :)

android/samples/sample-gltf-viewer/src/main/java/com/google/android/filament/gltf/MainActivity.kt
```diff
@@ -83,7 +83,7 @@ class MainActivity : Activity() {
         modelViewer.loadModelGltf(buffer) { uri ->
             readCompressedAsset("models/$uri")
         }
-        modelViewer.transformToUnitCube()
+        modelViewer.transformToUnitCube(0.0f, 0.0f, 4.0f)
```
romainguy, line 86: 👍

android/filament-utils-android/src/main/java/com/google/android/filament/utils/ModelViewer.kt
```diff
@@ -134,12 +134,13 @@ class ModelViewer : android.view.View.OnTouchListener {
     }
 
     @Suppress("unused")
-    constructor(textureView: TextureView) {
+    constructor(textureView: TextureView, engine: Engine = Engine.create()) : this(engine) {
         cameraManipulator = Manipulator.Builder()
                 .targetPosition(0.0f, 0.0f, -4.0f)
```
romainguy, line 139: The camera is always positioned at z=-4, should we make that an option too? Or should the manipulator be passed by the client? That way they could use a free fly camera if they so desire.

android/samples/sample-gltf-viewer/src/main/java/com/google/android/filament/gltf/MainActivity.kt
prideout: Wait, this should probably take a `Float3` since ModelViewer already has a dependency on your math library.

android/filament-utils-android/src/main/java/com/google/android/filament/utils/ModelViewer.kt
```diff
@@ -58,7 +58,8 @@ private const val kSensitivity = 100f
  *
  * See `sample-gltf-viewer` for a usage example.
  */
-class ModelViewer : android.view.View.OnTouchListener {
+class ModelViewer(val engine: Engine) : android.view.View.OnTouchListener {
+    private val DEFAULT_OBJECT_POSITION = Float3(0.0f, 0.0f, -4.0f)
```
romainguy, line 62: Note that this won't be a static field.

PR #2520, opened 2020-05-12
fix minor bug

locobrain:
1. fix link error
2. matc support debug


prideout:
These fixes look ok to me.

tools/matc/src/matc/CommandlineConfig.cpp
```diff
@@ -116,7 +116,7 @@ CommandlineConfig::CommandlineConfig(int argc, char** argv) : Config(), mArgc(ar
 }
 
 bool CommandlineConfig::parse() {
-    static constexpr const char* OPTSTR = "hlxo:f:dm:a:p:OSEr:vV:g";
+    static constexpr const char* OPTSTR = "hlxo:f:dm:a:p:OSEr:vV:gt";
```
romainguy, line 119: `t` was left out on purpose, the `-t` is not documented as a short option. I'm not against adding it but then you should add `-t` in the help string in the `usage()` function.

Thanks!

romainguy:
Thank you!

PR #2523, opened 2020-05-12
Very basic SwiftShader support

pixelflinger:
- we only provide SwiftShader's khornos headers in third_party

- swiftshader itself must be available on the host (for instance it
  can be compiled from source).

- to enable pass -DFILAMENT_USE_SWIFTSHADER=ON option to cmake

- only GLES 3.0 is supported. Vulkan is not yet supported.

CMake should find the swiftshader libraries automatically, but if
they're installed in a non-standard place, the environment variable
SWIFTSHADER_LD_LIBRARY_PATH can be set to that place.

We also use the GLES 3.0 headers everywhere, since we don't rely on 3.1
at this point.


filament/CMakeLists.txt
```diff
@@ -306,19 +298,6 @@ else()
     add_definitions(-DFILAMENT_ENABLE_MATDBG=0)
 endif()
 
-# Android, iOS, and WebGL do not use bluegl.
-if(NOT IOS AND NOT ANDROID AND NOT WEBGL)
-    target_link_libraries(${TARGET} PRIVATE bluegl)
-endif()
-
-if (FILAMENT_SUPPORTS_VULKAN)
-    target_link_libraries(${TARGET} PUBLIC bluevk vkmemalloc)
```
prideout, line 315: Why is this stuff removed?

prideout: Never mind, just saw your other PR.

PR #2526, opened 2020-05-12
debug option to track Entities

pixelflinger:
Set FILAMENT_UTILS_TRACK_ENTITIES to true when building libutils to
activate entity tracking. This adds two public methods:

getActiveEntities() and dumpActiveEntities() the later displays the
stack trace of where the remaining entities were allocated.

This is useful for tracking leaks.

### DIFF
--- a/libs/utils/include/utils/EntityManager.h
+++ b/libs/utils/include/utils/EntityManager.h
@@ -28,7 +28,7 @@
 #endif
 
 #if FILAMENT_UTILS_TRACK_ENTITIES
-#include <utils/Log.h>
+#include <utils/ostream.h>
 #include <vector>
 #endif
 
--- a/libs/utils/include/utils/EntityManager.h
+++ b/libs/utils/include/utils/EntityManager.h
@@ -23,6 +23,15 @@
 #include <utils/Entity.h>
 #include <utils/compiler.h>
 
+#ifndef FILAMENT_UTILS_TRACK_ENTITIES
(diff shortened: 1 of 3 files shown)


libs/utils/include/utils/EntityManager.h
```diff
@@ -23,6 +23,15 @@
 #include <utils/Entity.h>
 #include <utils/compiler.h>
 
+#ifndef FILAMENT_UTILS_TRACK_ENTITIES
+#define FILAMENT_UTILS_TRACK_ENTITIES false
+#endif
+
+#if FILAMENT_UTILS_TRACK_ENTITIES
+#include <utils/Log.h>
```
prideout, line 31:
```suggestion
#include <utils/ostream.h>
```

libs/utils/include/utils/EntityManager.h
```diff
@@ -23,6 +23,15 @@
 #include <utils/Entity.h>
 #include <utils/compiler.h>
 
+#ifndef FILAMENT_UTILS_TRACK_ENTITIES
+#define FILAMENT_UTILS_TRACK_ENTITIES false
```
prideout, line 27: To prevent code rot, maybe we should default this to `true` in debug builds (i.e. when `NDEBUG` is not defined).

pixelflinger: The problem is that capturing a stack trace can be very very slow.

PR #2529, opened 2020-05-13
FilamentApp: do not recreate SwapChain upon resize.

### LINKED ISSUE #2528: FilamentApp should not re-create SwapChain upon resize
Currently, FilamentApp recreates the Filament SwapChain after a resize event occurs.

I've manually confirmed that our Android samples do not do this. So, our desktop samples probably shouldn't either.

prideout:
This is done for consistency with the Android samples. Tested with
gltf_viewer against Vulkan, Metal, and OpenGL. Dragged window between
high DPI and low DPI displays.

Since resizing does not trigger the "NSView has changed" handler in
PlatformCocoaGL, it now must to detect a resize in order to make the
required call to NSOpenGLContext::update() as per the following
documentation.

https://developer.apple.com/documentation/appkit/nsopenglcontext/1436135-update

Fixes #2528

### DIFF
--- a/filament/backend/src/opengl/PlatformCocoaGL.mm
+++ b/filament/backend/src/opengl/PlatformCocoaGL.mm
@@ -145,27 +145,30 @@ bool PlatformCocoaGL::pumpEvents() noexcept {
 }
 
 void PlatformCocoaGLImpl::updateOpenGLContext(NSView *nsView, bool resetView) {
+    auto& v = mHeadlessSwapChains;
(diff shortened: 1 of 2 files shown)


prideout:
Testing reveals that this works with Vulkan, although it's a bit laggy.

With Metal, I think there's something missing... @bejado

bejado:
I can repro the Metal issue.. I'll take a look!

PR #2531, opened 2020-05-14
don't kick the driver thread after each framegraph pass

pixelflinger:
this was a misguided optimization, on low-end devices this actually
has a large overhead (presumably because of the smaller cache and
increased branch missprediction as well as extra synchronization).

on some low-end devices we see the gl thread go from ~40ms to ~10ms
with the gltf sample.


pixelflinger:
> Does it help on high end devices though?

Not really actually. in practice the gl thread is always way slower than our "framegraph thread", so it doesn't accomplish anything to kick a buffer early. also we're kicking just after the shadow-pass, so that usually keeps that thread busy.

PR #2537, opened 2020-05-14
Shadow cascades: adjust near / far planes to tightly bound scene

bejado:
When computing the cascade split positions, adjust the near and far camera planes to tightly bound the scene dimensions.


filament/src/ShadowMap.cpp
```diff
@@ -188,23 +226,9 @@ void ShadowMap::computeShadowCameraDirectional(
...
-                nearFar.x = std::max(nearFar.x, nf.x);  // near
-                nearFar.y = std::min(nearFar.y, nf.y);  // far
-            },
-            [&wsShadowReceiversVolume](Aabb receiver) {
-                wsShadowReceiversVolume.min = min(wsShadowReceiversVolume.min, receiver.min);
-                wsShadowReceiversVolume.max = max(wsShadowReceiversVolume.max, receiver.max);
-            }
-    );
+    Aabb& wsShadowCastersVolume = cascadeParams.wsShadowCastersVolume;
+    Aabb& wsShadowReceiversVolume = cascadeParams.wsShadowReceiversVolume;
```
pixelflinger, line 230: do these need to be non-const references?

filament/src/ShadowMap.cpp
```diff
@@ -188,23 +226,9 @@ void ShadowMap::computeShadowCameraDirectional(
...
+    float2& nearFar = cascadeParams.lsNearFar;
```
pixelflinger, line 231: does this need to be a reference at all?

filament/src/ShadowMap.cpp
```diff
@@ -103,9 +103,47 @@ void ShadowMap::render(DriverApi& driver, Handle<HwRenderTarget> rt,
...
+    const float3 lightPosition = camera.getPosition();
+    const float3 dir = lightData.elementAt<FScene::DIRECTION>(index);
+    const mat4f M = mat4f::lookAt(lightPosition, lightPosition + dir, float3{ 0, 1, 0 });
+    const mat4f Mv = FCamera::rigidTransformInverse(M);
+    const mat4f V = camera.view;
+
+    float2& lsNearFar = cascadeParams.lsNearFar;
+    float2& vsNearFar = cascadeParams.vsNearFar;
+    Aabb& wsShadowCastersVolume = cascadeParams.wsShadowCastersVolume;
+    Aabb& wsShadowReceiversVolume = cascadeParams.wsShadowReceiversVolume;
```
pixelflinger, line 119: instead of using references here, just use `cascadeParams` everywhere, including in the lambda below -- it'll be less confusing and more efficient.

filament/src/ShadowMap.cpp
```diff
@@ -711,19 +735,21 @@ void ShadowMap::intersectWithShadowCasters(Aabb& UTILS_RESTRICT lightFrustum,
 }
 
 void ShadowMap::computeFrustumCorners(float3* UTILS_RESTRICT out,
-        const mat4f& UTILS_RESTRICT projectionViewInverse, float csNear, float csFar) noexcept {
+        const mat4f& UTILS_RESTRICT projectionViewInverse, float2 csNearFar) noexcept {
 
     // compute view frustum in world space (from its NDC)
     // matrix to convert: ndc -> camera -> world
+    float& near = csNearFar.x;
+    float& far = csNearFar.y;
```
pixelflinger, line 743: don't use a reference for a basic type like `float`.

filament/src/ShadowMapManager.cpp
```diff
@@ -227,12 +235,20 @@ bool ShadowMapManager::updateCascadeShadowMaps(FEngine& engine, FView& view,
                 float3{0, normalBias * texelSizeWorldSpace, 0});
     }
 
+    // Adjust the near and far planes to tighly bound the scene.
+    float near = -viewingCameraInfo.zn;
+    float far = -viewingCameraInfo.zf;
```
pixelflinger, line 240: I like your convention of using ls, vs, cs prefixes, could be useful here too.

PR #2551, opened 2020-05-18
Fix buffer size bug in sample-page-curl.

prideout:


prideout:
Egregiously wrong buffer size, I noticed this only when using a debug build of Filament.

PR #2552, opened 2020-05-18
add support for 3D textures

pixelflinger:
Only gl backend in this PR.

### DIFF
--- a/filament/backend/src/opengl/OpenGLDriver.cpp
+++ b/filament/backend/src/opengl/OpenGLDriver.cpp
@@ -577,6 +577,7 @@ void OpenGLDriver::textureStorage(OpenGLDriver::GLTexture* t,
             glTexStorage2D(t->gl.target, GLsizei(t->levels), t->gl.internalFormat,
                     GLsizei(width), GLsizei(height));
             break;
+        case GL_TEXTURE_3D:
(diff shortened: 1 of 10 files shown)


filament/backend/src/opengl/OpenGLDriver.cpp
```diff
@@ -1642,6 +1648,22 @@ void OpenGLDriver::update2DImage(Handle<HwTexture> th,
...
+void OpenGLDriver::update3DImage(Handle<HwTexture> th,
+        uint32_t level, uint32_t xoffset, uint32_t yoffset, uint32_t zoffset,
+        uint32_t width, uint32_t height, uint32_t depth,
+        PixelBufferDescriptor&& data) {
+    DEBUG_MARKER()
+
+    GLTexture* t = handle_cast<GLTexture *>(th);
+    if (data.type == PixelDataType::COMPRESSED) {
+        setCompressedTextureData(t,
+                level, xoffset, yoffset, 0, width, height, 1, std::move(data), nullptr);
```
prideout, line 1660: why is depth=1 here?

pixelflinger: whoops, good catch!

PR #2553, opened 2020-05-19
tone mapping fast path on tilers

pixelflinger:
This allows to never store/load to main memory the HDR color buffer,
saving precious bandwidth. This mode relies on the
ext_framebuffer_fetch extension being present and can't be enabled
with certain features (currently MSAA, Bloom, DoF).

This will eventually be supported on Metal and Vulkan as well.

The current implementation is a slightly hackish: it uses a special tonemapping material which itself uses private material properties. Eventually we will find a better way to support Vulkan subpasses.

### DIFF
--- a/filament/src/RenderPass.cpp
+++ b/filament/src/RenderPass.cpp
@@ -171,18 +171,16 @@ RenderPass::Command* RenderPass::sortCommands() noexcept {
 void RenderPass::execute(const char* name,
         backend::Handle<backend::HwRenderTarget> renderTarget,
         backend::RenderPassParams params) const noexcept {
-
     FEngine& engine = mEngine;
-    Command const* const first = mCommands.begin();
-    Command const* const last = mCommands.end();
-
-    // Take care not to upload data within the render pass (synchronize can commit froxel data)
     DriverApi& driver = engine.getDriverApi();
-
-    // Now, execute all commands
     driver.beginRenderPass(renderTarget, params);
-    RenderPass::recordDriverCommands(driver, first, last);
+    executeCommands(name);
     driver.endRenderPass();
+}
+
+void RenderPass::executeCommands(const char* name) const noexcept {
+    DriverApi& driver = mEngine.getDriverApi();
+    RenderPass::recordDriverCommands(driver, mCommands.begin(), mCommands.end());
 }
 
 UTILS_NOINLINE // no need to be inlined
--- a/filament/src/materials/tonemappingWithSubpass.mat
+++ b/filament/src/materials/tonemappingWithSubpass.mat
@@ -1,0 +1,81 @@
+material {
+    name : tonemap,
+    parameters : [
+        {
+            type : int,
+            name : dithering
+        },
+        {
+            type : int,
+            name : fxaa
+        },
+        {
+            type : float4,
+            name : bloom
+        }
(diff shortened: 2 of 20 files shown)


filament/src/materials/tonemappingWithSubpass.mat
```diff
@@ -0,0 +1,81 @@
...
+vertex {
+    void postProcessVertex(inout PostProcessVertexInputs postProcess) {
+        postProcess.vertex.xy = postProcess.normalizedUV;
+    }
+}
+
+fragment {
+
+    // TODO: this should be specified as a parameter
+    layout (input_attachment_index = 0, set = 0, binding = 0) uniform mediump subpassInput colorBuffer;
```
romainguy, line 35: We should add new material properties to define those. Something like:
```
inputs: [
    { name: colorBuffer, type: subpassInput }
],
// etc.
```

pixelflinger: yes, but that will tie into MRTs which we also need something like that for.
e.g. I'm not sure it should be "inputs" because a subpass can be an output too.
i don't understand subpasses well enough at this point.

filament/src/RenderPass.cpp
```diff
@@ -171,20 +171,18 @@ RenderPass::Command* RenderPass::sortCommands() noexcept {
...
     DriverApi& driver = engine.getDriverApi();
-
-    // Now, execute all commands
     driver.beginRenderPass(renderTarget, params);
-    RenderPass::recordDriverCommands(driver, first, last);
+    executeCommands(name);
     driver.endRenderPass();
 }
 
+void RenderPass::executeCommands(const char* name) const noexcept {
```
bejado, line 181: Do we need the `name` argument?

libs/filamat/src/MaterialBuilder.cpp
```diff
@@ -468,8 +474,13 @@ bool MaterialBuilder::runSemanticAnalysis() noexcept {
 
     TargetApi targetApi = params.targetApi;
     assertSingleTargetApi(targetApi);
-    ShaderModel model = static_cast<ShaderModel>(params.shaderModel);
 
+    if (mEnableFramebufferFetch) {
+        // framebuffer fetch is only available with vulkan semantics
+        targetApi = TargetApi::VULKAN;
+    }
```
bejado, line 481: Why do we need to switch to `TargetApi::VULKAN` here, but not in other places, like in `findProperties` or `generateShaders`?

romainguy:
filamat_lite is not an issue because this is strictly a private API for us. We should not expose it to 3p developers.

libs/filamat/src/MaterialBuilder.cpp
pixelflinger: honestly I think the existing code doesn't completely make sense in the first place -- so I just tried to make the smallest change possible.

pixelflinger:
> @pixelflinger I'm confused, do we need to compile to SPIRV and back to GLSL to correctly handle framebuffer fetch? If so, we'll need to do something about filamat lite, which doesn't do any compilation.

Yeah so like @romainguy said, this is just private currently. However, when we want to expose vulkan's subpasses for real, we'll have to find a solution.

PR #2554, opened 2020-05-19
sample-page-curl: add inverseTonemap and rigidity param.

prideout:


android/samples/sample-page-curl/src/main/java/com/google/android/filament/pagecurl/MainActivity.java
```diff
@@ -152,19 +151,21 @@ protected void onCreate(Bundle savedInstanceState) {
 
         mCamera.lookAt(0, 0, 3, 0, 0, 0, 0, 1, 0);
 
+         // Use unit-less light intensities since this is UI-style rendering, not physically based.
+         mCamera.setExposure(1.0f);
```
bejado, line 155:
```suggestion
        // Use unit-less light intensities since this is UI-style rendering, not physically based.
        mCamera.setExposure(1.0f);
```

PR #2561, opened 2020-05-20
Add a new tone mapping option using a LUT

pixelflinger:
This is now the default, and the tone mapper used
is now the same on desktop and mobile.
In the future this will allow us to implement many features such as
color grading at no cpu cost.


filament/src/materials/tonemapping.mat
```diff
@@ -4,7 +4,12 @@ material {
         {
             type : sampler2d,
             name : colorBuffer,
-            precision: high
+            precision: medium
```
romainguy, line 7: Oh, wow. That was a silly precision to use :)

prideout:
I haven't yet added 3D texture support to Vulkan, can we wait to make it default until then?

pixelflinger:
> I haven't yet added 3D texture support to Vulkan, can we wait to make it default until then?

oh yeah, I completely forgot about that -- and metal too.

romainguy:
Where are the before/after screenshots? :)

filament/src/Tonemapper.cpp
```diff
@@ -0,0 +1,203 @@
...
+using namespace utils;
+using namespace math;
+using namespace backend;
+
+static constexpr size_t LUT_DIMENSION = 32u;
+
+Tonemapper::Tonemapper(FEngine& engine) {
+    DriverApi& driver = engine.getDriverApi();
+
+    // it's okay to allocate the data in the command stream because it's only 256 KiB
```
bejado, line 42: What does it mean to "allocate the data in the command stream"?

pixelflinger: how I should have removed this comment, I'm using malloc now.

you can call driver.allocate() to allocate memory in the command buffer, which has the same lifetime as the command buffer itself (so you don't need a callback to free it).

PR #2562, opened 2020-05-20
Fix compilation issues under ubuntu 20 and latest clang

neshume:


neshume:
Thx guy. 
I will fix that directly in the 3rd party and submit pull request there so that when you update it will be fixed. 👍🙏

PR #2563, opened 2020-05-20
filamat: generate spirv 1.0 rather than spirv 1.3

prideout:
Since we are still on Vulkan 1.0, we cannot use SPIR-V 1.3.
If we try to do so, this error is generated:

    Invalid SPIR-V binary version 1.3 for target environment SPIR-V 1.0

In non-optimized builds, we were already generating spirv 1.0, but
when we enabled the shader optimizer, we generated spirv 1.3.

This bug has actually been around forever, but we did not notice because
we were only invoking the optimizer in release builds, which does not
enable validation.

We cannot upgrade to Vulkan 1.1 because the latest LunarG SDK for macOS
does not support it.

### DIFF
--- a/libs/filamat/src/GLSLPostProcessor.cpp
+++ b/libs/filamat/src/GLSLPostProcessor.cpp
@@ -304,7 +304,7 @@ void GLSLPostProcessor::fullOptimization(const TShader& tShader,
     GlslangToSpv(*tShader.getIntermediate(), spirv, &options);
 
     // Run the SPIR-V optimizer
-    Optimizer optimizer(SPV_ENV_UNIVERSAL_1_3);
+    Optimizer optimizer(SPV_ENV_UNIVERSAL_1_0);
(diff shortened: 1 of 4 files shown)


romainguy:
Don't we need 1.1 for external memory (for camera streams)?

pixelflinger:
> Don't we need 1.1 for external memory (for camera streams)?

Which raises the question of how many devices support only 1.0?

prideout:
This PR does not change the version of Vulkan that we use.  We use Vulkan 1.0, and always have.  (Yes, I think we should consider upgrading to Vulkan 1.1 after the next release of the LunarG SDK for macOS, and after doing research at https://vulkan.gpuinfo.org/.)

This PR fixes an assert that started to occur after we enabled the SPIRV optimizer in debug builds, which revealed a mismatch between the SPIR-V version and the Vulkan API version.

PR #2575, opened 2020-05-22
Several improvements to SSAO quality

pixelflinger:


android/filament-android/src/main/java/com/google/android/filament/View.java
```diff
@@ -171,6 +171,14 @@
          */
         @NonNull
         public QualityLevel quality = QualityLevel.LOW;
+
+        /**
+         * The upsampling setting controls the quality of the ambient occlusion buffer upsampling.
+         * The default is QualityLevel.LOW and uses bilinear filtering, a value of
```
romainguy, line 177: What does MEDIUM do?

pixelflinger: Nothing for now. I left it open because we could maybe have LOW==NEAREST.

romainguy: What do you mean by nothing? You don't get SSAO or do you get the equivalent of LOW or the equivalent of HIGH?

pixelflinger: It's like LOW.

shaders/src/ambient_occlusion.fs
```diff
@@ -11,8 +11,50 @@
...
+
 float evaluateSSAO() {
-    return textureLod(light_ssao, uvToRenderTargetUV(getNormalizedViewportCoord().xy), 0.0).r;
+    highp vec2 uv = uvToRenderTargetUV(getNormalizedViewportCoord().xy);
+
+    // Upscale the SSAO buffer in real-time, in high quality mode we use a custom bilinear
+    // filter, and we only do this if we have textureGather (ES3.1)
+    // This adds about 2.5ms @ 250MHz on Pixel 4
+
+#if !defined(GL_ES) || (defined(GL_ES) && __VERSION__ >= 310)
```
pixelflinger, line 26: that won't work with spirvcross

android/filament-android/src/main/java/com/google/android/filament/View.java
romainguy: Well, you should have the documentation explain that then :p

PR #2597, opened 2020-05-27
PostProcessManager cleanup and improvements

pixelflinger:


bejado:
Windows error:

```
error C2926: 'filament::PostProcessManager::PostProcessMaterial::<unnamed-tag>::mEngine': a default member initializer is not allowed for a member of an anonymous struct within a union 
```

PR #2607, opened 2020-05-28
 gltfio: Add support for "software instancing".

### LINKED ISSUE #1513: How to render a renderable in different positions of one scene at a time
In our practical use, we need to render a renderable several times on one scene with different transforms at the same time. For now, we have tried to load a glb file for multiple times so that we have several filamentAssets at the same time. But it turns out to be a burden on the memory if the glb file has many textures, since each filament asset will load its own buffer and textures.
We wish to have an easy way to render a renderable more than once on one scene at a time, or to copy filamentAssets without load glb for multiple times. Is there a possible solution to this?

prideout:
This adds createInstancedAsset() to AssetLoader, which creates a master asset and a set of slave instances. Vertex buffers, index buffers, textures, and material instances are shared. Entities and components are duplicated.

Instances have their own API object that exposes a subset of the master asset's entity hierarchy and an optional Animator object.

This also adds a new sample app called `gltf_instances`.  It is similar to `gltf_viewer` but with no GUI.  It has additional command line arguments for a number of instances and animation.

Java and JavaScript bindings will be added in a subsequent PR.

Fixes #1513

curl 10 41 40 AM

### DIFF
--- a/libs/gltfio/include/gltfio/Animator.h
+++ b/libs/gltfio/include/gltfio/Animator.h
@@ -18,10 +18,14 @@
 #define GLTFIO_ANIMATOR_H
 
 #include <gltfio/FilamentAsset.h>
+#include <gltfio/FilamentInstance.h>
 
 namespace gltfio {
 
-namespace details { struct FFilamentAsset; }
+namespace details {
+    struct FFilamentAsset;
+    struct FFilamentInstance;
+}
 
 struct AnimatorImpl;
 
@@ -71,9 +75,11 @@ private:
 
     /*! \cond PRIVATE */
     friend struct details::FFilamentAsset;
+    friend struct details::FFilamentInstance;
     /*! \endcond */
 
     Animator(FilamentAsset* asset);
+    Animator(FilamentInstance* instance);
     ~Animator();
     AnimatorImpl* mImpl;
 };
--- a/libs/gltfio/include/gltfio/Animator.h
+++ b/libs/gltfio/include/gltfio/Animator.h
@@ -78,8 +78,7 @@ private:
     friend struct details::FFilamentInstance;
     /*! \endcond */
 
-    Animator(FilamentAsset* asset);
-    Animator(FilamentInstance* instance);
+    Animator(details::FFilamentAsset* asset, details::FFilamentInstance* instance);
     ~Animator();
     AnimatorImpl* mImpl;
 };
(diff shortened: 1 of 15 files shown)


libs/gltfio/include/gltfio/Animator.h
```diff
@@ -18,10 +18,14 @@
 #define GLTFIO_ANIMATOR_H
 
 #include <gltfio/FilamentAsset.h>
+#include <gltfio/FilamentInstance.h>
 
 namespace gltfio {
 
-namespace details { struct FFilamentAsset; }
+namespace details {
```
pixelflinger, line 25: in another PR, maybe we should get rid of details:: in gltfio too?

prideout: Yes I would definitely like to do that.

davidwarford:
Hay just a question about this part of the API. this seems to require you know how many instances you are going to need upfront. What is the intended method to add new instances at run time? I don't want to keep loading the asset every time I need to draw a new copy of my model.

prideout:
You can use the new `AssetLoader::createInstance()` method to dynamically add instances, but it has a few limitations.  Firstly, you still need to load the asset using `createInstancedAsset` (you can set numInstances to 1).  Secondly, the newly added instances do not support animation yet -- that's TBD.

davidwarford:
This is not exposed in the web api yet, any idea on when it could be?

prideout:
It is exposed in the web API although I haven't had a chance to test it on that platform:

https://github.com/google/filament/blob/main/web/filament-js/jsbindings.cpp#L1787
https://github.com/google/filament/blob/main/web/filament-js/filament.d.ts#L568

davidwarford:
oh, I'm sorry must have been added after I pulled or something, will have to pull again. Thank you!

davidwarford:
@prideout after pulling and some confusion, I noticed that yes the function is in the definition file and in the jsbindings file it does not exist in the filament.js file. Should this be elevated to an issue?

prideout:
`filament.js` is the glue layer between `filament.wasm` and the client application, it does not contain entry points unless they require specialized glue code (e.g. Scene's `addEntity` method isn't in there either).

However I can easily believe that there is an issue with the web bindings for this entry point since I have not yet tested.  Feel free to file an issue if you see an error message in Chrome's developer console when calling this method.

PR #2610, opened 2020-05-29
Metal: use render target height to adjust viewport when blitting

bejado:


shaders/src/post_process.vs
```diff
@@ -8,7 +8,7 @@ void main() {
 
 // In Vulkan and Metal, texture coords are Y-down. In OpenGL, texture coords are Y-up.
 #if defined(TARGET_METAL_ENVIRONMENT) || defined(TARGET_VULKAN_ENVIRONMENT)
-    inputs.texelCoords.y = frameUniforms.resolution.y - 1.0 - inputs.texelCoords.y;
+    inputs.texelCoords.y = frameUniforms.resolution.y - inputs.texelCoords.y;
```
bejado, line 11: This is somewhat related, but I'm not sure why I put that `-1.0` originally. This caused a one pixel shift vertically.

PR #2611, opened 2020-05-29
Prep work for proper color grading

romainguy:


filament/src/Renderer.cpp
```diff
@@ -307,15 +307,14 @@ void FRenderer::renderJob(ArenaScope& arena, FView& view) {
             .hasContactShadows = scene.hasContactShadows()
     };
 
-    const ToneMappingConfig toneMappingConfig{
-            .toneMappingAsSubpass =
-                    toneMapping &&
+    const ColorGradingConfig colorGradingConfig{
+            .asSubpass =
```
pixelflinger, line 311: This needs to be `false` if post processing is disabled, because in that case, color-grading is also disabled. We don't want to take the subpass path during the color pass.
I think it would be better to keep the `tonemapping` boolean (renamed) in this function.

filament/src/Renderer.cpp
```diff
@@ -396,25 +398,25 @@ void FRenderer::renderJob(ArenaScope& arena, FView& view) {
...
-                        scale, bloomOptions,
-                        toneMappingConfig.dithering);
+        if (colorGrading) {
+            if (!colorGradingConfig.asSubpass) {
+                input = ppm.colorGrading(fg, input,
+                                         colorGradingConfig.ldrFormat,
+                                         colorGradingConfig.translucent,
+                                         colorGradingConfig.fxaa,
+                                         scale, bloomOptions,
+                                         colorGradingConfig.dithering);
```
pixelflinger: continuation should be 8 spaces

filament/src/Renderer.cpp
```diff
@@ -558,10 +560,10 @@ FrameGraphId<FrameGraphTexture> FRenderer::refractionPass(FrameGraph& fg,
         config.refractionLodOffset = refractionLodOffset;
         config.clearFlags = TargetBufferFlags::NONE;
         output = colorPass(fg, "Color Pass (transparent)",
-                desc, config, toneMappingConfig, translucentPass, view);
+                           desc, config, colorGradingConfig, translucentPass, view);
```
pixelflinger, line 565: continuation 8 spaces

PR #2612, opened 2020-05-29
gltfio: Use JobSystem to compute tangents and AABB's.

prideout:
On a Pixel 3a, this reduces load time for Bistro from ~6.5 seconds to ~2.5 seconds, where "load
time" is the time between app launch and the first mesh being rendered.

After these optimizations, we still see some low CPU utilization during the load, which appears
to be due to AssetManager usage on the Java side. It might be possible to optimize this too.

Note that AABB computation is optional since glTF has built-in min/max values, but we currently
enable it by default for robustness. In a future PR I would like to change this from opt-out to
opt-in.


prideout:
(using glb instead of gltf would make this faster too)

PR #2613, opened 2020-05-30
Introduce ColorGrading API

romainguy:
A ColorGrading object will eventually be settable on a View to
choose the tone mapping and other color transformations.

There is currently only a single default ColorGrading instance
used by all views.

### DIFF
--- a/filament/src/ColorGrading.cpp
+++ b/filament/src/ColorGrading.cpp
@@ -62,7 +62,9 @@ ColorGrading* ColorGrading::Builder::build(Engine& engine) {
     return upcast(engine).createColorGrading(*this);
 }
 
-std::function<float3(float3)> selectToneMapping(ColorGrading::ToneMapping toneMapping) {
+using ToneMapper = float3(*)(float3);
+
+ToneMapper selectToneMapping(ColorGrading::ToneMapping toneMapping) {
     switch(toneMapping) {
         case ColorGrading::ToneMapping::LINEAR:
             return tonemap::Linear;
@@ -87,7 +89,7 @@ FColorGrading::FColorGrading(FEngine& engine, const Builder& builder) {
     constexpr size_t elementSize = sizeof(half4);
     void* const data = malloc(lutElementCount * elementSize);
 
-    auto toneMapper = selectToneMapping(builder->mToneMapping);
+    ToneMapper toneMapper = selectToneMapping(builder->mToneMapping);
 
     //auto now = std::chrono::steady_clock::now();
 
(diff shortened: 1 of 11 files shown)


filament/src/ColorGrading.cpp
```diff
@@ -37,7 +41,44 @@ using namespace backend;
...
+ColorGrading::Builder& ColorGrading::Builder::toneMapping(ToneMapping toneMapping) noexcept {
+    mImpl->mToneMapping = toneMapping;
+    return *this;
+}
+
+ColorGrading* ColorGrading::Builder::build(Engine& engine) {
+    return upcast(engine).createColorGrading(*this);
+}
+
+std::function<float3(float3)> selectToneMapping(ColorGrading::ToneMapping toneMapping) {
```
pixelflinger, line 65: It could be slightly more efficient if we just used a pointer-to-member, I think std::function<> is pretty big and has to do some type erasing, which  in this case is not needed, since we know the type.
You could just have a `using TonemapperFunc = float3(ColorGRading::*)(float3);`

filament/src/ColorGrading.cpp
```diff
@@ -54,8 +97,8 @@ Tonemapper::Tonemapper(FEngine& engine) {
     JobSystem& js = engine.getJobSystem();
     auto slices = js.createJob();
     for (size_t b = 0; b < LUT_DIMENSION; b++) {
-        auto job = js.createJob(slices,[data, b](JobSystem&, JobSystem::Job*) {
-            half4* UTILS_RESTRICT p = (half4*)data + b * LUT_DIMENSION * LUT_DIMENSION;
+        auto job = js.createJob(slices, [&, data, b](JobSystem&, JobSystem::Job*) {
```
pixelflinger, line 100: I prefer we list exactly what we need to capture. If you switch to a pointer-to-member it might be possible to capture `tonemapper` by value, which would be better.

filament/src/PostProcessManager.cpp
```diff
@@ -979,7 +975,7 @@ void PostProcessManager::colorGradingSubpass(DriverApi& driver,
     Handle<HwRenderPrimitive> const& fullScreenRenderPrimitive = engine.getFullScreenRenderPrimitive();
 
     FMaterialInstance* mi = mColorGradingAsSubpass.getMaterialInstance();
-    mi->setParameter("lut", mTonemapper->getHwHandle(), {
+    mi->setParameter("lut", mEngine.getDefaultColorGrading()->getHwHandle(), {
```
pixelflinger, line 978: This will come from the View eventually, right?

filament/src/ColorGrading.cpp
romainguy: Yeah I was hoping for some kind of small std::function optimization. But I can use a raw function pointer.

filament/src/ColorGrading.cpp
romainguy: We go over with an std::function, we should be good with the pointer. Eventually I will need to pass more data so all the inputs will be inside an intermediate struct anyway.

filament/src/PostProcessManager.cpp
romainguy: That's correct, this is temporary. We will still use this default color grading by default on Views though.

PR #2615, opened 2020-05-30
Add per-view color grading LUT

romainguy:
Each View can now have its own `ColorGrading` object, which defines various color transforms.
These transforms currently include:
- Tone mapping (linear, Reinhard, ACES, Filmic and display range for debugging)
- White balance (temperature/tint offsets)

Here is an example of a scene rendered using ACES:
Screen Shot 2020-05-30 at 11 20 42 PM

And the same scene using ACES and a custom white balance to emphasize the sunset in the IBL:
Screen Shot 2020-05-30 at 11 20 38 PM

### DIFF
--- a/filament/include/filament/ColorGrading.h
+++ b/filament/include/filament/ColorGrading.h
@@ -35,7 +35,7 @@ class UTILS_PUBLIC ColorGrading : public FilamentAPI {
 public:
     enum class ToneMapping : uint8_t {
         LINEAR        = 0,     //!< Linear tone mapping (i.e. no tone mapping)
-        ACES          = 1,     //!< ACES tone mapping (ODT+RRT in )
+        ACES          = 1,     //!< ACES tone mapping, with a brightness modifier
         FILMIC        = 2,     //!< Filmic tone mapping, modelled after ACES but applied in sRGB space
         REINHARD      = 3,     //!< Reinhard luma-based tone mapping
         DISPLAY_RANGE = 4,     //!< Debug tone mapping to validate scene exposure
@@ -52,6 +52,9 @@ public:
         Builder& operator=(Builder&& rhs) noexcept;
 
         Builder& toneMapping(ToneMapping toneMapping) noexcept;
+
+        // TODO: document: temperature from -1 to +1, tint from -1 to +1; clipped otherwise
+        Builder& whiteBalance(float temperature, float tint) noexcept;
 
         ColorGrading* build(Engine& engine);
 
--- a/filament/src/ColorGrading.cpp
+++ b/filament/src/ColorGrading.cpp
@@ -18,6 +18,7 @@
 
 #include "details/Engine.h"
 
+#include "ColorSpace.h"
 #include "FilamentAPI-impl.h"
 
 #include <math/vec2.h>
@@ -44,6 +45,10 @@ using namespace math;
 using namespace backend;
 
 static constexpr size_t LUT_DIMENSION = 32u;
+
+//------------------------------------------------------------------------------
+// Builder
+//------------------------------------------------------------------------------
 
 struct ColorGrading::BuilderDetails {
     ToneMapping toneMapping = ToneMapping::ACES;
@@ -75,6 +80,10 @@ ColorGrading* ColorGrading::Builder::build(Engine& engine) {
     return upcast(engine).createColorGrading(*this);
 }
 
+//------------------------------------------------------------------------------
+// Tone mapping
+//------------------------------------------------------------------------------
+
 using ToneMapper = float3(*)(float3);
 
 ToneMapper selectToneMapping(ColorGrading::ToneMapping toneMapping) {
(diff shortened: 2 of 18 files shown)


filament/include/filament/ColorGrading.h
```diff
@@ -53,6 +53,9 @@ class UTILS_PUBLIC ColorGrading : public FilamentAPI {
 
         Builder& toneMapping(ToneMapping toneMapping) noexcept;
 
+        // TODO: document: temperature from -1 to +1, tint from -1 to +1; clipped otherwise
```
pixelflinger, line 56: Shouldn't this be a value in K? or is there a way to go from either a value in K or from a standard white point (e.g. D65) to this -1/+1 value?

filament/src/ColorGrading.cpp
```diff
@@ -89,13 +102,13 @@ FColorGrading::FColorGrading(FEngine& engine, const Builder& builder) {
...
 
-    ToneMapper toneMapper = selectToneMapping(builder->mToneMapping);
+    ToneMapper toneMapper = selectToneMapping(builder->toneMapping);
 
     //auto now = std::chrono::steady_clock::now();
 
     // Multithreadedly generate the tone mapping 3D look-up table using 32 jobs
     // Slices are 8 KiB (128 cache lines) apart.
-    // This takes about 3-6ms on Android in Release mode
+    // This takes about 3-6ms on Android in Release mode on Pixel 4
```
pixelflinger, line 111: Actually this was 3ms on Pixel4 and 6ms on MotoG7, that's why I didn't specify Pixel4.

filament/src/ColorGrading.cpp
```diff
@@ -104,20 +117,24 @@ FColorGrading::FColorGrading(FEngine& engine, const Builder& builder) {
...
+
                     *p++ = half4{ v, 0 };
                 }
             }
         });
         js.run(job);
     }
+
+    // TODO: Should we do a runAndRetain() and defer the wait() + texture creation until
+    //       getHwHandle() is invoked?
```
pixelflinger, line 137: That's an interesting idea, but we've never really used the JobSystem in that way.
Also if we did this we'd have to handle the case of multiple calls in a row. 
As long as we're talking < 10ms I wouldn't go there.

filament/src/PostProcessManager.cpp
```diff
@@ -1027,17 +1028,17 @@ FrameGraphId<FrameGraphTexture> PostProcessManager::colorGrading(FrameGraph& fg,
         }
     }
 
-    auto& ppColorGrading = fg.addPass<PostProcessColorGrading>("colorGrading",
+    auto& ppColorGrading = fg.addPass<PostProcessColorGrading>("colorGradingOptions",
```
pixelflinger, line 1031: That's a strange name for a pass?

filament/src/details/View.h
```diff
@@ -205,6 +206,14 @@ class FView : public View {
         return mToneMapping;
     }
 
+    void setColorGrading(FColorGrading* colorGrading) noexcept {
+        mColorGrading = colorGrading;
```
pixelflinger, line 210: should we make it that if passing `nullptr` this reverts to the default one?

filament/include/filament/ColorGrading.h
romainguy: White balance can be applied in different way. It's common to treat color/white balance as an adjustment between -100 and +100, both for the temperature and the tint. Basically think of it as an offset from D65 both horizontally (temperature) and vertically (tint).

filament/src/ColorGrading.cpp
romainguy: Will remove then.

filament/src/ColorGrading.cpp
romainguy: Yeah that's why I just left it as a TODO. Handling multiple calls in a row wouldn't be difficult I think but it does complicate things a bit.

filament/src/PostProcessManager.cpp
romainguy: Argh automatic refactoring that was a bit too eager. Will fix.

filament/src/details/View.h
romainguy: Yes, good catch.

filament/include/filament/ColorGrading.h
pixelflinger: I imagine this depends from the color temperature of the screen.  I'm wondering if we shouldn't/couldn't use this to correct for screens that are too blue "because it looks better".

Shouldn't we have a utility then, that we could give it the assumed white-point (e.g. D65) and the desired one, and it would return the proper value for this function?

What does -100 or +100 mean? do they correspond to a known temperature? I think we should have more comments about what the parameters are and how to use them.

romainguy: Yeah I still need to write the documentation (and Java bindings). Remember the main use case for color grading is artistic control. This I not to fix broken displays (because you'd need to know what the display's true white point is as well). As for setting the temperature directly (like you'd do in Lightroom) it works only if you know the white point of the original image which you do get from a camera, but in our case we don't know what it is. We must also not confuse the white point of the color space (always D65 for us) and the white point of the image.

PR #2618, opened 2020-06-01
Add channel mixer and docs to ColorGrading

romainguy:
Example using channel mixer to create a sepia tone effect:

Screen Shot 2020-05-31 at 11 56 06 PM

### DIFF
--- a/filament/src/ColorGrading.cpp
+++ b/filament/src/ColorGrading.cpp
@@ -18,8 +18,14 @@
 
 #include "details/Engine.h"
 
-#include "ColorSpace.h"
 #include "FilamentAPI-impl.h"
+
+#include "ColorSpace.h"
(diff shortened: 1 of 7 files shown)


filament/src/ColorGrading.cpp
```diff
@@ -76,30 +78,16 @@ ColorGrading::Builder& ColorGrading::Builder::whiteBalance(float temperature, fl
     return *this;
 }
 
-ColorGrading* ColorGrading::Builder::build(Engine& engine) {
-    return upcast(engine).createColorGrading(*this);
+ColorGrading::Builder& ColorGrading::Builder::channelMixer(
```
pixelflinger, line 81: This is basically a 3x3, so it's basically a colorspace transform :-)

romainguy: Pretty much :)

PR #2622, opened 2020-06-02
gltfio: Add get-by-name and get-by-prefix methods.

prideout:
Java bindings will be added in a subsequent PR.

Note that this PR also adds Thibaut's trie implementation to third_party.


libs/gltfio/include/gltfio/FilamentAsset.h
```diff
@@ -137,6 +135,35 @@ class FilamentAsset {
...
+     *
+     * @param name Null-terminated string to match.
+     * @param entities Pointer to an array to populate. The array must be at least "count" large.
+     * @param count Maximum number of entities to retrieve.
+     *
+     * @return If entities is non-null, the number of entities written to the entity pointer.
+     * Otherwise this returns the number of entities with the given name.
+     */
+    size_t getEntitiesByName(const char* name, utils::Entity* entities, size_t maxCount) const
+            noexcept;
```
romainguy, line 152: Don't split the line at noexcept, split earlier. Same for the method below.

libs/gltfio/src/FFilamentAsset.h
```diff
@@ -150,46 +126,25 @@ struct FFilamentAsset : public FilamentAsset {
...
+    utils::Entity getFirstEntityByName(const char* name) noexcept;
 
-    utils::Entity getWireframe() noexcept {
-        if (!mWireframe) {
-            mWireframe = new Wireframe(this);
-        }
-        return mWireframe->mEntity;
-    }
+    size_t getEntitiesByName(const char* name, utils::Entity* entities, size_t maxCount) const
+            noexcept;
```
romainguy, line 134: Weird split here (method below too)

libs/gltfio/include/gltfio/FilamentAsset.h
prideout: Fixed.

PR #2628, opened 2020-06-02
Add new async method to ModelViewer.

prideout:
If you now try loading Bistro, the skybox shows up fairly quickly.
This works by moving AssetManager uncompression off the UI thread.

### DIFF
--- a/android/filament-utils-android/src/main/java/com/google/android/filament/utils/ModelViewer.kt
+++ b/android/filament-utils-android/src/main/java/com/google/android/filament/utils/ModelViewer.kt
@@ -16,7 +16,6 @@
 
 package com.google.android.filament.utils
 
-import android.os.AsyncTask
 import android.view.MotionEvent
 import android.view.Surface
 import android.view.SurfaceView
@@ -25,6 +24,7 @@ import com.google.android.filament.*
 import com.google.android.filament.android.DisplayHelper
 import com.google.android.filament.android.UiHelper
 import com.google.android.filament.gltfio.*
+import kotlinx.coroutines.*
 import java.nio.Buffer
 
 private const val kNearPlane = 0.5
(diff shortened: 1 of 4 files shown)


android/filament-utils-android/src/main/java/com/google/android/filament/utils/ModelViewer.kt
```diff
@@ -310,3 +346,5 @@ class ModelViewer(val engine: Engine) : android.view.View.OnTouchListener {
         private val kDefaultObjectPosition = Float3(0.0f, 0.0f, -4.0f)
     }
 }
+
+typealias ResourceTask = AsyncTask<Unit, Unit, Map<String, Buffer>>
```
romainguy, line 350: `AsyncTask` is deprecated and shouldn't be used anymore. You should look into coroutines instead since this is a Kotlin library.

prideout: Done, added an impl dep on kotlinx-coroutines too.

android/filament-utils-android/src/main/java/com/google/android/filament/utils/ModelViewer.kt
```diff
@@ -281,6 +297,25 @@ class ModelViewer(val engine: Engine) : android.view.View.OnTouchListener {
         return true
     }
 
+    private suspend fun fetchResources(callback: (String) -> Buffer) = withContext(Dispatchers.IO) {
```
romainguy, line 300: I don't think you need the `withContext(Dispatchers.IO)` here since you already launched the coroutine in the IO dispatcher.

android/samples/sample-gltf-viewer/src/main/java/com/google/android/filament/gltf/MainActivity.kt
```diff
@@ -80,9 +80,8 @@ class MainActivity : Activity() {
             input.read(bytes)
             ByteBuffer.wrap(bytes)
         }
-        modelViewer.loadModelGltf(buffer) { uri ->
-            readCompressedAsset("models/$uri")
-        }
+
+        modelViewer.loadModelGltfAsync(buffer, { uri -> readCompressedAsset("models/$uri") })
```
romainguy, line 84: You don't necessarily need to use the `Async` suffix since coroutine methods are marked in the IDE.

android/filament-utils-android/src/main/java/com/google/android/filament/utils/ModelViewer.kt
prideout: Nice!  Done.

android/samples/sample-gltf-viewer/src/main/java/com/google/android/filament/gltf/MainActivity.kt
prideout: Well I need to disambiguate it from its Sync counterpart, which I think should still exist for simple use cases.

PR #2629, opened 2020-06-02
Fix typescript View set/getAmbientOcclusion wrong type

MingweiSamuel:


MingweiSamuel:
@googlebot I signed it!

PR #2648, opened 2020-06-05
Implement framebuffer fetch for Metal backend

bejado:
This saves about .5ms on iOS.

### DIFF
--- a/filament/backend/src/metal/MetalDriver.mm
+++ b/filament/backend/src/metal/MetalDriver.mm
@@ -568,7 +568,11 @@ bool MetalDriver::isRenderTargetFormatSupported(TextureFormat format) {
 }
 
 bool MetalDriver::isFrameBufferFetchSupported() {
+#if defined(IOS)
+    return true;
+#else
     return false;
+#endif
 }
 
 bool MetalDriver::isFrameTimeSupported() {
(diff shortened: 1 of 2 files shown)


filament/backend/src/metal/MetalDriver.mm
```diff
@@ -917,8 +921,22 @@
...
+    }
+    if (any(buffers & TargetBufferFlags::COLOR2)) {
+        srcAttachIndex = 2;
+    }
+    if (any(buffers & TargetBufferFlags::COLOR3)) {
+        srcAttachIndex = 3;
+    }
+    MetalRenderTarget::Attachment srcColorAttachment = srcTarget->getColorAttachment(srcAttachIndex);
+
+    // We always blit to the COLOR0 attachment.
```
pixelflinger, line 939: This seems unrelated?

filament/backend/src/metal/MetalDriver.mm
```diff
@@ -934,9 +952,12 @@
             srcTarget->height - (NSUInteger) srcRect.bottom - srcRect.height,
             srcRect.width, srcRect.height);
 
+    const NSInteger dstHeight =
+            dstTarget->isDefaultRenderTarget() ?
+            mContext->currentSurface->getSurfaceHeight() : dstTarget->height;
     MTLRegion dstRegion = MTLRegionMake2D(
             (NSUInteger) dstRect.left,
-            dstTarget->height - (NSUInteger) dstRect.bottom - dstRect.height,
+            dstHeight - (NSUInteger) dstRect.bottom - dstRect.height,
```
pixelflinger, line 960: This also seems unrelated?

bejado: Technically unrelated, but this bug was "exposed" by framebuffer fetch.

filament/backend/src/metal/MetalDriver.mm
bejado: I originally added this because I thought we needed it (we're bliting from COLOR1), though it looks like you create a new RT for the blit so it isn't strictly needed. But the way it was written before was incorrect, so probably best to leave it.

PR #2651, opened 2020-06-06
Add new curves color grading tool

romainguy:
This change improves the UI of the color grading tools in `material_sandbox`. Screenshots of before color grading and after:

Color grading off

Color grading on


prideout:
Very cool UI.

PR #2652, opened 2020-06-06
Fix: cannot use parts index for material index.

roxlu:
Although I don't know the exact details of the `MeshWriter` and `MeshReader` I noticed that it's currently possible that there are more `Part`s in a filamesh file than that there are materials. This results in a segfault when using the `i` variable to select a material name. To fix this we have to use `parts[i].material` as that points to the correct and stored material.

To reproduce this, you have to create a particle system in blender and let it pick a random object from a collection. The number of used materials needs to be smaller then the number of objects in the collection. See the attached blend file; export this to .obj and convert that to .filamesh. 
[particles.tar.gz](https://github.com/google/filament/files/4740592/particles.tar.gz)


romainguy:
Thanks!

PR #2656, opened 2020-06-09
Add new color grading feature, vibrance

romainguy:
Compared to the saturation API, vibrance tries to not saturate already saturated colors and tries to preserve red hues. Vibrance also does not desaturate the image completely.

Saturation = 2.0
Screen Shot 2020-06-08 at 6 10 45 PM

Vibrance = 2.0
Screen Shot 2020-06-08 at 6 10 48 PM

Saturation = 0.0
Screen Shot 2020-06-08 at 6 11 03 PM

Vibrance = 0.0
Screen Shot 2020-06-08 at 6 10 58 PM

### DIFF
--- a/filament/src/ColorGrading.cpp
+++ b/filament/src/ColorGrading.cpp
@@ -63,8 +63,9 @@ struct ColorGrading::BuilderDetails {
     float3 offset           = {0.0f};
     float3 power            = {1.0f};
     // Color adjustments
-    float  saturation       = 1.0f;
     float  contrast         = 1.0f;
+    float  vibrance         = 1.0f;
+    float  saturation       = 1.0f;
     // Curves
     float3 shadowGamma      = {1.0f};
     float3 midPoint         = {1.0f};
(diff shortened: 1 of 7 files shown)


filament/src/ColorGrading.cpp
```diff
@@ -257,15 +263,30 @@ inline float3 colorDecisionList(float3 v, float3 slope, float3 offset, float3 po
...
+UTILS_ALWAYS_INLINE
+inline constexpr float3 saturation(float3 v, float saturation) {
+    const float3 y = dot(v, LUMA_REC709);
     return y + saturation * (v - y);
 }
 
+UTILS_ALWAYS_INLINE
+inline float3 vibrance(float3 v, float vibrance) {
+    float r = v.r - max(v.g, v.b);
+    float s = (vibrance - 1.0f) / (1.0f + exp(-r * 3.0f)) + 1.0f;
```
pixelflinger, line 280: use `std::exp` or, better, `using namespace std;` in the method.

romainguy: Ok!

PR #2662, opened 2020-06-09
Perform non-ACES grading in BT.2020

romainguy:


filament/src/ColorGrading.cpp
```diff
@@ -31,6 +31,7 @@
 #include <utils/JobSystem.h>
 #include <utils/Systrace.h>
 
+#include <cstring>
```
pixelflinger, line 34: do we need cstring? I don't trust that header :-)

filament/src/ColorGrading.cpp
```diff
@@ -148,6 +151,13 @@ ColorGrading::Builder& ColorGrading::Builder::curves(
 }
 
 ColorGrading* ColorGrading::Builder::build(Engine& engine) {
+    // We want to see if any of the default adjustment values have been modified
+    // We skip the tonemapping operator on purpose since we always want to apply it
+    BuilderDetails defaults;
+    defaults.toneMapping = mImpl->toneMapping;
+    bool hasAdjustments = memcmp(&defaults, mImpl, sizeof(BuilderDetails)) != 0;
```
pixelflinger, line 158: Be careful with `memcmp()` if `BuilderDetails` ever has padding. It would be better to add an `operator=`, even if all it does is `memcmp`, because then we can have a comment there that says the structure can't have padding.
This bit us before.

PR #2663, opened 2020-06-09
Add support to gltfio for glTF cameras

bejado:
Still TODO:
- [ ] Java bindings
- [ ] JS bindings

cameras_demo

### DIFF
--- a/filament/include/filament/Camera.h
+++ b/filament/include/filament/Camera.h
@@ -213,7 +213,7 @@ public:
      * aspect ratio:
      *     mat4::scaling(double3 {1.0, width / height, 1.0});
      *
-     * By default, this is an identiy matrix.
+     * By default, this is an identity matrix.
      *
      * @param scaling     matrix to be applied after the projection matrix.
      *
--- a/filament/include/filament/Camera.h
+++ b/filament/include/filament/Camera.h
@@ -217,7 +217,7 @@ public:
      *
      * By default, this is an identity matrix.
      *
-     * @param scaling     diaganol of the scaling matrix to be applied after the projection matrix.
+     * @param scaling     diagonal of the scaling matrix to be applied after the projection matrix.
      *
      * @see setProjection, setLensProjection, setCustomProjection
      */
@@ -246,7 +246,7 @@ public:
 
     /** Returns the scaling amount used to scale the projection matrix.
      *
-     * @return the diaganol of the scaling matrix applied after the projection matrix.
+     * @return the diagonal of the scaling matrix applied after the projection matrix.
      *
      * @see setScaling
      */
--- a/filament/include/filament/Camera.h
+++ b/filament/include/filament/Camera.h
@@ -209,7 +209,7 @@ public:
     /** Sets an additional matrix that scales the projection matrix.
      *
      * This is useful to adjust the aspect ratio of the camera independent from its projection.
-     * First, pass an aspect of 1.0 to setProjection. Then set the scaling with your desired aspect
+     * First, pass an aspect of 1.0 to setProjection. Then set the scaling with the desired aspect
      * ratio:
      *
      *     camera->setScaling(double4 {1.0, width / height, 1.0, 1.0});
@@ -217,7 +217,7 @@ public:
      *
      * By default, this is an identity matrix.
      *
-     * @param scaling     diaganol of the matrix to be applied after the projection matrix.
+     * @param scaling     diaganol of the scaling matrix to be applied after the projection matrix.
      *
      * @see setProjection, setLensProjection, setCustomProjection
      */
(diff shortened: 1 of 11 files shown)


filament/src/Camera.cpp
```diff
@@ -177,7 +181,7 @@ mat4f UTILS_NOINLINE FCamera::getViewMatrix() const noexcept {
 
 Frustum FCamera::getFrustum() const noexcept {
     // for culling purposes we keep the far plane where it is
-    return FCamera::getFrustum(mProjectionForCulling, getViewMatrix());
+    return FCamera::getFrustum(mScalingMatrix * mProjectionForCulling, getViewMatrix());
```
pixelflinger, line 184: use `getCullingProjectionMatrix()` here

filament/src/details/Camera.h
```diff
@@ -161,6 +163,7 @@ class FCamera : public Camera {
 
     math::mat4 mProjection;            // projection matrix (infinite far)
     math::mat4 mProjectionForCulling;  // projection matrix (with far plane)
+    math::mat4 mScalingMatrix;         // additional scaling applied to projection
```
pixelflinger, line 166: Store this as a `vec4` representing the diagonal, then everywhere you use `mScalingMatrix` replace with `mat4(mScalingDiagonal)` -- this will actually be a lot more efficient because you're telling the compiler that most coefficient in the matrix are zero (for the multiply), also takes less space.

libs/gltfio/src/AssetLoader.cpp
```diff
@@ -794,6 +798,56 @@ void FAssetLoader::createLight(const cgltf_node* node, Entity entity) {
...
+        filamentCamera->setProjection(yfovDegrees, aspect,
+                projection.znear, far,
+                filament::Camera::Fov::VERTICAL);
+    } else if (camera->type == cgltf_camera_type_orthographic) {
+        auto& projection = camera->data.orthographic;
+
+        const double left   = -projection.xmag / 2.0;
+        const double right  =  projection.xmag / 2.0;
+        const double bottom = -projection.ymag / 2.0;
+        const double top    =  projection.ymag / 2.0;
```
pixelflinger, line 837: `* 0.5` instead of `/ 2.0`

libs/gltfio/include/gltfio/FilamentAsset.h
```diff
@@ -76,6 +76,30 @@ class FilamentAsset {
...
+
+        // Name of the camera node present in the glTF file. May be null.
+        char* name;
+
+        // The aspect ratio present in the glTF camera node.
+        // A value of 0 means the user should adjust the aspect ratio to match the Viewport. In this
+        // case, gltfio sets the camera projection's aspect ratio to 1.0, and allows the user to
+        // adjust it via filament::Camera::setScalingMatrix.
+        double requestedAspectRatio;
+    };
```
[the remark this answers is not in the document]
bejado: Okay, for now I've decided to remove the that struct entirely, and instead set the scaling matrix for clients. They can adjust it / override it if they choose. The scenario I described probably isn't all that common, and it seems like other glTF viewers ignore aspect ratio entirely.

filament/include/filament/Camera.h
```diff
@@ -206,6 +206,23 @@ class UTILS_PUBLIC Camera : public FilamentAPI {
...
+     * This is useful to adjust the aspect ratio of the camera independent from its projection.
+     * First, pass an aspect of 1.0 to setProjection. Then set the scaling with the desired aspect
+     * ratio:
+     *
+     *     camera->setScaling(double4 {1.0, width / height, 1.0, 1.0});
+     *
+     *
+     * By default, this is an identity matrix.
+     *
+     * @param scaling     diaganol of the scaling matrix to be applied after the projection matrix.
```
prideout, line 220:
```suggestion
     * @param scaling     diagonol of the scaling matrix to be applied after the projection matrix.
```

libs/gltfio/include/gltfio/FilamentAsset.h
```diff
@@ -76,6 +76,27 @@ class FilamentAsset {
...
     size_t getLightEntityCount() const noexcept;
 
+    /**
+     * Gets the list of entities in the scene representing cameras. All of these have a \c Camera
+     * component.
+     *
+     * Note about aspect ratios:
+     * gltfio always uses an aspect ratio of 1.0 when setting the projection matrix for perspective
+     * cameras. gltfio then sets the camera's scaling matrix with the aspect ratio specified in the
+     * glTF file (if present).
```
prideout, line 86: 👍

filament/include/filament/Camera.h
prideout: Wups I misspelled it too actually, should be "diagonal"

PR #2664, opened 2020-06-10
Fix applyAnimation() scale when original scale is 0

iamlkj:
Fix wrong orientation when calling applyAnimation() with node scale close to 0. Originally it'll still compute orientation when scale is 0 which lead to a wrong initial quaternion(0.5,0,0,0). The fix gives default quaternion if scale is 0

### DIFF
--- a/libs/gltfio/src/math.h
+++ b/libs/gltfio/src/math.h
@@ -79,8 +79,8 @@ inline void decomposeMatrix(const filament::math::mat4f& mat, filament::math::fl
         *rotation = clone.toQuaternion();
     }
     else {
-      // Set to idendity if close to zero
-      *rotation = quatf(1);
+        // Set to idendity if close to zero
+        *rotation = quatf(1);
     }
 }
 
--- a/libs/gltfio/src/math.h
+++ b/libs/gltfio/src/math.h
@@ -80,7 +80,7 @@ inline void decomposeMatrix(const filament::math::mat4f& mat, filament::math::fl
     }
     else {
       // Set to idendity if close to zero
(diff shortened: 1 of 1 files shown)


libs/gltfio/src/math.h
```diff
@@ -75,10 +75,13 @@ inline void decomposeMatrix(const filament::math::mat4f& mat, filament::math::fl
         clone[0] /= s.x;
         clone[1] /= s.y;
         clone[2] /= s.z;
+        // Extract rotation.
+        *rotation = clone.toQuaternion();
+    }
```
romainguy, line 80: Looks good to me, just a simple formatting issue. This should written:
```
} else {
    *rotation = // ... note the 4 spaces indentation
}
```

libs/gltfio/src/math.h
```diff
@@ -75,10 +75,13 @@ inline void decomposeMatrix(const filament::math::mat4f& mat, filament::math::fl
...
+    else {
+      // Set to idendity if close to zero
+      *rotation = filament::math::quatf(1);
```
prideout, line 83:
```suggestion
      *rotation = quatf(1);
```

iamlkj:
> Good catch, I can see how this would be a problem for animations that reveal a mesh by scaling from 0. We can merge this after the two formatting issues are addressed.

The other formatting issue we can do is the o

romainguy:
Thanks, could you make sure to sign the CLA please? We can't merge your patch until you do that.

romainguy:
I've pushed the change with formatting corrected myself. Thank you for this suggestion!

PR #2678, opened 2020-06-12
Support Metal in the iOS simulator

### LINKED ISSUE #2144: Add iOS simulator to releases
It would be great to add the simulator build to the releases. Currently it’s the only platform not provided (as far as I can tell), and it’s very useful — would be great to not have to build filament at all and just use the provided releases, but the simulator build is too valuable to give up.

Are there any issues with this I’m not seeing?

bejado:
Also add simulator support to release and continuous builds.

Fixes #2144

### DIFF
--- a/filament/backend/src/metal/MetalState.mm
+++ b/filament/backend/src/metal/MetalState.mm
@@ -119,6 +119,17 @@ id<MTLSamplerState> SamplerStateCreator::operator()(id<MTLDevice> device,
     samplerDescriptor.compareFunction =
             params.compareMode == SamplerCompareMode::NONE ?
                 MTLCompareFunctionNever : getCompareFunction(params.compareFunc);
+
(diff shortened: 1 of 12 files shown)


filament/backend/src/metal/MetalState.mm
```diff
@@ -119,6 +119,17 @@
     samplerDescriptor.compareFunction =
             params.compareMode == SamplerCompareMode::NONE ?
                 MTLCompareFunctionNever : getCompareFunction(params.compareFunc);
+
+#if defined(IOS)
+    // Older Apple devices (and the simulator) don't support setting a comparison function in
+    // MTLSamplerDescriptor.
+    // In practice, this means shadows are not supported when running in the simulator.
```
romainguy, line 126: Waaaat? This simulator seems really bad overall.

PR #2693, opened 2020-06-16
Add JNI bindings for gltfio camera

bejado:


romainguy:
This needs to be rebased on top of `main`

PR #2694, opened 2020-06-16
Check if an existing material parameter is a sampler

roxlu:
New function which returns false/true wether an existing parameter is a sampler. This can be used to dynamically set a texture when a parameter is a sampler or otherwise a vec4 color.

### DIFF
--- a/filament/include/filament/Material.h
+++ b/filament/include/filament/Material.h
@@ -211,6 +211,9 @@ public:
     //! Indicates whether a parameter of the given name exists on this material.
     bool hasParameter(const char* name) const noexcept;
 
+    //! Indicates whether an existing parameter is a sampler or not.
+    bool hasSampler(const char* name) const noexcept;
+
     /**
      * Sets the value of the given parameter on this material's default instance.
      *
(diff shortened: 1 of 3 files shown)


filament/include/filament/Material.h
```diff
@@ -211,6 +211,9 @@ class UTILS_PUBLIC Material : public FilamentAPI {
     //! Indicates whether a parameter of the given name exists on this material.
     bool hasParameter(const char* name) const noexcept;
 
+    //! Indicates whether an existing parameter is a sampler or not.
+    bool hasSampler(const char* name) const noexcept;
```
romainguy, line 215: Wouldn't it make more sense for this to be called `isSampler`? `hasParameter` above will tell you whether a parameter exists on the material. Here you want to check the nature of said parameter. Actually a better API would be `getParameterType(const char*)`, so it would work with vectors, matrices, etc.

roxlu: I agree, I will rename it to `isSampler()`. 

I'm not sure about `getParameterType()` . What would this function return as the type is stored as a union:

```c++
    struct ParameterInfo {
        //! Name of the parameter.
        const char* name;
        //! Whether the parameter is a sampler (texture).
        bool isSampler;
        union {
            //! Type of the parameter if the parameter is not a sampler.
            ParameterType type;
            //! Type of the parameter if the parameter is a sampler.
            SamplerType samplerType;
        };
        //! Size of the parameter when the parameter is an array.
        uint32_t count;
        //! Requested precision of the parameter.
        Precision precision;
    };
```

PR #2700, opened 2020-06-17
New DoF implementation

pixelflinger:
This implementation is an order of magnitude more efficient, running
in about 4ms on Pixel 4.

We're using the ring binning technique described in "Life of a Bokeh",
SIGGRAPH 2018. An issue in spirvopt makes this implementation a bit
slower than it should be.

dof


filament/src/materials/dof.mat
```diff
@@ -49,17 +44,271 @@ fragment {
...
+    // constant can be used. Here we use MAX_COC_RADIUS * MAX_COC_RADIUS instead of 1/pi,
+    // because this allows us to fit in mediump precision.
+    return (MAX_COC_RADIUS * MAX_COC_RADIUS) / (max(coc * coc, 0.5));
+}
+
+float intersection(float border, float absCoc, float mip) {
+    return saturate((absCoc - border) * pow(0.5, mip) + 0.5);
+}
+
+float rcp(float x) {
```
romainguy, line 71: Should we just move this to `common_math.fs`?

filament/src/materials/dof.mat
```diff
@@ -49,17 +44,271 @@ fragment {
...
+void accumulateBackground(inout vec4 background, inout float opacity,
+        inout float i, highp vec2 pos, float radius, float mip) {
+    float coc = textureLod(materialParams_cocFgBg, pos, mip).g;
+    vec4 s = textureLod(materialParams_background, pos, mip);
+    float o = cocToAlpha(coc);
+    float w = intersection(radius, abs(coc), mip) * isBackground(coc);
+    background += mix(background * rcp(i), s, w);
+    opacity += mix(opacity * rcp(i), o, w);
+    i += 1.0;
+    //w *= sampleWeight(coc);
```
romainguy, line 99: Remove commented code

filament/src/materials/dof.mat
```diff
@@ -49,17 +44,271 @@ fragment {
...
+void accumulateForeground(inout vec4 foreground, inout float opacity, inout float i,
+        highp vec2 pos, float radius, float mip) {
+    float coc = textureLod(materialParams_cocFgBg, pos, mip).r;
+    vec4 s = textureLod(materialParams_foreground, pos, mip);
+    float o = cocToAlpha(coc);
+    float w = intersection(radius, abs(coc), mip) * isForeground(coc);
+    foreground += mix(foreground * rcp(i), s, w);
+    opacity += mix(opacity * rcp(i), o, w);
+    i += 1.0;
+    //w *= sampleWeight(coc);
```
romainguy, line 114: Same here

filament/src/materials/dof.mat
```diff
@@ -49,17 +44,271 @@ fragment {
...
+    //i += w;
+}
+
+void accumulateForeground(inout vec4 foreground, inout float opacity, inout float i,
+        highp vec2 pos, float radius, float mip) {
+    float coc = textureLod(materialParams_cocFgBg, pos, mip).r;
+    vec4 s = textureLod(materialParams_foreground, pos, mip);
+    float o = cocToAlpha(coc);
+    float w = intersection(radius, abs(coc), mip) * isForeground(coc);
+    foreground += mix(foreground * rcp(i), s, w);
```
romainguy, line 111: `rcp(i)` is invoked twice, since it's a division, worth factoring? Same in `accumulateBackground()`

filament/src/materials/dof.mat
```diff
@@ -49,17 +44,271 @@ fragment {
...
+    // the geometry.
+    
+    float coc0 = textureLod(materialParams_cocFgBg, center + offset, mip).r;
+    float coc1 = textureLod(materialParams_cocFgBg, center - offset, mip).r;
+    float coc = min(coc0, coc1);
+    vec4 s0 = textureLod(materialParams_foreground, center + offset, mip);
+    vec4 s1 = textureLod(materialParams_foreground, center - offset, mip);
+    float o = cocToAlpha(coc);
+    float w = intersection(radius, abs(coc), mip) * isForeground(coc);
+    foreground += mix(foreground * rcp(i), s0, w);
```
romainguy, line 157: rcp twice

filament/src/materials/dof.mat
```diff
@@ -49,17 +44,271 @@ fragment {
...
+    opacity += mix(opacity * rcp(i), o, w);
+    i += 1.0;
+    foreground += mix(foreground * rcp(i), s1, w);
+    opacity += mix(opacity * rcp(i), o, w);
+    i += 1.0;
+}
+
+float getMipLevel(float kernelSize) {
+    float mip = log2(kernelSize * rcp(RING_COUNT + 0.5));
+    //mip = floor(0.5 + mip);
```
romainguy, line 167: Remove?

filament/src/materials/dof.mat
```diff
@@ -49,17 +44,271 @@ fragment {
...
+    vec4 foreground = vec4(0.0);
+    vec4 background = vec4(0.0);
+    float fgOpacity = 0.0;
+    float bgOpacity = 0.0;
+
+    // we use the full resolution pixel size because that's the unit the CoC is in
+    highp vec2 fullResPixelSize = 0.5 / vec2(textureSize(materialParams_foreground, 0));
+
+    // When using mipmaping the noise seems to hurt more than it helps
+    //float randomAngle = random(gl_FragCoord.xy) * (2.0 * PI);
```
romainguy, line 194: Remove?

filament/src/materials/dof.mat
```diff
@@ -49,17 +44,271 @@ fragment {
...
+        float kernelSize = -tiles.g;
+        float mip = getMipLevel(kernelSize);
+        vec2 uvCenter = uv + noise * kernelSize * fullResPixelSize;
+        float c = 0.0;
+
+        // the center sample is handled separatly, since it's by itself
+        accumulateForegroundFirst(foreground, fgOpacity, c, uvCenter, 0.0, mip);
+
+        for (float i = 1.0; i < RING_COUNT; i += 1.0) {
+            float radius = (kernelSize / RING_COUNT) * i;
```
romainguy, line 252: Might be worth checking that the optimizer converts / RING_COUNT into a multiply by a constant

filament/src/materials/dofCombine.mat
```diff
@@ -0,0 +1,72 @@
...
+    void postProcessVertex(inout PostProcessVertexInputs postProcess) {
+    postProcess.vertex.xy = postProcess.normalizedUV;
+}
+}
+
+fragment {
+
+#include "dofUtils.fs"
+
+void dummy(){}
```
romainguy, line 43: remove?

filament/src/materials/dofDilate.mat
```diff
@@ -0,0 +1,80 @@
...
+    void postProcessVertex(inout PostProcessVertexInputs postProcess) {
+        postProcess.vertex.xy = postProcess.normalizedUV;
+    }
+}
+
+fragment {
+
+#include "dofUtils.fs"
+
+void dummy(){}
```
romainguy, line 28: Remove?

filament/src/materials/dof.mat
pixelflinger: I thought of it but the reason I didn't do it is that I'm unclear about what happens with precision qualifiers. e.g. should it specify `highp`? but if it doesn't and it's fed a `highp` value, will it loose its `highp`ness?

filament/src/materials/dof.mat
pixelflinger: Yeah, so I don't want to remove that one just yet, as we might switch to it soon, if I figure out some issues.

filament/src/materials/dof.mat
romainguy: I mean that's true of all of our math helpers. They all assume default precision right now, we can always make a rcpHighp if needed.

filament/src/materials/dof.mat
pixelflinger: yes, I'm relying on the compiler here -- but yeah I could do that.

filament/src/materials/dof.mat
pixelflinger: It's useful to have when testing in matdbg -- since it's not trivial to remember this code, I left it there with a comment.

filament/src/materials/dof.mat
pixelflinger: yes, I checked that, at least spirvtools does it.

filament/src/materials/dofCombine.mat
pixelflinger: ah yes. I have a bunch of those, that's to squash CLion syntax highlighting, it makes it hard to work on the code without it :-\

filament/src/materials/dofTiles.mat
```diff
@@ -0,0 +1,63 @@
...
+}
+
+void postProcess(inout PostProcessInputs postProcess) {
+    highp vec2 uv = variable_vertex.xy;
+    highp vec2 size = vec2(textureSize(materialParams_cocMaxMin, 0));
+
+    ivec2 i = ivec2(uv * size - 0.5);
+
+#if defined(TARGET_VULKAN_ENVIRONMENT) || !defined(TARGET_MOBILE)
+    vec4 f = textureGather(materialParams_cocMaxMin, uv, 0); // 01, 11, 10, 00
```
prideout, line 45: Wow, TIL about `textureGather`.

filament/src/materials/dof.mat
```diff
@@ -49,17 +44,423 @@ fragment {
...
+#else
+    return 0.0;
+#endif
+}
+
 void postProcess(inout PostProcessInputs postProcess) {
+    const float ringCountGather = float(RING_COUNT_GATHER);
+    const float ringCountFast   = float(RING_COUNT_FAST);
     highp vec2 uv = variable_vertex.xy;
+    vec2 tiles = textureLod(materialParams_tiles, uv, 0.0).rg;
```
prideout, line 318: Nit: I would have called this variable `tileInfo` rather than `tiles` to better indicate that it holds CoC information for a single tile.

filament/src/materials/dofCombine.mat
```diff
@@ -0,0 +1,72 @@
...
+        vertex
+    ],
+    domain : postprocess,
+    depthWrite : false,
+    depthCulling : false
+}
+
+vertex {
+    void postProcessVertex(inout PostProcessVertexInputs postProcess) {
+    postProcess.vertex.xy = postProcess.normalizedUV;
```
romainguy, line 35: Indentation :)

oHuflying:
Can you provide an example of using DOF on Android?

PR #2706, opened 2020-06-18
Improve JS API for SurfaceOrientation helper.

### LINKED ISSUE #2702: SurfaceOrientation$Builder and the triangles16 method: TypeError: Cannot convert "1,2,3" to long
**Describe the bug**

Are TS types for SurfaceOrientation$Builder correct? `triangles16`, `triangles32` and few other methods that (probably, see code snippets below for C++ and Java) should accept buffer are described here as a `number` type:

https://github.com/google/filament/blob/94776a01b38901d9b3a7592c555f67acdbca4eab/web/filament-js/filament.d.ts#L874-L879

That would not be a problem (only IDE complains "Argument type Uint16Array is not assignable to parameter type number ") but an attempt to use it ends with an error: `TypeError: Cannot convert "1,2,3" to long`.

Edit: I'm using the Javascript version.

**To Reproduce**
```ts
const sob = new Filament.SurfaceOrientation$Builder();

sob.triangles16(new Uint16Array([1,2,3]));
sob.triangleCount(trianglesCount);

const orientation = sob.build();
```

**Expected behavior**
I would expect that those methods correspond somehow to the `SurfaceOrientation.cpp` implementation and will trigger `buildWithFlatNormals`, `buildWithNormalsOnly` or `buildWithUvs` internally and I will be able to get tangents for triangle mesh imported from an STL file.

**Screenshots**
Screenshot 2020-06-17 at 19 47 00

 - OS: Mac OS
 - GPU: Intel Iris Plus Graphics 655 1536 MB
 - Backend: OpenGL probably

**Additional context**

In the C++ version those are buffers
https://github.com/google/filament/blob/94776a01b38901d9b3a7592c555f67acdbca4eab/libs/geometry/src/SurfaceOrientation.cpp#L103-L111

Same for Java bindings:
https://github.com/google/filament/blob/94776a01b38901d9b3a7592c555f67acdbca4eab/android/filament-android/src/main/java/com/google/android/filament/SurfaceOrientation.java#L116-L126

prideout:
Web developers should not be expected to call malloc on the emscripten
heap. This new API is better aligned with how we handle vertex buffers.

Fixes #2702.

### DIFF
--- a/web/filament-js/extensions.js
+++ b/web/filament-js/extensions.js
@@ -302,7 +302,7 @@ Filament.loadClassExtensions = function() {
         buffer = new Uint8Array(buffer.buffer, buffer.byteOffset, buffer.byteLength);
         this.posPointer = Filament._malloc(buffer.byteLength);
         Filament.HEAPU8.set(buffer, this.posPointer);
-        this._uvs(this.posPointer, stride);
+        this._positions(this.posPointer, stride);
     };
 
     Filament.SurfaceOrientation$Builder.prototype.triangles16 = function(buffer, stride = 0) {
--- a/web/filament-js/extensions.js
+++ b/web/filament-js/extensions.js
@@ -284,10 +284,60 @@ Filament.loadClassExtensions = function() {
         pbd.delete();
     }
 
+    Filament.SurfaceOrientation$Builder.prototype.normals = function(buffer, stride = 0) {
+        buffer = new Uint8Array(buffer.buffer, buffer.byteOffset, buffer.byteLength);
+        this.norPointer = Filament._malloc(buffer.byteLength);
+        Filament.HEAPU8.set(buffer, this.norPointer);
+        this._normals(this.norPointer, stride);
+    };
+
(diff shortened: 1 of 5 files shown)


web/filament-js/extensions.js
```diff
@@ -284,12 +284,62 @@ Filament.loadClassExtensions = function() {
...
+        this.uvsPointer = Filament._malloc(buffer.byteLength);
+        Filament.HEAPU8.set(buffer, this.uvsPointer);
+        this._uvs(this.uvsPointer, stride);
+    };
+
+    Filament.SurfaceOrientation$Builder.prototype.positions = function(buffer, stride = 0) {
+        buffer = new Uint8Array(buffer.buffer, buffer.byteOffset, buffer.byteLength);
+        this.posPointer = Filament._malloc(buffer.byteLength);
+        Filament.HEAPU8.set(buffer, this.posPointer);
+        this._uvs(this.posPointer, stride);
```
wojciechczerniak, line 305: Copy/paste typo?

```suggestion
        this._positions(this.posPointer, stride);
```

prideout: Thanks!

PR #2709, opened 2020-06-18
Fix gltfio-java build errors on Windows

bejado:


prideout:
Looks like this fails on macOS.

Issue #2714, opened 2020-06-20
Polly glTF model does not display correctly in WebGL environment


cx20:
**Describe the bug**

I tried displaying the [Polly glTF model](https://rawcdn.githack.com/KhronosGroup/glTF-Blender-Exporter/0e23c773bf27dad67d2c25f060370d6fa012d87d/polly/project_polly.gltf) in Filament `1.7.0`.
However, after a few seconds the display goes wrong.

![filament_polly_problem](https://user-images.githubusercontent.com/1833979/85203987-41bccc80-b34c-11ea-8acd-fe9d10757dfb.gif)

**To Reproduce**
https://cx20.github.io/jsdo.it-archives/cx20/M2lx/

**Desktop (please complete the following information):**
 - OS: [Windows 10]
 - Browser : [Chrome 83]
 - GPU: [Intel(R) UHD Graphics 617]
 - Backend: [ANGLE]

prideout:
Taking a look now.  I suspect that Animator is producing NaN's during the bad frames, although I already fixed a slew of quaternion-related NaN bugs a while back.  Happens on desktop too.

cx20:
BTW, when I tried to display Polly with the latest version of Filament, I found a new artifact on Polly's head. It looks like an antenna, but I'm wondering what it is.

[Filament latest version + Polly glTF model result:](https://cx20.github.io/gltf-test/examples/filament/index.html?url=https://rawcdn.githack.com/KhronosGroup/glTF-Blender-Exporter/64017da8f49a618dbeec8ff80f6053cbcb715ad5/polly/project_polly.gltf)
![image](https://user-images.githubusercontent.com/1833979/85330831-0ec33600-b510-11ea-9f53-b1cd347c5aa8.png)

prideout:
Ha, I thought that was the antenna but yes it seems to be an artifact!  I will take a look.

Regarding the original bug, it appears that our Animator is not applying CUBICSPLINE interpolation correctly for morph weights, although we *do* apply this interpolation correctly for other types of rigid body transforms.

The glTF spec is somewhat unclear about the buffer layout for this situation, and there are no models in the Khronos test suite that exercise cubic interpolation in conjunction with morph weights.  However [this Khronos issue](https://github.com/KhronosGroup/glTF-Tutorials/issues/36) has some helpful information that I will use to make our implementation more consistent with the ecosystem.

prideout:
[...]

PR #2720, opened 2020-06-22
Fix CUBICSPLINE interpolation for morph weights.

prideout:
Tested with Polly the robo-hound and ensured that InterpolationTest
did not regress.

Partial fix for #2714.

### DIFF
--- a/libs/gltfio/src/Animator.cpp
+++ b/libs/gltfio/src/Animator.cpp
@@ -238,11 +238,11 @@ void Animator::applyAnimation(size_t animationIndex, float time) const {
         // Compute the interpolant between 0 and 1.
         float prevTime = prevIter->first;
         float nextTime = nextIter->first;
-        float interval = nextTime - prevTime;
-        if (interval < 0) {
-            interval += anim.duration;
+        float deltaTime = nextTime - prevTime;
+        if (deltaTime < 0) {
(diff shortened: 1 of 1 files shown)


libs/gltfio/src/Animator.cpp
```diff
@@ -304,27 +304,38 @@ void Animator::applyAnimation(size_t animationIndex, float time) const {
...
+                    for (int comp = 0; comp < numComponents; ++comp) {
+                        float vert0 = splineVerts[comp + prevIndex * valuesPerKeyframe];
+                        float tang0 = outTangents[comp + prevIndex * valuesPerKeyframe];
+                        float tang1 = inTangents[comp + nextIndex * valuesPerKeyframe];
+                        float vert1 = splineVerts[comp + nextIndex * valuesPerKeyframe];
+                        weights[comp] = cubicSpline(vert0, tang0, vert1, tang1, t);
+                    }
+                } else {
+                    // Filament supports up to 4 morph targets, all others are ignored.
+                    const int numComponents = std::min(4, valuesPerKeyframe);
```
romainguy, line 331: I feel we should probably expose that constant 4 somewhere in our APIs for when we incraease it.

prideout: Good idea, fixed.

PR #2725, opened 2020-06-23
gltfio: various improvements to skinning normalization.

### LINKED ISSUE #2714: Polly glTF model does not display correctly in WebGL environment

prideout:
- Allow JavaScript and Java / Kotlin clients to configure this setting.
- Opt in by default (this was disabled in #2621)
- Optimize large assets like Bistro by iterating through prims only when the asset actually has skins.

Fixes #2714.

### DIFF
--- a/libs/gltfio/src/ResourceLoader.cpp
+++ b/libs/gltfio/src/ResourceLoader.cpp
@@ -296,7 +296,7 @@ bool ResourceLoader::loadResources(FFilamentAsset* asset, bool async) {
     }
     asset->mResourcesLoaded = true;
     mPool->addAsset(asset);
-    auto gltf = (cgltf_data*) asset->mSourceAsset;
+    const cgltf_data* gltf = asset->mSourceAsset;
     cgltf_options options {};
 
     // For emscripten and Android builds we have a custom implementation of cgltf_load_buffers which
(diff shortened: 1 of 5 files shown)


libs/gltfio/src/ResourceLoader.cpp
```diff
@@ -381,13 +381,19 @@ bool ResourceLoader::loadResources(FFilamentAsset* asset, bool async) {
...
-    // places, this will needlessly repeat the work. In the future we would like to remove this
-    // feature, and instead simply require correct models. See also:
-    // https://github.com/KhronosGroup/glTF-Sample-Models/issues/215
-    if (pImpl->mNormalizeSkinningWeights) {
-        normalizeSkinningWeights(asset);
+    // Normalize skinning weights, then "import" each skin into the asset by building a mapping of
+    // skins to their affected entities.
+    if (gltf->skins_count > 0) {
+        if (pImpl->mNormalizeSkinningWeights) {
+            normalizeSkinningWeights(asset);
```
romainguy, line 388: BTW if this causes performance issues we should be able to use the job system and write the code in the way that will produce vectorized code.

prideout: Yes this would be easy to do since we already use jobs for bounding boxes and tangent gen.  Something for the future.

PR #2729, opened 2020-06-23
Remove use of deprecated APIs in libfilamentapp

bejado:


filament/src/details/View.h
```diff
@@ -17,6 +17,12 @@
 #ifndef TNT_FILAMENT_DETAILS_VIEW_H
 #define TNT_FILAMENT_DETAILS_VIEW_H
 
+// Avoid warnings for using the ToneMapping API, which has been publically deprecated.
+#if defined(__clang__)
```
pixelflinger, line 21: This will not fix visual studio right?

bejado: Well, this warning silences warnings due to using APIs marked `__attribute(deprecated)`, which I don't think MSVC has.

PR #2733, opened 2020-06-24
Vulkan: fix bloom on Android by supporting LOAD_OP_LOAD.

prideout:
If a render target attachment is not cleared or discarded, then it should be preserved and therefore requires VK_ATTACHMENT_LOAD_OP_LOAD.

### DIFF
--- a/filament/backend/src/vulkan/VulkanFboCache.cpp
+++ b/filament/backend/src/vulkan/VulkanFboCache.cpp
@@ -24,8 +24,8 @@ namespace backend {
 bool VulkanFboCache::RenderPassEq::operator()(const RenderPassKey& k1,
         const RenderPassKey& k2) const {
     return
-            k1.finalColorLayout == k2.finalColorLayout &&
-            k1.finalDepthLayout == k2.finalDepthLayout &&
+            k1.colorLayout == k2.colorLayout &&
+            k1.depthLayout == k2.depthLayout &&
             k1.colorFormat == k2.colorFormat &&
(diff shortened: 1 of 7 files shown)


filament/backend/src/vulkan/VulkanFboCache.cpp
```diff
@@ -104,26 +111,46 @@ VkRenderPass VulkanFboCache::getRenderPass(RenderPassKey config) noexcept {
         .pDepthStencilAttachment = hasDepth ? &depthAttachmentRef : nullptr
     };
 
-    // The attachment description specifies the layout to transition to at the END of the render pass.
+    // Set up some const aliases for terseness.
+    const VkAttachmentLoadOp clear = VK_ATTACHMENT_LOAD_OP_CLEAR;
+    const VkAttachmentLoadOp dontCare = VK_ATTACHMENT_LOAD_OP_DONT_CARE;
+
+    // NOTE: We do not support LOAD_OP_LOAD for the swap chain. This is rarely what you want anyway,
```
pixelflinger, line 118: Wouldn't that break when rendering several Views?

prideout: That's an excellent point, multiple views seem to work on Molten but I bet they would break on Android for this reason.

For now, let me change the NOTE into TODO, then work on this after merging.

pixelflinger: And we have a nice test for this, btw.

PR #2741, opened 2020-06-26
Improve matc error reporting

bejado:
When `matc` reports an error, it will now include the name of the material file and the correct line number. For example:

```
ERROR: sandboxUnlit.mat:20: 'bafeColor' : undeclared identifier
ERROR: sandboxUnlit.mat:20: '' : compilation terminated
```

This works just as well with files `#include`d by a material:

```
ERROR: Unable to parse fragment shader:
ERROR: dofUtils.fs:30: 'assign' :  cannot convert from ' const 3-component vector of float' to ' temp 4-component vector of float'
ERROR: dofUtils.fs:30: '' : compilation terminated
```

and even with errors in Filament shaders in debug builds:

```
ERROR: shadowing.fs:323: 'ShadowSample_PCF_Low' : no matching overloaded function found
ERROR: shadowing.fs:323: '' : missing #endif
ERROR: shadowing.fs:323: '' : compilation terminated
ERROR: 3 compilation errors.  No code generated.
```

### DIFF
--- a/libs/filamat/include/filamat/IncludeCallback.h
+++ b/libs/filamat/include/filamat/IncludeCallback.h
@@ -24,34 +24,45 @@
 namespace filamat {
 
 struct IncludeResult {
-    /**
-     * The full contents of the include file. This may contain additional, recursive include
-     * directives.
-     */
-    utils::CString source;
+    // The line number for the first line of source (first line is 0).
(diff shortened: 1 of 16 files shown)


romainguy:
I could have used this 10 minutes ago 🤣

PR #2743, opened 2020-06-27
Add vignette post-processing effect

romainguy:
Screen Shot 2020-06-26 at 4 49 55 PM

Screen Shot 2020-06-26 at 4 53 21 PM

### DIFF
--- a/filament/src/PostProcessManager.cpp
+++ b/filament/src/PostProcessManager.cpp
@@ -1202,17 +1202,50 @@ FrameGraphId<FrameGraphTexture> PostProcessManager::bloomPass(FrameGraph& fg,
     return bloomUpsamplePass.getData().out;
 }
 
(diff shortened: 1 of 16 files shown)


filament/src/PostProcessManager.cpp
```diff
@@ -1202,17 +1202,50 @@ FrameGraphId<FrameGraphTexture> PostProcessManager::bloomPass(FrameGraph& fg,
...
+                mix(1.0f + 4.0f * (1.0f - options.feather), 1.0f, std::sqrtf(oval));
+
+        // Factor to transform oval into circle
+        float aspect = mix(1.0f, float(width) / float(height), circle);
+
+        return float4{midPoint, radius, aspect, options.feather};
+    }
+
+    // Set half-max to show disabled
+    return float4{65504.0f};
```
pixelflinger, line 1230: I _think_ you can actually use `std::numeric_limit<half>::max()`

PR #2754, opened 2020-06-30
filamat: Resolve string processing inefficiency

cxyzs7:
`std::string` doesn't have a `string(const char*, size_t pos, size_t len)` constructor: http://www.cplusplus.com/reference/string/string/string/ (Or at least not with MSVC compiler). Instead, it will use `string (const char* s)` [null-terminated character sequence (C-string) pointed by s], which scans the whole sequence to create a string first and use `string (const string& str, size_t pos, size_t len)` to get the substring. This is very inefficient as it scans the full shader text for each line and runs very in Debug mode with Visual Studio 2019.


bejado:
LGTM

Issue #2752, opened 2020-06-29
Vulkan: incorrect output from Depth Mipmap Pass


prideout:
When `sao.mat` samples any miplevel other than the base level in `materialParams_depth`, it always comes back zero.  This is likely why Vulkan and OpenGL produce visually inconsistent SSAO, especially when radius is large.

Mip 0 looks fine.

Note: this occurs with MoltenVK and Android.

prideout:
Ah, this issue is caused by the fact that VulkanDriver currently removes the fragment shader from depth-only passes as a workaround for a problem we encountered in the old days when we had a depth prepass.

PR #2757, opened 2020-06-30
Vulkan: add diagnostics to check if format is sampleable.

prideout:
Motivated by #2752.

### DIFF
--- a/filament/backend/src/vulkan/VulkanDriver.cpp
+++ b/filament/backend/src/vulkan/VulkanDriver.cpp
@@ -163,9 +163,29 @@ VulkanDriver::VulkanDriver(VulkanPlatform* platform,
 
     // Choose a depth format that meets our requirements. Take care not to include stencil formats
     // just yet, since that would require a corollary change to the "aspect" flags for the VkImage.
-    mContext.depthFormat = findSupportedFormat(mContext,
+    mContext.finalDepthFormat = findSupportedFormat(mContext,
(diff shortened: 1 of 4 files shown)


filament/backend/src/vulkan/VulkanDriver.cpp
```diff
@@ -163,9 +163,29 @@ VulkanDriver::VulkanDriver(VulkanPlatform* platform,
...
     // Choose a depth format that meets our requirements. Take care not to include stencil formats
     // just yet, since that would require a corollary change to the "aspect" flags for the VkImage.
-    mContext.depthFormat = findSupportedFormat(mContext,
+    mContext.finalDepthFormat = findSupportedFormat(mContext,
         { VK_FORMAT_D32_SFLOAT, VK_FORMAT_X8_D24_UNORM_PACK32 },
         VK_IMAGE_TILING_OPTIMAL, VK_FORMAT_FEATURE_DEPTH_STENCIL_ATTACHMENT_BIT);
+
+    // For diagnostic purposes, print useful information about available depth formats.
+    // Note that Vulkan is more constrained than OpenGL ES 3.1 in this area.
+    #if VK_ENABLE_VALIDATION
```
bejado, line 172: Left-align preprocessor directives? Same for VulkanHandles.cpp

prideout: Oh right.  Done.

PR #2760, opened 2020-06-30
Add matc support for command-line preprocessor defines

### LINKED ISSUE #2699: Is there a way to #define when calling `matc`?
GLSL supports C-style preprocessor directives but is there a way to set defines when calling `matc` like `gcc -D` does?

bejado:
Fixes #2699

### DIFF
--- a/tools/matc/src/matc/CommandlineConfig.cpp
+++ b/tools/matc/src/matc/CommandlineConfig.cpp
@@ -55,6 +55,10 @@ static void usage(char* name) {
             "       Specify the target API: opengl (default), vulkan, metal, or all\n"
             "       This flag can be repeated to individually select APIs for inclusion:\n"
             "           MATC --api opengl --api metal ...\n\n"
+            "   --define, -D\n"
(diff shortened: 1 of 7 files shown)


tools/matc/src/matc/CommandlineConfig.cpp
```diff
@@ -115,8 +119,35 @@ CommandlineConfig::CommandlineConfig(int argc, char** argv) : Config(), mArgc(ar
...
+            return;
+        }
+        std::string def(defineArg, p - defineArg);
+        defines.emplace(def, p + 1);
+        return;
+    }
+
+    // No explicit assignment, use a default value of 1.
+    std::string def(defineArg, p - defineArg);
+    defines.emplace(def, "1");
```
romainguy, line 146: We could just use empty string as well. That's enough for #ifdef or #if defined(…) to work.

bejado: True, this follows Clang's convention of setting it explicitly to 1.

PR #2766, opened 2020-07-01
Add error message when parsing material with missing backend

bejado:


filament/src/Engine.cpp
```diff
@@ -415,23 +415,7 @@ int FEngine::loop() {
...
-                backend = "Vulkan";
-                break;
-            case backend::Backend::METAL:
-                backend = "Metal";
-                break;
-            default:
-                backend = "Unknown";
-                break;
-        }
+        backend = backendToString(mBackend);
```
prideout, line 418: Maybe combine these codelines

PR #2771, opened 2020-07-03
filament.d.ts make init, fetch callbacks optional

MingweiSamuel:


romainguy:
Thanks for your contribution! This looks good to me but we'll wait for @prideout's review :)

PR #2778, opened 2020-07-08
Minor fixes for "matinfo --web-server"

prideout:


prideout:
I recently used `matinfo --web-server 8081` in order to inspect the final GLSL for materials that are not always loaded by the Engine, such as colorGradingAsSubpass.

PR #2783, opened 2020-07-08
Vulkan: add support for input attachment descriptors.

prideout:
This is step 1 for supporting subpasses (aka framebuffer fetch).

Vulkan requires us to explicitly bind the shader to any framebuffers
that we read from. These are called input attachments in Vulkan
parlance, and each binding is called a descriptor.

For simplicity we only create homogeneous descriptor sets where each
descriptor set is comprised of only one type of binding (i.e. all
uniforms or all samplers). I see no reason to break with that
architecture right now, so input attachments live in their own set.
In the long term, we should probably think about partitioning
descriptors differently (e.g. according to their frequency of change).

This PR also renames some internal fields in order to avoid nomenclature
confusion, such as `mDescriptorSets` => `mDescriptorBundles` . Each
entry in this map is a set of descriptor sets, or what we are now
calling a "bundle".

### DIFF
--- a/filament/backend/src/vulkan/VulkanBinder.cpp
+++ b/filament/backend/src/vulkan/VulkanBinder.cpp
@@ -37,8 +37,8 @@ static constexpr VkAllocationCallbacks* VKALLOC = nullptr;
 // TODO: Make VulkanBinder robust against a large number of descriptors.
 // There are several approaches we could take to deal with too many descriptors:
 // - Create another pool after hitting the limit of the first pool.
-// - Allow up to 3 uniform descriptors to bind simultaneously instead of just 1.
-static constexpr uint32_t MAX_NUM_DESCRIPTORS = 1000;
+// - Allow several uniform buffer descriptors to bind simultaneously instead of just one.
+static constexpr uint32_t MAX_DESCRIPTOR_SET_COUNT = 1500;
 
 static VulkanBinder::RasterState createDefaultRasterState();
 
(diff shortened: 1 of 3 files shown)


filament/backend/src/vulkan/VulkanBinder.cpp
```diff
@@ -37,8 +37,8 @@ static constexpr VkAllocationCallbacks* VKALLOC = nullptr;
 // TODO: Make VulkanBinder robust against a large number of descriptors.
 // There are several approaches we could take to deal with too many descriptors:
 // - Create another pool after hitting the limit of the first pool.
-// - Allow up to 3 uniform descriptors to bind simultaneously instead of just 1.
-static constexpr uint32_t MAX_NUM_DESCRIPTORS = 1000;
+// - Allow several uniform buffer descriptors to bind simultaneously instead of just one.
+static constexpr uint32_t MAX_DESCRIPTOR_SET_COUNT = 1500;
```
bejado, line 41: Just curious, how many descriptors do we typically generate? I'm surprised we get anywhere near to 1000

prideout: We go over 1000 now for some scenes because we never mutate descriptor sets; we create one for each unique combination of "bindings" across a given frame, where a binding is a single uniform buffer or sampler.  This strategy makes life easy because it avoids synchronization problems.

Since we sometimes run into the limit, we should consider switching to the strategy that is used in bgfx, which has one "uber" uniform buffer and doesn't need many descriptor sets.

bejado: Cool, I'm curious because Metal has something similar called "argument buffers" which we're not currently using, but I'm considering switching over to in the future

PR #2794, opened 2020-07-10
Prevent crash if the glTF doesn't have animations

romainguy:


prideout:
Thanks.  This prevents a crash when the client never calls getAnimator.

PR #2795, opened 2020-07-10
Add HelloCocoaPods iOS sample

bejado:


ios/samples/HelloCocoaPods/HelloCocoaPods/ViewController.m
```diff
@@ -0,0 +1,233 @@
...
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+
+#import "ViewController.h"
+
+#import <MetalKit/MTKView.h>
+
+// These are all C++ headers, so make sure the type of this file is Objective-C++ source.
```
prideout, line 21: Should this file have an `.mm` extension?

bejado: Changing the extension to `.mm` didn't seem to trigger Xcode into recognizing it in Objective-C++, so I had to explicitly change the file type. Doesn't hurt to do both, I suppose.

PR #2804, opened 2020-07-13
NoopDriver now uses ShaderModel::GL_ES_30 on WASM

lincolnfrog:
This allows us to do WASM-based unit tests using the NOOP driver.

### DIFF
--- a/filament/backend/src/noop/NoopDriver.cpp
+++ b/filament/backend/src/noop/NoopDriver.cpp
@@ -31,7 +31,7 @@ NoopDriver::NoopDriver() noexcept : DriverBase(new ConcreteDispatcher<NoopDriver
 NoopDriver::~NoopDriver() noexcept = default;
 
 backend::ShaderModel NoopDriver::getShaderModel() const noexcept {
-#if defined(ANDROID) || defined(IOS)
+#if defined(ANDROID) || defined(IOS) || defined(__EMSCRIPTEN__)
     return ShaderModel::GL_ES_30;
 #else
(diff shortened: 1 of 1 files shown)


filament/backend/src/noop/NoopDriver.cpp
```diff
@@ -31,7 +31,7 @@ NoopDriver::NoopDriver() noexcept : DriverBase(new ConcreteDispatcher<NoopDriver
 NoopDriver::~NoopDriver() noexcept = default;
 
 backend::ShaderModel NoopDriver::getShaderModel() const noexcept {
-#if defined(ANDROID) || defined(IOS)
+#if defined(ANDROID) || defined(IOS) || defined(EMSCRIPTEN)
```
romainguy, line 34: Everywhere else we use:
`#if defined(__EMSCRIPTEN__)`

@prideout can you confirm?

prideout: Yes good catch Romain, let's be consistent with the rest of Filament.

lincolnfrog: that's how I had it originally but I figured we should match ANDROID and IOS. It does work FWIW.

prideout: The underscored version is more correct as it allows for emscripten's "strict" mode to be enabled, and is consistent with the rest of Filament.

lincolnfrog: Done.

PR #2806, opened 2020-07-13
CMake: specify -fPIC at root level.

### LINKED ISSUE #1881: Linking with filament: -fPIE/-fPIC issue
I've just downloaded the prebuilt libraries for linux and tried to link them in my project which gave me this error: 

```
/usr/bin/ld: filament-prefix/src/filament/lib/x86_64//libimage.a(KtxBundle.cpp.o): 
relocation R_X86_64_32 against `.rodata.str1.1' can not be used when making a PIE object; 
recompile with -fPIE
```

If I understand that error correctly, it's telling me to recompile libimage.a with -fPIE. I had a discussion with someone who explained me -fPIE is an alias for -fPIC. After some reading up I found a [StackOverflow post with a command that allows you to check if an object/library was compiled with -fPIC](https://stackoverflow.com/questions/1340402/how-can-i-tell-with-something-like-objdump-if-an-object-file-has-been-built-wi). It seems libimage.a was compiled with -fPIC which makes me wonder why I get that error (especially if -fPIE is an alias for -fPIC). 

I did a couple of tests, enabling/disabling POSITION_INDEPENDENT_CODE in my CMakelists.txt but nothing solved the issue. 

I was able to compile and link filament by using the filament repository as an external project and building it from source:

```cmake
ExternalProject_Add(
    filament
    GIT_REPOSITORY https://github.com/google/filament.git
    CMAKE_ARGS -DCMAKE_INSTALL_PREFIX=<INSTALL_DIR>
)
``` 

I'm trying to figure out why I can't use the prebuilt libraries? Any thoughts? Maybe it's because of compiler defaults? (Using clang 9)

prideout:
Fixes #1881.


libs/ibl/CMakeLists.txt
```diff
@@ -43,8 +43,6 @@ target_link_libraries(${TARGET} math utils)
 # ==================================================================================================
 if (MSVC)
     target_compile_options(${TARGET} PRIVATE /fp:fast)
-else()
-    target_compile_options(${TARGET} PRIVATE -ffast-math $<$<PLATFORM_ID:Linux>:-fPIC>)
```
romainguy, line 47: You lost `-ffast-math`, and we really need it

prideout: Fixed.  I should probably move fast math to the global level too.

Issue #2780, opened 2020-07-08
colorGradingSubpass uploads uniforms inside render pass


prideout:
It's okay to call `mi->use()` but it's not okay to call `mi->commit()`.  Because Vulkan.

PR #2811, opened 2020-07-14
Add simple subpass support to the Driver API.

prideout:
This makes two changes to the Driver API:
1. Adds the beginSubpass() method.
2. Adds subpassMask to RenderPassParams.

Locally I've proven that this PR is sufficient by making a series of Vulkan changes; sneak peek is available at commit bfef6914a7d288bbd25926c02ec93d231a8c0d07.

Note that this PR also moves the new usage flag to a different texture because I had placed it on the wrong one.

See also #2780.

### DIFF
--- a/filament/backend/include/backend/DriverEnums.h
+++ b/filament/backend/include/backend/DriverEnums.h
@@ -838,8 +838,15 @@ struct RenderPassParams {
     //! Stencil value to clear the stencil buffer with
     uint32_t clearStencil = 0;
 
-    //! reserved, must be zero
-    uint32_t reserved1 = 0;
+    /**
+     * The subpass mask specifies which color attachments are designated for read-back in the second
+     * subpass. If this is zero, the render pass has only one subpass. The least significant bit
+     * specifies that the first color attachment in the render target is a subpass input.
+     *
+     * For now only 2 subpasses are supported, so only the lower 4 bits are used, one for each color
+     * attachment (see MRT::TARGET_COUNT).
+     */
+    uint32_t subpassMask = 0;
 };
 
 struct PolygonOffset {
--- a/filament/backend/src/opengl/OpenGLDriver.cpp
+++ b/filament/backend/src/opengl/OpenGLDriver.cpp
@@ -2192,6 +2192,9 @@ void OpenGLDriver::endRenderPass(int) {
 }
 
 
+void OpenGLDriver::nextSubpass(int) {}
+
+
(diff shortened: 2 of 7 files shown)


filament/backend/include/backend/DriverEnums.h
```diff
@@ -838,8 +838,8 @@ struct RenderPassParams {
     //! Stencil value to clear the stencil buffer with
     uint32_t clearStencil = 0;
 
-    //! reserved, must be zero
-    uint32_t reserved1 = 0;
+    //! If non-zero, specifies which color attachments are designated for read-back in the subpass.
+    uint32_t subpassMask = 0;
```
bejado, line 842: How does this field work, is it a bitmask? Maybe add an explanatory comment

prideout: Added a second commit with an explanatory comment.

filament/src/Renderer.cpp
```diff
@@ -695,11 +697,15 @@ FrameGraphId<FrameGraphTexture> FRenderer::colorPass(FrameGraph& fg, const char*
 
                 out.params.clearColor = data.clearColor;
 
-                driver.beginRenderPass(out.target, out.params);
-                pass.executeCommands(resources.getPassName());
-
                 if (colorGradingConfig.asSubpass) {
-                    // post-processing....
+                    // TODO: update params for the subpass material here (before beginRenderPass)
```
pixelflinger, line 701: it's not a "TODO" anymore right?

filament/backend/src/opengl/OpenGLDriver.cpp
```diff
@@ -2192,6 +2192,9 @@ void OpenGLDriver::endRenderPass(int) {
 }
 
 
+void OpenGLDriver::beginSubpass(int) {}
```
pixelflinger, line 2195: so this is were glMemoryBarrier would go if we were not using the coherent extension right?

prideout: Yes, I think so, although Vulkan is the motivation of course. I'm using it both to call vkCmdNextSubpass and to bind the input attachment to a descriptor.   Do you like the name?  Maybe I should just be "nextSubpass" since there's no corresponding end.

pixelflinger: I agree, I think `nextSubpass` is better, especially that in the future we might support more.

Issue #2812, opened 2020-07-14
Misalignment in Material parameter type and Sampler type


israelillan-verizonmedia:
**Describe the bug**
in filament/backend/include/backend/DriverEnums.h the c++ enum type for textures samplers is defined as follows:
```
//! Texture sampler type
enum class SamplerType : uint8_t {
    SAMPLER_2D,         //!< 2D texture
    SAMPLER_2D_ARRAY,   //!< 2D array texture
    SAMPLER_CUBEMAP,    //!< Cube map texture
    SAMPLER_EXTERNAL,   //!< External texture
};
```
This struct is translated to a java enum type which represents all types of material parameters (including samplers and no samplers), which is defined in android/filament-android/src/main/java/com/google/android/filament/Material.java like this:
```
        public enum Type {
            BOOL,
            BOOL2,
            BOOL3,
            BOOL4,
            FLOAT,
            FLOAT2,
            FLOAT3,
            FLOAT4,
            INT,
            INT2,
            INT3,
            INT4,
            UINT,
            UINT2,
            UINT3,
            UINT4,
            MAT3,
            MAT4,
            SAMPLER_2D,
            SAMPLER_CUBEMAP,
            SAMPLER_EXTERNAL
        }
```

there is one sampler missing: SAMPLER_2D_ARRAY, so when the java side receives 21 as type, it crashes because the enum has only 20 elements. The solution is just adding the missing type:
```
        public enum Type {
            BOOL,
            BOOL2,
            BOOL3,
            BOOL4,
            FLOAT,
            FLOAT2,
            FLOAT3,
            FLOAT4,
            INT,
            INT2,
            INT3,
            INT4,
            UINT,
            UINT2,
            UINT3,
            UINT4,
            MAT3,
            MAT4,
            SAMPLER_2D,
            SAMPLER_2D_ARRAY,  // <- missing type
            SAMPLER_CUBEMAP,
            SAMPLER_EXTERNAL
        }
```

**To Reproduce**
Just try to build a model containing a material with an external texture sampler using modelrenderable.builder

PR #2814, opened 2020-07-14
Fix material parameters enum

romainguy:
Fixes bug #2812.


romainguy:
We may have to cut another release to publish this fix. This will prevent developers from using external samplers from Java it seems.

PR #2817, opened 2020-07-14
Vulkan: add support for subpasses.

prideout:


filament/backend/src/vulkan/VulkanBinder.cpp
```diff
@@ -352,10 +353,11 @@ void VulkanBinder::bindRasterState(const RasterState& rasterState) noexcept {
     }
 }
 
-void VulkanBinder::bindRenderPass(VkRenderPass renderPass) noexcept {
-    if (mPipelineKey.renderPass != renderPass) {
+void VulkanBinder::bindRenderPass(VkRenderPass renderPass, int subpass) noexcept {
```
pixelflinger, line 356: why is it not a bool (or enum)?

filament/backend/src/vulkan/VulkanBinder.h
```diff
@@ -170,7 +170,8 @@ class VulkanBinder {
         VkShaderModule shaders[SHADER_MODULE_COUNT]; // 8*2 bytes
         RasterState rasterState; // 248 bytes
         VkRenderPass renderPass; // 8 bytes
-        VkPrimitiveTopology topology; // 4 bytes
+        VkPrimitiveTopology topology : 16; // 2 bytes
+        uint16_t subpass; // 2 bytes
```
pixelflinger, line 174: is this the same "subpass" value than above?  Maybe it should be its own typedef then?

filament/backend/src/vulkan/VulkanDriver.cpp
```diff
@@ -989,9 +994,38 @@ void VulkanDriver::endRenderPass(int) {
...
+        mContext.currentRenderPass.currentSubpass = 0;
+    }
     mContext.currentRenderPass.renderPass = VK_NULL_HANDLE;
 }
 
+void VulkanDriver::nextSubpass(int) {
+    assert(mContext.currentCommands);
+    assert(mContext.currentSurface);
+    assert(mCurrentRenderTarget);
+    assert(mContext.currentRenderPass.subpassMask);
```
pixelflinger, line 1010: should we also assert() mContext.currentRenderPass.currentSubpass ? It's limited to 1 right now, right?

filament/backend/src/vulkan/VulkanBinder.cpp
prideout: I'm trying to be forward looking for our future, it's basically a zero-based index.  Let me rename it to `subpassIndex`.

filament/backend/src/vulkan/VulkanBinder.h
prideout: Let me rename this to `subpassIndex` too, that's really what it is.

filament/backend/src/vulkan/VulkanDriver.cpp
prideout: Yes, currently I would expect `currentSubpass` to always be zero, but in the future we might allow several subpasses.

pixelflinger: I'm just saying if we do several calls to nextSubpass() in filament (by accident), it would be good to be caught here (and also make sure that it doesn't cause out-of-bounds accesses (even in release)).

prideout: Understood, amended the commit -- since you mentioned release, I used `ASSERT_PRECONDITION` for this instead of `assert`.

PR #2823, opened 2020-07-16
Add outputs block to material definition

bejado:
This adds a new `outputs` block to material definitions which allows specifying post process output variables. Support for framebuffer_fetch will be added later.


prideout:
🥳

shaders/src/post_process_inputs.fs
```diff
@@ -1,12 +1,24 @@
...
-LAYOUT_LOCATION(0) out vec4 fragColor;
-
 /** @public-api */
 highp vec2 getUV() {
     return vertex_uv;
 }
 
 struct PostProcessInputs {
-    vec4 color;
+#if defined(OUTPUT0)
```
romainguy, line 9: Is it possible to not define any output? Do we error out somewhere in that case (at compilation time)?

libs/filamat/src/shaders/CodeGenerator.cpp
```diff
@@ -215,6 +215,30 @@ io::sstream& CodeGenerator::generateShaderInputs(io::sstream& out, ShaderType ty
...
+    // TODO: add and support additional variable qualifiers
+    (void) qualifier;
+    assert(qualifier == MaterialBuilder::VariableQualifier::OUT);
+
+    const char* typeString = getOutputTypeName(outputType);
+
+    out << "\n#define OUTPUT" << index << " " << name.c_str() << "\n";
+    out << "\n#define OUTPUT_AT" << index << " output_" << name.c_str() << "\n";
+    out << "\n#define OUTPUT_TYPE" << index << " " << typeString << "\n";
+    out << "LAYOUT_LOCATION(" << index << ") out " << typeString <<
```
pixelflinger, line 235: Maybe rename those to "FRAGMENT_OUTPUT_" or "FRAG_OUTPUT_,  "OUTPUT" seems too generic to me, for a #define.

shaders/src/post_process_inputs.fs
bejado: `MaterialBuilder` will add a default color output if no other ones are specified. Otherwise yes, we would error at shader compilation time (structs can't be empty in GLSL)

shaders/src/post_process.fs
```diff
@@ -9,5 +8,19 @@ void main() {
     inputs.color = clamp(inputs.color, -MEDIUMP_FLT_MAX, MEDIUMP_FLT_MAX);
```
pixelflinger: don't we need to remove this line?
Also I'm wondering if all the color outputs need to be clamped to +/-MEDIUMP_FLT_MAX on mobile?

shaders/src/post_process_inputs.fs
romainguy: 👍

shaders/src/post_process.fs
romainguy: Ben already made that change, see the latest commit

bejado: Yeah this caused a build issue on mobile. My fix is to clamp all of the color outputs. Is there ever a case where we would have a `highp` color output on mobile?

pixelflinger: I think we can cross that bridge later. We might indeed need a precision qualifier in theory.

PR #2828, opened 2020-07-17
DoF: fix many quality/performance issues

pixelflinger:
- make sure all downsampling is exactly correct and use proper
  uv values to access these textures. This stabilizes the DoF a bit.

- for now only use NEAREST filtering, as bilinear can't be used naively.

- noise is now mandatory, fix noise radius

- add a median pass (performance is not impacted because DoF is faster
  due to using NEAREST filtering)

- fix  be consistant in how the kernel radius is computed, fix a couple
  errors in that area.

- change ring density from 8 to 4 (but starting with 8), so each new
  ring is 4 samples larger than the previous. This reduces the number
  of samples needed from 25 to 21 on mobile, and 49 to 37 on desktop. 
  It also makes it conceivable to use 5 rings (57 samples). Quality is
  not visibly impacted because the center ring is still 8 samples.
  This density, matches the number of samples required for "square"
  rings.

- fix scaling issue with diaphragm simulation

- fix typos in background accumulation code -- this fixes a lot of 
  quality issues.

- improve performance of "fast tiles" during both DoF and combine passes

Overall the effect runs about 0.5ms faster and quality is greatly
improved.

### DIFF
--- a/filament/src/materials/dof.mat
+++ b/filament/src/materials/dof.mat
@@ -23,7 +23,12 @@ material {
         },
         {
            type : float2,
-           name : fullResPixelSize,
+           name : cocToTexelOffset,
+           precision: high
+        },
+        {
+           type : float4,
+           name : uvscale,
            precision: high
         },
         {
@@ -41,7 +46,7 @@ material {
 
 vertex {
     void postProcessVertex(inout PostProcessVertexInputs postProcess) {
-        postProcess.vertex.xy = postProcess.normalizedUV;
+        postProcess.vertex = postProcess.normalizedUV.xyxy * materialParams.uvscale;
     }
 }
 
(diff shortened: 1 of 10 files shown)


filament/src/materials/dof.mat
```diff
@@ -445,32 +469,31 @@ void postProcess(inout PostProcessInputs postProcess) {
             foreground *= rcp(c);
             fgOpacity  *= rcp(sampleCount(ringCountGather));
         }
+        //foreground.r += 0.1;
```
romainguy, line 472: Is this supposed to be commented? Should it be removed?

pixelflinger: There are 3 of those, I use them for debugging with matdbg. It's basically you're first line of defense when something goes wrong, as it tells you in which type of tile the problem lies.

romainguy: You should leave an explanation so we don't get rid of them or stare at them wondering later on :)

PR #2830, opened 2020-07-18
Filament.d.ts: add Vector<T> interface, vectorToArray<T> fn, fix getResourceUris() return vec

MingweiSamuel:
Not really sure if Vector&lt;T&gt; should be an interface or a type alias, probably doesn't matter.

https://github.com/google/filament/blob/c6d9d4a9aec0ebbdd17c28eca3dc6c587d4a64c1/web/filament-js/extensions.js#L29-L35

https://github.com/google/filament/blob/c6d9d4a9aec0ebbdd17c28eca3dc6c587d4a64c1/web/filament-js/jsbindings.cpp#L1537-L1544

### DIFF
--- a/web/filament-js/filament.d.ts
+++ b/web/filament-js/filament.d.ts
@@ -46,6 +46,14 @@ export type mat3 = glm.mat3|number[];
 export type mat4 = glm.mat4|number[];
 export type quat = glm.quat|number[];
 
+/** A C++ std::vector. */
+export interface Vector<T> {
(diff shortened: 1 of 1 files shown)


MingweiSamuel:
Also not sure about versioning buy removing EntityVector and MaterialInstanceVector is a breaking change, could add back type aliases for them

prideout:
This is great, thank you for making these improvements to our TypeScript bindings.  :)

PR #2835, opened 2020-07-20
CSM: allow setting individual split positions

bejado:


filament/include/filament/LightManager.h
```diff
@@ -283,6 +314,75 @@ class UTILS_PUBLIC LightManager : public FilamentAPI {
...
+         * @param cascades          the number of shadow cascades
+         * @param near              the camera near plane
+         * @param far               the camera far plane
+         */
+        static void computeLogSplits(float splitPositions[3], uint8_t cascades,
+                float near, float far) {
+            size_t s = 0;
+            for (size_t c = 1; c < cascades; c++) {
+                splitPositions[s++] =
+                    (near * std::powf(far / near, (float) c / cascades) - near) / (far - near);
```
romainguy, line 350: We shouldn't use the STL in our public headers

filament/include/filament/LightManager.h
```diff
@@ -283,6 +314,75 @@ class UTILS_PUBLIC LightManager : public FilamentAPI {
...
+         * split scheme.
+         *
+         * @warning This API is still experimental and subject to change.
+         *
+         * @param splitPositions    a float array to write the split positions into
+         * @param cascades          the number of shadow cascades
+         * @param near              the camera near plane
+         * @param far               the camera far plane
+         */
+        static void computeLogSplits(float splitPositions[3], uint8_t cascades,
```
romainguy, line 345: Do we really want to always take an array of 3? It could just be a float* with the contract that we will write up to 3 cascades. Note that your implementation doesn't clamp the `cascades` parameter so you can write out of bounds anyway.

pixelflinger: Actually, I think we shouldn't have any code in public headers -- because you can never change it.

The only code that is OK in public headers is templates (sometimes), and helpers -- i.e. things that the user could do themselves.

bejado: I had this in the public header because I viewed these functions as "helpers," but the inability to change them + the use of an `std::` function has convinced me to make the implementations private.

PR #2840, opened 2020-07-21
gltfio: Add support for KHR_materials_transmission.

prideout:
This chooses refraction type=THIN, mode=SCREENSPACE, and blending=TRANSPARENT.

Due to sampler overload, we print a friendly error message and enable a
fallback if the user tries to simultaneously enable ubershader mode, the
clearcoat extension, and the transmission extension.

Screen Shot 2020-07-21 at 11 37 52 AM


libs/gltfio/src/AssetLoader.cpp
```diff
@@ -883,36 +884,36 @@ MaterialInstance* FAssetLoader::createMaterialInstance(const cgltf_material* inp
...
+        ccConfig.clearcoat_normal_texture.has_transform ||
+        trConfig.transmission_texture.has_transform;
 
     cgltf_texture_view baseColorTexture = mrConfig.base_color_texture;
     cgltf_texture_view metallicRoughnessTexture = mrConfig.metallic_roughness_texture;
 
     MaterialKey matkey {
-        .doubleSided = (bool) inputMat->double_sided,
-        .unlit = (bool) inputMat->unlit,
+        .doubleSided = !!inputMat->double_sided,
```
pixelflinger, line 894: the `bool` operator :-)

PR #2841, opened 2020-07-22
Introduce removeEntities API / fix leak in ModelViewer (!)

prideout:
For testing purposes, sample-gltf-viewer reloads the model when the
user double-taps the screen. This operation was causing a buildup of
blank entities in the Scene, because we were merely destroying the
components, not removing them from the Scene.

### DIFF
--- a/filament/include/filament/Scene.h
+++ b/filament/include/filament/Scene.h
@@ -96,7 +96,7 @@ public:
     void addEntity(utils::Entity entity);
 
     /**
-     * Adds a contiguous list of entities to the Scene.
+     * Adds a list of entities to the Scene.
      *
      * @param entities Array containing entities to add to the scene.
      * @param count Size of the entity array.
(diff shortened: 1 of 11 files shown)


filament/include/filament/Scene.h
```diff
@@ -111,6 +111,14 @@ class UTILS_PUBLIC Scene : public FilamentAPI {
      */
     void remove(utils::Entity entity);
 
+    /**
+     * Removes a contiguous list of entities to the Scene.
```
pixelflinger, line 115: I'm not sure "contiguous" is needed here -- when I read it the first time I thought the Entities needed to be contiguous.

I think we should state that this is equivalent to calling remove() on a single entity in a loop.

This is to address the case of the array containing the same Entity several times -- the effect will be "as if" it was removed several times.

Maybe specify what happens when entities is nullptr (e.g. it's forbidden, or it's a noop). I would go with noop (to mimic free()), but don't have a strong opinion.

filament/src/Scene.cpp
```diff
@@ -364,6 +364,12 @@ void FScene::remove(Entity entity) {
     mEntities.erase(entity);
 }
 
+void FScene::removeEntities(const Entity* entities, size_t count) {
+    for (size_t i = 0; i < count; ++i, ++entities) {
+        mEntities.erase(*entities);
```
pixelflinger, line 369: maybe this should just call `remove(Entity)`?

filament/include/filament/Scene.h
prideout: All good suggestions, I updated the commit with new comments.  (If any entity, including the null entity, does not exist in the scene, then the robin_set will silently ignore it.)

filament/src/Scene.cpp
prideout: Done

PR #2845, opened 2020-07-22
Fix KTX texture format for latest version

DDoS:
The *_ibl.ktx and *_skybox.ktx files generated by cmgen cannot be read in the latest KTX release (v4.0.0-beta3). This is because `glFormat` should be unsized, according to spec (so it should be RGB).


prideout:
Thanks!  We should also remove the last case in the switch in toPixelDataFormat in `KtxUtility.h`.

tools/cmgen/src/cmgen.cpp
```diff
@@ -933,7 +933,7 @@ void iblRoughnessPrefilter(
         .endianness = KtxBundle::ENDIAN_DEFAULT,
         .glType = KtxBundle::R11F_G11F_B10F,
         .glTypeSize = 1,
-        .glFormat = KtxBundle::R11F_G11F_B10F,
```
romainguy, line 936: Why was this removed? We need this format for our IBLs. It breaks all environment lighting.

prideout: We're still using the 11_11_10 (see the `glType` field), this needed to be unsized to be valid KTX.  Is stuff breaking for you?

romainguy: Actually the issue isn't here. The issue is above.

libs/image/include/image/KtxUtility.h
```diff
@@ -227,7 +227,6 @@ namespace ktx {
             case KtxBundle::RG: return PixelDataFormat::RG;
             case KtxBundle::RGB: return PixelDataFormat::RGB;
             case KtxBundle::RGBA: return PixelDataFormat::RGBA;
-            case KtxBundle::R11F_G11F_B10F: return PixelDataFormat::RGB;
```
romainguy, line 230: Removing this line breaks compatibility with existing assets.

PR #2849, opened 2020-07-22
DoF: bug fixes and quality improvements

pixelflinger:
- Use the random offset to calculate the correct intersection radius.
  For the center sample, the ring radius is otherwise zero, which means
  any sample in the random disk with a CoC radius > 0 would intersect,
  and that seems wrong.

- The selected mip was off by one.
  We were always picking a mip too high because the CoC are in
  high resolution, but our mip-chain is 1/4 res from that.
  This improves rendering quality quite a lot.

### DIFF
--- a/filament/src/materials/dof.mat
+++ b/filament/src/materials/dof.mat
@@ -179,7 +179,8 @@ void initBucket(out Bucket ring) {
     ring.coc = 0.0;
 }
 
-void initRing(const float i, const float ringCount, const float kernelSize, out float count, out mat2 r, out vec2 p) {
+void initRing(const float i, const float ringCount, const float kernelSize, const vec2 noise,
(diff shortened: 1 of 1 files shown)


filament/src/materials/dof.mat
```diff
@@ -192,6 +193,9 @@ void initRing(const float i, const float ringCount, const float kernelSize, out
 
     float firstSamplePosition = mod(i, 2.0) * 0.5 * inc;
     p = radius * vec2(cos(firstSamplePosition), sin(firstSamplePosition));
+
+    // it might be okay to do this only for the center sample (i.e.: offset = radius instead)
+    offset = length(p + noise * kernelSize);
```
romainguy, line 198: I assume we have to use a length here and we can't use the square distance?

PR #2858, opened 2020-07-23
Add quality option to ColorGrading

romainguy:
Low: 16x16x16 LUT in 10 bit
Medium: 32x32x32 LUT in 10 bit
High: 32x32x32 LUT in 16 bit
Ultra: 64x64x64 LUT in 16 bit

### DIFF
--- a/filament/src/ColorGrading.cpp
+++ b/filament/src/ColorGrading.cpp
@@ -39,13 +39,13 @@ using namespace utils;
 using namespace math;
 using namespace backend;
 
-static constexpr size_t LUT_DIMENSION = 32u;
-
 //------------------------------------------------------------------------------
 // Builder
 //------------------------------------------------------------------------------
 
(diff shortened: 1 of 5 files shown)


filament/src/ColorGrading.cpp
```diff
@@ -492,14 +523,16 @@ FColorGrading::FColorGrading(FEngine& engine, const Builder& builder) {
     //std::chrono::duration<float, std::milli> duration = std::chrono::steady_clock::now() - now;
     //slog.d << "LUT generation time: " << duration.count() << " ms" << io::endl;
 
+    TextureFormat textureFormat = selectLutTextureFormat(builder->quality);
+
     // create the texture, we use RGBA16F because RGB16F is not supported everywhere
```
pixelflinger, line 528: this comment should be moved to the where the format is selected

PR #2862, opened 2020-07-23
DoF: improve quality in "fast tiles"

pixelflinger:
fast tiles (i.e. where all pixels have the same circle-of-confusion)
are allowed to use bilinear filtering.
this results in better looking DoF in these areas.


filament/src/PostProcessManager.cpp
```diff
@@ -1093,6 +1099,13 @@ FrameGraphId<FrameGraphTexture> PostProcessManager::dof(FrameGraph& fg,
      * DoF recombine
      */
 
+    auto outForground = ppDoFMedian.getData().outForeground;
```
romainguy, line 1102: Typo in `outForground` --> `outForeground`

filament/src/materials/dof.mat
```diff
@@ -6,6 +6,11 @@ material {
            name : foreground,
            precision: medium
         },
+        {
+           type : sampler2d,
+           name : foreground_linear,
```
romainguy, line 11: `foregroundLinear`?

PR #2863, opened 2020-07-24
DoF: fix tiles dilate pass edges

pixelflinger:
We were using texelFetch which doesn't handle wrap modes, and this
resulted in all edge tiles to be undefined. Use textureLod() instead.

### DIFF
--- a/filament/src/materials/dofDilate.mat
+++ b/filament/src/materials/dofDilate.mat
@@ -30,8 +30,8 @@ void dummy(){}
 // Tiles of 16 pixels requires two dilate rounds to accomodate our max Coc of 32 pixels
 #define TILE_SIZE   16.0
 
-vec2 tap(const ivec2 uv, const ivec2 offset) {
-    return texelFetch(materialParams_tiles, uv + offset, 0).rg;
+vec2 tap(const vec2 uv, const vec2 offset) {
+    return textureLod(materialParams_tiles, uv + offset, 0.0).rg;
 }
 
 vec2 dilate(inout vec2 center, const vec2 tap) {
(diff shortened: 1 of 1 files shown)


filament/src/materials/dofDilate.mat
```diff
@@ -54,26 +54,32 @@ vec2 dilate(inout vec2 center, const vec2 tap) {
...
-    ivec2 xy = ivec2(uv * size);
+    const float radius = 1.0;
+    // we need highp here just to maintain precision of 1/x
+    highp vec2 size = vec2(textureSize(materialParams_tiles, 0));
+    highp vec2 isize = 1.0 / size;
+
+    vec2 radii = float(radius) * isize;
 
-    vec2 center = tap(xy, ivec2(0, 0));
+    vec2 uv = variable_vertex.xy;
```
romainguy, line 64: Shouldn't this be highp?

filament/src/materials/dofDilate.mat
```diff
@@ -54,26 +54,32 @@ vec2 dilate(inout vec2 center, const vec2 tap) {
...
-    // highp is not needed here, because we now the resolution is very low
-    vec2 uv = variable_vertex.xy;
-    vec2 size = vec2(textureSize(materialParams_tiles, 0));
-    ivec2 xy = ivec2(uv * size);
+    const float radius = 1.0;
+    // we need highp here just to maintain precision of 1/x
+    highp vec2 size = vec2(textureSize(materialParams_tiles, 0));
+    highp vec2 isize = 1.0 / size;
+
+    vec2 radii = float(radius) * isize;
```
romainguy, line 62: Highp?

filament/src/materials/dofDilate.mat
pixelflinger: ah no, because the resolution of tiles is very small -- we assume highp is not needed.

filament/src/materials/dofDilate.mat
pixelflinger: same thing here, we needed highp just for 1/x.

PR #2864, opened 2020-07-24
Fix Hierarchy Corruption

tpsiaki:
When a node is set to have no parent, its .next is not updated.  The result is that later operations, such as removing this node from the graph result in its previous next sibling node being corrupted.

Note that this failure is masked in debug builds because removeNode sets parent/prev/next to 0 in those cases.


tpsiaki:
It looks like I pulled in some stuff here in error.  I'm going to create a new PR.

PR #2870, opened 2020-07-24
fix a few issues with color grading

pixelflinger:
- we were not clamping the color grading output to [0,1]

- we were not using a supported input (cpu) pixel format for the
  3d LUT in low and medium quality

### DIFF
--- a/filament/src/ColorGrading.cpp
+++ b/filament/src/ColorGrading.cpp
@@ -33,6 +33,8 @@
 
 #include <functional>
 
+#include <math.h>
+
 namespace filament {
 
 using namespace utils;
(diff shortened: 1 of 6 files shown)


pixelflinger:
> I can test this out after you merge. We might need to update VulkanUtility.cpp by replacing VK_FORMAT_A2R10G10B10_UNORM_PACK32 with VK_FORMAT_A2B10G10R10_UNORM_PACK32

Yes, that was it! Thanks!

filament/src/ColorGrading.cpp
```diff
@@ -525,17 +545,41 @@ FColorGrading::FColorGrading(FEngine& engine, const Builder& builder) {
...
+    PixelDataFormat format;
+    PixelDataType type;
+    selectLutTextureParams(builder->quality, textureFormat, format, type);
 
     mLutHandle = driver.createTexture(SamplerType::SAMPLER_3D, 1, textureFormat, 0,
             config.lutDimension, config.lutDimension, config.lutDimension, TextureUsage::DEFAULT);
 
+    // convert input to UINT_2_10_10_10_REV if needed
+    if (type == PixelDataType::UINT_2_10_10_10_REV) {
+        uint32_t* const UTILS_RESTRICT dst = (uint32_t*)malloc(lutElementCount * sizeof(uint32_t));
```
romainguy, line 558: BTW we do we use mallocs in this file and now just new type[]?

filament/src/ColorGrading.cpp
```diff
@@ -525,17 +545,41 @@ FColorGrading::FColorGrading(FEngine& engine, const Builder& builder) {
...
+        half4 const* UTILS_RESTRICT src = (half4*)data;
+
+        // we use a vectorize width of 8 because, on ARMv8 it allows the compiler to write eight
+        // 32-bits results in one go.
+        const size_t count = lutElementCount & ~0x7; // tell the compiler that we're a multiple of 8
+        #pragma clang loop vectorize_width(8)
+        for (size_t i = 0; i < count; ++i) {
+            float4 v{ src[i] };
+            uint32_t r = uint32_t(v.x * 1023.0f);
```
romainguy, line 567: Do we need to do any rounding?

filament/src/ColorGrading.cpp
pixelflinger: note sure. doesn't really matter.

filament/src/ColorGrading.cpp
pixelflinger: yeah good point. we could round.

PR #2873, opened 2020-07-27
Fix Metal crash when >16 samplers are bound

bejado:


filament/backend/src/metal/MetalDriver.mm
```diff
@@ -1123,6 +1123,11 @@
     enumerateSamplerGroups(program, [this, &texturesToBind, &samplersToBind](
             const SamplerGroup::Sampler* sampler,
             uint8_t binding) {
+        // We currently only support a max of SAMPLER_BINDING_COUNT samplers. Ignore any additional
+        // samplers that may be bound.
+        if (binding >= SAMPLER_BINDING_COUNT) {
```
pixelflinger, line 1128: should we log something on debug builds? or is it going to be too anoying?

bejado: Yeah, I think that would just add log spew. Plus, we already log that the user has exceeded the max sampler count inside `SamplerBindingMap.cpp`.

PR #2879, opened 2020-07-28
ColorGrading: add JavaScript and TypeScript bindings.

prideout:
This is not tested but seems fairly straightforward.


prideout:
@elalish @sun765

PR #2880, opened 2020-07-28
attempt to reduce code size in PostProcessManager

pixelflinger:
The two main changes are:
- materials are now stored in a hash map which allows us to
  automatically destroy all of them using one loop -- instead of
  having to call terminate() for each of them.

- factored a very common sequence of code:
  commit()
  use()
  beginRenderPass()
  draw()
  endRenderPass()

  This makes the code easier to write and also results in less code.

There are also minor other changes, e.g. RenderPass constructor is not
inline anymore.

### DIFF
--- a/filament/src/PostProcessManager.cpp
+++ b/filament/src/PostProcessManager.cpp
@@ -228,7 +228,7 @@ void PostProcessManager::commitAndRender(FrameGraphRenderTarget const& out,
 
 UTILS_ALWAYS_INLINE
 void PostProcessManager::commitAndRender(FrameGraphRenderTarget const& out,
-        PostProcessMaterial const& material, DriverApi& driver) const noexcept{
+        PostProcessMaterial const& material, DriverApi& driver) const noexcept {
     commitAndRender(out, material, 0, driver);
 }
 
--- a/filament/src/PostProcessManager.cpp
+++ b/filament/src/PostProcessManager.cpp
@@ -57,7 +57,6 @@ PostProcessManager::PostProcessMaterial::PostProcessMaterial() noexcept {
     mEngine = nullptr;
     mData = nullptr;
     mMaterial = nullptr;            // aliased to mEngine
-    mMaterialInstance = nullptr;    // aliased to mData
 }
 
 PostProcessManager::PostProcessMaterial::PostProcessMaterial(FEngine& engine,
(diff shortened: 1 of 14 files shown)


filament/src/PostProcessManager.h
```diff
@@ -157,24 +168,10 @@ class PostProcessManager {
...
-    PostProcessMaterial mDoFTiles;
-    PostProcessMaterial mDoFDilate;
-    PostProcessMaterial mDoFMedian;
-    PostProcessMaterial mDoFCombine;
-    PostProcessMaterial mBloomDownsample;
-    PostProcessMaterial mBloomUpsample;
-    PostProcessMaterial mColorGradingAsSubpass;
-    PostProcessMaterial mColorGrading;
-    PostProcessMaterial mFxaa;
-    PostProcessMaterial mBlit[3];
```
romainguy, line 177: 👍

filament/src/PostProcessManager.cpp
```diff
@@ -195,25 +207,29 @@ void PostProcessManager::terminate(DriverApi& driver) noexcept {
...
+    auto last = mMaterialRegistry.end();
+    while (first != last) {
+        first.value().terminate(engine);
+        ++first;
+    }
+}
+
+UTILS_NOINLINE
+void PostProcessManager::commitAndRender(FrameGraphRenderTarget const& out,
+        PostProcessMaterial const& material, uint8_t variant, DriverApi& driver) const noexcept{
```
bejado, line 220:
```suggestion
        PostProcessMaterial const& material, uint8_t variant, DriverApi& driver) const noexcept {
```

filament/src/PostProcessManager.cpp
```diff
@@ -195,25 +207,29 @@ void PostProcessManager::terminate(DriverApi& driver) noexcept {
...
+    mi->commit(driver);
+    mi->use(driver);
+    driver.beginRenderPass(out.target, out.params);
+    driver.draw(material.getPipelineState(variant), mEngine.getFullScreenRenderPrimitive());
+    driver.endRenderPass();
+}
+
+UTILS_ALWAYS_INLINE
+void PostProcessManager::commitAndRender(FrameGraphRenderTarget const& out,
+        PostProcessMaterial const& material, DriverApi& driver) const noexcept{
```
bejado, line 231:
```suggestion
        PostProcessMaterial const& material, DriverApi& driver) const noexcept {
```

PR #2881, opened 2020-07-28
multithread the conversion to RGB10_A2

pixelflinger:
we just move the conversion code into the already existing jobs.

### DIFF
--- a/filament/src/ColorGrading.cpp
+++ b/filament/src/ColorGrading.cpp
@@ -453,9 +453,20 @@ FColorGrading::FColorGrading(FEngine& engine, const Builder& builder) {
     };
 
     size_t lutElementCount = config.lutDimension * config.lutDimension * config.lutDimension;
-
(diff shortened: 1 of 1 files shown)


filament/src/ColorGrading.cpp
```diff
@@ -453,10 +453,21 @@ FColorGrading::FColorGrading(FEngine& engine, const Builder& builder) {
...
     size_t lutElementCount = config.lutDimension * config.lutDimension * config.lutDimension;
-
     size_t elementSize = sizeof(half4);
     void* data = malloc(lutElementCount * elementSize);
 
+    TextureFormat textureFormat;
+    PixelDataFormat format;
+    PixelDataType type;
+    selectLutTextureParams(builder->quality, textureFormat, format, type);
+     assert(FTexture::validatePixelFormatAndType(textureFormat, format, type));
```
prideout, line 463: indention

PR #2885, opened 2020-07-29
WebGL: use color grading in demo, expose destroy.

prideout:


web/samples/helmet.html
```diff
@@ -57,7 +57,7 @@
         const sunlight = Filament.EntityManager.get().create();
         Filament.LightManager.Builder(LightType.SUN)
             .color(iblColor.slice(0, 3))
-            .intensity(150000)
```
romainguy, line 60: That's quite a change in intensity Oo

web/samples/helmet.html
```diff
@@ -95,13 +95,17 @@
 
         const cameraEntity = Filament.EntityManager.get().create();
         this.camera = engine.createCamera(cameraEntity);
-        this.camera.setExposure(32, 1/125.0, 100);
+
+        const colorGrading = Filament.ColorGrading.Builder()
+            .toneMapping(Filament.ColorGrading$ToneMapping.ACES)
```
romainguy, line 100: You could add a vignette if you want to make it obvious

web/samples/helmet.html
prideout: Fixed.

web/samples/helmet.html
prideout: Oh, actually I need to add bindings for that, it's a View API not a ColorGrading API.  Will do so next.

PR #2887, opened 2020-07-29
Add lightroom IBL, enable ColorGrading by default in gltf_viewer.

prideout:
This IBL was created at Google for SceneViewer, and is also used by `<model-viewer>`.  Before / After below.

double

The non-legacy ACES is now default, it looks much better.  I also tested this IBL with WebGL and Android.

lightroom


romainguy:
Looks great! Did you try to match the intensity as well?

samples/gltf_viewer.cpp
```diff
@@ -96,9 +96,9 @@ struct App {
     } scene;
 
     struct ColorGradingOptions {
-        bool enabled = false;
+        bool enabled = true;
         int quality = static_cast<int>(ColorGrading::QualityLevel::MEDIUM);
-        int toneMapping = static_cast<int>(ColorGrading::ToneMapping::ACES_LEGACY);
+        int toneMapping = static_cast<int>(ColorGrading::ToneMapping::ACES);
```
romainguy, line 101: I am not sure that's what we want to do here. This tone mapper is darker which will create bigger mismatches with other renderers and may make comparisons more difficult.

prideout: Non-legacy looks better but yeah good point -- I changed it back.

romainguy: It shouldn't: the only difference between legacy and non-legacy is exposure. Legacy is a slightly brighter version of non-legacy. You can get the same result by just changing your camera exposure and/or your light intensities.

romainguy: The exposure difference is 1/0.6.

prideout:
> Looks great! Did you try to match the intensity as well?

Just now made web match better with desktop, but I didn't try to match everything with model-viewer, they do stuff like setExposureDirect(1.0) which I'm not sure is best practice.

samples/gltf_viewer.cpp
prideout: Ah interesting -- understood!  I think it looks pretty good right now actually, I think I'll go ahead and merge this.

PR #2888, opened 2020-07-29
fix refraction roughness when specular AA is active

pixelflinger:


shaders/src/shading_lit.fs
```diff
@@ -162,7 +162,10 @@ void getRoughnessPixelParams(const MaterialInputs material, inout PixelParams pi
     float perceptualRoughness = material.roughness;
 #endif
 
-#if defined(GEOMETRIC_SPECULAR_AA)
+    // This is used by the refraction code and must be saved before we apply specular AA
+    pixel.perceptualRoughnessUnclamped = perceptualRoughness;
+
+    #if defined(GEOMETRIC_SPECULAR_AA)
```
romainguy: Indentation

PR #2910, opened 2020-08-04
Infrastructure for frame history data

pixelflinger:
We can now save an arbitrary (but decided at compile time) number of
entries containing information about previous frames and it for while
drawing the current frame. Typically this is used for such effects as
TAA, screen-space reflection, building probes over time, etc...

Currently the frame history consists of one frame per View. FView 
provides a new (internal) getFrameHistory() method used to access
previous frames entries as well as to save the current's frame
information.

The FrameGraphPassResources (which is available in the execute closure
of a FrameGraph pass) has a new detach() method which can be used to
remove that resource from the FrameGraph lifetime management, at which
point it becomes the responsibility of the caller. Typically this
resource will be stored in the frame history.


filament/src/FrameHistory.h
```diff
@@ -0,0 +1,75 @@
...
+// This is where we store all the history of a frame
+struct FrameHistoryEntry {
+    // history data goes here
+};
+
+/*
+ * This is a very simple FIFO that stores the history of previous frames.
+ */
+template<typename T, size_t SIZE = 1u>
+class TFrameHistory {
```
prideout, line 31: Currently, this is a very generic FIFO.  Since it is an internal class that is specifically called "FrameHistory", I would consider publicizing `commit()`, adding a public `drain()` method, removing the friend declaration, and removing the two new pass-through methods in FView in favor of `FView::getFrameHistory()`.

I would also consider moving the default size of "1" from the template arg to the "using" alias.

pixelflinger: all sounds good. will do.

filament/src/View.cpp
```diff
@@ -800,6 +802,19 @@ void FView::renderShadowMaps(FEngine& engine, FEngine::DriverApi& driver, Render
...
 }
 
+void FView::commitFrameHistory(FEngine& engine) noexcept {
+    // Here we need to destroy resources in mFrameHistory.back()
+    // and then push the new history entry to the history stack
+    mFrameHistory.commit();
+}
+
+void FView::drainFrameHistory(FEngine& engine) noexcept {
+    // make sure we free all resources in the history
```
bejado, line 812: What types of resources are you planning on storing in `FrameHistoryEntry`?

pixelflinger: It's going to be things like `Handle<>` which need to be destroyed on the main thread and need a `DriverApi&` for the destruction.

filament/src/FrameHistory.h
pixelflinger: btw, we can't remove the "pass-through" because in the future they won't be pass-through. This is where the clean-up of the resources stored in FrameEntry will happen. The dtor can't do it alone because it needs the allocator that might have been used for allocating the resources in the first place.

prideout: Ah ok, understood.

PR #2911, opened 2020-08-04
fix a FG crasher when we have more than 32 passes

pixelflinger:
The FrameGraph cannot store pointers internally because all objects
are stored in vectors. Unfortunately we were storing the writer of
resources as a pointer. 

This is fixed by storing an index instead.

### DIFF
--- a/filament/src/fg/FrameGraph.cpp
+++ b/filament/src/fg/FrameGraph.cpp
@@ -148,13 +148,14 @@ void FrameGraph::moveResourceBase(FrameGraphHandle fromHandle, FrameGraphHandle 
 
     // The pass that is writing to "fromHandle" no longer does (and might be culled)
     // (note: there can only be a single pass that can be a writer)
-    PassNode* const pass = from.writer;
-    if (pass) {
+    if (from.writerIndex.isValid()) {
+        PassNode* const pass = &mPassNodes[from.writerIndex.index];
+        assert(pass);
         auto pos = std::find_if(pass->writes.begin(), pass->writes.end(),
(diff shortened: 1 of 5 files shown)


filament/src/fg/FrameGraph.cpp
```diff
@@ -284,12 +285,11 @@ FrameGraph& FrameGraph::compile() noexcept {
             resourceNodes[resource.index]->readerCount++;
         }
 
-#ifndef NDEBUG
+        // set the writers
         for (FrameGraphHandle resource : pass.writes) {
             assert(resource.isValid());
-            assert(resourceNodes[resource.index]->writer == &pass);
+            resourceNodes[resource.index]->writer = &pass;
```
romainguy, line 291: Is it normal that the test becomes an assignment?

pixelflinger: yes -- it used to be this way (which is the correct way). But at one point I moved the assignment somewhere else (where it's incorrect), and I added the assert() here to make sure they matched.

filament/src/fg/fg/ResourceNode.h
```diff
@@ -34,6 +34,8 @@ struct ResourceNode { // 24
     ResourceNode(ResourceNode&&) noexcept = default;
     ResourceNode& operator=(ResourceNode const&) = delete;
 
+    FrameGraphHandle writerIndex;   // only needed by moreResource
```
prideout, line 37:
```suggestion
    FrameGraphHandle writerIndex;   // only needed by moveResource
```

PR #2914, opened 2020-08-05
fix a moveResource which was broken recently

pixelflinger:
a recent fix actually badly broke moveResource, we were using the
index of the newly created resource instead of the pass index.

### DIFF
--- a/filament/src/fg/fg/PassNode.cpp
+++ b/filament/src/fg/fg/PassNode.cpp
@@ -114,7 +114,7 @@ FrameGraphHandle PassNode::write(FrameGraph& fg, const FrameGraphHandle& handle)
     // record the write
     auto& newNode = fg.getResourceNodeUnchecked(r);
     assert(!newNode.writerIndex.isValid());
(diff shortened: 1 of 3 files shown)


filament/src/fg/fg/PassNode.cpp
```diff
@@ -114,7 +114,7 @@ FrameGraphHandle PassNode::write(FrameGraph& fg, const FrameGraphHandle& handle)
     // record the write
     auto& newNode = fg.getResourceNodeUnchecked(r);
     assert(!newNode.writerIndex.isValid());
-    newNode.writerIndex = r; // needed by move resources
+    newNode.writerIndex = FrameGraphHandle(this->id); // needed by move resources
```
romainguy, line 117: Should it still be called index?

pixelflinger: yes, it's very much the index of the PassNode in the array. I'm using a FrameGraphHandle just as a convenience because it's basically an index with a non-zero default state.

PR #2919, opened 2020-08-06
gltfio: fix some small memory leaks.

prideout:
This fixes two oversights in the FilamentAsset destructor:

- Destroy the faux root Entity.
- If it exists, remove the name component from each Entity.


prideout:
> BTW looks like we leak a ColorGrading object in gltf_viewer.

Fixed.

PR #2921, opened 2020-08-07
Implement basic temporal antialiasing (TAA)

pixelflinger:
while this started as a sample to demonstrate how to use the new
frame history feature, this is now solid enough to become a feature.

this version of TAA doesn't use motion vectors, so it is prone to
various artifacts (e.g. ghosting). It also has the typical issues
TAA has (e.g. transparents).

with that said, performance is good and quality is overall superior
to FXAA or MSAAx4.

taa

taa-zoom

### DIFF
--- a/filament/src/PostProcessManager.cpp
+++ b/filament/src/PostProcessManager.cpp
@@ -51,6 +51,19 @@ using namespace backend;
 static constexpr uint8_t kMaxBloomLevels = 12u;
 static_assert(kMaxBloomLevels >= 3, "We require at least 3 bloom levels");
 
+constexpr static float halton(unsigned int i, unsigned int b) noexcept {
+    // skipping a bunch of entries makes the average of the sequence closer to 0.5
+    i += 409;
+    float f = 1.0f;
+    float r = 0.0f;
+    while (i > 0u) {
+        f /= float(b);
+        r += f * float(i % b);
+        i /= b;
+    }
+    return r;
+}
+
 // ------------------------------------------------------------------------------------------------
 
 PostProcessManager::PostProcessMaterial::PostProcessMaterial() noexcept {
@@ -141,7 +154,26 @@ FMaterialInstance* PostProcessManager::PostProcessMaterial::getMaterialInstance(
 
 // ------------------------------------------------------------------------------------------------
 
-PostProcessManager::PostProcessManager(FEngine& engine) noexcept : mEngine(engine) {
+PostProcessManager::PostProcessManager(FEngine& engine) noexcept
+        : mEngine(engine),
+          mHaltonSamples{
+                  { filament::halton( 0, 2), filament::halton( 0, 3) },
+                  { filament::halton( 1, 2), filament::halton( 1, 3) },
+                  { filament::halton( 2, 2), filament::halton( 2, 3) },
+                  { filament::halton( 3, 2), filament::halton( 3, 3) },
+                  { filament::halton( 4, 2), filament::halton( 4, 3) },
+                  { filament::halton( 5, 2), filament::halton( 5, 3) },
+                  { filament::halton( 6, 2), filament::halton( 6, 3) },
+                  { filament::halton( 7, 2), filament::halton( 7, 3) },
+                  { filament::halton( 8, 2), filament::halton( 8, 3) },
+                  { filament::halton( 9, 2), filament::halton( 9, 3) },
+                  { filament::halton(10, 2), filament::halton(10, 3) },
+                  { filament::halton(11, 2), filament::halton(11, 3) },
+                  { filament::halton(12, 2), filament::halton(12, 3) },
+                  { filament::halton(13, 2), filament::halton(13, 3) },
+                  { filament::halton(14, 2), filament::halton(14, 3) },
+                  { filament::halton(15, 2), filament::halton(15, 3) }}
+{
 }
 
 UTILS_NOINLINE
@@ -172,6 +204,7 @@ void PostProcessManager::init() noexcept {
     registerPostProcessMaterial("colorGrading", MATERIAL(COLORGRADING));
     registerPostProcessMaterial("colorGradingAsSubpass", MATERIAL(COLORGRADINGASSUBPASS));
     registerPostProcessMaterial("fxaa", MATERIAL(FXAA));
+    registerPostProcessMaterial("taa", MATERIAL(TAA));
     registerPostProcessMaterial("dofDownsample", MATERIAL(DOFDOWNSAMPLE));
     registerPostProcessMaterial("dofMipmap", MATERIAL(DOFMIPMAP));
     registerPostProcessMaterial("dofTiles", MATERIAL(DOFTILES));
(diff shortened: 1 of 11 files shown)


pixelflinger:
**notes:** 
- doesn't work yet with metal or vulkan, in both cases I get a lot of flickering (probably a texture coordinate issue).
- java bindings missing

prideout:
Really nice quality!

filament/src/materials/antiAliasing/taa.mat
```diff
@@ -0,0 +1,280 @@
...
+
+    // untonemap result
+    result.rgb *= 1.0 / (1.0 - luma(result.rgb));
+
+#if USE_YCoCg
+    result.rgb = YCoCg_RGB(result.rgb);
+#endif
+
+    // store result (which will becomes new history)
+    result = max(vec4(0), result);
```
prideout, line 270: Interesting, I'm surprised this could ever be negative going in.

filament/src/materials/antiAliasing/taa.mat
```diff
@@ -0,0 +1,280 @@
...
+    result *= 1.0 / (k0 + k1 + k2 + k3 + k4);
+
+    return result;
+}
+
+void postProcess(inout PostProcessInputs postProcess) {
+    highp vec4 uv = variable_vertex.xyxy; // interpolated to pixel center
+
+    // read color buffer center sample
+    float depth = textureLod(materialParams_depth, uv.xy, 0.0).r;
```
prideout, line 181: The comment says "color" but the codeline says "depth"

filament/src/materials/antiAliasing/taa.mat
```diff
@@ -0,0 +1,280 @@
...
+
+#if HISTORY_REPORJECTION
```
prideout, line 183: spelling

filament/src/materials/antiAliasing/taa.mat
romainguy: That's because of the bicubic/catmull-rom spline interpolation I think. Other implementations end up with the same fix.

filament/src/PostProcessManager.cpp
```diff
@@ -1513,6 +1546,115 @@ FrameGraphId<FrameGraphTexture> PostProcessManager::fxaa(FrameGraph& fg,
...
+                };
+
+                FrameHistoryEntry& current = frameHistory.getCurrent();
+
+                float sum = 0.0;
+                float weights[9];
+                for (size_t i = 0; i < 9; i++) {
+                    float2 d = sampleOffsets[i] - current.jitter;
+                    d *= 1.0 / taaOptions.filterWidth;
+                    weights[i] = std::exp(-2.29f * (d.x * d.x + d.y * d.y));
```
romainguy, line 1623: Might be worth explaining what this formula is/where -2.29f comes from?

filament/src/materials/antiAliasing/taa.mat
```diff
@@ -0,0 +1,280 @@
...
+// accurate box clipping
+#define BOX_CLIPPING_ACCURATE   0
+// clamping instead of clipping
+#define BOX_CLIPPING_CLAMP      1
+// no clipping (for debugging only)
+#define BOX_CLIPPING_NONE       2
+
+#if defined(TARGET_MOBILE)
+#define BOX_CLIPPING            BOX_CLIPPING_ACCURATE
+#define USE_YCoCg               0
```
romainguy, line 56: Will any of these options be eventually enabled on mobile? Or just too expensive? Should these be part of the API as toggle or even as part of a quality setting?

filament/src/materials/antiAliasing/taa.mat
```diff
@@ -0,0 +1,280 @@
...
+#define FILTER_HISTORY          0
+#else
+#define BOX_CLIPPING            BOX_CLIPPING_ACCURATE
+#define USE_YCoCg               1
+#define PREVENT_FLICKERING      0   // FIXME: thin lines disapear
+#define FILTER_HISTORY          1
+#endif
+
+// disable for debugging only
+#define HISTORY_REPORJECTION    1
```
romainguy, line 67: Typo

filament/src/materials/antiAliasing/taa.mat
```diff
@@ -0,0 +1,280 @@
...
+
+void dummy() {} // twart IDE
```
romainguy, line 69: twart -> thwart

filament/src/materials/antiAliasing/taa.mat
```diff
@@ -0,0 +1,280 @@
...
+
+float luma(const vec3 color) {
+#if USE_YCoCg
+    return color.x;
+#else
+#if defined(TARGET_MOBILE)
+    return max3(color);
+#else
+    return dot(color, vec3(0.2126, 0.7152, 0.0722));
```
romainguy, line 78: We have a built-in `luminance(vec3 x)` function you can call instead of rolling your own.

filament/src/materials/antiAliasing/taa.mat
```diff
@@ -0,0 +1,280 @@
...
+#endif
+#endif
+}
+
+vec3 RGB_YCoCg(const vec3 c) {
```
romainguy, line 83: Should these two functions be moved to `common_graphics.fs`?

filament/src/materials/antiAliasing/taa.mat
```diff
@@ -0,0 +1,280 @@
...
+    }
+    return h;
+}
+
+// Samples a texture with Catmull-Rom filtering, using 9 texture fetches instead of 16.
+//      https://therealmjp.github.io/
+// Some optimizations from here:
+//      http://vec3.ca/bicubic-filtering-in-fewer-taps/ for more details
+// Optimized to 5 taps by removing the corner samples
+// And Modified for mediump support
```
romainguy, line 123: Modified -> modified

libs/gltfio/include/gltfio/SimpleViewer.h
```diff
@@ -461,8 +462,14 @@ void SimpleViewer::updateUserInterface() {
     if (ImGui::CollapsingHeader("View")) {
         ImGui::Indent();
         ImGui::Checkbox("Dithering", &mEnableDithering);
-        ImGui::Checkbox("FXAA", &mEnableFxaa);
         ImGui::Checkbox("MSAA 4x", &mEnableMsaa);
+        ImGui::Checkbox("TAA", &mTAAOptions.enabled);
+        // this clutters the UI and isn't that useful (except when working on TAA)
```
romainguy, line 467: `#ifndef NDEBUG` maybe?

bejado:
Windows error:

```
D:\a\filament\filament\filament\src\PostProcessManager.cpp(1640,19): error C2664: 'void filament::FMaterialInstance::setParameter(const char *,filament::backend::Handle<filament::backend::HwTexture>,filament::backend::SamplerParams) noexcept': cannot convert argument 2 from 'filament::backend::Handle<filament::backend::HwTexture>' to 'const filament::Texture *' [D:\a\filament\filament\out\cmake-mt\filament\filament.vcxproj]
```

filament/src/materials/antiAliasing/taa.mat
pixelflinger: I think probably at some point yes, once we understand better what works.

filament/src/materials/antiAliasing/taa.mat
pixelflinger: I think maybe at some point, if we keep them. Once we know what works better.
I think it's pretty specific to TAA too?

romainguy: I played with it for gamut mapping and tone mapping, but yeah we can move it later.

libs/gltfio/include/gltfio/SimpleViewer.h
pixelflinger: I really didn't think they were that useful other than when debugging TAA, I don't think people will tweak them in practice.

filament/src/PostProcessManager.cpp
```diff
@@ -141,7 +154,26 @@ FMaterialInstance* PostProcessManager::PostProcessMaterial::getMaterialInstance(
 
 // ------------------------------------------------------------------------------------------------
 
-PostProcessManager::PostProcessManager(FEngine& engine) noexcept : mEngine(engine) {
+PostProcessManager::PostProcessManager(FEngine& engine) noexcept
+        : mEngine(engine),
+          mHaltonSamples{
+                  { filament::halton( 0, 2), filament::halton( 0, 3) },
```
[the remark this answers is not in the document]
pixelflinger: yeah, crazy right? I think it has to actually.

Issue #2057, opened 2020-01-21
Option to configure `CommandBufferQueue` size.


roxlu:
I'm rendering a scene with ~3000 items that share ~100 meshes and it seems that the `CommandBufferQueue` is running out of space.  In `Allocators.h` I see the buffer sizes, but I'm wondering how to change the size; or maybe there is another way to render 3000 items?

romainguy:
I talked to @pixelflinger and this is pretty much what we wanted to add. It was just never a priority so far. We'd love a test scene if you have one (a glTF file would work great if that's not a problem for you).

pixelflinger:
I had to increase `CONFIG_PER_RENDER_PASS_ARENA_SIZE` to 3 MiB, `CONFIG_PER_FRAME_COMMANDS_SIZE` to 2 MiB and `CONFIG_MIN_COMMAND_BUFFERS_SIZE` to 3 MiB to be able to render a ~3400 items scene.

We will make this user-settable at Engine creation time eventually.  Also maybe we could increase the defaults a bit.

roxlu:
I want to look into this and wondering how we can make the buffer sizes configurable. I've looked into some files and found the following buffer sizes which should be configurable. In [Allocators.h](https://github.com/google/filament/blob/main/filament/src/details/Allocators.h ) we have: 

-  `CONFIG_PER_RENDER_PASS_ARENA_SIZE`
- `CONFIG_PER_FRAME_COMMANDS_SIZE`
- `CONFIG_MIN_COMMAND_BUFFERS_SIZE`

And in [OpenGLDriver.cpp](https://github.com/google/filament/blob/main/filament/backend/src/opengl/OpenGLDriver.cpp#L136) we have:

- `mHandleArena("Handles", 2U * 1024U * 1024U),`

I didn't find other other buffer sizes which should be configurable; but there might be more. 

@pixelflinger and @romainguy what are you thoughts on this? Should we create `Builders` for these too?

roxlu:
@romainguy what about using #defines so you can at least configure them? This could be added in 5min and we have a solution that works for now.

PR #2923, opened 2020-08-07
Configurable buffer sizes via CMake

roxlu:
This feature allows us to specify the buffer sizes that are used in Allocators.h and OpenGLDriver.cpp. Quick fix for https://github.com/google/filament/issues/2057 By increasing these buffers we can render more instances.


prideout:
This PR has two 1-month old commits in it, should it be rebased?

roxlu:
> This PR has two 1-month old commits in it, should it be rebased?

I created a feature branch from upstream/master. Not sure what happened.

romainguy:
We changed our branch from master to main a few weeks ago. And your change should define default values in the code, not just in the build system.

roxlu:
> We changed our branch from master to main a few weeks ago. 
Ah good! I'll rebase late.

> And your change should define default values in the code, not just in the build system.

https://github.com/roxlu/filament/commit/899f9a1d1fe88c2bf54711b9af7b6e25d591e05c

What do you mean; see above for the changes in Allocators.h and OpenGLDriver.cpp

romainguy:
That's exactly my point, your change *relies* on defines being set by the build system, not something we want. You should at least have a series of `#ifndef XXXX #define XXXX YYYY #endif`.

roxlu:
Ah I see what you mean; I wasn’t sure if you wanted to have such an implementation. I’ll make the changes.

romainguy:
The master branch has now been deleted to avoid further confusions. GitHub finally fixed the issue that was preventing us from getting rid of it.

romainguy:
Your change still touches many files that are not related to your change.

PR #2930, opened 2020-08-10
API CHANGE: make ssao options work like all other options

pixelflinger:
setAmbientOcclusion (and associates) is now deprecated (but still works),
in favor of the new .enabled field in AmbientOcclusionOptions.

### DIFF
--- a/android/filament-android/src/main/java/com/google/android/filament/View.java
+++ b/android/filament-android/src/main/java/com/google/android/filament/View.java
@@ -139,8 +139,7 @@ public class View {
     }
 
     /**
-     * Options for Ambient Occlusion
-     * @see #setAmbientOcclusion
+     * Options for screen space Ambient Occlusion
      */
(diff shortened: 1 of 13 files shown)


android/filament-android/src/main/java/com/google/android/filament/View.java
```diff
@@ -184,6 +183,11 @@
          */
         @NonNull
         public QualityLevel upsampling = QualityLevel.LOW;
+
+        /**
+         * enable or disable screen space ambient occlusion
+         */
+        public boolean enabled = false;
```
romainguy, line 190: We should update the sample apps

pixelflinger: I did that didn't I?

pixelflinger: oh the java apps. got it.

PR #2931, opened 2020-08-11
make use of textureLodOffet where appropriate

pixelflinger:
this really makes the code easier to read and has a chance to be
more efficient

### DIFF
--- a/filament/src/materials/blitHigh.mat
+++ b/filament/src/materials/blitHigh.mat
@@ -27,20 +27,20 @@ vertex {
 }
 
 fragment {
+    void dummy() {}
+
     void postProcess(inout PostProcessInputs postProcess) {
         highp vec2 uv = variable_vertex.xy; // interpolated to pixel center
-        highp float du = materialParams.resolution.z;
-        highp float dv = materialParams.resolution.w;
         vec4 c0, c1;
-        c0  = textureLod(materialParams_color, uv + vec2(-du, -dv), 0.0);
-        c0 += textureLod(materialParams_color, uv + vec2( du, -dv), 0.0);
-        c0 += textureLod(materialParams_color, uv + vec2( du,  dv), 0.0);
-        c0 += textureLod(materialParams_color, uv + vec2(-du,  dv), 0.0);
+        c0  = textureLodOffset(materialParams_color, uv, 0.0, ivec2(-1, -1));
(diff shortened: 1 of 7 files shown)


filament/src/materials/blitHigh.mat
```diff
@@ -27,20 +27,20 @@ vertex {
...
     void postProcess(inout PostProcessInputs postProcess) {
         highp vec2 uv = variable_vertex.xy; // interpolated to pixel center
-        highp float du = materialParams.resolution.z;
-        highp float dv = materialParams.resolution.w;
         vec4 c0, c1;
-        c0  = textureLod(materialParams_color, uv + vec2(-du, -dv), 0.0);
-        c0 += textureLod(materialParams_color, uv + vec2( du, -dv), 0.0);
-        c0 += textureLod(materialParams_color, uv + vec2( du,  dv), 0.0);
-        c0 += textureLod(materialParams_color, uv + vec2(-du,  dv), 0.0);
+        c0  = textureLodOffset(materialParams_color, uv, 0.0, ivec2(-1, -1));
```
romainguy, line 35: So one thing I'm wondering is whether textureLodOffset() does the equivalent of a textureSize() per call. Did you measure before/after?

filament/src/materials/dof/dofMedian.mat
```diff
@@ -92,23 +92,21 @@ void minLeftMaxRight(inout vec4 a, inout vec4 b, inout vec4 c, inout vec4 d, ino
...
 }
 
-vec4 makeSample(ivec2 ij) {
-    vec4 r = vec4(texelFetch(materialParams_dof, ij, 0).rgb,
-            texelFetch(materialParams_alpha, ij, 0).r);
-    return r;
-}
+// - we can't use a function because offset must be a constant expression
+// - we can't use line continuation because that's not supported on desktop
+#define makeSample(ij, o) vec4(texelFetchOffset(materialParams_dof, ij, 0, o).rgb, texelFetchOffset(materialParams_alpha, ij, 0, o).r)
```
romainguy, line 97: WTF GL :)

filament/src/materials/blitHigh.mat
pixelflinger: I did a quick test with bloom, it doesn't seem to make any difference (1ms @ 425MHz / 1080p / P4)

PR #2935, opened 2020-08-11
Fix resizing issues with MoltenVK.

prideout:
This removes the old getClientExtent stuff that I added a while back,
and replaces it with the strategy espoused by the MoltenVK author found
here:

https://github.com/KhronosGroup/MoltenVK/issues/428

I tested this PR on macOS with Vulkan and Metal. I tested both on high
and low DPI displays. I also tried dragging the window between the two
displays.

Android still has issues with rotation, I will deal with that next.

### DIFF
--- a/filament/backend/src/vulkan/VulkanContext.cpp
+++ b/filament/backend/src/vulkan/VulkanContext.cpp
@@ -299,7 +299,7 @@ void getPresentationQueue(VulkanContext& context, VulkanSurfaceContext& sc) {
     ASSERT_POSTCONDITION(sc.presentQueue, "Unable to obtain presentation queue.");
 }
 
-void getSurfaceCaps(VulkanContext& context, VulkanSurfaceContext& sc) {
+static VkSurfaceCapabilitiesKHR getSurfaceCaps(VulkanContext& context, VulkanSurfaceContext& sc) {
     VkResult result = vkGetPhysicalDeviceSurfaceCapabilitiesKHR(context.physicalDevice,
             sc.surface, &sc.surfaceCapabilities);
     ASSERT_POSTCONDITION(result == VK_SUCCESS, "vkGetPhysicalDeviceSurfaceCapabilitiesKHR error.");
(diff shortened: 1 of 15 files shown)


filament/backend/src/vulkan/VulkanContext.cpp
```diff
@@ -357,7 +359,17 @@ void createSwapChain(VulkanContext& context, VulkanSurfaceContext& surfaceContex
         .imageExtent = size,
         .imageArrayLayers = 1,
         .imageUsage = VK_IMAGE_USAGE_COLOR_ATTACHMENT_BIT | VK_IMAGE_USAGE_TRANSFER_DST_BIT,
+
+        // TODO: Setting the preTransform to IDENTITY means we are letting the Android Compositor
```
romainguy, line 363: With Vulkan we definitely need to handle pre-rotation correctly otherwise we'll get horrible performance in one of the display orientations.

libs/filamentapp/src/NativeWindowHelperCocoa.mm
```diff
@@ -42,6 +42,12 @@
     // full-screen, we can skip the macOS compositor if the size matches the display size.
     metalLayer.drawableSize = [view convertSizeToBacking:view.bounds.size];
 
+    // In its implementation of vkGetPhysicalDeviceSurfaceCapabilitiesKHR, MoltenVK takes into
+    // consideration both the size (in points) of the bounds, and the contentsScale of the
+    // CAMetalLayer from which the Vulkan surface was created.
+    // See also https://github.com/KhronosGroup/MoltenVK/issues/428
+    metalLayer.contentsScale = view.window.backingScaleFactor;
```
prideout, line 49: @bejado PTAL at this codeline and the one in resize, since these code paths are active with our Metal backend too.  AFAICT it has no effect with our Metal backend.

filament/backend/src/vulkan/VulkanContext.cpp
prideout: @romainguy According to the article, we would need to adjust our MVP as well as all our dFdx/dFdy calls in GLSL, which seems a bit painful.  :-\  How is this done in OpenGL?

romainguy: It's done magically by the driver on Android. The Graphics team can help us to do this right, and maybe glslang/spirv-* have support for this? Changing derivatives would be a major pain. Wouldn't that mean calling the *Grad() variants of texture calls too?

libs/filamentapp/src/NativeWindowHelperCocoa.mm
bejado: Well, this would affect the Metal backend, but only on desktop. This looks fine to me, if you run with the Metal backend does scaling work correctly still?

filament/backend/src/vulkan/VulkanContext.cpp
pixelflinger: I don't think we need to do anything because everything happens on the GPU and textures are all swizzled, I don't see why 90 rotation of a swizzled texture would be slower. On the Android UI things are different because the gralloc buffers are not swizzled (because they need to be read by the HWC).

The only case where this would be a problem is when we draw directly into the swapchain -- but that basically never happens.

romainguy: Right but our final buffer is not swizzled, the one coming from the Surface? Shouldn't we handle pre-rotation there at least?

libs/filamentapp/src/NativeWindowHelperCocoa.mm
prideout: @bejado Yep it looks fine with `-a metal` on both my monitors.

filament/backend/src/vulkan/VulkanContext.cpp
pixelflinger: Yes, yes. But I meant this will be much easier, because it's a post-process pass, often just a blit. So derivatives shouldn't be a problem, nor the MVP.

PR #2938, opened 2020-08-11
Add shadow map A/B controls

bejado:
This adds a tab to gltf_viewer that allows comparing two sets of shadow options:

shadow-comparison

### DIFF
--- a/libs/gltfio/include/gltfio/SimpleViewer.h
+++ b/libs/gltfio/include/gltfio/SimpleViewer.h
@@ -506,7 +506,7 @@ void SimpleViewer::updateUserInterface() {
             ImGui::SliderInt("Shadow map size", &o.shadowMapSizeExponent, 5, 13, label);
         };
 
-        if (ImGui::BeginTabBar("Shadow options")) {
+        if (ImGui::BeginTabBar("Shadow presets")) {
             if (ImGui::BeginTabItem("A")) {
                 shadowProperties(mShadowPropertiesA);
                 mShadowProperties = &mShadowPropertiesA;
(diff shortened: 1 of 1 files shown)


romainguy:
Does this make sense in gltf_viewer? SimpleViewer is intended to be used by third party devs I think.

bejado:
> Does this make sense in gltf_viewer? SimpleViewer is intended to be used by third party devs I think.

The sun entity belongs to SimpleViewer, so it made sense to put them there, but I agree, they're not interesting to third party developers. I suppose I could move all of the sun settings into gltf_viewer.

These "debug" settings really belong in the controller app, and once we have working that I can migrate them into there.

prideout:
I wonder if the `Shadow Options` group should be renamed to `Shadow Presets`?

`SimpleViewer` was intended for 3P usage, so we might keep it even after we build the new controller system.  If this is temporary or experimental, I think it should probably be moved elsewhere.  Otherwise it's probably fine, since we already have similar things like `Debug Cascades`.

bejado:
This change isn't meant to be temporary, but I think it pushes the line of what belongs inside SimpleViewer.h. So my suggestion is to merge it for now (because like you said, we already have Debug Cascades). After we have the controller app is finished, all of these "debug" type settings can be re-evaluated if they truly belong inside SimpleViewer. If not, we should move them to the controller.

bejado:
I'm also fine renaming to "Shadow Presets", though note this isn't visible in the UI.

prideout:
> This change isn't meant to be temporary, but I think it pushes the line of what belongs inside SimpleViewer.h. So my suggestion is to merge it for now (because like you said, we already have Debug Cascades). After we have the controller app is finished, all of these "debug" type settings can be re-evaluated if they truly belong inside SimpleViewer. If not, we should move them to the controller.

SGTM

pixelflinger (commented):
Is this going in eventually? Just curious.

bejado:
> Is this going in eventually? Just curious.

Let me revisit this with the new changes made to SimpleViewer.

PR #2940, opened 2020-08-11
Switch to C++17

bejado:
Just seeing if everything compiles first.


prideout:
Fingers crossed!

bejado:
Looks like some errors from `jsbindings.cpp`, but maybe these are legitimate?

prideout:
> Looks like some errors from `jsbindings.cpp`, but maybe these are legitimate?

Weird, let me pull the branch and play around with this.

PR #2945, opened 2020-08-12
Add public APIs for texture 3D and texture 2D arrays

pixelflinger:
Also cleanup all setImage() error reporting. All error checks are
changed from silent to NON_FATAL_ASSERT, which means they'll throw
if exceptions are enabled or do a no-op and log a message otherwise.

Also added more precondition checks.

### DIFF
--- a/filament/include/filament/Texture.h
+++ b/filament/include/filament/Texture.h
@@ -129,7 +129,7 @@ public:
          * effectively create a 3D texture.
          * @param depth Depth of the texture in texels (default: 1).
          * @return This Builder, for chaining calls.
-         * @attention This Texture instance must use Sampler::SAMPLER_2D_ARRAY or it has no effect.
+         * @attention This Texture instance must use Sampler::SAMPLER_3D or Sampler::SAMPLER_2D_ARRAY or it has no effect.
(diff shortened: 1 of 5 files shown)


filament/include/filament/Texture.h
```diff
@@ -331,7 +331,32 @@ class UTILS_PUBLIC Texture : public FilamentAPI {
...
+     * @param zoffset   Depth offset of the sub-region to update.
+     * @param width     Width of the sub-region to update.
+     * @param height    Height of the sub-region to update.
+     * @param depth     Depth of the sub-region to update.
+     * @param buffer    Client-side buffer containing the image to set.
+     *
+     * @attention \p engine must be the instance passed to Builder::build()
+     * @attention \p level must be less than getLevels().
+     * @attention \p buffer's Texture::Format must match that of getFormat().
+     * @attention This Texture instance must use Sampler::SAMPLER_3D.
```
prideout, line 352: Can it be `SAMPLER_2D_ARRAY`?

pixelflinger: good catch!  thanks.

PR #2951, opened 2020-08-13
Set the model matrix and projection separately for shadow cameras

bejado:
This is in preparation for VSM, which needs to access the light's position in the shaders.

### DIFF
--- a/filament/src/ShadowMap.cpp
+++ b/filament/src/ShadowMap.cpp
@@ -329,6 +329,7 @@ void ShadowMap::computeShadowCameraDirectional(
 
         // Compute the LiSPSM warping
         mat4f W, Wp;
+        mat4f L; // Rotation matrix in light space
         if (USE_LISPSM) {
             // Orient the shadow map in the direction of the view vector by constructing a
             // rotation matrix in light space around the z-axis, that aligns the y-axis with the camera's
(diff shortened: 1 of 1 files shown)


filament/src/ShadowMap.cpp
```diff
@@ -440,8 +440,16 @@ void ShadowMap::computeShadowCameraDirectional(
...
+        const mat4f b = mat4f::translation(dir * params.options.constantBias);
+        const mat4f Sb = S * b;
+
+        // It's important to set the light camera's model matrix separately from its projection, so
+        // that the cameraPosition uniform gets set correctly.
+        // mLightSpace is used in the shader to access the shadow map texture, and has the model
+        // matrix baked in.
+
+        mCamera->setModelMatrix(FCamera::rigidTransformInverse(b) * M);
+        mCamera->setCustomProjection(mat4(F * W * L * Mp), znear, zfar);
```
bejado, line 452: The change here results in an identical final transform.
```
F * W * L * Mp * Mv * b
F * W * L * Mp * (Mv * b)
F * W * L * Mp * (b * Mv)^-1
F * W * L * Mp * view matrix
view matrix = (b * Mv)^-1
model matrix = b * Mv
```

bejado:
> This makes me very nervous, I'm not sure why.

Understandable. I double checked and the resultant `clipFromWorld` remains unchanged.

PR #2960, opened 2020-08-14
Vulkan: add support for MSAA render targets.

prideout:
This does not include lazy allocation or support for an MSAA swap chain.
Those features will come later.

### DIFF
--- a/filament/backend/src/vulkan/VulkanHandles.cpp
+++ b/filament/backend/src/vulkan/VulkanHandles.cpp
@@ -88,53 +88,107 @@ VulkanProgram::~VulkanProgram() {
     vkDestroyShaderModule(context.device, bundle.fragment, VKALLOC);
 }
 
-static VulkanAttachment createOffscreenAttachment(VulkanAttachment source) {
-    VulkanTexture* tex = source.texture;
-    if (tex == nullptr) {
-        return source;
+static VulkanAttachment createAttachment(VulkanAttachment spec) {
+    if (spec.texture == nullptr) {
+        return spec;
     }
     return {
-        .format = tex->vkformat,
-        .image = tex->textureImage,
-        .view = VK_NULL_HANDLE,
-        .memory = VK_NULL_HANDLE,
-        .texture = tex,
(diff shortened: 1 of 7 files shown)


filament/backend/src/vulkan/VulkanHandles.cpp
```diff
@@ -140,9 +141,20 @@ VulkanRenderTarget::VulkanRenderTarget(VulkanContext& context, uint32_t width, u
             viewInfo.subresourceRange.layerCount = 1;
         }
         vkCreateImageView(context.device, &viewInfo, VKALLOC, &attachment.view);
+
+        // Create a sidecar MSAA texture if this attachment needs to resolve to a 1-sample texture.
+        if (texture->samples == 1 && samples > 1) {
+            mMsaaTextures[targetIndex] = new VulkanTexture(context, texture->target,
+                    texture->levels, texture->format, samples, texture->width, texture->height,
+                    texture->depth, texture->usage, stagePool);
```
bejado, line 149: One special case that I forgot to handle in my initial implementation:

If the render target is a layer in a `2D_TEXTURE_ARRAY`, then the sidecar texture only needs a single layer. The sidecar then should be configured to resolve to the target layer of the render target. I'm not sure how to handle this in Vulkan, and maybe you already are, but just something to be aware of.

prideout: Interesting, thanks!  I wonder if we have a test or sample that exercises that case, would be useful.

bejado: We will after VSM!

pixelflinger: wow...  actually I don't even know if that works in GLES -- I don't know if EXT_multisampled_render_to_texture is supposed to handle 2D arrays. so *if* this is not supported in GLES, then I would need to do the sidecar version in that case on ES. Time to read the spec...

prideout:
This is finally ready to go.  Took a while to iron out bugs found with Refraction.

PR #2966, opened 2020-08-17
Add documentation for cmgen

### LINKED ISSUE #454: Improve cmgen documentation
As @TimvanScherpenzeel pointed out, cmgen should better explain that --size applies to the DFG LUT as well and that a cubemap and the LUT cannot be generated at the same time.

m0rphtail:
resolve #454


tools/cmgen/README.md
```diff
@@ -0,0 +1,92 @@
+# cmgen
+
+`cmgen` is a command-line tool for generating SH and mipmap levels from an env map.
+Cubemaps and equirectangular formats are both supported, automatically detected according to the aspect ratio of the source image.
+
+The tool consumes a HDR environment map in latlong format, and produces two cubemap files: a mipmapped IBL(Image Based Lighting) and a blurry skybox.
```
romainguy, line 6: It *can* produce those two files, but not necessarily. The input HDR map doesn't have to be equirectilinear, it also accepts "cross" cubemap formats (vertical and horizontal), and row/column cubemap formats.

PR #2968, opened 2020-08-17
Add new shader variants for VSM

bejado:
This reserves 7 new variants for VSM (no shader code change yet):

```
    #31  gl41   fs 0x45   DIR|SRE|VSM
    #32  gl41   fs 0x46   DYN|SRE|VSM
    #33  gl41   fs 0x47   DIR|DYN|SRE|VSM
    #34  gl41   fs 0x50   DEP|VSM
    #35  gl41   fs 0x65   DIR|SRE|FOG|VSM
    #36  gl41   fs 0x66   DYN|SRE|FOG|VSM
    #37  gl41   fs 0x67   DIR|DYN|SRE|FOG|VSM

```

### DIFF
--- a/libs/filabridge/include/filament/MaterialEnums.h
+++ b/libs/filabridge/include/filament/MaterialEnums.h
@@ -27,7 +27,7 @@
 namespace filament {
 
 // update this when a new version of filament wouldn't work with older materials
-static constexpr size_t MATERIAL_VERSION = 7;
+static constexpr size_t MATERIAL_VERSION = 8;
 
 /**
  * Supported shading models
--- a/libs/filabridge/include/private/filament/Variant.h
+++ b/libs/filabridge/include/private/filament/Variant.h
@@ -21,9 +21,10 @@
 #include <cstddef>
 
 namespace filament {
-    static constexpr size_t VARIANT_COUNT = 64;
+    static constexpr size_t VARIANT_COUNT = 128;
 
     // IMPORTANT: update filterVariant() when adding more variants
+    // Also be sure to update formatVariantString inside CommonWriter.cpp
     struct Variant {
         Variant() noexcept = default;
         constexpr explicit Variant(uint8_t key) noexcept : key(key) { }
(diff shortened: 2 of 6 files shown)


bejado:
> Any chance this could not be a variant? Could you look at how much it increases the size of a lit material?

This has to be a variant because VSM uses a different sampler type (`sampler2d` as opposed to `sampler2dShadow`). Otherwise we'd have to burn another sampler slot.

sandboxLit.filamat increases by roughly 250 KiB. That also includes new shader code not in this PR.

romainguy:
An extra 250 KiB seems like a lot? Is our dictionary not working or is that without optimizations?

bejado:
Yeah, 250 KiB does seem like a lot. Compiling for mobile with only OpenGL, the size increase is 80 KiB, which is still more than I'd expect.

I think optimizations cause even small shader changes to generate a bunch of unique lines. For example, each of these lines comes from the same source line, but due to variable name changes are stored as separate unique lines:

```
vec3 _1570 = fragColor.xyz * _3218;
vec3 _2118 = fragColor.xyz * _4239;
vec3 _2439 = fragColor.xyz * _4810;
vec3 _2490 = fragColor.xyz * _5328;
```

romainguy:
Yeah optimizations can definitely cause a size increase. We should probably
move to a huffman encoding system at some point. The line dictionary is
neat but too limited. We could encode the dictionary itself vs all the
sources maybe?

bejado:
There's probably a lot we could experiment with, but I'm wondering how effective that would be considering the materials will most likely end up in a compressed AAR anyway.

romainguy:
On Android yes, I was thinking of other platforms (which are admittedly lower priority).

romainguy:
There are other possible options:
- VSM could be opt-in (we would need some kind of config file format/API so you can init an Engine/invoke matc with the same config)
- VSM should be a filterable variant (if you want to reduce the size of your shaders and you know you never turn on VSM, might as well not generate/store those shaders)

bejado:
Yeah, I think making VSM on by default but opt-out via a variant filter is a good compromise now. Let me add that to this PR.

PR #2969, opened 2020-08-17
Add new lightPosition uniform

bejado:


filament/src/ShadowMap.cpp
```diff
@@ -102,10 +102,41 @@ void ShadowMap::render(DriverApi& driver, Handle<HwRenderTarget> rt,
 }
 
 void ShadowMap::computeSceneCascadeParams(const FScene::LightSoa& lightData, size_t index,
-        FScene const* scene, filament::CameraInfo const& camera, uint8_t visibleLayers,
+        FView const& view, filament::CameraInfo const& camera, uint8_t visibleLayers,
         CascadeParameters& cascadeParams) {
+    // Calculate the directional light's "position".
+    // For VSM, we pick a point on the sphere that bounds the camera's culling frustum.
+    if (false /*view.hasVsm()*/) {
```
bejado, line 109: Will update in later PR

romainguy: SG. Speaking of which should VSMs be a view property or a light/shadowmap property?

bejado: A per-view property. If it was per-light, then we'd have to burn an additional sampler (surfaces could be affected by both PCF and VSM shadows).

romainguy: Doh, of course. Dammit, samplers @!#.

PR #2971, opened 2020-08-18
Add VSM shader changes

bejado:
`getPerViewSib` now takes a `uin8t_t` variant key, because the Sib changes slightly depending on if VSM is on or not.

### DIFF
--- a/libs/filabridge/src/SamplerInterfaceBlock.cpp
+++ b/libs/filabridge/src/SamplerInterfaceBlock.cpp
@@ -76,7 +76,7 @@ SamplerInterfaceBlock::Builder::~Builder() noexcept = default;
 
 SamplerInterfaceBlock::SamplerInterfaceBlock() = default;
 SamplerInterfaceBlock::SamplerInterfaceBlock(const SamplerInterfaceBlock& rhs) = default;
-SamplerInterfaceBlock::SamplerInterfaceBlock(SamplerInterfaceBlock&& rhs) noexcept /* = default */ {};
+SamplerInterfaceBlock::SamplerInterfaceBlock(SamplerInterfaceBlock&& rhs) noexcept = default;
 SamplerInterfaceBlock& SamplerInterfaceBlock::operator=(const SamplerInterfaceBlock& rhs) = default;
 SamplerInterfaceBlock& SamplerInterfaceBlock::operator=(SamplerInterfaceBlock&& rhs) /*noexcept*/ = default;
 SamplerInterfaceBlock::~SamplerInterfaceBlock() noexcept = default;
--- a/libs/filabridge/src/SibGenerator.cpp
+++ b/libs/filabridge/src/SibGenerator.cpp
@@ -16,19 +16,22 @@
 
 #include "private/filament/SibGenerator.h"
 
+#include "private/filament/Variant.h"
+
 #include <backend/DriverEnums.h>
 
 #include <private/filament/SamplerInterfaceBlock.h>
 
 namespace filament {
 
-SamplerInterfaceBlock const& SibGenerator::getPerViewSib() noexcept {
+SamplerInterfaceBlock const& SibGenerator::getPerViewSib(uint8_t variantKey) noexcept {
     using Type = SamplerInterfaceBlock::Type;
     using Format = SamplerInterfaceBlock::Format;
     using Precision = SamplerInterfaceBlock::Precision;
 
-    // TODO: ideally we'd want this to be constexpr, this is a compile time structure
-    static SamplerInterfaceBlock sib = SamplerInterfaceBlock::Builder()
+    // TODO: ideally we'd want these to be constexpr, these are compile time structures.
+
+    static SamplerInterfaceBlock sibPcf = SamplerInterfaceBlock::Builder()
             .name("Light")
             .add("shadowMap",     Type::SAMPLER_2D_ARRAY,   Format::SHADOW, Precision::MEDIUM)
             .add("records",       Type::SAMPLER_2D,         Format::UINT,   Precision::MEDIUM)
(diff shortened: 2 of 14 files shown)


libs/filabridge/src/SibGenerator.cpp
```diff
@@ -16,19 +16,22 @@
...
+SamplerInterfaceBlock const& SibGenerator::getPerViewSib(uint8_t variantKey) noexcept {
     using Type = SamplerInterfaceBlock::Type;
     using Format = SamplerInterfaceBlock::Format;
     using Precision = SamplerInterfaceBlock::Precision;
 
-    // TODO: ideally we'd want this to be constexpr, this is a compile time structure
-    static SamplerInterfaceBlock sib = SamplerInterfaceBlock::Builder()
+    // TODO: ideally we'd want these to be constexpr, these are compile time structures.
+
+    static SamplerInterfaceBlock sibPcf = SamplerInterfaceBlock::Builder()
```
romainguy, line 34: It might be worth having a shared function that fills in the non-specific parameters so we don't have to maintain two lists. Seems like an easy thing to mess up at some point :)

shaders/src/depth_main.vs
```diff
@@ -3,15 +3,17 @@
 void materialVertex(inout MaterialVertexInputs m) { }
 
 void main() {
-#if defined(VERTEX_DOMAIN_DEVICE)
-    gl_Position = getPosition();
-#else
     MaterialVertexInputs material;
```
romainguy, line 6: That's a big change here, should it be only for VSMs?

shaders/src/depth_main.vs
```diff
@@ -3,15 +3,17 @@
...
     initMaterialVertex(material);
     materialVertex(material);
+#if defined(VERTEX_DOMAIN_DEVICE)
+    gl_Position = getPosition();
+#else
     gl_Position = getClipFromWorldMatrix() * getWorldPosition(material);
 #endif
 
+    vertex_worldPosition = material.worldPosition.xyz;
```
romainguy, line 15: This will only be used by VSM variants, should it be #if'd?

shaders/src/shadowing.fs
```diff
@@ -327,3 +350,15 @@ float shadow(const lowp sampler2DArrayShadow shadowMap, const uint layer, const
     return ShadowSample_PCF_High(shadowMap, layer, size, shadowPosition);
 #endif
 }
+
+float shadowVsm(const lowp sampler2DArray shadowMap, const uint layer, const vec3 shadowPosition,
```
romainguy, line 354: You should double check the precision qualifiers here. shadowMap shouldn't be lowp but highp (for 32F depth maps, or mediump otherwise). The shadow position and fragDepth require highp as well.

shaders/src/shadowing.fs
```diff
@@ -306,6 +306,29 @@ float screenSpaceContactShadow(vec3 lightDirection) {
...
+
+float linstep(float a, float b, float v) {
+    return clamp((v - a) / (b - a), 0.0, 1.0);
+}
+
+float reduceLightBleed(float pMax, float amount) {
+    return linstep(amount, 1.0, pMax);
+}
+
+float chebyshevUpperBound(vec2 moments, float mean, float minVariance, float lightBleedReduction) {
```
romainguy, line 321: Check precisions

libs/filabridge/src/SamplerInterfaceBlock.cpp
```diff
@@ -76,7 +76,7 @@ SamplerInterfaceBlock::Builder::~Builder() noexcept = default;
 
 SamplerInterfaceBlock::SamplerInterfaceBlock() = default;
 SamplerInterfaceBlock::SamplerInterfaceBlock(const SamplerInterfaceBlock& rhs) = default;
-SamplerInterfaceBlock::SamplerInterfaceBlock(SamplerInterfaceBlock&& rhs) noexcept /* = default */ {};
+SamplerInterfaceBlock::SamplerInterfaceBlock(SamplerInterfaceBlock&& rhs) noexcept = default;
```
bejado, line 79: Doesn't look like it ever gets called, but not sure why this was commented out.

romainguy: That would be a good question to ask @pixelflinger  :)

shaders/src/depth_main.vs
bejado: That would add another depth variant, which might be acceptable to avoid computing this (and the `getPosition()` above).

libs/filabridge/src/SibGenerator.cpp
bejado: Yeah good point

shaders/src/depth_main.fs
```diff
@@ -13,4 +13,15 @@ void main() {
         discard;
     }
 #endif
+
+#if defined(HAS_VSM)
+    // Since we're rendering from the perspective of the light, frameUniforms.cameraPosition is the
+    // light position, in world space.
+    highp float depth = length(frameUniforms.cameraPosition.xyz - vertex_worldPosition);
```
pixelflinger, line 20: I'm confused by this code. It is calculating the distance from the light to the fragment and it's calling that the "depth". Is that different from the actual depth?

shaders/src/shadowing.fs
```diff
@@ -306,6 +306,29 @@ float screenSpaceContactShadow(vec3 lightDirection) {
...
 
+//------------------------------------------------------------------------------
+// VSM
+//------------------------------------------------------------------------------
+
+float linstep(float a, float b, float v) {
+    return clamp((v - a) / (b - a), 0.0, 1.0);
+}
+
+float reduceLightBleed(float pMax, float amount) {
```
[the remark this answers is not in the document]
bejado: P here stands for "probability" and directly corresponds to the equations listed in the VSM paper: https://www.punkuser.net/vsm/vsm_paper.pdf

shaders/src/depth_main.fs
bejado: By "actual depth", do you mean depth in clip space (`gl_FragDepth`)? We _could_ use `gl_FragDepth`, but we need a linear depth representation. `gl_FragDepth` is only linear in specific cases (non-LiSPM directional lights with orthographic projections).

libs/filabridge/src/SamplerInterfaceBlock.cpp
bejado: Looks like it succeeds now :)

shaders/src/shadowing.fs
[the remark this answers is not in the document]
pixelflinger: I would rename it to `maxProbability` and/or list the link to the paper in the comments

shaders/src/depth_main.fs
pixelflinger: You can easily linearize the depth, it's basically just a `madd`, this is done in `sao.mat`

But what I'm not sure about is that "the distance between the light and the fragment" is not the same than the "difference between the z of the camera and the fragment's" -- so which is the correct one?
So conceptually, what is it that we want? the depth or the distance to the camera?

bejado: Ah I see the confusion. "distance from the fragment to the light" is the metric I'm going for.

> A linear metric with more uniform precision, such as the distance to the light, is more appropriate and stable.

[GPU Gems source](https://developer.nvidia.com/gpugems/gpugems3/part-ii-light-and-shadows/chapter-8-summed-area-variance-shadow-maps)

shaders/src/depth_main.fs
```diff
@@ -17,16 +17,16 @@ void main() {
...
 
     highp float dx = dFdx(depth);
     highp float dy = dFdy(depth);
 
     // Output the first and second depth moments.
     // The first moment is mean depth.
-    // The second moment is depth squared.
+    // The second moment is mean depth squared.
     // These values are retrieved when sampling the shadow map to compute variance.
     highp float bias = 0.25 * (dx * dx + dy * dy);
```
pixelflinger, line 31: is the 0.25 the correct value, or is it a settable?

bejado: It's a constant value. Taken from section 8.4.2: https://developer.nvidia.com/gpugems/gpugems3/part-ii-light-and-shadows/chapter-8-summed-area-variance-shadow-maps

shaders/src/depth_main.vs
bejado: Actually, this would add 2 new depth variants. I'll go ahead and add the variants to make this conditional.. we can always remove the variants in the future without breaking materials if we decide to pay the computational cost instead.

PR #2973, opened 2020-08-18
fix translucent views with custom render targets

### LINKED ISSUE #2802: How to render into a transparent RenderTarget/Texture?
I create a `View` and set a `RenderTarget` which contains a color and depth attachment. 

The texture for the color attachment is created like:

```c++
fila_texture = filament::Texture::Builder()
      .width(settings.width)
      .height(settings.height)
      .levels(1)
      .usage(filament::Texture::Usage::COLOR_ATTACHMENT | filament::Texture::Usage::SAMPLEABLE)
      .format(filament::Texture::InternalFormat::RGBA16F)
      .build(settings.ctx->getEngineRef());
```

I've set the clear color using `ClearOptions` of the `Renderer` to an arbitary color with an alpha of 0.

Is this the correct way of rendering into a transparent buffer? 

I'm asking because the result I get seems to have an alpha of 1.0.

pixelflinger:
Marking a view with a custom render target will now disable blending
for that view but require an alpha channel for intermediate buffers.
Blending is disabled because it is meaningless.

fixes #2802

### DIFF
--- a/android/filament-android/src/main/java/com/google/android/filament/View.java
+++ b/android/filament-android/src/main/java/com/google/android/filament/View.java
@@ -713,6 +713,11 @@ public class View {
      * SwapChain associated with the engine.
      * </p>
      *
+     * <p>
+     * A view with a custom render target cannot rely on Renderer.ClearOptions, which only applies
+     * to the SwapChain. Such view can use a Skybox instead.
+     * </p>
+     *
      * @param target render target associated with view, or null for the swap chain
      */
     public void setRenderTarget(@Nullable RenderTarget target) {
--- a/filament/include/filament/View.h
+++ b/filament/include/filament/View.h
@@ -61,8 +61,6 @@ class Viewport;
  */
 class UTILS_PUBLIC View : public FilamentAPI {
 public:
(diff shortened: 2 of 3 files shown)


android/filament-android/src/main/java/com/google/android/filament/View.java
```diff
@@ -713,6 +713,11 @@ public void setShadowsEnabled(boolean enabled) {
      * SwapChain associated with the engine.
      * </p>
      *
+     * <p>
+     * A view with a custom render target cannot rely on Renderer.ClearOptions, which only applies
```
romainguy, line 717: That's a bit weird API-wise :/

pixelflinger: Well, I think it makes sense actually.

The ClearOptions clear the renderer -- not the view. When the view brings it own render target, it's just because it's an intermediate draw, that will ultimately be rendered (e.g. as a texture) into another "prime" view.

I think it would be weird for the Renderer.ClearOptions to clear some other render target.

We have a mechanism to clear a View now, and that's Skybox.

PR #2995, opened 2020-08-23
use generated SH in IBL::loadFromKtx

RYDB3RG:


RYDB3RG:
> @googlebot I signed it!

PR #3004, opened 2020-08-25
Add ShadowType API and VSM rendering logic

bejado:


filament/src/RenderPass.cpp
```diff
@@ -376,14 +376,16 @@ void RenderPass::generateCommandsImpl(uint32_t extraFlags,
     materialVariant.setDirectionalLighting(renderFlags & HAS_DIRECTIONAL_LIGHT);
     materialVariant.setDynamicLighting(renderFlags & HAS_DYNAMIC_LIGHTING);
     materialVariant.setFog(renderFlags & HAS_FOG);
+    materialVariant.setVsm(renderFlags & HAS_VSM && hasShadowing);
```
prideout, line 379: Clarify precedence with parens maybe.

filament/src/ShadowMapManager.cpp
```diff
@@ -91,31 +91,71 @@ void ShadowMapManager::prepare(FEngine& engine, DriverApi& driver, SamplerGroup&
...
     // destroy the current rendertargets and texture
     destroyResources(driver);
 
-    mShadowMapTexture = driver.createTexture(
-            SamplerType::SAMPLER_2D_ARRAY, 1, mTextureFormat, 1, dim, dim, layersNeeded,
-            TextureUsage::DEPTH_ATTACHMENT |TextureUsage::SAMPLEABLE);
+    // Shadows are either all PCF or all VSM.
+
+    if (mUseVsm) {
+        // TODO: support 16-bit VSM depth textures.
```
romainguy, line 107: If I remember correctly we can get away with using just the 16 bit variant right?

bejado: I haven't tried that yet, but it's on my list. Yeah, if 16 bits is sufficient, then we could just use that unconditionally.

romainguy: https://mynameismjp.wordpress.com/2013/09/10/shadow-maps/ has info on that I believe

PR #3005, opened 2020-08-26
Improvements to MSAA handling

pixelflinger:


pixelflinger:
note to reviewers: I recommend you look at the 5 commits separately.

filament/backend/include/backend/DriverEnums.h
```diff
@@ -505,6 +505,20 @@ enum class TextureSwizzle {
     CHANNEL_3
 };
 
+//! returns whether this format a depth format
+static constexpr bool isDepthFormat(TextureFormat format) noexcept {
```
prideout, line 509: Note that several backends, including GL, can now use this in their impl of `isTextureFormatMipmappable()`.

pixelflinger:
@bejado yes I'm setting the target bits on attachments that don't exist, assuming they'll be ignored -- there is nothing to do anyways.

filament/src/Renderer.cpp
```diff
@@ -716,7 +724,14 @@ FrameGraphId<FrameGraphTexture> FRenderer::colorPass(FrameGraph& fg, const char*
...
                             .width = colorBufferDesc.width,
                             .height = colorBufferDesc.height,
-                            .format = TextureFormat::DEPTH24
+                            // If the color attachment requested MS, we assume this means the MS buffer
+                            // must be kept, and for that reason we allocate the depth buffer with MS
+                            // as well. On the other hand, if the color attachment was allocated without
+                            // MS, no need to allocate the depth buffer with MS, if the RT is MS,
+                            // the tile depth buffer will be MS, but it'll be resolved to single
+                            // sample automatically -- which is what we want.
+                            .samples = colorBufferDesc.samples,
```
prideout, line 733: The right hand side of this assignment isn't a true sample count because "0" is a valid value, meaning "auto".

The left hand side is an actual sample count that should always be 1 or greater.

PR #3007, opened 2020-08-26
Handle blitting only depth in Metal backend

### LINKED ISSUE #2984: Refractive material and DOF with 2+ anti-alias samples crash in Metal
**Describe the bug**
Transparent shader with screen-space refraction and depth of field with 2+ anti-alias samples crash the engine when using the Metal backend.

```
Error message: Execution of the command buffer was aborted due to an error during execution. Caused GPU Hang Error (IOAF code 3)
validateAttachmentOnDevice:490: failed assertion `Memoryless attachment cannot be loaded as it have no memory to load from.'
```

Tested with hello-pbr from the samples.
Tested with filament 1.8.0 and 1.8.1.

**To Reproduce**
1. Use a transparent shader with screen-space refractions on the subject model. Both solid and thin types produce the bug.
2. Turn on depth of field effect. (default parameters)
3. Set viewSamples to or higher than 2.

**Expected behavior**
Rendering without crash.

**Smartphone:**
 - Device: iPad Pro 12.9" 2018
 - OS: iOS 13.6

bejado:
Fixes #2984

### DIFF
--- a/filament/backend/src/metal/MetalBlitter.h
+++ b/filament/backend/src/metal/MetalBlitter.h
@@ -40,12 +40,20 @@ public:
         struct Attachment {
             id<MTLTexture> color = nil;
             id<MTLTexture> depth = nil;
-            MTLRegion region;
+            MTLRegion region = {};
             uint8_t level = 0;
         };
 
         Attachment source, destination;
         SamplerMagFilter filter;
+
+        bool blitColor() const {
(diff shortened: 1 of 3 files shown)


bejado:
@prideout Mathias' latest changes add a depth-only blit. I turned on MSAA, DoF, and SSR with material_sandbox.

PR #3016, opened 2020-08-27
Do not pass a zero sample count to the backend.

prideout:
FrameGraphTexture::Descriptor uses "0" to mean "auto" but the
sample count that is passed to the backend should always be 1 or
greater.

See also #2957.

### DIFF
--- a/filament/src/ResourceAllocator.cpp
+++ b/filament/src/ResourceAllocator.cpp
@@ -124,6 +124,10 @@ backend::TextureHandle ResourceAllocator::createTexture(const char* name,
     }
 #endif
 
+    // The frame graph descriptor uses "0" to mean "auto" but the sample count that is passed to the
+    // backend should always be 1 or greater.
+    samples = samples ? samples : uint8_t(1);
+
     // do we have a suitable texture in the cache?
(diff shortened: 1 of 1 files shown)


filament/src/Renderer.cpp
```diff
@@ -731,7 +731,7 @@ FrameGraphId<FrameGraphTexture> FRenderer::colorPass(FrameGraph& fg, const char*
                             // MS, no need to allocate the depth buffer with MS, if the RT is MS,
                             // the tile depth buffer will be MS, but it'll be resolved to single
                             // sample automatically -- which is what we want.
-                            .samples = colorBufferDesc.samples,
+                            .samples = colorBufferDesc.samples == 0 ? uint8_t(1) : colorBufferDesc.samples,
```
romainguy, line 734: Could we make sure the sample count is never 0 in the description instead? It seems error prone if we have to make this check everywhere we read the sample count.

pixelflinger: Yeah I think we should fix that in the ResourceAllocator instead.

prideout:
Good point, fixed.

PR #3019, opened 2020-08-28
Migrate the SwiftShader build to Vulkan.

prideout:
This also adds a -t flag to the easy build script for convenience.

SwiftShader on Vulkan can now be used as follows.

    git clone https://github.com/google/swiftshader.git
    cd swiftshader/build
    cmake .. &&  make -j && export SWIFTSHADER_LD_LIBRARY_PATH=`pwd`
    cd ../filament
    ./build.sh -t debug && ./out/cmake-debug/samples/gltf_viewer -a vulkan

Note that `third_party/swiftshader` does not need to exist since we already have Vulkan headers. I will remove it in a subsequent PR.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -15,7 +15,7 @@ option(FILAMENT_ENABLE_JAVA "Compile Java projects, requires a JDK and the JAVA_
 
 option(FILAMENT_USE_EXTERNAL_GLES3 "Experimental: Compile Filament against OpenGL ES 3" OFF)
 
-option(FILAMENT_USE_SWIFTSHADER "Experimental: Compile Filament against SwiftShader" OFF)
+option(FILAMENT_USE_SWIFTSHADER "Compile Filament against SwiftShader" OFF)
 
 option(FILAMENT_ENABLE_LTO "Enable link-time optimizations if supported by the compiler" OFF)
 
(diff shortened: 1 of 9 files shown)


CMakeLists.txt
```diff
@@ -484,7 +483,11 @@ endif()
 # "bluevk" and "samples" targets.
 # ==================================================================================================
 
-if (FILAMENT_SUPPORTS_VULKAN AND APPLE)
+if (FILAMENT_USE_SWIFTSHADER)
```
bejado, line 486: Should we verify that `FILAMENT_SUPPORTS_VULKAN` is set here? SwiftShader is only useful with Vulkan support, correct?

filament/backend/src/Platform.cpp
```diff
@@ -35,7 +33,7 @@
         #include "vulkan/PlatformVkCocoaTouch.h"
     #endif
 #elif defined(__APPLE__)
-    #ifndef FILAMENT_USE_EXTERNAL_GLES3
+    #if !defined(FILAMENT_USE_EXTERNAL_GLES3) && !defined(FILAMENT_USE_SWIFTSHADER)
```
bejado, line 36: Why is this change is only necessary for APPLE platforms?

CMakeLists.txt
prideout: Good idea, added an error message here.

filament/backend/src/Platform.cpp
prideout: Excellent catch, added this for all desktop platforms.

PR #3020, opened 2020-08-28
Fix a few issues with alpha-masked objects

pixelflinger:
- don't apply SSAO to blended objects
  these are not drawn into the structure buffer, so they don't 
  participate in SSAO.

- remove unused code that was needed for the depth prepass, which
  simplifies handling of blended objects.

- always treat alpha-masked objects as opaque.

### DIFF
--- a/filament/src/RenderPass.cpp
+++ b/filament/src/RenderPass.cpp
@@ -246,7 +246,7 @@ void RenderPass::recordDriverCommands(FEngine::DriverApi& driver, const Command*
 /* static */
 UTILS_ALWAYS_INLINE // this function exists only to make the code more readable. we want it inlined.
 inline              // and we don't need it in the compilation unit
-void RenderPass::setupColorCommand(Command& cmdDraw, bool hasDepthPass,
+void RenderPass::setupColorCommand(Command& cmdDraw,
         FMaterialInstance const* const UTILS_RESTRICT mi, bool inverseFrontFaces) noexcept {
 
     FMaterial const * const UTILS_RESTRICT ma = mi->getMaterial();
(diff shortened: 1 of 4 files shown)


shaders/src/ambient_occlusion.fs
```diff
@@ -17,6 +17,7 @@ float unpack(vec2 depth) {
 }
 
 float evaluateSSAO() {
+#if defined(BLEND_MODE_OPAQUE) || defined(BLEND_MODE_MASKED)
```
romainguy, line 20: This is still something I'd like to think about. It seems correct but it bothers me.

pixelflinger: In what way? is it the implementation here, or just the idea of not applying SSAO to objets that are not in the depth buffer?

filament/src/RenderPass.cpp
```diff
@@ -260,8 +260,10 @@ void RenderPass::setupColorCommand(Command& cmdDraw, bool hasDepthPass,
     keyBlending |= uint64_t(Pass::BLENDED);
     keyBlending |= uint64_t(CustomCommand::PASS);
 
+    BlendingMode blendingMode = ma->getBlendingMode();
     bool hasScreenSpaceRefraction = ma->getRefractionMode() == RefractionMode::SCREEN_SPACE;
-    bool hasBlending = !hasScreenSpaceRefraction && ma->getRasterState().hasBlending();
+    bool blendingCommand = !hasScreenSpaceRefraction &&
```
bejado, line 265: Since it's a `bool`, maybe rename to a verb phrase? I liked `hasBlending`

PR #3022, opened 2020-08-28
Changed ifdef WIN32 to WIN32 or _WIN32

toppa102:
Changed if defined WIN32 to if defined WIN32 or _WIN32 because MSVC doesn't define WIN32 or WIN64.

### DIFF
--- a/libs/utils/include/utils/compiler.h
+++ b/libs/utils/include/utils/compiler.h
@@ -189,7 +189,7 @@
 
 
 // ssize_t is a POSIX type.
-#if defined(WIN32)
+#if defined(_WIN32)
 #include <BaseTsd.h>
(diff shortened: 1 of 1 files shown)


toppa102:
@googlebot I signed it!

libs/utils/include/utils/compiler.h
```diff
@@ -189,7 +189,7 @@
 
 
 // ssize_t is a POSIX type.
-#if defined(WIN32)
```
romainguy, line 192: There are a lot more places where we use `WIN32`, this should at least be `#if defined(WIN32) || defined(_WIN32)`

toppa102: Yeah i changed it to WIN32 or _WIN32.

PR #3025, opened 2020-08-29
fix refraction + MSAA + TAA (depth resolve)

pixelflinger:
the depth buffer wasn't resolved in this case. We fix this by always
creating a depth buffer resolve pass -- which is a lot more robust.
this pass automatically culled if the depth is not used (i.e. TAA or 
DoF are off). The pass is a no-op if resolving is not needed.


filament/src/Renderer.cpp
```diff
@@ -442,6 +442,13 @@ void FRenderer::renderJob(ArenaScope& arena, FView& view) {
             refractionPass(fg, config, colorGradingConfigForColor, pass, view);
     FrameGraphId<FrameGraphTexture> input = colorPassOutput;
 
+    // resolve depth -- which might be needed because of TAA. This pass will be culled
```
bejado, line 445: Maybe also note that DOF needs it as well?

PR #3026, opened 2020-08-29
Changed TargetBufferInfo default constructor and a calculation in algorithm.h

toppa102:
On MSVC Visual Studio 2019 Win 10 the default constructor of TargetBufferInfo creates the error: 
`C2512	'filament::backend::Handle<filament::backend::HwTexture>': no appropriate default constructor available`
but this change fixes it (there's probably other headers where this same thing occurs).

The calculation in algorithm.h creates the error:
`Error	C4146	unary minus operator applied to unsigned type, result still unsigned`
But using the other calculation mentioned in the comment makes it compile on MSVC.

### DIFF
--- a/filament/backend/include/backend/TargetBufferInfo.h
+++ b/filament/backend/include/backend/TargetBufferInfo.h
@@ -49,7 +49,7 @@ public:
         // for 3D textures
         uint16_t layer = 0;
     };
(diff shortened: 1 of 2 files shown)


toppa102:
Well now it broke everything...

PR #3029, opened 2020-09-01
optionally compress highlights before bloom

pixelflinger:
This avoid unrealistic looking bloom halos from very strong highlights.
This is enabled by default.

### DIFF
--- a/android/filament-android/src/main/java/com/google/android/filament/View.java
+++ b/android/filament-android/src/main/java/com/google/android/filament/View.java
@@ -283,6 +283,12 @@ public class View {
          * enable or disable bloom
          */
         public boolean enabled = false;
+
+        /**
+         * limit highlights to this value before bloom. Use +inf for no limiting.
+         * minimum value is 10.0.
+         */
+        public float highlight = 1000.0f;
     }
 
     /**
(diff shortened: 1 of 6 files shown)


android/filament-android/src/main/java/com/google/android/filament/View.java
```diff
@@ -283,6 +283,12 @@
          * enable or disable bloom
          */
         public boolean enabled = false;
+
+        /**
+         * limit highlights to this value before bloom. Use +inf for no limiting.
+         * minimum value is 4.0.
```
romainguy, line 289: Why 4? Maybe also explain that this is a post-exposure multiplier?

pixelflinger: I chose 4 because some value needs to be chosen. I even wanted to pick 10, because it makes no sense to pick a value that's too low. Technically 1.0 is the real minimum, but that would just disable bloom (not that a lot is left with 4).

romainguy: Yeah but I was wondering if it wouldn't make more sense to use the top of our tonemapping curve? (10)

pixelflinger: yeah okay. I was about to use 10 actually.

PR #3031, opened 2020-09-01
New API to create an Engine asynchronously

pixelflinger:
Because initializing the backend can be relatively slow, creating an
Engine can block the main thread, is this is a problem, this new API
can be used instead; while a bit more complex, it allow the engine to
be initialized asynchronously.

### DIFF
--- a/filament/include/filament/Engine.h
+++ b/filament/include/filament/Engine.h
@@ -187,6 +187,66 @@ public:
     static Engine* create(Backend backend = Backend::DEFAULT,
             Platform* platform = nullptr, void* sharedGLContext = nullptr);
 
+#if UTILS_HAS_THREADING
+    /**
+     * A callback used with Engine::createAsync() called once the engine is initialized and it is
+     * safe to call Engine::getEngine(token). This callback is invoked from an arbitrary worker
+     * thread. Engine::getEngine() CANNOT be called from that thread, instead it must be called
+     * from the same thread than Engine::createAsync() was called from.
+     *
+     * @param user   User provided parameter given in createAsync().
+     *
+     * @param token  An opaque token used to call Engine::getEngine().
+     */
+    using CreateCallback = void(void* user, void* token);
+
+    /**
+     * Creates an instance of Engine asynchronously
+     *
+     * @param callback          Callback called once the engine is initialized and it is safe to
+     *                          call Engine::getEngine.
+     *
+     * @param user              A user provided pointer that is given back to callback unmodified.
+     *
+     * @param backend           Which driver backend to use.
+     *
+     * @param platform          A pointer to an object that implements Platform. If this is
+     *                          provided, then this object is used to create the hardware context
+     *                          and expose platform features to it.
+     *
+     *                          If not provided (or nullptr is used), an appropriate Platform
+     *                          is created automatically.
+     *
+     *                          All methods of this interface are called from filament's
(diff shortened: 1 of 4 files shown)


filament/include/filament/Engine.h
```diff
@@ -187,6 +187,59 @@ class UTILS_PUBLIC Engine {
...
 
+#if UTILS_HAS_THREADING
+
+    /**
+     * A callback used with Engine::createAsync() called once the engine is initialized and it is
+     * safe to call Engine::getEngine(token).
+     * @param user   User provided parameter given in createAsync().
+     * @param token  An opaque token used to call Engine::getEngine().
+     */
+    using CreateCallback = void(void* user, void* token);
```
romainguy, line 198: API design question: why can't we just pass the Engine* in the callback instead of a token?

filament/include/filament/Engine.h
```diff
@@ -187,6 +187,59 @@ class UTILS_PUBLIC Engine {
     static Engine* create(Backend backend = Backend::DEFAULT,
             Platform* platform = nullptr, void* sharedGLContext = nullptr);
 
+#if UTILS_HAS_THREADING
+
+    /**
+     * A callback used with Engine::createAsync() called once the engine is initialized and it is
```
romainguy, line 193: Maybe mention that the callback will be invoked on an arbitrary worker thread and that the app should handle this with care.

filament/src/Engine.cpp
```diff
@@ -105,6 +102,47 @@ FEngine* FEngine::create(Backend backend, Platform* platform, void* sharedGLCont
...
+}
+
+FEngine* FEngine::getEngine(void* token) {
+
+    FEngine* instance = static_cast<FEngine*>(token);
+
+    ASSERT_PRECONDITION(instance->mMainThreadId == std::this_thread::get_id(),
+            "Engine::createAsync() and Engine::getEngine() must be called on the same thread.");
+
+    // we use mResourceAllocator has a proxy for "am I already initialized"
```
romainguy, line 131: has->as

filament/include/filament/Engine.h
romainguy: It's particularly important since they have to call getEngine() from the same thread as createAsync() so the app will need to do a bit of thread hopping here.

filament/include/filament/Engine.h
pixelflinger: Well, so it's because I have no way to call the callback from the main thread -- which is what I wanted to do ideally.
But also because I need to run some code on the main thread after the backend is initialized but before I return Engine*.

romainguy: Note that the Java bindings will be able to invoke the user callback on main thread in the end.

How expensive is the code that needs to run on the main thread when calling getEngine()?

pixelflinger: So I asked Dan and they don't need java bindings for this, so I'm just thinking of not providing them.

The stuff that needs to run on the main thread is mostly creating backend objects, so it's just writing commands in the command stream. Currently it also generates the color grading LUT. More refactoring is needed if we want to do this asynchronously too.

romainguy: Sounds good. The LUT is multithreaded btw so not a big deal, but we should check again how long it takes in wall time because that will contribute to the "latency" our customers are complaining about.

PR #3050, opened 2020-09-08
CoroutineScope job should be canceled before destory

daichen:
tombstone happens if run destory while loadModelGltfAsync in ModelViewer.kt

To Reproduce
1.loadModelGltfAsync
2.detach from the winodw quickly
3.tombstone ..

### DIFF
--- a/android/filament-utils-android/src/main/java/com/google/android/filament/utils/ModelViewer.kt
+++ b/android/filament-utils-android/src/main/java/com/google/android/filament/utils/ModelViewer.kt
@@ -85,6 +85,8 @@ class ModelViewer(val engine: Engine) : android.view.View.OnTouchListener {
     private var surfaceView: SurfaceView? = null
     private var textureView: TextureView? = null
 
+    private var fetchResJob: Job? = null
+
     private val renderer: Renderer
     private var swapChain: SwapChain? = null
     private var assetLoader: AssetLoader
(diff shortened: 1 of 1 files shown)


android/filament-utils-android/src/main/java/com/google/android/filament/utils/ModelViewer.kt
```diff
@@ -269,6 +272,7 @@ class ModelViewer(val engine: Engine) : android.view.View.OnTouchListener {
         view.addOnAttachStateChangeListener(object : android.view.View.OnAttachStateChangeListener {
             override fun onViewAttachedToWindow(v: android.view.View?) {}
             override fun onViewDetachedFromWindow(v: android.view.View?) {
+                fetchResJob?.cancel()
```
romainguy, line 275: destroyModel() is called below which will call cancel on the job, is it necessary to cancel here as well?

android/filament-utils-android/src/main/java/com/google/android/filament/utils/ModelViewer.kt
```diff
@@ -85,6 +85,8 @@ class ModelViewer(val engine: Engine) : android.view.View.OnTouchListener {
     private var surfaceView: SurfaceView? = null
     private var textureView: TextureView? = null
 
+    private var fetchResJob: Job? = null
```
romainguy, line 88: Please rename to `fetchResourcesJob`

romainguy (changes requested):
Thank you for your contribution! I only have a couple of comments

daichen:
sorry, I'm not familiar with git hub, I will close this request and create a new request

PR #3053, opened 2020-09-08
rework a bit dithering noise

pixelflinger:
- don't use frameUniform.time as the temporal noise seed because
  this has a cycle to it. It's not very visible on dithering but
  obvious when used for other things.

- don't use fract(time) since time was already in [0,1] -- though it's
  moot now.

- instead of time pass a random number in [0,1] generated on the CPU
  each frame.
  This will also allow to control temporal noise more easily if
  needed in the future (e.g. for accessibility or rendering tests)

### DIFF
--- a/shaders/src/dithering.fs
+++ b/shaders/src/dithering.fs
@@ -39,10 +39,9 @@ float interleavedGradientNoise(highp vec2 n) {
 // Dithering
 //------------------------------------------------------------------------------
 
-vec4 Dither_InterleavedGradientNoise(vec4 rgba) {
+vec4 Dither_InterleavedGradientNoise(vec4 rgba, const highp float temporalNoise01) {
     // Jimenez 2014, "Next Generation Post-Processing in Call of Duty"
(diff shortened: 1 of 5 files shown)


shaders/src/dithering.fs
```diff
@@ -53,10 +52,10 @@ vec4 Dither_InterleavedGradientNoise(vec4 rgba) {
     return rgba + vec4(noise / 255.0);
 }
 
-vec4 Dither_TriangleNoise(vec4 rgba) {
+vec4 Dither_TriangleNoise(vec4 rgba, const highp float temporalNoise01) {
```
romainguy, line 55: Why is it called 01?

romainguy (approved):

shaders/src/dithering.fs
pixelflinger: it's expected to be between 0 and 1 values. (e.g. as opposed to [-1,1] or [-0.5, 0.5]

PR #3054, opened 2020-09-08
attempt at a better noise for SSAO

pixelflinger:
We add noise do the radius of the samples on the spiral 
(instead of adding noise only to the angle), this reduces a visible
"echo", especially at higher radii.

before:
before

after:
after


romainguy (approved):
That's a cool idea 👍

PR #3056, opened 2020-09-09
CoroutineScope job should be canceled before destory

daichen:
tombstone happens if run destory while loadModelGltfAsync in ModelViewer.kt

To Reproduce
1.loadModelGltfAsync
2.detach from the winodw quickly
3.tombstone ..


romainguy (approved):
Thank you!

PR #3063, opened 2020-09-10
fix memory leak issue in reource loader

daichen:
shoule release texels before destory or reload

To Reproduce

1. start android sample-gltf-viewer
2. keep on double tabing the activity to reload the model
3. monitor the memory useage

### DIFF
--- a/libs/gltfio/src/ResourceLoader.cpp
+++ b/libs/gltfio/src/ResourceLoader.cpp
@@ -927,10 +927,10 @@ void ResourceLoader::Impl::computeTangents(FFilamentAsset* asset) {
 }
 
 ResourceLoader::Impl::~Impl() {
-    releaseTexels();
     if (mDecoderRootJob) {
         mEngine->getJobSystem().waitAndRelease(mDecoderRootJob);
     }
+    releaseTexels();
 }
 
 void ResourceLoader::applySparseData(FFilamentAsset* asset) const {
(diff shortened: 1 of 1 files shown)


libs/gltfio/src/ResourceLoader.cpp
```diff
@@ -907,6 +927,7 @@ void ResourceLoader::Impl::computeTangents(FFilamentAsset* asset) {
 }
 
 ResourceLoader::Impl::~Impl() {
+    releaseTexels();
```
romainguy, line 930: This needs to be done after waiting for the decoder job to finish below.

romainguy (changes requested):
Thanks for the fix, please see comments below.

prideout:
Search this file for `PixelBufferDescriptor` and note that the `texels` pointer should be freed when the PixelBufferDescriptor callback is triggered.  If you're seeing a leak, can you determine why the PixelBufferDescriptor callback isn't being triggered?

pixelflinger:
A quick inspection of the code doesn't explain how the callback could not be called after `setImage` has been called. So we need to investigate this more.

pixelflinger:
okay more details about this -- the leak happens only when triggering a reload (double-tap) while a reload is already happening.  My guess is that there is a code path where malloc() is called and `setImage` is not.

daichen:
It is because com.google.android.filament.utils.ModelViewer load the resource use asyncBeginLoad, it will not call  uploadPendingTextures in createTextures.
After ModelViewer call asyncUpdateLoad in render(), it will send image to gpu and set callback free the memory

daichen:
please reivew https://github.com/google/filament/pull/3063/commits/fa1bbe5c36e428be834f4d3aff71242c2ed4c55b

prideout:
I am back from vacation and will take a look at this soon.

prideout:
Thanks for bringing this to our attention!  I have an alternative fix up (#3109) that handles this in addition to a couple other missing cancellation features (e.g. an assert in DependencyGraph, zombie decoder threads)

PR #3064, opened 2020-09-10
Remove Google-style line directives driver-side

### LINKED ISSUE #3059: Runtime shader compilation error on Android devices that don't support GL_GOOGLE_cpp_style_line_directive
I think this another instance of [this problem](https://github.com/google/filament/issues/2857), which was fixed [by explicitly removing the unsupported directives](https://github.com/google/filament/pull/2860).

The problem occurs for me when I link to the debug version of the Filament AAR. The _skybox.mat_ material has been compiled into the AAR [without optimization](https://github.com/google/filament/blob/main/CMakeLists.txt#L409), which means the unsupported directives remain and it gives the following error:

```
E/Filament: COMPILE ERROR: 
    0:4: P0003: Warning: Extension 'GL_GOOGLE_cpp_style_line_directive' not supported
    0:9: L0001: Unknown character '"'(34)
    0:9: L0001: Unknown character '"'(34)
    0:44: L0001: Unknown character '"'(34)
    0:44: L0001: Unknown character '"'(34)
    0:207: L0001: Unknown character '"'(34)
    0:207: L0001: Unknown character '"'(34)
    0:347: L0001: Unknown character '"'(34)
    0:347: L0001: Unknown character '"'(34)
    0:370: L0001: Unknown character '"'(34)
    0:370: L0001: Unknown character '"'(34)
    0:446: L0001: Unknown character '"'
    1:   #version 300 es
    2:   
    3:   #define TARGET_MOBILE
    4:   #extension GL_GOOGLE_cpp_style_line_directive : enable
    5:   
    6:   precision highp float;
    7:   precision highp int;
    8:   #if defined(GL_GOOGLE_cpp_style_line_directive)
    9:   #line 0 "common_types.fs"
    ....
```

bejado:
Fixes #3059

### DIFF
--- a/filament/backend/include/private/backend/BackendUtils.h
+++ b/filament/backend/include/private/backend/BackendUtils.h
@@ -22,6 +22,22 @@ namespace filament {
 namespace backend {
 
 /**
+ * Returns true if the shader string requests the Google-style line directive extension.
+ */
+bool requestsGoogleLineDirectivesExtension(const char* shader, size_t length) noexcept;
+
+/**
+ * Edit a GLSL shader string in-place so any Google-style line directives are turned into regular
+ * line directives.
+ *
+ * E.g.:
+ * #line 100 "foobar.h"
+ * is transformed to (_ denotes a space)
+ * #line 100 __________
+ */
+void removeGoogleLineDirectives(char* shader, size_t length) noexcept;
+
+/**
  * Returns the number of bytes per pixel for the given format. For compressed texture formats,
  * returns the number of bytes per block.
  */
--- a/filament/backend/src/BackendUtils.cpp
+++ b/filament/backend/src/BackendUtils.cpp
@@ -16,8 +16,52 @@
 
 #include "private/backend/BackendUtils.h"
 
+#include <string_view>
+
 namespace filament {
 namespace backend {
+
+bool requestsGoogleLineDirectivesExtension(const char* shader, size_t length) noexcept {
+    std::string_view s(shader, length);
+    return s.find("GL_GOOGLE_cpp_style_line_directive") != std::string_view::npos;
+}
+
+void removeGoogleLineDirectives(char* shader, size_t length) noexcept {
+    std::string_view s(shader, length);
+
+    size_t pos = 0;
+    while (true) {
+        pos = s.find("#line", pos);
+        if (pos == std::string_view::npos) {
+            break;
+        }
+
+        pos += 5;
+
+        bool googleStyleDirective = false;
+        size_t len = 0;
+        size_t start = 0;
+        while (pos < length) {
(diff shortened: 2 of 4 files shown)


filament/backend/include/private/backend/BackendUtils.h
```diff
@@ -21,6 +21,22 @@
...
+bool requestsGoogleLineDirectivesExtension(const char* shader, size_t length) noexcept;
+
+/**
+ * Edit a GLSL shader string in-place so any Google-style line directives are turned into regular
+ * line directives.
+ *
+ * E.g.:
+ * #line 100 "foobar.h"
+ * is transformed to (_ denotes a space)
+ * #line 100 __________
```
romainguy, line 36: Smart!

filament/backend/src/BackendUtils.cpp
```diff
@@ -16,9 +16,53 @@
 
 #include "private/backend/BackendUtils.h"
 
+#include <string_view>
+
 namespace filament {
 namespace backend {
 
+bool requestsGoogleLineDirectivesExtension(const char* shader, size_t length) noexcept {
+    std::string_view s(shader, length);
```
romainguy, line 25: Could we just assume a shorter length? i.e. the directive will be in the first 1 KiB or less? It seems like there's no reason to parse the entire shader when the directive is not present (is it ever not present though? )

filament/backend/src/BackendUtils.cpp
```diff
@@ -16,9 +16,53 @@
...
+}
+
+void removeGoogleLineDirectives(char* shader, size_t length) noexcept {
+    std::string_view s(shader, length);
+
+    size_t pos = 0;
+    while (true) {
+        pos = s.find("#line", pos);
+        if (pos == std::string_view::npos) {
+            break;
```
romainguy, line 36: Doesn't this mean you will only fix the first #line directive?

romainguy (approved):

filament/backend/src/BackendUtils.cpp
bejado: It's not present in optimized GLSL. Yeah, I can limit this to the first 1 KiB.

filament/backend/src/BackendUtils.cpp
bejado: No, this breaks when there are no more #line directives in the string.

pixelflinger (approved):

filament/backend/src/BackendUtils.cpp
romainguy: Ah right of course. I keep thinking you're doing processing line by line.

Zenix27 (approved):

allenx (approved):
approved.

PR #3073, opened 2020-09-11
Render shadow passes with frame graph

bejado:


filament/src/Renderer.cpp
```diff
@@ -379,6 +359,22 @@ void FRenderer::renderJob(ArenaScope& arena, FView& view) {
     // TODO: the scaling should depends on all passes that need the structure pass
     ppm.structure(fg, pass, svp.width, svp.height, aoOptions.resolution);
 
+    // Apply the TAA jitter to everything after the structure pass, starting with the color pass.
+    if (taaOptions.enabled) {
```
bejado, line 363: @pixelflinger, I read your FIXME comment and I think I did this correctly, but please advise

filament/src/ShadowMapManager.cpp
```diff
@@ -193,47 +71,166 @@ void ShadowMapManager::addSpotShadowMap(size_t lightIndex) noexcept {
...
+                    } else {
+                        renderTarget.attachments = { {}, { data.shadow, 0u, i } };
+                        renderTarget.clearFlags = TargetBufferFlags::DEPTH;
+                    }
+                    data.rt[i] = builder.createRenderTarget("Shadow RT", renderTarget);
+                }
+            },
+            [=, passes = std::move(passes), &view](FrameGraphPassResources const& resources,
+                    auto const& data, DriverApi& driver) mutable {
+                if (UTILS_UNLIKELY(fillWithCheckerboard)) {
```
pixelflinger, line 167: you shouldn't have to do this, because this pass should be culled automatically. Is it not the case?

pixelflinger (changes requested):

filament/src/Renderer.cpp
```diff
@@ -686,6 +683,7 @@ FrameGraphId<FrameGraphTexture> FRenderer::colorPass(FrameGraph& fg, const char*
                 TargetBufferFlags clearColorFlags = config.clearFlags & TargetBufferFlags::COLOR;
                 data.clearColor = config.clearColor;
 
+                data.shadow = blackboard.get<FrameGraphTexture>("shadow");
```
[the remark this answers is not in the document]
bejado: Yup, good point.

filament/src/ShadowMapManager.cpp
bejado: Let me check. I guess it should be culled because I've only specified "write" in the "Shadow Debug Pattern Pass".

filament/src/ShadowMapManager.cpp
```diff
@@ -193,47 +71,166 @@ void ShadowMapManager::addSpotShadowMap(size_t lightIndex) noexcept {
...
-        mRenderPassParams.viewport = viewport;
-        map.getShadowMap()->render(driver, mRenderTargets[currentRt],
-                viewport, view.getVisibleDirectionalShadowCasters(), pass, mRenderPassParams, view);
+        map.getShadowMap()->render(driver, view.getVisibleDirectionalShadowCasters(), pass, view);
+
+        assert(map.getLayout().layer < mTextureRequirements.layers);
+        passes.emplace_back(&map, pass);
     }
-    for (size_t i = 0; i < mSpotShadowMaps.size(); i++, currentRt++) {
+    for (size_t i = 0; i < mSpotShadowMaps.size(); i++) {
```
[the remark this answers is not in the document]
bejado: Here, no, because of `VISIBLE_SPOT_SHADOW_CASTER_N(i)`

pixelflinger (approved):

PR #3074, opened 2020-09-12
Improvements to SSAO quality

pixelflinger:
- simulate a standard deviation of 8 by skipping every other pixel
  when blurring the AO buffer at LOW or MEDIUM quality. This creates
  a faint checker pattern, but does look much better than the low
  frequency noise that is introduced otherwise.

- tweak AO parameters to give a more uniform and smooth AO effect by
  default. This also makes AO even more consistant between quality
  modes.

- add a simple dithering stage to AO blur to help eliminate banding
  artifacts at higher sample counts.

before:
Screen Shot 2020-09-11 at 21 17 52
Screen Shot 2020-09-11 at 22 22 37

after:

Screen Shot 2020-09-11 at 21 18 17
Screen Shot 2020-09-11 at 22 22 12

### DIFF
--- a/filament/src/materials/ssao/sao.mat
+++ b/filament/src/materials/ssao/sao.mat
@@ -216,7 +216,7 @@ fragment {
             tapPosition = angleStep * tapPosition;
         }
 
-        float ao = max(0.0, 1.0 - occlusion * materialParams.intensity);
+        float ao = max(0.0, 1.0 - sqrt(occlusion * materialParams.intensity));
         ao = pow(ao, materialParams.power);
 
 #if defined(TARGET_MOBILE)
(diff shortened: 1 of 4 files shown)


filament/src/materials/ssao/sao.mat
```diff
@@ -216,7 +216,7 @@ fragment {
             tapPosition = angleStep * tapPosition;
         }
 
-        float ao = max(0.0, 1.0 - occlusion * materialParams.intensity);
+        float ao = max(0.0, 1.0 - sqrt(occlusion * materialParams.intensity));
```
romainguy, line 219: Is the extra sqrt necessary?

romainguy (approved):

filament/src/materials/ssao/sao.mat
pixelflinger: yeah, it helps a lot making the effect smoother and more consistant at all quality levels -- it's somewhat an artistic decision though. Before we achieved a similar effect by increasing the contrast, but it didn't look as  nice.

PR #3078, opened 2020-09-14
rename .blurScale to .apertureScale in dof options

pixelflinger:
.blurScale was in fact a scale factor applied to the circle of
confusion, which makes it indeed a "blur scale". 

However, it can also be seen as an aperture scale factor, 
which I think is more useful because it makes it more obvious that it 
can be used to set the DoF blur independently from the camera settings.


romainguy (approved):

pixelflinger:
closing because I got confused with a wrong comment -- will do a better one next.

PR #3079, opened 2020-09-14
rename .blurScale to .cocScale in dof options

pixelflinger:
.blurScale was in fact a scale factor applied to the circle of
confusion (which makes it indeed a "blur scale", but let's use
a more precise language here). 

Update comments to show how to  use .cocScale to control the DoF
effect independently from the camera aperture, which can be useful for
artistic reasons.


bejado (approved):
Does this warrant a middle number version change?

pixelflinger:
@bejado I would say no because it doesn't break binary compatibility, just source compatibility.

PR #3084, opened 2020-09-16
Add MSAA option to VSM

bejado:


filament/src/ShadowMapManager.cpp
```diff
@@ -487,18 +507,22 @@ void ShadowMapManager::calculateTextureRequirements(FEngine& engine,
     for (auto& cascade : mCascadeShadowMaps) {
         // Shadow map size should be the same for all cascades.
         const uint16_t dim = getShadowMapSize(cascade.getLightIndex());
```
pixelflinger, line 509: move getLightIndex into a local

filament/src/ShadowMapManager.cpp
```diff
@@ -487,18 +507,22 @@ void ShadowMapManager::calculateTextureRequirements(FEngine& engine,
...
         maxDimension = std::max(maxDimension, dim);
         cascade.setLayout({
             .layer = layer++,
-            .size = dim
+            .size = dim,
+            .vsmSamples = vsmSamples
         });
     }
     for (auto& spotShadowMap : mSpotShadowMaps) {
         const uint16_t dim = getShadowMapSize(spotShadowMap.getLightIndex());
```
pixelflinger, line 519: move getLightIndex into a local

libs/gltfio/include/gltfio/SimpleViewer.h
```diff
@@ -493,6 +494,9 @@ void SimpleViewer::updateUserInterface() {
         ImGui::Checkbox("Enable sunlight", &mEnableSunlight);
         ImGui::Checkbox("Enable shadows", &mEnableShadows);
         ImGui::Checkbox("Enable VSM", &mEnableVsm);
+        char label[32];
+        sprintf(label, "%d", 1 << mVsmMsaaSamplesLog2);
```
pixelflinger, line 498: use `snprintf` to make absolutely sure there are no overflows

filament/src/ShadowMapManager.cpp
```diff
@@ -154,10 +166,12 @@ void ShadowMapManager::render(FrameGraph& fg, FEngine& engine, FView& view,
                     if (view.hasVsm()) {
                         renderTarget.attachments = { { data.shadows, 0u, i }, { data.tempShadow } };
```
pixelflinger, line 167: rename `renderTarget` to `renderTargetDesc`  also rename `shadowTexture` to `shadowTextureDesc` (generally use the `fooDesc` convention for descriptors)

pixelflinger (changes requested):
minor naming changes

pixelflinger (changes requested):
FYI, this crashes with the GL backend on macos with this assert

`Assertion failed: (std::max(uint8_t(1), t->samples) == std::max(uint8_t(1), rt->gl.samples)), function framebufferTexture, file ../filament/backend/src/opengl/OpenGLDriver.cpp, line 927.
`

This happens because an attachment is not sampleable (meaning the framegraph never called sample() on it), so it's allocated as renderbuffers, but in this case the sample count must match the rentertarget's -- this might be a framegraph or gl backend issue, I'm not sure, we just never ran into this scenario before.

filament/src/ShadowMapManager.cpp
```diff
@@ -140,6 +151,7 @@ void ShadowMapManager::render(FrameGraph& fg, FEngine& engine, FView& view,
                         .width = mTextureRequirements.size, .height = mTextureRequirements.size,
                         .depth = 1,
                         .levels = 1,
+                        .samples = 1,
```
pixelflinger, line 154: This is what is causing the GL crash -- in fact this is always incorrect.  You're telling the framegraph you're requesting a depth buffer with a sample count of 1, but that's not true if you have MSAA.
By leaving the value to 0, you're telling the framegraph to "do the right thing" -- it'll automatically assign a sample count that matches the rendertarget's.

Now, I tried that, and I still get a crash in the GL driver, but I'm not sure why -- it looks  like a backend bug, I'm investigating.

Regardless I think setting samples=1 here is wrong, it should be either 0 or the requested msaa count.

pixelflinger (changes requested):

filament/src/ShadowMapManager.cpp
pixelflinger: Actually I see what you attempted to do now :-)

You wanted to use a non MS texture so you can share it with all rendertargets regardless of their MSAA, relying on the "magic resolve"; i.e. the tile would have the right MS count, and the buffer would be discarded anyways. Man this needs a comment!

Very smart :-)

So basically the problem is that it's not supported in the GL backend right now -- but you're right it should be. I'm looking into it.

pixelflinger (approved):

PR #3085, opened 2020-09-16
several fog fixes

### LINKED ISSUE #3069: Enabling fog gives a black screen on Android adreno GPU
If I enable fog for a View I can see the effect fine on my RockChip Mali device, but on my two Adreno devices (an SD 835 and an XR1 processor) the screen is all black.

One time I did see some of my scene, but there was a diagonal line cutting the screen in half with the clear part completely unchanged (no fog) and other part completely black. I can't actually reproduce this state reliably, but it implied some sort of threshold / rounding issue with fog shader calculation.

It seems to be the same with and without post-processing and with optimized and non-optimized material compilation.

Has anyone seen anything similar?

pixelflinger:
- inScatteringSize actually cannot be set to zero because it produces
  a 0^0 in the shader. So with this change, in-scattering must be
  strictly > 0 to enable.

- inScatteringSize on the java side had an incorrect default value of
  zero (which of course, now doesn't matter anymore).

- also clamp the fog altitude to 1mm which simplifies the shader quite
  a bit for the same result.

Fixes #3069

### DIFF
--- a/android/filament-android/src/main/java/com/google/android/filament/View.java
+++ b/android/filament-android/src/main/java/com/google/android/filament/View.java
@@ -335,9 +335,9 @@ public class View {
         public float inScatteringStart = 0.0f;
 
         /**
-         * size of in-scattering (>=0 to activate). Good values are >> 1 (e.g. ~10 - 100)
+         * size of in-scattering (>0 to activate). Good values are >> 1 (e.g. ~10 - 100)
          */
-        public float inScatteringSize = 0.0f;
+        public float inScatteringSize = -1.0f;
 
         /**
          * fog color will be modulated by the IBL color in the view direction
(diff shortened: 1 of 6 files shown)


android/filament-android/src/main/java/com/google/android/filament/View.java
```diff
@@ -335,9 +335,9 @@
         public float inScatteringStart = 0.0f;
 
         /**
-         * size of in-scattering (>=0 to activate). Good values are >> 1 (e.g. ~10 - 100)
+         * size of in-scattering (>0 to activate). Good values are >> 1 (e.g. ~10 - 100)
          */
-        public float inScatteringSize = 0.0f;
+        public float inScatteringSize = -1.0f;
```
romainguy, line 340: 0.0 would work with the new shader right?

romainguy (approved):

PR #3090, opened 2020-09-16
non-sampleable textures didn't support "magic resolve"

pixelflinger:
There was an inconsistency in the gles backend where non-sampleable,
non-MS textures failed (assert) if they were used with a render target 
with MS (i.e. "magic resolve" case).

We fix this by taking the same code path for textures and renderbuffers,
the only difference being the calls to glFrameBufferRenderbuffer
instead of glFramebufferTexture2D.


bejado (approved):

romainguy (approved):

bejado (approved):
Second commit LGTM

PR #3105, opened 2020-09-22
Fix the normal matrix when det < 0.

### LINKED ISSUE #3001: How to mirror Camera around X-axis?
I want to mirror my scene around the X-axis so I tried to flip the X-axis by setting the X scale of my camera matrix to -1 but this results in a pure black scene.  I'm trying to get the same behavior as in Blender. For example, consider this scene where I've set the X scaling to -1 of the default camera. When viewing through the _default viewport camera_ you see the normal scene (w/o the X-mirroring).

image

Then, when I look through the _camera with the -1 X scaling_, I get:

image

When I render my scene without the negative X scaling it renders fine. 

Is there an obvious answer to why this doesn't work in Filament?  I've got an idea what might cause this in my engine, but before I dive into that I was wondering if there is a simple explanation why this might be an issue with Filament.

prideout:
This was tested by replacing the node 0 scale in BusterDrone with [-1, 1, 1].

Fixes #3001.

### DIFF
--- a/filament/src/Scene.cpp
+++ b/filament/src/Scene.cpp
@@ -199,6 +199,12 @@ void FScene::updateUBOs(utils::Range<uint32_t> visibleRenderables, backend::Hand
         mat3f m = mat3f::getTransformForNormals(model.upperLeft());
         m *= mat3f(1.0f / std::sqrt(max(float3{length2(m[0]), length2(m[1]), length2(m[2])})));
 
+        // Since our "getTransformForNormals" matrix omits the divide-by-determinant step,
+        // it has the wrong sign when the determinant is negative.
+        if (sceneData.elementAt<REVERSED_WINDING_ORDER>(i)) {
+            m = -m;
+        }
+
         UniformBuffer::setUniform(buffer,
                 offset + offsetof(PerRenderableUib, worldFromModelNormalMatrix), m);
 
--- a/libs/math/include/math/mat3.h
+++ b/libs/math/include/math/mat3.h
@@ -260,6 +260,24 @@ public:
     /**
      * Returns a matrix suitable for transforming normals
      *
+     * The inverse-transpose of a matrix is equal to its cofactor matrix divided by its determinant:
+     *
+     *     transpose(inverse(M)) = cof(M) / det(M)
+     *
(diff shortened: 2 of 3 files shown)


prideout:
For future reference, the following commit is when we switched from transpose(inverse()) to cof()

https://github.com/google/filament/commit/f728776714d4521a44d2f21da473f9f96fd09907

filament/src/Scene.cpp
```diff
@@ -199,6 +199,12 @@ void FScene::updateUBOs(utils::Range<uint32_t> visibleRenderables, backend::Hand
         mat3f m = mat3f::getTransformForNormals(model.upperLeft());
         m *= mat3f(1.0f / std::sqrt(max(float3{length2(m[0]), length2(m[1]), length2(m[2])})));
 
+        // Since our "getTransformForNormals" matrix omits the divide-by-determinant step,
+        // it has the wrong sign when the determinant is negative.
+        if (sceneData.elementAt<REVERSED_WINDING_ORDER>(i)) {
```
romainguy, line 204: Should we do this in `getTransformForNormals`?

romainguy (approved):

filament/src/Scene.cpp
prideout: Maybe, but if we move it then we will be computing the determinant twice, which is somewhat redundant.  I'm on the fence.  @pixelflinger : you made the original switch to cof(), what do you think?

prideout: Also note that `getTransformForNormals` is used in a couple other places, e.g. when determining dominant light direction.  So yeah maybe we should make it slower and just add a divide-by-det in there.

prideout: Dale Weiler's [GitHub post](https://github.com/graphitemaster/normals_revisited) makes the argument that we should just use the raw cofactor matrix, but I didn't understand one of the steps in his algebra.  Moreover Lengyel never makes such a claim.  In practice, dividing the cof by the det (and therefore using the inverse-transpose) seems to fix this particular issue.

prideout:
Actually, let me close and research this a little more.  Using the cofactor matrix *should* be correct.

PR #3107, opened 2020-09-22
gltfio: Fix "uniform not found" errors.

prideout:
The equality operator for MaterialKey was missing some fields (also, its
ordering was inconsistent with the struct), which caused faulty behavior
in the material cache.

I ran into this bug when testing against a Blender-exported model that
has some materials with texture transforms, and some without.

### DIFF
--- a/libs/gltfio/src/MaterialProvider.cpp
+++ b/libs/gltfio/src/MaterialProvider.cpp
@@ -26,25 +26,28 @@ bool operator==(const MaterialKey& k1, const MaterialKey& k2) {
         (k1.unlit == k2.unlit) &&
         (k1.hasVertexColors == k2.hasVertexColors) &&
         (k1.hasBaseColorTexture == k2.hasBaseColorTexture) &&
-        (k1.hasMetallicRoughnessTexture == k2.hasMetallicRoughnessTexture) &&
         (k1.hasNormalTexture == k2.hasNormalTexture) &&
         (k1.hasOcclusionTexture == k2.hasOcclusionTexture) &&
         (k1.hasEmissiveTexture == k2.hasEmissiveTexture) &&
+        (k1.useSpecularGlossiness == k2.useSpecularGlossiness) &&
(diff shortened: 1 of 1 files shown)


libs/gltfio/src/MaterialProvider.cpp
```diff
@@ -26,25 +26,28 @@ bool operator==(const MaterialKey& k1, const MaterialKey& k2) {
...
         (k1.hasTransmission == k2.hasTransmission) &&
+        (k1.hasTextureTransforms == k2.hasTextureTransforms) &&
+        (k1.emissiveUV == k2.emissiveUV) &&
+        (k1.aoUV == k2.aoUV) &&
+        (k1.normalUV == k2.normalUV) &&
         (k1.hasTransmissionTexture == k2.hasTransmissionTexture) &&
-        (k1.clearCoatUV == k2.clearCoatUV) &&
-        (k1.clearCoatRoughnessUV == k2.clearCoatRoughnessUV) &&
-        (k1.clearCoatNormalUV == k2.clearCoatNormalUV);
+        (k1.transmissionUV == k2.transmissionUV);
```
romainguy, line 50: That's why I hate writing comparison functions for structs. I wish there was a safer way of doing this :/

romainguy (approved):

bejado (approved):

libs/gltfio/src/MaterialProvider.cpp
pixelflinger: for a struct like this, maybe we can have its ctors memset itself to make sure all padding is 0 and have the comparison operator just do memcmp?

pixelflinger (approved):

prideout:
@pixelflinger sorry I merged before reading your suggestion.  Yes, let's just migrate over to memcmp next time we need to touch this code.

I wish that C++ would do this for us, since it's just a POD.  C++20 makes it a little easier to generate comparison operators, but not equality.  Sigh.

PR #3108, opened 2020-09-22
Flip the shading normal when det < 0.

### LINKED ISSUE #3001: How to mirror Camera around X-axis?

prideout:
This is similar to my first attempt but now I'm more happy with the comments and explanations that I've added.

This was tested by replacing the node 0 scale in BusterDrone with [-1, 1, 1].

For future reference, commit f728776 shows when we switched from transpose(inverse()) to cof(). This was a good change, but before that particular change, we had a "two wrongs made a right" situation for mirrored normals.

Fixes #3001.

Before / After screenshots for Node_NegativeScale_07:

bad
good

### DIFF
--- a/filament/src/Scene.cpp
+++ b/filament/src/Scene.cpp
@@ -199,6 +199,13 @@ void FScene::updateUBOs(utils::Range<uint32_t> visibleRenderables, backend::Hand
         mat3f m = mat3f::getTransformForNormals(model.upperLeft());
         m *= mat3f(1.0f / std::sqrt(max(float3{length2(m[0]), length2(m[1]), length2(m[2])})));
 
+        // The shading normal must be flipped for mirror transformations.
+        // Basically we're shading the other side of the polygon and therefore need to negate the
+        // normal, similar to what we already do to support double-sided lighting.
+        if (sceneData.elementAt<REVERSED_WINDING_ORDER>(i)) {
+            m = -m;
+        }
+
         UniformBuffer::setUniform(buffer,
                 offset + offsetof(PerRenderableUib, worldFromModelNormalMatrix), m);
 
--- a/libs/math/include/math/mat3.h
+++ b/libs/math/include/math/mat3.h
@@ -260,6 +260,26 @@ public:
     /**
      * Returns a matrix suitable for transforming normals
      *
+     * Note that the inverse-transpose of a matrix is equal to its cofactor matrix divided by its
+     * determinant:
+     *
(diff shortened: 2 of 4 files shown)


romainguy (approved):

bejado (approved):

filament/src/Scene.cpp
```diff
@@ -199,6 +199,13 @@ void FScene::updateUBOs(utils::Range<uint32_t> visibleRenderables, backend::Hand
         mat3f m = mat3f::getTransformForNormals(model.upperLeft());
         m *= mat3f(1.0f / std::sqrt(max(float3{length2(m[0]), length2(m[1]), length2(m[2])})));
 
+        // The shading normal must be flipped for mirror transformations.
+        // Basically we're shading the other side of the polygon and therefore need to negate the
+        // normal, similar to what we already do to support double-sided lighting.
+        if (sceneData.elementAt<REVERSED_WINDING_ORDER>(i)) {
+            m = -m;
```
pixelflinger, line 206: I can't convince myself it is correct to flip T, B and N as this changes the handed-ness -- is this how others are doing it? Wouldn't that mirror the uv? is that what we want?

romainguy: That's with a negative scale. It reminds me of double sidedness where you can mirror or not mirror and I couldn't find a common solution (Unity HDRP lets you choose which behavior you want).

prideout: This has the effect of flipping T and N inside `main.vs`. 

B is still N x T (so handedness is the same) because we compute it in `shading_parameters.fs`.

ThreeJS renders the mirrored BusterDrone correctly without doing this explicitly because they use a traditional inverse-transpose rather than a cofactor matrix, and therefore already have the determinant sign accounted for.

pixelflinger (approved):

PR #3110, opened 2020-09-22
Implement Screen Space Cone Tracing (SSCT)

pixelflinger:
This is a technique from Naughty Dog used in TLOU2, where we compute
the "ambient" shadowing of a dominant light in screen-space using cone
tracing.

This is currently integrated to the SSAO pass.

ssct

### DIFF
--- a/android/filament-android/src/main/cpp/View.cpp
+++ b/android/filament-android/src/main/cpp/View.cpp
@@ -196,7 +196,7 @@ Java_com_google_android_filament_View_nGetAmbientOcclusion(JNIEnv*, jclass, jlon
 extern "C" JNIEXPORT void JNICALL
 Java_com_google_android_filament_View_nSetAmbientOcclusionOptions(JNIEnv*, jclass,
     jlong nativeView, jfloat radius, jfloat bias, jfloat power, jfloat resolution, jfloat intensity,
-    jint quality, jint upsampling, jboolean enabled, jfloat minHorizonAngleRad) {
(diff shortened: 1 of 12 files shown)


romainguy (approved):

android/filament-android/src/main/cpp/View.cpp
```diff
@@ -196,19 +196,42 @@ Java_com_google_android_filament_View_nGetAmbientOcclusion(JNIEnv*, jclass, jlon
...
+    options.lowPassFilter = (View::QualityLevel)lowPassFilter;
+    options.upsampling = (View::QualityLevel)upsampling;
+    options.enabled = (bool)enabled;
+    options.minHorizonAngleRad = minHorizonAngleRad;
+    view->setAmbientOcclusionOptions(options);
+}
+
+
+extern "C" JNIEXPORT void JNICALL
+Java_com_google_android_filament_View_nSetSSCTOptions(JNIEnv *, jclass, jlong nativeView,
```
prideout, line 217: Ah, took me a second to realize you had to split this in order to keep the arg count reasonable.  I wish JNI was nicer with POD structs.  :-\

prideout (approved):

romainguy (approved):

android/filament-android/src/main/cpp/View.cpp
pixelflinger: yeah, on the C++ I have to read the structure from filament, but that seemed easier to do it this way.

PR #3111, opened 2020-09-22
cmgen / mipgen: add opt-in for ASTC / ETC support.

prideout:
Block compression is currently only used when we invoke mipgen for the
suzanne demo. This PR makes it easy to disable.

Block compression is still enabled in our CMake build, but other build
systems can simply omit BlockCompression.{h,cpp} and it will "just
work".


libs/imageio/CMakeLists.txt
```diff
@@ -35,6 +35,11 @@ if (WIN32)
     target_link_libraries(${TARGET} PRIVATE wsock32)
 endif()
 
+# ==================================================================================================
+# Transitive macro definitions
+# ==================================================================================================
+target_compile_definitions(${TARGET} PUBLIC IMAGEIO_SUPPORTS_BLOCK_COMPRESSION)
```
romainguy, line 41: Shouldn't you not link against astcenc when IMAGEIO_SUPPORTS_BLOCK_COMPRESSION isn't set? cmgen will depend on imageio and imageio compils BlockCompression.cpp

prideout: Block compression is still always enabled in our CMake build, this change allows other build systems to make it optional.

romainguy: That's my question, how does this make it optional? Our code doesn't check for this define in imageio.

prideout: Ah the PR had a bad title, it actually makes it optional in cmgen / mipgen.

imageio itself doesn't need any changes because the build system can simply omit BlockCompression.{h,cpp}

romainguy (approved):

bejado (approved):

PR #3114, opened 2020-09-23
matdbg: allow edits without trailing newline.

prideout:


bejado (approved):

libs/matdbg/src/ShaderReplacer.cpp
```diff
@@ -300,11 +300,11 @@ void ShaderIndex::encodeShadersToIndices() {
         offset += sizeof(uint32_t);
 
         const char* s = record.decodedShaderText.c_str();
```
pixelflinger, line 302: `s` could be `const` here

libs/matdbg/src/ShaderReplacer.cpp
```diff
@@ -300,11 +300,11 @@ void ShaderIndex::encodeShadersToIndices() {
...
-        size_t length = record.decodedShaderText.length();
-        for (size_t cur = 0; s[cur] != '\0'; cur++) {
+        const size_t length = record.decodedShaderText.length();
+        for (size_t cur = 0; cur < length; cur++) {
```
pixelflinger, line 304: is it possible to have a `\0` before we reach `length`?

libs/matdbg/src/ShaderReplacer.cpp
```diff
@@ -300,11 +300,11 @@ void ShaderIndex::encodeShadersToIndices() {
...
             size_t pos = cur;
             size_t len = 0;
-            while (s[cur] != '\n') {
+            while (s[cur] != '\n' && s[cur] != '\0') {
```
pixelflinger, line 307: why look for the `\0` here instead of relying only on `length`, like for the outer loop?

pixelflinger (approved):

libs/matdbg/src/ShaderReplacer.cpp
prideout: Fixed.  Also renamed.

libs/matdbg/src/ShaderReplacer.cpp
prideout: Yes they are equivalent, fixed.

libs/matdbg/src/ShaderReplacer.cpp
prideout: This is not possible because length() counts up to the null.

Btw I just noticed that we are already inconsistent about using length() vs size()  (they are equivalent)

libs/matdbg/src/ShaderReplacer.cpp
prideout: Actually this was a good catch, this should use `length` rather than `\0` to robust against unicode strings.

PR #3116, opened 2020-09-24
matdbg: fix edits for large shaders. 

prideout:
- Fixed edits for large shaders.
    - If the shader source exceeds the websocket max size, we might receive it in chunks.
    - We now allow for this by including a length prefix in the EDIT command.
- Added support for UTF-8 string encoding.
- Display an orange warning when editing an unused shader variant.

### DIFF
--- a/libs/utils/src/CString.cpp
+++ b/libs/utils/src/CString.cpp
@@ -36,7 +36,6 @@ int StaticString::compare(const StaticString& rhs) const noexcept {
 UTILS_NOINLINE
 CString::CString(const char* cstr, size_t length) {
     if (length && cstr) {
-        assert(length == strlen(cstr));
         Data* p = (Data*)malloc(sizeof(Data) + length + 1);
         p->length = (size_type)length;
         mCStr = (value_type*)(p + 1);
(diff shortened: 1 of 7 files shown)


libs/utils/src/CString.cpp
```diff
@@ -36,7 +36,6 @@ int StaticString::compare(const StaticString& rhs) const noexcept {
 UTILS_NOINLINE
 CString::CString(const char* cstr, size_t length) {
     if (length && cstr) {
-        assert(length == strlen(cstr));
```
prideout, line 39: I removed this assert because this constraint doesn't seem necessary. CString works perfectly well for arbitrary blobs or alternative character encodings.

pixelflinger: okay, but that means that it's not really a  "c string" anymore. Could you maybe update the comments in the header that this is a supported use case?

pixelflinger (approved):

libs/utils/src/CString.cpp
prideout: Good catch because a comment definitely needed to be updated.  Fixed.  Note that CString now has similar constraints to std::string, which can also hold arbitrary data when using the constructor that takes a length.

PR #3123, opened 2020-09-25
Vulkan: fix final image barrier used for swap chain.

prideout:
This image barrier should wait until all write operations are complete. This was not set up correctly (src and dst flags were flipped) but this did not manifest in any visible problems or validation errors.

I also tried to document this better, since using an image barrier for the swap chain seems to be an unusual thing to do (e.g. the Sascha Willems demos do not do this).

### DIFF
--- a/filament/backend/src/vulkan/VulkanContext.cpp
+++ b/filament/backend/src/vulkan/VulkanContext.cpp
@@ -491,12 +491,22 @@ void destroySwapChain(VulkanContext& context, VulkanSurfaceContext& surfaceConte
     vkFreeMemory(context.device, surfaceContext.depth.memory, VKALLOC);
 }
 
-void transitionSwapChain(VulkanContext& context) {
+// makeSwapChainPresentable()
+//
+// Near the end of the frame, we transition the swap chain to VK_IMAGE_LAYOUT_PRESENT_SRC_KHR using
+// an image barrier rather than a render pass because each render pass does not know whether or not
+// it is the last pass in the frame. (This seems to be an atypical way of achieving the transition,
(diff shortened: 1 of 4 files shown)


romainguy (approved):

filament/backend/src/vulkan/VulkanContext.cpp
```diff
@@ -491,12 +491,22 @@ void destroySwapChain(VulkanContext& context, VulkanSurfaceContext& surfaceConte
     vkFreeMemory(context.device, surfaceContext.depth.memory, VKALLOC);
 }
 
-void transitionSwapChain(VulkanContext& context) {
+// makeSwapChainPresentable()
+//
+// Near the end of the frame, we transition the swap chain to VK_IMAGE_LAYOUT_PRESENT_SRC_KHR using
+// an image barrier rather than a render pass because each render pass does not whether or not it
```
pixelflinger, line 497: I think there is a typo here: "does not **know** whether or not"

filament/backend/src/vulkan/VulkanContext.cpp
```diff
@@ -491,12 +491,22 @@ void destroySwapChain(VulkanContext& context, VulkanSurfaceContext& surfaceConte
...
+// is the last pass in the frame. (This seems to be an atypical of way of achieving the transition,
```
pixelflinger, line 498: typo: "atypical way"

pixelflinger (approved):

prideout:
Thanks for catching those typos Mathias, fixed.  Also tested on Android.

PR #3130, opened 2020-09-28
Introduce libs/viewer and serializable view settings.

prideout:
NOTE: individual commits might be easier to review than the consolidated PR.

Adds a new lib that holds SimpleViewer and a new `Settings` structure that includes a JSON reader / writer. This will be used for automated testing and for client / server communication.

The `viewer::Settings` struct is closely associated with the `filament::View` API; when updating the latter we will often need to update the former, as well as some serialization code. This increases the maintenance burden and I think we should consider using a parser library like libclang or a macro-based reflection utility.

This PR also migrates SimpleViewer into libs/viewer and un-inlines its implementation. It does not belong in gltfio because it has an imgui dependency.


romainguy (approved):
👍

libs/viewer/src/Settings.cpp
```diff
@@ -0,0 +1,834 @@
...
+#include <viewer/Settings.h>
+
+#include <utils/Log.h>
+
+#include <sstream>
+#include <string>
+
+#include <assert.h>
+
+// NOTE: We are including cgltf only to get JSMN. We should consider adding jsmn to third_party.
```
bejado: Is this comment still needed / can we include the jsmn headers directly now?

bejado (approved):
Very excited for this!

libs/viewer/src/Settings.cpp
prideout: I removed this comment in a subsequent CL.

bejado: whoops, missed that

PR #3131, opened 2020-09-28
Use light space Z as VSM depth metric

bejado:
This switches to using the light space Z coordinate as the VSM depth metric. This fixes precision issues seen with a distant camera far plane.

### DIFF
--- a/filament/src/ShadowMap.cpp
+++ b/filament/src/ShadowMap.cpp
@@ -427,17 +427,7 @@ void ShadowMap::computeShadowCameraDirectional(
         }
         mLightSpace = St;
 
-        // The mLightSpace matrix transforms coordinates from world space into (u, v, z)
-        // coordinates, where (u, v) are used to access the shadow map, and z is the PCF
-        // comparison value between 0 and 1.
-        // For VSM, we don't want to project the z coordinate, but do want to transform it to linear
-        // light space. We then scale the z by zfar, which prevents us from overflowing when we
-        // square z in the depth shader.
-        mLightSpaceVsm = mLightSpace;
-        mLightSpaceVsm[0][2] = Mv[0][2] / std::abs(zfar);
-        mLightSpaceVsm[1][2] = Mv[1][2] / std::abs(zfar);
-        mLightSpaceVsm[2][2] = Mv[2][2] / std::abs(zfar);
-        mLightSpaceVsm[3][2] = Mv[3][2] / std::abs(zfar);
+        mLightSpaceVsm = computeVsmLightSpaceMatrix(mLightSpace, Mv, zfar);
 
         // We apply the constant bias in world space (as opposed to light-space) to account
         // for perspective and lispsm shadow maps. This also allows us to do this at zero-cost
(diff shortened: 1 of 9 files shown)


shaders/src/light_directional.fs
```diff
@@ -46,8 +46,10 @@ void evaluateDirectionalLight(const MaterialInputs material,
         if (hasDirectionalShadows && cascadeHasVisibleShadows) {
             uint layer = cascade;
 #if defined(HAS_VSM)
-            highp float fragDepth = length(vertex_worldPosition - frameUniforms.lightPosition.xyz);
-            visibility = shadowVsm(light_shadowMap, layer, getCascadeLightSpacePosition(cascade), fragDepth);
+            highp vec3 lightSpacePosition = getCascadeLightSpacePosition(cascade);
+            // Flip the sign of the Z coordinate; -Z is forward in light space.
+            lightSpacePosition.z *= -1.0;
```
pixelflinger, line 51: You should be able to include the negation in the matrix, right? (since it's already a weird matrix anyways) -- this way you wouldn't need a `#if HAS_VSM` case here.

shaders/src/light_punctual.fs
```diff
@@ -172,9 +172,10 @@ void evaluatePunctualLights(const PixelParams pixel, inout vec3 color) {
         if (light.NoL > 0.0){
             if (light.castsShadows) {
 #if defined(HAS_VSM)
-                highp float fragDepth = length(vertex_worldPosition - light.worldPosition);
-                visibility = shadowVsm(light_shadowMap, light.shadowLayer,
-                    getSpotLightSpacePosition(light.shadowIndex), fragDepth);
+                highp vec3 lightSpacePosition = getSpotLightSpacePosition(light.shadowIndex);
+                // Flip the sign of the Z coordinate; -Z is forward in light space.
+                lightSpacePosition.z *= -1.0;
```
pixelflinger, line 177: You should be able to include the negation in the matrix, right? (since it's already a weird matrix anyways) -- this way you wouldn't need a #if HAS_VSM case here.

pixelflinger (changes requested):

filament/src/ShadowMap.cpp
```diff
@@ -453,6 +427,18 @@ void ShadowMap::computeShadowCameraDirectional(
...
+        // coordinates, where (u, v) are used to access the shadow map, and z is the PCF
+        // comparison value between 0 and 1.
+        // For VSM, we don't want to project the z coordinate, but do want to transform it to linear
+        // light space. We then scale the z by zfar, which prevents us from overflowing when we
+        // square z in the depth shader.
+        mLightSpaceVsm = mLightSpace;
+        mLightSpaceVsm[0][2] = Mv[0][2] / std::abs(zfar);
+        mLightSpaceVsm[1][2] = Mv[1][2] / std::abs(zfar);
+        mLightSpaceVsm[2][2] = Mv[2][2] / std::abs(zfar);
+        mLightSpaceVsm[3][2] = Mv[3][2] / std::abs(zfar);
```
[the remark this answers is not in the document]
bejado: Sure, I notice we do this a lot. What's the reason?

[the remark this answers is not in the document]
pixelflinger: when you use `operator /` it's applied to all elements of the vector, so that's 4 divides for a float4, so it's a good habit. Here it's not that case, but you still have 4 divides that can be replaced 4 muls --  it wouldn't make a huge difference here, maybe not worth it, but still a good habit.

shaders/src/light_directional.fs
bejado: Yeah, I can fold that term into the matrix. I'll still need a `HAS_VSM` case, because we're calling a unique shadow function (`shadowVsm`).

pixelflinger (approved):

PR #3138, opened 2020-09-30
Introduce a configurable test case generator.

prideout:
This lets us generate long lists of settings permutations by writing a
simple JSON spec. For example, the following spec would generate six
`Settings` objects, all of which have SSAO enabled. See the unit test
for a larger example.

```
{
    "name": "viewopts",
    "base": { "view.ssao.enabled": true }
    "permute": {
        "view.dithering": ["NONE", "TEMPORAL"],
        "view.sampleCount": [1, 4, 8]
    }
}
```

### DIFF
--- a/libs/viewer/include/viewer/Automation.h
+++ b/libs/viewer/include/viewer/Automation.h
@@ -17,30 +17,46 @@
 #ifndef VIEWER_AUTOMATION_H
 #define VIEWER_AUTOMATION_H
 
-#include <string>
-#include <vector>
-
 #include <viewer/Settings.h>
 
 namespace filament {
 namespace viewer {
 
-// Named list of settings permutations.
+// Immutable list of settings permutations that are created and owned by AutomationList.
 struct AutomationSpec {
-    std::string name;
-    std::vector<Settings> cases;
+    char const* name;
+    size_t count;
+    Settings const* settings;
 };
 
-// Consumes a JSON string and produces a list of automation specs using a counting algorithm.
+// List of automation specs constructed from a JSON string.
 //
 // Each top-level item in the JSON is an object with "name", "base" and "permute".
 // The "base" object specifies a single set of changes to apply to default settings.
 // The optional "permute" object specifies a cross product of changes to apply to the base.
 // See the unit test for an example.
-//
-// - Returns true if successful.
-// - This function writes warnings and error messages into the utils log.
-bool generate(const char* jsonChunk, size_t size, std::vector<AutomationSpec>* out);
+class AutomationList {
+public:
+
+    // Parses a JSON spec, then generates a set of Settings lists.
+    // Returns null on failure (see utils log for warnings and errors).
+    // Clients should release memory using "delete".
+    static AutomationList* generate(const char* jsonChunk, size_t size);
+
+    // Returns the number of generated Settings lists.
+    size_t size() const;
+
+    // Returns a view onto a generated Settings list.
+    AutomationSpec get(size_t index) const;
+
+    // Frees all Settings objects and name strings.
+    ~AutomationList();
+
+private:
+    struct Impl;
+    AutomationList(Impl*);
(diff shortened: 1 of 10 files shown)


libs/viewer/include/viewer/Automation.h
```diff
@@ -0,0 +1,48 @@
...
+// Consumes a JSON string and produces a list of automation specs using a counting algorithm.
+//
+// Each top-level item in the JSON is an object with "name", "base" and "permute".
+// The "base" object specifies a single set of changes to apply to default settings.
+// The optional "permute" object specifies a cross product of changes to apply to the base.
+// See the unit test for an example.
+//
+// - Returns true if successful.
+// - This function writes warnings and error messages into the utils log.
+bool generate(const char* jsonChunk, size_t size, std::vector<AutomationSpec>* out);
```
romainguy, line 43: If this is intended to be a public header (and not just for us), don't use STL

libs/viewer/src/parse.h
```diff
@@ -0,0 +1,33 @@
+/*
```
romainguy, line 1: Maybe use a more specific file name?

libs/viewer/include/viewer/Automation.h
prideout: It's mostly intended for us but let me clean this up anyway and make it into a proper class without any STL in the header, should be easy.

libs/viewer/src/parse.h
prideout: Fixed.

libs/viewer/include/viewer/Automation.h
prideout: Fixed.

pixelflinger: I think the "no stl in headers" rule is for libfilament, libutils and libmath -- i.e. what you need to use filament.  I don't think we need to apply it for tools (?)

romainguy: If it's just for us or mostly for us yeah that's fine

libs/viewer/src/Automation.cpp
```diff
@@ -0,0 +1,262 @@
+/*
+ * Copyright (C) 2020 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by mIcable law or agreed to in writing, software
```
bejado, line 10: mIcable -> applicable (I think Automation.h, jsonParseUtils.h and SimpleViewer.h have this mutation too)

libs/viewer/src/Automation.cpp
```diff
@@ -0,0 +1,262 @@
...
+            }
+        }
+
+        // Apply changes to the settings object.
+        for (const auto& iter : iters) {
+            const std::string& jsonString = *iter;
+            if (VERBOSE) {
+                slog.i << "  Applying " << jsonString.c_str() << io::endl;
+            }
+            if (!readJson(jsonString.c_str(), jsonString.size(), &base)) {
```
bejado, line 198: I assume after this call to `readJson`, we would do something interesting with `base` (like actually testing Filament with it). Am I understanding correctly?

bejado (approved):
This is really cool

libs/viewer/src/Automation.cpp
prideout: Correct, this returns a list of Settings that can then be applied to Filament; my next PR will make use of it.  The current PR merely generates the specified permutations and sets up a unit test for them.

libs/viewer/src/Automation.cpp
prideout: Wups, will fix!

PR #3146, opened 2020-10-01
gltf_viewer: add --settings, prep for automation.

prideout:


samples/gltf_viewer.cpp
```diff
@@ -402,8 +431,8 @@ static void colorGradingUI(App& app) {
         colorGrading.toneMapping = (decltype(colorGrading.toneMapping)) toneMapping;
 
         if (ImGui::CollapsingHeader("White balance")) {
-            ImGui::SliderInt("Temperature", &colorGrading.temperature, -100, 100);
-            ImGui::SliderInt("Tint", &colorGrading.tint, -100, 100);
+            ImGui::SliderFloat("Temperature", &colorGrading.temperature, -1.0f, 1.0f);
```
prideout, line 434: The actual API takes a float between -1 and 1, and the `Settings` struct is 1:1 with the API, so I think this is a bit more straightforward.

samples/gltf_viewer.cpp
```diff
@@ -721,11 +764,15 @@ int main(int argc, char** argv) {
             }
 
             colorGradingUI(app);
-        });
 
-        // Leave FXAA enabled but we also enable MSAA for a nice result. The wireframe looks
-        // much better with MSAA enabled.
-        view->setSampleCount(4);
```
prideout, line 728: This codeline became bogus after we added a checkbox for MSAA.  It did no harm because it got clobbered later in the execution floor.

samples/gltf_viewer.cpp
romainguy: The UI uses -100/+100 because that's what tools like Photoshop/Lightroom/etc. expose.  There are other settings that I think we need to expose differently in the UI than in our API (for instance angles should be in degrees in the UI, not in radians).

samples/CMakeLists.txt
```diff
@@ -240,7 +240,7 @@ if (NOT ANDROID)
 
     # Sample app specific
     target_link_libraries(frame_generator PRIVATE imageio)
-    target_link_libraries(gltf_viewer PRIVATE gltf-resources gltfio viewer)
+    target_link_libraries(gltf_viewer PRIVATE gltf-resources gltfio viewer imageio)
```
prideout, line 243: Accidentally added this dep, it'll be needed for the next PR though since it will add screenshot capability.

samples/gltf_viewer.cpp
prideout: Ah, got it.  I'll fix this using temporary integers.

prideout: Fixed.

romainguy (approved):

PR #3147, opened 2020-10-02
VSM: render both shadow casters and receivers

bejado:
VSM is unique in that both shadow casters AND shadow receivers must be rendered into the shadow map.  This differs slightly from PCF shadows:

|caster|receiver|PCF rendered|VSM rendered|example|
|--:|--:|--:|--:|--:|
|0|0|NO|NO|skybox|
|0|1|NO|YES|ground plane|
|1|0|YES|YES|an offscreen [cucoloris](https://en.wikipedia.org/wiki/Cucoloris)|
|1|1|YES|YES|typical object|

The second scenario, where PCF and VSM differ, is what is being updated here.

- Rename `VISIBLE_DIR_SHADOW_CASTER` type visibility bits to `VISIBLE_DIR_SHADOW_RENDERABLE`, to make them more general
- Update `computeVisibilityMasks` logic to mark all shadow receivers visible during VSM shadow passes.

### DIFF
--- a/filament/src/View.cpp
+++ b/filament/src/View.cpp
@@ -578,19 +578,22 @@ void FView::computeVisibilityMasks(
         //
         // It is written without if statements to avoid branches, which allows it to be vectorized 16x.
 
-        bool visRenderables   = (!v.culling || (mask & VISIBLE_RENDERABLE))    && inVisibleLayer;
-        bool visShadowRenderable =
-            (!v.culling || (mask & VISIBLE_DIR_SHADOW_RENDERABLE)) && inVisibleLayer
-            && (v.castShadows || (hasVsm && v.receiveShadows));
+        const bool visRenderables   = (!v.culling || (mask & VISIBLE_RENDERABLE))    && inVisibleLayer;
+        const bool vvsmRenderShadow = hasVsm && v.receiveShadows;
+        const bool visShadowParticipant = v.castShadows || vvsmRenderShadow;
+        const bool visShadowRenderable =
+            (!v.culling || (mask & VISIBLE_DIR_SHADOW_RENDERABLE)) && inVisibleLayer && visShadowParticipant;
         visibleMask[i] = Culler::result_type(visRenderables) |
                 Culler::result_type(visShadowRenderable << 1u);
         // this loop gets fully unrolled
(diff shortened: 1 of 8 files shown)


filament/src/View.cpp
```diff
@@ -555,7 +555,7 @@ void FView::computeVisibilityMasks(
         uint8_t visibleLayers,
         uint8_t const* UTILS_RESTRICT layers,
         FRenderableManager::Visibility const* UTILS_RESTRICT visibility,
-        uint8_t* UTILS_RESTRICT visibleMask, size_t count) {
+        uint8_t* UTILS_RESTRICT visibleMask, size_t count, bool hasVsm) {
```
pixelflinger, line 558: please make sure that adding this `bool` doesn't impact how this loop is vectorized.

pixelflinger (changes requested):
Not a change -- but I just want to check that the loop is still vectorized after the addition of the boolean, before we check in.

filament/src/View.cpp
bejado: Sounds good, will do.

bejado: It wasn't getting vectorized, but splitting the statement up into individual booleans did the trick.

romainguy:
LGTM with the loop now vectorized

romainguy (approved):

shaders/src/shadowing.fs
```diff
@@ -360,6 +360,11 @@ highp float shadowVsm(const highp sampler2DArray shadowMap, const uint layer,
     highp vec2 moments = texture(shadowMap, vec3(shadowPosition.xy, layer)).xy;
     highp float depth = shadowPosition.z;
 
+    // depth must be clamped to support floating-point depth formats. This is to avoid comparing a
+    // value from the depth texture (which is never greater than 1.0) with a greater-than-one
+    // comparison value (which is possible with floating-point formats).
+    depth = min(depth, 1.0f);
```
romainguy, line 366: Note that a saturate (clamp(0, 1, x)) is free on some (all?) architecture. It might be worth using a clamp here. Not sure if min gives you that. Not a big deal either way.

romainguy (approved):

PR #3149, opened 2020-10-02
gltf_viewer: Add automation UI and functionality.

prideout:
This adds `AutomationEngine` to libs/viewer, which iterates through
`Settings` instances that were generated from a JSON spec and applies
them to a Filament `View`. It can be configured to sleep between tests
using a time delay or a frame count.

This also adds command line arguments and user-interface elements to
`gltf_viewer` for automated testing.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,5 @@ filament/docs/html/**
 civetweb.txt
 /TAGS
 settings.json
+test*.png
+test*.json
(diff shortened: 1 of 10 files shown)


romainguy (approved):
We should be able to use this to replace my terrible frame_generator right? It basically did the same thing but with material parameters. I used it to generate all the rows of spheres in the documentation. It would be useful to test the shader too eventually.

prideout:
> We should be able to use this to replace my terrible frame_generator right? It basically did the same thing but with material parameters. I used it to generate all the rows of spheres in the documentation. It would be useful to test the shader too eventually.

Ha sure, in fact I looked at your screenshot code for inspiration.  Currently the JSON spec does not let you tweak material parameters, but it should be an easy feature to add.  After that we can remove frame_generator.

PR #3161, opened 2020-10-06
Add Live Wallpaper Android sample

bejado:
Super simple sample right now, just updates the color of the Skybox.


romainguy (approved):

prideout (approved):
Neat.  I wonder if there's something that we could demonstrate here that involves camera positioning.

bejado:
> Neat.  I wonder if there's something that we could demonstrate here that involves camera positioning.

Definitely, I have some ideas on how to improve it that would involve loading a glTF model and animating the camera.

pixelflinger:
Doing something based on the accelerometer/gyro could be kind of cool!

PR #3167, opened 2020-10-07
Support clang 8 by adding more #includes, etc.

prideout:


romainguy:
Is this for Linux? We build with clang 7 in CI and there are no issues. We use clang 11 and 12 on macOS without issues either.

prideout:
Yes, I had to make these changes to build on a cloud Linux machine that had clang 8.

This PR does not need to be merged, but I thought it was interesting.

prideout:
Closing the PR but keeping the brach intact, in case we need this in the future.

PR #3172, opened 2020-10-07
Fix Vulkan crash when using shadow cascades

bejado:


filament/backend/src/vulkan/VulkanHandles.cpp
```diff
@@ -600,8 +601,9 @@ VulkanTexture::VulkanTexture(VulkanContext& context, SamplerType target, uint8_t
 
     if (any(usage & (TextureUsage::COLOR_ATTACHMENT | TextureUsage::DEPTH_ATTACHMENT))) {
         auto transition = [=](VulkanCommandBuffer commands) {
+            const uint32_t layers = target == SamplerType::SAMPLER_2D_ARRAY ? depth : 1;
```
romainguy, line 604: The name of the variable "depth" is confusing here.

romainguy (approved):

prideout (approved):

filament/backend/src/vulkan/VulkanHandles.cpp
bejado: I'll add a clarification comment.

PR #3173, opened 2020-10-07
Add Dockerfile and test script for SwiftShader.

prideout:
This adds a Dockerfile and a new bash script that makes it
east to invoke the appopriate Docker commands.

This does not yet enable a GitHub Action because of intermittent
issues that we have not yet ironed out.

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -468,3 +468,19 @@ export SWIFTSHADER_LD_LIBRARY_PATH=`pwd`
 ```
 
 Next, go to your Filament repo and use the [easy build](#easy-build) script with `-t`.
+
+## SwiftShader for CI
+
+Continuous testing turnaround can be quite slow if you need to build SwiftShader from scratch, so we
+provide an Ubuntu-based Docker image that has it already built. The Docker image also includes
+everything necessary for building Filament. You can fetch and run the image as follows:
+
+```
+docker pull ghcr.io/filament-assets/swiftshader
+docker run -it ghcr.io/filament-assets/swiftshader
+```
+
+To do more with the container, see the helper script at `build/swiftshader/test.sh`.
+
+If you are a team member, you can update the public image to the latest SwiftShader by
+following the instructions at the top of `build/swiftshader/Dockerfile`.
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,4 @@ civetweb.txt
 settings.json
 test*.png
 test*.json
+results
--- a/build/swiftshader/Dockerfile
+++ b/build/swiftshader/Dockerfile
@@ -1,0 +1,53 @@
+# Build the image:
+#   docker build --no-cache --tag ssfilament -f build/swiftshader/Dockerfile .
+#   docker tag ssfilament ghcr.io/filament-assets/swiftshader
+#
+# Publish the image:
+#   docker login ghcr.io --username <user> --password <token>
+#   docker push ghcr.io/filament-assets/swiftshader
+#
+# Run the image and mount the current directory:
+#   docker run -it -v `pwd`:/trees/filament -t ssfilament
+
+FROM ubuntu:focal
+WORKDIR /trees
+ARG DEBIAN_FRONTEND=noninteractive
+ENV SWIFTSHADER_LD_LIBRARY_PATH=/trees/swiftshader/build
+ENV CXXFLAGS='-fno-builtin -Wno-pass-failed'
+
+RUN apt-get update && \
+    apt-get --no-install-recommends install -y \
+	apt-transport-https \
+	apt-utils \
+	build-essential \
+	cmake \
+	ca-certificates \
+	git \
+	ninja-build \
+	python \
+	python3 \
+	xorg-dev \
+	clang-7 \
+	libc++-7-dev \
+	libc++abi-7-dev \
+	lldb
+
+# Ensure that clang is used instead of gcc.
+RUN set -eux ;\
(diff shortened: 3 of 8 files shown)


BUILDING.md
```diff
@@ -468,3 +468,29 @@ export SWIFTSHADER_LD_LIBRARY_PATH=`pwd`
 ```
 
 Next, go to your Filament repo and use the [easy build](#easy-build) script with `-t`.
+
+## SwiftShader for CI
+
+Continuous testing turnaround can be quite slow if you need build SwiftShader from scratch, so we
```
romainguy, line 474: Now that one of us knows how to build Docker image, should we do the same for our regular Linux builds? We currently have to download LLVM and CMake every time.

BUILDING.md
```diff
@@ -468,3 +468,29 @@ export SWIFTSHADER_LD_LIBRARY_PATH=`pwd`
...
+provide an Ubuntu-based Docker image that has it ready to go. The Docker image also includes
+everything necessary for building Filament and running tests. You can fetch and run the image as
+follows:
+
+```
+docker pull ghcr.io/prideout/ssfilament
```
romainguy, line 480: Should we create a `filament` account?

build/Dockerfile
```diff
@@ -0,0 +1,49 @@
...
+	apt-utils \
+	build-essential \
+	cmake \
+	ca-certificates \
+	git \
+	ninja-build \
+	python \
+	python3 \
+	xorg-dev \
+	clang-7 \
```
romainguy, line 18: Could we use clang-9+ if we can choose our own Ubuntu? We were restricted to 7 because of the ancient Ubuntu in CI (14 or 16)

romainguy (approved):

.github/workflows/swiftshader.yml
```diff
@@ -0,0 +1,40 @@
+name: SwiftShader
+
+on: push
```
bejado, line 3: This action will trigger whenever any branch is pushed. I don't think we want that. Did you mean to replace this with the `workflow_dispatch` below?

.github/workflows/swiftshader.yml
```diff
@@ -0,0 +1,40 @@
...
+# on:
+#   workflow_dispatch:
+
+jobs:
+  swiftshader:
+    name: Test Filament + Vulkan + SwiftShader
+    runs-on: ubuntu-latest
+    steps:
+
+    - uses: actions/checkout@v2.0.0
```
bejado, line 14: Do we still need to checkout Filament?

BUILDING.md
prideout: Yes we could this.  We could simply use the top part of the Dockerfile that I already wrote.

BUILDING.md
prideout: Someone already has that namespace.  It may make sense to use the "google" account but I have not figured out how to get permission for that.

build/Dockerfile
prideout: Yes, this is something we can upgrade after we move to a Docker-based workflow for our Linux builds.

.github/workflows/swiftshader.yml
prideout: I actually wasn't sure, I wanted to ask you guys about that.  I was leaning towards the manual trigger via workflow_dispatch.

.github/workflows/swiftshader.yml
prideout: This ties in with your previous question a little; the docker run command uses `-v` to access the host copy of Filament, which would be the branch that's being pushed.  I think the checkout action is necessary to give the host access to this.

.github/workflows/swiftshader.yml
bejado: I'm fine with a manual trigger, but it's a bit tricky to set up. You have to get a GitHub personal access token and then make a POST request to a specific GitHub API: https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#repository_dispatch

.github/workflows/swiftshader.yml
bejado: Got it, sounds good

.github/workflows/swiftshader.yml
```diff
@@ -25,15 +25,16 @@ jobs:
 
     - name: Build Filament with SwiftShader Support
       run: |
-          docker run -v `pwd`:/trees/branch ssfilament branch/build.sh -t release
+          docker run -v `pwd`:/trees/branch ssfilament branch/build.sh -t debug
```
romainguy, line 28: Do we need to build everything? Could we just build gltf_viewer to reduce turnaround times?

prideout: That's a good idea, will do.  (The last few commit were tests btw, I will remove them.)

BUILDING.md
prideout: Created an account called `filament-assets` for this.

prideout:
There are a few different segfaults that are preventing me from merging this, they all seem to occur either within SwiftShader or spirv-tools.  They are all intermittent.

The PR now includes a workaround patch for one of these segfaults, which simply removes one of the spirv-tools optimization passes.

However there's another intermittent segfault that occurs while creating a fence.  Here's the stack trace:
https://gist.github.com/prideout/ab3c926290c18bd45e716fc4bcae1042

We will need to report these issues to the SwiftShader team.

PR #3185, opened 2020-10-12
VSM: add support for mipmaps and anisotropic sampling

bejado:
Anisotropic filtering helps improve VSM shadow quality at glancing angles.

vsm-anisotropy

### DIFF
--- a/android/filament-android/src/main/java/com/google/android/filament/View.java
+++ b/android/filament-android/src/main/java/com/google/android/filament/View.java
@@ -604,14 +604,14 @@ public class View {
          *
          * <strong>Warning: This API is still experimental and subject to change.</strong>
          */
-        public ShadowType shadowType;
+        public @NonNull ShadowType shadowType = ShadowType.PCF;
 
         /**
          * Sets the number of anisotropic samples to use when sampling a VSM shadow map. If greater
          * than 0, mipmaps will automatically be generated each frame for all lights.
          *
          * <p>
-         * The number of anisotropic samples = 2 ^ vsmAnisotropyLog2.
+         * The number of anisotropic samples = 2 ^ vsmAnisotropy.
          * </p>
          *
          * <p>
@@ -620,7 +620,7 @@ public class View {
          *
          * <strong>Warning: This API is still experimental and subject to change.</strong>
          */
-        public int vsmAnisotropyLog2;
+        public int vsmAnisotropy = 0;
     }
 
     /**
@@ -1215,7 +1215,7 @@ public class View {
     public void setShadowOptions(@NonNull ShadowOptions options) {
         mShadowOptions = options;
         nSetShadowOptions(getNativeObject(), options.shadowType.ordinal(),
-                options.vsmAnisotropyLog2);
+                options.vsmAnisotropy);
     }
 
     /**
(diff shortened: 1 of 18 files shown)


filament/src/materials/vsmMipmap.mat
```diff
@@ -0,0 +1,40 @@
...
+    domain : postprocess,
+    depthWrite : false,
+    depthCulling : false,
+    culling: none
+}
+
+fragment {
+    void postProcess(inout PostProcessInputs postProcess) {
+        highp vec2 uv = gl_FragCoord.xy * 2.0f /
+            vec2(textureSize(materialParams_color, materialParams.level).xy);
```
romainguy, line 34: How easy would it be to just pass the texture size as a uniform? Not a big deal but on some GPUs it's equivalent to a texture read.

android/filament-android/src/main/java/com/google/android/filament/View.java
```diff
@@ -582,6 +583,46 @@
...
+         * The number of anisotropic samples = 2 ^ vsmAnisotropyLog2.
+         * </p>
+         *
+         * <p>
+         * Only applicable when shadowType is set to ShadowType::VSM.
+         * </p>
+         *
+         * <strong>Warning: This API is still experimental and subject to change.</strong>
+         */
+        public int vsmAnisotropyLog2;
```
romainguy, line 623: Should we have a separate struct for VSM options? I imagine there could be others (MSAA for instance). Also just call the parameter vsmAnisotropy.

android/filament-android/src/main/java/com/google/android/filament/View.java
```diff
@@ -582,6 +583,46 @@
...
+         * {@link ShadowType#VSM} imposes a restriction on marking renderables as only shadow
+         * receivers (but not casters). To ensure correct shadowing with VSM, all shadow participant
+         * renderables should be marked as both receivers and casters. Objects that are guaranteed
+         * to not cast shadows on themselves or other objects (such as flat ground planes) can be
+         * set to not cast shadows, which might improve shadow quality.
+         * </p>
+         *
+         * <strong>Warning: This API is still experimental and subject to change.</strong>
+         */
+        public ShadowType shadowType;
```
romainguy, line 607: You should set the default value of this field to PCF, it will be null otherwise. Also maybe mark this field as @NonNull

filament/include/filament/View.h
```diff
@@ -272,13 +272,46 @@ class UTILS_PUBLIC View : public FilamentAPI {
...
+         * Sets the number of anisotropic samples to use when sampling a VSM shadow map. If greater
+         * than 0, mipmaps will automatically be generated each frame for all lights.
+         *
+         * The number of anisotropic samples = 2 ^ vsmAnisotropyLog2.
+         *
+         * Only applicable when shadowType is set to ShadowType::VSM.
+         *
+         * @warning This API is still experimental and subject to change.
+         */
+        uint8_t vsmAnisotropyLog2 = 0;
```
romainguy, line 312: See comment above for the name

filament/src/materials/vsmMipmap.mat
pixelflinger: Ans if you did that, you could also avoid the divide -- you'd pass 2/size(level) in the uniform.

bejado: Sure, I can do that

android/filament-android/src/main/java/com/google/android/filament/View.java
bejado: Hm, yeah that might make more sense. I was trying to avoid having structs with only a single option in them. MSAA is set on a per-light basis, but I can see needing to add "blur amount" in the future.

bejado: So in that case, I'll revert the removal of `setShadowType` and rename this struct to `VsmOptions`.

prideout (approved):
I think you will need to update the big string in `test_settings.cpp` to avoid unit test failure.

romainguy (approved):

PR #3187, opened 2020-10-13
Various changes to help clients upgrade Filament.

prideout:
The non-const getSkybox() method is especially useful to client code
that needs to change the clear color, but does not have direct
access to the Renderer object.

### DIFF
--- a/filament/include/filament/Renderer.h
+++ b/filament/include/filament/Renderer.h
@@ -25,6 +25,8 @@
 
 #include <backend/PresentCallable.h>
 
+#include <math/vec4.h>
+
 #include <stdint.h>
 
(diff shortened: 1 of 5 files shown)


romainguy (approved):

filament/include/filament/Renderer.h
```diff
@@ -25,6 +25,8 @@
 
 #include <backend/PresentCallable.h>
 
+#include <math/vec4.h>
```
pixelflinger, line 28: mathfwd should be enough, no?

prideout: Ah, good catch!

PR #3188, opened 2020-10-14
Fix: when IBL loadFromKtx, use the spherical harmonic if available

todaylg:


prideout:
We removed this intentionally in December, see dd8c0a7.

prideout:
There are three high-level IBL utilities: this one for C++ clients (FilamentApp), the one for Java / Kotlin (KtxLoader), and the one for JavaScript.  They all use a common low-level utility (KtxUtility.h), which merely creates a Texture object not the IBL.

Anyway, yes: maybe we should add an option to the three high-level utilities, and add a checkbox to gltf_viewer.

todaylg:
Got it, thanks

PR #3193, opened 2020-10-16
Add subpass parameter type to materials

bejado:


tools/matc/src/matc/ParametersProcessor.cpp
```diff
@@ -219,7 +243,7 @@ static bool processParameters(MaterialBuilder& builder, const JsonishValue& v) {
     bool ok = true;
     for (auto value : jsonArray->getElements()) {
         if (value->getType() == JsonishValue::Type::OBJECT) {
-            ok |= processParameter(builder, *value->toJsonObject());
```
romainguy, line 222: Damn booleans :)

romainguy (approved):
Does matinfo need to be updated too?

bejado:
> Does matinfo need to be updated too?

matinfo will work, although it won't report the subpass as a parameter. In order to do so we'd have to store the subpass in a new material chunk, which I thought was a bit overkill for now.

libs/filamat/src/shaders/CodeGenerator.cpp
```diff
@@ -339,6 +339,29 @@ io::sstream& CodeGenerator::generateSamplers(
...
+        SubpassInfo subpass) const {
+    if (!subpass.isValid) {
+        return out;
+    }
+
+    CString subpassName =
+            SamplerInterfaceBlock::getUniformName("MaterialParams", subpass.name.c_str());
+
+    char const* const typeName = "subpassInput";
+    // In our Vulkan backend, subpass inputs always live in descriptor set 2. (ignored for GLES)
```
prideout, line 352: 👍

prideout (approved):

romainguy:
Yeah I agree it's overkill but we already list the parameters and their types in a chunk. Could we just reuse the existing mechanism for subpass input?

bejado:
> Yeah I agree it's overkill but we already list the parameters and their types in a chunk. Could we just reuse the existing mechanism for subpass input?

It's difficult to reuse the existing mechanism, but I looked into it and I can just add a new chunk type for subpasses.

romainguy:
The Material API lets you query all the declared parameters, it's not just for matinfo, it can be used from Filament directly. We need to make sure this query API works at the very least.

bejado:
> The Material API lets you query all the declared parameters, it's not just for matinfo, it can be used from Filament directly. We need to make sure this query API works at the very least.

Agreed. I'll put that change up in a separate PR to make it more straightforward to review.

PR #3196, opened 2020-10-18
Fix TypeScript binding

andreperezmaselco:
Fix a use of `Texture$Usage` to the new name `TextureUsage`.
This new name was introduced by the commit [5549adc](https://github.com/google/filament/commit/5549adc0206162b5fd26117b1b05b11c44990cde).


prideout:
Thanks for catching this, I really need to set up some automated tests for the typescript bindings.  The actual fix should be to consume a "number" here (because it is a bitmask). Moreover the enum should provide powers of 2 rather than using default values.   I've replaced this PR with #3197.

PR #3198, opened 2020-10-19
Add optional XCB support to PlatformVkLinux.

prideout:
This allows clients to build the Vulkan backend with either XLIB or XCB
support. I did a quick smoke test of the XCB option, but for now we are
continuing to default to XLIB. Note that the native window type used to
create the swap chain differs between these two API's.


romainguy (approved):
Is this something we should somehow handle at runtime?

prideout:
Ah, I forgot that createSwapChain has a "flags" argument.  We could add a flag for XCB, that way we could support both.

prideout:
Oh, actually we should make it possible to build FIlament with neither XLIB nor XCB support; I've already hit a use case for this, when headless rendering on a system that does not have any X11 libraries.  So I'll need to follow this up with another PR.

PR #3203, opened 2020-10-19
Update Material query API to work with subpasses

bejado:
This allows us to query materials for subpass parameters.


bejado:
Still working on updating the JNI layer.

Bestmaker602:
Okay

pixelflinger (commented):
Nice

PR #3206, opened 2020-10-21
Add PNG output to test_ReadPixels.

prideout:


filament/backend/test/test_ReadPixels.cpp
```diff
@@ -16,15 +16,38 @@
...
 #include <fstream>
 
+using namespace filament;
+using namespace filament::backend;
+using namespace image;
+
 namespace {
 
+template<typename T>
+static LinearImage toLinear(size_t w, size_t h, size_t bpr, const uint8_t* src) {
```
romainguy, line 37: Don't we have multiple copies of this method over the code base now? Worth having only one?

romainguy (approved):

filament/backend/test/test_ReadPixels.cpp
prideout: Good catch, added a commit.

prideout:
Hm, iOS is failing, I'll need to figure that out.  This PR isn't really important, I found it somewhat useful though.

filament/backend/CMakeLists.txt
```diff
@@ -307,6 +307,8 @@ if (APPLE)
         filabridge
         getopt
         gtest
+        image
+        imageio
```
bejado, line 311: You should also add these to `BACKEND_TEST_DEPS`.

bejado (changes requested):
The iOS build is failing because the `clang` invocation is missing the `imageio` include directory. Not sure why though..

PR #3212, opened 2020-10-22
Vulkan: fix thread safety regression in Disposer.

prideout:
This fixes a regression introduced by 548b28c6e manifesting as
intermittent assertions. The VulkanDisposer gc() should only be called
from the driver thread.

### DIFF
--- a/filament/backend/src/vulkan/VulkanDriver.h
+++ b/filament/backend/src/vulkan/VulkanDriver.h
@@ -77,17 +77,6 @@ private:
     VulkanDriver(VulkanDriver const&) = delete;
     VulkanDriver& operator = (VulkanDriver const&) = delete;
 
-    void purge() noexcept override {
-        // First we trigger garbage collection of Vulkan resources. This ensures that transient
-        // resources (e.g. the ReadPixels staging buffer) are removed if their refcount is 0, which
-        // allows related BufferDescriptors to move to the purge list.
-        mDisposer.gc();
-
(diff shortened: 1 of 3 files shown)


filament/backend/src/vulkan/VulkanDriver.h
```diff
@@ -77,17 +77,6 @@ class VulkanDriver final : public DriverBase {
     VulkanDriver(VulkanDriver const&) = delete;
     VulkanDriver& operator = (VulkanDriver const&) = delete;
 
-    void purge() noexcept override {
-        // First we trigger garbage collection of Vulkan resources. This ensures that transient
-        // resources (e.g. the ReadPixels staging buffer) are removed if their refcount is 0, which
-        // allows related BufferDescriptors to move to the purge list.
-        mDisposer.gc();
```
bejado, line 84: Does `mDispose.gc()` need to be called elsewhere now?

bejado (approved):

filament/backend/src/vulkan/VulkanDriver.h
prideout: The gc() func is already called in beginFrame().  I recently added a new call inside purge() in order to force the callback to trigger in unit tests that draw only one frame, but that was a wrong thing to do, thread-wise.

If the unit test issue comes up again then yeah we might need to add a call to gc() somewhere.  This PR is basically a partial revert.

PR #3214, opened 2020-10-22
PlatformVkLinux now supports all combos of XLIB and XCB.

prideout:
You can now build Filament with support for both X11 APIs, or neither.
If both are supported, run-time selection is achieved using a SwapChain
flag.

Supporting only one API at build time (or neither) is useful because our
list of "required" VkInstance extensions can vary according to which
API's are supported. During VkInstance creation, we do not have a priori
knowledge about what kinds of swap chains will be created. (headless vs
non-headless, XCB vs XLIB, etc)

Note that some Vulkan implementation (e.g. some builds of SwiftShader)
only support XCB.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -21,6 +21,10 @@ option(FILAMENT_ENABLE_LTO "Enable link-time optimizations if supported by the c
 
 option(FILAMENT_SKIP_SAMPLES "Don't build samples" OFF)
 
+option(FILAMENT_SUPPORTS_XCB "Include XCB support in Linux builds" ON)
+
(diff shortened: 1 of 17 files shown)


CMakeLists.txt
```diff
@@ -84,6 +84,7 @@ if (UNIX AND NOT APPLE AND NOT ANDROID AND NOT WEBGL)
 endif()
 
 if (LINUX)
+    add_definitions(-DFILAMENT_SUPPORTS_XCB -DFILAMENT_SUPPORTS_XLIB)
```
bejado, line 87: Should these be CMake options instead?

bejado (approved):

CMakeLists.txt
prideout: Good idea, done.

PR #3215, opened 2020-10-23
Vulkan: make ReadPixels synchronous.

prideout:
This makes it so that the PixelBufferDescriptor callback triggers at a
time consistent with other backends.

ReadPixels is still asynchronous in the sense that the callback is
triggered on the main thread. However, it is now guaranteed to trigger
during (or before) flushAndWait(), which is less surprising behavior.


prideout:
NOTE: The PR title is a little misleading; ReadPixels is still asynchronous in the sense that the callback is triggered on the main thread.  The only difference is that it is now guaranteed to trigger by the time the following flushAndWait() is invokes, which is consistent with the expectations of some of Google's automated tests.

romainguy (approved):

PR #3222, opened 2020-10-26
Fix colorGradingAsSubpass when dithering is off.

prideout:
This fixes the blank screen seen with Vulkan when all view options are
disabled.


romainguy (approved):

bejado (approved):
Ah that was my fault—thanks for the fix!

PR #3225, opened 2020-10-28
gltfio: add createInstance() to AssetLoader.

### LINKED ISSUE #3137: gltfio: Create additional FilamentInstances dynamically
Related to https://github.com/google/filament/issues/1513, which introduced asset instance creation.
However the current API only allows creating instances when the asset is first loaded, i.e. the number of instances has to be known from the beginning. 

```
 filamentAsset = assetLoader.createInstancedAsset(buffer, filamentInstances)
```

This is not sufficient in our application, as we want to load the asset into memory and then allow the user to place and delete instances dynamically without having to load the asset into memory again and without having a fixed number of instances allocated from the beginning.

I would like to have an API that allows me to create and destroy additional asset instances from a FilamentAsset dynamically.

It could work like this:
```
filamentAsset = assetLoader.createInstancedAsset(buffer)

// ... time passes...

filamentInstance = assetLoader.createInstance(filamentAsset)
```

```
assetLoader.destroyInstance(filamentAsset, filamentInstance)
```

prideout:
Note that this API is on the loader rather than the asset. This is
because the loader knows how to create Filament entities by traversing
a cgltf node hierarchy.

We did not add destroyInstance() because gltfio favors flat arrays for
long term storage of entity lists and instance lists, which would be
slow to shift. We also wish to discourage create/destroy churn since it
is more efficient to pre-allocate instances and selectively add them
into the scene.

Fixes #3137.

### DIFF
--- a/android/gltfio-android/src/main/java/com/google/android/filament/gltfio/AssetLoader.java
+++ b/android/gltfio-android/src/main/java/com/google/android/filament/gltfio/AssetLoader.java
@@ -153,6 +153,32 @@ public class AssetLoader {
     }
 
     /**
+     * Adds a new instance to an instanced asset.
+     *
+     * Use this with caution. It is more efficient to pre-allocate a max number of instances, and
+     * gradually add them to the scene as needed. Instances can also be "recycled" by removing and
+     * re-adding them to the scene.
+     *
+     * NOTE: destroyInstance() does not exist because gltfio favors flat arrays for storage of
+     * entity lists and instance lists, which would be slow to shift. We also wish to discourage
+     * create/destroy churn, as noted above.
+     *
+     * This cannot be called after FilamentAsset#releaseSourceData().
+     * This cannot be called on a non-instanced asset.
+     * Animation is not supported in new instances.
+     * See also AssetLoader#createInstancedAsset().
+     */
+    @Nullable
+    @SuppressWarnings("unused")
+    public FilamentInstance createInstance(@NonNull FilamentAsset asset) {
+        long nativeInstance = nCreateInstance(mNativeObject, asset.getNativeObject());
+        if (nativeInstance == 0) {
+            return null;
+        }
+        return new FilamentInstance(nativeInstance);
+    }
+
+    /**
      * Allows clients to enable diagnostic shading on newly-loaded assets.
      */
     @SuppressWarnings("unused")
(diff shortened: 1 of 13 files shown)


android/gltfio-android/src/main/java/com/google/android/filament/gltfio/AssetLoader.java
```diff
@@ -152,6 +152,32 @@ public FilamentAsset createInstancedAsset(@NonNull Buffer buffer,
...
     }
 
+    /**
+     * Adds a new instance to an instanced asset.
+     *
+     * Use this with caution. It is more efficient to pre-allocate a max number of instances, and
+     * gradually add them to the scene as needed. Instances can also be "recycled" by removing and
+     * re-adding them to the scene.
+     *
+     * NOTE: destroyInstance() does not exist because gltfio favors flat arrays for storage of
```
romainguy, line 162: Would it be that slow? Those arrays won't be big.

romainguy (approved):

android/gltfio-android/src/main/java/com/google/android/filament/gltfio/AssetLoader.java
```diff
@@ -152,6 +152,32 @@ public FilamentAsset createInstancedAsset(@NonNull Buffer buffer,
         return new FilamentAsset(mEngine, nativeAsset);
     }
 
+    /**
+     * Adds a new instance to an instanced asset.
+     *
+     * Use this with caution. It is more efficient to pre-allocate a max number of instances, and
```
bejado, line 158: How do you pre-allocate instances? Maybe add some example code. Also, when do the instances get destroyed? When calling `destroyAsset`?

bejado (approved):

android/gltfio-android/src/main/java/com/google/android/filament/gltfio/AssetLoader.java
prideout: Not super slow, but in some cases, the number of entities in an instance can be large.  Also the texture dependency graph has to be re-checked in case a large texture are still being loaded during instance creation.

android/gltfio-android/src/main/java/com/google/android/filament/gltfio/AssetLoader.java
prideout: Pre-allocation is done via createInstancedAsset, which is the "see also" for this doc block.  The doc block there already says: "Instances are freed when the primary asset is freed."

bejado: Ah I missed that, got it!

PR #3229, opened 2020-10-28
Add support for swizzling readPixels

bejado:
Add support for swizzling during a readPixel operation on Metal.

Question for reviewers: Would it make more sense to add a new `readPixelsSwizzle` driver method?

### DIFF
--- a/filament/backend/include/private/backend/DriverAPI.inc
+++ b/filament/backend/include/private/backend/DriverAPI.inc
@@ -442,7 +442,11 @@ DECL_DRIVER_API_N(readPixels,
         uint32_t, y,
         uint32_t, width,
         uint32_t, height,
-        backend::PixelBufferDescriptor&&, data)
+        backend::PixelBufferDescriptor&&, data,
+        backend::TextureSwizzle, r,
+        backend::TextureSwizzle, g,
+        backend::TextureSwizzle, b,
+        backend::TextureSwizzle, a)
 
 DECL_DRIVER_API_N(readStreamPixels,
         backend::StreamHandle, sh,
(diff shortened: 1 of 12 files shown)


filament/backend/include/private/backend/DriverAPI.inc
```diff
@@ -442,7 +442,11 @@ DECL_DRIVER_API_N(readPixels,
         uint32_t, y,
         uint32_t, width,
         uint32_t, height,
-        backend::PixelBufferDescriptor&&, data)
+        backend::PixelBufferDescriptor&&, data,
+        backend::TextureSwizzle, r,
+        backend::TextureSwizzle, g,
+        backend::TextureSwizzle, b,
+        backend::TextureSwizzle, a)
```
prideout, line 449: I think this is fine, I guess we can't re-use `ReadPixelsOptions` here because of DriverAPI macro magic, right?

bejado: We could in theory, but we'd have to include the public `Renderer.h` header in `DriverAPI.inc`, which requires libbackend to depend on libfilament.

prideout (approved):

filament/backend/include/private/backend/DriverAPI.inc
pixelflinger: The way to do this is to move the struct to DriverEnums.h, and do an alias in Renderer.h.
I've done this a lot before.

pixelflinger (changes requested):
I think we need to implement the swizzle in all backends.
For example by doing the swizzle on the cpu side if needed.

bejado:
> I think we need to implement the swizzle in all backends.
> For example by doing the swizzle on the cpu side if needed.

That wouldn't be fun. I'm going to close this for now, as I don't think there's any immediate need for it.

PR #3232, opened 2020-10-29
Vulkan: fix color attachment list in first subpass.

prideout:
This fixes a slew of validation warnings and errors seen when multiple
subpasses are enabled, starting with:

    Attachment 1 not written by fragment shader; undefined values will
    be written to attachment

The Vulkan driver was using the same color attachment list for both
subpasses, so the first subpass had unused attachments.

Note that Vulkan makes a distinction between color attachments and input
attachments and requires separate lists to be supplied for each subpass,
but our Driver API consolidates everything into a single list. This
should perhaps be refactored at a later date.

### DIFF
--- a/filament/backend/src/vulkan/VulkanBinder.h
+++ b/filament/backend/src/vulkan/VulkanBinder.h
@@ -103,7 +103,7 @@ public:
         VkPipelineColorBlendAttachmentState blending;
         VkPipelineDepthStencilStateCreateInfo depthStencil;
         VkPipelineMultisampleStateCreateInfo multisampling;
-        uint32_t getColorTargetCount;
+        uint32_t colorTargetCount;
     };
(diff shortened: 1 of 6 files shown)


filament/backend/src/vulkan/VulkanBinder.h
```diff
@@ -103,7 +103,7 @@ class VulkanBinder {
         VkPipelineColorBlendAttachmentState blending;
         VkPipelineDepthStencilStateCreateInfo depthStencil;
         VkPipelineMultisampleStateCreateInfo multisampling;
-        uint32_t getColorTargetCount;
```
romainguy, line 106: Eh. good rename :)

romainguy (approved):

bejado (approved):

PR #3243, opened 2020-11-02
Fix bug with Fence timeouts

bejado:
We use the highest value of a `uint64_t` in nanoseconds to denote an infinite timeout for fences. However, `std::chrono::nanoseconds` uses a signed `int64_t`. This causes a negative timeout when passing the value to a `std::cv`.


prideout (approved):
Nice catch!

romainguy (approved):

PR #3247, opened 2020-11-02
Fix FENCE_WAIT_FOR_EVER in Linux.

prideout:
The number of infinite nanoseconds was negative because we asked
chrono for a signed integer, so "wait forever" really meant "do not
wait at all".

### DIFF
--- a/libs/utils/include/utils/linux/Condition.h
+++ b/libs/utils/include/utils/linux/Condition.h
@@ -68,7 +68,7 @@ public:
     std::cv_status wait_until(std::unique_lock<Mutex>& lock,
             const std::chrono::time_point<std::chrono::steady_clock, D>& timeout_time) noexcept {
         // convert to nanoseconds
-        int64_t ns = std::chrono::duration<int64_t, std::nano>(timeout_time.time_since_epoch()).count();
+        uint64_t ns = std::chrono::duration<uint64_t, std::nano>(timeout_time.time_since_epoch()).count();
         using sec_t = decltype(timespec::tv_sec);
         using nsec_t = decltype(timespec::tv_nsec);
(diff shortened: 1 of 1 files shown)


romainguy (approved):
Yay signed vs unsigned

libs/utils/include/utils/linux/Condition.h
```diff
@@ -68,7 +68,7 @@ class Condition {
     std::cv_status wait_until(std::unique_lock<Mutex>& lock,
             const std::chrono::time_point<std::chrono::steady_clock, D>& timeout_time) noexcept {
         // convert to nanoseconds
-        int64_t ns = std::chrono::duration<int64_t, std::nano>(timeout_time.time_since_epoch()).count();
+        uint64_t ns = std::chrono::duration<uint64_t, std::nano>(timeout_time.time_since_epoch()).count();
```
bejado, line 71: What's the type of `D` here? Is it `uint64_t`? Could we use `D` instead of hard-coding `uint64_t`?

bejado (approved):

libs/utils/include/utils/linux/Condition.h
prideout: Ohhhhh let me give that a try and make sure the test still passes.

prideout: D is a duration but replacing `uint64_t` with `typename D::rep` might work.

prideout: Hm, it fails if I use `D::rep` so I'll just stick to uint64_t

bejado: Sure, was worth a shot. Yay templates

PR #3260, opened 2020-11-06
Regenerate BlueVK, use a C++ namespace.

prideout:
This moves all Vulkan entry points into the bluevk namespace
to avoid conflicts with other Vulkan loaders.

This also updates the header files and loader code to version 1.2.159 of
the Vulkan spec, which is currently available at the following links.

https://khronos.org/registry/vulkan/specs/1.2/html/vkspec.html
https://github.com/KhronosGroup/Vulkan-Headers/commit/87451c55aa8b4f00831fab6ffc4b721d0f148102

### DIFF
--- a/filament/backend/src/vulkan/PlatformVkAndroid.cpp
+++ b/filament/backend/src/vulkan/PlatformVkAndroid.cpp
@@ -24,6 +24,8 @@
 
 #include <android/native_window.h>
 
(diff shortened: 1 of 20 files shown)


romainguy (approved):

prideout:
I'll fix the build breakage before merging.

PR #3267, opened 2020-11-09
NOOP backend should not care about GLSL vs SPIRV.

prideout:
This fixes errors that would occur when using the NOOP backend with
materials that were built without OpenGL support.

### DIFF
--- a/filament/src/Material.cpp
+++ b/filament/src/Material.cpp
@@ -44,6 +44,34 @@ namespace filament {
 
 using namespace backend;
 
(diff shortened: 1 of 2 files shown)


romainguy (approved):

bejado (approved):
Surprised we haven't run into this before.

prideout:
>Surprised we haven't run into this before.

It's pretty rare that we build materials without GL support because our root-level CMake unconditionally includes GL in the matc base flags.  At some point we should add a build option that opts out of GL support.

PR #3270, opened 2020-11-10
Add callback API to SwapChain

bejado:
This adds new new APIs to SwapChain:

- `FrameScheduledCallback` (previously `FrameFinishedCallback`). Called when a frame is scheduled by the OS for GPU execution. This is needed on iOS to correctly synchronize between UI elements and 3D rendering.
- `FrameCompletedCallback`. Called when a frame has completed on the GPU (similar to a hard fence).

### DIFF
--- a/filament/backend/include/backend/DriverEnums.h
+++ b/filament/backend/include/backend/DriverEnums.h
@@ -22,6 +22,8 @@
 #include <utils/BitmaskEnum.h>
 #include <utils/unwindows.h> // Because we define ERROR in the FenceStatus enum.
 
+#include <backend/PresentCallable.h>
+
 #include <math/vec4.h>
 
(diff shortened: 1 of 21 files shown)


prideout (approved):

romainguy (approved):
Don't forget the Java APIs :)

prideout:
By the way, I noticed that the comments say the callback is called on an arbitrary thread...  Is this really true?  I'm asking because Filament's other callbacks do not work that way.

bejado:
@romainguy Yes, will update in separate PR.

@prideout That's a good point. At the moment, yes, they really do happen on a separate arbitrary thread. But perhaps it's better to call it on the user thread like we do for `BufferDescriptor`s.

prideout (approved):
That's a clever way of invoking it!

PR #3281, opened 2020-11-12
Vulkan: improve robustness by providing 1x1 fallback.

prideout:


filament/backend/src/vulkan/VulkanContext.cpp
```diff
@@ -957,5 +957,14 @@ void blitColor(VulkanContext* context, const VulkanRenderTarget* dstTarget,
     }
 }
 
+void createEmptyTexture(VulkanContext& context, VulkanStagePool& stagePool) {
+    context.emptyTexture = new VulkanTexture(context, SamplerType::SAMPLER_2D, 1,
```
romainguy, line 961: I assume in theory we should provide similar fallback for texture arrays, 3D textures, etc.?

romainguy (approved):

filament/backend/src/vulkan/VulkanContext.cpp
prideout: Yes, that's what my comment in the other file is mentioning, and unfortunately there is no easy way to determine what kind of placeholder texture to create. I think we would need to reflect over the SPIRV to do this right.

bejado (approved):

PR #3296, opened 2020-11-16
Fix, matc crash when building mobile materials

bejado:


romainguy (approved):

prideout (approved):
Thanks for the fix!  I'm surprised that the commit message does not mention Vulkan.

bejado:
> Thanks for the fix! I'm surprised that the commit message does not mention Vulkan.

Technically this was a matc bug and this could have happened when compiling Metal shaders as well, but we only ever use `textureGather` in Vulkan mobile shaders. I'll update the description though.

PR #3303, opened 2020-11-20
Vulkan: change vkWaitForFences usage for SwiftShader.

prideout:
When passing only 1 fence to vkWaitForFences, the `waitAll` argument should not have any effect, but SwiftShader seems to skip the wait when this argument is set to VK_FALSE.

More specifically, the failure to wait in `acquireWorkCommandBuffer` allows the subsequent destruction of an in-use fence, which causes a TSAN failure with Google's internal tests.

I am consulting with the SwiftShader team on a real fix, meanwhile we can commit this easy workaround.

Note that we have 5 usages of vkWaitForFences, one of which uses multiple fences and should have used VK_TRUE anyway.

### DIFF
--- a/filament/backend/src/vulkan/VulkanContext.cpp
+++ b/filament/backend/src/vulkan/VulkanContext.cpp
@@ -595,7 +595,7 @@ void waitForIdle(VulkanContext& context) {
             }
         }
         if (nfences > 0) {
-            vkWaitForFences(context.device, nfences, fences, VK_FALSE, ~0ull);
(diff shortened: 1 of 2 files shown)


romainguy (approved):

bejado (approved):

prideout:
The SwiftShader team has posted an actual fix for this:

https://swiftshader-review.googlesource.com/c/SwiftShader/+/50628

However I will still merge this, since it permits usage of older SwiftShader and one of the changes improves correctness on our end.

PR #3304, opened 2020-11-20
Fix Windows build.

prideout:
Newer compilers are picky about variable initialization and goto statements.


romainguy (approved):

bejado (approved):

bejado:
Grr looks like Windows is still failing :/

PR #3315, opened 2020-11-23
Update native window documentation

### LINKED ISSUE #3307: Document correct SwapChain native window type for macOS/Metal
**Describe the bug**
The doxygen markup in include/filament/SwapChain.h says that the native window parameter type for OS X is `NSView*`.  But if you're trying to use Metal (as opposed to OpenGL), I'm told that the correct type is `CAMetalLayer*`.

bejado:
Fixes #3307


prideout:
Thanks for fixing this!  ( I suppose it's technically called macOS instead of OSX nowadays. )

prideout (approved):

prideout (approved):

PR #3323, opened 2020-11-27
Fix clearing Rendertarget multiple times

MasTraER:
This fixes a bug Renderer clearing a render target multiple times when the first time and back again.

Usually the default RenderTarget is unintentionally cleared/discarded in the current behavior.
1) Renderer renders a view without custom RenderTarget having full-screen viewport,
2) Renderer renders a view with custom RenderTarget
3) And Renderer renders an another view without custom RenderTarget having different viewport from 1)
You may see junk or empty pixels in outside of the viewport of the view 3), depending on ClearOptions and the backend.

This fix make Renderer put RenderTargets into a set and check them to be cleared only for the first time, instead of comparing the previous RenderTarget to the current RenderTarget.

### DIFF
--- a/filament/src/Renderer.cpp
+++ b/filament/src/Renderer.cpp
@@ -229,9 +229,9 @@ void FRenderer::renderJob(ArenaScope& arena, FView& view) {
     }
 
     FRenderTarget* currentRenderTarget = upcast(view.getRenderTarget());
-    if (mPreviousRenderTarget != currentRenderTarget) {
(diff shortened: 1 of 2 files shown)


MasTraER:
@googlebot I fixed it..

prideout (approved):
Thanks!  This change makes sense to me but I would like one other person to review before we merge it.

It should be noted that #3276 introduced the concept of "previous render target", which was needed to fix a 1P bug.

bejado (approved):
LGTM

PR #3332, opened 2020-12-02
Android: do not export internal symbols.

prideout:
We now export only two groups of symbols: (1) JNI and (2) dynamic linker
symbols used by the companion libraries.

This commit has the following impact:

- Shrinks the arm64 so from 1024 KiB to 954 KiB.
- Shrinks the arm64 aar from 560 KiB to 538 KiB.

Note that the patterns in the map file are tested against mangled names
and therefore should not be used to match the double-colon operator.

### DIFF
--- a/android/filament-android/libfilament-jni.map
+++ b/android/filament-android/libfilament-jni.map
@@ -1,4 +1,20 @@
 LIBFILAMENT {
-  global: *filament*; JNI*;
+  global:
+    Java_com_google_android_filament_*;
+    JNI*;
+    *filament*Camera*;
+    *filament*Skybox*;
+    *filament*Engine*;
+    *filament*RenderableManager*;
+    *filament*Aabb*;
+    *filament*IndirectLight*;
+    *filament*LightManager*;
+    *filament*Transform*;
+    *filament*Material*;
+    *filament*IndexBuffer*;
+    *filament*VertexBuffer*;
+    *filament*Texture*;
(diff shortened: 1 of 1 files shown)


prideout:
For historical purposes, I've attached a dump of `nm -D | grep filament`, applied to the build _after_ this commit:

[symbols.txt](https://github.com/google/filament/files/5630970/symbols.txt)

romainguy (commented):
We should probably automate the generation of this list at some point. It will be easy to forget including new symbols required by companion libraries.

romainguy (approved):

android/filament-android/libfilament-jni.map
```diff
@@ -1,4 +1,20 @@
 LIBFILAMENT {
-  global: *filament*; JNI*;
+  global:
+    Java_com_google_android_filament_*;
+    JNI*;
+    *filament*Camera*;
```
bejado, line 5: Just curious, is `*filament*` not sufficient here?

bejado (approved):

android/filament-android/libfilament-jni.map
romainguy: It should be but it will export more than what we need.

prideout: It is sufficient but it is too wide of a net, e.g. it exposes UibGenerator.

PR #3336, opened 2020-12-03
Fix computeTangents

cjsliuj:
I load a gltf file, then assert failed at this point:
image
After spending some time to investigate, the problem is found here:
image

This is the log's out puts:
image


romainguy (approved):

romainguy:
Asking @prideout to confirm the fix :)

prideout (approved):
Excellent catch, thanks!

(a better commit message would be "gltfio: fix tangents with morphing"

PR #3341, opened 2020-12-05
Add zbloat tool.

prideout:
This tool makes it easy to analyze the composition of Android applications that use Filament.  It generates treemaps that include materials if resgen was invoked with the JSON flag.

See [README.md](https://github.com/google/filament/blob/6ec897f4f79cdb88fa6aa89d7a57953318483504/tools/zbloat/README.md) for usage instructions.  Go [here](https://prideout.net/webtreemap/demo.html) for a demo.

Screen Shot 2020-12-05 at 12 58 33 PM

### DIFF
--- a/tools/zbloat/Dockerfile
+++ b/tools/zbloat/Dockerfile
@@ -1,0 +1,32 @@
+# To run the image and mount the current directory:
+#   docker run --rm -it -v `pwd`:/filament -t zbloat
+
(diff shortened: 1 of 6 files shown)


romainguy (approved):

romainguy:
We should investigate if we can reduce the size of those bundled materials.

PR #3345, opened 2020-12-09
libibl: create "lite" target, remove std::function

prideout:


bejado:
iOS is failing because of some changes needed in the samples. I can make the changes and push to this PR

romainguy (approved):

libs/ibl/CMakeLists.txt
```diff
@@ -35,21 +35,27 @@ include_directories(${PUBLIC_HDR_DIR})
...
     target_compile_options(${TARGET} PRIVATE -ffast-math)
+    target_compile_options(${TARGET}_lite PRIVATE -ffast-math)
 endif()
 
 
 # ==================================================================================================
 # Installation
 # ==================================================================================================
-install(TARGETS ${TARGET} ARCHIVE DESTINATION lib/${DIST_DIR})
+install(TARGETS ${TARGET}_lite ARCHIVE DESTINATION lib/${DIST_DIR})
```
bejado, line 60: Are we no longer distributing the full `ibl` library, only the `ibl_lite` one now? I think it's best to continuing distributing both for now, otherwise this could break clients (as it broke the iOS sample projects)

romainguy: Also we should call it ibl-lite not ibl_lite to be consistent with the other libraries. I noticed we have gltfio_core but all other libraries (.so, .jar, .aar, etc.) use dashes and not underscores.

prideout: Added installation for both variants of the library, and renamed it to `ibl-lite`.  Note that `filamat_lite` uses an underscore.

bejado (approved):

PR #3346, opened 2020-12-09
filamat: minify struct fields.

prideout:
This shrinks the arm64 so file by 24 KiB.  Tested on Android.


bejado (approved):

romainguy (approved):
Unless we *must* support struct minification in filamat-lite, would it be possible/safer to do the renames with glslang's AST APIs?

prideout:
> Unless we _must_ support struct minification in filamat-lite, would it be possible/safer to do the renames with glslang's AST APIs?

@romainguy I tried getting started on this and it's pretty painful, I feel like the AST is hard to use, especially when using it for code mutation rather than analysis.

PR #3348, opened 2020-12-09
math: reduce template bloat for matrices

prideout:
This does not change our API, it merely reduces the number of
non-inlined function instantiations appearing in non-optimized binaries.


bejado (approved):

romainguy (approved):
\o/

I prefer code that uses the proper types :)

PR #3354, opened 2020-12-10
Fix RenderableManager docs for "index"

### LINKED ISSUE #3351: Is the index offset in bytes or number of indices for RenderableManager::Buildfer::geometry()?
I just ran into this `[entity=1, primitive @ 4] offset (48) + count (6) > indexCount (36)`  which is triggered in the `RenderableManager::Builder`: 

```c++
RenderableManager::Builder::Result RenderableManager::Builder::build(Engine& engine, Entity entity) {
        // ....
        // reject invalid geometry parameters
        if (!ASSERT_PRECONDITION_NON_FATAL(entry.offset + entry.count <= entry.indices->getIndexCount(),
                "[entity=%u, primitive @ %u] offset (%u) + count (%u) > indexCount (%u)",
                i, entity.getId(),
                entry.offset, entry.count, entry.indices->getIndexCount())) {
            entry.vertices = nullptr;
            return Error;
        }
        // ....
}
```

The function description says:
```
         * @param offset specifies where in the index buffer to start reading (expressed as a number of bytes)
```

But precondition check assumes offset is specified as an index not in bytes; which makes me wonder if the offset description in the header should be changed/updated.

prideout:
I also double-checked the backends, they indeed treat it as an index
count.

Fixes #3351

### DIFF
--- a/android/filament-android/src/main/java/com/google/android/filament/RenderableManager.java
+++ b/android/filament-android/src/main/java/com/google/android/filament/RenderableManager.java
@@ -131,7 +131,7 @@ public class RenderableManager {
          * @param type specifies the topology of the primitive (e.g., PrimitiveType.TRIANGLES)
          * @param vertices specifies the vertex buffer, which in turn specifies a set of attributes
          * @param indices specifies the index buffer (either u16 or u32)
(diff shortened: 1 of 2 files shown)


romainguy (approved):
Thanks Philip!

Issue #3128, opened 2020-09-26
WebGL triangle demo has GL_INVALID_FRAMEBUFFER_OPERATION


prideout:
The web demos all work except the triangle demo, which does not render anything and emits the following in the console:

     glBlitFramebufferCHROMIUM: framebuffer incomplete
     
The triangle demo is the only WebGL demo that enables multisampling.

prideout:
This GL error occurs when the framebuffer status is the following:

`GL_FRAMEBUFFER_INCOMPLETE_MULTISAMPLE`

romainguy:
Is this fixed now?

prideout:
I can still reproduce this (Chrome 87 + macOS Catalina)

Mivr:
Hi,

I experience the same error ([.WebGL-0x395e5c67d000] GL_INVALID_FRAMEBUFFER_OPERATION: Framebuffer is incomplete.).
In the following demos/tutorials:
https://google.github.io/filament/webgl/tutorial_triangle.html
https://google.github.io/filament/webgl/tutorial_redball.html
https://google.github.io/filament/webgl/tutorial_suzanne.html
https://google.github.io/filament/webgl/parquet.html
https://google.github.io/filament/webgl/suzanne.html
https://google.github.io/filament/webgl/helmet.html

The system is Ubuntu 20.04 Chrome version: 87.0.4280.88

P.S. Surprisingly https://prideout.net/knotess/ demo works just fine for me.

cx20:
@Mivr  This is a known issue. Try Nightly Build.

https://github.com/google/filament/issues/3352

prideout:
On my machine (2017 MacBook), Chrome 87 does not support EXT_multisampled_render_to_texture so the following path is taken:

OpenGLDriver::createRenderTarget is creating two attachments: COLOR_ATTACHMENT0 and DEPTH_ATTACHMENT.  The render target has 4 samples.  The color attachment has 1 attachment and the depth attachment has 4 samples.  The usage flags are appropriately marked such that the color attachment is SAMPLEABLE and the depth attachment is not.

The `here we emulate ext.EXT_multisampled_render_to_texture` path is taken for both attachments.   Thus, ultimately there are four attachments:
1) renderbuffer attachment with 4 samples for color
2) non-ms texture attachment  for color
3) renderbuffer attachment with 4 samples for depth
4) non-ms texture attachment  for depth

[...]

PR #3375, opened 2020-12-18
Fix for: WebGL triangle demo has GL_INVALID_FRAMEBUFFER_OPERATION #3128

jnmaloney:


romainguy:
It seems more like a workaround than an actual fix?

jnmaloney:
Ok sure. The demo runs but won't include the multisample feature. I'm interested to look more into it because I'm seeing GL_INVALID_FRAMEBUFFER_OPERATION for Incomplete Frame Buffer in another application.

prideout:
I would rather we figure out the actual cause, rather than mask the issue.  Another workaround would be to disable post-processing.

Issue #3250, opened 2020-11-03
Web API reference incomplete?


davidwarford:
the javascript API reference is here https://google.github.io/filament/webgl/reference.html
but it seems to be incredibly incomplete, for instance it does not have scene.addEntity or really much of anything, if you look at the filament.d.ts file it declares much more than is present in that reference page, is there a more complete version available?

prideout:
Yes it needs to be fleshed out, the typescript bindings are more up-to-date.  The most authoritative source is `jsbindings.cpp` and our C++ headers.

davidwarford:
Okay. So for the time being I should read the c++ headers then?

prideout:
Yes the C++ headers and the demos in `web/samples` are your best bet.

I think we should probably simplify our JavaScript documentation process.  Currently we use a custom python script to slurp up scattered comments.  Instead, we should simply consolidate all the documentation into the TypeScript definitions file, so that it's one place.

davidwarford:
Yes if the typescript definition file had comments that would have been ideal for me.

prideout:
I added some notes to #5530 yesterday, in a sense this a dup of that bug, closing.

PR #3385, opened 2020-12-21
build.sh: Add TypeScript bindings to the tgz file.

### LINKED ISSUE #3268: filament.d.ts should be included in the web release
Currently you need to pull both the web release and the source code release so that you can get the correct version of filament.d.ts

add filament.d.ts to the web release going forward.

prideout:
This makes the tgz consistent with the npm package. Fixes #3268.

This also removes a stale comment in the script (we stopped packaging the old HTML docs a while back).

Going forward, we would like to use the TS declarations as documentation because they must be kept up to date anyway, and they already contain type information. This was discussed in #3250.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -241,14 +241,10 @@ function build_webgl_with_target {
 
         if [[ "${ISSUE_ARCHIVES}" == "true" ]]; then
             echo "Generating out/filament-${lc_target}-web.tgz..."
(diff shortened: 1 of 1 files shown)


bejado (approved):

romainguy (approved):

davidwarford:
Thank you for your hard work!

PR #3394, opened 2020-12-27
Add fuzzer to integrate Filament with OSS-fuzz

AdamKorcz:
This PR adds a build script to integrate Filament with OSS-fuzz.

I will shortly be setting up the integration on the OSS-fuzz side which will use the build file in this PR.

The build script builds the existing libFuzzer fuzzers and their corpora. Having Filament being fuzzed by OSS-fuzz will allow contributors to OSS-fuzz to work on Filaments fuzzing suite, and it will allow more resources to run the fuzzers.

I will be happy to both work on more fuzzers for the Filament project as well as gradually integrate the oss-fuzz build into Filaments build system.

Signed-off-by: AdamKorcz <adam@adalogics.com>


romainguy:
Thanks for this PR but could you explain the point of it? That build file is added to a third party dependency (spirv) but build what seems to be random parts of our build tree. This doesn't seem to be the right location for this build script at all. Also this seems like it should be part of the build system itself? Last but not least this seems to fuzz our 3p libraries; is the goal to test those libraries? If so this is definitely outside of the scope of our repository.

AdamKorcz:
I apologize for the confusion in the original PR message.
Fuzzing is a way of testing software applications whereby pseudo-random data is passed to a target application with the goal of finding bugs and vulnerabilities. OSS-fuzz is a project by Google that offers free resources and infrastructure for critical open-source projects.
The goal of integrating into OSS-fuzz is to allow continuous fuzzing of Filaments existing fuzzers as well as future fuzzers. This will allow Google to run the fuzzers continuously and notify maintainers if/when bugs are found. 
Looking forward, more fuzzers should be added to the project to test other parts of the code, and including these in the OSS-fuzz integration will allow them to run for longer time and look for harder-to-find bugs.

In that context, I will be happy to provide answers to your specific questions.

> That build file is added to a third party dependency (spirv) but build what seems to be random parts of our build tree.

`oss_fuzz_build.sh` builds the parts of Filament that are required for the existing fuzzers. Some parts are left out as they will not build in the OSS-fuzz environment. `oss_fuzz_build.sh` is meant to be used to build the fuzzers on the OSS-fuzz platform. I have set up a draft of the files on the OSS-fuzz side, and in the Dockerfile (https://github.com/google/oss-fuzz/pull/4896/files#) `oss_fuzz_build.sh` is used to build the fuzzers.

> This doesn't seem to be the right location for this build script at all

The location of the build script is a suggestion. If it does not fit into the structure of the project I am happy to relocate it. Let me know whereto.

> this seems like it should be part of the build system itself?

Absolutely! Integration into the build system can be a large project in itself, and my preference is to begin with a build script as it is in this PR and then move on to integrating this process into the build system. I will be happy to commit to this after Filament has integrated with OSS-fuzz.

> Last but not least this seems to fuzz our 3p libraries; is the goal to test those libraries?

The build script builds all the libFuzzer fuzzers in the Filament repository except for the fuzzer at `third_party/cgltf/fuzz/main.c` which I will be happy to add. The existing fuzzers like you mention target third party libraries. 
The reason that fuzzers for third party libraries are not only hosted in Filaments repository but also run through Chromium fuzzing (based on the existing `BUILD.gn` file) I cannot comment on. I am merely providing the script that also includes those fuzzers in the build process. The Spirv-tools fuzzers were integrated by @bejado.
If it is not important to run the Spirv-tools fuzzers continuously, I will be happy to remove those from the build script.


Finally, I will be happy to write more fuzzers for the Filament project progressively.

romainguy:
Thanks for the context. I know what fuzzing is about but it wasn’t clear what the intentions were in the context of Filament. I don’t think we should build any 3p fuzzer. It’s not Filament’s responsibility and we certainly don’t want to become a point of contact for issues in 3p libraries that already have their own fuzzers. Having fuzzers that are specific to Filament would make more sense (where it’s relevant of course since filament only processes data in a few places, mostly in offline tools).

AdamKorcz:
Sounds fair to me.

Are you open to progressing with OSS-fuzz integration with the build script as it is now as a first step? If so, then I suggest that I add a fuzzer that targets a core part of Filament so we get Filament up and running through OSS-fuzz.

This may be a matter for a separate issue but would you like a PR that removes the existing 3p fuzzers?

AdamKorcz:
I have now removed the build script and added a fuzzer for the material parser. For now we can host the build script for the fuzzers on the OSS-fuzz repository. I have added a build script on the OSS-fuzz side for the fuzzer in this PR: https://github.com/google/oss-fuzz/pull/4896.

Once the fuzzer here is merged, the fuzzer in this PR will be able to run continuously through OSS-fuzz.

I will be happy to progressively add more fuzzers to Filament. There are a few targets that seem ideal to fuzz, but I do need a bit more time to study the codebase.

AdamKorcz:
@romainguy Kind reminder. Let me know if there is anything I can do from my side for landing this.

romainguy:
Thanks for your submission but this is not something we want to take on at this point. I love the idea of fuzzing our code but I don't want to maintain builds in another context than ours.

AdamKorcz:
> I don't want to maintain builds in another context than ours

The build for these fuzzers can be integrated into Filaments CMake build system. I will be happy to do this, if it means that Filament can be fuzzed continuously.

PR #3407, opened 2021-01-06
Fix Filament RenderingTest

bejado:
`RenderingTest` is failing due to our change to use a LUT for tonemapping.

Even when set to `LINEAR` and `ULTRA`, a linear value of 1.0f tonemaps to 254.

### DIFF
--- a/filament/test/filament_rendering_test.cpp
+++ b/filament/test/filament_rendering_test.cpp
@@ -31,10 +31,6 @@
 using namespace filament;
 using namespace backend;
 
-static LinearColor inverseTonemapSRGB(sRGBColor x) {
-    return (x * -0.155) / (x - 1.019);
-}
-
 class RenderingTest : public testing::Test {
 protected:
     Engine* mEngine = nullptr;
@@ -118,7 +114,9 @@ private:
 };
 
 TEST_F(RenderingTest, ClearRed) {
-    mSkybox->setColor({ inverseTonemapSRGB({1, 0, 0}), 1.0f });
+    // We need to clear red to >1 here to ensure a tonemapped value of 255 regardless of LUT
+    // precision.
+    mSkybox->setColor({ 2.0f, 0.0f, 0.0f, 1.0f });
     mView->setDithering(View::Dithering::NONE);
     runTest([this](uint8_t const* rgba, uint32_t width, uint32_t height) {
         EXPECT_EQ(rgba[0], 0xff);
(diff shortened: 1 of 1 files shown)


filament/test/filament_rendering_test.cpp
```diff
@@ -31,6 +31,10 @@
 using namespace filament;
 using namespace backend;
 
+static LinearColor inverseTonemapSRGB(sRGBColor x) {
+    return (x * -0.155) / (x - 1.019);
```
romainguy, line 35: Note this will be just an approximation, esp. with the new default tone mapper. Maybe our tests should just disable tone mapping?

romainguy (approved):

filament/test/filament_rendering_test.cpp
bejado: Is there a way to disable it? We already set tonemapping to `ColorGrading::ToneMapping::LINEAR` for this test.

romainguy: That should disable tone mapping.

bejado: Doesn't `LINEAR` still apply an sRGB transform? Note that inside gltf_viewer, we also use `inverseTonemapSRGB` when setting the background color: https://github.com/google/filament/blob/main/samples/gltf_viewer.cpp#L982-L984

romainguy: You still have the sRGB transform yes. But `inverseTonemapSRGB()` is to invert the tonemapper in an sRGB color. But sRGB shouldn't matter because 0 and 1 will still be 0 and 1 after sRGB transform. So a red background 1,0,0 should still produce 1,0,0 after sRGB transform. What value do we get without `inverseTonemapSRGB`?

bejado: Without it, we get 249 instead of 255.

romainguy: I wonder if that's just an approximation issue caused by the LUT (LUT UVs slightly off for instance). If it was tone mapping I would expect ~225 or 224, not 249 (based on the formula we use for tone mapping).

bejado: Yeah, sounds like I need to take a closer look at what's going on.

bejado: It's definitely a LUT precision issue. When I set the quality to `ULTRA`, I see 254 read back.

romainguy: One thing you could check is whether we don't incorrectly compute the UVs. It's easy to have off by one pixel or half a pixel with LUTs. ULTRA prob helps because it's a higher resolution LUT which hides the problem.

bejado: I did the math by hand, and I don't see anything wrong with conversion to LogC or UV calculation. The LUT values appear correct too. I think it's just an unlucky precision issue. In the case of (1, 0, 0), we're in between two LUT texels.

For the test, we could just bump the clear values to 2.0f, which will put us well in 255 territory.

romainguy (approved):
Thanks. We should also eventually see if we could tweak the LUT parameterization so that values at 0 and 1 could yield better results.

PR #3414, opened 2021-01-12
gltf_viewer: fix sporadic crash when exiting.

prideout:
This could occur when trying to exit the app while textures are still being decoded.

### DIFF
--- a/samples/gltf_viewer.cpp
+++ b/samples/gltf_viewer.cpp
@@ -893,6 +893,7 @@ int main(int argc, char** argv) {
 
     auto cleanup = [&app](Engine* engine, View*, Scene*) {
         app.automationEngine->terminate();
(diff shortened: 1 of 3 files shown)


samples/gltf_viewer.cpp
```diff
@@ -893,6 +893,8 @@ int main(int argc, char** argv) {
 
     auto cleanup = [&app](Engine* engine, View*, Scene*) {
         app.automationEngine->terminate();
+        app.resourceLoader->asyncCancelLoad();
```
bejado, line 896: Should the resource loader be responsible for the `flushAndWait`?

bejado (approved):

samples/gltf_viewer.cpp
prideout: That's a good idea, will make this change.

PR #3417, opened 2021-01-13
fix material use

mirsfang:
This line of code appears to affect Android phones on Qualcomm's platform，This line of code will not work without it

### DIFF
--- a/filament/src/PostProcessManager.cpp
+++ b/filament/src/PostProcessManager.cpp
@@ -1482,6 +1482,7 @@ void PostProcessManager::colorGradingSubpass(DriverApi& driver,  bool translucen
     Handle<HwRenderPrimitive> const& fullScreenRenderPrimitive = engine.getFullScreenRenderPrimitive();
 
     auto const& material = getPostProcessMaterial("colorGradingAsSubpass");
+    material.getMaterialInstance()->use(driver);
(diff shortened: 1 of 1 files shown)


mirsfang:
My device is Samsung S20+(865) and Pixel 3xl (845) , build main branch android is not work , but  xiaomi k30 ultra is normal,  i find main branch this commit `57df4481e99dd45af04e47c9259b44187862bff5` results in，when i revert this line , every one is ok

filament/src/PostProcessManager.cpp
```diff
@@ -1482,6 +1482,7 @@ void PostProcessManager::colorGradingSubpass(DriverApi& driver,  bool translucen
     Handle<HwRenderPrimitive> const& fullScreenRenderPrimitive = engine.getFullScreenRenderPrimitive();
 
     auto const& material = getPostProcessMaterial("colorGradingAsSubpass");
+    material.getMaterialInstance()->use(driver);
```
pixelflinger, line 1485:
```suggestion
    FMaterialInstance* mi = material.getMaterialInstance();
    mi->use(driver);
```

romainguy (approved):

pixelflinger:
nice fix thank you

romainguy:
Thank you!

PR #3424, opened 2021-01-14
basic support for Apple Silicon

pixelflinger:
This set of CLs allows to build and run filament on an aarch64 mac.

This supports setting `CMAKE_OSX_ARCHITECTURES` to either arm64 or x86_64, but not both.
This also supports not setting anything on an Apple Silicon mac.

There are a few things that need fixing before we can do fat binaries:
- we have a .a combine script that doesn't handle fat binaries


prideout (approved):
Nice!  One note: instead of invoking uname from CMake, we might want the easy build script to do it.  When it invokes CMake, it would pass `-DCMAKE_OSX_ARCHITECTURES=arm64`.  In theory this would allow our CMake build to support cross-compilation from a non-M1.

bejado:
I like Philip's suggestion. We can merge this as-is and I can take over that work if you prefer.

pixelflinger:
Yeah `CMAKE_OSX_ARCHITECTURES` looks better -- I'm going to give it a go. I feel rusty.

PR #3445, opened 2021-01-22
Add swap() methods to Allocator.h

AdrianAtGoogle:
This allows StructureOfArraysBase's move constructor to compile.
Removed an apparent workaround in SingleInstanceComponentManager.

### DIFF
--- a/libs/utils/include/utils/Allocator.h
+++ b/libs/utils/include/utils/Allocator.h
@@ -416,6 +416,12 @@ public:
     void* end() const noexcept { return mEnd; }
     size_t getSize() const noexcept { return uintptr_t(mEnd) - uintptr_t(mBegin); }
 
+    friend void swap(HeapArea& lhs, HeapArea& rhs) {
+        using std::swap;
+        swap(lhs.mBegin, rhs.mBegin);
+        swap(lhs.mEnd, rhs.mEnd);
+    }
+
 private:
     void* mBegin = nullptr;
     void* mEnd = nullptr;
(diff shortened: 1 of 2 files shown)


libs/utils/include/utils/Allocator.h
```diff
@@ -416,6 +416,12 @@ class HeapArea {
     void* end() const noexcept { return mEnd; }
     size_t getSize() const noexcept { return uintptr_t(mEnd) - uintptr_t(mBegin); }
 
+    friend void swap(HeapArea& lhs, HeapArea& rhs) {
```
pixelflinger, line 419: noexcept

libs/utils/include/utils/Allocator.h
```diff
@@ -631,6 +637,15 @@ class Arena {
     Arena(Arena const& rhs) noexcept = delete;
     Arena& operator=(Arena const& rhs) noexcept = delete;
 
+    friend void swap(Arena& lhs, Arena& rhs) {
```
pixelflinger, line 640: noexcept

pixelflinger:
Nice one! Thanks!

pixelflinger (changes requested):
I think the swap() can be noexcept.

romainguy:
Note this change breaks the Linux build:

```
In file included from ../../libs/utils/include/utils/NameComponentManager.h:26:
../../libs/utils/include/utils/SingleInstanceComponentManager.h:75:37: error: exception specification of explicitly defaulted move assignment operator does not match the calculated one
    SingleInstanceComponentManager& operator=(SingleInstanceComponentManager&& rhs) noexcept = default;
                                    ^
../../libs/utils/include/utils/NameComponentManager.h:70:50: note: in instantiation of template class 'utils::SingleInstanceComponentManager<utils::details::SafeString>' requested here
class UTILS_PUBLIC NameComponentManager : public SingleInstanceComponentManager<details::SafeString> {
                                                 ^
```

libs/utils/include/utils/Allocator.h
pixelflinger, line 419:
```suggestion
    friend void swap(HeapArea& lhs, HeapArea& rhs) noexcept {
```

libs/utils/include/utils/Allocator.h
pixelflinger, line 640:
```suggestion
    friend void swap(Arena& lhs, Arena& rhs) noexcept {
```

AdrianAtGoogle:
Whoops - sorry about that; I took off the SingleInstanceComponentManager change

pixelflinger (approved):

PR #3447, opened 2021-01-22
Update combine-static-libs.sh script to work with universal binaries

### LINKED ISSUE #3432: combine-static-libs.sh doesn't handle fat .a
`combine-static-libs.sh` doesn't handle fat binaries which prevents us from building for arm64+x86_64 on new macs

To reproduce the issue:

```
mkdir out
cmake -GNinja -Bout -DCMAKE_OSX_ARCHITECTURES="arm64;x86_64"
ninja -C out
```

bejado:
Fixes #3432

### DIFF
--- a/build/linux/combine-static-libs.sh
+++ b/build/linux/combine-static-libs.sh
@@ -1,8 +1,9 @@
 #!/usr/bin/env bash
 set -e
 
+SELF_NAME=$(basename "$0")
+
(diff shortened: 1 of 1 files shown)


romainguy (approved):

pixelflinger (commented):
Tested on the macmini M1, works great.

PR #3448, opened 2021-01-23
start work on fg2

pixelflinger:


filament/src/fg2/FrameGraph.h
```diff
@@ -0,0 +1,270 @@
...
+/** A handle on a resource */
+class FrameGraphHandle;
+
+/** A typed handle on a resource */
+template<typename RESOURCE>
+class FrameGraphId<RESOURCE> : public FrameGraphHandle {
+public:
+    using FrameGraphHandle::FrameGraphHandle;
+    FrameGraphId() noexcept = default;
+    explicit FrameGraphId(FrameGraphHandle r) : FrameGraphHandle(r) { }
```
prideout, line 58: Small idea: make this into a method called `cast<Foo>`, which would potentially help newcomers understand what the code is doing when it constructs one handle from another.  Otherwise they wonder if the resource is getting copied.

filament/src/fg2/FrameGraph.h
```diff
@@ -0,0 +1,270 @@
...
+
+    backend::Handle<backend::HwTexture> texture;
+};
+
+struct TextureSubresource {
+
+};
+
+
+/** A handle on a resource */
```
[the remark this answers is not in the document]
prideout: Sometimes you mention "virtual resource" and "concrete resource".  It might be useful to explain the difference somewhere (maybe here?)

filament/src/fg2/FrameGraph.h
pixelflinger: ah... in fact this is not intended to be public I realize. I mean the user facing API should never use/see a non-typed framegraphid

filament/src/fg2/FrameGraph.h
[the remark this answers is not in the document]
pixelflinger: virtual resource is essentially a framegraph resource, concrete is the corresponding backend ressource (i.e. Handle<>)

filament/src/fg2/details/DependencyGraph.h
```diff
@@ -0,0 +1,189 @@
...
+    DependencyGraph& operator=(const DependencyGraph&) noexcept = delete;
+
+    using NodeID = uint32_t;
+
+    class Node;
+
+    /**
+     * A link between two nodes
+     */
+    class Edge {
```
prideout, line 42: This might look a little cleaner if it were a `struct`.  Technically this merely lets you remove the public keywords (here, as well as two places in the descendant) but it also serves as a hint to humans: "hey, edges are merely holding some data, they do not have any logic"

pixelflinger: Good idea

Issue #2338, opened 2020-04-03
SSAO, Bloom need ping-pong implementation for WebGL


prideout:
WebGL 2.0 on Chrome and Firefox do not allow what they call "feedback loops" for depth, and there are Khronos conformance tests in existence to enforce this.  For more information, see:

https://bugs.chromium.org/p/chromium/issues/detail?id=1066201

We need to provide an alternate FrameGraph for SSAO on some platforms that employs a ping-pong strategy as follows.  Say the mipchain is 5 levels, you create another one that's 4 levels. You do the processing from the first(level 0) to the 2nd, (level 0) then blit the 2nd (level 0) to the first(1), repeat.

MingweiSamuel:
Does this issue prevent SSAO from working on Chrome/Firefox?

prideout:
Yes, last time I checked it appeared to be more than just a warning, it looked wrong.

MingweiSamuel:
I'm wondering since I've been trying to test SSAO but haven't been able to get it to do anything. Should I be seeing the `GL_INVALID_OPERATION` error in the dev console? (I'm not)

For example I tried going to https://google.github.io/filament/webgl/helmet.html and running
```js
app.view.setAmbientOcclusionOptions({
        radius: 0.3,
        power: 1.0,
        bias: 0.0005,
        resolution: 0.5,
        intensity: 1.0,
        quality: 3,
});
app.view.setAmbientOcclusion('SSAO'); // or true
```
But nothing happens, so I'm not sure if that is this issue or I'm doing something wrong.

prideout:
The correct way to enable it is `view.setAmbientOcclusion(Filament.View$AmbientOcclusion.SSAO)`.

I recommend using our TypeScript bindings to help catch stuff like this.

prideout:
This is an issue for bloom as well.

prideout:
Also: bloom does not work on desktop OpenGL on two of my home machines (it floods the viewport with white and flashes), and these are the two machines in my house that happen that have Intel GPU's.

So, in addition to the WebGL problem (`GL ERROR :GL_INVALID_OPERATION : glDrawElements: Source and destination textures of the draw are the same.`) we might need to do this for Desktop GL, at least for Intel GPU's.

prideout:
This is something we've gone back-and-forth on, we've implemented it then removed it, etc.  :)

The bug itself is very old and I'm surprised it's still open.  Closing.

PR #3451, opened 2021-01-26
Bloom: add optional ping-pong implementation.

prideout:
For now, this is only active for WebGL but we will likely activate it for Intel GPU's in an upcoming PR.

#2338

### DIFF
--- a/filament/src/PostProcessManager.cpp
+++ b/filament/src/PostProcessManager.cpp
@@ -1308,17 +1308,29 @@ FrameGraphId<FrameGraphTexture> PostProcessManager::bloomPass(FrameGraph& fg,
         });
     }
 
+    // TODO: this is a feature flag that should be passed in from above.
+    // Reading from one mip while writing to a different mip is covered in the OpenGL ES 3.0
+    // specification under the heading  "Feedback Loops Between Textures and the Framebuffer".
+    // Some OpenGL implementations do not permit this behavior (e.g. macOS Chrome + WebGL 2.0).
+#if defined(__EMSCRIPTEN__)
+    const bool disableFeedback = true;
+#else
+    const bool disableFeedback = false;
+#endif
+
     struct BloomPassData {
         FrameGraphId<FrameGraphTexture> in;
         FrameGraphId<FrameGraphTexture> out;
+        FrameGraphId<FrameGraphTexture> stage;
         FrameGraphRenderTargetHandle outRT[kMaxBloomLevels];
+        FrameGraphRenderTargetHandle stageRT[kMaxBloomLevels];
     };
 
     // downsample phase
     auto& bloomDownsamplePass = fg.addPass<BloomPassData>("Bloom Downsample",
             [&](FrameGraph::Builder& builder, auto& data) {
                 data.in = builder.sample(input);
-                data.out = builder.createTexture("Bloom Texture", {
+                data.out = builder.createTexture("Bloom Out Texture", {
                         .width = width,
                         .height = height,
                         .levels = bloomOptions.levels,
@@ -1326,9 +1338,23 @@ FrameGraphId<FrameGraphTexture> PostProcessManager::bloomPass(FrameGraph& fg,
                 });
                 data.out = builder.write(builder.sample(data.out));
 
+                if (disableFeedback) {
+                    data.stage = builder.createTexture("Bloom Stage Texture", {
+                            .width = width,
+                            .height = height,
+                            .levels = bloomOptions.levels,
+                            .format = outFormat
+                    });
+                    data.stage = builder.write(builder.sample(data.stage));
+                } else {
+                    data.stage = builder.write(builder.sample(data.out));
+                }
+
                 for (size_t i = 0; i < bloomOptions.levels; i++) {
-                    data.outRT[i] = builder.createRenderTarget("Bloom target", {
+                    data.outRT[i] = builder.createRenderTarget("Bloom Out Target", {
                             .attachments = {{ data.out, uint8_t(i) }} });
+                    data.stageRT[i] = builder.createRenderTarget("Bloom Stage Target", {
+                            .attachments = {{ data.stage, uint8_t(i) }} });
                 }
             },
             [=](FrameGraphPassResources const& resources,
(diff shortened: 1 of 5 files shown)


prideout:
To recap, some browsers do not allow reading from a mip while writing to a different mip. In the OpenGL ES 3.0 specification this is covered in section 4.4.3, "Feedback Loops Between Textures and the Framebuffer".

Chromium bug at https://bugs.chromium.org/p/chromium/issues/detail?id=1066201

filament/src/PostProcessManager.cpp
```diff
@@ -1308,27 +1308,50 @@ FrameGraphId<FrameGraphTexture> PostProcessManager::bloomPass(FrameGraph& fg,
         });
     }
 
+    // TODO: this is a feature flag that should be passed in from above.
```
bejado, line 1311: I would maybe add a comment here explaining what "feedback" means and why it's necessary to disable it in certain cases.

bejado (approved):

filament/src/PostProcessManager.cpp
prideout: Good catch, this term is from the GL spec, I amended the comment by fleshing out this comment block.

filament/src/PostProcessManager.cpp
```diff
@@ -1308,27 +1308,53 @@ FrameGraphId<FrameGraphTexture> PostProcessManager::bloomPass(FrameGraph& fg,
...
+                if (disableFeedback) {
+                    data.stage = builder.createTexture("Bloom Stage Texture", {
+                            .width = width,
+                            .height = height,
+                            .levels = bloomOptions.levels,
+                            .format = outFormat
+                    });
+                    data.stage = builder.write(builder.sample(data.stage));
+                } else {
+                    data.stage = builder.write(builder.sample(data.out));
```
pixelflinger, line 1350: one thing that makes me nervous here is that `data.out` becomes invalid after calling `sample`. So it's dangerous to have it around. At least we need a comment, and maybe reset it to 0.

filament/src/PostProcessManager.cpp
```diff
@@ -1308,27 +1308,53 @@ FrameGraphId<FrameGraphTexture> PostProcessManager::bloomPass(FrameGraph& fg,
...
+                }
+
                 for (size_t i = 0; i < bloomOptions.levels; i++) {
-                    data.outRT[i] = builder.createRenderTarget("Bloom target", {
+                    data.outRT[i] = builder.createRenderTarget("Bloom Out Target", {
                             .attachments = {{ data.out, uint8_t(i) }} });
```
pixelflinger, line 1355: I dont know why it doesn't fail -- but data.out is invalid here in the normal case (see my comment above)

filament/src/PostProcessManager.cpp
```diff
@@ -1353,45 +1380,53 @@ FrameGraphId<FrameGraphTexture> PostProcessManager::bloomPass(FrameGraph& fg,
...
     // upsample phase
     auto& bloomUpsamplePass = fg.addPass<BloomPassData>("Bloom Upsample",
             [&](FrameGraph::Builder& builder, auto& data) {
-                data.in = builder.sample(input);
-                data.out = builder.write(input);
+                data.out = builder.write(builder.sample(output));
+                if (disableFeedback) {
+                    data.stage = builder.write(builder.sample(stage));
+                } else {
+                    data.stage = builder.write(data.out);
```
pixelflinger, line 1416: same problem here

filament/src/PostProcessManager.cpp
```diff
@@ -1353,45 +1380,53 @@ FrameGraphId<FrameGraphTexture> PostProcessManager::bloomPass(FrameGraph& fg,
                 mi->setParameter("invHighlight", std::isinf(bloomOptions.highlight) ? 0.0f : 1.0f / bloomOptions.highlight);
 
                 for (size_t i = 0; i < bloomOptions.levels; i++) {
-                    auto hwOutRT = resources.get(data.outRT[i]);
+                    const bool parity = (i % 2) == 0;
```
pixelflinger, line 1383: does this code work if the parity if `bloomOptions.levels` changes?

web/filament-js/extensions.js
```diff
@@ -251,11 +251,12 @@ Filament.loadClassExtensions = function() {
 
     /// setBloomOptions ::method::
     /// overrides ::argument:: Dictionary with one or more of the following properties: \
-    /// dirtStrength, strength, resolution, anomorphism, levels, blendMode, threshold, enabled.
+    /// enabled, strength, resolution, anomorphism, levels, blendMode, threshold, highlight.
     /// NOTE: dirt texture is not yet supported in the JavaScript API.
     Filament.View.prototype.setBloomOptions = function(overrides) {
         const options = {
             dirtStrength: 0.2,
+            highlight: 1000.0,
```
pixelflinger, line 259: should be in another PR?

pixelflinger (changes requested):
That makes me very nervous!  My preference would be to disable bloom entirely on these GPUs.

web/filament-js/extensions.js
prideout: At least it's a separate commit but yeah it should've made a sep PR too.

romainguy:
Mathias, that would mean disabling bloom entirely on Intel GPUs (so most laptops basically) and in WebGL.

filament/src/PostProcessManager.cpp
prideout: Yes, I had to think about this: the parity of the downsample pass is straightforward but the parity of the upsample is adjusted accordingly.

filament/src/PostProcessManager.cpp
prideout: The current code is trying to make it so that "stage" is just an alias for "out".  Would is make sense to add an alias method to the builder for this?

Another idea is leaving `bloomPass` exactly as it is, and instead simply create a brand new method called `bloomPingPong`.  There will be some code duplication but maybe it will be safer.

@pixelflinger WDYT?

web/filament-js/extensions.js
prideout: Gah, you're right, there is some cross-over between the two CL's.

prideout:
Discarding this in favor of a new cleaner approach, stay tuned.

PR #3453, opened 2021-01-26
Add unit test for feedback loops.

prideout:
This test passes on my Radeon GPU MacBook but fails with an Intel GPU.

### DIFF
--- a/filament/backend/test/test_FeedbackLoops.cpp
+++ b/filament/backend/test/test_FeedbackLoops.cpp
@@ -1,0 +1,197 @@
+/*
+ * Copyright (C) 2021 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


filament/backend/test/test_FeedbackLoops.cpp
```diff
@@ -0,0 +1,197 @@
...
+////////////////////////////////////////////////////////////////////////////////////////////////////
+
+static std::string downsampleVs = R"(#version 450 core
+layout(location = 0) in vec4 mesh_position;
+void main() {
+    // Hack: move and scale triangle so that it covers entire viewport.
+    gl_Position = vec4((mesh_position.xy + 0.5) * 5.0, 0.0, 1.0);
+})";
+
+static std::string downsampleFs = R"(#version 450 core
```
romainguy, line 44: These shaders won't work on GLES/WebGL right? If so maybe make sure you don't build the test on those platforms? Or at least not run it on those platforms.

romainguy (approved):

filament/backend/test/test_FeedbackLoops.cpp
prideout: The existing sister tests (eg `test_MRT.cpp`) use version 450 as well, and yeah they only run on macOS right now.

bejado (approved):
Yay backend test!

pixelflinger (approved):

PR #3456, opened 2021-01-26
Plumbing for disabling feedback loops on Intel.

prideout:
This is a follow-on to #3451, which adds optional ping-pong support to the bloom generator.

### DIFF
--- a/filament/backend/src/opengl/OpenGLContext.cpp
+++ b/filament/backend/src/opengl/OpenGLContext.cpp
@@ -71,6 +71,13 @@ OpenGLContext::OpenGLContext() noexcept {
         }
     } else if (strstr(renderer, "Intel")) {
         bugs.vao_doesnt_store_element_array_buffer_binding = true;
+
+        // We have observed bloom flashing artifacts (as well as failure in test_FeedbackLoop) on a
+        // MacBook Air with an Intel GPU, as well as an Intel NUC Windows PC. For now we're assuming
+        // that all Intel GPU's cannot handle feedback loops properly.
+        // TODO: match a more specific string pattern after Intel discrete GPU's arrive.
+        bugs.disable_feedback_loops = true;
+
     } else if (strstr(renderer, "PowerVR") || strstr(renderer, "Apple")) {
     } else if (strstr(renderer, "Tegra") || strstr(renderer, "GeForce") || strstr(renderer, "NV")) {
     } else if (strstr(renderer, "Vivante")) {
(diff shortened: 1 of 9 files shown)


filament/backend/src/opengl/OpenGLContext.cpp
```diff
@@ -71,6 +71,7 @@ OpenGLContext::OpenGLContext() noexcept {
         }
     } else if (strstr(renderer, "Intel")) {
         bugs.vao_doesnt_store_element_array_buffer_binding = true;
+        bugs.disable_feedback_loops = true;
```
romainguy, line 74: We may want to leave a TODO to re-evaluate once Intel's discrete GPUs hit the market.

romainguy (approved):

filament/backend/src/opengl/OpenGLContext.cpp
prideout: Done.

filament/src/PostProcessManager.cpp
```diff
@@ -1272,7 +1272,7 @@ FrameGraphId<FrameGraphTexture> PostProcessManager::dof(FrameGraph& fg,
 
 FrameGraphId<FrameGraphTexture> PostProcessManager::bloomPass(FrameGraph& fg,
         FrameGraphId<FrameGraphTexture> input, TextureFormat outFormat,
-        View::BloomOptions& bloomOptions, float2 scale) noexcept {
+        View::BloomOptions& bloomOptions, float2 scale, bool disableFeedbackLoops) noexcept {
```
pixelflinger, line 1275: I think this state should exist in the postprocess manager.  You can initilize it in `PostProcessManager::init`. This way we don't have to pass the flag in all APIs.

pixelflinger (changes requested):

filament/src/PostProcessManager.cpp
prideout: Ah, good idea, done.

filament/src/PostProcessManager.cpp
```diff
@@ -187,7 +187,8 @@ PostProcessManager::PostProcessMaterial& PostProcessManager::getPostProcessMater
 
 #define MATERIAL(n) MATERIALS_ ## n ## _DATA, MATERIALS_ ## n ## _SIZE
 
-void PostProcessManager::init() noexcept {
+void PostProcessManager::init(bool disableFeedbackLoops) noexcept {
+    mDisableFeedbackLoops = disableFeedbackLoops;
```
pixelflinger, line 191: you don't need to pass the boolean here. You have access to `DriverApi` just below.

pixelflinger (changes requested):

filament/src/PostProcessManager.cpp
prideout: Gah, I'm so out of it.  Fixed, and squashed.

PR #3459, opened 2021-01-26
Modulate emissive by alpha on blended objects

### LINKED ISSUE #3454: Render transparent and emissive gltf model incorrectly
I was using this [gltf model](https://sketchfab.com/3d-models/paladins-book-ancient-knights-secrets-07bc6364101c4bd4adaa3d0cee1aaa3e).
It was expected to look like the preview provided on Sketchfab.
image

But when I use the Modelviewer class to load and view it on an Android Device,  it looks incorrectly rendered as following. It seems that the alpha map was ignored when rendering emissive.
image

Besides, when the transparent part overlays with other parts, it looks wrong from some angle. I've gone through issue #2744, but I haven't figured out how to set the transparency mode to twoPassesTwoSides in java when using gltf model.
image

romainguy:
This prevents light from being emitted from completely transparent pixels.

Fixes #3454.

### DIFF
--- a/shaders/src/shading_lit.fs
+++ b/shaders/src/shading_lit.fs
@@ -299,7 +299,7 @@ void addEmissive(const MaterialInputs material, inout vec4 color) {
 #if defined(MATERIAL_HAS_EMISSIVE)
     highp vec4 emissive = material.emissive;
     highp float attenuation = mix(1.0, frameUniforms.exposure, emissive.w);
-    color.rgb += emissive.rgb * attenuation;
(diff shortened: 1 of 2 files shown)


prideout (approved):
Seems very reasonable.  I wasn't able to find anything about this in the glTF spec or bug tracker.

romainguy:
The spec says *nothing* about this. And the asset to repro the issue crashes Babylon.js and Don McCurdy's viewer… :)

PR #3461, opened 2021-01-27
gltfio: disable shader optimizations by default

prideout:


prideout:
Note that ubershaders (which are always used on Android and WebGL) are still optimized by default.

romainguy (approved):
Could you also update the Java API please?

romainguy:
Thanks Philip! I started looking at multithreading filamat, it seems doable, I just need to cleanup some of our code first (we use class members as "globals" between methods, gross).

romainguy:
Oh and you should update RELEASE_NOTES so devs are aware of this change.

prideout:
> Could you also update the Java API please?

The Java API is hardcoded to use ubershaders, I think I was trying to avoid a filamat dependency there.

prideout:
> Oh and you should update RELEASE_NOTES so devs are aware of this change.

Done.

romainguy:
> > Could you also update the Java API please?
> 
> The Java API is hardcoded to use ubershaders, I think I was trying to avoid a filamat dependency there.

Ah right we never created gltfio-lite and gltfio-full. Übershaders come with limitations unfortunately, so ideally we'd want a version of gltfio without spirvopt/cross dependencies but shader generation. That's a problem for another day.

PR #3464, opened 2021-01-28
Multi-thread shaders generation when compiling materials

romainguy:
We currently generate 128 shader variants per target API per material. On a 12-core MacPro, running `matc` on
`sandboxCloth.mat` takes ~12 seconds. After multi-threading using `JobSystem` it takes ~2 seconds. There are
many optimizations we could do to further improve these numbers, mainly around allocations, especially in
`GLSLPostProcessor`.

Unfortunately we rely on `glslang` and `spirv-*` which don't make any thread safety guarantees. Actually `glslang`
is particularly hostile since it uses globals in unsafe ways in various places. This is worked around by first running
a single job by itself that ensures `glslang` is properly initialized.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -8,7 +8,7 @@ A new header is inserted each time a *tag* is created.
 - Added `sheenColor` and `sheenRoughness` properties to materials to create cloth/fabric
 - gltfio: added support for `KHR_materials_sheen`
 - gltfio: shader optimizations are now disabled by default, unless opting in or using ubershaders
-- `MaterialBuilder::build()` now excepts a reference to a `JobSystem` to multi-thread shaders
+- `MaterialBuilder::build()` now expects a reference to a `JobSystem` to multi-thread shaders
   generation. A `JobSystem` can be obtained with `Engine::getJobSystem()` when using Filament,
   or created directly otherwise (⚠️ **API change**).
 
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -8,6 +8,9 @@ A new header is inserted each time a *tag* is created.
 - Added `sheenColor` and `sheenRoughness` properties to materials to create cloth/fabric
 - gltfio: added support for `KHR_materials_sheen`
 - gltfio: shader optimizations are now disabled by default, unless opting in or using ubershaders
+- `MaterialBuilder::build()` now excepts a reference to a `JobSystem` to multi-thread shaders
+  generation. A `JobSystem` can be obtained with `Engine::getJobSystem()` when using Filament,
+  or created directly otherwise (⚠️ **API change**).
 
 ## v1.9.10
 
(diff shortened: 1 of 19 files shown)


libs/filamat/src/MaterialBuilder.cpp
```diff
@@ -641,101 +653,138 @@ bool MaterialBuilder::generateShaders(const std::vector<Variant>& variants, Chun
...
+
+                    textDictionary.addText(metalEntry.shader);
+                    metalEntries.push_back(metalEntry);
+                }
 #endif
+            });
+
+            // NOTE: We run the first job separately to work the lack of thread safety
+            //       guarantees in glslang. This library performs unguarded global
+            //       operations on first use.
```
bejado, line 774: Really? Wth glslang...

bejado (approved):
Nice! Thank you for doing this

RELEASE_NOTES.md
```diff
@@ -8,6 +8,9 @@ A new header is inserted each time a *tag* is created.
 - Added `sheenColor` and `sheenRoughness` properties to materials to create cloth/fabric
 - gltfio: added support for `KHR_materials_sheen`
 - gltfio: shader optimizations are now disabled by default, unless opting in or using ubershaders
+- `MaterialBuilder::build()` now excepts a reference to a `JobSystem` to multi-thread shaders
```
prideout, line 11:
```suggestion
- `MaterialBuilder::build()` now expects a reference to a `JobSystem` to multi-thread shaders
```

android/filamat-android/src/main/java/com/google/android/filament/filamat/MaterialBuilder.java
```diff
@@ -421,9 +433,55 @@ public MaterialBuilder variantFilter(int variantFilter) {
...
+     * or an invalid job system provider, a temporary job system will be created which
+     * is less efficient than reusing an existing job system.
+     *
+     * Currently the only valid type of job system provider is an <code>Engine</code>
+     * instance from the main Filament library (<code>com.google.android.filament.Engine</code>).
+     *
+     * If you are using Filament and the filamat library together you <em>must</em> pass an
+     * <code>Engine</code> as the job system provider, <em>or</em> invoke
+     * <code>MaterialBuilder</code> from a thread other than the thread used to invoke Filament
+     * APIs.
```
prideout, line 466: 👍

android/samples/sample-material-builder/src/main/java/com/google/android/filament/material_builder/MainActivity.kt
```diff
@@ -216,7 +216,9 @@ class MainActivity : Activity() {
                 // variant of the filamat library.
                 .optimization(MaterialBuilder.Optimization.NONE)
 
-                .build()
+                // When compiling more than one material it is more efficient to pass an Engine
```
prideout, line 219:
```suggestion
                // When compiling more than one material variant, it is more efficient to pass an Engine
```

prideout (approved):

PR #3472, opened 2021-01-29
Rewrite and enhance test_FeedbackLoops.

prideout:
Previously this test was actually finding a ReadPixels failure on Intel GPU's, rather than a failure of feedback loop behavior. This was due to the fact that it was calling ReadPixels on a non-zero miplevel, which seems to be unsupported on Intel GPU's.

The test now follows the downsample phase with an upsample phase, just like the real bloom effect. This allows us to issue ReadPixels on the base level, thus making the test more targetted to rendering feedback loops.

In a subsequent PR, I will add `setMinMaxLevels` to the driver API and modify the test accordingly.  Good (Radeon) and Bad (Intel) screenshots are below.

good
bad

### DIFF
--- a/filament/backend/test/test_FeedbackLoops.cpp
+++ b/filament/backend/test/test_FeedbackLoops.cpp
@@ -21,6 +21,9 @@
 #include "ShaderGenerator.h"
 #include "TrianglePrimitive.h"
 
+#include <utils/Hash.h>
+#include <utils/Log.h>
+
 #include <fstream>
 
 #ifndef IOS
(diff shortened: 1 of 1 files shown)


filament/backend/test/test_FeedbackLoops.cpp
```diff
@@ -34,29 +36,81 @@ using namespace image;
...
+static std::string fullscreenVs = R"(#version 450 core
 layout(location = 0) in vec4 mesh_position;
 void main() {
     // Hack: move and scale triangle so that it covers entire viewport.
     gl_Position = vec4((mesh_position.xy + 0.5) * 5.0, 0.0, 1.0);
 })";
 
-static std::string downsampleFs = R"(#version 450 core
+static std::string fullscreenFs = R"(#version 450 core
+precision mediump int; precision highp float;
```
romainguy, line 47: Using highp shouldn't matter on desktop, but it's nice for readability

filament/backend/test/test_FeedbackLoops.cpp
```diff
@@ -66,132 +120,128 @@ TEST_F(BackendTest, FeedbackLoops) {
...
+            backend::SamplerGroup samplers(1);
+            backend::SamplerParams sparams = {};
+            sparams.filterMag = SamplerMagFilter::LINEAR;
+            sparams.filterMin = SamplerMinFilter::LINEAR_MIPMAP_NEAREST;
+            samplers.setSampler(0, texture, sparams);
+            auto sgroup = getDriverApi().createSamplerGroup(samplers.getSize());
+            getDriverApi().updateSamplerGroup(sgroup, std::move(samplers.toCommandStream()));
+            auto ubuffer = getDriverApi().createUniformBuffer(sizeof(MaterialParams),
+                    backend::BufferUsage::STATIC);
+            getDriverApi().makeCurrent(swapChain, swapChain);
```
romainguy, line 183: Maybe just create a local reference to the driver API object so you don't call the method all over the place? It would be just a bit cleaner

filament/backend/test/test_FeedbackLoops.cpp
```diff
@@ -66,132 +120,128 @@ TEST_F(BackendTest, FeedbackLoops) {
...
-                {});                                       // stencil
-
-        // Fill the texture with interesting colors.
-        const size_t size = 512 * 512 * 4;
+            SamplerType::SAMPLER_2D, kNumLevels, kTexFormat, 1, kTexWidth, kTexHeight, 1, usage);
+
+        // Create a RenderTarget for each miplevel.
+        Handle<HwRenderTarget> renderTargets[kNumLevels];
+        for (uint8_t level = 0; level < kNumLevels; level++) {
+            printf("Level %d: %d x %d\n", level, kTexWidth >> level, kTexHeight >> level);
```
romainguy, line 143: You should use our iostream/logger because it'll work nicer on Android if we ever run this test on Android.

romainguy (approved):

PR #3473, opened 2021-01-29
Compile Android libraries with fPIC

romainguy:


romainguy:
The Android build is currently broken because it seems we need to build our libraries with `-fPIC`.

filament/CMakeLists.txt
```diff
@@ -390,7 +390,6 @@ set(LINUX_LINKER_OPTIMIZATION_FLAGS
 )
 
 set(LINUX_COMPILER_FLAGS
-        -fPIC
 )
```
prideout, line 393: Remove these codelines entirely?

prideout (approved):

PR #3474, opened 2021-01-30
Add setMinMaxLevels to the DriverAPI for spec compliance.

prideout:
This adds setMinMaxLevels to the DriverAPI and calls it from PostProcessManager and the new unit test. The allows us to be compliant to the OpenGL ES 3.x specifications, which all state:

```
...if TEXTURE_BASE_LEVEL and TEXTURE_MAX_LEVEL exclude any levels containing
image(s) attached to the currently bound draw framebuffer object, then [...]
the above rule will not cause the values of rendered fragments to be undefined.
```

The above quote is taken from the **Rendering Feedback Loops** section.

The new unit test now looks much better on Intel GPUs.  Unfortunately, Chrome still gives an INVALID_OPERATION and my Intel GPU still exhibits issues with the actual bloom effect. So, for now we need to keep the ping-pong implementation for Chrome and for Intel.

### DIFF
--- a/filament/src/PostProcessManager.cpp
+++ b/filament/src/PostProcessManager.cpp
@@ -1350,6 +1350,7 @@ FrameGraphId<FrameGraphTexture> PostProcessManager::bloomPass(FrameGraph& fg,
                         .filterMin = SamplerMinFilter::LINEAR /* level is always 0 */
                 });
                 mi->setParameter("level", 0.0f);
+                driver.setMinMaxLevels(hwIn, 0, 0);
                 mi->setParameter("threshold", bloomOptions.threshold ? 1.0f : 0.0f);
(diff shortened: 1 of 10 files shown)


romainguy (approved):
Can't wait to ditch GL...

bejado (approved):

filament/src/PostProcessManager.cpp
```diff
@@ -1373,6 +1374,7 @@ FrameGraphId<FrameGraphTexture> PostProcessManager::bloomPass(FrameGraph& fg,
                             .filterMin = SamplerMinFilter::LINEAR_MIPMAP_NEAREST
                     });
                     mi->setParameter("level", float(i));
+                    driver.setMinMaxLevels(hwIn, i, i);
```
pixelflinger, line 1377:
```suggestion
                    driver.setMinMaxLevels(hwIn, i, i); // safe because we're using LINEAR_MIPMAP_NEAREST
```

pixelflinger (approved):

PR #3484, opened 2021-02-01
Do not pass texture id when renderbuffer id is expected.

### LINKED ISSUE #3128: WebGL triangle demo has GL_INVALID_FRAMEBUFFER_OPERATION

prideout:
glFramebufferRenderbuffer expects a renderbuffer, not a texture.
Trying to pass a texture results in INVALID_OPERATION on some platforms.

Fixes #3128.

### DIFF
--- a/filament/backend/src/opengl/OpenGLDriver.cpp
+++ b/filament/backend/src/opengl/OpenGLDriver.cpp
@@ -929,10 +929,6 @@ void OpenGLDriver::framebufferTexture(backend::TargetBufferInfo const& binfo,
                 if (any(t->usage & TextureUsage::SAMPLEABLE)) {
                     glFramebufferTexture2D(GL_FRAMEBUFFER, attachment,
                             target, t->gl.id, binfo.level);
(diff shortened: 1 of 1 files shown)


filament/backend/src/opengl/OpenGLDriver.cpp
```diff
@@ -929,10 +929,6 @@ void OpenGLDriver::framebufferTexture(backend::TargetBufferInfo const& binfo,
                 if (any(t->usage & TextureUsage::SAMPLEABLE)) {
                     glFramebufferTexture2D(GL_FRAMEBUFFER, attachment,
                             target, t->gl.id, binfo.level);
-                } else {
-                    assert(target == GL_TEXTURE_2D);
-                    glFramebufferRenderbuffer(GL_FRAMEBUFFER, attachment,
-                            GL_RENDERBUFFER, t->gl.id);
```
prideout, line 935: Note that the renderbuffer was already attached on line 913, no need to attach it twice.

Issue #3485, opened 2021-02-02
Vulkan does not support depth resolves.


prideout:
Transmissive models do not display correctly.  To repro, download the attached JSON file and do:

```
gltf_viewer -a vulkan -u --settings viewopts04.json TransmissionTest.gltf
```

This causes validation errors and triggers an assert in debug builds.

[viewopts04.json.zip](https://github.com/google/filament/files/5914246/viewopts04.json.zip)

PR #3487, opened 2021-02-03
matc: add --raw argument for internal use.

prideout:
This feature can generate SPIRV blobs and will be used by VulkanBlitter.

(#3485)

### DIFF
--- a/tools/matc/src/matc/MaterialCompiler.cpp
+++ b/tools/matc/src/matc/MaterialCompiler.cpp
@@ -16,7 +16,6 @@
 
 #include "MaterialCompiler.h"
 
-#include <functional>
 #include <memory>
 #include <iostream>
 
@@ -32,6 +31,10 @@
 #include "JsonishLexer.h"
 #include "JsonishParser.h"
 #include "ParametersProcessor.h"
+
+#include <GlslangToSpv.h>
+
+#include "sca/builtinResource.h"
 
 using namespace utils;
 using namespace filamat;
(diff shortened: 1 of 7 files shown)


tools/matc/src/matc/MaterialCompiler.cpp
```diff
@@ -441,4 +453,49 @@ bool MaterialCompiler::parseMaterial(const char* buffer, size_t size,
...
+    msg = (EShMessages) (EShMessages::EShMsgVulkanRules | EShMessages::EShMsgSpvRules);
+
+    tShader.setAutoMapBindings(true);
+
+    bool ok = tShader.parse(&DefaultTBuiltInResource, version, false, msg);
+    if (!ok) {
+        std::cerr << "ERROR: Unable to parse " << ext << ":" << std::endl;
+        std::cerr << tShader.getInfoLog() << std::endl;
+        return false;
+    }
```
bejado, line 479: I think we need to make sure we link the program, just as we do for normal materials:
https://github.com/google/filament/blob/9f68daeb8f18b8d13e66105a009e59644531a8a2/libs/filamat/src/GLSLPostProcessor.cpp#L202-L209

bejado (changes requested):

tools/matc/src/matc/MaterialCompiler.cpp
prideout: Good catch, fixed.

bejado:
Hm, Windows fails with

```
CUSTOMBUILD : error : Unable to parse vs: [D:\a\filament\filament\out\cmake-mt\filament\backend\vkshaders.vcxproj]
CUSTOMBUILD : error : 0:12: '' :  syntax error, unexpected DOT [D:\a\filament\filament\out\cmake-mt\filament\backend\vkshaders.vcxproj]
CUSTOMBUILD : error : 1 compilation errors.  No code generated. [D:\a\filament\filament\out\cmake-mt\filament\backend\vkshaders.vcxproj]
```

I can't think of why though..

prideout:
> Hm, Windows fails

I have a theory that this is due to a lack of null terminator.  Amended a commit, fingers crossed.

PR #3491, opened 2021-02-03
Fix MSAA for OpenGL ES 3.0 core.

### LINKED ISSUE #3128: WebGL triangle demo has GL_INVALID_FRAMEBUFFER_OPERATION

prideout:
This fixes `GL_FRAMEBUFFER_INCOMPLETE_MULTISAMPLE` caused by the following spec clause being violated:

```
The value of RENDERBUFFER_SAMPLES [must be] the same for all
attached renderbuffers and, if the attached images are a mix of
renderbuffers and textures, the value of RENDERBUFFER_SAMPLES
is zero.
```

Fixes #3128.


pixelflinger:
btw, I think the language in the spec makes sense for ES3.0 because ES3.0 doesn't support multi sampled textures, so if there is a mix of renderbuffers and textures, since textures can't be multi sampled, it implies that all attachments can't be.

So is  the problem that we ended up with a mix of renderbuffers/textures and multisampling was enabled?  This shouldn't be the case for the resolve FBO, by definition it's not multisampled.

PR #3492, opened 2021-02-03
Upgrade MoltenVK.

prideout:


romainguy (approved):

prideout (commented):
@bejado do we need the ios library here?  I can't remember why we added it, it is fairly large.

bejado:
@prideout No, we shouldn't need it anymore. My original iOS Filament port used MoltenVK, but that's been long since abandoned.

bejado (approved):

prideout:
> @prideout No, we shouldn't need it anymore. My original iOS Filament port used MoltenVK, but that's been long since abandoned.

Ah, that makes sense thanks.  I amended the commit...23.5 MB removed from the repo!

PR #3493, opened 2021-02-03
Add getVertexIndex API for vertex shaders

bejado:
In order to support vertex indices (OpenGL: `gl_VertexID` or Vulkan: `gl_VertexIndex`), we need to
update some glslang options when compiling GLSL to SPIR-V under OpenGL semantics.

This change is a bit risky, and I'm still in the process of checking that nothing breaks.

As far as I can tell, there's only one scenario where we use GLSL Vulkan semantics but still compile
to OpenGL-flavored GLSL, and that's for framebuffer fetch. I treat this as a special case, which
we're already doing somewhat.


romainguy (approved):
Thank you Ben!

prideout (approved):
Nice.

pixelflinger (changes requested):
dumb question: couldn't we fix this by "simply" always using the vulkan syntax and always including a #define on opengl.
we'd document that our filament glsl must use the vulkan syntax.

romainguy:
No because the Vulkan syntax is not compatible with GL.

pixelflinger (approved):

PR #3494, opened 2021-02-03
Fix MSAA for OpenGL ES 3.0 core. (third attempt)

### LINKED ISSUE #3128: WebGL triangle demo has GL_INVALID_FRAMEBUFFER_OPERATION

prideout:
This fixes `GL_FRAMEBUFFER_INCOMPLETE_MULTISAMPLE` caused by the following spec clause being violated:

```
The value of RENDERBUFFER_SAMPLES [must be] the same for all
attached renderbuffers and, if the attached images are a mix of
renderbuffers and textures, the value of RENDERBUFFER_SAMPLES
is zero.
```

Tested on Chrome with Triangle and Helmet demos (after enabling MSAA).
Also tried running the automation suite in a debug build on macOS, to see if any asserts or GL errors would be triggered.

Fixes #3128.

### DIFF
--- a/filament/backend/src/opengl/OpenGLDriver.cpp
+++ b/filament/backend/src/opengl/OpenGLDriver.cpp
@@ -773,35 +773,38 @@ void OpenGLDriver::framebufferTexture(backend::TargetBufferInfo const& binfo,
     assert(rt->width  <= valueForLevel(binfo.level, t->width) &&
            rt->height <= valueForLevel(binfo.level, t->height));
 
+    // Declare a small mask of bits that will later be OR'd into the texture's resolve mask.
+    TargetBufferFlags resolveFlags = {};
+
     GLRenderTarget::GL::RenderBuffer const* pRenderBuffer = nullptr;
     switch (attachment) {
         case GL_COLOR_ATTACHMENT0:
-            rt->gl.resolve |= TargetBufferFlags::COLOR0;
+            resolveFlags = TargetBufferFlags::COLOR0;
             pRenderBuffer = &rt->gl.color[0];
             break;
         case GL_COLOR_ATTACHMENT1:
-            rt->gl.resolve |= TargetBufferFlags::COLOR1;
+            resolveFlags = TargetBufferFlags::COLOR1;
             pRenderBuffer = &rt->gl.color[1];
             break;
         case GL_COLOR_ATTACHMENT2:
-            rt->gl.resolve |= TargetBufferFlags::COLOR2;
+            resolveFlags = TargetBufferFlags::COLOR2;
             pRenderBuffer = &rt->gl.color[2];
             break;
         case GL_COLOR_ATTACHMENT3:
-            rt->gl.resolve |= TargetBufferFlags::COLOR3;
+            resolveFlags = TargetBufferFlags::COLOR3;
             pRenderBuffer = &rt->gl.color[3];
             break;
         case GL_DEPTH_ATTACHMENT:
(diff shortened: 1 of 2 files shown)


filament/backend/src/opengl/OpenGLDriver.cpp
```diff
@@ -899,21 +902,43 @@ void OpenGLDriver::framebufferTexture(backend::TargetBufferInfo const& binfo,
         }
     } else
 #endif
-    { // here we emulate ext.EXT_multisampled_render_to_texture
+    if (!any(t->usage & TextureUsage::SAMPLEABLE)) {
```
pixelflinger, line 905: I think this test is not enough. We also need a sidecar if the target render buffer is not multi-sampled.
That is, if the user is asking for a "magic resolve" into a render buffer.

filament/backend/src/opengl/OpenGLDriver.cpp
```diff
@@ -899,21 +902,43 @@ void OpenGLDriver::framebufferTexture(backend::TargetBufferInfo const& binfo,
...
         gl.bindFramebuffer(GL_FRAMEBUFFER, rt->gl.fbo);
+        glFramebufferRenderbuffer(GL_FRAMEBUFFER, attachment, GL_RENDERBUFFER, t->gl.id);
+
+        // Clear the resolve bit for this particular attachment. Note that other attachment(s)
+        // might be sampleable, so this does not necessarily prevent the resolve from occuring.
+        resolveFlags = TargetBufferFlags::NONE;
 
+    } else {
+        assert(rt->gl.samples > 1);
+        assert(glIsTexture(t->gl.id));
```
pixelflinger, line 922: per the comment above, this check should be removed.

filament/backend/src/opengl/OpenGLDriver.cpp
```diff
@@ -926,14 +951,8 @@ void OpenGLDriver::framebufferTexture(backend::TargetBufferInfo const& binfo,
...
             case GL_TEXTURE_2D:
-                if (any(t->usage & TextureUsage::SAMPLEABLE)) {
-                    glFramebufferTexture2D(GL_FRAMEBUFFER, attachment,
-                            target, t->gl.id, binfo.level);
-                } else {
-                    assert(target == GL_TEXTURE_2D);
-                    glFramebufferRenderbuffer(GL_FRAMEBUFFER, attachment,
-                            GL_RENDERBUFFER, t->gl.id);
-                }
+                glFramebufferTexture2D(GL_FRAMEBUFFER, attachment,
```
pixelflinger, line 954: Per the comment above, we can still end-up here with a render buffer

pixelflinger (changes requested):

filament/backend/src/opengl/OpenGLDriver.cpp
prideout: If the target render buffer is not multi-sampled, then this code is unreachable; see line 850.

pixelflinger: No no that's not what I mean. I meant if the render buffer is not multi-sampled, but the render target is. Again, this is the case of a "magic resolve" into a render buffer.

prideout: This is complicated!  Okay I added another commit to make this change, and thankfully WebGL is still working.

pixelflinger (approved):

PR #3508, opened 2021-02-05
Do not exclude Vulkan from continuous Android CI.

prideout:


build/android/build.sh
```diff
@@ -50,6 +50,11 @@ elif [[ "$LC_UNAME" == "darwin" ]]; then
 fi
 source `dirname $0`/../common/build-common.sh
 
+# For continuous builds, do not exclude Vulkan.
+if [[ ! -z "${BUILD_DEBUG}" ]] && [[ ! -z "${BUILD_RELEASE}" ]] ; then
```
romainguy, line 54: We probably still want to exclude from presubmits otherwise it takes forever. But we can include in continuous (post submit) builds

romainguy (changes requested):

build/android/build.sh
prideout: According to `build-common.sh`, our presubmits are release only, so this conditional will return false (thus excluding Vulkan).  The continuous builds are the only ones that build both.

romainguy: Makes sense, but the condition would be easier to read if you checked the build target instead:

```
if [[ "$TARGET" == "continuous" ]]; then
```

romainguy (approved):

build/android/build.sh
prideout: Doh!  Fixed.

PR #3524, opened 2021-02-08
matc: Use consistent params for semantic code analysis

bejado:
When running semantic analysis on a material, we were arbitrarily choosing the first code gen permutation to analyze. So, running `matc` with arguments `--api metal` versus `--api all` would run analysis on slightly different shader code. This causes bugs when flags passed to glslang differ during semantic analysis. This change updates all semantic analysis to always use the same shader code.

### DIFF
--- a/libs/filamat/include/filamat/MaterialBuilder.h
+++ b/libs/filamat/include/filamat/MaterialBuilder.h
@@ -111,7 +111,7 @@ protected:
         TargetLanguage targetLanguage;
     };
     std::vector<CodeGenParams> mCodeGenPermutations;
-    // For finding properties and running semantic analaysis, we always use the same code gen
+    // For finding properties and running semantic analysis, we always use the same code gen
     // permutation. This is the first permutation generated with default arguments passed to matc.
     CodeGenParams mSemanticCodeGenParams = {
         .shaderModel = (int) ShaderModel::GL_ES_30,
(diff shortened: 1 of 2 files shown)


libs/filamat/include/filamat/MaterialBuilder.h
```diff
@@ -111,6 +111,13 @@ class UTILS_PUBLIC MaterialBuilderBase {
         TargetLanguage targetLanguage;
     };
     std::vector<CodeGenParams> mCodeGenPermutations;
+    // For finding properties and running semantic analaysis, we always use the same code gen
```
prideout, line 114:
```suggestion
    // For finding properties and running semantic analysis, we always use the same code gen
```

prideout (approved):

libs/filamat/include/filamat/MaterialBuilder.h
```diff
@@ -111,6 +111,13 @@ class UTILS_PUBLIC MaterialBuilderBase {
         TargetLanguage targetLanguage;
     };
     std::vector<CodeGenParams> mCodeGenPermutations;
+    // For finding properties and running semantic analysis, we always use the same code gen
+    // permutation. This is the first permutation generated with default arguments passed to matc.
+    CodeGenParams mSemanticCodeGenParams = {
```
romainguy, line 116: Can you make it `const` as well?

PR #3527, opened 2021-02-09
Improve docs generation

romainguy:
`build.sh -w` now performs a full install of dependencies instead of relying on local installation tricks.


prideout (approved):
Looks fine to me, at least `node_modules` will be in out/web-docs rather than the cwd.

romainguy:
It was always in out/web-doxs already. But dependencies weren't getting pulled because wat.

PR #3533, opened 2021-02-12
Add backend unit test for blit.

prideout:
This is fairly simple for now, only does color.
Will enhance in a future PR.

### DIFF
--- a/filament/backend/test/test_Blit.cpp
+++ b/filament/backend/test/test_Blit.cpp
@@ -28,14 +28,6 @@
 #include <image/ColorTransform.h>
 #endif
 
-static uint32_t sPixelHashResult = 0;
-
-static constexpr int kDstTexWidth = 384;
-static constexpr int kDstTexHeight = 384;
-static constexpr auto kDstTexFormat = filament::backend::TextureFormat::RGBA8;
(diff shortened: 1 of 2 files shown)


filament/backend/test/test_Blit.cpp
```diff
@@ -0,0 +1,260 @@
...
+    getDriverApi().endFrame(0);
+
+    // Wait for the ReadPixels result to come back.
+    api.finish();
+    executeCommands();
+    getDriver().purge();
+
+    // Check if the image matches perfectly to our golden run.
+    const uint32_t expected = 0xe2353ca6;
+    printf("Computed hash is 0x%8.8x, Expected 0x%8.8x\n", sPixelHashResult, expected);
```
romainguy, line 249: This seems quite brittle to rely on a global static field, esp. if we add more tests. Couldn't you instead pass a reference of some kind to dumpScreenshot to it writes the result in a safe place?

prideout: Yes this was sloppy.  I added a commit to the PR to fix this.

romainguy (approved):
👍 Thanks!

PR #3544, opened 2021-02-17
Remove Windows line endings.

prideout:
These 19 files have line endings that cause annoying diffs when comparing Filament's GitHub source with its twin sister within Google.


romainguy (approved):

bejado (approved):
Thanks Philip!

PR #3545, opened 2021-02-17
Avoid uninitialized reads in computeVisibilityMasks.

prideout:
I verified that this code is not enabled in our GitHub builds, and I
verified that the MSAN error goes away.

### DIFF
--- a/filament/src/Scene.cpp
+++ b/filament/src/Scene.cpp
@@ -167,6 +167,16 @@ void FScene::prepare(const mat4f& worldOriginTransform) {
     for (size_t i = lightData.size(), e = (lightData.size() + 3u) & ~3u; i < e; i++) {
         new(lightData.data<POSITION_RADIUS>() + i) float4{ 0, 0, 0, 1 };
     }
+
+    // Purely for the benefit of MSAN, we can avoid uninitialized reads by zeroing out the
+    // unused scene elements between the end of the array and the rounded-up count.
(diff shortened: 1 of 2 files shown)


romainguy (approved):

pixelflinger (approved):

filament/src/Scene.cpp
```diff
@@ -167,6 +167,16 @@ void FScene::prepare(const mat4f& worldOriginTransform) {
...
         new(lightData.data<POSITION_RADIUS>() + i) float4{ 0, 0, 0, 1 };
     }
+
+    // Purely for the benefit of MSAN, we can avoid uninitialized reads by zeroing out the
+    // unused scene elements between the end of the array and the rounded-up count.
+    if (UTILS_HAS_SANITIZE_MEMORY) {
+        for (size_t i = sceneData.size(), e = (sceneData.size() + 0xFu) & ~0xFu; i < e; i++) {
+            sceneData.data<LAYERS>()[i] = 0;
+            sceneData.data<VISIBLE_MASK>()[i] = 0;
+            sceneData.data<VISIBILITY_STATE>()[i] = {};
```
pixelflinger, line 177: Side note: you could also write it like if you prefer.
```
sceneData.elementAt<LAYERS>(i) = 0;
...
```

The most efficient way to write it would be to call `.data` outside the loop. I don't think it matters here though.

bejado (approved):

PR #3554, opened 2021-02-19
Cherry pick some fixes into rc/1.9.13

prideout:
This PR merges a small selection of innocuous commits from main to our current release candidate.  It would be nice to get these into Google3 on Monday.


bejado (approved):
In general I think we want to avoid cherry-picking non-bug fixes into an rc branch, but since these are relatively safe commits I think it's fine.

My recommendation is to "Rebase and merge" these (as opposed to traditional squashing), so it's easy to see which commits were cherry-picked.

pixelflinger (approved):

prideout:
> In general I think we want to avoid cherry-picking non-bug fixes into an rc branch, but since these are relatively safe commits I think it's fine.
> 
> My recommendation is to "Rebase and merge" these (as opposed to traditional squashing), so it's easy to see which commits were cherry-picked.

These are both good points, agreed 100%.  Let's try to avoid these types of cherry-picks in the future.

My excuse for breaking the rule is that I'm currently in the mode of trying to improve parity between the codebases.

I will be sure not to squash.

PR #3560, opened 2021-02-21
Attempt at improving bloom/emissive with glTF files

romainguy:
Without this change, a valid glTF file only supplies values between 0 and 1 for `emissiveFactor` which means emissive surfaces never bloom. I used a somewhat arbitrary EV factor (see comment for reasons) but we could also apply the inverse tonemap operator to `emissiveFactor`.


prideout (approved):
Thanks, good change. This likely affects the look of the default loaded model in gltf_viewer since it has emissive portions (DamagedHelmet).

PR #3565, opened 2021-02-22
assert_invariant() now terminates the program in failure

pixelflinger:
rename logAndPanic to panicLog, since it doesn't actually panics.

### DIFF
--- a/libs/utils/include/utils/Panic.h
+++ b/libs/utils/include/utils/Panic.h
@@ -374,7 +374,7 @@ private:
 
 namespace details {
 // these are private, don't use
-void logAndPanic(
+void panicLog(
         char const* function, char const* file, int line, const char* format, ...) noexcept;
 }  // namespace details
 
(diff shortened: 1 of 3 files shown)


romainguy (approved):

prideout (commented):
Why is it called panicLog if doesn't actually panic? Typically the word "panic" is reserved for unrecoverable errors.

pixelflinger:
> Why is it called panicLog if doesn't actually panic? Typically the word "panic" is reserved for unrecoverable errors.

here "panic" is more like a prefix for the panic stuff. it mimics a macro called PANIC_LOG (as opposed to PANIC_ASSERT).

kpeeters:
The `assert_invariant` macro clashes with e.g. nlohmann::json (https://github.com/nlohmann/json), which has that as a member function on some of its objects. Might be better to give it an explicit `filament` prefix (`filament_assert_invariant`) and perhaps also capitalise it to make clear that it is a macro.

PR #3570, opened 2021-02-23
DO NOT MERGE: Reproduce Metal backend assertion

bejado:


pixelflinger:
I think we can close this.

PR #3576, opened 2021-02-25
FrameGraph rewrite from the ground up

pixelflinger:
FrameGraph rewrite from the ground up

The main goal of this rewrite was to make the code
simpler and easier to maintain.

The API is mostly unchanged, however there are some differences.
- we now have the concept of subresources, e.g. for Textures, as
  subresource is a mip level or layer.
- RenderTargets are no longer resources, instead they are transiant
  objects associated to a pass and are now called RenderPasses.
- RenderPasses take subresources for attachments.
- We have better validation of graph building.

We should also compute discard flags more accurately.


bejado:
FYI Windows is failing with

```
D:\a\filament\filament\filament\src\fg2/FrameGraphId.h(67,14): error C2039: 'swap': is not a member of 'std' [D:\a\filament\filament\out\cmake-mt\filament\filament.vcxproj]
C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\VC\Tools\MSVC\14.28.29333\include\limits(25): message : see declaration of 'std' [D:\a\filament\filament\out\cmake-mt\filament\filament.vcxproj]
```

filament/src/PostProcessManager.cpp
```diff
@@ -281,94 +282,80 @@ FrameGraphId<FrameGraphTexture> PostProcessManager::structure(FrameGraph& fg,
...
     // We limit the lowest lod size to 32 pixels (which is where the -5 comes from)
-    const size_t levelCount = FTexture::maxLevelCount(width, height) - 5;
+    const size_t levelCount = std::min(8, FTexture::maxLevelCount(width, height) - 5);
     assert_invariant(levelCount >= 1);
 
     // generate depth pass at the requested resolution
     auto& structurePass = fg.addPass<StructurePassData>("Structure Pass",
             [&](FrameGraph::Builder& builder, auto& data) {
-                data.depth = builder.createTexture("Structure Buffer", {
+                data.depth = builder.create<FrameGraphTexture>("Structure Buffer", {
```
bejado, line 301: Why not use `builder.createTexture` here?

filament/src/fg2/details/DependencyGraph.h
```diff
@@ -0,0 +1,199 @@
...
+     * @return A list of outgoing edges
+     */
+    EdgeContainer getOutgoingEdges(Node const* node) const noexcept;
+
+    Node const* getNode(NodeID id) const noexcept;
+
+    Node* getNode(NodeID id) noexcept;
+
+    //! cull unreferenced nodes. Links ARE NOT removed, only reference counts are updated.
+    void cull() noexcept;
```
bejado, line 165: I'm having some trouble understanding the algorithm behind cull. When is a node culled? When there's no path from it to a "target" node?

filament/src/PostProcessManager.cpp
pixelflinger: That's an oversight. Thanks!

filament/src/fg2/details/DependencyGraph.h
pixelflinger: Yes excatly.

PR #3583, opened 2021-02-26
Label Metal render encoders with group markers

bejado:
This makes debugging Metal captures much easier to follow.


prideout (approved):
Looks good.  Note that <stack> includes <deque> which includes <algorithm>.  After Mathias makes some containers we should do a big cleanup.

bejado:
> Note that `<stack>` includes `<deque>` which includes `<algorithm>`. After Mathias makes some containers we should do a big cleanup.

Agreed.

PR #3587, opened 2021-02-26
Remove hat-trie from third_party.

prideout:
This is unused, my fault for adding it in the first place.


romainguy (approved):

prideout:
Oh we're using this for the name-to-entity stuff in gltfio.

PR #3589, opened 2021-02-26
A replacement for std::vector<> of trivial types

pixelflinger:
The aim here is to reduce code bloat.

### DIFF
--- a/libs/utils/include/utils/vector.h
+++ b/libs/utils/include/utils/vector.h
@@ -19,6 +19,7 @@
 
 #include <utils/Allocator.h>
 #include <utils/compiler.h>
+#include <utils/compressed_pair.h>
 
 #include <algorithm>
 #include <utility>
@@ -70,21 +71,36 @@ protected:
     using size_type = uint32_t;
     using allocator_type = ALLOCATOR;
 
+    // our compressed_pair works only with classes
+    struct Capacity {
+        Capacity() = default;
+        Capacity(size_type v) : value(v) {}
+        Capacity(Capacity const& rhs) noexcept = default;
+        Capacity(Capacity&& rhs) noexcept {
+            std::swap(value, rhs.value);
+        }
+        size_type value = 0;
+    };
+
     char* mBegin = nullptr;
     size_type mItemCount = 0;
-    size_type mCapacity = 0;
-    ALLOCATOR mAllocator{};
+    utils::compressed_pair<Capacity, allocator_type> mCompressedCapacity;
 
-    TrivialVectorBase() noexcept = default;
+    size_type& capacity() { return mCompressedCapacity.first().value; }
+    allocator_type& allocator() { return mCompressedCapacity.second(); }
 
-    explicit TrivialVectorBase(const allocator_type& allocator) noexcept : mAllocator(allocator) { }
+    TrivialVectorBase() = default;
 
-    TrivialVectorBase(size_type itemSize, size_type count, const allocator_type& allocator) noexcept
-            : mItemCount(count), mCapacity(count), mAllocator(allocator) {
-        mBegin = (char*)mAllocator.alloc(mCapacity * itemSize);
+    explicit TrivialVectorBase(const allocator_type& allocator)
+            : mCompressedCapacity(0, allocator) {
     }
 
-    TrivialVectorBase(size_type itemSize, size_type count) noexcept
+    TrivialVectorBase(size_type itemSize, size_type count, const allocator_type& a)
+            : mItemCount(count), mCompressedCapacity(count, a) {
+        mBegin = (char*)allocator().alloc(capacity() * itemSize);
+    }
+
+    TrivialVectorBase(size_type itemSize, size_type count)
             : TrivialVectorBase(itemSize, count, allocator_type()) {
     }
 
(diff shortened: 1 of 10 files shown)


libs/utils/include/utils/vector.h
```diff
@@ -55,6 +55,214 @@ static inline bool insert_sorted_unique(std::vector<T>& v, T item) {
...
+        set_capacity(c, itemSize);
+    }
+
+    UTILS_NOINLINE
+    void set_capacity(size_type n, size_type itemSize) {
+        // FIXME: move this in cpp
+        if (n == mCapacity) {
+            return;
+        }
+        char* addr = (char*)malloc(n * itemSize);
```
romainguy, line 101: You should include `<string.h>` and `<stdlib.h>` for malloc/free/memcpy

libs/utils/include/utils/vector.h
```diff
@@ -55,6 +55,214 @@ static inline bool insert_sorted_unique(std::vector<T>& v, T item) {
...
+    void push_back(const_reference v) {
+        assert_capacity(size() + 1, sizeof(T));
+        *end() = v;
+        mItemCount++;
+    }
+
+    template<typename ... ARGS>
+    reference emplace_back(ARGS&& ... args) {
+        assert_capacity(size() + 1, sizeof(T));
+        T::T(end())(std::forward<ARGS>(args)...);
```
romainguy, line 209: include `<utility>`

libs/utils/include/utils/vector.h
```diff
@@ -55,6 +55,214 @@ static inline bool insert_sorted_unique(std::vector<T>& v, T item) {
...
+        }
+        char* addr = (char*)malloc(n * itemSize);
+        memcpy(addr, mBegin, ((mItemCount < n) ? mItemCount : n) * itemSize);
+        free(mBegin);
+        mBegin = addr;
+        mCapacity = n;
+    }
+
+    void swap(TrivialVectorBase& other) {
+        using std::swap;
```
romainguy, line 109: include `<algorithm>`

libs/utils/include/utils/vector.h
```diff
@@ -55,6 +55,214 @@ static inline bool insert_sorted_unique(std::vector<T>& v, T item) {
...
+    }
+
+    char* begin() noexcept { return mBegin; }
+    char* end(size_type itemSize) noexcept { return mBegin + mItemCount * itemSize; }
+    char const* begin() const noexcept { return const_cast<TrivialVectorBase*>(this)->begin(); }
+    char const* end(size_type itemSize) const noexcept { return const_cast<TrivialVectorBase*>(this)->end(itemSize); }
+};
+
+template<typename T, std::enable_if_t<std::is_trivial<T>::value, bool> = true>
+class vector : private TrivialVectorBase {
```
romainguy, line 122: Should we give it a different name to avoid using the wrong one/collisions when/if we do using utils and using std in the same file? I think we rarely if ever do using std but…

libs/utils/include/utils/vector.h
```diff
@@ -14,11 +14,11 @@
  * limitations under the License.
  */
 
-#include <vector>
-
 #ifndef UTILS_VECTOR_H
 #define UTILS_VECTOR_H
 
+#include <vector>
```
bejado, line 20: Why is this still necessary? Doesn't this default the whole purpose?

bejado: Oh, I see `std::vector` is still used elsewhere in the file.. I'm guessing a later PR will handle switching over to the new `vector`.

romainguy: This file contains utility functions for std::vector too.

pixelflinger: We can't just blindly switch either because utils::vector works only for trivial types, also I won't be able to make it work with STL allocators, because of how they work. They'll work with our filament allocator though.

libs/utils/include/utils/vector.h
bejado, line 209: I'm probably missing something obvious, but I can't parse this line. What's `T::T` suppose to be? When I try to create a `struct Foo {}; utils::vector<Foo>`, I get an error.

romainguy: It's a `T::T(…)`, a constructor.

bejado: This is what I mean:

```
struct Foo { };
utils::vector<Foo> v;
v.emplace_back();
```

```
../../libs/utils/include/utils/vector.h:209:12: error: no member named 'T' in 'Foo'
        T::T(end())(std::forward<ARGS>(args)...);
```

pixelflinger: Yeah I completely goofed up there.
The code should have been `new T(end())(...)`.  Because it's a template it didn't fail at compile time until I actually tried to use it :-)

libs/utils/include/utils/vector.h
pixelflinger: The final code is not using these.

libs/utils/include/utils/vector.h
bejado: I was thinking this was some bizarre new C++ syntax I hadn't seen before, which wouldn't have surprised me :)

bejado (approved):

PR #3590, opened 2021-02-27
Add automation to Android build, add JNI bindings, change PNG to PPM.

prideout:
This adds `libs/viewer` to `filament-utils-android` and provides JNI bindings for AutomationEngine.

This commit also changes the AutomationEngine to use PPM for screenshots
rather than PNG. The reasoning is that it is wasteful to include stb and
our two image-related libraries just to dump screenshots, especially for
a feature we might not use on Android.


libs/viewer/src/AutomationEngine.cpp
```diff
@@ -65,12 +60,10 @@ void exportScreenshot(View* view, Renderer* renderer, std::string filename,
...
             const Viewport& vp = state->view->getViewport();
-            LinearImage image(toLinear<uint8_t>(vp.width, vp.height, vp.width * 3,
-                    static_cast<uint8_t*>(buffer)));
             Path out(state->filename);
-            std::ofstream outputStream(out, std::ios::binary | std::ios::trunc);
-            ImageEncoder::encode(outputStream, ImageEncoder::Format::PNG, image, "",
-                    state->filename);
+            std::ofstream ppmStream(out);
+            ppmStream << "P6 " << vp.width << " " << vp.height << " " << 255 << std::endl;
+            ppmStream.write(static_cast<char*>(buffer), vp.width * vp.height * 3);
```
prideout, line 66: I ❤️ ppm.

PR #3599, opened 2021-03-01
sample-gltf-viewer: introduce QR code feature.

prideout:
This works in tandem with the new page at https://google.github.io/filament/qrcode/

Screen Shot 2021-03-01 at 9 46 36 AM

This feature did not require any changes to `ModelViewer.kt` or any of our libraries. It involves the following additions to the app MainActivity:

- Add a button to the app bar, which requires AppCompatActivity as per
  https://developer.android.com/training/appbar/setting-up

- Add a PreviewView widget from androidx.

- Add usage of BarcodeScanner in mlkit.

- Add a coroutine for downloading data via HttpURLConnection.

We would like to add more features in the future, including the ability
to drop additional models without rescanning and the ability to drop
zip files.

This feature was inspired by a Model Viewer editor feature, thanks @elalish

### DIFF
--- a/android/samples/sample-gltf-viewer/src/main/java/com/google/android/filament/gltf/MainActivity.kt
+++ b/android/samples/sample-gltf-viewer/src/main/java/com/google/android/filament/gltf/MainActivity.kt
@@ -45,13 +45,14 @@ import java.net.URL
 import java.nio.Buffer
 import java.nio.ByteBuffer
 import java.util.concurrent.Executors
+import java.util.zip.ZipInputStream
 
 class MainActivity : AppCompatActivity() {
 
     companion object {
         // Load the library for the utility layer, which in turn loads gltfio and the Filament core.
         init { Utils.init() }
-        private const val TAG = "gltf"
+        private const val TAG = "gltf-viewer"
     }
 
     private lateinit var surfaceView: SurfaceView
@@ -127,8 +128,8 @@ class MainActivity : AppCompatActivity() {
                 .setBackpressureStrategy(ImageAnalysis.STRATEGY_KEEP_ONLY_LATEST)
                 .build()
 
-        imageAnalysis.setAnalyzer(Executors.newSingleThreadExecutor(), {
-            imageProxy -> scanQRCode(imageProxy)
+        imageAnalysis.setAnalyzer(Executors.newSingleThreadExecutor(), { imageProxy ->
+            scanQRCode(imageProxy)
         })
 
         try {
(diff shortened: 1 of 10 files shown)


elalish:
Let me get this straight: you also created a project called `ModelViewer`? 🤦‍♂️

prideout:
> Let me get this straight: you also created a project called `ModelViewer`? 🤦‍♂️

Well it's just the name of a Kotlin class, it seemed like the most appropriate name at the time.  :)

android/samples/sample-gltf-viewer/src/main/java/com/google/android/filament/gltf/MainActivity.kt
```diff
@@ -74,6 +111,126 @@ class MainActivity : Activity() {
         modelViewer.view.bloomOptions = bloomOptions
     }
 
+    private fun bindPreview(cameraProvider: ProcessCameraProvider) {
+        val preview : Preview = Preview.Builder().build().also {
```
bejado, line 115: I think you can leave off the `: Preview` here and `: CameraSelector` down below

android/samples/sample-gltf-viewer/src/main/java/com/google/android/filament/gltf/MainActivity.kt
```diff
@@ -74,6 +111,126 @@ class MainActivity : Activity() {
...
+        if (previewView.visibility != View.VISIBLE) {
+            imageProxy.close()
+            return
+        }
+        val options = BarcodeScannerOptions.Builder().setBarcodeFormats(Barcode.FORMAT_QR_CODE).build()
+        val image = InputImage.fromMediaImage(mediaImage, imageProxy.imageInfo.rotationDegrees)
+        val scanner = BarcodeScanning.getClient(options)
+        scanner.process(image)
+                .addOnSuccessListener { barcodes ->
+                    for (barcode in barcodes) {
```
bejado, line 173: If multiple barcodes are detected, does it make sense to call `downloadFromBarcode` for each? Perhaps we just want to download the first one we find.

bejado (approved):

android/samples/sample-gltf-viewer/src/main/java/com/google/android/filament/gltf/MainActivity.kt
prideout: Good catch, I will simplify this.

PR #3603, opened 2021-03-02
gltf_viewer: avoid writeJson when automation is off.

prideout:


romainguy (approved):

libs/viewer/src/Settings.cpp
```diff
@@ -1061,6 +1061,30 @@ std::string writeJson(const ViewSettings& in) {
     return oss.str();
 }
 
+bool ColorGradingSettings::operator==(const ColorGradingSettings &rhs) const {
```
bejado, line 1064: It might be worth putting a `static_assert(sizeof(ColorGradingSettings) == NNN)` here to remind us to update this operator when we add a new field. This has certainly bit me before.

prideout: Good idea, done.

PR #3606, opened 2021-03-03
sample-gltf-viewer: add drag-and-drop feature.

prideout:
This adds a WebSockets server to filament-utils-android.

Unlike my first attempt (#3599), this does not use a piping server and
is therefore much faster to use over a LAN. It also has lower user
friction because there is no need to touch a button in the app or
scan a QR code.

https://user-images.githubusercontent.com/1288904/109844924-74200880-7c01-11eb-8572-fe369d60d84b.mov


android/samples/sample-gltf-viewer/src/main/java/com/google/android/filament/gltf/MainActivity.kt
```diff
@@ -105,6 +122,73 @@ class MainActivity : Activity() {
...
+        message.buffer.get(zipfileBytes)
+
+        var gltfPath: String? = null
+        val pathToBufferMapping = withContext(Dispatchers.IO) {
+            val deflater = ZipInputStream(ByteArrayInputStream(zipfileBytes))
+            val mapping = HashMap<String, Buffer>()
+            while (true) {
+                val entry = deflater.nextEntry ?: break
+                if (entry.isDirectory) continue
+                if (entry.name.startsWith("__MACOSX")) continue
```
bejado, line 150: Is `__MACOSX` not recognized as a directory? I wonder if `.DS_Store` files would cause a similar issue.

third_party/civetweb/tnt/CMakeLists.txt
```diff
@@ -32,3 +32,5 @@ target_include_directories(${TARGET} PUBLIC ${PUBLIC_HDR_DIR})
 target_link_libraries(${TARGET} ${CMAKE_DL_LIBS})
 
 target_compile_options(${TARGET} PRIVATE $<$<PLATFORM_ID:Linux>:-fPIC>)
+
+install(TARGETS ${TARGET} ARCHIVE DESTINATION third_party/${DIST_DIR})
```
bejado, line 36: Is there a reason we need a new `third_party` directory? Seems like we already put third party binaries inside `lib`, like `libsmol-v.a` and `libdracodec.a`.

prideout: Excellent catch, fixing this thanks!

android/samples/sample-gltf-viewer/src/main/java/com/google/android/filament/gltf/MainActivity.kt
prideout: This is really just an optimization, maybe it's questionable, I noticed these junky things appearing when I create ZIP files with macOS, and there's no reason to percolate them all the way down into gltfio.  Good point, I also see `.DS_Store` files in there so I think I should ignore those too.

dandingol03:
> This adds a WebSockets server to filament-utils-android.
> 
> Unlike my first attempt (#3599), this does not use a piping server and is therefore much faster to use over a LAN. It also has lower user friction because there is no need to touch a button in the app or scan a QR code.
> 
>  lower_friction.mov

what a amazing work!!!

Issue #3562, opened 2021-02-22
Windows Vulkan problem


FakerShy:
**Describe the bug**
When I finished building the filament on Windows，The configured backend is opengl.When I change the configuration to Vulkan ，The library doesn't work.

The program gave me some logs.But i dont know how to fix it.So i need some help.
VULKAN ERROR: (VUID-vkMapMemory-size-00680) Validation Error: [ VUID-vkMapMemory-size-00680 ] Object 0: handle = 0x8e7343000000014d, type = VK_OBJECT_TYPE_DEVICE_MEMORY; | MessageID = 0xff4787ab | VkMapMemory: Attempting to map memory range of size zero The Vulkan spec states: If size is not equal to VK_WHOLE_SIZE, size must be greater than 0 (https://vulkan.lunarg.com/doc/view/1.2.141.0/windows/1.2-extensions/vkspec.html#VUID-vkMapMemory-size-00680)

**To Reproduce**
config.backend = filament::Engine::Backend::VULKAN;

 - OS: [e.g. windows]
 - GPU: [e.g. NVIDIA GTX 1660ti]
 - Backend: [Vulkan]

prideout:
Thanks, we currently trigger some known validation messages, but this one is new to me.  Which sample app are you running?  Does it work in release configuration?

FakerShy:
Sample is suzanne. My program runs in win10, vs version is 2019, and the downloaded version is 1.9.12.I found the compilation option of cmakelist in windows，I changed option(FILAMENT_SUPPORTS_VULKAN "Include the Vulkan backend" ON).But it 
didn't work,I only changed the compilation options and configuration in the project. OpenGL works, Vulkan doesn't.
I've tried many examples, all of which are about this problem. I doubt the version of Vulkan. 1.1.0 1.2 doesn't work. My own program has no problem using Vulkan, so I need your help.

wumo:
Also failed due to validation error. OS: Windows 10, Nvidia RTX 3080.
```
filament\cmake-build-debug\samples\gltf_viewer.exe --api vulkan
FEngine (64 bits) created at 000002714928B080 (threading is enabled)
FEngine resolved backend: Vulkan
Vulkan device driver: NVIDIA 461.40
Selected physical device 'GeForce RTX 3080' from 1 physical devices. (vendor 0x10de, device 0x2206, driver 0x734a0000, a
pi 1.2)
[...]

PR #3609, opened 2021-03-04
PostProcessManager: fix setMinMaxLevels for StructureMipmap.

prideout:


romainguy (approved):

bejado (approved):

filament/src/PostProcessManager.cpp
```diff
@@ -350,7 +350,7 @@ FrameGraphId<FrameGraphTexture> PostProcessManager::structure(FrameGraph& fg,
                     mi->setParameter("level", uint32_t(level));
                     commitAndRender(out, material, driver);
                 }
-                driver.setMinMaxLevels(in, 0, levelCount);
+                driver.setMinMaxLevels(in, 0, levelCount - 1);
```
pixelflinger, line 353: out of curiosity, does it matter? I thought the default value was like 1000 on gl?

pixelflinger (approved):

filament/src/PostProcessManager.cpp
prideout: Yeah, to pass validation Vulkan cannot create an ImageView with a SubresourceRange that exceeds the bounds of the image.  I should perhaps make the backend auto-clamp.  Either way this codeline was wrong.

pixelflinger: autoclamp seems like a good idea, for when we'll make these api public. We want to avoid errors in the backend.

prideout:
This fixes an issue mentioned in #3562.

filament/src/PostProcessManager.cpp
prideout: Done.

PR #3610, opened 2021-03-04
Make material compilation deterministic

bejado:
With multithreaded variant compilation, compiled materials differ based on which variants finish first. This change ensures that compiled materials are always bit-for-bit identical.

This is helpful when comparing two compiled materials. Hashing the material files can be used to check for equality. It's also convenient for variants to occupy the same index when using matinfo's `--print-glsl` flag, for example.

### DIFF
--- a/libs/filamat/src/eiff/ShaderEntry.h
+++ b/libs/filamat/src/eiff/ShaderEntry.h
@@ -18,6 +18,7 @@
 #define TNT_FILAMAT_SHADER_ENTRY_H
 
 #include <string>
+#include <vector>
 
 namespace filamat {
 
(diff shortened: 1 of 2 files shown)


romainguy (commented):
Awesome catch :) I will bow my head in shame for not thinking about this in the first place!

libs/filamat/src/eiff/ShaderEntry.h
```diff
@@ -34,6 +35,9 @@ struct SpirvEntry {
     uint8_t variant;
     uint8_t stage;
     size_t dictionaryIndex;
+
+    // temporarily holds this entry's spirv until added to the dictionary
+    std::vector<uint32_t> spirv;
```
romainguy, line 40: #ifndef FILAMAT_LITE ?

romainguy (approved):

PR #3611, opened 2021-03-04
More code size optimizations

pixelflinger:
MaterialInstance: generate versions of setParameter() methods that
are templated on the parameter sizeof(), which allows us to
coalesce implementations, e.g. float4 and int4 are now
the same. We make the compiler generate non-inline version of all
these methods, the typed version just call through, taking
advantage of the tail-call optimization in clang.

MaterialParser: improve code to avoid calling multiple methods that
all did more or less the same thing and were inlined.

PostProcessManager: use a loop to initialize the materials

ToneMapping: de-dup ACES implementation

And other more minor optimizations.

### DIFF
--- a/filament/CMakeLists.txt
+++ b/filament/CMakeLists.txt
@@ -76,6 +76,7 @@ set(SRCS
         src/Stream.cpp
         src/SwapChain.cpp
         src/Texture.cpp
+        src/ToneMapping.cpp
         src/UniformBuffer.cpp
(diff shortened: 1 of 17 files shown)


prideout (approved):

romainguy (approved):
Did you measure the impact of not inlining the tone mappers btw? Since we run them when we initialize Filament I'd like to make sure it doesn't make a meaningful difference.

PR #3614, opened 2021-03-05
new option to disable PRECONDITION checks at compile time

pixelflinger:
the new -DFILAMENT_PRECONDITION_CHECKS=0 flag can be used to disable
all preconditions checks that uses libutil's Panic.h facilities.

this saves about 21KiB from the android release aar.

also removed the PANIC_ARITHMETIC macros, as these are just a type
of post-condition panics and were not used anywhere.


prideout (approved):

romainguy (approved):

bejado (approved):

pixelflinger:
Thinking more about it, this change is not not quite what we want. We need to discuss this more.

PR #3615, opened 2021-03-06
Add a "shift" parameter to Camera.

pixelflinger:
This has the effect of translating the viewport, without changing its
size. This is an effect similar to using a shift lens.

### DIFF
--- a/filament/include/filament/Camera.h
+++ b/filament/include/filament/Camera.h
@@ -227,6 +227,18 @@ public:
      */
     void setScaling(math::double4 const& scaling) noexcept;
 
+    /**
+     * Sets an additional matrix that shifts the projection matrix.
+     * By default, this is an identity matrix.
+     *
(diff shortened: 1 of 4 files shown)


pixelflinger:
In this PR I temporarily changed how gltf_viewer implements the sidebar (it's actually kind of cool because it doesn't require a viewport change, which would trigger the froxelizer update).
This miss the Java API, which is why it's a draft.

romainguy (approved):

bejado (approved):

prideout (approved):
Looks fine, what's the motivation for this?

romainguy:
Apparently useful to our users. Except to do it right now they need to hack the projection matrix. Unity has a similar camera feature. It replicates the shift feature of tilt shift lenses too.

pixelflinger:
And I like that it allows to implement something very similar to moving the viewport, but without having to reset it, meaning no relocation of the targets or update to the froxels.

PR #3616, opened 2021-03-06
Fix blanking windows not being drawn into on macOS

prewettg:
Multiple windows on macOS currently flicker: as one is drawn into, the other one is blanked to the NSView color. This is caused by NSOpenGLContext -clearDrawable, which not only disassociates the view from the context as the documentation says, but also blanks the view before disassociating from it. This seems to be necessary when the view is resized, but is not desirable in other situations. This PR only calls -clearDrawable when it can detect that the NSView has resized, otherwise it calls -setView: if necessary and -update.


prewettg:
CI is failing on Windows, but it seems to be because the compiler runs out of heap space on gltf_viewer, which hasn't changed in this PR. The only change was in an Objective-C++ file, so there should be no impact on Windows at all.

filament/backend/src/opengl/PlatformCocoaGL.mm
```diff
@@ -122,14 +129,27 @@
...
         pImpl->mCurrentView = nsView;
-        pImpl->updateOpenGLContext(nsView, true);
+        bool needsClear = false;
+        auto rectIter = pImpl->mPreviousSwapChainBounds.find(nsView);
+        if (rectIter == pImpl->mPreviousSwapChainBounds.end() ||
+            !NSEqualSizes(currentBounds.size, rectIter->second.size)) {
+            needsClear = true;
+        }
+        pImpl->updateOpenGLContext(nsView, true, needsClear);
     } else if (!CGRectEqualToRect(currentBounds, pImpl->mPreviousBounds)) {
```
prideout, line 147: Does the `mPreviousBounds` field need to exist?  Seems like this is duplicated data, since the conditional here is equivalent to: `!CGRectEqualToRect(currentBounds, pImpl->mPreviousSwapChainBounds[nsView])`.

prideout:
Thanks!  It might be useful to add a multi-window app to `samples/` to demonstrate the issue that is being fixed here.

filament/backend/src/opengl/PlatformCocoaGL.mm
```diff
@@ -28,6 +28,7 @@
 #include <OpenGL/OpenGL.h>
 #include <Cocoa/Cocoa.h>
 
+#include <map>
```
pixelflinger, line 31: use unordered_map instead

filament/backend/src/opengl/PlatformCocoaGL.mm
```diff
@@ -38,8 +39,9 @@
     NSOpenGLContext* mGLContext = nullptr;
     NSView* mCurrentView = nullptr;
     std::vector<NSView*> mHeadlessSwapChains;
-    NSRect mPreviousBounds = {};
-    void updateOpenGLContext(NSView *nsView, bool resetView);
+    std::map<NSView*, NSRect> mPreviousSwapChainBounds;
```
pixelflinger, line 42: 1. use unordered_map instead
2. we in fact don't need a map at all. But the implementation of SwapChain needs to change. Currently, we're simply casting the nsView to a Platform::SwapChain -- instead, we need to allocate a small CocoaGLSwapChain that stores the nsView and the bounds, and return that (and of course destroy that small structure when the swapchain is destroyed)

pixelflinger (changes requested):
The use of a map<> shouldn't be needed. The more correct solution is slightly more complex (but will be tons more efficient and save binary size).

filament/backend/src/opengl/PlatformCocoaGL.mm
prideout: Agreed with @pixelflinger.  @prewettg can you make these changes?

prewettg: @prideout Sure, will do. Do you have a preference on which of the options to go with?  Option 2, changing SwapChain, seems like it would simplify things a lot, so I'm inclined to go with that one, but can do whichever works best for the project.

prideout: Thanks!  We definitely want option 2.  :)

prideout:
Thank you, we will review this soon.

filament/backend/src/opengl/PlatformCocoaGL.mm
prewettg: Implemented option 2. I found some additional problems with our (Open3D) case, when resizing. The resizing problems don't seem to happen with the SDL sample I added, it may be related to our use of GLFW instead of SDL.

filament/backend/src/opengl/PlatformCocoaGL.mm
prewettg: No longer applicable because of the refactor below.

filament/backend/src/opengl/PlatformCocoaGL.mm
prewettg: I believe this was for the case when the context did not need to switch views, but the view had changed size. However, it is no longer needed after the refactor.

prewettg:
I added a sample as requested. I didn't use FilamentApp as this test is actually the guts, and it seemed simplest (from a reading-code perspective) to take just the parts from FilamentApp that are required to implement multiple windows. I implemented using SDL, since that's how FilamentApp does it. However, GLFW seems to have different behavior (at least within the Open3D framework). I'm not sure if there should be a GLFW version, or if that is out of scope, so I went with the latter.

filament/backend/src/opengl/PlatformCocoaGL.mm
```diff
@@ -34,12 +34,17 @@
 
 using namespace backend;
 
+struct CocoaGLSwapChain {
```
pixelflinger, line 37:
```suggestion
struct CocoaGLSwapChain : public Platform::SwapChain {
```

filament/backend/src/opengl/PlatformCocoaGL.mm
```diff
@@ -86,50 +91,90 @@
...
     // then we need to force an update to occur in the subsequent makeCurrent, which can be done by
     // simply resetting the current view. In multi-window situations, this happens automatically.
-    if (pImpl->mCurrentView == nsView) {
-        pImpl->mCurrentView = nullptr;
+    if (pImpl->mCurrentSwapChain && pImpl->mCurrentSwapChain->view == nsView) {
+        pImpl->mCurrentSwapChain = nullptr;
     }
 
-    return (SwapChain*) nativewindow;
+    return (Platform::SwapChain*)swapChain;
```
pixelflinger, line 106:
```suggestion
    return swapChain;
```

filament/backend/src/opengl/PlatformCocoaGL.mm
```diff
@@ -86,50 +91,90 @@
...
     // adding the pointer to the array retains the NSView
     pImpl->mHeadlessSwapChains.push_back(nsView);
-    return (__bridge SwapChain*)nsView;
+
+    CocoaGLSwapChain* swapChain = new CocoaGLSwapChain();
+    swapChain->view = nsView;
+    swapChain->previousBounds = NSZeroRect;
+    swapChain->previousWindowFrame = NSZeroRect;
+
+    return (Platform::SwapChain*)swapChain;
```
pixelflinger, line 120:
```suggestion
    return swapChain;
```

filament/backend/src/opengl/PlatformCocoaGL.mm
```diff
@@ -86,50 +91,90 @@
...
 }
 
 void PlatformCocoaGL::destroySwapChain(Platform::SwapChain* swapChain) noexcept {
+    CocoaGLSwapChain* cocoaSwapChain = (CocoaGLSwapChain*)swapChain;
```
pixelflinger, line 124:
```suggestion
    CocoaGLSwapChain* cocoaSwapChain = static_cast<CocoaGLSwapChain*>(swapChain);
```

pixelflinger (changes requested):
great, thanks! I suggested some changes to get better type safety.

prewettg:
@pixelflinger Sorry, I never actually checked Platform::SwapChain and was still in the mental mode of casting from the original code. Fixed to inherit from Platform::SwapChain.

pixelflinger:
This doesn't seem to build on Windows:
```
sdl2main.lib(SDL_windows_main.obj) : error LNK2019: unresolved external symbol SDL_main referenced in function main_getcmdline [D:\a\filament\filament\out\cmake-mt\samples\multiple_windows.vcxproj]
```

pixelflinger (approved):

pixelflinger (approved):

PR #3629, opened 2021-03-10
RemoteServer code cleanup and logcat cleanup.

prideout:
This fixes two different logcat messages that would occur when scrubbing
a slider in the UI:

- "Discarding message, message queue overflow."
- "Downloading model..." would repeat itself needlessly


prideout:
`rendertarget` and `hellotriangle` are both broken with OpenGL too, not just Metal.

PR #3630, opened 2021-03-10
Improve Actions hygiene

sethvargo:
👋 hello there! I'm a fellow Googler who works on projects that leverage GitHub Actions for CI/CD. Recently I noticed a large increase in our queue time, and I've tracked it down to the [limit of 180 concurrent jobs](https://docs.github.com/en/actions/reference/usage-limits-billing-and-administration) for an organization. To help be better citizens, I'm proposing changes across a few repositories that will reduce GitHub Actions hours and consumption. I hope these changes are reasonable and I'm happy to talk through them in more detail.

- Only run GitHub Actions for pushes and PRs against the main branch of the repository. If your team uses a forking model, this change will not affect you. If your team pushes branches to the repository directly, this changes actions to only run against the primary branches or if you open a Pull Request against a primary branch.

There are other changes you could make, depending on your project (but I'm not an expert):

- If you have tests that should only run when a subset of code changes, consider gating your workflow to particular file paths. For example, we have some jobs that do Terraform linting, but [they only run when Terraform files are changed](https://github.com/google/exposure-notifications-verification-server/blob/c4f59fee71042cf668747e599e7c769fca736554/.github/workflows/terraform.yml#L3-L11).

Hopefully these changes are not too controversial and also hopefully you can see how this would reduce actions consumption to be good citizens to fellow Googlers. If you have any questions, feel free to respond here or ping me on chat. Thank you!

### DIFF
--- a/.github/workflows/presubmit.yml
+++ b/.github/workflows/presubmit.yml
@@ -1,6 +1,12 @@
 name: Presubmit
 
-on: [pull_request]
+on:
+  push:
+    branches:
+    - main
+  pull_request:
+    branches:
+    - main
 
 jobs:
   build-desktop:


romainguy:
@sethvargo Can you explain this change in more detail please? Every PR we put up is in a branch branched off main. Would this change still run presubmits against those PRs? If so, I doubt it will reduce the load much (we do very little work in branches outside of PRs). If not, we would lose presubmits?

sethvargo:
Hey @romainguy 

Sure. Right now, you branch off of main on this repo, suppose that branch is named "my-pr" for easy reference. When you push that branch (`git push origin my-pr`) to google/filament, it immediately starts building because the current configuration is to run actions against all branches. Then, when you open the pull request, GitHub Actions does a ff-merge and builds _again_. You end up doubling the number of builds required. With this change, the build doesn't start until the user opens the PR, and only one build is triggered.

romainguy (approved):

sethvargo:
I think it's what most people want, but the default behavior is surprisingly resource-intensive. Please ping me if you have any questions!

romainguy:
Thanks! I'm surprised that the default does what you describe, I would expect `on: [pull_requests]` to trigger actions when PRs are actually open. I really appreciate your help on this!

bejado:
@sethvargo 
Why was "on push" added? We have a separate continuous workflow for commits made to `main`.

```
on:
  push:
    branches:
    - main
```

Wouldn't this be sufficient?

```
on:
  pull_request:
    branches:
    - main
```

PR #3631, opened 2021-03-10
Fix dominant light shadows in Metal and Vulkan

### LINKED ISSUE #3619: Dominant Light Shadows on Metal
**Describe the bug**
Dominant Light Shadows feature looks different and glitchy on Metal compared to OpenGL.

**To Reproduce**
Compare the same models with the same lighting parameters on OpenGL and Metal in the gltf_viewer sample.

**Expected behavior**
Dominant Light Shadows should look the same on Metal and OpenGL.

**Screenshots**
See added screenshot.

 - OS: macOS Big Sur 11.2.1
 - GPU: Intel Iris Plus Graphics 645
 - Backend: OpenGL/Metal

**Additional context**
On Metal seems like the direction of DLS is a different vector and the camera angle has a much more dramatic influence on it than on OpenGL.

dominant-shadows
.

bejado:
SAO calculations rely on two coordinate spaces:
- UV-space: framebuffer/texture space, graphics API dependent.
- screen-space: origin bottom left, normalized coordinates.

We didn't see any issues with OpenGL, as its UV-space matches screen-space (Y up). However, Metal and Vulkan's UV-space is vertically flipped (Y down).

This is my attempt to resolve the discrepancy by adding in conversions between the spaces when necessary. @pixelflinger I'm curious if you have other ideas or prefer different terminology.

Fixes #3619

### DIFF
--- a/filament/src/materials/ssao/sao.mat
+++ b/filament/src/materials/ssao/sao.mat
@@ -125,6 +125,12 @@ material {
 vertex {
     void postProcessVertex(inout PostProcessVertexInputs postProcess) {
         postProcess.vertex.xy = postProcess.normalizedUV;
+#if defined(TARGET_METAL_ENVIRONMENT) || defined(TARGET_VULKAN_ENVIRONMENT)
+        // On metal/vulkan postProcess.normalizedUV has its origin at the left-top, but we need a
+        // uniform coordinate space for SAO. So, we flip the y coordinate so we have bottom-left UV
+        // coordinates across all backends.
+        postProcess.vertex.y = 1.0 - postProcess.vertex.y;
+#endif
     }
 }
 
@@ -153,7 +159,8 @@ fragment {
         return fract(m.z * fract(dot(w, m.xy)));
     }
 
-    highp vec2 getApiAgnosticFragCoords() {
+    // returns the frag coord in the GL convention with (0, 0) at the bottom-left
+    highp vec2 getFragCoord() {
 #if defined(TARGET_METAL_ENVIRONMENT) || defined(TARGET_VULKAN_ENVIRONMENT)
         return vec2(gl_FragCoord.x, materialParams.resolution.y - gl_FragCoord.y);
 #else
@@ -161,8 +168,8 @@ fragment {
 #endif
     }
 
-    highp vec3 computeViewSpacePositionFromDepth(highp vec2 ss, highp float linearDepth) {
-        return vec3((0.5 - ss) * materialParams.positionParams.xy * linearDepth, linearDepth);
+    highp vec3 computeViewSpacePositionFromDepth(highp vec2 uv, highp float linearDepth) {
+        return vec3((0.5 - uv) * materialParams.positionParams.xy * linearDepth, linearDepth);
     }
 
     highp vec3 faceNormal(highp vec3 dpdx, highp vec3 dpdy) {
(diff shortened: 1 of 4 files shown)


filament/src/materials/ssao/sao.mat
```diff
@@ -153,8 +153,16 @@ fragment {
         return fract(m.z * fract(dot(w, m.xy)));
     }
 
-    highp vec3 computeViewSpacePositionFromDepth(highp vec2 uv, highp float linearDepth) {
-        return vec3((0.5 - uv) * materialParams.positionParams.xy * linearDepth, linearDepth);
+    highp vec2 getApiAgnosticFragCoords() {
```
romainguy, line 156: Maybe just getFragCoord()? (at least drop the s at the end)

filament/src/materials/ssao/ssaoUtils.fs
```diff
@@ -13,5 +13,16 @@ highp float sampleDepthLinear(const sampler2D depthTexture, const vec2 uv, float
     return linearizeDepth(textureLod(depthTexture, uv, lod).r, depthParams);
 }
 
+highp vec2 ssFromUv(highp vec2 uv) {
```
romainguy, line 16: I read this as "screenSpace", it's not really screen space, it's still a UV. Not sure what a better name would be. screenUvFromUv()?

filament/src/materials/ssao/sao.mat
pixelflinger: +1 for  `getFragCoord`

filament/src/materials/ssao/sao.mat
```diff
@@ -316,7 +332,7 @@ fragment {
 
 #if defined(TARGET_MOBILE)
         // this line is needed to workaround what seems to be a bug on qualcomm hardware
-        aoVisibility += gl_FragCoord.x * MEDIUMP_FLT_MIN;
+        aoVisibility += getApiAgnosticFragCoords().x * MEDIUMP_FLT_MIN;
```
pixelflinger, line 335: this one is just a workaround for qualcomm, I would just leave it.

filament/src/materials/ssao/ssaoUtils.fs
pixelflinger: Agreed, I also thought it was "screen-space" -- but it's like a normalized screenspace, isn't that "NDC", Normalized device coordinate?
This method needs a comment about what the input/output are.
Maybe `NdcFromUv` ?

filament/src/materials/ssao/ssaoUtils.fs
```diff
@@ -13,5 +13,16 @@ highp float sampleDepthLinear(const sampler2D depthTexture, const vec2 uv, float
...
+#if defined(TARGET_METAL_ENVIRONMENT) || defined(TARGET_VULKAN_ENVIRONMENT)
+    return vec2(uv.x, 1.0 - uv.y);
+#endif
+    return uv;
+}
+
+highp vec2 uvFromSs(highp vec2 ss) {
+    return ssFromUv(ss);
```
pixelflinger, line 24: I would just reimplement the code here -- it's super confusing to call the opposite function.

filament/src/materials/ssao/sao.mat
pixelflinger: We need a comment that explains that it's returning in the gl convention (0,0) at bottom-left.

filament/src/materials/ssao/sao.mat
```diff
@@ -287,18 +298,23 @@ fragment {
...
 
+    // For SAO calculations, we make use of two coordinate spaces:
+    //     UV-space: framebuffer/texture space, graphics API dependent.
+    //     screen-space: origin bottom left, normalized coordinates.
+    // Textures must always be sampled with UV coordinates.
+
     void postProcess(inout PostProcessInputs postProcess) {
         highp vec2 uv = variable_vertex.xy; // interpolated to pixel center
 
         highp float depth = sampleDepthLinear(materialParams_depth, uv, 0.0, materialParams.depthParams);
```
pixelflinger, line 316: if you made `sampleDepthLinear` take a `ss` you'd be able to remove a lot of calls to `uvFromSs` -- might simplify things.

pixelflinger: Same thing with `computeViewSpaceNormal`.

I feel like we should do all the work in `ss` and pass `ss` everywhere, except at texture sampling time, where you can do a `uvFromSs`.

pixelflinger (changes requested):

filament/src/materials/ssao/ssaoUtils.fs
romainguy: It's not NDC either; NDC uses -1…1 coordinates, here it's 0…1.

bejado: Yeah this isn't NDC. It's a left hand, normalized screen space. I'm fine with using "screenUV" as the terminology, but I'm not sure how to abbreviate it. `suv`?

bejado:
So, I ended up making this more complicated than necessary. There's no need to introduce any new terminology. On Metal/Vulkan, we can simply flip the UV coordinates in the vertex shader and then flip again when we go to sample the depth texture.

pixelflinger (approved):
much nicer! Thanks!

PR #3634, opened 2021-03-11
Fixes some issues with imported rendertargets

### LINKED ISSUE #3628: Mac rendering broken by commit [1189c9e29]
**Describe the bug**
In recent builds, I'm seeing incorrect rendering on macOS.  If I had to guess what's going wrong by looking at the output, I'd say that the depth buffer isn't getting cleared between frames.

**To Reproduce**
1. Invoke the `rendertarget` sample on the command line, like `rendertarget -a metal`.

**Expected behavior**
The sample is supposed to show a rotating head and a mirror reflection.

**Actual behavior**
The direct image looks OK, but the mirror image kind of disappears piecemeal.

**Screenshots**
I have a hard time snapping a screen shot fast enough, but here's one with just a little of the reflection still present.

Screen Shot 2021-03-10 at 12 58 31 PM

 - OS: macOS 11.3 Beta
 - GPU: Radeon Pro 560X 4 GB
 - Backend: Metal

**Additional context**
If I run the `rendertarget` sample using OpenGL, then the reflection is completely missing from the start.

It appears that the problem started with the commit "FrameGraph rewrite from the ground up  [1189c9e29]".

@pixelflinger

pixelflinger:
We were not declaring the attachments it was using.

Fixes #3628

### DIFF
--- a/filament/src/RenderTarget.cpp
+++ b/filament/src/RenderTarget.cpp
@@ -119,6 +119,12 @@ FRenderTarget::FRenderTarget(FEngine& engine, const RenderTarget::Builder& build
     mHandle(createHandle(engine, builder)) {
     mAttachments[COLOR] = builder.mImpl->mAttachments[COLOR];
     mAttachments[DEPTH] = builder.mImpl->mAttachments[DEPTH];
+    if (mAttachments[COLOR].texture) {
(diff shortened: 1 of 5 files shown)


romainguy (approved):

pixelflinger:
Ah good point. Definitely yes.

bejado (approved):

PR #3639, opened 2021-03-11
gltfio: Fix tangents when morphing is enabled.

### LINKED ISSUE #1609: GLTFIO morph target tangents
I'm not sure that morph target tangents are being processed properly when importing GLTF models into Filament. Consider [AnimatedMorphCube](https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/2.0/AnimatedMorphCube), which has a sloping animated face. With a single directional light source we would expect to see some variation on the sloping face as seen in [this PR](https://github.com/KhronosGroup/glTF-Sample-Models/pull/171). Instead we see something like this:

Screenshot 2019-09-09 at 14 56 31
Screenshot 2019-09-09 at 14 56 19

I'm trying to do the same import in my own engine and I was hoping to copy from GLTFIO. Looking at the code I can't quite work out how it can use the same function to combine GLTF normals and tangents in the morph targets that is uses for the base primitive normals and tangents. The morph targets are deltas, but the base attributes are absolute values. I may of course be missing something subtle that is happening prior to getting to that point in the code.

prideout:
The vertex data for glTF morph targets are deltas rather than final
values, so we were passing incorrect data into the SurfaceOrientation
helper.

This has been a long-standing bug that affected models with morphing,
including Buster Drone. However in practice we have observed visual
issues only with AnimatedMorphCube.

Tested against AnimatedMorphCube, Buster Drone, Cesium Man, Fox, and
Littlest Tokyo.

Fixes #1609

### DIFF
--- a/libs/gltfio/src/ResourceLoader.cpp
+++ b/libs/gltfio/src/ResourceLoader.cpp
@@ -784,43 +784,43 @@ void ResourceLoader::Impl::computeTangents(FFilamentAsset* asset) {
         const cgltf_primitive& prim = *params->prim;
         const uint8_t slot = params->slot;
         const int morphTargetIndex = params->morphTargetIndex;
+        const bool isMorphTarget = morphTargetIndex != kMorphTargetUnused;
 
-        // Declare vectors of normals and tangents, which we'll extract & convert from the source.
-        std::vector<float3> fp32Normals;
-        std::vector<float4> fp32Tangents;
-        std::vector<float3> fp32Positions;
-        std::vector<float2> fp32TexCoords;
-        std::vector<uint3> ui32Triangles;
+        // Declare storage for data that has been unpacked and converted from the source buffers.
+        // This data needs to be held until after the SurfaceOrientation helper consumes it.
+        std::vector<float3> unpackedNormals;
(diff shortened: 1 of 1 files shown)


libs/gltfio/src/ResourceLoader.cpp
```diff
@@ -784,114 +784,130 @@ void ResourceLoader::Impl::computeTangents(FFilamentAsset* asset) {
...
+        }
+        if (isMorphTarget) {
             const cgltf_morph_target& morphTarget = prim.targets[morphTargetIndex];
             for (cgltf_size aindex = 0; aindex < morphTarget.attributes_count; aindex++) {
                 const cgltf_attribute& attr = morphTarget.attributes[aindex];
                 if (attr.index == 0) {
-                    accessors[attr.type] = attr.data;
-                    vertexCount = attr.data->count;
+                    assert(baseAccessors[attr.type] &&
+                            "Morph target data has no corresponding base vertex data.");
```
romainguy, line 818: I never thought of doing that to add a message to an assert!

romainguy (approved):
🥳

pixelflinger (approved):

PR #3647, opened 2021-03-12
Revert GL backend handle tracking

pixelflinger:
This wasn't very useful in the first place because we're recycling
handles very quickly. Additionally there was a race condition
which cause false positives.

This reverts commit bc6acd5c5a1d7ee4b703fe9facfd8bf695b617d0.
This reverts commit 3a15756c78605395c4bdd2fa29d25587cc37e9f7.


romainguy (approved):

prideout (approved):
I'll cherry pick this into the rc.

PR #3655, opened 2021-03-16
fix rendertarget sample app

pixelflinger:
we can't use a projection with infinite far as a custom
projection matrix, because this is not suited for culling.

recently a new version of setCustomProjection was added which allows
to specify both the rendering and culling projections.


romainguy (approved):

prideout (approved):
THANKS!!

PR #3656, opened 2021-03-16
Fix RenderTarget slice bug

bejado:
Inside `RenderTarget`, we're only ever constructing `TargetBufferInfo` with a `TextureCubemapFace`. This is problematic, because `TargetBufferInfo` uses a `union` to keep track of the face or layer:

```
    union {
        // face if texture is a cubemap
        TextureCubemapFace face;
        // for 3D textures
        uint16_t layer = 0;
    };
```

`TextureCubemapFace` occupies only 8 bits, so even if we pass in `POSITIVE_X` (value of 0), the full 16 bits of the union aren't set, and when reading back `layer` we read garbage.

### DIFF
--- a/filament/src/RenderTarget.cpp
+++ b/filament/src/RenderTarget.cpp
@@ -28,7 +28,7 @@ namespace filament {
 using namespace backend;
 
 struct RenderTarget::BuilderDetails {
-    FRenderTarget::Attachment mAttachments[RenderTarget::ATTACHMENT_COUNT];
+    FRenderTarget::Attachment mAttachments[RenderTarget::ATTACHMENT_COUNT] = {};
 };
 
 using BuilderType = RenderTarget;
@@ -102,12 +102,14 @@ FRenderTarget::HwHandle FRenderTarget::createHandle(FEngine& engine, const Build
     // note that post-processing includes FXAA by default.
     const uint8_t samples = 1;
 
-    const TargetBufferInfo cinfo(upcast(color.texture)->getHwHandle(),
-            color.mipLevel, color.face);
+    const TargetBufferInfo cinfo =
+            upcast(color.texture)->getTarget() == Texture::Sampler::SAMPLER_CUBEMAP ?
+            TargetBufferInfo(upcast(color.texture)->getHwHandle(), color.mipLevel, color.face) :
+            TargetBufferInfo(upcast(color.texture)->getHwHandle(), color.mipLevel, color.layer);
 
     const TargetBufferInfo dinfo(
             depth.texture ? upcast(depth.texture)->getHwHandle() : TextureHandle(),
-            color.mipLevel, color.face);
+            color.mipLevel, color.layer);
 
     const uint32_t width = FTexture::valueForLevel(color.mipLevel, color.texture->getWidth());
     const uint32_t height = FTexture::valueForLevel(color.mipLevel, color.texture->getHeight());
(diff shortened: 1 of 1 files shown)


filament/src/RenderTarget.cpp
```diff
@@ -102,12 +102,14 @@ FRenderTarget::HwHandle FRenderTarget::createHandle(FEngine& engine, const Build
...
-            color.mipLevel, color.face);
+    const TargetBufferInfo cinfo =
+            upcast(color.texture)->getTarget() == Texture::Sampler::SAMPLER_CUBEMAP ?
+            TargetBufferInfo(upcast(color.texture)->getHwHandle(), color.mipLevel, color.face) :
+            TargetBufferInfo(upcast(color.texture)->getHwHandle(), color.mipLevel, color.layer);
 
     const TargetBufferInfo dinfo(
             depth.texture ? upcast(depth.texture)->getHwHandle() : TextureHandle(),
-            color.mipLevel, color.face);
+            color.mipLevel, color.layer);
```
bejado, line 112: I think this should be `depth.mipLevel` and `color.layer`, no?

prideout: Yes this must have been my typo.  😭

filament/src/RenderTarget.cpp
```diff
@@ -28,7 +28,7 @@ namespace filament {
 using namespace backend;
 
 struct RenderTarget::BuilderDetails {
-    FRenderTarget::Attachment mAttachments[RenderTarget::ATTACHMENT_COUNT];
+    FRenderTarget::Attachment mAttachments[RenderTarget::ATTACHMENT_COUNT] = {};
```
prideout, line 31: Was this necessary?  If so I don't totally understand how C++ field initializers work.

romainguy: An array is left uninitialized otherwise.

filament/src/RenderTarget.cpp
```diff
@@ -102,12 +102,14 @@ FRenderTarget::HwHandle FRenderTarget::createHandle(FEngine& engine, const Build
     // note that post-processing includes FXAA by default.
     const uint8_t samples = 1;
 
-    const TargetBufferInfo cinfo(upcast(color.texture)->getHwHandle(),
-            color.mipLevel, color.face);
+    const TargetBufferInfo cinfo =
+            upcast(color.texture)->getTarget() == Texture::Sampler::SAMPLER_CUBEMAP ?
+            TargetBufferInfo(upcast(color.texture)->getHwHandle(), color.mipLevel, color.face) :
+            TargetBufferInfo(upcast(color.texture)->getHwHandle(), color.mipLevel, color.layer);
```
prideout, line 108: 🙏

romainguy (approved):

filament/src/RenderTarget.cpp
prideout: But `Attachment` has field initializers.  Just now did a test in glot.io:

https://glot.io/snippets/fwstbhey6q

filament/src/RenderTarget.cpp
pixelflinger: Wait, wouldn't this need to also check the sampler type to use either .face or .layer?

filament/src/RenderTarget.cpp
pixelflinger: yeah, that seems strange that it's needed. I don't think it'll hurt too much anyways, since we already have a default initializer.

filament/src/RenderTarget.cpp
bejado: Hmm, yeah I think it should for completeness. My initial thought was that a depth cubemap doesn't make sense, but I suppose there are use-cases (like point light shadow maps).

prideout (approved):

pixelflinger (approved):

PR #3657, opened 2021-03-16
break dependency on libutils

pixelflinger:
it was only used of Path which now can be replaced bu
std::filesystem::path.


pixelflinger:
Looks like iOS does't have std::filesystem before 10.15

bejado:
I think we could move off of macOS 10.14 if that's the only requirement blocking this.

romainguy:
Do we have 10.15 in CI?

bejado:
Yes, `macos-latest` is 10.15.

PR #3658, opened 2021-03-16
gltfio: Refactor the tangents job.

prideout:
This cleans up the glTF tangents generator and allows it to be used in other parts of the library.  Also removed un-necessary usage of std vector.

### DIFF
--- a/libs/gltfio/src/ResourceLoader.cpp
+++ b/libs/gltfio/src/ResourceLoader.cpp
@@ -18,6 +18,7 @@
 #include <gltfio/Image.h>
 
 #include "FFilamentAsset.h"
+#include "TangentsJob.h"
 #include "upcast.h"
 
 #include <filament/Engine.h>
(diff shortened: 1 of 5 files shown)


libs/gltfio/src/ResourceLoader.cpp
```diff
@@ -932,36 +781,36 @@ void ResourceLoader::Impl::computeTangents(FFilamentAsset* asset) {
...
 
     // Create a job description for each primitive.
-    std::vector<JobParams> jobParams;
+    std::vector<TangentsJob::Params> jobParams;
     for (auto pair : asset->mPrimitives) {
         VertexBuffer* vb = pair.second;
         auto iter = baseTangents.find(vb);
         if (iter != baseTangents.end()) {
-            jobParams.emplace_back(JobParams { pair.first, vb, iter->second, kMorphTargetUnused });
+            jobParams.push_back({{ pair.first }, {vb, iter->second }});
```
pixelflinger, line 789: could you use `emplace_back` here?

pixelflinger (approved):

libs/gltfio/src/ResourceLoader.cpp
prideout: Yes although C++ is unable to infer the type when using emplace.  I'll revert these lines, it was unintentional.

PR #3659, opened 2021-03-17
gltfio: Refactor Animator.

prideout:
This does not change any behavior, it just prepares for the new MorphHelper class that will arrive in the next PR.

### DIFF
--- a/libs/gltfio/src/Animator.cpp
+++ b/libs/gltfio/src/Animator.cpp
@@ -44,9 +44,9 @@ using namespace utils;
 
 namespace gltfio {
 
-using TimeValues = std::map<float, size_t>;
-using SourceValues = std::vector<float>;
-using BoneVector = std::vector<filament::math::mat4f>;
+using TimeValues = map<float, size_t>;
+using SourceValues = vector<float>;
+using BoneVector = vector<filament::math::mat4f>;
 
 struct Sampler {
     TimeValues times;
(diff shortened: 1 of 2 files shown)


libs/gltfio/src/Animator.cpp
```diff
@@ -453,4 +337,139 @@ const char* Animator::getAnimationName(size_t animationIndex) const {
...
+    cgltf_animation_channel* srcChannels = srcAnim.channels;
+    cgltf_animation_sampler* srcSamplers = srcAnim.samplers;
+    const Sampler* samplers = dst.samplers.data();
+    for (cgltf_size j = 0, nchans = srcAnim.channels_count; j < nchans; ++j) {
+        const cgltf_animation_channel& srcChannel = srcChannels[j];
+        auto iter = nodeMap.find(srcChannel.target_node);
+        if (iter == nodeMap.end()) {
+            slog.w << "No scene root contains node ";
+            if (srcChannel.target_node->name) {
+                slog.w << "'" << srcChannel.target_node->name << "' ";
```
pixelflinger, line 352: We probably don't wait logs in release builds. Also, the condition for the log needs a `unlikely`.

pixelflinger (approved):
There are a bunch if stray logs, which we probably don't want in release builds. Also, the conditions for the log need a `unlikely`.

libs/gltfio/src/Animator.cpp
prideout: Fixed, added a macro to help me omit these in release build.  There are lots of non-catastrophic errors that can occur when loading a gltf so I should probably do this elsewhere in the library too.

Issue #1852, opened 2019-11-06
Horse.glb does not animate correctly


cx20:
**Describe the bug**
I tried to display the [horse model](https://github.com/mrdoob/three.js/blob/dev/examples/models/gltf/Horse.glb) in the latest version of Filament.
https://github.com/mrdoob/three.js/blob/dev/examples/models/gltf/Horse.glb
However, it does not seem to animate correctly.

**To Reproduce**
[Filament + Hourse.glb result:](https://rawcdn.githack.com/cx20/gltf-test/53d24ea498413ad6b1c1f93a1a9e00afa5154d90/examples/filament/index.html?url=https://rawcdn.githack.com/mrdoob/three.js/aea0b4bc96a340cc0bc420971efe71ed5194062b/examples/models/gltf/Horse.glb&scale=0.01)
![filament_Horse](https://user-images.githubusercontent.com/1833979/68300430-177e0d80-00e1-11ea-9483-94368e793bfc.gif)

**Expected behavior**
[three.js + Hourse.glb result:](https://rawcdn.githack.com/cx20/gltf-test/53d24ea498413ad6b1c1f93a1a9e00afa5154d90/examples/threejs/index.html?url=https://rawcdn.githack.com/mrdoob/three.js/aea0b4bc96a340cc0bc420971efe71ed5194062b/examples/models/gltf/Horse.glb&scale=0.01)
![three_Horse_20190302](https://user-images.githubusercontent.com/1833979/68300464-295fb080-00e1-11ea-8154-0515fd8beb91.gif)

**Desktop (please complete the following information):**
- OS: Windows 10 + Chrome 78
- GPU: Intel HD Graphics 520
- Backend: WebGL/ANGLE (Intel(R) HD Graphics 520 Direct3D11 vs_5_0 ps_5_0)

prideout:
This mesh has 15 morph targets but Filament only supports 4 morph targets right now.  See also #1487.

As an aside, this mesh does not have normals.

donmccurdy:
> This mesh has 15 morph targets but Filament only supports 4 morph targets right now.

Note that three.js also does not support more than 4-8 morph targets, but automatically selects the morph targets with the highest weights if there are more than can be applied. For this style of animation (which is basically an MDD file baked into a glTF morph target sequence) there are never more than ~2 morph targets active at a time, so selecting the highest weight targets works well.

prideout:
[...]

PR #3661, opened 2021-03-17
gltfio: Add internal MorphHelper, enable up to 255 targets.

### LINKED ISSUE #1487: support blend shapes more than 8

Our company HyprSense is making an avatar animation technology (like iPhone X Animoji). We were looking for rendering engines to showcase 3D characters, and they usually have 40~50 blend shapes rigged in.

With Unity / Unreal Engine / Apple SceneKit, we had no problem because their vertex shader supports that number of blend shapes for a skinned mesh renderer. We are having difficulties with Sceneform, where the number of blend shapes is limited to 8 (or 4 with blending normals).

We want you to ask if you have a plan to implement a feature to support 40+ blend shapes. 

You might want to store the blend shape vertex and normal deformation information in a data texture, and load from the vertex shader to apply the deformation. There are similar threads discussing about this topic in the three.js repository, and they came up with the approach I have mentioned.

https://github.com/mrdoob/three.js/issues/15556
https://github.com/mrdoob/three.js/issues/14441

**OS and backend**
All OSes and backends

prideout:
horse2

This allows us to handle many more models, such as those exported by Modo, and the horse model in #1852.

We do this by picking the highest 4 weights. The indices for these weights are called the "primary indices", and we construct a new VertexBuffer for each unique set of primary indices.

In the near future this will be made more efficient by adding a new VertexBuffer API to the Filament core.

Fixes #1487.

### DIFF
--- a/libs/gltfio/src/FilamentAsset.cpp
+++ b/libs/gltfio/src/FilamentAsset.cpp
@@ -30,8 +30,6 @@ using namespace utils;
 namespace gltfio {
 
 FFilamentAsset::~FFilamentAsset() {
-    delete mAnimator;
-
     releaseSourceData();
 
     // The only things we need to free in the instances are their animators.
@@ -41,6 +39,7 @@ FFilamentAsset::~FFilamentAsset() {
         delete instance;
     }
 
+    delete mAnimator;
     delete mWireframe;
 
     mEngine->destroy(mRoot);
--- a/libs/gltfio/src/FilamentAsset.cpp
+++ b/libs/gltfio/src/FilamentAsset.cpp
@@ -30,6 +30,8 @@ using namespace utils;
 namespace gltfio {
 
 FFilamentAsset::~FFilamentAsset() {
+    delete mAnimator;
+
     releaseSourceData();
 
     // The only things we need to free in the instances are their animators.
@@ -39,7 +41,6 @@ FFilamentAsset::~FFilamentAsset() {
         delete instance;
     }
 
-    delete mAnimator;
     delete mWireframe;
 
     mEngine->destroy(mRoot);
--- a/libs/gltfio/src/MorphHelper.cpp
+++ b/libs/gltfio/src/MorphHelper.cpp
@@ -1,0 +1,366 @@
+/*
+ * Copyright (C) 2021 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "MorphHelper.h"
+
+#include <filament/RenderableManager.h>
+#include <filament/VertexBuffer.h>
+
+#include "GltfEnums.h"
+#include "TangentsJob.h"
+
+using namespace filament;
+using namespace filament::math;
+using namespace utils;
+
+static const auto kFreeCallback = [](void* mem, size_t, void*) { free(mem); };
+static constexpr uint8_t kUnused = 0xff;
+
+namespace gltfio {
+
+uint32_t computeBindingSize(const cgltf_accessor* accessor);
+uint32_t computeBindingOffset(const cgltf_accessor* accessor);
+
+// Returns true if a is a subset of b.
+static bool isSubsetOf(ubyte4 a, ubyte4 b) {
+    if (a.x != kUnused && a.x != b.x && a.x != b.y && a.x != b.z && a.x != b.w) return false;
+    if (a.y != kUnused && a.y != b.x && a.y != b.y && a.y != b.z && a.y != b.w) return false;
+    if (a.z != kUnused && a.z != b.x && a.z != b.y && a.z != b.z && a.z != b.w) return false;
+    if (a.w != kUnused && a.w != b.x && a.w != b.y && a.w != b.z && a.w != b.w) return false;
+    return true;
+}
+
+MorphHelper::MorphHelper(FFilamentAsset* asset, FFilamentInstance* inst) : mAsset(asset),
+        mInstance(inst) {
+    NodeMap& sourceNodes = asset->isInstanced() ? asset->mInstances[0]->nodeMap : asset->mNodeMap;
+    for (auto pair : sourceNodes) {
+        mMorphTable[pair.second].sourceNode = pair.first;
+    }
+}
+
+MorphHelper::~MorphHelper() {
+    auto engine = mAsset->mEngine;
+    for (auto& entry : mMorphTable) {
+        for (auto& perm : entry.second.permutations) {
+            if (perm.owner) {
+                for (auto& prim : perm.primitives) {
+                    engine->destroy(prim.vertices);
+                }
+            }
+        }
+    }
+}
+
+void MorphHelper::applyWeights(Entity entity, float const* weights, size_t count) {
+    auto renderableManager = &mAsset->mEngine->getRenderableManager();
+    auto renderable = renderableManager->getInstance(entity);
+
+    // If there are 4 or fewer targets, we can simply re-use the original VertexBuffer.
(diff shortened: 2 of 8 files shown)


romainguy (approved):
That's an awesome change!

libs/gltfio/src/FilamentAsset.cpp
```diff
@@ -30,6 +30,8 @@ using namespace utils;
 namespace gltfio {
 
 FFilamentAsset::~FFilamentAsset() {
+    delete mAnimator;
```
pixelflinger, line 33: curious why this change? if order is important, maybe a comment?

libs/gltfio/src/MorphHelper.h
```diff
@@ -0,0 +1,86 @@
...
+        std::vector<Permutation> permutations;
+        const cgltf_node* sourceNode;
+    };
+
+    void addPermutation(Entity targetEntity, filament::math::ubyte4 primaryIndices);
+
+    filament::VertexBuffer* createVertexBuffer(const cgltf_primitive& prim, const UvMap& uvmap,
+            filament::math::ubyte4 primaryIndices);
+
+    bool mEnabled = true;
```
pixelflinger, line 79: move the bool at the end of the structure, to potentially avoid padding

libs/gltfio/src/MorphHelper.h
```diff
@@ -0,0 +1,86 @@
...
+/**
+ * Internal class that partitions lists of morph weights and maintains a cache of VertexBuffer
+ * objects for each partition. This allows gltfio to support up to 255 morph targets.
+ *
+ * Each partition is associated with an unordered set of 4 (or fewer) morph target indices, which
+ * we call the "primary indices" for that time slice.
+ *
+ * Animator has ownership over a single instance of MorphHelper, thus it is 1:1 with FilamentAsset.
+ */
+class MorphHelper {
```
pixelflinger, line 40: add the `noexpect` decorator where needed (everywhere?)

libs/gltfio/src/MorphHelper.h
```diff
@@ -0,0 +1,86 @@
...
+
+    /**
+     * Picks the 4 most influential weights and applies them to the target entity.
+     */
+    void applyWeights(Entity targetEntity, float const* weights, size_t count);
+
+    /**
+     * Allows the caller to prime the MorphHelper cache without actually affecting any renderables.
+     */
+    void enableWrites(bool enable) { mEnabled = enable; }
```
pixelflinger, line 54: the description and name of the method seem very different? Also the name of the variable "mEnabled"

libs/gltfio/src/MorphHelper.h
```diff
@@ -0,0 +1,86 @@
...
+
+private:
+    struct GltfPrim {
```
pixelflinger, line 57: `GltfPrimitive` (we try to not abbreviate names in our style guide)

libs/gltfio/src/MorphHelper.cpp
```diff
@@ -0,0 +1,366 @@
...
+
+void MorphHelper::applyWeights(Entity entity, float const* weights, size_t count) {
+    auto renderableManager = &mAsset->mEngine->getRenderableManager();
+    auto renderable = renderableManager->getInstance(entity);
+
+    // If there are 4 or fewer targets, we can simply re-use the original VertexBuffer.
+    if (count <= 4) {
+        if (mEnabled) {
+            float4 vec(0, 0, 0, 0);
+            for (size_t i = 0; i < count; i++) vec[i] = weights[i];
```
pixelflinger, line 75: use { }

libs/gltfio/src/MorphHelper.cpp
```diff
@@ -0,0 +1,366 @@
...
+}
+
+void MorphHelper::applyWeights(Entity entity, float const* weights, size_t count) {
+    auto renderableManager = &mAsset->mEngine->getRenderableManager();
+    auto renderable = renderableManager->getInstance(entity);
+
+    // If there are 4 or fewer targets, we can simply re-use the original VertexBuffer.
+    if (count <= 4) {
+        if (mEnabled) {
+            float4 vec(0, 0, 0, 0);
```
pixelflinger, line 74: `float4 vec{}` -- slightly less verbose, not a biggie.

libs/gltfio/src/MorphHelper.cpp
```diff
@@ -0,0 +1,366 @@
...
+    for (size_t primIndex = 0; primIndex < permutation->primitives.size(); ++primIndex) {
+        const GltfPrim& prim = permutation->primitives[primIndex];
+        renderableManager->setGeometryAt(renderable, primIndex, prim.type, prim.vertices,
+               prim.indices, 0, prim.indices->getIndexCount());
+    }
+
+    // Finally, set the 4-tuple uniform for the weight values by derefing the primary indices.
+    // Note that we first create a "safe set" by replacing the unused sentinel with zero.
+    ubyte4 safe = permutation->primaryIndices;
+    for (int i = 0; i < 4; i++) if (safe[i] == kUnused) safe[i] = 0;
```
pixelflinger, line 137: omg! use `{}` here :-)

libs/gltfio/src/MorphHelper.cpp
```diff
@@ -0,0 +1,366 @@
...
+
+    for (cgltf_size pi = 0; pi < prim_count; ++pi) {
+        const cgltf_primitive& prim = prims[pi];
+        RenderableManager::PrimitiveType prim_type;
+        getPrimitiveType(prim.type, &prim_type);
+        const auto& gltfioPrim = mAsset->mMeshCache.at(mesh)[pi];
+        VertexBuffer* const vertices = result.owner ?
+                createVertexBuffer(prim, gltfioPrim.uvmap, primaryIndices) :
+                gltfioPrim.vertices;
+        result.primitives.push_back({
```
pixelflinger, line 165: could it be `emplace_back` here?

libs/gltfio/src/MorphHelper.cpp
```diff
@@ -0,0 +1,366 @@
...
+    }();
+
+    // This creates a copy because we don't know when the user will free the cgltf source data.
+    // For non-morphed vertex buffers, we use a sharing mechanism to prevent copies, but here
+    // we just want to keep it as simple as possible.
+    auto createBufferDescriptor = [](const cgltf_accessor* accessor) {
+        auto bufferData = (const uint8_t*) accessor->buffer_view->buffer->data;
+        const uint8_t* data = computeBindingOffset(accessor) + bufferData;
+        const uint32_t size = computeBindingSize(accessor);
+        uint8_t* clone = (uint8_t*) malloc(size);
```
pixelflinger, line 205: this makes me very nervous, the `malloc` happens at a different place frmo the `free`. Could this be refactored to ahve the malloc/free in the same object, or, use a `unique_ptr`?

pixelflinger (changes requested):

libs/gltfio/src/FilamentAsset.cpp
prideout: Accidental movement, reverted.

libs/gltfio/src/MorphHelper.h
prideout: Done.

libs/gltfio/src/MorphHelper.h
prideout: Reworded the comment, also renamed the method to `disableWrites`.

libs/gltfio/src/MorphHelper.h
prideout: Done.

libs/gltfio/src/MorphHelper.cpp
prideout: Don.e

libs/gltfio/src/MorphHelper.cpp
prideout: Done.

libs/gltfio/src/MorphHelper.cpp
prideout: Done :)

libs/gltfio/src/MorphHelper.cpp
prideout: The `free` is actually just two lines after the malloc but you may've missed it because it was wrapped.  I fixed this to be more obvious.  (emscripten does not let me pass `free` directly into the callback)

pixelflinger: ah yes... duh.

pixelflinger (approved):

PR #3662, opened 2021-03-17
gltfio: Add support for data:// in image URI's.

prideout:


romainguy (approved):
I trust that the string parsing bits are safe :)

PR #3677, opened 2021-03-22
CString can construct from a string_view

AdrianAtGoogle:
(string_views lack a null terminator)

### DIFF
--- a/libs/utils/src/CString.cpp
+++ b/libs/utils/src/CString.cpp
@@ -41,7 +41,7 @@ CString::CString(const char* cstr, size_t length) {
         mCStr = (value_type*)(p + 1);
         // we don't use memcpy here to avoid a call to libc, the generated code is pretty good.
         std::uninitialized_copy_n(cstr, length, mCStr);
-        mCStr[length + 1] = '\0';
+        mCStr[length] = '\0';
     }
 }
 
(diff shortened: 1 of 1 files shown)


romainguy (commented):
LGTM, but I'll let @pixelflinger approve

pixelflinger (approved):
wow. good one.

pixelflinger:
strangely it looks like this is breaking stuff. I'm going to have a look next.

libs/utils/src/CString.cpp
```diff
@@ -40,7 +40,8 @@ CString::CString(const char* cstr, size_t length) {
         p->length = (size_type)length;
         mCStr = (value_type*)(p + 1);
         // we don't use memcpy here to avoid a call to libc, the generated code is pretty good.
-        std::uninitialized_copy_n(cstr, length + 1, mCStr);
+        std::uninitialized_copy_n(cstr, length, mCStr);
+        mCStr[length + 1] = '\0';
```
pixelflinger, line 44:
```suggestion
        mCStr[length] = '\0';
```

pixelflinger (changes requested):

AdrianAtGoogle:
lol thanks for the off-by-1 fix :)

PR #3679, opened 2021-03-22
Dof: cleanup and improve significantly foreground layer

pixelflinger:
We're implementing the two-layer technique described in
Jimenez DoF paper, where the foreground field is split in two layers, front
and back. In addition we compute accurately the foreground field alpha.

From a performance standpoint, most of the extra work is increased
register pressure and a bit of ALU.

before
after

### DIFF
--- a/filament/src/materials/dof/dof.mat
+++ b/filament/src/materials/dof/dof.mat
@@ -106,6 +106,29 @@ float sampleCount(const float ringCount) {
     return s * s;
 }
 
+float getMipLevel(const float ringCount, const float kernelSize) {
+    #if KERNEL_USE_MIPMAP
+    // note: the 0.5 is to convert from highres to our downslampled texture
+    float ringDistanceInTexels = 0.5 * kernelSize * rcp(ringCount - 0.5);
+    float mip = log2(ringDistanceInTexels);
+
+    // We should round-up to avoid too large gaps in the kernel. Small gaps are filled by the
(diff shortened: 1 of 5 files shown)


filament/src/materials/dof/dof.mat
```diff
@@ -181,11 +200,10 @@ void initBucket(out Bucket ring) {
 
 void initRing(const float i, const float ringCount, const float kernelSize, const vec2 noise,
         out float offset, out float count, out mat2 r, out vec2 p) {
-    const float ringDensity = float(RING_DENSITY);
     float radius = (kernelSize / (ringCount - 0.5)) * i;
 
     // this is never called with i == 0
-    count = ringDensity * i;
+    count = 8.0 * i;
```
romainguy, line 206: Why 8?

romainguy (approved):

filament/src/materials/dof/dof.mat
pixelflinger: Oh, this is because we want to create a (2k+1)*(2k+1) box, that is a n*n grid with n odd. Each "ring" is the outside layer of that box. So ring 0 obviously is the 1x1 box, ring 1 is the 3x3 box, which has 8 samples on its outside layer. This is where the 8 comes from, then all subsequent rings, are built from ring 1.
Each "box" is then "morphed" to the shape we want for the bokeh (i.e. a circle).

romainguy: That would be a perfect comment for that line :p

PR #3680, opened 2021-03-22
DoF: add a "filter" parameter to the DoF settings

pixelflinger:
Currently the only NONE and MEDIAN are available options, at some 
point we may add a "MAX" option which is cheaper.

There is no real practical uses for this option with these only two
choices, other than for debugging.

### DIFF
--- a/web/filament-js/jsenums.cpp
+++ b/web/filament-js/jsenums.cpp
@@ -136,6 +136,10 @@ enum_<View::AntiAliasing>("View$AntiAliasing")
     .value("NONE", View::AntiAliasing::NONE)
     .value("FXAA", View::AntiAliasing::FXAA);
 
+enum_<View::DepthOfFieldOptions::Filter>("View$DepthOfFieldOptions$Filter")
+    .value("NONE", View::DepthOfFieldOptions::Filter::NONE)
+    .value("MEDIAN", View::DepthOfFieldOptions::Filter::MEDIAN);
+
 enum_<Camera::Fov>("Camera$Fov")
     .value("VERTICAL", Camera::Fov::VERTICAL)
     .value("HORIZONTAL", Camera::Fov::HORIZONTAL);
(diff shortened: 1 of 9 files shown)


romainguy (approved):

prideout (commented):
We also need to update `filament.d.ts`, `jsenums.cpp`, `jsbindings.cpp`, and `Settings.cpp`.

pixelflinger:
> We also need to update `filament.d.ts`, `jsenums.cpp`, `jsbindings.cpp`, and `Settings.cpp`.

I did update `Settings.cpp` right? Is there more to do there?
I'll update the other ones too.

web/filament-js/jsenums.cpp
```diff
@@ -136,6 +136,10 @@ enum_<View::AntiAliasing>("View$AntiAliasing")
     .value("NONE", View::AntiAliasing::NONE)
     .value("FXAA", View::AntiAliasing::FXAA);
 
+enum_<View::DepthOfFieldOptions::Filter>("View$DepthOfFieldOptions$Filter")
+    .value("NONE", View::DepthOfFieldOptions::Filter::NONE)
+    .value("MEDIAN", View::DepthOfFieldOptions::Filter::MEDIAN);
```
prideout, line 141: I think this will bind MEDIAN to 1 instead of 2.  Can you put a placeholder for 1?

prideout (approved):
Oh didn't see Settings.cpp, looks fine.  I think the only remaining issue is that JavaScript needs a placeholder value in the enum.

web/filament-js/jsenums.cpp
pixelflinger: I was wondering about that so the right side "View::DepthOfFieldOptions::Filter::MEDIAN" that's not the C++ value?

prideout: God I'm having a bad day, you're right Mathias, the code is fine as is.

prideout (approved):

prideout (commented):
One last thing.  :)  Settings.cpp needs to be updated in `static std::ostream& operator<<(std::ostream& out, const DepthOfFieldOptions& in)`

pixelflinger:
> One last thing. :) Settings.cpp needs to be updated in `static std::ostream& operator<<(std::ostream& out, const DepthOfFieldOptions& in)`

OK, Columbo! :-)

PR #3681, opened 2021-03-23
Introduce BufferObject API.

prideout:
NOTE: after Metal and Vulkan support lands, we can potentially simplify
the Driver API by removing some of the old VertexBuffer entry points.

This introduces a new API-level object called `BufferObject`, and a new
backend-level object called `HwBufferObject`. This encapsulates a single
VBO. It's especially useful when clients need to share data between
multiple VertexBuffer objects. (e.g. for morphing).

In the future, buffer objects could perhaps be used for non-vertex data
(e.g. for compute).

The existing `VertexBuffer` class can now be configured to use buffer
objects by calling `enableBufferObjects` on the builder. By default its
API is unchanged. If buffer objects are enabled, then clients should use
`setBufferObjectAt` (which takes a buffer object) rather than
`setBufferAt` (which takes CPU data).

OpenGL is the trickiest backend for this, due to existence of VAOs.
The proposed implementation in this PR uses a version-tracking scheme to
figure out when to update the VAO.

### DIFF
--- a/filament/backend/include/backend/DriverEnums.h
+++ b/filament/backend/include/backend/DriverEnums.h
@@ -263,6 +263,11 @@ enum class ElementType : uint8_t {
     HALF4,
 };
 
+//! Buffer object binding type
+enum class BufferObjectBinding : uint8_t {
+    VERTEX,
+};
+
 //! Face culling Mode
 enum class CullingMode : uint8_t {
     NONE,               //!< No culling, front and back faces are visible
--- a/filament/backend/src/opengl/OpenGLDriver.cpp
+++ b/filament/backend/src/opengl/OpenGLDriver.cpp
@@ -412,6 +412,10 @@ Handle<HwIndexBuffer> OpenGLDriver::createIndexBufferS() noexcept {
     return initHandle<GLIndexBuffer>();
 }
 
+Handle<HwBufferObject> OpenGLDriver::createBufferObjectS() noexcept {
+    return initHandle<GLBufferObject>();
+}
+
 Handle<HwRenderPrimitive> OpenGLDriver::createRenderPrimitiveS() noexcept {
     return initHandle<GLRenderPrimitive>();
 }
@@ -478,12 +482,13 @@ void OpenGLDriver::createVertexBufferR(
         uint8_t attributeCount,
         uint32_t elementCount,
         AttributeArray attributes,
-        BufferUsage usage) {
+        BufferUsage usage,
+        bool bufferObjectsEnabled) {
     DEBUG_MARKER()
 
     auto& gl = mContext;
     GLVertexBuffer* vb = construct<GLVertexBuffer>(vbh,
-            bufferCount, attributeCount, elementCount, attributes);
+            bufferCount, attributeCount, elementCount, attributes, bufferObjectsEnabled);
 
     GLsizei n = GLsizei(vb->bufferCount);
 
@@ -521,6 +526,24 @@ void OpenGLDriver::createIndexBufferR(
     gl.bindVertexArray(nullptr);
     gl.bindBuffer(GL_ELEMENT_ARRAY_BUFFER, ib->gl.buffer);
     glBufferData(GL_ELEMENT_ARRAY_BUFFER, size, nullptr, getBufferUsage(usage));
+    CHECK_GL_ERROR(utils::slog.e)
+}
+
+void OpenGLDriver::createBufferObjectR(
+        Handle<HwBufferObject> boh,
+        uint32_t byteCount,
+        BufferObjectBinding bindingType) {
+    DEBUG_MARKER()
+
+    auto& gl = mContext;
+    GLBufferObject* bo = construct<GLBufferObject>(boh);
+    glGenBuffers(1, &bo->gl.id);
+    gl.bindVertexArray(nullptr);
+
+    assert_invariant(bindingType == BufferObjectBinding::VERTEX);
+
+    gl.bindBuffer(GL_ARRAY_BUFFER, bo->gl.id);
+    glBufferData(GL_ARRAY_BUFFER, byteCount, nullptr, GL_STATIC_DRAW);
     CHECK_GL_ERROR(utils::slog.e)
 }
 
@@ -755,6 +778,57 @@ void OpenGLDriver::importTextureR(Handle<HwTexture> th, intptr_t id,
     }
 
     CHECK_GL_ERROR(utils::slog.e)
+}
+
+void OpenGLDriver::updateVertexArrayObject(GLRenderPrimitive* rp, GLVertexBuffer const* vb) {
+
+    auto& gl = mContext;
+
+    // NOTE: this is called from draw() and must be as efficient as possible.
+
+    // The VAO for the given render primitive must already be bound.
+    #ifndef NDEBUG
+    GLint vaoBinding;
+    glGetIntegerv(GL_VERTEX_ARRAY_BINDING, &vaoBinding);
+    assert_invariant(vaoBinding == (GLint) rp->gl.vao);
+    #endif
+
+    rp->gl.vertexBufferVersion = vb->bufferObjectsVersion;
+    rp->maxVertexCount = vb->vertexCount;
+    for (size_t i = 0, n = vb->attributes.size(); i < n; i++) {
+        const auto& attribute = vb->attributes[i];
+        if (attribute.buffer != Attribute::BUFFER_UNUSED) {
+            uint8_t bi = attribute.buffer;
+            gl.bindBuffer(GL_ARRAY_BUFFER, vb->gl.buffers[bi]);
+            if (UTILS_UNLIKELY(attribute.flags & Attribute::FLAG_INTEGER_TARGET)) {
+                glVertexAttribIPointer(GLuint(i),
+                        getComponentCount(attribute.type),
+                        getComponentType(attribute.type),
+                        attribute.stride,
+                        (void*) uintptr_t(attribute.offset));
+            } else {
+                glVertexAttribPointer(GLuint(i),
+                        getComponentCount(attribute.type),
+                        getComponentType(attribute.type),
+                        getNormalization(attribute.flags & Attribute::FLAG_NORMALIZED),
+                        attribute.stride,
+                        (void*) uintptr_t(attribute.offset));
+            }
+
+            gl.enableVertexAttribArray(GLuint(i));
+        } else {
+
+            // In some OpenGL implementations, we must supply a properly-typed placeholder for
+            // every integer input that is declared in the vertex shader.
+            if (UTILS_UNLIKELY(attribute.flags & Attribute::FLAG_INTEGER_TARGET)) {
+                glVertexAttribI4ui(GLuint(i), 0, 0, 0, 0);
+            } else {
+                glVertexAttrib4f(GLuint(i), 0, 0, 0, 0);
+            }
+
+            gl.disableVertexAttribArray(GLuint(i));
+        }
+    }
 }
 
 void OpenGLDriver::framebufferTexture(backend::TargetBufferInfo const& binfo,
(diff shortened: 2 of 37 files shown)


filament/backend/src/opengl/OpenGLDriver.cpp
```diff
@@ -1617,6 +1626,7 @@ void OpenGLDriver::updateVertexBuffer(Handle<HwVertexBuffer> vbh,
 
     auto& gl = mContext;
     GLVertexBuffer* eb = handle_cast<GLVertexBuffer *>(vbh);
+    assert(!eb->bufferObjectsEnabled && "Please use setVertexBufferObject() instead.");
```
pixelflinger, line 1629:
```suggestion
    assert_invariant(!eb->bufferObjectsEnabled && "Please use setVertexBufferObject() instead.");
```

filament/backend/src/opengl/OpenGLDriver.cpp
```diff
@@ -1626,6 +1636,28 @@ void OpenGLDriver::updateVertexBuffer(Handle<HwVertexBuffer> vbh,
     CHECK_GL_ERROR(utils::slog.e)
 }
 
+void OpenGLDriver::setVertexBufferObject(Handle<HwVertexBuffer> vbh,
+        size_t index, Handle<HwBufferObject> boh, uint32_t byteOffset) {
+   DEBUG_MARKER()
+
+    GLVertexBuffer* vb = handle_cast<GLVertexBuffer *>(vbh);
+    assert(vb->bufferObjectsEnabled && "Please use updateVertexBuffer() instead.");
```
pixelflinger, line 1644:
```suggestion
    assert_invariant(vb->bufferObjectsEnabled && "Please use updateVertexBuffer() instead.");
```

filament/backend/src/opengl/OpenGLDriver.h
```diff
@@ -69,6 +69,10 @@ class OpenGLDriver final : public backend::DriverBase {
         backend::BufferUsage usage = {};
     };
 
+    struct GLBufferObject : public backend::HwBufferObject {
+        GLuint id = 0;
```
pixelflinger, line 73: The convention I have used elsewhere is that all OpenGL state variables go into an anonymous `struct { } gl;` structure.

filament/include/filament/BufferObject.h
```diff
@@ -0,0 +1,109 @@
...
+        ~Builder() noexcept;
+        Builder& operator=(Builder const& rhs) noexcept;
+        Builder& operator=(Builder&& rhs) noexcept;
+
+        /**
+         * Size of the buffer in bytes.
+         * @param byteCount Maximum number of bytes the BufferObject can hold.
+         * @return A reference to this Builder for chaining calls.
+         */
+        Builder& size(uint32_t byteCount) noexcept;
```
pixelflinger, line 69: Do we need a `.type()` method to distinguish between SSBO, VBO, etc?
Maybe we should add it now with only a single value possible and documenting it defaults to VBO?

filament/src/VertexBuffer.cpp
```diff
@@ -188,6 +196,7 @@ size_t FVertexBuffer::getVertexCount() const noexcept {
 
 void FVertexBuffer::setBufferAt(FEngine& engine, uint8_t bufferIndex,
         backend::BufferDescriptor&& buffer, uint32_t byteOffset) {
+    ASSERT_PRECONDITION_NON_FATAL(!mBufferObjectsEnabled, "Please use setBufferObjectAt()");
```
pixelflinger, line 199: You cannot (shouldn't) use `ASSERT_PRECONDITION_NON_FATAL` here because you cannot return the failure to the user in any meaningful way. Use `ASSERT_PRECONDITION` instead.

filament/src/VertexBuffer.cpp
```diff
@@ -197,6 +206,19 @@ void FVertexBuffer::setBufferAt(FEngine& engine, uint8_t bufferIndex,
     }
 }
 
+void FVertexBuffer::setBufferObjectAt(FEngine& engine, uint8_t bufferIndex,
+        FBufferObject const * bufferObject, uint32_t byteOffset) {
+    ASSERT_PRECONDITION_NON_FATAL(mBufferObjectsEnabled, "Please use setBufferAt()");
```
pixelflinger, line 211:
```suggestion
    ASSERT_PRECONDITION(mBufferObjectsEnabled, "Please use setBufferAt()");
```

filament/src/VertexBuffer.cpp
```diff
@@ -197,6 +206,19 @@ void FVertexBuffer::setBufferAt(FEngine& engine, uint8_t bufferIndex,
...
+    if (bufferIndex < mBufferCount) {
+        auto hwBufferObject = bufferObject->getHwHandle();
+        engine.getDriverApi().setVertexBufferObject(mHandle,
+                bufferIndex, hwBufferObject, byteOffset);
+    } else {
+        ASSERT_PRECONDITION_NON_FATAL(bufferIndex < mBufferCount,
```
pixelflinger, line 217: Same here.

filament/backend/src/opengl/OpenGLDriver.cpp
```diff
@@ -3112,7 +3145,17 @@ void OpenGLDriver::draw(PipelineState state, Handle<HwRenderPrimitive> rph) {
...
     useProgram(p);
 
-    const GLRenderPrimitive* rp = handle_cast<const GLRenderPrimitive *>(rph);
+    GLRenderPrimitive* rp = handle_cast<GLRenderPrimitive *>(rph);
+
+    // If necessary, mutate the bindings in the VAO.
+    VertexBufferHandle vbwo = rp->gl.vertexBufferWithObjects;
+    if (UTILS_UNLIKELY(vbwo)) {
+        const GLVertexBuffer const* vb = handle_cast<GLVertexBuffer*>(vbwo);
+        if (vb->bufferObjectsVersion != rp->gl.vertexBufferVersion) {
```
pixelflinger, line 3154: wouldn't it be *slightly* better to swap the sides of the `!=`, i.e. "if the rp is out of date".

pixelflinger (commented):
looks pretty good!

bejado:
Do the Metal/Vulkan driver need stubs for the new driver API functions?

pixelflinger:
> Do the Metal/Vulkan driver need stubs for the new driver API functions?

I think stuff like that is why it's still a draft :-)

filament/include/filament/BufferObject.h
prideout: Yes I love this idea.  Perhaps it could be called `bindingTarget`.

I suppose we could even support two of these for now: STATIC_VERTEX and DYNAMIC_VERTEX.  However I'm really not sure if GL's STATIC vs DYNAMIC hint actually does anything...

filament/backend/src/opengl/OpenGLDriver.cpp
```diff
@@ -1626,6 +1669,28 @@ void OpenGLDriver::updateVertexBuffer(Handle<HwVertexBuffer> vbh,
...
+    GLBufferObject* bo = handle_cast<GLBufferObject *>(boh);
+
+    // If the specified VBO handle is different from what's already in the slot, then update the
+    // slot and bump the cyclical version number. Dependent VAOs use the version number to detect
+    // when they should be updated.
+    static constexpr uint32_t kMaxVer = std::numeric_limits<decltype(vb->bufferObjectsVersion)>::max();
+    if (vb->gl.buffers[index] != bo->gl.id) {
+        vb->gl.buffers[index] = bo->gl.id;
+        const uint32_t version = vb->bufferObjectsVersion;
+        vb->bufferObjectsVersion = (version + 1) % kMaxVer;
```
pixelflinger, line 1688: kVersionMax or kMaxVersion -- we try to not abbreviate.

filament/backend/src/opengl/OpenGLDriver.cpp
```diff
@@ -757,6 +779,15 @@ void OpenGLDriver::importTextureR(Handle<HwTexture> th, intptr_t id,
     CHECK_GL_ERROR(utils::slog.e)
 }
 
+void OpenGLDriver::updateRenderPrimitiveBuffer(GLRenderPrimitive* rp, GLVertexBuffer const* vb) {
+
+    // TODO: This will bind the VAO, then bind a new VBO to the appropriate slot. The implementation
+    // of this method will be similar to setRenderPrimitiveBuffer() except that it will not bother
+    // updating the index buffer.
+
+    rp->gl.vertexBufferVersion = vb->bufferObjectsVersion;
```
pixelflinger, line 788: maybe add a comment that this is called from draw() and must be as efficient as possible.

bejado (approved):
LGTM, Mathias to approve OpenGL bits

filament/backend/src/opengl/OpenGLDriver.cpp
```diff
@@ -757,6 +780,57 @@ void OpenGLDriver::importTextureR(Handle<HwTexture> th, intptr_t id,
...
+                        getComponentType(attribute.type),
+                        getNormalization(attribute.flags & Attribute::FLAG_NORMALIZED),
+                        attribute.stride,
+                        (void*) uintptr_t(attribute.offset));
+            }
+
+            gl.enableVertexAttribArray(GLuint(i));
+        } else {
+
+            // In some OpenGL implementations, we must supply a properly-typed placeholder for
```
pixelflinger, line 821: `vb->attributes.size()` is a constant right? -- so that means we will generate like N `glVertexAttrib` calls? 
So can we make those calls part of the tracked state?

filament/backend/src/vulkan/VulkanHandles.cpp
```diff
@@ -403,18 +403,18 @@ bool VulkanRenderTarget::invalidate() {
...
         uint32_t size = 0;
         for (auto const& item : attributes) {
             if (item.buffer == bufferIndex) {
-              uint32_t end = item.offset + elementCount * item.stride;
+                uint32_t end = item.offset + elementCount * item.stride;
                 size = std::max(size, end);
             }
         }
-        buffers.emplace_back(new VulkanBuffer(context, stagePool, disposer, this,
+        buffers.push_back(std::make_shared<VulkanBuffer>(context, stagePool, disposer, this,
```
pixelflinger, line 417: One thing I don't like about `shared_ptr` is that it's unclear where and when the destructor is called, which can be a problem if the destructor has side-effects or in a multithreaded environment.

I don't know if there would be a problem here (I'm not familiar with the vulkan code enough), but I want to call it out, so we check that all the places where a reference can drop are safe for the dtor to be called (e.g. locks are not held --or-- always held, etc...).

pixelflinger (changes requested):

filament/backend/src/vulkan/VulkanHandles.cpp
prideout: Yes I was hesitant about this but I felt it was the right thing to use here, since the ownership truly is shared between two objects.  I thought about it and the destructor will be invoked at the correct time.

filament/backend/src/opengl/OpenGLDriver.cpp
prideout: Yes it's a fixed size.  This is code is actually unchanged from before (see `setRenderPrimitiveBuffer`), it's just factored into a new function so that it can be called when the VAO needs to be mutated (which is rare).

filament/backend/include/backend/DriverEnums.h
```diff
@@ -263,6 +263,11 @@ enum class ElementType : uint8_t {
     HALF4,
 };
 
+//! Buffer object binding type
+enum class BufferObjectBinding : uint8_t {
```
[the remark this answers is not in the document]
prideout: Done!

filament/backend/src/opengl/OpenGLDriver.cpp
pixelflinger: ah... okay then.

pixelflinger (approved):

PR #3686, opened 2021-03-23
Move Camera settings from gltf_viewer to SimpleViewer.

prideout:
This allows users to play with DoF sliders on Android.

Screen Shot 2021-03-23 at 11 50 50 AM


libs/viewer/src/SimpleViewer.cpp
```diff
@@ -658,6 +659,56 @@ void SimpleViewer::updateUserInterface() {
         ImGui::Unindent();
     }
 
+    if (ImGui::CollapsingHeader("Camera")) {
+        ImGui::Indent();
+        //ImGui::SliderFloat("Focal length (mm)", &FilamentApp::get().getCameraFocalLength(), 16.0f, 90.0f);
```
romainguy, line 664: Why don't we keep the focal length and exposure parameters? They are super useful

prideout: Doh!  Totally an accident.  Did this too fast.

pixelflinger (approved):

libs/viewer/src/SimpleViewer.cpp
prideout: Fixed this, thanks for catching it.

PR #3688, opened 2021-03-23
DoF: support full-resolution DoF

pixelflinger:
This is mostly used for debugging and having somewhat of a reference.
Change to the shader are minimal because all CoC values were always
treated as full-resolution. We just need to replace the downsampling
shader by one that just calculates the CoC. And we need to introduce
a CoC scale factor to compute the LOD to use (this was hardcoded
before).

On the C++ side, it's mostly a matter of replacing hardcoded buffer size
scales by a runtime or compile time variable.

### DIFF
--- a/filament/src/PostProcessManager.cpp
+++ b/filament/src/PostProcessManager.cpp
@@ -209,6 +209,7 @@ static const MaterialInfo sMaterialList[] = {
         { "fxaa",                  MATERIAL(FXAA) },
         { "taa",                   MATERIAL(TAA) },
         { "dofDownsample",         MATERIAL(DOFDOWNSAMPLE) },
+        { "dofCoc",                MATERIAL(DOFCOC) },
         { "dofMipmap",             MATERIAL(DOFMIPMAP) },
         { "dofTiles",              MATERIAL(DOFTILES) },
         { "dofDilate",             MATERIAL(DOFDILATE) },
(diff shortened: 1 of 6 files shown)


filament/src/PostProcessManager.cpp
```diff
@@ -915,12 +916,17 @@ FrameGraphId<FrameGraphTexture> PostProcessManager::dof(FrameGraph& fg,
     auto depth = blackboard.get<FrameGraphTexture>("depth");
     assert_invariant(depth);
 
-    // the downsampled target is multiple of 8, so we can have 4 clean mipmap levels
-    constexpr const uint32_t maxMipLevels = 4u;
+    /*
+     * dofResolution is used (at compile time for now) to chose between full- or quarter-resolution
+     * for the DoF calculations. Set to [1] for full resolution or [2] for quarter-resolution.
+     */
+    const uint32_t dofResolution = 2;
```
prideout, line 923: If this were a #define then you could use it to swap between dofCoc and dofDownsample in sMaterialList, which would allow the binary size to stay as small as possible.

prideout (approved):

romainguy (approved):

filament/src/PostProcessManager.cpp
pixelflinger: True -- but I was thinking of possibly making it a runtime option, e.g. for a photo/very high quality mode.

PR #3689, opened 2021-03-23
fix JNI setImage and friends buffer size calculation

### LINKED ISSUE #3685: filament-android/src/main/cpp/Texture.cpp nSetImage sizeInBytes calculate wrong when set sub image
Hi,guys

when I try update Texture's subImage with
`void com.google.android.filament.Texture.setImage(@NonNull Engine engine, @IntRange(from=0) int level, @IntRange(from=0) int xoffset, @IntRange(from=0) int yoffset, @IntRange(from=0) int width, @IntRange(from=0) int height, @NonNull PixelBufferDescriptor buffer)`

In my case,important arguments are:
image

But I always get
`java.nio.BufferOverflowException`
I look into `filament-android/src/main/cpp/Texture.cpp` in line 184,I found getTextureDataSize not pass height to calculate need bytes , getTextureDataSize just use whole texture's height instead... 

so is it means if we try to set sub image,because sub image's height maybe less than texture height,code in `line 188` will always `return -1`? 

 
Best Regards

pixelflinger:
- setImage() used the width of the texture for calculating the required
  user buffer size, but this was too large when only updating a
  portion of the texture. It would result in a BufferOverFlowException.

- setImage3D() had the opposite problem where it didn't take into
  account the depth of the texture.

- the offset in the user buffer was called "bottom" on the C++ JNI side,
  but it is called "top" on the java and filament side.

Fixes #3685

### DIFF
--- a/android/filament-android/src/main/cpp/Texture.cpp
+++ b/android/filament-android/src/main/cpp/Texture.cpp
@@ -34,12 +34,13 @@
 using namespace filament;
 using namespace backend;
 
-static size_t getTextureDataSize(const Texture *texture, size_t level,
-        Texture::Format format, Texture::Type type, size_t stride, size_t alignment) {
+static size_t getTextureDataSize(const Texture *texture,
+        size_t level, Texture::Format format, Texture::Type type,
+        size_t stride, size_t height, size_t alignment) {
     // Zero stride implies tight row-to-row packing.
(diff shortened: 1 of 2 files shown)


romainguy (approved):

prideout (approved):
O_O

bejado (approved):

PR #3691, opened 2021-03-24
Experiment: generate AST for Settings.h

prideout:
This generates a JSON file that has 3.7 million lines, so I think
we'd want to parse it with a fairly efficient JSON parser!

The CMake stuff for this is pretty hacky but I wasn't able to do it
any other way.

### DIFF
--- a/libs/viewer/CMakeLists.txt
+++ b/libs/viewer/CMakeLists.txt
@@ -32,13 +32,33 @@ target_link_libraries(${TARGET} PUBLIC imgui filament gltfio_core filagui jsmn c
 target_include_directories(${TARGET} PUBLIC ${PUBLIC_HDR_DIR})
 
 # ==================================================================================================
+# Generate AST (experimental)
+# ==================================================================================================
+
+set(filament_includes "-I$<JOIN:$<TARGET_PROPERTY:filament,INCLUDE_DIRECTORIES>,;-I>")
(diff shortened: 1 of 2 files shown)


romainguy:
3.7 million lines from Settings.h? Wth. Is this because of includes?

prideout:
> 3.7 million lines from Settings.h? Wth. Is this because of includes?

Yup it includes almost the entire Filament API.  (also, I should remove `<string>` from there)

Crazy idea:

To make it much simpler we could create a `Settings.h` in our public API, and move all our various options structs into that.  e.g. `View::BloomOptions` would be an alias for `Settings::BloomOptions`.

romainguy:
Maybe we should just parse it and see how fast/slow it is before worrying about optimizations. I know there are good, fast JSON parsers out there.

romainguy:
And it's not something we'll do at runtime, it can't be slower than generating our env maps through cmgen :p

prideout:
> Maybe we should just parse it and see how fast/slow it is before worrying about optimizations. I know there are good, fast JSON parsers out there.

Yeah.  We already use jsmn for the actual serialization, which is pretty damn fast, so we could use that.  Although, it's a C/C++ library and I'm not sure if C/C++ is the right tool for codegen...  On the other hand C++ is the lingua franca...

pixelflinger:
It's an opportunity to re-double check that we are not including things that we don't need in public headers and that we're fwd declaring everything we can.

PR #3693, opened 2021-03-24
Fix Metal bug when setGeometryAt is called multiple times

### LINKED ISSUE #3687: MTLVertexDescriptor error with morphing
When I try to run gltf_viewer with Horse.glb (link is below) I see the following Metal error:

```
-[MTLVertexDescriptorInternal newSerializedDescriptor]:726: failed assertion `Serialized Descriptor Creation
None of the attributes set bufferIndex to 13, but MTLVertexDescriptor set buffer layout[13].stride(12).
```

This might be something wrong in gltfio but if so it's not obvious; OpenGL and Vulkan look okay.

https://github.com/mrdoob/three.js/blob/dev/examples/models/gltf/Horse.glb

bejado:
Fixes #3687


prideout (approved):
🙏

pixelflinger (approved):

Issue #3692, opened 2021-03-24
Formatting improvement for detailed description of BloomOptions


jwwalker:
In `filament/include/filament/View.h`, the detailed description of `struct BloomOptions` lists descriptions of each member of the structure.  When processed by doxygen, these all run together in a single paragraph, which becomes quite hard to read.  The simple fix is to add blank lines in the source to become paragraph breaks in the documentation.

romainguy:
@bejado @pixelflinger

PR #3694, opened 2021-03-24
attempt to improve doxygen formatting

pixelflinger:
fixes: #3692


bejado:
Maybe it would be better if we moved each field's description into the struct itself, like how we do for `ShadowOptions`.

PR #3695, opened 2021-03-24
Minor Driver API simplification.

prideout:
As far as I can tell, there is no need for the `enabledAttributes`
argument in `setRenderPrimitiveBuffer`. This helps pave the way for
the upcoming BufferObject API.


bejado (approved):

prideout:
@pixelflinger  you okay with this one?

pixelflinger (approved):
Yeah, that looks okay.  I wonder why I did it this way originally.

Issue #2856, opened 2020-07-23
The skinning animation is broken in the latest version of Filament


cx20:
**Describe the bug**
I updated the Filament sample from `v1.8.0` to the [latest version](https://github.com/google/filament/actions/runs/179130980).
However, in the latest version of Filament, the model of skinning animation appears broken.

[Filament latest + RiggedFigure.gltf result:](https://cx20.github.io/gltf-test/examples/filament/index.html?model=RiggedFigure&scale=1)
![image](https://user-images.githubusercontent.com/1833979/88278572-92459200-cd1d-11ea-84a0-e700f65d3111.png)

**To Reproduce**
Filament latest + RiggedFigure.gltf result:
https://cx20.github.io/gltf-test/examples/filament/index.html?model=RiggedFigure&scale=1

**Expected behavior**
Filament v1.8.0+ RiggedFigure.gltf result:
![image](https://user-images.githubusercontent.com/1833979/88279547-3845cc00-cd1f-11ea-8f76-20a0cf32afba.png)

**Desktop (please complete the following information):**
 - OS: Windows 10 + Chrome 84
 - GPU: Intel UHD Graphics 617
 - Backend: WebGL/ANGLE (Intel(R) UHD Graphics 617 Direct3D11 vs_5_0 ps_5_0)

romainguy:
We haven't made changes to skinning related code recently. We did update cgltf though. @prideout any ideas?

romainguy:
Where can we get this asset? Rigging works fine in other gltf files.

prideout:
I have reproduced this, not sure what's going on yet.

prideout:
This only reproduces with our WASM build, other platforms are fine.

cx20:
@romainguy The assets are available from the Khronos glTF Sample Models repository.
https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/2.0#more-complex-models
|Model|Screenshot|
|:-------:|:------------:|
|[Rigged Figure](https://github.com/KhronosGroup/glTF-Sample-Models/blob/master/2.0/RiggedFigure)|![](https://raw.githubusercontent.com/KhronosGroup/glTF-Sample-Models/master/2.0/RiggedFigure/screenshot/screenshot.gif)|

[...]

PR #3698, opened 2021-03-24
Remove old emscripten workaround from GL backend.

prideout:
This was put in as a workaround for #2856. I made a fix to emscripten
itself (yay open source) in July. So, this workaround is no longer
necessary.

Verified with the rigged figure model.

https://github.com/emscripten-core/emscripten/pull/11742


romainguy (approved):

prideout:
When I removed this workaround, I failed to double-check the emsdk version used by GitHub Actions.  Resolved via https://github.com/google/filament/pull/3763.

PR #3701, opened 2021-03-25
move the focus distance from DofOptions to Camera

pixelflinger:
The DofOption parameter is now deprecated, but will still work if the
focus distance is not set on Camera.

### DIFF
--- a/filament/include/filament/Camera.h
+++ b/filament/include/filament/Camera.h
@@ -191,12 +191,13 @@ public:
 
     /** Sets the projection matrix from the focal length.
      *
-     * @param focalLength lens's focal length in millimeters. \p focalLength > 0.
+     * @param focalLengthInMillimeters lens's focal length in millimeters. \p focalLength > 0.
      * @param aspect      aspect ratio \f$ \frac{width}{height} \f$. \p aspect > 0.
      * @param near        distance in world units from the camera to the near plane. \p near > 0.
      * @param far         distance in world units from the camera to the far plane. \p far > \p near.
      */
-    void setLensProjection(double focalLength, double aspect, double near, double far) noexcept;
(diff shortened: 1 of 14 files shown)


pixelflinger:
What do you guys think of this change? I think it's more logical this way.

filament/include/filament/Camera.h
```diff
@@ -433,6 +434,19 @@ class UTILS_PUBLIC Camera : public FilamentAPI {
     //! returns this camera's sensitivity in ISO
     float getSensitivity() const noexcept;
 
+    //! returns the focal length in meters [m] for a 35mm camera
+    double getFocalLength() const noexcept;
+
+    /**
+     * Sets the camera focus distance. This is used by the Depth-of-field PostProcessing effect.
+     * @param distance Distnace from the camera to the plane of focus in world units.
```
bejado, line 442:
```suggestion
     * @param distance Distance from the camera to the plane of focus in world units.
```

bejado (approved):

libs/viewer/src/Settings.cpp
```diff
@@ -1171,6 +1175,7 @@ static std::ostream& operator<<(std::ostream& out, const ViewerOptions& in) {
         << "\"skyboxEnabled\": " << to_string(in.skyboxEnabled) << ",\n"
         << "\"backgroundColor\": " << (in.backgroundColor) << ",\n"
         << "\"cameraFocalLength\": " << (in.cameraFocalLength) << "\n"
```
prideout, line 1177: Add a trailing comma to cameraFocalLength.  I really need to make this automated.

prideout (changes requested):
I don't see `filament.d.ts` in here, I think that's the only missing piece (plus the comma thingie).

PR #3716, opened 2021-03-29
Remote UI now handles keyboard input.

prideout:
This lets you ctrl+click on a slider and type numbers.


bejado (approved):

libs/viewer/src/SimpleViewer.cpp
```diff
@@ -418,34 +418,78 @@ void SimpleViewer::applyAnimation(double currentTime) {
...
-void SimpleViewer::renderUserInterface(float timeStepInSeconds, View* guiView, float pixelRatio,
-        float mouseX, float mouseY, bool mouseButton, float mouseWheelY) {
+void SimpleViewer::renderUserInterface(float timeStepInSeconds, View* guiView, float pixelRatio) {
     if (mImGuiHelper == nullptr) {
         mImGuiHelper = new ImGuiHelper(mEngine, guiView, "");
 
+        auto& io = ImGui::GetIO();
+
+        // The following table assumes that keyCode arises from a web keydown event.
+        io.KeyMap[ImGuiKey_Tab] = 9;
```
romainguy, line 428: Could those keycodes be exposed as constants in the header file? What if we want to invoke the key handles from C++/Java? It seems the keycodes are just regular ASCII/ANSI codes so maybe document this in the key handler APIs.

libs/viewer/src/SimpleViewer.cpp
```diff
@@ -418,34 +418,78 @@ void SimpleViewer::applyAnimation(double currentTime) {
...
+        io.MouseWheel += mouseWheelY;
+        io.MouseDown[0] = mouseButton != 0;
+        io.MouseDown[1] = false;
+        io.MouseDown[2] = false;
+        io.KeyCtrl = control;
+    }
+}
+
+void SimpleViewer::keyDownEvent(int keyCode) {
+    if (mImGuiHelper && keyCode < 512) {
```
romainguy, line 476: Why 512?

libs/viewer/src/SimpleViewer.cpp
```diff
@@ -418,34 +418,78 @@ void SimpleViewer::applyAnimation(double currentTime) {
...
+        ImGui::GetIO().KeysDown[keyCode] = true;
+    }
+}
+
+void SimpleViewer::keyUpEvent(int keyCode) {
+    if (mImGuiHelper && keyCode < 512) {
```
romainguy, line 482: Same here

romainguy (approved):

romainguy (approved):

libs/viewer/src/SimpleViewer.cpp
prideout: It's a safety thing since this is a fixed-size C array.  Good catch though, replaced the magic number with IM_ARRAYSIZE.

libs/viewer/src/SimpleViewer.cpp
prideout: Yeah they're just ANSI codes. not really web-specific like the comment implies.  To address this I changed the wording and added a comment to the header file.

pixelflinger (approved):

PR #3718, opened 2021-03-30
DoF: use only a single lowres source for both fg and bg

pixelflinger:
This PR also includes support for swizzling in the framegraph, which is needed for the DoF change.


romainguy (approved):

pixelflinger:
There are other problems that need to be fixed / understood before we do this.

PR #3719, opened 2021-03-30
Fix VSM with MSAA on Android

bejado:
When using >1 MSAA samples, VSM renders into a multisampled render tarrget that resolves into a single-sampled texture. On Android, we attempt to use `GL_EXT_multisampled_render_to_texture`, however this does not support 2D texture arrays.


romainguy (approved):

pixelflinger (approved):
good one!

PR #3728, opened 2021-03-31
Revert misguided attempt to fix bloom with glTF

romainguy:


romainguy:
@cx20 FYI

prideout (approved):

libs/gltfio/src/AssetLoader.cpp
```diff
@@ -1159,7 +1155,6 @@ void FAssetLoader::addTextureBinding(MaterialInstance* materialInstance, const c
     }
 
     mResult->mTextureSlots.push_back({
-        .texture = srcTexture,
```
prideout: Is removal of this line intentional?

romainguy: No it wasn't, I already uploaded a revert of this

PR #3732, opened 2021-03-31
Fix Metal texture swizzling

bejado:
Metal textures can't have the `.swizzle` property set if they're used as render targets. So instead, we create a swizzled "texture view" which is used when binding the texture for sampling.

### DIFF
--- a/filament/backend/src/metal/MetalHandles.h
+++ b/filament/backend/src/metal/MetalHandles.h
@@ -183,6 +183,12 @@ struct MetalTexture : public HwTexture {
     MetalContext& context;
     MetalExternalImage externalImage;
     id<MTLTexture> texture = nil;
+
+    // If non-nil, a swizzled texture view to use instead of "texture".
+    // Filament swizzling only affects texture reads, so this should not be used when the texture is
+    // bound as a render target attachment.
+    id<MTLTexture> textureView = nil;
+
     uint8_t bytesPerElement; // The number of bytes per pixel, or block (for compressed texture formats).
     uint8_t blockWidth; // The number of horizontal pixels per block (only for compressed texture formats).
     uint8_t blockHeight; // The number of vertical pixels per block (only for compressed texture formats).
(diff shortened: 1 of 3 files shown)


filament/backend/src/metal/MetalHandles.h
```diff
@@ -183,6 +183,12 @@ struct MetalTexture : public HwTexture {
     MetalContext& context;
     MetalExternalImage externalImage;
     id<MTLTexture> texture = nil;
+
+    // If non-nil, a swizzled texture view to use instead of "texture".
+    // Filament swizzling only affects texture reads, so this should not be used when the texture is
+    // bound as a render target attachment.
+    id<MTLTexture> textureView = nil;
```
prideout, line 190: Maybe this could be named this be `swizzledTextureView`?  NBD.

filament/backend/src/metal/MetalHandles.mm
```diff
@@ -525,6 +516,30 @@ static MTLPixelFormat decidePixelFormat(id<MTLDevice> device, TextureFormat form
...
+    // If swizzling is set, set up a swizzled texture view that we'll use when sampling this texture.
+    const bool isDefaultSwizzle =
+            r == TextureSwizzle::CHANNEL_0 &&
+            g == TextureSwizzle::CHANNEL_1 &&
+            b == TextureSwizzle::CHANNEL_2 &&
+            a == TextureSwizzle::CHANNEL_3;
+    // If texture is nil, then it must be a SAMPLER_EXTERNAL texture. We'll ignore this case for now.
+    // TODO: implement swizzling for external textures.
+    if (!isDefaultSwizzle && texture) {
+        if (@available(macOS 10.15, iOS 13, *)) {
```
prideout, line 529: Maybe call `self->isTextureSwizzleSupported` here?  This ObjC stuff looks like a Harry Potter incantation to me.  :P

prideout (approved):

pixelflinger (approved):

filament/backend/src/metal/MetalHandles.mm
bejado: So the compiler issues warnings when we use an Apple API newer than our iOS deployment target (iOS 11). The `@available` guard silences this warnings. The compiler isn't smart enough to propagate the check if we moved the `@available` check into a helper function, unfortunately.

Here's more information on [@available](https://clang.llvm.org/docs/LanguageExtensions.html#objective-c-available).

PR #3734, opened 2021-04-01
DoF: add many user settings

pixelflinger:
- native/half resolution
- gather kernel ring counts
- circle-of-confusion radius clamps

### DIFF
--- a/filament/CMakeLists.txt
+++ b/filament/CMakeLists.txt
@@ -191,6 +191,11 @@ list(APPEND SRCS ${RESGEN_SOURCE})
 # Configuration
 # ==================================================================================================
 
+# whether we're building for mobile -- this can affect some default quality settings
+if (IS_MOBILE_TARGET)
+    add_definitions(-DFILAMENT_TARGET_MOBILE=1)
+endif()
+
 # Size of the DFG lookup table
 if (NOT DFG_LUT_SIZE)
     if (IS_MOBILE_TARGET)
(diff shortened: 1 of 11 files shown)


romainguy (approved):

web/filament-js/jsbindings.cpp
```diff
@@ -369,6 +369,12 @@ value_object<filament::View::DepthOfFieldOptions>("View$DepthOfFieldOptions")
     .field("maxApertureDiameter", &filament::View::DepthOfFieldOptions::maxApertureDiameter)
     .field("enabled", &filament::View::DepthOfFieldOptions::enabled)
     .field("filter", &filament::View::DepthOfFieldOptions::filter);
+    .field("nativeResolution", &filament::View::DepthOfFieldOptions::nativeResolution);
+    .field("foregroundRingCount", &filament::View::DepthOfFieldOptions::foregroundRingCount);
+    .field("backgroundRingCount", &filament::View::DepthOfFieldOptions::backgroundRingCount);
+    .field("fastGatherRingCount", &filament::View::DepthOfFieldOptions::fastGatherRingCount);
+    .field("maxForegroundCOC", &filament::View::DepthOfFieldOptions::maxForegroundCOC);
+    .field("maxBackgroundCOC", &filament::View::DepthOfFieldOptions::maxBackgroundCOC);
```
prideout, line 377: This won't build, it is a daisy chain so does not need semicolons.

prideout (commented):
Don't forget `filament.d.ts`

filament/CMakeLists.txt
```diff
@@ -191,6 +191,11 @@ list(APPEND SRCS ${RESGEN_SOURCE})
 # Configuration
 # ==================================================================================================
 
+# whether we're building for mobile -- this can affect some default quality settings
+if (IS_MOBILE_TARGET)
+    add_definitions(-DTARGET_MOBILE=1)
```
bejado, line 196: Can you preface this definition with `FILAMENT_`?

bejado (approved):

filament/CMakeLists.txt
pixelflinger: good idea, I went with the same define we use in shaders, but really this one should've had the prefix too.

PR #3742, opened 2021-04-02
Vulkan: add swizzle support.

prideout:


bejado (approved):

pixelflinger (approved):

pixelflinger (changes requested):
I think you need to return "true" from isTextureSwizzleSupported too right?

prideout:
> I think you need to return "true" from isTextureSwizzleSupported too right?

Good catch, fixed.

PR #3749, opened 2021-04-02
Implement Metal BufferObjects

bejado:


filament/backend/src/metal/MetalHandles.h
```diff
@@ -112,12 +112,24 @@ class MetalSwapChain : public HwSwapChain {
...
+
+};
+
 struct MetalVertexBuffer : public HwVertexBuffer {
     MetalVertexBuffer(MetalContext& context, uint8_t bufferCount, uint8_t attributeCount,
             uint32_t vertexCount, AttributeArray const& attributes, bool bufferObjectsEnabled);
-    ~MetalVertexBuffer();
 
-    std::vector<MetalBuffer*> buffers;
+    std::vector<std::shared_ptr<MetalBuffer>> buffers;
```
romainguy, line 132: Do we really need shared_ptr here instead of unique_ptr? Is MetalBufferObject the only other place where those pointers are used? Can a MetalbufferObject outlive this vector? If not, it doesn't look like we need shared_ptr.

prideout: In the future, we might be able to simplify the Driver API and make it so that buffers are always owned by a HwBufferObject.  This would let us remove the shared ownership.

Vulkan is also using shared_ptr (for now) because each buffer in this list was created either by a HwVertexBuffer or by a HwBufferObject, and it lets you avoid the ownership bookkeeping.

prideout (approved):

filament/backend/src/metal/MetalHandles.h
bejado: Yeah, what Philip said. If we refactor Filament's VertexBuffer to always use BufferObjects under the hood (and I think we should), then we won't need the additional bookkeeping here.

PR #3762, opened 2021-04-06
fix output location in fragment shaders

pixelflinger:
It looks like the layout(location) is needed for output parameters of
the fragment shader in GLES. Adreno dirver didn't seem to care, but
Mali drivers do. 

This fixes shaders that use multiple out buffers in debug mode.

### DIFF
--- a/libs/filamat/src/shaders/CodeGenerator.cpp
+++ b/libs/filamat/src/shaders/CodeGenerator.cpp
@@ -232,7 +232,7 @@ utils::io::sstream& CodeGenerator::generateOutput(utils::io::sstream& out, Shade
     out << "\n#define FRAG_OUTPUT" << index << " " << name.c_str() << "\n";
     out << "\n#define FRAG_OUTPUT_AT" << index << " output_" << name.c_str() << "\n";
     out << "\n#define FRAG_OUTPUT_TYPE" << index << " " << typeString << "\n";
-    out << "LAYOUT_LOCATION(" << index << ") out " << typeString <<
+    out << "layout(location=" << index << ") out " << typeString <<
         " output_" << name.c_str() << ";\n";
 
     return out;


libs/filamat/src/shaders/CodeGenerator.cpp
```diff
@@ -232,7 +232,7 @@ utils::io::sstream& CodeGenerator::generateOutput(utils::io::sstream& out, Shade
     out << "\n#define FRAG_OUTPUT" << index << " " << name.c_str() << "\n";
     out << "\n#define FRAG_OUTPUT_AT" << index << " output_" << name.c_str() << "\n";
     out << "\n#define FRAG_OUTPUT_TYPE" << index << " " << typeString << "\n";
-    out << "LAYOUT_LOCATION(" << index << ") out " << typeString <<
```
romainguy, line 235: I'm not sure this is right, here's our #define in the shaders:
```
#if !defined(TARGET_MOBILE) || defined(TARGET_LANGUAGE_SPIRV)
#define LAYOUT_LOCATION(x) layout(location = x)
#else
#define LAYOUT_LOCATION(x)
#endif
```
Is `layout(location = x)` valid in GLES 3.0? If so, remove the defines in the shaders (`common_types.fs`).

romainguy (changes requested):

libs/filamat/src/shaders/CodeGenerator.cpp
pixelflinger: I assumed this was there for a reason?

I assumed this was needed for out variables of the fragment shader only?

romainguy: No I mean we need to keep the layout() directive sometimes, but your change effectively removes the conditional from the shaders.

pixelflinger: No but the question is are there times where we *must not* use it? And why? It seems the condition is "if it's GLES then don't use it". Except that breaks the `out` variables on Mali.
So I *assumed* this meant that `out` variable needed it always, so that's what I did.
Note that my change only affects `out` variables in the fragment shader.

romainguy: I checked the spec and it looks valid in GLES 3.0. This code predates our move to GitHub so I don't remember why it is this way.

pixelflinger: ok, I can try to remove them all and see what happens.

romainguy: Oh wait I think I know: the location is supported on the out parameter of the fragment, but not between vertex out/fragment in maybe. So it's prob safe to remove it here but not from the other places.

romainguy: For posterity, here is the initial commit message:

```
Layout qualifiers are valid on vertex input and fragment output

Filament does not bind nor query the location of the attributes. While
layout qualifiers are invalid on ES 3.0 on varyings, they are valid on
vertex inputs and fragment outputs. We therefore restore the use of
these qualifiers on vertex inputs/fragment outputs. Without this change,
any attribute except for position is misinterpreted.
```

PR #3763, opened 2021-04-06
Update emscripten in Google Actions.

prideout:
This now matches what is already specified in BUILDING.md


romainguy (approved):

prideout:
This will make it so that our release builds include the fix that I made to emscripten itself (it had a bug in its implementation of glVertexAttribIPointer, which affected skinning.)

https://github.com/emscripten-core/emscripten/pull/11742

PR #3772, opened 2021-04-07
Fix some mediump related issues

pixelflinger:
- DoF: don't convert gl_FragCoord to mediump when generating noise
- SSAO: add a bunch of missing highp qualifiers
- all matrices and UV should be highp
- Also fix some typo in comments.

### DIFF
--- a/shaders/src/getters.fs
+++ b/shaders/src/getters.fs
@@ -7,14 +7,14 @@ vec4 getColor() {
 
 #if defined(HAS_ATTRIBUTE_UV0)
 /** @public-api */
-vec2 getUV0() {
+highp vec2 getUV0() {
     return vertex_uv01.xy;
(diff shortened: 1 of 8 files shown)


shaders/src/getters.fs
```diff
@@ -7,14 +7,14 @@ vec4 getColor() {
 
 #if defined(HAS_ATTRIBUTE_UV0)
 /** @public-api */
-vec2 getUV0() {
+highp vec2 getUV0() {
```
prideout, line 10: This one is an actual bug!

prideout (approved):

romainguy (approved):

pixelflinger:
@prideout this change might cause failures in g3 wrt to the golden images.

PR #3775, opened 2021-04-07
VertexBuffer cleanup and simplification.

prideout:
This includes changes to OpenGL, Metal, and Vulkan backends.

At the backend level, vertex buffers are now always composed of buffer
objects. This lets us simplify the Driver API and some bookkeeping in
the backends.

This change also splits MAX_VERTEX_ATTRIBUTE_COUNT into two constants
because the maximum number of bound buffers is a separate concept from
the maximum number of attribute semantics (e.g. consider interleaving).
For now these two constants are set to the same value.

We also now store a byte count in HwBufferObject, which allows us to
remove the byte count from the Metal-specific handle, and to add some
asserts to debug builds to prevent size overflow.

### DIFF
--- a/filament/backend/src/metal/MetalHandles.mm
+++ b/filament/backend/src/metal/MetalHandles.mm
@@ -267,7 +267,7 @@ void MetalSwapChain::scheduleFrameCompletedCallback() {
 }
 
 MetalBufferObject::MetalBufferObject(MetalContext& context, uint32_t byteCount)
-        : byteCount(byteCount), buffer(std::make_shared<MetalBuffer>(context, byteCount)) {}
+        : HwBufferObject(byteCount), buffer(std::make_unique<MetalBuffer>(context, byteCount)) {}
 
 void MetalBufferObject::updateBuffer(void* data, size_t size, uint32_t byteOffset) {
     assert_invariant(byteOffset + size <= byteCount);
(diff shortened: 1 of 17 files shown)


filament/backend/src/opengl/OpenGLDriver.cpp
```diff
@@ -3229,11 +3211,9 @@ void OpenGLDriver::draw(PipelineState state, Handle<HwRenderPrimitive> rph) {
 
     // If necessary, mutate the bindings in the VAO.
     VertexBufferHandle vbwo = rp->gl.vertexBufferWithObjects;
-    if (UTILS_UNLIKELY(vbwo)) {
-        const GLVertexBuffer* vb = handle_cast<GLVertexBuffer*>(vbwo);
-        if (rp->gl.vertexBufferVersion != vb->bufferObjectsVersion) {
-            updateVertexArrayObject(rp, vb);
-        }
+    const GLVertexBuffer* vb = handle_cast<GLVertexBuffer*>(vbwo);
+    if (UTILS_UNLIKELY(rp->gl.vertexBufferVersion != vb->bufferObjectsVersion)) {
```
pixelflinger, line 3215: we're sure that `vb` can't be `nullptr` here? by construction?

pixelflinger (approved):
very nice!

filament/backend/src/opengl/OpenGLDriver.cpp
prideout: That's a good point, it's possible for Filament to create a render primitive and never set it up with a vertex buffer and index buffer.  However I would argue that calling draw() with an uninitialized render primitive is a mis-use of the Driver API.

Maybe this is a good candidate for assert_invariant.

filament/backend/src/metal/MetalHandles.mm
```diff
@@ -267,42 +267,18 @@ BufferDescriptor b(nullptr, 0u, [](void* buffer, size_t size, void* user) {
...
-            }
-        }
-
-        std::shared_ptr<MetalBuffer> buffer = nullptr;
-        if (size > 0) {
-            buffer = std::make_shared<MetalBuffer>(context, size);
-        }
-        buffers.push_back(buffer);
-    }
+    buffers.resize(bufferCount);
```
bejado, line 281: I think we can remove the call to `buffers.reserve` here.

bejado (approved):

filament/backend/src/opengl/OpenGLDriver.cpp
pixelflinger: No I think it make more sense to do nothing in that case. It's like drawing an empty draw call.

filament/backend/src/metal/MetalHandles.mm
prideout: Good catch, removed.

filament/backend/src/opengl/OpenGLDriver.cpp
prideout: Ok, added an 'if unlikely" for the empty draw call case.

PR #3780, opened 2021-04-07
rework the "debugCommand" feature of backend

pixelflinger:
- it was called on the main thread side, which didn't allow to,
  for instance, systrace commands on the backend thread side.

- moved to a begin/end callback model

- implemented basic logging/systracing for all backends in Driver.cpp
  (and removed this functionality for the metal driver).

### DIFF
--- a/filament/backend/include/private/backend/CommandStream.h
+++ b/filament/backend/include/private/backend/CommandStream.h
@@ -213,37 +213,64 @@ public:
 
 // ------------------------------------------------------------------------------------------------
 
-#if defined(NDEBUG)
-    #define DEBUG_COMMAND(methodName, ...)
-#else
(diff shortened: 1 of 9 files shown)


prideout (approved):

prideout:
Note that the ubuntu build failure looks like an actual error.

pixelflinger:
> Note that the ubuntu build failure looks like an actual error.

I didn't understand what it was?

I thought it was because of the #pragma -- which I just removed anyways.

prideout:
According to the logs, clang is having trouble with the default member initializer for mDispatcher, which I think is bogus but should be easy to work around.

https://stackoverflow.com/questions/43819314/default-member-initializer-needed-within-definition-of-enclosing-class-outside

bejado (approved):

PR #3787, opened 2021-04-09
better handle anisotropic filtering with various drivers

pixelflinger:
Some drivers claim to support anisotropic filtering but don't allow
setting it via glSamplerParameter. Until now we were disabling 
anisotropic filtering entirely on these drivers at compile time.

With this change, we should detect at runtime which drivers have this
issue and use a workaround -- instead of disabling anisotropic 
filtering entirely.

### DIFF
--- a/filament/backend/src/opengl/OpenGLContext.cpp
+++ b/filament/backend/src/opengl/OpenGLContext.cpp
@@ -70,7 +70,6 @@ OpenGLContext::OpenGLContext() noexcept {
             bugs.dont_use_timer_query = true;
         }
         if (strstr(renderer, "Mali-G")) {
-            bugs.disable_texture_filter_anisotropic = true;
             // note: We have verified that timer queries work well at least on some Mali-G.
         }
     } else if (strstr(renderer, "Intel")) {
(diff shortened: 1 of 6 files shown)


pixelflinger:
@prideout would you be able to check that this doesn't break webgl at runtime? I did change the logic for webgl, assuming that the issue was similar to other drivers.

romainguy (approved):

prideout:
Sure I will check...

prideout (approved):
WebGL looks fine.

filament/backend/src/opengl/OpenGLContext.cpp
```diff
@@ -147,8 +146,22 @@ OpenGLContext::OpenGLContext() noexcept {
...
 #ifdef GL_EXT_texture_filter_anisotropic
-    if (ext.EXT_texture_filter_anisotropic && !bugs.disable_texture_filter_anisotropic) {
+    if (ext.EXT_texture_filter_anisotropic) {
+        // make sure we don't have any error flag
+        while (glGetError() != GL_NO_ERROR) { }
+
+        // query max anisotropy
         glGetFloatv(GL_MAX_TEXTURE_MAX_ANISOTROPY_EXT, &gets.maxAnisotropy);
+
+        // check that we can actually set the anysotropy on the sampler
```
bejado, line 156:
```suggestion
        // check that we can actually set the anisotropy on the sampler
```

bejado (approved):

PR #3793, opened 2021-04-12
enable contact-shadows functionality in mobile gltf-viewer

pixelflinger:
contact-shadows needs to be enabled per object.

### DIFF
--- a/android/filament-utils-android/src/main/java/com/google/android/filament/utils/ModelViewer.kt
+++ b/android/filament-utils-android/src/main/java/com/google/android/filament/utils/ModelViewer.kt
@@ -267,9 +267,14 @@ class ModelViewer(val engine: Engine) : android.view.View.OnTouchListener {
     }
 
     private fun populateScene(asset: FilamentAsset) {
+        val rcm = engine.renderableManager
         var count = 0
         val popRenderables = {count = asset.popRenderables(readyRenderables); count != 0}
         while (popRenderables()) {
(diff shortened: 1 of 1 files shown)


prideout (approved):
Sure but maybe make this optional by adding `var enableContactShadows = true` similar to the other options like `recomputeBoundingBoxes`.

pixelflinger:
> Sure but maybe make this optional by adding `var enableContactShadows = true` similar to the other options like `recomputeBoundingBoxes`.

It's not needed because contact shadows also need to be enabled on the light, and that's what we control with the checkbox in the UI.

prideout:
> > Sure but maybe make this optional by adding `var enableContactShadows = true` similar to the other options like `recomputeBoundingBoxes`.
> 
> It's not needed because contact shadows also need to be enabled on the light, and that's what we control with the checkbox in the UI.

Ok sounds good.  (In that case I wonder why Filament's doesn't just do this automatically.)

pixelflinger:
> > > Sure but maybe make this optional by adding `var enableContactShadows = true` similar to the other options like `recomputeBoundingBoxes`.
> > 
> > 
> > It's not needed because contact shadows also need to be enabled on the light, and that's what we control with the checkbox in the UI.
> 
> Ok sounds good. (In that case I wonder why Filament's doesn't just do this automatically.)

This is so that you *can* enable contact shadow on a per-renderable basis, since it's expensive.

PR #3794, opened 2021-04-12
silently clamp MSAA sample count to maximum allowed

pixelflinger:
It's an error to ask for a higher MSAA sample count than supported by
the gl driver, and results in corrupted rendering.
We now always clamp the requested sample count to the maximum allowed 
but the gl driver.

includes some minor refactoring.


pixelflinger:
@prideout @bejado I think we need something similar for the vk and mtl backends.

prideout (approved):

PR #3795, opened 2021-04-12
sample-gltf-viewer: Prevent OOM for large models like Bistro.

prideout:


android/samples/sample-gltf-viewer/src/main/java/com/google/android/filament/gltf/MainActivity.kt
```diff
@@ -154,12 +156,27 @@ class MainActivity : Activity() {
...
-        message.buffer.get(zipfileBytes)
+        // To alleviate memory pressure, remove the old model before deflating the zip.
+        withContext(Dispatchers.Main) {
+            modelViewer.destroyModel()
+        }
 
+        // Large zip files should first be written to a file to prevent OOM.
+        // It is also crucial that we null out the message "buffer" field.
+        val zipStream = {
+            val file = File.createTempFile("incoming", "zip", this.cacheDir);
```
pixelflinger, line 167: should we unlink the temp file ourselves, just to make sure it goes away as soon as we don't need it?

pixelflinger (approved):

android/samples/sample-gltf-viewer/src/main/java/com/google/android/filament/gltf/MainActivity.kt
prideout: Sure, Java provides a convenient `deleteOnExit` method for this but I think deleting it ASAP would be better.  I'll do this.

PR #3796, opened 2021-04-12
allow the public RenderTarget API to use MRT

pixelflinger:


filament/src/RenderTarget.cpp
```diff
@@ -61,74 +64,93 @@ RenderTarget::Builder& RenderTarget::Builder::layer(AttachmentPoint pt, uint32_t
...
+        mAttachmentMask |= TargetBufferFlags::COLOR2;
+        setColorAttachment(mrt[2], COLOR2);
+    }
+    if (mAttachments[COLOR3].texture) {
+        mAttachmentMask |= TargetBufferFlags::COLOR3;
+        setColorAttachment(mrt[3], COLOR3);
     }
     if (mAttachments[DEPTH].texture) {
         mAttachmentMask |= TargetBufferFlags::DEPTH;
+        setColorAttachment(dinfo, DEPTH);
```
prideout, line 148: Rename the lambda maybe, since it's bue

prideout (approved):
Don't forget `filament.d.ts`

PR #3799, opened 2021-04-13
matdbg robustness improvements / cleanup.

prideout:


prideout:
Oh I forgot about the filamat dependency, I might need to back out some of the simplifications.

prideout:
> Oh I forgot about the filamat dependency, I might need to back out some of the simplifications.

Fixed.

bejado (approved):

filament/src/Material.cpp
```diff
@@ -313,6 +311,15 @@ FMaterial::FMaterial(FEngine& engine, const Material::Builder& builder)
...
 FMaterial::~FMaterial() noexcept {
+
+#if FILAMENT_ENABLE_MATDBG
+    // Unregister the material with matdbg.
+    matdbg::DebugServer* server = upcast(mEngine).debug.server;
+    if (UTILS_UNLIKELY(server)) {
+        server->removeMaterial(mDebuggerId);
+    }
+#endif
+
```
pixelflinger, line 322: it would be better to move this to `terminate()`. Anything that could fail shouldn't be in a dtor.

libs/matdbg/src/DebugServer.cpp
```diff
@@ -421,7 +421,7 @@ class WebSocketHandler : public CivetWebSocketHandler {
     }
 
     // Notify the JavaScript client that a new material package has been loaded.
-    void notify(const DebugServer::MaterialRecord& material) {
+    void addMaterial(const DebugServer::MaterialRecord& material) {
         if (mConnection) {
             char matid[9] = {};
             sprintf(matid, "%8.8x", material.key);
```
pixelflinger, line 427: snprintf would be safer

pixelflinger (changes requested):
move the removing of material to `terminate` instead of dtor.

filament/src/Material.cpp
prideout: Done.

PR #3808, opened 2021-04-14
Improvements to Renderer

pixelflinger:
- more aggressively skip lighting/shadowing code that won't be 
  needed by the current view

- Allow up to 64 KiB to be allocated from the render stream, up from
  1 KiB

- Renderer::renderStandaloneView() is a new method that can be used
  outside of beginFrame/endFrame on Views that have a RenderTarget
  associated. This can be used as a poorman's compute API.


android/filament-android/src/main/java/com/google/android/filament/Renderer.java
```diff
@@ -356,6 +356,48 @@ public void render(@NonNull View view) {
...
+     * {@link #endFrame}.</li>
+     *
+     * <li><code>renderStandaloneView()</code> must be called from the {@link Engine}'s main thread
+     * (or external synchronization must be provided). In particular, calls to
+     * <code>renderStandaloneView()</code> on different <code>Renderer</code> instances <b>must</b>
+     * be synchronized.</li>
+     *
+     * <li><code>renderStandaloneView()</code> performs potentially heavy computations and cannot be
+     * multi-threaded. However, internally, it is highly multi-threaded to both improve performance
+     * and mitigate he call's latency.</li>
```
bejado, line 387:
```suggestion
     * and mitigate the call's latency.</li>
```

filament/include/filament/Renderer.h
```diff
@@ -329,69 +259,44 @@ class UTILS_PUBLIC Renderer : public FilamentAPI {
...
      */
-    void readPixels(uint32_t xoffset, uint32_t yoffset, uint32_t width, uint32_t height,
+    void readPixels(RenderTarget* renderTarget,
+            uint32_t xoffset, uint32_t yoffset, uint32_t width, uint32_t height,
             backend::PixelBufferDescriptor&& buffer);
 
-
     /**
-     * Reads back the content of the provided RenderTarget.
+     * Render a standalone View into its associated RenderTarget
```
bejado, line 267: I think `renderStandaloneView`'s signature should come after `render`, since `render` is still the preferred method (if I'm understanding correctly). Also, I think we should better document what the differences between the two are. Why would someone call `renderStandaloneView` versus `render`?

filament/src/Renderer.cpp
```diff
@@ -921,10 +952,6 @@ bool FRenderer::beginFrame(FSwapChain* swapChain, uint64_t vsyncSteadyClockTimeN
     auto beginFrameInternal = [=]() {
         FEngine& engine = getEngine();
         FEngine::DriverApi& driver = engine.getDriverApi();
-
-        if (callback) {
-            driver.setFrameScheduledCallback(swapChain->getHwHandle(), callback, user);
```
bejado, line 926: We still need to support this somehow.

bejado (changes requested):

android/filament-android/src/main/java/com/google/android/filament/Renderer.java
```diff
@@ -356,6 +356,48 @@ public void render(@NonNull View view) {
         nRender(getNativeObject(), view.getNativeObject());
     }
 
+   /**
+     * Renders a standalone {@link View} into its associated <code>RenderTarget</code>.
+     *
+     * <p>
+     * The purpose of the <code>renderStandaloneView()</code> function is to generate render
+       commands which are asynchronously executed by the {@link Engine}'s render thread.
```
prideout, line 364: This is somewhat cryptic, end users don't know what "render commands" are.  Maybe we should instead say that this can be used for doing work that is not associated with any particular swap chain?  Also I think a `*` is missing.

prideout (commented):
Would it make sense to change our rendertarget mirror demo to use this new API?

I'm not saying we should to thatin this PR, I'm just curious if that's a good example use case, or if this API is really more about doing compute-style stuff like cmgen.

prideout:
Also your commit message isn't great since this introduces a new API.  :)   (makes it harder t owrite good release notes)

filament/src/Renderer.cpp
pixelflinger: But that API was marked deprecated. Do we need to keep it?

pixelflinger:
> Would it make sense to change our rendertarget mirror demo to use this new API?
> 
> I'm not saying we should to thatin this PR, I'm just curious if that's a good example use case, or if this API is really more about doing compute-style stuff like cmgen.

Yes I think it would make sense -- in this case though, both ways of doing this are fine.

android/filament-android/src/main/java/com/google/android/filament/Renderer.java
pixelflinger: This is just a copy of the language for render() -- which is also kind of cryptic :-)

pixelflinger: Okay I tried to remove the confusing language.

android/filament-android/src/main/java/com/google/android/filament/Renderer.java
```diff
@@ -331,31 +331,62 @@ public void endFrame() {
...
+     * (or external synchronization must be provided). In particular, calls to
+     * <code>renderStandaloneView()</code> on different <code>Renderer</code> instances <b>must</b>
+     * be synchronized.</li>
+     *
+     * <li><code>renderStandaloneView()</code> performs potentially heavy computations and cannot be
+     * multi-threaded. However, internally, it is highly multi-threaded to both improve performance
+     * and mitigate the call's latency.</li>
+     *
+     * <li><code>renderStandaloneView()</code> is typically called once per frame
+     * (but not necessarily).</li>
```
prideout, line 380: Note sure if this particular bullet should be here?

filament/src/Renderer.cpp
bejado: I'm fine with getting rid of it, but I think that would constitute an API breaking change. I'm not sure if any Google clients are still using it, but it should be easy to migrate them.

prideout (approved):

RELEASE_NOTES.md
```diff
@@ -7,19 +7,21 @@ A new header is inserted each time a *tag* is created.
 
 ## v1.9.21
 
+- engine: Add new `Renderer::renderStandaloneView()` API.
```
bejado, line 10: Please add this under the `## Next release (main branch)` header. It will go out with v1.9.22.

pixelflinger: wow good catch!

PR #3816, opened 2021-04-16
Disable VSM MSAA on Adreno devices

bejado:
On Adreno devices, a driver bug prevents bliting from a renderbuffer into certain layers of a texture array, which is precicely what we do for VSM MSAA. For now, we'll just disable support for MSAA and simply fallback to normal, non-MSAA rendering.

It'd be nice to not have to disable this for all Adreno devices, but for now I can only assume the bug is present on all of them. 😕 

I also attempted to clarify the logic inside `OpenGLDriver::framebufferTexture`.

### DIFF
--- a/filament/backend/src/opengl/OpenGLDriver.cpp
+++ b/filament/backend/src/opengl/OpenGLDriver.cpp
@@ -937,36 +937,17 @@ void OpenGLDriver::framebufferTexture(backend::TargetBufferInfo const& binfo,
         attachmentTypeNotSupportedByMSRTT = true;
     }
 
-    // The following 4 techniques are different mechanisms to provide storage to the FBO.
-    // We'll use the first technique that is sufficient.
-    // nonMSAA:         MSAA was not requested, or we're rendering to an MS texture which
-    //                  doesn't require any special setup here
-    // useMSRTT:        use the EXT_multisampled_render_to_texture extension
-    // renderDirect:    texture is not sampleable, render directly into a renderbuffer
-    // emulateMSRTT:    prior 3 techniques are unsuitable, emulate MSRTT extension support
-
-    const bool nonMSAA =
-            rt->gl.samples <= 1 ||
-            (rt->gl.samples > 1 && t->samples > 1 && gl.features.multisample_texture);
-
-    // EXT_multisampled_render_to_texture only support GL_COLOR_ATTACHMENT0
-    const bool useMSRTT =
-            !attachmentTypeNotSupportedByMSRTT && (t->depth <= 1) &&
-            ((gl.ext.EXT_multisampled_render_to_texture && attachment == GL_COLOR_ATTACHMENT0) ||
-             gl.ext.EXT_multisampled_render_to_texture2);
-
-    const bool renderDirect = !any(t->usage & TextureUsage::SAMPLEABLE) && t->samples > 1;
-
-    const bool emulateMSRTT = !nonMSAA && !useMSRTT && !renderDirect;
-
     // There's a bug with certain drivers preventing us from emulating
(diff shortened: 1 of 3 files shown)


filament/backend/src/opengl/OpenGLDriver.cpp
```diff
@@ -937,8 +937,36 @@ void OpenGLDriver::framebufferTexture(backend::TargetBufferInfo const& binfo,
         attachmentTypeNotSupportedByMSRTT = true;
     }
 
-    if (rt->gl.samples <= 1 ||
-        (rt->gl.samples > 1 && t->samples > 1 && gl.features.multisample_texture)) {
+    // The following 4 techniques are different mechanisms to provide storage to the FBO.
+    // We'll use the first technique that is sufficient.
+    // nonMSAA:         MSAA was not requested, or we're rendering to an MS texture which
```
pixelflinger, line 942: maybe instead of "nonMSAA" it should be more like "noAutomaticMsaaResolve"? i.e. We won't do an automatic MSAA resolve (because either we don't have MSAA at all, or the attachment itself is MS and the resolve will have to be custom). Either way, we're not handling it. 

*However* because MS textures are not supported in ES3, we just do nothing. Meaning we effectively disable MSAA. This is *wrong* if we wanted to support ES3.1 MS textures.

filament/backend/src/opengl/OpenGLDriver.cpp
```diff
@@ -937,8 +937,36 @@ void OpenGLDriver::framebufferTexture(backend::TargetBufferInfo const& binfo,
...
+    //                  doesn't require any special setup here
+    // useMSRTT:        use the EXT_multisampled_render_to_texture extension
```
pixelflinger, line 944: Similarly, this is more "useMsrttMsaaRsolve". i.e. MSAA resolve will be handled by MSRTT.

filament/backend/src/opengl/OpenGLDriver.cpp
```diff
@@ -937,8 +937,36 @@ void OpenGLDriver::framebufferTexture(backend::TargetBufferInfo const& binfo,
...
+    // renderDirect:    texture is not sampleable, render directly into a renderbuffer
```
pixelflinger, line 945: So this is similar to `nonMSAA`, except here, this is supported by ES3.0 because the texture is not sampleable, so we can use a renderbuffer (which, well, is not a texture). But the point here is that there won't be any automatic resolve either.

Maybe it would make sense to refactor this so that the first case is just "noMSAA" and this case handles both sampleable and non sampleable MS textures, but in the case of sampleable MS textures it would "do nothing", i.e. revert to no MSAA.

filament/backend/src/opengl/OpenGLDriver.cpp
```diff
@@ -937,8 +937,36 @@ void OpenGLDriver::framebufferTexture(backend::TargetBufferInfo const& binfo,
...
+    // emulateMSRTT:    prior 3 techniques are unsuitable, emulate MSRTT extension support
```
pixelflinger, line 946: Well, it's not that the 3 prior techniques are unsuitable -- it's more that we need an automatic resolve and that it can't be done by MSRTT.
And we need an automatic resolve because:
- the attachment **is not** MS
- the RenderTarget **is** MS

Which is our definition of "auto resolve" in our API.

pixelflinger (changes requested):
Juste rename `bugs.disable_sidecar_blit`, this gives the impression that all sidecar blits need to be disabled.

My other comments are just informative, you may want to clarify your own comments.

pixelflinger (approved):

PR #3818, opened 2021-04-16
gltfio: robustness improvements for Draco meshes.

prideout:
These changes are motivated by #3815.

See also:
    https://github.com/KhronosGroup/glTF-Blender-IO/issues/1326
    https://github.com/KhronosGroup/glTF-Blender-IO/issues/1345

### DIFF
--- a/libs/geometry/src/SurfaceOrientation.cpp
+++ b/libs/geometry/src/SurfaceOrientation.cpp
@@ -17,6 +17,7 @@
 #include <geometry/SurfaceOrientation.h>
 
 #include <utils/Panic.h>
+#include <utils/debug.h>
 
 #include <math/mat3.h>
 #include <math/norm.h>
@@ -226,6 +227,7 @@ SurfaceOrientation* OrientationBuilderImpl::buildWithUvs() {
     memset(tan2.data(), 0, sizeof(float3) * vertexCount);
     for (size_t a = 0; a < triangleCount; ++a) {
         uint3 tri = triangles16 ? uint3(triangles16[a]) : triangles32[a];
+        assert_invariant(tri.x < vertexCount && tri.y < vertexCount && tri.z < vertexCount);
         const float3& v1 = positions[tri.x];
         const float3& v2 = positions[tri.y];
         const float3& v3 = positions[tri.z];
@@ -334,6 +336,7 @@ SurfaceOrientation* OrientationBuilderImpl::buildWithFlatNormals() {
     float3* normals = new float3[vertexCount];
     for (size_t a = 0; a < triangleCount; ++a) {
         const uint3 tri = triangles16 ? uint3(triangles16[a]) : triangles32[a];
+        assert_invariant(tri.x < vertexCount && tri.y < vertexCount && tri.z < vertexCount);
         const float3 v1 = positions[tri.x];
         const float3 v2 = positions[tri.y];
         const float3 v3 = positions[tri.z];
--- a/libs/gltfio/src/DracoCache.cpp
+++ b/libs/gltfio/src/DracoCache.cpp
@@ -117,10 +117,10 @@ DracoMesh* DracoMesh::decode(const uint8_t* data, size_t dataSize) {
     return new DracoMesh(new DracoMeshDetails { std::move(meshStatus).value() });
 }
 
-void DracoMesh::getFaceIndices(cgltf_accessor* target) const {
+bool DracoMesh::getFaceIndices(cgltf_accessor* target) const {
     // Return early if we've already decompressed this data.
     if (target->buffer_view) {
-        return;
+        return true;
     }
 
     draco::Mesh* mesh = mDetails->mesh.get();
@@ -131,7 +131,7 @@ void DracoMesh::getFaceIndices(cgltf_accessor* target) const {
     if (target->count != count) {
         slog.e << "The glTF accessor wants " << target->count << " indices, "
                << "but the decoded Draco mesh has " <<  count << " indices." << io::endl;
-        return;
+        return false;
     }
 
     cgltf_buffer_view* view = new cgltf_buffer_view;
@@ -149,8 +149,9 @@ void DracoMesh::getFaceIndices(cgltf_accessor* target) const {
         case cgltf_component_type_r_8u: convertFaces<uint8_t>(target, mesh); break;
         default:
             slog.e << "Unexpected component type for Draco indices." << io::endl;
-            break;
+            return false;
     }
+    return true;
 }
 
 bool DracoMesh::getVertexAttributes(uint32_t attributeId, cgltf_accessor* target) const {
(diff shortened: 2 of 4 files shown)


libs/geometry/src/SurfaceOrientation.cpp
```diff
@@ -226,6 +227,7 @@ SurfaceOrientation* OrientationBuilderImpl::buildWithUvs() {
     memset(tan2.data(), 0, sizeof(float3) * vertexCount);
     for (size_t a = 0; a < triangleCount; ++a) {
         uint3 tri = triangles16 ? uint3(triangles16[a]) : triangles32[a];
+        assert_invariant(tri.x < vertexCount && tri.y < vertexCount && tri.z < vertexCount);
```
pixelflinger, line 230: So is this checking for possible bugs in our code?

If this assert can fail due to something the client does, this should use the `PANIC` stuff and the method should not be `noexcept`. 

If this code is not directly called by our client, but rather used by filament itself, then `assert_invariant` is fine -- it would just mean that the caller (i.e. us) did something wrong, in which case we need to make sure this doesn't happen.

libs/gltfio/src/DracoCache.cpp
```diff
@@ -117,10 +117,10 @@ DracoMesh* DracoMesh::decode(const uint8_t* data, size_t dataSize) {
     return new DracoMesh(new DracoMeshDetails { std::move(meshStatus).value() });
 }
 
-void DracoMesh::getFaceIndices(cgltf_accessor* target) const {
+bool DracoMesh::getFaceIndices(cgltf_accessor* target) const {
     // Return early if we've already decompressed this data.
     if (target->buffer_view) {
-        return;
+        return true;
```
pixelflinger, line 123: is a precondition violated here? Or is it legitimate to have a null buffer_view?

libs/gltfio/src/DracoCache.cpp
```diff
@@ -131,7 +131,7 @@ void DracoMesh::getFaceIndices(cgltf_accessor* target) const {
     if (target->count != count) {
         slog.e << "The glTF accessor wants " << target->count << " indices, "
                << "but the decoded Draco mesh has " <<  count << " indices." << io::endl;
-        return;
+        return false;
```
pixelflinger, line 134: If precondition is violated this should use the `ASSERT_PRECONDITION`.

I'm very uneasy about having logs in release builds. It should not be our decision to log stuff (i.e. this could be leaking user data). If this error is a precondition violation (i.e. user didn't set thing correctly), we should throw, i.e. use `ASSERT_PRECONDITION`.  

Because our clients may not want to use exceptions **but** they may also want to handle the error in that case (which I understand, but don't agree with), I think it would be fine to use `ASSERT_PRECONDITION_NON_FATAL`, which will throw in DEBUG (as it should) but return `false` in release (and gives us the ability to control this policy).

If this code has java bindings, I think the bindings should catch errors and convert them to java exceptions.

pixelflinger (changes requested):
Mainly I would like to avoid logging in release builds. logging is a response to an error, and that should be decided by the caller -- not us.
Maybe this PR is not the right one to rework this, so I'm fine with this change, if we circle back and clean this up later.

libs/geometry/src/SurfaceOrientation.cpp
prideout: This checks for badly formed model data.  This is a bounds check that I think should not occur in release builds since it's in a tight loop.  I just want to make it easier to catch this issue using a debug build without the help of ASAN.

The client directly calls this method, but since I didn't want to add a bounds check to release builds, I felt that an assert was the way to go.

libs/gltfio/src/DracoCache.cpp
prideout: This is not a precondition violation, it just checks if the data has already been decompressed.

libs/gltfio/src/DracoCache.cpp
prideout: This change does not add new logs, it simply returns false so that a higher layer can skip further processing and thus avoid the heap overflow that was reported by a user.

Regarding the existing issue of too many logs, we might need to do a sweep through the gltfio library in order to apply the policy that you're describing.

Writing a loader library for a fairly new file format like glTF makes it tricky to make robustness vs correctness decisions.   For minor spec violations, sometimes the pragmatic thing to do is for the loader library to try to recover rather than panic.

pixelflinger: Agreed. But maybe that means the logs should be in DEBUG only (and/or protected by a compile option). Also when we are trying to recover, I think a comment is useful (to disambiguate with a legit error/exception).

prideout:
> Mainly I would like to avoid logging in release builds. logging is a response to an error, and that should be decided by the caller -- not us.

This makes sense.  So...I have an idea (for the future):

If gltfio needs to be robust against certain types of badly formed data, maybe it can recover but still tell clients about it by using an "info log" string, similar to how OpenGL handles shader errors.  e.g. Instead of emitting to the log, ResourceLoader could provide a `getInfoLog()` query.

pixelflinger:
> > Mainly I would like to avoid logging in release builds. logging is a response to an error, and that should be decided by the caller -- not us.
> 
> This makes sense.  So...I have an idea (for the future):
> 
> If gltfio needs to be robust against certain types of badly formed data, maybe it can recover but still tell clients about it by using an "info log" string, similar to how OpenGL handles shader errors.  e.g. Instead of emitting to the log, ResourceLoader could provide a `getInfoLog()` query.
> 

Yes, I think it makes sense for gltfio to adopt a defensive stance. Which is different from the philosophy in filament.

But the question is, do we make all the checks in debug and release? If we are truly robust against ill formed files only in debug, then why not log only in debug too?

I'm starting to think that for gltfio the robustness checks should be a compile option. Because if you know for a fact that your files are good, you shouldn't have to pay the price of the checks (performance and size). On the other hand, if you can't guarantee it because you don't control the files, you might want to use the "robust" version. In debug builds, the checks and logs could be always enabled. In release builds, it would be a compile time option for both the checks and the logs. 


I don't know if it's feasible, but maybe we need a separate validation library. People who need the robustness, could use to first validate the files? This way gltfio itself can stay lean and mean.

prideout:
> > > Mainly I would like to avoid logging in release builds. logging is a response to an error, and that should be decided by the caller -- not us.
> > 
> > 
> > This makes sense.  So...I have an idea (for the future):
> > If gltfio needs to be robust against certain types of badly formed data, maybe it can recover but still tell clients about it by using an "info log" string, similar to how OpenGL handles shader errors.  e.g. Instead of emitting to the log, ResourceLoader could provide a `getInfoLog()` query.
> 
> Yes, I think it makes sense for gltfio to adopt a defensive stance. Which is different from the philosophy in filament.
> 
> But the question is, do we make all the checks in debug and release? If we are truly robust against ill formed files only in debug, then why not log only in debug too?
> 
> I'm starting to think that for gltfio the robustness checks should be a compile option. Because if you know for a fact that your files are good, you shouldn't have to pay the price of the checks (performance and size). On the other hand, if you can't guarantee it because you don't control the files, you might want to use the "robust" version. In debug builds, the checks and logs could be always enabled. In release builds, it would be a compile time option for both the checks and the logs.
> 
> I don't know if it's feasible, but maybe we need a separate validation library. People who need the robustness, could use to first validate the files? This way gltfio itself can stay lean and mean.

Yes, I like the idea of a compile-time option.  We actually already ask cgltf (the 3P library) to perform an explicit validation step in debug builds, but it does not catch everything.

I'm a little unsure about creating a separate validation library since that implies some duplication of effort: the "robust" version of gltfio would need to detect & recover errors, and that same detection would exist in the validation library.

Again, I like the idea of a "fast" version of gltfio.  For example a generic viewer should probably be robust, but a game with fixed assets does not.

pixelflinger:
> For example a generic viewer should probably be robust, but a game with fixed assets does not.

exactly.

PR #3822, opened 2021-04-19
GPU based IBL prefiltering library

pixelflinger:


libs/iblprefilter/src/materials/iblprefilter.mat
```diff
@@ -0,0 +1,148 @@
...
+
+vertex {
+    void postProcessVertex(inout PostProcessVertexInputs postProcess) {
+        postProcess.vertex.xy = postProcess.normalizedUV;
+    }
+}
+
+fragment {
+
+void dummy() {}
```
prideout, line 68: what's this?

pixelflinger: oh it's just to thwart CLion -- it doesn't like our jsonish meta data stuff at the top of the materials.

libs/iblprefilter/include/iblprefilter/IBLPrefilterContext.h
```diff
@@ -0,0 +1,196 @@
...
+        filament::Texture* createReflectionsTexture(ReflectionsOptions options);
+
+        /**
+         * Generates a prefiltered cubemap.
+         * @param options                   Options for this environment
+         * @param environmentCubemap        Environment cubemap (input). Can't be null.
+         * @param outReflectionsTexture     Output prefiltered texture or if null it is created.
+         * @return returns outReflectionsTexture
+         */
+        filament::Texture* operator()(Options options,
```
bejado, line 155: I think it's a bit unexpected to make this API a function call operator, when we also have the `createReflectionsTexture` method. I'd probably make this into a method called `createPrefilteredCubemapTexture` or something.

libs/iblprefilter/src/IBLPrefilterContext.cpp
```diff
@@ -0,0 +1,387 @@
...
+            weight += NoL; // note: L.z == NoL
+            p[i] = { L, mipLevel };
+        }
+
+        // upload the coefficients
+        weights->setImage(engine, 0, {
+                p, effectiveSampleCount * sizeof(half4),
+                PixelDataFormat::RGBA, PixelDataType::HALF });
+
+        mi->setParameter("kernel", weights, TextureSampler{ SamplerMagFilter::NEAREST });
```
bejado, line 353: I would maybe rename either `kernel` or `weights` to match the other.

libs/iblprefilter/include/iblprefilter/IBLPrefilterContext.h
bejado: Actually, I think my question is rather: why is `createReflectionsTexture` public?

libs/iblprefilter/src/IBLPrefilterContext.cpp
pixelflinger: sounds good

libs/iblprefilter/include/iblprefilter/IBLPrefilterContext.h
pixelflinger: I wanted to avoid "createXXX" because it's more than that -- it's launch the computation. I didn't want to call it `.filter()`. I though of `.process()`. But then I thought in fact it doesn't need a name. It's just the "do it" method.

pixelflinger: `createReflectionsTexture` is more like a helper, it'll create the texture with the right `Usage` for instance. But you don't have to use it.

libs/iblprefilter/include/filament/iblprefilter/IBLPrefilterContext.h
```diff
@@ -0,0 +1,203 @@
...
+ * IBLPrefilterContext context(engine);
+ * IBLPrefilterContext::SpecularFilter filter(context);
+ * Texture* texture = filter(environment_cubemap);
+ *
+ * IndirectLight* indirectLight = IndirectLight::Builder()
+ *     .reflections(texture)
+ *     .build(engine);
+ * ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+ */
+class IBLPrefilterContext {
```
prideout, line 57: Nice comment block!  If I had to prefilter several cubemaps, I guess it would be best to create only one context and one filter, then invoke the filter multiple times, right?

prideout: Is there an alternate usage pattern that uses `createReflectionsTexture`, which could be mentioned here?

libs/iblprefilter/src/IBLPrefilterContext.cpp
```diff
@@ -0,0 +1,418 @@
...
+
+    Engine& engine = mContext.mEngine;
+    View* const view = mContext.mView;
+    Renderer* const renderer = mContext.mRenderer;
+    Texture* const kernel = mKernelTexture;
+    MaterialInstance* const mi = mContext.mMaterial->createInstance("IBLPrefilterContext::Filter");
+
+    RenderableManager& rcm = engine.getRenderableManager();
+    rcm.setMaterialInstanceAt(
+            rcm.getInstance(mContext.mFullScreenQuadEntity), 0, mi);
```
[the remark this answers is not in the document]
pixelflinger: Ah yes, this is the remnants from something else I tried. Good catch.

libs/iblprefilter/include/filament/iblprefilter/IBLPrefilterContext.h
pixelflinger: > Nice comment block! If I had to prefilter several cubemaps, I guess it would be best to create only one context and one filter, then invoke the filter multiple times, right?

Yes, it would be a lot more efficient -- especially if you can reuse the same `SpecularFilter`.

bejado (approved):

PR #3824, opened 2021-04-19
DO NOT MERGE: Prefilter test

pixelflinger:


prideout:
Note: Vulkan support is taking a few days because I'm doing a fairly significant refactor of command buffer management to allow for this.

prideout:
I think we can close this, I'm able to test prefiltering on main.

PR #3825, opened 2021-04-19
don't use TARGET_MOBILE for quality settings

pixelflinger:
Until now we were using TARGET_MOBILE in our shaders for both 
quality setting and correctness.

With this PR we introduce a new define FILAMENT_QUALITY which currently
can take two value FILAMENT_QUALITY_LOW and FILAMENT_QUALITY_HIGH and
we use that instead of TARGET_MOBILE for quality settings.

Currently, FILAMENT_QUALITY_LOW is selected when TARGET_MOBILE is set.


prideout (approved):

bejado (approved):
Should we also rename the [`FILAMENT_TARGET_MOBILE`](https://github.com/google/filament/blob/032ce5ff83572e11a16f0c3c4fbe85756c5598e6/filament/CMakeLists.txt#L199) cpp definition?

pixelflinger:
> Should we also rename the [`FILAMENT_TARGET_MOBILE`](https://github.com/google/filament/blob/032ce5ff83572e11a16f0c3c4fbe85756c5598e6/filament/CMakeLists.txt#L199) cpp definition?

probably -- I'm thinking as a separate PR.

PR #3837, opened 2021-04-22
Add Transcoder API for C++ clients.

prideout:
We will add JavaScript and Java support in a later PR.

### DIFF
--- a/filament/include/filament/VertexBuffer.h
+++ b/filament/include/filament/VertexBuffer.h
@@ -116,6 +116,9 @@ public:
          * @warning VertexAttribute::TANGENTS must be specified as a quaternion and is how normals
          *          are specified.
          *
+         * @warning Not all backends support 3-component attributes that are not floats. For help
+         *          with conversion, see geometry::Transcoder.
+         *
          * @see VertexAttribute
          *
          * This is a no-op if the \p attribute is an invalid enum.
--- a/libs/geometry/include/geometry/Transcoder.h
+++ b/libs/geometry/include/geometry/Transcoder.h
@@ -20,6 +20,7 @@
 #include <utils/compiler.h>
 
 #include <stddef.h>
+#include <stdint.h>
 
 namespace filament {
 namespace geometry {
@@ -46,8 +47,8 @@ enum class ComponentType {
  * Transcoder transcode({
  *     .componentType = ComponentType::BYTE,
  *     .normalized = true,
- *     .numComponents = 3,
- *     .strideBytes = 0
+ *     .componentCount = 3,
+ *     .inputStrideBytes = 0
  * });
  *
  * transcode(outputPtr, inputPtr, count);
@@ -65,8 +66,8 @@ public:
     struct Config {
         ComponentType componentType;
         bool normalized;
-        int numComponents;
-        int strideBytes = 0; //!< If stride is 0, the transcoder assumes tight packing.
+        uint32_t componentCount;
+        uint32_t inputStrideBytes = 0; //!< If stride is 0, the transcoder assumes tight packing.
     };
 
     /**
--- a/libs/geometry/include/geometry/Transcoder.h
+++ b/libs/geometry/include/geometry/Transcoder.h
@@ -75,7 +75,7 @@ public:
      *
      * The config is not passed by const reference to allow for type inference at the call site.
      */
-    Transcoder(Config config) : mConfig(config) {}
+    Transcoder(Config config) noexcept : mConfig(config) {}
 
     /**
      * Converts arbitrary data into tightly packed 32-bit floating point values.
@@ -91,7 +91,8 @@ public:
      * @return Number of bytes required to contain "count" items after conversion to packed floats
      *
      */
-    size_t operator()(float* target, void const* source, size_t count) const noexcept;
+    size_t operator()(float* UTILS_RESTRICT target, void const* UTILS_RESTRICT source,
+            size_t count) const noexcept;
 
 private:
     const Config mConfig;
--- a/libs/geometry/include/geometry/Transcoder.h
+++ b/libs/geometry/include/geometry/Transcoder.h
@@ -1,0 +1,102 @@
+/*
+ * Copyright (C) 2021 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef TNT_GEOMETRY_TRANSCODER_H
+#define TNT_GEOMETRY_TRANSCODER_H
+
+#include <utils/compiler.h>
+
+#include <stddef.h>
+
+namespace filament {
+namespace geometry {
+
+enum class ComponentType {
+    BYTE,   //!< If normalization is enabled, this maps from [-127,127] to [-1,+1]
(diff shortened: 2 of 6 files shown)


filament/include/filament/VertexBuffer.h
```diff
@@ -116,6 +116,9 @@ class UTILS_PUBLIC VertexBuffer : public FilamentAPI {
          * @warning VertexAttribute::TANGENTS must be specified as a quaternion and is how normals
          *          are specified.
          *
+         * @warning Not all backends support 3-component attributes that are not floats. For help
```
romainguy, line 119: Do we have an API apps can query to know that?

libs/geometry/src/Transcoder.cpp
```diff
@@ -0,0 +1,119 @@
...
+using filament::math::half;
+
+namespace filament {
+namespace geometry {
+
+// The internal workhorse function of the Transcoder, which takes arbitrary input but always
+// produced packed floats. We expose a more readable interface than this to users, who often have
+// untyped blobs of interleaved data.
+template<typename SOURCE_TYPE, int NORMALIZATION_FACTOR>
+void convert(float* target, void const* source, size_t count, int numComponents,
```
romainguy, line 30: Can you check (maybe with Mathias' help) that those loops get nicely SIMDified by the compiler? They seem to be perfect candidates for this (or even for job system for large data sets; if you have millions of vertices for instance).

romainguy (approved):

filament/include/filament/VertexBuffer.h
prideout: Mathias and I had discussed this, we wish to avoid adding too many queries.  One day in the future we might be able to do this conversion automatically.

libs/geometry/src/Transcoder.cpp
prideout: The resulting ARM code is fairly tight when using -O3 with clang.  For example, the multiply gets eliminated when it is 1 (which is why I made it consexpr) and the `scvtf` instruction is used for the cast:

https://gcc.godbolt.org/z/jfzMbfW8M

libs/geometry/include/geometry/Transcoder.h
```diff
@@ -0,0 +1,102 @@
...
+ *     .numComponents = 3,
+ *     .strideBytes = 0
+ * });
+ *
+ * transcode(outputPtr, inputPtr, count);
+ * ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+ *
+ * The interpretation of signed normalized data is consistent with Vulkan and OpenGL ES 3.0+.
+ * Note that this slightly differs from earlier versions of OpenGL ES.  For example, a signed byte
+ * value of -127 maps exactly to -1.0f under ES3 and VK rules, but not ES2.
```
prideout, line 58: This was an interesting thing that I learned while writing this.

libs/geometry/src/Transcoder.cpp
romainguy: The output is even better if you make `numComponents` a template argument: https://gcc.godbolt.org/z/ecsdxv6hx. Realistically numComponents will mostly be 3, maybe 1 or 2, so it might be worth it.

prideout: I had thought of doing that but it is debatable if it is worth it.  The outer loop in my version has fewer instructions, yes there's a branch in there but this is a RISC architecture.

prideout: Oh I see, it got simdified.  Sorry about that, haven't woken up yet.  Yeah I'll go ahead and do this for 3 & 4 components.

prideout: Added a commit for this.

bejado (approved):

filament/include/filament/VertexBuffer.h
pixelflinger: Yes so our current stance is that "vk doesn't support it" (even if some vk might).

libs/geometry/include/geometry/Transcoder.h
```diff
@@ -0,0 +1,102 @@
...
+ */
+class UTILS_PUBLIC Transcoder {
+public:
+    /**
+     * Describes the format of all input data that get passed to this transcoder object.
+     */
+    struct Config {
+        ComponentType componentType;
+        bool normalized;
+        int numComponents;
```
pixelflinger, line 68: `numComponents` should be `componentCount`
also it should be unsigned, use `uint32_t` or `size_t`.

libs/geometry/include/geometry/Transcoder.h
```diff
@@ -0,0 +1,102 @@
...
+        int strideBytes = 0; //!< If stride is 0, the transcoder assumes tight packing.
```
pixelflinger, line 69: Should the stride be unsigned too? I guess it's possible to have a negative stride? 
Also maybe rename to `inputStrideBytes`

libs/geometry/include/geometry/Transcoder.h
```diff
@@ -0,0 +1,102 @@
...
+    };
+
+    /**
+     * Creates an immutable function object with the specified configuration.
+     *
+     * The config is not passed by const reference to allow for type inference at the call site.
+     */
+    Transcoder(Config config) : mConfig(config) {}
```
pixelflinger, line 77: `noexcept`

libs/geometry/include/geometry/Transcoder.h
```diff
@@ -0,0 +1,102 @@
...
+     *
+     * If target is null, returns the number of bytes required.
+     *
+     * @param target Client owned area to write into, or null for a size query
+     * @param source Pointer to the data to read from (does not get retained)
+     * @param count The maximum number of items to write (i.e. number of float3 values, not bytes)
+     * @return Number of bytes required to contain "count" items after conversion to packed floats
+     *
+     */
+    size_t operator()(float* target, void const* source, size_t count) const noexcept;
```
pixelflinger, line 93: make these pointers `UTILS_RESTRICT` this will help with vectorizing.

libs/geometry/src/Transcoder.cpp
```diff
@@ -0,0 +1,193 @@
...
+
+namespace filament {
+namespace geometry {
+
+// The internal workhorse function of the Transcoder, which takes arbitrary input but always
+// produced packed floats. We expose a more readable interface than this to users, who often have
+// untyped blobs of interleaved data. Note that this variant takes an arbitrary number of
+// components, we also have a fixed-size variant for better compiler output.
+template<typename SOURCE_TYPE, int NORMALIZATION_FACTOR>
+void convert(float* target, void const* source, size_t count, int numComponents,
```
pixelflinger, line 31: `UTILS_RESTRICT` pointers will help with vectorizing / unrolling.

pixelflinger: I *think* C++ 17 allows `float` template parameters (e.g. for NORMALIZATION_FACTOR)

pixelflinger: `numComponents` -> `componentCount`

pixelflinger (changes requested):

libs/geometry/src/Transcoder.cpp
prideout: > I _think_ C++ 17 allows `float` template parameters (e.g. for NORMALIZATION_FACTOR)

I tried this and failed.

libs/geometry/include/geometry/Transcoder.h
prideout: Done

libs/geometry/include/geometry/Transcoder.h
prideout: Done

libs/geometry/src/Transcoder.cpp
prideout: Done

libs/geometry/include/geometry/Transcoder.h
prideout: Done

PR #3840, opened 2021-04-22
gltfio: Use the new Transcoder API in some cases.

prideout:
This fixes issues seen seen with Vulkan in conjunction with quantized model variants in the glTF samples repo such as the Avocado.

We already support the quantized glTF extension, which is when this
comes up. Technically this is only needed for Vulkan but gltfio does
not (and should not) know which backend is being used, so we simply
apply these transformations assuming the worst.

In practice this code usually won't be activated, and if even when it
is, it won't hurt loading time as much as other things like large
textures.

### DIFF
--- a/android/filament-android/CMakeLists.txt
+++ b/android/filament-android/CMakeLists.txt
@@ -97,7 +97,6 @@ target_link_libraries(filament-jni
     PRIVATE backend
     PRIVATE filaflat
     PRIVATE filabridge
-    PRIVATE geometry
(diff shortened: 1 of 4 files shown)


pixelflinger (approved):
is this tested with some specific gltf files?

prideout:
> is this tested with some specific gltf files?

Yes the commit message says `Fixes issue seen with Vulkan + the quantized version of Avocado.`, I think I added that after I made the PR.

PR #3850, opened 2021-04-24
New experimental tone mapper

romainguy:
The new tone mapper proposed in this change replaces Uchimura (see below) and is composed of
two distinct algorithms:

- The first is the customizable compression curve from Timothy Lotte (from GDC 2015). The curve's
  default parameters were chosen to match the overall appearance (contrast) of the ACES curves
  used as our current default tone mapper. This was done to provide a good migration path to
  making this new solution the default in a future version of Filament.
  The compression curve can be customized via its contrast, shoulder, input middle gray, output
  middle gray, and maximum "HDR value"(scene referred). It is important to note that the compression
  is only applied to the luminance of the input scene referred value.

- The second is a novel approach by Troy Sobotka called Exposure Value Invariant Luminance Scaling
  (or EVILS). This solution works on the largest chromatic channel of the input value as opposed
  to mapping all RGB channels as we do in our other tone mappers. The algorithm effectively scales
  all channels by a ratio that depends on both the compressed luminance from the first stage and
  the luminance of input value normalized by the largest channel.

This new solution offers important benefits as it manages to preserve chromaticity for high input
values. It is also designed to provide a desirable rolloff to white on overexposure. This means that
even without full gamut mapping (since we only , this solution avoids many hue skews (red to orange,
blue to purple, etc.). The results provided by this solution are much more natural and make using
other color grading tools much easier.

We replace Uchimura in this change — technically an API breakage — because Uchimura suffers
from heavy hue skews and does not offer any benefits over this new solution. Our other tone
mappers remain useful for backward compatibility (`Filmic`) or for validation with other
tools/engines (`ACES`, `Reinhard`).

Analysis of EVILS can be found in the updated Mathematica notebook `Tone mapping Operators`
and in the new notebook `EVILS`. The latter provides an interactive tool to try the algorithm
on an image (note that this notebook uses a different compression curve).

Here are various comparisons with `ACES (Legacy)`, our default tone mapper.

Bistro scene with ACES, notice the skew on the red awning:
Screen Shot 2021-04-23 at 7 51 31 PM

With EVILS, the skew disappears in favor of a rolloff to white:
Screen Shot 2021-04-23 at 7 51 26 PM

Primaries test scene with ACES, notice how the blue sphere skews to purple, the red sphere to
orange, and the green plane to yellow:
Screen Shot 2021-04-23 at 7 52 04 PM

With EVILS, chromaticity is properly retained with a rolloff to white on the most intense areas
directly illuminated by the ~150,000 lux directional light:
Screen Shot 2021-04-23 at 7 52 01 PM

Blue light saber with ACES, notice the hue skew again:
Screen Shot 2021-04-23 at 8 37 57 PM

With EVIL, the skew is gone:
Screen Shot 2021-04-23 at 8 37 48 PM

Red light saber with ACES, notice how the core skews to orange:
Screen Shot 2021-04-23 at 8 39 06 PM

With EVILS, the skew is gone:
Screen Shot 2021-04-23 at 8 39 09 PM


filament/src/ColorGrading.cpp
```diff
@@ -573,10 +576,10 @@ FColorGrading::FColorGrading(FEngine& engine, const Builder& builder) {
...
                 #pragma clang loop vectorize_width(8)
                 for (size_t i = 0; i < count; ++i) {
-                    float4 v{ src[i] };
-                    uint32_t r = uint32_t(floorf(v.x * 1023.0f + 0.5f));
-                    uint32_t g = uint32_t(floorf(v.y * 1023.0f + 0.5f));
-                    uint32_t b = uint32_t(floorf(v.z * 1023.0f + 0.5f));
+                    float4 v{src[i]};
+                    uint32_t r = uint32_t(floor(v.x * 1023.0f + 0.5f));
+                    uint32_t g = uint32_t(floor(v.y * 1023.0f + 0.5f));
+                    uint32_t b = uint32_t(floor(v.z * 1023.0f + 0.5f));
```
pixelflinger, line 582: usually we use `std::floor`

filament/src/ToneMapping.cpp
```diff
@@ -15,7 +15,7 @@
  */
 
 #include "ToneMapping.h"
-
+#include <iostream>
```
pixelflinger, line 18: we shouldn't need `iostream`

filament/src/ToneMapping.cpp
```diff
@@ -187,33 +186,62 @@ float3 ACES_Legacy(float3 x) noexcept {
...
-    float3 T = m * pow(x / m, c) + b;
-    float3 S = P - (P - S1) * exp(CP * (x - S0));
-    float3 L = m + a * (x - m);
-
-    return T * w0 + L * w1 + S * w2;
+float genericTonemap(float x, float contrast, float shoulder,
+        float midGreyIn, float midGreyOut, float hdrMax) noexcept {
+    // Lottes, 2016,"Advanced Techniques and Optimization of VDR Color Pipelines"
+    // https://gpuopen.com/wp-content/uploads/2016/03/GdcVdrLottes.pdf
+    float mc = pow(midGreyIn, contrast);
```
pixelflinger, line 193: `std::pow` everywhere

pixelflinger (changes requested):

filament/src/ColorGrading.cpp
romainguy: Done

filament/src/ToneMapping.cpp
romainguy: Done

filament/src/ToneMapping.cpp
romainguy: Done

pixelflinger (approved):

prideout (approved):

bejado (approved):

PR #3851, opened 2021-04-24
Improve libiblprefilter

pixelflinger:
- move public headers from filament/foo to filament-foo/
  to avoid confusion with libfilament's headers
- add support for equirectangular to cubemap conversion
- add support for using an .hdr file directly in all our samples


romainguy (approved):

libs/iblprefilter/src/materials/equirectToCube.mat
```diff
@@ -0,0 +1,78 @@
...
+
+fragment {
+
+void dummy() {}
+
+highp vec2 toEquirect(const highp vec3 s) {
+    highp float xf = atan2(s.x, s.z) * (1.0 / PI);  // range [-1.0, 1.0]
+    highp float yf = asin(s.y) * (2.0 / PI);        // range [-1.0, 1.0]
+    xf = (xf + 1.0) * 0.5;                          // range [0, 1.0 [
+    yf = (1.0 - yf) * 0.5;                          // range [0, 1.0 [
```
prideout, line 54: either a typo or a range notation that I'm not familiar with :)

prideout (approved):
This is sooooo cool.  Do we need to update the usage message in the samples?  For example, gltf_viewer --help says:

```
        "   --ibl=<path to cmgen IBL>, -i <path>\n"
        "       Override the built-in IBL\n\n"
```

libs/iblprefilter/src/materials/equirectToCube.mat
romainguy: It means the 1.0 is exclusive.

pixelflinger: Ah yeah, correct, it's actually inclusive -- this is because this comment comes from the CPU version.

libs/filamentapp/src/IBL.cpp
```diff
@@ -53,6 +55,65 @@ IBL::~IBL() {
...
+    Texture::PixelBufferDescriptor buffer(
+            data, size,Texture::Format::RGB, Texture::Type::FLOAT,
+            [](void* buffer, size_t size, void* user) { stbi_image_free(buffer); });
+
+    Texture* const equirect = Texture::Builder()
+            .width((uint32_t)w)
+            .height((uint32_t)h)
+            .levels(0xff)
+            .format(Texture::InternalFormat::R11F_G11F_B10F)
+            .sampler(Texture::Sampler::SAMPLER_2D)
```
bejado, line 89: Do you need to set `usage` here to `COLOR_ATTACHMENT | SAMPLEABLE`?

bejado (approved):

PR #3858, opened 2021-04-26
Cleanup some shader #defines

pixelflinger:
- create a new FILAMENT_HAS_FEATURE_TEXTURE_GATHER, which can be used
  in the shaders to use textureGather.

- A new TARGET_GLES_ENVIRONMENT and TARGET_GL_ENVIRONMENT in addition
  to TARGET_METAL_ and TARGET_VULKAN_, for completeness.

- Rename TARGET_LANGUAGE_SPIRV to FILAMENT_VULKAN_SEMANTICS and add
  FILAMENT_OPENGL_SEMANTICS for completeness.

- expose the shader model to shaders as: FILAMENT_SHADER_MODEL_ES30 and
  FILAMENT_SHADER_MODEL_GL41

Use these new #define where appropriate.

The main change here is that TARGET_MOBILE is now only used for dealing
with the `mediump` default on mobile.

This change also enables textureGather on Metal.


shaders/src/common_types.fs
```diff
@@ -1,4 +1,4 @@
-#if !defined(TARGET_MOBILE) || defined(TARGET_LANGUAGE_SPIRV)
+#if defined(FILAMENT_SHADER_MODEL_GL41) || defined(FILAMENT_VULKAN_SEMANTICS)
 #define LAYOUT_LOCATION(x) layout(location = x)
```
bejado, line 2: Do we really need the `layout(location = x)` for OpenGL desktop?

bejado (commented):
Can we add a table to our documentation somewhere?

Also, I'd be in favor of renaming `MODEL_ES30` to `MODEL_MOBILE` and `MODEL_GL41` to `MODEL_DESKTOP`.

| Target        | `FILAMENT_SHADER_MODEL_ES30` | `FILAMENT_SHADER_MODEL_GL41` | `FILAMENT_VULKAN_SEMANTICS` | `FILAMENT_OPENGL_SEMANTICS` |
|---------------------------|------------------------------|------------------------------|-----------------------------|-----------------------------|
| Metal on iOS              |               1              |               0              |              1              |              0              |
| Metal on macOS            |               0              |               1              |              1              |              0              |
| Vulkan on Android         |               1              |               0              |              1              |              0              |
| Vulkan on Desktop         |               0              |               1              |              1              |              0              |
| OpenGL ES 3.0+ on Android |               1              |               0              |              0              |              1              |
| OpenGL on Desktop         |               0              |               1              |              0              |              1              |

shaders/src/common_types.fs
pixelflinger: looks like not!

bejado (approved):

libs/filamat/src/shaders/CodeGenerator.cpp
```diff
@@ -453,6 +465,36 @@ io::sstream& CodeGenerator::generateMaterialProperty(io::sstream& out,
...
 
+io::sstream& CodeGenerator::generateQualityDefine(io::sstream& out, ShaderQuality quality) const {
+    out << "#define FILAMENT_QUALITY_LOW    0\n";
+    out << "#define FILAMENT_QUALITY_NORMAL 1\n";
+    out << "#define FILAMENT_QUALITY_HIGH   2\n";
+
+    switch (quality) {
+        case ShaderQuality::DEFAULT:
+            switch (mShaderModel) {
+                default:                        goto quality_normal;
```
romainguy, line 476: You broke my brain with this :)

romainguy (approved):

PR #3864, opened 2021-04-27
Enable Vulkan in our Android Release package.

prideout:
This makes release builds consistent with continuous builds and default
local builds. It also makes it easier to test the Vulkan backend.
Clients who need a smaller APK size can always make their own build.

This also makes filament-android-release.aar go from 1.8M to 2.3M.
Note that we are now including bluevk, vkshaders (for the blit shader),
smol-v (for SPIRV decompression) and vk_mem_alloc (for mem allocation).

### DIFF
--- a/build/android/build.sh
+++ b/build/android/build.sh
@@ -11,9 +11,6 @@
 NDK_VERSION="ndk;22.0.7026061"
 ANDROID_NDK_VERSION=22
 
-# Exclude Vulkan from CI builds for Android. (It is enabled for other platforms.)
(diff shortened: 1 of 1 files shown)


romainguy (approved):

pixelflinger (approved):

bejado:
We want this for Maven artifacts too, yes? Starting with v1.9.23, I'll exclude the `-v` flag when building.

romainguy:
That's correct yes.

PR #3870, opened 2021-04-28
Fix incorrect units documented for getIntensity

bejado:


romainguy (approved):
I wonder if we should also keep the value set by the user so they have an API to retrieve what they set

bejado:
> I wonder if we should also keep the value set by the user so they have an API to retrieve what they set

I think we could add a `getIntensityLumens` to return whatever the user called `setIntensity` with.

pixelflinger:
I would like this stuff to work like std::Chrono::duration ideally.
That is, an intensity is always in some canonical unit, and the user can specify/convert to whatever they like.

bejado:
> the user can specify/convert to whatever they like

The problem here is that this conversion also depends on the cone angle of the spot light, so it wouldn't work quite as cleanly as`std::chrono::duration` does. I think there's a chance for a better API here, but I need to think about it a bit.

PR #3871, opened 2021-04-28
Fix memory leak inside utils::details::SafeString

bejado:
Calls to `strdup` must be balanced with calls to `free`.

### DIFF
--- a/libs/utils/include/utils/NameComponentManager.h
+++ b/libs/utils/include/utils/NameComponentManager.h
@@ -26,6 +26,8 @@
 #include <utils/EntityInstance.h>
 #include <utils/SingleInstanceComponentManager.h>
 
+#include <functional>
+
 namespace utils {
 
 class EntityManager;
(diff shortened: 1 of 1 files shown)


prideout (approved):
Looks legit to me.  Nice one.

pixelflinger (commented):
Why would this be needed? We're in the ctor, mCstr should be nullptr.

libs/utils/include/utils/NameComponentManager.h
```diff
@@ -37,6 +37,7 @@ class SafeString {
     explicit SafeString(const char* str) noexcept : mCStr(strdup(str)) { }
     SafeString(SafeString&& rhs) noexcept : mCStr(rhs.mCStr) { rhs.mCStr = nullptr; }
     SafeString& operator=(SafeString&& rhs) noexcept {
+        free((void*)mCStr);
```
pixelflinger, line 40: Actually, it would be better to implement this with swap().

pixelflinger (changes requested):

pixelflinger (approved):

PR #3873, opened 2021-04-28
Add gltf-viewer iOS sample

bejado:
This sample is akin to Android's `sample-gltf-viewer`. It allows loading glTF files and updating settings through the Filament Remote site. The idea is for this to replace the `hello-gltf` sample, which has grown stale. It doesn't have complete feature parity yet; namely loading zipped glTF files is not yet supported.

One hitch is that the Filament remote site will need to be updated to support specifying an IP address, as iOS does not have the equivalent of `adb forward`. I'll add this in a separate PR.

Screen Shot 2021-04-28 at 3 00 58 PM


prideout:
Super cool!  Looking over it now...

prideout (approved):
Looks great, let me know if you need help with the HTML / JS side of things.

pixelflinger:
Nice!

PR #3875, opened 2021-04-30
Vulkan: big refactor for command buffer management.

prideout:
This removes a lot of caveats that we had in the Vulkan backend. Many driver API calls were unsupported due to a potential "missing" command buffer, but in the new scheme, a command buffer is always made available, and there's no weird separate "work and swap chain command buffers".

Here's how various driver API entry points now map into Vulkan actions:

```
makeCurrent => Acquire Swap Chain.
beginFrame => Do nothing.
endFrame => Submit command buffer and create a new one.
commit => Submit commit buffer again if necessary, Present Swap Chain.
```

This handles both `copyFrame` (which does not do beginFrame / endFrame) and `renderStandaloneView` (which does not do makeCurrent / commit).

### DIFF
--- a/filament/backend/src/vulkan/VulkanDisposer.cpp
+++ b/filament/backend/src/vulkan/VulkanDisposer.cpp
@@ -17,9 +17,14 @@
 #include "vulkan/VulkanDisposer.h"
 
 #include <utils/debug.h>
+#include <utils/Log.h>
 
 namespace filament {
 namespace backend {
+
+// Always wait at least 3 frames after a DriverAPI-level resource has been destroyed for safe
+// destruction, due to potential usage by outstanding command buffers and triple buffering.
+static constexpr uint32_t FRAMES_BEFORE_EVICTION = 3;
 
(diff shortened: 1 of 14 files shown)


filament/backend/src/vulkan/VulkanDisposer.cpp
```diff
@@ -17,10 +17,15 @@
...
 
 #include <utils/debug.h>
+#include <utils/Log.h>
 
 namespace filament {
 namespace backend {
 
+// Always wait at least 3 frames after a DriverAPI-level resource has been destroyed for safe
+// destruction, due to potential usage by outstanding command buffers and triple buffering.
+static constexpr uint32_t FRAMES_BEFORE_EVICTION = 3;
```
bejado, line 27: Do we need to do something special now in `VulkanDriver::terminate` (like call `mDisposer.gc()` 3 times) to ensure that all resources are actually destroyed?

prideout: Good question: terminate calls `mDisposer.reset()` which has been updated so that it now ensures that all resources are actually destroyed.

PR #3882, opened 2021-05-03
Support a maximum of 8 render buffers (from 4)

pixelflinger:
We chose 8 renderbuffer because it seems to be commonly supported
on mobile gpu and is more than 6 which is needed for generating
cubemaps. 4 was the minimum guaranteed by ES3.0.

The main changes here are:
- TargetBufferFlags is now a uint32_t instead of uint8_t
- The number of render buffer supported by filament and supported by
  the hardware can now be different. Applications/users now must
  query this value if more than 4 is needed.
- added a backend query (used gl terminology)

This is a redo of a previous attempt. There was an issue in the GL
backend that caused MRT to fail on some devices.

### DIFF
--- a/filament/backend/include/backend/DriverEnums.h
+++ b/filament/backend/include/backend/DriverEnums.h
@@ -86,22 +86,37 @@ static constexpr const char* backendToString(backend::Backend backend) {
 /**
  * Bitmask for selecting render buffers
  */
-enum class TargetBufferFlags : uint8_t {
+enum class TargetBufferFlags : uint32_t {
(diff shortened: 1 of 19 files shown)


romainguy (approved):

bejado (approved):

prideout (changes requested):
Mathias can you also update `VulkanDriver::blit`?  There is a sequence of `if` statements in there that should probably be changed into a loop.

prideout (approved):
Thanks!

PR #3883, opened 2021-05-03
Vulkan: fix image view type for render target attachments.

prideout:
When rendering into a single cubemap face, we should use
VK_IMAGE_VIEW_TYPE_2D rather than VK_IMAGE_VIEW_TYPE_CUBE.

### DIFF
--- a/filament/backend/src/vulkan/VulkanHandles.h
+++ b/filament/backend/src/vulkan/VulkanHandles.h
@@ -143,7 +143,9 @@ struct VulkanTexture : public HwTexture {
     void setPrimaryRange(uint32_t minMiplevel, uint32_t maxMiplevel);
 
     // Gets or creates a cached VkImageView for a range of miplevels and array layers.
-    VkImageView getImageView(VkImageSubresourceRange range);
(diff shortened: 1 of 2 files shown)


filament/backend/src/vulkan/VulkanHandles.h
```diff
@@ -143,7 +143,9 @@ struct VulkanTexture : public HwTexture {
     void setPrimaryRange(uint32_t minMiplevel, uint32_t maxMiplevel);
 
     // Gets or creates a cached VkImageView for a range of miplevels and array layers.
-    VkImageView getImageView(VkImageSubresourceRange range);
+    // If force2D is true, this always returns VK_IMAGE_VIEW_TYPE_2D, regardless of the
```
pixelflinger, line 146: it's not so much that it returns that constant right? It returns an VkImageView of that type, correct?

pixelflinger (approved):

filament/backend/src/vulkan/VulkanHandles.h
prideout: Correct, I tweaked the wording.

PR #3884, opened 2021-05-03
Vulkan: enable anisotropic filtering if supported.

prideout:


filament/backend/src/vulkan/VulkanContext.cpp
```diff
@@ -189,6 +189,7 @@ void createLogicalDevice(VulkanContext& context) {
     VkPhysicalDeviceFeatures enabledFeatures {
         .textureCompressionETC2 = supportedFeatures.textureCompressionETC2,
         .textureCompressionBC = supportedFeatures.textureCompressionBC,
+        .samplerAnisotropy = supportedFeatures.samplerAnisotropy,
```
pixelflinger, line 192: apparently the failure on window is because the designated initializers must be in order

pixelflinger (changes requested):

PR #3891, opened 2021-05-04
Fix Filament remote file upload

bejado:
With the latest Chrome update, `websocket.send(file)` no longer works. Chrome sends a binary message of length 0 bytes. `.send` accepts a `Blob` object, which `File` supposedly inherits from, but I bet we have to actually _read_ the entire file stream before sending it.

### DIFF
--- a/web/samples/remote.html
+++ b/web/samples/remote.html
@@ -210,8 +210,10 @@ class App {
         const connection = this.websocket;
         if (connection && connection.readyState == connection.OPEN) {
             console.info(`Uploading ${file.name}`);
-            connection.send(file.name);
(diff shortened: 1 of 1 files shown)


prideout (approved):
Wow thanks!!

prideout:
If you want you can also push a direct update to the public-facing page in `docs`

bejado:
Sure, I'll do that after I land my other changes allowing the remote page to work with iOS devices.
By the way, I think this might have just been a bug in the Chrome developer tools. It showed a length of 0 bytes for binary messages, though I suspect still sent the message in full.

PR #3893, opened 2021-05-04
ios: Don't destroy gltfio materials between asset loads

bejado:
This causes a crash because `materialProvider` deletes an internal `dummyTexture`, which is never re-created.


prideout (approved):
Interesting, we should consider adding an evict() method to MaterialProvider for clients who need to free up memory.

bejado:
> Interesting, we should consider adding an evict() method to MaterialProvider for clients who need to free up memory.

Yeah, this could be useful.

PR #3895, opened 2021-05-04
metal: Query max color render targets

bejado:


filament/backend/src/metal/MetalDriver.mm
```diff
@@ -673,7 +684,7 @@
 }
 
 uint8_t MetalDriver::getMaxDrawBuffers() {
-    return backend::MRT::MIN_SUPPORTED_RENDER_TARGET_COUNT; // TODO: query real value
+    return mContext->maxColorRenderTargets;
```
pixelflinger, line 687: this should never return more than `MIN_SUPPORTED_RENDER_TARGET_COUNT`, so it might be safer for the future to do a `min()` here.

pixelflinger (changes requested):

filament/backend/src/metal/MetalDriver.mm
bejado: You mean `MAX_SUPPORTED_RENDER_TARGET_COUNT`, right? But yeah, good idea.

pixelflinger (approved):

PR #3899, opened 2021-05-05
Use 16 bit format for VSM

bejado:
iOS (and possibly Android) doesn't support filtering from RG32F textures. There is no discernible quality difference between 32 and 16 bit, so I'll just make this the default for now.


romainguy (approved):

romainguy:
Good read on this: https://mynameismjp.wordpress.com/2013/09/10/shadow-maps/

Couldn't we use a 16 bit UNORm format for VSM? EVSM needs floats I think but not VSM.

bejado:
Yeah we may be able to, I'll try experimenting with that as well.

PR #3906, opened 2021-05-06
Second memory leak fix in GL backend.

### LINKED ISSUE #3888: Release resources allocated with ResourceLoader::loadResources
**Describe the bug**
Not really sure if this is a bug, or a problem in our code.
We're trying to render a large number of gltf models, one at a time.

Code looks like this:
```
auto asset = assetLoader->createAssetFromBinary(
        reinterpret_cast<const uint8_t*>(gltfData.c_str()), gltfData.length());
    resourceLoader->loadResources(asset);
    scene->addEntities(asset->getEntities(), asset->getEntityCount());
    
    std::vector<uint8_t> outputImage(width * height * 4);
    backend::PixelBufferDescriptor pd(
        outputImage.data(), outputImage.size(), backend::PixelDataFormat::RGBA,
        backend::PixelDataType::UBYTE, nullptr, nullptr);
    
    renderer->beginFrame(swapchain);
    renderer->render(view);
    renderer->readPixels(0, 0, width, height, std::move(pd));
    renderer->endFrame();
    engine->flushAndWait();
    ...
    <save the buffer into a file>
    ...
    scene->removeEntities(asset->getEntities(), asset->getEntityCount());
    assetLoader->destroyAsset(asset);
```

This runs in a loop, receiving new models and render. It works as expected, however it looks like the resources loaded by `resourceLoader->loadResources(asset)` are not released and memory is "leaking". So after rendering several thousand models (all different), we run out of memory.

Is there a missing step here required to release these resources?

prideout:
The handle still needs to be destroyed.

Fixes #3888.

### DIFF
--- a/filament/backend/src/opengl/OpenGLDriver.cpp
+++ b/filament/backend/src/opengl/OpenGLDriver.cpp
@@ -1275,7 +1275,10 @@ void OpenGLDriver::createTimerQueryR(Handle<HwTimerQuery> tqh, int) {
 
 void OpenGLDriver::destroyVertexBuffer(Handle<HwVertexBuffer> vbh) {
     DEBUG_MARKER()
-    // Do nothing, because a VertexBuffer is just a collection of BufferObject handles.
+    if (vbh) {
+        GLVertexBuffer const* vb = handle_cast<const GLVertexBuffer*>(vbh);
+        destruct(vbh, vb);
+    }
 }
 
 void OpenGLDriver::destroyIndexBuffer(Handle<HwIndexBuffer> ibh) {


filament/backend/src/opengl/OpenGLDriver.cpp
```diff
@@ -1276,6 +1276,11 @@ void OpenGLDriver::createTimerQueryR(Handle<HwTimerQuery> tqh, int) {
 void OpenGLDriver::destroyVertexBuffer(Handle<HwVertexBuffer> vbh) {
     DEBUG_MARKER()
     // Do nothing, because a VertexBuffer is just a collection of BufferObject handles.
```
romainguy, line 1278: Remove this comment?

romainguy (approved):

PR #3908, opened 2021-05-07
new screen space lens flare effect

pixelflinger:
- this PR also improves bloom slightly

Note: for now this is not available on WebGL due to complications
in the bloom pass.

flare


filament/src/PostProcessManager.cpp
```diff
@@ -254,22 +255,33 @@ void PostProcessManager::init() noexcept {
...
+
     PixelBufferDescriptor dataOne(driver.allocate(4), 4, PixelDataFormat::RGBA, PixelDataType::UBYTE);
     PixelBufferDescriptor dataOneArray(driver.allocate(4), 4, PixelDataFormat::RGBA, PixelDataType::UBYTE);
     PixelBufferDescriptor dataZero(driver.allocate(4), 4, PixelDataFormat::RGBA, PixelDataType::UBYTE);
+    PixelBufferDescriptor dataStarburst(driver.allocate(256), 256, PixelDataFormat::R, PixelDataType::UBYTE);
     *static_cast<uint32_t *>(dataOne.buffer) = 0xFFFFFFFF;
     *static_cast<uint32_t *>(dataOneArray.buffer) = 0xFFFFFFFF;
     *static_cast<uint32_t *>(dataZero.buffer) = 0;
+    std::generate_n((uint8_t*)dataStarburst.buffer, 256,
+            [&dist = mUniformDistribution, &gen = mEngine.getRandomEngine()]() {
```
romainguy, line 269: Neat

filament/src/PostProcessManager.cpp
```diff
@@ -1467,10 +1480,56 @@ FrameGraphId<FrameGraphTexture> PostProcessManager::bloomPass(FrameGraph& fg,
...
+
+                    auto const& material = getPostProcessMaterial("flare");
+                    FMaterialInstance* mi = material.getMaterialInstance();
+
+                    mi->setParameter("color", in, {
+                            .filterMag = SamplerMagFilter::LINEAR,
+                            .filterMin = SamplerMinFilter::LINEAR_MIPMAP_NEAREST
+                    });
+
+                    mi->setParameter("level", 1.0f);    // adjust with resolution
```
romainguy, line 1513: Is a todo needed here? Or an option?

filament/src/materials/colorGrading/colorGrading.mat
```diff
@@ -65,66 +75,104 @@ fragment {
...
+
+vec3 bloom(const vec3 color) {
+    highp vec2 uv = variable_vertex.xy;
+
+    vec3 result = vec3(0.0);
+
+    vec3 bloom = textureLod(materialParams_bloomBuffer, uv, 0.0).rgb;
+    result += bloom * materialParams.bloom.x;
+
+    float starburstMask = starburst(uv);
```
romainguy, line 104: Wait the lens flare cannot be disabled?

romainguy: and the starburst too.

filament/src/materials/flare/flare.mat
```diff
@@ -0,0 +1,162 @@
...
+    float2 aspectRatio = materialParams.aspectRatio;
+
+    // make sure to avoid divide-by-zero in normalize
+    vec2 v = vec2(0.5001) - uv;
+    // handle aspect ratio
+    v.x *= aspectRatio.x;
+    v = normalize(v);
+    v.x *= aspectRatio.y;
+
+    vec3 s = sampleColor(uv + v * haloRadius);
```
romainguy, line 138: Can’t we skip this if chromatic aberration is turned off?

romainguy (changes requested):

filament/src/materials/colorGrading/colorGrading.mat
pixelflinger: The flare can be disabled, but not the starburst (when the flare is active).

filament/src/materials/flare/flare.mat
pixelflinger: We could make sampleColor do a single fetch of RGB instead of three r,g and b. We could add a bool in uniform to decide or have two shaders.
The cache should absorb all that though.

filament/src/materials/colorGrading/colorGrading.mat
pixelflinger: Oh I see what you mean. Yes, I can probably do better here. I was just relying on the texture being a 1x1. But we can save all that ALU.

Good point.

romainguy: Yeah and I would also be careful with the texture unit cache. Given that color grading samples the 3D LUT + the color buffer + the bloom buffer + maybe the dirt buffer I don't know how much will actually fit in the cache? Anyway it seems worth doing a couple of uniform checks to avoid the texture fetches anyway + all the ALUs indeed (just like we do with the dirt texture basically).

filament/src/PostProcessManager.cpp
pixelflinger: I think 1/16 res is a good choice for now (1/4 of the bloom buffer, which is already 1/4)

romainguy (approved):

PR #3925, opened 2021-05-10
Add reusable animator

zbai-sc:
- animation and model can be loaded separately as different assets
- we can create a reusable animator from an animation asset and apply it to model assets later


zbai-sc:
Created by mistake. Please delete.

PR #3942, opened 2021-05-11
Do not merge: switch to UBOs for froxels / records

bejado:


pixelflinger:
@bejado I think we can close this one?

bejado:
Yup, this was just for reference.

PR #3950, opened 2021-05-13
Allow using specific NDK version, for CI don't use latest but required

alexey-pelykh:
**Current behavior**

Build uses the newest NDK that's present in the system. CI build script installs the specific version that is ignored if the CI system has a newer NDK version.

**Suggested behavior**

Build uses the newest NDK that's present in the system _within the same major NDK release chain_ unless the NDK version is specified explicitly using `FILAMENT_NDK_VERSION` envvar. The envvar can specify the version partially like `22` or `22.0`. CI build script installs the specific version that is enforced for usage.

**Rationale**

When newer NDK comes out, older releases are not guaranteed to be buildable by it. Removing/installing newer NDK to get old releases building is a headache :)


romainguy:
Thanks for catching this. I need to think whether that's what we want. We should definitely fix the toolchain files to use an installed NDK that matches at least the major version we specify in our build script instead of a potentially newer one. I'm not convinced we want to always build with the newest NDK installed by default.

build/android/ndk.version
```diff
@@ -0,0 +1 @@
+22.1.7171670
```
romainguy, line 1: Thanks I've been meaning to do something like this for a while. The version number was in too many places.

alexey-pelykh:
> I'm not convinced we want to always build with the newest NDK installed by default.

I've kept this part to make less breaking change :) `build.sh` can be patched to select only lastest-within-major by default it explicit is not set.

build/toolchain-aarch64-linux-android.cmake
```diff
@@ -31,10 +31,13 @@ set(DIST_ARCH arm64-v8a)
 string(TOLOWER ${CMAKE_HOST_SYSTEM_NAME} HOST_NAME_L)
 file(TO_CMAKE_PATH $ENV{ANDROID_HOME} ANDROID_HOME_UNIX)
 
-file(GLOB NDK_VERSIONS LIST_DIRECTORIES true ${ANDROID_HOME_UNIX}/ndk/*)
-list(SORT NDK_VERSIONS)
-list(GET NDK_VERSIONS -1 NDK_VERSION)
-get_filename_component(NDK_VERSION ${NDK_VERSION} NAME)
+set(NDK_VERSION $ENV{FILAMENT_NDK_VERSION})
```
romainguy, line 34: So here I'd rather we matched the major version number and pick for instance the newest NDK 22 installed. We could also print a warning if the major version matches but not the rest of the version number.

romainguy (changes requested):

build/toolchain-aarch64-linux-android.cmake
alexey-pelykh: Should that be done in a CMake part or in `build.sh`, assuming CMake was not planned to be standalone as far as I've understood?

alexey-pelykh: Added some changes to address that.

romainguy: `build.sh` is just a helper, the CMake files can be used standalone.

build/toolchain-aarch64-linux-android.cmake
```diff
@@ -31,7 +31,7 @@ set(DIST_ARCH arm64-v8a)
 string(TOLOWER ${CMAKE_HOST_SYSTEM_NAME} HOST_NAME_L)
 file(TO_CMAKE_PATH $ENV{ANDROID_HOME} ANDROID_HOME_UNIX)
 
-file(GLOB NDK_VERSIONS LIST_DIRECTORIES true ${ANDROID_HOME_UNIX}/ndk/*)
+file(GLOB NDK_VERSIONS LIST_DIRECTORIES true ${ANDROID_HOME_UNIX}/ndk/$ENV{FILAMENT_NDK_VERSION}*)
```
romainguy, line 34: So this requires setting and exporting FILAMENT_NDK_VERSION. I'd rather build.sh sets is explicitly when invoking CMake with a -D or better yet, why not read the ndk.version file here? That would keep the CMake files "standalone".

romainguy (changes requested):

build/toolchain-aarch64-linux-android.cmake
alexey-pelykh: Reading `ndk.version` from here technically can be done, if `FILAMENT_NDK_VERSION` is not set, as a final fallback. Yet that way `ndk.version` would be read in 3 places. And since this code is in 4 files, for sure some refactoring would be needed to make that nice and tidy.

alexey-pelykh: In any case, what next steps you have in mind?

romainguy: I think for now just reading ndk.version would be best. Those 4 files contain a lot of duplicated code, another thing I've been meaning to fix for a long time. But it's not urgent as we rarely touch those files

alexey-pelykh: Okay, moved things around and did mini-cleanup

alexey-pelykh: Now it would work like that:

```
filament % FILAMENT_NDK_VERSION=28 ./build.sh -i -p android release
Error: Android NDK side-by-side version 28 or higher must be installed, exiting
```

but

```
filament % FILAMENT_NDK_VERSION=22 ./build.sh -i -p android release
```
and
```
filament % ./build.sh -i -p android release
```

would succeed if that NDK 22 is present.

Also, this would work for checking with minor NDK version changes

```
filament % FILAMENT_NDK_VERSION=22.2 ./build.sh -i -p android release
Error: Android NDK side-by-side version 22.2 or higher must be installed, exiting
```

build.sh
```diff
@@ -369,16 +368,9 @@ function ensure_android_build {
...
-        local ndk_version=$(ls "${ndk_side_by_side}" | sort -V | tail -n 1 | cut -f 1 -d ".")
-        if [[ ${ndk_version} -lt ${ANDROID_NDK_VERSION} ]]; then
-            echo "Error: Android NDK side-by-side version ${ANDROID_NDK_VERSION} or higher must be installed, exiting"
-            exit 1
-        fi
-    else
-        echo "Error: Android NDK side-by-side version ${ANDROID_NDK_VERSION} or higher must be installed, exiting"
+    # shellcheck disable=SC2012
+    if [[ -z $(ls "${ANDROID_HOME}/ndk/" | sort -V | grep "${FILAMENT_NDK_VERSION}") ]]; then
+        echo "Error: Android NDK side-by-side version ${FILAMENT_NDK_VERSION} or higher must be installed, exiting"
```
romainguy, line 373: The error message is incorrect now since the NDK version must match exactly.

build/android/build.sh
```diff
@@ -47,16 +44,14 @@ elif [[ "$LC_UNAME" == "darwin" ]]; then
 fi
 source `dirname $0`/../common/build-common.sh
 
-# Only update and install the NDK if necessary, as this can be slow
+# Unless explicitly specified, NDK version will be set to match exactly the required one
+export FILAMENT_NDK_VERSION=${FILAMENT_NDK_VERSION:-$(cat `dirname $0`/ndk.version)}
```
romainguy, line 48: Does this need to be exported here? The root build.sh already exports it

build.sh
```diff
@@ -104,12 +104,11 @@ function print_matdbg_help {
 }
 
 # Unless explicitly specified, NDK version will be selected as highest available version within same major release chain
-export FILAMENT_NDK_VERSION=${FILAMENT_NDK_VERSION:-$(cat `dirname $0`/build/android/ndk.version | cut -f 1 -d ".").}
+export FILAMENT_NDK_VERSION=${FILAMENT_NDK_VERSION:-$(cat `dirname $0`/build/android/ndk.version | cut -f 1 -d ".")}
```
romainguy, line 107: This shouldn't be exported but passed to CMake as a -D argument like we do for `FILAMENT_ENABLE_JAVA` for instance and various other options.

romainguy (changes requested):
Thank you I really appreciate your contribution. Just a few more comments :)

build.sh
alexey-pelykh: Not in all cases, but I'll adjust it to be "or compatible" instead of "or higher". Since FILAMENT_NDK_VERSION can be 20, 20.0, or 20.0.12312312313

build/android/build.sh
alexey-pelykh: Why not? root `build.sh` accepts it from the environment.

build.sh
romainguy: Ah good point, yeah that's nice. The error message just won't work with a fully specified version number but that's not a big deal. Although the problem with partial version numbers is that calls to sdkmanager in the other build.sh won't work (last I checked sdkmanager only accepts full version numbers now).

build.sh
alexey-pelykh: changed to avoid env

build.sh
alexey-pelykh: > with partial version numbers is that calls to sdkmanager

true, but it's not expected that CI script will work with non-qualified versions

alexey-pelykh:
Eagerly waiting for another set of comments :)

romainguy (approved):
Thank you! I greatly appreciate your patience and this awesome contribution. It's a great step towards cleaner builds.

alexey-pelykh:
Anything else I need to do on my end?

romainguy:
@alexey-pelykh Nope, I was just waiting for all the builds to go green!

PR #3955, opened 2021-05-13
Add URL input to Filament Remote Viewer

bejado:
This adds a URL input to the Remote Viewer that lets users specify a device IP to connect to. This is useful when connecting to iOS devices, which don't have the equivalent of "adb forward".

Screen Shot 2021-05-13 at 12 12 45 PM

I refactored the WebSocket logic into a `Connection` class to make the logic a bit simplier.

### DIFF
--- a/web/samples/remote.html
+++ b/web/samples/remote.html
@@ -40,6 +40,41 @@ body {
     font-size: 12px;
 }
 
+.connection-settings {
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    width: 100%;
+    height: 100%;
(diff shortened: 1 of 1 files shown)


prideout:
Very interesting, let me play with this a little before I approve it.  At first glance it seems fine.

prideout (approved):
I like the refactoring.  The invalid URL style is very cool.  And, the Android workflow is unchanged.  Nice work!

romainguy (approved):

pixelflinger (approved):
nice!

PR #3958, opened 2021-05-14
Use a UBO instead of a texture for froxels data structure

pixelflinger:
MATERIAL BREAKAGE: this change breaks materials

In order to use one less sampler, we now use a UBO instead of a texture
to store the "froxel records". Currently this UBO is limited to 16KiB
vs. 64KiB before with the texture.

We also handle running out of record space better, by using a predefined
record that has all the lights in the scene. This way the scene will be
rendered properly, albeit at a potentially large performance cost.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -3,7 +3,9 @@
 This file contains one line summaries of commits that are worthy of mentioning in release notes.
 A new header is inserted each time a *tag* is created.
 
-## v1.9.26 (currently main branch)
+## v1.10.0 (currently main branch)
+
+- engine: User materials can now use 9 samplers instead of 8 [⚠️ **Material breakage**].
 
 ## v1.9.25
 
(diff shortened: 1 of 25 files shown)


prideout (changes requested):
Yay, another sampler!!

MaterialEnums.h has a `MATERIAL_VERSION` that needs to be bumped.

By the way, why did we combine the two binding counts?  Seems like the number of sampler bindings could be different from the number of UBO bindings.

pixelflinger:
> Yay, another sampler!!
> 
> MaterialEnums.h has a `MATERIAL_VERSION` that needs to be bumped.
> 
> By the way, why did we combine the two binding counts? Seems like the number of sampler bindings could be different from the number of UBO bindings.

I combined them because there is a single enum. I think if we wanted to keep them separate we should use two different enums, otherwise it's a bit confusing, also currently if you set them to different values, this would create a memory corruption if you're not careful.

prideout (approved):
> I combined them because there is a single enum.

Makes sense!  Btw some glTF models have multiple light sources if you want an additional way of testing the froxelizer.

PR #3964, opened 2021-05-14
add java API for Texture::Builder::import()

pixelflinger:
deprecate Stream::stream(intptr_t).

This method was needed for ARCore back in the days, but there is now a
zero-copy way to achieve the same thing. This API shouldn't be used anymore.

### DIFF
--- a/filament/include/filament/Stream.h
+++ b/filament/include/filament/Stream.h
@@ -131,7 +131,9 @@ public:
          * @return This Builder, for chaining calls.
          *
          * @see Texture::setExternalStream()
+         * @deprecated this method existed only for ARCore which doesn't need this anymore, use Texture::import() instead.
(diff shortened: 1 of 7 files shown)


filament/include/filament/Stream.h
```diff
@@ -131,7 +131,9 @@ class UTILS_PUBLIC Stream : public FilamentAPI {
          * @return This Builder, for chaining calls.
          *
          * @see Texture::setExternalStream()
+         * @deprecated this method existed only for ARCore which doesn't need this anymore, use Texture::import() instead.
          */
+        [[deprecated]]
```
bejado, line 136: `UTILS_DEPRECATED` ?

bejado (approved):

PR #3966, opened 2021-05-14
Add a FILAMENT_SUPPORTS_OPENGL CMake option

bejado:


BUILDING.md
```diff
@@ -84,6 +84,7 @@ The following CMake options are boolean options specific to Filament:
 - `FILAMENT_ENABLE_JAVA`:          Compile Java projects: requires a JDK and the JAVA_HOME env var
 - `FILAMENT_ENABLE_LTO`:           Enable link-time optimizations if supported by the compiler
 - `FILAMENT_BUILD_FILAMAT`:        Build filamat and JNI buildings
+- `FILAMENT_SUPPORTS_OPENGL`:      Include the OpenGL backend
```
romainguy, line 87: Maybe we should test in the CMake that one of OPENGL/METAL/VULKAN is set. Otherwise it's not going to produce a very useful Filament :)

romainguy (approved):

BUILDING.md
pixelflinger: but it'll be very small -- you get what you pay for! :-)

pixelflinger (approved):

BUILDING.md
bejado: I think technically you could still use the NOOP driver

PR #3971, opened 2021-05-17
getExtras() API to expose GLTF extras

alexey-pelykh:


libs/gltfio/include/gltfio/FilamentInstance.h
```diff
@@ -49,6 +55,12 @@ class FilamentInstance {
     /** Gets the transform root for the instance, which has no matching glTF node. */
     utils::Entity getRoot() const noexcept;
 
+    /** Gets the NameComponentManager label for the given entity, if it exists. */
+    const char* getName(utils::Entity) const noexcept;
+
+    /** Gets the application-specific extras loaded from asset for specific entity, if such exist. */
+    std::string getExtras(utils::Entity entity) const noexcept;
```
romainguy, line 62: Please don't expose STL types. Return const char* instead, or our own CString (although I don't know if it's public).

alexey-pelykh: Removed

android/gltfio-android/src/main/java/com/google/android/filament/gltfio/FilamentInstance.java
```diff
@@ -66,6 +67,20 @@ void clearNativeObject() {
         return result;
     }
 
+    /**
+     * Gets the <code>NameComponentManager</code> label for the given entity, if it exists.
+     */
+    public String getName(@Entity int entity) {
```
prideout, line 73: Can't  users just do `instance->getOwner()->getName(..)`?

libs/gltfio/src/FFilamentAsset.h
```diff
@@ -268,9 +270,11 @@ struct FFilamentAsset : public FilamentAsset {
     std::vector<TextureSlot> mTextureSlots;
     std::vector<const char*> mResourceUris;
     NodeMap mNodeMap; // unused for instanced assets
+    ReverseNodeMap mReverseNodeMap; // unused for instanced assets
     std::vector<std::pair<const cgltf_primitive*, filament::VertexBuffer*> > mPrimitives;
     MatInstanceCache mMatInstanceCache;
     MeshCache mMeshCache;
+    tsl::robin_map<const void *, std::string> mExtras;
```
prideout, line 277: Instead of a entity=>node map a node=>string map, can't we just have a entity=>string map?

android/gltfio-android/src/main/java/com/google/android/filament/gltfio/FilamentInstance.java
alexey-pelykh: There's no `getOwner()` in Android

libs/gltfio/src/FFilamentAsset.h
alexey-pelykh: Not really, extras are retrieved before entities are created.

alexey-pelykh:
@romainguy @prideout any feedback to address?

libs/gltfio/src/FFilamentAsset.h
prideout: > @romainguy @prideout any feedback to address?

It would be good to fix the "two maps" thing somehow.  Note that the cgltf node pointers are transient and become invalid when `releaseSourceData` is called, which clears out `mNodeMap` but not your new reverse node map.

Maybe we need two maps (or two passes) during the loading process, but surely it's possible to have only one map as a member variable.

alexey-pelykh: I'll double-check how it handles the described case, but maybe I'll have to add a third "squashed" map and clean-up others after loading is complete

alexey-pelykh: Okay, got your point, a very valid one.

alexey-pelykh: There are still two maps, but one is used as cache and later wiped out after loading is complete.

prideout (changes requested):
Can we use something other than `shared_ptr<const char>`?  There's no shared ownership going on here.  Maybe `unique_ptr` or `utils::CString`.

alexey-pelykh:
@prideout there is? When instances are used, many entities refer to same extra data via shared ptr. Otherwise I'd use simple string. Alternative is shared ptr to a string object but there's no point in having object

romainguy:
It may be shared by objects but the shared ownership doesn't help with cleanup here since it's in a map. It might as well be a const char*. The only time the shared_ptr is useful here is when destroying `mEntityToExtras` since nothing seems to be ever removed from the map until destruction? We might as well just loop over the entries in the destructor and free memory so we don't have to pay the overhead of shared_ptr. An alternative, like @prideout mentioned, is to use unique_ptr just for the purpose of memory management and give raw pointers to the entities. The public API `getExtras(Entity)` returns a const char* anyway.

alexey-pelykh:
Good point that destroying instance needs to remove from the map as well

alexey-pelykh:
@romainguy:

> nothing seems to be ever removed from the map until destruction?

correct, I've double-checked that nothing is removed from `mEntityToExtras` and neither it should unless someday dynamic instance destruction is implemented

> We might as well just loop over the entries in the destructor and free memory so we don't have to pay the overhead of shared_ptr.

Then it's either std::vector of pointers or strings like `mExtras` and `mEntityToExtras` that points to them to avoid duplicate usage of memory by number of instances. So the overhead is either shared_ptr or extra std::vector - what you prefer better?

alexey-pelykh:
@prideout @romainguy bump, need your further input

romainguy:
@prideout Are you happy with this change?

prideout:
> @prideout Are you happy with this change?

I'll take a look today.

prideout:
This has too much memory overhead for assets that have no extras whatsoever (it adds a map entry per node).  I'm working on a new PR to replace this.

PR #3972, opened 2021-05-17
getChannelsCount() / getTargets() API

alexey-pelykh:


libs/gltfio/include/gltfio/Animator.h
```diff
@@ -20,6 +20,8 @@
 #include <gltfio/FilamentAsset.h>
 #include <gltfio/FilamentInstance.h>
 
+#include <vector>
```
romainguy, line 23: Please don't expose STL types in public headers

romainguy (commented):
Could you please explain the purpose of this change?

prideout:
I can't think of any reason for clients to require access to the low-level cgltf data.  There is an existing backdoor (see `FilamentAsset::getSourceAsset()`) and I don't think we need another one.

alexey-pelykh:
@romainguy @prideout how otherwise we can sync ex-Sceneform scene graph with transforms?
```
    private fun updateJointNodesFromFilament(animationInstance: AnimationInstance, node: Node) {
        if (node.filamentEntity in animationInstance.targets) {
            node.updateTransformFromFilament()
        }

        for (child in node.children) {
            updateJointNodesFromFilament(animationInstance, child)
        }
    }
```

prideout:
FilamentAsset has a `getEntities()` method that returns an entity for every glTF node, including joint nodes.  Every entity that it returns has a Transform component.

alexey-pelykh:
@prideout how to discover if that's a joint node from that list?

prideout:
Joint nodes aren't special, they are just nodes with transforms.  You can do any of the following: (a) synchronize _all_ transform nodes instead of a subset (b) access the underlying cgltf data using getSourceAsset()  (c) avoid synchronization by using Filament's ECS directly, rather than duplicating the transform data.

alexey-pelykh:
(a) is inefficient
(b) is backdorish instead of being normal API and usable only in cpp
(c) kind-a defies the purpose

In the end, Filament does not have a suitable API to get only joints, thus there's a PR.

prideout:
gltfio is just a loader library and there are many parts of the glTF structure that its API does not expose, so I'm struggling to understand why the joints are special.  For example, gltfio is very unsuitable for building a model editor.

Can you help us understand why you need to synchronize the joint transforms between Filament and an outside system?

FabianRueckert:
@prideout We are managing our own scene graph, like Sceneform did, mirroring Filament's scene graph. One of the reasons for that is that it allows us to store translation/rotation/scale separately and only build the transform matrix as a last step.
This is to avoid accumulating rounding errors and distinguish a rotation by 180 degrees from a reflection along two axes.

prideout:
Interesting!  Thanks for helping me understand this.  Do you have an example model that reveals an issue caused by the fact that we do not store separate translation/rotation/scale transforms for each node?

FabianRueckert:
It can happen with all models due to incremental rounding errors accumulating over time (eg. rotating n degrees per frame)
Here is a good explanation: https://gamedev.stackexchange.com/questions/114282/do-matrix-manipulations-for-3d-graphics-ever-produce-singular-matrices/114284#114284

prideout:
At each time step, our Animator decomposes the matrix into three components, then completely overwrites one or more of those components before re-assembling the matrix.

Since the individual transform components are clobbered at each time step, I'm not confident that incremental rounding errors can arise.  An example model would be great.

FabianRueckert:
> At each time step, our Animator decomposes the matrix into three components, then completely overwrites one or more of those components before re-assembling the matrix. The new transform component is derived from original sampler values.

Yes in that case there is no problem with rounding errors inside the Animator.

However such rounding errors could occur when transforming the matrices of Transformmanager manually, for example when programmatically rotating an object in a game and incrementally transforming the current frame's transform.
This is why we are managing our own scene graph on top of Filament's scene graph.
And because we have that, we need these API functions on the Animator to update our scene graph after Filament has modified it.

prideout:
Given your need for synchronization, I'm starting to understand the rationale for `getTargets`.

What about getChannelsCount / getSkinCount / getJoints / getSkinTargets?  What is the need for those?

PR #3973, opened 2021-05-17
Material::Builder::name() API

alexey-pelykh:


romainguy:
Could you explain the reasoning behind this change? A material's name can already be set with `MaterialBuilder` when creating it. This seems to be here to override the name when loading from a binary blob? If so, why add the ability to override the name but not anything else? Either way, `Material::Builder` is not intended to be used to modify the material blobs.

alexey-pelykh:
> A material's name can already be set with MaterialBuilder when creating it

We're not using Uber-shader or on-demand shader compilation because of the size constraints. See #3975

> This seems to be here to override the name when loading from a binary blob?

Correct, for cases when the same shader blob is used by multiple materials where the name is the only difference.

> Either way, Material::Builder is not intended to be used to modify the material blobs.

But blob is not changed, and names baked into blobs is a bit debatable, imho.

romainguy:
> Correct, for cases when the same shader blob is used by multiple materials where the name is the only difference.

If the shader blob is the same, the name should be the same as well.

romainguy:
It's the name of the material, it's useful for debugging or for UI? Also you should not load the same material binary several times if possible. It's expensive. That's why we have material instances.

PR #3974, opened 2021-05-17
GetBackendHandle() API

alexey-pelykh:


romainguy:
Thank you but those handles are a private implementation detail and should not be exposed via public APIs.

alexey-pelykh:
True, though there are cases when access to OpenGL handle is needed outside of the filament itself 🤔

romainguy:
Filament might not have a GL backend, we do not want to expose GL specific things.

alexey-pelykh:
Likewise it was implemented for Vulkan as well, so it's not GL specific..

PR #3975, opened 2021-05-17
External material provider

alexey-pelykh:


romainguy (commented):
Please describe the PR. It's a lot of code and exposes things that I'm not sure we want to expose.

alexey-pelykh:
We're not using uber-shader or shader builder due to AAR size concerns, so instead, shaders are precompiled and are bundled side-by-side with GLTF, and we require a way to provide materials loaded "externally" and use `Material::Builder::payload`

prideout:
This is definitely useful, but we were planning on exposing this in a different way that is a bit more 1:1 with the existing C++ functionality.

More specifically, we would like to change `MaterialProvider` into an interface class (similar to its existing C++ counterpart) and clients can write their own classes that implement this interface.  Then, they can simply pass an instance of their custom provider into the ModelViewer constructor.

PR #3977, opened 2021-05-17
MaterialInstance.getParameter() API

alexey-pelykh:


FabianRueckert:
@googlebot I consent.

romainguy:
Could you please explain the reason of this change?

android/filament-android/src/main/cpp/MaterialInstance.cpp
```diff
@@ -260,6 +441,48 @@ Java_com_google_android_filament_MaterialInstance_nSetParameterTexture(
...
+
+    const char *name = env->GetStringUTFChars(name_, 0);
+    auto success = instance->getParameter(name, texture, sampler);
+    env->ReleaseStringUTFChars(name_, name);
+
+    if (!success) {
+        return NULL;
+    }
+
+    const auto jTextureClass = env->FindClass("com/google/android/filament/Texture");
```
romainguy, line 460: These need to be looked up once at initialization, not every time the function is called

android/filament-android/src/main/java/com/google/android/filament/MaterialInstance.java
```diff
@@ -232,6 +299,28 @@ public void setParameter(@NonNull String name,
         nSetParameterTexture(getNativeObject(), name, texture.getNativeObject(), sampler.mSampler);
     }
 
+    public static class TextureAndSampler {
```
romainguy, line 302: Missing documentation, also the name is awkward. I'd prefer to have separate getTexture/getSampler calls.

alexey-pelykh: > Missing documentation

Valid point

> also the name is awkward

Open to suggestions :)

> I'd prefer to have separate getTexture/getSampler calls.

It was done as one call to mirror how it's being set, yet see no reason why not to change that and drop the structure

pixelflinger:
Not having getters for `MaterialInstance` parameters was by design -- we didn't want to *have to* keep a copy on the CPU side at all times. Another problem is that this adds tons of code for little benefit. Often, applications already have a representation of the parameters they set and don't need to be able to query them.
The increased code size is a real issue for us -- we'd rather spend that budget somewhere else.

Can give more explanations as to why these APIs are useful or needed (meaning there is no other way to achieve what you need)?

We are (probably) going to add an API to clone a `MaterialInstance`, would that help in any way?

jwwalker:
@pixelflinger Personally, I was wishing there was an API to get just the names, not the values, of the parameters.  Or even just to query whether the material instance has a parameter of a given name.  This would be for debugging, to make sure I've set values for all the parameters I defined.

romainguy:
@jwwalker You mean like `Material::hasParameter()` and `Material::getParameters()`? :)

jwwalker:
@romainguy Oh, duh!  Yes, it makes perfect sense that these would be methods of `Material` and not `MaterialInstance`.  Thanks for the quick correction.

PR #3978, opened 2021-05-17
KTXLoader.getSphericalHarmonics

alexey-pelykh:
Co-authored-by: Fabian Rueckert <fabi.rueckert@gmail.com>


FabianRueckert:
@googlebot I consent.

android/filament-utils-android/src/main/java/com/google/android/filament/utils/KtxLoader.kt
```diff
@@ -76,7 +76,14 @@ object KtxLoader {
         return Skybox(engine, nativeSkybox)
     }
 
+    fun getSphericalHarmonics(buffer: Buffer): FloatArray {
+        val floatBuffer = FloatBuffer.allocate(3 * 9)
```
romainguy, line 80: Why go through a `FloatBuffer`, why not use a regular array?

romainguy (commented):
Could you explain why this change is needed?

FabianRueckert:
> Could you explain why this change is needed?

To be able to separately load a KTX texture and spherical harmonics, that may then be modified on the client side, for example modulated using AR light estimation and used to create a filament indirect light from these modified values.
@romainguy

android/filament-utils-android/src/main/cpp/Utils.cpp
```diff
@@ -73,6 +73,13 @@ static jlong nCreateSkybox(JNIEnv* env, jclass,
     return (jlong) Skybox::Builder().environment(cubemap).showSun(true).build(*engine);
 }
 
+static void nGetSphericalHarmonics(JNIEnv* env, jclass, jobject javaBuffer, jint remaining, jobject outBuffer) {
+    AutoBuffer buffer(env, javaBuffer, remaining);
+    KtxBundle bundle((const uint8_t*) buffer.getData(), buffer.getSize());
+    AutoBuffer outAutoBuffer(env, outBuffer, 3 * 9, true);
+    bundle.getSphericalHarmonics((float3*) outAutoBuffer.getData());
```
romainguy, line 80: What happens btw if the KTX doesn't contain spherical harmonics here?

romainguy (changes requested):

prideout:
LGTM if Romain's comments are addressed and it is rebased to the latest main (we named KtxLoader to KTXLoader)

prideout (approved):
LGTM if Romain's comments are addressed and it is rebased to the latest main (we named KtxLoader to KTXLoader)

android/filament-utils-android/src/main/java/com/google/android/filament/utils/KtxLoader.kt
alexey-pelykh: Changed to `FloatArray`

android/filament-utils-android/src/main/cpp/Utils.cpp
alexey-pelykh: Now it would return null. Btw, `nCreateIndirectLight` also does no checks.

PR #3979, opened 2021-05-17
LightManager: cone angles API

alexey-pelykh:


FabianRueckert:
@googlebot I consent.

filament/src/components/LightManager.cpp
```diff
@@ -482,6 +482,13 @@ float LightManager::getSpotLightOuterCone(Instance i) const noexcept {
     return upcast(this)->getSpotParams(i).outerClamped;
 }
 
+float LightManager::getSpotLightInnerConeApproximate(Instance i) const noexcept {
```
romainguy, line 485: Why is this called "approximate"?

FabianRueckert: Because this function is recomputing the inner spot angle based on the current `SpotParams`. This results in some rounding errors for retrieving the inner cone angle.
The alternative would be to store the inner cone angle inside the `SpotParams` when it's being set.

romainguy: If that data is not passed to the shaders, that's fine.

pixelflinger (changes requested):
Java APIs and C++ API should be consistant.
So please use `getSpotLightOuterCone` everywhere on the java side.
Also use `getSpotLightInnerCone` and document that the value may differ slightly from the value set.
Also don't implement `getSpotLightInnerCone` in `LightManager`, this must be in `FLightManager`, please follow the same pattern than for the other methods.

filament/src/components/LightManager.cpp
```diff
@@ -482,6 +482,13 @@ float LightManager::getSpotLightOuterCone(Instance i) const noexcept {
...
+    float cosOuter = fast::cos(getSpotLightOuterCone(i));
```
pixelflinger, line 486: use `std::cos` here -- since we're computing some value back, might as well try to be precise.

pixelflinger (changes requested):

filament/src/components/LightManager.cpp
alexey-pelykh: 👍

alexey-pelykh:
Thanks for the feedback, the PR has been amended.

pixelflinger (approved):

alexey-pelykh:
@romainguy co-authorship got lost somehow, would it be possible to fix that?

PR #3980, opened 2021-05-17
TextureHelper: bitmap internal buffer API

alexey-pelykh:


FabianRueckert:
@googlebot I consent.

romainguy:
Thanks for the PR but this change adds functionality that is not specific to Filament and should therefore be in a different library. It also has arbitrary limitations (why only `ANDROID_BITMAP_FORMAT_RGBA_8888` bitmaps?). Note that there is already an API to manipulate bitmaps via buffers (`Bitmap.copyTo/FromBuffer`) on Android.

PR #3981, opened 2021-05-17
View: culling results access API

alexey-pelykh:


FabianRueckert:
@googlebot I consent.

romainguy:
Thank you but this change exposes implementation details (`getVisibilityMasks()`, `getRenderableInstances()`) that should not be exposed.

PR #3982, opened 2021-05-17
Indirect light KTX

alexey-pelykh:


romainguy:
Thank you but this change duplicates existing functionality of filament-utils and adds a large dependency to Filament itself (`libimage`). In addition it adds something we do not with to have in Filament, I/O operations. We strive to keep the core library as small as possible.

alexey-pelykh:
@romainguy understood, fair point.

PR #3987, opened 2021-05-18
matc: detect missing end brace.

prideout:
matc was failing to report certain kinds of syntax errors and would
read out-of-bounds memory.

This change casuses the flare material to fail.


romainguy (approved):

bejado (approved):
Nice

PR #3990, opened 2021-05-18
Vulkan: change depth layout to permit feedback loops.

prideout:
Vulkan does not permit the GENERAL layout when the depth buffer that is
currently being used for depth testing is also sampled from.

This (finally) fixes the last validation error that we were silencing,
which would occur only when SSAO was enabled.

### DIFF
--- a/filament/backend/src/vulkan/VulkanContext.h
+++ b/filament/backend/src/vulkan/VulkanContext.h
@@ -56,6 +56,7 @@ struct VulkanRenderPass {
     VkRenderPass renderPass;
     uint32_t subpassMask;
     int currentSubpass;
+    VulkanTexture* depthFeedback;
 };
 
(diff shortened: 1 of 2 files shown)


bejado (approved):

romainguy (approved):

pixelflinger:
Ah right, because we use depth testing during the ssao path to automatically reject the skybox. This was a big optimization.
This is not a feedback loop in that case though -- we're not writing to the depth buffer.

prideout:
Agreed that the term "feedback" is debatable here, I used it because the GL spec uses this language, e.g. section 9.3 in the ES 3.2 spec says:

```
note that even if writing to a given texture is disabled (e.g. via the color, depth, or stencil writemasks) during rendering, a feedback loop will still occur.
```

PR #3998, opened 2021-05-19
Add HDRLoader for JVM clients.

prideout:
This adds support for decoding HDR files, dropping them into the remote
web page, and sending them over the wire.  However the decoded results
are not yet passed into IBLPrefilterContext, that will be subsequent
PR.

Rename KtxLoader to KTXLoader.

The remote UI has been tweaked a bit, so that it shows a more helpful
message when disconnected:

disconnected
connected

### DIFF
--- a/android/filament-utils-android/CMakeLists.txt
+++ b/android/filament-utils-android/CMakeLists.txt
@@ -2,6 +2,7 @@ cmake_minimum_required(VERSION 3.19)
 project(filament-utils-android)
 
 set(FILAMENT_DIR ${FILAMENT_DIST_DIR})
+set(IMAGEIO_DIR ../../libs/imageio)
 
 add_subdirectory(${CMAKE_CURRENT_SOURCE_DIR}/../gltfio-android ${CMAKE_CURRENT_BINARY_DIR}/gltfio-android)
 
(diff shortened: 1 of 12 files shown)


android/filament-utils-android/CMakeLists.txt
```diff
@@ -38,6 +39,12 @@ add_library(filament-utils-jni SHARED
         ../common/NioUtils.cpp
 )
 
+target_include_directories(filament-utils-jni PRIVATE
+        ${FILAMENT_DIR}/include
+        ..
+        ../../third_party/stb
```
romainguy, line 45: Any chance we could reuse our own HDR decoder from libimage? Even if we compile it separately. I'm asking because there are some subtle differences between HDR decoders and it would be good to match cmgen.

android/filament-utils-android/src/main/java/com/google/android/filament/utils/HDRLoader.kt
```diff
@@ -0,0 +1,50 @@
...
+import com.google.android.filament.Texture
+
+import java.nio.Buffer
+
+/**
+ * Utility for decoding an HDR file and producing a Filament texture.
+ */
+object HDRLoader {
+    class Options {
+        var desiredFormat: Texture.InternalFormat = Texture.InternalFormat.RGB16F
```
romainguy, line 29: `: Texture.InternalFormat` is redundant here

romainguy (approved):

bejado (approved):

android/filament-utils-android/CMakeLists.txt
prideout: Good point I'll give it a try.  It would give us EXR reading capability too.

android/filament-utils-android/src/main/java/com/google/android/filament/utils/HDRLoader.kt
prideout: Good catch

android/filament-utils-android/CMakeLists.txt
romainguy: Well I'm wondering if we want to link the entire libimage, it's not small (and supports PNG that we don't care about here). We could however compile in only our own HDRDecoder class.

pixelflinger (approved):

PR #4006, opened 2021-05-21
added MaterialInstance::duplicate(...)

pixelflinger:
It's now possible to duplicate a MaterialInstance.

### DIFF
--- a/filament/src/MaterialInstance.cpp
+++ b/filament/src/MaterialInstance.cpp
@@ -177,26 +177,45 @@ template UTILS_PUBLIC void MaterialInstance::setParameter<mat4f>   (const char* 
 
 FMaterialInstance::FMaterialInstance() noexcept = default;
 
-FMaterialInstance::FMaterialInstance(FEngine& engine, FMaterial const* material, const char* name) :
-        mName(name) {
+FMaterialInstance::FMaterialInstance(FEngine& engine,
+        FMaterialInstance const* other, const char* name)
+        : mMaterial(other->mMaterial),
+          mPolygonOffset(other->mPolygonOffset),
+          mCulling(other->mCulling),
+          mColorWrite(other->mColorWrite),
+          mDepthWrite(other->mDepthWrite),
+          mDepthFunc(other->mDepthFunc),
+          mScissorRect(other->mScissorRect),
+          mName(name ? CString(name) : other->mName) {
+
     FEngine::DriverApi& driver = engine.getDriverApi();
+    FMaterial const* const material = other->getMaterial();
 
     if (!material->getUniformInterfaceBlock().isEmpty()) {
-        mUniforms.setUniforms(material->getDefaultInstance()->getUniformBuffer());
+        mUniforms.setUniforms(other->getUniformBuffer());
         mUbHandle = driver.createUniformBuffer(mUniforms.getSize(), backend::BufferUsage::DYNAMIC);
     }
 
     if (!material->getSamplerInterfaceBlock().isEmpty()) {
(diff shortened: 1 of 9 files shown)


prideout:
#3977 #3992

android/samples/sample-gltf-viewer/src/main/java/com/google/android/filament/gltf/MainActivity.kt
```diff
@@ -70,7 +71,7 @@ class MainActivity : Activity() {
 
         doubleTapDetector = GestureDetector(applicationContext, doubleTapListener)
 
-        modelViewer = ModelViewer(surfaceView)
+        modelViewer = ModelViewer(surfaceView, Engine.create(Engine.Backend.VULKAN))
```
prideout, line 74: Unintentional?

prideout (changes requested):

filament/src/MaterialInstance.cpp
```diff
@@ -177,24 +177,39 @@ template UTILS_PUBLIC void MaterialInstance::setParameter<mat4f>   (const char*
...
-        mName(name) {
+FMaterialInstance::FMaterialInstance(FEngine& engine,
+        FMaterialInstance const* other, const char* name)
+        : mPolygonOffset(other->mPolygonOffset),
+          mCulling(other->mCulling),
+          mColorWrite(other->mColorWrite),
+          mDepthWrite(other->mDepthWrite),
+          mDepthFunc(other->mDepthFunc),
+          mScissorRect(other->mScissorRect),
+          mName(name ? CString(name) : other->mName) {
```
prideout, line 188: Will these get clobbered later when `initialize()` is called?

filament/src/MaterialInstance.cpp
```diff
@@ -177,24 +177,39 @@ template UTILS_PUBLIC void MaterialInstance::setParameter<mat4f>   (const char*
...
         mUbHandle = driver.createUniformBuffer(mUniforms.getSize(), backend::BufferUsage::DYNAMIC);
     }
 
     if (!material->getSamplerInterfaceBlock().isEmpty()) {
-        mSamplers.setSamplers(material->getDefaultInstance()->getSamplerGroup());
+        mSamplers.setSamplers(other->getSamplerGroup());
         mSbHandle = driver.createSamplerGroup(mSamplers.getSize());
     }
 
     initialize(material);
```
jimbz, line 203: Inside `initialize` there are calls like `setMaskThreshold(material->getMaskThreshold());` that to my understanding modify the uniforms. If we do something like this:
```C++
MaterialInstance *mi = material->createInstance("original");
mi->setMaskThreshold(0.5f);
MaterialInstance duplicate = mi->duplicate("duplicate");
```
wouldn't we expect the mask threshold of the duplicate to be the overridden value instead of the material's default?

android/samples/sample-gltf-viewer/src/main/java/com/google/android/filament/gltf/MainActivity.kt
pixelflinger: oh damn yes. thanks!

filament/src/MaterialInstance.cpp
pixelflinger: So I changed initialize() so that doesn't happen.

filament/src/MaterialInstance.cpp
pixelflinger: yes very good point, I suppose we don't want the value from the material in that case.

prideout (approved):

PR #4007, opened 2021-05-21
Add getPlatform API to Engine, move functionality into MetalPlatform

bejado:
This change allows access to the Engine's underlying Platform object. This is useful for Metal clients to access a new method on `MetalPlatform`, `createAndEnqueueCommandBuffer`.


filament/backend/include/private/backend/MetalPlatform.h
```diff
@@ -17,28 +17,36 @@
...
-
-class PlatformMetal final : public backend::MetalPlatform {
-public:
-    backend::Driver* createDriver(void* sharedContext) noexcept override;
+    Driver* createDriver(void* sharedContext) noexcept override;
     int getOSVersion() const noexcept override { return 0; }
-    ~PlatformMetal() override;
+
+    virtual id<MTLDevice> createDevice() noexcept;
+    virtual id<MTLCommandQueue> createCommandQueue(id<MTLDevice> device) noexcept;
```
pixelflinger, line 36: Maybe some documentation comments?

pixelflinger (approved):

filament/backend/include/private/backend/MetalPlatform.h
prideout: One interesting thing about this is API is that these calls do not occur in the driver thread.

Presumably this isn't a problem in Metal because there's no notion of "current context" like here is in GL.

(In Vulkan there's no notion of "current context" either, but users are still expected to to avoid calling certain methods from more than one thread.)

Just thought I'd mention it in case there's a gotcha we haven't thought of yet.

prideout: I think `createDevice` is used by our driver and it's not really intended for outside parties, right?  Maybe we need a small note to warn users about that.

filament/backend/include/private/backend/MetalPlatform.h
```diff
@@ -17,28 +17,36 @@
...
+
+    /**
+     * Obtain a MTLCommandBuffer enqueued on this Platform's MTLCommandQueue. The command buffer is
+     * guaranteed to execute before all subsequent command buffers.
```
prideout, line 40: Does "subsequent command buffers" refer to the command buffers created by the method, or the ones created privately in the driver?  Or both?

bejado: Both. Good question, I'll update the comment.

prideout (approved):

filament/backend/include/private/backend/MetalPlatform.h
bejado: No, these calls happen on the driver thread (they're called from the `MetalDriver` constructor. The idea is that outside parties could override `createDevice` and/or `createCommandQueue` if they wanted to.. the platform might want to select a specific device, for example.

Issue #3321, opened 2020-11-25
macOS main thread checker violations


jwwalker:
**Describe the bug**
When trying to use Filament under the Xcode debugger, the method `PlatformCocoaGL::makeCurrent` produces a runtime error message

    Main Thread Checker: UI API called on a background thread: -[NSView bounds]

and if I let it run further it also complains that `-[NSView convertRectToBacking:]` should only be used on the main thread.  This is happening in a thread called `Engine::loop`.

I was trying to get Filament to work on macOS without SDL by imitating the project at iOS/samples/hello-triangle.

**Desktop:**
 - OS: macOS Big Sur 11.1 Beta (20C5048k)
 - GPU: Radeon Pro 560X 4 GB
 - Backend: OpenGL

**Additional context**
Xcode 12.3 beta

bejado:
This should have been fixed with https://github.com/google/filament/pull/3273. You may have to wait for the fix to make it into the next release or build Filament yourself.

jwwalker:
@bejado No, I did build from source.  That revision made thread-safety changes in `PlatformCocoaGLImpl::updateOpenGLContext`, whereas I am talking about `PlatformCocoaGL::makeCurrent`.

bejado:
`makeCurrent` calls `updateOpenGLContext`, which is why I suspect it's the same bug.

jwwalker:
@bejado Looks like a similar bug, but not the same.  Comments in `updateOpenGLContext` say it's trying to make sure that the methods `-[NSOpenGLContext setView:]` and `-[NSOpenGLContext update]` are called only from the main thread, and that this is a "hard requirement" as of macOS 10.15.  Whereas my bug is about methods of `NSView`, and maybe it's not a hard requirement.

bejado:
Modifying our calls to `convertRectToBacking` and `bounds` so that they happen synchronously on the main thread drastically hurts performance. Given that these don't seem to be hard requirements and that OpenGL is deprecated on macOS anyway, I don't think this will be a priority for us to fix.

jwwalker:
@bejado Of course I wouldn't have expected you to just synchronously do it on the main thread.  A more reasonable approach would be to subscribe to notifications of frame geometry updates, and when such a notification is received (on the main thread), cache the bounds.  It looks like `NSViewFocusDidChangeNotification` would be the best notification, though for some reason it's deprecated.

PR #4011, opened 2021-05-22
fix macOS main thread checker warnings with OpenGL

jwwalker:
Fixes issue #3321 

Xcode's main thread checker reported runtime warnings about several `NSView` methods being called off the main thread when using an OpenGL back end on macOS.  This fix uses notifications to keep track of and cache the necessary data on the main thread.


filament/backend/src/opengl/PlatformCocoaGL.mm
```diff
@@ -35,9 +35,15 @@
 using namespace backend;
 
 struct CocoaGLSwapChain : public Platform::SwapChain {
+    CocoaGLSwapChain( NSView* inView );
+    ~CocoaGLSwapChain() noexcept;
+
     NSView* view;
     NSRect previousBounds;
     NSRect previousWindowFrame;
+    __strong NSMutableArray< id<NSObject> >* observers;
```
romainguy, line 44: Please fix formatting: remove extra empty spaces (same above in the constructor)

filament/backend/src/opengl/PlatformCocoaGL.mm
```diff
@@ -35,9 +35,15 @@
...
+    NSRect curBounds;
```
romainguy, line 45: Spell current don't use abbreviations

filament/backend/src/opengl/PlatformCocoaGL.mm
```diff
@@ -47,6 +53,78 @@
...
+                    usingBlock: noteHandler];
+                [observers addObject: observer];
+                
+                aView = aView.superview;
+            }
+        }
+    });
+}
+
+CocoaGLSwapChain::~CocoaGLSwapChain() noexcept
```
romainguy, line 121: Please put { at the end of the previous line

filament/backend/src/opengl/PlatformCocoaGL.mm
```diff
@@ -47,6 +53,77 @@
...
+CocoaGLSwapChain::CocoaGLSwapChain( NSView* inView )
+        : view(inView)
+        , previousBounds(NSZeroRect)
+        , previousWindowFrame(NSZeroRect)
+        , observers([NSMutableArray array])
+        , currentBounds(NSZeroRect)
+        , currentWindowFrame(NSZeroRect) {
+    __weak NSView* weakView = view;
+    
+    void (^noteHandler)(NSNotification *note) = ^(NSNotification *note) {
```
bejado, line 65: This block captures the `this` pointer implicitly. Can you make that more obvious by prefixing the member var access with `this->`?

I think the `SwapChain` object could get destroyed before this notification block executes, in which case we'd crash.

filament/backend/src/opengl/PlatformCocoaGL.mm
```diff
@@ -47,6 +53,77 @@
...
+        }
+    };
+    
+    // Various methods below should only be called from the main thread:
+    // -[NSView bounds], -[NSView convertRectToBacking:], -[NSView window],
+    // -[NSWindow frame], -[NSView superview],
+    // -[NSView setPostsFrameChangedNotifications:],
+    // -[NSView setPostsBoundsChangedNotifications:]
+    dispatch_async(dispatch_get_main_queue(), ^(void) {
+        if (weakView != nil) {
```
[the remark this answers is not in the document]
jwwalker: will do

filament/backend/src/opengl/PlatformCocoaGL.mm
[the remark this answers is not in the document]
jwwalker: will do

jwwalker: I was thinking that the fact that the destructor removes notifications would prevent the notification from running after that, but I put in a weak reference check to the `observers` member to be sure.

But that made me realize that the anonymous block (the one that installs notifications) posted to the main queue also references member variables `observers`, `currentBounds`, and `currentWindowFrame`.  I'll make that explicit and use a weak reference to `observers`.    Doing a check that the weak reference to `observers` is not nil handles the case that the block runs after the `CocoaGLSwapChain` is destroyed.  And then I thought, it's unlikely, but what if the `CocoaGLSwapChain` gets destroyed *while* the block is running, so that some observers get installed after the chance to remove them has passed?  I put in `@synchronized` checks to handle that.

filament/backend/src/opengl/PlatformCocoaGL.mm
```diff
@@ -47,6 +53,83 @@
...
+    
+    // Various methods below should only be called from the main thread:
+    // -[NSView bounds], -[NSView convertRectToBacking:], -[NSView window],
+    // -[NSWindow frame], -[NSView superview],
+    // -[NSView setPostsFrameChangedNotifications:],
+    // -[NSView setPostsBoundsChangedNotifications:]
+    dispatch_async(dispatch_get_main_queue(), ^(void) {
+        NSView* strongView = weakView;
+		NSMutableArray* strongObservers = weakObservers;
+        if ((weakView == nil) || (weakObservers == nil)) {
```
bejado, line 82: Please fix indentation here and throughout the rest of the file (we use 4 spaces, not tabs).

filament/backend/src/opengl/PlatformCocoaGL.mm
bejado: Alright, that sounds fine to me.

filament/backend/src/opengl/PlatformCocoaGL.mm
jwwalker: done!

bejado (approved):

jwwalker:
Do I need to do anything here?  Since my change does not affect any Windows code, I don't know what to make of a Windows test failure here.

bejado:
> Do I need to do anything here? Since my change does not affect any Windows code, I don't know what to make of a Windows test failure here.

No, the Windows test failure is unrelated.

PR #4013, opened 2021-05-23
fix MRTs >= COLOR4 in the GL backend

pixelflinger:
These would be ignored.

### DIFF
--- a/filament/backend/src/opengl/OpenGLDriver.h
+++ b/filament/backend/src/opengl/OpenGLDriver.h
@@ -21,6 +21,8 @@
 #include "DriverBase.h"
 #include "OpenGLContext.h"
 
+#include "backend/TargetBufferInfo.h"
+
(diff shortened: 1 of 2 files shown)


filament/backend/src/opengl/OpenGLDriver.h
```diff
@@ -382,7 +384,8 @@ class OpenGLDriver final : public backend::DriverBase {
         return mSamplerBindings;
     }
 
-    static GLsizei getAttachments(std::array<GLenum, 6>& attachments,
+    using AttachmentArray = std::array<GLenum, backend::MRT::MAX_SUPPORTED_RENDER_TARGET_COUNT + 2>;
```
romainguy, line 387: Maybe add a comment to explain the + 2

romainguy (approved):

PR #4018, opened 2021-05-24
build.sh -c option entirely cleans the android/ directory

pixelflinger:
The android/ directory is now left as if it was freshly checked out, 
all non-git files (including those in .gitignore) are deleted.

This is to prevent a recurring problem where "clean" builds where not
actually clean in some cases.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -17,6 +17,10 @@ function print_help {
     echo "        Generate .tgz build archives, implies -i."
     echo "    -c"
     echo "        Clean build directories."
+    echo "    -C"
+    echo "        Clean build directories and revert android/ to a freshly sync'ed state."
+    echo "        All (and only) git-ignored files under android/ are deleted."
+    echo "        This is sometimes needed instead of -c (which still misses some clean steps)."
     echo "    -d"
     echo "        Enable matdbg and disable material optimization."
     echo "    -f"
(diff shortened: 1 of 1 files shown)


romainguy (commented):
Doesn't this mean that you could lose any changes you haven't staged/committed if you run build.sh -c during your testing?

pixelflinger:
> Doesn't this mean that you could lose any changes you haven't staged/committed if you run build.sh -c during your testing?

I will double check, but I don't think so because the `X` option only looks at gitignored files (as opposed to the `x` option).

romainguy:
So `-fX` will delete files listed in .gitignore? Doesn't that mean losing IDE configurations from the `.idea/` folder, your local Gradle properties generated when importing the project in Studio, etc.? I love the idea but I want to make sure we won't blow up more than we should.

pixelflinger:
> So `-fX` will delete files listed in .gitignore? Doesn't that mean losing IDE configurations from the `.idea/` folder, your local Gradle properties generated when importing the project in Studio, etc.? I love the idea but I want to make sure we won't blow up more than we should.

Yes it'll blow away all that stuff :-\
I didn't notice any major issues with the IDE, though.

prideout (approved):
Personally I love this.  For those who are faint of heart, I suppose the script could explain what will happen and ask "Are you sure? y/N" ... but I think it's fine as is.

romainguy:
I still think it's unnecessarily dangerous during development. You may have your IDE open and blow up project files because you're just trying a clean build. I'd rather this became a new "nuclear clean" option in build.sh.

romainguy (approved):

prideout (approved):

bejado (approved):

PR #4023, opened 2021-05-25
A fixed capacity vector.

pixelflinger:
This is a drop-in replacement for std::vector, the main 
difference is that FixedCapacityVector has a fixed (at runtime)
capacity, which makes it a lot more efficient and small for operations
like push_back();

In addition a FixedCapacityVector<> instance is only 16 bytes instead
of 24 for std::vector<>.

The aim here is to reduce code bloat.

### DIFF
--- a/libs/utils/include/utils/FixedCapacityVector.h
+++ b/libs/utils/include/utils/FixedCapacityVector.h
@@ -1,0 +1,316 @@
+/*
+ * Copyright (C) 2021 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef TNT_UTILS_FIXEDCAPACITYFACTOR_H
+#define TNT_UTILS_FIXEDCAPACITYFACTOR_H
+
+#include <utils/compressed_pair.h>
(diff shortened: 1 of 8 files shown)


libs/utils/include/utils/FixedCapacityVector.h
```diff
@@ -0,0 +1,316 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef TNT_UTILS_FIXEDCAPACITYFACTOR_H
+#define TNT_UTILS_FIXEDCAPACITYFACTOR_H
```
prideout, line 18: FACTOR?

prideout (commented):
How do you construct an empty FixedCapacityVector with a certain capacity?  The existing constructors here seem to only handle the case where initial size == capacity.

pixelflinger:
> How do you construct an empty FixedCapacityVector with a certain capacity?  The existing constructors here seem to only handle the case where initial size == capacity

You call `reserve`!

prideout:
> > How do you construct an empty FixedCapacityVector with a certain capacity?  The existing constructors here seem to only handle the case where initial size == capacity
> 
> You call `reserve`!

LOL fine!

In that case I will have a lot of two-liners: declare-and-reserve.  Would be nice to have a constructor or static method for convenience, even though STL does not have this feature.

libs/utils/include/utils/FixedCapacityVector.h
pixelflinger: lol, I misspelled the file and I forget to fix that one :-)

pixelflinger:
> > > How do you construct an empty FixedCapacityVector with a certain capacity?  The existing constructors here seem to only handle the case where initial size == capacity
> > 
> > 
> > You call `reserve`!
> 
> LOL fine!
> 
> In that case I will have a lot of two-liners: declare-and-reserve. Would be nice to have a constructor or static method for convenience, even though STL does not have this feature.

PR #4025, opened 2021-05-25
Add support for drag-and-drop HDR.

prideout:
This adds Java bindings for IBLPrefilterContext and uses them in `sample-gltf-viewer`.


android/samples/sample-gltf-viewer/src/main/java/com/google/android/filament/gltf/MainActivity.kt
```diff
@@ -158,12 +160,30 @@ class MainActivity : Activity() {
...
-                // TODO: Add Java bindings for IBLPrefilterContext and use them here.
                 setStatusText("Successfully decoded HDR file.")
+
+                val context = IBLPrefilterContext(engine)
+                val equirectToCubemap = IBLPrefilterContext.EquirectangularToCubemap(context)
+                val skyboxTexture = equirectToCubemap.run(equirect)!!
+                engine.destroyTexture(equirect)
+
+                val specularFilter = IBLPrefilterContext.SpecularFilter(context)
+                val reflections = specularFilter.run(skyboxTexture)
```
romainguy, line 176: Does this mean the specular filter operation is blocking? Any chance our API could be made asynchronous or at least designed so it could be invoked from a coroutine in Kotlin?

romainguy: I understand it makes Filament calls but since it returns a `Texture`, does this mean it waits for the GPU to complete? What I don't like about this API is that it doesn't give the app the *possibility* of doing the work asynchronously. We force the app to block the Filament thread, which is likely to be the app's UI/main thread.

prideout: Well I made the Java 1:1 with C++, and there's no call to `flushAndWait()` behind the scenes so it's actually already asynchronous I think.  @pixelflinger might have thoughts, it's his baby.

pixelflinger: It returns a `Texture` but doesn't wait for the GPU to be done. At the point it returns the Texture, it's just a handle, it's not even created yet (on the backend). 
There is nothing blocking here -- it'll just send a lot of work to the GPU, which means the GPU will miss a few frames.

pixelflinger (approved):

android/samples/sample-gltf-viewer/src/main/java/com/google/android/filament/gltf/MainActivity.kt
romainguy: Do we have a callback then to let the app know when the texture is ready to be used?

romainguy (approved):

android/samples/sample-gltf-viewer/src/main/java/com/google/android/filament/gltf/MainActivity.kt
pixelflinger: you don't need one -- everything is pipelined into our command queue. As long as you don't try to share it with another engine of something like that.
Really this works just like any other filament API -- e.g. you don't need a callback for generateMipmap, for instance.

PR #4036, opened 2021-05-26
VulkanSwapChain: simplify and repair swap chain resizing.

prideout:
Newer versions of MoltenVK seem to behave similarly to Android,
so there's no need to separate codepaths. I tested this by not only
resizing the window, but also dragging between displays with different
pixel ratios, and rotating on Android with & without `orientation` in
`configChanges`. I also tried our headless unit tests.


romainguy (approved):
Yay, less code!

PR #4040, opened 2021-05-26
slog must be thread safe

pixelflinger:


romainguy (approved):

pixelflinger:
Unfortunately, on Android we can't use `thread_local` and instead we're using a wrapper, but this can't work with the current syntax of `slog` (because we can't override `.` in C++).
it looks like `thread_local` is now supported on android, but it's unclear to me if a binary built with a recent NDK will work on older versions of Android.

romainguy (approved):

PR #4046, opened 2021-05-27
Add support for KHR_materials_specular to glTF

romainguy:
This change makes available the existing IOR material parameter to materials that don't have refraction enabled.

Here is a glTF model with an IOR of 1:
Screen Shot 2021-05-27 at 3 06 58 PM

And the same model with an IOR of 2:
Screen Shot 2021-05-27 at 3 06 36 PM

Note that using this parameter on a non-refractive object is equivalent to setting the reflectance parameter.

IOR support is implemented either as IOR or reflectance in ubershaders to use a more optimized code path
for the common case (no IOR set).

### DIFF
--- a/libs/gltfio/src/AssetLoader.cpp
+++ b/libs/gltfio/src/AssetLoader.cpp
@@ -1138,7 +1138,7 @@ MaterialInstance* FAssetLoader::createMaterialInstance(const cgltf_material* inp
         }
     }
 
-    // IOR can be implemented either has IOR or reflectance because of ubershaders
+    // IOR can be implemented as either IOR or reflectance because of ubershaders
     if (matkey.hasIOR) {
         if (mi->getMaterial()->hasParameter("ior")) {
             mi->setParameter("ior", inputMat->ior.ior);
--- a/libs/gltfio/src/AssetLoader.cpp
+++ b/libs/gltfio/src/AssetLoader.cpp
@@ -46,6 +46,7 @@
 
 #include <tsl/robin_map.h>
 
+#include <cmath>
 #include <vector>
 
 #define CGLTF_IMPLEMENTATION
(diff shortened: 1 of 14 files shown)


romainguy:
I need to fix an issue with ubershaders first.

libs/gltfio/src/AssetLoader.cpp
```diff
@@ -926,31 +926,32 @@ MaterialInstance* FAssetLoader::createMaterialInstance(const cgltf_material* inp
...
+        .hasTransmissionTexture = trConfig.transmission_texture.texture != nullptr,
         .transmissionUV = (uint8_t) trConfig.transmission_texture.texcoord,
-        .hasSheenColorTexture = !!shConfig.sheen_color_texture.texture,
+        .hasSheenColorTexture = shConfig.sheen_color_texture.texture != nullptr,
         .sheenColorUV = (uint8_t) shConfig.sheen_color_texture.texcoord,
-        .hasSheenRoughnessTexture = !!shConfig.sheen_roughness_texture.texture,
+        .hasSheenRoughnessTexture = shConfig.sheen_roughness_texture.texture != nullptr,
         .sheenRoughnessUV = (uint8_t) shConfig.sheen_roughness_texture.texcoord,
         .hasSheen = !!inputMat->has_sheen,
+        .hasIOR = !!inputMat->has_ior,
```
pixelflinger, line 954: why change all the `!!` except here?

pixelflinger (approved):

prideout (commented):
Shouldn't we update `ubershader.mat.in` or any of the other .mat.in files?

prideout:
> Shouldn't we update `ubershader.mat.in` or any of the other .mat.in files?

Oh nm you just added it

romainguy:
@prideout I initially didn't implement the feature in ubershaders but then realized I could do it. The implementation got a bit subtle since it uses either IOR or reflectance, but that's because of other extensions like transmission/volume/specular that also rely on IOR. I may revisit how ubershaders deal with all this once I make my way through specular and volume.

libs/gltfio/src/AssetLoader.cpp
```diff
@@ -1137,8 +1138,18 @@ MaterialInstance* FAssetLoader::createMaterialInstance(const cgltf_material* inp
         }
     }
 
+    // IOR can be implemented either has IOR or reflectance because of ubershaders
```
prideout, line 1141: Grammar

prideout (commented):
Does `FAssetLoader::createMaterialInstance` need to transfer the `hasReflectance` property from cgltf to the MaterialKey?

romainguy:
I forgot to remove `hasReflectance`, it's now unused.

prideout (approved):
I would've put core Filament changes into a sep commit from gltfio changes but that's minor :)

alexey-pelykh:
@romainguy but the commit is about `KHR_materials_ior`, no?

PR #4061, opened 2021-06-02
VulkanPipelineCache: do not hash uninitialized data.

prideout:
This fixes an issue caught by a memory sanitizer, whereby padding in the
cache key structure was being consumed by a hashing function. In
practice this oversight was fairly harmless, but this change makes the
code more safe and predictable.

### DIFF
--- a/filament/backend/src/vulkan/VulkanPipelineCache.cpp
+++ b/filament/backend/src/vulkan/VulkanPipelineCache.cpp
@@ -39,6 +39,7 @@ VulkanPipelineCache::VulkanPipelineCache() : mDefaultRasterState(createDefaultRa
     markDirtyDescriptor();
     markDirtyPipeline();
     mDescriptorKey = {};
+    mPipelineKey = {};
 
     mDummyBufferWriteInfo.sType = VK_STRUCTURE_TYPE_WRITE_DESCRIPTOR_SET;
     mDummyBufferWriteInfo.pNext = nullptr;
--- a/filament/backend/src/vulkan/VulkanPipelineCache.cpp
+++ b/filament/backend/src/vulkan/VulkanPipelineCache.cpp
@@ -83,8 +83,9 @@ bool VulkanPipelineCache::bindDescriptors(VkCommandBuffer cmdbuffer) noexcept {
         return false;
     }
     if (bind) {
-        vkCmdBindDescriptorSets(cmdbuffer, VK_PIPELINE_BIND_POINT_GRAPHICS, mPipelineLayout, 0,
-                VulkanPipelineCache::DESCRIPTOR_TYPE_COUNT, descriptors, 0, nullptr);
+        vkCmdBindDescriptorSets(cmdbuffer, VK_PIPELINE_BIND_POINT_GRAPHICS,
+                mPipelineLayout, 0, VulkanPipelineCache::DESCRIPTOR_TYPE_COUNT, descriptors,
+                0, nullptr);
     }
     return true;
 }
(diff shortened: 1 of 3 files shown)


filament/backend/src/vulkan/VulkanPipelineCache.h
```diff
@@ -77,13 +77,15 @@ class VulkanPipelineCache : public CommandBufferObserver {
     // The RasterState POD contains standard graphics-related state like blending, culling, etc.
     // Note that several fields are unused (sType etc) so we could shrink this by avoiding the Vk
     // structures. However it's super convenient just to use standard Vulkan structs here.
-    struct alignas(8) RasterState {
+    #pragma pack(push, 1)
+    struct UTILS_PACKED RasterState {
```
pixelflinger, line 81: this doesn't have performance impact? -- if fields are not aligned to their natural alignement it can cause all sorts of issues -- it's sometimes okay, but it's hard to tell just looking at this structure.

I think it's better to show the padding explicitly and have a static_assert<> on the size.

prideout: I'm using structures from the Vulkan header inside this struct, so I can't make padding explicit or rearrange their members.  I do not store these in a contiguous array, so I don't think this will impact performance in a significant way.

pixelflinger: and you never pass a pointer to them to vk (or to a piece of code that wouldn't expect the packed format)?

prideout: Good point, I decided that using the Vk spec structs for state tracking is just plain wrong, added a commit to fix this.

filament/backend/src/vulkan/VulkanPipelineCache.cpp
```diff
@@ -423,31 +464,48 @@ void VulkanPipelineCache::bindProgramBundle(const ProgramBundle& bundle) noexcep
...
-    const VkPipelineMultisampleStateCreateInfo& ms1 = rasterState.multisampling;
+    auto& raster0 = mPipelineKey.rasterState.rasterization;
+    const auto& raster1 = rasterState.rasterization;
+    auto& blend0 = mPipelineKey.rasterState.blending;
+    const auto& blend1 = rasterState.blending;
+    auto& ds0 = mPipelineKey.rasterState.depthStencil;
+    const auto& ds1 = rasterState.depthStencil;
+    auto& ms0 = mPipelineKey.rasterState.multisampling;
+    const auto& ms1 = rasterState.multisampling;
     if (
```
pixelflinger, line 475: any sense of if this should be likely or unlikely?

filament/backend/src/vulkan/VulkanPipelineCache.cpp
```diff
@@ -423,31 +464,48 @@ void VulkanPipelineCache::bindProgramBundle(const ProgramBundle& bundle) noexcep
...
 void VulkanPipelineCache::bindRasterState(const RasterState& rasterState) noexcept {
-    VkPipelineRasterizationStateCreateInfo& raster0 = mPipelineKey.rasterState.rasterization;
-    const VkPipelineRasterizationStateCreateInfo& raster1 = rasterState.rasterization;
-    VkPipelineColorBlendAttachmentState& blend0 = mPipelineKey.rasterState.blending;
-    const VkPipelineColorBlendAttachmentState& blend1 = rasterState.blending;
-    VkPipelineDepthStencilStateCreateInfo& ds0 = mPipelineKey.rasterState.depthStencil;
-    const VkPipelineDepthStencilStateCreateInfo& ds1 = rasterState.depthStencil;
-    VkPipelineMultisampleStateCreateInfo& ms0 = mPipelineKey.rasterState.multisampling;
-    const VkPipelineMultisampleStateCreateInfo& ms1 = rasterState.multisampling;
+    auto& raster0 = mPipelineKey.rasterState.rasterization;
```
pixelflinger, line 467: to make the code more readable, you *could* write a standalone `operator=` for each of these struct.

pixelflinger (approved):

filament/backend/src/vulkan/VulkanPipelineCache.cpp
prideout: Good Q.  Let me add an "unlikely" over the entire expression, that would help with horrible models like car.zip

filament/backend/src/vulkan/VulkanPipelineCache.cpp
prideout: Yeah, the sub-structs are anonymous right now but I'll consider doing this in the future.

PR #4062, opened 2021-06-02
use c++ thread_local now that it's supported on Android

pixelflinger:
there are cases where thread_local still doesn't work well, but
we're not using those cases.


romainguy (approved):

filament/backend/src/android/ExternalStreamManagerAndroid.cpp
```diff
@@ -39,7 +39,7 @@ static void loadSymbol(T*& pfn, const char *symbol) noexcept {
 
 ExternalStreamManagerAndroid& ExternalStreamManagerAndroid::get() noexcept {
     // declaring this thread local, will ensure it's destroyed with the calling thread
-    static UTILS_DECLARE_TLS(ExternalStreamManagerAndroid) instance;
+    static thread_local ExternalStreamManagerAndroid instance;
```
bejado, line 42: What happens now if `thread_local` isn't supported?

libs/utils/include/utils/compiler.h
```diff
@@ -158,14 +158,9 @@
 #endif
 
 #if defined(_MSC_VER) && _MSC_VER >= 1900
-#       define UTILS_HAS_FEATURE_CXX_THREAD_LOCAL 1
+#   define UTILS_HAS_FEATURE_CXX_THREAD_LOCAL 1
```
bejado, line 161: Do we still need this define? It doesn't look like it's getting used elsewhere.

filament/backend/src/android/ExternalStreamManagerAndroid.cpp
pixelflinger: We're only using it on Android, where it's supported now. So I think it doesn't matter much (?)

libs/utils/include/utils/compiler.h
pixelflinger: Yes it's not used right now, but I guess if we need some thread local stuff in the common code, it might still be useful.

bejado (approved):

PR #4063, opened 2021-06-02
attempt to fix windows build

pixelflinger:


build/windows/build-github.bat
```diff
@@ -111,6 +111,7 @@ cmake ..\.. ^
     -DCMAKE_INSTALL_PREFIX=..\%variant% ^
     -DFILAMENT_WINDOWS_CI_BUILD:BOOL=ON ^
     -DFILAMENT_SUPPORTS_VULKAN=ON ^
+    -D_SILENCE_CXX17_ITERATOR_BASE_CLASS_DEPRECATION_WARNING ^
```
bejado, line 114: This sets a CMake option, but I think you mean to set a compiler option.

bejado:
I get this error locally too, so we should find a solution that doesn't rely on setting CI-only flags. See #4065.

PR #4065, opened 2021-06-02
spirv-tools: turn off warnings as errors to fix Windows build

bejado:
This started happening with the Visual Studio 2019 16.10 update.


bejado:
I get this error locally too, so we should find a solution that doesn't rely on setting CI-only flags. See #4065.

romainguy (approved):
Good call on the patch file

pixelflinger (approved):

PR #4074, opened 2021-06-02
add system property to override the backend selection

pixelflinger:
On Android, set `debug.filament.backend` to the desired backend:
- 0: default
- 1: gl/gles
- 2: vulkan
- 3: metal
- 4: noop

example:
   `adb shell setprop debug.filament.backend 2`

This will affect all filament apps on the device.

This sets the backed as if it was set from the public API.

### DIFF
--- a/filament/backend/src/Platform.cpp
+++ b/filament/backend/src/Platform.cpp
@@ -20,6 +20,7 @@
 #include <utils/debug.h>
 
 #if defined(ANDROID)
+    #include <sys/system_properties.h>
     #if defined(FILAMENT_SUPPORTS_OPENGL) && !defined(FILAMENT_USE_EXTERNAL_GLES3)
         #include "opengl/PlatformEGLAndroid.h"
     #endif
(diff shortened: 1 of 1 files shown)


prideout (approved):

filament/backend/src/Platform.cpp
```diff
@@ -82,6 +83,15 @@ Platform::~Platform() noexcept = default;
 DefaultPlatform* DefaultPlatform::create(Backend* backend) noexcept {
     SYSTRACE_CALL();
     assert_invariant(backend);
+
+#if defined(ANDROID)
+    char scratch[PROP_VALUE_MAX + 1];
+    int length = __system_property_get("debug.filament.backend", scratch);
+    if (length > 0) {
+        *backend = Backend(atoi(scratch));
```
romainguy, line 91: You should check the result of `atoi()`

romainguy (approved):

bejado (approved):

filament/backend/src/Platform.cpp
pixelflinger: You can't.
`On success, the function returns the converted integral number as an int value.
If the converted value would be out of the range of representable values by an int, it causes undefined behavior.`

I think this is good enough for a debug feature.

PR #4079, opened 2021-06-03
Feature/gltfio header refactor for ResourceLoader and AssetLoader

imbrig:
These changes make both `ResourceLoader` and `AssetLoader` more inline to the code design of other classes such as `FilamentAsset`, which uses an internal struct `FBarClass` that derives from the public regular `BarClass` and has that in a separate header in the `src` folder. Current functionality should remain intact, with functions being declared in these `FClass` headers. The `Impl` struct was removed from `ResourceLoader` and its (Impl) functions where put into `FResourceLoader`. Projects that where working before on Android and Windows where tested.

Why are these changes important? I believe it brings some consistency to these objects that we see in all other objects. It also makes it easier (than before) to extend gltf loading and usage of additional extensions. While ideally I would like to have complete access (putting these headers in the include path) to `FFilamentAsset`, `FAssetLoader`, and `FResourceLoader`, and deriving from those into custom classes for handling additional gltf extensions. I can still work around it by using forward declarations and namespace functions. This would also have the benefit of not intermingling the additional extension code with what’s already in the gltfio library.
I’m open to other suggestions for accomplishing this.


libs/gltfio/src/ResourceLoader.cpp
```diff
@@ -1061,4 +1011,45 @@ void ResourceLoader::updateBoundingBoxes(FFilamentAsset* asset) const {
...
+    return new FResourceLoader(config);
+}
+
+void ResourceLoader::destroy(ResourceLoader** loader) {
+    delete *loader;
+    *loader = nullptr;
+}
+
+void ResourceLoader::addResourceData(const char* uri, BufferDescriptor&& buffer) {
+    upcast(this)->addResourceData(uri, std::move(buffer));
```
imbrig, line 1024: There's gotta be a better way of passing BufferDescriptor&& buffer through than using std::move(buffer) twice along the way.

romainguy:
Thanks. Could you explain a bit more the reasons for this change? Consistency, while nice, is not always necessary. The design of Foo classes and FFoo implementations exists mostly to hide away internals and preserve binary compatibility as much as possible (and to avoid STL types in our public headers). Is this what you are trying to achieve here?

prideout:
I would be interested in learning how you'd like to extend the library.  If you need to process a custom glTF extension, you can look at the raw cgltf data using `getSourceAsset()`, then create your own Filament primitives accordingly.

libs/gltfio/src/ResourceLoader.cpp
pixelflinger: `std::move(buffer)` is essentially a no-op -- that's the correct way to do this.
`BufferDescriptor` can never be copied (because it holds a callback).

imbrig:
@romainguy @prideout Thank you for taking the time to look at this PR and for replying.

Currently I’ve added custom gltf extensions that are parsed through cgltf, with custom object members kept in FFilamentAsset, and loaded with both Asset and Resource loaders. For example, apng & video textures, node triggers for events, and other effects that are defined in the gltf file. This code is intertwined and prone to merge conflicts making it hard to manage. My intention is to extend / derive from these 3 internal classes, where the custom changes would live without affecting any of the original Filament code.

I understand why there are internal FFoo implementations that derive from interface Foo classes, although this make extending a little tricky but not impossible if I can access the FFoo from a header, even if it’s in the src directory. My intention is not to completely change the design, but use precedence for a design pattern already being used in gltfio.

The way I would extend these objects is by deriving from the FFoo struct with MyFFoo struct and using a series of public functions that would call the hidden MyFoo. So far I've done this on Windows with the gltf_viewer project without issues. It's a bit of extra hoops to jump through but it should keep those headers in the scr without moving them to include.
For instance the idea is,
JNI <- FilamentAsset <- FFilamentAsset -> MyFilamentAsset -> Interface_MyFilamentAsset -> MyJNI
           public header       private header       private header            public header functions
                                                                                                    fwd declaration MyFilamentAsset

The catch here is that I must do this for ResourceLoader and AssetLoader. It's not pretty but should work without changing much in Filament.

romainguy:
Thanks for the clarification. Just to be clear any `FFoo` class/struct in Filament is *private* APIs and may (will) break at any time as we make changes.

imbrig:
> Thanks for the clarification. Just to be clear any `FFoo` class/struct in Filament is _private_ APIs and may (will) break at any time as we make changes.

That is fine and if I implement things on my end correctly, there should be no issues when these changes do occur.

libs/gltfio/src/ResourceLoader.cpp
imbrig: Thanks for the reply.

imbrig:
@prideout @romainguy I was trying out `getSourceAsset()` pulling the `cgltf_data` to create the extension objects. It should take care of all of my current use cases, haven't converted everything but looks promising. I'll continue with this and if everything works out, I'll close this PR or report back if there was an issue. Thank you for the help.

imbrig:
Closing this PR. See my previous comment.

PR #4096, opened 2021-06-07
Ensure FILAMENT_MIN_COMMAND_BUFFERS_SIZE_IN_MB is available.

prideout:
If your build system does not provide a definition for this macro, then
you cannot compile this cpp file.  Since Allocators.h provides a
backup value, it needs to be included.

Note that this makes it so that the backend includes a file from
src/details, but I did not see any better alternative.

### DIFF
--- a/filament/backend/include/private/backend/BackendUtils.h
+++ b/filament/backend/include/private/backend/BackendUtils.h
@@ -18,6 +18,10 @@
 
 #include <stddef.h>
 
+#ifndef FILAMENT_MIN_COMMAND_BUFFERS_SIZE_IN_MB
+#    define FILAMENT_MIN_COMMAND_BUFFERS_SIZE_IN_MB 1
+#endif
+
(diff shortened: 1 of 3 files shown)


pixelflinger:
Oh.. I think we need to move this constant somewhere in the backend/ then and have Allocator.h include that.

prideout:
> Oh.. I think we need to move this constant somewhere in the backend/ then and have Allocator.h include that.

@pixelflinger Done!

bejado (approved):

pixelflinger (commented):
Technically this doesn't need to be in a public header. But I'm not sure which header would be best.

prideout:
> Technically this doesn't need to be in a public header. But I'm not sure which header would be best.

Moved to private/backend/BackendUtils.h

pixelflinger (approved):
Great! Thank you!

PR #4099, opened 2021-06-07
Include sample-gltf-viewer with Android releases

bejado:
This builds, signs, and uploads sample-gltf-viewer with releases.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -100,7 +100,7 @@ jobs:
         run: |
           echo "${APK_KEYSTORE_BASE64}" > filament.jks.base64
           base64 --decode filament.jks.base64 > filament.jks
-          BUILD_TOOLS_VERSION=30.0.3
+          BUILD_TOOLS_VERSION=$(ls ${ANDROID_HOME}/build-tools | sort -V | tail -n 1)
           APKSIGNER=${ANDROID_HOME}/build-tools/${BUILD_TOOLS_VERSION}/apksigner
           IN_FILE="out/sample-gltf-viewer-release.apk"
           OUT_FILE="out/filament-gltf-viewer-${TAG}-android.apk"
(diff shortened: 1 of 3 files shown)


prideout (commented):
I defer to @romainguy for review.  :)

.github/workflows/release.yml
```diff
@@ -96,6 +96,20 @@ jobs:
       - name: Run build script
         run: |
           cd build/android && printf "y" | ./build.sh release
+      - name: Sign sample-gltf-viewer
+        run: |
+          echo "${APK_KEYSTORE_BASE64}" > filament.jks.base64
+          base64 --decode filament.jks.base64 > filament.jks
+          BUILD_TOOLS_VERSION=30.0.3
```
romainguy, line 103: Do we need to do anything to ensure we have that version installed? Could we just take the most recent that's already installed instead?

build.sh
```diff
@@ -406,6 +413,24 @@ function build_android {
 
     build_desktop "${MOBILE_HOST_TOOLS}"
 
+    # When building the samples, we need to partially "install" the host tools so Gradle can see
```
romainguy, line 416: Why not do an install then?

romainguy (approved):

build.sh
bejado: Doing a full `install` builds everything, which is wasteful since we only need the host tools. What we want is something like `ninja install matc cmgen`, but that doesn't exist.

romainguy: Makes sense 👍

.github/workflows/release.yml
bejado: According to https://github.com/actions/virtual-environments/blob/main/images/macos/macos-11-Readme.md#android, it should be pre-installed on macOS runners. I don't know of any way to just use the latest version

romainguy: I would just ls the directory, sort, uniq and take the first item. Here's how I used to do it for the NDK:

```
local ndk_version=$(ls "${ndk_side_by_side}" | sort -V | tail -n 1 | cut -f 1 -d ".")
```

bejado: Cool, I'll try that.

pixelflinger (approved):

PR #4122, opened 2021-06-10
gltfio: Cache the dummy buffer to optimize load time.

prideout:
This greatly improves the load time on some Mali devices. On one device,
loading the car model went from 85 seconds to 35 seconds.

### DIFF
--- a/libs/gltfio/src/AssetLoader.cpp
+++ b/libs/gltfio/src/AssetLoader.cpp
@@ -156,6 +156,9 @@ struct FAssetLoader : public AssetLoader {
     const char* mDefaultNodeName;
     bool mError = false;
     bool mDiagnosticsEnabled = false;
+
+    // Weak reference to the largest dummy buffer so far in the current loading phase.
+    BufferObject* mDummyBufferObject;
 };
 
 FILAMENT_UPCAST(AssetLoader)
@@ -265,6 +268,7 @@ void FAssetLoader::createAsset(const cgltf_data* srcAsset, size_t numInstances) 
     #endif
 
     mResult = new FFilamentAsset(mEngine, mNameManager, &mEntityManager, srcAsset);
+    mDummyBufferObject = nullptr;
 
     // If there is no default scene specified, then the default is the first one.
     // It is not an error for a glTF file to have zero scenes.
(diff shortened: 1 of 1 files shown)


libs/gltfio/src/AssetLoader.cpp
```diff
@@ -156,6 +156,9 @@ struct FAssetLoader : public AssetLoader {
     const char* mDefaultNodeName;
     bool mError = false;
     bool mDiagnosticsEnabled = false;
+
+    // Weak reference to the largest dummy buffer so far in the current loading phase.
+    BufferObject* mDummyBufferObject;
```
romainguy, line 161: Aren't you leaking the object? I see no destruction anywhere.

libs/gltfio/src/AssetLoader.cpp
```diff
@@ -785,14 +789,16 @@ bool FAssetLoader::createPrimitive(const cgltf_primitive* inPrim, Primitive* out
...
-        mResult->mBufferObjects.push_back(bufferObject);
-        uint32_t* dummyData = (uint32_t*) malloc(size);
-        memset(dummyData, 0xff, size);
-        VertexBuffer::BufferDescriptor bd(dummyData, size, FREE_CALLBACK);
-        bufferObject->setBuffer(*mEngine, std::move(bd));
-        vertices->setBufferObjectAt(*mEngine, slot, bufferObject);
+        const uint32_t requiredSize = sizeof(ubyte4) * vertexCount;
+        if (mDummyBufferObject == nullptr || requiredSize > mDummyBufferObject->getByteCount()) {
+            mDummyBufferObject = BufferObject::Builder().size(requiredSize).build(*mEngine);
+            mResult->mBufferObjects.push_back(mDummyBufferObject);
```
romainguy, line 795: Can we end up pushing back the same mDummyBufferObject multiple times?

libs/gltfio/src/AssetLoader.cpp
```diff
@@ -785,14 +789,16 @@ bool FAssetLoader::createPrimitive(const cgltf_primitive* inPrim, Primitive* out
...
+            uint32_t* dummyData = (uint32_t*) malloc(requiredSize);
+            memset(dummyData, 0xff, requiredSize);
```
romainguy, line 797: Is the memset necessary?

libs/gltfio/src/AssetLoader.cpp
prideout: It is owned by the asset, in its mBufferObjects array.

libs/gltfio/src/AssetLoader.cpp
prideout: No, this code is executed only when the required size exceeds the current dummy size.

libs/gltfio/src/AssetLoader.cpp
prideout: Yes, ubershader handles per-vertex color, which needs to be set to 1.0

romainguy (approved):

PR #4128, opened 2021-06-11
Vulkan: Use custom VMA pools to speed up Mali.

prideout:
This makes the car load in 3 seconds instead of 30 seconds on Mali.

On devices that have bufferImageGranularity set to 1 in the
VkPhysicalDeviceLimits caps structure, we observe that vmaCreateBuffer
can be two orders of magnitude faster than with devices that have a
higher value for their bufferImageGranularity.

According to the Vulkan spec, the bufferImageGranularity aligment
constraint needs to be honored only when linear and non-linear resources
live next to each other in memory.  Since all VkBuffer objects are
linear, we can allocate them all from the same pool and ignore the
alignment constraint.

### DIFF
--- a/filament/backend/src/vulkan/VulkanContext.cpp
+++ b/filament/backend/src/vulkan/VulkanContext.cpp
@@ -275,6 +275,37 @@ void VulkanContext::createLogicalDevice() {
     };
     vmaCreateAllocator(&allocatorInfo, &allocator);
 
+    const uint32_t memTypeBits = UINT32_MAX;
+
+    // Create a GPU memory pool for VkBuffer objects that ignores the bufferImageGranularity field
+    // in VkPhysicalDeviceLimits. We have observed that honoring bufferImageGranularity can cause
+    // the allocator to slow to a crawl.
+    uint32_t memTypeIndex = UINT32_MAX;
+    const VmaAllocationCreateInfo gpuInfo = { .usage = VMA_MEMORY_USAGE_GPU_ONLY };
+    UTILS_UNUSED_IN_RELEASE VkResult res = vmaFindMemoryTypeIndex(allocator, memTypeBits, &gpuInfo,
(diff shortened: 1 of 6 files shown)


prideout:
For reference, I've also filed a bug against VMA here: https://github.com/GPUOpen-LibrariesAndSDKs/VulkanMemoryAllocator/issues/178

filament/backend/src/vulkan/VulkanContext.cpp
```diff
@@ -275,6 +275,38 @@ void VulkanContext::createLogicalDevice() {
...
+    // observed that honoring "bufferImageGranularity" can cause the allocator to slow to a crawl.
+    uint32_t memTypeIndex = UINT32_MAX;
+    const VmaAllocationCreateInfo gpuInfo = { .usage = VMA_MEMORY_USAGE_GPU_ONLY };
+    UTILS_UNUSED_IN_RELEASE VkResult res = vmaFindMemoryTypeIndex(allocator, memTypeBits, &gpuInfo,
+            &memTypeIndex);
+    assert_invariant(res == VK_SUCCESS && memTypeIndex != UINT32_MAX);
+    const VmaPoolCreateInfo gpuPoolInfo {
+        .memoryTypeIndex = memTypeIndex,
+        .flags = VMA_POOL_CREATE_IGNORE_BUFFER_IMAGE_GRANULARITY_BIT,
+        .blockSize = 128ull * 1024 * 1024,
```
romainguy, line 292: Would it be useful to make this a C++ constant somewhere?

romainguy (approved):

filament/backend/src/vulkan/VulkanContext.cpp
prideout: Doh, in fact, we have an entire file for stuff like this (VulkanConstants.h).  Fixed.

PR #4137, opened 2021-06-15
filament-utils-android: fix string literal conversions

bejado:
This is annoying, but we're technically doing something bad here. The `JNINativeMethod` struct looks like this:
```
typedef struct {
    char *name;
    char *signature;
    void *fnPtr;
} JNINativeMethod;
```

To comply with Google3's stricter build rules, we have to provide a non-const `char*` (string literals are constant). Otherwise, I see the following error:
```
error: ISO C++11 does not allow conversion from string literal to 'char *' [-Werror,-Wwritable-strings]
```

Alternatively, if we assume that JNI won't try to mutate the string, we could just cast to a `char*`. I think that's a bit riskier though.

### DIFF
--- a/android/filament-utils-android/src/main/cpp/Utils.cpp
+++ b/android/filament-utils-android/src/main/cpp/Utils.cpp
@@ -102,20 +102,25 @@ JNIEXPORT jint JNI_OnLoad(JavaVM* vm, void*) {
     jclass ktxloaderClass = env->FindClass("com/google/android/filament/utils/KTXLoader");
     if (ktxloaderClass == nullptr) return JNI_ERR;
     static const JNINativeMethod ktxMethods[] = {
-        {"nCreateKTXTexture", "(JLjava/nio/Buffer;IZ)J", reinterpret_cast<void*>(nCreateKTXTexture)},
-        {"nCreateIndirectLight", "(JLjava/nio/Buffer;IZ)J", reinterpret_cast<void*>(nCreateIndirectLight)},
-        {"nCreateSkybox", "(JLjava/nio/Buffer;IZ)J", reinterpret_cast<void*>(nCreateSkybox)},
-        {"nGetSphericalHarmonics", "(Ljava/nio/Buffer;I[F)Z", reinterpret_cast<void*>(nGetSphericalHarmonics)},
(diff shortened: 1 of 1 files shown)


romainguy (approved):

bejado:
Grr.. it looks like with some versions of JNI, `JNINativeMethod` _does_ use `const char*`, which means the call to `free` will fail... Maybe it's fine to just do the cast to `char*`.

android/filament-utils-android/src/main/cpp/Utils.cpp
```diff
@@ -102,25 +102,20 @@ JNIEXPORT jint JNI_OnLoad(JavaVM* vm, void*) {
...
-    free(ktxMethods[2].name); free(ktxMethods[2].signature);
-    free(ktxMethods[3].name); free(ktxMethods[3].signature);
     if (rc != JNI_OK) return rc;
 
     // HDRLoader
     jclass hdrloaderClass = env->FindClass("com/google/android/filament/utils/HDRLoader");
     if (hdrloaderClass == nullptr) return JNI_ERR;
     static const JNINativeMethod hdrMethods[] = {
-        {strdup("nCreateHDRTexture"), strdup("(JLjava/nio/Buffer;II)J"), reinterpret_cast<void*>(nCreateHDRTexture)},
+        {(char*)"nCreateHDRTexture", (char*)"(JLjava/nio/Buffer;II)J", reinterpret_cast<void*>(nCreateHDRTexture)},
```
romainguy, line 117: We could also use an std::string. I'm surprised this works without a const_cast.

romainguy (approved):

android/filament-utils-android/src/main/cpp/Utils.cpp
pixelflinger: It's clearly an oversight. Let's just cast to `char*`.

pixelflinger: `const_cast` would be cleaner -- but the basic C cast always wins (it's stronger than `reinterpret_cast`). I think this is fine. 
I'm surprised that the `reinterpret_cast` for the function pointer is needed actually, any pointer should automatically cast to void. oh well.

pixelflinger (approved):

PR #4140, opened 2021-06-15
Fix several FrameGraph issues related to imported render targets

### LINKED ISSUE #4085: RenderTarget crashes with transparent material
**Describe the bug**
RenderTarget crashes when there is transparent material in the scene.

**To Reproduce**
1. Add the following lines to aiDefaultMat.mat - which is applied to the object in the rendertarget example application.

```
    refractionType : solid,
    refractionMode : screenspace,
```
2. Try running the recompiled rendertarget example which will crash immediately:
```
in void filament::backend::metal::MetalDriver::createRenderTargetR(Handle<filament::backend::HwRenderTarget>, filament::backend::TargetBufferFlags, uint32_t, uint32_t, uint8_t, backend::MRT, filament::backend::TargetBufferInfo, filament::backend::TargetBufferInfo):273
reason: The COLOR0 flag was specified, but no color texture provided.
```

**Expected behavior**
No crash with transparent materials when using render targets.

 - OS: MacOS
 - Backend: OpenGL/Metal

pixelflinger:
Imported render targets are a special case in the frame graph, in particular, the texture resource that refers to them is not a real texture with a HwTexture handle, it can be converted back to the render target, but it can never be used as an actual texture -- in particular it can't be used to create a HwRenderTarget.

There was other problem related to clearing render targets, individual fixes are explained in their respective CL.

This should fix #4085

### DIFF
--- a/libs/viewer/src/SimpleViewer.cpp
+++ b/libs/viewer/src/SimpleViewer.cpp
@@ -600,9 +600,14 @@ void SimpleViewer::updateUserInterface() {
     if (ImGui::CollapsingHeader("View")) {
         ImGui::Indent();
 
-        bool dither = mSettings.view.dithering == Dithering::TEMPORAL;
-        ImGui::Checkbox("Dithering", &dither);
-        enableDithering(dither);
+        ImGui::Checkbox("Post-processing", &mSettings.view.postProcessingEnabled);
+        ImGui::Indent();
+            bool dither = mSettings.view.dithering == Dithering::TEMPORAL;
+            ImGui::Checkbox("Dithering", &dither);
+            enableDithering(dither);
+            ImGui::Checkbox("Bloom", &mSettings.view.bloom.enabled);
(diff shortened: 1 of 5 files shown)


libs/viewer/src/SimpleViewer.cpp
```diff
@@ -600,9 +600,14 @@ void SimpleViewer::updateUserInterface() {
     if (ImGui::CollapsingHeader("View")) {
         ImGui::Indent();
 
-        bool dither = mSettings.view.dithering == Dithering::TEMPORAL;
-        ImGui::Checkbox("Dithering", &dither);
-        enableDithering(dither);
+        ImGui::Checkbox("Post-processing", &mSettings.view.postProcessingEnabled);
```
prideout, line 603: Oh nice, this setting already existed so we should have exposed it anyway.  We might want to update the remote web page at some point.

prideout (approved):
LGTM.  Do we have a way of testing for the problems that this fixes?

romainguy (approved):

PR #4142, opened 2021-06-16
Materials can now provide custom lighting/shading

romainguy:
When a material uses the "lit" shading model, customSurfaceShading can
be enabled to replace Filament's lighting implementation. When this
feature is enabled, the material *must* provide the following function
in the fragment shader block:

```
    vec3 surfaceShading(
            const MaterialInputs materialInputs,
            const ShadingData shadingData,
            const LightData lightData
    ) {
        return vec3(1.0); // custom lighting here
    }
```

Please refer to the docs in Materials.html for more information about
this feature and the different values provided by the data structures
passed to the function.

Here is an example of a simple toon shading material:

```
material {
    name : Toon,
    shadingModel : lit,
    parameters : [
        {
            type : float3,
            name : baseColor
        }
    ],
    customSurfaceShading : true
}

fragment {
    void material(inout MaterialInputs material) {
        prepareMaterial(material);
        material.baseColor.rgb = materialParams.baseColor;
    }

    vec3 surfaceShading(
            const MaterialInputs materialInputs,
            const ShadingData shadingData,
            const LightData lightData
    ) {
        // Number of visible shade transitions
        const float shades = 5.0;
        // Ambient intensity
        const float ambient = 0.1;

        float toon = max(ceil(lightData.NdotL * shades) / shades, ambient);

        // Shadowing and attenuation
        toon *= lightData.visibility * lightData.attenuation;

        // Color and intensity
        vec3 light = lightData.colorIntensity.rgb * lightData.colorIntensity.w;

        return shadingData.diffuseColor * light * toon;
    }
}
```

And the result in action:

https://user-images.githubusercontent.com/869684/122141705-beb93f80-ce02-11eb-9254-a9c985af411a.mov


docs/Materials.md.html
```diff
@@ -1869,6 +1889,165 @@
...
+struct ShadingData {
+    // The material's diffuse color, as derived from baseColor and metallic.
+    // This color is pre-multiplied by alpha and in the linear sRGB color space.
+    vec3  diffuseColor;
+
+    // The material's specular color, as derived from baseColor and metallic.
+    // This color is pre-multiplied by alpha and in the linear sRGB color space.
+    vec3  f0;
+
+    // The perceptual roughness it the roughness value set in MaterialInputs,
```
prideout, line 1948:
```suggestion
    // The perceptual roughness is the roughness value set in MaterialInputs,
```

docs/Materials.md.html
```diff
@@ -1869,6 +1889,165 @@
...
+    // with extra processing:
+    // - Clamped to safe values
+    // - Filtered if specularAntiAliasing is enabled
+    // This value is between 0.0 and 1.0.
+    float perceptualRoughness;
+
+    // The roughness value expected by BRDFs. This value is the square of
+    // perceptualRoughness. This value is between 0.0 and 1.0.
+    float roughness;
+};
```
prideout, line 1958: Minor: in the actual struct the field ordering looks slightly different (perceptualRoughness, f0, roughness)

prideout (approved):

docs/Materials.md.html
romainguy: It is; the order in GLSL is to align on vec4s, but I used a more natural order for the docs.

docs/Materials.md.html
```diff
@@ -1869,6 +1889,165 @@
...
+    // The normalized light vector, in world space (direction from the
+    // current fragment's position to the light).
+    vec3  l;
+
+    // The dot product of the shading normal (with normal mapping applied)
+    // and the light vector. This value is equal to the result of
+    // saturate(dot(getWorldSpaceNormal(), lightData.l)).
+    // This value is always between 0.0 and 1.0. When the value is <= 0.0,
+    // the current fragment is not visible from the light and lighting
+    // computations can be skipped.
```
pixelflinger, line 1980: should we skip calling the user function entirely if nol<0?

docs/Materials.md.html
```diff
@@ -1869,6 +1889,165 @@
...
+    float NdotL;
+
+    // The position of the light in world space.
+    vec3  worldPosition;
```
pixelflinger, line 1984: We need more details here. This is the "fake" world space, right? I mean, the world space offset by the camera position.
Meaning it's not the same world space that the actual world. i.e. if they pass some objects position by hand, they need to adjust with the camera position.

pixelflinger (approved):

docs/Materials.md.html
romainguy: So, we do that for *our* lighting, but like mentioned somewhere else in the doc, I did it this way because it can be useful for stylized effects.

docs/Materials.md.html
romainguy: It's the same as all the world position in materials, this is documented somewhere else in this file. But yeah we could detail more.

PR #4145, opened 2021-06-16
Repair FeedbackLoops test.

prideout:


prideout:
btw @bejado I noticed that Metal asserts `mUsedStages.empty()` but this test is a little non-standard so I'm not sure if that's a real bug or not.

romainguy (approved):

bejado:
@prideout Yeah, looks like it is failing. Probably not urgent, but I'll take a look. Remind me, what are "feedback loops" used for in Filament? Some post-processing step?

prideout:
> @prideout Yeah, looks like it is failing. Probably not urgent, but I'll take a look. Remind me, what are "feedback loops" used for in Filament? Some post-processing step?

When bloom generates a mipmap chain, it uses the same texture for reading & writing.  In GL we have to set up BASE_LEVEL and MAX_LEVEL for this work across all vendors.

Since bloom looks reasonable with Metal, I'm guessing the Metal issue has more to do with the weirdness of the test than what it's actually intended to test.  Therefore, yes totally not urgent. :)

prideout:
@bejado never mind, I made some additional fixes to the test itself (#4152) and now the Metal assert is gone.

PR #4155, opened 2021-06-17
Backend::DEFAULT now selects a backend most suited for the platform

pixelflinger:
- OpenGL ES on Android
- Metal on MacOS/iOS
- Vulkan everywhere else (Linux/Windows)

The main change here is that OpenGL is no longer the default on
desktop.

Most of our samples have a `-a` option that can be use to specify the
desired backend.


romainguy (approved):
Please document this in `RELEASE_NOTES` because this may affect how you generate the materials.

bejado (approved):

prideout (approved):

PR #4158, opened 2021-06-17
Include ProGuard rules in our AARs

romainguy:


prideout (approved):

pixelflinger:
What does that actually do?

bejado (approved):

romainguy:
It copies our proguard-rules.pro files into the AAR, so other AARs or APKs that depend on Filament will properly preserve the classes we marked `UsedBy`. Otherwise, using R8 might remove classes/methods needed by our runtime.

PR #4176, opened 2021-06-22
FixedCapacityVector: add note about default construction.

prideout:


romainguy (approved):

libs/utils/include/utils/FixedCapacityVector.h
```diff
@@ -46,6 +46,11 @@ namespace utils {
  * An empty vector with a given capacity is created with
  *   FixedCapacityVector<T>::with_capacity( capacity );
  *
+ * NOTE: One important difference betwen std::vector and FixedCapacityVector is that the former
```
pixelflinger, line 49: To perfectly correct, default construction *is* performed if an only if it is *not trivial*.  Or said differently, only trivial default construction is skipped.

pixelflinger (changes requested):

libs/utils/include/utils/FixedCapacityVector.h
prideout: Good catch, I added a commit.

pixelflinger (approved):

PR #4178, opened 2021-06-22
VSM improvements and fixes

pixelflinger:
- use EVSM (exponential VSM), helps a lot with light bleeding
- don't apply shadow biases with VSM
- improve VSM user settings: added more controls, expose to java
- minor code cleanups:
   - rename some structures
   - remove unused parameters
   - remove unused uniforms

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -4,6 +4,7 @@ This file contains one line summaries of commits that are worthy of mentioning i
 A new header is inserted each time a *tag* is created.
 
 ## v1.10.6 (currently main branch)
+- engine: Use exponential VSM and improve VSM user settings [⚠️ **Recompile Materials for VSM**].
 
 ## v1.10.5
 
--- a/android/filament-android/src/main/java/com/google/android/filament/View.java
+++ b/android/filament-android/src/main/java/com/google/android/filament/View.java
@@ -687,20 +687,44 @@ public class View {
     /**
      * View-level options for VSM shadowing.
      *
+     * <strong>Warning: This API is still experimental and subject to change.</strong>
+     *
      * @see View#setVsmShadowOptions
      */
     public static class VsmShadowOptions {
         /**
          * Sets the number of anisotropic samples to use when sampling a VSM shadow map. If greater
          * than 0, mipmaps will automatically be generated each frame for all lights.
+         * This implies mipmapping below.
          *
          * <p>
          * The number of anisotropic samples = 2 ^ vsmAnisotropy.
          * </p>
          *
-         * <strong>Warning: This API is still experimental and subject to change.</strong>
          */
         public int anisotropy = 0;
+
+        /**
+         * Whether to generate mipmaps for all VSM shadow maps.
+         */
+        public boolean mipmapping = false;
+
+        /**
+         * EVSM exponent
+         * The maximum value permissible is 5.54 for a shadow map in fp16, or 42.0 for a
+         * shadow map in fp32. Currently the shadow map bit depth is always fp16.
+         */
+        public float exponent = 5.54f;
+
+        /**
+         * VSM minimum variance scale, must be positive.
+         */
+        public float minVarianceScale = 1.0f;
+
+        /**
+         * VSM light bleeding reduction amount, between 0 and 1.
+         */
+        public float lightBleedReduction = 0.2f;
     }
 
     /**
@@ -1308,7 +1332,8 @@ public class View {
      */
     public void setVsmShadowOptions(@NonNull VsmShadowOptions options) {
         mVsmShadowOptions = options;
-        nSetVsmShadowOptions(getNativeObject(), options.anisotropy);
+        nSetVsmShadowOptions(getNativeObject(), options.anisotropy, options.mipmapping,
+                options.exponent, options.minVarianceScale, options.lightBleedReduction);
     }
 
     /**
@@ -1522,7 +1547,7 @@ public class View {
     private static native void nSetRenderQuality(long nativeView, int hdrColorBufferQuality);
     private static native void nSetDynamicLightingOptions(long nativeView, float zLightNear, float zLightFar);
     private static native void nSetShadowType(long nativeView, int type);
-    private static native void nSetVsmShadowOptions(long nativeView, int anisotropy);
+    private static native void nSetVsmShadowOptions(long nativeView, int anisotropy, boolean mipmapping, float exponent, float minVarianceScale, float lightBleedReduction);
     private static native void nSetColorGrading(long nativeView, long nativeColorGrading);
     private static native void nSetPostProcessingEnabled(long nativeView, boolean enabled);
     private static native boolean nIsPostProcessingEnabled(long nativeView);
--- a/filament/include/filament/LightManager.h
+++ b/filament/include/filament/LightManager.h
@@ -230,15 +230,17 @@ public:
 
         /** Constant bias in world units (e.g. meters) by which shadows are moved away from the
          * light. 1mm by default.
+         * This is ignored when the View's ShadowType is set to VSM.
          */
         float constantBias = 0.001f;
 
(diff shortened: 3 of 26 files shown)


pixelflinger:
I marked the PR as draft because I think I need to rework the `VsmShadowOptions` so the variance/light-bleed parameters are per-light.

pixelflinger:
> I marked the PR as draft because I think I need to rework the `VsmShadowOptions` so the variance/light-bleed parameters are per-light.

It's actually pretty involved to move those settings per-light, so if we do it, it'll be in a separate PR.

romainguy:
Can we even move those settings per light? We would have to store more data in the froxels (another case were deferred rendering shines :(). We would also want to keep a global set of options to apply by default to all the lights.

pixelflinger:
> Can we even move those settings per light? We would have to store more data in the froxels (another case were deferred rendering shines :(). We would also want to keep a global set of options to apply by default to all the lights.

I don't think so, it's just a couple of uniforms for the the light-bleed and min variance. I'm not suggesting to be able to chose PCF/VSM per light.

We can't  move mipmaping/anisotropy per-light either because we only have one sampler for all the shadow maps.

romainguy (commented):
That looks good but our API names and comments are a bit confusing because we mention both VSM and EVSM. We should make it clear that VSM means EVSM in our case.

romainguy:
> I don't think so, it's just a couple of uniforms for the the light-bleed and min variance. I'm not suggesting to be able to chose PCF/VSM per light.

It's two uniforms *per light* though, so they have to go in the froxels. I know we're not talking about choosing PCF/VSM per light.

pixelflinger:
> > I don't think so, it's just a couple of uniforms for the the light-bleed and min variance. I'm not suggesting to be able to chose PCF/VSM per light.
> 
> It's two uniforms _per light_ though, so they have to go in the froxels. I know we're not talking about choosing PCF/VSM per light.

It's not needed per-froxel, these values are not different per froxel. But in fact we don't even need them per-light, because we only support a limited number of shadows -- so we just need these 2 uniforms per shadowmap, and we already have a UBO per spotlight shadow, so there is a precedent for doing this.

romainguy:
Yes, sorry I meant "light record" not "froxel". But yeah if it's only for the shadow maps that's easier.

RELEASE_NOTES.md
```diff
@@ -13,6 +13,7 @@ A new header is inserted each time a *tag* is created.
 - engine: `Backend::DEFAULT` now selects the most appropriate backend for the platform, rather than
   always `OPENGL`. On Android the default is `OPENGL`, on Apple platforms the default is `METAL` and
   on all other platforms that default is `VULKAN`.
+- engine: Use exponential VSM and improve VSM user settings.
```
bejado, line 16: Make a note about needing to rebuild materials for EVSM?

android/filament-android/src/main/java/com/google/android/filament/View.java
```diff
@@ -701,6 +702,26 @@
          * <strong>Warning: This API is still experimental and subject to change.</strong>
```
bejado, line 702: Move this Warning to `VsmShadowOptions` like you did for C++.

android/filament-android/src/main/java/com/google/android/filament/View.java
```diff
@@ -701,6 +702,26 @@
...
          */
         public int anisotropy = 0;
+
+        /**
+         * Whether to generate mipmaps for all VSM shadow maps.
+         */
+        public boolean mipmapping = false;
+
+        /**
+         * EVSM exponent
```
bejado, line 712: Copy over C++ documentation for exponent?

filament/include/filament/LightManager.h
```diff
@@ -230,11 +230,13 @@ class UTILS_PUBLIC LightManager : public FilamentAPI {
 
         /** Constant bias in world units (e.g. meters) by which shadows are moved away from the
          * light. 1mm by default.
+         * This is ignored when VSM are used.
```
bejado, line 233: I'd phrase this as:

```
This is ignored when the View's ShadowType is set to VSM.
```

Same for `normalBias`.

filament/src/ShadowMap.cpp
```diff
@@ -90,27 +90,25 @@ void ShadowMap::render(DriverApi& driver, FView::Range const& range, RenderPass&
...
-    const mat4f M = mat4f::lookAt(lightPosition, lightPosition + dir, float3{ 0, 1, 0 });
-    const mat4f Mv = FCamera::rigidTransformInverse(M);
+mat4f ShadowMap::getLightViewMatrix(float3 position, float3 direction) noexcept {
+    const mat4f M = mat4f::lookAt(position, position + direction, float3{ 0, 1, 0 });
+    return FCamera::rigidTransformInverse(M);
+}
+
+void ShadowMap::computeSceneCascadeParams(float3 dir,
+        FScene const& scene, filament::CameraInfo const& camera, uint8_t visibleLayers,
+        SceneInfo& cascadeParams) {
```
bejado, line 100: Rename `cascadeParams` to `sceneInfo`?

filament/src/ShadowMap.cpp
```diff
@@ -428,32 +431,37 @@ void ShadowMap::computeShadowCameraDirectional(
...
         // It's important to set the light camera's model matrix separately from its projection, so
         // that the cameraPosition uniform gets set correctly.
         // mLightSpace is used in the shader to access the shadow map texture, and has the model
         // matrix baked in.
 
-        mCamera->setModelMatrix(FCamera::rigidTransformInverse(b) * M);
+        // The model matrix below is in fact inverted to get the view matrix and passed to the
+        // shader as 'viewFromWorldMatrix', and is used in the VSM case to compute the depth metric.
+        // (see depth_main.fs). Note that in the case of VSM, 'b' below is identity.
+        mCamera->setModelMatrix(inverse(Mv * b));
```
bejado, line 454: Can we not use `rigidTransformInverse` here?

filament/src/ShadowMap.cpp
```diff
@@ -624,22 +636,22 @@ mat4f ShadowMap::getTextureCoordsMapping() const noexcept {
...
+    // (the normalization factor is therefore -1/zfar).
+    //
     // When sampling a VSM shadow map, the shader follows suit, and doesn't divide by w for the z
     // coordinate. See getters.fs.
-    math::mat4f lightSpaceVsm = lightSpacePcf;
-    lightSpaceVsm[0][2] = Mv[0][2] * (-1.0f / std::abs(zfar));
-    lightSpaceVsm[1][2] = Mv[1][2] * (-1.0f / std::abs(zfar));
-    lightSpaceVsm[2][2] = Mv[2][2] * (-1.0f / std::abs(zfar));
-    lightSpaceVsm[3][2] = Mv[3][2] * (-1.0f / std::abs(zfar));
+    math::mat4f lightSpaceVsm{ lightSpacePcf };
```
bejado, line 650: I think we can remove the negative sign here, and also in the shader.

bejado (commented):
Overall LGTM. I'll make another pass through later

filament/src/ShadowMap.cpp
pixelflinger: I realized later that this is both a scene info AND a cascade parameter :-)

filament/src/ShadowMap.cpp
pixelflinger: Yes, good point.

pixelflinger:
I added a CL that fixes the "receivers are also casters" with VSM in a more robust way.

romainguy (approved):

filament/src/ShadowMap.cpp
pixelflinger: I think it'll be less confusing to leave it this way, so we get values between 0 and 1.

bejado: Yeah, nevermind I was looking at the wrong equation

PR #4179, opened 2021-06-22
Vulkan: add support for depth resolves.

prideout:
The DepthResolve test now passes with Vulkan, although DoF still
triggers validation errors, which I'll look into next.


filament/backend/CMakeLists.txt
```diff
@@ -235,8 +235,8 @@ function(build_vkshader SOURCE TARGET_PATH)
             DEPENDS matc ${SOURCE_PATH}
             COMMENT "Building SPIR-V")
 endfunction()
-build_vkshader("BlitColor.vs" "BlitColorVs.spv")
-build_vkshader("BlitColor.fs" "BlitColorFs.spv")
+build_vkshader("BlitDepth.vs" "BlitDepthVs.spv")
+build_vkshader("BlitDepth.fs" "BlitDepthFs.spv")
```
romainguy, line 239: Do we smol-v compress shaders built outside of materials?

prideout: That's an interesting idea, currently our compileRawShader function does _not_ apply smolv compression...

prideout: Let me give this a shot.

prideout: @romainguy I implemented your idea and added a commit.

pixelflinger (approved):

PR #4184, opened 2021-06-23
Vulkan: fix swizzle support.

prideout:
When creating a texture with a swizzle, the swizzle should be applied to
the image view only when sampling from that texture.  When a swizzled
texture is used as a render target attachment, the image view should not
have any swizzle applied.


bejado (approved):

pixelflinger (approved):
nice one

PR #4185, opened 2021-06-23
Metal: ensure sample counts are supported by device

bejado:


pixelflinger (approved):

filament/backend/src/metal/MetalContext.h
```diff
@@ -55,6 +58,9 @@ struct MetalContext {
     bool supportsTextureSwizzling = false;
     uint8_t maxColorRenderTargets = 4;
 
+    // sampleCountLookup[requestedSamples] gives a <= sample count supported by the device.
+    std::array<uint8_t, MAX_SAMPLE_COUNT + 1> sampleCountLookup;
```
prideout, line 62: This is a neat way of doing it.  I might do it differently since Vulkan has multiple kinds of MSAA limits, depending on usage.

prideout (approved):

PR #4187, opened 2021-06-24
Rework a bit UniformBuffer and friends

pixelflinger:
Filament uses a handful of known UBO, the code for creating their
interface block is not needed in filament (via filabridge), so
reorganize things so that we can move that code to filamat.

On the other hand, filament needs the C structure corresponding to
the UBO, to find the fields offset and ubo size and name.

We now have a UibStructs.h in filabridge which only defines C struct
with a static name for the interface block's.

This should slightly reduce the size of filament proper.

We also now have a new helper TypedUniformBuffer<> which simply holds
one of these structs (or an array), and allows to set the fields
directly and more naturally from C++. std140 alignment is now left to
the caller, when going through that class.

### DIFF
--- a/filament/src/TypedUniformBuffer.h
+++ b/filament/src/TypedUniformBuffer.h
@@ -1,0 +1,76 @@
+/*
+ * Copyright (C) 2021 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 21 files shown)


romainguy (approved):

filament/src/TypedUniformBuffer.h
```diff
@@ -0,0 +1,77 @@
...
+ * limitations under the License.
+ */
+
+#ifndef TNT_FILAMENT_TYPEDUNIFORMBUFFER_H
+#define TNT_FILAMENT_TYPEDUNIFORMBUFFER_H
+
+#include "private/backend/DriverApi.h"
+
+#include <utils/compiler.h>
+#include <utils/debug.h>
```
prideout, line 23: I don't see panic or assert_invariant in this file, not sure if this include is needed.

For new files I usually put 2021 for the copyright notice.

prideout (approved):

filament/src/TypedUniformBuffer.h
pixelflinger: good points

PR #4189, opened 2021-06-24
gltfio: introduce support for extras strings.

prideout:
This allows clients to obtain the extras strings for the asset and
for individual nodes.  Note that extras for buffer views, accessors, etc
are not supported, although C++ clients can access the raw cgltf data
directly if they need to.


romainguy (approved):

mseare:
Hello!  How would one use this to get the extras from a Scene?  For example, in the attached sample glTF file, I have one scene and some extra data on it.  I'd like to get that extra data.
[extra_data_on_scene.gltf.zip](https://github.com/google/filament/files/10581849/extra_data_on_scene.gltf.zip)

PR #4194, opened 2021-06-24
Implement blurring of VSM shadow maps

pixelflinger:
This is still somewhat work-in-progress, but good enough to be useful.

Some caveats:
- large blurs don't work well with cascades
- the implementation uses more memory than it could

### DIFF
--- a/android/filament-android/src/main/java/com/google/android/filament/LightManager.java
+++ b/android/filament-android/src/main/java/com/google/android/filament/LightManager.java
@@ -326,6 +326,18 @@ public class LightManager {
          */
         @IntRange(from = 1)
         public int vsmMsaaSamples = 1;
+
+        /**
+         * Standard deviation of the VSM blur. Zero do disable.
+         * The maximum value is 21, which corresponds to a gaussian blur filter width
+         * of 125 pixels. The relation between the filter width and the standard deviation
+         * is roughly: stddev = (kernelWidth + 1) / 6.
+         * Some common values for blurStandardDeviation:
+         *  3x3 gaussian : 0.6667
+         *  5x5 gaussian : 1.0
+         *  9x9 gaussian : 1.6667
+         */
+        public float blurStandardDeviation = 0.0f;
     }
 
     public static class ShadowCascades {
@@ -454,7 +466,8 @@ public class LightManager {
                     options.mapSize, options.shadowCascades, options.cascadeSplitPositions,
                     options.constantBias, options.normalBias, options.shadowFar, options.shadowNearHint,
                     options.shadowFarHint, options.stable, options.screenSpaceContactShadows,
-                    options.stepCount, options.maxShadowDistance, options.vsmMsaaSamples);
+                    options.stepCount, options.maxShadowDistance, options.vsmMsaaSamples,
+                    options.blurStandardDeviation);
             return this;
         }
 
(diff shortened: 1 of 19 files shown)


android/filament-android/src/main/java/com/google/android/filament/LightManager.java
```diff
@@ -326,6 +326,18 @@ public void destroy(@Entity int entity) {
...
         @IntRange(from = 1)
         public int vsmMsaaSamples = 1;
+
+        /**
+         * Standard deviation of the VSM blur. Zero do disable.
+         * The maximum value is 21, which corresponds to a gaussian blur filter width
+         * of 125 pixels. The relation between the filter width and the standard deviation
+         * is roughly: stddev = (kernelWidth + 1) / 6.
+         * Some common values for blurStandardDeviation:
+         *  3x3 gaussian : 0.6667
```
romainguy, line 336: Should this API take a blur radius and we compute the standard deviation from that? It would be more user friendly.

libs/viewer/src/Settings.cpp
```diff
@@ -1168,7 +1170,8 @@ static std::ostream& operator<<(std::ostream& out, const LightManager::ShadowOpt
     math::float3 splitsVector = { splits[0], splits[1], splits[2] };
     return out << "{\n"
         << "\"vsm\": {\n"
-        << "\"msaaSamples\": " << int(in.vsm.msaaSamples) << "\n"
+        << "\"msaaSamples\": " << int(in.vsm.msaaSamples) << ",\n"
+        << "\"blurStandardDeviation\": " << int(in.vsm.blurStandardDeviation) << "\n"
```
prideout, line 1174: I think the cast-to-int is unintentional here.

android/filament-android/src/main/java/com/google/android/filament/LightManager.java
pixelflinger: Yes, I can probably do that.

libs/viewer/src/Settings.cpp
pixelflinger: Thanks!

filament/src/components/LightManager.cpp
```diff
@@ -172,19 +172,24 @@ void FLightManager::create(const FLightManager::Builder& builder, utils::Entity
...
-        shadowParams.options.stable = builder->mShadowOptions.stable;
-        shadowParams.options.polygonOffsetConstant = builder->mShadowOptions.polygonOffsetConstant;
-        shadowParams.options.polygonOffsetSlope = builder->mShadowOptions.polygonOffsetSlope;
-        shadowParams.options.screenSpaceContactShadows = builder->mShadowOptions.screenSpaceContactShadows;
-        shadowParams.options.stepCount = builder->mShadowOptions.stepCount;
-        shadowParams.options.maxShadowDistance = builder->mShadowOptions.maxShadowDistance;
+        shadowParams.options = builder->mShadowOptions;
+
+        // validate all shadow options
+        shadowParams.options.mapSize = clamp(shadowParams.options.mapSize, 0u, 2048u);
```
romainguy, line 178: We could allow up to 4096 right?

romainguy (approved):

prideout (approved):

romainguy (approved):

filament/src/components/LightManager.cpp
romainguy: Yeah, but looking at the list of GPUs that support only [2048](https://opengles.gpuinfo.org/listreports.php?capability=GL_MAX_TEXTURE_SIZE&esversion=2&value=2048) and not 4096, I'm not sure it's worth having that limitation. Devices that support 2048 represent [3.7% of the total population](https://opengles.gpuinfo.org/displaycapability.php?name=GL_MAX_TEXTURE_SIZE&esversion=2). Everything else supports 4096 or more.

PR #4207, opened 2021-06-28
fix LISPSM shadow virtual near/far planes shadow hint

pixelflinger:


filament/src/ShadowMapManager.cpp
```diff
@@ -175,6 +175,7 @@ void ShadowMapManager::render(FrameGraph& fg, FEngine& engine, FView& view,
                         // When rendering VSM shadow maps, we still need a depth texture for sorting.
                         auto depth = builder.createTexture("Temporary VSM Depth Texture", {
                                 .width = textureRequirements.size, .height = textureRequirements.size,
+                                .samples = options->vsm.msaaSamples,
```
bejado, line 178: Why is this needed? This seems to contradict the above comment.

pixelflinger: Yeah it's a bit confusing. So it's not strictly needed in fact. If you don't specify samples, the depth buffer will be created in "auto resolve" mode. However, on GLES, "auto resolve" of depth is not supported by the MSRTT extension, so it's done manually with a blit (meaning that internally a MS renderbuffer is created). However, in this case, the depth is temporary, so it's never resolved anyways. So the net result is that a temporary MS renderbuffer is created and discarded.

Or, you can specify the sample count (which also creates a MS renderbuffer), in which case the MSRTT extension is happy, you're essentially telling it that you won't need a "auto resolve".

bejado: Got it, thanks! 👍  Maybe add a quick note about explicitly setting the depth samples to avoid "auto resolve"

bejado (approved):

PR #4231, opened 2021-07-01
gltfio: Allow Java / Kotlin clients to customize MaterialProvider.

prideout:
This adds some significant JNI glue in order to make the Java API for
gltfio materials much more consistent with C++, and create feature
parity between the two languages.

The motivation is to allow Java clients to provide their own
ubershader-based solution or filamat-based solution.

- The `MaterialProvider` Java class is now an interface.
- The `UbershaderLoader` Java class is now a concrete implementation.
- Users can now write their own provider class using Java / Kotlin.

To streamline the amount of JNI work that occurs at run time, the
MaterialProvider Java interface uses ints in a couple spots instead
of enums. This sacrafices some readability for Java clients who wish
to write their own provider, but makes the JNI much simpler.

This replaces #3975.


android/filament-utils-android/src/main/java/com/google/android/filament/utils/ModelViewer.kt
```diff
@@ -110,7 +110,7 @@ class ModelViewer(val engine: Engine) : android.view.View.OnTouchListener {
         view.scene = scene
         view.camera = camera
 
-        assetLoader = AssetLoader(engine, MaterialProvider(engine), EntityManager.get())
+        assetLoader = AssetLoader(engine, UbershaderLoader(engine), EntityManager.get())
```
prideout, line 113: This Java was badly named, it was actually an ubershader loader. After this PR lands, it will be easy to add a feature to ModelViewer that allows clients to pass in a custom provider object.

android/gltfio-android/src/main/cpp/MaterialKey.cpp
```diff
@@ -0,0 +1,103 @@
...
+    dst.transmissionUV = env->GetIntField(src, field("transmissionUV", "I"));
+    dst.hasSheenColorTexture = env->GetBooleanField(src, field("hasSheenColorTexture", "Z"));
+    dst.sheenColorUV = env->GetIntField(src, field("sheenColorUV", "I"));
+    dst.hasSheenRoughnessTexture = env->GetBooleanField(src, field("hasSheenRoughnessTexture", "Z"));
+    dst.sheenRoughnessUV = env->GetIntField(src, field("sheenRoughnessUV", "I"));
+    dst.hasSheen = env->GetBooleanField(src, field("hasSheen", "Z"));
+    dst.hasIOR = env->GetBooleanField(src, field("hasIOR", "Z"));
+}
+
+void nativeToJava(JNIEnv* env, MaterialKey& src, jobject dst) {
```
pixelflinger, line 65: these could be private native getters.

android/gltfio-android/src/main/cpp/MaterialKey.h
```diff
@@ -0,0 +1,24 @@
...
+ */
+
+#include <jni.h>
+
+#include <gltfio/MaterialProvider.h>
+
+#define JAVA_MATERIAL_KEY "com/google/android/filament/gltfio/MaterialProvider$MaterialKey"
+
+void nativeFromJava(JNIEnv* env, gltfio::MaterialKey& dst, jobject src);
+void nativeToJava(JNIEnv* env, gltfio::MaterialKey& src, jobject dst);
```
pixelflinger, line 24: `src` should be const&

prideout:
NOTE: In a future PR we should probably add a fast path for the built-in UbershaderLoader thats skips the JNI back-and-forth.

android/gltfio-android/src/main/cpp/AssetLoader.cpp
```diff
@@ -23,18 +23,151 @@
...
+        assert_invariant(mDestroyMaterials);
+    }
+
+    ~JavaMaterialProvider() override {
+        mEnv->DeleteGlobalRef(mMaterialKeyClass);
+        mEnv->DeleteGlobalRef(mJavaProvider);
+        delete mPreviousMaterials;
+    }
+
+    MaterialInstance* createMaterialInstance(MaterialKey* config, UvMap* uvmap, const char* label) override {
```
pixelflinger, line 92: Is it not possible to call up to Java, and have all that java stuff be donne in Java?

pixelflinger (commented):
I feel like there is too much "java stuff" that's done on the C++ side

android/gltfio-android/src/main/cpp/MaterialKey.cpp
```diff
@@ -0,0 +1,103 @@
...
+
+#include <jni.h>
+
+#include <gltfio/MaterialProvider.h>
+
+#include "MaterialKey.h"
+
+using namespace gltfio;
+
+void nativeFromJava(JNIEnv* env, MaterialKey& dst, jobject src) {
```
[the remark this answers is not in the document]
prideout: There are many fields here, so it would be like over 30 arguments, are we sure that's an okay thing to do?  Agreed that I should at least cache all the field ids.

android/gltfio-android/src/main/cpp/MaterialKey.h
prideout: Good catch

android/gltfio-android/src/main/cpp/AssetLoader.cpp
prideout: This is called from the native code.  The point of this PR is to allow Java clients to write their own MaterialProvider class.

android/gltfio-android/src/main/cpp/AssetLoader.cpp
```diff
@@ -23,18 +23,151 @@
...
+    JavaMaterialProvider(JNIEnv* env, jobject provider) : mEnv(env),
+            mJavaProvider(env->NewGlobalRef(provider)) {
+        mMaterialKeyClass = env->FindClass(JAVA_MATERIAL_KEY);
+        mMaterialKeyClass = (jclass) env->NewGlobalRef(mMaterialKeyClass);
+        assert_invariant(mMaterialKeyClass);
+
+        mMaterialKeyConstructor = env->GetMethodID(mMaterialKeyClass, "<init>", "()V");
+        assert_invariant(mMaterialKeyConstructor);
+
+        jclass materialInstanceClass = env->FindClass("com/google/android/filament/MaterialInstance");
```
bejado, line 61: Just as a heads up, like with https://github.com/google/filament/commit/0a6a74c78ce18d3ad03a22c9aee408ce67f4f5ea, we'll have to cast the string literals to `char*`if this ever ends up in Google3.

android/gltfio-android/src/main/cpp/MaterialKey.cpp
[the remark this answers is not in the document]
pixelflinger: I think it's okay, and it'll prevent mistakes (like typos) by failing at compile time.  From a performance standpoint (and maybe even code size), it would probably be better. 
But I realize that this is called from native -- so if you're going to do a native->java transition for that, it's probably not worth it. However, if you already have a switch to java, you should do as much as possible there.

bejado (approved):

android/gltfio-android/src/main/cpp/AssetLoader.cpp
pixelflinger: yes but could use have a java method that's like "doAllTheJavaStuff()" that you call from native and does all that?

android/gltfio-android/src/main/cpp/AssetLoader.cpp
pixelflinger: oh god. good catch.

prideout: Luckily I think I'm in the clear, these env methods seem to take `const char*` rather than `char*`.

android/gltfio-android/src/main/cpp/AssetLoader.cpp
prideout: In this particular case, this idea would not result in an simplification, most of the work is copying between JNI and native.

android/gltfio-android/src/main/cpp/MaterialKey.cpp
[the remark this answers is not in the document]
prideout: @pixelflinger okay I added a commit that adds a helper class for caching these field ids.  Since the material key is an in-out argument, I think this is a better solution than a long argument list.

pixelflinger (approved):

android/gltfio-android/src/main/cpp/AssetLoader.cpp
bejado: I think the latest JDK correctly has them take a `const char*`, but Google3's version (for whatever reason) takes a `char*`.

prideout: Gah, okay, thanks for the heads up.

alexey-pelykh:
There's an issue with NoSuchMethod exception, see #4389

PR #4233, opened 2021-07-02
RenderPass and ShadowMap code cleanup

pixelflinger:
This is split in 3 CLs, but the main changes are:
- rework RenderPass allocations
- simplify ShadowMap cpde
See each CL for details.

### DIFF
--- a/filament/src/ShadowMapManager.cpp
+++ b/filament/src/ShadowMapManager.cpp
@@ -77,7 +77,7 @@ void ShadowMapManager::addSpotShadowMap(size_t lightIndex) noexcept {
 }
 
 void ShadowMapManager::render(FrameGraph& fg, FEngine& engine, FView& view,
-        backend::DriverApi& driver, RenderPass& pass) noexcept {
+        backend::DriverApi& driver, RenderPass const& pass) noexcept {
 
     constexpr size_t MAX_SHADOW_LAYERS =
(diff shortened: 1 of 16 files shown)


romainguy (approved):

filament/src/ShadowMapManager.cpp
```diff
@@ -88,37 +88,45 @@ void ShadowMapManager::render(FrameGraph& fg, FEngine& engine, FView& view,
...
+        if (!map.hasVisibleShadows() || range.empty()) {
             continue;
         }
-        pass.setVisibilityMask(VISIBLE_SPOT_SHADOW_RENDERABLE_N(i));
-        map.getShadowMap().render(driver, view.getVisibleSpotShadowCasters(), pass, view);
-        pass.clearVisibilityMask();
-        passes.emplace_back(&map, pass);
+        auto& entry = passList.emplace_back(&map, pass);
+        entry.pass.setVisibilityMask(VISIBLE_SPOT_SHADOW_RENDERABLE_N(i));
+        map.getShadowMap().render(driver, range, &entry.pass, view);
```
bejado, line 132: Should we still reset the visibility mask after rendering the shadowmap?

bejado (approved):

filament/src/ShadowMapManager.cpp
pixelflinger: No because `emplace_back` creates a new `RenderPass` each time.

PR #4237, opened 2021-07-02
moved HandleAllocator outside of the OpenGL backend

pixelflinger:
This means it can easily be used by our other backend.

HandleAllocator uses 3 pool (which currently can't be configured),
and is very efficient in terms of memory overhead and handle to
pointer conversion.

A Handle<> is just an offset into a large heap managed by the 3 pools.


prideout (approved):
Thanks for doing this, I'll be able to use this.

bejado (approved):
Nice, the API LGTM

PR #4238, opened 2021-07-02
Refactor Metal texture uploads

bejado:
Previously, Metal was never performing any texture conversions during upload. For example, `iblprefilter` loads `FLOAT` image data into the `R11F_G11_B10F` texture format. This PR fixes the issue, refactors the upload pipeline, and adds a new backend test for coverage.


pixelflinger:
Is this going in? :-)

bejado:
@pixelflinger Yes, I still need to fix sRGB texture uploads, then it will be ready for review.

pixelflinger (approved):

filament/backend/src/BackendUtils.cpp
```diff
@@ -326,6 +328,55 @@ size_t getBlockHeight(TextureFormat format) noexcept {
     }
 }
 
+bool reshape(const PixelBufferDescriptor& data, PixelBufferDescriptor& reshaped) {
+    if (data.format != PixelDataFormat::RGB && data.format != PixelDataFormat::RGB_INTEGER) {
```
pixelflinger, line 332: can we get a comment for this test, I'm not understanding it.

filament/backend/src/metal/MetalHandles.h
```diff
@@ -164,6 +163,19 @@ struct MetalProgram : public HwProgram {
     bool isValid = false;
 };
 
+struct PixelBufferShape {
+    size_t bytesPerPixel;
+    size_t bytesPerRow;
+    size_t bytesPerSlice;
+    size_t totalBytes;
```
pixelflinger, line 170: do all these need to be `size_t` as opposed to `uint32_t`?

filament/backend/src/metal/MetalHandles.mm
```diff
@@ -537,162 +521,242 @@ static MTLPixelFormat decidePixelFormat(id<MTLDevice> device, TextureFormat form
...
+    const size_t stagingBufferSize = shape.totalBytes;
+    const bool largeUpload = stagingBufferSize > deviceMaxBufferLength;
+
+    ASSERT_PRECONDITION(!nonBlittableTexture || !conversionNecessary,
+            "SAMPLER_2D_ARRAY, SAMPLER_3D, and SAMPLER_CUBEMAP texture uploads"
+            "do not support format conversions.");
+
+    ASSERT_PRECONDITION(!nonBlittableTexture || !largeUpload,
+            "SAMPLER_2D_ARRAY, SAMPLER_3D, and SAMPLER_CUBEMAP texture uploads"
+            "have a max size of %d bytes.", deviceMaxBufferLength);
```
pixelflinger, line 674: Is this a "todo" or just something we don't support in filament? If it's the later, we should probably have the ASSERT_PRECONDITION on the user facing API, and just use an assert() here.

pixelflinger (approved):
I'm not 100% sure I understand all the steps needed. If I'm getting this right, we use the DataReshaper (CPU) only when the source format is RGB (but all other cases, R, RG, RGBA bypass the DataReshaper), then the actual texture upload is done with the GPU? Correct?

What I'm not clear about is if the srd->dst conversions we handle are the same and exactly the same than GLES' (i.e. no more, no less)?

bejado:
> If I'm getting this right, we use the DataReshaper (CPU) only when the source format is RGB (but all other cases, R, RG, RGBA bypass the DataReshaper), then the actual texture upload is done with the GPU? Correct?

Correct. Since Metal does not support 3-component textures, we have to reshape the data on the CPU first. Then we use the GPU to upload to the actual texture, which also performs any necessary conversions.

> What I'm not clear about is if the srd->dst conversions we handle are the same and exactly the same than GLES' (i.e. no more, no less)?

I made sure to support at least what GLES supports. In practice however, the Metal backend should support additional conversions, for instance, `RGB32F` to `RGBA16F`. I don't think GLES explicitly supports that one.

filament/backend/src/metal/MetalHandles.mm
bejado: This is a limitation of the Metal backend at the moment, so more of a "todo". I'll add a TODO comment.

filament/backend/src/metal/MetalHandles.h
bejado: `uint32_t` should be fine.

PR #4247, opened 2021-07-06
ShadowMapManager: fix MSAN use-of-uninitialized-value.

prideout:
The operator!= in std::array compares SPLIT_COUNT elements, which
is potentially greater than cascadeCount, which was the number of
initialized elements in splitPercentages.

### DIFF
--- a/filament/src/ShadowMapManager.cpp
+++ b/filament/src/ShadowMapManager.cpp
@@ -380,8 +380,7 @@ ShadowMapManager::ShadowTechnique ShadowMapManager::updateCascadeShadowMaps(
 
     // We divide the camera frustum into N cascades. This gives us N + 1 split positions.
     // The first split position is the near plane; the last split position is the far plane.
-    std::array<float, CascadeSplits::SPLIT_COUNT> splitPercentages;
-    splitPercentages[0] = 0.0f;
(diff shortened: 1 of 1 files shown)


romainguy (approved):

romainguy:
This reminds me why I don't like `std::array` :)

filament/src/ShadowMapManager.cpp
```diff
@@ -380,7 +380,7 @@ ShadowMapManager::ShadowTechnique ShadowMapManager::updateCascadeShadowMaps(
 
     // We divide the camera frustum into N cascades. This gives us N + 1 split positions.
     // The first split position is the near plane; the last split position is the far plane.
-    std::array<float, CascadeSplits::SPLIT_COUNT> splitPercentages;
+    std::array<float, CascadeSplits::SPLIT_COUNT> splitPercentages{};
     splitPercentages[0] = 0.0f;
```
pixelflinger, line 384: This line is not needed anymore.

pixelflinger (changes requested):

filament/src/ShadowMapManager.cpp
prideout: Good catch.

Issue #4246, opened 2021-07-04
DataReshaper


Joaopmoliveira:
**Is your feature request related to a problem? Please describe.**
When reading the pixels from the GPU there is a stage related with reshaping the original image, i was wondering if this could be removed in specific use cases. 

**Describe the solution you'd like**
My question is if this step could be removed when rendering a standalone view with a user provided texture, thus the image layout is known and the copy into cpu could be speed up.

**Describe alternatives you've considered**
I do not totally understand the implications of what I am describing, but I assume in the Vulkan case a VKImage is created with a given layout, if this image is already in a layout which is the same as the Texture provided as a render target then it is expected that the user would provide a PixelBufferDescriptor with the same layout and thus a simple copy would suffice after the vkCmdBlitImage command.

**OS and backend**
Not specific to any device.

**Extra info**
I saw that in the backend tests (test_Blit.cpp) there is a call to a api.blit() method, is this is dangerous to turn into a public interface?

romainguy:
Reshaping only occurs for formats allowed by our API but not natively handled by Vulkan. If you use a format that Vulkan supports natively, there shouldn’t be any overhead. Do you have a specific example?

Joaopmoliveira:
This is more or less the code I have been using for testing. 

```c++
using closure_t = std::function<void(uint8_t* rgba, uint32_t width, uint32_t height)>;

	Engine* mEngine = Engine::create();

	View* mView = mEngine->createView();
	mView->setViewport({ 0, 0, swapchain_width, swapchain_height });

	const Viewport& vp = mView->getViewport();

	Texture* offscreenColorTexture = Texture::Builder()
		.width(vp.width).height(vp.height).levels(1)
		.usage(Texture::Usage::SAMPLEABLE | Texture::Usage::UPLOADABLE | Texture::Usage::COLOR_ATTACHMENT)
		.format(Texture::InternalFormat::RGBA8).build(*mEngine);

	RenderTarget* offscreenRenderTarget  = RenderTarget::Builder()
		.texture(RenderTarget::AttachmentPoint::COLOR, offscreenColorTexture)
		.build(*mEngine);

	Renderer* mRenderer = mEngine->createRenderer();

	Scene* mScene = mEngine->createScene();

	utils::Entity mCameraEntity = utils::EntityManager::get().create();
	Camera* mCamera = mEngine->createCamera(mCameraEntity);

	mView->setScene(mScene);
	mView->setCamera(mCamera);
[...]

PR #4248, opened 2021-07-06
Add simple ReadPixels performance test.

prideout:
Add simple ReadPixels performance test.

This simply lets the gtest framework report the total test time, which includes startup and teardown.  However we perform 100 runs of the render+readback code to reduce the run-to-run variance. For 100 runs of a 2000x2000 buffer, I see the following results on my mac (total time for all 100 runs):

- Vulkan: 1881 ms
- Metal: 1625 ms
- OpenGL: 978 ms

The test can be built & invoked like this:

```
./build.sh release backend_test_mac && \
    ./out/cmake-release/filament/backend/backend_test_mac -a opengl --gtest_filter="*ReadPixelsPerformance"
```

See also issue #4246.

### DIFF
--- a/filament/backend/test/test_ReadPixels.cpp
+++ b/filament/backend/test/test_ReadPixels.cpp
@@ -62,7 +62,12 @@ void main() {
 
 namespace test {
 
-TEST_F(BackendTest, ReadPixels) {
+class ReadPixelsTest : public BackendTest {
+public:
+    bool readPixelsFinished = false;
+};
+
+TEST_F(ReadPixelsTest, ReadPixels) {
     // These test scenarios use a known hash of the result pixel buffer to decide pass / fail,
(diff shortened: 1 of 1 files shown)


romainguy:
Are those numbers for all of the 100 runs, or per run?

filament/backend/test/test_ReadPixels.cpp
```diff
@@ -318,4 +318,101 @@ TEST_F(BackendTest, ReadPixels) {
     getDriver().purge();
 }
 
+static bool gReadPixelsFinished = false;
```
romainguy, line 321: It would be better to write a unit test that uses a custom class+instance instead of relying on a global field.

prideout:
> Are those numbers for all of the 100 runs, or per run?

All 100 runs. (gtest reports the total time of the test)

pixelflinger (approved):

PR #4261, opened 2021-07-07
Allow MSAA when post-processing is disabled

### LINKED ISSUE #2533: View.sampleCount not applied when post-processing disabled
There is code [in the Renderer](https://github.com/google/filament/blob/master/filament/src/Renderer.cpp#L418) to handle MSAA when rendering directly (i.e. no post-processing), but there is [also code that overrides the sample count](https://github.com/google/filament/blob/master/filament/src/Renderer.cpp#L212) when post-processing is disabled. MSAA appears to work if you remove that line, but I'm not sure if there are consequences I am missing.

Related: are their performance implications to using textures for MSAA rather than just enabling EGL_SAMPLE_BUFFERS when the EGL context is created? I can't quite work out what the [opaqueBlit](https://github.com/google/filament/blob/master/filament/src/PostProcessManager.cpp#L478) function actually does.

pixelflinger:
There was no good reason not allow this.
Fixes #2533

### DIFF
--- a/filament/src/Renderer.cpp
+++ b/filament/src/Renderer.cpp
@@ -239,7 +239,6 @@ void FRenderer::renderJob(ArenaScope& arena, FView& view) {
     auto taaOptions = view.getTemporalAntiAliasingOptions();
     if (!hasPostProcess) {
         // disable all effects that are part of post-processing
-        msaa = 1;
         dofOptions.enabled = false;
         bloomOptions.enabled = false;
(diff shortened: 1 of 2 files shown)


bejado (approved):
If you haven't already, might be worth running an automation test in gltf_viewer to make sure this works with all settings permutations.

romainguy (approved):

PR #4262, opened 2021-07-07
enable up to 6 spot shadows

pixelflinger:
This is a trivial change since we moved the lightspace computation
in the fragment shader. The limitation to 6 is due to how we're storing
the culling result.


romainguy:
Wait, wasn't the limitation also because of the number of varyings required?

romainguy (approved):

bejado (approved):
@romainguy The varyings limit is no longer an issue as the light-space position computation is now done in the fragment shader.

PR #4267, opened 2021-07-08
sample-gltf-viewer: fix remote control of IndirectLight.

prideout:


pixelflinger:
The ibl rotation didn't seem to work as well, maybe it's the same reason?

prideout:
Yes this fixes both  issues.

PR #4274, opened 2021-07-08
Skip lighting/shadowing when possible

pixelflinger:
We can skip lighting in the areas fully shadowed, we can also skip
shadowing outside the radius of punctual lights.

### DIFF
--- a/shaders/src/light_directional.fs
+++ b/shaders/src/light_directional.fs
@@ -35,6 +35,12 @@ void evaluateDirectionalLight(const MaterialInputs material,
 
     Light light = getDirectionalLight();
 
+#if defined(MATERIAL_CAN_SKIP_LIGHTING)
+    if (light.NoL <= 0.0) {
+        return;
+    }
+#endif
+
     float visibility = 1.0;
 #if defined(HAS_SHADOWING)
     if (light.NoL > 0.0) {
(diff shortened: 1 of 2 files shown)


shaders/src/light_directional.fs
```diff
@@ -58,13 +64,12 @@ void evaluateDirectionalLight(const MaterialInputs material,
         #if defined(MATERIAL_HAS_AMBIENT_OCCLUSION)
         visibility *= computeMicroShadowing(light.NoL, material.ambientOcclusion);
         #endif
-    } else {
+    }
 #if defined(MATERIAL_CAN_SKIP_LIGHTING)
+    if (visibility <= 0.0) {
```
romainguy, line 69: Visibility cannot be <= 0.0 if we don't enter the shadowing code, move inside the `if (light.NoL > 0.0) {`

shaders/src/light_punctual.fs
```diff
@@ -190,7 +196,7 @@ void evaluatePunctualLights(const MaterialInputs material,
         }
 #endif
 #if defined(MATERIAL_CAN_SKIP_LIGHTING)
-        if (light.NoL <= 0.0 || light.attenuation <= 0.0) {
+        if (visibility <= 0.0) {
```
romainguy, line 199: Same here

romainguy (approved):

PR #4277, opened 2021-07-09
Remote UI: auto load last loaded model.

prideout:
The most recently loaded model is now automatically re-loaded if you re-start the `sampler-gltf-viewer` app while keeping the remote page open.

Super useful in my case for `rgb-lights.glb` :)


pixelflinger (approved):
Yay!

romainguy (approved):

PR #4278, opened 2021-07-09
Work around Qualcomm issue with point lights.

### LINKED ISSUE #3602: Android: Point lights don't work with Vulkan + Qualcomm
Point lights don't seem to work in `sample-gltf-viewer` when using the Vulkan backend.

**To Reproduce**

I'm using the QR code branch to make it easy to upload a .glb test file: https://github.com/google/filament/pull/3599

1. Build Filament for Android:

```
./build.sh -c -i -p android release
```

2. Patch `sample-gltf-viewer` to enable the Vulkan backend and turn down IBL intensity:

```
--- a/android/samples/sample-gltf-viewer/src/main/java/com/google/android/filament/gltf/MainActivity.kt
+++ b/android/samples/sample-gltf-viewer/src/main/java/com/google/android/filament/gltf/MainActivity.kt
@@ -30,6 +30,7 @@ import androidx.camera.core.Preview
 import androidx.camera.lifecycle.ProcessCameraProvider
 import androidx.core.content.ContextCompat
 import androidx.lifecycle.LifecycleOwner
+import com.google.android.filament.Engine
 import com.google.android.filament.utils.KtxLoader
 import com.google.android.filament.utils.ModelViewer
 import com.google.android.filament.utils.Utils
@@ -89,7 +90,7 @@ class MainActivity : AppCompatActivity() {

         doubleTapDetector = GestureDetector(applicationContext, doubleTapListener)

-        modelViewer = ModelViewer(surfaceView)
+        modelViewer = ModelViewer(surfaceView, Engine.create(Engine.Backend.VULKAN))

         surfaceView.setOnTouchListener { _, event ->
             modelViewer.onTouchEvent(event)
@@ -256,7 +257,7 @@ class MainActivity : AppCompatActivity() {
         val ibl = "default_env"
         readCompressedAsset("envs/$ibl/${ibl}_ibl.ktx").let {
             scene.indirectLight = KtxLoader.createIndirectLight(engine, it)
-            scene.indirectLight!!.intensity = 30_000.0f
+            scene.indirectLight!!.intensity = 3000.0f
         }
         readCompressedAsset("envs/$ibl/${ibl}_skybox.ktx").let {
             scene.skybox = KtxLoader.createSkybox(engine, it)
```

3. Install on device

```
./gradlew samples:sample-gltf-viewer:InstallDebug
```

4. Run and upload [this glb test file](https://github.com/google/filament/files/6071072/rgb-lights.glb.zip), which contains 3 point lights.

5. Notice the difference between Vulkan / OpenGL:

Vulkan:

Screenshot_20210301-164027

OpenGL:

Screenshot_20210301-164122

It's peculiar on Vulkan, there are black smears where the point lights should be:

Screenshot_20210301-164027

 - Device: Pixel 4a
 - OS: Android 11 userdebug RQ1A.210205.004

prideout:
Fixes #3602.

### DIFF
--- a/shaders/src/shading_lit.fs
+++ b/shaders/src/shading_lit.fs
@@ -296,7 +296,7 @@ vec4 evaluateLights(const MaterialInputs material) {
 void addEmissive(const MaterialInputs material, inout vec4 color) {
 #if defined(MATERIAL_HAS_EMISSIVE)
     highp vec4 emissive = material.emissive;
-    highp float attenuation = mix(1.0, frameUniforms.exposure, emissive.w);
+    highp float attenuation = mix(1.0, getExposure(), emissive.w);
     color.rgb += emissive.rgb * (attenuation * color.a);
 #endif
 }
--- a/shaders/src/shading_unlit.fs
+++ b/shaders/src/shading_unlit.fs
@@ -1,7 +1,7 @@
 void addEmissive(const MaterialInputs material, inout vec4 color) {
 #if defined(MATERIAL_HAS_EMISSIVE)
     highp vec4 emissive = material.emissive;
-    highp float attenuation = mix(1.0, frameUniforms.exposure, emissive.w);
+    highp float attenuation = mix(1.0, getExposure(), emissive.w);
     color.rgb += emissive.rgb * (attenuation * color.a);
 #endif
 }
--- a/libs/filamat/src/UibGenerator.cpp
+++ b/libs/filamat/src/UibGenerator.cpp
@@ -66,7 +66,7 @@ UniformInterfaceBlock const& UibGenerator::getPerViewUib() noexcept  {
             // ibl
             .add("iblLuminance",            1, UniformInterfaceBlock::Type::FLOAT)
             // camera
-            .add("exposure",                1, UniformInterfaceBlock::Type::FLOAT)
+            .add("exposure",                1, UniformInterfaceBlock::Type::FLOAT, Precision::HIGH) // high precision to work around #3602 (qualcomm)
             .add("ev100",                   1, UniformInterfaceBlock::Type::FLOAT)
             // ibl
             .add("iblSH",                   9, UniformInterfaceBlock::Type::FLOAT3)
--- a/shaders/src/common_getters.fs
+++ b/shaders/src/common_getters.fs
@@ -64,6 +64,8 @@ highp float getUserTimeMod(float m) {
 
 /** @public-api */
 float getExposure() {
+    // NOTE: this is a highp uniform only to work around #3602 (qualcomm)
+    // We are intentionally casting it to mediump here, as per the Materials doc.
     return frameUniforms.exposure;
 }
 


prideout:
This is not a precision bug, but specifying highp here is the easiest way to trick the optimizer into generating SPIRV code that doesn't break Qualcomm.  I tried lots of other hacks to no avail.

romainguy (approved):
@pixelflinger

It seems fine but I want to make sure we don't end up doing extra work in fp32 because of this. Everything related to exposure was carefully designed to be in fp16 as much as possible.

pixelflinger:
Woah, this is wild.

So it's this a highp problem in the end? I don't even understand why this specific uniform has an effect.

Did this end up changing a lot of things in the final optimized shader?

prideout:
> Woah, this is wild.
> 
> So it's this a highp problem in the end? I don't even understand why this specific uniform has an effect.
> 
> Did this end up changing a lot of things in the final optimized shader?

This uniform is used to compute an intensity that seems to get zeroed out by the qcom driver.  By using highp here, we are forcing the optimizer to avoid simplification of an expression that seems to trigger the bug.

So... it's actually not highp related at all, but I wasn't able to trick the optimizer in any other way.

shaders/src/light_punctual.fs
```diff
@@ -128,7 +128,7 @@ Light getLight(const uint index) {
     // and populate the Light structure
     Light light;
     light.colorIntensity.rgb = colorIntensity.rgb;
-    light.colorIntensity.w = computePreExposedIntensity(colorIntensity.w, frameUniforms.exposure);
+    light.colorIntensity.w = computePreExposedIntensity(colorIntensity.w, getExposure());
```
romainguy, line 131: Sooo… here it would make more sense to keep accessing the uniform since this particular function will make the value highp again to do a highp multiplication.

shaders/src/shading_lit.fs
```diff
@@ -296,7 +296,7 @@ vec4 evaluateLights(const MaterialInputs material) {
 void addEmissive(const MaterialInputs material, inout vec4 color) {
 #if defined(MATERIAL_HAS_EMISSIVE)
     highp vec4 emissive = material.emissive;
-    highp float attenuation = mix(1.0, frameUniforms.exposure, emissive.w);
+    highp float attenuation = mix(1.0, getExposure(), emissive.w);
```
pixelflinger, line 299: I don't think attenuation needs to be `highp` here, actually -- probably for another PR though.

shaders/src/shading_unlit.fs
```diff
@@ -1,7 +1,7 @@
 void addEmissive(const MaterialInputs material, inout vec4 color) {
 #if defined(MATERIAL_HAS_EMISSIVE)
     highp vec4 emissive = material.emissive;
-    highp float attenuation = mix(1.0, frameUniforms.exposure, emissive.w);
+    highp float attenuation = mix(1.0, getExposure(), emissive.w);
```
pixelflinger, line 4: I don't think attenuation needs to be `highp` here, actually -- probably for another PR though.

romainguy: It has to be, it's effectively an exposure value and exposure needs to be highp. It's pre-exposed intensities (lightIntensity * exposure) that do not need to be highp.

PR #4285, opened 2021-07-09
Fix, MSVC's std::sort fails on Zip2Iterator

### LINKED ISSUE #3395: Point lights and spot lights don't work on desktop
Never been able to get lights other than directional to actually work

https://user-images.githubusercontent.com/1667289/103274569-e6822b80-49b9-11eb-9b1a-3a78c714d13a.mp4

bejado:
We should never be returning `value_type` from the iterator. That would only ever make sense, for example, in a `ConstZip2Iterator` class, _not_ for a `const Zip2Iterator`.

This led to `std::sort` not working correctly on Windows, which caused flickering punctual lights. Fixes #3395

### DIFF
--- a/libs/utils/include/utils/Zip2Iterator.h
+++ b/libs/utils/include/utils/Zip2Iterator.h
@@ -51,7 +51,6 @@ public:
 
     using value_type = std::pair<Val1, Val2>;
     using reference = Ref;
-    using pointer = value_type*;
     using difference_type = ptrdiff_t;
     using iterator_category = std::random_access_iterator_tag;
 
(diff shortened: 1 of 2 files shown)


libs/utils/include/utils/Zip2Iterator.h
```diff
@@ -60,11 +60,9 @@ class Zip2Iterator {
     Zip2Iterator(Zip2Iterator const& rhs) noexcept = default;
     Zip2Iterator& operator=(Zip2Iterator const& rhs) = default;
 
-    value_type operator*() const { return { *mIt.first, *mIt.second }; }
-    reference  operator*()       { return { *mIt.first, *mIt.second }; }
+    reference operator*() const { return { *mIt.first, *mIt.second }; }
```
pixelflinger, line 63: shouldn't this be a `const_reference`?

libs/utils/include/utils/Zip2Iterator.h
```diff
@@ -60,11 +60,9 @@ class Zip2Iterator {
...
 
-    const value_type operator[](size_t n) const { return *(*this + n); }
-          reference  operator[](size_t n)       { return *(*this + n); }
+    reference operator[](size_t n) const { return *(*this + n); }
```
pixelflinger, line 65: shouldn't this be a `const_reference`?

pixelflinger (changes requested):
nice one

libs/utils/include/utils/Zip2Iterator.h
bejado: No, that would only be if this were `ConstZip2Iterator`

pixelflinger (approved):

PR #4287, opened 2021-07-09
API CHANGE: honor user-defined precision, rename SamplerPrecision.

prideout:
matc now honors 'precision' on non-samplers in mat files, and
the filamat API now allows clients to specify precision for
non-samplers.

This involved flattening a union that is internal to filamat.

### DIFF
--- a/android/filamat-android/src/main/java/com/google/android/filament/filamat/MaterialBuilder.java
+++ b/android/filamat-android/src/main/java/com/google/android/filament/filamat/MaterialBuilder.java
@@ -247,13 +247,30 @@ public class MaterialBuilder {
 
     @NonNull
     public MaterialBuilder uniformParameter(@NonNull UniformType type, String name) {
-        nMaterialBuilderUniformParameter(mNativeObject, type.ordinal(), name);
+        nMaterialBuilderUniformParameter(mNativeObject, type.ordinal(), name,
+            ParameterPrecision.DEFAULT.ordinal());
+        return this;
+    }
+
+    @NonNull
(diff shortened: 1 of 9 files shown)


android/filamat-android/src/main/java/com/google/android/filament/filamat/MaterialBuilder.java
```diff
@@ -259,7 +259,7 @@ public MaterialBuilder uniformParameterArray(@NonNull UniformType type, int size
 
     @NonNull
     public MaterialBuilder samplerParameter(@NonNull SamplerType type, SamplerFormat format,
```
romainguy, line 261: You should add new methods for non-sampler parameters to specify the precision. Java doesn't have default parameter values, so just use overloading instead.

romainguy (changes requested):

pixelflinger (approved):

android/filamat-android/src/main/java/com/google/android/filament/filamat/MaterialBuilder.java
prideout: Good catch, added a commit.

bejado (commented):
> Will this require a middle version number bump? What about `MATERIAL_VERSION`?

Due to the enum rename, I do think we should bump the middle version number, yes.  It doesn't break backward compatibility of materials so I don't think we need to bump MATERIAL_VERSION.

romainguy (approved):

romainguy:
Merging this PR because I need it to update mine :)

PR #4290, opened 2021-07-09
gltfio: fix precision in KHR_texture_transform

### LINKED ISSUE #4253: Vulkan: texture filtering looks wrong (Qualcomm)
On a Pixel 4, load up bistro and disable scale-to-unit cube to move the camera close to the floor.  Does not occur on Mali devices or MoltenVK.

The can also be reproduced using the default BusterDrone model, just zoom in to the floor.  GL and Vulkan comparisons are attached.

OpenGL
Vulkan

prideout:
Fixes #4253 but also requires #4287 in order to work.


prideout:
Oh, this is failing presubmit because I need to merge the other PR.

romainguy (approved):

PR #4291, opened 2021-07-10
Add support for KHR_materials_volume

romainguy:
Adds support for `KHR_materials_volume` to `gltfio`. It is important to note that we don't currently handle the special
case of thickness set to 0 which requires to change the refraction model to a thin model. We do however — thanks to a
fairly gross hack — handle thickness scaling.

Screen Shot 2021-07-09 at 4 58 11 PM

This PR also addresses rendering issues with transmission, most notably by using alpha-as-coverage instead of the fade blending mode.

Screen Shot 2021-07-09 at 7 04 25 PM


pixelflinger (approved):

alexey-pelykh:
Will there be a support for MaterialProvider in Android?

romainguy:
@alexey-pelykh How so? This extension is handled by gltfio so it will work on Android.

alexey-pelykh:
@romainguy but it's not exposed in `MaterialProvider.MaterialKey`?

romainguy:
Ah you're right, that's an oversight. I'll fix it.

PR #4296, opened 2021-07-12
default to openGL when windows does not support VULKAN

bigianb:
My windows laptop does not support vulkan (or at least the cmake does not detect it). That means the samples crash because the backend is defaulting to vulkan. This change just makes the backend default openGL in the case that Vulkan is not available.


prideout:
Interesting, this change seems reasonable but we don't intend to keep supporting OpenGL on desktop forever.  So I'll defer to @pixelflinger.

pixelflinger (approved):

PR #4299, opened 2021-07-12
remove unneeded namespace qualifications

pixelflinger:


bejado (approved):

prideout (approved):
Side note: `std::string` is the only place where we should keep the namespace qualifier.

PR #4301, opened 2021-07-13
Add new sample app: image_viewer

romainguy:
This app shows how to load and visualize images. It's also useful to work
on color grading and other post-processing effects by just loading an HDR
or EXR image as a reference.

This change also adds a new exposure setting to the color grading LUT.

Comparisons of the "Filmic" and "EVILS" tone mappers using image_viewer:

Screen Shot 2021-07-12 at 6 39 33 PM
Screen Shot 2021-07-12 at 6 39 36 PM

### DIFF
--- a/android/filament-utils-android/src/main/java/com/google/android/filament/utils/Manipulator.java
+++ b/android/filament-utils-android/src/main/java/com/google/android/filament/utils/Manipulator.java
@@ -267,7 +267,7 @@ public class Manipulator {
          * @return this <code>Builder</code> object for chaining calls
          */
         @NonNull
-        public Builder imageEntity(float a, float b, float c, float d) {
+        public Builder groundPlane(float a, float b, float c, float d) {
             nBuilderGroundPlane(mNativeBuilder, a, b, c, d);
             return this;
         }
--- a/android/filament-utils-android/src/main/java/com/google/android/filament/utils/Manipulator.java
+++ b/android/filament-utils-android/src/main/java/com/google/android/filament/utils/Manipulator.java
@@ -267,7 +267,7 @@ public class Manipulator {
          * @return this <code>Builder</code> object for chaining calls
          */
         @NonNull
-        public Builder groundPlane(float a, float b, float c, float d) {
+        public Builder imageEntity(float a, float b, float c, float d) {
             nBuilderGroundPlane(mNativeBuilder, a, b, c, d);
             return this;
         }
--- a/filament/include/filament/ColorGrading.h
+++ b/filament/include/filament/ColorGrading.h
@@ -158,7 +158,8 @@ public:
          * Adjusts the exposure of this image. The exposure is specified in stops:
          * each stop brightens (positive values) or darkens (negative values) the image by
          * a factor of 2. This means that an exposure of 3 will brighten the image 8 times
-         * more than an exposure of 0 (2^3 = 8 and 2^0 = 1).
+         * more than an exposure of 0 (2^3 = 8 and 2^0 = 1). Contrary to the camera's exposure,
+         * this setting is applied after all post-processing (bloom, etc.) are applied.
          *
          * @param exposure Value in EV stops. Can be negative, 0, or positive.
          *
--- a/filament/include/filament/ColorGrading.h
+++ b/filament/include/filament/ColorGrading.h
@@ -64,6 +64,7 @@ class FColorGrading;
  * ========
  *
  * The various transforms held by ColorGrading are applied in the following order:
+ * - Exposure
  * - White balance
  * - Channel mixer
  * - Shadows/mid-tones/highlights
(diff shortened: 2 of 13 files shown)


filament/include/filament/ColorGrading.h
```diff
@@ -152,6 +154,18 @@ class UTILS_PUBLIC ColorGrading : public FilamentAPI {
          */
         Builder& toneMapping(ToneMapping toneMapping) noexcept;
 
+        /**
+         * Adjusts the exposure of this image. The exposure is specified in stops:
```
pixelflinger, line 158: Maybe we can have a small explanation of how that differs from the camera exposure and how the two interact?
e.g. I could imagine scenarios where people do it wrong and adjust this exposure usual is the camera's?

romainguy: It does the same thing except because it's in the color grading pass it's applied post screen space effects. I don't think it makes sense to try and compare it to the camera.

romainguy: Done.

samples/materials/image.mat
```diff
@@ -0,0 +1,39 @@
...
+        uv.t = 1.0 - uv.t;
+
+        material.baseColor = textureLod(materialParams_image, uv.st, 0.0);
+    }
+}
+
+vertex {
+    void materialVertex(inout MaterialVertexInputs material) {
+        material.imageUV.st = getPosition().st * 0.5 + 0.5;
+    }
```
pixelflinger, line 38: Generally we try to get the vertex{} section appear before the fragment{} section.

samples/materials/image.mat
```diff
@@ -0,0 +1,39 @@
+material {
+    name : Image,
+    parameters : [
+        {
+            type : sampler2d,
+            name : image
+        },
+        {
+            type : mat3,
+            name : transform
```
pixelflinger, line 10: precision should probably be `highp`

samples/materials/image.mat
```diff
@@ -0,0 +1,39 @@
...
+    culling: none
+}
+
+fragment {
+    void material(inout MaterialInputs material) {
+        prepareMaterial(material);
+
+        vec2 uv = (materialParams.transform * vec3(saturate(variable_imageUV.st), 1.0)).st;
+        if (uv.s > 1.0 || uv.s < 0.0 || uv.t < 0.0 || uv.t > 1.0) discard;
+        uv.t = 1.0 - uv.t;
```
pixelflinger, line 29: is this correct on all backends?

pixelflinger (changes requested):
I think the `mat3` in the material should be `highp`

samples/materials/image.mat
romainguy: I'm not using NDC or frag coords to compute the UVs. They are based solely on the vertex positions I've specified so it's backend independent (tested on Metal, Vulkan and GL).

android/filament-utils-android/src/main/java/com/google/android/filament/utils/Manipulator.java
```diff
@@ -267,7 +267,7 @@ public Builder flightMoveDamping(float damping) {
          * @return this <code>Builder</code> object for chaining calls
          */
         @NonNull
-        public Builder groundPlane(float a, float b, float c, float d) {
```
prideout, line 270: Was this intentional?

samples/image_viewer.cpp
```diff
@@ -0,0 +1,386 @@
...
+#include <getopt/getopt.h>
+
+#include <math/half.h>
+#include <math/vec3.h>
+#include <math/vec4.h>
+#include <math/mat3.h>
+#include <math/norm.h>
+
+#include <imgui.h>
+#include <filagui/ImGuiExtensions.h>
```
prideout, line 48: Is this include necessary?  Also, I use copyright 2021 for new files.

samples/image_viewer.cpp
```diff
@@ -0,0 +1,386 @@
...
+                -tx,        -ty,         1.0f
+            );
+
+            app.scene.imageMaterial->setDefaultParameter("transform", transform);
+        }
+    };
+
+    auto postRender = [&app](Engine* engine, View* view, Scene* scene, Renderer* renderer) {
+        // TODO
+    };
```
prideout, line 375: You can remove this, FilamentApp supplies a default argument.

samples/image_viewer.cpp
```diff
@@ -0,0 +1,386 @@
...
+
+    FilamentApp& filamentApp = FilamentApp::get();
+
+    filamentApp.setDropHandler([&] (std::string path) {
+        loadImage(app, app.engine, Path(path));
+    });
```
prideout, line 381: Neat!

android/filament-utils-android/src/main/java/com/google/android/filament/utils/Manipulator.java
romainguy: Grrr, that's CLion's refactoring tools that do things where they are not supposed to. Good catch.

romainguy: I renamed a variable called `groundPlane` in C++ code using the rename refactor, and it renamed a Java method. Wth.

samples/image_viewer.cpp
romainguy: Ah yeah I know, but I was originally planning on using it.

prideout (approved):

pixelflinger (approved):

PR #4305, opened 2021-07-13
Improvements to image_viewer

romainguy:
Image loading errors are handled more gracefully by showing a
background color.

Added UI to set the background color.

Replaced fragment discards with solid background color.

### DIFF
--- a/samples/materials/image.mat
+++ b/samples/materials/image.mat
@@ -9,6 +9,14 @@ material {
             type : mat3,
             name : transform,
             precision : high
+        },
+        {
+            type : float3,
+            name : backgroundColor
+        },
+        {
+            type : int,
+            name : showImage
         }
     ],
     variables : [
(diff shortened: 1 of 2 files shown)


samples/materials/image.mat
```diff
@@ -9,6 +9,14 @@ material {
...
             name : transform,
             precision : high
+        },
+        {
+            type : float3,
+            name : backgroundColor
+        },
+        {
+            type : int,
+            name : showImage
```
prideout, line 19: We support `bool` here.

prideout (approved):

samples/materials/image.mat
romainguy: We do but there are bugs with qcom drivers so we've been using int instead in several materials.

prideout: Interesting -- I never knew or have completely forgotten.

PR #4307, opened 2021-07-13
Prep for using new handle allocator in Vulkan.

prideout:
Previously we were zero filling the data blob used to store all Vulkan
handles. This occured in VulkanDriver::alloc_handle().

There were several places that depended on the zero fill. These
have now been fixed up so that the handle can be constructed in a
more natural way.

Also: remove the OPENGL specific constant from HandleAllocator. This
breaks in non-CMake build systems (e.g. Bazel) that do not provide the
constant at a global level.

### DIFF
--- a/filament/backend/src/HandleAllocator.cpp
+++ b/filament/backend/src/HandleAllocator.cpp
@@ -74,8 +74,7 @@ HandleBase::HandleId HandleAllocator::allocateHandleSlow(size_t size) noexcept {
 
     if (UTILS_UNLIKELY(id == (HEAP_HANDLE_FLAG|1u))) { // meaning id was zero
         PANIC_LOG("HandleAllocator arena is full, using slower system heap. Please increase "
-                  "FILAMENT_OPENGL_HANDLE_ARENA_SIZE_IN_MB which is currently set to %u MiB.",
-                FILAMENT_OPENGL_HANDLE_ARENA_SIZE_IN_MB);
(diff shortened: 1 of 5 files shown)


bejado (approved):
Nice, I'll probably need to do something similar for Metal.

filament/backend/src/HandleAllocator.cpp
```diff
@@ -74,8 +74,7 @@ HandleBase::HandleId HandleAllocator::allocateHandleSlow(size_t size) noexcept {
 
     if (UTILS_UNLIKELY(id == (HEAP_HANDLE_FLAG|1u))) { // meaning id was zero
         PANIC_LOG("HandleAllocator arena is full, using slower system heap. Please increase "
-                  "FILAMENT_OPENGL_HANDLE_ARENA_SIZE_IN_MB which is currently set to %u MiB.",
-                FILAMENT_OPENGL_HANDLE_ARENA_SIZE_IN_MB);
+                  "the appropriate constant (e.g. FILAMENT_OPENGL_HANDLE_ARENA_SIZE_IN_MB).");
```
bejado, line 77: We should probably rename this to something more generic now that other backends are going to use it, like `FILAMENT_HANDLE_ARENA_SIZE_IN_MB`.

prideout: Well, Vulkan might need a larger arena size since stuff like its HwRenderPrimitive handle is (by necessity) much larger.  I guess I was thinking of making separate constants for each backend...at least for now.

bejado: Got it, yeah that makes sense.

PR #4309, opened 2021-07-13
Parameterize HandleAllocator and use it in VulkanDriver.

prideout:


prideout:
Please review CL's individually.

filament/backend/include/private/backend/HandleAllocator.h
```diff
@@ -36,6 +36,7 @@ namespace filament::backend {
 /*
  * A utility class to efficiently allocate and manage Handle<>
  */
+template <int P0, int P1, int P2>
```
pixelflinger, line 39: use `size_t` instead of `int` (or at least `unsigned`)

filament/backend/src/HandleAllocator.cpp
```diff
@@ -24,26 +24,29 @@ namespace filament::backend {
...
+template <int P0, int P1, int P2>
 UTILS_NOINLINE
-HandleAllocator::Allocator::Allocator(AreaPolicy::HeapArea const& area)
+HandleAllocator<P0, P1, P2>::Allocator::Allocator(AreaPolicy::HeapArea const& area)
         : mArea(area) {
     // TODO: we probably need a better way to set the size of these pools
     const size_t unit = area.size() / 32;
     const size_t offsetPool1 =      unit;
     const size_t offsetPool2 = 16 * unit;
     char* const p = (char*)area.begin();
```
pixelflinger, line 35: We probably also need some parameters to define the relative size of each pool arena, can be in another PR.

filament/backend/src/opengl/OpenGLDriver.h
```diff
@@ -249,7 +249,8 @@ class OpenGLDriver final : public backend::DriverBase {
 
     // Memory management...
 
-    backend::HandleAllocator mHandleAllocator;
+    // See also the explicit template instantiation in HandleAllocator.cpp
+    backend::HandleAllocator<16, 64, 208> mHandleAllocator;
```
pixelflinger, line 253: You could define an alias inside the `#if FILAMENT_SUPPORTS_FOO`, for e.g.
`using HandleAllocatorGL = HandleAllocator<16,64,208>;`
So that we don't have to repeat the sizes here?

pixelflinger (changes requested):
very nice

filament/backend/src/opengl/OpenGLDriver.h
prideout: Yeah but the `#if FILAMENT_SUPPORTS_FOO` thingie is in a cpp file, and C++ doesn't let you instantiate a template using an alias, as far as I can tell.

prideout: Maybe I could do this with a macro...

filament/backend/include/private/backend/HandleAllocator.h
prideout: Done.

filament/backend/src/opengl/OpenGLDriver.h
prideout: Done.

pixelflinger (approved):

PR #4317, opened 2021-07-14
matdbg: fix active variant list, update README.

prideout:


libs/matdbg/include/matdbg/ShaderInfo.h
```diff
@@ -22,9 +22,15 @@
...
 #include <filaflat/MaterialChunk.h>
 
+#include <private/filament/Variant.h>
+
+#include <utils/bitset.h>
+
 namespace filament {
 namespace matdbg {
 
+using VariantList = utils::bitset<uint64_t, VARIANT_COUNT>;
```
romainguy, line 32: Should we share this type between filament and matdbg? It seems it would be easy to change it on one side without updating the other.

romainguy (approved):

bejado (approved):

libs/matdbg/include/matdbg/ShaderInfo.h
prideout: Yes I struggled to find a good place.  Added a commit that consolidates it to `Variant.h`.

PR #4319, opened 2021-07-14
Implement sharing of bones for skinning

pixelflinger:
The API now allows to create a standalone BufferObject containing
Bones for skinning. This BufferObject can even be very large and
contains bone data for several renderable.

A new API on RenderableManager allows to set the BufferObject to use.

This is still WIP:
- currently not possible to set the binding offset
- currently not possible to go back to the old API after using the new one

### DIFF
--- a/filament/include/filament/RenderableManager.h
+++ b/filament/include/filament/RenderableManager.h
@@ -36,11 +36,13 @@ namespace utils {
 
 namespace filament {
 
+class BufferObject;
 class Engine;
 class IndexBuffer;
 class Material;
 class MaterialInstance;
 class Renderer;
+class SkinningBuffer;
 class VertexBuffer;
 
 class FEngine;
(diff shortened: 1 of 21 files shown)


filament/include/filament/RenderableManager.h
```diff
@@ -239,9 +241,40 @@ class UTILS_PUBLIC RenderableManager : public FilamentAPI {
...
+         * Enables GPU vertex skinning for up to 255 bones, 0 by default.
+         *
+         * Skinning Buffer mode must be enabled.
+         *
+         * Each vertex can be affected by up to 4 bones simultaneously. The attached
+         * VertexBuffer must provide data in the \c BONE_INDICES slot (uvec4) and the
+         * \c BONE_WEIGHTS slot (float4).
+         *
+         * See also RenderableManager::setSkinningBuffer(), which can be called on a per-frame basis
+         * to advance the animation.
```
prideout, line 264: Nit with this comment.  It would be unusual to advance an animation by swapping out the skinning buffer.  It would be more common for clients to update the bone matrices to their interpolated values by calling `SkinningBuffer::setBones()` on a per-frame basis.

filament/include/filament/SkinningBuffer.h
```diff
@@ -0,0 +1,109 @@
...
+         * @param engine Reference to the filament::Engine to associate this IndexBuffer with.
+         *
+         * @return pointer to the newly created object or nullptr if exceptions are disabled and
+         *         an error occurred.
+         *
+         * @exception utils::PostConditionPanic if a runtime error occurred, such as running out of
+         *            memory or other resources.
+         * @exception utils::PreConditionPanic if a parameter to a builder function was invalid.
+         *
+         * @see IndexBuffer::setBuffer
```
prideout, line 83: Probably a type to mention `IndexBuffer` here?

filament/include/filament/SkinningBuffer.h
```diff
@@ -0,0 +1,120 @@
...
+         * some memory overhead. This memory overhead can be mitigated by using the same
+         * SkinningBuffer to store the bone information for multiple RenderPrimitives.
+         *
+         * @param boneCount Number of bones the skinning buffer can hold.
+         * @return A reference to this Builder for chaining calls.
+         */
+        Builder& boneCount(uint32_t boneCount) noexcept;
+
+        /**
+         * The new buffer is created with identity bones
```
prideout, line 66: Might want to mention in the comment what the default state for this (true or false).

filament/src/View.cpp
```diff
@@ -574,10 +574,6 @@ void FView::prepare(FEngine& engine, backend::DriverApi& driver, ArenaScope& are
     s.fogInscatteringSize  = fogOptions.inScatteringSize;
     s.fogColorFromIbl      = fogOptions.fogColorFromIbl ? 1.0f : 0.0f;
 
-    // upload the renderables's dirty UBOs
-    engine.getRenderableManager().prepare(driver,
-            renderableData.data<FScene::RENDERABLE_INSTANCE>(), merged);
-
```
prideout, line 580: Interesting that this can be removed.  I don't totally get it, but I believe it.  :)

pixelflinger: Yeah, this is because before we were storing the data in a staging "heap" area, but now we issue a driver command each time the user calls setBones.

prideout (approved):

August1996:
are there any samples for using `SkinningBuffer`?😊

PR #4326, opened 2021-07-15
Switch Metal driver to use HandleAllocator

bejado:
Everything seems to be working smoothly. I want to run a few performance tests before merging.


filament/backend/src/metal/MetalHandles.h
```diff
@@ -258,7 +258,9 @@ class MetalRenderTarget : public HwRenderTarget {
     id<MTLTexture> multisampledDepth = nil;
 };
 
-class MetalFence : public HwFence {
+// MetalFence is used to implement both Fences and Syncs.
+// There's no diamond problem, because HwBase (superclass of HwFence and HwSync) is empty.
+class MetalFence : public HwFence, public HwSync {
```
pixelflinger, line 263: maybe add a static_assert for the emptyness of HwBase?

pixelflinger (approved):

filament/backend/src/metal/MetalHandles.h
bejado: Good idea.

filament/backend/src/metal/MetalDriver.mm
```diff
@@ -55,7 +55,8 @@
 MetalDriver::MetalDriver(backend::MetalPlatform* platform) noexcept
         : DriverBase(new ConcreteDispatcher<MetalDriver>()),
         mPlatform(*platform),
-        mContext(new MetalContext) {
+        mContext(new MetalContext),
+        mHandleAllocator("Handles", FILAMENT_OPENGL_HANDLE_ARENA_SIZE_IN_MB * 1024U * 1024U) {
```
prideout, line 59: I created a new constant for VULKAN so we might as well do the same for Metal, at least for now, and make it the same value.  Otherwise it's strange to refer to an OPENGL thing here.

bejado: Yeah, sounds good.

PR #4330, opened 2021-07-16
New tone mapper API

romainguy:
- `ColorGrading::Builder::toneMapping(ColorGrading::ToneMapping)` is now deprecated
- New API: `ColorGrading::Builder::toneMapper(const ToneMapper*)` 
- New tone mapper: `GenericToneMapper`, a configurable tone mapper
- Removed `REINHARD` tonemap operator


filament/src/ToneMapper.cpp
```diff
@@ -14,11 +14,22 @@
...
-using namespace filament::math;
+#include "ColorSpace.h"
+
+#include <math/vec3.h>
+#include <math/scalar.h>
+
+#include <iostream>
+
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wweak-vtables"
```
pixelflinger, line 27: We shouldn't need that.

filament/src/ToneMapper.cpp
```diff
@@ -14,11 +14,22 @@
...
-#include "ToneMapping.h"
+#include <filament/ToneMapper.h>
 
-using namespace filament::math;
+#include "ColorSpace.h"
+
+#include <math/vec3.h>
+#include <math/scalar.h>
+
+#include <iostream>
```
pixelflinger, line 24: We shouldn't need iostream

filament/src/ToneMapper.cpp
romainguy: Oh yeah that's a leftover

filament/src/ToneMapper.cpp
romainguy: Yeah same it's from debugginh

filament/include/filament/ToneMapper.h
```diff
@@ -0,0 +1,53 @@
...
+/**
+ *
+ */
+struct UTILS_PUBLIC ToneMapper {
+    virtual ~ToneMapper() noexcept = default;
+
+    virtual math::float3 operator()(math::float3 c) const noexcept;
+};
+
+struct ACESToneMapper : public ToneMapper {
```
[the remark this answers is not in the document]
romainguy: Should I do copy/move/assignment as well?

pixelflinger (approved):

filament/include/filament/ToneMapper.h
```diff
@@ -0,0 +1,230 @@
...
+ * The tone mapping curve is defined by 5 parameters:
+ * - contrast: controls the contrast of the curve
+ * - shoulder: controls the shoulder of the curve, i.e. how quickly scene
+ *             referred values map to output white
+ * - midGrayIn: sets the input middle gray
+ * - midGrayOut: sets the output middle gray
+ * - hdrMax: defines the maximum input value that will be mapped to
+ *           output white
+ */
+struct UTILS_PUBLIC GenericToneMapper final : public ToneMapper {
```
pixelflinger, line 133: Because `GenericToneMapper` has an allocated field (`mOptions`) it's not safe to copy it. I would suggest to at least add a deleted copy ctor and operator= and implement the move versions.
```
GenericToneMapper(GenericToneMapper const&) = delete;
GenericToneMapper& operator=(GenericToneMapper const&) = delete;
GenericToneMapper(GenericToneMapper&& rhs)  noexcept : mOptions(rhs.mOptions) { rhs.mOptions = nullptr; }
GenericToneMapper& operator=(GenericToneMapper& rhs) noexcept { /* same as above */ }
```

pixelflinger (approved):
LGTM but `GenericToneMapper` could be made safer.

filament/include/filament/ToneMapper.h
romainguy: Done.

PR #4341, opened 2021-07-19
Add bilateral threshold in SSAO options.

pixelflinger:
This can have a dramatic effect on SSAO and needs 
to be adjusted based on the scene's scale/units.

The default value used to be about 6cm, and has 
been changed to 5mm.

### DIFF
--- a/android/filament-android/src/main/java/com/google/android/filament/View.java
+++ b/android/filament-android/src/main/java/com/google/android/filament/View.java
@@ -170,6 +170,12 @@ public class View {
         public float intensity = 1.0f;
 
         /**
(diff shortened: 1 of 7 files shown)


android/filament-android/src/main/java/com/google/android/filament/View.java
```diff
@@ -169,6 +169,12 @@
          */
         public float intensity = 1.0f;
 
+        /**
+         * Depth distance that constitute an edge for filtering. Must be positive.
+         * Default is 5mm.
```
romainguy, line 174: Maybe mention it should be adjusted based on the scene/scale.

romainguy (approved):

PR #4342, opened 2021-07-20
associate sidecar MS buffer to its texture instead of RenderTarget

### LINKED ISSUE #4338: Performance issue with render target created/destroyed on high resolution displays
**Describe the bug**
Using a 4K display and MSAA x2, creating and destroying render targets at each frame has a big performance impact when they could be re-used from frame to frame. The issue is less glaring when not using MSAA or PP, which is a hint the performance hit is related to the size of the RTs.

This is very similar to this issue I reported 2 years ago, but I'm not sure if I can or should reopen it ?
https://github.com/google/filament/issues/1460

**To Reproduce**
Displaying a simple scene with MSAAx2 and profiling it, I get these results which highlights the costs of recreating the RT :

Function Name	Total CPU [unit, %]
| - filament::backend::ConcreteDispatcher<filament::OpenGLDriver>::createRenderTarget	492 (21.53%)
| - filament::backend::invoke<filament::OpenGLDriver,void __cdecl(filament::backend::Handle<filament::backend::HwRenderTarget>,enum filament::backend::TargetBufferFlags,unsigned int,unsigned int,unsigned char,filament::backend::MRT,filament::backend::TargetBufferInfo,filament::backend::TargetBufferInfo),filament::OpenGLDriver &,filament::backend::Handle<filament::backend::HwRenderTarget>,enum filament::backend::TargetBufferFlags,unsigned int,unsigned int,unsigned char,filament::backend::MRT,filament::backend::TargetBufferInfo,filament::backend::TargetBufferInfo>	492 (21.53%)
| - filament::OpenGLDriver::createRenderTargetR	492 (21.53%)
| - filament::OpenGLDriver::framebufferTexture	488 (21.36%)
| - filament::backend::ConcreteDispatcher<filament::OpenGLDriver>::destroyRenderTarget	405 (17.72%)
| - filament::OpenGLDriver::destroyRenderTarget	405 (17.72%)

Edit: I tried to profile in the sandbox app (out of my project) but I don't have symbols in release there. I expect it'd show the same results though.

 - OS: Windows
 - GPU: NVIDIA GTX 1080
 - Backend: OpenGL

pixelflinger:
this allows the sidecar buffer to have the same lifetime than the
texture it represents, and therefore participate to the caching of
that texture.

Fixes #4338

### DIFF
--- a/filament/backend/src/opengl/OpenGLDriver.h
+++ b/filament/backend/src/opengl/OpenGLDriver.h
@@ -113,7 +113,8 @@ public:
             GLuint id = 0;          // texture or renderbuffer id
             GLenum target = 0;
             GLenum internalFormat = 0;
-            mutable GLsync fence = nullptr;
+            GLuint sidecarRenderBufferMS = 0;  // multi-sample sidecar renderbuffer
+            mutable GLsync fence = {};
 
             // texture parameters go here too
             GLfloat anisotropy = 1.0;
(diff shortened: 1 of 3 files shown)


romainguy (approved):

filament/backend/src/opengl/OpenGLDriver.h
```diff
@@ -113,7 +113,8 @@ class OpenGLDriver final : public backend::DriverBase {
             GLuint id = 0;          // texture or renderbuffer id
             GLenum target = 0;
             GLenum internalFormat = 0;
-            mutable GLsync fence = nullptr;
+            GLuint sidecarRenderBufferMS = 0;  // multi-sample sidecar renderbuffer
```
prideout, line 116: I will do something similar except in my case I will probably have a pointer here instead of a low-level handle, because there are several low-level objects associated with a single texture (VkDeviceMemory, VkImage, VkImageView...).  Since Vulkan handles are 8 bytes, a pointer isn't any worse.

PR #4344, opened 2021-07-20
 matdbg: add support for editing Vulkan shaders

prideout:
matdbg now links in filamat and provides a language tab at the top of the UI.

If a Vulkan shader is selected and edits are made in the GLSL tab, then glslang is used to regenerate the SPIR-V.

In the future we would like to support editing the original GLSL (not the transpiled GLSL) but this provides a short-term solution to Vulkan editing.

### DIFF
--- a/libs/matdbg/src/ShaderReplacer.cpp
+++ b/libs/matdbg/src/ShaderReplacer.cpp
@@ -193,10 +193,14 @@ bool ShaderReplacer::replaceShaderSource(ShaderModel shaderModel, uint8_t varian
     }
 
     // Copy the new package from the stringstream into a ChunkContainer.
+    // The memory gets freed by DebugServer, which has ownership over the material package.
     const size_t size = tstream.str().size();
     uint8_t* data = new uint8_t[size];
     memcpy(data, tstream.str().data(), size);
+
+    assert_invariant(mEditedPackage == nullptr);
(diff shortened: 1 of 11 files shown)


libs/matdbg/src/ShaderReplacer.cpp
```diff
@@ -159,6 +200,93 @@ bool ShaderReplacer::replaceShaderSource(backend::ShaderModel shaderModel, uint8
...
+    // Append the new chunks for Dictionary* and Material*.
+    if (!shaderIndex.isEmpty()) {
+        shaderIndex.replaceShader(shaderModel, variant, stage, source, sourceLength);
+        shaderIndex.writeBlobsChunk(mDictionaryTag, tstream);
+        shaderIndex.writeShadersChunk(mMaterialTag, tstream);
+    }
+
+    // Copy the new package from the stringstream into a ChunkContainer.
+    const size_t size = tstream.str().size();
+    uint8_t* data = new uint8_t[size];
```
bejado, line 284: Does this leak memory?

prideout: Yeah it gets freed by the DebugServer, which isn't obvious so I added some comments in a new commit.  Thanks for pointing this out.

bejado: I see, thanks!

bejado (approved):

PR #4349, opened 2021-07-20
Metal: correctly handle byte offsets for updateIndexBuffer and updateBufferObject

bejado:


filament/backend/src/opengl/OpenGLDriver.cpp
```diff
@@ -1643,6 +1643,7 @@ void OpenGLDriver::updateBufferObject(
 
     if (bo->gl.binding == GL_UNIFORM_BUFFER) {
         // TODO: use updateBuffer() for all types of buffer? Make sure GL supports that.
+        // TODO: take byteOffset into account.
```
romainguy, line 1646: Seems like a pretty important TODO?

bejado: I thought so too, but none of the backends (except for Metal now, after this PR lands) are respecting it.

romainguy: Why does Metal need to respect it and not the others then?

romainguy (approved):

filament/backend/src/opengl/OpenGLDriver.cpp
bejado: There's no reason in particular Metal needs to, other than for correctly implementing the driver API.

@pixelflinger We should either fix the other backends, or remove `byteOffset` from the driver API.

pixelflinger: This is completely an oversight from my part. It should take this into account. This never was a problem until now because the old UniformBuffer code didn't take this parameter because we didn't need it. With the new code, we're always passing 0 for UBOs.
In any case, we should fix that.
Thank you for catching this!

pixelflinger: One more semi important thing...

When `BufferUsage::STREAM` is specified, `byteOffset` is meaningless and must always be zero. I realize I have not explained well what `STREAM` mode means.  This was a way to do very efficient partial updates of UBOs automatically, using `glMapBuffer`.
In `STREAM` mode, we have a large buffer, and each time we update it we "advance" into it, the backend is supposed to keep track of the where we're at and bind the buffer at that location. So the "byteOffset" is managed automatically.

The API allows for completely ignoring the `STREAM` mode, In this case, the backend only sees partial updates, as usual.

`STREAM` mode is useful for the per-renderable UBO, because this way we have a single large UBO where we store all the transforms (less overhead than having one per renderable), but since that changes each frame due to culling, we end-up uploading *less* than the UBO size, which implies that we're doing partial updates, which are notoriously slow. The idea is that we would use a different region of the buffer each time, until it's full, at which point a new one is allocated (and the old one orphaned).
It might be worth implementing this in metal/vulkan and you can look at the gl backend code for reference, I think it's pretty self-explanatory, but I can help too. But that's just a potential optimization.

BTW, an implication of `STREAM` mode is that even if you don't implement it with "map buffer", you are allowed to create a new buffer instead of uploading a sub-region of the existing one (and orphan the old one). In GL speak, this means you can do a 'glBufferData()` instead of a `glBufferSubData()` which is always slower.

bejado: Cool, thanks for the explanation. I think `STREAM` mode would be fairly easy to implement with Metal. I'll try experimenting with it.

PR #4350, opened 2021-07-21
Optimize matrix multiplications to avoid allocations

romainguy:
The previous code was a lot cleaner but caused 1+N (where N is
the dimension of the matrix) on each multiply.


prideout (approved):
Did not check the math, I trust you. :)

romainguy:
> Did not check the math, I trust you. :)

I added unit tests before making the change in my upstream repo to make sure I would not mess up the math :)

PR #4351, opened 2021-07-21
gltfio: fix AssetLoader leak, remove unwanted destructor.

prideout:
This fixes two bugs that combined in a way that prevented a double-free
error.

The destructor of the impl class for AssetLoader was never called, which
was sort of good, because it should not destroy the material provider.
The asset loader was never meant to have ownership over the provider.

### DIFF
--- a/libs/gltfio/src/AssetLoader.cpp
+++ b/libs/gltfio/src/AssetLoader.cpp
@@ -102,11 +102,9 @@ struct FAssetLoader : public AssetLoader {
         FilamentInstance** instances, size_t numInstances);
     FilamentInstance* createInstance(FFilamentAsset* primary);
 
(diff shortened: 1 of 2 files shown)


romainguy (approved):

libs/gltfio/src/AssetLoader.cpp
```diff
@@ -1268,7 +1264,7 @@ AssetLoader* AssetLoader::create(const AssetConfiguration& config) {
 }
 
 void AssetLoader::destroy(AssetLoader** loader) {
```
pixelflinger, line 1266: If we follow the Filament convention, `AssetLoader::destroy` should call `FAssetLoader::destroy` which would definitely prevent this kinds of mistakes.

pixelflinger (changes requested):

PR #4362, opened 2021-07-21
Metal/Vulkan: Fix reshaped image uploads with buffer padding or offsets

bejado:
When reshaping pixels, we need to make sure the new `PixelBufferDescriptor` has the same attributes, like stride, left, and top.


romainguy (approved):

pixelflinger (approved):
good one!

PR #4363, opened 2021-07-21
fix an issue where a sampler could fail to be updated

pixelflinger:
SamplerGroup was comparing texture handles to decide if a texture needed
to be updated, however, texture handles are (quickly) recycled and
therefore can't be used for that purpose. e.g. if a texture is destroyed,
its handle could be reused quickly by another texture, if that texture 
is now set on the SamplerGroup, it will ignore it, thinking it's not
different.

### DIFF
--- a/filament/backend/src/SamplerGroup.cpp
+++ b/filament/backend/src/SamplerGroup.cpp
@@ -62,13 +62,10 @@ SamplerGroup& SamplerGroup::setSamplers(SamplerGroup const& rhs) noexcept {
     return *this;
 }
 
-void SamplerGroup::setSampler(size_t index, Sampler const& sampler) noexcept {
-    if (index < mBuffer.size()) {
-        auto& cur = mBuffer[index];
-        if (cur.t != sampler.t || cur.s.u != sampler.s.u) {
-            cur = sampler;
-            mDirty.set(index);
-        }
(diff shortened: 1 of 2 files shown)


romainguy (approved):

bejado (approved):
We should also cherry-pick to rc/1.11.1 branch

filament/backend/src/SamplerGroup.cpp
```diff
@@ -62,13 +62,10 @@ SamplerGroup& SamplerGroup::setSamplers(SamplerGroup const& rhs) noexcept {
...
 
-void SamplerGroup::setSampler(size_t index, Sampler const& sampler) noexcept {
-    if (index < mBuffer.size()) {
-        auto& cur = mBuffer[index];
-        if (cur.t != sampler.t || cur.s.u != sampler.s.u) {
-            cur = sampler;
-            mDirty.set(index);
-        }
+void SamplerGroup::setSampler(size_t index, Sampler sampler) noexcept {
+    if (UTILS_LIKELY(index < mBuffer.size())) {
```
pixelflinger, line 66: could you put a comment in the code that we can't compare Handle<> and also a TODO we need to find a way to not mark dirty if the texture doesn't actually change

PR #4372, opened 2021-07-22
don't use Vulkan subpasses on Apple M1

pixelflinger:
It looks like they are not working, at least with the current version
of the drivers and/or MoltenVK. It is not impossible that we have a
bug in how we set the subpasses as well. This will need more investigation
later.

### DIFF
--- a/filament/backend/src/vulkan/VulkanDriver.cpp
+++ b/filament/backend/src/vulkan/VulkanDriver.cpp
@@ -806,6 +806,13 @@ bool VulkanDriver::isRenderTargetFormatSupported(TextureFormat format) {
 }
 
 bool VulkanDriver::isFrameBufferFetchSupported() {
+    // TODO: We might need a better way to handle vendor-specific workaround
+    // Disable framebuffer fetch on APPLE M1 (vendor 0x106b, device 0xa140)
+    // Subpasses, don't seen to work on M1, this possibly needs more investigation.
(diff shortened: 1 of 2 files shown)


filament/backend/src/vulkan/VulkanDriver.cpp
```diff
@@ -806,6 +806,13 @@ bool VulkanDriver::isRenderTargetFormatSupported(TextureFormat format) {
 }
 
 bool VulkanDriver::isFrameBufferFetchSupported() {
+    // TODO: We might need a better way to handle vendor-specific workaround
+    // Disable framebuffer fetch on APPLE M1 (vendor 0x106b, device 0xa140)
+    // Subpasses, don't seen to work on M1, this possibly needs more investigation.
+    VkPhysicalDeviceProperties const& deviceProperties = mContext.physicalDeviceProperties;
+    if ((deviceProperties.vendorID == 0x106b) && (deviceProperties.deviceID == 0xa140)) {
```
romainguy, line 813: Could we create constants for those?

romainguy (approved):

filament/backend/src/vulkan/VulkanDriver.cpp
pixelflinger: Since this is the only hack of this type we have right now, I want to let @prideout decide how to handle these, I don't want to make more changes than necessary.

bejado (approved):

PR #4374, opened 2021-07-22
cleanup colorGradingAsSubpass shader

pixelflinger:
It was confusing how it reused the input as a temporary variable.

### DIFF
--- a/filament/src/materials/colorGrading/colorGradingAsSubpass.mat
+++ b/filament/src/materials/colorGrading/colorGradingAsSubpass.mat
@@ -36,7 +36,7 @@ material {
     ],
     outputs : [
         {
-            name : color,
+            name : dummy,
             target : color,
             type : float4,
             location : 0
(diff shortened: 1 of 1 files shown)


romainguy (approved):

filament/src/materials/colorGrading/colorGradingAsSubpass.mat
```diff
@@ -36,7 +36,7 @@ material {
     ],
     outputs : [
         {
-            name : color,
+            name : dummy,
```
bejado, line 39: Why do we still need this output?

pixelflinger: Ah yes... we do. Because matc crashes without it. This is something I wanted to investigate next.

bejado: Hm, yeah that sounds like a bug.

PR #4392, opened 2021-07-27
Implement specular AO with screen space bent normals

pixelflinger:
SSAO is optionally augmented to evaluate bent normals, later used for specular AO.

before:
before
after:
after

### DIFF
--- a/filament/src/materials/ssao/bilateralBlurBentNormals.mat
+++ b/filament/src/materials/ssao/bilateralBlurBentNormals.mat
@@ -1,5 +1,5 @@
 material {
-    name : bilateralBlur,
+    name : bilateralBlurBentNormals,
     parameters : [
         {
             type : sampler2dArray,
@@ -22,6 +22,18 @@ material {
         {
             type : float[16],
             name : kernel
+        }
+    ],
+    outputs : [
+        {
+            name : aoData,
+            target : color,
+            type : float3
+        },
+        {
+            name : bnData,
+            target : color,
+            type : float3
         }
     ],
     variables : [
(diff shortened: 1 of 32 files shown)


bejado (approved):
Does this require recompiling materials due to the use of an ARRAY texture now?

filament/src/materials/ssao/bilateralBlurBentNormals.mat
```diff
@@ -0,0 +1,135 @@
...
+vertex {
+    void postProcessVertex(inout PostProcessVertexInputs postProcess) {
+        postProcess.vertex.xy = postProcess.normalizedUV;
+    }
+}
+
+fragment {
+    #include "ssaoUtils.fs"
+
+    void dummy(){}
```
romainguy, line 56: Why is this needed again?

filament/src/materials/ssao/bilateralBlurBentNormals.mat
```diff
@@ -0,0 +1,135 @@
...
+    void postProcess(inout PostProcessInputs postProcess) {
+        highp vec2 uv = variable_vertex.xy; // interpolated at pixel's center
+
+        float ao;
+        highp float depth;
+        highp float sampleDepth;
+        vec3 bn;
+
+        vec3 data = textureLod(materialParams_ssao, vec3(uv, 0.0), 0.0).rgb;
+        ao = data.r;
```
romainguy, line 85: Why not declare the variables here were you initialize them?

filament/src/materials/ssao/bilateralBlurBentNormals.mat
```diff
@@ -0,0 +1,135 @@
...
+        depth = unpack(data.gb);
+
+        if (data.g * data.b == 1.0) {
```
romainguy, line 88: Should it be >= 1.0 just in case?

filament/src/materials/ssao/bilateralBlurBentNormals.mat
```diff
@@ -0,0 +1,135 @@
...
+            bilateral = weight * bilateralWeight(depth, sampleDepth);
+            totalWeight += bilateral;
+            tapBN(materialParams_ssao, uv - offset, bn);
+            sumAO += ao * bilateral;
+            sumBN += bn * bilateral;
+
+            offset += materialParams.axis;
+        }
+
+        ao = sumAO * (1.0 / totalWeight);
```
romainguy, line 125: Factor 1/totalWeight in a variable to help the compiler?

libs/filabridge/src/SibGenerator.cpp
```diff
@@ -45,7 +45,7 @@ SamplerInterfaceBlock const& SibGenerator::getPerViewSib(uint8_t variantKey) noe
             .add("froxels",       Type::SAMPLER_2D,         Format::UINT,    Precision::MEDIUM)
             .add("iblDFG",        Type::SAMPLER_2D,         Format::FLOAT,   Precision::MEDIUM)
             .add("iblSpecular",   Type::SAMPLER_CUBEMAP,    Format::FLOAT,   Precision::MEDIUM)
-            .add("ssao",          Type::SAMPLER_2D,         Format::FLOAT,   Precision::MEDIUM)
+            .add("ssao",          Type::SAMPLER_2D_ARRAY,   Format::FLOAT,   Precision::MEDIUM)
```
romainguy, line 48: We need to increase the material version number because of this

romainguy (approved):

filament/src/materials/ssao/bilateralBlurBentNormals.mat
pixelflinger: It's just for CLion's glsl editor

filament/src/materials/ssao/bilateralBlurBentNormals.mat
pixelflinger: No it can't be > 1, since `.g` and `.b` are 0 to 1, also .`gb` encode a depth scaled between 0 and 1 and this encoding exists only because we're storing the depth in 8 bits channels.

libs/filabridge/src/SibGenerator.cpp
pixelflinger: ah good point.

PR #4398, opened 2021-07-28
MaterialKey: expose constrainMaterial to Android

alexey-pelykh:
cc @romainguy

### DIFF
--- a/android/gltfio-android/src/main/cpp/MaterialKey.cpp
+++ b/android/gltfio-android/src/main/cpp/MaterialKey.cpp
@@ -69,11 +69,6 @@ void MaterialKeyHelper::init(JNIEnv* env) {
     hasIOR = field("hasIOR", "Z");
 }
 
-extern "C" JNIEXPORT void JNICALL
-Java_com_google_android_filament_gltfio_MaterialProvider_00024MaterialKey_nGlobalInit(JNIEnv* env, jclass) {
-    MaterialKeyHelper::get().init(env);
-}
-
 void MaterialKeyHelper::copy(JNIEnv* env, MaterialKey& dst, jobject src) {
     dst.doubleSided = env->GetBooleanField(src, doubleSided);
     dst.unlit = env->GetBooleanField(src, unlit);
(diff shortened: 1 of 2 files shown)


android/gltfio-android/src/main/cpp/MaterialKey.cpp
```diff
@@ -72,6 +72,32 @@ Java_com_google_android_filament_gltfio_MaterialProvider_00024MaterialKey_nGloba
...
+    auto& helper = MaterialKeyHelper::get();
+    helper.copy(env, nativeMaterialKey, materialKey);
+
+    UvMap nativeUvMap = {};
+    constrainMaterial(&nativeMaterialKey, &nativeUvMap);
+
+    // Copy the UvMap results from the native array into the JVM array.
+    jint* elements = env->GetIntArrayElements(uvMap, nullptr);
+    if (elements) {
+        const size_t javaSize = env->GetArrayLength(uvMap);
```
romainguy, line 90: Use `std::copy_n` instead. For example:

```
std::copy_n(asset->getEntities(),
            std::min(available, (jsize) asset->getEntityCount()), entities);
```

android/gltfio-android/src/main/java/com/google/android/filament/gltfio/MaterialProvider.java
```diff
@@ -70,6 +70,12 @@ public MaterialKey() {}
             nGlobalInit();
         }
         private static native void nGlobalInit();
+
+        public void constrainMaterial(@NonNull @Size(min = 8) int[] uvmap) {
```
romainguy, line 74: Please keep the static native methods last like we do in the other JNI bindings. Also, could you add a documentation block to explain what this API is for?

romainguy (changes requested):

android/gltfio-android/src/main/cpp/MaterialKey.cpp
alexey-pelykh: Sure, but I've copied that snippet from the other part of Filament, literally neighboring file...

android/gltfio-android/src/main/java/com/google/android/filament/gltfio/MaterialProvider.java
alexey-pelykh: `constrainMaterial` has an explanation in native code, it's just a wrapper. Should I copy it over?

romainguy: That's usually what we do yes. More importantly the doc should mention why a user of gltfio should care or use this API.

romainguy (approved):
Thank you!

android/gltfio-android/src/main/cpp/MaterialKey.cpp
alexey-pelykh: Ugh, copy_n won't work - types are different :|

romainguy: Argh, sorry about that. That explains the original code :(

romainguy (approved):

PR #4400, opened 2021-07-28
gltfio: support material extras

alexey-pelykh:


libs/gltfio/src/AssetLoader.cpp
```diff
@@ -1012,9 +1020,17 @@ MaterialInstance* FAssetLoader::createMaterialInstance(const cgltf_material* inp
             break;
     }
 
+    // Check if this material has an extras string.
+    CString extras;
+    const cgltf_size extras_size = inputMat->extras.end_offset - inputMat->extras.start_offset;
+    if (extras_size > 0) {
```
alexey-pelykh, line 1026: @pixelflinger @prideout this one would be as optimal as it can get using your approach to extras.

android/gltfio-android/src/main/java/com/google/android/filament/gltfio/MaterialProvider.java
```diff
@@ -99,9 +99,10 @@ public void constrainMaterial(@NonNull @Size(min = 8) int[] uvmap) {
      * @param uvmap Output argument that gets populated with a small table that maps from a glTF uv
      *              index to a Filament uv index (0 = UNUSED, 1 = UV0, 2 = UV1).
      * @param label Optional tag that is not a part of the cache key.
+     * @param extras Optional extras that are not a part of the cache key.
      */
     public @Nullable MaterialInstance createMaterialInstance(MaterialKey config,
-            @NonNull @Size(min = 8) int[] uvmap, @Nullable String label);
+            @NonNull @Size(min = 8) int[] uvmap, @Nullable String label, @Nullable String extras);
```
romainguy, line 105: This is unnecessary API breakage. You should keep the existing method, create a new one with the new argument, and have the old method call the new one by passing `null` for the extras.

android/gltfio-android/src/main/java/com/google/android/filament/gltfio/UbershaderLoader.java
```diff
@@ -53,7 +53,7 @@ public void destroy() {
     }
 
     public @Nullable MaterialInstance createMaterialInstance(MaterialKey config,
-            @NonNull @Size(min = 8) int[] uvmap, @Nullable String label) {
+            @NonNull @Size(min = 8) int[] uvmap, @Nullable String label, @Nullable String extras) {
         long nativeMaterialInstance = nCreateMaterialInstance(mNativeObject, config, uvmap, label);
         return nativeMaterialInstance == 0 ? null : new MaterialInstance(null, nativeMaterialInstance);
```
romainguy, line 58: Same here

libs/gltfio/src/AssetLoader.cpp
```diff
@@ -1012,9 +1020,17 @@ MaterialInstance* FAssetLoader::createMaterialInstance(const cgltf_material* inp
...
+        extras = CString(srcAsset->json + inputMat->extras.start_offset, extras_size);
+    }
+
     // This not only creates a material instance, it modifies the material key according to our
     // rendering constraints. For example, Filament only supports 2 sets of texture coordinates.
-    MaterialInstance* mi = mMaterials->createMaterialInstance(&matkey, uvmap, inputMat->name);
+    MaterialInstance* mi = mMaterials->createMaterialInstance(&matkey, uvmap, inputMat->name,
```
romainguy, line 1032: You would have to document in createMaterialInstance() that the lifetime of `extras` is only valid for the duration of that function call since you allocate the `CString` on the stack.

third_party/models/BusterDrone/scene.gltf
```diff
@@ -9101,7 +9101,10 @@
         0,
         0,
         0
-      ]
+      ],
+      "extras": {
```
romainguy, line 9105: Please don't change this model. It's not ours and we don't want to fork it.

romainguy (changes requested):

android/gltfio-android/src/main/java/com/google/android/filament/gltfio/MaterialProvider.java
alexey-pelykh: May I ask who actually uses that API except for Filament itself?

romainguy: You apparently for starters :) And it doesn't matter, it's a public API and should be treated as such. We sometimes break them when we can't do otherwise but this is very easy to support properly.

romainguy: BTW you should also document the format of that string. What does it contain? JSON?

alexey-pelykh: > You apparently for starters :) 

Kind-a yes, since extras support was our initiative.

> And it doesn't matter, it's a public API and should be treated as such. We sometimes break them when we can't do otherwise but this is very easy to support properly.

The API was introduced a month ago, and Filament breaks API every now and then. In patch releases.

> engine: Removed the "EVILS" tone mapping operator [⚠️ API Change].

And at the same time requesting to be strict in backward compatibility?

> BTW you should also document the format of that string. What does it contain? JSON?

Yes, JSON or null. Valid point.

libs/gltfio/src/AssetLoader.cpp
alexey-pelykh: Reasonable point.

third_party/models/BusterDrone/scene.gltf
alexey-pelykh: Reasonable point.

android/gltfio-android/src/main/java/com/google/android/filament/gltfio/MaterialProvider.java
romainguy: > And at the same time requesting to be strict in backward compatibility?

Yes we do break backward compatibility sometimes, that doesn't mean we do it *every* time. For instance we recently changed how the tone mappers work, but we kept the older API and re-implemented it on top of the new API for compatibility. This is a very easy change to make here so there's no reason not to do it.

alexey-pelykh: I should note that there's no base class for `MaterialProvider`, so the interface would have to have two methods, and it will be a breaking change in any case since all derived classes would have to implement the extra method.
I'm not sure I follow which would be a better way.

romainguy: Default methods could work here.

romainguy: That said, if this is only called by the JNI code then I guess we can break the API here. It would otherwise be complicated to figure out which version to invoke.

alexey-pelykh: If I got that right, the sole purpose of MaterialProvider existence is to be used from JNI side.

libs/gltfio/src/AssetLoader.cpp
```diff
@@ -531,8 +537,8 @@ void FAssetLoader::createRenderable(const cgltf_node* node, Entity entity, const
     }
 }
 
-bool FAssetLoader::createPrimitive(const cgltf_primitive* inPrim, Primitive* outPrim,
-        const UvMap& uvmap, const char* name) {
+bool FAssetLoader::createPrimitive(const cgltf_data* srcAsset, const cgltf_primitive* inPrim,
```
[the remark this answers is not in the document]
alexey-pelykh: Good question. No specific reason, apart from keeping signatures of similar methods similar. And some groundwork for using extras in mentioned methods as well.

From a purist perspective, these changes are not meaningful within the scope of this exact PR alone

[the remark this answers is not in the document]
pixelflinger: Yes, please don't make changes that are not necessary for the scope of this PR -- this is because it'll make things more confusing in the future, when we go back to this change. Thanks!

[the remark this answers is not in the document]
alexey-pelykh: Updated

alexey-pelykh:
Change requests have been addressed

android/gltfio-android/src/main/java/com/google/android/filament/gltfio/MaterialProvider.java
prideout: Yes this is actually an "internal API" whose raison d'être is our JNI layer, so it's probably okay to break it as long we note it in the release notes.

I know it's been a loooong time  (my apologies) but I actually think we can merge this PR if it doesn't break anything.  I'll go ahead and do so.

prideout (approved):
I know it's been a loooong time (my apologies) but I actually think we can merge this PR, it's a useful feature and not dangerous.

alexey-pelykh:
@prideout do you need me to rebase it and/or do anything else?

prideout:
> @prideout do you need me to rebase it and/or do anything else?

Nope, we can merge this after we cut our next release (which will likely be Tuesday).

PR #4403, opened 2021-07-28
Implement light channels

pixelflinger:
We use fp16 for most values (e.g.: direction, color), which 
ultimately frees a vec4 (Currently unused) and a few other bits.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -7,6 +7,8 @@ A new header is inserted each time a *tag* is created.
 
 - engine: Option to automatically compute bent normals from SSAO & apply to specular AO
   [⚠️ **Material breakage**].
+- engine: New APIs: Light channels. Geometry and lights now have a channel associated to them, at 
+  least one channel must match for lighting to occur [⚠️ **Material breakage**].
 
 ## v1.11.2
 
--- a/libs/filamat/src/shaders/CodeGenerator.cpp
+++ b/libs/filamat/src/shaders/CodeGenerator.cpp
@@ -111,6 +111,30 @@ io::sstream& CodeGenerator::generateProlog(io::sstream& out, ShaderType type,
 
     out << SHADERS_COMMON_TYPES_FS_DATA;
 
+    if (mTargetApi == TargetApi::OPENGL && mShaderModel == ShaderModel::GL_CORE_41) {
+        // Tragically, OpenGL 4.1 doesn't support unpackHalf2x16 and
+        // MacOS doesn't support GL_ARB_shading_language_packing
+
+        // TODO: inject this from the backend based on the GL version and extension presence
+
+        out << R"(
+
+float u16tofp32(highp uint v) {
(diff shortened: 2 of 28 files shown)


libs/filamat/src/shaders/CodeGenerator.cpp
```diff
@@ -111,6 +111,30 @@ io::sstream& CodeGenerator::generateProlog(io::sstream& out, ShaderType type,
...
+float u16tofp32(highp uint v) {
+    // this doesn't handle denormals, NaNs or inf
+    v <<= 16u;
+    highp uint s = v & 0x80000000u;
+    highp uint n = v & 0x7FFFFFFFu;
+    highp uint nz = n == 0u ? 0u : 0xFFFFFFFF;
+    return uintBitsToFloat(s | ((((n >> 3u) + (0x70u << 23))) & nz));
+}
+
+vec2 unpackHalf2x16(highp uint v) {
```
romainguy, line 131: Do we expect naming conflicts depending on the GL version? Should we use a different name?

romainguy (approved):

libs/filamat/src/shaders/CodeGenerator.cpp
pixelflinger: Yeah, I'm afraid of that. That's why I need to try on a windows or Linux with at least 4.2.

romainguy (approved):

RELEASE_NOTES.md
```diff
@@ -7,6 +7,8 @@ A new header is inserted each time a *tag* is created.
 
 - engine: Option to automatically compute bent normals from SSAO & apply to specular AO
   [⚠️ **Material breakage**].
+- engine: New APIs: Light channels. Geometry and lights now have a channel associated to them, at 
+  least one channel must match for lighting to occur [⚠️ **Material breakage**].
```
bejado, line 11: Should we bump this release to v1.12.0?

shaders/src/getters.vs
```diff
@@ -69,14 +69,14 @@ vec4 getPosition() {
 
 #if defined(HAS_SKINNING_OR_MORPHING)
 
-    if (objectUniforms.morphingEnabled == 1) {
+    if ((objectUniforms.flags & 0x2u) != 0u) {
```
bejado, line 72: Make some defines for clarity?
```
#define SKINNING_ENABLED_BIT 0x1u
#define MORPHING_ENABLED_BIT 0x2u
#define CONTACT_SHADOWS_BIT 0x4u
```

RELEASE_NOTES.md
pixelflinger: I think so.

shaders/src/getters.vs
pixelflinger: Now that's a good idea!

RELEASE_NOTES.md
romainguy: Yes definitely.

PR #4411, opened 2021-07-30
Fade specular AO from bent normals at grazing angles

romainguy:
This helps prevent over-darkening but it's an artistic hack.
We need to check how we compute the bent normals as it seems we
find occlusion where there is no occlusion. The effect is
particularly visible on clear coat materials like a car's body.

### DIFF
--- a/shaders/src/ambient_occlusion.fs
+++ b/shaders/src/ambient_occlusion.fs
@@ -105,7 +105,7 @@ float sphericalCapsIntersection(float cosCap1, float cosCap2, float cosDistance)
 }
 
 // This function could (should?) be implemented as a 3D LUT instead, but we need to save samplers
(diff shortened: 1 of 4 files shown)


shaders/src/ambient_occlusion.fs
```diff
@@ -141,12 +140,13 @@ float computeSpecularAO(const highp vec2 uv, float NoV, float visibility, float
 #if defined(BLEND_MODE_OPAQUE) || defined(BLEND_MODE_MASKED)
 
 #if SPECULAR_AMBIENT_OCCLUSION == SPECULAR_AO_SIMPLE
+    // TODO: Should we even bother computing this when screen space bent normals are enabled?
```
pixelflinger, line 143: I was also wondering about that. But since it is a material parameter, this is the responsibility of the user, and since it doesn't do anything "wrong", I feel we should leave it.

pixelflinger (approved):

PR #4414, opened 2021-07-30
Fix a possible out of bound access on the GPU

### LINKED ISSUE #4388: a crash when viewing glTF with point/spot lights with filament-1.11.0 on iOS
screenshot

pixelflinger:
When calculating the froxel z slice, a depth buffer value of exactly
1 (i.e. far plane) would map to after the last slice, this value
must be clamped.

made the code more readable, hopefully.

Fixes #4388

### DIFF
--- a/shaders/src/light_punctual.fs
+++ b/shaders/src/light_punctual.fs
@@ -30,9 +30,19 @@ uvec3 getFroxelCoords(const highp vec3 fragCoords) {
     froxelCoord.xy = uvec2(fragCoords.xy * frameUniforms.resolution.xy *
             vec2(frameUniforms.oneOverFroxelDimension, frameUniforms.oneOverFroxelDimensionY));
 
-    froxelCoord.z = uint(max(0.0,
-            log2(frameUniforms.zParams.x * fragCoords.z + frameUniforms.zParams.y) *
-                    frameUniforms.zParams.z + frameUniforms.zParams.w));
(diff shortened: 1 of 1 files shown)


shaders/src/light_punctual.fs
```diff
@@ -30,9 +30,15 @@ uvec3 getFroxelCoords(const highp vec3 fragCoords) {
...
             vec2(frameUniforms.oneOverFroxelDimension, frameUniforms.oneOverFroxelDimensionY));
 
-    froxelCoord.z = uint(max(0.0,
-            log2(frameUniforms.zParams.x * fragCoords.z + frameUniforms.zParams.y) *
-                    frameUniforms.zParams.z + frameUniforms.zParams.w));
+    // go from screen-space (non-inverted) Z to normalized view-space Z (i.e. scaled by 1/zLightFar)
+    // (see Froxelizer.cpp)
+    highp float viewSpaceNormalizedZ = frameUniforms.zParams.x * fragCoords.z + frameUniforms.zParams.y;
+    // compute the sliceZ mapping
+    highp float sliceZ = log2(viewSpaceNormalizedZ) * frameUniforms.zParams.z + frameUniforms.zParams.w;
```
romainguy, line 37: Does the slice need to be highp? That seems unnecessary

romainguy (approved):

bejado (approved):

shaders/src/light_punctual.fs
pixelflinger: yes good point

PR #4424, opened 2021-07-30
Fix specular AO for the sheen and clear coat layers

romainguy:
We used a single specular AO term for the sheen and clear coat layers,
computed for the base layer. This was creating occlusion that was way
too strong when the roughness from the top layers was very different from
the roughness of the base layer. The effect was particularly obvious
on clear coat materials (car paint for instance).

With this change we can also remove the "grazing angle fade" artistic
hack that was previously introduced to combat this effect.

Ambient occlusion:
Screen Shot 2021-07-30 at 4 33 37 PM

Specular ambient occlusion from screen-space bent normals with this fix:
Screen Shot 2021-07-30 at 4 33 34 PM

### DIFF
--- a/shaders/src/light_indirect.fs
+++ b/shaders/src/light_indirect.fs
@@ -386,7 +386,8 @@ void evaluateClothIndirectDiffuseBRDF(const PixelParams pixel, inout float diffu
 #endif
 }
 
-void evaluateSheenIBL(const PixelParams pixel, float specularAO, inout vec3 Fd, inout vec3 Fr) {
+void evaluateSheenIBL(const PixelParams pixel, float diffuseAO,
+        const in SSAOInterpolationCache cache, inout vec3 Fd, inout vec3 Fr) {
 #if !defined(SHADING_MODEL_CLOTH) && !defined(SHADING_MODEL_SUBSURFACE)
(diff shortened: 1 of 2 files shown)


shaders/src/light_indirect.fs
```diff
@@ -592,6 +591,16 @@ void evaluateIBL(const MaterialInputs material, const PixelParams pixel, inout v
     vec3 E = vec3(0.0); // TODO: fix for importance sampling
     Fr = isEvaluateSpecularIBL(pixel, shading_normal, shading_view, shading_NoV);
 #endif
+
+    SSAOInterpolationCache interpolationCache;
+#if defined(BLEND_MODE_OPAQUE) || defined(BLEND_MODE_MASKED)
+    interpolationCache.uv = uvToRenderTargetUV(getNormalizedViewportCoord().xy);
+#endif
```
pixelflinger, line 598: That seems a bit dangerous -- could we remove the `uv` field from `SSAOInterpolationCache` when `defined(BLEND_MODE_OPAQUE) || defined(BLEND_MODE_MASKED)` is not true?

pixelflinger (changes requested):

pixelflinger (approved):

PR #4435, opened 2021-08-02
Temporarily switch back to REC709 for color grading

bejado:
@romainguy Feel free to revert after `inverseTonemap` has been updated.

### DIFF
--- a/filament/src/ColorGrading.cpp
+++ b/filament/src/ColorGrading.cpp
@@ -518,7 +518,7 @@ FColorGrading::FColorGrading(FEngine& engine, const Builder& builder) {
                     }
 
                     // Convert to color grading color space
-                    v = sRGB_to_REC2020 * v;
+                    // v = sRGB_to_REC2020 * v;
 
                     if (builder->hasAdjustments) {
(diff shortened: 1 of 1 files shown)


romainguy (commented):
Ah I didn't realize you would do this directly on our side. Note this will break ACES_LEGACY (our default) and ACES. In the function aces::ACES() of ToneMapper.cpp, we have this:

`float3 ap0 = REC2020_to_AP0 * color;`

And it would now need to be `REC709_to_AP0`. And at the end of the function `AP1_to_REC2020` should become `AP1_to_REC709`.

bejado:
@romainguy I think we should be consistent with the releases. Otherwise 1.11.2 will break `inverseTonemap` for 3P clients.

I don't see a definition for `REC709_to_AP0` or `AP1_to_REC709`, can you help me figure out what they should be?

romainguy:
I'll make the change, you can close this PR.

bejado:
Thanks!

romainguy:
Yeah I agree but I didn't really care that much since inverseToneMap() was always somewhat broken anyway :)

PR #4440, opened 2021-08-03
Use a clip space that matches our screen space for the z axis

pixelflinger:
We now set the projection matrix so that the clip space for the z
axis follows the inverted DX convention, i.e.: z_clip between 1 (near)
and 0 (far), which matches our screen space.

Note that the actual clip space in the GPU can revert to the "inverted
GL convention", i.e.: z_clip between -1 and 1, if the clip_control
extension is not present. In that case the precision benefits are 
mostly lost.

Also note that there shouldn't be any user facing changes. 
The projection matrix at the API level, still follows the GL convention.

Also, currently, the VERTEX_DOMAIN_DEVICE for materials also keeps
the GL convention (this is corrected in the vertex shader).

The gist of this change is to move the "reversed DX" mapping from the
vertex shader to the projection matrix.

### DIFF
--- a/shaders/src/getters.vs
+++ b/shaders/src/getters.vs
@@ -150,6 +150,6 @@ vec4 computeWorldPosition() {
     }
     return position * (1.0 / position.w);
 #else
-#error "Unknown Vertex Domain"
+#error Unknown Vertex Domain
 #endif
 }
--- a/shaders/src/getters.vs
+++ b/shaders/src/getters.vs
@@ -142,12 +142,14 @@ vec4 computeWorldPosition() {
     mat4 transform = getWorldFromViewMatrix();
     vec3 position = getPosition().xyz;
     return mulMat4x4Float3(transform, position);
-#else
+#elif defined(VERTEX_DOMAIN_DEVICE)
     mat4 transform = getWorldFromClipMatrix();
     vec4 position = transform * getPosition();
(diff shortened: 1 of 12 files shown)


romainguy (approved):

shaders/src/getters.vs
```diff
@@ -142,12 +142,14 @@ vec4 computeWorldPosition() {
...
-#else
+#elif defined(VERTEX_DOMAIN_DEVICE)
     mat4 transform = getWorldFromClipMatrix();
     vec4 position = transform * getPosition();
     if (abs(position.w) < MEDIUMP_FLT_MIN) {
         position.w = position.w < 0.0 ? -MEDIUMP_FLT_MIN : MEDIUMP_FLT_MIN;
     }
     return position * (1.0 / position.w);
+#else
+#error "Unknown Vertex Domain"
```
bejado, line 153:
```suggestion
#error Unknown Vertex Domain
```

IIRC Chrome can't handle quotes inside shader text. I think these are technically disallowed characters.

romainguy: Also is it actually needed? These defines are generated by filamat. It shouldn't ever get to the #else case.

pixelflinger: Well, unless we add a constant one day and have a bug or something.

PR #4442, opened 2021-08-03
views with custom render target should be blendable

### LINKED ISSUE #4430: Fail to Blend current rendering results with color buffer of framebuffer when post-processing enabled
This problem originates from #4404 . Basically, I have two views and one render target and I want to render the 2nd view on top of the 1st view by alpha blending:
```cpp
filament::Scene *first_scene = nullptr;
filament::Scene *second_scene = nullptr;
filament::View* first_view     = nullptr;
filament::View* second_view    = nullptr;
filament::Renderer *renderer = nullptr;
filament::RenderTarget *target = nullptr;

//this will cause the 1st view overlaid by the 2nd view, disabling post-processing will overcome it.
second_view->setBlendMode(View::BlendMode::TRANSLUCENT);
second_view->setPostProcessingEnabled(true);
second_view->setAntiAliasing(filament::View::AntiAliasing::FXAA);
//codes to initialize views, render and target
...

first_view->setScene(first_scene);
second_view->setScene(second_scene);
first_view->setRenderTarget(target);
second_view->setRenderTarget(target);
....
renderer->render(first_view);
renderer->render(second_view);
```
There are only one entity in both scenes, material of entity in the first scene is `bakedTexture` and that in the second scene is a trivial modification of `bakedColor.mat`:
```glsl
material {
    name : BakedColor,
    requires : [
        color
    ],
    blending: transparent,
    shadingModel : unlit,
    culling : none
}

fragment {
    void material(inout MaterialInputs material) {
        prepareMaterial(material);
        material.baseColor = getColor();
        material.baseColor.a = 0.2;
}
```
Result is as expected when disabling post-processing:
img

While contents of the 1st view was completely overlaid by that of the 2nd view if open post-processing:
img1

Post-processing is necessary to me due to anti-aliasing.

pixelflinger:
We made this wrong assumption that views with a custom rendertarget
are never blended, but it's not true -- e.g. two views can use the
same rendertarget.

Fixes #4430
May impact #2802

### DIFF
--- a/filament/src/Renderer.cpp
+++ b/filament/src/Renderer.cpp
@@ -352,9 +352,7 @@ void FRenderer::renderJob(ArenaScope& arena, FView& view) {
             }, viewRenderTarget);
 
     const bool blendModeTranslucent = view.getBlendMode() == View::BlendMode::TRANSLUCENT;
-    const bool hasCustomRenderTarget = viewRenderTarget != mRenderTarget;
-    // "blending" is meaningless when the view has a custom rendertarget because it's not composited
-    const bool blending = !hasCustomRenderTarget && blendModeTranslucent;
+    const bool blending = blendModeTranslucent;
     // If the swapchain is transparent or if we blend into it, we need to allocate our intermediate
     // buffers with an alpha channel.
     const bool needsAlphaChannel = (mSwapChain ? mSwapChain->isTransparent() : false) || blendModeTranslucent;


pixelflinger:
@roxlu this change may impact the resolution to #2802, you might want to try it and let us know if this causes a problem.

romainguy (approved):

PR #4443, opened 2021-08-03
Properly address the color grading LUT

romainguy:
With this change we properly sample from the texel centers when we should, thus avoiding a bias toward the low values. This was particularly problematic in the darks, resulting in darker results than they should have been.

Before:
Screen Shot 2021-08-03 at 2 32 47 PM

After:
Screen Shot 2021-08-03 at 2 32 31 PM

### DIFF
--- a/filament/src/materials/colorGrading/colorGrading.mat
+++ b/filament/src/materials/colorGrading/colorGrading.mat
@@ -12,6 +12,10 @@ material {
             precision: medium
         },
         {
+            type : float2,
+            name : lutSize
+        },
+        {
             type : sampler2d,
             name : bloomBuffer,
             precision: medium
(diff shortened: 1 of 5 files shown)


filament/src/materials/colorGrading/colorGrading.mat
```diff
@@ -126,6 +130,10 @@ vec3 colorGrade(mediump sampler3D lut, const vec3 x) {
     const float c = 0.244161 / log2(10.0);
     const float d = 0.386036;
     vec3 logc = c * log2(a * x + b) + d;
+
+    // Remap to sample pixel centers
+    logc = (0.5 + logc * materialParams.lutSize.x) * materialParams.lutSize.y;
```
pixelflinger, line 135: if you massaged `lutSize.xy` you could use just a madd:

```
logc  = (0.5 * materialParams.lutSize.y) + logc * (materialParams.lutSize.x * materialParams.lutSize.y)
```

pixelflinger (approved):
Could use just a madd in the shader with minor changes on the CPU side.

filament/src/materials/colorGrading/colorGrading.mat
romainguy: Great idea, done!

pixelflinger (approved):

PR #4449, opened 2021-08-05
double-precision accurate translation

pixelflinger:
Augment the TransformManager so it can maintain the 4th column as
double precision; we accomplish this by storing an extra float3 per matrix.
It's calculated as `doubleTranslation - float3{doubleTranslation}`.

We also add methods to set and get a transform as mat4, internally only
the 4th column is kept as double precision.

Finally when we calculate the worldTransform, we take this extra data into
account, but only for calculating the new 4th column, so the extra work
is small.

model/view matrices on Camera now store their translation as double
double precision. This allows filament to handle a very large world
space (thousands of kilometers).

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -5,8 +5,10 @@ A new header is inserted each time a *tag* is created.
 
 ## v1.12.1 (currently main branch)
 
-- engine: `double` precision model (and view) matrix on `Camera`. This is only an
-  API change, internal precision is still `float` [⚠️ **API Change**].
+- engine: `double` precision translation support in TransformManager. Disabled by default. 
+  Augment model (and view) matrix on `Camera` to accept double precision matrices. When enabled,
+  double precision translations allow filament to handle a very large world space.
+  [**New API**].
 
 ## v1.12.0
 
--- a/android/filament-android/src/main/java/com/google/android/filament/TransformManager.java
+++ b/android/filament-android/src/main/java/com/google/android/filament/TransformManager.java
@@ -84,6 +84,35 @@ import androidx.annotation.Size;
     }
 
     /**
+     * Enables or disable the accurate translation mode. Disabled by default.
+     *
+     * When accurate translation mode is active, the translation component of all transforms is
+     * maintained at double precision. This is only useful if the mat4 version of setTransform()
+     * is used, as well as getTransformAccurate().
+     *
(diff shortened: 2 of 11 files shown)


RELEASE_NOTES.md
```diff
@@ -5,8 +5,10 @@ A new header is inserted each time a *tag* is created.
 
 ## v1.12.1 (currently main branch)
 
-- engine: `double` precision model (and view) matrix on `Camera`. This is only an
-  API change, internal precision is still `float` [⚠️ **API Change**].
+- engine: `double` precision translation support in TransformManager. Disabled by default. 
+  Augment model (and view) matrix on `Camera` to accept double precision matrices. When enabled,
+  double precision translations allow filament to handle a very large world space.
+  [⚠️ **API Change**].
```
romainguy, line 11: Is this an API change or a new API? Afaict it doesn't break existing APIs?

android/filament-android/src/main/java/com/google/android/filament/TransformManager.java
```diff
@@ -83,6 +83,35 @@ public int getInstance(@Entity int entity) {
...
+     *
+     * @param enable true to enable the accurate translation mode, false to disable.
+     *
+     * @see #isAccurateTranslationsEnabled
+     * @see #create(int, int, double[])
+     * @see #setTransform(int, double[])
+     * @see #getTransform(int, double[])
+     * @see #getWorldTransform(int, double[])
+     */
+    public void enableAccurateTranslations(boolean enable) {
```
romainguy, line 101: It should be setAccurateTranslationsEnabled() to match the getter is*().

filament/include/filament/TransformManager.h
```diff
@@ -112,6 +112,30 @@ class UTILS_PUBLIC TransformManager : public FilamentAPI {
...
+     *
+     * @param enable true to enable the accurate translation mode, false to disable.
+     *
+     * @see isAccurateTranslationsEnabled
+     * @see create(utils::Entity, Instance, const math::mat4&);
+     * @see setTransform(Instance, const math::mat4&)
+     * @see getTransformAccurate
+     * @see getWorldTransformAccurate
+     */
+    void enableAccurateTranslations(bool enable) noexcept;
```
romainguy, line 130: See above

romainguy (approved):

PR #4457, opened 2021-08-05
Switch white balance to CAT16 and add gamut mapping code

romainguy:
CAT16 is the transform from CIECAM16 which improves on CIECAM02/CAT02, with more accurate perceptual results.

The gamut mapping code is based on the Oklab color space and projects out-of-gamut colors onto the gamut volume. We use adaptive L0=0.5 and a default alpha of 0.05 but we should probably explore other projections (with Lcusp, etc.). This code is currently not used but will/may serve as the foundation for gamut mapping once we start using wider color gamuts during the color grading pass.

### DIFF
--- a/filament/CMakeLists.txt
+++ b/filament/CMakeLists.txt
@@ -51,6 +51,7 @@ set(SRCS
         src/Camera.cpp
         src/Color.cpp
         src/ColorGrading.cpp
+        src/ColorSpace.cpp
         src/Culler.cpp
         src/DFG.cpp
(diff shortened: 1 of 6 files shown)


bejado (approved):
Will `inverseTonemap` still work correctly?

pixelflinger (approved):
Looks like there is 3 independent changes in this one PR.

romainguy:
> Will `inverseTonemap` still work correctly?

It doesn't affect tone mapping.

PR #4476, opened 2021-08-10
improve ortho projection support, camera code and comments

pixelflinger:


libs/math/include/math/mat4.h
```diff
@@ -557,6 +557,26 @@ constexpr typename TMat44<T>::col_type MATH_PURE operator*(const TMat44<T>& lhs,
...
+    double4 result{};
+    result += lhs[0] * rhs[0];
+    result += lhs[1] * rhs[1];
+    result += lhs[2] * rhs[2];
+    result += lhs[3] * rhs[3];
+    return float4{ result };
+}
+
+// mat4 * mat4, with double precision intermediates
+constexpr mat4f multiply(mat4f const& lhs, mat4f const& rhs) noexcept {
```
bejado, line 571: It may be worth renaming to something other than simply `multiply`, to clearly distinguish it from the `*` operator.

bejado (approved):

libs/math/include/math/mat4.h
pixelflinger: Yes, I couldn't find something satisfying. It's like... `highPrecisionMultiply`. Any idea?

bejado: Yeah, `highPrecisionMultiply` sounds fine to me.

PR #4487, opened 2021-08-11
Fix CSM glitch with ground plane

bejado:
`shading_unlit.fs` was missing a check :)

### DIFF
--- a/shaders/src/shading_unlit.fs
+++ b/shaders/src/shading_unlit.fs
@@ -42,8 +42,10 @@ vec4 evaluateMaterial(const MaterialInputs material) {
 #if defined(HAS_DIRECTIONAL_LIGHTING)
 #if defined(HAS_SHADOWING)
     float visibility = 1.0;
-    if ((frameUniforms.directionalShadows & 1u) != 0u) {
-        uint cascade = getShadowCascade();
+    uint cascade = getShadowCascade();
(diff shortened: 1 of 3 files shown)


shaders/src/shading_unlit.fs
```diff
@@ -42,8 +42,10 @@ vec4 evaluateMaterial(const MaterialInputs material) {
 #if defined(HAS_DIRECTIONAL_LIGHTING)
 #if defined(HAS_SHADOWING)
     float visibility = 1.0;
-    if ((frameUniforms.directionalShadows & 1u) != 0u) {
-        uint cascade = getShadowCascade();
+    uint cascade = getShadowCascade();
+    bool cascadeHasVisibleShadows = bool(frameUniforms.cascades & (1u << cascade << 8u));
```
pixelflinger, line 46: can we just add some parens here (around the << operators)?

pixelflinger (changes requested):

shaders/src/shading_unlit.fs
pixelflinger:
```suggestion
    bool cascadeHasVisibleShadows = bool(frameUniforms.cascades & ((1u << cascade) << 8u));
```

pixelflinger (approved):

PR #4496, opened 2021-08-13
move private headers around + cleanups

pixelflinger:
src/details now only has the private headers of public types/classes.
Private/implementation classes have their header next to them under src.


bejado:
LGTM at first glance. Are the two commits independent? If so, let's _not_ do a squash commit when merging. This way they can be reverted separately if need be.

pixelflinger:
Yes, I wasn't going to do a squash. I think it will be easy to revert the header moves. However, pretty much any other PR after that one will probably conflict.

PR #4509, opened 2021-08-17
Metal: skip loop when unbinding uniform buffer objects

bejado:


romainguy (approved):

filament/backend/src/metal/MetalHandles.h
```diff
@@ -119,8 +119,14 @@ class MetalBufferObject : public HwBufferObject {
...
     MetalBuffer* getBuffer() { return &buffer; }
 
+    // Flags that this BufferObject is used as a uniform buffer.
+    // If this isn't set, we can skip a check when destroying this BufferObject.
+    void flagUniformUsage() { isUniform = true; }
+    bool isUsedAsUniform() const { return isUniform; }
+
 private:
     MetalBuffer buffer;
+    bool isUniform;
```
pixelflinger, line 129: instead of a bool, you could use a `bitset32` and set the `index` bit when you bind (and unset it when you unbind) -- this way you could avoid to loop on the whole list, since each bit set would tel you which index has this ubo bound to.
We even have a `forEachBitSet` method you could use instead of the loop.

pixelflinger (commented):
lgtm, but maybe consider the proposal in comments?

filament/backend/src/metal/MetalHandles.h
bejado: Good suggestion, I like that

pixelflinger (approved):

PR #4521, opened 2021-08-18
enable GL_ARB_shading_language_packing on GL 4.1

pixelflinger:


libs/filamat/src/shaders/CodeGenerator.cpp
```diff
@@ -64,6 +64,7 @@ io::sstream& CodeGenerator::generateProlog(io::sstream& out, ShaderType type,
                 out << "#version 450 core\n\n";
             } else {
                 out << "#version 410 core\n\n";
+                out << "#extension GL_ARB_shading_language_packing : enable\n\n";
```
romainguy, line 67: Are we guaranteed to have this extension?

prideout:
Confirmed that this fixes the issue on my machine, although maybe the #extension line needs to be conditional.

romainguy:
> Confirmed that this fixes the issue on my machine, although maybe the #extension line needs to be conditional.

The problem is that this code is in `filamat`, so it's not necessarily executed at runtime. We could inject that line at runtime when needed though.

libs/filamat/src/shaders/CodeGenerator.cpp
pixelflinger: No but it doesn't mater right? When you specify `enable` vs. `require`. I'm checking at runtime and injecting the appropriate code if it's not there. The spec says that if the extension is not supported, it's just a warning.

pixelflinger: > Confirmed that this fixes the issue on my machine, although maybe the #extension line needs to be conditional.

@prideout did you check with a debug or release build?

prideout: The bug repro and the verification of the fix were both using a release build of Filament.  (I'm actually unable to build a debug build of Filament on my Windows machine because matc crashes during the build, I haven't looked into that yet.)

pixelflinger: Nice, I think if it works with a release build, we're good.

bejado (approved):

prideout:
Uh oh, my verification was wrong (I'm distracted nowadays) because I just now managed to reproduce the issue again in release config, with the Mathias fix!

So, I think the `#extension` line that Mathias added getting removed by the optimization phase...  Are we going to need another hack like what we do towards the end of `CodeGenerator::fixupExternalSamplers`?  😢

pixelflinger:
okay it needs more work.

Issue #4473, opened 2021-08-09
Allow more shadow-casting lights


jwwalker:
Please permit far more shadow-casting lights.  Currently there is a limit of `CONFIG_MAX_SHADOW_CASTING_SPOTS = 6`.  (And by the way, more than the current 8 light channels would also be good.)

bejado:
This is because we use a `uint8_t` for visibility culling: https://github.com/google/filament/blob/bbee5599f13e49267936aa2c5dc979026af07b8a/filament/src/details/View.h#L100-L101

In the past @pixelflinger and I have discussed changing this to a `uint32_t`, which would allow for a max of 30 lights.

jwwalker:
@bejado Ideally I'd like even more than 30, especially if I have to build my own shadow-casting point lights using 4 spot lights each.

pixelflinger:
Performance will be abysmal though. And handling so many shadowmaps is not a priority for us (for mobile).

PR #4522, opened 2021-08-19
Increase the number of spot shadows to 14 (from 6)

pixelflinger:
Adresses #4473

### DIFF
--- a/filament/src/Culler.cpp
+++ b/filament/src/Culler.cpp
@@ -22,11 +22,9 @@
 
 using namespace filament::math;
 
-// We use a vectorize width of 8 because, on ARMv8 it allows the compiler to write eight
-// 8-bits results in one go. Without this it has to do 4 separate byte writes, which
-// ends-up being slower.
-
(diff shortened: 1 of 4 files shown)


romainguy (approved):

filament/src/Culler.cpp
```diff
@@ -22,11 +22,9 @@
 
 using namespace filament::math;
 
-// We use a vectorize width of 8 because, on ARMv8 it allows the compiler to write eight
-// 8-bits results in one go. Without this it has to do 4 separate byte writes, which
-// ends-up being slower.
-
-#define FILAMENT_CULLER_VECTORIZE_HINT 8
+// use 8 if Culler::result_type is 8-bits, on ARMv8 it allows the compiler to write eight
```
bejado, line 25: I'm confused here.. isn't `Culler::result_type` now 16 bits?

pixelflinger: right it's 16 bits now, so we use a value of 4 -- if it was 8 bits we'd need to use 8 (that's what the comment is trying to say).

PR #4524, opened 2021-08-19
Use wide gamut color grading and gamut mapping

romainguy:
This change introduces a new API on ColorGrading to enable or disable gamut mapping at will. This feature is currently off
by default and while useful and recommended, it isn't cheap and we need to figure out what we want our defaults to be specially on mobile targets.

Gamut mapping can be combined with luminance scaling to properly control hue skews.

Importantly, this change introduces a compatibility mode for our legacy tone mapping API `ToneMapping::FILMIC`. When the tone mapping operator is selected via this API, backward compatibility is enabled and color grading is performed in sRGB instead of the wider BT.2020 color space. This is important to retain compatibility with the inverse tone mapping operation available to shaders (this could be worked around by performing a conversion to/from Rec.2020 in the shader during inverse tone mapping but is it worth the cost?).

Here is an example of HDR content tone mapped normally:
Screen Shot 2021-08-19 at 4 42 55 PM

And the same content with the gamut mapping step (observe the red shirt):
Screen Shot 2021-08-19 at 4 42 57 PM

Gamut mapping can be combined with luminance scaling. Here is luminance scaling alone:
Screen Shot 2021-08-19 at 4 43 00 PM

And combine with gamut mapping (observe the red shirt again):
Screen Shot 2021-08-19 at 4 43 02 PM


filament/src/ColorGrading.cpp
```diff
@@ -511,26 +552,28 @@ FColorGrading::FColorGrading(FEngine& engine, const Builder& builder) {
                     // LogC encoding
                     v = LogC_to_linear(v);
 
-                    // TODO: Performed in sRGB, should be in Rec.2020 or AP1
+                    // In this section we perform a few adjustments in sRGB
+                    if (builder->hasAdjustments) {
+                    }
```
pixelflinger, line 557: What adjustments? :-)

pixelflinger (commented):
Add something in Release Notes?

PR #4529, opened 2021-08-20
Add FFX FSR as a dynamic resolution setting

pixelflinger:


filament/backend/src/opengl/OpenGLProgram.cpp
```diff
@@ -92,7 +92,28 @@ float u16tofp32(highp uint v) {
...
+    } else if (exponent > -15) {
+        exponent += 15;
+        mantissa >>= 13;
+        f16 = sign | uint(exponent << 10) | mantissa;
+    } else {
+        f16 = sign;
+    }
+    return f16;
+}
+highp uint packHalf2x16(vec2 v) {
```
romainguy, line 112: Can we use a built-in when it's available?

filament/src/PostProcessManager.cpp
```diff
@@ -226,6 +227,8 @@ static const MaterialInfo sMaterialList[] = {
         { "separableGaussianBlur",      MATERIAL(SEPARABLEGAUSSIANBLUR) },
         { "taa",                        MATERIAL(TAA) },
         { "vsmMipmap",                  MATERIAL(VSMMIPMAP) },
+        { "fsr_easu",                  MATERIAL(FSR_EASU) },
```
romainguy, line 230: Nit: alignment on `MATERIAL` (same the line below)

romainguy (approved):

filament/backend/src/opengl/OpenGLProgram.cpp
pixelflinger: Well, it should inject that code only on GL 4.1 when the extension is not available.

romainguy: Ah right, sorry the diff on GitHub wasn't showing enough context.

PR #4556, opened 2021-08-27
change backend  triangle test attributes build

mirsfang:
backend triangle test build AttributeArray is intricate，i feel this change looks better


mirsfang:
yes ， hand shook

PR #4559, opened 2021-08-28
Add night adaptation control to ColorGrading

romainguy:
In low-light conditions, peak luminance sensitivity of the eye shifts toward the blue end of the visible spectrum. This effect called the Purkinje effect occurs during the transition from photopic (cone-based) vision to scotopic (rod-based) vision. Because the rods and cones use the same neural pathways, a color shift is introduced as the rods take over to
improve low-light perception.

This function aims to (somewhat) replicate this color shift and peak luminance sensitivity increase to more faithfully reproduce scenes in low-light conditions as they would be perceived by a human observer (as opposed to an artificial observer such as a camera sensor).

The night adaptation can be controlled using a 0..1 factor for artistic reasons using the new `ColorGrading::Builder::nightAdaptation(float)` API.

Color spectrum without and with night adaptation:
Screen Shot 2021-08-28 at 1 23 05 PM
Screen Shot 2021-08-28 at 1 23 02 PM

HDR spectrum (0EV, +1EV, +2EV, +4EV) without and with night adaptation, with luminance scaling and gamut mapping always on:
Screen Shot 2021-08-28 at 1 24 33 PM
Screen Shot 2021-08-28 at 1 24 36 PM

Example of fully adapted scene:
Screen Shot 2021-08-21 at 12 14 35 AM

Scene without any adaptation:
Screen Shot 2021-08-21 at 12 28 26 AM

And with full adaptation:
Screen Shot 2021-08-21 at 12 28 44 AM


tools/rgb-to-lmsr/src/main.cpp
```diff
@@ -0,0 +1,1365 @@
+/*
+ * Copyright (C) 2018 The Android Open Source Project
```
bejado, line 2: 2021 ?

bejado (approved):

tools/rgb-to-lmsr/src/main.cpp
romainguy: Good catch, copy paste from other tools ^^

PR #4581, opened 2021-09-02
Fix color format for the external image

kirinzer:


pixelflinger:
> Thank you but could you explain the purpose of this change? What does it fix?

Yes, and would you mind creating an issue as well, so we can associate this fix with an issue for future reference. Thank you.

kirinzer:
> Thank you but could you explain the purpose of this change? What does it fix?

Sorry for no description provided. I thought it is clearly. The question is `createTextureFromImage`, this func only pass a `glFormat` param, but for iOS platform your `pixelBuffer` format usually is `kCVPixelFormatType_32BGRA`, it's not same order to GL_RGBA, so that I modify the `createTextureFromImage` func, and insert a new param `pixelFormat` to distinguish the different case.

romainguy:
Is the format of an external image always BGRA? Shouldn't this query the image's format?

kirinzer:
> > Thank you but could you explain the purpose of this change? What does it fix?
> 
> Yes, and would you mind creating an issue as well, so we can associate this fix with an issue for future reference. Thank you.

Absolutely, I will 

> Is the format of an external image always BGRA? Shouldn't this query the image's format?

Good question! Actually the best method is query format that no doubt. But in this case , the `ASSERT_POSTCONDITION` had filtered the others.
```
    ASSERT_POSTCONDITION(formatType == kCVPixelFormatType_32BGRA ||
                         formatType == kCVPixelFormatType_420YpCbCr8BiPlanarFullRange,
            "iOS external images must be in either 32BGRA or 420f format.");

```

kirinzer:
We only need  to focus two format types (32BGRA or 420f)

kirinzer:
> > Thank you but could you explain the purpose of this change? What does it fix?
> 
> Yes, and would you mind creating an issue as well, so we can associate this fix with an issue for future reference. Thank you.

I thought we need to confirm the solution first and make sure that is the right way to fix the issue. At last, I will create the issue if it is necessary.

filament/backend/src/opengl/CocoaTouchExternalImage.mm
```diff
@@ -228,6 +229,21 @@ void main() {
...
 }
 
+CVOpenGLESTextureRef CocoaTouchExternalImage::createTextureFromImage(CVPixelBufferRef image, GLuint
+        glFormat, GLuint pixelFormat, size_t plane) noexcept {
+    const size_t width = CVPixelBufferGetWidthOfPlane(image, plane);
+    const size_t height = CVPixelBufferGetHeightOfPlane(image, plane);
+
+    CVOpenGLESTextureRef texture = nullptr;
+    UTILS_UNUSED_IN_RELEASE CVReturn success =
+            CVOpenGLESTextureCacheCreateTextureFromImage(kCFAllocatorDefault,
```
bejado, line 239: According to [Apple's documentation](https://developer.apple.com/documentation/corevideo/1619860-cvopenglestexturecachecreatetext?language=objc), even for BGRA buffers `GL_RGBA` should be passed:

```
//Mapping a BGRA buffer as a source texture:
CVOpenGLESTextureCacheCreateTextureFromImage(kCFAllocatorDefault, textureCache, pixelBuffer, NULL, GL_TEXTURE_2D, GL_RGBA, width, height, GL_RGBA, GL_UNSIGNED_BYTE, 0, &outTexture);
```

Does changing it to `GL_BGRA` have an effect?

pixelflinger: Interesting, I wonder if there is a typo in their sample code there?

kirinzer: > According to [Apple's documentation](https://developer.apple.com/documentation/corevideo/1619860-cvopenglestexturecachecreatetext?language=objc), even for BGRA buffers `GL_RGBA` should be passed:
> 
> ```
> //Mapping a BGRA buffer as a source texture:
> CVOpenGLESTextureCacheCreateTextureFromImage(kCFAllocatorDefault, textureCache, pixelBuffer, NULL, GL_TEXTURE_2D, GL_RGBA, width, height, GL_RGBA, GL_UNSIGNED_BYTE, 0, &outTexture);
> ```
> 
> Does changing it to `GL_BGRA` have an effect?

In my case, I decoded a video by AVAssetReader, and I got a CVPixelBuffer object. if I does not change the type, I mean pass the same type GL_RGBA, the code could work, but the texture color looks like more blue than normal. After that I change the type like my commit, everything works fine.

kirinzer: > Interesting, I wonder if there is a typo in their sample code there?

Apple's documentation just provides "Some example mappings can be seen in the following code snippet."
So there are not all cases. It seems to me that the image type passed should be true order with pixelBuffer, and this func will convert it to GL_RGBA for create texture.

kirinzer:
@pixelflinger How are things going?

filament/backend/src/opengl/CocoaTouchExternalImage.mm
```diff
@@ -228,6 +229,21 @@ void main() {
     return texture;
 }
 
+CVOpenGLESTextureRef CocoaTouchExternalImage::createTextureFromImage(CVPixelBufferRef image, GLuint
```
bejado, line 232: This doesn't need to be a new method. Add the `GLuint pixelFormat` argument to the original `createTextureFromImage` and pass `GL_RGBA` for the other calls.

filament/backend/src/opengl/CocoaTouchExternalImage.mm
bejado: I think I'm fine with making the change, though I need to test a few things first to make sure it won't break any of our use-cases.

bejado (changes requested):

bejado:
Closed in favor of #4600

PR #4600, opened 2021-09-08
Fix, specify BGRA channel ordering for iOS OpenGL single-plane external images

bejado:


filament/backend/src/opengl/CocoaTouchExternalImage.h
```diff
@@ -63,7 +63,7 @@ class CocoaTouchExternalImage final {
 
     void release() noexcept;
     CVOpenGLESTextureRef createTextureFromImage(CVPixelBufferRef image, GLuint glFormat,
-            size_t plane) noexcept;
+            GLenum format, size_t plane) noexcept;
```
romainguy, line 66: Should/can we query the format from the `CVPixelBufferRef`?

romainguy (approved):

filament/backend/src/opengl/CocoaTouchExternalImage.h
bejado: We do query the format [here](https://github.com/google/filament/pull/4600/files#diff-4f5b8d8c432c66967e5696e4856f2fb7501c6e594b2620581a980c117c0a9c25R137). Since it's a single planed image, it must be BGRA.

PR #4607, opened 2021-09-11
Use more sensible defaults for spot light inner outer cone angles

bejado:
The max is `f::PI_2`, so we should never have been setting this to more than that. >= `f::PI_2` causes shadows to disappear. `f::PI_4` is a much better default.


romainguy (approved):
Thanks! Should we give a slightly smaller inner angle too to get a soft falloff?

pixelflinger (approved):

bejado:
> Thanks! Should we give a slightly smaller inner angle too to get a soft falloff?

Yeah, I like that.

romainguy (approved):

PR #4613, opened 2021-09-13
fix a race in jobsystem

pixelflinger:
We were decrementing activeJobCount after removing the job from the
queue, which could cause other threads in the pool to preempt us before
the decrement, causing them to spin forever trying to get a non-existant
job, until the decrement actually happened.

Now we always decrement first and fix-up the count if we couldn't get
a job from the queues. The race is inverted, and doesn't cause threads
to spin a long time.

### DIFF
--- a/libs/utils/include/utils/JobSystem.h
+++ b/libs/utils/include/utils/JobSystem.h
@@ -354,19 +354,56 @@ private:
         assert(job);
         size_t index = job - mJobStorageBase;
         assert(index >= 0 && index < MAX_JOB_COUNT);
+
+        // put the job into the queue first
         workQueue.push(uint16_t(index + 1));
+        // then increase our active job count
+        uint32_t oldActiveJobs = mActiveJobs.fetch_add(1, std::memory_order_relaxed);
+        // but it's possible that the job has already been picked-up, so oldActiveJobs could be
+        // negative for instance. We signal only if that's not the case.
+        if (oldActiveJobs >= 0) {
+            wakeOne(); // wake-up a thread if needed...
+        }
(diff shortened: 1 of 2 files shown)


libs/utils/include/utils/JobSystem.h
```diff
@@ -354,19 +354,56 @@ class JobSystem {
...
+        // negative for instance. We signal only if that's not the case.
+        if (oldActiveJobs >= 0) {
+            wakeOne(); // wake-up a thread if needed...
+        }
     }
 
     Job* pop(WorkQueue& workQueue) noexcept {
+        // decrement mActiveJobs first, this is to ensure that if there is only a single job left
+        // (and we're about to pick it up), other threads don't loop trying to do the same.
+        mActiveJobs.fetch_sub(1, std::memory_order_relaxed);
```
romainguy, line 372: Can you end up with mActiveJobs < 0 here but workQueue.pop() returning a valid job?

romainguy (approved):

libs/utils/include/utils/JobSystem.h
pixelflinger: Yes, you could but it wouldn't be a problem. This would mean than another thread is for e.g. doing a steal() at the same time. So, if we do get a valid job from the queue, this means that the other thread didn't, and it will "fix" the counter by doing a fetch_add.
On the other hand, if we don't get a job, then we are fixing the counter.

The key is to do the correction based on whether we actually got a job or not, NOT based on the value of the counter.

The other key is to signal if needed when fixing the counter.

PR #4618, opened 2021-09-15
Implement lod bias for user materials

pixelflinger:
A lod bias can be applied to textures when dynamic resolution is
enabled. A proper lod bias is calculated and applied for quality levels
HIGH and ULTRA.

Lod Bias is implemented in glslang, which mean it's ignored when
glslang is not used, which currently happens with OpenGL backends with
optimization turned off (which includes filament LITE). This may be
fixed in the future.

### DIFF
--- a/filament/src/PerViewUniforms.cpp
+++ b/filament/src/PerViewUniforms.cpp
@@ -87,6 +87,16 @@ void PerViewUniforms::prepareCamera(const CameraInfo& camera) noexcept {
     s.clipControl = mClipControl;
 }
 
+void PerViewUniforms::prepareUpscaler(math::float2 scale,
+        DynamicResolutionOptions const& options) noexcept {
+    auto& s = mPerViewUb.edit();
+    if (options.quality >= QualityLevel::HIGH) {
+        s.lodBias = std::log2(std::min(scale.x, scale.y));
+    } else {
+        s.lodBias = 0.0f;
+    }
+}
+
 void PerViewUniforms::prepareExposure(float ev100) noexcept {
     const float exposure = Exposure::exposure(ev100);
     auto& s = mPerViewUb.edit();
--- a/libs/filamat/src/sca/ASTHelpers.cpp
+++ b/libs/filamat/src/sca/ASTHelpers.cpp
@@ -359,4 +359,91 @@ void getFunctionParameters(TIntermAggregate* func, std::vector<FunctionParameter
     }
 }
 
+template <typename F>
+class TraverserAdapter: public TIntermTraverser {
+    F closure;
+public:
+    explicit TraverserAdapter(F closure)
+        : TIntermTraverser(true, false, false, false),
+          closure(closure) {
+    }
+    bool visitAggregate(TVisit visit, TIntermAggregate* node) override {
+        return closure(visit, node);
+    }
+};
+
+void textureLodBias(TIntermediate* intermediate, TIntermNode* root,
+        const char* lodBiasSymbolName) {
+    // we need to run this only from the user's main entry point
+
+    // First, find the "lodBias" symbol
+    TIntermSymbol* pIntermSymbolLodBias = nullptr;
+    TraverserAdapter findLodBiasSymbol(
+            [&, done = false](TVisit visit, TIntermAggregate* node) mutable {
+        if (node->getOp() == glslang::EOpSequence) {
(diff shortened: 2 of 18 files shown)


filament/src/PerViewUniforms.cpp
```diff
@@ -87,6 +87,16 @@ void PerViewUniforms::prepareCamera(const CameraInfo& camera) noexcept {
     s.clipControl = mClipControl;
 }
 
+void PerViewUniforms::prepareUpscaler(math::float2 scale,
+        DynamicResolutionOptions const& options) noexcept {
+    auto& s = mPerViewUb.edit();
+    if (options.quality >= QualityLevel::HIGH) {
+        s.lodBias = std::log2(std::min(scale.x, scale.y));
```
romainguy, line 94: Should we do it any time FSR is active?

libs/filamat/src/sca/ASTHelpers.cpp
```diff
@@ -359,4 +359,91 @@ void getFunctionParameters(TIntermAggregate* func, std::vector<FunctionParameter
...
+                }
+            }
+        }
+        return false;
+    });
+    root->traverse(&findLodBiasSymbol);
+
+    if (!pIntermSymbolLodBias) {
+        // something went wrong
+        utils::slog.e << "lod bias ignored becase \"" << lodBiasSymbolName << "\" was not found!" << utils::io::endl;
```
romainguy, line 404: Typo in "because"

romainguy (approved):

filament/src/PerViewUniforms.cpp
pixelflinger: The FSR doc recommends to do it when FSR is active. Here it's done only at HIGH and  ULTRA quality levels. The idea is that lod bias will impact bandwidth. The quality improvement is very subtle as well.
Also if the scale factor is 1.0 the log2() ends up being 0.0.

romainguy: Right I wasn't worried about scale 1, I was mostly wondering if we should enable it at medium as well. But good point about bandwidth.

romainguy (approved):

romainguy (approved):

libs/filamat/src/sca/GLSLTools.cpp
```diff
@@ -359,4 +359,12 @@ void GLSLTools::prepareShaderParser(MaterialBuilder::TargetApi targetApi, glslan
     }
 }
 
+void GLSLTools::textureLodBias(TShader& shader) {
+    TIntermediate* intermediate = shader.getIntermediate();
+    TIntermNode* root = intermediate->getTreeRoot();
+    ASTUtils::textureLodBias(intermediate, root,
+            "material(struct-MaterialInputs",
```
bejado, line 366: Why can't we put the full signature here? Also I'm confused by the `struct-` but I'm guessing that's a glslang thing

libs/filabridge/include/filament/MaterialEnums.h
```diff
@@ -27,7 +27,7 @@
 namespace filament {
 
 // update this when a new version of filament wouldn't work with older materials
-static constexpr size_t MATERIAL_VERSION = 12;
+static constexpr size_t MATERIAL_VERSION = 13;
```
bejado, line 30: Do we need to bump the material version? Older materials still work, albeit without lod bias, no?

libs/filamat/src/sca/GLSLTools.cpp
romainguy: The signature changes when we change MaterialInputs :(( It's apparently something like struct-MaterialInputs-v4f-v1f-etc. Apparently the struct flattened??

pixelflinger: It's because the signature seems to change based on what is inside `MaterialInputs` and yes the `struct` is part of the signature, apparently.

libs/filabridge/include/filament/MaterialEnums.h
pixelflinger: Yes I think it's true that they will work. I added a field to `FrameUniforms` but in this case it shouldn't be a problem.

So... good good point. Maybe we don't need to upgrade the material version.

bejado (approved):

PR #4630, opened 2021-09-16
break dependency on libutils

pixelflinger:
it was only used of Path which now can be replaced bu
std::filesystem::path.


romainguy (approved):

pixelflinger:
argh windows still doesn't support this.

PR #4657, opened 2021-09-21
Implement picking functionality

pixelflinger:
This is a pixel accurate implementation of picking. Picking queries
can be created on view, and upon completion a user provided callback
is called with the Entity of the renderable at the queried coordinates
in the viewport.

Picking queries typically have 1 or 2 frame of latency and may impact
performance on some drivers.

It is mostly intended for use by editors, or when latency is not a major
concern. This api should not be used for dragging/moving objects, it is
intended for initial picking only.

Picking is implemented in the structure pass.

### DIFF
--- a/android/filament-android/src/main/cpp/View.cpp
+++ b/android/filament-android/src/main/cpp/View.cpp
@@ -20,6 +20,8 @@
 #include <filament/View.h>
 #include <filament/Viewport.h>
 
+#include "common/CallbackUtils.h"
+
 using namespace filament;
 
 extern "C" JNIEXPORT void JNICALL
@@ -381,4 +383,42 @@ Java_com_google_android_filament_View_nIsScreenSpaceRefractionEnabled(JNIEnv *, 
         jlong nativeView) {
     View* view = (View*) nativeView;
     return (jboolean)view->isScreenSpaceRefractionEnabled();
+}
+
+extern "C"
+JNIEXPORT void JNICALL
+Java_com_google_android_filament_View_nPick(JNIEnv* env, jclass clazz,
+        jlong nativeView,
+        jint x, jint y, jobject handler, jobject internalCallback) {
+
+    // jniState will be initialized the first time this method is called
+    static const struct JniState {
+        jclass internalOnPickCallbackClass;
+        jfieldID renderableFieldId;
+        jfieldID depthFieldId;
+        jfieldID fragCoordXFieldId;
+        jfieldID fragCoordYFieldId;
+        jfieldID fragCoordZFieldId;
+        explicit JniState(JNIEnv* env) noexcept {
+            internalOnPickCallbackClass = env->FindClass("com/google/android/filament/View$InternalOnPickCallback");
+            renderableFieldId = env->GetFieldID(internalOnPickCallbackClass, "mRenderable", "I");
+            depthFieldId = env->GetFieldID(internalOnPickCallbackClass, "mDepth", "F");
+            fragCoordXFieldId = env->GetFieldID(internalOnPickCallbackClass, "mFragCoordsX", "F");
+            fragCoordYFieldId = env->GetFieldID(internalOnPickCallbackClass, "mFragCoordsY", "F");
+            fragCoordZFieldId = env->GetFieldID(internalOnPickCallbackClass, "mFragCoordsZ", "F");
+        }
+    } jniState(env);
+
+    View* view = (View*) nativeView;
+    JniCallback *callback = JniCallback::make(env, handler, internalCallback);
+    view->pick(x, y, [callback](View::PickingQueryResult const& result) {
+        jobject obj = callback->getCallbackObject();
+        JNIEnv* const env = callback->getJniEnv();
+        env->SetIntField(obj, jniState.renderableFieldId, (jint)result.renderable.getId());
+        env->SetFloatField(obj, jniState.depthFieldId, result.depth);
+        env->SetFloatField(obj, jniState.fragCoordXFieldId, result.fragCoords.x);
+        env->SetFloatField(obj, jniState.fragCoordYFieldId, result.fragCoords.y);
+        env->SetFloatField(obj, jniState.fragCoordZFieldId, result.fragCoords.z);
+        JniCallback::invoke(callback);  // this destroys JniCallback
+    });
 }
--- a/android/filament-android/src/main/java/com/google/android/filament/View.java
+++ b/android/filament-android/src/main/java/com/google/android/filament/View.java
@@ -1548,6 +1548,71 @@ public class View {
         return mDepthOfFieldOptions;
     }
 
+    /**
+     * A class containing the result of a picking query
+     */
+    public static class PickingQueryResult {
+        /** The entity of the renderable at the picking query location */
+        @Entity public int renderable;
+        /** The value of the depth buffer at the picking query location */
+        public float depth;
+        /** The fragment coordinate in GL convention at the the picking query location */
+        @NonNull public float[] fragCoords = new float[3];
+    };
+
+    /**
+     * An interface to implement a custom class to receive results of picking queries.
+     */
+    public interface OnPickCallback {
+        /**
+         * onPick() is called by the specified Handler in {@link View#pick} when the picking query
+         * result is available.
+         * @param result An instance of {@link PickingQueryResult}.
+         */
+        void onPick(@NonNull PickingQueryResult result);
+    }
+
+    /**
+     * Creates a picking query. Multiple queries can be created (e.g.: multi-touch).
+     * Picking queries are all executed when {@link Renderer#render} is called on this View.
+     * The provided callback is guaranteed to be called at some point in the future.
+     *
+     * Typically it takes a couple frames to receive the result of a picking query.
+     *
+     * @param x        Horizontal coordinate to query in the viewport with origin on the left.
+     * @param y        Vertical coordinate to query on the viewport with origin at the bottom.
+     * @param handler  An {@link java.util.concurrent.Executor Executor}.
+     *                 On Android this can also be a {@link android.os.Handler Handler}.
+     * @param callback User callback executed by <code>handler</code> when the picking query
+     *                 result is available.
+     */
+    public void pick(int x, int y,
+            @Nullable Object handler, @Nullable OnPickCallback callback) {
+        InternalOnPickCallback internalCallback = new InternalOnPickCallback(callback);
+        nPick(getNativeObject(), x, y, handler, internalCallback);
+    }
+
+    private static class InternalOnPickCallback implements Runnable {
+        public InternalOnPickCallback(OnPickCallback mUserCallback) {
+            this.mUserCallback = mUserCallback;
+        }
+        @Override
+        public void run() {
+            mPickingQueryResult.renderable = mRenderable;
+            mPickingQueryResult.depth = mDepth;
+            mPickingQueryResult.fragCoords[0] = mFragCoordsX;
+            mPickingQueryResult.fragCoords[1] = mFragCoordsY;
+            mPickingQueryResult.fragCoords[2] = mFragCoordsZ;
+            mUserCallback.onPick(mPickingQueryResult);
+        }
+        private final OnPickCallback mUserCallback;
+        private final PickingQueryResult mPickingQueryResult = new PickingQueryResult();
+        @Entity int mRenderable;
+        float mDepth;
+        float mFragCoordsX;
+        float mFragCoordsY;
+        float mFragCoordsZ;
+    }
 
     public long getNativeObject() {
         if (mNativeObject == 0) {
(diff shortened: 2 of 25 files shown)


libs/filabridge/include/private/filament/Variant.h
```diff
@@ -40,24 +40,33 @@ namespace filament {
         // SRE: Shadow Receiver
         // SKN: Skinning
         // DEP: Depth only
-        // FOG: Fog
+        // FOG: Fog (or picking)
```
bejado, line 43: I'm confused what fog has to do with picking. If picking is enabled, doesn't that mean fog would have to be as well (and vice-versa)?

pixelflinger: So in fact it's because I use the same bit for both FOG and PICKING.  This is because you can never have the FOG bit set when the DEPTH bit is set. So, in fact, the DEPTH bit defines what the other bits mean.

pixelflinger:
reviewers: it's almost ready, missing only the java API I think and maybe a sample.

filament/include/filament/View.h
```diff
@@ -571,6 +572,104 @@ class UTILS_PUBLIC View : public FilamentAPI {
...
+        PickingQuery& query = pick(x, y, [](PickingQueryResult const& result, PickingQuery* pq) {
+            void* user = pq->storage;
+            (*static_cast<T**>(user)->*method)(result);
+        });
+        query.storage[0] = instance;
+    }
+
+    /**
+     * Helper for creating a picking query from Foo::method, by copy for a small object
+     * e.g.: auto* query = pick<Foo, &Foo::bar>(x, y, foo);
```
romainguy, line 614: The function returns void, I don’t understand the example

filament/include/filament/View.h
```diff
@@ -571,6 +572,104 @@ class UTILS_PUBLIC View : public FilamentAPI {
...
+    struct PickingQuery {
+        void* storage[8];
+    };
+
+    /** callback type used for picking queries. */
+    using PickingQueryResultCallback = void(*)(PickingQueryResult const& result, PickingQuery* pq);
+
+    /**
+     * Helper for creating a picking query from Foo::method, by pointer.
+     * e.g.: auto* query = pick<Foo, &Foo::bar>(x, y, &foo);
```
romainguy, line 596: Pick() returns void

filament/include/filament/View.h
```diff
@@ -571,6 +572,104 @@ class UTILS_PUBLIC View : public FilamentAPI {
...
+            T* that = static_cast<T*>(user);
+            (that->*method)(result);
+            that->~T();
+        });
+        new(query.storage) T(std::move(instance));
+    }
+
+    /**
+     * Helper for creating a picking query from a small functor
+     * e.g.: auto* query = pick(x, y, [](PickingQueryResult const& result){});
```
romainguy, line 636: Same here

libs/filabridge/include/private/filament/Variant.h
```diff
@@ -97,28 +99,29 @@ namespace filament {
...
-            return isValidDepthVariant(key);
-        }
-
         inline static constexpr bool isValidDepthVariant(uint8_t variantKey) noexcept {
-            // For a variant to be a valid depth variant, all of the bits in DEPTH_MASK must be 0,
-            // except for DEPTH.
-            return (variantKey & DEPTH_MASK) == DEPTH_VARIANT;
+            // VSM and PICKING are mutually exclusive for DEPTH variants
+            return (variantKey & DEPTH_MASK) == DEPTH_VARIANT &&
+                    variantKey != 0b1110000u &&
```
romainguy, line 123: Constants?

libs/filabridge/include/private/filament/Variant.h
```diff
@@ -128,26 +131,38 @@ namespace filament {
             // 2. If SRE is set, either DYN or DIR must also be set (it makes no sense to have
             // shadows without lights).
             // 3. If VSM is set, then SRE must be set.
-            return
-                ((variantKey & DEPTH) && !isValidDepthVariant(variantKey)) ||
-                (variantKey & 0b0010111u) == 0b0000100u ||
-                (variantKey & 0b1010100u) == 0b1000000u;
+            return ((variantKey & DEPTH) && !isValidDepthVariant(variantKey)) ||
+                    (variantKey & 0b0010111u) == 0b0000100u ||
```
romainguy, line 135: Constants?

romainguy (approved):

bejado:
Should we add picking to matc's `--variant-filter`?

filament/backend/src/opengl/OpenGLDriver.cpp
```diff
@@ -2842,7 +2843,8 @@ void OpenGLDriver::readPixels(Handle<HwRenderTarget> src,
         PixelBufferDescriptor& p = *pUserBuffer;
         auto& gl = mContext;
         gl.bindBuffer(GL_PIXEL_PACK_BUFFER, pbo);
-        void* vaddr = glMapBufferRange(GL_PIXEL_PACK_BUFFER, 0,  p.size, GL_MAP_READ_BIT);
+        void* vaddr = glMapBufferRange(GL_PIXEL_PACK_BUFFER, 0,  p.size,
```
bejado, line 2846: According to https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/glMapBufferRange.xhtml `glMapBufferRange` generates `GL_INVALID_OPERATION` if both `GL_MAP_READ_BIT` and `GL_MAP_UNSYNCHRONIZED_BIT` are set.

filament/include/filament/View.h
pixelflinger: Ah yes, good point, the example is out of daten

filament/backend/src/opengl/OpenGLDriver.cpp
pixelflinger: Heh. What do you know! Good catch.

filament/include/filament/View.h
```diff
@@ -571,6 +572,115 @@ class UTILS_PUBLIC View : public FilamentAPI {
...
+        utils::Entity renderable{};     //! RenderableManager Entity at the queried coordinates
+        float depth{};                  //! Depth buffer value (1 (near plane) to 0 (infinity))
+        uint32_t reserved1{};
+        uint32_t reserved2{};
+        /**
+         * screen space coordinates in GL convention, this can be used to compute the view or
+         * world space position of the picking hit. For e.g.:
+         *   clip_space_position  = (fragCoords.xy / viewport.wh, fragCoords.z) * 2.0 - 1.0
+         *   view_space_position  = inverse(projection) * clip_space_position
+         *   world_space_position = model * view_space_position
```
romainguy, line 587: Don't you mean `inverse(view_space_position)` here? And you don't need the model matrix?

pixelflinger: `view_space_position` is a `vec3` so `inverse` of that is not possible.
The world space position is the `modelMatrix * viewSpacePosition`. Of course, `the modelMatrix` is the `inverse(viewMatrix)`. But `Camera` has a `getModelMatrix()`, so that's why I used this convention.

romainguy: Ah here you mean model matrix of the camera, I thought you meant of a renderable. It makes sense in hindsight but maybe clarify. In this case, yes, that's the correct formula (and I mean inverse(viewMatrix) not of the view position, sorry).

romainguy (approved):

android/filament-android/src/main/cpp/View.cpp
```diff
@@ -382,3 +384,29 @@ Java_com_google_android_filament_View_nIsScreenSpaceRefractionEnabled(JNIEnv *,
...
+extern "C"
+JNIEXPORT void JNICALL
+Java_com_google_android_filament_View_nPick(JNIEnv* env, jclass clazz,
+        jlong nativeView,
+        jint x, jint y, jobject handler, jobject internalCallback) {
+    View* view = (View*) nativeView;
+    JniCallback *callback = JniCallback::make(env, handler, internalCallback);
+    view->pick(x, y, [callback, env, internalCallback](View::PickingQueryResult const& result) {
+
+        auto internalOnPickCallbackClass = env->FindClass("com/google/android/filament/View$InternalOnPickCallback");
```
romainguy, line 397: Ideally you'd make those queries only once at init time. But For picking it's probably not an issue.

android/filament-android/src/main/java/com/google/android/filament/View.java
```diff
@@ -1548,6 +1548,63 @@ public DepthOfFieldOptions getDepthOfFieldOptions() {
...
+    static class PickingQueryResult {
+        /** The entity of the renderable at the picking query location */
+        @Entity int renderable;
+        /** The value of the depth buffer at the picking query location */
+        float depth;
+        /** The fragment coordinate in GL convention at the the picking query location */
+        @NonNull float[] fragCoords = new float[3];
+    };
+
+    public interface OnPickCallback {
```
romainguy, line 1563: Javadoc?

android/filament-android/src/main/java/com/google/android/filament/View.java
```diff
@@ -1548,6 +1548,63 @@ public DepthOfFieldOptions getDepthOfFieldOptions() {
...
+            mPickingQueryResult.depth = mDepth;
+            mPickingQueryResult.fragCoords[0] = mFragCoordsX;
+            mPickingQueryResult.fragCoords[1] = mFragCoordsY;
+            mPickingQueryResult.fragCoords[2] = mFragCoordsZ;
+            mUserCallback.onPick(mPickingQueryResult);
+        }
+        private final OnPickCallback mUserCallback;
+        private final PickingQueryResult mPickingQueryResult = new PickingQueryResult();
+        @Entity public int mRenderable;
+        public float mDepth;
```
romainguy, line 1603: You don't need those fields, the JNI could directly set the values in `mPickingQueryResult`.

romainguy (approved):

android/filament-android/src/main/cpp/View.cpp
pixelflinger: Yes, I was not sure where to stash the result?

android/filament-android/src/main/java/com/google/android/filament/View.java
pixelflinger: Yes, it seemed harder to do, because I would have to retrieve the object's field from jni, or pass it to the jni, and then I would have to capture it in the lambda. And technically, I was wondering if I would have to take a global ref on it (it just rely on the fact that there is one on the object that owns it).
Just seemed more complicated :-)

romainguy: Yeah but that's no different from what you do now. Except on a different class.

android/filament-android/src/main/cpp/View.cpp
romainguy: A global structure is what we use in the platform. Usually we look them up in the JNI OnLoad but probably not worth it here.

libs/filabridge/include/private/filament/Variant.h
pixelflinger: maybe later.

libs/filabridge/include/private/filament/Variant.h
pixelflinger: maybe later.

android/filament-android/src/main/java/com/google/android/filament/View.java
pixelflinger: I was going to go down this route, and then realized that I would need to manage the array of 3 floats from JNI, so decided to keep it simple for now.

romainguy: Using an array from JNI is easier than looking up/setting 3 fields :) Anyway it doesn't matter much, it would just make the code simpler.

zcufo123:
@pixelflinger 

Thank you for providing this handy API for us and it can make our lives easier.

I have one question for you: If one picking query is requested and multiple entity is selected, how would the PickingQueryResult return these multiple entities?

romainguy:
It won't, it will return the closest Entity.

ozzbI:
Is there any way to get picked material or primitive? Or idea what should be done to make it possible?

flykule:
> Is there any way to get picked material or primitive? Or idea what should be done to make it possible?

The pick result callback has returned nearest entity. You can use renderableManager.getMaterialInstance to get selected material .

ozzbI:
> > Is there any way to get picked material or primitive? Or idea what should be done to make it possible?
> 
> The pick result callback has returned nearest entity. You can use renderableManager.getMaterialInstance to get selected material .

Thank you.
In my case instance can have multiple primitives with different materials attached to each primitive.
I can iterate over primitives and get all materials, but I am looking how to get particular material (or primitive index) which was picked.

PR #4658, opened 2021-09-22
Podspec: include headers in nested directories

alexey-pelykh:


ios/CocoaPods/Filament.podspec
```diff
@@ -50,12 +50,12 @@ Pod::Spec.new do |spec|
   end
 
   spec.subspec "gltfio_core" do |ss|
-    ss.source_files = "include/gltfio/*.h"
+    ss.source_files = "include/gltfio/**/*.h"
```
romainguy, line 53: gltfio doesn't have nested directories for its public headers

alexey-pelykh: https://github.com/google/filament/blob/main/libs/utils/include/utils/Mutex.h#L23 that looks like include of a nested directory to me

romainguy: That's in `utils`, not `gltfio`.

alexey-pelykh: 🤦‍♂️ sorry, you're right. Let me adjust that

alexey-pelykh: done

romainguy:
Thanks! I'll let @bejado finish the review because he's our iOS expert. I don't see any issues but I'm not familiar with podspecs so… :)

bejado (approved):
Thank you!

PR #4674, opened 2021-09-24
Fix, missing call to setTransparencyMode

bejado:


romainguy (approved):
Ouch good catch. I'm surprised this didn't cause more issues.

PR #4678, opened 2021-09-28
rework Vulkan image layout transitions

pixelflinger:
The main issue that gets fixed here are:

- VK_IMAGE_LAYOUT_UNDEFINED cannot be used if the image content is to 
be preserved after the transition. Unfortunately, there is more than a 
few places in the VK backend where we use VK_IMAGE_LAYOUT_UNDEFINED on
images that are later used as sources.

- accessing an image memory from the host is only supported if the the 
layout is PREINITIALIZED or GENERAL. However, images in the staging
pool were transitioned to other layouts and put back into the pool,
later they would be mapped/memcpy'ed into. Instead we now always use
the GENERAL layout and never change it.

- images that where in the pool where transitioned using the miplevel
of the texture we were copying into, however this didn't make sense
because images from the pool don't have mip levels.

As part of these fixes, we also refactor imageLayoutTranstion(),
because there was two copies of this function, one specific to texture
and another one more generic. We now always use the more generic one,
with a new helper for textures.

Note: didn't fix "readPixels" which has at least problem (1), because
it's a tricky implementation, and it requires more toughts.

### DIFF
--- a/filament/backend/src/vulkan/VulkanBlitter.cpp
+++ b/filament/backend/src/vulkan/VulkanBlitter.cpp
@@ -38,30 +38,6 @@ struct BlitterUniforms {
     float inverseSampleCount;
 };
 
-// Helper function for populating barrier fields based on the desired image layout.
-// This logic is specific to blitting, please keep this private to VulkanBlitter.
-static VulkanLayoutTransition transitionHelper(VulkanLayoutTransition transition) {
-    switch (transition.newLayout) {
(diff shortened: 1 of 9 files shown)


romainguy (approved):
Everything looks fine but I can't comment on the validity of the Vulkan code. Based on the PR description, the code seems correct though.

chrisforbes:
Looks at least *better* from vulkan correctness pov.

PR #4679, opened 2021-09-28
fix readPixels vertical offset in vk backend

pixelflinger:
vulkan expected a vertical offset from the top, but the api is in
gl convention with the offset from the bottom.


romainguy (approved):

prideout (approved):
Interestingly, the RHS is unsigned and the LHS is signed, but this change looks fine to me!

PR #4682, opened 2021-09-28
Metal: support integer formats with readPixels

bejado:
This change enables more formats for readPixels, notably `UINT`/`R_INTEGER` for picking. Formats that require a conversion to an integer type are not yet supported, due to current limitations of `MetalBlitter`.

### DIFF
--- a/filament/backend/test/BackendTest.cpp
+++ b/filament/backend/test/BackendTest.cpp
@@ -178,5 +178,65 @@ int runTests() {
     return RUN_ALL_TESTS();
 }
 
+void getPixelInfo(PixelDataFormat format, PixelDataType type, size_t& outComponents, int& outBpp) {
+    switch (format) {
+        case PixelDataFormat::UNUSED:
+        case PixelDataFormat::R:
+        case PixelDataFormat::R_INTEGER:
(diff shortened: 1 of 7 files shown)


filament/backend/test/BackendTest.cpp
```diff
@@ -178,5 +178,65 @@ int runTests() {
...
+            break;
+        case PixelDataFormat::RGBA:
+        case PixelDataFormat::RGBA_INTEGER:
+            outComponents = 4;
+            break;
+    }
+
+    outBpp = outComponents;
+    switch (type) {
+        case PixelDataType::COMPRESSED:
```
romainguy, line 207: Isn't the compressed case more complicated and depends on the compression format?

romainguy (approved):

pixelflinger (approved):

filament/backend/test/BackendTest.cpp
bejado: Yeah, I adapted this from `PixelBufferDescriptor::computeDataSize` so the `COMPRESSION` case is just to silence warnings.

bejado: I'll add an assert for this.

PR #4687, opened 2021-09-30
Update BUILDING.md

wingdi:
NDK 23.0.7599858 don't have aarch64-linux-android-ar, use it will build fail.


romainguy:
We'll fix the build instead. Thanks for the contribution though.

PR #4688, opened 2021-09-30
improve user callbacks latency

pixelflinger:
A new Engine::pumpMessageQueues() method can be used to trigger all
pending user callbacks right away. This can help reducing latency
of certain callbacks such as picking queries.
This is typically called once before calculating the next frame's
parameters (e.g. object positions or appearance that may depend on a 
callback result).

Fix some comments and documentation.


romainguy (approved):
👍

bejado (approved):

PR #4702, opened 2021-10-06
better API for MSAA options

pixelflinger:


android/filament-android/src/main/java/com/google/android/filament/View.java
```diff
@@ -291,6 +292,29 @@
...
+         *      1: no anti-aliasing
+         *      n: sample count. Effective sample could be different depending on the
+         *         GPU capabilities.
+         */
+        public int sampleCount = 4;
+
+        /**
+         * custom resolve improves quality for HDR scenes, but may impact performance.
+         */
+        public boolean customResolve = false;
```
romainguy, line 315: Should this be an enum if we ever want to add different filters?

romainguy (approved):

android/filament-android/src/main/java/com/google/android/filament/View.java
pixelflinger: Yes, this is kind of a placeholder for now, it'll probably change.

PR #4748, opened 2021-10-20
Support for DPCF

pixelflinger:


pixelflinger:
Good question about the material version. I think old materials should work as long as the new feature is not used. Right?

PR #4749, opened 2021-10-20
Fix, call VirtualMachineEnv::JNI_OnLoad for non-Android Java builds

bejado:
We need to make sure that `VirtualMachineEnv` is initialized now for non-Android builds.


bejado:
I think we could simply call `::filament::VirtualMachineEnv::JNI_OnLoad` unconditionally, but we can revisit for the next release

PR #4758, opened 2021-10-23
[WIP] Wayland Support

jwinarske:
Preliminary Wayland Support for Desktop Filament.  WIP.

Building with:
```
CC=$LLVM_ROOT/bin/clang CXX=$LLVM_ROOT/bin/clang++ cmake .. -D BUILD_SHARED_LIBS=OFF -D CMAKE_BUILD_TYPE=Release -D FILAMENT_SUPPORTS_VULKAN=TRUE -D FILAMENT_SUPPORTS_OPENGL=FALSE -D FILAMENT_SUPPORTS_XCB=FALSE -D FILAMENT_SUPPORTS_XLIB=FALSE -D FILAMENT_SUPPORTS_WAYLAND=TRUE
```


jwinarske:
@romainguy If you're open to it, I will create a new PR that only has the Vulkan Wayland changes.  It would not include any libsdl2 changes, as not everyone wants or uses SDL2.  That would make demo apps require X11; until latest sdl2 release is pulled in, or a change set on top of what you already have.

.github/workflows/android-continuous.yml
```diff
@@ -1,11 +1,13 @@
 name: Android
 
 on:
+  workflow_dispatch:
```
romainguy, line 4: Are those workflow changes only there while you're working on the PR?

CMakeLists.txt
```diff
@@ -1,7 +1,7 @@
 # ==================================================================================================
 # CMake
 # ==================================================================================================
-cmake_minimum_required(VERSION 3.19)
+cmake_minimum_required(VERSION 3.16.5)
```
romainguy, line 4: Don't lower the CMake requirements

CMakeLists.txt
```diff
@@ -94,22 +96,42 @@ if (UNIX AND NOT APPLE AND NOT ANDROID AND NOT WEBGL)
...
+            add_definitions(-DFILAMENT_SUPPORTS_X11)
+            set(FILAMENT_SUPPORTS_X11 TRUE)
+        endif()
     endif()
 
-    if (FILAMENT_SUPPORTS_XLIB)
-        add_definitions(-DFILAMENT_SUPPORTS_XLIB)
+    get_property(LIB64 GLOBAL PROPERTY FIND_LIBRARY_USE_LIB64_PATHS)
+
+    if ("${LIB64}" STREQUAL "TRUE")
```
romainguy, line 119: What is this change for?

CMakeLists.txt
```diff
@@ -94,22 +96,42 @@ if (UNIX AND NOT APPLE AND NOT ANDROID AND NOT WEBGL)
...
+        set(LIBSUFFIX 64)
+    else()
+        set(LIBSUFFIX "")
     endif()
 endif()
 
 if (ANDROID OR WEBGL OR IOS)
     set(IS_MOBILE_TARGET TRUE)
-endif()
+    message(STATUS "Target Build")
```
romainguy, line 128: This message isn't necessary and unrelated to your changes

CMakeLists.txt
```diff
@@ -94,22 +96,42 @@ if (UNIX AND NOT APPLE AND NOT ANDROID AND NOT WEBGL)
...
+endif ()
 
 if (NOT ANDROID AND NOT WEBGL AND NOT IOS)
     set(IS_HOST_PLATFORM TRUE)
-endif()
+    message(STATUS "Host Build")
```
romainguy, line 133: This message isn't necessary and unrelated to your changes

CMakeLists.txt
```diff
@@ -696,6 +704,10 @@ if (IS_HOST_PLATFORM)
     add_subdirectory(${TOOLS}/specular-color)
 endif()
 
+if (FILAMENT_SUPPORTS_VULKAN)
+    add_subdirectory(${LIBRARIES}/bluevk)
```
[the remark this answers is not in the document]
romainguy: I still need to thing about it, I really don't like relying on external build systems this way. Forcing configs to a few options is a feature in our case.

filament/backend/src/Platform.cpp
```diff
@@ -115,7 +119,11 @@ DefaultPlatform* DefaultPlatform::create(Backend* backend) noexcept {
             #elif defined(IOS)
                 return new PlatformVkCocoaTouch();
             #elif defined(__linux__)
-                return new PlatformVkLinux();
+                #if defined (FILAMENT_SUPPORTS_WAYLAND)
+                    return new PlatformVkLinuxWayland();
```
romainguy, line 123: So what happens when support for X11 is off but the user tries to use OpenGL instead of Vulkan?

filament/backend/CMakeLists.txt
```diff
@@ -372,7 +376,7 @@ install(DIRECTORY ${PUBLIC_HDR_DIR}/backend DESTINATION include)
 # ==================================================================================================
 option(INSTALL_BACKEND_TEST "Install the backend test library so it can be consumed on iOS" OFF)
 
-if (APPLE)
+if (APPLE OR LINUX)
```
romainguy, line 379: Good catch

romainguy (commented):
Additional question: is there any way to support both X11 and Wayland easily at runtime? I'm not familiar enough with the implications.

.github/workflows/android-continuous.yml
jwinarske: So I can trigger workflow since I'm not merging into main

CMakeLists.txt
jwinarske: lib vs lib64.  Fedora prefers lib64, whereas Ubuntu does not.  It's a Linux compatibility thing.

CMakeLists.txt
jwinarske: will remove

CMakeLists.txt
jwinarske: will remove

filament/backend/src/Platform.cpp
jwinarske: I hadn't got that far on testing.  Both should co-exist.

CMakeLists.txt
jwinarske: I will revert.  Is there a feature currently implemented that's dependent on 3.19?

Yocto dunfell currently has CMake at 3.16.5.  I will need to create bbappend recipe to build with newer CMake, and that will be a dependency for filament recipe (klunky).

jwinarske:
> Additional question: is there any way to support both X11 and Wayland easily at runtime? I'm not familiar enough with the implications.

A couple of scenarios come to mind why to have them mutually exclusive:
* Minimize code size
* In most embedded Linux builds X11 support is not present, only Wayland.  Fedora and Ubuntu 20 ship standard with Wayland enabled, and X11 support has to be manually added.  Yocto project recently made Wayland default in SATO builds, it was X prior.
* If you have Wayland available why run a slower frame rate via WaylandX (X emulation)?
* If you have a Wayland system it will have WaylandX, which X11 build would run on (slower frame rate).

CMakeLists.txt
jwinarske: Currently the only consumer is libsdl2, but really it's a global thing.  I suspect it would just be moved later

jwinarske:
Closing in favor of https://github.com/google/filament/pull/4803

PR #4802, opened 2021-11-02
Enforce readPixels is called within frame

bejado:


filament/src/Renderer.cpp
```diff
@@ -1171,6 +1173,10 @@ void FRenderer::endFrame() {
 
 void FRenderer::readPixels(uint32_t xoffset, uint32_t yoffset, uint32_t width, uint32_t height,
         PixelBufferDescriptor&& buffer) {
+#ifndef NDEBUG
```
romainguy, line 1176: Is there anything useful we could do in release build?

romainguy (approved):

filament/src/Renderer.cpp
pixelflinger: The only thing we could do is skip the call (and call the callback). I'm against logging something because it can quickly become log spam, unless we have a way to log just once.

filament/src/Renderer.cpp
```diff
@@ -1057,6 +1057,7 @@ bool FRenderer::beginFrame(FSwapChain* swapChain, uint64_t vsyncSteadyClockTimeN
         FEngine::DriverApi& driver = engine.getDriverApi();
 
         driver.beginFrame(appVsync.time_since_epoch().count(), mFrameId);
+        mWithinFrame = true;
```
pixelflinger, line 1060: I think you should just use `mSwapChain` instead of adding a boolean.

pixelflinger (changes requested):

filament/src/Renderer.cpp
romainguy: I was also wondering if we could change the API to return true/false for instance or something like this. I'm afraid just doing a no-op would lead to annoying bug reports.

pixelflinger: It would be a bit weird because basically the return value would be completely determined by whether it's called issue l inside begin/end frame. So it's not very different from just documenting it.

Also if we return an error, what do we do with the callback? 

It would be weird to call it if there was a reported error, on the other hand, if we don't, we could leak memory.

Usually filament silently fails when you don't use it properly, I think it should be the same here. Logging something just once could be nice.

pixelflinger (approved):

filament/src/Renderer.cpp
bejado: Note that `readPixels` in release builds will continue to work as before. The `NDEBUG` check is just there to mitigate risk and can be removed in the future.

prideout:
The "Export Screenshot" checkbox in gltf_viewer (in the Automation section) triggers this assert.

PR #4803, opened 2021-11-02
Vukan Wayland

jwinarske:
@romainguy I think this PR is more inline with your expectations around minimal build changes.

I have all the plumbing mostly there, currently working through EGL BE bring-up.  If there is no interest to support EGL BE, I'll remove it.  We (Toyota) are moving entirely to Vulkan, and this EGL addition was just to complete the thought.

@pixelflinger Any strong opinions on [vulkan/vulkan.hpp](https://github.com/KhronosGroup/Vulkan-Hpp)?  The author recently added RAII support.  I'm looking at using vulkan/vulkan.hpp for a Flutter Vulkan mock-up.


jwinarske:
I believe SDL needs to move forward from 2.0.8 to 2.0.15 in order to support Wayland EGL.  SDL 2.0.15 exposures egl_window pointer via SDL_GetWindowWMInfo:
https://github.com/jwinarske/SDL/blob/main/src/video/wayland/SDL_waylandwindow.c#L637

Keeping in mind Wayland blocks the use of pbuffer surfaces in favor of using FBO.

filament/backend/src/opengl/platforms/PlatformEGLWayland.h
```diff
@@ -0,0 +1,95 @@
...
+#include <EGL/eglext.h>
+#include <wayland-egl.h>
+
+#include <backend/DriverEnums.h>
+
+#include "private/backend/OpenGLPlatform.h"
+
+namespace filament {
+
+class PlatformEGLWayland : public backend::OpenGLPlatform {
```
pixelflinger, line 32: This should derive from `PlatformEGL`, there should be a lot of code that can be shared between both.
Is there/what are the limitations in `PlatformEGL` that prevent it from being used here?

pixelflinger (changes requested):
Thank you, I think this looks reasonable.  The only change I would like to make here is to use `PlatformEGL` for the wayland EGL implementation.

.github/workflows/android-continuous.yml
```diff
@@ -6,6 +6,7 @@ on:
       - main
       - release
       - rc/**
+  workflow_dispatch:
```
bejado, line 9: Please remove these GitHub workflow changes

filament/backend/CMakeLists.txt
```diff
@@ -98,14 +98,26 @@ if (FILAMENT_SUPPORTS_OPENGL AND NOT FILAMENT_USE_EXTERNAL_GLES3 AND NOT FILAMEN
...
+        endif()
     elseif (WIN32)
         list(APPEND SRCS src/opengl/platforms/PlatformWGL.cpp)
     else()
         list(APPEND SRCS src/opengl/platforms/PlatformDummyGL.cpp)
     endif()
 endif()
 
+if (FILAMENT_SUPPORTS_WAYLAND)
+    add_compile_definitions(
```
bejado, line 114: Can we use `target_compile_definitions` here instead?

libs/filamentapp/include/filamentapp/FilamentApp.h
```diff
@@ -185,6 +185,15 @@ class FilamentApp {
         filament::Renderer* mRenderer = nullptr;
         filament::Engine::Backend mBackend;
 
+        #if defined(FILAMENT_SUPPORTS_WAYLAND)
```
bejado, line 188:
```suggestion
#if defined(FILAMENT_SUPPORTS_WAYLAND)
```

libs/filamentapp/include/filamentapp/FilamentApp.h
```diff
@@ -185,6 +185,15 @@ class FilamentApp {
...
+        struct {
+            void *eglSharedContext;
+            struct wl_display *display;
+            struct wl_surface *surface;
+            struct wl_egl_window *egl_window;
+        } mWayland{};
+        #endif
```
bejado, line 195:
```suggestion
#endif
```

libs/filamentapp/include/filamentapp/FilamentApp.h
```diff
@@ -185,6 +185,15 @@ class FilamentApp {
         filament::Renderer* mRenderer = nullptr;
         filament::Engine::Backend mBackend;
 
+        #if defined(FILAMENT_SUPPORTS_WAYLAND)
+        struct {
```
bejado, line 189: Can this be moved to the cpp file?

.github/workflows/android-continuous.yml
jwinarske: done

filament/backend/CMakeLists.txt
jwinarske: good point.  I made them target specific and moved them after the backend target def.

filament/backend/CMakeLists.txt
```diff
@@ -98,14 +98,26 @@ if (FILAMENT_SUPPORTS_OPENGL AND NOT FILAMENT_USE_EXTERNAL_GLES3 AND NOT FILAMEN
...
+        WL_EGL_PLATFORM
+        EGL_NO_X11
+        MESA_EGL_NO_X11_HEADERS
```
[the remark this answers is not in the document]
jwinarske: done.  moved to target after backed target def.

libs/filamentapp/include/filamentapp/FilamentApp.h
jwinarske: moved to cpp.

jwinarske:
@googlebot I fixed it

filament/backend/src/opengl/platforms/PlatformEGLWayland.h
pixelflinger: Okay, so in that case, I think it is better to pull out all the GL/EGL changes from this PR, and keep only the Vulkan ones.  We can discuss/work on the GL/EGL stuff separately, in another PR. It's better to not tie them together.
Thanks.

pixelflinger: BTW, Is Wayland guaranteed to have `EGL_KHR_no_config_context`? If so, we don't need the "dummy" pbuffer and we could change EGLPlatform to skip creating the pbuffer in that case.

jwinarske: @pixelflinger Sounds good.  Will do.

Wayland does not guarantee `EGL_KHR_no_config_context`.  The source shows it is just checking if it's present, so whatever the driver provides.  I'm finding it not present on a few machines.
https://github.com/wayland-project/weston/search?q=EGL_KHR_no_config_context

jwinarske: What is odd is that I'm running fairly Mesa drivers from last two releases and PlatformEGL is not detecting the `EGL_KHR_no_config_context` extension.  I would expect it should be present.  I'll dig into that a bit more prior to pulling it.

jwinarske: I played around a bit extending PlatformEGL.  Passing `EGL_RECORDABLE_ANDROID` generates `EGL_BAD_CONFIG` error, and it exits.  Wrapping line 113 with `#if defined(__ANDROID__)` I moved on.  Commenting out eglCreatePbufferSurface (mEGLDummySurface == nullptr) context is created and function returns.  

Later hitting 
```
in static filament::backend::Driver *filament::OpenGLDriver::create(filament::backend::OpenGLPlatform *const, void *const):135
in file /b/toyota/filament/filament/backend/src/opengl/OpenGLDriver.cpp
reason: OpenGL ES 3.0 minimum needed (current 0.0)
```

What about the primary EGL surface created in initial SDL create window call?  This EGL surface pointer is only exposed in SDL >= 2.0.15.

jwinarske: @pixelflinger I backed out Wayland EGL.

libs/filamentapp/include/filamentapp/FilamentApp.h
```diff
@@ -73,8 +73,8 @@ class FilamentApp {
 
     void setDropHandler(DropCallback handler) { mDropHandler = handler; }
 
-    void run(const Config& config, SetupCallback setup, CleanupCallback cleanup,
-            ImGuiCallback imgui = ImGuiCallback(), PreRenderCallback preRender = PreRenderCallback(),
+    void run(const Config& config, const SetupCallback& setup, const CleanupCallback& cleanup,
+            ImGuiCallback imgui = ImGuiCallback(), const PreRenderCallback& preRender = PreRenderCallback(),
```
pixelflinger, line 77: We appreciate this change, but it's not related to this PR -- so I would prefer to revert it here. It's just for clarity and not be confused later.

libs/filamentapp/src/FilamentApp.cpp
```diff
@@ -73,9 +75,9 @@ View* FilamentApp::getGuiView() const noexcept {
     return mImGuiHelper->getView();
 }
 
-void FilamentApp::run(const Config& config, SetupCallback setupCallback,
-        CleanupCallback cleanupCallback, ImGuiCallback imguiCallback,
-        PreRenderCallback preRender, PostRenderCallback postRender,
+void FilamentApp::run(const Config& config, const SetupCallback& setupCallback,
+        const CleanupCallback& cleanupCallback, ImGuiCallback imguiCallback,
+        const PreRenderCallback& preRender, PostRenderCallback postRender,
```
pixelflinger, line 80: change not related to this PR.

libs/filamentapp/src/FilamentApp.cpp
```diff
@@ -532,15 +534,41 @@ FilamentApp::Window::Window(FilamentApp* filamentApp,
...
+        SDL_VERSION(&wmi.version);
+        ASSERT_POSTCONDITION(SDL_GetWindowWMInfo(mWindow, &wmi), "SDL version unsupported!");
+        if (wmi.subsystem == SDL_SYSWM_WAYLAND) {
+            wayland.eglSharedContext = nullptr;
+            wayland.display = wmi.info.wl.display;
+            wayland.surface = wmi.info.wl.surface;
+            wayland.egl_window = nullptr; //TODO - needs to come from SDL -> only 2.0.15+ exposes
+        }
+        void* nativeWindow = &wayland;
+
```
pixelflinger, line 556: I think I'm confused here.  I thought this change only added support for Wayland on Vulkan, correct? So why is there any "EGL" stuff here?

pixelflinger (changes requested):
Thank you for the changes. Sorry I missed one before about the added 'const', which are not related to this PR. 
I also have a question about some EGL stuff, since this PR is about addind Vulkan Wayland support.

libs/filamentapp/src/FilamentApp.cpp
jwinarske: removed egl related variables in wayland struct.

libs/filamentapp/include/filamentapp/FilamentApp.h
jwinarske: removed const.

libs/filamentapp/src/FilamentApp.cpp
jwinarske: removed const.

jwinarske:
@pixelflinger What is left on this?

pixelflinger:
> @pixelflinger What is left on this

Probably not much, other than my being on vacations this week and last approval from @bejado and @romainguy 

Thank you again for all the great work.

bejado (approved):

PR #4811, opened 2021-11-04
Add a (crude) way to have structs in our UBOs

pixelflinger:
The struct must be declared in common_type.fs, so custom
structures are not supported.

### DIFF
--- a/libs/filabridge/include/private/filament/UniformInterfaceBlock.h
+++ b/libs/filabridge/include/private/filament/UniformInterfaceBlock.h
@@ -63,10 +63,21 @@ public:
         Builder& add(utils::StaticString const& uniformName, size_t size,
                 Type type, Precision precision = Precision::DEFAULT);
 
+        // Add a known struct
+        Builder& add(utils::CString const& uniformName, size_t size,
+                utils::CString const& structName, size_t stride);
+
         template<size_t N>
         Builder& add(utils::StringLiteral<N> const& uniformName, size_t size,
                 Type type, Precision precision = Precision::DEFAULT) {
             return add(utils::StaticString{ uniformName }, size, type, precision);
+        }
+
+        template<size_t N0, size_t N1>
+        Builder& add(utils::StringLiteral<N0> const& uniformName, size_t size,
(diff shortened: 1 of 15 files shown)


shaders/src/common_types.fs
```diff
@@ -22,3 +22,10 @@
 
 #define float3x3 mat3
 #define float4x4 mat4
+
+struct ShadowData {
+    highp mat4 spotlightFromWorldMatrix;
```
romainguy, line 27: This won't work on Adreno, `highp` is ignored inside `struct`...

romainguy: Although I guess that will work inside a UBO? If so you should probable leave a comment explaining this struct MUST be used in a UBO.

romainguy (approved):

libs/filabridge/include/private/filament/UniformInterfaceBlock.h
```diff
@@ -63,23 +63,38 @@ class UniformInterfaceBlock {
...
+            return add(utils::StaticString{ uniformName }, size,
+                    utils::StaticString{ structName }, stride);
+        }
+
         // build and return the UniformInterfaceBlock
         UniformInterfaceBlock build();
     private:
         friend class UniformInterfaceBlock;
         struct Entry {
             Entry(utils::CString name, uint32_t size, Type type, Precision precision) noexcept
```
bejado, line 88: Let's initialize `stride` here for non-struct types with `strideForType(type)`. That way `stride` is always accurate.

libs/filabridge/src/UniformInterfaceBlock.cpp
```diff
@@ -89,7 +95,7 @@ UniformInterfaceBlock::UniformInterfaceBlock(Builder const& builder) noexcept
     uint16_t offset = 0;
     for (auto const& e : builder.mEntries) {
         size_t alignment = baseAlignmentForType(e.type);
-        uint8_t stride = strideForType(e.type);
+        uint8_t stride = strideForType(e.type, e.stride);
```
bejado, line 98: You should be able to simply do `e.stride` now.

PR #4823, opened 2021-11-09
clean-up filamat

pixelflinger:
- workaround a CLion bug (IDE only)
- simplify how we decide to use the optimized depth vertex shader
- get rid of depth_main.vs


libs/filamat/src/MaterialBuilder.cpp
```diff
@@ -967,10 +961,30 @@ Package MaterialBuilder::build(JobSystem& jobSystem) noexcept {
...
+bool MaterialBuilder::hasCustomVaryings() const noexcept {
+    for (const auto& variable : mVariables) {
+        if (!variable.empty()) {
+            return true;
+        }
+    }
+    return false;
+}
+
+bool MaterialBuilder::needsStandardDepthProgram() const noexcept {
```
romainguy, line 973: By "standard" here you mean vertex+fragment we use for the color pass, so not the "depth" program right?

romainguy (approved):
I think it's correct :)

libs/filamat/src/MaterialBuilder.cpp
pixelflinger: No actually, I meant the standard vertex shader + the depth fragment shader.

At no time do we use the standard (ie color) fragment shader with the DEPTH variant.

I think this is correct, there is nothing in the standard fragment shader that we need when we're only doing the depth.

romainguy: All we sometimes need is alpha testing (we need to discard based on alpha). As long as we have that test when needed we're good.

pixelflinger: Ah yes, `depth_main.fs` does that indeed:
```
#if defined(BLEND_MODE_MASKED)
    if (alpha < getMaskThreshold()) {
        discard;
    }
#endif
```

Issue #4822, opened 2021-11-09
Configuration cache issues when building sample-gltf-viewer on Android


bejado:
**To Reproduce**
1. Do a clean build for Android and desktop:
```
git clean -qfX android
./build.sh -i -p android release
./build.sh -i -p desktop release
```

2. Attempt to build sample-gltf-viewer.

```
cd android
./gradlew -Pcom.google.android.filament.dist-dir=../out/android-release/filament :samples:sample-gltf-viewer:assembleRelease
```

**Error message:**

```
FAILURE: Build failed with an exception.

* What went wrong:
Configuration cache problems found in this build.

1 problem was found storing the configuration cache.
- Task `:samples:sample-gltf-viewer:collectReleaseDependencies` of type `com.android.build.gradle.internal.tasks.PerModuleReportDependenciesTask`: invocation of 'Task.project' at execution time is unsupported.
  See https://docs.gradle.org/7.0.2/userguide/configuration_cache.html#config_cache:requirements:use_project_during_execution

See the complete report at file:///Users/bendoherty/code/filament/android/build/reports/configuration-cache/agn5rul0upywryazkatbyjcut/configuration-cache-report.html
> Invocation of 'Task.project' by task ':samples:sample-gltf-viewer:collectReleaseDependencies' at execution time is unsupported.

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output. Run with --scan to get full insights.

* Get more help at https://help.gradle.org

BUILD FAILED in 22s
141 actionable tasks: 68 executed, 73 up-to-date
Configuration cache entry discarded with 1 problem.
```

PR #4831, opened 2021-11-09
Skip task incompatible with configuration caching

romainguy:
Fixes issue #4822

### DIFF
--- a/android/samples/sample-gltf-viewer/build.gradle
+++ b/android/samples/sample-gltf-viewer/build.gradle
@@ -4,7 +4,7 @@ plugins {
     id 'filament-tools-plugin'
 }
 
-project.ext.isSample = true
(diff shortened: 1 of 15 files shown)


android/samples/sample-gltf-viewer/build.gradle
```diff
@@ -4,7 +4,7 @@ plugins {
     id 'filament-tools-plugin'
 }
 
-project.ext.isSample = true
+//project.ext.isSample = true
```
pixelflinger, line 7: Is this intentional?

pixelflinger (approved):

android/samples/sample-gltf-viewer/build.gradle
romainguy: Ah no it's not, good catch.

PR #4834, opened 2021-11-10
Use workflow_dispatch instead of repository_dispatch

bejado:
This will allow us to trigger release builds from the GitHub UI.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,14 +1,25 @@
 name: Release
 
 # This Workflow can be triggered two ways:
-# 1. A GitHub release is created (using the GitHub web UI). This triggers all of the platforms to build and upload assets.
-# 2. A repository_dispatch API event is sent. This triggers a build for only the platform specified in the dispatch event.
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -1,22 +1,33 @@
...
 env:
-  RELEASE_TAG: ${{ github.event.client_payload.release_tag }}
+  RELEASE_TAG: ${{ github.event.inputs.release_tag }}
 
 on:
-  repository_dispatch:
+  workflow_dispatch:
+    inputs:
+      platform:
+        description: 'Platform to build (desktop, web, android, ios, windows)'
```
romainguy, line 16: Is windows not desktop/is desktop Linux/macOS?

romainguy (approved):

.github/workflows/release.yml
bejado: Yes, desktop is Linux/macOS. We should probably separate those two out at some point.

PR #4842, opened 2021-11-11
speed up under ThreadSanitizer

bejado:


libs/utils/include/utils/compiler.h
```diff
@@ -33,7 +33,7 @@
 #if __has_attribute(visibility)
 #    define UTILS_PUBLIC  __attribute__((visibility("default")))
 #else
-#    define UTILS_PUBLIC
```
romainguy, line 36: This doesn't match the indent above

libs/utils/include/utils/compiler.h
```diff
@@ -65,19 +65,21 @@
 #endif
 
 #define UTILS_NO_SANITIZE_THREAD
-#if defined(__has_feature)
-#    if __has_feature(thread_sanitizer)
-#        undef UTILS_NO_SANITIZE_THREAD
-#        define UTILS_NO_SANITIZE_THREAD __attribute__((no_sanitize("thread")))
-#    endif
+#if __has_feature(thread_sanitizer)
```
bejado, line 68: I think we still need to check for `defined(__has_feature)`, otherwise MSVC complains.

dvyukov: There is:
```
#ifndef __has_feature
#define __has_feature(x) 0
#endif
```
at the top of this file and existing uses of __has_feature w/o the check, e.g.:
```
#elif __has_feature(cxx_thread_local)
```

bejado: Ah thanks I see that now.

pixelflinger (approved):
the gcc/clang one is a very good catch!

PR #4843, opened 2021-11-11
several improvements to shadow mapping code

pixelflinger:
This PR contains several independent CLs that improve shadow mapping code:

- store world-space texel size in shadow UBO
- simplify shadow position computation
- move shadow position computation inside `shadow()`
- rearrange EVSM code so it's easy to test the full EVSM calculation
- apply gaussian blur only on channels present in the destination


filament/CMakeLists.txt
```diff
@@ -198,7 +198,10 @@ set(MATERIAL_SRCS
         src/materials/skybox.mat
         src/materials/ssao/sao.mat
         src/materials/ssao/saoBentNormals.mat
-        src/materials/separableGaussianBlur.mat
+        src/materials/separableGaussianBlur1.mat
```
romainguy, line 201: Did you test the impact on binary size before/after?

shaders/src/depth_main.fs
```diff
@@ -38,8 +40,20 @@ void main() {
 
 #if defined(HAS_VSM)
     // interpolated depth is stored in vertex_worldPosition.w (see main.vs)
-    highp float depth = exp(vertex_worldPosition.w);
+    highp float depth = vertex_worldPosition.w;
+    depth = exp(depth);
+    fragColor.xy = computeDepthMomentsVSM(depth);
+    fragColor.zw = vec2(0.0);
+    //fragColor.zw = computeDepthMomentsVSM(-1.0/depth);
```
romainguy, line 47: remove?

shaders/src/shadowing.fs
```diff
@@ -167,16 +172,46 @@ float ShadowSample_VSM(const mediump sampler2DArray shadowMap,
  */
 
 // PCF sampling
-float shadow(const mediump sampler2DArrayShadow shadowMap,
-        const uint layer, const highp vec4 shadowPosition) {
+float shadow(const bool DIRECTIONAL,
```
romainguy, line 175: I see what you are doing but it's weird to have a function parameter in upper case.

romainguy (approved):

filament/CMakeLists.txt
pixelflinger: no, but do we even have some guidance about that? Generally we have prioritized performance.

shaders/src/depth_main.fs
pixelflinger: no I want to keep it so it's easy to activate in `matdbg` -- e.g. when seeing artifacts with EVSM, it's nice to be able to quickly see if it's because of that.

filament/CMakeLists.txt
romainguy: It depends on what the increase is. If it adds 100 KiB to a single target, I'd say it's not worth it.

shaders/src/depth_main.fs
romainguy: Then leave a comment because otherwise I'm sure I'll remove this random piece of code next time I see it since I will have forgotten why you want it.

shaders/src/shadowing.fs
pixelflinger: Yeah -- it's because it's a constant at compile time, so it's like a template parameter really. I'm open to better ideas.
One idea would be to `#define constexpr const` so we'd use that in the declaration (but the IDE might choke).

romainguy: I would just call it `directional` and leave it at at that 🤷.

shaders/src/shadowing.fs
```diff
@@ -167,16 +172,46 @@ float ShadowSample_VSM(const mediump sampler2DArray shadowMap,
...
     // note: shadowPosition.z is in the [1, 0] range (reversed Z)
     return ShadowSample_PCF(shadowMap, layer, position);
 }
 
 // VSM sampling
-float shadow(const mediump sampler2DArray shadowMap,
-        const uint layer, const highp vec4 shadowPosition) {
+float shadow(const bool DIRECTIONAL,
+        const mediump sampler2DArray shadowMap,
+        const uint layer, const uint index, const uint cascade) {
```
bejado, line 200: Can we make `index`/`cascade` into a single argument?

pixelflinger: Yes we could, but I thought it would be nice to keep them separate, in case we want to have several directional shadow in the future.

PR #4844, opened 2021-11-13
Fix spotlight normal bias calculation

pixelflinger:
The calculation was completely wrong due to several typos more or less
canceling each other, that led to some misunderstanding!


romainguy (approved):

bejado (approved):
Does the material version need to be bumped?

pixelflinger:
> Does the material version need to be bumped?

mh I think so yes. good thinking.

PR #4848, opened 2021-11-15
Enhance vkRenderPass

YuqiaoZhang:
The InputAttachment should not be used as the ColorAttachment and thus the StoreOp can be set to DONT_CARE


YuqiaoZhang:
Sign CLA

Issue #4768, opened 2021-10-26
Validation error about vulkan image layout


wumo:
**Describe the bug**
Running sample gltf_viewer got vulkan validation error on version 1.12.9

**To Reproduce**
1. Compile and run gltf_viewer.

**Expected behavior**
Got validation error:

```shell
VULKAN ERROR: (UNASSIGNED-CoreValidation-DrawState-InvalidImageLayout) Validation Error: [ UNASSIGNED-CoreValidation-Dra
wState-InvalidImageLayout ] Object 0: handle = 0x1b4d3eebb88, type = VK_OBJECT_TYPE_COMMAND_BUFFER; | MessageID = 0x4dae
5635 | vkQueueSubmit(): pSubmits[0].pCommandBuffers[0] command buffer VkCommandBuffer 0x1b4d3eebb88[] expects VkImage 0x
54fef00000000a6[] (subresource: aspectMask 0x1 array layer 0, mip level 0) to be in layout VK_IMAGE_LAYOUT_SHADER_READ_O
NLY_OPTIMAL--instead, current layout is VK_IMAGE_LAYOUT_UNDEFINED.
```

**Desktop (please complete the following information):**
 - OS: Windows 10
 - GPU: NVIDIA RTX 3080
 - Backend: Vulkan (Using Vulkan SDK 1.2.189.2 for validation layer)
 - filament: 1.12.9

Additional Context:

This validation error won't happen before this [merge](https://github.com/google/filament/pull/4678).

prideout:
I reproduced this today with MoltenVK + debug build of `gltf_viewer`.

PR #4852, opened 2021-11-16
Fix khronos validation layer error for the hellopbr sample on Linux with Vulkan backend

YuqiaoZhang:
The newly created texture is LAYOUT_UNDEFINED, and thus an error will occur from the khronos validation layer if we blindly use the LAYOUT_SHADER_READ_ONLY_OPTIMAL.

VULKAN ERROR: (UNASSIGNED-CoreValidation-DrawState-InvalidImageLayout) Validation Error: [ UNASSIGNED-CoreValidation-DrawState-InvalidImageLayout ] Object 0: handle = 0x7f33dca4bae8, type = VK_OBJECT_TYPE_COMMAND_BUFFER; | MessageID = 0x4dae5635 | Submitted command buffer expects VkImage 0x200000000020[] (subresource: aspectMask 0x1 array layer 0, mip level 0) to be in layout VK_IMAGE_LAYOUT_SHADER_READ_ONLY_OPTIMAL--instead, current layout is VK_IMAGE_LAYOUT_UNDEFINED.

We may use LAYOUT_UNDEFINED if we can tell the update is not partial. Besides, at most time we perform the full update for an uninitialized texture and thus we bypass this bug.

### DIFF
--- a/filament/backend/src/vulkan/VulkanTexture.cpp
+++ b/filament/backend/src/vulkan/VulkanTexture.cpp
@@ -245,12 +245,14 @@ void VulkanTexture::updateWithCopyBuffer(const PixelBufferDescriptor& hostData, 
 
     const VkCommandBuffer cmdbuffer = mContext.commands->get().cmdbuffer;
 
-    // We can't blindly use LAYOUT_UNDEFINED because it may destroy the data, and because
-    // we're potentially updating only a sub-region it would be a problem.
+    // We may use LAYOUT_UNDEFINED if we can tell the update is not partial.
+    // Besides, the newly created texture is LAYOUT_UNDEFINED, and thus an error will occur from
+    // the khronos validation layer if we blindly use the LAYOUT_SHADER_READ_ONLY_OPTIMAL.
+    bool fullupdate = (width == this->width) && (height == this->height) && (depth == this->depth);
     VkImageLayout textureLayout = mContext.getTextureLayout(usage);
     transitionImageLayout(cmdbuffer, textureTransitionHelper({
             .image = mTextureImage,
-            .oldLayout = textureLayout,
+            .oldLayout = fullupdate ? VK_IMAGE_LAYOUT_UNDEFINED : textureLayout,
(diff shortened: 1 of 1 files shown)


filament/backend/src/vulkan/VulkanTexture.cpp
```diff
@@ -245,12 +245,14 @@ void VulkanTexture::updateWithCopyBuffer(const PixelBufferDescriptor& hostData,
...
-    // we're potentially updating only a sub-region it would be a problem.
+    // We may use LAYOUT_UNDEFINED if we can tell the update is not partial.
+    // Besides, the newly created texture is LAYOUT_UNDEFINED, and thus an error will occur from
+    // the khronos validation layer if we blindly use the LAYOUT_SHADER_READ_ONLY_OPTIMAL.
+    bool fullupdate = (width == this->width) && (height == this->height) && (depth == this->depth);
     VkImageLayout textureLayout = mContext.getTextureLayout(usage);
     transitionImageLayout(cmdbuffer, textureTransitionHelper({
             .image = mTextureImage,
-            .oldLayout = textureLayout,
+            .oldLayout = fullupdate ? VK_IMAGE_LAYOUT_UNDEFINED : textureLayout,
```
pixelflinger, line 255: It seems to me that the problem is that `textureLayout` is not initialized properly somewhere else.
This fix is more like a workaround.

pixelflinger (changes requested):

pixelflinger:
I think this would be a fix for #4768

filament/backend/src/vulkan/VulkanTexture.cpp
YuqiaoZhang: Actually, we need some additional information to indicate whether the texture is newly created. However, the additional information may make the program not efficient.

YuqiaoZhang: Yes. This patch doesn't totally fix this bug. However, this patch may boost performance since we use the LAYOUT_UNDEFINED.

prideout:
This was fixed with https://github.com/google/filament/pull/5052

PR #4858, opened 2021-11-16
Remove the feedback loop inside the renderpass

YuqiaoZhang:
The InputAttachment should not be used as the ColorAttachment, which may introduce the "feedback loop", which is harmful by the vkspec, and thus the StoreOp can be set to DONT_CARE.

The initalLayout can be set to LAYOUT_UNDEFINED if the LoadOp is CLEAR or DONT_CARE since it's not reasonable to use the image when the LoadOp is set to CLEAR or DONT_CARE.

The layout inside the renderpass can be deduced and thus we can remove some code.

### DIFF
--- a/filament/backend/src/vulkan/VulkanFboCache.cpp
+++ b/filament/backend/src/vulkan/VulkanFboCache.cpp
@@ -125,8 +125,21 @@ VkRenderPass VulkanFboCache::getRenderPass(RenderPassKey config) noexcept {
         iter.value().timestamp = mCurrentTime;
         return iter->second.handle;
     }
-    const bool isSwapChain = config.colorLayout[0] == VK_IMAGE_LAYOUT_PRESENT_SRC_KHR;
-    const bool hasSubpasses = config.subpassMask != 0;
+
+    // The Driver API consolidates all color attachments from the first and second subpasses
+    // into a single list, and uses a bitmask to mark attachments that belong only to the
+    // second subpass and should be available as inputs.
+
+    // The attachment list contains: Color Attachments, Resolve Attachments, and Depth Attachment.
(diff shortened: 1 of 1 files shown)


filament/backend/src/vulkan/VulkanFboCache.cpp
```diff
@@ -215,53 +198,54 @@ VkRenderPass VulkanFboCache::getRenderPass(RenderPassKey config) noexcept {
...
+                colorAttachmentRefs[1][index].attachment = attachmentIndex;
+            }
         }
 
         const TargetBufferFlags flag = TargetBufferFlags(int(TargetBufferFlags::COLOR0) << i);
         const bool clear = any(config.clear & flag);
-        const bool discard = any(config.discardStart & flag);
+        const bool discardStart = any(config.discardStart & flag);
+
+        // The mission here is to ensure that the initial layout matches the current layout of the image.
```
prideout, line 233: It might be better to track layout than to try to deduce it with difficult logic.

prideout (changes requested):
This change might be fine, but first I need a little more info.  Does this fix an actual validation error or warning?  Has this been tested while verifying that `colorGradingConfig.asSubpass` is enabled? Thanks!

prideout:
@YuqiaoZhang thanks for this fix, it looks good but I haven't had a chance to play with it yet.

I'm curious how you discovered this issue, did you come across a validation error?

prideout:
Closing because we need more information and the PR now has merge conflicts.  Feel free to re-open this PR against the latest main branch.  Be sure to mention which validation messages this fixes (or, cite the relevant text from the Vulkan specification).

PR #4863, opened 2021-11-17
Android: re-enable VSM cascade fix

bejado:
I don't understand why, but I'm able to reproduce the VSM/cascade issue on Android again. I want to bisect recent commits and see if I can track down where this started reproducing again, but for now I'm going to re-enable the fix.


romainguy (approved):
Mathias was messing with the shadowing code lately so maybe one of his commits?

PR #4885, opened 2021-11-23
Use locale-independent string->float conversion

### LINKED ISSUE #4883: loading shperical harmonics from *_ibl.ktx fails on systems locals using `,` instead of `.` for decimal separator
**Describe the bug**

When building my own 3d viewer in open3d which uses filament for rendering i observed that setting indirect lighting  by loading *_ibl.ktx file faild with inability to load spherical harmonics from *_ibl.ktx file. The file contains proper 3 x 9 table for spherical harmonics.

After investigation i figured that the reason for this locals sensitivity is the `std::strtof` function used on line `ilament/libs/image/src/KtxBundle.cpp:298` of current master branch to parse the spherical harmonics list. See also [std::strtof reference](https://en.cppreference.com/w/cpp/string/byte/strtof).

**To Reproduce**
1. set system locals to for example `de_DE` or `de_UTF8`
2. run your program which uses indirect lighting only
3. surfaces will be black only

4. Open editor (eg vi/vim)
5. replace all `.` in spherical harmonics list (key `sh`) with `,`
6. rerun demo
7. Surface colors will be back

**Expected behavior**

That `KtxBundle::getSphericalHarmonics` returns valid spherical harmonics list independent of which locals are set on the system filament or filament based library is used. 

**Screenshots**
`std::strtof` failing on eg.: German locale 
std_strtof_failing_on_locale
with `,` as decimal separator eg: in German locale
with_as_decimal_separator

**Logs**

 - OS: [e.g. iOS] Mint 20 i386 
 - GPU: [e.g. NVIDIA GTX 1080] irrelevant
 - Backend: [OpenGL/Vulkan] guess opengl through intel/open3d python wheel

 - Device: [e.g. Pixel 2] 
 - OS: [e.g. Android Pie 9.0]

**Additional context**

romainguy:
strtof and friends are locale aware and won't parse decimal numbers
with a period ("12.6" for instance) in locales that use another
character for the decimal period ("," in French for instance).

This change introduces a new function called strtof_c that forces
the use of a specific locale (called "C") to make sure we always
parse floats in the desired "C" format ("12.6").

With C++17 we should be able to use std::from_chars but this API
is not implemented in clang for floats at the moment.

Fixes #4883.

### DIFF
--- a/libs/image/src/KtxBundle.cpp
+++ b/libs/image/src/KtxBundle.cpp
@@ -17,10 +17,11 @@
 #include <image/KtxBundle.h>
 
 #include <utils/Panic.h>
+#include <utils/string.h>
 
 #include <string>
-#include <vector>
 #include <unordered_map>
+#include <vector>
 
 namespace {
 
@@ -295,7 +296,7 @@ bool KtxBundle::getSphericalHarmonics(filament::math::float3* result) {
     // 3 bands, 9 RGB coefficients for a total of 27 floats.
     for (int i = 0; i < 9 * 3; i++) {
         char* next;
-        *flat++ = std::strtof(src, &next);
+        *flat++ = utils::strtof_c(src, &next);
         if (next == src) {
             return false;
         }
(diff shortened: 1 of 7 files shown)


bejado (approved):
LGTM after Linux failure is fixed

PR #4902, opened 2021-11-29
Cache calls to Enum.values()

romainguy:
Each call to .values() allocates a new array. Let's cache them once
and for all to avoid unncessary allocations.

### DIFF
--- a/android/filament-android/src/main/java/com/google/android/filament/Material.java
+++ b/android/filament-android/src/main/java/com/google/android/filament/Material.java
@@ -37,6 +37,20 @@ import java.util.Set;
  * @see <a href="https://google.github.io/filament/Materials.html">Filament Materials Guide</a>
  */
 public class Material {
+    static final class EnumCache {
+        private EnumCache() { }
+
+        static final Shading[] sShadingValues = Shading.values();
+        static final Interpolation[] sInterpolationValues = Interpolation.values();
+        static final BlendingMode[] sBlendingModeValues = BlendingMode.values();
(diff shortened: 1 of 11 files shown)


android/filament-android/src/main/java/com/google/android/filament/Material.java
```diff
@@ -37,6 +37,20 @@
...
+
+        static final Shading[] sShadingValues = Shading.values();
+        static final Interpolation[] sInterpolationValues = Interpolation.values();
+        static final BlendingMode[] sBlendingModeValues = BlendingMode.values();
+        static final RefractionMode[] sRefractionModeValues = RefractionMode.values();
+        static final RefractionType[] sRefractionTypeValues = RefractionType.values();
+        static final VertexDomain[] sVertexDomainValues = VertexDomain.values();
+        static final CullingMode[] sCullingModeValues = CullingMode.values();
+        static final VertexBuffer.VertexAttribute[] sVertexAttributeValues =
+                VertexBuffer.VertexAttribute.values();
```
pixelflinger, line 51: I wouldn't use the `s` prefix here, just because they're accessed through `EnumCache`. It looks slightly better in the code.

pixelflinger (approved):

android/filament-android/src/main/java/com/google/android/filament/Material.java
romainguy: Sure but that'd be a weird one off change to our style. I don't care either way.

PR #4907, opened 2021-11-30
Pin web docs Filament version

bejado:


romainguy (approved):

romainguy:
Should/can we delete filament.js  and filament.wasm if we always pull from npm?

bejado:
> Should/can we delete filament.js and filament.wasm if we always pull from npm?

Only the tutorials pull from npm, the other demos use filament.js/wasm.

PR #4919, opened 2021-12-02
Expose Engine::destroy(SkinningBuffer*)

MasTraER:
This commit enables Engine::destroy(SkinningBuffer*), which is already implemented internally in FEngine.

I implement a shared animator class for multiple renderables utilizing SkinningBuffer instances. I found FEngine already has destroy function for FSkinningBuffer, but there is no trampoline somehow.


romainguy (approved):

romainguy:
Thanks!

Note to self: we need to expose this in JNI as well.

PR #4925, opened 2021-12-03
Add support for basic PCSS

pixelflinger:


romainguy (approved):
You need to expose the new shadow techniques in Java too

pixelflinger:
> You need to expose the new shadow techniques in Java too

bejado (approved):

PR #4928, opened 2021-12-03
Fade lights out when close to light far plane

pixelflinger:
the fade takes place over 10% of the light far distance.

### DIFF
--- a/shaders/src/light_punctual.fs
+++ b/shaders/src/light_punctual.fs
@@ -106,8 +106,12 @@ float getSquareFalloffAttenuation(float distanceSquare, float falloff) {
 float getDistanceAttenuation(const highp vec3 posToLight, float falloff) {
     float distanceSquare = dot(posToLight, posToLight);
     float attenuation = getSquareFalloffAttenuation(distanceSquare, falloff);
+    // light far attenuation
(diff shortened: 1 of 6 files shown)


shaders/src/light_punctual.fs
```diff
@@ -106,8 +106,11 @@ float getSquareFalloffAttenuation(float distanceSquare, float falloff) {
 float getDistanceAttenuation(const highp vec3 posToLight, float falloff) {
     float distanceSquare = dot(posToLight, posToLight);
     float attenuation = getSquareFalloffAttenuation(distanceSquare, falloff);
+    // light far attenuation
+    float d = length(getWorldPosition() - getWorldCameraPosition());
```
romainguy, line 110: We already have the result of that subtraction, it's `posToLight`. Also we have the square distance above, just sqrt() it instead of calling length (or better, work in square distance if you can make it work?)

romainguy (changes requested):

shaders/src/light_punctual.fs
pixelflinger: No, `posToLight` is not the same thing -- it's the position of the fragment to the light (`positionFalloff.xyz - worldPosition`). I need the distance from the camera to the fragment.

PR #4932, opened 2021-12-08
Rework color history in preparation for SSR

bejado:
Call `import` and `detach` outside of TAA, since SSR will also need color history.

### DIFF
--- a/filament/src/Renderer.cpp
+++ b/filament/src/Renderer.cpp
@@ -588,16 +588,18 @@ void FRenderer::renderJob(ArenaScope& arena, FView& view) {
 
     // TAA for color pass
     if (taaOptions.enabled) {
-        input = ppm.taa(fg, input, view.getFrameHistory(), getColorHistory(fg, view), taaOptions,
-                colorGradingConfig);
+        input = ppm.taa(fg, input, view.getFrameHistory(),
+                getColorHistory(fg, view.getFrameHistory()), taaOptions, colorGradingConfig);
     }
 
-    struct SaveColorHistoryData {
+    struct ExportColorHistoryData {
         FrameGraphId<FrameGraphTexture> color;
     };
     if (taaOptions.enabled) {
-        fg.addPass<SaveColorHistoryData>("Save color history", [&](FrameGraph::Builder& builder,
+        fg.addPass<ExportColorHistoryData>("Export color history", [&](FrameGraph::Builder& builder,
                 auto& data) {
+            // We need to use sideEffect here to ensure this pass won't be culled. The "output" of
+            // this pass is going to be used during the next frame as an "import".
             builder.sideEffect();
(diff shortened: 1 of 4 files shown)


filament/src/Renderer.cpp
```diff
@@ -1247,6 +1267,24 @@ void FRenderer::initializeClearFlags() {
...
+    if (auto history = blackboard.get<FrameGraphTexture>("colorHistory")) {
+        return history;
+    }
+
+    const FrameHistory& frameHistory = view.getFrameHistory();
+    FrameHistoryEntry const& entry = frameHistory[0];
+    if (UTILS_UNLIKELY(!entry.color.handle)) {
+        return {};
+    }
+    FrameGraphId<FrameGraphTexture> history = fg.import("Color history", entry.colorDesc,
```
pixelflinger, line 1282: rename to `colorHistory`

filament/src/Renderer.cpp
```diff
@@ -1247,6 +1267,24 @@ void FRenderer::initializeClearFlags() {
             | TargetBufferFlags::DEPTH_AND_STENCIL;
 }
 
+FrameGraphId<FrameGraphTexture> FRenderer::getColorHistory(FrameGraph& fg, FView& view) noexcept {
```
pixelflinger, line 1270: pass the `FrameHistory const&` instead of the `FView&`.

filament/src/Renderer.cpp
```diff
@@ -588,7 +588,23 @@ void FRenderer::renderJob(ArenaScope& arena, FView& view) {
...
+                colorGradingConfig);
+    }
+
+    struct SaveColorHistoryData {
+        FrameGraphId<FrameGraphTexture> color;
+    };
+    if (taaOptions.enabled) {
+        fg.addPass<SaveColorHistoryData>("Save color history", [&](FrameGraph::Builder& builder,
+                auto& data) {
+            builder.sideEffect();
```
pixelflinger, line 601: explain that we need to use a `sideEffect` here (i.e.: pass won't be culled) because the "output" of this pass is going to be used during the next frame as an "import".

pixelflinger (changes requested):
main structure is good, just needs more comments and minor other things.

filament/src/Renderer.cpp
```diff
@@ -675,6 +691,10 @@ void FRenderer::renderJob(ArenaScope& arena, FView& view) {
 
     fg.execute(driver);
 
+    if (auto history = blackboard.get<FrameGraphTexture>("colorHistory")) {
+        blackboard.remove("colorHistory");
+    }
+
```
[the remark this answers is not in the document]
bejado: Ah yes, good point.

pixelflinger (approved):

PR #4939, opened 2021-12-10
Fix skinning calculations

### LINKED ISSUE #4887: Weird distortion when rendering model with skinned animation
**Describe the bug**
I have a model with skeletal animation that renders weirdly distorted in filament.

**To Reproduce**
- Unzip [Suzanne_glb.zip](https://github.com/google/filament/files/7595692/Suzanne_glb.zip) 
- Render Suzanne in gltf_viewer sample (v1.14.1)

**Expected behavior**
Suzanne should not be distorted.

**Screenshots**
### [Babylon sandbox](https://sandbox.babylonjs.com)
babylon

### [glTF sample viewer](https://github.khronos.org/glTF-Sample-Viewer-Release/)
gltf_sample_viewer

### [ModelViewer](https://modelviewer.dev/editor/)
modelviewer

### Filament (v1.14.1)
filament

pixelflinger:
it was incorrect to store bone matrices as quaternions because it is
allowed for these matrices to have a skew component.

We now store the 4x4 matrix -- which is the same amount of data as 
before, however we compute the cofactor matrix in the vertex shader when
transforming normals.

fixes #4887


pixelflinger:
This is a draft because I need to run more tests and also because I'm going improve performance before actual comit. But this new version should be functional.

shaders/src/getters.vs
```diff
@@ -22,44 +22,31 @@ mat3 getWorldFromModelNormalMatrix() {
...
-    // apply the rigid transform (valid only for unit quaternions)
-    n += 2.0 * cross(q.xyz, cross(q.xyz, n) + q.w * n);
-
-    return n;
+    highp mat3 transform = mat3(
+        bonesUniforms.transform[i][0].xyz,
+        bonesUniforms.transform[i][1].xyz,
+        bonesUniforms.transform[i][2].xyz
+    );
+    return normalize(cofactor(transform) * n);
```
romainguy, line 30: Should we even normalize here if we normalize in the fragment shader?

romainguy (approved):

shaders/src/getters.vs
pixelflinger: yes, unfortunately, because of that:
```
void skinNormal(inout vec3 n, const uvec4 ids, const vec4 weights) {
    n =   mulBoneNormal(n, ids.x) * weights.x
        + mulBoneNormal(n, ids.y) * weights.y
        + mulBoneNormal(n, ids.z) * weights.z
        + mulBoneNormal(n, ids.w) * weights.w;
}
```

shaders/src/common_types.fs
```diff
@@ -34,3 +34,8 @@ struct ShadowData {
     float bulbRadiusLs;
     float nearOverFarMinusNear;
 };
+
+struct BoneData {
+    highp mat3x4 transform;    // bone transform is mat4x3 stored in row-major (last row [0,0,0,1])
```
romainguy, line 39: Won't we run into the adreno bug with highps inside struct being ignored?

shaders/src/common_types.fs
```diff
@@ -34,3 +34,8 @@ struct ShadowData {
...
+    highp uvec4 cof;           // 8 first cofactor matrix of transform's upper left
```
romainguy, line 40: Could these be floats to avoid int<>float conversions on the GPU and avoiding the int pipeline?

romainguy (approved):

shaders/src/common_types.fs
pixelflinger: I'm thinking not because it's in an UBO?

shaders/src/common_types.fs
pixelflinger: no because we can't store fp16 floats in ubos. we can only pack two fp16 into one `int`.

shaders/src/common_types.fs
romainguy: Ah right yeah that's fine.

shaders/src/common_types.fs
romainguy: Ok makes sense

PR #4946, opened 2021-12-10
Add initial glossy screen-space reflections

bejado:
Currently only surfaces with near-zero roughness are eligible for SSR.

By default SSR are off. To enable, turn them on in materials via the new `reflections: screenspace` material property and set `ScreenSpaceReflectionsOptions::enabled` to `true`.

Future improvements:
- Java/JavaScript bindings
- non-glossy reflections
- support for simultaneous screen-space reflections/refractions

https://user-images.githubusercontent.com/5298046/148319050-5b620050-5e22-4ac5-94bb-59ead7541c79.mp4


romainguy:
"non-glossy" is confusing, to me it means "rough". Which is the opposite of this PR ^^

RELEASE_NOTES.md
```diff
@@ -6,6 +6,8 @@ A new header is inserted each time a *tag* is created.
 ## v1.15.2 (currently main branch)
 
 - engine: add support for PCSS (Percentage Closer Soft Shadows)
+- engine: add experimental non-glossy screen-space reflections [⚠️ **Materials need to be rebuilt to
```
romainguy: You should sync, Mathias just revved materials for v1.16

filament/include/filament/Material.h
```diff
@@ -193,9 +193,12 @@ class UTILS_PUBLIC Material : public FilamentAPI {
     //! Returns the refraction mode used by this material.
     RefractionMode getRefractionMode() const noexcept;
 
-    // Return the refraction type used by this material.
+    //! Return the refraction type used by this material.
     RefractionType getRefractionType() const noexcept;
 
+    //! Returns the reflections mode used by this material.
+    ReflectionsMode getReflectionsMode() const noexcept;
```
romainguy, line 200: Should we use the singular to match `RefractionMode`?

libs/filamat/src/UibGenerator.cpp
```diff
@@ -117,8 +117,15 @@ UniformInterfaceBlock const& UibGenerator::getPerViewUib() noexcept  {
             .add("nearOverFarMinusNear",    1, UniformInterfaceBlock::Type::FLOAT, Precision::HIGH)
             .add("temporalNoise",           1, UniformInterfaceBlock::Type::FLOAT, Precision::HIGH)
 
+            // Screen-space reflection parameters
+            .add("ssrReprojection",         1, UniformInterfaceBlock::Type::MAT4, Precision::HIGH)
+            .add("ssrThickness",            1, UniformInterfaceBlock::Type::FLOAT)
+            .add("ssrBias",                 1, UniformInterfaceBlock::Type::FLOAT)
+            .add("ssrDistance",             1, UniformInterfaceBlock::Type::FLOAT)
+            .add("ssrEnabled",              1, UniformInterfaceBlock::Type::UINT)
```
romainguy, line 125: Could we skip this uniform and just use ssrDistance > 0 for instance?

bejado:
> "non-glossy" is confusing, to me it means "rough". Which is the opposite of this PR ^^

Yeah, you're absolutely right 🤦

shaders/src/light_indirect.fs
```diff
@@ -539,20 +539,27 @@ void applyRefraction(
...
-    // TODO: Remove this factor temporarily until we find a better solution
-    //       This overblurs many scenes and needs a more principled approach
-    // float tweakedPerceptualRoughness = perceptualRoughness * 1.74;
-    float tweakedPerceptualRoughness = perceptualRoughness;
-    float lod = max(0.0, 2.0 * log2(tweakedPerceptualRoughness) + frameUniforms.refractionLodOffset);
-
-    vec3 Ft = textureLod(light_ssr, p.xy, lod).rgb;
+    vec3 Ft;
+    if (frameUniforms.ssrEnabled > 0u) {
+        // if screen-space reflections is enabled, we can't read from the screen-space buffer, so
```
romainguy, line 544: Interesting, why can't we do both at the same time?

shaders/src/light_indirect.fs
```diff
@@ -595,6 +602,16 @@ void evaluateIBL(const MaterialInputs material, const PixelParams pixel, inout v
     Fr = isEvaluateSpecularIBL(pixel, shading_normal, shading_view, shading_NoV);
 #endif
 
+#if defined(HAS_REFLECTIONS) && REFLECTIONS_MODE == REFLECTIONS_MODE_SCREEN_SPACE
+    // Up to this point Fr is the Cubemap reflection only.
+    // evaluateScreenSpaceReflections will replace the value of Fr if there's a hit.
+    // TODO: do we want iblLuminance to control screen-space reflections?
+    if (pixel.roughness <= 0.01f && frameUniforms.ssrEnabled > 0u) {
```
romainguy, line 609: Should we optimize by only evaluating the IBL when there's no hit instead?

shaders/src/light_reflections.fs
```diff
@@ -0,0 +1,269 @@
...
+//------------------------------------------------------------------------------
+
+#if defined(HAS_REFLECTIONS) && REFLECTIONS_MODE == REFLECTIONS_MODE_SCREEN_SPACE
+
+// Copied from depthUtils.fs
+highp float linearizeDepth(highp float depth) {
+    // Our far plane is at infinity, which causes a division by zero below, which in turn
+    // causes some issues on some GPU. We workaround it by replacing "infinity" by the closest
+    // value representable in  a 24 bit depth buffer.
+    const float preventDiv0 = 1.0 / 16777216.0;
```
romainguy, line 12: This needs to be marked `highp`

shaders/src/light_reflections.fs
```diff
@@ -0,0 +1,269 @@
...
+    mat4 p = getViewFromClipMatrix();
```
romainguy, line 13: highp maybe?

shaders/src/light_reflections.fs
```diff
@@ -0,0 +1,269 @@
...
+    highp vec3 vsOrigin = vsRayStart.xyz;
+    highp vec3 vsDirection = vsRayDirection.xyz;
+    highp float vsZThickness = frameUniforms.ssrThickness;
+    // TODO: use the actual near plane.
+    highp float nearPlaneZ = -0.1f;
+    highp float stride = 1.0;
+    // TODO: jitterFraction should be a between 0 and 1, but anything < 1 gives banding artifacts.
+    highp float jitterFraction = 1.0f;
+    // TODO: set this to the larger of the viewport dimensions.
+    highp float maxSteps = 1000.0;
```
romainguy, line 233: This seems way too high. Should it be user parameterized as well?
And do all those values need to be `highp`?

shaders/src/light_indirect.fs
bejado: Mathias and I discussed this a bit, we think we can but it's a bit more involved and will likely require some frame graph changes.

RELEASE_NOTES.md
bejado: That PR hasn't been merged yet, but I'll update this after he does.

shaders/src/light_indirect.fs
bejado: That's a good idea for now, but I think with blurry reflections we'll want to sample both anyway. I'm not sure yet.

libs/filamat/src/UibGenerator.cpp
bejado: I think so, yes. Good idea.

shaders/src/light_reflections.fs
```diff
@@ -0,0 +1,276 @@
...
+
+float distanceSquared(vec2 a, vec2 b) {
+    a -= b;
+    return dot(a, a);
+}
+
+// Note: McGuire and Mara use the "cs" prefix to stand for "camera space", equivalent to Filament's
+// "view space". "cs" has been replaced with "vs" to avoid confusion.
+bool traceScreenSpaceRay(Point3 vsOrigin, Vector3 vsDirection, mat4x4 projectToPixelMatrix,
+        sampler2D vsZBuffer, float vsZThickness, float nearPlaneZ, float stride,
```
pixelflinger, line 67: I think you have to specify `mediump` here

shaders/src/light_reflections.fs
```diff
@@ -0,0 +1,276 @@
...
+}
+
+float distanceSquared(vec2 a, vec2 b) {
+    a -= b;
+    return dot(a, a);
+}
+
+// Note: McGuire and Mara use the "cs" prefix to stand for "camera space", equivalent to Filament's
+// "view space". "cs" has been replaced with "vs" to avoid confusion.
+bool traceScreenSpaceRay(Point3 vsOrigin, Vector3 vsDirection, mat4x4 projectToPixelMatrix,
```
pixelflinger, line 66: mark all constant params as `const`.
Also I think most values in this function should be `highp`, they're basically "vertices". Only "colors" can be`mediump`

shaders/src/light_reflections.fs
```diff
@@ -0,0 +1,276 @@
...
+
+#undef Point3
+#undef Point2
+#undef Vector2
+#undef Vector3
+#undef Vector4
+
+// -- end "BSD 2-clause license" -------------------------------------------------------------------
+
+mat4x4 scaleMatrix(float x, float y) {
```
pixelflinger, line 195: From here and below, a lot of `highp` are missing. Also use `mat4` instead of `mat4x4`.

shaders/src/light_reflections.fs
pixelflinger: @romainguy no, I think this is just the actual maximum to can possibly do before it exists the viewport. It's not the number of step that will actually be used (I think).

shaders/src/light_reflections.fs
```diff
@@ -0,0 +1,276 @@
...
+            maxRayTraceDistance, hitPixel, vsHitPoint)) {
+        vec2 ssrRes = vec2(textureSize(light_ssr, 0).xy);
+        vec4 reprojected = scaleMatrix(ssrRes.x, ssrRes.y) *
+                translateMatrix(0.5f, 0.5f) *
+#if defined(TARGET_METAL_ENVIRONMENT) || defined(TARGET_VULKAN_ENVIRONMENT)
+                scaleMatrix(0.5f, -0.5f) *
+#else
+                scaleMatrix(0.5f, 0.5f) *
+#endif
+                frameUniforms.ssrReprojection *
```
pixelflinger, line 267: This can be in a uniform. And even if not, it's almost the same computation than above.

shaders/src/light_reflections.fs
```diff
@@ -0,0 +1,276 @@
...
+    vec2 res = vec2(textureSize(light_structure, 0).xy);
+    highp mat4x4 projectToPixelMatrix =
+        scaleMatrix(res.x, res.y) *
+        translateMatrix(0.5f, 0.5f) *
+#if defined(TARGET_METAL_ENVIRONMENT) || defined(TARGET_VULKAN_ENVIRONMENT)
+        scaleMatrix(0.5f, -0.5f) *  // Metal and Vulkan's texture space is y-inverted
+#else
+        scaleMatrix(0.5f, 0.5f) *
+#endif
+        getClipFromViewMatrix();
```
pixelflinger, line 250: This can be passed in a unifrom

pixelflinger (changes requested):

pixelflinger (changes requested):
This PR definitely needs a screenshot!

shaders/src/light_reflections.fs
```diff
@@ -0,0 +1,276 @@
...
+        // ensure correct ordering
+        rayZMin = prevZMaxEstimate;
+
+        // Compute the value at 1/2 pixel into the future
+        rayZMax = (dQ.z * 0.5 + Q.z) / (dk * 0.5 + k);
+        prevZMaxEstimate = rayZMax;
+        if (rayZMin > rayZMax) { swap(rayZMin, rayZMax); }
+
+        // View-space z of the background
+        sceneZMax = linearizeDepth(texelFetch(vsZBuffer, int2(hitPixel), 0).r);
```
pixelflinger, line 178: We need to explore if we could avoid the linearization here (future TODO) -- `sceneMax` is only used to check if we have a "hit", it's just a compare with the "z" of the ray, taking the thickness into account, those values could be "unlinearized" before the loop.  I'm not sure if it's feasible with `vsZThickness` though.

shaders/src/light_reflections.fs
```diff
@@ -0,0 +1,276 @@
...
+                translateMatrix(0.5f, 0.5f) *
+#if defined(TARGET_METAL_ENVIRONMENT) || defined(TARGET_VULKAN_ENVIRONMENT)
+                scaleMatrix(0.5f, -0.5f) *
+#else
+                scaleMatrix(0.5f, 0.5f) *
+#endif
+                frameUniforms.ssrReprojection *
+                vec4(vsHitPoint, 1.0f);
+        reprojected *= (1.0 / reprojected.w);
+        Fr = texelFetch(light_ssr, int2(reprojected.x, reprojected.y), 0).rgb;
```
pixelflinger, line 270: We should fade the contribution of the reflection near the edges, to avoid a sharp cutoff (See screenshot).
Screen Shot 2021-12-12 at 11 15 52 PM

We already do something like that for the contact shadows, this is the code we use there:
```
    vec2 fade = max(12.0 * abs(ray.xy - 0.5) - 5.0, 0.0);
    occlusion *= saturate(1.0 - dot(fade, fade));
```
This can be adapted here, using a mix().

We should also probably fade based on the max ray-length vs hitpoint.

pixelflinger: Why use `texelFetch` here? This prevents using bilinear filtering and causes issues at the edges. Use `textureLod` instead, the only modification to the code is to not scale by `ssrRes` when calculating `reprojected` above.
This will also help with the fading mentioned above.

shaders/src/light_reflections.fs
bejado: I think `vsZBuffer` should be `highp` (`light_structure` is `highp`), but yes I need to mark it as such.

shaders/src/light_reflections.fs
bejado: Most of it I can fold into a uniform, but I can't seem to find an easy way to get the dimensions of the structure buffer on the CPU side.

shaders/src/light_reflections.fs
bejado: Right, this is just a fail-safe to avoid tracing further than necessary. The equivalent user param is `ssrDistance`. And yes, some of these values don't need to be `highp`.

pixelflinger (approved):

shaders/src/light_indirect.fs
```diff
@@ -539,20 +539,27 @@ void applyRefraction(
...
-    // cubemap and screen-space modes match at perceptualRoughness 0.125
-    // TODO: Remove this factor temporarily until we find a better solution
-    //       This overblurs many scenes and needs a more principled approach
-    // float tweakedPerceptualRoughness = perceptualRoughness * 1.74;
-    float tweakedPerceptualRoughness = perceptualRoughness;
-    float lod = max(0.0, 2.0 * log2(tweakedPerceptualRoughness) + frameUniforms.refractionLodOffset);
-
-    vec3 Ft = textureLod(light_ssr, p.xy, lod).rgb;
+    vec3 Ft;
+    if (frameUniforms.ssrDistance > 0.0f) {
```
pixelflinger, line 543: This test shouldn't even be there if we're not in `REFLECTION_MODE_SCREEN_SPACE` right?

shaders/src/light_reflections.fs
```diff
@@ -0,0 +1,254 @@
...
+        out highp vec2 hitPixel, out highp vec3 vsHitPoint) {
+    // Clip ray to a near plane in 3D (doesn't have to be *the* near plane, although that would be a
+    // good idea)
+    highp float rayLength = ((vsOrigin.z + vsDirection.z * maxRayTraceDistance) > nearPlaneZ) ?
+            (nearPlaneZ - vsOrigin.z) / vsDirection.z : maxRayTraceDistance;
+    highp vec3 vsEndPoint = vsDirection * rayLength + vsOrigin;
+
+    // Project into screen space
+    highp vec4 H0 = projectToPixelMatrix * vec4(vsOrigin, 1.0);
+    highp vec4 H1 = projectToPixelMatrix * vec4(vsEndPoint, 1.0);
```
pixelflinger, line 73: you could use our `mulMat4x4Float3` for these

shaders/src/light_reflections.fs
```diff
@@ -0,0 +1,254 @@
...
+ *
+ * If there is no hit, Fr is unmodified.
+ */
+void evaluateScreenSpaceReflections(vec3 r, inout vec4 Fr) {
+    vec3 wsRayDirection = r;
+    highp vec3 wsRayStart = shading_position + frameUniforms.ssrBias * wsRayDirection;
+
+    // ray start/end in view space
+    highp vec4 vsRayStart = getViewFromWorldMatrix() * vec4(wsRayStart, 1.0);
+    highp vec4 vsRayDirection = getViewFromWorldMatrix() * vec4(wsRayDirection, 0.0);
```
pixelflinger, line 208: you could use `mulMat4x4Float3` for those

shaders/src/light_reflections.fs
```diff
@@ -0,0 +1,254 @@
...
+
+    highp vec3 vsOrigin = vsRayStart.xyz;
+    highp vec3 vsDirection = vsRayDirection.xyz;
+    float vsZThickness = frameUniforms.ssrThickness;
+    // TODO: use the actual near plane.
+    const float nearPlaneZ = -0.1f;
```
pixelflinger, line 214: You can get `nearPlaneZ` as:
```
    nearPlaneZ = -frameUniforms.nearOverFarMinusNear / frameUniforms.oneOverFarMinusNear;
```

shaders/src/light_reflections.fs
```diff
@@ -0,0 +1,254 @@
...
+    const float stride = 1.0;
+    // TODO: jitterFraction should be between 0 and 1, but anything < 1 gives banding artifacts.
+    const float jitterFraction = 1.0f;
+    float maxRayTraceDistance = frameUniforms.ssrDistance;
+
+    highp vec2 res = vec2(textureSize(light_structure, 0).xy);
+    highp mat4x4 projectToPixelMatrix =
```
pixelflinger, line 221:
```suggestion
    highp mat4 projectToPixelMatrix =
```

shaders/src/light_reflections.fs
```diff
@@ -0,0 +1,254 @@
...
+        // Fade the reflections out near the edges.
+        vec2 edgeFactor = max(fadeRate * abs(reprojected.xy - 0.5f) - (fadeRate / 2.0f - 1.0f), 0.0f);
+        float edgeFade = saturate(1.0 - dot(edgeFactor, edgeFactor));
+
+        // Fade the reflections out near maxRayTraceDistance.
+        float t = distance(vsOrigin, vsHitPoint) / maxRayTraceDistance;
+        float distFactor = max(fadeRate * (t - 0.5f) - (fadeRate / 2.0f - 1.0f), 0.0f);
+        float distanceFade = saturate(1.0 - (distFactor, distFactor));
+
+        Fr = vec4(textureLod(light_ssr, reprojected.xy, 0.0f).rgb, edgeFade * distanceFade);
```
pixelflinger, line 250: Usually we use premultiplied alpha everywhere.  Maybe it would be better to stick to that convention. This would just change the `mix` at the call site.

shaders/src/light_reflections.fs
```diff
@@ -0,0 +1,254 @@
...
+    // ray start/end in view space
+    highp vec4 vsRayStart = getViewFromWorldMatrix() * vec4(wsRayStart, 1.0);
+    highp vec4 vsRayDirection = getViewFromWorldMatrix() * vec4(wsRayDirection, 0.0);
+
+    highp vec3 vsOrigin = vsRayStart.xyz;
+    highp vec3 vsDirection = vsRayDirection.xyz;
+    float vsZThickness = frameUniforms.ssrThickness;
+    // TODO: use the actual near plane.
+    const float nearPlaneZ = -0.1f;
+    const float stride = 1.0;
```
pixelflinger, line 215: make stride a parameter (at least have a TODO here about it)

shaders/src/light_reflections.fs
```diff
@@ -0,0 +1,254 @@
...
+}
+
+highp float distanceSquared(highp vec2 a, highp vec2 b) {
+    a -= b;
+    return dot(a, a);
+}
+
+// Note: McGuire and Mara use the "cs" prefix to stand for "camera space", equivalent to Filament's
+// "view space". "cs" has been replaced with "vs" to avoid confusion.
+bool traceScreenSpaceRay(const highp vec3 vsOrigin, const highp vec3 vsDirection,
```
pixelflinger, line 60: We talked about  moving the raycast code into its own file so it can be used for other things eventually. Ok to do in a later PR.

pixelflinger (changes requested):

pixelflinger (approved):

bejado:
Looking into an OpenGL issue, then will merge.

PR #4973, opened 2021-12-15
fix aabb misalignment for skinned gltf model

Ankbzpx:
For a skinned gltf model, its mesh primitives world transform may not be the same as its bind transform. This pr aims to recompute aabb based on bind transform to fix misalignment issue.

| | Suzanne     | torus | 
|----------|-----------|----------|
| Before | Suzanne_skinning_before   | torus_before |
| After | Suzanne_skinning_after  | torus_after |

Test models: [test_models_glb.zip](https://github.com/google/filament/files/7717362/test_models_glb.zip)

### DIFF
--- a/libs/gltfio/include/gltfio/ResourceLoader.h
+++ b/libs/gltfio/include/gltfio/ResourceLoader.h
@@ -53,8 +53,8 @@ struct ResourceConfiguration {
     //! do not need this, but it is useful for robustness.
     bool recomputeBoundingBoxes;
 
-    //! If true, ignore bind transform for skinned primitives when compute bounding box. Only 
-    //! applicable when recomputeBoundingBoxes is set to true
+    //! If true, ignore bind transform for skinned primitives when compute bounding box. Implicitly true 
+    //! for instanced asset. Only applicable when recomputeBoundingBoxes is set to true
     bool ignoreBindTransform;
 };
 
--- a/libs/gltfio/include/gltfio/ResourceLoader.h
+++ b/libs/gltfio/include/gltfio/ResourceLoader.h
@@ -151,7 +151,7 @@ private:
     bool loadResources(FFilamentAsset* asset, bool async);
     void applySparseData(FFilamentAsset* asset) const;
     void normalizeSkinningWeights(FFilamentAsset* asset) const;
-    void updateBoundingBoxes(FFilamentAsset* asset) const;
+    void updateBoundingBoxes(FFilamentAsset* asset, void* cgltfSkinBaseAddress) const;
     AssetPool* mPool;
     struct Impl;
     Impl* pImpl;
--- a/libs/gltfio/include/gltfio/ResourceLoader.h
+++ b/libs/gltfio/include/gltfio/ResourceLoader.h
@@ -53,7 +53,7 @@ struct ResourceConfiguration {
     //! do not need this, but it is useful for robustness.
     bool recomputeBoundingBoxes;
 
-    //! If true, ignore bind transform for skinned primitives when compute bounding box. Implicitly true 
+    //! If true, ignore skinned primitives bind transform when compute bounding box. Implicitly true 
     //! for instanced asset. Only applicable when recomputeBoundingBoxes is set to true
     bool ignoreBindTransform;
 };
--- a/libs/gltfio/include/gltfio/ResourceLoader.h
+++ b/libs/gltfio/include/gltfio/ResourceLoader.h
@@ -52,6 +52,10 @@ struct ResourceConfiguration {
     //! If true, computes the bounding boxes of all \c POSITION attibutes. Well formed glTF files
     //! do not need this, but it is useful for robustness.
     bool recomputeBoundingBoxes;
+
+    //! If true, ignore bind transform for skinned primitives when compute bounding box. Only 
+    //! applicable when recomputeBoundingBoxes is set to true
+    bool ignoreBindTransform;
 };
 
 /**
@@ -151,7 +155,7 @@ private:
     bool loadResources(FFilamentAsset* asset, bool async);
     void applySparseData(FFilamentAsset* asset) const;
     void normalizeSkinningWeights(FFilamentAsset* asset) const;
-    void updateBoundingBoxes(FFilamentAsset* asset, void* cgltfSkinBaseAddress) const;
+    void updateBoundingBoxes(FFilamentAsset* asset) const;
     AssetPool* mPool;
     struct Impl;
     Impl* pImpl;
--- a/libs/gltfio/src/ResourceLoader.cpp
+++ b/libs/gltfio/src/ResourceLoader.cpp
@@ -1020,8 +1020,7 @@ void ResourceLoader::updateBoundingBoxes(FFilamentAsset* asset) const {
     const size_t skinningAttrSize = cgltf_num_components(cgltf_type_vec4);
     auto computeBoundingBoxSkinned = [&](const cgltf_primitive* prim, const Skin* skin, Aabb* result) {
         Aabb aabb;
-        std::vector<mat4f> inverseGlobalTransforms;
-        inverseGlobalTransforms.resize(skin->targets.size());
+        std::vector<mat4f> inverseGlobalTransforms(skin->targets.size());
         for (size_t i = 0; i < skin->targets.size(); i++) {
             auto xformable = tm.getInstance(skin->targets[i]);
             if (xformable) {
--- a/libs/gltfio/src/ResourceLoader.cpp
+++ b/libs/gltfio/src/ResourceLoader.cpp
@@ -465,7 +465,9 @@ bool ResourceLoader::loadResources(FFilamentAsset* asset, bool async) {
 
     if (pImpl->mRecomputeBoundingBoxes) {
         // asset->mSkins is unused for instanced assets
-        pImpl->mIgnoreBindTransform = asset->isInstanced();
+        if (!pImpl->mIgnoreBindTransform) {
+            pImpl->mIgnoreBindTransform = asset->isInstanced();
+        }
         if (!pImpl->mIgnoreBindTransform) {
            pImpl->cgltfSkinBaseAddress = &gltf->skins[0];
         }
--- a/libs/gltfio/src/ResourceLoader.cpp
+++ b/libs/gltfio/src/ResourceLoader.cpp
@@ -468,9 +468,7 @@ bool ResourceLoader::loadResources(FFilamentAsset* asset, bool async) {
         if (!pImpl->mIgnoreBindTransform) {
             pImpl->mIgnoreBindTransform = asset->isInstanced();
         }
-        if (!pImpl->mIgnoreBindTransform) {
-           pImpl->cgltfSkinBaseAddress = &gltf->skins[0];
-        }
+        pImpl->cgltfSkinBaseAddress = &gltf->skins[0];
         updateBoundingBoxes(asset);
     }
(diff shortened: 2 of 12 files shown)


libs/gltfio/src/ResourceLoader.cpp
```diff
@@ -985,46 +985,109 @@ void ResourceLoader::updateBoundingBoxes(FFilamentAsset* asset) const {
...
+                cgltf_accessor_unpack_floats(accessor, &rawJoints[0], accessor->count * skinningDim);
+            }
+            if (attr.type == cgltf_attribute_type_weights && cgltf_num_components(accessor->type) >= skinningDim) {
+                weights.resize(accessor->count * skinningDim);
+                cgltf_accessor_unpack_floats(accessor, &weights[0], accessor->count * skinningDim);
+            }
+        }
+        if (skin) {
+            std::vector<size_t> jointIndices(rawJoints.begin(), rawJoints.end());
+            auto primCount = static_cast<size_t>(verts.size() / posDim);
```
romainguy, line 1038: primitiveCount

libs/gltfio/src/ResourceLoader.cpp
```diff
@@ -985,46 +985,109 @@ void ResourceLoader::updateBoundingBoxes(FFilamentAsset* asset) const {
...
+            for (size_t i = 0; i < primCount; i++) {
+                float3 pt(verts[posDim * i], verts[posDim * i + 1], verts[posDim * i + 2]);
```
romainguy, line 1040: point not pt

libs/gltfio/src/ResourceLoader.cpp
```diff
@@ -985,46 +985,109 @@ void ResourceLoader::updateBoundingBoxes(FFilamentAsset* asset) const {
...
+                    auto globalJointTransform = tm.getWorldTransform(tm.getInstance(jointEntity));
+                    auto inverseBindMatrix = skin->inverseBindMatrices[jointIndex];
+                    tmp += jointWeight * globalJointTransform * inverseBindMatrix;
+                }
+                for (const auto& inverseGlobalTransform: inverseGlobalTransforms) {
+                    auto skinMatrix = inverseGlobalTransform * tmp;
+                    if (normalizeWeight) {
+                        skinMatrix /= skinMatrix[3].w;
+                    }
+                    auto ptTransformed = (pt.x * skinMatrix[0] + pt.y * skinMatrix[1] + pt.z * skinMatrix[2] + skinMatrix[3]).xyz;
```
romainguy, line 1055: skinnedPoint or transformedPoint

romainguy (changes requested):
Thanks for the submission, I left a few comments, mostly stylistic (names and abuse of auto). Note that AABBs were ignoring skinning on purpose for two reasons: load times and not generating unnecessarily large bounding boxes. I think your solution is a fine default but it would be nice to have a way to turn it off potentially?

libs/gltfio/include/gltfio/ResourceLoader.h
```diff
@@ -52,6 +52,10 @@ struct ResourceConfiguration {
     //! If true, computes the bounding boxes of all \c POSITION attibutes. Well formed glTF files
     //! do not need this, but it is useful for robustness.
     bool recomputeBoundingBoxes;
+
+    //! If true, ignore bind transform for skinned primitives when compute bounding box. Only 
+    //! applicable when recomputeBoundingBoxes is set to true
+    bool ignoreBindTransform;
```
Ankbzpx, line 58: @romainguy Do you think adding this config flag is acceptable? If it is, I will try to add it to Android and Web binding

pixelflinger: yes I think it's fine.

pixelflinger (changes requested):
thanks looks nice. Just would like better comments here and there.

libs/gltfio/src/ResourceLoader.cpp
```diff
@@ -985,46 +996,115 @@ void ResourceLoader::updateBoundingBoxes(FFilamentAsset* asset) const {
...
-            for (cgltf_size index = 0, nprims = mesh->primitives_count; index < nprims; ++index) {
-                prims.push_back(&mesh->primitives[index]);
+        const Skin* skin = nullptr;
+        if (!ignoreBindTransform) {
+            cgltf_skin* const cgltfSkin = iter.first->skin;
+            if (cgltfSkin) {
+                // importSkins unpacked cgltfSkin into FFilamentInstance::SkinVector bijectively so that
+                // the unpacked Skin can be retrieved given cgltfSkin index
+                int skinIndex = cgltfSkin - pImpl->cgltfSkinBaseAddress;
+                skin = &asset->mSkins[skinIndex];
```
Ankbzpx, line 1088: I just realized I forgot instanced asset. `asset->mSkins` was not used for instanced asset so skin will always be `nullptr`. It feels very expensive to compute aabb with skin bindTransform considered for each instance. So I implicitly disable it with `pImpl->mIgnoreBindTransform = asset->isInstanced();`. Any suggestions?

Ankbzpx:
My apologies, I couldn't figure out web binding. I attempted to add it, with `gltf` file passed (It appears that filament js only use `ResourceLoader` for `gltf` file), logs captured in `ResourceLoader::updateBoundingBoxes` side, it is still not functioning.

pixelflinger (approved):

libs/gltfio/src/ResourceLoader.cpp
```diff
@@ -985,46 +998,115 @@ void ResourceLoader::updateBoundingBoxes(FFilamentAsset* asset) const {
         tm.setParent(tm.getInstance(e), 0);
     }
 
-    auto computeBoundingBox = [](const cgltf_primitive* prim, Aabb* result) {
+    using SkinnedPrimitive = std::pair<cgltf_primitive*, const Skin*>;
```
bejado, line 1001: The name `SkinnedPrimitive` is confusing to me because `Skin*` can be null.

libs/gltfio/src/ResourceLoader.cpp
```diff
@@ -985,46 +998,115 @@ void ResourceLoader::updateBoundingBoxes(FFilamentAsset* asset) const {
...
+
+    const size_t posAttrSize = cgltf_num_components(cgltf_type_vec3);
+    const size_t skinningAttrSize = cgltf_num_components(cgltf_type_vec4);
+    const bool weightNormalized = pImpl->mNormalizeSkinningWeights;
+    const bool ignoreBindTransform = pImpl->mIgnoreBindTransform;
+    auto computeBoundingBox = [&](const SkinnedPrimitive skinnedPrimitive, Aabb* result) {
```
bejado, line 1007: This lambda is a bit difficult to read due to the the repeated `!ignoreBindTransform && skin` conditions and interleaved code.  I think this can be refactored into two separate functions: a `computeBoundingBox(const cgltf_primitive*, Aabb*)` which is identical to the current one, and a new `computeBoundingBoxSkinned(const cgltf_primitive*, const Skin*, Aabb*)` function that contains the new skinning logic.

Then, the logic below that invokes `computeBoundingBox` can decide which one to invoke based on if skinning is present and flags like `ignoreBindTransform`.

libs/gltfio/src/ResourceLoader.cpp
```diff
@@ -985,46 +998,115 @@ void ResourceLoader::updateBoundingBoxes(FFilamentAsset* asset) const {
...
+                if (ignoreBindTransform || !skin) {
+                    for (cgltf_size i = 0, j = 0, n = accessor->count; i < n; i++, j += posAttrSize) {
+                        float3 pt(verts[j + 0], verts[j + 1], verts[j + 2]);
+                        aabb.min = min(aabb.min, pt);
+                        aabb.max = max(aabb.max, pt);
+                    }
+                    break;
+                }
+            }
+            if (!ignoreBindTransform) {
```
bejado, line 1041: Should this be `!ignoreBindTransform && skin`?

samples/gltf_viewer.cpp
```diff
@@ -133,6 +134,8 @@ static void printUsage(char* name) {
         "       Do not scale the model to fit into a unit cube\n\n"
         "   --recompute-aabb, -r\n"
         "       Ignore the min/max attributes in the glTF file\n\n"
+        "   --ignore-bind-transform, -g\n"
+        "       Ignore bind transform when recompute aabb\n\n"
```
bejado, line 138:
```suggestion
        "       Ignore bind transform when recomputing aabb\n\n"
```

bejado (approved):

libs/gltfio/src/ResourceLoader.cpp
```diff
@@ -1005,27 +1016,99 @@ void ResourceLoader::updateBoundingBoxes(FFilamentAsset* asset) const {
...
+                Entity jointEntity = skin->joints[jointIndex];
+                mat4f globalJointTransform = tm.getWorldTransform(tm.getInstance(jointEntity));
+                mat4f inverseBindMatrix = skin->inverseBindMatrices[jointIndex];
+                tmp += jointWeight * globalJointTransform * inverseBindMatrix;
+            }
+            for (const auto& inverseGlobalTransform: inverseGlobalTransforms) {
+                mat4f skinMatrix = inverseGlobalTransform * tmp;
+                if (!pImpl->mNormalizeSkinningWeights) {
+                    skinMatrix /= skinMatrix[3].w;
+                }
```
prideout, line 1066: @Ankbzpx why do we care about weight normalization here?  Weight normalization is a separate feature of the loader and it occurs earlier in the pipeline.

Ankbzpx: Here was my take:

The `skinMatrix` is the weighted sum of each `jointMatrice`. If the weights were not normalized, the weighted sum of the `m[3].w` (the last element) of `jointMatrics` is not guaranteed to be 1, so the resulting transformation matrix `skinMatrix` needs to be normalized.

prideout: Thanks, I understand now, but the glTF spec says that they should add as close to 1.0 as possible (section 3.7.3.3).  We added the `mNormalizeSkinningWeights` feature only for non-compliant assets.  If include this `/= w` thing, then we're adding yet another accommodation for non-complaint assets, so I don't think it's necessary.  Moreover it does not match what Filament's vertex shader does.

Ankbzpx: Ah, I see. The logic does sound. Thanks for clarifying and please help remove it then

prideout: By the way @Ankbzpx , is the AABB computation feature something that you're actively using?  It defaults to off and I'd be surprised if many folks are using it.

I'm asking because it is posing some challenges with some upcoming features, and we are considering removing this feature.

Ankbzpx: I no longer work there so I am not sure anymore. Here was our use case back then:

We occasionally received animated `FBX` models authored on client side involving many modeling tricks, that when converted to `glb`, had non identity bind transforms (i.e. in Blender, model shown in the `Object` mode and `Edit` mode varies substantially). Due to nested hierarchy and animation, they were non-trivial to fix in modeling software (it's possible to bake inverse bind transform directly onto geometry, but at risk of breaking it). Thus, this PR aimed to relax those situations.

I believe it falls into category of accommodating non-complaint assets, and since it also hinders your development, it would be logical to remove it for now.

PR #4999, opened 2021-12-28
Compute morphing on GPU

### LINKED ISSUE #4772: Support more than 4 blend shapes
### Feature Request
The ability to access the underlying cgltf data on Android.

I need to find a way around the 4 blend shapes limitation. Accessing the underlying cgltf data to update the vertex manually seems to be the most straight forward solution (`asset->getSourceAsset()` and potentially more?).

Related discussions about the state of blend shapes, [here](https://github.com/google/filament/issues/1487) and [here](https://github.com/google/filament/issues/3726) 

- Ideally morph targets via data texture ([mentioned here](https://github.com/google/filament/issues/1487#issuecomment-533820189) - long-term solution)
or
- The ability the access the underlying cgltf data to update the vertex manually (short-term solution).
or
- Support more than 4 blend shapes out of the box via `RenderableManager::setMorphWeights()`

I tried converting the morphs to animations inside the GLTF file, however animations aren't blending making it an unviable solution.
[Discussion here](https://github.com/google/filament/discussions/4754)

**OS and backend**
Android.

Thanks.

daemyung:
This patch enables to compute morphing on GPU. Filament only used 4 morph targets for morphing but the number of morph targets can be more than 4. Especially 4 morph targets aren't enough to sufficient facial expression. So this patch uploads morphing targets to a texture array and compute morphing on GPU.

Fixes #4772


filament/include/filament/RenderableManager.h
```diff
@@ -463,6 +465,20 @@ class UTILS_PUBLIC RenderableManager : public FilamentAPI {
...
+     */
+    void setMorphWeightsAt(Instance instance,
+            size_t primitiveIndex, const float* weights, int count) noexcept;
+
+    /**
+     * Updates the morphing targets for the given primitive.
+     *
+     * Texture sampler must be SAMPLER_2D_ARRAY and the array size must be greater than weight count.
+     */
+    void setMorphTargetsAt(Instance instance,
```
romainguy, line 479: The fact that a texture is used should be an implementation detail and not exposed to the user. Also why a texture array? You don't seem to use slices anywhere?

filament/src/Material.cpp
```diff
@@ -378,6 +378,10 @@ Handle<HwProgram> FMaterial::getSurfaceProgramSlow(uint8_t variantKey) const noe
 
     if (Variant(variantKey).hasSkinningOrMorphing()) {
         pb.setUniformBlock(BindingPoints::PER_RENDERABLE_BONES, PerRenderableUibBone::_name);
+        pb.setUniformBlock(BindingPoints::PER_RENDER_PRIMITIVE_MORPHING, PerRenderPrimitiveMorphingUib::_name);
+
+        addSamplerGroup(pb, BindingPoints::PER_RENDER_PRIMITIVE_MORPHING,
```
romainguy, line 383: We have a very limited number of samplers on mobile (16) so we probably can't use one just for morphing. This needs to be investigated to see what other sampler could be shared instead.

filament/src/RenderPass.h
```diff
@@ -208,27 +209,28 @@ class RenderPass {
...
-    struct PrimitiveInfo { // 24 bytes
-        FMaterialInstance const* mi = nullptr;                          // 8 bytes (4)
-        backend::Handle<backend::HwRenderPrimitive> primitiveHandle;    // 4 bytes
-        backend::RasterState rasterState;                               // 4 bytes
-        uint16_t index = 0;                                             // 2 bytes
-        Variant materialVariant;                                        // 1 byte
-        uint8_t reserved[13 - sizeof(void*)] = {};                      // 5 byte (9)
+    struct PrimitiveInfo { // 32 bytes
+        FMaterialInstance const* mi = nullptr;                            // 8 bytes (4)
+        PerRenderPrimitiveMorphing const* primitiveMorphing = nullptr;    // 8 bytes (4)
```
romainguy, line 214: This is a big increase per primitive for something that won't be used by most primitives

libs/gltfio/src/MorphHelper.cpp
```diff
@@ -155,13 +105,40 @@ void MorphHelper::addPrimitive(cgltf_mesh const* mesh, int primitiveIndex, Table
...
+            targetsTextureSize.x * sizeof(float4), targetsTextureSize.y, 1);
+
+    enum MorphTargetAttributes { POSITION = 0, NORMAL = 1, COUNT = 2 };
+
+    if (prim.targets_count) {
+        targets.targets = Texture::Builder()
+                .width(targetsTextureSize.x)
+                .height(targetsTextureSize.y)
+                .depth(prim.targets_count * MorphTargetAttributes::COUNT)
+                .format(Texture::InternalFormat::RGBA32F)
```
romainguy, line 139: RGBA32F seems overkill. Why not RGBA16F or even normalized integer formats?

filament/src/Material.cpp
daemyung: That's fine. I have checked whether exceeding the limted number of samplers on iOS, Android and macOS and there was no issues if we use one sampler for morphing. Also after adding one sampler, all sampler slots are occupied.

filament/src/RenderPass.h
daemyung: Yes I know but I couldn't find other ways. Each primitive can have a morphing target. I thought that putting morphing info into HwRenderPrimitive but it's not a good idea. Because PerRenderPrimitiveMorphing has the sampling binding location info and the driver layer shouldn't know about this. If you have good ideas, please let me know.

libs/gltfio/src/MorphHelper.cpp
daemyung: I try to optimize texture format. However the precision of RGBA16F or normalized integer formats is not enough for morph target position.

filament/include/filament/RenderableManager.h
daemyung: Because gltfio can only see public headers in filament and morph targets are created by gltfio. Thus I need a way setting morph targets texture to the primitive of renderable.

The number of morph targets could be a lot. In my case it's 52. If we create a texture per a morph target, we can't binding all samplers because the maximum sampler binding count is 16. So each primitive has a sampler array and each target are uploaded into each slice(even index is for position and odd index is for normal).

romainguy: The public API should receive morph weights as a float array and handle textures internally.

filament/src/Material.cpp
romainguy: That is not fine because user materials (or extensions in gltf) may need samplers. We are already at the limit with some of our users.

filament/include/filament/RenderableManager.h
daemyung: Ok I will do that internally like SkinningBuffer. I think this class is very similar to this case.

From OpenGL ES 3.0 Spec, the minimum number of layers allowed in an array texture is at least 256. Thus if we make a limitation of Filament then it will be fine using texture array.

filament/src/Material.cpp
daemyung: Hm... then there is no more sample binding space? Actually current morphing implementation is not fine as I thought. I want to hear your ideas... or Do you have a plan to detach Sampler from Texture? If we do it that we can binding more texture than 16.

daemyung:
@romainguy Create the MorphingBuffer and Users can update weights, positions and tangets via this class.

filament/src/RenderPass.h
daemyung: I found a way to pack the size of command by 32 bytes. We don't have to worry about the big increasement per the command.

daemyung:
I tried to use SSBO for morphing. Unfortunately this approch is impossible. Because Mali doesn't support SSBO on the vertex shader. I have tested it on Mali-G78 and GL_MAX_VERTEX_SHADER_STORAGE_BLOCKS returns 0. If you want to review my SSBO patch, you can find it at [here](https://github.com/daemyung/filament/commit/2fd5de84d4535e1ff5a37fd6ab4916d05931f1c8).

filament/src/Material.cpp
daemyung: I tried to use SSBO but Mali doesn't support binding SSBO on the vertex shader. Also ARM doesn't have a plan to support this. ARM's answer is [here](https://community.arm.com/support-forums/f/graphics-gaming-and-vr-forum/13645/can-a-mali-g71-use-a-ssbo-in-a-vertex-shader). So it's only way uploading morphing data to a texture.

filament/src/details/MorphTargets.h
```diff
@@ -0,0 +1,96 @@
...
+        void setPositions(math::float3 const* positions, int count) noexcept;
+        void setPositions(math::float4 const* positions, int count) noexcept;
+        void setTangents(math::short4 const* tangents, int count) noexcept;
+
+        bool isAnyDirty(uint8_t dirtyFlags) const noexcept;
+        void clearDirty() const noexcept;
+
+        inline auto getSize() const noexcept { return mSize; }
+        inline auto getPositions() const noexcept { return mPositions; }
+        inline auto getTangents() const noexcept { return mTangents; }
```
pixelflinger, line 74: don't use `auto` for the return type

filament/src/details/MorphTargets.h
```diff
@@ -0,0 +1,96 @@
...
+
+    private:
+        int mSize;
+        utils::HeapAllocator mAllocator;
```
pixelflinger, line 78: Unless I'm mistaken, we know upfront the number of targets and all Target have the same size, am I correct?
If so, we don't need an allocator per target, all we need is a single malloc() to hold all the Target's data.

Also are the `Target` retained for the whole lifetime of `MorphTarget`? (because it's using quite a bit of memory potentially), are they freed once the data is uploaded into the texture/gpu?

libs/filabridge/include/private/filament/UibStructs.h
```diff
@@ -227,6 +226,14 @@ struct PerRenderableUibBone { // NOLINT(cppcoreguidelines-pro-type-member-init)
 static_assert(CONFIG_MAX_BONE_COUNT * sizeof(PerRenderableUibBone) <= 16384,
         "PerRenderableUibBone exceed max UBO size");
 
+struct alignas(16) PerRenderableMorphingUib {
+  static constexpr utils::StaticString _name{ "MorphingUniforms" };
+  int count;
+  math::float4 weights[CONFIG_MAX_MORPH_TARGET_COUNT];
+};
```
pixelflinger, line 233: Indentation is 4 spaces. Definitely don't use `int` here because that's 8 bytes. Use `uint32_t`. Make sure to make the alignment apparent. Either use alignas(16) or  use padding.

I'm confused here, why do we have 4 floats per morph target?

Also I think we should use 16 bits float for the weights. So I think it should be something like:

`math::uint4 weights[CONFIG_MAX_MORPH_TARGET_COUNT / 8];  // 8 weights per "line"`

Or am I missing something?

pixelflinger (changes requested):
Thank you for this amazing change!

A few general comments:

Don't use `int` for count, use `uint32_t` instead to be consistant with the rest of the code. Also counts can't be negative, si a signed integer is not appropriate.

don't use `std::for_each` it's often unnecessary and we're not using it elsewhere.

There are also more changes needed inline.

libs/filabridge/src/SibGenerator.cpp
```diff
@@ -66,12 +66,26 @@ SamplerInterfaceBlock const& SibGenerator::getPerViewSib(uint8_t variantKey) noe
     return v.hasVsm() ? sibVsm : sibPcf;
 }
 
+SamplerInterfaceBlock const& SibGenerator::getPerRenderPrimitiveMorphingSib(uint8_t variantKey) noexcept {
+    using Type = SamplerInterfaceBlock::Type;
+    using Format = SamplerInterfaceBlock::Format;
+
+    static SamplerInterfaceBlock sib = SamplerInterfaceBlock::Builder()
+            .name("Morphing")
+            .add("targets", Type::SAMPLER_2D_ARRAY, Format::FLOAT)
```
bejado, line 75: Should precision be set to `HIGHP`?

bejado (commented):
Please bump [`MATERIAL_VERSION`](https://github.com/google/filament/blob/2f3aac29da7e833378d75dd89fadcfc4bcb11623/libs/filabridge/include/filament/MaterialEnums.h#L30) to 17.

bejado:
Also, some builds are failing on CI.

Windows build:

```
D:\a\filament\filament\filament\src\MorphTargets.cpp(111,13): error C2664: 'void filament::backend::SamplerGroup::setSampler(size_t,filament::backend::Handle<filament::backend::HwTexture>,filament::backend::SamplerGroup::SamplerParams)': cannot convert argument 3 from 'initializer list' to 'filament::backend::SamplerGroup::SamplerParams' [D:\a\filament\filament\out\cmake-mt\filament\filament.vcxproj]
```

Web build:

```
/Users/runner/work/filament/filament/web/filament-js/jsbindings.cpp:1052:53: error: too few arguments to function call, expected 3, have 2
        self->setMorphWeights(instance, {x, y, z, w});
        ~~~~~~~~~~~~~~~~~~~~~        
```

filament/include/filament/MorphTargets.h
```diff
@@ -0,0 +1,116 @@
...
+
+#include <filament/Engine.h>
+
+#include <math/mathfwd.h>
+
+
+namespace filament {
+
+/**
+ * MorphTargets is used to hold morphing data (weights, positions and tangents).
```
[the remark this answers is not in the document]
daemyung: Correct. I will delete that.

filament/include/filament/RenderableManager.h
```diff
@@ -456,12 +455,16 @@ class UTILS_PUBLIC RenderableManager : public FilamentAPI {
     /**
      * Updates the vertex morphing weights on a renderable, all zeroes by default.
      *
-     * This is specified using a 4-tuple, one float per morph target. If the renderable has fewer
-     * than 4 morph targets, then clients should fill the unused components with zeroes.
-     *
      * The renderable must be built with morphing enabled, see Builder::morphing().
      */
-    void setMorphWeights(Instance instance, math::float4 const& weights) noexcept;
+    void setMorphWeights(Instance instance, float const* weights, int count) noexcept;
```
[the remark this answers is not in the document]
daemyung: Got it.

filament/include/filament/MorphTargets.h
```diff
@@ -0,0 +1,116 @@
...
+ */
+class UTILS_PUBLIC MorphTargets : public FilamentAPI {
```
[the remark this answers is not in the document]
daemyung: How's about MorphingBuffer like VertexBuffer?

filament/src/MorphTargets.cpp
```diff
@@ -0,0 +1,260 @@
...
+            stride, getHeight(vertexCount), 1);
+}
+
+static inline auto getPositionIndex(int targetIndex) {
+    return targetIndex * 2 + 0;
+}
+
+static inline auto getTangentIndex(int targetIndex) {
+    return targetIndex * 2 + 1;
+}
```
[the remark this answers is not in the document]
daemyung: Ok. It's done.

filament/src/MorphTargets.cpp
```diff
@@ -0,0 +1,260 @@
...
+
+void FMorphTargets::terminate(FEngine& engine) {
+    FEngine::DriverApi& driver = engine.getDriverApi();
+
+    driver.destroySamplerGroup(mSbHandle);
+    driver.destroyTexture(mTbHandle);
+
+    std::for_each(std::begin(mTargets), std::end(mTargets),
+            [](Target& t) { t.terminate(); });
```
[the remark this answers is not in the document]
daemyung: Ok. It's done.

filament/src/MorphTargets.cpp
```diff
@@ -0,0 +1,260 @@
...
+            "MorphTargets (size=%lu) overflow (size=%lu)",
+            target.getSize(), sizeof(math::short4) * count);
+
+    target.setTangents(tangents, count);
+}
+
+void FMorphTargets::commit(FEngine& engine) const noexcept {
+    FEngine::DriverApi& driver = engine.getDriverApi();
+
+    for (auto ti = 0; ti != mCount; ++ti) {
```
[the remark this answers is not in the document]
daemyung: Ok. It's done.

filament/src/MorphTargets.cpp
```diff
@@ -0,0 +1,260 @@
...
+    if (mSBuffer.isDirty()) {
+        driver.updateSamplerGroup(mSbHandle, std::move(mSBuffer.toCommandStream()));
+    }
+}
+
+void FMorphTargets::bind(backend::DriverApi& driver) const noexcept {
+    driver.bindSamplers(BindingPoints::PER_RENDERABLE_MORPHING, mSbHandle);
+}
+
+FMorphTargets::Target::Target(int size) noexcept
```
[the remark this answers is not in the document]
daemyung: Ok. It's done.

filament/src/components/RenderableManager.h
```diff
@@ -165,27 +174,32 @@ class FRenderableManager : public RenderableManager {
         uint16_t offset = 0;
         bool skinningBufferMode = false;
     };
-
     static_assert(sizeof(Bones) == 12);
 
+    struct MorphWeights {
+        backend::Handle<backend::HwBufferObject> handle;
+        int count = 0;
```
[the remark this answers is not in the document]
daemyung: Ok it's done.

filament/src/components/RenderableManager.cpp
```diff
@@ -562,9 +580,39 @@ void FRenderableManager::setSkinningBuffer(FRenderableManager::Instance ci,
     bones.offset = uint16_t(offset);
 }
 
-void FRenderableManager::setMorphWeights(Instance ci, const float4& weights) noexcept {
-    if (ci) {
-        mManager[ci].morphWeights = weights;
+static void setMorphWeights(FEngine& engine, backend::Handle<backend::HwBufferObject> handle,
+        float const* weights, int count) noexcept {
```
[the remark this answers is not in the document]
daemyung: Ok it's done.

filament/src/components/RenderableManager.h
```diff
@@ -200,13 +214,13 @@ class FRenderableManager : public RenderableManager {
...
             union {
                 // this specific usage of union is permitted. All fields are identical
-                Field<AABB>         aabb;
-                Field<LAYERS>       layers;
+                Field<AABB>          aabb;
+                Field<LAYERS>        layers;
+                Field<CHANNELS>      channels;
+                Field<VISIBILITY>    visibility;
+                Field<PRIMITIVES>    primitives;
+                Field<BONES>         bones;
```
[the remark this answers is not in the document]
daemyung: I want to collect related data. Primitives, bones and morphing are related. If you don't want reordering, I will rollback.

filament/src/details/MorphTargets.h
```diff
@@ -0,0 +1,96 @@
...
+
+        bool isAnyDirty(uint8_t dirtyFlags) const noexcept;
+        void clearDirty() const noexcept;
+
+        inline auto getSize() const noexcept { return mSize; }
+        inline auto getPositions() const noexcept { return mPositions; }
+        inline auto getTangents() const noexcept { return mTangents; }
+
+    private:
+        int mSize;
```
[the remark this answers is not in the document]
daemyung: Ok It's done.

filament/src/details/MorphTargets.h
daemyung: Ok I will use return type explicitly.

filament/include/filament/MorphTargets.h
```diff
@@ -0,0 +1,116 @@
...
+        ~Builder() noexcept;
+        Builder& operator=(Builder const& rhs) noexcept;
+        Builder& operator=(Builder&& rhs) noexcept;
+
+        /**
+         * Size of the morph targets in vertex counts.
+         * @param vertexCount Number of vertex counts the morph targets can hold.
+         * @return A reference to this Builder for chaining calls.
+         */
+        Builder& vertexCount(int vertexCount) noexcept;
```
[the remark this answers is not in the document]
daemyung: Ok it's done. @pixelflinger I will use size_t for count instead of uint32_t.

libs/filabridge/include/private/filament/UibStructs.h
daemyung: Because the uniform block follows std140 memory layout. In std140, the array stride(the bytes between array elements) is always rounded up to the size of a vec4.

libs/filabridge/src/SibGenerator.cpp
daemyung: `MEDIUMP` was enough when I tested on macOS, iOS and Android. If you want to set `HIGHP`precision I will do that.

filament/src/RenderPass.h
```diff
@@ -209,12 +209,11 @@ class RenderPass {
     }
 
     struct PrimitiveInfo { // 24 bytes
-        FMaterialInstance const* mi = nullptr;                          // 8 bytes (4)
-        backend::Handle<backend::HwRenderPrimitive> primitiveHandle;    // 4 bytes
-        backend::RasterState rasterState;                               // 4 bytes
-        uint16_t index = 0;                                             // 2 bytes
-        Variant materialVariant;                                        // 1 byte
-        uint8_t reserved[13 - sizeof(void*)] = {};                      // 5 byte (9)
+        FRenderPrimitive const* primitive = nullptr;    // 8 bytes (4)
```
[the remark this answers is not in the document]
daemyung: Unfortunately the reserved space is not enough to store `FRenderPrimitive const* primitive`. We need 8 bytes but 4 bytes are remained. So if we keep `FMaterialInstance const* mi` and add `FRenderPrimitive const* primitive` then the size of `PrimitiveInfo` should be increased. @romainguy said that it's a big increase. Is it fine the size of `PrimitiveInfo` is increased?

filament/src/RenderPass.cpp
```diff
@@ -577,7 +576,19 @@ void RenderPass::Executor::recordDriverCommands(backend::DriverApi& driver,
                         skinning.offset * sizeof(PerRenderableUibBone),
                         CONFIG_MAX_BONE_COUNT * sizeof(PerRenderableUibBone));
             }
-            driver.draw(pipeline, info.primitiveHandle);
+
+            auto morphing = soaMorphing[info.index];
+            if (UTILS_UNLIKELY(morphing.handle)) {
+                driver.bindUniformBuffer(BindingPoints::PER_RENDERABLE_MORPHING, morphing.handle);
+            }
```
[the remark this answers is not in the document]
daemyung: Yes we can save it. It's done.

filament/src/details/MorphTargets.h
daemyung: Yes you are very correct. But if we don't keep targets, you have to upload target data to buffer every time. I though that it can waste the memory bandwidth. If it's find upload target data to buffer right away, we don't have to target data in staging memory.

You are right. I will figure out to release it.

daemyung:
Sorry for loosing the review history by doing force-push. I worry about the history be tainted by merge commits. I realized that the history is very clean when PR is merged into the main branch. I will do not force-push again.

libs/filabridge/src/SibGenerator.cpp
bejado: Generally we use `highp` for position-related data. @pixelflinger what do you think?

filament/src/components/RenderableManager.h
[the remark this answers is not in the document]
pixelflinger: Because this is a SoA there is no performance change when reordering. So I would prefer to not have the reorder so that it's less confusing in the review. Thank you.

libs/filabridge/include/private/filament/UibStructs.h
pixelflinger: Right, right. But the way to fix this is to unroll the loop 4 (or 8 times) in the shader. Something like:

```
for (i=0 ; i<count ; i+=4) {
  float w[4] = { weight[i/4].x, weight[i/4].y, weight[i/4].z, weight[i/4].w };
  for (j=0 ; j<4 && i+j < count ; j++) {
     // use w[j] here
  }
}
```

I can live with this for now, but please have a comment here (and in the shader) that explains the situation.

libs/filabridge/src/SibGenerator.cpp
pixelflinger: It definitely should be `HIGHP` -- if MEDIUMP was really enough, the texture could be RGBA16F. My guess is that the driver doesn't honor MEDIUMP properly.

filament/include/filament/MorphTargets.h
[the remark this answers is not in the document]
pixelflinger: Yes I like that better, could also be `MorphTargetBuffer`?

android/filamat-android/src/main/java/com/google/android/filament/filamat/MaterialBuilder.java
```diff
@@ -114,14 +114,14 @@
...
         UNUSED,                 // reserved for future use
-        CUSTOM0,                // custom or MORPH_POSITION_0
-        CUSTOM1,                // custom or MORPH_POSITION_1
-        CUSTOM2,                // custom or MORPH_POSITION_2
-        CUSTOM3,                // custom or MORPH_POSITION_3
-        CUSTOM4,                // custom or MORPH_TANGENTS_0
-        CUSTOM5,                // custom or MORPH_TANGENTS_1
-        CUSTOM6,                // custom or MORPH_TANGENTS_2
-        CUSTOM7                 // custom or MORPH_TANGENTS_3
+        CUSTOM0,                // custom
```
pixelflinger, line 117: This is actually a neat side effect of this change!

filament/src/Material.cpp
```diff
@@ -378,6 +378,10 @@ Handle<HwProgram> FMaterial::getSurfaceProgramSlow(uint8_t variantKey) const noe
 
     if (Variant(variantKey).hasSkinningOrMorphing()) {
         pb.setUniformBlock(BindingPoints::PER_RENDERABLE_BONES, PerRenderableUibBone::_name);
+        pb.setUniformBlock(BindingPoints::PER_RENDERABLE_MORPHING, PerRenderableMorphingUib::_name);
```
pixelflinger, line 381: One thing I'm wondering here (maybe for a follow-up PR), we only need the MorphTargets texture in the vertex shader, so... it might not be too bad because in practice we're texture limited in the fragment shader.

From the GLES spec:
```
GL_MAX_COMBINED_TEXTURE_IMAGE_UNITS
params returns one value, the maximum supported texture image units that can be used to access texture maps from the vertex shader and the fragment processor combined. If both the vertex shader and the fragment processing stage access the same texture image unit, then that counts as using two texture image units against this limit. The value must be at least 32.
```
So it  sounds like we can use 32 samplers/textures as long as it's 16 in each stage.

Now the problem is that filament doesn't distinguish between fragment and vertex shader. I need to think more about this.

filament/src/MorphTargets.cpp
```diff
@@ -0,0 +1,219 @@
...
+    return *this;
+}
+
+MorphTargets* MorphTargets::Builder::build(Engine& engine) {
+    return upcast(engine).createMorphTargets(*this);
+}
+
+// ------------------------------------------------------------------------------------------------
+
+constexpr size_t MAX_TEXTURE_WIDTH = 4096;
```
pixelflinger, line 63: add a comment explaining where this is coming from. Also 4096 is not guaranteed to be supported everywhere. The GLES3 spec says:
```
GL_MAX_TEXTURE_SIZE
params returns one value. The value gives a rough estimate of the largest texture that the GL can handle. The value must be at least 2048. See glTexImage2D.
```
So unfortunately, this must be 2048.


On a side note, 2D arrays the numbers of layers is limited to 256 in the minspec:
```
GL_MAX_ARRAY_TEXTURE_LAYERS
params returns one value. The value indicates the maximum number of layers allowed in an array texture, and must be at least 256. See glTexImage2D.
```

filament/src/MorphTargets.cpp
```diff
@@ -0,0 +1,219 @@
...
+
+static const auto FREE_CALLBACK = [](void* mem, size_t, void*) { free(mem); };
```
pixelflinger, line 65:
```suggestion
static const auto FREE_CALLBACK = [](void* mem, size_t, void*) { ::free(mem); };
```

filament/src/MorphTargets.cpp
```diff
@@ -0,0 +1,219 @@
...
+void FMorphTargets::setPositionsAt(FEngine& engine, size_t targetIndex, math::float3 const* positions, size_t count) {
+    ASSERT_PRECONDITION(targetIndex < mCount, "targetIndex must be < count");
+
+    auto size = getSize(mVertexCount);
+
+    ASSERT_PRECONDITION((int)sizeof(math::float3) * count <= size,
+            "MorphTargets (size=%lu) overflow (size=%lu)",
+            size, sizeof(math::float3) * count);
+
+    auto* out = (float4*) malloc(size);
```
pixelflinger, line 127: Maybe add a comment here that we could use a pool instead of `malloc()` directly.

filament/src/RenderPass.cpp
```diff
@@ -577,7 +576,17 @@ void RenderPass::Executor::recordDriverCommands(backend::DriverApi& driver,
                         skinning.offset * sizeof(PerRenderableUibBone),
                         CONFIG_MAX_BONE_COUNT * sizeof(PerRenderableUibBone));
             }
-            driver.draw(pipeline, info.primitiveHandle);
+
+            auto morphWeights = soaMorphing[info.index];
```
pixelflinger, line 580: this can be moved inside the `if`

filament/src/RenderPass.h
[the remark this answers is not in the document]
pixelflinger: Ah yes. I really don't like this extra dereference in this main loop. Let me think about this more.

libs/filabridge/include/private/filament/UibStructs.h
```diff
@@ -227,6 +226,14 @@ struct PerRenderableUibBone { // NOLINT(cppcoreguidelines-pro-type-member-init)
 static_assert(CONFIG_MAX_BONE_COUNT * sizeof(PerRenderableUibBone) <= 16384,
         "PerRenderableUibBone exceed max UBO size");
 
+struct alignas(16) PerRenderableMorphingUib {
+    static constexpr utils::StaticString _name{ "MorphingUniforms" };
+    uint32_t count;
```
pixelflinger, line 231: Actually I think the count shouldn't be here, it should be in the `PerRenderableUib`

libs/filabridge/include/private/filament/UibStructs.h
```diff
@@ -149,8 +149,7 @@ struct alignas(256) PerRenderableUib { // NOLINT(cppcoreguidelines-pro-type-memb
     static constexpr utils::StaticString _name{ "ObjectUniforms" };
     math::mat4f worldFromModelMatrix;
     math::mat3f worldFromModelNormalMatrix;   // this gets expanded to 48 bytes during the copy to the UBO
-    alignas(16) math::float4 morphWeights;    // morph weights (we could easily have 8 using half)
```
pixelflinger, line 152: store the count here  (instead of morphWeights) and instead of storing it in `PerRenderableMorphingUib`

pixelflinger (changes requested):
Thank you for all the changes. I have more comments. Please bear with me, this is a pretty large change!

filament/src/RenderPass.h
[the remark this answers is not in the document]
pixelflinger: Okay I think what we should do at this point in increase the size of `PrimitiveInfo` (like in your first attempt) to 40 bytes, i.e. add the `FRenderPrimitive` pointer + 8 bytes of extra padding.
This will increase the size of `Command` to 48 bytes (from 32), which is I think is acceptable. We might want to increase the size of the command buffer in `Allocator.h`, I would set `FILAMENT_PER_RENDER_PASS_ARENA_SIZE_IN_MB` to 3 and `FILAMENT_PER_FRAME_COMMANDS_SIZE_IN_MB` to 2 -- that's just an increase of 1 MB, it's not the end of the world.

pixelflinger (changes requested):
I commented on what I think we should do for `PrimitiveInfo`

filament/include/filament/MorphTargets.h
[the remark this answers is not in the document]
daemyung: Look good. It's done.

filament/src/components/RenderableManager.h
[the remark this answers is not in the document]
daemyung: Ok I see. This changes are reverted.

filament/src/RenderPass.cpp
```diff
@@ -555,9 +554,9 @@ void RenderPass::Executor::recordDriverCommands(backend::DriverApi& driver,
             // per-renderable uniform
             const PrimitiveInfo info = first->primitive;
             pipeline.rasterState = info.rasterState;
-            if (UTILS_UNLIKELY(mi != info.mi)) {
+            if (UTILS_UNLIKELY(mi != info.primitive->getMaterialInstance())) {
                 // this is always taken the first time
-                mi = info.mi;
+                mi = info.primitive->getMaterialInstance();
```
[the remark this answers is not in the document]
daemyung: Ok it's done.

filament/src/RenderPass.h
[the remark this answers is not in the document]
daemyung: Sounds good! It's done.

libs/filabridge/src/SibGenerator.cpp
daemyung: Ok It's done. As far as I know, the driver doesn't have mediump. They have only lowp and highp.

android/filamat-android/src/main/java/com/google/android/filament/filamat/MaterialBuilder.java
daemyung: Sorry I don't understand your comment. Is it fine or we should keep MORPH_XXXX_X?

filament/src/MorphTargets.cpp
daemyung: Thanks. I changed this value and add comments.

libs/filabridge/include/private/filament/UibStructs.h
daemyung: Ok. It's moved to `PerRenderableUib`.

libs/filabridge/include/private/filament/UibStructs.h
daemyung: Thanks for guide. It's done.

libs/filabridge/include/private/filament/UibStructs.h
daemyung: Add a comment and your ways looks very good. I never thought your way before.

filament/src/MorphTargets.cpp
daemyung: Ok it's done.

filament/src/MorphTargets.cpp
daemyung: Ok it's done.

filament/src/Material.cpp
daemyung: Right. I'm working on this at an other branch because the driver code needs to be chagned.

android/filamat-android/src/main/java/com/google/android/filament/filamat/MaterialBuilder.java
pixelflinger: It's fine!

filament/src/Material.cpp
daemyung: It is solved by #5036.

filament/src/RenderPass.cpp
```diff
@@ -577,6 +580,16 @@ void RenderPass::Executor::recordDriverCommands(backend::DriverApi& driver,
                         skinning.offset * sizeof(PerRenderableUibBone),
                         CONFIG_MAX_BONE_COUNT * sizeof(PerRenderableUibBone));
             }
+
+            auto morphTargetBuffer = info.morphTargetBuffer;
+            if (UTILS_UNLIKELY(morphTargetBuffer)) {
+                auto morphing = soaMorphing[info.index];
+                assert_invariant(morphing.handle);
+                driver.bindUniformBuffer(BindingPoints::PER_RENDERABLE_MORPHING, morphing.handle);
```
pixelflinger, line 588: Add a comment that instead of using a UBO per primitive, we could also have a single UBO for all primitives and use `bindUniformBufferRange` which might be more efficient.

filament/src/components/RenderableManager.cpp
```diff
@@ -372,6 +374,16 @@ void FRenderableManager::create(
                 }
             }
         }
+
+        if (builder->mMorphingEnabled) {
+            MorphWeights& morphWeights = manager[ci].morphWeights;
+            morphWeights = MorphWeights {
+                .handle = driver.createBufferObject(
```
pixelflinger, line 381: add a comment (same as above) that we could use a single UBO for several primitives and use `bindUniformBufferRange`, which might be more efficient.

pixelflinger (approved):
Just have some minor comments. 
How is this tested? Do you have an asset that exercises the new code?

daemyung:
> Just have some minor comments. How is this tested? Do you have an asset that exercises the new code?

I have an asset to test it. You can download the free version at [here](https://www.polywink.com/9-automatic-expressions-blendshapes-on-demand.html) but you need to convert this asset to gltf. I have the converted file but not sure I can share to you.

filament/src/RenderPass.cpp
daemyung: It's done. It's a good point!

filament/src/components/RenderableManager.cpp
daemyung: It's done.

will4621d:
Here's a asset with 50 blendshapes, https://github.com/dilmerv/UnityARBlendShapes/tree/master/Assets/Models/sloth.
I've tested these changes in Android and it works fine. Thanks for all your efforts. @daemyung

pixelflinger:
Build error on Windows:
```
D:\a\filament\filament\filament\src\MorphTargetBuffer.cpp(110,107): error C2440: '<function-style-cast>': cannot convert from 'initializer list' to 'filament::backend::SamplerParams' [D:\a\filament\filament\out\cmake-mt\filament\filament.vcxproj]
D:\a\filament\filament\filament\src\MorphTargetBuffer.cpp(110,26): message : Invalid aggregate initialization [D:\a\filament\filament\out\cmake-mt\filament\filament.vcxproj]
```

daemyung:
> Build error on Windows:
> 
> ```
> D:\a\filament\filament\filament\src\MorphTargetBuffer.cpp(110,107): error C2440: '<function-style-cast>': cannot convert from 'initializer list' to 'filament::backend::SamplerParams' [D:\a\filament\filament\out\cmake-mt\filament\filament.vcxproj]
> D:\a\filament\filament\filament\src\MorphTargetBuffer.cpp(110,26): message : Invalid aggregate initialization [D:\a\filament\filament\out\cmake-mt\filament\filament.vcxproj]
> ```

Thanks. Now it should be fine. Sorry for a build error on Windows.

daemyung:
@bejado 

> Build error on Windows:
> 
> ```
> D:\a\filament\filament\filament\src\MorphTargetBuffer.cpp(110,107): error C2440: '<function-style-cast>': cannot convert from 'initializer list' to 'filament::backend::SamplerParams' [D:\a\filament\filament\out\cmake-mt\filament\filament.vcxproj]
> D:\a\filament\filament\filament\src\MorphTargetBuffer.cpp(110,26): message : Invalid aggregate initialization [D:\a\filament\filament\out\cmake-mt\filament\filament.vcxproj]
> ```

@pixelflinger Thanks. Now it should be fine. Sorry for a build error on Windows.

daemyung:
@bejado I would like to happy if you review and approved this change.

61315:
I'm happy with this change. I mean, just look at them go.

@pixelflinger Tested this PR using one of the  @readyplayerme [avatars](https://github.com/mrdoob/three.js/issues/21636#issuecomment-893520036). One should be able to inspect the morph targets of the glTF asset [here](https://gltf-viewer.donmccurdy.com/).  Note their asset is available under CC BY-NC-SA 4.0.



https://user-images.githubusercontent.com/46559594/149135221-f59cdba8-ffc5-449e-b333-6a49cb0f84df.mp4

pixelflinger (approved):

bejado (approved):

PR #5016, opened 2022-01-05
Added support for triangle strips

jeanlemotan:
Initially discussed [here](https://github.com/google/filament/issues/4508l)
Strips are supported on all current backends and can be useful in some cases.
The changes are minimal.


jeanlemotan:
Oops, some other commits slipped in in PR. I will redo it, closing.

PR #5018, opened 2022-01-05
Fixed cubemap update

jeanlemotan:
Cubemap update was broken by this commit: [6fa82ea73d7263931d6c96d56a42b5cca20a07a0](https://github.com/google/filament/commit/6fa82ea73d7263931d6c96d56a42b5cca20a07a0)
Cubemap update call setcompressedTextureData/setTextureData with 0, 0 as width and height as this was relying on those methods to just use the t->width, t->height instead of the function arguments.
Not sure if that is legal, maybe an assert should check for that case or it should be handled in the function (width = width ? width : t->width)

### DIFF
--- a/filament/backend/src/opengl/OpenGLDriver.cpp
+++ b/filament/backend/src/opengl/OpenGLDriver.cpp
@@ -1807,10 +1807,12 @@ void OpenGLDriver::updateCubeImage(Handle<HwTexture> th, uint32_t level,
     DEBUG_MARKER()
 
     GLTexture* t = handle_cast<GLTexture *>(th);
+    auto width = std::max(1u, t->width >> level);
+    auto height = std::max(1u, t->height >> level);
     if (data.type == PixelDataType::COMPRESSED) {
-        setCompressedTextureData(t, level, 0, 0, 0, std::max(1u, t->width >> level), std::max(1u, t->height >> level), 0, std::move(data), &faceOffsets);
+        setCompressedTextureData(t, level, 0, 0, 0, width, height, 0, std::move(data), &faceOffsets);
     } else {
-        setTextureData(t, level, 0, 0, 0, std::max(1u, t->width >> level), std::max(1u, t->height >> level), 0, std::move(data), &faceOffsets);
+        setTextureData(t, level, 0, 0, 0, width, height, 0, std::move(data), &faceOffsets);
     }
 }
 
--- a/filament/backend/src/opengl/OpenGLDriver.cpp
+++ b/filament/backend/src/opengl/OpenGLDriver.cpp
@@ -1808,9 +1808,9 @@ void OpenGLDriver::updateCubeImage(Handle<HwTexture> th, uint32_t level,
 
     GLTexture* t = handle_cast<GLTexture *>(th);
     if (data.type == PixelDataType::COMPRESSED) {
-        setCompressedTextureData(t, level, 0, 0, 0, t->width >> level, t->height >> level, 0, std::move(data), &faceOffsets);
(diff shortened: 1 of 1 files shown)


bejado (approved):
LGTM. @pixelflinger for approval

filament/backend/src/opengl/OpenGLDriver.cpp
```diff
@@ -1808,9 +1808,9 @@ void OpenGLDriver::updateCubeImage(Handle<HwTexture> th, uint32_t level,
 
     GLTexture* t = handle_cast<GLTexture *>(th);
     if (data.type == PixelDataType::COMPRESSED) {
-        setCompressedTextureData(t, level, 0, 0, 0, 0, 0, 0, std::move(data), &faceOffsets);
+        setCompressedTextureData(t, level, 0, 0, 0, t->width >> level, t->height >> level, 0, std::move(data), &faceOffsets);
```
pixelflinger, line 1811: This doesn't work because `x >>  level` could be 0 when it needs to be 1.  We should use `max(1, x>>level)` instead, like it is done in other places.

pixelflinger (changes requested):

filament/backend/src/opengl/OpenGLDriver.cpp
jeanlemotan: Done in [a80174417c67a87ecb82d7a68197e5706cebf81f](https://github.com/google/filament/pull/5018/commits/a80174417c67a87ecb82d7a68197e5706cebf81f)

filament/backend/src/opengl/OpenGLDriver.cpp
```diff
@@ -1808,9 +1808,9 @@ void OpenGLDriver::updateCubeImage(Handle<HwTexture> th, uint32_t level,
 
     GLTexture* t = handle_cast<GLTexture *>(th);
     if (data.type == PixelDataType::COMPRESSED) {
-        setCompressedTextureData(t, level, 0, 0, 0, 0, 0, 0, std::move(data), &faceOffsets);
+        setCompressedTextureData(t, level, 0, 0, 0, std::max(1u, t->width >> level), std::max(1u, t->height >> level), 0, std::move(data), &faceOffsets);
     } else {
-        setTextureData(t, level, 0, 0, 0, 0, 0, 0, std::move(data), &faceOffsets);
+        setTextureData(t, level, 0, 0, 0, std::max(1u, t->width >> level), std::max(1u, t->height >> level), 0, std::move(data), &faceOffsets);
     }
```
pixelflinger, line 1814: The lines are too long, please use this instead:

```
auto width = std::max(1u, t->width >> level);
auto height = std::max(1u, t->height >> level);
if (data.type == PixelDataType::COMPRESSED) {
    setCompressedTextureData(t, level, 0, 0, 0, width, height, 0, std::move(data), &faceOffsets);
} else {
    setTextureData(t, level, 0, 0, 0, width, height, 0, std::move(data), &faceOffsets);
}
```

pixelflinger (changes requested):
Sorry for the last change request, I just would like the lines to not be too long. This is a very good catch, thank you for the PR!

filament/backend/src/opengl/OpenGLDriver.cpp
jeanlemotan: Done in [1eb243ff4694b43b7d29bf3bfca3567435cfd655](https://github.com/google/filament/pull/5018/commits/1eb243ff4694b43b7d29bf3bfca3567435cfd655)

jeanlemotan:
> Sorry for the last change request, I just would like the lines to not be too long. This is a very good catch, thank you for the PR!

Absolutely no problem, I agree 100% with your reasoning.

pixelflinger (approved):

PR #5020, opened 2022-01-05
ModelViewer move UiHelper to constructor 

srMarlins:
Moving `UiHelper` to the constructor which allows for more composable control of the underlying views & surface.

More specifically, I needed to set the opacity of the underlying views within `ModelViewer`.

### DIFF
--- a/android/filament-utils-android/src/main/java/com/google/android/filament/utils/ModelViewer.kt
+++ b/android/filament-utils-android/src/main/java/com/google/android/filament/utils/ModelViewer.kt
@@ -58,7 +58,10 @@ private const val kSensitivity = 100f
  *
  * See `sample-gltf-viewer` for a usage example.
  */
-class ModelViewer(val engine: Engine) : android.view.View.OnTouchListener {
+class ModelViewer(
+        val engine: Engine,
+        private val uiHelper: UiHelper
+) : android.view.View.OnTouchListener {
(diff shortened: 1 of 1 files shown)


android/filament-utils-android/src/main/java/com/google/android/filament/utils/ModelViewer.kt
```diff
@@ -82,9 +85,9 @@ class ModelViewer(val engine: Engine) : android.view.View.OnTouchListener {
     val view: View
     val camera: Camera
     val renderer: Renderer
-    @Entity val light: Int
+    @Entity
```
romainguy, line 88: Please revert this change

romainguy (changes requested):

android/filament-utils-android/src/main/java/com/google/android/filament/utils/ModelViewer.kt
srMarlins: my bad, should I revert all the format changes?

romainguy: No just this one is fine, thanks!

romainguy (approved):

srMarlins:
Android studio auto imported `import android.animation.Animator` without me realizing so had to delete that

Issue #4724, opened 2021-10-11
ModelViewer + SurfaceView within a RecyclerView crash


srMarlins:
**Describe the bug**
When attaching a `SurfaceView` to a `ModelViewer` within a `RecyclerView`, the view lifecycle isn't properly handled. A `SurfaceView` is detached from the window but not destroyed. When the `SurfaceHolder.Callback` attached within the `UiHelper` callbacks are invoked, the engine has been already been destroyed resulting in the provided stack trace.

**To Reproduce**
1. Put multiple `SurfaceView` backed by `ModelViewer` into a `RecyclerView`
2. Scroll up or down until a view is detached from the window, but not destroyed, then scroll back to that view 

**Expected behavior**
Any cleanup/configuration (specifically with the engine) should be properly cleaned up/recreated when the view detaches/reattaches itself without relying on complete destruction/reconstruction of the view itself. 

**Logs**
```
java.lang.IllegalStateException: Calling method on destroyed Engine
        at com.google.android.filament.Engine.getNativeObject(Engine.java:649)
        at com.google.android.filament.Engine.createSwapChain(Engine.java:293)
        at com.google.android.filament.Engine.createSwapChain(Engine.java:271)
        at com.google.android.filament.utils.ModelViewer$SurfaceCallback.onNativeWindowChanged(ModelViewer.kt:362)
        at com.google.android.filament.android.UiHelper.createSwapChain(UiHelper.java:570)
        at com.google.android.filament.android.UiHelper.access$100(UiHelper.java:120)
        at com.google.android.filament.android.UiHelper$1.surfaceCreated(UiHelper.java:405)
        at android.view.SurfaceView.updateSurface(SurfaceView.java:1173)
        at android.view.SurfaceView.lambda$new$0$SurfaceView(SurfaceView.java:173)
        at android.view.-$$Lambda$SurfaceView$w68OV7dB_zKVNsA-r0IrAUtyWas.onPreDraw(Unknown Source:2)
        at android.view.ViewTreeObserver.dispatchOnPreDraw(ViewTreeObserver.java:1093)
        at android.view.ViewRootImpl.performTraversals(ViewRootImpl.java:3096)
        at android.view.ViewRootImpl.doTraversal(ViewRootImpl.java:1959)
```

**Smartphone (please complete the following information):**
 - Device: Sony Xperia 5ii
 - OS: Android 11

**Additional context**
[...]

PR #5031, opened 2022-01-07
Add view.addOnAttachStateChangeListener to ModelViewer constructor to…

braveBo:
… support using ModelViewer within RecyclerView

Issue: https://github.com/google/filament/issues/4724

example of using the custom addOnAttachStateChangeListener:

        modelViewer = ModelViewer(surfaceView, engine, object : android.view.View.OnAttachStateChangeListener {
            override fun onViewAttachedToWindow(v: android.view.View?) {}
            override fun onViewDetachedFromWindow(v: android.view.View?) {
                modelViewer.destroyResource()
                // avoid call engine.destroy() to prevent crash
            }
        })

### DIFF
--- a/android/filament-utils-android/src/main/java/com/google/android/filament/utils/ModelViewer.kt
+++ b/android/filament-utils-android/src/main/java/com/google/android/filament/utils/ModelViewer.kt
@@ -58,7 +58,10 @@ private const val kSensitivity = 100f
  *
  * See `sample-gltf-viewer` for a usage example.
  */
-class ModelViewer(val engine: Engine) : android.view.View.OnTouchListener {
+class ModelViewer(
+    val engine: Engine,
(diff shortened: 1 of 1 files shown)


romainguy:
Thanks for the contribution but please see my comments in the issue. This is not the right way to fix it. `ModelViewer` is not properly designed for this use case. Destroying resources and the engine on detach is going to be incredibly expensive. I've linked to an example of how to do this right in the issue, please take inspiration from it. This change also exposes internals (destroyEngine, etc.) that could easily be misused and lead to issues.

braveBo:
Hey @romainguy. Thank you so much for your quick review and providing feedback.

1. Could I understand that you suggest creating our own ModelViewer to share the Engine and various other resources between multiple Views? with names such as 'singleEngineModelViewer' or something like that.
2. Is it okay for me to create the new 'xxModelViewer' class under google/android/utils/ folder, making it shareable?

PR #5038, opened 2022-01-10
refactor ostream and fix threading

pixelflinger:


romainguy (approved):

libs/utils/src/ostream.cpp
```diff
@@ -87,142 +94,116 @@ const char* ostream::getFormat(ostream::type t) const noexcept {
...
 
-void ostream::growBufferIfNeeded(size_t s) noexcept {
-    Buffer& buf = getBuffer();
-    if (UTILS_UNLIKELY(buf.size < s)) {
-        size_t used = buf.curr - buf.buffer;
-        size_t newCapacity = std::max(size_t(32), used + (s * 3 + 1) / 2); // 32 bytes minimum
-        buf.reserve(newCapacity);
-        assert(buf.size >= s);
+UTILS_NOINLINE
+ostream& ostream::print(const char* format, ...) noexcept {
```
bejado, line 98: Do we have any need for more than 1 argument? Or are you future-proofing?

libs/utils/src/Log.cpp
```diff
@@ -46,6 +46,7 @@ class LogStream : public ostream {
 };
 
 ostream& LogStream::flush() noexcept {
+        std::lock_guard lock(mLock);
```
bejado, line 49:
```suggestion
    std::lock_guard lock(mLock);
```

bejado (approved):

libs/utils/src/ostream.cpp
pixelflinger: actually it's not about have one argument, it's about not knowing what type it is! I didn't want to use templates and generate tons of code.

PR #5041, opened 2022-01-12
Replace generic tonemap curve with a simpler one

romainguy:
This change removes the shoulder parameter which had inconsistent
behaviors. The new curve is simpler and designed to still match
by default the gray point and the contrast of ACES in a bright
surround.


pixelflinger (approved):
I really prefer that we don't refactor and change behavior at the same time.

pixelflinger:
@bejado this might require updating the goldens

romainguy:
What refactor? Removing the parameter? The new behavior is linked to the refactor, doing them as two separate changes wouldn't make sense. If you are talking about the internal only constant renames… Really not worth doing it separately either.

romainguy:
Note that this doesn't affect the default behavior of Filament so golden images don't need to be re-generated. In addition, the new default values make generic behave even more closely to ACES (ACES remains the default, something we should change some day…).

PR #5046, opened 2022-01-13
Optimize memory of MorphTargetBuffer

daemyung:
This is my last PR for morphing. After #5036 is merged, we can have more sampler binding space because we can bind sampler each shaders. So we can upload positions of morph target and tangents of morph target to each texture. We can save memory footprint and also memory bandwidth.

Thanks everyone to review and test my PRs. Especially really thanks to @pixelflinger, @bejado, @prideout and @romainguy reviewing very big changes.


filament/src/MorphTargetBuffer.cpp
```diff
@@ -74,114 +74,110 @@ static inline size_t getHeight(size_t vertexCount) noexcept {
...
 void FMorphTargetBuffer::setPositionsAt(FEngine& engine, size_t targetIndex, math::float3 const* positions, size_t count) {
     ASSERT_PRECONDITION(targetIndex < mCount, "targetIndex must be < count");
 
-    auto size = getSize(mVertexCount);
+    auto size = getSize<VertexAttribute::POSITION>(mVertexCount);
 
-    ASSERT_PRECONDITION((int)sizeof(math::float3) * count <= size,
-            "MorphTargetBuffer (size=%lu) overflow (size=%lu)",
-            size, sizeof(math::float3) * count);
+    ASSERT_PRECONDITION(sizeof(float3) * count <= size,
```
pixelflinger, line 130: maybe put `sizeof(float3)*count` into a local

filament/src/MorphTargetBuffer.cpp
```diff
@@ -74,114 +74,110 @@ static inline size_t getHeight(size_t vertexCount) noexcept {
...
 
-    Texture::PixelBufferDescriptor buffer(out, size, Texture::Format::RGBA, Texture::Type::FLOAT,
+    Texture::PixelBufferDescriptor buffer(out, size, Texture::Format::RGB, Texture::Type::FLOAT,
             FREE_CALLBACK);
-
     FEngine::DriverApi& driver = engine.getDriverApi();
-    driver.update3DImage(mTbHandle, 0, 0, 0, getPositionIndex(targetIndex),
+    driver.update3DImage(mPbHandle, 0, 0, 0, targetIndex,
             getWidth(mVertexCount), getHeight(mVertexCount), 1,
             std::move(buffer));
```
pixelflinger, line 163: you could factor that code out, probably, because it's identical with the float4 version. Might save a few bytes of code.

libs/filabridge/src/SibGenerator.cpp
```diff
@@ -33,7 +33,8 @@ SamplerInterfaceBlock const& SibGenerator::getPerViewSib(uint8_t variantKey) noe
         auto builder = SamplerInterfaceBlock::Builder();
 
         builder
-            .name("Light");
+            .name("Light")
+            .stageFlags({ .fragment = true });
```
pixelflinger, line 37: good catch

shaders/src/getters.vs
```diff
@@ -78,17 +78,18 @@ void skinPosition(inout vec3 p, const uvec4 ids, const vec4 weights) {
...
+        p += morphingUniforms.weights[i] * texelFetch(morphTargetBuffer_positions, texcoord, 0);
     }
 }
 
 void morphNormal(inout vec3 n) {
     ivec3 texcoord = ivec3(getVertexIndex() % MAX_MORPH_TARGET_BUFFER_WIDTH, getVertexIndex() / MAX_MORPH_TARGET_BUFFER_WIDTH, 0);
     for (uint i = 0u; i < objectUniforms.morphTargetCount; ++i) {
-        texcoord.z = int(i) * 2 + 1;
+        texcoord.z = int(i);
+        vec4 tangent = normalize(vec4(texelFetch(morphTargetBuffer_tangents, texcoord, 0)));
```
pixelflinger, line 90: The `normalize()`  is tricky, and I think this deserves at least a comment and possibly some precision adjustment.

The data in the texture is 16-bit signed integer, meaning texelFetch() will return a *float* value between -32768.0f and 32767.0f, and that doesn't fit in `mediump`, so the sampler might need to actually be `highp`.

What should happen, in theory, is a multiply by `1.0f / 32767.0f` and normalize shouldn't be needed (because the data is supposed to be mormalized AFAIK).

I *think* it's better to do the multiply by `1.0f / 32767.0f` to avoid overflow is normalize(), but  maybe we can live without the normalize() at all?

pixelflinger (changes requested):

shaders/src/getters.vs
romainguy: Not only that but `normalize()` is much more expensive too.

filament/src/MorphTargetBuffer.cpp
daemyung: Ok it's done.

filament/src/MorphTargetBuffer.cpp
daemyung: Ok it's done.

filament/src/MorphTargetBuffer.cpp
```diff
@@ -74,114 +74,99 @@ static inline size_t getHeight(size_t vertexCount) noexcept {
...
 
 void FMorphTargetBuffer::terminate(FEngine& engine) {
     FEngine::DriverApi& driver = engine.getDriverApi();
-
     driver.destroySamplerGroup(mSbHandle);
     driver.destroyTexture(mTbHandle);
+    driver.destroyTexture(mPbHandle);
 }
 
 void FMorphTargetBuffer::setPositionsAt(FEngine& engine, size_t targetIndex, math::float3 const* positions, size_t count) {
```
bejado, line 125: nit: keep line length <= 100 (here and a few other places in this file)

bejado (approved):

libs/filabridge/include/filament/MaterialEnums.h
```diff
@@ -27,7 +27,7 @@
 namespace filament {
 
 // update this when a new version of filament wouldn't work with older materials
-static constexpr size_t MATERIAL_VERSION = 17;
+static constexpr size_t MATERIAL_VERSION = 18;
```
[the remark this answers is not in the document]
daemyung: Got it. Thanks :)

filament/src/MorphTargetBuffer.cpp
daemyung: It's updated. 100 is too short nowadays. Do you have plan increasing to 120? :)

filament/src/MorphTargetBuffer.cpp
```diff
@@ -74,114 +77,102 @@ static inline size_t getHeight(size_t vertexCount) noexcept {
...
 
-static inline size_t getSize(size_t vertexCount) noexcept {
-    const size_t stride = getWidth(vertexCount) * sizeof(float4);
+template<>
+inline size_t getSize<VertexAttribute::POSITION>(size_t vertexCount) noexcept {
+    const size_t stride = getWidth(vertexCount) * sizeof(float3);
+    const size_t height = getHeight(vertexCount);
     return Texture::PixelBufferDescriptor::computeDataSize(
-            Texture::PixelBufferDescriptor::PixelDataFormat::RGBA,
+            Texture::PixelBufferDescriptor::PixelDataFormat::RGB,
```
romainguy, line 88: Are we sure we want RGB here? I'm not sure RGB formats are natively supported by any driver/GPU,, so it might turn into RGBA anyway. Besides I think the Vulkan/Metal backends might just expand to RGBA, and it would be more expensive to do it in two steps. @prideout and @bejado to check me

filament/src/MorphTargetBuffer.cpp
```diff
@@ -74,114 +77,102 @@ static inline size_t getHeight(size_t vertexCount) noexcept {
...
         : mSBuffer(PerRenderPrimitiveMorphingSib::SAMPLER_COUNT),
           mVertexCount(builder->mVertexCount),
           mCount(builder->mCount) {
     FEngine::DriverApi& driver = engine.getDriverApi();
-
     mSbHandle = driver.createSamplerGroup(PerRenderPrimitiveMorphingSib::SAMPLER_COUNT);
-    mTbHandle = driver.createTexture(SamplerType::SAMPLER_2D_ARRAY, 1, TextureFormat::RGBA32F, 1,
-            getWidth(mVertexCount), getHeight(mVertexCount), getDepth(mCount),
-            TextureUsage::DEFAULT);
+    mPbHandle = driver.createTexture(SamplerType::SAMPLER_2D_ARRAY, 1, TextureFormat::RGB32F, 1,
```
romainguy, line 109: See my previous comment. Is RGB32F natively supported? If not we should just request RGBA32F (also could we use RGBA16F instead?)

filament/src/MorphTargetBuffer.cpp
daemyung: Yes you are right. I will revert it. Thanks.

filament/src/MorphTargetBuffer.cpp
daemyung: No it's not natively supported. I will reverted it.

daemyung: When I have tested it RGBA16F isn't enough. This data is very small so the precision is not enough.

daemyung: @romainguy I'm wrong. RGBA16F is enough for positions of morph targets. So I change this format to RGBA16F. But it will fallback to slow copy pass. I will enhance it later.

shaders/src/getters.vs
daemyung: @pixelflinger is right. `unpackHalf2x16` is improper because packing way is different. I have changed sampler precision and devide by `32767.0f`

filament/src/MorphTargetBuffer.cpp
daemyung: I have checked using RGBA16F instead of RGBA32F and it looks fine. I have changed it.

filament/src/MorphTargetBuffer.cpp
```diff
@@ -106,7 +106,7 @@ FMorphTargetBuffer::FMorphTargetBuffer(FEngine& engine, const Builder& builder)
           mCount(builder->mCount) {
     FEngine::DriverApi& driver = engine.getDriverApi();
     mSbHandle = driver.createSamplerGroup(PerRenderPrimitiveMorphingSib::SAMPLER_COUNT);
-    mPbHandle = driver.createTexture(SamplerType::SAMPLER_2D_ARRAY, 1, TextureFormat::RGB32F, 1,
+    mPbHandle = driver.createTexture(SamplerType::SAMPLER_2D_ARRAY, 1, TextureFormat::RGBA16F, 1,
```
pixelflinger, line 109: I'm not understanding this change. I don't think positions can be stored as fp16, it's not enough precision. It's okay for normals usually, though.

pixelflinger (changes requested):

filament/src/MorphTargetBuffer.cpp
daemyung: Ok I will revert it. I agree because fp32 is used for position at the vertex input. Could you talk about this with @romainguy? He want to this.

pixelflinger: I think we could make it an option, but if it's not an option it should be "correct first", so it has to be RGBA32F for positions. Sorry for the back-and-forth.

pixelflinger:
There seem to be a conflict, could you resolve it? Thanks!

filament/src/MorphTargetBuffer.cpp
daemyung: That's fine. :) I also think position's format should be RGBA32F.

daemyung:
> There seem to be a conflict, could you resolve it? Thanks!

Can I do this after #5036 is merged? If we do this now, it will exceeds max sampler binding count.

filament/src/MorphTargetBuffer.cpp
daemyung: It's reverted. Thanks.

pixelflinger (commented):
looks good, just have a few small comments inline.

filament/src/MorphTargetBuffer.cpp
```diff
@@ -41,8 +41,10 @@ BuilderType::Builder::Builder() noexcept = default;
 BuilderType::Builder::~Builder() noexcept = default;
 BuilderType::Builder::Builder(BuilderType::Builder const& rhs) noexcept = default;
 BuilderType::Builder::Builder(BuilderType::Builder&& rhs) noexcept = default;
-BuilderType::Builder& BuilderType::Builder::operator=(BuilderType::Builder const& rhs) noexcept = default;
-BuilderType::Builder& BuilderType::Builder::operator=(BuilderType::Builder&& rhs) noexcept = default;
+BuilderType::Builder&
+BuilderType::Builder::operator=(BuilderType::Builder const& rhs) noexcept = default;
+BuilderType::Builder&
+BuilderType::Builder::operator=(BuilderType::Builder&& rhs) noexcept = default;
```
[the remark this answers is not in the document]
daemyung: Ok It's reverted. The reason why I changed is that I saw `100 columns` in CODE_STYLE.md.

filament/src/MorphTargetBuffer.cpp
```diff
@@ -64,7 +66,8 @@ MorphTargetBuffer* MorphTargetBuffer::Builder::build(Engine& engine) {
 // When you change this value, you must change MAX_MORPH_TARGET_BUFFER_WIDTH at getters.vs
 constexpr size_t MAX_MORPH_TARGET_BUFFER_WIDTH = 2048;
 
-static const auto FREE_CALLBACK = [](void* mem, size_t, void*) { ::free(mem); };
+static const auto FREE_CALLBACK =
+        [](void* mem, size_t, void*) { ::free(mem); };
```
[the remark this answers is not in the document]
daemyung: Ok it's reverted.

filament/src/MorphTargetBuffer.cpp
```diff
@@ -100,117 +106,127 @@ FMorphTargetBuffer::FMorphTargetBuffer(FEngine& engine, const Builder& builder)
...
     // create buffer (here a texture) to store the morphing vertex data
-    mTbHandle = driver.createTexture(SamplerType::SAMPLER_2D_ARRAY, 1,
+    mPbHandle = driver.createTexture(SamplerType::SAMPLER_2D_ARRAY, 1,
             TextureFormat::RGBA32F, 1,
             getWidth(mVertexCount),
             getHeight(mVertexCount),
-            getDepth(mCount),
+            mCount,
+            TextureUsage::DEFAULT);
+    mTbHandle = driver.createTexture(SamplerType::SAMPLER_2D_ARRAY, 1,
```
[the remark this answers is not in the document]
daemyung: Ok it's updated.

shaders/src/getters.vs
```diff
@@ -78,17 +78,18 @@ void skinPosition(inout vec3 p, const uvec4 ids, const vec4 weights) {
...
 
 void morphNormal(inout vec3 n) {
     ivec3 texcoord = ivec3(getVertexIndex() % MAX_MORPH_TARGET_BUFFER_WIDTH, getVertexIndex() / MAX_MORPH_TARGET_BUFFER_WIDTH, 0);
     for (uint i = 0u; i < objectUniforms.morphTargetCount; ++i) {
-        texcoord.z = int(i) * 2 + 1;
+        texcoord.z = int(i);
+        ivec4 tangent = texelFetch(morphTargetBuffer_tangents, texcoord, 0);
         vec3 normal;
-        toTangentFrame(texelFetch(morphing_targets, texcoord, 0), normal);
+        toTangentFrame(float4(tangent) / 32767.0, normal);
```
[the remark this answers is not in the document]
daemyung: It's update. BTW is there any different between your and my code?

filament/src/MorphTargetBuffer.cpp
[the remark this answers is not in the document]
pixelflinger: yes, but for large changes like that, I prefer we do these reformatings separately. thank you though!

shaders/src/getters.vs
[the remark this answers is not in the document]
pixelflinger: technically yours does 4 divides,   mine is rcp + 4 muls. It's unclear what the compiler will actually do, so I prefer to be explicit. No major differences.

pixelflinger (approved):

filament/src/MorphTargetBuffer.cpp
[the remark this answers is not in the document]
romainguy: Not only that but we don't strictly enforce 100 columns. If the lines goes above 100 for a few characters and it's still more readable than splitting the line, we'll leave it as is. In this case it looks/reads better without splitting the line. Very subjective but we're flexible :)

prideout (approved):

prideout:
nit: in the future please provide good description in the commit itself, e.g. in this case you could have explained that using separate textures for positions and tangents allows tangents to become SHORT4 instead of FLOAT4, thus reducing the texture footprint.

daemyung:
> nit: in the future please provide good description in the commit itself, e.g. in this case you could have explained that using separate textures for positions and tangents allows tangents to become SHORT4 instead of FLOAT4, thus reducing the texture footprint.

Yes I see. It’s my fault. I will give more detail information from now on. Thanks!

PR #5055, opened 2022-01-14
Fix, assertion with shadowReceiver = false and VSM

### LINKED ISSUE #5053: Turning off receiveShadows with VSM causes crash
**Describe the bug**
If a gltf lit shader has receiveShadows turned off when VSM is enabled, there is a consistent crash when rendering the scene.

**To Reproduce**
Turn off receiveShadows on an imported gltf model with the lit shader on while using VSM type shadows.

**Expected behavior**
Rendering without a crash.

**Logs**
```
Panic
in filament::backend::Program filament::FMaterial::getProgramBuilderWithVariants(uint8_t, uint8_t, uint8_t) const:434
reason: The material 'ubershader_lit_opaque' has not been compiled to include the required GLSL or SPIR-V chunks for the fragment shader (variant=0x41, filtered=0x41).
```
 - Backend: Metal

bejado:
When VSM is turned on but shadow receiver is turned off for a renderable, we're hitting an assertion:

```
reason: ../../filament/src/Material.cpp:366: failed assertion `!Variant::isReserved(variantKey)'
```

The variant in question is 0x41, which is a reserved variant:

```
//                      +-----+-----+-----+-----+-----+-----+-----+-----+
// Variant              |  0  | VSM | FOG | DEP | SKN | SRE | DYN | DIR |   128
//                      +-----+-----+-----+-----+-----+-----+-----+-----+
//  0x41                   0     1     0     0     0     0     0     1
```

It's reserved because it doesn't make sense to have shadows if the shadow receiver bit is off.

This variant should get filtered to 0x01. We were checking for this in `isValidStandardVariant`, but not accounting for it in `filterVariant`.

Fixes #5053

### DIFF
--- a/libs/filabridge/include/private/filament/Variant.h
+++ b/libs/filabridge/include/private/filament/Variant.h
@@ -185,6 +185,10 @@ namespace filament {
             if (isValidDepthVariant(variantKey)) {
                 return variantKey;
             }
+            // if shadow receiver is disabled, turn off VSM
+            if (!(variantKey & SHADOW_RECEIVER)) {
+                return variantKey & ~(VSM);
+            }
             // when the shading mode is unlit, remove all the lighting variants
             return isLit ? variantKey : (variantKey & UNLIT_MASK);
         }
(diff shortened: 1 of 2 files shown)


bejado:
Hmm, I need to double-check that the logic in `filterVariant` is correct.

romainguy (approved):

pixelflinger (approved):

PR #5058, opened 2022-01-15
add basic support for Android Thermal Manager

pixelflinger:


libs/utils/include/utils/android/ThermalManager.h
```diff
@@ -0,0 +1,58 @@
...
+#include <utils/compiler.h>
+
+struct AThermalManager;
+
+namespace utils {
+
+class ThermalManager {
+public:
+    enum class ThermalStatus : int8_t {
+        ERROR = -1,
```
romainguy, line 29: Could the enum be shared between both headers?

romainguy (approved):

libs/utils/include/utils/android/ThermalManager.h
pixelflinger: We should probably revisit how we handle platform specific implementations in `libutils` but the way it's done currently, is by duplicating the headers  and including the right one at compile time. Sharing the types/enums between them makes things complicated, e.g. introducing a new header that's public but not to be included by the user.

PR #5060, opened 2022-01-16
Fix, potential null dereferences in OpenGLProgram

### LINKED ISSUE #5059: Compute morphing on GPU #4999 breaks bones
hi,

#4999 breaks renderables with bones
`samplerBindings` has nullptr entries, which crashes when accessed later:
https://github.com/google/filament/blob/8d1402d37b0b87cc64ef64550b1f000b9b0a16fd/filament/backend/src/opengl/OpenGLProgram.cpp#L274

I can reproduce it with opengl backend on windows with unmodified `gltf_viewer` and an animated model from gltf sample repro, such as [BrainStem.gltf](https://github.com/KhronosGroup/glTF-Sample-Models/blob/master/2.0/BrainStem/glTF-Embedded/BrainStem.gltf)

daemyung:
Morphing is enabled but no morph targets in some models.

Fixes #5059

### DIFF
--- a/filament/backend/src/opengl/OpenGLProgram.cpp
+++ b/filament/backend/src/opengl/OpenGLProgram.cpp
@@ -273,10 +273,6 @@ void OpenGLProgram::updateSamplers(OpenGLDriver* gld) noexcept {
         BlockInfo blockInfo = blockInfos[i];
         HwSamplerGroup const * const UTILS_RESTRICT hwsb = samplerBindings[blockInfo.binding];
         if (UTILS_UNLIKELY(!hwsb)) {
-#ifndef NDEBUG
-            slog.w << "In material " << name.c_str()
-                   << ": no sampler group to " << (uint32_t) blockInfo.binding << io::endl;
-#endif
             continue;
         }
         SamplerGroup const& UTILS_RESTRICT sb = *(hwsb->sb);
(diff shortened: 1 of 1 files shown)


pixelflinger (commented):
I don't think we should log anything here. Isn't this case completely normal?

daemyung:
> I don't think we should log anything here. Isn't this case completely normal?

Metal backend prints logs in this case. So I print logs at here. If you still want to remove log, please let me know.
Now it's normal behavior. Because `morphing_targets` is required even if only skinning is enabled. So texture can be null.

pixelflinger:
> > I don't think we should log anything here. Isn't this case completely normal?
> 
> Metal backend prints logs in this case. So I print logs at here. If you still want to remove log, please let me know. Now it's normal behavior. Because `morphing_targets` is required even if only skinning is enabled. So texture can be null.

Yes, I think we need to remove the logs for normal behavior, i.e.: if there are no mistake from the user. You can leave a comment instead.

Thanks!

daemyung:
> > > I don't think we should log anything here. Isn't this case completely normal?
> > 
> > 
> > Metal backend prints logs in this case. So I print logs at here. If you still want to remove log, please let me know. Now it's normal behavior. Because `morphing_targets` is required even if only skinning is enabled. So texture can be null.
> 
> Yes, I think we need to remove the logs for normal behavior, i.e.: if there are no mistake from the user. You can leave a comment instead.
> 
> Thanks!

Yes I was wrong. There is no log in `MetalDriver` for no binding sampler groups. I removed it.

daemyung:
@pixelflinger BTW you don't have to review my code on your weekend. Actually it's my personality... I can't stand anything to do. You can deserve your weekend.

pixelflinger (approved):

PR #5062, opened 2022-01-16
Create and bind morphing buffer if skinning is enabled

### LINKED ISSUE #5061: A validation error on Metal
**Describe the bug**
`-[MTLDebugRenderCommandEncoder validateCommonDrawErrors:]:5252: failed assertion `Draw Errors Validation
Vertex Function(main0): missing buffer binding at index 3 for morphingUniforms[0].`

daemyung:
The morphing buffer is needed even if skinning is enabled
because shader is generated by HAS_SKINNING_OR_MORPHING.

Fixes #5061


filament/src/RenderPass.cpp
```diff
@@ -581,15 +581,16 @@ void RenderPass::Executor::recordDriverCommands(backend::DriverApi& driver,
                         CONFIG_MAX_BONE_COUNT * sizeof(PerRenderableUibBone));
             }
 
-            auto morphTargetBuffer = info.morphTargetBuffer;
-            if (UTILS_UNLIKELY(morphTargetBuffer)) {
+            auto morphing = soaMorphing[info.index];
```
pixelflinger, line 584: I would like to avoid this dereference here. We should be able to set things up in a way that the info structure is enough to know if we need to take the if or not.

pixelflinger (changes requested):

filament/src/RenderPass.cpp
daemyung: Ok it's done. After that we don't have to increase `PrimitiveInfo` size. Shoud I revert related changes?

pixelflinger: Yes, definitely -- if we can go back to the old size, absolutely. Thanks.
In your last change, I still see the dereference (soaMorphing[info.index];)?

daemyung: Oh I see... I have also removed that. Now we need more space. :)

pixelflinger (approved):

filament/src/RenderPass.cpp
```diff
@@ -581,15 +591,15 @@ void RenderPass::Executor::recordDriverCommands(backend::DriverApi& driver,
...
                 // Instead of using a UBO per primitive, we could also have a single UBO for all primitives
                 // and use bindUniformBufferRange which might be more efficient.
-                auto morphing = soaMorphing[info.index];
-                assert_invariant(morphing.handle);
-                driver.bindUniformBuffer(BindingPoints::PER_RENDERABLE_MORPHING, morphing.handle);
-                assert_invariant(morphing.count <= morphTargetBuffer->getCount());
-                morphTargetBuffer->bind(driver);
+                driver.bindUniformBuffer(BindingPoints::PER_RENDERABLE_MORPHING,
+                        info.morphWeightBuffer);
+                if (UTILS_UNLIKELY(info.morphTargetBuffer)) {
```
prideout, line 599: Add a comment describing the conditions in which `morphWeightBuffer` is true, but `morphTargetBuffer` is false.  I think this can occur when skinning is enabled but morphing is not, right?

filament/src/components/RenderableManager.cpp
```diff
@@ -375,7 +375,7 @@ void FRenderableManager::create(
             }
         }
 
-        if (builder->mMorphingEnabled) {
+        if (UTILS_UNLIKELY(count > 0 || builder->mMorphingEnabled)) {
```
prideout, line 378: We should rename `count` to `boneCount` because it used very far from its definition.  Also a comment would be helpful to explain why this is necessary even when morphing is disabled.

prideout (changes requested):

bejado (approved):

filament/src/RenderPass.cpp
```diff
@@ -372,19 +373,25 @@ void RenderPass::generateCommandsImpl(uint32_t extraFlags,
         const bool writeDepthForShadowCasters = depthContainsShadowCasters & shadowCaster;
 
         const Slice<FRenderPrimitive>& primitives = soaPrimitives[i];
+        const auto& morphing = soaMorphing[i];
```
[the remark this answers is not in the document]
daemyung: Yeap it's done.

filament/src/RenderPass.cpp
daemyung: Yes it's exactly.

filament/src/components/RenderableManager.cpp
daemyung: Yes that's good point. It's updated.

prideout (approved):

PR #5063, opened 2022-01-16
Add FilamentAsset Java bindings

daemyung:


pixelflinger (approved):

prideout (changes requested):
I do not understand why these were made public when texture-based morphing was added.   Morphing weights can already be set by RenderableManager.

In general, the gltfio API does not provide duplicate API's for all the Filament API's.

daemyung:
> I do not understand why these were made public when texture-based morphing was added. Morphing weights can already be set by RenderableManager.
> 
> In general, the gltfio API does not provide duplicate API's for all the Filament API's.

I just thought it will be helper class. I will close this PR and also delete APIs.

daemyung:
> I do not understand why these were made public when texture-based morphing was added. Morphing weights can already be set by RenderableManager.
> 
> In general, the gltfio API does not provide duplicate API's for all the Filament API's.

@prideout Ok I see. I removed duplicated APIs.

prideout (changes requested):
Looks good but one more question about the `getMorphTargetCount` method that you added.  What does this do if a mesh has several primitives?  Does the glTF spec require all primitives in a given mesh to have the same number of targets?

prideout:
> Looks good but one more question about the `getMorphTargetCount` method that you added. What does this do if a mesh has several primitives? Does the glTF spec require all primitives in a given mesh to have the same number of targets?

Next mind, found this in the glTF spec: `All primitives MUST have the same number of morph targets in the same order.`

prideout (approved):
Thank you!  In the future, I think we should consider removing the internal MorphHelper class.  Its main purpose was to partition the morph weights in order to support > 4 targets.  It's no longer necessary now that we use texture-based morphing.

daemyung:
@prideout Actually I can't find it from GLTF spec... but I guess it's true. You can see `cgltf_mesh` has `target_names_count`

```
typedef struct cgltf_mesh {
	char* name;
	cgltf_primitive* primitives;
	cgltf_size primitives_count;
	cgltf_float* weights;
	cgltf_size weights_count;
	char** target_names;
	cgltf_size target_names_count;
	cgltf_extras extras;
	cgltf_size extensions_count;
	cgltf_extension* extensions;
} cgltf_mesh;
```

So current Filament's morphing count implementation is not perfect. I will fix it.

daemyung:
> Thank you! In the future, I think we should consider removing the internal MorphHelper class. Its main purpose was to partition the morph weights in order to support > 4 targets. It's no longer necessary now that we use texture-based morphing.

Yes right. I will merge `MorphHelper` into `AssetLoader`. I wanted to do this when morphing PR is reviewed but that PR is too big so I want to do that in the different PR.

PR #5066, opened 2022-01-17
Get morph target names from FilamentAsset

### LINKED ISSUE #5065: Can't get morph target names
There is no way to retrieve the morph target name.

Get morph target names from `FilamentAsset`.

daemyung:
Fixes #5065


daemyung:
스크린샷 2022-01-18 오전 12 49 20

libs/viewer/include/viewer/SimpleViewer.h
```diff
@@ -243,6 +243,7 @@ class UTILS_PUBLIC SimpleViewer {
     Settings mSettings;
     int mSidebarWidth;
     uint32_t mFlags;
+    utils::Entity mMorphEntity;
```
prideout, line 246: I don't understand what this is.  I think it's the first morphable entity in the asset?  What if the user wants to play with morphing on an entity other than the first one?

libs/viewer/src/SimpleViewer.cpp
```diff
@@ -985,17 +988,12 @@ void SimpleViewer::updateUserInterface() {
             ImGui::Unindent();
         }
 
-        if (!mMorphWeights.empty() && ImGui::CollapsingHeader("Morphing")) {
```
prideout, line 988: Do the sliders move if an animation is underway?  Should they be disabled while an animation is underway?  How do we guarantee that the weights sum to 1.0?

prideout (changes requested):
The getName API looks good but I have some questions about the new gltf_viewer functionality.

libs/viewer/include/viewer/SimpleViewer.h
pixelflinger: Yeah, I think this just assumes that. 
I think it's still useful for testing morphing though. It's not a great UI. Do we want a list of all morphable entities? Maybe this can be done later?

libs/viewer/src/SimpleViewer.cpp
pixelflinger: I think sliders should move when the animation is underway :-) That goes well with the ImGUI realtime everything way.

I think we don't need to enforce in the UI that all weights sum to 1 -- it's good for testing. However, we could have a text box that displays the sum. @daemyung can you add that?

libs/viewer/include/viewer/SimpleViewer.h
prideout: Another idea is to move this UI into the existing Hierarchy section, and simply parent the sliders under the node that they are associated with.

In the short term this could be useful for testing, but we could at least rename the variable.

libs/viewer/src/SimpleViewer.cpp
```diff
@@ -985,17 +988,12 @@ void SimpleViewer::updateUserInterface() {
...
-                std::string label = ss.str();
-                ImGui::SliderFloat(label.data(), &mMorphWeights[i], 0.0f, 1.0);
-            }
-            for (size_t i = 0, c = mAsset->getEntityCount(); i != c; ++i) {
-                mAsset->setMorphWeights(mAsset->getEntities()[i],
-                        mMorphWeights.data(), mMorphWeights.size());
+                ImGui::SliderFloat(mAsset->getMorphTargetNameAt(mMorphEntity, i),
+                        &mMorphWeights[i], 0.0f, 1.0);
             }
+            mAsset->setMorphWeights(mMorphEntity, mMorphWeights.data(), mMorphWeights.size());
```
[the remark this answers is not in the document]
daemyung: Yes it's done by #5063.

libs/viewer/include/viewer/SimpleViewer.h
daemyung: That's good idea. It's chagned.

daemyung:
Sorry guys I merged main branch incorrectly so I did force-push.

libs/viewer/src/SimpleViewer.cpp
prideout: Re: summing to 1.0, doh, I was thinking of skinning!

@daemyung this is a really cool feature, but I want to avoid getting bug reports like: "the sliders don't work" because users try them on models that have animation that affect morph weights, in which case the sliders won't do anything.

Note that when `mCurrentAnimation == 0`  in `SimpleViewer`, that means animation is disabled, so we could perhaps gray out the sliders when that's false.

For the "sliders move on their own" idea that Mathias likes, maybe we could expose more of the Animator internals, but I'm reluctant to do that.  I would like to keep the simplicity of the Animator API (gltfio is just a loader).  Is there any another way we could do it?  For example maybe it wouldn't be so bad to track the values in the app.

prideout:
@daemyung can this be split into two different PR's?  The target name queries look fine, and we could merge that before settling on the gltf_viewer feature.

daemyung:
> @daemyung can this be split into two different PR's? The target name queries look fine, and we could merge that before settling on the gltf_viewer feature.

Yes no problem :)

libs/viewer/src/SimpleViewer.cpp
daemyung: @prideout I think to gray out the sliders is the best solution now. I will do it that on the different PR.

daemyung:
> @daemyung can this be split into two different PR's? The target name queries look fine, and we could merge that before settling on the gltf_viewer feature.

@pixelflinger Yes it's not a problem. Changes in SimpleViewer is reverted.

libs/gltfio/src/MorphHelper.cpp
```diff
@@ -154,4 +165,21 @@ void MorphHelper::addPrimitive(cgltf_mesh const* mesh, int primitiveIndex, Entit
...
+    const auto count = mesh->target_names_count;
+    if (count == 0) {
+        return;
+    }
+
+    auto& entry = mMorphTable[entity];
+    auto& names = entry.targetNames;
+
+    assert_invariant(names.empty());
+    names = utils::FixedCapacityVector<utils::CString>::with_capacity(count);
```
prideout, line 178: I think you can call "reserve" instead of doing a copy assignment.  Not a big deal.

prideout (approved):

libs/gltfio/src/MorphHelper.cpp
daemyung: Thanks it's updated.

pixelflinger (approved):

PR #5068, opened 2022-01-18
Reduce cascade of run-time warnings.

prideout:
The `uint8_t` here needs to be cast to an integer to be printed
correctly. Also these warnings seem to be quite common so we should
hide them in Release builds.

These warnings probably indicate an actual problem that I do not fully
understand yet.

### DIFF
--- a/filament/backend/src/metal/MetalDriver.mm
+++ b/filament/backend/src/metal/MetalDriver.mm
@@ -1388,8 +1388,8 @@ void MetalDriver::enumerateSamplerGroups(
         }
         const auto* metalSamplerGroup = mContext->samplerBindings[samplerGroupIdx];
         if (!metalSamplerGroup) {
-            utils::slog.w << "Program has non-empty samplerGroup (index " << samplerGroupIdx <<
-                    ") but has not bound any samplers." << utils::io::endl;
+            // Do not emit warning here. For example this can arise when skinning is enabled
+            // and the morphing texture is unused.
             continue;
         }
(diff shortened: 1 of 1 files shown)


pixelflinger:
@prideout, I think this is due to the morphing change -- it made it "not an error" to have BindingPoints that are not set. I think it's because morphing and skinning are using the same variant, but skinning doesn't use a texture in the vtx shader, so it's binding point is empty (and there is a binding point because it's used for morphing... which is the same variant).

filament/backend/src/metal/MetalDriver.mm
```diff
@@ -1388,8 +1388,11 @@
         }
         const auto* metalSamplerGroup = mContext->samplerBindings[samplerGroupIdx];
         if (!metalSamplerGroup) {
-            utils::slog.w << "Program has non-empty samplerGroup (index " << samplerGroupIdx <<
-                    ") but has not bound any samplers." << utils::io::endl;
+#ifndef NDEBUG
```
pixelflinger, line 1391: I believe we don't need this even in DEBUG.

prideout: Agreed.  Changed the PR accordingly.

pixelflinger (approved):

bejado (commented):
Hm, my initial reaction is that this should be fixed at the Filament level. I need to investigate this a bit more.

daemyung:
> @daemyung, am I correct here?
Yes it its.

bejado (approved):

PR #5069, opened 2022-01-18
allow a scene to have both transmission and reflection

pixelflinger:
what's not allowed at this point is an object which has both screen
space refraction and reflection.

this is accomplished by just always disabling screen-space reflections
for the translucent step of the refraction pass
Screen Shot 2022-01-18 at 2 44 41 PM
.

### DIFF
--- a/filament/src/Renderer.cpp
+++ b/filament/src/Renderer.cpp
@@ -813,6 +813,7 @@ FrameGraphId<FrameGraphTexture> FRenderer::refractionPass(FrameGraph& fg,
 
         config.refractionLodOffset = refractionLodOffset;
         config.clearFlags = TargetBufferFlags::NONE;
+        config.hasScreenSpaceReflections = false;
         output = colorPass(fg, "Color Pass (transparent)",
                 desc, config, colorGradingConfig, translucentPass, view);
 
--- a/filament/src/Renderer.cpp
+++ b/filament/src/Renderer.cpp
@@ -813,7 +813,6 @@ FrameGraphId<FrameGraphTexture> FRenderer::refractionPass(FrameGraph& fg,
 
         config.refractionLodOffset = refractionLodOffset;
         config.clearFlags = TargetBufferFlags::NONE;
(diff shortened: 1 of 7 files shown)


prideout (approved):

bejado (approved):
So easy!

romainguy (approved):

filament/src/Renderer.cpp
```diff
@@ -972,32 +975,23 @@ FrameGraphId<FrameGraphTexture> FRenderer::colorPass(FrameGraph& fg, const char*
...
+                            getClipSpaceToTextureSpaceMatrix() *
+                            cameraInfo.projection;
+                    reprojection =
+                            getClipSpaceToTextureSpaceMatrix() *
+                            historyProjection *
+                            inverse(cameraInfo.view * cameraInfo.worldOrigin);
                 }
+                TextureHandle ssrHandle = data.ssr ?
+                        resources.getTexture(data.ssr) : ppm.getOneTexture();
+                view.prepareSSR(ssrHandle,config.refractionLodOffset,
```
bejado, line 992:
```suggestion
                view.prepareSSR(ssrHandle, config.refractionLodOffset,
```

filament/src/Renderer.cpp
```diff
@@ -972,32 +975,23 @@ FrameGraphId<FrameGraphTexture> FRenderer::colorPass(FrameGraph& fg, const char*
                         resources.getTexture(data.structure) : ppm.getOneTexture());
 
                 // set screen-space reflections or screen-space refractions
```
bejado, line 977:
```suggestion
                // set screen-space reflections and screen-space refractions
```

bejado (changes requested):
Can you update [this comment](https://github.com/google/filament/blob/4b572cf0657ce49b7e5aafd8420072655f6e71e7/libs/filamat/src/shaders/ShaderGenerator.cpp#L311-L314) in ShaderGenerator.cpp? Other than that, LGTM!

PR #5070, opened 2022-01-19
utils: Introduce RangeMap container and unit test.

prideout:
This will allow the Vulkan backend to efficiently track the subresource
image layouts for each texture.

This is a sparse container for a series of ordered non-overlapping
integer intervals, where each interval maps to a concrete value.


daemyung:
@prideout Do you have a plan to support recording command buffers on many threads in Vulkan backend? If you have this plan, it's not good to track the current image layout.

prideout:
> @prideout Do you have a plan to support recording command buffers on many threads in Vulkan backend? If you have this plan, it's not good to track the current image layout.

If we add multi-threading to the vk backend, then yes we would need to manage each thread's access to textures.  This might be a general concern, not just for layouts but also for texel data.

I'm curious what other Vulkan engines do out there, so I also looked at the MoltenVK source and they use a custom vector type to track each subresource.  I think my RangeMap solution is better for overall memory footprint, but worse in terms of heap allocations.  In the future I should probably use a custom allocator.

prideout:
cc @chrisforbes in case he has thoughts about tracking image layout.

bejado (approved):

libs/utils/include/utils/RangeMap.h
```diff
@@ -0,0 +1,309 @@
...
+ * RangeMap has a low memory footprint if it contains fairly homogeneous data. Internally, the
+ * intervals are automatically split and merged as elements are added or removed.
+ *
+ * Each interval maps to an instance of ValueType, which should support cheap equality checks
+ * and copy assignment. (simple concrete types are ideal)
+ *
+ * KeyType should support operator< because intervals are internally sorted using std::map.
+ */
+template<typename KeyType, typename ValueType>
+class RangeMap {
```
pixelflinger, line 40: note: when actually using this class, it *might* be good to explicitly instantiate it, because it looks like all these methods could generate a lot of code.

One question I have is, is it possible to refactor this so either:
- some code could be moved into a .cpp (i.e. code that doesn't depend on the templated types)
- or, make the templated class just a wrapper (for type safety) around a generic class

libs/utils/include/utils/RangeMap.h
```diff
@@ -0,0 +1,309 @@
...
+public:
+    /**
+     *  Replaces all slots between first (inclusive) and last (exclusive).
+     */
+    void add(KeyType first, KeyType last, const ValueType& value) noexcept {
+        // First check if an existing range contains "first".
+        iterator iter = find(first);
```
pixelflinger, line 47: I would rename `find` to something more explicit. I initially thought it was `std::find`.

libs/utils/include/utils/RangeMap.h
```diff
@@ -0,0 +1,309 @@
...
+    const_iterator end() const noexcept { return mMap.end(); }
+
+    iterator begin() noexcept { return mMap.begin(); }
+    iterator end() noexcept { return mMap.end(); }
+
+    Range<KeyType>& get_range(iterator iter) const { return iter->second.first; }
+    ValueType& get_value(iterator iter) const { return iter->second.second; }
+
+    const Range<KeyType>& get_range(const_iterator iter) const { return iter->second.first; }
+    const ValueType& get_value(const_iterator iter) const { return iter->second.second; }
```
pixelflinger, line 198: this naming convention is not filament's. This should be `getRange()`, etc...

libs/utils/include/utils/RangeMap.h
```diff
@@ -0,0 +1,309 @@
...
+    // Erase the given range if it contains no elements.
+    void clean(iterator iter) {
+        Range<KeyType>& range = get_range(iter);
+        assert_invariant(range.first <= range.last);
+        if (range.first == range.last) {
+            mMap.erase(iter);
+        }
+    }
+
+    const_iterator find(KeyType key) const noexcept {
```
pixelflinger, line 275: can one `find` be implemented in term of the other one, e.g. using `const_cast`?  It seems strange to have the same exact code duplicated twice.

pixelflinger (changes requested):
Mainly the naming conventions don't match filament's. e.g. `foo_bar()` should be `fooBar()`

libs/utils/include/utils/RangeMap.h
```diff
@@ -0,0 +1,269 @@
...
+ */
+
+#ifndef TNT_UTILS_RANGEMAP_H
+#define TNT_UTILS_RANGEMAP_H
+
+#include <utils/Panic.h>
+#include <utils/Range.h>
+#include <utils/debug.h>
+
+#include <map>
```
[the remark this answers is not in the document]
prideout: We need an ordered map for this to work.  std::maps also features logarithmic complexity when calling `lower_bound`, which is heavily leveraged here.

Interestingly Tessil does have some "sparse container" stuff but they don't suit our needs here.

libs/utils/include/utils/RangeMap.h
prideout: Fixed.  Tried a few different ways (including `const_cast` and `std::as_const`) but settled on using a template helper method because that's the easiest way to handle both `const_iterator` and `iterator` return types.

libs/utils/include/utils/RangeMap.h
prideout: Fixed.

libs/utils/include/utils/RangeMap.h
prideout: Renamed to findRange.

libs/utils/include/utils/RangeMap.h
prideout: Hmm I tried but wasn't able to move any code into a cpp file, since all methods use template types.  Keeping it all in the header might not be so bad, at least it ensures that we don't bloat libutils for clients who don't need RangeMap.

FWIW, the Vulkan backend will likely instantiate this with KeyType=uint32 and ValueType=VkImageLayout.

pixelflinger (approved):
I tried to review the algorithm but it's a bit difficult -- seems right to me, so relying on test for this sounds like a plan :-)

prideout:
By the way I am intentionally waiting to merge this, I'm currently using it with a local PR that I'll put up soon.

PR #5074, opened 2022-01-19
Fix integration of screen-space reflections with lighting

pixelflinger:
screen-space reflections are integrated in the indirect light computation
(similarly to screen-space refraction), this is because it can be seen as
the specular term of screen-space GI.

However, because this effect is screen-space, the data it accesses has 
the ibl luminance (and exposure) applied.

Besides some minor refactoring, this change just applies the 
screen-space reflections *after* the ibl's specular term is exposed.

### DIFF
--- a/shaders/src/light_reflections.fs
+++ b/shaders/src/light_reflections.fs
@@ -191,17 +191,17 @@ highp mat4 scaleMatrix(const highp float x, const highp float y) {
 }
 
 /**
- * Evaluates screen-space reflections, storing a premultiplied color in Fr.rgb if there's a hit.
- * r is the desired reflected vector.
+ * Evaluates screen-space reflections, returning a color if there's a hit.
+ * wsRayDirection is the desired reflected vector.
  *
- * Fr.a is set to a value between [0, 1] representing the "opacity" of the reflection. 1.0f is full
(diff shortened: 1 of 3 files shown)


prideout (approved):
Looks okay but I'm not deeply familiar with this.

shaders/src/light_reflections.fs
```diff
@@ -195,13 +195,12 @@ highp mat4 scaleMatrix(const highp float x, const highp float y) {
  * r is the desired reflected vector.
  *
  * Fr.a is set to a value between [0, 1] representing the "opacity" of the reflection. 1.0f is full
- * screen-space reflection. Values < 1.0f should be blended with the scene's IBL. Fr.rgb is
- * premultiplied by Fr.a.
+ * screen-space reflection. Values < 1.0f should be blended with the scene's IBL.
```
bejado, line 198: Remove the word "premultiplied" from line 194

bejado (approved):

shaders/src/light_reflections.fs
pixelflinger: ah yeah, sorry about that btw. When working on the code I realized that it was more logical for it to not be premultiplied. Similarly to the refraction.

PR #5077, opened 2022-01-20
Get the morph target count from the renderable

daemyung:


filament/src/components/RenderableManager.cpp
```diff
@@ -596,14 +596,22 @@ static void updateMorphWeights(FEngine& engine, backend::Handle<backend::HwBuffe
     driver.updateBufferObject(handle, { out, size }, 0);
 }
 
-void FRenderableManager::setMorphWeights(Instance instance, float const* weights, size_t count) noexcept {
+void FRenderableManager::setMorphTargetCount(Instance instance, size_t count) noexcept {
     if (instance) {
+        ASSERT_PRECONDITION(count < CONFIG_MAX_MORPH_TARGET_COUNT,
```
pixelflinger, line 601: when using `ASSERT_*` the method cannot be `noexcept`, because these will throw if exceptions are enabled.

filament/src/components/RenderableManager.cpp
```diff
@@ -596,14 +596,22 @@ static void updateMorphWeights(FEngine& engine, backend::Handle<backend::HwBuffe
...
+                "Only %d morph targets are supported (count=%d)", CONFIG_MAX_MORPH_TARGET_COUNT, count);
+
         MorphWeights& morphWeights = mManager[instance].morphWeights;
         morphWeights.count = count;
+    }
+}
 
+void FRenderableManager::setMorphWeights(Instance instance, float const* weights, size_t count) noexcept {
+    if (instance) {
         ASSERT_PRECONDITION(count < CONFIG_MAX_MORPH_TARGET_COUNT,
```
pixelflinger, line 611: same here, can't be `noexcept` if `ASSERT_*` is used.

pixelflinger (changes requested):
Can you explain why we need this and how this interacts with `MorphTargetBuffer`?

I thought `MorphTargetBuffer` was the one with a count of target? Should `setCount` be on `MorphTargetBuffer`?

It seems to me that if we have a `setMorphTargetCount` on the renderable, we should also have a `setOffset` so that you can use a range inside a `MorphTargetBuffer`.

Please le me know if I'm misunderstanding.

daemyung:
> Can you explain why we need this and how this interacts with `MorphTargetBuffer`?

Sorry don't explain details. glTF spec says that `All primitives MUST have the same number of morph targets in the same order`. Thus I think it's more proper having the morph target count in the renderable. Also when the morph target buffer is added to a renderable, the morph target count must same to the renderable's morph target count. I think the morph target count should be set by 'RenderableManager::Builder`. I will fix.

> I thought `MorphTargetBuffer` was the one with a count of target? Should `setCount` be on `MorphTargetBuffer`?

Yes your understanding is right. However I don't think `setCount` be on `MorphTargetBuffer` unless if we give flexibility to `MorphTargetBuffer`

> It seems to me that if we have a `setMorphTargetCount` on the renderable, we should also have a `setOffset` so that you can use a range inside a `MorphTargetBuffer`.

Yes right. I will update it.

filament/src/components/RenderableManager.cpp
daemyung: Thanks it's updated.

filament/src/components/RenderableManager.cpp
daemyung: Thanks it's updated.

filament/include/filament/RenderableManager.h
```diff
@@ -443,29 +443,34 @@ class UTILS_PUBLIC RenderableManager : public FilamentAPI {
...
 
     /**
      * Updates the vertex morphing weights on a renderable, all zeroes by default.
      *
      * The renderable must be built with morphing enabled, see Builder::morphing().
      */
-    void setMorphWeights(Instance instance, float const* weights, size_t count) noexcept;
-
+    void setMorphWeights(Instance instance,
+            float const* weights, size_t count, size_t offset = 0);
```
pixelflinger, line 461: Please describe  all parameters in the comment (especially offset) -- it should have been done in the other APIs, sorry about that, we will fix.

filament/src/components/RenderableManager.h
```diff
@@ -106,13 +106,14 @@ class FRenderableManager : public RenderableManager {
...
-    inline void setBones(Instance instance, math::mat4f const* transforms, size_t boneCount, size_t offset = 0) noexcept;
+    inline void setBones(Instance instance, Bone const* transforms, size_t boneCount, size_t offset = 0);
+    inline void setBones(Instance instance, math::mat4f const* transforms, size_t boneCount, size_t offset = 0);
     inline void setSkinningBuffer(Instance instance, FSkinningBuffer* skinningBuffer,
-            size_t count, size_t offset) noexcept;
-    void setMorphWeights(Instance instance, float const* weights, size_t count) noexcept;
+            size_t count, size_t offset);
+    void setMorphWeights(Instance instance, float const* weights, size_t count, size_t offset);
     void setMorphTargetBufferAt(Instance instance,
             size_t primitiveIndex, FMorphTargetBuffer* morphTargetBuffer) noexcept;
```
pixelflinger, line 115: Reading this code again, I think `setMorphTargertBufferAt` should work like `setGeometryAt`. 

So I think it should have `uint8_t level` parameter, which for now will always be 0, because we dont support LODs yet. But the API should have it.

This API should also have a `size_t offset, size_t count`, which define a range *inside* the `MorphTargetBuffer`. Then you need to check that:

1. this range fits in the `MorphTargetBuffer`
2. the count matches the renderable's morph target count
3. make the appropriate changes in the shader so that the offset is taken into account

This will make the API more similar to `setGeometryAt` and it will allow us to use a larger `MorphTargetBuffer` for several primitives (or in the future for lods).

I think this change should actually be done first and independently from this PR. This is my bad for not catching this before.

Please let me know what you think. Thank you!

pixelflinger (changes requested):
Please see comments inline, I think we need to change `setMorphTargetBufferAt`, sorry for not catching this earlier.

filament/src/components/RenderableManager.h
daemyung: @pixelflinger Thanks. I agree `setMorphTargetBufferAt` shoud be work like `setGeometryAt` but I have some different idea.

I think `setMorphTargetBufferAt` should be changed to `setMorphTargetsAt` like `setGeometryAt`. Now we can know how many morph targets are needed when `Renderable` is created. Thus instead of creating and setting `MorphTargetBuffer` outside of `RenderPrimitive` , we can create `MorphTargetBuffer` in `RenderPrimitive` and update morph target data like `setGeometryAt`.

Your idea is great that we can share large `MorphTargetBuffer` with many `RenderPrimitive`s but I'm not sure sharing `MorphTargetBuffer` will be very rare and I guess it will be not happend. Because internally `MorphTargetBuffer`'s resource type is `Texture` not `Buffer`. If it is `Buffer` your idea will be very useful.

Please let me know your idea. Thank you!

daemyung: @pixelflinger Sorry ignore my previous comment. Yes your idea is correct. The vertex count exactly same at geometry and morphing. Thanks.

pixelflinger (approved):

prideout (approved):

pixelflinger:
@daemyung this now has conflicts because of my last change. Sorry.

daemyung:
> @daemyung this now has conflicts because of my last change. Sorry.

@pixelflinger No problem at all. :) I resolved conflicts.

filament/src/components/RenderableManager.cpp
```diff
@@ -595,30 +595,42 @@ static void updateMorphWeights(FEngine& engine, backend::Handle<backend::HwBuffe
...
-            updateMorphWeights(mEngine, morphWeights.handle, weights, count);
+            updateMorphWeights(mEngine, morphWeights.handle, weights, count, offset);
         }
     }
 }
 
 void FRenderableManager::setMorphTargetBufferAt(Instance instance,
         size_t primitiveIndex, FMorphTargetBuffer* morphTargetBuffer) noexcept {
     if (instance) {
+#if !defined(NDEBUG)
```
bejado, line 615: I don't think we need to surround `assert_invariant` with a `NDEBUG` check, we already do this inside [`debug.h`](https://github.com/google/filament/blob/4060b558c79207ad9098c9cd7eb130f95c683a14/libs/utils/include/utils/debug.h#L26-L30).

daemyung: I think `ASSERT_PRECONDITION`will be better.

filament/include/filament/RenderableManager.h
daemyung: Ok it's update. Sorry for forgetting your comment.

filament/src/components/RenderableManager.cpp
```diff
@@ -588,12 +588,11 @@ void FRenderableManager::setSkinningBuffer(FRenderableManager::Instance ci,
...
-    auto size = sizeof(float4) * count;
-    auto* UTILS_RESTRICT out = (PerRenderableMorphingUib*)driver.allocate(size);
-    memset(out, 0, size);
+    auto size = sizeof(PerRenderableMorphingUib);
+    auto* UTILS_RESTRICT out = driver.allocatePod<PerRenderableMorphingUib>(1);
     std::transform(weights, weights + count, out->weights,
-            [](float value) { return float4(value); });
-    driver.updateBufferObject(handle, { out, size }, sizeof(float4) * offset);
+            [](float value) { return float4(value, 0, 0, 0); });
+    driver.updateBufferObject(handle, { out, size }, 0);
```
pixelflinger, line 595: Actually your code is much better! (and also it handles `offset`, which this one doesn't).

But please keep `return float4(value, 0, 0, 0);`

pixelflinger (changes requested):

filament/src/components/RenderableManager.cpp
daemyung: Oh it's my mistake. Thanks!

filament/src/components/RenderableManager.cpp
```diff
@@ -592,7 +592,7 @@ static void updateMorphWeights(FEngine& engine, backend::Handle<backend::HwBuffe
     auto* UTILS_RESTRICT out = driver.allocatePod<PerRenderableMorphingUib>(1);
     std::transform(weights, weights + count, out->weights,
```
pixelflinger, line 593: your original code here was better -- because this allocates the whole 16KiB, but we only need `sizeof(float)*count`.

I'm sorry I caused confusion with my change, it's because your first code was using `sizeof(PerRenderableMorphingUib)`

```suggestion
    auto* UTILS_RESTRICT out = (float4*)driver.allocate(sizeof(float4) * count);
    std::transform(weights, weights + count, out,
```

pixelflinger (changes requested):

filament/src/components/RenderableManager.cpp
daemyung: Ah I understand. Thanks!

pixelflinger (approved):

pixelflinger (approved):

pixelflinger:
Thank you so much @daemyung !

PR #5078, opened 2022-01-21
Make morphable entities selectable in UI

daemyung:
Morphable entities can selectable and morphing is grayed out when playing animations.

스크린샷 2022-01-21 오전 11 03 31

### DIFF
--- a/libs/viewer/include/viewer/SimpleViewer.h
+++ b/libs/viewer/include/viewer/SimpleViewer.h
@@ -236,13 +236,15 @@ private:
     std::function<void()> mCustomUI;
 
     // Properties that can be changed from the UI.
-    int mCurrentAnimation = 1;
+    int mCurrentAnimation = 1; // It is a 1-based index and 0 means not playing animation
     bool mResetAnimation = true;
     bool mEnableWireframe = false;
     int mVsmMsaaSamplesLog2 = 1;
     Settings mSettings;
     int mSidebarWidth;
     uint32_t mFlags;
+    utils::Entity mCurrentMorphingEntity;
+    std::vector<float> mMorphWeights;
 
     // 0 is the default "free camera". Additional cameras come from the gltf file (1-based index).
     int mCurrentCamera = 0;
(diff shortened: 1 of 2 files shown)


pixelflinger (changes requested):
Is this updating the version of ImGui to a more recent one and is it required?

If so I would like this to be a separate, independent PR.

daemyung:
> Is this updating the version of ImGui to a more recent one and is it required?

Yes it's required. `ImGui::BeginDisabled()` and `ImGui::EndDisabled()` are needed to gray out. But if you want a seperate PR I can do that.

daemyung:
Do not merge until #5083 is merged.

pixelflinger (commented):
looks good to me, once it's rebased on top of the imgui update.

libs/viewer/include/viewer/SimpleViewer.h
```diff
@@ -236,13 +236,15 @@ class UTILS_PUBLIC SimpleViewer {
     std::function<void()> mCustomUI;
 
     // Properties that can be changed from the UI.
-    int mCurrentAnimation = 1;
+    int mCurrentAnimation = 0;
```
prideout, line 239: Personally I kind of like the auto-play feature that we had.  We could keep this set to `1` if we were to change the conditional as I've described in my subsequent comment.

By the way, we should probably have a comment here that explains that it is a 1-based index and 0 means "not playing animation".  That's my fault.

libs/viewer/src/SimpleViewer.cpp
```diff
@@ -980,6 +991,23 @@ void SimpleViewer::updateUserInterface() {
             ImGui::Unindent();
         }
 
+        if (mCurrentMorphingEntity && ImGui::CollapsingHeader("Morphing")) {
+            if (mCurrentAnimation != 0) {
```
prideout, line 995: Instead of testing current animation against 0 in multiple places, you could just say:

```
const bool isAnimation = mCurrentAnimation != && mAnimator->->getAnimationCount() > 0;
```

This would allow us to keep our auto-play feature.

prideout (changes requested):
Looks good, just a comment about auto-play.  Also needs a rebase like Mathias mentioned.

libs/viewer/include/viewer/SimpleViewer.h
daemyung: Ok it's reverted.

libs/viewer/src/SimpleViewer.cpp
daemyung: Ok it's updated.

pixelflinger (approved):

prideout (approved):

PR #5079, opened 2022-01-21
attempt to enable rough screen-space reflections

pixelflinger:


filament/src/PostProcessManager.h
```diff
@@ -144,6 +143,10 @@ class PostProcessManager {
     FrameGraphId<FrameGraphTexture> resolve(FrameGraph& fg,
             const char* outputBufferName, FrameGraphId<FrameGraphTexture> input) noexcept;
 
+    FrameGraphId<FrameGraphTexture> resolveNoCheck(FrameGraph& fg,
```
bejado, line 146: Add a comment describing what this new method does.. and yeah, I think I prefer `resolveBaseLevel`. At least, I can't think of anything better.

shaders/src/light_reflections.fs
```diff
@@ -251,7 +251,11 @@ vec4 evaluateScreenSpaceReflections(const PixelParams pixel, const vec3 wsRayDir
         float distanceFade = saturate(1.0 - (distFactor, distFactor));
         float fade = edgeFade * distanceFade;
 
-        Fr = vec4(textureLod(light_ssr, reprojected.xy, 0.0f).rgb, fade);
+        //float tweakedPerceptualRoughness = pixel.perceptualRoughness;
```
bejado, line 254: Remove

PR #5084, opened 2022-01-21
Improve (again) comments in Variants.h

pixelflinger:
- Improved formating
- Added more compile-time sanity checks
- Renamed Variant bits to 3 letters

### DIFF
--- a/libs/filabridge/include/private/filament/Variant.h
+++ b/libs/filabridge/include/private/filament/Variant.h
@@ -53,131 +53,117 @@ namespace filament {
         //
         // Standard variants:
         //                      +-----+-----+-----+-----+-----+-----+-----+-----+
-        //                      |  0  | VSM | FOG |  0  | SKN | SRE | DYN | DIR |    40 (-24)
+        //                      |  0  | VSM | FOG |  0  | SKN | SRE | DYN | DIR |    64 - 24 = 40
         //                      +-----+-----+-----+-----+-----+-----+-----+-----+
         //      Vertex shader            0     0     0     X     X     X     X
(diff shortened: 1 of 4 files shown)


libs/filabridge/include/private/filament/Variant.h
```diff
@@ -53,144 +53,133 @@ namespace filament {
...
-        //           Reserved            X     X     1     X     1     0     0
-        //           Reserved            X     X     1     X     1     0     1
-        //           Reserved            X     X     1     X     1     1     0
-        //           Reserved            X     X     1     X     1     1     1
+        //           Reserved            1     1     1     X     0     0     0     [ -2]
         //
-        // 46 variants used, 82 reserved
+        // 46 variants used, 82 reserved (128 - 46)
         //
+        // note: a valid variant can be neither a valid vertex nor a valid fragment variant
```
bejado, line 73: This note confuses me. Do you mean a "_reserved_ variant can be neither..."

pixelflinger: No, I meant that a valid variant (i.e. `!isReserved()`) can be invalid as both a fragment and vertex variant. This is counter intuitive and that's why I noted it here.

For e.g. a variant that as FOG|SKN is absolutely valid. But, it's not a valid fragment variant nor a valid vertex variant. This is because fragment variants don't allow SKN and vertex variants don't allow FOG.

bejado: I see. I would replace "can be" with "could potentially be".

I think the usage of "valid" to mean "not reserved" is confusing, since some "valid" variants will still get filtered out. Maybe we should add a note about that.

PR #5096, opened 2022-01-24
Fixed VulkanTexture::copyBufferToImage for 2D texture arrays

ondys:
2D array textures are initialized with .depth = 1 and layerCount = depth (see line 58 in VulkanTexture.cpp). By setting layerCount = 1 only the first layer was copied over, leaving the other layers uninitialized.


prideout (approved):

prideout:
Thanks!  Out of curiosity, how did you come across this bug?

ondys:
Just a personal project that uses texture arrays. I was trying to upload all layers of the texture array using filament::Texture::setImage() and it always updated only the first layer.

PR #5104, opened 2022-01-25
Create a dummy MorphTargetBuffer if skinning is enabled

### LINKED ISSUE #5085: Skinning breaks validation on Vulkan (unbound morphing sampler)
This error is seen with Vulkan but the fix will need to live in the Filament layer.  Note that Filament already binds dummy textures (e.g. a black reflection map) when a shader variant is activated that samples from a binding point that does not have a bound texture, even if it samples from it conditionally.

We can't correctly handle this in the backend because we don't know what type of image to create (cubemap, 3D, array).  That's why we currently see this error with skinned models:

```
Descriptor in binding #7 index 0 requires an image view of type VK_IMAGE_VIEW_TYPE_2D_ARRAY but got VK_IMAGE_VIEW_TYPE_2D.
```

daemyung:
Fixes #5085

### DIFF
--- a/libs/gltfio/src/MorphHelper.cpp
+++ b/libs/gltfio/src/MorphHelper.cpp
@@ -52,10 +52,11 @@ MorphHelper::MorphHelper(FFilamentAsset* asset, FFilamentInstance* inst) : mAsse
     for (auto pair : sourceNodes) {
         cgltf_node const* node = pair.first;
         cgltf_mesh const* mesh = node->mesh;
+        cgltf_skin const* skin = node->skin;
         if (mesh) {
             cgltf_primitive const* prims = mesh->primitives;
             for (cgltf_size pi = 0, count = mesh->primitives_count; pi < count; ++pi) {
-                addPrimitive(mesh, pi, pair.second);
+                addPrimitive(mesh, pi, skin->joints_count, pair.second);
             }
             addTargetNames(mesh, pair.second);
         }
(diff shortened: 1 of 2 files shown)


daemyung:
Nice to see you a fox again!
스크린샷 2022-01-25 오전 11 27 14

libs/gltfio/src/MorphHelper.cpp
```diff
@@ -52,10 +52,11 @@ MorphHelper::MorphHelper(FFilamentAsset* asset, FFilamentInstance* inst) : mAsse
     for (auto pair : sourceNodes) {
         cgltf_node const* node = pair.first;
         cgltf_mesh const* mesh = node->mesh;
+        cgltf_skin const* skin = node->skin;
         if (mesh) {
             cgltf_primitive const* prims = mesh->primitives;
             for (cgltf_size pi = 0, count = mesh->primitives_count; pi < count; ++pi) {
-                addPrimitive(mesh, pi, pair.second);
+                addPrimitive(mesh, pi, skin->joints_count, pair.second);
```
pixelflinger, line 59:
```suggestion
                addPrimitive(mesh, pi, (bool)skin->joints_count, pair.second);
```

pixelflinger (changes requested):
@daemyung sorry this is not the right approach. This logic should reside in filament, not in gltfio.

I will take a look at fixing this today. 

Thank you again!

pixelflinger:
I'm closing this PR because it is not the right approach, this logic should not be in `gltfio`.

daemyung:
@pixelflinger Yes it's just W/A. I don't know where is the best place to create dummy resource.

PR #5110, opened 2022-01-25
Fix typo in BlueVK script that caused segfault.

### LINKED ISSUE #5026: vkGetInstanceProcAddr should not be initialized again
https://github.com/google/filament/blob/2f3aac29da7e833378d75dd89fadcfc4bcb11623/libs/bluevk/src/BlueVK.cpp#L360
vkGetInstanceProcAddr should not be initialized again. It was initialized in function bluevk::initialize() and got NULL after reinitalizing. THEN CRASH!!!

prideout:
Fixes #5026.


romainguy:
Ooops, good catch :)

PR #5113, opened 2022-01-25
Simplify and optimize morphing

pixelflinger:
There is no need to go through all primitive and call updateSamplerGroup,
because the SamplerGroup never changes -- it just hold the texture that
holds the data.


pixelflinger:
@daemyung FYI, I think we can do this simplification.

prideout (approved):

romainguy (approved):

daemyung:
> @daemyung FYI, I think we can do this simplification.

That's good point! It's legacy because `Texture` could be changed at the first implementation.

PR #5114, opened 2022-01-25
Vulkan: clean up image layout management.

prideout:
This fixes validation errors and makes a first pass at simplification.
VulkanTexture now tracks image layout using RangeMap, which paves the
way for further simplification.

### DIFF
--- a/filament/backend/src/vulkan/VulkanDriver.cpp
+++ b/filament/backend/src/vulkan/VulkanDriver.cpp
@@ -1412,8 +1412,8 @@ void VulkanDriver::stopCapture(int) {
 void VulkanDriver::readPixels(Handle<HwRenderTarget> src, uint32_t x, uint32_t y,
         uint32_t width, uint32_t height, PixelBufferDescriptor&& pbd) {
     const VkDevice device = mContext.device;
-    const VulkanRenderTarget* srcTarget = handle_cast<VulkanRenderTarget*>(src);
-    const VulkanTexture* srcTexture = srcTarget->getColor(mContext.currentSurface, 0).texture;
+    VulkanRenderTarget* srcTarget = handle_cast<VulkanRenderTarget*>(src);
+    VulkanTexture* srcTexture = srcTarget->getColor(mContext.currentSurface, 0).texture;
     const VkFormat srcFormat = srcTexture ? srcTexture->getVkFormat() :
             mContext.currentSurface->surfaceFormat.format;
     const bool swizzle = srcFormat == VK_FORMAT_B8G8R8A8_UNORM;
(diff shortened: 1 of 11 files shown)


filament/backend/src/vulkan/VulkanDriver.cpp
```diff
@@ -1510,11 +1509,10 @@ void VulkanDriver::readPixels(Handle<HwRenderTarget> src, uint32_t x, uint32_t y
         .layerCount = 1,
     };
 
-    // FIXME: the content of the source may be destroyed because of VK_IMAGE_LAYOUT_UNDEFINED
-    VkImage srcImage = srcTarget->getColor(mContext.currentSurface, 0).image;
+    VkImage srcImage = srcAttachment.image;
     transitionImageLayout(cmdbuffer, {
         .image = srcImage,
-        .oldLayout = VK_IMAGE_LAYOUT_UNDEFINED,
+        .oldLayout = VK_IMAGE_LAYOUT_COLOR_ATTACHMENT_OPTIMAL,
```
pixelflinger, line 1515: so how do we know that it's OPTIMAL here?

filament/backend/src/vulkan/VulkanDriver.cpp
```diff
@@ -1526,30 +1524,19 @@ void VulkanDriver::readPixels(Handle<HwRenderTarget> src, uint32_t x, uint32_t y
     // Perform the blit.
 
     vkCmdCopyImage(cmdbuffer, srcTarget->getColor(mContext.currentSurface, 0).image,
-            VK_IMAGE_LAYOUT_TRANSFER_SRC_OPTIMAL, stagingImage, VK_IMAGE_LAYOUT_TRANSFER_DST_OPTIMAL,
+            VK_IMAGE_LAYOUT_TRANSFER_SRC_OPTIMAL, stagingImage, VK_IMAGE_LAYOUT_GENERAL,
```
pixelflinger, line 1527: Personally, I would like a comment everywhere there is a hardcoded LAYOUT constant, explaining how/why that constant was chosen.

pixelflinger (approved):
I like it a lot!

filament/backend/src/vulkan/VulkanDriver.cpp
prideout: Added a comment

filament/backend/src/vulkan/VulkanDriver.cpp
prideout: I agree, added a comment here but I would like to do another pass in a future PR and make it even more clear.

Side note: another thing I don't like is that there is a lot of special case code to handle situations that involve the SwapChain (as opposed to a "normal" Texture), so I am considering adding a proxy VulkanTexture to VulkanSwapChain.  This might also allow us to consolidate all the layout logic into the VulkanTexture wrapper, rather than sprinkling it throughout the codebase.

Issue #5109, opened 2022-01-25
Horse morph animation looks incorrect with Vulkan


prideout:
This has 15 morph targets and it worked prior to texture-based morphing.  Looks fine on Metal and OpenGL.

https://github.com/mrdoob/three.js/blob/dev/examples/models/gltf/Horse.glb

This error occurs with release configuration, it is not a validation issue and is therefore separate from #5085.

prideout:
In debug builds, these models trigger the following assert.

```
reason: ../../filament/backend/src/vulkan/VulkanDriver.cpp:882: failed assertion `xoffset == 0 && yoffset == 0 && zoffset == 0 && "Offsets not yet supported."'
```

prideout:
These models are now fixed.

PR #5115, opened 2022-01-25
Vulkan now supports offsets when uploading texture data.

prideout:
This allows `MorphStressTest` to work on Vulkan.

However, `Horse` is still broken because it provides positions but not
tangents.  Separate fix for that is coming.

Partial fix for #5109.

### DIFF
--- a/filament/backend/src/vulkan/VulkanTexture.cpp
+++ b/filament/backend/src/vulkan/VulkanTexture.cpp
@@ -195,13 +195,8 @@ VulkanTexture::~VulkanTexture() {
     }
 }
 
-void VulkanTexture::update2DImage(const PixelBufferDescriptor& data, uint32_t width,
-        uint32_t height, int miplevel) {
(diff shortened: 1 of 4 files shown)


filament/backend/src/vulkan/VulkanTexture.cpp
```diff
@@ -213,41 +208,67 @@ void VulkanTexture::update3DImage(const PixelBufferDescriptor& data, uint32_t wi
...
+    VkImageSubresourceRange transitionRange = {
+        .aspectMask = mAspect,
+        .baseMipLevel = miplevel,
+        .levelCount = 1,
+        .baseArrayLayer = 0,
+        .layerCount = 1
+    };
 
-    transitionLayout(cmdbuffer, range, textureLayout);
+    // Vulkan specifies subregions for 3D textures differently than from 2D arrays.
```
pixelflinger, line 253: of course it does! :-)

pixelflinger (approved):

PR #5119, opened 2022-01-26
fix build android aar on Linux because of case sensitive

HanYunChenLuo:
"#include <android/Thermal.h>" will cause some errors when building aar on Linux,
the corrret answer is "#include <android/thermal.h>"


pixelflinger (approved):
Thank you for catching this.

HanYunChenLuo:
Sorry, maybe I click update branch button in a mistake, neecd to be merged manully.

PR #5120, opened 2022-01-26
WebGL: Support compile filament.wasm on Linux platform.

HanYunChenLuo:
Now， Filament WebGL could  be compiled on MacOS. I add somethings in webgl/common-ci.sh scripts to support linux platform compile.


prideout (approved):
Looks ok to me if it looks ok to others.

pixelflinger (approved):

PR #5122, opened 2022-01-26
gltfio morphing: generate normals when they are missing.

prideout:
This allows the `Horse.glb` model to work with Vulkan without any
validation errors. As a reminder, the glTF spec says:

```
When normals are not specified, client implementations MUST calculate
flat normals
```

Fixes #5115.

### DIFF
--- a/libs/gltfio/src/MorphHelper.cpp
+++ b/libs/gltfio/src/MorphHelper.cpp
@@ -119,6 +119,7 @@ void MorphHelper::addPrimitive(cgltf_mesh const* mesh, int primitiveIndex, Entit
     const cgltf_accessor* previous = nullptr;
     for (int targetIndex = 0; targetIndex < prim.targets_count; targetIndex++) {
         const cgltf_morph_target& morphTarget = prim.targets[targetIndex];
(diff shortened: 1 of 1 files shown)


romainguy (approved):

libs/gltfio/src/MorphHelper.cpp
```diff
@@ -162,6 +164,18 @@ void MorphHelper::addPrimitive(cgltf_mesh const* mesh, int primitiveIndex, Entit
...
+
+        // Generate flat normals if necessary.
+        if (!hasNormals && !prim.material->unlit) {
+            TangentsJob job;
+            TangentsJob::Params params = { .in = { &prim, targetIndex } };
+            TangentsJob::run(&params);
+            if (params.out.results) {
+                morphHelperPrim.targets->setTangentsAt(engine, targetIndex,
+                        params.out.results, params.out.vertexCount);
+                free(params.out.results);
```
pixelflinger, line 176: It would be nicer to have `TangentsJob::free(&params)` to not have this "unbalanced" call to free.

pixelflinger (approved):

Issue #5111, opened 2022-01-25
Headless rendering issue with Metal and Vulkan


prideout:
Running gltf_viewer in automated testing mode crashes on my MacBook.  Try running it with `--batch=default --headless`

For Vulkan, there are existing image layout issues that I'm already working through which need to be fixed first.

For Metal, the call stack seems to indicate a problem deep inside our readPixels invocation:

```
  * frame #0: 0x00007ffa21a10332 AMDRadeonX6000MTLDriver`GFX10_Tile_2D_D_BPP4_ReadVec(void*, void*, unsigned int) + 8
    frame #1: 0x00007ffa21a0f6f9 AMDRadeonX6000MTLDriver`GFX10_GenerateTileAddresses(CpuTexInterfaceRec const*, ATIMipmapBufferHeader const*, unsigned int const*, unsigned int, unsigned int, unsigned int, unsigned int, unsigned int, unsigned int, unsigned int, unsigned int, unsigned int, void (*)(unsigned long long, unsigned int, unsigned int, void*), void*) + 3076
    frame #2: 0x00007ffa21a0e4f5 AMDRadeonX6000MTLDriver`GFX10_TextureReadTiles(CpuTexInterfaceRec const*, ATIMipmapBufferHeader const*, unsigned int const*, amd_gfx10_format_info const*, unsigned int, unsigned int, unsigned long, unsigned int, unsigned long, unsigned long, unsigned int, unsigned int, unsigned int, unsigned int, unsigned int) + 293
[...]

PR #5125, opened 2022-01-26
gltf_viewer: fix bad ReadPixels size in headless mode

prideout:
This fixes the gltf_viewer automated testing mode, invoked via
`--batch=default --headless`.

ReadPixels was being issued with an out-of-bounds size on high DPI
platforms, which caused weird backend errors.

Since we create an invisible SDL window while in headless mode, the
presence of the SDL window does *not* indicate headless vs non-headless.

Partial fix for #5111


romainguy (approved):

pixelflinger (approved):
Nice

PR #5131, opened 2022-01-28
fix an out-of-bound access when setting morphing buffer

pixelflinger:
Because the morphing buffer is actually a texture, it's possible that
the last line is not a full-width line when updating the buffer.

Because textures are updated via a rectangle, it was possible to read
past the malloc'ed buffer during the update.

Moreover, because the count can be specified, we can't just blindly send
more "dummy" data. 

Finally, added an offset to match the skinning API, and in that case, 
the first line can also be incomplete.

We fix this by issuing up to 3 texture update calls if needed.

The java/js API update will come later.

### DIFF
--- a/filament/include/filament/MorphTargetBuffer.h
+++ b/filament/include/filament/MorphTargetBuffer.h
@@ -88,7 +88,8 @@ public:
      * @param count number of position elements in positions
      * @see setTangentsAt
      */
-    void setPositionsAt(Engine& engine, size_t targetIndex, math::float3 const* positions, size_t count);
+    void setPositionsAt(Engine& engine, size_t targetIndex,
+            math::float3 const* positions, size_t count, size_t offset = 0);
 
     /**
      * Updates the position of morph target at the index.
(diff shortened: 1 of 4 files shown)


pixelflinger:
@daemyung you might be interested in having a look at this one.

daemyung (approved):
LGTM.

BTW is there any performance regression doing copy 3 times?

pixelflinger:
> LGTM.
> 
> BTW is there any performance regression doing copy 3 times? I know there is no way doing this unless we keep data or use `glMapBufferRange`.
> 
> I think MorphTargetBuffer will be modified very rarely so... using device local memory is best and using `glMapBufferRange` will be improper. However filament is mobile first engine and mobile use unified memory so there is no different between host and device memory so... using `glMapBufferRange` can be fine.

Partial texture uploads are never good, for sure.

For `glMapBufferRange` I don't think you can do that with a texture? 
I don't see how it would be efficient because map/unmap would have to do the swizzling.
What you can do is use a PBO and map that, but you'd still have the same number of calls to `glTexImage2D`.

romainguy (approved):

Issue #5142, opened 2022-01-31
VulkanPipelineCache Regression: intermittent seg fault and validation errors


prideout:
Clicking `Run batch test` in a debug build of `gltf_viewer` now results in intermittent SEGV if you leave "Export Screenshots" unchecked.  (Exporting screenshots triggers a known validation error that I'm already working on.)

Sometimes, before the segfault, I see this in the console:

```
Cannot call vkUpdateDescriptorSets() to perform write update on VkDescriptorSet [...]
allocated with VkDescriptorSetLayout [...] which has been destroyed.
```

With MoltenVK, sometimes the SEGV occurs in `MVKPipelineLayout::bindDescriptorSets`, and sometimes it occurs in `vkUpdateDescriptorSets`.

These errors started with 41cf50774.  cc @daemyung

PR #5144, opened 2022-02-01
Fix Vulkan crash regression

prideout:
This PR has two commits that should not be squashed.  The first commit is a cleanup pass.  The second commit contains the actual fix for #5142.


bejado (approved):
LGTM. Please also cherry-pick into rc/1.18.0

prideout:
> LGTM. Please also cherry-pick into rc/1.18.0

Done!

daemyung (commented):
So nice! 👍

PR #5146, opened 2022-02-02
Fix mismatching sampler binding indices between Program and BindingIndexMap

### LINKED ISSUE #5139: samples/heightfield doesn't work on Metal
Tried a debug build, when launching  `samples/heightfield   -a metal` it just doesn't draw anything. This is on M1 macmini.

This works on with opengl and vulkan/molten.

daemyung:
BindingIndexMap is generated with the dummy variant but Program isn't. It
causes mismatching sampler binding indices because the morphing sampler group
is added always with the dummy variant but it can be added or not depending
on the variant on Program.

Fixes #5139


pixelflinger (approved):
seems reasonable to me.

romainguy (approved):

bejado (approved):

PR #5151, opened 2022-02-03
Deprecate MorphHelper

daemyung:
These changes bring the loading performance optimization by removing flush and wait.


daemyung:
This [model](https://github.com/mrdoob/three.js/issues/21636#issuecomment-893520036) and below code are used to measure the performance.
```
auto begin = std::chrono::system_clock::now();
app.resourceLoader->asyncBeginLoad(app.asset);
app.engine->flushAndWait();
auto end = std::chrono::system_clock::now();
std::cout << "compute: " << std::chrono::duration_cast<std::chrono::nanoseconds>(end - begin).count() << std::endl;
```

The loading performance is enhanced around 3.7x(from 531338666 ns to 143684666 ns) on Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz.

filament/src/components/RenderableManager.cpp
```diff
@@ -783,6 +783,12 @@ void RenderableManager::setMorphTargetBufferAt(Instance instance, uint8_t level,
             upcast(morphTargetBuffer), 0, count);
 }
 
+void RenderableManager::setMorphTargetBufferAt(Instance instance, uint8_t level, size_t primitiveIndex,
+        MorphTargetBuffer* morphTargetBuffer) {
+    upcast(this)->setMorphTargetBufferAt(instance, level, primitiveIndex,
+            upcast(morphTargetBuffer), 0, morphTargetBuffer->getVertexCount());
+}
+
```
pixelflinger, line 791: We have a rule that those "trampolines" call the real implementation with the same exact parameters. So you need to create a version without 'count' in the `FRenderableManager`, and have that one call the version with the count.

filament/include/filament/RenderableManager.h
```diff
@@ -488,6 +488,9 @@ class UTILS_PUBLIC RenderableManager : public FilamentAPI {
     void setMorphTargetBufferAt(Instance instance, uint8_t level, size_t primitiveIndex,
             MorphTargetBuffer* morphTargetBuffer, size_t count); //!< \overload
 
+    void setMorphTargetBufferAt(Instance instance, uint8_t level, size_t primitiveIndex,
+            MorphTargetBuffer* morphTargetBuffer); //!< \overload
+
```
pixelflinger, line 493: This needs a java version.
Maybe it would make more sense to remove the overload with only a count. Also I think this API change should be in a separate PR.

romainguy (commented):
This change will break the Android builds. You need to update the CMake files in android/gltfio-android I believe (and maybe the JNI?)

filament/include/filament/RenderableManager.h
```diff
@@ -488,6 +488,9 @@ class UTILS_PUBLIC RenderableManager : public FilamentAPI {
     void setMorphTargetBufferAt(Instance instance, uint8_t level, size_t primitiveIndex,
             MorphTargetBuffer* morphTargetBuffer, size_t count); //!< \overload
 
+    void setMorphTargetBufferAt(Instance instance, uint8_t level, size_t primitiveIndex,
```
romainguy, line 491: This needs to be added to the Java APIs as well.

pixelflinger (changes requested):

libs/gltfio/src/ResourceLoader.cpp
```diff
@@ -503,17 +503,28 @@ bool ResourceLoader::loadResources(FFilamentAsset* asset, bool async) {
             slot.vertexBuffer->setBufferObjectAt(engine, slot.bufferIndex, bo);
             continue;
         }
-        assert(slot.indexBuffer);
-        if (accessor->component_type == cgltf_component_type_r_8u) {
-            const size_t size16 = size * 2;
-            uint16_t* data16 = (uint16_t*) malloc(size16);
-            convertBytesToShorts(data16, data, size);
-            IndexBuffer::BufferDescriptor bd(data16, size16, FREE_CALLBACK);
+        else if (slot.indexBuffer) {
```
bejado, line 506: nit: move the `else if` so it's on the same line as the `}` (or just delete the `else`)

filament/include/filament/RenderableManager.h
daemyung: Ok I see. I will revert this change now and will prepare a new PR for this.

filament/include/filament/RenderableManager.h
daemyung: Yes I will create a new PR for these changes.

libs/gltfio/src/ResourceLoader.cpp
daemyung: Thanks it's fixed. BTW what is nit?

romainguy: nit is short for [nitpick](https://www.google.com/search?q=nitpick&oq=nitpick&aqs=chrome..69i57j69i59l3j69i60l4.742j0j7&sourceid=chrome&ie=UTF-8)

daemyung:
> This change will break the Android builds. You need to update the CMake files in android/gltfio-android I believe (and maybe the JNI?)

@romainguy Yes it's updated.

daemyung:
The loading performance is enhanced around 2.7X(from 292623567 ns to 107575567 ns) on Samsung Exynos 2100 2.91 GHz. But there is no performance enhancement on iOS 7 Plus. I guess iOS 7 Plus has few CPU cores.

pixelflinger (approved):

pixelflinger (approved):

prideout (approved):

PR #5153, opened 2022-02-03
Rewrite VulkanPipelineCache (without changing its API).

prideout:
All three types of internal caches (descriptor sets, pipelines, and pipeline
layouts) are now managed in exactly the same way. They all use an LRU
eviction scheme that is based on a count of command buffer flush
events.

Vulkan objects can only be destroyed if there are no in-flight command
buffers that reference them, so an easy way to know when it is safe to
evict a given entry is to wait for "N" flushes after its last use, where
"N" is the number of command buffers in the command buffer ring.

Another big simplification is that there are no more dirty flags,
instead we store two sets of state vectors for each type of cache: the
"currently bound" state, and the "current requirements" state.

### DIFF
--- a/filament/backend/src/vulkan/VulkanPipelineCache.cpp
+++ b/filament/backend/src/vulkan/VulkanPipelineCache.cpp
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2018 The Android Open Source Project
+ * Copyright (C) 2022 The Android Open Source Project
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -19,7 +19,6 @@
 
 #include <utils/Log.h>
 #include <utils/Panic.h>
-#include <utils/trap.h>
 
 #include "VulkanConstants.h"
 #include "VulkanHandles.h"
@@ -66,11 +65,6 @@ static getPipelineLayoutKey(const Program::SamplerGroupInfo& samplerGroupInfo) n
 }
 
 VulkanPipelineCache::VulkanPipelineCache() : mDefaultRasterState(createDefaultRasterState()) {
-    markDirtyDescriptor();
-    markDirtyPipeline();
-    mDescriptorKey = {};
-    mPipelineKey = {};
-
     mDummyBufferWriteInfo.sType = VK_STRUCTURE_TYPE_WRITE_DESCRIPTOR_SET;
     mDummyBufferWriteInfo.pNext = nullptr;
     mDummyBufferWriteInfo.dstArrayElement = 0;
@@ -110,11 +104,10 @@ void VulkanPipelineCache::setDevice(VkDevice device, VmaAllocator allocator) {
     mAllocator = allocator;
     mDescriptorPool = createDescriptorPool(mDescriptorPoolSize);
 
-    // Next, create a small dummy UBO. Filament's fixed-sized binding arrays are nice because they
-    // allow us to track only 1 VkPipelineLayout object. However, we occasionally need to clear
-    // out an unused descriptor set slot. Since Vulkan does not allow specifying VK_NULL_HANDLE
-    // without the robustness2 extension, we are forced to clear out unused UBO slots using a dummy
-    // resource.
+    // Formulate some dummy objects and dummy descriptor info used only for clearing out unused
+    // bindings. This is especially crucial after a texture has been destroyed. Since core Vulkan
+    // does not allow specifying VK_NULL_HANDLE without the robustness2 extension, we are forced to
+    // bind dummy objects.
 
     VkBufferCreateInfo bufferInfo {
         .sType = VK_STRUCTURE_TYPE_BUFFER_CREATE_INFO,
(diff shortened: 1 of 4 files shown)


filament/backend/src/vulkan/VulkanPipelineCache.h
```diff
@@ -261,77 +263,90 @@ class VulkanPipelineCache : public CommandBufferObserver {
...
+    // being used by the GPU, and is therefore safe to destroy or reclaim.
+    using Timestamp = uint64_t;
+    Timestamp mCurrentTime = 0;
 
-    struct PipelineVal {
-        VkPipeline handle;
+    // The descriptor set cache entry is a group of descriptor sets that are bound simultaneously.
+    struct DescriptorCacheEntry {
+        std::array<VkDescriptorSet, DESCRIPTOR_TYPE_COUNT> handles;
+        Timestamp last_used;
```
pixelflinger, line 279: should be `lastUsed`

filament/backend/src/vulkan/VulkanPipelineCache.h
```diff
@@ -261,77 +263,90 @@ class VulkanPipelineCache : public CommandBufferObserver {
...
-        // which is a crucial feature because VkPipeline construction is very slow.
-        uint32_t age;
     };
 
-    using LayoutMap = tsl::robin_map<PipelineLayoutKey , LayoutBundle, LayoutKeyHashFn, LayoutKeyEqual>;
-    using PipelineMap = tsl::robin_map<PipelineKey, PipelineVal, PipelineHashFn, PipelineEqual>;
-    using DescriptorMap = tsl::robin_map<DescriptorKey, DescriptorBundle, DescHashFn, DescEqual>;
+    struct PipelineCacheEntry {
+        VkPipeline handle;
+        Timestamp last_used;
```
pixelflinger, line 285: `lastUsed`

filament/backend/src/vulkan/VulkanPipelineCache.h
```diff
@@ -261,77 +263,90 @@ class VulkanPipelineCache : public CommandBufferObserver {
...
+    };
 
-    struct CmdBufferState {
-        PipelineVal* currentPipeline = nullptr;
-        DescriptorBundle* currentDescriptorBundle = nullptr;
-        VkRect2D scissor = {};
+    struct PipelineLayoutCacheEntry {
+        VkPipelineLayout handle;
+        Timestamp last_used;
```
pixelflinger, line 290: `lastUsed`

filament/backend/src/vulkan/VulkanPipelineCache.h
```diff
@@ -261,77 +263,90 @@ class VulkanPipelineCache : public CommandBufferObserver {
...
 
-    LayoutMap mLayouts;
-    PipelineMap mPipelines;
-    DescriptorMap mDescriptorBundles;
-    uint32_t mCmdBufferIndex = 0;
+    // Current state for scissoring.
+    VkRect2D mCurrentScissor = {};
 
     VkDescriptorPool mDescriptorPool;
     uint32_t mDescriptorPoolSize = 500;
```
pixelflinger, line 352: maybe use a define? Also why not 512?

pixelflinger (approved):
wow it's super clean!

daemyung (approved):

filament/backend/src/vulkan/VulkanPipelineCache.cpp
```diff
@@ -154,109 +142,99 @@ void VulkanPipelineCache::setDevice(VkDevice device, VmaAllocator allocator) {
...
+        descriptorIter.value().last_used = mCurrentTime;
+        return true;
     }
-    if (bind) {
-        LayoutBundle* layout = getOrCreatePipelineLayout();
-        vkCmdBindDescriptorSets(cmdbuffer, VK_PIPELINE_BIND_POINT_GRAPHICS,
-                layout->pipelineLayout, 0, VulkanPipelineCache::DESCRIPTOR_TYPE_COUNT,
-                descriptors, 0, nullptr);
+
+    // If a cached object exists, re-use it, otherwise a create a new one.
```
bejado, line 154:
```suggestion
    // If a cached object exists, re-use it, otherwise create a new one.
```

filament/backend/src/vulkan/VulkanPipelineCache.cpp
```diff
@@ -154,109 +142,99 @@ void VulkanPipelineCache::setDevice(VkDevice device, VmaAllocator allocator) {
...
+    if (PipelineEqual equals; UTILS_LIKELY(equals(mBoundPipeline, mPipelineRequirements))) {
+        assert_invariant(pipelineIter != mPipelines.end());
+        pipelineIter.value().last_used = mCurrentTime;
+        return;
     }
+
+    // If a cached object exists, re-use it, otherwise a create a new one.
+    PipelineCacheEntry* cacheEntry = UTILS_LIKELY(pipelineIter != mPipelines.end()) ?
+            &pipelineIter.value() : createPipeline();
+
```
bejado, line 186: Do we need to check if `cacheEntry` is `nullptr` here (like you did in `bindDescriptors`)? Likewise maybe `bindPipeline` should also return a boolean.

daemyung: `createPipeline` must not be failed as I think.

bejado: `createPipeline` returns `nullptr` if `vkCreateGraphicsPipelines` fails, though I suppose we also have an assert for this case.

PR #5157, opened 2022-02-03
Minor cleanups

pixelflinger:
- Rename projectToPixelMatrix to uvFromViewMatrix, this match our 
existing style better.

- slightly more efficient direction transform

- limit roughness to 0.1 instead of 0.01, because in a lot of cases, 
the roughness increases with the LOD, for instance if it comes from
a texture (to combat aliasing).

### DIFF
--- a/shaders/src/light_indirect.fs
+++ b/shaders/src/light_indirect.fs
@@ -541,7 +541,7 @@ vec3 evaluateRefraction(
     vec3 Ft;
 
     // compute the point where the ray exits the medium, if needed
-    vec4 p = vec4(frameUniforms.clipFromWorldMatrix * vec4(ray.position, 1.0));
+    vec4 p = vec4(getClipFromWorldMatrix() * vec4(ray.position, 1.0));
     p.xy = uvToRenderTargetUV(p.xy * (0.5 / p.w) + 0.5);
 
     // perceptualRoughness to LOD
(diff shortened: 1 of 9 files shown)


romainguy:
I don't see a change related to roughness clamping?

bejado (approved):

romainguy:
Never mind it showed up after your new push.

shaders/src/light_indirect.fs
```diff
@@ -580,7 +580,7 @@ void evaluateIBL(const MaterialInputs material, const PixelParams pixel, inout v
     vec4 Fssr = vec4(0.0f);
     // evaluateScreenSpaceReflections will set the value of ssr if there's a hit.
     // ssr.a contains the reflection's contribution.
-    if (pixel.roughness <= 0.01f && frameUniforms.ssrDistance > 0.0f) {
+    if (pixel.roughness <= 0.1f && frameUniforms.ssrDistance > 0.0f) {
```
romainguy, line 583: Could we introduce a #define for this? MAX_SSR_ROUGHNESS or something.

romainguy (approved):

shaders/src/light_indirect.fs
pixelflinger: no because it's going away soon.

PR #5160, opened 2022-02-04
Update RELEASE_NOTES.md

daemyung:


pixelflinger (approved):

prideout (commented):
I would add a few more words since this is a new API, e.g.

"engine: Add MorphTargetBuffer API for texture-based morphing, up to 256 targets."

prideout (approved):
Never mind, I was thinking of the previous release and that ship has sailed.

PR #5163, opened 2022-02-04
Add a MAX_LEGACY_MORPH_TARGETS definition

bejado:
This is useful for clients who previously relied on `MAX_MORPH_TARGETS` being 4.


pixelflinger (approved):
okay, but at some point all the "legacy" morphtarget stuff will be deprecated, right?

bejado:
> okay, but at some point all the "legacy" morphtarget stuff will be deprecated, right?

Yes, that's the goal.

PR #5168, opened 2022-02-05
Enable rough screen-space reflections

pixelflinger:
This works by first generating a reflection buffer which gets blurred,
then the color pass samples from this buffer according to the roughness
of the surface being rendered.

A lot of the changes here involve utilizing "reserved" variants for
the new "SSR" pass and all the fallout from that.

### DIFF
--- a/filament/src/Material.cpp
+++ b/filament/src/Material.cpp
@@ -221,6 +221,10 @@ FMaterial::FMaterial(FEngine& engine, const Material::Builder& builder)
             mRasterState.depthWrite = true;
             break;
         case BlendingMode::MASKED:
+            // MASKED mode now leaves destination alpha intact.
+            // This prevents strange behavior with semi-transparent render targets, which the
+            // model viewer team discovered when testing against the Khronos alpha test
+            // conformance model.
             mRasterState.blendFunctionSrcRGB   = BlendFunction::ONE;
             mRasterState.blendFunctionSrcAlpha = BlendFunction::ZERO;
             mRasterState.blendFunctionDstRGB   = BlendFunction::ZERO;
--- a/filament/src/PostProcessManager.cpp
+++ b/filament/src/PostProcessManager.cpp
@@ -30,6 +30,7 @@
 #include "fg2/FrameGraphResources.h"
 
 #include "fsr.h"
+#include "PerViewUniforms.h"
 #include "RenderPass.h"
 
 #include "details/Camera.h"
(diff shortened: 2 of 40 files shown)


filament/src/Material.cpp
```diff
@@ -221,6 +221,10 @@ FMaterial::FMaterial(FEngine& engine, const Material::Builder& builder)
             mRasterState.depthWrite = true;
             break;
         case BlendingMode::MASKED:
+            // MASKED mode now leaves destination alpha intact.
```
romainguy, line 224: Thanks for adding this comment!

filament/src/RenderPass.cpp
```diff
@@ -199,6 +199,12 @@ void RenderPass::setupColorCommand(Command& cmdDraw,
 
     cmdDraw.key = isBlendingCommand ? keyBlending : keyDraw;
     cmdDraw.primitive.rasterState = ma->getRasterState();
+    if (!isBlendingCommand && Variant::isSSRVariant(variant)) {
```
romainguy, line 202: Is the compiler smart enough to get rid of the branch? Should we instead put the condition in the assignment with bit masking?

filament/src/RenderPass.cpp
```diff
@@ -348,7 +354,9 @@ void RenderPass::generateCommandsImpl(uint32_t extraFlags,
 
         cmdColor.key = makeField(soaVisibility[i].priority, PRIORITY_MASK, PRIORITY_SHIFT);
         cmdColor.primitive.index = (uint16_t)i;
-        variant.setShadowReceiver(soaVisibility[i].receiveShadows & hasShadowing);
+        if (!Variant::isSSRVariant(variant)) {
```
romainguy, line 357: Same thing here? Is the compiler doing the right thing?

shaders/src/light_indirect.fs
```diff
@@ -575,53 +551,61 @@ void evaluateIBL(const MaterialInputs material, const PixelParams pixel, inout v
...
-        Fssr = evaluateScreenSpaceReflections(pixel, r);
+#if defined(MATERIAL_HAS_REFLECTIONS)
+    vec4 ssrFr = vec4(0.0f);
+#if defined(BLEND_MODE_OPAQUE) || defined(BLEND_MODE_MASKED)
+    // do the uniform based test first
+    if (frameUniforms.ssrDistance > 0.0f) {
+        // There is no point doing SSR for very high roughness because we're limited by the fov
+        // of the screen, in addition it doesn't really add much to the final image.
+        // TODO: maybe make this a parameter
+        const float maxPerceptualRoughness = sqrt(0.5);
```
romainguy, line 568: Just hardcode 0.125 in case the shader compiler doesn't inline the sqrt()? (and leave a comment about where the value comes from)

romainguy (approved):

shaders/src/light_indirect.fs
pixelflinger: because it's `const`, which really means `constexpr`, it has no choice but to inline `sqrt`.

filament/src/RenderPass.cpp
pixelflinger: okay, made it more branchless.

filament/src/RenderPass.cpp
pixelflinger: okay, made it more branchless.

libs/filabridge/include/private/filament/Variant.h
```diff
@@ -206,85 +212,15 @@ struct Variant {
...
+
 private:
     void set(bool v, type_t mask) noexcept {
         key = (key & ~mask) | (v ? mask : type_t(0));
     }
 };
 
-namespace details {
-
-// compile time sanity-check tests
```
bejado, line 217: Could you move `isFragmentVariant` and `isVertexVariant` to the cpp as well, since they're only called by these compile-time checks?

libs/filabridge/include/private/filament/Variant.h
```diff
@@ -206,85 +212,15 @@ struct Variant {
         return Variant(key & rhs);
     }
 
+    static Variant filterVariant(
```
bejado, line 248: I think it's confusing that we now have two `filterVariant` methods. Can you add a comment describing what each one does and/or rename one of them?

bejado (changes requested):

libs/filabridge/include/private/filament/Variant.h
pixelflinger: well they are for the moment, but I think they could be part of the Variant API.

libs/filabridge/include/private/filament/Variant.h
pixelflinger: ah yes... good point.

filament/src/PostProcessManager.cpp
```diff
@@ -437,6 +438,111 @@ FrameGraphId<FrameGraphTexture> PostProcessManager::structure(FrameGraph& fg,
...
+                FrameGraphTexture::Usage::SAMPLEABLE, entry.ssr.color);
+        historyProjection = entry.ssr.projection;
+    }
+
+    auto const& uvFromClipMatrix = mEngine.getUvFromClipMatrix();
+
+    auto& ssrPass = fg.addPass<SSRPassData>("SSR Pass",
+            [&](FrameGraph::Builder& builder, auto& data) {
+
+                // create our reflection buffer. We need an alpha chanel, so we have to use RGBA16F
```
bejado, line 480:
```suggestion
                // create our reflection buffer. We need an alpha channel, so we have to use RGBA16F
```

filament/src/PostProcessManager.cpp
```diff
@@ -437,6 +438,111 @@ FrameGraphId<FrameGraphTexture> PostProcessManager::structure(FrameGraph& fg,
...
+                        .width = desc.width, .height = desc.height,
+                        .format = TextureFormat::DEPTH32F });
+
+                // we're writing to both these buffers
+                data.reflections = builder.write(data.reflections,
+                        FrameGraphTexture::Usage::COLOR_ATTACHMENT);
+                data.depth = builder.write(data.depth,
+                        FrameGraphTexture::Usage::DEPTH_ATTACHMENT);
+
+                // finaly declare our render target
```
bejado, line 496:
```suggestion
                // finally declare our render target
```

shaders/src/shading_reflections.fs
```diff
@@ -0,0 +1,23 @@
+/*
+ * screen-space reflection shading
+ */
+vec4 evaluateMaterial(const MaterialInputs material) {
+
+#if defined(MATERIAL_HAS_REFLECTIONS)
+    PixelParams pixel;
+    //getAnisotropyPixelParams(inputs, pixel);
```
bejado, line 8: Remove?

shaders/src/shading_reflections.fs
```diff
@@ -0,0 +1,23 @@
...
+    vec4 color = vec4(0.0);
+    if (frameUniforms.ssrDistance > 0.0) {
```
bejado, line 10: Do we still need to check `ssrDistance` here? Seems that it's redundant now that SSR is its own pass.

shaders/src/shading_reflections.fs
```diff
@@ -0,0 +1,23 @@
...
+        vec3 r = getReflectedVector(pixel, shading_view, shading_normal);
+        // evaluateScreenSpaceReflections will set the value of color if there's a hit.
+        // color.a contains the reflection's contribution.
+        color = evaluateScreenSpaceReflections(r);
+    }
+#else
+    // objects with reflections just erase the framebuffer, they need to be drawn in the
```
bejado, line 17: Do you mean "without reflections"?

romainguy (approved):

shaders/src/shading_reflections.fs
pixelflinger: yes because the material could have it enabled, while the View disables it. In this case we don't want to do the ray-marching.

shaders/src/shading_reflections.fs
pixelflinger: uh. yes.

shaders/src/shading_reflections.fs
pixelflinger: Actually that reminds me that I need to fix that :-)

libs/filabridge/include/private/filament/Variant.h
pixelflinger: done

PR #5169, opened 2022-02-07
Add function overloading for setMorphTargetBufferAt

daemyung:


pixelflinger (changes requested):
I agree with @prideout, this is adding too much stuff for very little benefit.

Here is whatI think we should do:

- only keep the version that takes an `offset` and `count` (i.e.: remove the version that takes only a `count` completely).
- add an `inline` in `RenderableManager.h` that takes neither, only a C++ helper:
```
inline void setMorphTargetBufferAt(Instance instance, uint8_t level, size_t primitiveIndex,
            MorphTargetBuffer* morphTargetBuffer) {
setMorphTargetBufferAt(instance, level, primitiveIndex, morphTargetBuffer, 0, morphTargetBuffer->getCount());
}
```

daemyung:
> * add an `inline` in `RenderableManager.h` that takes neither, only a C++ helper:
> 
> ```
> inline void setMorphTargetBufferAt(Instance instance, uint8_t level, size_t primitiveIndex,
>             MorphTargetBuffer* morphTargetBuffer) {
> setMorphTargetBufferAt(instance, level, primitiveIndex, morphTargetBuffer, 0, morphTargetBuffer->getCount());
> }
> ```

@pixelflinger We need to include `/Users/danny/Repos/filament/filament/src/MorphTargetBuffer.cpp` at `components/RenderableManager.h` to make helper function as inline. Is it fine?

filament/include/filament/RenderableManager.h
```diff
@@ -486,7 +486,7 @@ class UTILS_PUBLIC RenderableManager : public FilamentAPI {
             MorphTargetBuffer* morphTargetBuffer, size_t offset, size_t count);
 
     void setMorphTargetBufferAt(Instance instance, uint8_t level, size_t primitiveIndex,
-            MorphTargetBuffer* morphTargetBuffer, size_t count); //!< \overload
+            MorphTargetBuffer* morphTargetBuffer); //!< \overload
```
pixelflinger, line 489: make this one just an `inline` right there. No need to create a function for it.  We're just going to say that it's a helper. The "real" API is the one that takes the offset and count.

android/filament-android/src/main/java/com/google/android/filament/RenderableManager.java
```diff
@@ -821,6 +816,7 @@ public long getNativeObject() {
     private static native int nSetBonesAsQuaternions(long nativeObject, int i, Buffer quaternions, int remaining, int boneCount, int offset);
     private static native void nSetMorphWeights(long nativeObject, int instance, float[] weights, int offset);
     private static native void nSetMorphTargetBufferAt(long nativeObject, int i, int level, int primitiveIndex, long nativeMorphTargetBuffer, int offset, int count);
+    private static native void nSetMorphTargetBufferAt(long nativeObject, int i, int level, int primitiveIndex, long nativeMorphTargetBuffer);
```
pixelflinger, line 819: Just implement this one in Java, by calling the other one and your new `getMorphTargetCount()`. It's just easier. In the javadoc just say it's a helper.

pixelflinger (changes requested):
thanks for the changes! I think we can remove more code even, because it's really not a performance or binary compatibility issue here, just convenience. See comments inline.

filament/include/filament/RenderableManager.h
daemyung: This way is good to add utility functions. :)

android/filament-android/src/main/java/com/google/android/filament/RenderableManager.java
daemyung: Ok it's done. nit: Are you saying new `SetMorphTargetBufferAt()` instead of `getMorphTargetCount()` right?

libs/gltfio/src/MorphHelper.cpp
```diff
@@ -103,7 +103,7 @@ void MorphHelper::addPrimitive(cgltf_mesh const* mesh, int primitiveIndex, Entit
 
         auto& rcm = engine.getRenderableManager();
         rcm.setMorphTargetBufferAt(rcm.getInstance(entity), 0, primitiveIndex,
-                morphHelperPrim.targets, vertexBuffer->getVertexCount());
+                morphHelperPrim.targets);
```
pixelflinger, line 106: is that correct? `morphHelperPrim.targets` is the count of targets or count of vertices?

pixelflinger (commented):
looks good but I'm not sure about the use of `morphHelperPrim.targets`.

libs/gltfio/src/MorphHelper.cpp
daemyung: Yes it's correct. We are set `MorphTargetBuffer` for `geometry` so it should be match `geometry`'s count. Also `MorphTargetBuffer`'s count(morph target count) should be match renderable's morph target count.

daemyung:
> looks good but I'm not sure about the use of `morphHelperPrim.targets`.

@pixelflinger If you think that we should understand association between `geometry` and `MorphTargetBuffer` and setting `offset` and `count` manually help to understand this concept then I think having only one API will be better. One API is  `void setMorphTargetBufferAt(Instance instance, uint8_t level, size_t primitiveIndex, MorphTargetBuffer* morphTargetBuffer, size_t offset, size_t count);`.

libs/gltfio/src/MorphHelper.cpp
pixelflinger: oh you're right I misread the line. Sorry.

pixelflinger:
> > looks good but I'm not sure about the use of `morphHelperPrim.targets`.
> 
> @pixelflinger If you think that we should understand association between `geometry` and `MorphTargetBuffer` and setting `offset` and `count` manually help to understand this concept then I think having only one API will be better. One API is `void setMorphTargetBufferAt(Instance instance, uint8_t level, size_t primitiveIndex, MorphTargetBuffer* morphTargetBuffer, size_t offset, size_t count);`.

yes yes, nevermind my comment -- I didn't read the line properly. It's all good.

pixelflinger (approved):

libs/gltfio/src/MorphHelper.cpp
daemyung: No worry. Honestly it's very confused.

PR #5170, opened 2022-02-07
fix typo in RenderTarget::build attachment size validation

pixelflinger:
the validation could fail if attachments used different LODs but the
sizes still matched.


prideout (approved):
Nice one.

PR #5174, opened 2022-02-08
Make sure one use the right combination of version

eliemichel:
To avoid `Material version mismatch`


pixelflinger (approved):

prideout (changes requested):
Ah thanks, yes it's hard to find bandwidth to always update the tutorials, this can prevent some frustration.

Maybe you could also update the HTML in `/docs/webgl/tutorial_redball.html`?

eliemichel:
idk, I was just reporting a small change on the fly but if need to clone the project + install build script dependencies and generate the html it becomes a bit painful, my bet is that it'll be much quicker if you do it. ;)

prideout:
I assumed you would manually modify the HTML since the change is so small.  But yes, either of us could do it.

PR #5179, opened 2022-02-09
Set the morph target buffer by builder

daemyung:


filament/include/filament/RenderableManager.h
```diff
@@ -362,6 +382,11 @@ class UTILS_PUBLIC RenderableManager : public FilamentAPI {
             MaterialInstance const* materialInstance = nullptr;
             PrimitiveType type = PrimitiveType::TRIANGLES;
             uint16_t blendOrder = 0;
+            struct {
+                MorphTargetBuffer* buffer = nullptr;
+                size_t offset = 0;
+                size_t count = 0;
```
bejado, line 388: I'm confused, I don't see where `offset` or `count` are getting read?

android/filament-android/src/main/java/com/google/android/filament/RenderableManager.java
```diff
@@ -425,6 +425,44 @@ public Builder morphing(@IntRange(from = 0, to = 255) int targetCount) {
...
+        /**
+         * Utility method to specify morph target buffer for a primitive.
+         * For details, see the {@link RenderableManager.Builder#morphing}.
+         */
+        @NonNull
+        public Builder morphing(@IntRange(from = 0) int level,
+                                @IntRange(from = 0) int primitiveIndex,
+                                @NonNull MorphTargetBuffer morphTargetBuffer) {
+            nBuilderMorphing(mNativeBuilder, level, primitiveIndex,
+                    morphTargetBuffer.getNativeObject(), 0, morphTargetBuffer.getVertexCount();
```
bejado, line 462: From CI:

```
RenderableManager.java:462: error: ')' expected
```

daemyung: Thanks! It's fixed.

android/filament-android/src/main/java/com/google/android/filament/RenderableManager.java
```diff
@@ -425,6 +425,44 @@ public Builder morphing(@IntRange(from = 0, to = 255) int targetCount) {
...
+         * The morph target buffer must have an associated renderable and geometry. Especially
+         * the count of morph target buffer must equals to the renderable's morph target count
+         * and the count or the vertex count of morph target buffer must equals to the geometry's
+         * count.
+         *
+         * @param level the level of detail (lod), only 0 can be specified
+         * @param primitiveIndex zero-based index of the primitive, must be less than the count passed to Builder constructor
+         * @param morphTargetBuffer specifies the morph target buffer
+         * @param offset specifies where in the morph target buffer to start reading
+         * @param count number of the morph target buffer to read, must be equal to geometry's count.
```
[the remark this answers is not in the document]
daemyung: Yes it's updated. Could you review updated comments?

filament/include/filament/RenderableManager.h
daemyung: Right now it's not used and need to implement support these values.

android/filament-android/src/main/java/com/google/android/filament/RenderableManager.java
```diff
@@ -425,6 +425,44 @@ public Builder morphing(@IntRange(from = 0, to = 255) int targetCount) {
...
+         * The morph target buffer must have an associated renderable and geometry. Especially
+         * the count of morph target buffer must equals to the renderable's morph target count
+         * and the count or the vertex count of morph target buffer must equals to the geometry's
+         * count.
+         *
+         * @param level the level of detail (lod), only 0 can be specified
+         * @param primitiveIndex zero-based index of the primitive, must be less than the count passed to Builder constructor
+         * @param morphTargetBuffer specifies the morph target buffer
+         * @param offset specifies where in the morph target buffer to start reading (expressed as a number of vertices)
+         * @param count number of the morph target buffer to read, must be equal to geometry's count (for triangles, this should be a multiple of 3)
```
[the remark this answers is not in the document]
daemyung: Really thanks!

android/filament-android/src/main/java/com/google/android/filament/RenderableManager.java
```diff
@@ -425,6 +425,44 @@ public Builder morphing(@IntRange(from = 0, to = 255) int targetCount) {
             return this;
         }
 
+        /**
+         * Specifies the morph target buffer for a primitive.
+         *
+         * The morph target buffer must have an associated renderable and geometry. Especially
+         * the count of morph target buffer must equals to the renderable's morph target count
+         * and the count or the vertex count of morph target buffer must equals to the geometry's
+         * count.
```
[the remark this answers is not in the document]
daemyung: Thanks! It's more clear!

android/filament-android/src/main/java/com/google/android/filament/RenderableManager.java
```diff
@@ -809,6 +848,7 @@ public long getNativeObject() {
     private static native int nBuilderSkinningBones(long nativeBuilder, int boneCount, Buffer bones, int remaining);
     private static native void nBuilderSkinningBuffer(long nativeBuilder, long nativeSkinningBuffer, int boneCount, int offset);
     private static native void nBuilderMorphing(long nativeBuilder, int targetCount);
+    private static native void nBuilderMorphing(long nativeBuilder, int level, int primitiveIndex, long nativeMorphTargetBuffer, int offset, int count);
```
pixelflinger, line 851: please change the name of this one (to avoid overloads for the native JNI), for instance `nBuilderMorphingWithData`

pixelflinger (changes requested):

android/filament-android/src/main/java/com/google/android/filament/RenderableManager.java
daemyung: Do we have to use the different function name for this? It will be handled by JNI automatically if we add postfix.

pixelflinger: Yes but I don't want to have to deal with name mangling if we don't have to. I prefer to use a different name. This is all private code, so it doesn't matter too much if the jni implementation name doesn't match exactly the public api.

daemyung: Ok I understand but could you consider again? Because it's java compiler feature. You can generate jni header by `javac -h jni_header file.java`. You don't have to deal with name mangling. If you still want to use a different name, I will change a function name.

daemyung: @pixelflinger Ok I understand but could you consider again? Because it's java compiler feature. You can generate jni header by `javac -h jni_header file.java`. You don't have to deal with name mangling. If you still want to use a different name, I will change a function name.

romainguy: @daemyung No, those auto-generated bindings are absolutely horrible and make incremental work a major pain. Note that ideally we'd switch of all our JNI to manual registration from `JNI_OnLoad` to catch mistakes in the bindings at load time, but also to get rid of symbols and shrink the final .so.

daemyung: Ok I see what you want. I will change a function name.

pixelflinger (approved):

prideout (approved):

PR #5189, opened 2022-02-11
Crucial refactoring of VulkanFboCache and VulkanSwapChain.

prideout:
Two big changes to simplify and fix layout transitions:

(1) The RenderPassKey struct has been reduced from 88 bytes to 56 bytes
by using custom enum types and bitmasks. It also now has separate fields
for "final" and "initial" image layouts.

(2) VulkanSwapChain now leverages VulkanTexture, which enables it to
share tracking functionality, avoid duplicated state, and have less code
overall.

In a future PR we will clean up the ReadPixels and Blit paths, and also
remove some hacks that appease Vulkan when SSAO is enabled.

### DIFF
--- a/filament/backend/src/vulkan/VulkanFboCache.h
+++ b/filament/backend/src/vulkan/VulkanFboCache.h
@@ -28,6 +28,13 @@
 namespace filament {
 namespace backend {
 
+// Avoid using VkImageLayout since it requires 4 bytes.
+enum class VulkanDepthLayout : uint8_t {
+    UNDEFINED, // VK_IMAGE_LAYOUT_UNDEFINED
+    GENERAL,   // VK_IMAGE_LAYOUT_GENERAL
+    READ_ONLY, // VK_IMAGE_LAYOUT_DEPTH_STENCIL_READ_ONLY_OPTIMAL.
+};
+
 // Simple manager for VkFramebuffer and VkRenderPass objects.
 //
 // Note that a VkFramebuffer is just a binding between a render pass and a set of image views. So,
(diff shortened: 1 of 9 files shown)


filament/backend/src/vulkan/VulkanFboCache.h
```diff
@@ -39,25 +46,36 @@ class VulkanFboCache {
...
+        // For each target, we need to know three image layouts: the layout BEFORE the pass, the
+        // layout DURING the pass, and the layout AFTER the pass. Here are the rules:
+        // - For depth, we explicitly specify all three layouts.
+        // - Color targets have their initial image layout specified with a bitmask.
+        // - For each color target, the pre-existing layout is either UNDEFINED (0) or GENERAL (1).
+        // - The render pass and final images layout for color buffers is always GENERAL.
+        uint8_t initialColorLayoutMask;          // 1 byte
+        VulkanDepthLayout initialDepthLayout;    // 1 byte
+        VulkanDepthLayout renderPassDepthLayout; // 1 byte
+        VulkanDepthLayout finalDepthLayout;      // 1 byte (for now this is always GENERAL)
```
pixelflinger, line 58: You could probably use a bitfield (`:2`) here and "compress" those 3 bytes into one?

filament/backend/src/vulkan/VulkanTexture.cpp
```diff
@@ -447,6 +474,17 @@ void VulkanTexture::transitionLayout(VkCommandBuffer commands, const VkImageSubr
     }
 }
 
+// Notifies the texture that a particular subresource's layout has changed.
+void VulkanTexture::trackLayout(uint32_t miplevel, uint32_t layer, VkImageLayout layout) {
+    const uint32_t first = (layer << 16) | miplevel;
+    const uint32_t last = (layer << 16) | (miplevel + 1);
```
pixelflinger, line 480: makes me a little nervous to see a `uint32_t` shifted by 16 into another `uint32_t`. Could you just use a `std::pair` or your own little structure?
If not, at least have an `assert()` that we don't have overflows, or use a `uint16_t` for `layer` and `miplevel`.

pixelflinger (approved):

filament/backend/src/vulkan/VulkanTexture.cpp
prideout: Good catch, I'll add an assert_invariant here and elsewhere.  ( I sort of want to keep uint32_t in the func sig to be consistent with the Vulkan structures, which are fairly stable. )

PR #5198, opened 2022-02-14
Vulkan: clean up ReadPixels and Blit paths.

prideout:
VulkanSwapChain is now always backed by VulkanTexture, so these
code paths can be simplified.

### DIFF
--- a/filament/backend/src/vulkan/VulkanDriver.cpp
+++ b/filament/backend/src/vulkan/VulkanDriver.cpp
@@ -1512,7 +1512,8 @@ void VulkanDriver::readPixels(Handle<HwRenderTarget> src, uint32_t x, uint32_t y
             VK_IMAGE_LAYOUT_TRANSFER_SRC_OPTIMAL, stagingImage, VK_IMAGE_LAYOUT_GENERAL,
             1, &imageCopyRegion);
 
-    // Restore the source image layout.
+    // Restore the source image layout. Between driver API calls, color images are always kept in
+    // UNDEFINED layout or in their "usage default" layout (see comment for getDefaultImageLayout).
 
     srcTexture->transitionLayout(cmdbuffer, srcRange,
             getDefaultImageLayout(TextureUsage::COLOR_ATTACHMENT));
--- a/filament/backend/src/vulkan/VulkanDriver.cpp
+++ b/filament/backend/src/vulkan/VulkanDriver.cpp
@@ -1408,8 +1408,8 @@ void VulkanDriver::readPixels(Handle<HwRenderTarget> src, uint32_t x, uint32_t y
     const VkDevice device = mContext.device;
     VulkanRenderTarget* srcTarget = handle_cast<VulkanRenderTarget*>(src);
     VulkanTexture* srcTexture = srcTarget->getColor(mContext.currentSurface, 0).texture;
-    const VkFormat srcFormat = srcTexture ? srcTexture->getVkFormat() :
-            mContext.currentSurface->surfaceFormat.format;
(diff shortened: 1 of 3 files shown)


filament/backend/src/vulkan/VulkanDriver.cpp
```diff
@@ -1506,41 +1503,19 @@ void VulkanDriver::readPixels(Handle<HwRenderTarget> src, uint32_t x, uint32_t y
...
-            .newLayout = VK_IMAGE_LAYOUT_COLOR_ATTACHMENT_OPTIMAL,
-            .subresources = srcRange,
-            .srcStage = VK_PIPELINE_STAGE_TRANSFER_BIT,
-            .srcAccessMask = VK_ACCESS_TRANSFER_WRITE_BIT,
-            .dstStage = VK_PIPELINE_STAGE_FRAGMENT_SHADER_BIT,
-            .dstAccessMask = VK_ACCESS_SHADER_READ_BIT,
-        });
-    }
+    srcTexture->transitionLayout(cmdbuffer, srcRange,
+            getDefaultImageLayout(TextureUsage::COLOR_ATTACHMENT));
```
pixelflinger, line 1518: question, do we have to restore the layout here, or can we just let it be, and have it transition automatically next time it's used for something?

pixelflinger (approved):

filament/backend/src/vulkan/VulkanDriver.cpp
daemyung: @pixelflinger We can put off layout transition as I think because now filament tracks the current layout.

prideout: The FboCache key uses a bitfield for space efficiency, so there are only two choices for "existing" color layout: UNDEFINED and the default.  Since the layout here is neither undefined nor default, we need to transition it back to the default layout.

pixelflinger: @prideout maybe a comment about that would be useful.

prideout: Done.

filament/backend/src/vulkan/VulkanDriver.cpp
```diff
@@ -1558,10 +1534,7 @@ void VulkanDriver::readPixels(Handle<HwRenderTarget> src, uint32_t x, uint32_t y
     vkMapMemory(device, stagingMemory, 0, VK_WHOLE_SIZE, 0, (void**) &srcPixels);
     srcPixels += subResourceLayout.offset;
 
-    // TODO: investigate why this Y-flip exists. This conditional seems to work with both
-    // test_ReadPixels.cpp (readpixels from a normal render target with texture attachment) and
-    // viewer_basic_test.cc (readpixels from an offscreen swap chain)
-    const bool flipY = srcTexture ? true : false;
+    const bool flipY = false;
```
bejado, line 1537: So there's no longer a difference with surfaces versus textures in regards to flipping y?

bejado (approved):

filament/backend/src/vulkan/VulkanDriver.cpp
prideout: Yes I tested ReadPixels with headless and non-headless and they both seem right side up now.

To be honest I never understood why this was necessary in the first place.  Maybe there was a bug in older versions of Molten, or something like that.

bejado: 👍

PR #5208, opened 2022-02-15
Update Gradle

romainguy:


build/android/build.sh
```diff
@@ -56,7 +56,7 @@ fi
 # Continuous builds will build everything
 ANDROID_ABIS=
 if [[ "$TARGET" == "presubmit" ]]; then
-  ANDROID_ABIS="-q arm64-v8a,x86"
```
bejado, line 59: Is this to further reduce build times? Can you update the comment above?

bejado (approved):

build/android/build.sh
romainguy: It's just temporary, I'm trying to figure out a fix.

PR #5209, opened 2022-02-16
Linux: Avoid identifer collision with glibc

100espressos:


pixelflinger (approved):

pixelflinger:
@100espressos thanks for the contribution. You need to sign the CLA before we can accept it. See "Details" above.

100espressos:
While I understand some of the motivations, some agreements and signatories will tend to discourage the tradition of FOSS drive-by patching. On a variety of established grounds I don't anticipate creating a Google account or entering into a contract with Google. ..but FOSS forever ❤️

PR #5216, opened 2022-02-17
Decouple the morphing from the render primitive

daemyung:
There is no reason the morphing is stored in the render primitive.


bejado:
This looks fine to me, @pixelflinger will probably have some thoughts though.

pixelflinger (changes requested):

filament/src/components/RenderableManager.cpp
```diff
@@ -408,10 +409,12 @@ void FRenderableManager::create(
...
                 .count = targetCount };
 
-            for (size_t i = 0, c = builder->mEntries.size(); i < c; ++i) {
+            MorphTargets* morphTargets = new MorphTargets[entryCount];
+            for (size_t i = 0; i < entryCount; ++i) {
                 const auto& morphing = builder->mEntries[i].morphing;
-                rp[i].set(upcast(morphing.buffer));
+                morphTargets[i] = { upcast(morphing.buffer), morphing.offset, morphing.count };
             }
+            mManager[ci].morphTargets = { morphTargets, size_type(entryCount) };
```
pixelflinger, line 417: Why not store `MorphTargets` directly into `mManager[ci].morphTargets` instead of allocating an array on the heap?
There is no advantage because the array is always allocated, even if there is no morphing.

pixelflinger (changes requested):

filament/src/RenderPass.cpp
```diff
@@ -389,17 +389,18 @@ void RenderPass::generateCommandsImpl(uint32_t extraFlags,
...
-            FMorphTargetBuffer const* const morphTargetBuffer = primitive.getMorphTargetBuffer();
             if constexpr (isColorPass) {
                 cmdColor.primitive.primitiveHandle = primitive.getHwHandle();
                 cmdColor.primitive.materialVariant = variant;
                 RenderPass::setupColorCommand(cmdColor, mi, inverseFrontFaces);
 
                 cmdColor.primitive.morphWeightBuffer = morphing.handle;
-                if (UTILS_UNLIKELY(morphTargetBuffer)) {
-                    cmdColor.primitive.morphTargetBuffer = morphTargetBuffer->getHwHandle();
+                if (UTILS_UNLIKELY(soaVisibility[i].morphing)) {
```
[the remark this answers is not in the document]
daemyung: Ok it's done.

filament/src/RenderPass.cpp
```diff
@@ -495,8 +496,9 @@ void RenderPass::generateCommandsImpl(uint32_t extraFlags,
                 cmdDepth.primitive.rasterState.culling = mi->getCullingMode();
 
                 cmdDepth.primitive.morphWeightBuffer = morphing.handle;
-                if (UTILS_UNLIKELY(morphTargetBuffer)) {
-                    cmdDepth.primitive.morphTargetBuffer = morphTargetBuffer->getHwHandle();
+                if (UTILS_UNLIKELY(soaVisibility[i].morphing)) {
```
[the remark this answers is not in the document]
daemyung: Ok it's done.

filament/src/components/RenderableManager.cpp
daemyung: Because each primitive needs a different morph target buffer. If we store `MorphTargets` directly into `mManager[ci].morphTargets` then all primitives have a same morph target buffer.

This array will be allocated if skinning or morphing is enabled.

daemyung: The goal of this PR is not saving memory. The goal is that changing code to make sense more.

Having morph targets in primitive is a concept of glTF. Filament doesn't have to follow this concept. Actually the resposiblity of `FRenderPrimitive` is to define vertex input stream because `FRenderPrimitive` map to `HwRenderPrimitive` in Filament. If we store `MorphTargetBuffer` into `FRenderPrimitive` then it exceeds the resposibility of `FRenderPrimitive`. Becase `MorphTargetBuffer` doesn't have any information how to set up vertex input stream. So handling morphing like skinning will be better.

prideout (approved):

filament/src/RenderPass.cpp
```diff
@@ -390,17 +391,18 @@ void RenderPass::generateCommandsImpl(uint32_t extraFlags,
...
                 cmdColor.primitive.primitiveHandle = primitive.getHwHandle();
                 cmdColor.primitive.materialVariant = variant;
                 RenderPass::setupColorCommand(cmdColor, mi, inverseFrontFaces);
 
                 cmdColor.primitive.morphWeightBuffer = morphing.handle;
-                if (UTILS_UNLIKELY(morphTargetBuffer)) {
-                    cmdColor.primitive.morphTargetBuffer = morphTargetBuffer->getHwHandle();
+                if (UTILS_UNLIKELY(hasMorphing)) {
+                    auto& morphTargets = morphing.targets[pi];
+                    cmdColor.primitive.morphTargetBuffer = morphTargets.buffer->getHwHandle();
```
pixelflinger, line 405: I was just looking at this yesterday. I really don't like that this adds (added) a branch in our inner hot loop.
Is there a way we can replace this whole `if` by just:

```
cmdColor.primitive.morphTargetBuffer = primitive.getMorphTargetHandle();
```

The RenderableManager should be able to populate this. In other words, could we use FRenderPrimitive has a staging cache for `soaMorphing[i].targets[pi].buffer->getHwHandle();`

filament/src/RenderPass.cpp
```diff
@@ -496,8 +498,9 @@ void RenderPass::generateCommandsImpl(uint32_t extraFlags,
                 cmdDepth.primitive.rasterState.culling = mi->getCullingMode();
 
                 cmdDepth.primitive.morphWeightBuffer = morphing.handle;
-                if (UTILS_UNLIKELY(morphTargetBuffer)) {
-                    cmdDepth.primitive.morphTargetBuffer = morphTargetBuffer->getHwHandle();
+                if (UTILS_UNLIKELY(hasMorphing)) {
+                    auto& morphTargets = morphing.targets[pi];
+                    cmdDepth.primitive.morphTargetBuffer = morphTargets.buffer->getHwHandle();
```
pixelflinger, line 503: same comment here.

pixelflinger (changes requested):
I would like to avoid the extra branch in the inner hot loop, if possible.

filament/src/RenderPass.cpp
daemyung: Ok... I'm done by creating `MorphTargets` always and set dummy `MorphTargetBuffers`. Please review new changes. By these changes, I can remove more branches in hot loops.

filament/src/components/RenderableManager.h
```diff
@@ -64,6 +64,12 @@ class FRenderableManager : public RenderableManager {
 
     static_assert(sizeof(Visibility) == sizeof(uint16_t), "Visibility should be 16 bits");
 
+    struct MorphTargets {
+        FMorphTargetBuffer* buffer = nullptr;
+        size_t offset = 0;
+        size_t count = 0;
```
pixelflinger, line 70: I think we can use `uint32_t` here, this way this structure would become 16 bytes instead of 24.

pixelflinger (changes requested):
Thank you so much, I like this version!  I just had a couple very minor requests.

filament/src/components/RenderableManager.cpp
```diff
@@ -394,6 +395,13 @@ void FRenderableManager::create(
             }
         }
 
+        // Create and initialize all needed MorphTargets. It's required to avoid branches in hot loops.
+        MorphTargets* morphTargets = new MorphTargets[entryCount];
```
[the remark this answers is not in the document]
daemyung: Yes it is. Actually we can have less `MorphTargets` than `FRenderPrimitive` but we need branchs in hot loop.

filament/src/details/MorphTargetBuffer.h
```diff
@@ -37,6 +37,11 @@ class FEngine;
 
 class FMorphTargetBuffer : public MorphTargetBuffer {
 public:
+    class DummyMaterialBuilder : public MorphTargetBuffer::Builder {
```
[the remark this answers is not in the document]
daemyung: Ok it's done.

filament/src/components/RenderableManager.h
```diff
@@ -142,6 +148,7 @@ class FRenderableManager : public RenderableManager {
     struct MorphingBindingInfo {
         backend::Handle<backend::HwBufferObject> handle;
         uint32_t count;
+        const MorphTargets* targets; // Pointer to Slice<MorphTargets> at a renderable.
```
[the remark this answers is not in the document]
daemyung: Ok it's done.

filament/src/components/RenderableManager.h
daemyung: Ok it's done.

pixelflinger (approved):

PR #5225, opened 2022-02-18
Vulkan: ensure that desired present mode is actually supported.

prideout:


filament/backend/src/vulkan/VulkanDriver.cpp
```diff
@@ -278,23 +278,23 @@ VulkanDriver::VulkanDriver(VulkanPlatform* platform,
...
-    for (VkFormat format = (VkFormat) 1;;) {
-        VkFormatProperties props;
-        vkGetPhysicalDeviceFormatProperties(mContext.physicalDevice, format, &props);
-        if ((props.optimalTilingFeatures & required) == required) {
-            utils::slog.i << format << " ";
-        }
-        if (format == VK_FORMAT_ASTC_12x12_SRGB_BLOCK) {
-            utils::slog.i << utils::io::endl;
-            break;
+    if (VK_ENABLE_VALIDATION && FILAMENT_VULKAN_VERBOSE) {
```
pixelflinger, line 281: I started to use `if constexpr` for these cases.

pixelflinger (approved):

filament/backend/src/vulkan/VulkanDriver.cpp
prideout: Nice, done.

PR #5226, opened 2022-02-18
Workaround feedback loops limitation in GLES

pixelflinger:


prideout (approved):
Personally I am more familiar with  DEPTH_STENCIL terminology than ANCILLARY.  (is that a term in the GL spec?)

bejado (approved):

PR #5230, opened 2022-02-18
Fix web apps that have multiple Filament viewers.

prideout:
This fixes the emscripten binding errors that we've been seeing
with the <filament-viewer> test page, which prevented us from
including web in the last few Filament releases.

The binding errors were caused by double-initializing the emscripten
module.

I fixed this by allowing clients (e.g. FilamentViewer) to call
Filament.init() more than once. We now accumulate a list of "on ready"
callbacks that get triggered after the emscripten module becomes ready.

As far as I can tell, multiple canvases were actually always broken, and
the viewer test page worked in the past only because we got lucky.

### DIFF
--- a/web/filament-js/filament-viewer.js
+++ b/web/filament-js/filament-viewer.js
@@ -17,20 +17,15 @@
 // If you are bundling this with rollup, webpack, or esbuild, the following URL should be trimmed.
 import { LitElement, html, css } from "https://unpkg.com/lit-element?module";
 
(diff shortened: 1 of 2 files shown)


pixelflinger (approved):

bejado:
Cherry-pick into rc/1.19.1?

prideout:
>Cherry-pick into rc/1.19.1?

Done!

PR #5243, opened 2022-02-23
Bugfix: lookAt z axis negated

chubei-oppen:
Reference: https://github.com/google/filament/blob/49b8e5e5d2c90b62cada2b77941db1f8bcfdfbc9/libs/math/include/math/mat4.h#L538


pixelflinger:
@chubei-oppen we need you to sign the CLA before we can accept your change. Thank you!

chubei-oppen:
I messed up the CLA. CLA team is working on helping fix it. I'll let you know once everything is okay.

chubei-oppen:
> Also, has this been fixed in the upstream repo? https://github.com/romainguy/kotlin-math

Looks like upstream has the same bug.

romainguy:
This change might break all the existing Android samples who rely on the current behavior. They’ll need to be fixed as well.

chubei-oppen:
Did a quick search in Android samples. No one is using `lookAt`. They're all using `Camera::lookAt`, which behaves like after this PR.

romainguy:
👍 Thanks for checking

PR #5249, opened 2022-02-23
move public API thunks to their own cpp

pixelflinger:
This reduces code duplication by preventing some unneeded aggressive inlining.


prideout (approved):
LGTM, although I don't understand the Windows failure.

bejado:
Copying an example Windows failure here for reference:

```
filament.lib(DebugRegistry.cpp.obj) : error LNK2019: unresolved external symbol "private: bool __cdecl filament::FDebugRegistry::hasProperty(char const *)const " (?hasProperty@FDebugRegistry@filament@@AEBA_NPEBD@Z) referenced in function "public: bool __cdecl filament::DebugRegistry::hasProperty(char const *)const " (?hasProperty@DebugRegistry@filament@@QEBA_NPEBD@Z) [D:\a\filament\filament\out\cmake-mt\samples\gltf_viewer.vcxproj]
```

pixelflinger:
> LGTM, although I don't understand the Windows failure.

There was some templated function in a cpp file used by another cpp file -- so the later couldn't instantiate them. I think that was at least part of the problem. not sure why it worked on other platforms. I added explicit template instantiation for those.

PR #5250, opened 2022-02-23
Vulkan: Add asserts that check render target size.

prideout:


bejado (approved):

filament/backend/src/vulkan/VulkanContext.cpp
```diff
@@ -42,7 +42,12 @@ VkImageLayout VulkanAttachment::getLayout() const {
     return texture ? texture->getVkLayout(layer, level) : VK_IMAGE_LAYOUT_UNDEFINED;
 }
 
-VkImageView VulkanAttachment::getImageView(VkImageAspectFlags aspect) {
+VkExtent2D VulkanAttachment::getExtent2D() const {
+    assert_invariant(texture);
+    return { texture->width >> level, texture->height >> level };
```
pixelflinger, line 47: You need to max(1, ...) here.

pixelflinger (changes requested):

filament/backend/src/vulkan/VulkanContext.cpp
prideout: Nice catch, fixed.

pixelflinger (approved):

PR #5252, opened 2022-02-23
Engine: fill dummy textures with zeroes.

prideout:
We should to initialize these textures to prevent warnings with Vulkan.

### DIFF
--- a/filament/src/Engine.cpp
+++ b/filament/src/Engine.cpp
@@ -259,12 +259,6 @@ void FEngine::init() {
             .format(Texture::InternalFormat::RGBA8)
             .sampler(Texture::Sampler::SAMPLER_CUBEMAP)
             .build(*this));
-    static uint32_t pixel = 0;
-    Texture::PixelBufferDescriptor buffer(
-            &pixel, 4, // 4 bytes in 1 RGBA pixel
-            Texture::Format::RGBA, Texture::Type::UBYTE);
-    Texture::FaceOffsets offsets = {};
-    mDefaultIblTexture->setImage(*this, 0, std::move(buffer), offsets);
 
     // 3 bands = 9 float3
     const float sh[9 * 3] = { 0.0f };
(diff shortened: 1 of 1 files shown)


bejado (approved):
LGTM, but maybe we only need to do this in debug builds? We shouldn't be sampling from these textures I don't think.

prideout:
Interesting idea, we could skip this in release builds but then I would need to add some logic to the Vulkan backend to ensure that the image layout gets transitioned, since it wouldn't be spec compliant otherwise.

bejado:
> Interesting idea, we could skip this in release builds but then I would need to add some logic to the Vulkan backend to ensure that the image layout gets transitioned, since it wouldn't be spec compliant otherwise.

I see, probably not worth the extra complexity then since these are super small textures.

pixelflinger (changes requested):
WTH, we were not doing that? What was I drinking.

Actually I think some of those textures are actually read. That's why they're called DummyOne or DummyZero.

So the content needs to be correct, i.e. initialized with 0 or 1 (0xFF for RGBA8).

prideout:
Doh, I updated the commit to fill the "ones" ones with ones.  :)

One interesting detail is that the tangents are **signed** integers, so I filled those with 0x7f...

filament/src/Engine.cpp
```diff
@@ -288,12 +282,35 @@ void FEngine::init() {
...
             TextureFormat::RGBA8I, 1, 1, 1, 1, TextureUsage::DEFAULT);
 
     mDummyZeroTexture = driverApi.createTexture(SamplerType::SAMPLER_2D, 1,
             TextureFormat::RGBA8, 1, 1, 1, 1, TextureUsage::DEFAULT);
 
+    // initialize the dummy textures so that their contents are not undefined
+
+    using PixelBufferDescriptor = Texture::PixelBufferDescriptor;
+
+    static uint32_t zeroes = 0;
```
pixelflinger, line 295: Use static const.

pixelflinger (changes requested):
Use static const for the pixel variable.

filament/src/Engine.cpp
prideout: Done.

PR #5258, opened 2022-02-24
Vulkan: Fix regression with IridescentDishWithOlives.

prideout:
When we introduced texture-based morphing, our custom depth resolver
needed to be updated to set up samplers correctly. This fixes a
validation error about shader stage access.

### DIFF
--- a/filament/backend/src/vulkan/VulkanBlitter.cpp
+++ b/filament/backend/src/vulkan/VulkanBlitter.cpp
@@ -222,6 +222,10 @@ void VulkanBlitter::lazyInit() noexcept {
     VkShaderModule fragmentShader = decode(VKSHADERS_BLITDEPTHFS_DATA, VKSHADERS_BLITDEPTHFS_SIZE);
     mDepthResolveProgram = new VulkanProgram(mContext, vertexShader, fragmentShader);
 
+    // Allocate one anonymous sampler at slot 0.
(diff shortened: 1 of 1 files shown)


filament/backend/src/vulkan/VulkanBlitter.cpp
```diff
@@ -222,6 +222,10 @@ void VulkanBlitter::lazyInit() noexcept {
     VkShaderModule fragmentShader = decode(VKSHADERS_BLITDEPTHFS_DATA, VKSHADERS_BLITDEPTHFS_SIZE);
     mDepthResolveProgram = new VulkanProgram(mContext, vertexShader, fragmentShader);
 
+    // Allocate one anonymous sampler at slot 0.
+    mDepthResolveProgram->samplerGroupInfo[0].samplers.reserve(1);
+    mDepthResolveProgram->samplerGroupInfo[0].samplers.resize(1);
+
```
pixelflinger, line 228: I'm a little concerned that a change such as the morphing change requires changes in the backend. This should have been a "filament side" change, AFAICT.

pixelflinger (approved):

filament/backend/src/vulkan/VulkanBlitter.cpp
prideout: Oh it actually wasn't the texture morphing change, it was the "split up samplers by stage" change.

PR #5261, opened 2022-02-25
Vulkan: assert if Filament asks to resolve depth in the render pass.

prideout:


bejado (approved):

prideout:
Rather than asserting, we decided that backends should offer a query to let Filament know if this is supported.

Since that work is still WIP, I changed this PR back to a draft.

PR #5265, opened 2022-02-25
more code size optimizations

pixelflinger:


prideout (approved):

libs/utils/include/utils/compiler.h
```diff
@@ -167,6 +167,14 @@
 #   define UTILS_HAS_FEATURE_CXX_THREAD_LOCAL 0
 #endif
 
+#if defined(_MSC_VER)
+// MSVC does not support loop unrolling hints
```
bejado, line 171: Huh, surprising

bejado (approved):

PR #5267, opened 2022-02-25
Vulkan: shrink PipelineKey from 608 bytes to 304 bytes, et al.

prideout:
This optimizes our memory footprint and does some cleanup. 
 There are no behavioral changes.

This PR is composed of 3 commits that should stay separate.

### DIFF
--- a/filament/backend/src/vulkan/VulkanPipelineCache.h
+++ b/filament/backend/src/vulkan/VulkanPipelineCache.h
@@ -282,26 +282,33 @@ private:
             return *this;
         }
         operator VkDescriptorImageInfo() const { return { sampler, imageView, imageLayout }; }
+
+        // TODO: replace the 64-bit sampler handle with `uint32_t samplerParams` and remove the
+        // padding field. This is possible if we have access to the VulkanSamplerCache.
         VkSampler sampler;
+
         VkImageView imageView;
         VkImageLayout imageLayout;
         uint32_t padding;
     };
+
+    // We store size with 32 bits, so our "WHOLE" sentinel is different from Vk.
+    static const uint32_t WHOLE_SIZE = 0xffffffffu;
 
     // Represents all the Vulkan state that comprises a bound descriptor set.
     struct DescriptorKey {
         VkBuffer uniformBuffers[UBUFFER_BINDING_COUNT];             // 0
         DescriptorImageInfo samplers[SAMPLER_BINDING_COUNT];        // 64
         DescriptorImageInfo inputAttachments[TARGET_BINDING_COUNT]; // 832
-        VkDeviceSize uniformBufferOffsets[UBUFFER_BINDING_COUNT];   // 1024
-        VkDeviceSize uniformBufferSizes[UBUFFER_BINDING_COUNT];     // 1088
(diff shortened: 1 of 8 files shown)


bejado (approved):

filament/backend/src/vulkan/VulkanPipelineCache.h
```diff
@@ -208,32 +205,63 @@ class VulkanPipelineCache : public CommandBufferObserver {
...
-        VkPrimitiveTopology topology;                // 4 bytes
-        VkRenderPass renderPass;                     // 8 bytes
-        uint16_t subpassIndex;                       // 2 bytes
-        uint16_t padding0;                           // 2 bytes
-        VkVertexInputAttributeDescription vertexAttributes[VERTEX_ATTRIBUTE_COUNT]; // 256 bytes
-        VkVertexInputBindingDescription vertexBuffers[VERTEX_ATTRIBUTE_COUNT];      // 192 bytes
-        uint32_t padding1;                                                          // 4 bytes
+        VkShaderModule shaders[SHADER_MODULE_COUNT]; // 0
+        VkRenderPass renderPass;                     // 16
+        uint16_t topology : 16;                      // 24
```
pixelflinger, line 251: i s the bitfield needed here, since it's a `uint16_t`?

pixelflinger: I think it would be very useful to have the size of each field in comment as well

filament/backend/src/vulkan/VulkanSwapChain.h
```diff
@@ -38,19 +38,17 @@ struct VulkanSwapChain : public HwSwapChain {
     void destroy();
     void makePresentable();
     bool hasResized() const;
-    VulkanAttachment getColorAttachment() const; // TODO: remove
-    VulkanAttachment getDepthAttachment() const; // TODO: remove
-    VulkanTexture& getColorTexture() const;
 
-    // TODO: privatize more fields
+    VulkanTexture& getColorTexture() const;
+    VulkanTexture& getDepthTexture() const;
```
pixelflinger, line 43: seems a bit "dangerous" to have a `const` method return a non-const reference.

filament/backend/src/vulkan/VulkanSwapChain.cpp
```diff
@@ -343,18 +347,12 @@ bool VulkanSwapChain::hasResized() const {
...
-    VulkanTexture& tex = *this->mDepth.get();
-    return VulkanAttachment { .texture = &tex, .level = 0, .layer = 0 };
+VulkanTexture& VulkanSwapChain::getColorTexture() const {
+    return *mColor[mCurrentSwapIndex];
 }
 
-VulkanTexture& VulkanSwapChain::getColorTexture() const {
-    return *this->mColor[this->currentSwapIndex];
+VulkanTexture& VulkanSwapChain::getDepthTexture() const {
+    return *mDepth;
```
pixelflinger, line 355: should we worried that this could be `nullptr`?

prideout:
Thanks for the good feedback @pixelflinger , addressed all your comments except the last one, the swap chain is always backed by color & depth textures.

pixelflinger (approved):

PR #5275, opened 2022-03-01
DriverAPI: add isAutoDepthResolveSupported query.

prideout:


filament/src/details/View.h
```diff
@@ -217,6 +217,8 @@ class FView : public View {
     }
 
     void setRenderTarget(FRenderTarget* renderTarget) noexcept {
+        assert_invariant(!mMultiSampleAntiAliasingOptions.enabled ||
+                !renderTarget->hasSampleableDepth());
```
pixelflinger, line 221: I think we could be setting a `nullptr` renderTarget here (e.g. to "unset" it).

prideout: Excellent catch, fixed.

prideout:
@pixelflinger can you point me in the right direction for the next step?  We need to use this query from PostProcessManager, right?

prideout:
@pixelflinger this one is ready to go, I think.

pixelflinger (approved):

PR #5276, opened 2022-03-01
more code size optimizations

pixelflinger:


prideout (approved):
Does the removal of lock guards from ResourceList change anything?

PR #5277, opened 2022-03-01
Bugfix: Mat4 from Quaternion was transposed

chubei-oppen:


romainguy (approved):

romainguy:
Note: this fix came from upstream

chubei-oppen:
@romainguy Do you mind if I copy https://github.com/romainguy/kotlin-math/pull/41 to this repo too? I hope these fixes get into `1.20.1` or `1.21.0` so I can stop using extension methods in my project.

romainguy:
You can copy all the changes. What we should really do is add kotlin-math as a dependency :/

PR #5278, opened 2022-03-02
Vulkan: fix logic that determines when to grow descriptor pool.

prideout:
When we added per-layout arenas for each of the 3 descriptor types, we did not
account for them when determining how much of the Vk Pool is in use.
Therefore the "growth" (really a re-creation) of the Vk Pool wasn't
always occurring when necessary, causing descriptor set allocation
to fail with large models.

### DIFF
--- a/filament/backend/src/vulkan/VulkanPipelineCache.cpp
+++ b/filament/backend/src/vulkan/VulkanPipelineCache.cpp
@@ -232,9 +232,18 @@ VulkanPipelineCache::DescriptorCacheEntry* VulkanPipelineCache::createDescriptor
     // are no longer used. This occurs during the cleanup phase during command buffer submission.
     auto& descriptorSetArenas = layoutCacheEntry->descriptorSetArenas;
     if (descriptorSetArenas[0].empty()) {
-        if (mDescriptorSets.size() >= mDescriptorPoolSize) {
+
+        // If allocating a new descriptor set from the pool would cause it to overflow, then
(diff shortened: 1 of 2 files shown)


bejado (approved):

pixelflinger (approved):

filament/backend/src/vulkan/VulkanPipelineCache.cpp
```diff
@@ -232,9 +232,18 @@ VulkanPipelineCache::DescriptorCacheEntry* VulkanPipelineCache::createDescriptor
...
+
+        // If allocating a new descriptor set from the pool would cause it to overflow, then
+        // recreate the pool. The number of descriptor sets that have already been allocated from
+        // the pool is the sum of the "active" descriptor sets (mDescriptorSets) and the "dormant"
+        // descriptor sets (mDescriptorArenasCount).
+        //
+        // NOTE: technically both sides of the inequality below should be multiplied by
+        // DESCRIPTOR_TYPE_COUNT to get the true number of descriptor sets.
+        if (mDescriptorSets.size() + mDescriptorArenasCount + 1 > mDescriptorPoolSize) {
             growDescriptorPool();
```
jeanlemotan, line 244: I think there is a problem here - the `growDescriptorPool()` method clears the `mDescriptorSets`, invalidating the `descriptorCacheEntry` at line 224 dangling.

prideout: Yes I think that's a problem!  Will fix.

PR #5281, opened 2022-03-02
Apply some clang-tidy fixes

bejado:


libs/utils/include/utils/Invocable.h
```diff
@@ -18,6 +18,7 @@
 #define TNT_UTILS_INVOKABLE_H
 
 #include <type_traits>
+#include <utility>
```
romainguy, line 21: @pixelflinger Is that the type of STL includes we should strive to remove from our public headers?

romainguy (approved):

libs/utils/include/utils/Invocable.h
pixelflinger: yeah, I'm not sure why this one is needed. We need to see if we can replace whatever it does easily. I think generally <utility> is okay though.

pixelflinger (commented):
Is there a way we can avoid <utility> ?

bejado:
`<utility>` is needed for `std::forward`

PR #5285, opened 2022-03-03
Android Viewer: fix URI bug for some zips files.

prideout:
Dragging a zip file into the Remote Viewer page would fail if it
contained a glTF with `./` prefixes in the `images` section. The best
fix is to use Java's proper URI object rather than trying to parse the
string.


romainguy (approved):
Nice Kotlinification :)

PR #5286, opened 2022-03-04
fix TargetBufferInfo uses an union for the face (uint8_t) and layer (uint16_t)

### LINKED ISSUE #5273: TargetBufferInfo uses an union for the face (uint8_t) and layer (uint16_t)
from @jeanlemotan 

> The TargetBufferInfo uses an union for the face (uint8_t) and layer (uint16_t), and the vulkan driver relies on this fact and always uses the layer, even for cubemaps.
> However, the constructor of TargetBufferInfo for cubemaps sets the face in the initializer list, and this causes the layer to not be initialized anymore, resulting in the face-layer union to have the lower 8 bits set with the proper value (the face) but the upper 8 bits random (or 0xCC in MSVC Debug).
>

pixelflinger:
Instead of having 2 fields we're following the vulkan convention of
having only the layer field.

The layer/face convention is still maintained in the public APIs.

fixes #5273

### DIFF
--- a/filament/backend/src/opengl/GLUtils.h
+++ b/filament/backend/src/opengl/GLUtils.h
@@ -168,8 +168,9 @@ constexpr inline GLenum getComponentType(backend::ElementType type) noexcept {
     }
 }
 
-constexpr inline GLenum getCubemapTarget(backend::TextureCubemapFace face) noexcept {
-    return GL_TEXTURE_CUBE_MAP_POSITIVE_X + GLenum(face);
+constexpr inline GLenum getCubemapTarget(uint16_t layer) noexcept {
+    assert_invariant(layer <= 5);
+    return GL_TEXTURE_CUBE_MAP_POSITIVE_X + layer;
 }
 
 constexpr inline GLenum getWrapMode(backend::SamplerWrapMode mode) noexcept {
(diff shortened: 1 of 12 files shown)


filament/src/details/RenderTarget.cpp
```diff
@@ -133,7 +133,7 @@ FRenderTarget::FRenderTarget(FEngine& engine, const RenderTarget::Builder& build
         info.handle = t->getHwHandle();
         info.level  = attachment.mipLevel;
         if (t->getTarget() == Texture::Sampler::SAMPLER_CUBEMAP) {
-            info.face = attachment.face;
+            info.layer = uint16_t(attachment.face);
```
bejado, line 136: Use the `+` operator here?

filament/backend/src/opengl/GLUtils.h
```diff
@@ -168,8 +168,8 @@ constexpr inline GLenum getComponentType(backend::ElementType type) noexcept {
     }
 }
 
-constexpr inline GLenum getCubemapTarget(backend::TextureCubemapFace face) noexcept {
-    return GL_TEXTURE_CUBE_MAP_POSITIVE_X + GLenum(face);
+constexpr inline GLenum getCubemapTarget(uint16_t layer) noexcept {
+    return GL_TEXTURE_CUBE_MAP_POSITIVE_X + (GLenum(layer) % 6);
```
bejado, line 172: Would it be better to assert that layer is <= 5?

bejado (approved):

prideout (approved):

filament/src/details/RenderTarget.cpp
pixelflinger: good point

romainguy (approved):

PR #5288, opened 2022-03-07
Compute the rotation between two vectors

daemyung:


libs/math/include/math/quat.h
```diff
@@ -126,6 +126,45 @@ class MATH_EMPTY_BASES TQuaternion :
     constexpr static TQuaternion MATH_PURE fromAxisAngle(const TVec3<A>& axis, B angle) {
         return TQuaternion(std::sin(angle * 0.5) * normalize(axis), std::cos(angle * 0.5));
     }
+
+    // constructs a quaternion from orig to dest. from and dest must be normalized.
+    template<typename A, typename = enable_if_arithmetic_t<A>>
+    constexpr static TQuaternion MATH_PURE fromOrigToDest(const TVec3<A>& orig, const TVec3<A>& dest) {
```
daemyung, line 132: How's about fromTo(...)?

libs/math/include/math/quat.h
```diff
@@ -126,6 +126,43 @@ class MATH_EMPTY_BASES TQuaternion :
     constexpr static TQuaternion MATH_PURE fromAxisAngle(const TVec3<A>& axis, B angle) {
         return TQuaternion(std::sin(angle * 0.5) * normalize(axis), std::cos(angle * 0.5));
     }
+
+    // constructs a quaternion from orig to dest. orig and dest must be normalized.
+    template<typename A, typename B, typename = enable_if_arithmetic_t<A, B>>
+    constexpr static TQuaternion MATH_PURE fromOrigToDest(const TVec3<A>& orig, const TVec3<B>& dest) {
```
pixelflinger, line 132: I don't like the name because it uses abbreviated words, which we are trying to avoid in our APIs. I agree that to be consistent with the other APIs it should be `fromXXX`.  Maybe simply `fromVectorAtoB`?

pixelflinger (changes requested):

libs/math/include/math/quat.h
romainguy: It could be just `from`. Or `rotation`.

pixelflinger: Yes, but there are some constraints on the name, I think:
1) `fromXXX()` is more in line with `fromAxisAngle()` for instance. It shows it's a factory.
2) We want a name that implies the direction of the transform w.r.t. the parameters. e.g. `from()` or `rotation()` are unclear about is it from param1 to param2 or vice-versa.
So I thought `fromVectorAtoB(vec a, vec b)` was clearer, even if it sounds strange.

romainguy: `fromVectorAtoB()` reads a like a repetition of the arguments to me. `fromDirectedRotation()` or similar would make more sense I think (I guess this implies a src/dst order, but that doesn't seem far fetched to make that assumption).

libs/math/include/math/quat.h
```diff
@@ -126,6 +126,43 @@ class MATH_EMPTY_BASES TQuaternion :
...
+            }
+
+            rotationAxis = normalize(rotationAxis);
+            return fromAxisAngle(rotationAxis, F_PI);
+        }
+
+        // implementation from Stan Melax's Game Programming Gems 1 article
+        rotationAxis = cross(orig, dest);
+
+        const T s = sqrt((static_cast<T>(1) + cosTheta) * static_cast<T>(2));
```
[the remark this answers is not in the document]
daemyung: It's updated.

libs/math/include/math/quat.h
```diff
@@ -126,6 +126,43 @@ class MATH_EMPTY_BASES TQuaternion :
...
     }
+
+    // constructs a quaternion from orig to dest. orig and dest must be normalized.
+    template<typename A, typename B, typename = enable_if_arithmetic_t<A, B>>
+    constexpr static TQuaternion MATH_PURE fromOrigToDest(const TVec3<A>& orig, const TVec3<B>& dest) {
+        // see the implementation of glm/gtx/quaternion.hpp
+        T cosTheta = dot(orig, dest);
+        TVec3<T> rotationAxis;
+
+        if(cosTheta >= static_cast<T>(1) - std::numeric_limits<T>::epsilon()) {
```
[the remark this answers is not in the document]
daemyung: It's updated.

libs/math/include/math/quat.h
```diff
@@ -126,6 +126,43 @@ class MATH_EMPTY_BASES TQuaternion :
...
+            // orig and dest point in the same direction
+            return TQuaternion(1, 0, 0, 0);
+        }
+
+        if(cosTheta < static_cast<T>(-1) + std::numeric_limits<T>::epsilon()) {
```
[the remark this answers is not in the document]
daemyung: It's updated.

libs/math/include/math/quat.h
```diff
@@ -126,6 +126,43 @@ class MATH_EMPTY_BASES TQuaternion :
...
+
+            rotationAxis = normalize(rotationAxis);
+            return fromAxisAngle(rotationAxis, F_PI);
+        }
+
+        // implementation from Stan Melax's Game Programming Gems 1 article
+        rotationAxis = cross(orig, dest);
+
+        const T s = sqrt((static_cast<T>(1) + cosTheta) * static_cast<T>(2));
+        return TQuaternion(s * static_cast<T>(0.5f),
```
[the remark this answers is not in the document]
daemyung: It's updated.

libs/math/include/math/quat.h
```diff
@@ -126,6 +126,43 @@ class MATH_EMPTY_BASES TQuaternion :
...
+                rotationAxis.x / s, rotationAxis.y / s, rotationAxis.z / s);
```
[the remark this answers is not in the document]
daemyung: `s` can't be zero. It's guaranteed by two upper if statements.

libs/math/include/math/quat.h
```diff
@@ -126,6 +126,43 @@ class MATH_EMPTY_BASES TQuaternion :
     constexpr static TQuaternion MATH_PURE fromAxisAngle(const TVec3<A>& axis, B angle) {
         return TQuaternion(std::sin(angle * 0.5) * normalize(axis), std::cos(angle * 0.5));
     }
+
+    // constructs a quaternion from orig to dest. orig and dest must be normalized.
```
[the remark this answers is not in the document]
daemyung: It's updated.

libs/math/include/math/quat.h
daemyung: Change name to `fromDirectedRotation`.

pixelflinger (approved):

libs/math/include/math/quat.h
```diff
@@ -126,6 +126,44 @@ class MATH_EMPTY_BASES TQuaternion :
     constexpr static TQuaternion MATH_PURE fromAxisAngle(const TVec3<A>& axis, B angle) {
         return TQuaternion(std::sin(angle * 0.5) * normalize(axis), std::cos(angle * 0.5));
     }
+
+    // constructs a quaternion from orig to dest.
+    // it returns the shortest arc and orig and dest must be normalized.
```
pixelflinger, line 131:
```suggestion
    // it returns the shortest arc and `from` and `to` must be normalized.
```

pixelflinger (changes requested):

pixelflinger (approved):

PR #5297, opened 2022-03-09
Retrieve morph target count from the primitive

### LINKED ISSUE #5289: Morphing does not work anymore
**Describe the bug**
I used version `1.17.0` and applied morphing with the `setMorphWeights(i, weights)` method. This version threw some "Fatal signal 11" errors after several calls to `loadModelGlb()` so I updated to the latest one `1.20.1`. Performance is better and I don't see errors but morphing stopped working. Filament ignores `setMorphWeights` calls.  

**To Reproduce**
1. Implement morphing with version `1.17.0`
2. Update to `1.20.1`
3. Change `setMorphWeights(i, weights)` to `setMorphWeights(i, weights, 0)` to have successful compilation 
4. Morphing does not work anymore

**Expected behavior**
Morphing should work like before. 

 - OS: Android
 - Backend: OpenGL

 - Device: OnePlus 6
 - OS: Android 11

**Additional context**
I see that there's a new API – `MorphTargetBuffer` but I don't see any example regarding how to use it. All I need is a simple call that lets me morph a blend shape with a given value. `setMorphWeights(i, weights)` worked fine. 

Also, I see that it's possible to enable legacy morphing using `MaterialBuilder` but it's a part of the Filamat and I don't have this dependency.

daemyung:
Morph weights can be zero even if the primitive has morph targets.

Fixes #5289

### DIFF
--- a/libs/gltfio/src/AssetLoader.cpp
+++ b/libs/gltfio/src/AssetLoader.cpp
@@ -445,7 +445,7 @@ void FAssetLoader::createRenderable(const cgltf_data* srcAsset, const cgltf_node
     Aabb aabb;
 
     // glTF spec says that all primitives MUST have the same number of morph targets in the same order.
-    const cgltf_size numMorphTargets = mesh->weights_count;
+    const cgltf_size numMorphTargets = inputPrim ? inputPrim->targets_count : 0;
     builder.morphing(numMorphTargets);
 
     // For each prim, create a Filament VertexBuffer, IndexBuffer, and MaterialInstance.
@@ -455,7 +455,7 @@ void FAssetLoader::createRenderable(const cgltf_data* srcAsset, const cgltf_node
             slog.e << "Unsupported primitive type in " << name << io::endl;
         }
 
-        if (numMorphTargets > 0 && inputPrim->targets_count != numMorphTargets) {
+        if (numMorphTargets != inputPrim->targets_count) {
             slog.e << "Sister primitives must all have the same number of morph targets."
                    << io::endl;
             mError = true;
@@ -535,8 +535,8 @@ void FAssetLoader::createRenderable(const cgltf_data* srcAsset, const cgltf_node
     // node weights are provided, they override the ones specified on the mesh.
     if (numMorphTargets > 0) {
         RenderableManager::Instance renderable = mRenderableManager.getInstance(entity);
-        const auto size = std::min(MAX_MORPH_TARGETS, std::max(mesh->weights_count, node->weights_count));
-        std::vector<float> weights(size);
+        const auto size = std::min(MAX_MORPH_TARGETS, numMorphTargets);
+        FixedCapacityVector<float> weights(size, 0.0f);
         for (cgltf_size i = 0, c = std::min(size, mesh->weights_count); i < c; ++i) {
             weights[i] = mesh->weights[i];
         }
--- a/libs/viewer/src/SimpleViewer.cpp
+++ b/libs/viewer/src/SimpleViewer.cpp
@@ -998,8 +998,8 @@ void SimpleViewer::updateUserInterface() {
             }
             for (int i = 0; i != mMorphWeights.size(); ++i) {
                 const char* name = mAsset->getMorphTargetNameAt(mCurrentMorphingEntity, i);
-                ImGui::SliderFloat(name ? name : "Unnamed target",
-                        &mMorphWeights[i], 0.0f, 1.0);
+                std::string label = name ? name : "Unnamed target " + std::to_string(i);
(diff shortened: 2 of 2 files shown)


libs/gltfio/src/AssetLoader.cpp
```diff
@@ -444,22 +444,20 @@ void FAssetLoader::createRenderable(const cgltf_data* srcAsset, const cgltf_node
...
         if (!getPrimitiveType(inputPrim->type, &primType)) {
             slog.e << "Unsupported primitive type in " << name << io::endl;
         }
 
-        if (numMorphTargets > 0 && inputPrim->targets_count != numMorphTargets) {
-            slog.e << "Sister primitives must all have the same number of morph targets."
-                   << io::endl;
-            mError = true;
-            continue;
+        if (inputPrim->targets_count > 0) {
```
bejado, line 455: I don't think this will catch cases where some primitives have no morph targets, but others do. Is that allowed? For example:
```
Primitive 0: 0 morph targets
Primitive 1: 5 morph targets
Primitive 2: 5 morph targets
```

libs/gltfio/src/AssetLoader.cpp
```diff
@@ -444,22 +444,20 @@ void FAssetLoader::createRenderable(const cgltf_data* srcAsset, const cgltf_node
...
+            if (numMorphTargets > 0 && inputPrim->targets_count != numMorphTargets) {
+                slog.e << "Sister primitives must all have the same number of morph targets."
+                       << io::endl;
+            }
```
bejado, line 459: Should we still set `mError = true` here?

libs/gltfio/src/AssetLoader.cpp
daemyung: This case is invalid glTF spec so it must be not allowed.

libs/gltfio/src/AssetLoader.cpp
daemyung: Actually I just revert changes. Your suggestion will be great. I will update this.

libs/gltfio/src/AssetLoader.cpp
```diff
@@ -535,8 +536,8 @@ void FAssetLoader::createRenderable(const cgltf_data* srcAsset, const cgltf_node
     // node weights are provided, they override the ones specified on the mesh.
     if (numMorphTargets > 0) {
         RenderableManager::Instance renderable = mRenderableManager.getInstance(entity);
-        const auto size = std::min(MAX_MORPH_TARGETS, std::max(mesh->weights_count, node->weights_count));
-        std::vector<float> weights(size);
+        const auto size = std::min(MAX_MORPH_TARGETS, numMorphTargets);
+        std::vector<float> weights(size, 0.0f);
```
prideout, line 540: I realize this is pre-existing, but as a cleanup can we change this to FixedCapacityVector?  std::vector isn't used anywhere else in this file.  In fact I would also like SkinVector to also be FixedCapacityVector.

prideout (approved):
Looks fine, however I do have one a cleanup idea (std::vector => FixedCapacityVector).

libs/gltfio/src/AssetLoader.cpp
bejado: My point is that the scenario where the first primitive has no morph targets (but other primitives do) will _not_ be reported as "Sister primitives must all have the same number of morph targets." It's an invalid glTF file, yes, but so are the ones where we show that message. We should be consistent.

```
            "primitives" : [
                {
                    "attributes" : {
                        "POSITION" : 0,
                        "NORMAL" : 1,
                        "TEXCOORD_0" : 2
                    },
                    "indices" : 3,
                    "material" : 0
                    "targets" : [
                    ]
                },
                {
                    "attributes" : {
                        "POSITION" : 8,
                        "NORMAL" : 9,
                        "TEXCOORD_0" : 10
                    },
                    "indices" : 3,
                    "material" : 1,
                    "targets" : [
                        {
                            "POSITION" : 11,
                            "NORMAL" : 12
                        },
                        {
                            "POSITION" : 13,
                            "NORMAL" : 14
                        }
                    ]
                },
                // etc..
            ],

```

bejado (changes requested):

libs/gltfio/src/AssetLoader.cpp
daemyung: I get your point. I have updated code. First get the morph target count from the first primitive if the first primitive is exist. Second check other primitives have same number of morph target count.

libs/gltfio/src/AssetLoader.cpp
daemyung: It's updated.

pixelflinger (approved):

bejado (approved):

PR #5300, opened 2022-03-09
Vulkan: support multiple renderers

### LINKED ISSUE #5291: Vulkan multiple renderers
Vulkan backend doesn't support 2 or more Renderers / SwapChains

When creating 2 renderers, an assert will trigger in `void VulkanDriver::createDefaultRenderTargetR(Handle<HwRenderTarget> rth, int)` because the Vulkan driver stores the default render target in the context member - and assumes only one swap chain.

The `defaultRenderTarget` is used in the `makeCurrent` calls, to bind it to the swap chain buffers.
There are 2 simple solutions:
1. Always pass the render target to the makeCurrent calls, along with the swapChain
2. Store this default render target in the swapChain instead of the context.

prideout:
The first commit is the important one, it fixes #5291.  The other two commits are cleanup.

I tested this with our `multiple_windows` sample.

### DIFF
--- a/filament/backend/src/vulkan/VulkanContext.h
+++ b/filament/backend/src/vulkan/VulkanContext.h
@@ -89,7 +89,7 @@ struct VulkanContext {
     bool portabilitySubsetSupported = false;
     bool maintenanceSupported[3] = {};
     VulkanPipelineCache::RasterState rasterState;
-    VulkanSwapChain* currentSurface;
+    VulkanSwapChain* currentSwapChain;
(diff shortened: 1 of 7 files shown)


bejado (approved):
LGTM but I guess I'm confused why Filament ever really needs to create more than one default render target. I see that each `Renderer` gets its own default RT, so I get why this happens, but why couldn't we have a single default render target for the whole `Engine`?

prideout:
Closing this as per discussion with @pixelflinger , will make a purely cleanup PR.

Issue #5299, opened 2022-03-09
Bone pose not applied if glTF model doesn't have animations


hybridherbst:
**Describe the bug**
Seems that when a glTF model has bones and those have transformations, those transformations are not applied when the model isn't also animated (e.g. it's only "pushed into pose" through bones).

**To Reproduce**
1. Open https://google.github.io/filament/viewer/
2. Drag in [BendyCylinder.zip](https://github.com/google/filament/files/8216673/BendyCylinder.zip)

**Expected behavior**
![image](https://user-images.githubusercontent.com/2693840/157492154-d47efdfd-a334-4544-a76e-9e275f5f1948.png)
(in modelviewer & babylon sandbox)

**Screenshots**
![image](https://user-images.githubusercontent.com/2693840/157492252-1127ba52-5f0a-434b-8374-83040091ed71.png)

**Desktop (please complete the following information):**
 - OS: [Windows 11, Chrome]
 - GPU: [NVIDIA RTX 2070]

prideout:
Our Android viewer properly calls `updateBoneMatrices` regardless of the presence of animation, but our web and desktop viewers do not.  So, it probably works as expected on Android, although `transformToUnitCube` probably doesn't work as expected.  That's something I need to to think about.

Incidentally this issue has given me merits an API tweak, which I proposed with #5301.

PR #5301, opened 2022-03-09
gltfio: simplify API, make animator always available.

prideout:
In the past there was an API gotcha because users had to "get" the
animator before releasing the glTF source data. This could have been
surprising because it was a getter method, not a factory method.

This was due to overeager optimization on my part, I wanted to avoid
animator overhead for non-animated models, when in fact it has very
little overhead.

Moreover, the animator is conceivably useful even when there are no
pre-supplied animations (e.g. for applying skins), so let's just create
it unconditionally.

Motivated by #5299.

### DIFF
--- a/libs/gltfio/include/gltfio/FilamentAsset.h
+++ b/libs/gltfio/include/gltfio/FilamentAsset.h
@@ -194,14 +194,14 @@ public:
     const char* getExtras(utils::Entity entity = {}) const noexcept;
 
     /**
-     * Lazily creates the animation engine or returns it from the cache.
+     * Returns the animation engine.
      *
      * The animator is owned by the asset and should not be manually deleted.
-     * The first time this is called, it must be called before FilamentAsset::releaseSourceData().
+     * Must be called after loadResources or asyncBeginLoad, otherwise returns null.
      * If the asset is instanced, this returns a "primary" animator that controls all instances.
      * To animate each instance individually, use \see FilamentInstance.
      */
-    Animator* getAnimator() noexcept;
+    Animator* getAnimator() const noexcept;
 
     /**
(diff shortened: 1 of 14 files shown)


libs/gltfio/include/gltfio/FilamentAsset.h
```diff
@@ -194,14 +194,13 @@ class UTILS_PUBLIC FilamentAsset {
...
-     * Lazily creates the animation engine or returns it from the cache.
+     * Returns the animation engine.
      *
      * The animator is owned by the asset and should not be manually deleted.
-     * The first time this is called, it must be called before FilamentAsset::releaseSourceData().
      * If the asset is instanced, this returns a "primary" animator that controls all instances.
      * To animate each instance individually, use \see FilamentInstance.
      */
-    Animator* getAnimator() noexcept;
+    Animator* getAnimator() const noexcept;
```
bejado, line 203: Can this only be called after `ResourceLoader::loadResources`?

prideout: Yes this is still true, good catch, I will update the docs.

pixelflinger: `Animator` should have a private destructor (to prevent the user calling delete on it), it should also have deleted copy/move ctors and copy operators (or private if you need that internally).

prideout: It already has a private destructor, I'll go ahead and delete copy/move methods.

prideout: Done.

pixelflinger (approved):

PR #5303, opened 2022-03-10
gltfio: decouple MaterialProvider lifetime from AssetLoader.

### LINKED ISSUE #5132: ubershaderLoader.destroy() crashes when called after assetLoader.destroy()
Hi,

Android Filament 1.12.8

I create an AssetLoader with an UbershaderLoader (previously, Filament 1.10.6, I used a MaterialProvider instance).
  val materialProvider = UbershaderLoader(engine) // Pre Filament 1.10.7 MaterialProvider(engine)
  val assetLoader = AssetLoader(engine, materialProvider, EntityManager.get())

Filament now crashes when I destroy the UbershaderLoader (did not happen with the old MaterialProvider)
  assetLoader.destroy()
  materialProvider.destroy()
    --------- beginning of crash
A/libc: Fatal signal 11 (SIGSEGV), code 1 (SEGV_MAPERR), fault addr 0x6f635f61766152 in tid 24346 (filamenttracker), pid 24346 (filamenttracker)

Is this a bug or do I not need to call the UbershaderLoader's destroy method if I have destroyed the AssetLoader?

Thanks in advance,
Steve

prideout:
This is a Java-only gltfio API change (!)

Jave clients now need to call destroy on MaterialProvider.

Previously, the Java AssetLoader took ownership of the native
material provider upon construction, but this was neither documented
nor consistent with the C++ layer. (This is historical; in the past we
did not expose MaterialProvider to Java.)

One motivation for this (aside from API consistency) is that users may
wish to preserve the material cache from one run to another.

Fixes #5132.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -5,6 +5,8 @@ A new header is inserted each time a *tag* is created.
 
 ## v1.20.4 (currently main branch)
 
+- gltfio: Java clients must now destroy the MaterialProvider [⚠️ **API Change**].
+
 ## v1.20.3
 
 ## v1.20.2
--- a/android/filament-utils-android/src/main/java/com/google/android/filament/utils/ModelViewer.kt
+++ b/android/filament-utils-android/src/main/java/com/google/android/filament/utils/ModelViewer.kt
@@ -98,6 +98,7 @@ class ModelViewer(
 
     private var swapChain: SwapChain? = null
     private var assetLoader: AssetLoader
+    private var materialProvider: MaterialProvider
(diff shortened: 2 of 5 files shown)


pixelflinger (changes requested):
since this is an API change, we probably need to update the Release Notes

romainguy (approved):
Make sure to update the README as suggested by Mathias.

prideout:
> since this is an API change, we probably need to update the Release Notes

Done.

pixelflinger (approved):

PR #5304, opened 2022-03-10
create only a single "default render target"

### LINKED ISSUE #5291: Vulkan multiple renderers

pixelflinger:
we didn't need one per Renderer, and doing so created complexity in the
backends.

fixes #5291


romainguy (approved):

prideout (approved):
thanks!

PR #5309, opened 2022-03-11
Make public gltfio's math

daemyung:


romainguy:
Can you explain why these are made public?

daemyung:
> Can you explain why these are made public?

I need to interpolate translation, rotation and scale separately however `transform manager` only returns matrix. Thus somehow I have to decompose this matrix and `gltfio` already have these functions. So I make these functions to public and will use it outside of `filament`. If you don't want to expose these apis then please close this pr. I can copy & paste this function.

prideout (approved):

PR #5310, opened 2022-03-11
Retrieve skins from FilamentAsset

### LINKED ISSUE #5290: How can I get skins in glTF?
I need skins to do the inverse kinematics. However there is no way to get this at `FilamentAsset`.

Supports getter function for skins at `FilamentAsset`.

Actually I only needs the skin name and joints for the inverse kinematics. So supporting skin name and joints gatter function will be enough.

daemyung:
Fixes #5290

### DIFF
--- a/libs/gltfio/include/gltfio/FilamentAsset.h
+++ b/libs/gltfio/include/gltfio/FilamentAsset.h
@@ -204,6 +204,26 @@ public:
     Animator* getAnimator() const noexcept;
 
     /**
+     * Gets the number of skins.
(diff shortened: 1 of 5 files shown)


prideout (approved):
Looks good.  None of these queries work until after the user calls loadResources() or asyncBeginLoad(), but we already have some methods that work that way.

In the future maybe we could prevent the API gotcha by moving all such methods off into a new interface (`FinalizedAsset` or something)

Anyway, fine for now.

PR #5311, opened 2022-03-11
Update blending equation for MASKED (core Filament change)

### LINKED ISSUE #4576: glTF conformance: labels do not appear
Several khronos examples use simple quads with text as labels, but none seem to be showing up in Filament (regression): 
https://modelviewer.dev/fidelity/#khronos-TextureEncodingTest
https://modelviewer.dev/fidelity/#khronos-TextureLinearInterpolationTest
etc.

prideout:
Our existing behavior was surprising for users who draw opaque objects
into a semitransparent views, and this was especially evident with
the labels in `TextureLinearInterpolationTest`.

Yes, once again the labels in a glTF test turned out to be more useful
than the test itself. :)

After discussion with Romain we realized that you don't always want to
leave destination alpha intact.

I also made sure that this does not regress `AlphaBlendModeTest`.

Fixes #4576.

### DIFF
--- a/filament/src/details/Material.cpp
+++ b/filament/src/details/Material.cpp
@@ -220,15 +220,13 @@ FMaterial::FMaterial(FEngine& engine, const Material::Builder& builder)
             mRasterState.blendFunctionDstAlpha = BlendFunction::ZERO;
             mRasterState.depthWrite = true;
             break;
+        // Do not change this behavior without checking for regressions with AlphaBlendModeTest
+        // and TextureLinearInterpolationTest, with and without View::BlendMode::TRANSLUCENT.
         case BlendingMode::MASKED:
-            // This leaves destination alpha intact if it is already opaque, but makes it opaque if
-            // it is already semi-transparent. This is usually what you want when drawing a scene.
-            // Do not change this behavior without checking for regressions with AlphaBlendModeTest
-            // and TextureLinearInterpolationTest, with and without View::BlendMode::TRANSLUCENT.
             mRasterState.blendFunctionSrcRGB   = BlendFunction::ONE;
-            mRasterState.blendFunctionSrcAlpha = BlendFunction::SRC_ALPHA;
-            mRasterState.blendFunctionDstRGB   = BlendFunction::ZERO;
-            mRasterState.blendFunctionDstAlpha = BlendFunction::ONE;
+            mRasterState.blendFunctionSrcAlpha = BlendFunction::ONE;
+            mRasterState.blendFunctionDstRGB   = BlendFunction::ONE_MINUS_SRC_ALPHA;
+            mRasterState.blendFunctionDstAlpha = BlendFunction::ONE_MINUS_SRC_ALPHA;
             mRasterState.depthWrite = true;
             break;
         case BlendingMode::TRANSPARENT:
--- a/filament/src/details/Material.cpp
+++ b/filament/src/details/Material.cpp
@@ -221,12 +221,12 @@ FMaterial::FMaterial(FEngine& engine, const Material::Builder& builder)
             mRasterState.depthWrite = true;
             break;
         case BlendingMode::MASKED:
-            // MASKED mode now leaves destination alpha intact.
-            // This prevents strange behavior with semi-transparent render targets, which the
-            // model viewer team discovered when testing against the Khronos alpha test
-            // conformance model.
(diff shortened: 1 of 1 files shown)


romainguy (approved):
Yay!

prideout:
Thanks for your help on this one Romain.  :)   Would the team be okay with me cherry-picking this into Monday's release?  That way @elalish could pick it up sooner.

romainguy:
We should cherry pick it yes

pixelflinger:
That seems wrong to me. Let's not rush.

filament/src/details/Material.cpp
```diff
@@ -221,12 +221,12 @@ FMaterial::FMaterial(FEngine& engine, const Material::Builder& builder)
...
-            // This prevents strange behavior with semi-transparent render targets, which the
-            // model viewer team discovered when testing against the Khronos alpha test
-            // conformance model.
+            // This leaves destination alpha intact if it is already opaque, but makes it opaque if
+            // it is already semi-transparent. This is usually what you want when drawing a scene.
+            // Do not change this behavior without checking for regressions with AlphaBlendModeTest
+            // and TextureLinearInterpolationTest, with and without View::BlendMode::TRANSLUCENT.
             mRasterState.blendFunctionSrcRGB   = BlendFunction::ONE;
-            mRasterState.blendFunctionSrcAlpha = BlendFunction::ZERO;
+            mRasterState.blendFunctionSrcAlpha = BlendFunction::SRC_ALPHA;
```
pixelflinger, line 229: If I'm not mistaken, with this setup the alpha value is calculated as:

`result_alpha = src_alpha * src_alpha + dst_alpha`

this just seems wrong, This can lead to alpha values > 1, and I don't think there is a justification for the src_alpha^2.

Why not use the standard equation: ONE / ONE_MINUS_SRC_ALPHA, which would lead to:

`result_alpha = src_alpha + (1 - src_alpha) * dst_alpha`
`result_alpha = dst_alpha + src_alpha * (1 - dst_alpha)`

The destination alpha can only increase, i.e. only get more opaque. If src_alpha is 1, then result is 1 and if dst_alplha is 1, then it stays to 1. I think this is what you wanted.

pixelflinger (changes requested):

filament/src/details/Material.cpp
romainguy: alpha>1 doesn't matter since it will be clamped on output anyway. But yeah, `ONE_MINUS_SRC_ALPHA` is better for DstAlpha.

romainguy:
And ideally when the target swap chain is not transparent we would disable blending for masked objects. It's unnecessarily expensive on non tilers.

filament/src/details/Material.cpp
pixelflinger: I don't think it's clamped for float targets.

prideout:
@pixelflinger I think you're right, the simplest solution is the best solution, which is to use classic src-over blending.  I'm not sure why this didn't occur to me last week.  I test this, then added another commit.

filament/src/details/Material.cpp
```diff
@@ -220,17 +220,9 @@ FMaterial::FMaterial(FEngine& engine, const Material::Builder& builder)
             mRasterState.blendFunctionDstAlpha = BlendFunction::ZERO;
             mRasterState.depthWrite = true;
             break;
+        // Do not change this behavior without checking for regressions with AlphaBlendModeTest
+        // and TextureLinearInterpolationTest, with and without View::BlendMode::TRANSLUCENT.
         case BlendingMode::MASKED:
```
romainguy, line 225: I don't like this fix because if your GPU is not a tiler then there is 0 benefit to using alpha masking over transparent blending. It would make rendering vegetations on desktop for instance incredibly expensive.

In addition you removed `depthWrite = true`, it needs to be brought back.

The blend functions should be the same as `OPAQUE` when the render target is opaque, and the same as `TRANSPARENT` (+ `depthWrite=true`) when the render target is transparent.

romainguy (changes requested):

filament/src/details/Material.cpp
prideout: Good catch on the `depthWrite` line, fixed.

prideout:
We are going to make a second attempt.  Closing.

PR #5315, opened 2022-03-14
Make ios/samples support source code debugging

DevinShine:
Current `ios/samples` do not allow easy debugging of filament source code, because filament is relied on by the demo as a compilation product and not in source form.
I added three CMakelists.txt to facilitate source debugging of `ios/samples` against samples, which I think will be more friendly to iOS developers and make it easier for them to debug filament.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -727,6 +727,12 @@ if (IS_HOST_PLATFORM)
     add_subdirectory(${TOOLS}/specular-color)
 endif()
 
+if (IOS AND IOS_DEBUG_SAMPLES)
+    add_subdirectory(${FILAMENT}/ios/samples/hello-triangle)
+    add_subdirectory(${FILAMENT}/ios/samples/hello-pbr)
+    add_subdirectory(${FILAMENT}/ios/samples/hello-gltf)
+endif()
+
 # Generate exported executables for cross-compiled builds (Android, WebGL, and iOS)
 if (NOT CMAKE_CROSSCOMPILING)
     export(TARGETS matc cmgen filamesh mipgen resgen glslminifier FILE ${IMPORT_EXECUTABLES})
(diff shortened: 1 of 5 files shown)


bejado:
Thanks for your contribution, but I'm not sure why this is necessary. Currently, I can build Filament for iOS with debug symbols and set Xcode breakpoints in Filament source code just fine.

DevinShine:
> Thanks for your contribution, but I'm not sure why this is necessary. Currently, I can build Filament for iOS with debug symbols and set Xcode breakpoints in Filament source code just fine.

Maybe I'm not very good with xcode, for example, I want to add a breakpoint to the hello-gltf example and I can only do it in xcode via adding a Symbolic Breakpoint.

image

image

This is the way I debug hello-gltf. There may be other tricks I haven't mastered, so I don't find this very convenient and added the corresponding CMakelists.txt.
The purpose of my modification is just to make it easy to read the filament source code in an xcode project and to add breakpoints.

As you can see below, the filament source and samples are in one xcode project, so you can easily jump to the corresponding cpp file and add breakpoints.
If you want to modify the filament source code, you can do so directly in this project and run the samples to see the effect.
image

I need to interrupt frequently and modify the source code to enhance my understanding in the process of learning filament, that's why I have this modification suggestion. If there is a better way or technique I would be happy to discuss it. :)

bejado:
Thanks for the explanation. I agree that this makes things slightly easier, but we don't have the bandwidth to support two different build configurations for iOS samples.

Note that even if the sample apps rely on pre-compiled Filament, you can still edit Filament source code in Xcode and set breakpoints the "usual" way. Open up Finder and drag the `filament` folder into the Xcode sample project, making sure to uncheck "Add to targets". This way you can browse, edit, and set breakpoints in Filament source. You'll just need to re-run `build.sh -i -p ios debug` before running the sample each time, which is annoying, but IMO an acceptable trade-off.

Now that I think about it, you could even add the `build.sh -i -p ios debug` step as another "Run Script Phase" to avoid having to leave Xcode.

PR #5316, opened 2022-03-14
New behavior for BlendingMode::MASKED.

### LINKED ISSUE #4576: glTF conformance: labels do not appear

prideout:
The existing behavior was surprising for users who draw opaque objects
into a semitransparent views, and this was especially evident with
the labels in `TextureLinearInterpolationTest`.

We now disable blending for MASKED, which is what our existing materials
documentation already says. We also now set the fragment alpha to 1 when
the fragment is not discarded, which prevents the "punch through"
effect. This is consistent with ThreeJS.

Fixes #4576.

### DIFF
--- a/docs/Materials.md.html
+++ b/docs/Materials.md.html
@@ -1280,7 +1280,8 @@ Description
       render target is brightened.
     - **Masked**: blending is disabled. This blending mode enables alpha masking. The alpha channel
       of the material's output defines whether a fragment is discarded or not. Additionally,
-      ALPHA_TO_COVERAGE is enabled. See the maskThreshold section for more information.
+      ALPHA_TO_COVERAGE is enabled for non-translucent views. See the maskThreshold section for more
+      information.
 
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ JSON
 material {
--- a/docs/Materials.md.html
+++ b/docs/Materials.md.html
@@ -1279,8 +1279,8 @@ Description
     - **Screen**: blending is enabled. Effectively the opposite of the `multiply`, the content of the
       render target is brightened.
     - **Masked**: blending is disabled. This blending mode enables alpha masking. The alpha channel
-      of the material's output defines whether a fragment is discarded or not. See the maskThreshold
-      section for more information.
+      of the material's output defines whether a fragment is discarded or not. Additionally,
+      ALPHA_TO_COVERAGE is enabled. See the maskThreshold section for more information.
 
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ JSON
 material {
@@ -1364,8 +1364,9 @@ Value
 
 Description
 :     Sets the minimum alpha value a fragment must have to not be discarded when the `blending` mode
-      is set to `masked`. When the blending mode is not `masked`, this value is ignored. This value
-      can be used to controlled the appearance of alpha-masked objects.
+      is set to `masked`. If the fragment is not discarded, its source alpha is set to 1. When the
+      blending mode is not `masked`, this value is ignored. This value can be used to controlled the
+      appearance of alpha-masked objects.
 
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ JSON
 material {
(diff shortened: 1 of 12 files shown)


pixelflinger (approved):

docs/Materials.md.html
```diff
@@ -1279,8 +1279,8 @@
     - **Screen**: blending is enabled. Effectively the opposite of the `multiply`, the content of the
       render target is brightened.
     - **Masked**: blending is disabled. This blending mode enables alpha masking. The alpha channel
-      of the material's output defines whether a fragment is discarded or not. See the maskThreshold
-      section for more information.
+      of the material's output defines whether a fragment is discarded or not. Additionally,
+      ALPHA_TO_COVERAGE is enabled. See the maskThreshold section for more information.
```
romainguy, line 1283: This is irrelevant since we always now set alpha to 1 for fragments that pass the test. See below.

shaders/src/shading_lit.fs
```diff
@@ -2,30 +2,40 @@
...
     // Use derivatives to smooth alpha tested edges
     return (a - getMaskThreshold()) / max(fwidth(a), 1e-3) + 0.5;
 }
-#endif
+
+float computeDiffuseAlpha(float a) {
+    // If we reach this point in the code, we already know that the fragment is not discarded due
+    // to the threshold factor. Therefore we can just output 1.0, which prevents a "punch thru"
+    // effect from occuring.
+    return 1.0;
```
romainguy, line 15: This should `return frameUniforms.transparentTarget ? 1.0 : a`.

shaders/src/shading_unlit.fs
```diff
@@ -34,6 +34,8 @@ vec4 evaluateMaterial(const MaterialInputs material) {
     color.a = computeMaskedAlpha(color.a);
     if (color.a <= 0.0) {
         discard;
+    } else {
+        color.a = 1.0;
```
romainguy, line 38: This should `if (frameUniforms.transparentTarget) color.a = 1.0`.

romainguy (changes requested):

shaders/src/shading_unlit.fs
pixelflinger: @prideout you'll need to refactor things a bit to make that work.

Renderer.cpp already has a `needsAlphaChannel` but it's computed way after the call to `view.prepare()` which is where the per-view UBO is set. You'll need to move the calculation of `needsAlphaChannel` up and pass it to `prepare`. I think that should do the trick.

romainguy (approved):
Thanks @prideout! One of those issues that end up being much harder to fix than originally thought. I like where you landed with this and hopefully it's the last time we had to think about this :))

shaders/src/shading_lit.fs
```diff
@@ -2,30 +2,44 @@
...
+
+float computeDiffuseAlpha(float a) {
+    // If we reach this point in the code, we already know that the fragment is not discarded due
+    // to the threshold factor. Therefore we can just output 1.0, which prevents a "punch through"
+    // effect from occuring. We do this only for TRANSLUCENT views in order to prevent breakage
+    // of ALPHA_TO_COVERAGE.
+    if (frameUniforms.needsAlphaChannel == 1.0) {
+        return 1.0;
+    }
+    return a;
```
pixelflinger, line 19: you could actually do `return frameUniforms.needsAlphaChannel*(1 - a) + a;` to avoid the conditional. Not sure if it would be best :-)

pixelflinger (approved):

libs/filabridge/include/private/filament/UibStructs.h
```diff
@@ -65,7 +65,7 @@ struct PerViewUib { // NOLINT(cppcoreguidelines-pro-type-member-init)
     math::float4 sun; // cos(sunAngle), sin(sunAngle), 1/(sunAngle*HALO_SIZE-sunAngle), HALO_EXP
 
     math::float2 lightFarAttenuationParams;     // a, a/far (a=1/pct-of-far)
-    float padding0;
+    float needsAlphaChannel;
```
bejado, line 68: Does this need to be a float? Could we use a single bit instead?

bejado (approved):
Can you make a note in `RELEASE_NOTES` that this fix requires re-compiling materials?

libs/filabridge/include/private/filament/UibStructs.h
pixelflinger: It's just that ubos can't have bitfields it bools

pixelflinger (approved):

PR #5321, opened 2022-03-15
ModelViewer: fix Engine ownership ambiguity.

prideout:
ModelViewer was destroying the Filament engine in its detach handler,
but a custom engine may have been passed into the constructor, in which
case it was surprising that ModelViewer took ownership over the engine's
lifetime.

This change makes it so that the engine is *always* owned by the caller,
and it is the caller's responsibility to destroy the engine.

Tested using orientation changes, with and without the custom
`configChanges` property in the manifest.

Discussed in #4724

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -5,6 +5,8 @@ A new header is inserted each time a *tag* is created.
 
 ## v1.20.5 (currently main branch)
 
+- Android ModelViewer no longer owns the Engine; requires clients to pass one in [⚠️ **API Change**].
+
 ## v1.20.4
 
 - gltfio: Java clients must now destroy the MaterialProvider [⚠️ **API Change**].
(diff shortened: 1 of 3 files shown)


romainguy (commented):
Can you also update the release notes? This is a pretty important API change.

prideout:
> Can you also update the release notes? This is a pretty important API change.

Yes,  I feel lukewarm about this though, do you think it's a good change?  For example, note that the Engine is now "reused" and never destroyed, which works but feels a little asymmetrical.

romainguy (changes requested):

romainguy:
> > Can you also update the release notes? This is a pretty important API change.
> 
> Yes, I feel lukewarm about this though, do you think it's a good change? For example, note that the Engine is now "reused" and never destroyed, which works but feels a little asymmetrical.

Yeah, see my comments. You need to destroy the engine.

I think the change is good but I wonder if we shouldn't keep the case of ModelViewer creating and owning the engine. This would require proper documentation at the class level to explain when you'd want to pass your own Engine. At this with your PR this is very explicit.

PR #5323, opened 2022-03-16
Avoid linking clashes with BlueGL and OpenGL

bejado:
Some clients need to link against both Filament and OpenGL. This can lead to symbol clashes under certain circumstances. As a fix, prefix all of our `bluegl` symbols with `bluegl_`. Previously we made this naming change for Windows, but now we'll do it on all platforms for simplicity.

bluegl files generated via:

```
cd libs/bluegl
./bluegl-gen.py
```


romainguy (approved):
Thanks Ben!

romainguy:
Great readme btw :)

prideout (approved):

pixelflinger (approved):

PR #5326, opened 2022-03-17
steps to switch SSR to using a 2D array instead of a sampler2d

pixelflinger:


filament/CMakeLists.txt
```diff
@@ -229,6 +229,10 @@ set(MATERIAL_SRCS
         src/materials/separableGaussianBlur2.mat
         src/materials/separableGaussianBlur3.mat
         src/materials/separableGaussianBlur4.mat
+        src/materials/separableGaussianBlur1L.mat
```
romainguy, line 232: Why the suffix `L` by the way? For `Layer`?

romainguy (approved):

PR #5332, opened 2022-03-18
Update MAX_MORPH_TARGETS to 256

### LINKED ISSUE #5327: Crash happens when more than 128 morph target are used
**Describe the bug**

the model have 245 morph targets, it renders fine with vesion 1.13.0. but crashs on vesion 1.20.2

the log is as follow:

```
W/Filament: WARNING: Exceeded max morph target count of 128
E/Filament: Panic
   in setTangentsAt:195
   reason: 128 target index must be < 128
E/Filament: --------- beginning of crash
A/libc: Fatal signal 6 (SIGABRT), code -1 (SI_QUEUE) in tid 30264 (com.demo.k3d), pid 30264 (com.demo.k3d)
```

daemyung:
Fixes #5327

### DIFF
--- a/libs/filabridge/include/filament/MaterialEnums.h
+++ b/libs/filabridge/include/filament/MaterialEnums.h
@@ -151,7 +151,7 @@ enum VertexAttribute : uint8_t {
 };
 
 static constexpr size_t MAX_LEGACY_MORPH_TARGETS = 4;
-static constexpr size_t MAX_MORPH_TARGETS = 128; // this is limited by filament::CONFIG_MAX_MORPH_TARGET_COUNT
+static constexpr size_t MAX_MORPH_TARGETS = 256; // this is limited by filament::CONFIG_MAX_MORPH_TARGET_COUNT
 static constexpr size_t MAX_CUSTOM_ATTRIBUTES = 8;
 
 /**


libs/filabridge/include/filament/MaterialEnums.h
```diff
@@ -151,7 +151,7 @@ enum VertexAttribute : uint8_t {
 };
 
 static constexpr size_t MAX_LEGACY_MORPH_TARGETS = 4;
-static constexpr size_t MAX_MORPH_TARGETS = 128; // this is limited by filament::CONFIG_MAX_MORPH_TARGET_COUNT
+static constexpr size_t MAX_MORPH_TARGETS = 256; // this is limited by filament::CONFIG_MAX_MORPH_TARGET_COUNT
```
pixelflinger, line 154: the problem is that GLES3.0 doesn't guarantee more  than 256 layers in a 2D array, and with 256 morph targets we'd need 512 guaranteed.

pixelflinger (changes requested):

prideout:
@pixelflinger I thought we had two separate textures for positions and tangents?  If so 256 shouldn't be a problem.  I would like to use 256, otherwise we have a feature regression on our hands.  In the past we supported 255 targets with 4 active simultaneous frames.  I believe Modo tends to export asset that have many morph targets but uses a max of 2 of them simultaneously.

pixelflinger:
> @pixelflinger I thought we had two separate textures for positions and tangents? If so 256 shouldn't be a problem. I would like to use 256, otherwise we have a feature regression on our hands. In the past we supported 255 targets with 4 active simultaneous frames. I believe Modo tends to export asset that have many morph targets but uses a max of 2 of them simultaneously.

You're completely right.

pixelflinger (approved):

prideout (approved):

PR #5336, opened 2022-03-19
greatly simplify SamplerBinderMap

pixelflinger:
The main simplification is that we don't need an actual map to retrieve
the global offset of a sampler, it can be easily calculated as:
  global_offset = offset_of_the_block + local_offset

### DIFF
--- a/libs/filabridge/include/private/filament/SamplerBindingMap.h
+++ b/libs/filabridge/include/private/filament/SamplerBindingMap.h
@@ -19,25 +19,7 @@
 
 #include <private/filament/EngineEnums.h>
 
-#include <tsl/robin_map.h>
-
 namespace filament {
-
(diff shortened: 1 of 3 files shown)


romainguy (approved):

libs/filabridge/include/private/filament/SamplerBindingMap.h
```diff
@@ -53,36 +35,22 @@ class SamplerBindingMap {
 
     // Given a valid Filament binding point and an offset within the block, returns true and sets
     // the output argument 'globalOffset' to the globally unique binding index.
```
prideout, line 37: This comment is now out of date, it should explain that the return value is the global offset.

prideout (approved):
Wow nice!  I noticed an outdated comment.

PR #5339, opened 2022-03-21
The backend CANNOT depend on filabridge

pixelflinger:
This would mean that the backend is not independent of our material
system. Thankfully we didn't have too many dependencies that leaked
there.

### DIFF
--- a/filament/backend/src/vulkan/VulkanHandles.cpp
+++ b/filament/backend/src/vulkan/VulkanHandles.cpp
@@ -76,10 +76,8 @@ VulkanProgram::VulkanProgram(VulkanContext& context, const Program& builder) noe
     // Make a copy of the binding map
     samplerGroupInfo = builder.getSamplerGroupInfo();
     if constexpr (FILAMENT_VULKAN_VERBOSE) {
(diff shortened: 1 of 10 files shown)


filament/backend/src/vulkan/VulkanHandles.cpp
```diff
@@ -76,10 +76,8 @@ VulkanProgram::VulkanProgram(VulkanContext& context, const Program& builder) noe
     // Make a copy of the binding map
     samplerGroupInfo = builder.getSamplerGroupInfo();
     if constexpr (FILAMENT_VULKAN_VERBOSE) {
-        utils::slog.d << "Created VulkanProgram " << builder.getName().c_str()
-                    << ", variant = (" << utils::io::hex
-                    << (int) builder.getVariant().key << utils::io::dec << "), "
-                    << "shaders = (" << bundle.vertex << ", " << bundle.fragment << ")"
+        utils::slog.d << "Created VulkanProgram " << builder
```
prideout, line 79: Cool.

prideout (approved):

PR #5342, opened 2022-03-21
cleanups and size optimizations

pixelflinger:


romainguy (approved):

prideout (approved):

romainguy (approved):

filament/backend/src/ostream.cpp
```diff
@@ -497,6 +497,17 @@ io::ostream& operator<<(io::ostream& out, MRT const& mrt) {
     return out;
 }
 
+io::ostream& operator<<(io::ostream& stream, ShaderStageFlags stageFlags) {
+    unsigned int v = +stageFlags.vertex + 2u * +stageFlags.fragment;
+    switch (v) {
```
bejado, line 502: Match the `to_string` implementation?
```suggestion
    switch (v & 3u) {
```

libs/filamat/src/GLSLPostProcessor.cpp
```diff
@@ -30,60 +27,63 @@
...
     const auto stageFlags = sib.getStageFlags();
-    if (!stageFlags.hasShaderType(config.shaderType)) {
-        return;
-    }
-    const auto& infoList = sib.getSamplerInfoList();
-    for (const auto& info : infoList) {
-        auto uniformName = filament::SamplerInterfaceBlock::getUniformName(
-                sib.getName().c_str(), info.name.c_str());
-        map[uniformName.c_str()] = map.size();
+    if (stageFlags.hasShaderType(config.shaderType)) {
```
bejado, line 56: Is there a size advantage to inverting this if statement? I much prefer using a guard clause.

bejado (approved):

libs/filamat/src/GLSLPostProcessor.cpp
pixelflinger: oh not at all, I thought it was more readable :-)

PR #5343, opened 2022-03-21
Moved reference to std::mutex out of public include header ostream.h …

### LINKED ISSUE #5331: OStream mutex addition breaks managed code (C++/CLI) compilations.
### Discussed in https://github.com/google/filament/discussions/5329

<div type='discussions-op-text'>

<sup>Originally posted by **BStringhamVRSK** March 17, 2022</sup>
@pixelflinger @prideout @bejado 
We are using Filament in a C# managed code environment.  **std::mutex** is not supported when C++ code is compiled with the /clr flag.  Because the std::mutex declaration appears in the include header file, it breaks our builds.

There are two ways forward that I can see:

1.  Move the mutex into ostream.cpp as a global.  That means all ostream instances will share the same mutex.  PROS:  This is probably the simplest change.  CONS:  This could have some impact on performance, but I don't know if it's a meaningful difference.
2. Create an **interface** for ostream.  That would require also creating a "factory" function to create ostream instances, rather than using the new operator.  PROS:  ostream would continue to work as before.  CONS:  Breaks any code currently using "new ostream()"

Any other thoughts or suggestions on this?  
I can do the programming changes, but I'd like to hear the code owners' opinions before diving in.
</div>

BStringhamVRSK:
I chose the simplest solution and moved the reference to **std::mutex** out of **ostream.h** and created a global in **ostream.cpp**.
Fixes #5331

### DIFF
--- a/libs/utils/include/utils/ostream.h
+++ b/libs/utils/include/utils/ostream.h
@@ -17,7 +17,6 @@
 #ifndef TNT_UTILS_OSTREAM_H
 #define TNT_UTILS_OSTREAM_H
 
-#include <mutex>
 #include <string>
 #include <utility>
 
@@ -86,7 +85,6 @@ protected:
         size_t capacity = 0;        // total capacity of the buffer
     };
 
-    std::mutex mLock;
     Buffer mData;
     Buffer& getBuffer() noexcept { return mData; }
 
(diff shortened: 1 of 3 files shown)


pixelflinger (changes requested):
using a global lock is not an acceptable solution.
instead we need to use a pimpl pattern for ostream.

pixelflinger:
working on a fix...

BStringhamVRSK:
Thanks for the fix @pixelflinger

PR #5349, opened 2022-03-22
Remove lite flavor from android-utils.

prideout:
We noticed that the lite flavor isn't actually smaller, plus we need to
rework our Android library hierarchy anyway, so for now let's remove
this.


romainguy (commented):
Have we figured out why lite was bigger?

prideout:
> Have we figured out why lite was bigger?

Unfortunately no, I gave up since it felt like beating a dead horse.  According to zbloat it is due to <stream> but I was unable to determine why this would occur in only the lite variant.

pixelflinger (approved):

bejado (approved):

PR #5351, opened 2022-03-22
skip work when morphtarget weight is zero

### LINKED ISSUE #5344: Rendering frame rate is low with a lot of morph targets
**Describe the bug**
My model has 244 morph targets and the frame rate is very low when rendered with filament, but not with babylon.

At the same time I found that as the number of morph target increases, the filament performance gets worse.

pixelflinger:
fixes #5344


romainguy (approved):

prideout (commented):
Thanks!  I would use `!= 0` since the spec allows signed types.

pixelflinger:
> Thanks! I would use `!= 0` since the spec allows signed types.

done!

PR #5354, opened 2022-03-23
Fix shadow frustum computation wrt projection matrix

pixelflinger:
Instead of recomputing the Frustum from its 8 vertices, we compute it
from the projection matrix directly. This ensures this works with any
projection matrix.  Building a Frustum object from 8 vertices required
these vertices to be given in a certain order which wasn't correct if
the projection was flipped for instance.

Removed the API to construct a Frustum from vertices.

fixes b/225975881


romainguy (approved):

prideout:
Nice fix!

PR #5355, opened 2022-03-23
Fix typo.

Michael-Lfx:


filament/src/RenderPass.cpp
```diff
@@ -444,7 +444,7 @@ void RenderPass::generateCommandsImpl(uint32_t extraFlags,
                     // correct for TransparencyMode::DEFAULT -- i.e. cancel the command
                     key |= select(mode == TransparencyMode::DEFAULT);
 
-                    // cancel command if asked to filter translucents
+                    // cancel command if asked to filter translucent
```
romainguy, line 447:
```suggestion
                    // cancel command if asked to filter translucent objects
```

romainguy (commented):
`translucents` is not a typo, more an abuse of language. I proposed something better

PR #5360, opened 2022-03-24
cleanup OpenGLProgram

pixelflinger:
This is the first step before we defer a lot of the glGet* calls.

### DIFF
--- a/filament/backend/src/opengl/OpenGLProgram.cpp
+++ b/filament/backend/src/opengl/OpenGLProgram.cpp
@@ -35,7 +35,7 @@ using namespace backend;
 
 static void logCompilationError(utils::io::ostream& out,
         backend::Program::Shader shaderType, const char* name,
-        GLuint shaderId) noexcept;
+        GLuint shaderId, CString sourceCode) noexcept;
 
 static void logProgramLinkError(utils::io::ostream& out,
         const char* name, GLuint program) noexcept;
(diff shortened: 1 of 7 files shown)


prideout (approved):

filament/backend/src/opengl/OpenGLProgram.cpp
```diff
@@ -35,43 +35,92 @@ using namespace backend;
...
+
+    // check status of program linking and shader compilation, logs error and free all resources
+    // in case of error.
+    // TODO: defer this until first use
+    bool success = OpenGLProgram::checkProgramStatus(name.c_str_safe(),
+            gl.program, gl.shaders);
+
+    // Failing to compile a program can't be fatal, because this will happen a lot in
+    // the material tools. We need to have a better way to handle these errors and
+    // return to the editor. Also note the early "return" statements in this function.
```
bejado, line 65: Hm, I don't see any `return` statements in `OpenGLProgram::OpenGLProgram`

bejado (approved):

filament/backend/src/opengl/OpenGLProgram.cpp
```diff
@@ -356,24 +400,6 @@ void logCompilationError(io::ostream& out, Program::Shader shaderType,
     out << "Compilation error in " << to_string(shaderType) << " shader \"" << name << "\":\n"
         << "\"" << error << "\""
         << io::endl;
-
-    size_t lc = 1;
```
romainguy, line 360: Any reason to remove this? It's quite useful for debugging to see the line numbers and the shader source when we dump an error.

pixelflinger: Yes, if we want to keep it, we'll have to keep the source of all programs until we get at least to the first draw.
Even if we do this, then we'll have to free that memory on first draw, but that would be (potentially many) calls to free() -- which could block, etc...

romainguy: Could we keep it in debug builds?

filament/backend/src/opengl/OpenGLProgram.cpp
pixelflinger: good catch.

filament/backend/src/opengl/OpenGLProgram.cpp
pixelflinger: yeah we could.

pixelflinger: done

romainguy (approved):

PR #5361, opened 2022-03-24
gltfio: add support for KHR_materials_variants 

prideout:
Tested the Java API by hacking ModelViewer.
Tested the JavaScript API by adding to filament-viewer-test.


web/samples/test-filament-viewer.html
```diff
@@ -83,7 +83,8 @@ <h3>Skybox, fixed size</h3>
 <h3>No skybox, no IBL, fixed size, round border</h3>
 <p>
 <filament-viewer
-    src="https://prideout.net/models/dino/trex.gltf"
+    src="https://prideout.net/models/MaterialsVariantsShoe.glb"
```
romainguy, line 86: We should prob not rely on your domain name btw.

romainguy (approved):
Looks great to me. Would it make sense to have an API to get the material/materialinstances associated with a given variant?

pixelflinger (approved):

prideout:
In a subsequent PR, I would like to add an applyMaterialVariant method to FilamentInstance.

prideout:
> Looks great to me. Would it make sense to have an API to get the material/materialinstances associated with a given variant?

Yes it might make sense to expose the MaterialInstance objects associated with a given variant, I might do that in my follow up PR.  (Makes less sense to expose Material, those are owned by the MaterialProvider and are widely shared.)

web/samples/test-filament-viewer.html
prideout: Maybe we should make an assets repo at some point.  I tried replacing these with blob URL's from the Khronos repo and ran into fun CORS errors:

```
Access to fetch at 'https://github.com/KhronosGroup/glTF-Sample-Models/raw/master/2.0/MaterialsVariantsShoe/glTF-Binary/MaterialsVariantsShoe.glb' from origin 'http://127.0.0.1:8080' has been blocked by CORS policy: The 'Access-Control-Allow-Origin' header has a value 'https://render.githubusercontent.com' that is not equal to the supplied origin. Have the server send the header with a valid value, or, if an opaque response serves your needs, set the request's mode to 'no-cors' to fetch the resource with CORS disabled.
```

PR #5363, opened 2022-03-25
Add Java binding to get skins in FilamentAsset

### LINKED ISSUE #5290: How can I get skins in glTF?

daemyung:
Fixes #5290


bejado (approved):

prideout (changes requested):
@daemyung we usually expose lists as Java arrays, we only expose them this way in C++ to avoid including STL containers in header files.  Do you think you could change these Java bindings to return arrays, like we do for the new variant names?

daemyung:
@prideout Ok no problem. However before doing it, I want to clear what I have to change.

From:

```
public int getJointCountAt(int skinIndex);
```

To: 

```
public int[] getJointCount;
```

Is this change what you want?

romainguy:
If you are going to return an array you need to accept an array as input. If the input is null, return a new array, otherwise use the one passed as a parameter. You also need to check the array's length and throw if too small.

prideout:
The existing precedent in gltfio is to create arrays from scratch, under the assumption that queries are infrequent.  This unfortunately allows users to kill performance, so we should revisit it in the future.

For now, I would prefer that this class be self-consistent.

Further complicating matters is the fact that skinning joints are conceptually arrays-of-arrays, which I'd rather not deal with.

I propose the following, just two methods:

```
public @NonNull String[] getSkinNames() { ... }
public @NonNull @Entity int[] getJointsAt(int skinIndex) { ... }
```

While not ideal, it's at least similar to existing gltfio Java API's.

daemyung:
@prideout Thanks for your guidance. It's updated and please check it.

prideout (changes requested):
Apologies for the delay, I forgot about this PR.  It looks good except that  `nGetMorphTargetNameAt` was added, it does not exist.

daemyung:
@prideout No worries and I fixed my huge mistake. Thanks for pointing out.

prideout (approved):

PR #5364, opened 2022-03-25
defer gl program link + query to first use (draw)

pixelflinger:
in an upcoming PR we'll move linking back a bit earlier to allow
more batching.

### DIFF
--- a/filament/backend/src/opengl/OpenGLProgram.h
+++ b/filament/backend/src/opengl/OpenGLProgram.h
@@ -38,12 +38,17 @@ class OpenGLProgram : public backend::HwProgram {
 public:
 
     OpenGLProgram() noexcept;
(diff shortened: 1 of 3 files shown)


romainguy (approved):

filament/backend/src/opengl/OpenGLProgram.h
```diff
@@ -99,19 +106,41 @@ class OpenGLProgram : public backend::HwProgram {
         static_assert(backend::Program::BINDING_COUNT <= 8, "BINDING_COUNT must be <= 8");
     };
 
+    // keep these sway from of other class attributes
```
prideout, line 109:
```suggestion
    // keep these away from of other class attributes
```

PR #5366, opened 2022-03-25
gltfio: material variants can now be applied to FilamentInstance.

prideout:
instanced shoes

### DIFF
--- a/libs/gltfio/src/AssetLoader.cpp
+++ b/libs/gltfio/src/AssetLoader.cpp
@@ -140,6 +140,8 @@ struct FAssetLoader : public AssetLoader {
     void addTextureBinding(MaterialInstance* materialInstance, const char* parameterName,
             const cgltf_texture* srcTexture, bool srgb);
     bool primitiveHasVertexColor(const cgltf_primitive* inPrim) const;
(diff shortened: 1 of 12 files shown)


romainguy (approved):

libs/gltfio/src/AssetLoader.cpp
```diff
@@ -572,6 +566,36 @@ void FAssetLoader::createRenderable(const cgltf_data* srcAsset, const cgltf_node
     }
 }
 
+void FAssetLoader::createMaterialVariants(const cgltf_data* srcAsset, const cgltf_mesh* mesh,
+        Entity entity, FFilamentInstance* instance) {
+    UvMap uvmap {};
+    for (cgltf_size prim = 0, n = mesh->primitives_count; prim < n; ++prim) {
+        const cgltf_primitive* srcPrim = &mesh->primitives[prim];
+        for (size_t i = 0, n = srcPrim->mappings_count; i < n; i++) {
```
bejado, line 574: Use a different loop variable other than `n` for clarity?

bejado (approved):

libs/gltfio/src/AssetLoader.cpp
prideout: Good catch, fixed.

pixelflinger (approved):

PR #5367, opened 2022-03-25
programs must now be created outside of renderpasses

pixelflinger:
from now on, the backends can assume that programs are created outside
of begin/endRenderpass.


prideout (approved):
`Material::loadProgram` should perhaps have a comment explaining what it does and when it should be called.  Maybe it could be renamed to `prepareProgram`?

bejado (approved):

pixelflinger:
> `Material::loadProgram` should perhaps have a comment explaining what it does and when it should be called. Maybe it could be renamed to `prepareProgram`?

done!

prideout (approved):

PR #5368, opened 2022-03-25
batch all program linking at begin renderpass time

pixelflinger:
With this change, shaders are compiled at program creation, programs
are linked at the start of the next render pass and queries are
deferred until first use.

This should improve overall performance on some drivers.

### DIFF
--- a/filament/backend/src/opengl/OpenGLDriver.cpp
+++ b/filament/backend/src/opengl/OpenGLDriver.cpp
@@ -1215,6 +1215,7 @@ void OpenGLDriver::destroyProgram(Handle<HwProgram> ph) {
     DEBUG_MARKER()
     if (ph) {
         OpenGLProgram* p = handle_cast<OpenGLProgram*>(ph);
+        cancelRunAtNextPassOp(p);
         destruct(ph, p);
     }
 }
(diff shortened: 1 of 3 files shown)


filament/backend/src/opengl/OpenGLDriver.cpp
```diff
@@ -2927,6 +2931,25 @@ void OpenGLDriver::executeEveryNowAndThenOps() noexcept {
...
+    mRunAtNextRenderPassOps[token] = std::move(fn);
+}
+
+void OpenGLDriver::cancelRunAtNextPassOp(void* token) noexcept {
+    mRunAtNextRenderPassOps.erase(token);
+}
+
+void OpenGLDriver::executeRenderPassOps() noexcept {
+    auto& ops = mRunAtNextRenderPassOps;
+    if (!ops.empty()) {
```
prideout, line 2945: UTILS_UNLIKELY maybe?

prideout (approved):

filament/backend/src/opengl/OpenGLDriver.cpp
pixelflinger: I was going to use it and then I thought executeRenderPassOps() is actually generic, so we can't really tell if it's likely or not.

PR #5371, opened 2022-03-27
Fix a typo.

Michael-Lfx:


pixelflinger:
The use of "a" is correct in this case.

PR #5379, opened 2022-03-29
Fix 12x overallocation of memory in MorphTargetBuffer.

prideout:
I noticed this bug while helping someone at Google understand our new morph API.

### DIFF
--- a/filament/src/details/MorphTargetBuffer.cpp
+++ b/filament/src/details/MorphTargetBuffer.cpp
@@ -77,7 +77,7 @@ inline size_t getSize(size_t vertexCount) noexcept;
 
 template<>
 inline size_t getSize<VertexAttribute::POSITION>(size_t vertexCount) noexcept {
-    const size_t stride = getWidth(vertexCount) * sizeof(float3);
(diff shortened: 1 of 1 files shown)


romainguy (approved):
Nice catch ^^

bejado (approved):

filament/src/details/MorphTargetBuffer.cpp
```diff
@@ -77,22 +77,22 @@ inline size_t getSize(size_t vertexCount) noexcept;
 
 template<>
 inline size_t getSize<VertexAttribute::POSITION>(size_t vertexCount) noexcept {
-    const size_t stride = getWidth(vertexCount) * sizeof(float3);
+    const size_t strideInPixels = getWidth(vertexCount);
```
pixelflinger, line 80: I'm okay with renaming to `strideInPixels` but this *is* the convention we use everywhere in filament. In the rare case that we need to deal with a stride that is in byte, we explicitly call it `strideInBytes`.
So to me, this is not necessary and different from other places in filament.

pixelflinger (approved):
good catch!

PR #5386, opened 2022-03-31
repair and optimize matdbg

pixelflinger:
we add a checkProgramEdits() method that must be called periodically
on all materials. This is currently done at the beginning of a frame.


romainguy (approved):

prideout (approved):
Looks GREAT, thanks for fixing this!

PR #5389, opened 2022-04-01
one more step towards reflection+refraction support

pixelflinger:
With this change we are now generating refraction and reflections
buffer into their own layer of a 2D array -- so they can coexist.

This change doesn't actually enable both at the same time yet.

There are downsides to this approach:
- a 2d array of 2 layers is created even if only one of reflection or
  refraction is active. This could be fixed at the cost of more
  complexity.
- if reflections are active then refraction must use a RGBA16F texture
  instead of RGB_11_11_10, because they share the texture


bejado (approved):
This reminds me, did we do a material version bump for `light_ssr` now being a 2d texture array?

pixelflinger:
> This reminds me, did we do a material version bump for `light_ssr` now being a 2d texture array?

mh... I think not.

PR #5390, opened 2022-04-01
enable both screen-space refraction and reflection

pixelflinger:
Both "SSR" can now be used on the same material. This is made possible
by using a 2D array to store both buffers
ssr
.


prideout:
🍾

prideout (approved):

bejado (approved):

PR #5391, opened 2022-04-01
update release notes and bump material version

pixelflinger:


bejado (changes requested):
It looks like 1.20.6 is the version that introduced the change from `SAMPLER_2D` to `SAMPLER_2D_ARRAY`. So we should really bump that version.

bejado:
Actually, it looks like the original commit 85135e60c7 made it into 1.20.5, which has already been released. So we should have bumped that version.

bejado (approved):

PR #5394, opened 2022-04-03
Use index offset provided by ImGui when rendering UI

borodust:
This is the fix for the same problem in rendering code as described in https://github.com/ocornut/imgui/issues/4863


prideout (approved):
Excellent, thanks!

PR #5397, opened 2022-04-05
Remove deprecated Stream APIs

pixelflinger:


RELEASE_NOTES.md
```diff
@@ -6,6 +6,7 @@ A new header is inserted each time a *tag* is created.
 ## v1.21.1 (currently main branch)
 
 - engine: allow both screen-space refraction and screen-space reflections on the same object [⚠️ **Material breakage**].
+- remove deprecated `Stream` APIs [⚠️ **API Change**].
```
bejado, line 9: Please rebase on main. This should go under v1.21.2.

bejado (changes requested):

kpeeters:
This means that you now require API 26, right? Any way you might reconsider this decision? It removes quite a large chunk of devices still out there.

prideout (commented):
We need to update `StreamHelper.kt` and make sure that `sample-stream-test` still builds and runs.

pixelflinger:
> We need to update `StreamHelper.kt` and make sure that `sample-stream-test` still builds and runs.

How come everything built?

pixelflinger:
> This means that you now require API 26, right? Any way you might reconsider this decision? It removes quite a large chunk of devices still out there.

Sorry why is that? I think the solution that uses `import()` works everywhere.

bejado (approved):

prideout (approved):

pixelflinger:
converted to draft until we're ready to submit.

bejado:
Please rebase on main and update the release notes for 1.21.3

PR #5398, opened 2022-04-05
Attempt to fix GitHub release asset upload

bejado:


.github/workflows/release.yml
```diff
@@ -50,6 +50,7 @@ jobs:
           cd build/$WORKFLOW_OS && printf "y" | ./build.sh release
       - name: Upload release assets
         run: |
+          pip3 install wheel
```
romainguy, line 53: Should we do something like ``pip3 install `cat release_dependencies.txt` `` at some point?

romainguy (approved):

.github/workflows/release.yml
bejado: Yeah I was thinking something like that too. Unfortunately I'm not sure if this will even fix things yet but once I get it working again I'll revisit.

PR #5411, opened 2022-04-07
Shrink BasisU transcoder.

prideout:
We do not support these formats and they add bloat to the basis tables.
In a release x86 build, this reduces the transcoder size from 774 KiB to
475 KiB.

(This library is not yet used, stay tuned.)

### DIFF
--- a/third_party/basisu/tnt/CMakeLists.txt
+++ b/third_party/basisu/tnt/CMakeLists.txt
@@ -2,6 +2,11 @@ project(basisu)
 
 cmake_minimum_required(VERSION 3.19)
 
+set(TRANSCODER_SRC
+    ../transcoder/basisu_transcoder.cpp
+    ../zstd/zstd.c
+)
+
 set(ENCODER_SRC
     ../encoder/basisu_backend.cpp
     ../encoder/basisu_basis_file.cpp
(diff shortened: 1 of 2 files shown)


bejado (approved):

third_party/basisu/tnt/CMakeLists.txt
```diff
@@ -24,19 +24,33 @@ set(ENCODER_SRC
...
 
+# Filament does not support certain compressed formats that BasisU can transcode
+# to, so we disable them in the build to save space. Some of these mappings are
+# confusing. (e.g., DXT5A corresponds to transcoder_texture_format::cTFBC4_R).
+set (BASIS_CONFIG
+    BASISD_SUPPORT_KTX2=1
+    BASISD_SUPPORT_KTX2_ZSTD=1
+    BASISD_SUPPORT_ATC=0
+    BASISD_SUPPORT_DXT5A=0
+    BASISD_SUPPORT_BC7=0
```
romainguy, line 35: Isn't BC7 super useful on desktop?

prideout: Filament does not support BC7 yet.  Here's a quote from the BasisU wiki:

> If you know your platform doesn't support BC7 or it's not necessary, be sure to set BASISD_SUPPORT_BC7=0. The tables for the mode 6 transcoder are very large (they bloat the compiled WebAssembly transcoder by around 400-500k). We'll probably be removing the mode 6 transcoder in a future update.

pixelflinger (approved):

PR #5418, opened 2022-04-11
Export Metal Device to public.

Michael-Lfx:
It's necessary to share the same Metal Device for the rendering pipeline when using filament as a part of offscreen rendering frameworks in a video processing app.

### DIFF
--- a/filament/backend/include/backend/Platform.h
+++ b/filament/backend/include/backend/Platform.h
@@ -66,6 +66,8 @@ public:
      * thread, or if the platform does not need to perform any special processing.
      */
     virtual bool pumpEvents() noexcept { return false; }
+    
+    virtual void* getMTLDevice() noexcept { return nullptr; }
 };
 
 
--- a/filament/backend/include/private/backend/MetalPlatform.h
+++ b/filament/backend/include/private/backend/MetalPlatform.h
@@ -32,7 +32,7 @@ public:
     Driver* createDriver(void* sharedContext) noexcept override;
     int getOSVersion() const noexcept override { return 0; }
     
-    void* getMTLDevice() noexcept override;
+    void* getMTLDevice() noexcept override { return mDevice; }
 
     /**
      * Obtain the preferred Metal device object for the backend to use.
@@ -59,6 +59,7 @@ public:
 
 private:
     id<MTLCommandQueue> mCommandQueue = nil;
+    id<MTLDevice> mDevice = nil;
 
 };
 
--- a/filament/backend/include/private/backend/MetalPlatform.h
+++ b/filament/backend/include/private/backend/MetalPlatform.h
@@ -31,6 +31,8 @@ public:
 
     Driver* createDriver(void* sharedContext) noexcept override;
     int getOSVersion() const noexcept override { return 0; }
+    
+    void* getMTLDevice() noexcept override;
 
     /**
      * Obtain the preferred Metal device object for the backend to use.
--- a/filament/backend/include/private/backend/MetalPlatform.h
+++ b/filament/backend/include/private/backend/MetalPlatform.h
@@ -32,7 +32,7 @@ public:
     Driver* createDriver(void* sharedContext) noexcept override;
     int getOSVersion() const noexcept override { return 0; }
     
-    void* getMTLDevice() noexcept override { return mDevice; }
+    void* getMTLDevice() noexcept override { return (__bridge void *)mDevice; }
 
     /**
      * Obtain the preferred Metal device object for the backend to use.
--- a/filament/backend/src/metal/MetalPlatform.mm
+++ b/filament/backend/src/metal/MetalPlatform.mm
@@ -55,6 +55,8 @@ id<MTLDevice> MetalPlatform::createDevice() noexcept {
     {
         result = MTLCreateSystemDefaultDevice();
     }
+    
+    mDevice = result;
 
     utils::slog.i << "Selected physical device '"
                   << [result.name cStringUsingEncoding:NSUTF8StringEncoding] << "'"


filament/backend/include/backend/Platform.h
```diff
@@ -66,6 +66,8 @@ class UTILS_PUBLIC Platform {
      * thread, or if the platform does not need to perform any special processing.
      */
     virtual bool pumpEvents() noexcept { return false; }
+    
+    virtual void* getMTLDevice() noexcept { return nullptr; }
```
pixelflinger, line 70: no we can't have a metal specific API on Platform. This would have to go on MetalPlatform only.

pixelflinger (changes requested):

pixelflinger:
I'm not comfortable with this addition. Please provide more context.
I think it would be better to start by filing an issue (or a discussion) and decide there how we address the problem.

filament/backend/include/backend/Platform.h
Michael-Lfx: MetalPlatform.h is a private header. We can't access it unless making it public.

romainguy: Could you describe your use case in more details then? We keep the backends private for good reasons.

Michael-Lfx: There are multiple rendering framework in our video processing app, snapchat/TikTok for example. The whole frame is consist of 2D renderer(such as GPUImage), 3D renderer(such as filament) and then 2D renderer(such as GPUImage). It's good to keep the same physical devices such MTLDevice and EGLContext between them. Although filament provide Platform creation, we just can't use those gpu hardware it created. Hence here's a proposal to make it accessible.

pixelflinger: We don't think "it's good" to share contextes.  In particular sharing the opengl context means your app (or other libraries) can screw-up our context, leading to impossible to debug bugs. There are also threading implications.

I still don't understand what you cannot do with the current system.

bejado: @Michael-Lfx assume Filament always simply called `MTLCreateSystemDefaultDevice`—if you also called `MTLCreateSystemDefaultDevice`, would that solve your problems (at least for Metal)?

Michael-Lfx: > We don't think "it's good" to share contextes. In particular sharing the opengl context means your app (or other libraries) can screw-up our context, leading to impossible to debug bugs. There are also threading implications.
> 
> I still don't understand what you cannot do with the current system.

How can we share GPU texture between different rendering frameworks which owns their own E(A)GLContext? ReadPixels back to CPU and then upload to GPU to pass data between them will greatly impact performance. 

Filament support `static Engine* create(Backend backend = Backend::DEFAULT, Platform* platform = nullptr, void* sharedGLContext = nullptr)` to make sure resources created inside filament are visible to caller. Say we Texture::import(a_gl_tex_id) to be a sample source or color attachment for framebuffer.

Same situation for Metal and Vulkan. Using the same MTLDevice/VkPhysical/VkDevice to create GPU resources are visible for other rendering frameworks to sample or draw to filament created textures/buffers.

romainguy: The support for a shared EGL context is something we did reluctantly for legacy code at Google. The goal was never to make Filament resources available externally, but the other way around. And even then, there are ways to share texture data efficiently via other mechanisms that don't require context sharing.

Michael-Lfx: > @Michael-Lfx assume Filament always simply called `MTLCreateSystemDefaultDevice`—if you also called `MTLCreateSystemDefaultDevice`, would that solve your problems (at least for Metal)?

Yes. Calling `MTLCreateSystemDefaultDevice ` both on iOS and macOS can solve my problem. How to suit the Android Vulkan case? Our app is running on macOS, iOS, Android and Windows.

Another plan is changing 
`static Engine* create(Backend backend = Backend::DEFAULT, Platform* platform = nullptr, void* sharedGLContext = nullptr)`
to
```
struct EngineCreateInfo {
    Backend backend = Backend::DEFAULT;
    Platform* platform = nullptr;
    void* sharedGLContext = nullptr;
    // For Metal
    void* mtlDevice = nullptr;
    // For Vulkan (maybe missing VkPhysicalDevice/VkInstance)
    void* vkDevice = nullptr;
};
static Engine* create(EngineCreateInfo &createInfo);
```

romainguy: *If* the goal was to extend sharing to Vulkan and Metal, you could repurpose the existing pointer parameter, there's no need for an entire structure. As for sharing on Android, you should be rendering into `Surface` objects that come from a `SurfaceTexture` or an `ImageReader` and use that. There shouldn't be a reason to share the contexts.

Michael-Lfx:
> I'm not comfortable with this addition. Please provide more context.
> I think it would be better to start by filing an issue (or a discussion) and decide there how we address the problem.

Sorry!

filament/backend/include/backend/Platform.h
Michael-Lfx: Sounds like `sharedGLContext` of `static Engine* create(Backend backend = Backend::DEFAULT, Platform* platform = nullptr, void* sharedGLContext = nullptr)` should be interpret by backends. It can be cast from `void*` to `id<MTLDevice>` inside `Driver* MetalPlatform::createDriver(void* sharedContext) noexcept`.

pixelflinger: I really would like to move this discussion to a proper discussion on GitHub so it can be useful to other people.

We should discuss this before we implement solutions that come with their own problems.

I'm going to close this PR. Again let's discuss the problem, and figure out what existing facilities you can use.

PR #5420, opened 2022-04-11
mipgen: fixups / clarification regarding sRGB.

prideout:
The `g_linearized` variable was being used for two purposes: to denote
the linearity of the source format AND the linearity of the destination
format. This was confusing and is now split is `sourceIsLinear` and
`destIsLinear`.

This change has no effect on the the look of the suzanne demo.

### DIFF
--- a/tools/mipgen/src/main.cpp
+++ b/tools/mipgen/src/main.cpp
@@ -48,7 +48,7 @@ static bool g_addAlpha = false;
 static bool g_stripAlpha = false;
 static bool g_grayscale = false;
 static bool g_ktxContainer = false;
-static bool g_linearized = false;
+static bool g_sourceIsLinear = false;
 static bool g_quietMode = false;
 static uint32_t g_mipLevelCount = 0;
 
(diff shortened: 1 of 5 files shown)


tools/mipgen/src/main.cpp
```diff
@@ -71,13 +71,13 @@ If the output format is a container format like KTX, then
    --license, -L
        print copyright and license information
    --linear, -l
-       assume that image pixels are already linearized
+       specifies that the source image transforms linearly to floats
```
romainguy, line 74: What do you mean by this? That the data is linear or that it will go through an transfer function?

prideout: Heh, I thought I was making it _more_ clear but you're right it has "transform" in there which is confusing.  I meant transform in the sense of bytes-to-floats.  How about we say simply: "specifies that the source image is linear".

pixelflinger: you can say "converts to float" instead of "transform"?

PR #5423, opened 2022-04-12
Fix sRGB correctness in Ktx1Reader.

### LINKED ISSUE #2422: mipgen linearized/srgb issue
It looks like if you are writing ktx with astc compression from mipgen, that the `.srgb` parameter in the astc configuration is not initialized or set? In the case of a non-ktx file, `g_linearized` is used to decide if the final data output is `fromLinearToRGB` or `fromLinearTosRGB`... however I don't see matching logic to control astc output from ktx code. I think the files are always being created as linear, which would be a quality loss for color images.

Perhaps I'm missing something somewhere, but it seems like a line along the lines of:
```
config.astc.srgb = !g_linearized;
```
is required after `parseOptionString` to output the correct color space? Or perhaps input params for astc compression need expanding to `astc_[fast|thorough]_[srgb|linear]_[ldr|hdr]_WxH`?

prideout:
When creating a KTX file (e.g. with mipgen), you can of course choose
whether or not you want to use an sRGB format.

At load time, your app has an expectation about whether a particular
texture has an sRGB format. e.g. the glTF convention is that albedo
textures should be sRGB, while roughness textures are linear.

Before this commit, our KTX loader would look at the app's expectation,
then silently re-interpret the bits in the glTF texture in order to
match this expectation. This seems very wrong.

Instead, if the expectated format does not match the format in the KTX
file, we now print a warning.

Fixes #2422, which will also be obsoleted by our upcoming KTX2 support.

Suzanne Demo Before / After:

Screen Shot 2022-04-12 at 10 20 31 AM

### DIFF
--- a/libs/ktxreader/include/ktxreader/Ktx1Reader.h
+++ b/libs/ktxreader/include/ktxreader/Ktx1Reader.h
@@ -55,7 +55,7 @@ namespace Ktx1Reader {
      *
      * @param engine Used to create the Filament Texture
      * @param ktx In-memory representation of a KTX file
-     * @param srgb Forces the KTX-specified format into an SRGB format if possible
+     * @param srgb Requests an sRGB format from the KTX file
      * @param callback Gets called after all texture data has been uploaded to the GPU
      * @param userdata Passed into the callback
      */
@@ -68,7 +68,7 @@ namespace Ktx1Reader {
      *
      * @param engine Used to create the Filament Texture
      * @param ktx In-memory representation of a KTX file
-     * @param srgb Forces the KTX-specified format into an SRGB format if possible
+     * @param srgb Requests an sRGB format from the KTX file
      */
     Texture* createTexture(Engine* engine, Ktx1Bundle* ktx, bool srgb);
 
@@ -80,7 +80,7 @@ namespace Ktx1Reader {
 
     bool isCompressed(const KtxInfo& info);
 
-    TextureFormat toSrgbTextureFormat(TextureFormat format);
+    bool isSrgbTextureFormat(TextureFormat format);
 
     TextureFormat toTextureFormat(const KtxInfo& info);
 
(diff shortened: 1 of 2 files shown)


romainguy (approved):
This reminds me that at some point we should cleanup our naming conventions for sRGB and linear. We're not doing anything "wrong" but we make the same mistake as everybody else of simplifying the names to the point that it creates confusion. This will become problematic if we ever add proper support for wide-gamut inputs. Ideally we'd use fully-qualified colorspace names: Rec.7090-sRGB-D65 (== "sRGB") and Rec.709-Linear-D65 (== "linear"). Then with wide-gamut you'd have P3-sRGB-D65, P3-Linear-D65, etc. And output becomes clearer too: Rec.709-Linear-D65 for the output of the lighting pass, and Rec.7090-sRGB-D65 for the color grading output, but it could be P3-sRGB-D65 for wide-gamut, or even Rec.2020-PQ-D65 for HDR.

Anyway, this PR looks great, thanks for fixing this!

pixelflinger:
> This reminds me that at some point we should cleanup our naming conventions for sRGB and linear. We're not doing anything "wrong" but we make the same mistake as everybody else of simplifying the names to the point that it creates confusion. This will become problematic if we ever add proper support for wide-gamut inputs. Ideally we'd use fully-qualified colorspace names: Rec.7090-sRGB-D65 (== "sRGB") and Rec.709-Linear-D65 (== "linear"). Then with wide-gamut you'd have P3-sRGB-D65, P3-Linear-D65, etc. And output becomes clearer too: Rec.709-Linear-D65 for the output of the lighting pass, and Rec.7090-sRGB-D65 for the color grading output, but it could be P3-sRGB-D65 for wide-gamut, or even Rec.2020-PQ-D65 for HDR.

So in this naming scheme, "sRGB" stands for the gamma curve -- not the primaries?

romainguy:
The sRGB spec uses the Rec.709 primaries but a different transfer function. Whereas DisplayP3 for instance uses the sRGB transfer function and white point but the primaries of DCI-P3.

Issue #4771, opened 2021-10-27
Support KTX2 texture


0k1019:
**Is your feature request related to a problem? Please describe.**

Filament can't render GLB/GLTB file containing KTX2 texture.
The explanation of [KTX2](https://www.khronos.org/ktx/) is in Khronos group website.
[Examples GLB files](https://github.com/KhronosGroup/3D-Formats-Guidelines/tree/main/samples/StainedGlassLamp) containing KTX2.

**Describe the solution you'd like**
I wish filament supports to render KTX2 texture. KTX2 is really nice for rendering quality and compression 3D file.

**Describe alternatives you've considered**
[cgltf.h](https://github.com/google/filament/blob/1182d30fb909fee4af1d9ebab14c15dd434f3655/third_party/cgltf/cgltf.h#L3964): This line means GLB with ktx2 texture convert to `basisu_image` not `image`.

[AssetLoader.cpp](https://github.com/google/filament/blob/1182d30fb909fee4af1d9ebab14c15dd434f3655/libs/gltfio/src/AssetLoader.cpp#L1210): This line means `basisu_image` can't load to render. 

Accept `basisu_image` in AssetLoader.

Testing in [filament viewer](https://google.github.io/filament/viewer/) between [GLB(with png)](https://raw.githubusercontent.com/KhronosGroup/3D-Formats-Guidelines/main/samples/StainedGlassLamp/StainedGlassLamp-JPG-PNG-Volume.glb) and [GLB(with KTX2)](https://raw.githubusercontent.com/KhronosGroup/3D-Formats-Guidelines/main/samples/StainedGlassLamp/StainedGlassLamp-KTX-BasisU-Volume.glb)
![image](https://user-images.githubusercontent.com/11485593/138989119-c90a8c31-e8e0-4f86-90c7-9e0467fdf399.png)

**OS and backend**
macOS Big Sur 11.6
Chrome 95.0.4638.54 (x86_64) 

Thank you for @codpot's help.

August1996:
May I ask which version of the features of ktx2 is planned to support?😊

prideout:
We intend to fully support `KHR_texture_basisu` and use that as the driving force for our ktx2 support.  The only reason we would drop features would be to reduce binary size.

August1996:
> We intend to fully support `KHR_texture_basisu` and use that as the driving force for our ktx2 support. The only reason we would drop features would be to reduce binary size.

very good feature👍. have you started developing this feature?

[...]

PR #5431, opened 2022-04-13
Add Ktx2Reader and BasisEncoder and use them in samples.

prideout:
mipgen can now emit basis-encoded KTX2 files. Both the desktop and
web "suzanne" samples use this as a test for compressed textures.

This PR does not add KTX2 support to glTF, but it's on the way.

`BasisEncoder` has a builder style API that calls the basis encoder to
create KTX2 files. This hides some low-level BasisU features that we are
not using, like file I/O and mipmap generation.

`Ktx2Reader` is an easy-to-use API for creating Filament textures from
KTX2 files. Its API primarily consists of these two methods:

    bool requestFormat(Texture::InternalFormat format);
    Filament::Texture* load(const uint8_t* data, size_t size);

The first method is used to build an ordered list of formats that are
supported by your hardware. The second method consumes the contents of a
basis-encoded KTX2 file and attempts to produce a Filament texture with
a preferred format.

IMPORTANT: Our tools still let you use KTX1 for non-compressed images
because it is useful for HDR, but you can no longer use KTX1 for
block-compressed data.

Partial fix for #4771.


libs/imageio/src/BasisEncoder.cpp
```diff
@@ -0,0 +1,267 @@
...
+BasisEncoder::BasisEncoder(BasisEncoderImpl* impl) noexcept : mImpl(impl) {}
+
+BasisEncoder::~BasisEncoder() noexcept {
+    delete mImpl->encoder;
+    delete mImpl->jobs;
+    delete mImpl;
+    basisu::basisu_encoder_deinit();
+}
+
+BasisEncoder::BasisEncoder(BasisEncoder&& that) noexcept {
```
bejado, line 203: I think you need to initialize `mImpl` here, otherwise it doesn't get initialized.

libs/ktxreader/include/ktxreader/Ktx2Reader.h
```diff
@@ -0,0 +1,92 @@
...
+    class ktx2_transcoder;
+}
+
+namespace ktxreader {
+
+class Ktx2Reader {
+    public:
+        using Engine = filament::Engine;
+        using Texture = filament::Texture;
+        enum TransferFunction { LINEAR, sRGB };
```
bejado, line 41: `enum class`?

libs/ktxreader/include/ktxreader/Ktx2Reader.h
```diff
@@ -0,0 +1,92 @@
...
+
+        Ktx2Reader(Engine& engine, bool quiet = false);
+        ~Ktx2Reader();
+
+        /**
+         * Requests that the reader constructs Filament textures with given internal format.
+         *
+         * This MUST be called at least once before calling load().
```
bejado, line 49: Could we instead specify a default format?

prideout: There might be cases where the caller considers the use of a certain formats to be "wrong" so I feel we shouldn't make assumptions.  We could maybe make it less tedious for simple cases, by allowing a format to be specified in the constructor.  The problem there is that I'm not sure what its priority should be.  Since formats are requested in priority order, would the default format be highest priority or lowest priority?  Probably the latter, but I don't think the API would be as obvious and self-explanatory as it is now.

Thanks for the great feedback in this PR by the way!  Working on fixing all the other stuff you noticed.

bejado: Got it, makes sense! 👍

libs/imageio/src/BasisEncoder.cpp
prideout: Addressed this and the other items in the latest commit.

bejado (approved):
🎉

PR #5433, opened 2022-04-14
propagate the matdbg flag to graddle

pixelflinger:


android/build.gradle
```diff
@@ -59,6 +59,11 @@ buildscript {
             .forUseAtConfigurationTime()
             .isPresent()
 
+    def matdbg = providers
```
romainguy, line 62: You should also document that property in the big comment at the top of the file

romainguy (approved):

android/build.gradle
pixelflinger: done

PR #5434, opened 2022-04-14
fix bloom downscaling (+ minor improvements)

pixelflinger:


filament/src/materials/bloom/bloomDownsample.mat
```diff
@@ -65,24 +60,27 @@ fragment {
...
+        vec3 lt2 = textureLodOffset(materialParams_source, uv, lod, ivec2(-2, -2)).rgb;
+        vec3 rt2 = textureLodOffset(materialParams_source, uv, lod, ivec2( 2, -2)).rgb;
+        vec3 rb2 = textureLodOffset(materialParams_source, uv, lod, ivec2( 2,  2)).rgb;
+        vec3 lb2 = textureLodOffset(materialParams_source, uv, lod, ivec2(-2,  2)).rgb;
 
-        vec3 l = textureLodOffset(materialParams_source, uv, lod, ivec2(-1,  0)).rgb;
-        vec3 t = textureLodOffset(materialParams_source, uv, lod, ivec2( 0, -1)).rgb;
-        vec3 r = textureLodOffset(materialParams_source, uv, lod, ivec2( 1,  0)).rgb;
-        vec3 b = textureLodOffset(materialParams_source, uv, lod, ivec2( 0,  1)).rgb;
+        vec3 l   = textureLodOffset(materialParams_source, uv, lod, ivec2(-2,  0)).rgb;
```
romainguy, line 80: Is it really supposed to be -2 on those?

romainguy (approved):

filament/src/materials/bloom/bloomDownsample.mat
pixelflinger: yes, that's the fix. 2 texels of "source" (texture) == 1 texel of "dest" (target) -- this is the downscale case.

PR #5435, opened 2022-04-14
don't issue a flush systematically after framegraph's execute

pixelflinger:
This is only a good place for a flush *if* we're going to draw another
view (so the GPU can start working). Otherwise, it can be costly
on some drivers and unnecessary since swapbuffers will do the same.

### DIFF
--- a/filament/src/details/Renderer.cpp
+++ b/filament/src/details/Renderer.cpp
@@ -186,13 +186,21 @@ void FRenderer::render(FView const* view) {
 
     assert_invariant(mSwapChain);
 
-    if (mBeginFrameInternal) {
+    if (UTILS_UNLIKELY(mBeginFrameInternal)) {
+        // this should not happen, the user should not call render() if we returned false from
+        // beginFrame(). But because this is allowed, we handle it gracefully.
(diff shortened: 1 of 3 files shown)


filament/src/details/Renderer.cpp
```diff
@@ -186,13 +186,21 @@ void FRenderer::render(FView const* view) {
...
+        // beginFrame(). But because this is allowed, we handle it gracefully.
         mBeginFrameInternal();
         mBeginFrameInternal = {};
     }
 
     if (UTILS_LIKELY(view && view->getScene())) {
+        if (mViewRenderedCount) {
+            // this is a good place to kick the GPU, since we've rendered a View before
+            // and we're about to render another one.
+            mEngine.getDriverApi().flush();
```
romainguy, line 200: Will this be done for every view or just after the first one?

romainguy (approved):

filament/src/details/Renderer.cpp
pixelflinger: This is only done starting with the 2nd view (and obviously not done for the last one -- which is a better way to look at it).

PR #5437, opened 2022-04-15
Remove WebGL workaround for bloom.

prideout:
Bloom now works fine (tested with Chrome 100 on a x86 MacBook) without
this workaround for overly aggressive feedback loop detection.


prideout:
Note: do not merge, sounds like we still need this for ANGLE.

pixelflinger (approved):

PR #5440, opened 2022-04-15
Audit backends and test cases for correct Y orientation.

bejado:
This PR:
- switches the test_blit test image from a circle to triangle to make failure cases more obvious
- for backend tests, flips image data vertically when uploading into a `Texture` so we can correctly `readPixel` it
- adds some documentation diagrams to help clarify the pipeline

Here's what now works across all backends:
1. Rendering into a `SwapChain` and then issuing a `readPixels`.
1. Rendering into a `RenderTarget` backend by a `Texture` and then issuing a `readPixels` (this previousy did not work for Vulkan).

The downside:

With OpenGL only, uploading image data into a `Texture` and then issuing a `readPixels` on that `Texture` (or any `Texture`s that it has been blitted into) will return y-flipped data. This is due to how OpenGL flips image data during upload in order to match its inverted texture coordinates. Thankfully this is probably not a common operation for clients, however, it affects backend tests. The fix is to flip the image data before upload.


pixelflinger (approved):
thanks! very nice!

prideout (approved):
Nice!  I tried running gltf_viewer batch mode against your PR with and without `--headless` and everything looks good.  :)

PR #5447, opened 2022-04-18
a lot of our workaround don't apply when using ANGLE

pixelflinger:


romainguy (approved):

blueswhen:
hi，can you tell me how to run filament in the ANGLE context? I have another sdk which can only run in ANGLE, so if i want to the sdk communicated with the filament, I also need filament running in ANGLE too.

PR #5453, opened 2022-04-20
gltfio: introduce TextureProvider API.

prideout:
This allows clients to provide their own asynchronous texture decoders
for various mime-typed images. This is a plug-in component for gltfio,
in some ways similar to MaterialProvider.

There are two motivations for this: to decouple gltfio from STB and
to make it easier to integrate support for BasisU textures.

This also has the side effect of simplifying ResourceLoader, since the
texture decoding jobs have been moved out.

As part of this work, I made the "stb" CMake target into a traditional
static library. Previously we had several files called `Image.cpp`
whose sole purpose was to enable STB_IMAGE_IMPLEMENTATION.


prideout:
Gah I just realized I'm using the JobSystem incorrectly, e.g. there is no need to retain the job pointer.  Let me fix that up.

libs/gltfio/src/ResourceLoader.cpp
```diff
@@ -533,347 +517,192 @@ bool ResourceLoader::loadResources(FFilamentAsset* asset, bool async) {
...
-    auto iter = mUriDataCache.find(uri);
-    if (iter != mUriDataCache.end()) {
-        const uint8_t* sourceData = (const uint8_t*) iter->second.buffer;
-        if (!stbi_info_from_memory(sourceData, iter->second.size, &entry->width,
-                &entry->height, &entry->numComponents)) {
-            slog.e << "Unable to decode " << uri << " : " << stbi_failure_reason() << io::endl;
-            mUriTextureCache.erase(uri);
-        }
-        return;
+    TextureProvider* provider = mTextureProviders[mime];
```
bejado, line 593: Maybe convert `mime` to lower-case?

libs/gltfio/src/StbProvider.cpp
```diff
@@ -0,0 +1,236 @@
...
+
+    struct TextureInfo {
+        Texture* texture;
+        TextureState state;
+        std::atomic<intptr_t> decodedTexelsBaseMipmap;
+        std::vector<uint8_t> sourceBuffer;
+        JobSystem::Job*  decoderJob;
+    };
+
+    static const intptr_t DECODING_ERROR = 0x1;
```
bejado, line 74: Maybe a constant for `DECODING_NOTSTARTED = 0x0;` or something. I think that'd help readability.

bejado (approved):
Nice, well-thought out API as always!

libs/gltfio/src/StbProvider.cpp
```diff
@@ -0,0 +1,236 @@
...
+                slog.e << "stb_image failed to decode texture." << io::endl;
+                info->state = TextureState::READY;
+                ++mDecodedCount;
+                continue;
+            }
+            Texture::PixelBufferDescriptor pbd((uint8_t*) data,
+                    texture->getWidth() * texture->getHeight() * 4, Texture::Format::RGBA,
+                    Texture::Type::UBYTE, [](void* mem, size_t, void*) { stbi_image_free(mem); });
+            texture->setImage(*mEngine, 0, std::move(pbd));
+            texture->generateMipmaps(*mEngine);
```
bejado, line 184: Seems `generateMipmaps` gets called unconditionally. Do we want a way for clients to control this?

prideout: Good question!  The tl;dr is that I think this is fine, given that glTF is the only user of this interface.  The TextureProvider docstring says that all miplevels should be ready after the decoding signal, and this is done because glTF uses linear mipmap filtering.

The Stb provider can only handle single-image formats so it calls generateMipmaps unconditionally.  The Ktx2 provider will only do this if the ktx file doesn't have all miplevels.

libs/gltfio/src/StbProvider.cpp
prideout: Good idea, I'll do this.

PR #5457, opened 2022-04-21
Complete the OGL backend's stencil support and add test case

bejado:
A first pass at implementing complete stencil buffer support. We still need to support other backends and a user-facing API.

### DIFF
--- a/filament/backend/src/opengl/OpenGLContext.h
+++ b/filament/backend/src/opengl/OpenGLContext.h
@@ -102,7 +102,7 @@ public:
     inline void depthMask(GLboolean flag) noexcept;
     inline void depthFunc(GLenum func) noexcept;
     inline void stencilFunc(GLenum func, GLint ref, GLuint mask) noexcept;
-    inline void stencilOp(GLenum op) noexcept;
+    inline void stencilOp(GLenum sfail, GLenum dpfail, GLenum dppass) noexcept;
     inline void stencilMask(GLuint mask) noexcept;
     inline void polygonOffset(GLfloat factor, GLfloat units) noexcept;
     inline void beginQuery(GLenum target, GLuint query) noexcept;
@@ -300,7 +300,14 @@ private:
                     return func != rhs.func || ref != rhs.ref || mask != rhs.mask;
                 }
             } stencilFunc;
-            GLenum stencilOp            = GL_KEEP;
+            struct StencilOp {
+                GLenum sfail            = GL_KEEP;
+                GLenum dpfail           = GL_KEEP;
+                GLenum dppass           = GL_KEEP;
+                bool operator != (StencilOp const& rhs) const noexcept {
+                    return sfail != rhs.sfail || dpfail != rhs.dpfail || dppass != rhs.dppass;
+                }
+            } stencilOp;
             GLuint stencilMask          = ~GLuint(0);
         } raster;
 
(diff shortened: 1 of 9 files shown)


filament/backend/src/opengl/OpenGLContext.h
```diff
@@ -634,6 +647,27 @@ void OpenGLContext::depthFunc(GLenum func) noexcept {
...
+    // WARNING: don't call this without updating mRasterState
+    update_state(state.raster.stencilFunc, {func, ref, mask}, [&]() {
+        glStencilFunc(func, ref, mask);
+    });
+}
+
+void OpenGLContext::stencilOp(GLenum op) noexcept {
+    // WARNING: don't call this without updating mRasterState
+    update_state(state.raster.stencilOp, op, [&]() {
+        glStencilOp(GL_KEEP, GL_KEEP, op);
```
pixelflinger, line 660: should we be able to set sfail and dpfail too?

filament/backend/src/opengl/OpenGLDriver.cpp
```diff
@@ -294,6 +294,17 @@ void OpenGLDriver::setRasterStateSlow(RasterState rs) noexcept {
         gl.depthMask(GLboolean(rs.depthWrite));
     }
 
+    // stencil test / operation
+    if (rs.stencilFunc == RasterState::StencilFunction::A && !rs.stencilWrite) {
+        gl.disable(GL_STENCIL_TEST);
+    } else {
+        gl.enable(GL_STENCIL_TEST);
+        gl.stencilFunc(getStencilFunc(rs.stencilFunc), rs.stencilRef, ~GLuint(0));
```
pixelflinger, line 302: why is the mask set to all 1's here?

filament/backend/src/opengl/OpenGLContext.h
bejado: Yeah I was wondering about that too. I think we might as well. I was concerned it would add additional fields to `RasterState` but I suppose if we go with 64 bits now we have plenty.

filament/backend/src/opengl/OpenGLDriver.cpp
bejado: If I understand correctly, this mask is separate from `glStencilMask`. I suppose we could add it to the API as well, but the default is all 1's. According to the spec:

> Specifies a mask that is ANDed with both the reference value and the stored stencil value when the test is done. The initial value is all 1's.

pixelflinger: maybe we just make it part of the API? Maybe it could be used so that we could reserve some bits for us vs. some bits for the users?

bejado: Yeah that sounds good.

bejado: Do you have any thoughts on where to store the mask value? Even with 64 bits, `RasterState` doesn't have enough to store a full 32 bit mask, let alone two masks (a stencil read mask and write mask).

bejado: After discussing offline the resolution is to ignore the mask values for now.

filament/src/RenderPass.h
```diff
@@ -213,30 +213,33 @@ class RenderPass {
...
         backend::Handle<backend::HwRenderPrimitive> primitiveHandle;    // 4 bytes
         backend::Handle<backend::HwBufferObject> morphWeightBuffer;     // 4 bytes
         backend::Handle<backend::HwSamplerGroup> morphTargetBuffer;     // 4 bytes
-        backend::RasterState rasterState;                               // 4 bytes
         uint16_t index = 0;                                             // 2 bytes
         uint16_t instanceCount;                                         // 2 bytes
+        backend::RasterState rasterState;                               // 8 bytes
         Variant materialVariant;                                        // 1 byte
-        uint8_t reserved[11 - sizeof(void*)] = {};                      // 3 bytes (7)
+        uint8_t reserved1[7] = {};                                      // 7 bytes
```
pixelflinger, line 228: So close to fit in 32 bytes :-\

pixelflinger (approved):

filament/src/RenderPass.h
bejado: Yeah :/

Issue #5465, opened 2022-04-25
Compiling for C++/CLI support requires c++17


BStringhamVRSK:
**Describe the bug**
**Note: I will be submitting a code pull request soon to fix this.**
We are using a C++/CLI wrapper for Filament in order to use it with managed C# code. Currently, C++/CLI only supports c++17.  A struct initializer (ALL_SHADER_STAGE_FLAGS in DriverEnums.h line 917) uses "designated initializers," a feature which requires c++20.  A simple modification will eliminate that dependency.

**To Reproduce**
We get this error when the compiler encounters the ALL_SHADER_STAGE_FLAGS initializer, because we compile with c++17:
`include\backend\DriverEnums.h(917,62): error C7555: use of designated initializers requires at least '/std:c++latest'`

**Recommended fix**
I will be submitting a pull request which will change `DriverEnums.h line 917` from:
`static constexpr ShaderStageFlags ALL_SHADER_STAGE_FLAGS = { .vertex = true, .fragment = true };`
to
`static constexpr ShaderStageFlags ALL_SHADER_STAGE_FLAGS = { true, true };`

**Desktop:**
 - OS: Win 10
 - Backend: OpenGL

PR #5466, opened 2022-04-25
Changed ALL_SHADER_STATE_FLAGS initializer to support c++17, which is…

BStringhamVRSK:
… required for C++/CLI support. (#5465)


pixelflinger (approved):

pixelflinger:
I think this is okay, but we're using designated initializer in *a lot* of places and we don't want to change that.

PR #5467, opened 2022-04-26
various cleanups

pixelflinger:
it's easier to review the CL separately.

Most of the cleanup are about removing unused code or data, or simplifying existing code.

### DIFF
--- a/filament/src/details/Camera.cpp
+++ b/filament/src/details/Camera.cpp
@@ -156,8 +156,8 @@ math::mat4 FCamera::getProjectionMatrix() const noexcept {
     // Note that this math ends up setting the projection matrix' p33 to 0, which is where we're
     // getting back a lot of precision in the depth buffer.
     const mat4 m{ mat4::row_major_init{
-            mScaling.x, 0.0, 0.0, mShiftCS.x,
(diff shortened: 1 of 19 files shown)


romainguy (approved):

filament/src/details/Camera.cpp
```diff
@@ -286,14 +286,13 @@ CameraInfo::CameraInfo(FCamera const& camera, const math::mat4& worldOriginCamer
     cullingProjection  = mat4f{ camera.getCullingProjectionMatrix() };
     model              = mat4f{ modelMatrix };
     view               = mat4f{ inverse(modelMatrix) };
+    worldOrigin        = worldOriginCamera;
```
bejado, line 289: Why not add a `userViewMatrix` field here instead of having to "undo" the world origin later in `getUserViewMatrix()`?

bejado (approved):

filament/src/details/Camera.cpp
pixelflinger: a-ha, good question! In fact this change was partly so that I could do that easily in the future! :-)
The downside would be that this structure (which is already big) would increase.

bejado: 👍 got it!

PR #5468, opened 2022-04-26
remove getUV() from post-process materials

pixelflinger:
it made it actually more confusing to have two version of the uv
(normalized and not). Also the unnormalized version was computed from
frameUniforms which is not great in the case of post-processing
(it just happened to work).

### DIFF
--- a/filament/src/materials/colorGrading/colorGrading.mat
+++ b/filament/src/materials/colorGrading/colorGrading.mat
@@ -98,9 +98,7 @@ float starburst(const vec2 uv) {
     return saturate(mask + (1.0 - smoothstep(0.0, 0.3, d)));
 }
 
-vec3 bloom(const vec3 color) {
-    highp vec2 uv = variable_vertex.xy;
-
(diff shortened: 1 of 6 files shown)


filament/src/materials/colorGrading/colorGrading.mat
```diff
@@ -124,27 +122,28 @@ vec3 bloom(const vec3 color) {
     return result;
 }
 
-vec4 resolveFragment(const ivec2 uv) {
+vec4 resolveFragment(const highp vec2 uv) {
 #if POST_PROCESS_OPAQUE
-    return vec4(texelFetch(materialParams_colorBuffer, uv, 0).rgb, 1.0);
+    return vec4(textureLod(materialParams_colorBuffer, uv, 0.0).rgb, 1.0);
```
romainguy, line 127: Why are you going from texelFetch to textureLod here? Isn't texelFetch what we want?

bejado (approved):

PR #5484, opened 2022-04-28
fixed a bug preventing compilation os msvc

AetherstormInteractive:


libs/utils/include/utils/algorithm.h
```diff
@@ -60,7 +60,7 @@ template<typename T, typename = std::enable_if_t<std::is_unsigned<T>::value>>
 constexpr inline T ctz(T x) noexcept {
     static_assert(sizeof(T) * CHAR_BIT <= 64, "details::ctz() only support up to 64 bits");
     T c = sizeof(T) * CHAR_BIT;
-    x &= -x;    // equivalent to x & (~x + 1)
+    x & (~x + 1);    // equivalent to x & (~x + 1)
```
romainguy, line 63: The computed value isn't assigned, a bug is being introduced here

romainguy (changes requested):

romainguy:
Could you explain what "bug" this fixes please?

AetherstormInteractive:
> Could you explain what "bug" this fixes please?

There is a math calculation in algorithm.h under utils that was recently changed, I believe Line 63, and this prevented compilation when hooked into another app, it would continuously give me C4146 until I changed it back to what it said it was equivalent to in its comment.

AetherstormInteractive:
It could be windows or machine specific only, but I was able to reproduce the bug on my laptop as well.

AetherstormInteractive:
I couldn't find where Line 63 was changed in the commits so I had rolled back to a 1.12 and this completely alleviated the issue.

romainguy:
What's the compilation issue? Please show the error message

AetherstormInteractive:
image

AetherstormInteractive:
this occurred in algorithm.h of utils lib

pixelflinger:
applying the unary `operator -` to an unsigned type *is not an error* and is well defined. I'm not sure why msvc complains but it is wrong.

If you need to fix it, please use a solution that is specific to MSVC, for e.g.:
```
#if defined(_MSC_VER)
    x &= (~x + 1u);
#else
    x &= -x;
#endif
```

AetherstormInteractive:
Alright

PR #5486, opened 2022-04-28
iOS: implement headleass swapchain

kmw4988:
This patch enables a headless swapchain on iOS.
So, I added code to make a CAEAGLLayer.


filament/backend/src/opengl/platforms/PlatformCocoaTouchGL.mm
```diff
@@ -108,8 +108,9 @@
 }
 
 Platform::SwapChain* PlatformCocoaTouchGL::createSwapChain(uint32_t width, uint32_t height, uint64_t& flags) noexcept {
-    // TODO: implement headless SwapChain
-    return nullptr;
+    CAEAGLLayer* glLayer = [CAEAGLLayer layer];
+    glLayer.frame = CGRectMake(0, 0, width, height);
+    return (__bridge SwapChain*) glLayer;
```
bejado, line 113: Won't this pointer be invalid when `glLayer` goes out of scope?

kmw4988: I think, there seems to be no problem.

I tested it with the code below :

```cpp
filament::SwapChain* swap_chain_ =
      engine_->createSwapChain(argear_view_info.width, argear_view_info.height,
                               filament::SwapChain::CONFIG_TRANSPARENT);

std::cout<<"create swapchain : "<< swap_chain_ << std::endl;
```

printed : 
```
create swapchain : 0x1055217a0
```

bejado: `[CAEAGLLayer layer]` returns an autoreleased object. Since nothing is strongly referencing `glLayer`, it will get released at some point after this method returns. The `SwapChain*` pointer will then be pointing to dead memory, which could lead to a crash.

We need to somehow hold a reference to the layer for the lifetime of the `SwapChain`.

bejado: Instead of using `__bridge`, use `CFBridgingRetain` to keep the layer alive. Then inside `destroySwapChain`, use `CFBridgingRelease`.

kmw4988: Oh, I see.
For some reason, I thought it was okay because the memory was alive even when connected to the c++ code.
I overlooked the instability.

I'll use `CFBridgingRetain `and `CFBridgingRelease`.

filament/backend/src/opengl/platforms/PlatformCocoaTouchGL.mm
```diff
@@ -108,11 +108,17 @@
...
 Platform::SwapChain* PlatformCocoaTouchGL::createSwapChain(uint32_t width, uint32_t height, uint64_t& flags) noexcept {
-    // TODO: implement headless SwapChain
-    return nullptr;
+    CAEAGLLayer* glLayer = [CAEAGLLayer layer];
+    glLayer.frame = CGRectMake(0, 0, width, height);
+    return (SwapChain*) CFBridgingRetain(glLayer);
 }
 
 void PlatformCocoaTouchGL::destroySwapChain(Platform::SwapChain* swapChain) noexcept {
+    CAEAGLLayer* glLayer = (CAEAGLLayer*) CFBridgingRelease(swapChain);
```
bejado, line 117: The problem with this approach is that we'll always call `release` on `swapChain`, even if it's not a headless one (and we didn't instantiate it ourselves).

In the [desktop implementation](https://github.com/google/filament/blob/dbf51c1c7de27407a25eecd31a78a08ff3341aed/filament/backend/src/opengl/platforms/PlatformCocoaGL.mm#L192-L193), the headless SwapChains get added to a `mHeadlessSwapChains` vector (which maintains their reference count) and then get removed (freeing their reference count) inside `destroySwapChain`.

 Come to think of it, I like this style better than calling `CFBridgingRetain`/`CFBridgingRelease`.

bejado (changes requested):

bejado (approved):
Thank you!

PR #5487, opened 2022-04-28
Ensure we always have 4 "safe" mip levels

pixelflinger:
A "safe" mip-level is one that has exactly half the dimensions of
the current level. In other words, we guarantee that we can halve the
dimensions 4 times without loss (i.e. odd dimension).

This is achieved by effectively padding the viewport (i.e. making the
viewport's dimensions multiple of 16). The post-processing passes
eventually take care of cropping the extra padding so the final
image is unchanged.

This is a more generic solution to dynamic resolution padding, helps
with memory allocations and helps all the algorithms that depend on
mipmaping (e.g. DoF, SSAO, Bloom).

One important improvement brought by this change is that all "final"
post-processing effects are now able to handle a sub-region of the
source, including: color-gradding, fxaa and upscaling.

### DIFF
--- a/filament/src/PostProcessManager.cpp
+++ b/filament/src/PostProcessManager.cpp
@@ -1340,7 +1340,7 @@ FrameGraphId<FrameGraphTexture> PostProcessManager::generateMipmapSSR(
             input = ppm.resolveBaseLevel(fg, "ssr", input);
             // then blit into an appropriate texture
             // this handles scaling, format conversion and mipmaping
-            input = ppm.opaqueBlit(fg, input, outDesc);
+            input = ppm.opaqueBlit(fg, input, { 0, 0, desc.width, desc.height }, outDesc);
         }
     }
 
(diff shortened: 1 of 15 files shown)


filament/src/PostProcessManager.cpp
```diff
@@ -2399,6 +2402,12 @@ FrameGraphId<FrameGraphTexture> PostProcessManager::colorGrading(FrameGraph& fg,
                 mi->setParameter("vignetteColor", vignetteOptions.color);
                 mi->setParameter("fxaa", colorGradingConfig.fxaa);
                 mi->setParameter("temporalNoise", temporalNoise);
+                mi->setParameter("viewport", float4{
```
bejado, line 2405: Maybe call this `normalizedViewport` (and other places)?

filament/src/details/Renderer.cpp
```diff
@@ -260,31 +260,108 @@ void FRenderer::renderJob(ArenaScope& arena, FView& view) {
...
+        // TODO: use this feature to implement guard-bands
+        //
+        // TODO: Should we enable when we don't have post-processing?
+        //       Without post-processing, we usually draw directly into
+        //       the SwapChain, and we might want to keep it this way.
+
+        // compute the new rendering width and height, multiple of 16.
+        constexpr uint32_t rounding = 16u;
+        float width  = float( (svp.width  + (rounding - 1u)) & ~(rounding - 1u) );
+        float height = float( (svp.height + (rounding - 1u)) & ~(rounding - 1u) );
```
bejado, line 335: Make these `const`.

filament/src/details/Renderer.cpp
```diff
@@ -436,7 +497,29 @@ void FRenderer::renderJob(ArenaScope& arena, FView& view) {
...
                 uniforms.prepareCamera(cameraInfo);
-                uniforms.prepareViewport(svp, 0, 0);
+
+                // The code here is a little fragile. In theory, we need to call prepareViewport()
+                // for each render pass, because the viewport parameters depend on the resolution.
+                // However, in practice, we only have two resolutions: the color pass resolution,
+                // and the structure pass which is governed by aoOptions.resolution (this could
+                // change in the future).
+                // So here we set the parameters for the structure pass and SSAO passes which
+                // are always done first. The SSR pass will aslo use these parameters which
```
bejado, line 507:
```suggestion
                // are always done first. The SSR pass will also use these parameters which
```

filament/src/details/Renderer.cpp
```diff
@@ -436,7 +497,29 @@ void FRenderer::renderJob(ArenaScope& arena, FView& view) {
...
+                // is wrong if it doesn't run at the same resolution as SSAO.
+                // preapreViewport() is called again during the color pass, which resets the
+                // values correctly for the colorpass, however, this will be again wrong
+                // for passes that comme after the colorpass, such as DoF.
```
bejado, line 511:
```suggestion
                // for passes that come after the colorpass, such as DoF.
```

bejado (approved):

PR #5489, opened 2022-04-28
WebGL: fix isTextureFormatSupported for ETC2 formats.

prideout:
It was assumed that ETC2 is supported on all ES3 devices, but WebGL on
desktop machines does not support ETC2.

Note that WebGL has its own special extension strings for compressed
textures.


bejado (approved):

pixelflinger (approved):
should we use the webgl extension string instead, so we can use it when it is available?

prideout:
> should we use the webgl extension string instead, so we can use it when it is available?

The webgl extension string is already used in an "or" condition.

PR #5490, opened 2022-04-28
Add async interface to Ktx2Reader.

prideout:


libs/ktxreader/src/Ktx2Reader.cpp
```diff
@@ -137,14 +176,166 @@ void Ktx2Reader::unrequestFormat(Texture::InternalFormat format) {
 }
 
 Texture* Ktx2Reader::load(const uint8_t* data, size_t size, TransferFunction transfer) {
```
bejado, line 178: This method seems pretty similar to `doTranscoding`, is there a way for them to share their implementation? Maybe a helper method that returns an `std::optional<Texture::PixelBufferDescriptor>`.

libs/ktxreader/include/ktxreader/Ktx2Reader.h
```diff
@@ -85,16 +93,107 @@ class Ktx2Reader {
...
+
+            /**
+             * Uploads all mipmaps to the texture.
+             *
+             * This calls Texture::setImage() and potentially generates missing mipmaps.
+             * If transcoding has not yet completed, returns INCOMPLETE.
+             *
+             * NOTE: This can be called only from the foreground thread.
+             */
+            void uploadImages();
```
pixelflinger, line 140: What happens if it's not called from the right thread? 

It says it returns `INCOMPLETE` in some cases, but it doesn't have a return value.

libs/ktxreader/include/ktxreader/Ktx2Reader.h
```diff
@@ -85,16 +93,107 @@ class Ktx2Reader {
...
+         * platform's capabilities and its availability from the transcoder. When a suitable format
+         * is determined, it then performs lossless decompression (zstd) before transcoding the data
+         * into the final format.
+         *
+         * The transfer function specified here is used in two ways:
+         *   1) It is checked against the transfer function that was specified as metadata
+         *      in the KTX2 blob. If they do not match, this method fails.
+         *   2) It is used as a filter when determining the final internal format.
+         */
+        Async* asyncCreate(const uint8_t* data, size_t size, TransferFunction transfer);
```
pixelflinger, line 175: shouldn't `data` be a `void*`?

libs/ktxreader/include/ktxreader/Ktx2Reader.h
```diff
@@ -85,16 +93,107 @@ class Ktx2Reader {
...
+
+        /**
+         * Frees the given async object if it is non-null.
+         *
+         * This frees the original source data (i.e. the raw content of the KTX2 file) but does not
+         * free the associated Texture object. This can be after transcoding has finished.
+         */
+        void asyncDestroy(Async* async);
```
pixelflinger, line 183: I like to pass a `Async**` in the destroy functions, so that the passed pointer can be set to nullptr. This makes it impossible (or hard) to use after calling destroy:
e.g.:
```
  asyncDestroy(&mAsync);
```

libs/ktxreader/include/ktxreader/Ktx2Reader.h
```diff
@@ -85,16 +93,107 @@ class Ktx2Reader {
...
+
     private:
         Ktx2Reader(const Ktx2Reader&) = delete;
         Ktx2Reader& operator=(const Ktx2Reader&) = delete;
         Ktx2Reader(Ktx2Reader&& that) noexcept = delete;
         Ktx2Reader& operator=(Ktx2Reader&& that) noexcept = delete;
 
+        Texture* createTexture(basist::ktx2_transcoder* transcoder, const uint8_t* data,
```
pixelflinger, line 191: same here I think `void*` is more appropriate.

libs/ktxreader/include/ktxreader/Ktx2Reader.h
```diff
@@ -85,16 +93,107 @@ class Ktx2Reader {
...
+                size_t size, TransferFunction transfer);
+
         Engine& mEngine;
         bool mQuiet;
```
pixelflinger, line 195: move the `bool` to the end of the class.

libs/ktxreader/src/Ktx1Reader.cpp
```diff
@@ -83,9 +83,11 @@ Texture* createTexture(Engine* engine, const Ktx1Bundle& ktx, bool srgb,
 
     auto texformat = toTextureFormat(ktxinfo);
     if (srgb && !isSrgbTextureFormat(texformat)) {
-        utils::slog.w << "Requested sRGB format but KTX contains a linear format. ";
+        utils::slog.w << "Requested sRGB format but KTX contains a linear format. "
+            << utils::io::endl;
     } else if (!srgb && isSrgbTextureFormat(texformat)) {
-        utils::slog.w << "Requested linear format but KTX contains a sRGB format. ";
+        utils::slog.w << "Requested linear format but KTX contains a sRGB format. "
+            << utils::io::endl;
```
pixelflinger, line 90: I feel like logs like this should be on debug builds only.

Either that or this condition is an error, and preconditions are violated. I guess... is this a misuse of the API? If so, I think it should be documented and we should use ASSERT_PRECONDITION and method should not be `noexcept`.

If we want to allow it, but warn that it's wrong, I think DEBUG build only is more appropriate.

libs/ktxreader/include/ktxreader/Ktx2Reader.h
```diff
@@ -85,16 +93,107 @@ class Ktx2Reader {
...
+        class Async {
+        public:
+            /**
+             * Retrieves the Texture object.
+             *
+             * The texture is available immediately, but does not have its miplevels ready until
+             * after uploadImages(). The caller has ownership over this texture and is responsible
+             * for freeing it.
+             */
+            Texture* getTexture() const noexcept;
```
pixelflinger, line 123: Is it safe then destroy the texture immediately? 

What happens if I delete the texture and then call `uploadImages()`?

maybe `uploadImages()` should take that `Texture*` as a parameter.

pixelflinger (changes requested):

libs/ktxreader/include/ktxreader/Ktx2Reader.h
prideout: Will do.  I noticed that most of our `Engine::destroy` methods don't do this, but I'm guessing that's historical.

prideout:
Thanks guys for the excellent feedback!  I clarified some comments, and incorporated all the ideas except adding a `Texture*` argument to `uploadImages` -- I understand that it would make it more obvious that the texture should still be alive, but it also opens the door to passing in an arbitrary texture.  I think the pros and cons cancel out.

By the way, the primary consumer of this API is gltfio, I don't expect many external clients to use it.

libs/ktxreader/include/ktxreader/Ktx2Reader.h
```diff
@@ -83,19 +91,111 @@ class Ktx2Reader {
...
+         *      in the KTX2 blob. If they do not match, this method fails.
+         *   2) It is used as a filter when determining the final internal format.
+         */
+        Async* asyncCreate(const void* data, size_t size, TransferFunction transfer);
+
+        /**
+         * Frees the given async object and sets it to null.
+         *
+         * This frees the original source data (i.e. the raw content of the KTX2 file) but does not
+         * free the associated Texture object. This can be after transcoding has finished.
```
pixelflinger, line 182:
```suggestion
         * free the associated Texture object. This can be done after transcoding has finished.
```

libs/ktxreader/include/ktxreader/Ktx2Reader.h
```diff
@@ -83,19 +91,111 @@ class Ktx2Reader {
...
         Ktx2Reader& operator=(Ktx2Reader&& that) noexcept = delete;
 
-        Engine& mEngine;
-        bool mQuiet;
-        basist::ktx2_transcoder* const mTranscoder;
+        Texture* createTexture(basist::ktx2_transcoder* transcoder, const void* data,
+                size_t size, TransferFunction transfer);
 
+        Engine& mEngine;
+        std::unique_ptr<basist::ktx2_transcoder> const mTranscoder;
```
pixelflinger, line 196: since this is a public header, are we okay with exposing `std`?

libs/ktxreader/src/Ktx2Reader.cpp
```diff
@@ -136,15 +223,99 @@ void Ktx2Reader::unrequestFormat(Texture::InternalFormat format) {
...
+    ktx2_transcoder_state basisThreadState;
+    basisThreadState.clear();
+
+    for (uint32_t levelIndex = 0; levelIndex < mTranscoder->get_levels(); levelIndex++) {
+        Texture::PixelBufferDescriptor* pbd;
+        Result result = transcodeImageLevel(*mTranscoder, basisThreadState, texture->getFormat(),
+                levelIndex, &pbd);
+        if (result != Result::SUCCESS) {
+            mEngine.destroy(texture);
+            utils::slog.e << "Failed to transcode level " << levelIndex << utils::io::endl;
```
pixelflinger, line 249: guard with `mQuiet`

libs/ktxreader/src/Ktx2Reader.cpp
```diff
@@ -136,15 +223,99 @@ void Ktx2Reader::unrequestFormat(Texture::InternalFormat format) {
...
+    }
+
+    ktx2_transcoder_state basisThreadState;
+    basisThreadState.clear();
+
+    for (uint32_t levelIndex = 0; levelIndex < mTranscoder->get_levels(); levelIndex++) {
+        Texture::PixelBufferDescriptor* pbd;
+        Result result = transcodeImageLevel(*mTranscoder, basisThreadState, texture->getFormat(),
+                levelIndex, &pbd);
+        if (result != Result::SUCCESS) {
```
pixelflinger, line 247: `UTILS_UNLIKELY`

libs/ktxreader/src/Ktx2Reader.cpp
```diff
@@ -136,15 +223,99 @@ void Ktx2Reader::unrequestFormat(Texture::InternalFormat format) {
...
+        }
+        texture->setImage(mEngine, levelIndex, std::move(*pbd));
+    }
+    return texture;
+}
+
+Result FAsync::doTranscoding() {
+    ktx2_transcoder_state basisThreadState;
+    basisThreadState.clear();
+    for (uint32_t levelIndex = 0; levelIndex < mTranscoder->get_levels(); levelIndex++) {
```
pixelflinger, line 260:
```suggestion
    for (uint32_t levelIndex = 0, c = mTranscoder->get_levels(); levelIndex < c; levelIndex++) {
```

libs/ktxreader/src/Ktx2Reader.cpp
```diff
@@ -136,15 +223,99 @@ void Ktx2Reader::unrequestFormat(Texture::InternalFormat format) {
...
+            return result;
+        }
+        mTranscoderResults[levelIndex].store(pbd);
+    }
+    return Result::SUCCESS;
+}
+
+void FAsync::uploadImages() {
+    size_t levelIndex = 0;
+    for (TranscoderResult& level : mTranscoderResults) {
```
pixelflinger, line 274: I suggest `UTILS_NOUNROLL` here because the compiler usually compeltely unrolls loops when the count is known at compile time.

libs/ktxreader/include/ktxreader/Ktx2Reader.h
```diff
@@ -83,19 +91,111 @@ class Ktx2Reader {
...
+            Result doTranscoding();
+
+            /**
+             * Uploads pending mipmaps to the texture.
+             *
+             * This can safely be called while doTranscoding() is still working in another thread.
+             * Since this calls Texture::setImage(), it should be called from the foreground thread;
+             * see "Thread safety" in the documentation for filament::Engine.
+             */
+            void uploadImages();
```
pixelflinger, line 141: now I'm wondering if `uploadImages` should return whether there is more?

prideout: The caller already knows, because they know if `doTranscoding()` has completed or not.  Note that the Ktx2Reader implementation does not create threads, it's the caller's responsibility to do so if they wish.  I think this is a reasonable separation of concerns.  (The caller in this case will be the gltfio TextureProvider that I'm adding in my next PR.)

pixelflinger (approved):

PR #5491, opened 2022-04-28
improvements to shadow UBO handling

pixelflinger:


filament/src/ShadowMap.cpp
```diff
@@ -60,6 +60,9 @@ ShadowMap::ShadowMap(FEngine& engine) noexcept :
     }
 }
 
+void ShadowMap::terminate(FEngine& engine) {
```
bejado, line 63: Is this necessary?

filament/src/PerViewUniforms.h
```diff
@@ -125,6 +133,8 @@ class PerViewUniforms {
     backend::Handle<backend::HwBufferObject> mUniformBufferHandle;
     backend::Handle<backend::HwSamplerGroup> mSamplerGroupHandle;
     std::uniform_real_distribution<float> mUniformDistribution{ 0.0f, 1.0f };
+    void prepareShadowSampling(PerViewUib& uniforms,
```
bejado, line 136: Make `static`?

bejado (approved):

filament/src/ShadowMap.cpp
pixelflinger: not in this PR, I added it for consistency with other classes. It would be useful if for e.g. ShadowMap had its own UBO.

filament/src/PerViewUniforms.h
pixelflinger: yes good point.

PR #5504, opened 2022-05-02
Metal: add RelaxedtoHalfPass for improved ALU performance

### LINKED ISSUE #5449: Use half in metal shader for texture read/write
Xcode suggest that we should use `half` instead of `float` to reduce ALU cost in some cases such as customize post processing passes or unlit materials. 
I have tried `precision` of `sampler2d ` but doesn't work. The translated MSL is still `texture2d<float>`.

Maybe we should proper handle the `precision` of `samplerX` or  adding a new property in `samplerX` to indicate we want to use `half` and `float` to sample our texture image data.

No.

**OS and backend**
Is this feature OS/backend specific? If so, explain why.
Currently, Metal only.

bejado:
Fixes #5449

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -5,6 +5,8 @@ A new header is inserted each time a *tag* is created.
 
 ## v1.22.1 (currently main branch)
 
+- Metal: Shaders now use `half` floating-point arithmetic when possible for improved performance. [⚠️ **Recompile Materials**]
+
 ## v1.22.0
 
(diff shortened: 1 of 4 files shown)


pixelflinger (approved):

Michael-Lfx:
Thank you! How can we use it in `MaterialVertexInputs` and `MaterialInputs` to specify `half`?

romainguy:
We use half by default on mobile.

PR #5507, opened 2022-05-03
Cleanup Renderer.cpp

pixelflinger:
The main changes are:

- ColorPass() and RefractionPass() are now static and moved into a 
  new RendererUtils file. These methods don't need FRenderer and are 
  more like a big "script".

- Cleaned-up includes

- Reformatted/reordered methods


prideout (approved):
LGTM.  I guess this could be a namespace instead of a static class.  I don't have a strong opinion on which is better.

PR #5511, opened 2022-05-04
Various glTF-related cleanup and enhancements.

prideout:
- For completion, FilamentAsset now has getRenderableEntities(). This is
  similar to sister methods getLightEntities() and getCameraEntities()
  except there is no need to store a separate array.

- The web helmet demo does not need to enable shadows, they are already
  enabled.

- The ViewerGui populateAsset() method was doing two things that are
  now decoupled for clarity: setAsset() and populateAsset().

- The updateRootTransform() method is now called only when the autoscale
  checkbox is toggled, instead of every frame.

- The getFooEntities() methods in Java now skip doing work for empty
  lists. Actually these should not return arrays at all, but let's fix
  that later, since it will break backwards compatibility.


romainguy (approved):
About returning arrays, we can keep these functions, and add `Foo[] getFoo(Foo[])` to avoid allocations.

PR #5518, opened 2022-05-05
fix a crash in assetloader

pixelflinger:
use std::partition instead of std::sort to partition the entities 
such as the renderables are first.


prideout (approved):
Excellent, this totally makes sense.

PR #5521, opened 2022-05-06
Implement an optional guard band for screen-space effects

pixelflinger:
When the guard band is enabled, the effective rendering area is
increased by a certain amount (currently 16 pixels on each side) so
that screen-space effects (e.g. SSAO, SSR, DoF) behave better around
the edges of the screen. Of course, this comes at a performance and
memory cost.

Currently the guard band is not configurable other than being optional.

Added c++, java and js bindings.

### DIFF
--- a/libs/viewer/src/AutomationSpec.cpp
+++ b/libs/viewer/src/AutomationSpec.cpp
@@ -35,10 +35,8 @@ using std::vector;
 
 static const bool VERBOSE = false;
 
-namespace filament {
-namespace viewer {
+namespace filament::viewer {
 
-// The default spec generates 66 test cases.
 static const char* DEFAULT_AUTOMATION = R"TXT([
     {
         "name": "ppoff",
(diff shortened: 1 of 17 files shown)


libs/viewer/src/AutomationSpec.cpp
```diff
@@ -64,6 +64,7 @@ static const char* DEFAULT_AUTOMATION = R"TXT([
             "view.screenSpaceReflections.enabled": [false, true]
             "view.bloom.enabled": [false, true],
             "view.dof.enabled": [false, true]
+            "view.guardBand.enabled": [false, true]
```
prideout, line 67: We should delete the comment at the top about "66 test cases".

libs/viewer/src/Settings.cpp
```diff
@@ -1551,6 +1575,12 @@ static std::ostream& operator<<(std::ostream& out, const VsmShadowOptions& in) {
         << "}";
 }
 
+static std::ostream& operator<<(std::ostream& out, const GuardBandOptions& in) {
+    return out << "{\n"
+        << "\"enabled\": " << in.enabled << "\n"
```
prideout, line 1580: Needs `to_string` because it's a bool

libs/viewer/src/AutomationSpec.cpp
prideout: Trailing comma is missing.

prideout (approved):

PR #5522, opened 2022-05-06
Introduce gltfio::NodeManager, remove some robin_map.

prideout:


libs/gltfio/src/AssetLoader.cpp
```diff
@@ -147,10 +157,11 @@ struct FAssetLoader : public AssetLoader {
 
     EntityManager& mEntityManager;
     RenderableManager& mRenderableManager;
-    NameComponentManager* mNameManager;
+    NameComponentManager* const mNameManager;
```
pixelflinger, line 160: Why a pointer and not a `&` just like the ones above?

libs/gltfio/src/AssetLoader.cpp
```diff
@@ -147,10 +157,11 @@ struct FAssetLoader : public AssetLoader {
...
     TransformManager& mTransformManager;
-    MaterialProvider* mMaterials;
-    Engine* mEngine;
+    MaterialProvider* const mMaterials;
+    Engine* const mEngine;
```
pixelflinger, line 163: These should be references too I think.

libs/gltfio/src/AssetLoader.cpp
```diff
@@ -147,10 +157,11 @@ struct FAssetLoader : public AssetLoader {
...
+    FNodeManager* const mNodeManager;
```
pixelflinger, line 164: Looks like this could just be a member.

libs/gltfio/src/AssetLoader.cpp
```diff
@@ -270,7 +281,7 @@ void FAssetLoader::createAsset(const cgltf_data* srcAsset, size_t numInstances)
     }
     #endif
 
-    mResult = new FFilamentAsset(mEngine, mNameManager, &mEntityManager, srcAsset);
+    mResult = new FFilamentAsset(mEngine, mNameManager, &mEntityManager, mNodeManager, srcAsset);
```
pixelflinger, line 284: I really don't like `mResult` as a name, especially as a member.

libs/gltfio/src/AssetLoader.cpp
```diff
@@ -402,8 +414,8 @@ void FAssetLoader::createEntity(const cgltf_data* srcAsset, const cgltf_node* no
     const cgltf_size extras_size = node->extras.end_offset - node->extras.start_offset;
     if (extras_size > 0) {
         const cgltf_data* srcAsset = mResult->mSourceAsset->hierarchy;
-        mResult->mNodeExtras[entity] = CString(
-                srcAsset->json + node->extras.start_offset, extras_size);
+        CString extras(srcAsset->json + node->extras.start_offset, extras_size);
+        mNodeManager->setExtras(mNodeManager->getInstance(entity), std::move(extras));
```
pixelflinger, line 418: I wouldn't even use a temporary `extras` here:
```
mNodeManager->setExtras(mNodeManager->getInstance(entity), {
    srcAsset->json + node->extras.start_offset, extras_size });
```

libs/gltfio/src/AssetLoader.cpp
```diff
@@ -525,12 +537,12 @@ void FAssetLoader::createRenderable(const cgltf_data* srcAsset, const cgltf_node
         }
     }
 
-    auto& morphTargetNames = mResult->mMorphTargetNames[entity];
-    assert_invariant(morphTargetNames.empty());
-    morphTargetNames = FixedCapacityVector<CString>(numMorphTargets);
+    FixedCapacityVector<CString> morphTargetNames(numMorphTargets);
     for (cgltf_size i = 0, c = mesh->target_names_count; i < c; ++i) {
-        morphTargetNames[i] = StaticString::make(mesh->target_names[i]);
+        morphTargetNames[i] = CString(mesh->target_names[i]);
```
pixelflinger, line 542: we need to use a `CString` here because the `target_names[i]` is not persistant?

libs/gltfio/src/FFilamentAsset.h
```diff
@@ -255,9 +258,10 @@ struct FFilamentAsset : public FilamentAsset {
...
-    filament::Engine* mEngine;
-    utils::NameComponentManager* mNameManager;
-    utils::EntityManager* mEntityManager;
+    filament::Engine* const mEngine;
+    utils::NameComponentManager* const mNameManager;
+    utils::EntityManager* const mEntityManager;
+    NodeManager* const mNodeManager;
     std::vector<utils::Entity> mEntities; // sorted such that renderables come first
     std::vector<utils::Entity> mLightEntities;
     std::vector<utils::Entity> mCameraEntities;
```
pixelflinger, line 267: could these be FixedCapacityVector?

libs/gltfio/src/FFilamentAsset.h
```diff
@@ -255,9 +258,10 @@ struct FFilamentAsset : public FilamentAsset {
 
     void createAnimators();
 
-    filament::Engine* mEngine;
-    utils::NameComponentManager* mNameManager;
-    utils::EntityManager* mEntityManager;
+    filament::Engine* const mEngine;
+    utils::NameComponentManager* const mNameManager;
+    utils::EntityManager* const mEntityManager;
+    NodeManager* const mNodeManager;
```
pixelflinger, line 264: These could be references.

libs/gltfio/src/FNodeManager.h
```diff
@@ -0,0 +1,152 @@
...
+    #ifndef NDEBUG
+            utils::slog.d << "cleaning up " << manager.getComponentCount()
+                << " leaked node components" << utils::io::endl;
+    #endif
+            while (!manager.empty()) {
+                Instance ci = manager.end() - 1;
+                manager.removeComponent(manager.getEntity(ci));
+            }
+        }
+    }
```
pixelflinger, line 54: don't put the implementation in the header file.

pixelflinger (changes requested):

libs/gltfio/include/gltfio/NodeManager.h
```diff
@@ -0,0 +1,107 @@
...
+     * @param e An entity.
+     *
+     * @see create()
+     */
+    void destroy(Entity e) noexcept;
+
+    void setMorphTargetNames(Instance ci, utils::FixedCapacityVector<CString>&& names) noexcept;
+    const utils::FixedCapacityVector<CString>& getMorphTargetNames(Instance ci) const noexcept;
+
+    void setExtras(Instance ci, CString&& extras) noexcept;
```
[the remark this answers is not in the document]
prideout: Pass-by-value would create a deep copy of the CString, would it not?  The only setter caller is the loader, which definitely does not need to retain a copy, so I wanted to force a movement to occur.  (This API is public just in case people wanted to use the ECS getters instead of the FilamentAsset bridge.)

`extras` is a glTF term, it's their version of userdata, it's just a string metadata.  We ignored it for a long time until a 3P contribution added support.

libs/gltfio/src/AssetLoader.cpp
prideout: Correct. This codeline was an error introduced by a 3P contribution.

libs/gltfio/include/gltfio/NodeManager.h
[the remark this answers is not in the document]
pixelflinger: it would do deep copy only if the caller doesn't use `std::move()` -- but what if the caller wanted to keep its copy of extras?

I'm okay leaving it this way, but it's slightly strange for a public API.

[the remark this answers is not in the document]
prideout: Ah, good point, since I'm in control of the loader, I can be smart there while giving users a choice.  I agree with you now.   In fact I think the vector should be pass-by-value too.

libs/gltfio/include/gltfio/NodeManager.h
```diff
@@ -0,0 +1,107 @@
...
+
+    /**
+     * Destroys this component from the given entity.
+     * @param e An entity.
+     *
+     * @see create()
+     */
+    void destroy(Entity e) noexcept;
+
+    void setMorphTargetNames(Instance ci, utils::FixedCapacityVector<CString>&& names) noexcept;
```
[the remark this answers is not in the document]
prideout: Changed to pass-by-value.

libs/gltfio/include/gltfio/NodeManager.h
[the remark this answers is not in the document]
prideout: Fixed.

libs/gltfio/src/AssetLoader.cpp
```diff
@@ -95,8 +96,13 @@ struct FAssetLoader : public AssetLoader {
             mTransformManager(config.engine->getTransformManager()),
             mMaterials(config.materials),
             mEngine(config.engine),
+            mNodeManager(new FNodeManager()),
```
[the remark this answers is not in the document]
prideout: Fixed.

libs/gltfio/src/AssetLoader.cpp
prideout: Fixed.

libs/gltfio/src/FFilamentAsset.h
prideout: I tried to do this in the past, I might try again on a rainy day but I would need to do the load in the two passes.  I do not know the counts apriori because the entities list is built recursively only from active root nodes.

libs/gltfio/src/FFilamentAsset.h
prideout: Done.  (except for name manager because it's optional)

libs/gltfio/src/FNodeManager.h
prideout: Done.

libs/gltfio/src/AssetLoader.cpp
prideout: Fixed.

pixelflinger (approved):

PR #5532, opened 2022-05-09
header files added, for compatibility of clang c++17. Don't forget to…

jiapei100:
… add -std=c++17.

### DIFF
--- a/filament/backend/include/private/backend/CommandStream.h
+++ b/filament/backend/include/private/backend/CommandStream.h
@@ -38,6 +38,7 @@
 #include <tuple>
 #include <thread>
 #include <utility>
+#include <cstddef>
 
 #include <assert.h>
 #include <stddef.h>
--- a/filament/backend/include/private/backend/CommandStream.h
+++ b/filament/backend/include/private/backend/CommandStream.h
@@ -34,11 +34,11 @@
 #include <utils/compiler.h>
 #include <utils/ThreadUtils.h>
 
+#include <cstddef>
 #include <functional>
 #include <tuple>
 #include <thread>
 #include <utility>
-#include <cstddef>
 
 #include <assert.h>
 #include <stddef.h>
--- a/libs/image/src/ImageSampler.cpp
+++ b/libs/image/src/ImageSampler.cpp
@@ -38,7 +38,7 @@ struct FilterFunction {
     bool rejectExternalSamples = true;
 };
 
-constexpr float M_PIf = float(filament::math::F_PI);
+constexpr float MM_PIf = float(filament::math::F_PI);
 
 const FilterFunction Box {
     .fn = [](float t) { return t <= 0.5f ? 1.0f : 0.0f; },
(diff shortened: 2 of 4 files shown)


filament/backend/include/private/backend/CommandStream.h
```diff
@@ -38,6 +38,7 @@
 #include <tuple>
 #include <thread>
 #include <utility>
+#include <cstddef>
```
romainguy, line 41: Please keep includes information alphabetical order

libs/image/src/ImageSampler.cpp
```diff
@@ -38,7 +38,7 @@ struct FilterFunction {
     bool rejectExternalSamples = true;
 };
 
-constexpr float M_PIf = float(filament::math::F_PI);
+constexpr float MM_PIf = float(filament::math::F_PI);
```
romainguy, line 41: Why MM?

jiapei100: Sorry, but, even if I configured **clang** as my compiler, it seems, it'll search for the particular file `/usr/include/math.h` by default, which is actually from **GNU** , in which **M_Pif** is already defined.

I've got **NO** idea why, but this is one of the conflicts I met...

```C++
/* GNU extension to provide float constants with similar names.  */
#ifdef __USE_GNU
# define M_Ef		2.7182818284590452354f	/* e */
# define M_LOG2Ef	1.4426950408889634074f	/* log_2 e */
# define M_LOG10Ef	0.43429448190325182765f	/* log_10 e */
# define M_LN2f		0.69314718055994530942f	/* log_e 2 */
# define M_LN10f	2.30258509299404568402f	/* log_e 10 */
# define M_PIf		3.14159265358979323846f	/* pi */
# define M_PI_2f	1.57079632679489661923f	/* pi/2 */
# define M_PI_4f	0.78539816339744830962f	/* pi/4 */
# define M_1_PIf	0.31830988618379067154f	/* 1/pi */
# define M_2_PIf	0.63661977236758134308f	/* 2/pi */
# define M_2_SQRTPIf	1.12837916709551257390f	/* 2/sqrt(pi) */
# define M_SQRT2f	1.41421356237309504880f	/* sqrt(2) */
# define M_SQRT1_2f	0.70710678118654752440f	/* 1/sqrt(2) */
#endif
```

filament/backend/include/private/backend/CommandStream.h
pixelflinger: please use `<stddef.h>`. We don't use the c++ version of the C headers when possible.

filament/backend/include/private/backend/CommandStream.h
```diff
@@ -34,6 +34,7 @@
 #include <utils/compiler.h>
 #include <utils/ThreadUtils.h>
 
+#include <cstddef>
```
pixelflinger, line 37: use `<stddef.h>`

pixelflinger (changes requested):

filament/backend/include/private/backend/CommandStream.h
jiapei100: No... It's because of 
`error: no member named 'max_align_t' in namespace 'std'`

- it seems **max_align_t** is **ONLY** in <cstddef>,
- and <stddef.h> is **already** included in the next couple lf lines in the same file...

pixelflinger: okay, sounds good then. Sorry about the confusion.

pixelflinger (approved):

PR #5534, opened 2022-05-09
Add code generator for Settings.

prideout:
This adds a C++ code generator, implemented in Go. For now this only
generates the JSON reader / writer code.

This PR also prepares `Options.h` (the ground truth) by simplifying its
syntax just a bit. The ground truth file must have very simple C++
syntax, which is described in the README:

https://github.com/google/filament/blob/pr/codegen2/tools/codegen-options/README.md

This process revealed a small bug: `Filter.MEDIAN` was bound to the
incorrect value in Java.

Still TBD is the codegen for Java and JavaScript bindings.


tools/codegen-options/main.go
```diff
@@ -0,0 +1,64 @@
+/*
```
bejado, line 1: Tabs 😱  is that a Go requirement?

bejado (commented):
Our first Go contribution!

tools/codegen-options/main.go
prideout: LOL yup, the language comes with an opinionated format via the gofmt tool, which uses tabs for indention and spaces for various alignment things, e.g. making struct fields line up.

bejado (approved):
Unfamiliar with Go, but LGTM!

PR #5535, opened 2022-05-10
Fix texture coordinate calculations in post-process materials

pixelflinger:
The most important change here is that we no longer apply the
backend (e.g. Metal/Vulkan) UV transformation before calling the
user's vertex shader. This is an API change for post-process materials
(but they're not public).
Now the user is responsible for using uvToRenderTargetUV() in their
shaders (either in the fragment or vertex as appropriate).

This makes it easier to handle offsets/transforms with all APIs.

Conceptually, the only thing that is needed is to call 
uvToRenderTargetUV() just before making a texture call.

This fixes a couple of issues:
- some image shifting in metal/vk
- flare in metal/vk was upside down

We also simplify the DoF code quite a bit now that we can rely on the
rendertargets begin multiple of 16.

We also "fix" SSAO that was working by accident on  metal/vk. The UV
correction was applied 3 times 2 of which were canceling each other.

### DIFF
--- a/filament/src/materials/dof/dofDownsample.mat
+++ b/filament/src/materials/dof/dofDownsample.mat
@@ -20,8 +20,8 @@ material {
             name : cocClamp
         },
         {
-            type : float4,
-            name : uvscale,
+            type : float2,
+            name : texelSize,
             precision: high
         }
     ],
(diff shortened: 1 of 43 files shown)


filament/src/materials/dof/dofDownsample.mat
```diff
@@ -50,7 +50,21 @@ vertex {
...
+        // It works to fix-up the UV *before* applying the offsets because materialParams.texelSize
+        // corresponds to one texel.
+        // on GL:           v0 = v - 0.5 * texel                                                (1)
+        //                  v1 = v - 0.5 * texel + 1 * texel
+        //                     = v + 0.5 * texel                                                (2)
+        // on MTL/VK:       v0 = 1 - ((1 - v) - 0.5 * texel)
+        //                     = v + 0.5 * texel                                                (3)
+        //                  v1 = 1 - ((1 - v) - 0.5 * texel + 1 * texel)
+        //                     = v - 0.5 * texel                                                (4)
+        // Because the all the operations on the 4 texels in the fragment shader don't depend on
```
bejado, line 62: Can you elaborate on this? It seems that `outColor` depends on `w` which depends on `hdr`, which depends on the ordering of the texels in `s00`, `s01`, etc.

bejado (approved):
LGTM just one question

filament/src/materials/dof/dofDownsample.mat
pixelflinger: Thank you for checking this -- I checked again and I think it is good.

So I think it works because the ordering of `s00`, etc... matches the ordering of `d[]` so, in turn, `c` follows the ordering of `d[]` and then `w` does too.

I don't think `hdr` "depends on hdr, which depends on the ordering of the texels in s00, s01, etc.". i.e. `hdr[0]` always depend on `s01`.


Another way to think of it is that we're "replacing" 4 texels by a single one. And those 4 texels are all interchangeable, meaning they don't depend on their neighbor. The downsampling uses the `CoC`, but it too is fetched in the same order (`d[]`).

What do you think?

bejado: Okay yeah, that makes sense. I think I interpreted your comment meaning `s00` might refer to a different texel depending on which backend was used. But after examining it again, I don't think that's the case, correct?

pixelflinger: No that's the case! Depending on the backend `s01` and `s00` are swapped and `s11` and `s10` are swapped. But I claim that it doesn't change the final result because `d[]` is swapped in the same way. And `hdr` only works on a single value (it does the max(r,g,b)).

bejado: 👍

PR #5537, opened 2022-05-11
Added config object to engine initialization (#2057) (#5470)

BStringhamVRSK:
This code adds a config object, "ConfigParams," to Engine creation parameters for setting memory buffer sizes for command buffers and driver handle arena sizes.  It will use #define values (FILAMENT_PER_FRAME_COMMANDS_SIZE_IN_MB, etc.) as defaults (and as minimum acceptable values) if the user does not provide values.  It attempts to validate the values given to prevent the user from creating a unusable state.

**NOTE:** I have tried to be very careful in making changes that affect all platforms, but I am not set up to build, run, and test anything other than PC desktop at the moment.  All other platforms should build and run correctly, but they've not been directly tested.

### DIFF
--- a/filament/backend/src/metal/MetalDriver.mm
+++ b/filament/backend/src/metal/MetalDriver.mm
@@ -40,20 +40,16 @@
 namespace filament {
 namespace backend {
 
-Driver* MetalDriverFactory::create(MetalPlatform* const platform, const Platform::DriverConfig& driverConfig) {
+Driver* MetalDriverFactory::create(MetalPlatform* const platform, Platform::DriverConfig& driverConfig) {
     return MetalDriver::create(platform, driverConfig);
 }
 
 UTILS_NOINLINE
-Driver* MetalDriver::create(MetalPlatform* const platform, const Platform::DriverConfig& driverConfig) {
+Driver* MetalDriver::create(MetalPlatform* const platform, Platform::DriverConfig& driverConfig) {
     assert_invariant(platform);
-    return new MetalDriver(platform, driverConfig);
-}
-
-size_t MetalDriver::getHandleArenaSize(const Platform::DriverConfig& driverConfig) noexcept {
-    size_t configSize = driverConfig.getHandleArenaSize();
     size_t defaultSize = FILAMENT_METAL_HANDLE_ARENA_SIZE_IN_MB * 1024U * 1024U;
-    return configSize > defaultSize ? configSize : defaultSize;
+    driverConfig.handleArenaSize = std::max(driverConfig.handleArenaSize, defaultSize);
+    return new MetalDriver(platform, driverConfig);
 }
 
 Dispatcher MetalDriver::getDispatcher() const noexcept {
@@ -63,7 +59,7 @@ Dispatcher MetalDriver::getDispatcher() const noexcept {
 MetalDriver::MetalDriver(MetalPlatform* platform, const Platform::DriverConfig& driverConfig) noexcept
         : mPlatform(*platform),
           mContext(new MetalContext),
-          mHandleAllocator("Handles", getHandleArenaSize(driverConfig)) {
+          mHandleAllocator("Handles", driverConfig.handleArenaSize) {
     mContext->driver = this;
 
     mContext->device = mPlatform.createDevice();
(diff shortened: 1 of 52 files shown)


pixelflinger:
thank you for the PR, please give us a bit of time for the review

filament/include/filament/Engine.h
```diff
@@ -54,6 +54,22 @@ class LightManager;
...
+// size of the high-level draw commands buffer (comes from the per-render pass allocator)
+#    define FILAMENT_PER_FRAME_COMMANDS_SIZE_IN_MB 2
+#endif
+
+#ifndef FILAMENT_MIN_COMMAND_BUFFERS_SIZE_IN_MB
+// size of a command-stream buffer (comes from mmap -- not the per-engine arena)
+#    define FILAMENT_MIN_COMMAND_BUFFERS_SIZE_IN_MB 1
+#endif
+
+
```
pixelflinger, line 72: I don't think we need this in the public header.

filament/include/filament/Engine.h
```diff
@@ -151,6 +167,70 @@ class UTILS_PUBLIC Engine {
...
     using Platform = backend::Platform;
     using Backend = backend::Backend;
+    using DriverConfig = backend::Platform::DriverConfig;
+
+    /**
+    * ConfigParams are used to define the memory footprint used by the engine, such as the
+    * command buffer size. ConfigParams can be used to customize engine requirements based 
+    * on the applications needs.
+    */
+    class ConfigParams {
```
pixelflinger, line 177: Make this a simple `struct` with public fields. Also rename to `Config`. e.g.:

```
struct Config {
    uint8_t commandBufferSize = FILAMENT_COMMAND_BUFFER_SIZE_IN_MB;   //!< size of command buffer in MiB
    ...
};
```

Validate the parameters in `Engine::create`. No need to store `DriverConfig` in this structure -- just create one when needed to pass to the driver.

We basically don't want any private fields visible in public headers.

pixelflinger (changes requested):
There will be more comments, but I thought I'd make those for now as it's going to be a big change already.

Thank you for the PR.

filament/include/filament/Engine.h
BStringhamVRSK: Ok.  I'm using them as default values for parameters.  I will replace them with straight literals.  I'll continue to use the #define's inside the functions.

pixelflinger: Actually you're right, after you make the change where `Config` is a simple struct with default values, you'll need those in the public header -- so let's keep them here.

filament/backend/src/metal/MetalDriver.mm
```diff
@@ -40,24 +40,30 @@
...
+Driver* MetalDriver::create(MetalPlatform* const platform, const Platform::DriverConfig& driverConfig) {
     assert_invariant(platform);
-    return new MetalDriver(platform);
+    return new MetalDriver(platform, driverConfig);
+}
+
+size_t MetalDriver::getHandleArenaSize(const Platform::DriverConfig& driverConfig) noexcept {
+    size_t configSize = driverConfig.getHandleArenaSize();
+    size_t defaultSize = FILAMENT_METAL_HANDLE_ARENA_SIZE_IN_MB * 1024U * 1024U;
+    return configSize > defaultSize ? configSize : defaultSize;
```
pixelflinger, line 56: use `std::max` also document in `DriverConfig` that at least the default value will be used -- which provides a way to overrides `DriverConfig` by a build provided value.

pixelflinger (changes requested):

BStringhamVRSK:
Arghh!  I did a rebase of my changes, deleted this branch and then pushed it up again.  It closed my PR, which I did not want to happen.  Can someone reopen this PR for me?  I don't have permissions, apparently.

BStringhamVRSK:
I pushed my branch up again, and it appears to be there.  Not sure what the problem is.

BStringhamVRSK:
Should I create a new PR?

romainguy:
Seems like it :/

BStringhamVRSK:
Created new pull request: #5556

PR #5542, opened 2022-05-11
Use codegen for TypeScript definitions.

prideout:
This is a bit fancy because it modifies a file rather than generating
one from scratch.


tools/codegen-options/README.md
```diff
@@ -28,13 +28,13 @@ include:
 - If the default value of a field is a vector, it must be in the form: `{ x, y, z }`.
 - The type of every struct field must be specified without any whitespace (e.g. `Texture*`).
 - There must be no string literals that contain keywords.
-- Enum names must be disambiguated when they appear in a struct field default value.
+- Enums must be fully qualified when they appear in struct field types and default values.
```
pixelflinger, line 31: do you think it's a restriction we could lift later?

pixelflinger (approved):

tools/codegen-options/README.md
prideout: Yes I think we should try to make this better in the future.  Another weird thing is that the JS bindings scope all these things under "View", which I'm not sure is still appropriate, now that we have top-level options.  It works because of the aliases.

By the way, initially I thought it would be cool to use C++ as ground truth but I'm less confident now.  (eg. think of all the different ways to initialize a float3...). However I'm 100% sure that we made the right decision with Go, it has a great templating system and it will be fairly easy to migrate it over to an IDL if we want.

pixelflinger: One downside of using an IDL is that until you generate the files you don't have IDE support for the changes/types. So if it's part of the build system (which we will probably want eventually), it becomes really annoying during development.

I think I still prefer to keep C++ as the ground truth and limit the allowed syntax.

prideout: Okay, that's an excellent point.  The  script needs to be better at catching unsupported syntax though.  Baby steps.

PR #5546, opened 2022-05-12
codegen: improve handling of enums

prideout:


pixelflinger (approved):
nice. Is fully qualification supported but not necessary?

PR #5556, opened 2022-05-13
Bstringham/geosp 263 config params resubmit (#2057) (#5470) (#5537)

BStringhamVRSK:
Resubmitting this PR because the other one (#5537) was closed and couldn't be reopened.


filament/src/details/Engine.cpp
```diff
@@ -188,6 +217,13 @@ FEngine::FEngine(Backend backend, Platform* platform, void* sharedGLContext) :
     // (it may not be the case)
     mJobSystem.adopt();
 
+    const size_t MB = 1024U * 1024U;
```
pixelflinger, line 220: constexpr and call it `MiB` which is more accurate (MB refers to 1000x1000)

filament/src/details/Engine.cpp
```diff
@@ -583,7 +619,10 @@ int FEngine::loop() {
     JobSystem::setThreadName("FEngine::loop");
     JobSystem::setThreadPriority(JobSystem::Priority::DISPLAY);
 
-    mDriver = mPlatform->createDriver(mSharedGLContext);
+    DriverConfig driverConfig { .handleArenaSize = mDriverHandleArenaSize };
+    mDriver = mPlatform->createDriver(mSharedGLContext, driverConfig);
+    mDriverHandleArenaSize = driverConfig.handleArenaSize;      // Get updated handle arena size in case driver changed it
```
pixelflinger, line 624: Same comments as above

filament/src/details/Engine.h
```diff
@@ -126,19 +126,25 @@ class FEngine : public Engine {
...
-    static constexpr size_t CONFIG_PER_RENDER_PASS_ARENA_SIZE   = filament::CONFIG_PER_RENDER_PASS_ARENA_SIZE;
-    static constexpr size_t CONFIG_PER_FRAME_COMMANDS_SIZE      = filament::CONFIG_PER_FRAME_COMMANDS_SIZE;
-    static constexpr size_t CONFIG_MIN_COMMAND_BUFFERS_SIZE     = filament::CONFIG_MIN_COMMAND_BUFFERS_SIZE;
-    static constexpr size_t CONFIG_COMMAND_BUFFERS_SIZE         = filament::CONFIG_COMMAND_BUFFERS_SIZE;
+    // Creation parameters
+    size_t mMinCommandBufferSize;           // minimum size of command buffer (in bytes)
+    size_t mCommandBufferSize;              // size of command buffer (in bytes)
+    size_t mPerFrameCommandsSize;           // size of the high-level draw commands buffer (in bytes)
+    size_t mPerRenderPassArenaSize;         // size of the per-pass arena buffer (in bytes)
+    size_t mDriverHandleArenaSize;          // size of driver handle arena (in bytes)
```
pixelflinger, line 134: move these in the `private` section towards the bottom (with all the other private fields)

filament/src/details/Engine.h
```diff
@@ -126,19 +126,25 @@ class FEngine : public Engine {
...
 
 public:
     static FEngine* create(Backend backend = Backend::DEFAULT,
-            Platform* platform = nullptr, void* sharedGLContext = nullptr);
+            Platform* platform = nullptr, void* sharedGLContext = nullptr,
+            const Config* config = nullptr);
+
+    static void validateConfig(const Config* srcConfig, Config& validConfig) noexcept;
```
pixelflinger, line 141: this shouldn't be public.

filament/src/details/Engine.cpp
```diff
@@ -167,7 +196,7 @@ static constexpr float4 sFullScreenTriangleVertices[3] = {
 // these must be static because only a pointer is copied to the render stream
 static const uint16_t sFullScreenTriangleIndices[3] = { 0, 1, 2 };
 
-FEngine::FEngine(Backend backend, Platform* platform, void* sharedGLContext) :
+FEngine::FEngine(Backend backend, Platform* platform, void* sharedGLContext, const Config& configParams) :
```
pixelflinger, line 199: I think it's fine to use a reference here (so it can't be nullptr) but in that case, I would move this parameter in 2nd position after `backend`.

Also name the parameters the same in the header and in the cpp (e.g. `config` in both cases).

pixelflinger (changes requested):

filament/backend/src/opengl/OpenGLDriver.cpp
```diff
@@ -136,7 +136,9 @@ Driver* OpenGLDriver::create(
         }
     }
 
-    OpenGLDriver* const driver = new OpenGLDriver(ec);
+    size_t defaultSize = FILAMENT_OPENGL_HANDLE_ARENA_SIZE_IN_MB * 1024U * 1024U;
+    driverConfig.handleArenaSize = std::max(driverConfig.handleArenaSize, defaultSize);
```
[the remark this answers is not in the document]
BStringhamVRSK: The problem is that each backend has a (potentially) different default handle arena size, driven by a different #define value: 
FILAMENT_OPENGL_HANDLE_ARENA_SIZE_IN_MB = 4 FILAMENT_METAL_HANDLE_ARENA_SIZE_IN_MB = 8
FILAMENT_VULKAN_HANDLE_ARENA_SIZE_IN_MB = 8

That is also why the DriverConfig passed into createDriver is not a const.  Only the driver knows how to validate the handle arena size because only it has the default values.  CreateDriver potentially modifies the DriverConfig.handleArenaSize when it validates the value.  Engine can then look at the modified value, if it desires.  Currently, Engine stores the validated value for reference after createDriver is called.

filament/backend/include/backend/Platform.h
```diff
@@ -56,7 +60,7 @@ class UTILS_PUBLIC Platform {
      *
      * @return nullptr on failure, or a pointer to the newly created driver.
      */
-    virtual backend::Driver* createDriver(void* sharedContext) noexcept = 0;
+    virtual backend::Driver* createDriver(void* sharedContext, DriverConfig& driverConfig) noexcept = 0;
```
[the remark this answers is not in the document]
BStringhamVRSK: See my other comment about default handle arena sizes.

filament/backend/include/backend/Platform.h
```diff
@@ -37,6 +37,10 @@ class UTILS_PUBLIC Platform {
         uintptr_t image = 0;
     };
 
+    struct DriverConfig {
+        size_t handleArenaSize = 0; // size of handle arena in bytes
```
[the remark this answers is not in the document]
BStringhamVRSK: Fixed

filament/src/details/Engine.cpp
```diff
@@ -61,11 +62,13 @@ namespace filament {
...
 using namespace filaflat;
 
-FEngine* FEngine::create(Backend backend, Platform* platform, void* sharedGLContext) {
+FEngine* FEngine::create(Backend backend, Platform* platform, void* sharedGLContext, const Config *config) {
     SYSTRACE_ENABLE();
     SYSTRACE_CALL();
 
-    FEngine* instance = new FEngine(backend, platform, sharedGLContext);
+    Config validConfig;
+    validateConfig(config, validConfig);
```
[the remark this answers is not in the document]
BStringhamVRSK: Fixed in my next update.

filament/src/details/Engine.cpp
```diff
@@ -83,7 +86,10 @@ FEngine* FEngine::create(Backend backend, Platform* platform, void* sharedGLCont
             delete instance;
             return nullptr;
         }
-        instance->mDriver = platform->createDriver(sharedGLContext);
+        DriverConfig driverConfig{ .handleArenaSize = instance->mDriverHandleArenaSize };
```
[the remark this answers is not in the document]
BStringhamVRSK: See my other comment about default handle arena size.

filament/src/details/Engine.cpp
BStringhamVRSK: Fixed in my next update.

filament/src/details/Engine.h
BStringhamVRSK: Fixed in my next update.

filament/src/details/Engine.h
BStringhamVRSK: Fixed in my next update.

filament/src/details/Engine.cpp
BStringhamVRSK: Fixed in my next update.

filament/backend/src/opengl/OpenGLDriver.cpp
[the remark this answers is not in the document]
pixelflinger: It doesn't seem like FEngine needs it actually. So let's not store it in FEngine there is no point.
I really don't like in/out values -- so I think we don't store the modified value in FEngine (Which we don't need), then we can keep the parameter as const.

filament/src/details/Engine.cpp
```diff
@@ -130,6 +138,27 @@ void FEngine::createAsync(CreateCallback callback, void* user,
...
+        return;
+    }
+
+    validConfig.minCommandBufferSizeMB = max((uint32_t)FILAMENT_MIN_COMMAND_BUFFERS_SIZE_IN_MB, config->minCommandBufferSizeMB);
+    validConfig.minCommandBufferSizeMB = max(config->perFrameCommandsSizeMB, validConfig.minCommandBufferSizeMB);                   // As a rule of thumb use the same value as perFrameCommandsMB
+    validConfig.commandBufferSizeMB = max(config->commandBufferSizeMB, validConfig.minCommandBufferSizeMB * NUM_FRAMES_OVERLAP);
+    validConfig.perFrameCommandsSizeMB = max((uint32_t)FILAMENT_PER_FRAME_COMMANDS_SIZE_IN_MB, config->minCommandBufferSizeMB);
+    validConfig.perRenderPassArenaSizeMB = max((uint32_t)FILAMENT_PER_RENDER_PASS_ARENA_SIZE_IN_MB, config->perRenderPassArenaSizeMB);
+    validConfig.perRenderPassArenaSizeMB = max(validConfig.perRenderPassArenaSizeMB, validConfig.perFrameCommandsSizeMB + ARENA_COMMANDS_DELTA);    // Enforce pre-render-pass arena rule-of-thumb
+    validConfig.driverHandleArenaSizeMB = config->driverHandleArenaSizeMB;      // This value gets validated during driver creation, so pass it through
```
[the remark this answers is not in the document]
BStringhamVRSK: Marking this as resolved, based on our other conversations.  The driver will validate, but not modify the config value.

filament/backend/src/opengl/OpenGLDriver.cpp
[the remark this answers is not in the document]
BStringhamVRSK: Based on this and our other conversations, I will take this approach:
* Engine::Config will continue to contain a parameter for the driver handle arena size and pass it through to Driver.
* The driver will validate the value, but not modify the config struct value.
* DriverConfig will be passed as a const.
* FEngine does need to store the requested driver handle arena size so it can be used later in FEngine::loop(), but I will not read back the value to discover the validated size.

filament/src/details/Engine.cpp
```diff
@@ -130,6 +135,27 @@ void FEngine::createAsync(CreateCallback callback, void* user,
...
 }
 
+Engine::Config FEngine::validateConfig(const Config* config) noexcept
+{
+    constexpr uint32_t ARENA_COMMANDS_DELTA = 1;    // Rule of thumb: perRenderPassArenaMB must be roughly 1 MB larger than perFrameCommandsMB
+    constexpr uint32_t NUM_FRAMES_OVERLAP = 3;      // Number of potential concurrent frames in flight
+
+    Config validConfig;
+    if (config == nullptr)
+    {
```
pixelflinger, line 145:
```suggestion
    if (!config) {
```

filament/src/details/Engine.h
```diff
@@ -356,8 +353,16 @@ class FEngine : public Engine {
     backend::Handle<backend::HwTexture> getZeroTextureArray() const { return mDummyZeroTextureArray; }
     backend::Handle<backend::HwTexture> getOneIntegerTextureArray() const { return mDummyOneIntegerTextureArray; }
 
+    size_t getMinCommandBufferSize()    { return mMinCommandBufferSize; }
+    size_t getCommandBufferSize()       { return mCommandBufferSize; }
+    size_t getPerFrameCommandsSize()    { return mPerFrameCommandsSize; }
+    size_t getPerRenderPassArenaSize()  { return mPerRenderPassArenaSize; }
+    size_t getRequestedDriverHandleArenaSize() { return mRequestedDriverHandleArenaSize; }
```
pixelflinger, line 360: all these should be `const noexcept`

pixelflinger (changes requested):
Just a few comments (see inlines). Looks pretty good. Thank you!

bejado (approved):

filament/backend/src/metal/MetalDriver.mm
```diff
@@ -40,24 +40,26 @@
...
+    return MetalDriver::create(platform, driverConfig);
 }
 
 UTILS_NOINLINE
-Driver* MetalDriver::create(MetalPlatform* const platform) {
+Driver* MetalDriver::create(MetalPlatform* const platform, const Platform::DriverConfig& driverConfig) {
     assert_invariant(platform);
-    return new MetalDriver(platform);
+    size_t defaultSize = FILAMENT_METAL_HANDLE_ARENA_SIZE_IN_MB * 1024U * 1024U;
+    Platform::DriverConfig validConfig { .handleArenaSize = std::max(driverConfig.handleArenaSize, defaultSize) };
```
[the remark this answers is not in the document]
BStringhamVRSK: I would have done that, but the constructor passes the handleArenaSize to the mHandleAllocator via a member initializer list.  I would have to inline all of that validation, which seems pretty messy.

filament/src/details/Engine.cpp
BStringhamVRSK: Yeah, that's my C# showing through :) .  Most of my career has been spent in C++, but recent years have focused on C#.

romainguy:
Related task in our projects: https://github.com/google/filament/projects/10#card-12101085

romainguy:
This changes fixes issue #2057

BStringhamVRSK:
@romainguy @pixelflinger @bejado 
Can someone help with the JS bindings?  I am out of my depth there, trying to figure out how the emscripten bindings and argument pass-throughs work.

Take a look at my most recent changes to `JSBindings.cpp` and `extensions.js`.  Clearly doing something wrong, but unsure how to fix it.

prideout:
@[BStringhamVRSK](https://github.com/BStringhamVRSK) you can just revert the JavaScript and TypeScript stuff for now, we can use default config and I can expose this stuff to web later.

pixelflinger (changes requested):
I proposed better explanations for the buffers.

BStringhamVRSK:
> 

@prideout I have rolled back my changes to `extensions.js`, `filament.d.ts`, and `jsbindings.cpp`.
I'm not sure that Web will compile, but it wasn't compiling before.

BStringhamVRSK:
@pixelflinger 
I committed your suggested changes to the buffer explanations.  The commit that GitHub created lists you as a co-author, which is fine.  But, now it is saying that you do not have an authorized `Contributer Licensing Agreement`, and is now failing all builds because of that.  I assume you can work around this?

pixelflinger:
@BStringhamVRSK you now need to update the `RELEASE_NOTES.md`.  Add a line there, something like "- engine: Add memory configuration parameters to Engine"
Sorry about that!

BStringhamVRSK:
> @BStringhamVRSK you now need to update the `RELEASE_NOTES.md`. Add a line there, something like "- engine: Add memory configuration parameters to Engine" Sorry about that!

No problem.  Just pushed up that change.

pixelflinger:
The web build is still failing, @prideout? (can you advise?).

prideout (approved):
LGTM, when merging this should definitely be squashed.

PR #5568, opened 2022-05-17
fix trailing zero when setting shader sources

pixelflinger:
the shader source blob's size included the null terminating character,
which ended up being passed to the opengl/metal API.


prideout (approved):
I double checked the docs for glShaderSource and this is correct, good catch.  (apparently we can also pass `-1` into the last arg of glShaderSource to make it scan for the null terminator, but that seems unsafe to me)

romainguy (approved):

PR #5570, opened 2022-05-18
beamsplitter: use generated code for View.java

prideout:
Note that JNI c++ functions still need to be manually maintained, as well as the setter / getter
methods.

Everything before the following codeline must be manually maintained:

```
// The remainder of this file is generated by beamsplitter
```

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -5,6 +5,8 @@ A new header is inserted each time a *tag* is created.
 
 ## main branch
 
+- Java View has several minor changes due to generated code, such as field ordering.
+
 ## v1.22.2
 
 - Java minor API change: rename ssctStartTraceDistance to ssctShadowDistance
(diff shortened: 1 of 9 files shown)


prideout:
The diff in `View.java` is hard to read, but it basically moves all options-related type definitions to the bottom.  That's just a side effect of the codegen.  In theory there no zero API changes caused by this PR.  Even the annotations like `@Nullable` should still be there.

pixelflinger (approved):
that's amazing!

pixelflinger:
> The diff in `View.java` is hard to read, but it basically moves all options-related type definitions to the bottom. That's just a side effect of the codegen. In theory there no API changes caused by this PR. Even the annotations like `@Nullable` should still be there.

So that means that we now have the javadoc only in one place (the cpp file)?

bejado (approved):

PR #5580, opened 2022-05-20
gltfio: Clear texture caches before loading resources

### LINKED ISSUE #5527: Crashed when load model after destory model
wecom-temp-45b2fd26107254a2917b3f8585bd405c

Load a glb model, and then destory the model, and load the glb model again. Crash happened when Load the glb model again.

bejado:
Some textures would persist in the cache between calls to `loadResources`, but had since been destroyed. This caused a crash in the iOS gltf-viewer sample when reloading the model.

Fixes #5527


prideout (approved):
Thanks for diagnosing this, sorry for the trouble!

PR #5583, opened 2022-05-21
Fix not calculating the normal while morphing in the legacy path

daemyung:
When the material doesn't need TBN, Filament doesn't calculate morph target's normal. As a result, the geometry is changed but the geometry's normal is not changed while morphing. This PR fixes this issue and now calculating normals while morphing.

### DIFF
--- a/shaders/src/main.vs
+++ b/shaders/src/main.vs
@@ -70,9 +70,26 @@ void main() {
         toTangentFrame(mesh_tangents, material.worldNormal);
 
         #if defined(VARIANT_HAS_SKINNING_OR_MORPHING)
-            if ((objectUniforms.flags & FILAMENT_OBJECT_SKINNING_ENABLED_BIT) != 0u) {
-                skinNormal(material.worldNormal, mesh_bone_indices, mesh_bone_weights);
-            }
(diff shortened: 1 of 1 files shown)


prideout (commented):
Thanks for the fix!  I think this makes sense, but the commit message should be better.  It looks like morphing was not being applied at all in the legacy path, but only in certain conditions: (a) when tangents were provided *and* (b) the material did not use anisotropy or normal mapping.

@daemyung  Does this fix also need to be applied to the case where HAS_ATTRIBUTE_TANGENTS is false?

bejado (approved):
This LGTM, but agreed that we need a better commit message.

daemyung:
@pixelflinger Sorry for no description. I updated more information. Please let me know if it's not enough. I will try to write more description.

@pixelflinger @prideout @bejado How's about the current commit message which is `Fix not calculating the normal while morphing when the material doesn't need TBN`?

daemyung:
> @daemyung Does this fix also need to be applied to the case where HAS_ATTRIBUTE_TANGENTS is false?

Maybe not...? It seems necessary that `HAS_ATTRIBUTE_TANGENTS` is true. From my understanding, if `HAS_ATTRIBUTE_TANGENTS` is false then we don't need to calculate normal.

pixelflinger (approved):

PR #5595, opened 2022-05-25
gltfio: Streamline the AABB + skinning computation.

prideout:
This optimizes and cleans up some code from a 3P contributor.

When computing a bounding box, there was no need for an inner loop
through the entire skins array.

Tested using the torus model in #4973 and the `-r` flag in gltf_viewer.

### DIFF
--- a/libs/gltfio/src/ResourceLoader.cpp
+++ b/libs/gltfio/src/ResourceLoader.cpp
@@ -85,10 +85,6 @@ struct ResourceLoader::Impl {
     bool mIgnoreBindTransform;
     std::string mGltfPath;
 
-    // This is used to calculate skinIndex when updateBoundingBoxes, so that the mapping between
-    // cgltf_node* and FFilamentInstance::Skin can be retrieved. This pointer doesn't need to be freed.
-    cgltf_skin* mSkinBaseAddress;
-
     // User-provided resource data with URI string keys, populated with addResourceData().
     // This is used on platforms without traditional file systems, such as Android, iOS, and WebGL.
     UriDataCache mUriDataCache;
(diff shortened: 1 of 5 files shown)


libs/gltfio/src/ResourceLoader.cpp
```diff
@@ -902,101 +897,105 @@ void ResourceLoader::updateBoundingBoxes(FFilamentAsset* asset) const {
...
-    const size_t skinningAttrSize = cgltf_num_components(cgltf_type_vec4);
-    auto computeBoundingBoxSkinned = [&](const cgltf_primitive* prim, const Skin* skin, Aabb* result) {
-        Aabb aabb;
-        std::vector<mat4f> inverseGlobalTransforms(skin->targets.size());
-        for (size_t i = 0; i < skin->targets.size(); i++) {
-            auto xformable = tm.getInstance(skin->targets[i]);
-            if (xformable) {
-                inverseGlobalTransforms[i] = inverse(tm.getWorldTransform(xformable));
-            }
-        }
```
prideout, line 915: This is one of the two loops that I removed, because a given primitive can only be influenced by a single skin.

libs/gltfio/src/ResourceLoader.cpp
```diff
@@ -902,101 +897,105 @@ void ResourceLoader::updateBoundingBoxes(FFilamentAsset* asset) const {
...
+            for (size_t j = 0; j < 4; j++) {
+                size_t jointIndex = joints[i][j];
+                Entity jointEntity = prim.skin->joints[jointIndex];
                 mat4f globalJointTransform = tm.getWorldTransform(tm.getInstance(jointEntity));
-                mat4f inverseBindMatrix = skin->inverseBindMatrices[jointIndex];
-                tmp += jointWeight * globalJointTransform * inverseBindMatrix;
+                mat4f inverseBindMatrix = prim.skin->inverseBindMatrices[jointIndex];
+                tmp += weights[i][j] * globalJointTransform * inverseBindMatrix;
             }
-            for (const auto& inverseGlobalTransform: inverseGlobalTransforms) {
```
prideout, line 951: This is the other inner loop that I removed.

PR #5598, opened 2022-05-25
Add JS bindings for Texture class methods

### LINKED ISSUE #4492: JavaScript Request: Expose Texture methods getWidth(), getHeight(), getDepth(), getLevels(), etc.
**Problem:**
After using Engine.createTextureFromKtx/jpeg/Png, I need to query the texture width and height so I can generate proper UV coordinates for font display.

**Proposed solution**
Expose the Texture.getWidth(), getHeight(), etc. methods to get texture information.  I currently only need width and height, but I can see usage cases for many of the other "getters" available in the C++ interface.

**OS and backend**
WebGL

domcorvasce:
This PR introduces JS bindings for the following methods: `getWidth`, `getHeight`, `getDepth`, and `getLevels`

Closes #4492

### DIFF
--- a/docs/remote/filament.js
+++ b/docs/remote/filament.js
@@ -611,6 +611,22 @@ Filament.loadClassExtensions = function() {
         pbd.delete();
     }
 
+    Filament.Texture.prototype.getWidth = function(engine, level = 0) {
+        return this._getWidth(engine, level);
+    }
+
+    Filament.Texture.prototype.getHeight = function(engine, level = 0) {
+        return this._getHeight(engine, level);
+    }
+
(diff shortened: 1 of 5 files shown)


prideout (changes requested):
Thanks!  Can you add a `?` after the optional parameters in the typescript definitions?  Also, please revert the stuff in `docs/`, I prefer matching the built javascript artifacts with a particular release.

prideout (commented):
Thanks, almost there!  Looks like `docs/webgl/filament.js` is still modified?

prideout (approved):

PR #5604, opened 2022-05-26
Update spirv-tools to get new code folding optimization.

prideout:
This change includes a fix for an issue we filed against spirv-tools:
https://github.com/KhronosGroup/SPIRV-Tools/issues/4371

So, it should reduce the number of extraneous matrix copies in finalized
GLSL code.

I did quick size total of all filamat files (*) before and after this
update, it went from 11.3 MB to 11.1 MB.

(*) All backends are enabled, includes built-in Filament shaders and
gltfio ubershaders.


romainguy (approved):
It might also fix driver bugs we were running into because of the generated code.

PR #5605, opened 2022-05-26
extend `blendOrder` functionality

pixelflinger:
blendOrder was used to control the draw order of blended render
primitive within a Renderable. We now have an option to make the
blend order global, in this case those primitives with a global blend
order are always sorted solely using the blend order value (i.e. the 
distance from the camera is not take into account).

### DIFF
--- a/android/filament-android/src/main/java/com/google/android/filament/RenderableManager.java
+++ b/android/filament-android/src/main/java/com/google/android/filament/RenderableManager.java
@@ -191,7 +191,9 @@ public class RenderableManager {
         }
 
         /**
-         * Sets an ordering index for blended primitives that all live at the same Z value.
+         * Sets the drawing order for blended primitives. The drawing order is either global or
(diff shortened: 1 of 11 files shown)


prideout (approved):
I can't find anything wrong!

android/filament-android/src/main/java/com/google/android/filament/RenderableManager.java
```diff
@@ -203,6 +205,18 @@ public Builder blendOrder(@IntRange(from = 0) int index,
             return this;
         }
 
+       /**
+         * Sets whether the blend order is global or local to this Renderable (by default).
```
bejado, line 209: Maybe put the "(by default)" right after "local":

```suggestion
         * Sets whether the blend order is global or local (by default) to this Renderable.
```

Same suggestion elsewhere.

bejado (approved):

PR #5624, opened 2022-06-01
PlatformCocoaGL: If a share context is specified, use its pixel format

jfaust:
This is the CocoaGL equivalent to https://github.com/google/filament/pull/1456

I was unable to figure out the right settings to make a GLFW shared context work without this, despite, as far as I can tell, the pixel format attributes ending up identical.

There's a demo/test of this at https://github.com/jfaust/filament-glfw-share-context. It will crash on a NULL context without this change.

### DIFF
--- a/filament/backend/src/opengl/platforms/PlatformCocoaGL.mm
+++ b/filament/backend/src/opengl/platforms/PlatformCocoaGL.mm
@@ -150,7 +150,13 @@ Driver* PlatformCocoaGL::createDriver(void* sharedContext) noexcept {
     };
 
     NSOpenGLContext* shareContext = (__bridge NSOpenGLContext*) sharedContext;
-    NSOpenGLPixelFormat* pixelFormat = [[NSOpenGLPixelFormat alloc] initWithAttributes:pixelFormatAttributes];
+    NSOpenGLPixelFormat* pixelFormat;
+    // If a share context is specified, use its pixel format
+    if (shareContext) {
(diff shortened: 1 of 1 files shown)


jfaust:
Absolutely - I think that applies to what's already happening in the GLX context though.

I could add some code to check that the pixel format is compatible, and abort if not. If I'm understanding correctly that would mean checking that all attributes specified in the `pixelFormatAttributes` array match what's expected (presumably minus 3_2Core, since 4_1Core should work?):
```
NSOpenGLPixelFormatAttribute pixelFormatAttributes[] = {
            NSOpenGLPFAOpenGLProfile, NSOpenGLProfileVersion3_2Core,
            NSOpenGLPFADepthSize,    (NSOpenGLPixelFormatAttribute) 24,
            NSOpenGLPFADoubleBuffer, (NSOpenGLPixelFormatAttribute) true,
            NSOpenGLPFAAccelerated,  (NSOpenGLPixelFormatAttribute) true,
            NSOpenGLPFANoRecovery,   (NSOpenGLPixelFormatAttribute) true,
            0, 0,
    };
```

jfaust:
Ah, that made me realize what was missing from the GLFW context: `NSOpenGLPFANoRecovery`. Which they don't provide a way to set.

romainguy:
Nice find. Maybe a better solution would be to remove that attribute then? I don't think we care whether that's set really.

jfaust:
Closing in favor of https://github.com/google/filament/pull/5626

PR #5632, opened 2022-06-02
disable user scissor while rendering the shadowmaps

### LINKED ISSUE #5607: setScissors influences shadows
When I do a `setScissors` on all my material instances, the rendering is restricted to the rectangle set there, as expected. However, shadows are truncated in a region which is *smaller* than the rectangle fed to `setScissors`, so that parts of my scene no longer receive shadows. It looks like the scissor test is applied to some aspects of the shadow computations, instead of just preventing pixels from being drawn.

Does this ring any bells? If it doesn't, then I'll try to create a minimal example.

pixelflinger:
fixes #5607

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -6,6 +6,7 @@ A new header is inserted each time a *tag* is created.
 ## main branch
 
 - gltfio: fix morphing for un-packed accessors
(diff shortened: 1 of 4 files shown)


bejado (approved):
Please update RELEASE_NOTES :)

PR #5633, opened 2022-06-02
GitHub Actions: verify RELEASE_NOTES is updated for each PR

bejado:
This check will verify that all PRs henceforth modify RELEASE_NOTES.md. To bypass, apply the 'internal' label to your PR.

Note: the check will fail for this PR since it checks out the `.github/` path from main, so won't work until this PR is merged.


prideout (approved):
Our first custom action...and it run Python in a Docker container :P

PR #5644, opened 2022-06-04
Metal: add support for scissor

MasTraER:
This patch enables user scissor in Metal backend; There was no implementation for it.
While the absence of the feature in Metal does not incur serious problem, many rendering glitches were found in apps using ImGui due to it.


bejado:
@MasTraER after looking into this today, I realize there's some details I still need to work out before scissor support can be implemented. Mainly, Filament assumes an OpenGL (bottom-left origin) coordinate system, but Metal uses a top-left origin. The existing viewport calculation inside of `beginRenderPass` doesn't take this correctly into account (it seems like it does, but the calculation is off due to the difference between render target versus texture dimensions).

I'll give it some more consideration, and probably update this PR with the necessary changes. Thanks again!

prideout:
@bejado Just as an FYI, ImGui uses scissoring but the `gltf_viewer` rarely exercises it, which is why this hasn't been a noticeable problem.  One way to exercise this is to hack ViewerGui.cpp by adding a long combo after `ImGui::Begin()` like this:

```
    bool foo = true;
    if (ImGui::BeginCombo("Test", "hi")) {
        for (int i = 0; i < 100; i++) ImGui::Selectable("hello", &foo);
        ImGui::EndCombo();
    }
```

Screen Shot 2022-06-29 at 9 58 20 AM

bejado (approved):

prideout (approved):

pixelflinger (approved):

PR #5661, opened 2022-06-08
Added a new `instanced` material parameter

pixelflinger:
This boolean parameter is used with materials that need to access 
`getInstanceIndex()` (filament's equivalent to `gl_InstanceIndex`).
It is false by default, and getInstanceIndex() is not accessible.
This is intended to be used in concert with
`RenderableManager::Builder::instances()`.


docs/Materials.md.html
```diff
@@ -1119,6 +1119,27 @@
...
+Type
+:    `boolean`
+
+Value
+:     `true` or `false`. Defaults to `false`.
+
+Description
+:    Allows a material to access the instance index (i.e.: **`gl_InstanceIndex`**) of instanced
+     primitives using `getInstanceIndex()` in the material's shader code. Never use
+     **`gl_InstanceIndex`** directly. This is typically used with
```
bejado, line 1133: Maybe we could add a warning to `matc` if we detect someone using `gl_InstanceIndex` / `gl_InstanceID`

bejado (approved):

docs/Materials.md.html
pixelflinger: that's a good idea. probably same with gl_FragCoord.

PR #5675, opened 2022-06-10
OpenGL: add WebGL support for ReadPixels.

prideout:


filament/backend/src/opengl/OpenGLDriver.cpp
```diff
@@ -38,6 +38,10 @@
 // Support is currently disabled as we don't need it
 #define TEXTURE_2D_MULTISAMPLE_SUPPORTED false
 
+extern "C" {
+void glGetBufferSubData(GLenum target, GLintptr offset, GLsizeiptr size, void *data);
+}
+
```
pixelflinger, line 44: I would define that in the proper place in `gl_headers.h`

filament/backend/src/opengl/OpenGLDriver.cpp
```diff
@@ -2622,7 +2626,15 @@ void OpenGLDriver::readPixels(Handle<HwRenderTarget> src,
         PixelBufferDescriptor& p = *pUserBuffer;
         auto& gl = mContext;
         gl.bindBuffer(GL_PIXEL_PACK_BUFFER, pbo);
-        void* vaddr = glMapBufferRange(GL_PIXEL_PACK_BUFFER, 0,  p.size, GL_MAP_READ_BIT);
+        void* vaddr = nullptr;
+        FixedCapacityVector<uint8_t> clientBuffer;
```
pixelflinger, line 2630: maybe use a unique_ptr with malloc or new?

pixelflinger (changes requested):

filament/backend/src/opengl/gl_headers.h
```diff
@@ -115,6 +115,13 @@
 
 #endif
 
+// This is an odd duck function that exists in WebGL 2.0 but not in OpenGL ES.
+#if defined(__EMSCRIPTEN__)
```
pixelflinger, line 119: if you're using __EMSCRIPTEN__ here, I think you should use __EMSCRIPTEN__ in OpenGLDriver.cpp where you're using `glGetBufferSubData`. It would be cleaner.

pixelflinger (approved):

filament/backend/src/opengl/OpenGLDriver.cpp
```diff
@@ -2622,7 +2622,15 @@ void OpenGLDriver::readPixels(Handle<HwRenderTarget> src,
...
         auto& gl = mContext;
         gl.bindBuffer(GL_PIXEL_PACK_BUFFER, pbo);
-        void* vaddr = glMapBufferRange(GL_PIXEL_PACK_BUFFER, 0,  p.size, GL_MAP_READ_BIT);
+        void* vaddr = nullptr;
+        std::unique_ptr<uint8_t> clientBuffer;
+        if constexpr (HAS_MAPBUFFERS) {
+            vaddr = glMapBufferRange(GL_PIXEL_PACK_BUFFER, 0,  p.size, GL_MAP_READ_BIT);
+        } else {
+            clientBuffer.reset(new uint8_t(p.size));
+            glGetBufferSubData(GL_PIXEL_PACK_BUFFER, 0, p.size, clientBuffer.get());
```
pixelflinger, line 2631: I guess you can't use an undefined function with `if constexpr`, you'd have to use #ifdef

pixelflinger (changes requested):

pixelflinger (approved):

PR #5679, opened 2022-06-10
Always use homogeneous attachment sizes

bejado:
Homogeneous attachments are a requirement for Metal (and possibly Vulkan) correctness. If rendering to a RenderTarget that has a smaller height than the underlying Texture, the Metal backend must set the Viewport to be the bottom-left of the texture, to match OGL. However, it can only do this correctly if all attachments have the same size.

### DIFF
--- a/filament/backend/src/metal/MetalHandles.mm
+++ b/filament/backend/src/metal/MetalHandles.mm
@@ -28,6 +28,9 @@
 #include <utils/trap.h>
 #include <utils/debug.h>
 
+#include <math/vec2.h>
+#include <math/scalar.h>
+
 #include <math.h>
 
 namespace filament {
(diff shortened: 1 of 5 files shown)


bejado:
This causes an [assertion](https://github.com/google/filament/blob/9508bf2a0c996b47785fbc244a530e5287a37c80/filament/src/PostProcessManager.cpp#L1476) to fire when DOF is turned on:

```
// compute numbers of "safe" levels (should be 4, but can be 3 at half res)
const uint8_t mipmapCount = std::min(maxMipLevels, ctz(width | height));
assert_invariant(mipmapCount == maxMipLevels || mipmapCount == maxMipLevels-1);
```

pixelflinger (changes requested):
I think we should add/update the asserts in the backends and in the framefraph.

filament/backend/src/metal/MetalHandles.mm
```diff
@@ -816,6 +823,13 @@ BufferDescriptor b(nullptr, 0u, [](void* buffer, size_t size, void* user) {
                 "MetalRenderTarget was initialized with a MSAA COLOR%d texture, but sample count is %d.",
                 i, samples);
 
+        auto t = color[i].metalTexture;
+        const auto twidth = std::max(1u, t->width >> color[i].level);
+        const auto theight = std::max(1u, t->height >> color[i].level);
+        tmin = { std::min(tmin.x, twidth), std::min(tmin.y, theight) };
+        tmax = { std::max(tmax.x, twidth), std::max(tmax.y, theight) };
+        attachmentCount++;
+
```
pixelflinger, line 832: this min/max method is good, but I think we can do better, you can check that the bitwise OR and the bitwise AND of all the values are the same. You can use the fact that:
if a==b, then (a&b) == (a|b) and by extension
if a==b and b==c, then (a&b&c) == (a|b|c)

pixelflinger (changes requested):

pixelflinger (approved):

prideout (approved):

Issue #5659, opened 2022-06-08
stencil separate support


Michael-Lfx:
**Is your feature request related to a problem? Please describe.**
```c++
        glStencilFunc( GL_ALWAYS, 0, 0xff );
        glStencilOpSeparate( GL_FRONT, GL_KEEP, GL_INCR_WRAP, GL_INCR_WRAP );
        glStencilOpSeparate( GL_BACK, GL_KEEP, GL_DECR_WRAP, GL_DECR_WRAP );
        glDrawArrays( GL_TRIANGLES, 0, fcount );     
```
https://github.com/astiopin/sdf_atlas/blob/2158930242546df5827df75418ef5d046c10b97d/src/sdf_gl.cpp#L112

As shown above sdf rendering requires stencil buffer op. Currently filament only support stencil in OpenGL backend. And users can't  configure front or back buffer separately.

**Describe the solution you'd like**
It will be nice to export some fields in `.mat` file for stencil op.

**Describe alternatives you've considered**
No.

**OS and backend**
No.

romainguy:
Signed distance fields do not need stencil operations to work, it can be done entirely in the fragment shader.

bejado:
We do have plans to implement stencil support for all backends and exposed via material properties.

Michael-Lfx:
Thank you!  I manage to add some front back stencil ops in `struct RasterState` and Metal/GL backend.  It renders the same after replace sdf_gl.cpp mentioned above with a new introduced stencil field in material/.mat file. It will be nice to have an official support version to deprecate our weird impl. Here is a CL from our impl and there are some questions about fields memory bit align and allocation size.

```c++
//! Raster state descriptor
struct RasterState {

    using CullingMode = backend::CullingMode;
    using DepthFunc = backend::SamplerCompareFunc;
    using BlendEquation = backend::BlendEquation;
    using BlendFunction = backend::BlendFunction;
    using StencilFunction = backend::SamplerCompareFunc;
    using StencilOperation = backend::StencilOperation;

    RasterState() noexcept { // NOLINT
        static_assert(sizeof(RasterState) == 2 * sizeof(uint64_t),
                "RasterState size not what was intended");
        culling = CullingMode::BACK;
        blendEquationRGB = BlendEquation::ADD;
        blendEquationAlpha = BlendEquation::ADD;
        blendFunctionSrcRGB = BlendFunction::ONE;
        blendFunctionSrcAlpha = BlendFunction::ONE;
[...]

PR #5686, opened 2022-06-14
WIP: Add front & back stencil state

Michael-Lfx:
Try to address https://github.com/google/filament/issues/5659.
Could you review it? @bejado

### DIFF
--- a/filament/backend/include/backend/DriverEnums.h
+++ b/filament/backend/include/backend/DriverEnums.h
@@ -832,7 +832,7 @@ struct RasterState {
     using StencilOperation = backend::StencilOperation;
 
     RasterState() noexcept { // NOLINT
-        static_assert(sizeof(RasterState) == sizeof(uint64_t),
+        static_assert(sizeof(RasterState) == 2 * sizeof(uint64_t),
                 "RasterState size not what was intended");
         culling = CullingMode::BACK;
         blendEquationRGB = BlendEquation::ADD;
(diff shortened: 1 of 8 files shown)


bejado:
I think we do eventually want control over the front & back stencil states, but at this time I'm not sure we want to grow `RasterState` to 128 bits. @pixelflinger any thoughts? We may need to store the stencil state elsewhere.

romainguy:
That's a huge increase just for stencil ops. And again, SDF do *not* need this feature. Until there's a clear use case for this, we should not pay such a cost.

pixelflinger:
> I think we do eventually want control over the front & back stencil states, but at this time I'm not sure we want to grow `RasterState` to 128 bits. @pixelflinger any thoughts? We may need to store the stencil state elsewhere.

I agree, Stencil is seldom used and it takes so much state space. Maybe we need an indirection for stencil, but I'm not sure how to do that. A related issue is that we need to compare the RasterState for auto-instancing, so I don't like it becoming even bigger.

Michael-Lfx:
Same worry about increasing `RasterState` to 128 bits. Maybe use `uint64_t u; uint16_t v;` instead.  Again the PR is not intend to merge but just for seeking help from you. Should I push more commits?

Michael-Lfx:
Could you take another look? @bejado @pixelflinger @romainguy

bejado:
@Michael-Lfx We're going to hold off on this for now as we're having some internal discussions about how to best move forward with stencil state.

pixelflinger:
we can probably close this.

PR #5693, opened 2022-06-14
Update instructions for Vulkan + macOS.

prideout:


BUILDING.md
```diff
@@ -144,6 +144,9 @@ make sure the command line tools are setup by running:
 $ xcode-select --install
 ```
 
+If you wish to run the Vulkan backend instead of the default Metal backend, you must install
+the LunarG SDK, enable "System Global Components", and reboot your machine.
+
```
pixelflinger, line 149: This is not actually needed for building correct? it's just to be able to run the apps.

Maybe this should be (or also be) in `filament/README.md`, where it has this section like:
```
To use the Vulkan backend on macOS you must also make the following libraries available at runtime:
```

Maybe the informations contained in both places should be consolidated and copied (or referenced) in both places?

pixelflinger (changes requested):

BUILDING.md
prideout: Good catch, I updated `filament/README.md`.  I think duplication is fine here, asking the reader to "refer to" another section isn't as friendly as repeating the info.

pixelflinger (approved):

PR #5694, opened 2022-06-15
Fix HwRenderPrimitiveFactory with NoopDriver

bejado:
Problem: the `NoopDriver` always returns the same handle for RenderPrimitives (even if the buffers are different). This break's `HwRenderPrimitiveFactory`'s assumption that `createRenderPrimitive` will always be unique. What ends up happening is multiple `RenderPrimitive`s map to the same `HwRenderPrimitive`, which causes an assertion to fail upon a Google3 test termination.

### DIFF
--- a/filament/src/HwRenderPrimitiveFactory.cpp
+++ b/filament/src/HwRenderPrimitiveFactory.cpp
@@ -89,6 +89,17 @@ RenderPrimitiveHandle HwRenderPrimitiveFactory::create(DriverApi& driver,
         // create the backend object
         auto handle = driver.createRenderPrimitive(vbh, ibh,
                 type, offset, minIndex, maxIndex, count);
+
+        // First, check if the handle is already in our set. This shouldn't normally happen, except
+        // when the driver always returns the same handle for render primitives (such as the
+        // NoopDriver).
+        auto pos = mMap.find(handle.getId());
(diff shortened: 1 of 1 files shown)


bejado:
Actually, I'm not sure this is the correct solution, I need to think about it some more..

bejado:
A better solution might be to simply return incremental handles from `NoopDriver` instead of a static handle ID.

filament/src/HwRenderPrimitiveFactory.cpp
```diff
@@ -89,6 +89,17 @@ RenderPrimitiveHandle HwRenderPrimitiveFactory::create(DriverApi& driver,
...
+        // First, check if the handle is already in our set. This shouldn't normally happen, except
+        // when the driver always returns the same handle for render primitives (such as the
+        // NoopDriver).
+        auto pos = mMap.find(handle.getId());
+        if (UTILS_UNLIKELY(pos != mMap.end())) {
+            auto ipos = pos->second;
+            ipos->refs++;
+            return ipos->handle;
+        }
+
```
pixelflinger, line 102: would it work instead to have the noop driver return a random number or a counter? I don't like that this adds overhead in the real case.

pixelflinger (changes requested):

PR #5700, opened 2022-06-15
Vk: do not raise SIGINT for validation errors.

prideout:
This was occasionally useful for debugging specific problems, but it is
much too intrusive to enable in all debug builds.

### DIFF
--- a/filament/backend/src/vulkan/VulkanDriver.cpp
+++ b/filament/backend/src/vulkan/VulkanDriver.cpp
@@ -28,7 +28,6 @@
 #include <utils/CString.h>
 #include <utils/FixedCapacityVector.h>
 #include <utils/Panic.h>
-#include <utils/trap.h>
 
 #ifndef NDEBUG
 #include <set>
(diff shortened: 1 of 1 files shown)


filament/backend/src/vulkan/VulkanDriver.cpp
```diff
@@ -53,7 +52,6 @@ VKAPI_ATTR VkBool32 VKAPI_CALL debugReportCallback(VkDebugReportFlagsEXT flags,
         int32_t messageCode, const char* pLayerPrefix, const char* pMessage, void* pUserData) {
     if (flags & VK_DEBUG_REPORT_ERROR_BIT_EXT) {
         utils::slog.e << "VULKAN ERROR: (" << pLayerPrefix << ") " << pMessage << utils::io::endl;
-        utils::debug_trap();
     } else {
         utils::slog.w << "VULKAN WARNING: (" << pLayerPrefix << ") "
                 << pMessage << utils::io::endl;
```
pixelflinger, line 57: kind of strange to always log something.

pixelflinger (approved):

filament/backend/src/vulkan/VulkanDriver.cpp
prideout: This is never called in release builds.

PR #5701, opened 2022-06-15
BlueVKDarwin: Simplify the build and loading process.

prideout:
Some of our CMake logic was not necessary because we do not staticly
link against anything for Vulkan.  All entry points are dynamically
loaded.

Some of the load-time code was also needlessly complicated. In fact the
ICD environment variable that we were setting is now deprecated.

I tested this PR with:

- macOS + June LunarG SDK
- Android on a Pixel 6


romainguy (approved):

libs/bluevk/src/BlueVKDarwin.cpp
```diff
@@ -15,41 +15,31 @@
...
-    const Path dylibPath = VKLIBRARY_PATH;
-
-    // Provide a value for VK_ICD_FILENAMES only if it has not already been set.
-    const char* icd = getenv("VK_ICD_FILENAMES");
-    if (icd == nullptr) {
-        const Path jsonPath = "/usr/local/share/vulkan/icd.d/MoltenVK_icd.json";
-        setenv("VK_ICD_FILENAMES", jsonPath.c_str(), 1);
-    }
+#ifdef IOS
+    const char* dylibPath = "Frameworks/libMoltenVK.dylib";
```
bejado, line 30: We can probably get rid of this, since we don't support MoltenVK on iOS anymore

bejado (approved):

pixelflinger (approved):

PR #5704, opened 2022-06-16
basic automatic-instancing

pixelflinger:


pixelflinger:
reviewers: might be easier to look at each change separately. thanks.

libs/viewer/include/viewer/Settings.h
```diff
@@ -213,6 +213,7 @@ struct ViewerOptions {
     float cameraFocalLength = 28.0f;
     float cameraFocusDistance = 10.0f;
     bool autoScaleEnabled = true;
+    bool autoInstancingEnabled = true;
```
prideout, line 216: The doesn't match its Java corollary and should probably default to "false" since it doesn't work with morphing / skinning.

prideout (changes requested):
Nice!  See my comment wrt the default value in `ViewerOptions`.

Also, in debug builds, is there a way we could assert that morphing & skinning are disabled?

filament/include/filament/Engine.h
```diff
@@ -335,6 +335,28 @@ class UTILS_PUBLIC Engine {
      */
     void enableAccurateTranslations() noexcept;
 
+    /**
+     * Enables or disables automatic instancing of render primitives. Instancing of render primitive
```
bejado, line 339:
```suggestion
     * Enables or disables automatic instancing of render primitives. Instancing of render primitives
```

filament/src/RenderPass.cpp
```diff
@@ -186,8 +185,126 @@ void RenderPass::sortCommands() noexcept {
...
+
+    if (mEngine.isAutomaticInstancingEnabled()) {
+        instanceify();
+    }
+}
+
+void RenderPass::instanceify() noexcept {
+    SYSTRACE_NAME("instanceify");
+
+    uint32_t instanceCount_debug = 0;
```
bejado, line 197: `UTILS_UNUSED_IN_RELEASE`?

libs/utils/include/utils/algorithm.h
```diff
@@ -203,6 +203,19 @@ bit_cast(const From &src) noexcept {
     return reinterpret_cast<const To&>(src);
 }
 
+template<typename T, typename Pred>
+T const* find_range(T const* const begin, T const* end, Pred pred) noexcept {
```
bejado, line 207: The name `find_range` is confusing to me, since this returns a single item. Maybe `find_first_nonequal`?

Also, I think you can make `end` also be `T const* const end`.

shaders/src/getters.vs
```diff
@@ -18,7 +18,11 @@ mat4 getLightFromWorldMatrix() {
 }
 
 PerRenderableData getObjectUniforms() {
+#if defined(MATERIAL_HAS_INSTANCES)
```
bejado, line 21: This probably warrants a comment (and getters.fs) since it's pretty counter intuitive :)

filament/src/RenderPass.cpp
```diff
@@ -639,11 +756,23 @@ void RenderPass::Executor::recordDriverCommands(FEngine& engine, backend::Driver
 
             pipeline.program = ma->getProgram(info.materialVariant);
 
-            if (uboDataIndex != info.index) {
+//            if (uboDataIndex != info.index)
```
bejado, line 759: Remove?

pixelflinger:
> Nice!  See my comment wrt the default value in `ViewerOptions`.
> 
> Also, in debug builds, is there a way we could assert that morphing & skinning are disabled?

It doesn't have to be disabled, it just won't instance that primitive unless skinning/morphing are the same.

libs/viewer/include/viewer/Settings.h
pixelflinger: ok, will set to false.

filament/include/filament/Engine.h
pixelflinger: will fix.

filament/src/RenderPass.cpp
pixelflinger: good catch.

libs/utils/include/utils/algorithm.h
pixelflinger: very good point. In fact we can use `std::find_if_not` instead. thanks!

filament/src/RenderPass.cpp
pixelflinger: yes, in fact this wasn't a good optimization because the backends should already do this (at least the GLES one does), and the backend knows better if this operation is heavy or not (it shouldn't).

pixelflinger:
note: this is causing the vulkan backend to fail.

PR #5705, opened 2022-06-16
smolv blobs are now 8-byte aligned in filamat. (breaks materials)

prideout:


libs/filamat/src/eiff/Flattener.h
```diff
@@ -125,6 +125,18 @@ class Flattener {
         mCursor += 4;
     }
 
+    // This first writes "n" (which is a single byte), followed by "n" zeroes.
+    // The subsequent write is guaranteed to be on a 8-byte boundary.
+    void writeAlignmentPadding() {
```
bejado, line 130: It seems even if we're already on an 8-byte boundary, we still have to write 8 bytes of padding (a 7 plus 7 zeroes). Maybe the reader could be slightly more intelligent and automatically align reads to 8 bytes to avoid the wasted space.

prideout: That's brilliant!  Added a commit.

libs/filamat/src/eiff/Flattener.h
```diff
@@ -125,6 +125,20 @@ class Flattener {
...
+    // 8-byte boundary. Note that the reader must perform a similar calculation to figure out
+    // how many bytes to skip.
+    void writeAlignmentPadding() {
+        const intptr_t offset = mCursor - mStart;
+        const uint8_t padSize = 8 - (offset % 8);
+        if (padSize != 8) {
+            for (uint8_t i = 0; i < padSize; i++) {
+                writeUint8(0);
+            }
+        }
```
pixelflinger, line 138:
```suggestion
        const uint8_t padSize = (8 - (offset % 8)) % 8;
        for (uint8_t i = 0; i < padSize; i++) {
            writeUint8(0);
        }
```

pixelflinger (approved):

libs/filamat/src/eiff/Flattener.h
prideout: Ah nice one.  Done.

PR #5715, opened 2022-06-17
Quick repair to matdbg for alignment padding.

prideout:
This needs to be refactored for a proper fix that shares code
with the flattener / unflattener pipeline.


prideout:
Closing without merging, this isn't quite right.

PR #5718, opened 2022-06-17
implement automatic instancing of render primitives

pixelflinger:


prideout (approved):

samples/image_viewer.cpp
```diff
@@ -292,6 +292,7 @@ int main(int argc, char** argv) {
         app.engine = engine;
         app.viewer = new ViewerGui(engine, scene, view, 410);
         app.viewer->getSettings().viewer.autoScaleEnabled = false;
+        app.viewer->getSettings().viewer.autoInstancingEnabled = true;
```
prideout, line 295: @pixelflinger was this line really intended for `image_viewer`? doesn't this app draw like 1 quad?

PR #5734, opened 2022-06-23
Check the font path is the file

daemyung:
If the font path is the directory then it's undefined behavior.


daemyung:
It's internal changes. Should I update RELEASE_NOTES.md?

PR #5739, opened 2022-06-27
matc: add --template option.

prideout:
This makes it easier to generate ubershaders without the fancy CMake machinery available in its `configure_file` command.

### DIFF
--- a/tools/matc/src/matc/CommandlineConfig.cpp
+++ b/tools/matc/src/matc/CommandlineConfig.cpp
@@ -59,6 +59,11 @@ static void usage(char* name) {
             "       Add a preprocessor define macro via <macro>=<value>. <value> defaults to 1 if omitted.\n"
             "       Can be repeated to specify multiple definitions:\n"
             "           MATC -Dfoo=1 -Dbar -Dbuzz=100 ...\n\n"
+            "   --template <macro>=<string>, -T<macro>=<string>\n"
+            "       Replaces ${MACRO} with specified string before parsing\n"
+            "       Unlike --define, this applies to the material specification, not GLSL.\n"
+            "       Can be repeated to specify multiple macros:\n"
+            "           MATC -TBLENDING=fade -TDOUBLESIDED=false ...\n\n"
             "   --reflect, -r\n"
             "       Reflect the specified metadata as JSON: parameters\n\n"
             "   --variant-filter=<filter>, -V <filter>\n"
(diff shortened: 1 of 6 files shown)


tools/matc/src/matc/CommandlineConfig.h
```diff
@@ -74,12 +74,12 @@ class FilesystemInput : public Config::Input {
...
     std::unique_ptr<const char[]> read() noexcept override {
-        auto mMaterialBuffer = std::unique_ptr<const char[]>(new char[mFilesize]);
-        if (!mFile.read(const_cast<char*>(mMaterialBuffer.get()), mFilesize)) {
+        auto buffer = std::make_unique<char[]>(mFilesize);
+        if (!mFile.read(buffer.get(), mFilesize)) {
             std::cerr << "Unable to read material source file '" << mPath << "'" << std::endl;
             return nullptr;
         }
-         return mMaterialBuffer;
+        return std::unique_ptr<const char[]>(buffer.release());
```
bejado, line 82: Why do we need to call `buffer.release()` here?

tools/matc/src/matc/Config.h
```diff
@@ -41,6 +41,10 @@ class Config {
     using TargetApi = filamat::MaterialBuilder::TargetApi;
     using Optimization = filamat::MaterialBuilder::Optimization;
 
+    // For defines and template args, we use an ordered map with a transparent comparator
+    // to enable efficient lookups using string_view keys.
+    using StringViewMap = std::map<std::string, std::string, std::less<>>;
```
bejado, line 46: I'm a bit confused by the name/comment since we're not using `std::string_view` as the key. Maybe rename to `StringSubstitutionMap` or something?

prideout: Good point, I renamed the type and rewrote the comment.

tools/matc/src/matc/CommandlineConfig.cpp
```diff
@@ -59,6 +59,10 @@ static void usage(char* name) {
             "       Add a preprocessor define macro via <macro>=<value>. <value> defaults to 1 if omitted.\n"
             "       Can be repeated to specify multiple definitions:\n"
             "           MATC -Dfoo=1 -Dbar -Dbuzz=100 ...\n\n"
+            "   --template <macro>=<string>, -T<macro>=<string>\n"
+            "       Replaces ${MACRO} with specified string before parsing\n"
+            "       Can be repeated to specify multiple macros:\n"
+            "           MATC -TBLENDING=fade -TDOUBLESIDED=false ...\n\n"
```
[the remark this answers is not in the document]
prideout: Done.

tools/matc/src/matc/CommandlineConfig.h
bejado: Ah got it, thanks!

PR #5748, opened 2022-06-29
ImGuiHelper: add support for Y flip.

prideout:
Reflects a change from Betty and should be cherry picked to v1.23.3

Users could customize the ImGuiHelper camera, but they had no control
over the scissor coordinates. This allows them to use vertically flipped
coordinates, which, unfortunately, is required for MediaPipe
integration.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -11,6 +11,7 @@ A new header is inserted each time a *tag* is created.
 
 ## v1.23.3
 
+- ImGuiHelper: add support for Y flip.
 - materials: add a new `instanced` material parameter that is now mandatory in order to call `getInstanceIndex()`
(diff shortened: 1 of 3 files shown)


RELEASE_NOTES.md
```diff
@@ -11,6 +11,7 @@ A new header is inserted each time a *tag* is created.
 
 ## v1.23.3
 
+- ImGuiHelper: add support for Y flip.
```
bejado, line 14: Are you cherry-picking this into v1.23.3? If not, this should go under `main branch`.

bejado (approved):

RELEASE_NOTES.md
prideout: Yes I intend to do a cherry-pick to v1.23.3.

bejado: Sounds good!

Issue #5667, opened 2022-06-10
What is the minimum android version that filament supports？


caorm:
I also have the same crash：

https://github.com/google/filament/issues/5426

the official model in the apk is fine，the crash info for sample-gltf-viewer which i replace my model.

thank you！

romainguy:
Filament supports Android 21+.

pixelflinger:
duplicate of #5426

caorm:
> Filament supports Android 21+.

hello

the crash info for the official sample-gltf-viewer demo

vivo X9s Plus
OS: Android 7.1.1
Pixel: 1920x1080
Display: 5.87

2022/6/13 14:32:4914577F*** *** *** *** *** *** *** *** *** *** *** *** *** *** *** ***
2022/6/13 14:32:4914577FBuild fingerprint: 'vivo/PD1635/PD1635:7.1.1/NMF26F/compil10251828:user/release-keys'
2022/6/13 14:32:4914577FRevision: '0'
2022/6/13 14:32:4914577FABI: 'arm64
'2022/6/13 14:32:4914577Fpid: 14527, tid: 14548, name: FEngine::loop >>> com.google.android.filament.gltf <<<
2022/6/13 14:32:49 14577 F signal 11 (SIGSEGV), code 1 (SEGV_MAPERR), fault addr 0x44
2022/6/13 14:32:49 14577 F x0 0000000000000bd0 x1 0000007f68ddeb40 x2 0000000000000001 x3 0000007f6411f000
2022/6/13 14:32:49 14577 F x4 0000000000000000 x5 0000007f77433000 x6 0000000000000000 x7 0000000000000004
2022/6/13 14:32:49 14577 F x8 0000000000000000 x9 0000000000000004 x10 0000000000000001 x11 0000007f64bb0a50
2022/6/13 14:32:49 14577 F x12 0000000000000000 x13 0000007f557b96a0 x14 0000000000000000 x15 0000000000000020
2022/6/13 14:32:49 14577 F x16 0000000000000030 x17 0000000000000038 x18 000000000000ffff x19 0000000000000010
2022/6/13 14:32:49 14577 F x20 0000007f77901800 x21 0000000000000001 x22 0000000000000000 x23 0000000000000000
2022/6/13 14:32:49 14577 F x24 0000007f68ddeb40 x25 0000000000000000 x26 0000000000000000 x27 0000000000000847
[...]

Issue #5426, opened 2022-04-13
crash on Android 8.1.0


August1996:
**Describe the bug**
some devices are works, but some not.

**crash info**
[bugreport files](https://github.com/August1996/filament-bugreport/blob/main/bugreport-PD1709-OPM1.171019.011-2022-04-13-14-21-14.zip)

crash detail info refer to file `tombstone_00`

**device info**
https://www.tme.net/device/vivo-x20a/

Android 8.1.0

August1996:
there are some logs, crash happens when i rendering complex glb model.
```
2022-04-13 17:10:27.546 28266-28350/com.tme.k3d W/Adreno-GSL: <ioctl_kgsl_syncobj_create:3977>: (37, 19, 4233) fail 22 Invalid argument
2022-04-13 17:10:27.561 28266-28350/com.tme.k3d W/Adreno-GSL: <gsl_ldd_control:541>: ioctl fd 54 code 0xc0200933 (IOCTL_KGSL_TIMESTAMP_EVENT) failed: errno 22 Invalid argument
2022-04-13 17:10:27.561 28266-28350/com.tme.k3d W/Adreno-GSL: <ioctl_kgsl_syncobj_create:3977>: (37, 19, 4233) fail 22 Invalid argument
2022-04-13 17:10:27.564 28266-28350/com.tme.k3d W/Adreno-GSL: <gsl_ldd_control:541>: ioctl fd 54 code 0xc0200933 (IOCTL_KGSL_TIMESTAMP_EVENT) failed: errno 22 Invalid argument
2022-04-13 17:10:27.564 28266-28350/com.tme.k3d W/Adreno-GSL: <ioctl_kgsl_syncobj_create:3977>: (37, 19, 4233) fail 22 Invalid argument
```

August1996:
i found that most of devices under api level 29 will carsh.

August1996:
i have try many versions, and found that the problem first appears on v1.18.0.

pixelflinger:
As you can see in the tombstone, this is not a crash in filament. It's a crash in libhwui:

```
    #00 pc 000000000001e728  /system/lib64/libc.so (abort+120)
    #01 pc 0000000000008348  /system/lib64/liblog.so (__android_log_assert+296)
    #02 pc 00000000000469f8  /system/lib64/libhwui.so (android::uirenderer::debug::GlesErrorCheckWrapper::assertNoErrors(char const*)+384)
[...]

PR #5754, opened 2022-07-02
fix adreno driver crash related to morph target change.

hzzengxiaoqi:
For some android gpu drivers, Adreno (TM) 506 in my case, some uniform arrays should be iniatialized to be used in the shader. Even when the actual shader code would not be called due to the condition. I used Opengl backend by the way.

This kind of initialization wouldn't cause any trouble in other platforms, so I fix it without any platform related macros.
These issues was introduced along with changes for GPU Morph Target. I think it should be introduced after version 1.17 or 1.18. I met them in 1.19.
I triggered these issues using two different gltf models.
1. One model is without morph target but with bones, so the morph weights("morphingUniforms") is used in the shader of that variant(key is '\t', with DIR and SKN bits set). The morph target number is zero, so the code reading morphingUniforms should not run, but on Adreno (TM) 506, the uniniatialized morph weights uniform caused crash. This is why I change the code in AssetLoader.cpp.
2. Another model is the official gltf model used in android sample project sample-gltf-viewer. It's the animated BusterDrone model, path of it in project:"third_party/models/BusterDrone". There's a mesh with one primitive, whose name is "1" and its parents named "Drone_IEye". The primitive have no bones, but it has morph targets which makes the material variant to be '\t' (same as the first case, with DIR and SKN set). So in this case, the bone uniforms appears in the shader. Although it should not be called due to the condition(flag of bone is not set), it crashed the drawing due to its existence. Initialize the bone uniforms beforehand would get rid of this crash. This is why I change the code in RenderableManager.cpp.

The crash stack is the same as [5667](https://github.com/google/filament/issues/5667) and [5426](https://github.com/google/filament/issues/5426)

The crash may ultimately caused by some driver issue, but my code may be a good workaround to fix it, without any side effects.


libs/gltfio/src/AssetLoader.cpp
```diff
@@ -605,6 +605,12 @@ void FAssetLoader::createRenderable(const cgltf_data* srcAsset, const cgltf_node
         }
         mRenderableManager.setMorphWeights(renderable, weights.data(), size);
     }
+    else
+    {
+        const float initWeights[1]={0};
+        RenderableManager::Instance renderable = mRenderableManager.getInstance(entity);
+        mRenderableManager.setMorphWeights(renderable, initWeights, 1);
+    }
```
pixelflinger, line 613: Is that part of the change absolutely needed too? I really don't like that it's in the client code -- that means that we need to *document* this behavior.

Is there a way to do this inside the filament library instead?

pixelflinger (commented):
Thank you so much for figuring this one out! 👏 🥇 

I have two big requests before for this fix though:
1. use the workaround facility in the backend api 
2. the fix/workaround should be entirely inside the filament lib, it can't be in gltfio, otherwise it becomes a public API we need to document.

thanks again!

prideout (changes requested):
Wow great work debugging this!  Personally I'm okay with this "fix", but only if @pixelflinger is okay with it too.  One minor note: our coding style is `} else {` without newlines.

RELEASE_NOTES.md
```diff
@@ -20,6 +20,8 @@ A new header is inserted each time a *tag* is created.
 - gltfio: UbershaderProvider now takes the ubershader archive in its constructor [⚠️ **API Change**]
 - gltfio: Fix morphing with sparse accessors.
 - gltfio: Fix models that use signed integers for morphing.
+- gltfio: Fix adreno gpu crash introduced from gpu morph target change
```
bejado, line 23: Please put the release note under the `main branch` heading. Also, we can stick with a single release note with the `engine` tag.

bejado (changes requested):

libs/gltfio/src/AssetLoader.cpp
hzzengxiaoqi: Hi, @pixelflinger , thanks for the review and the recommendation.

1. The morphWeights uniform array is also needed to be initialized to get rid of this crash on some Adreno gpu phones. Any models with mesh downloaded from mixamo website would trigger this crash, since there's no morph target, but there's bone skinning. For example, use any model with mesh from this link(in my case, my phone is with Adreno (TM) 506):
https://www.mixamo.com/#/?page=1&query=danc

I moved the code change to a suitable place in RenderableManager.cpp, I tested it and it still works.

2. I added a workaround just as you said. Since the renderer is already recorded in mContext.state, I added a get method to get it. I also added some short comments in the code.

RELEASE_NOTES.md
hzzengxiaoqi: Hi, @bejado , thanks for the review and the recommendation.
I moved the note to main branch, and since now there's only change in engine, I removed gltfio line.

filament/backend/src/opengl/OpenGLContext.h
```diff
@@ -720,6 +722,11 @@ GLuint OpenGLContext::getQuery(GLenum target) const noexcept {
             return 0;
     }
 }
+
+const char* OpenGLContext::getRenderer() const noexcept {
+    return state.renderer;
+}
+
```
pixelflinger, line 729: this is not needed, see below.

filament/backend/src/opengl/OpenGLContext.h
```diff
@@ -109,6 +109,8 @@ class OpenGLContext {
     inline void endQuery(GLenum target) noexcept;
     inline GLuint getQuery(GLenum target) const noexcept;
 
+    inline const char* getRenderer() const noexcept;
+
```
pixelflinger, line 113: not needed, see below

filament/backend/src/opengl/OpenGLDriver.cpp
```diff
@@ -1581,6 +1581,9 @@ bool OpenGLDriver::isWorkaroundNeeded(Workaround workaround) {
             return mContext.bugs.split_easu;
         case Workaround::ALLOW_READ_ONLY_ANCILLARY_FEEDBACK_LOOP:
             return mContext.bugs.allow_read_only_ancillary_feedback_loop;
+        case Workaround::ADRENO_UNIFORM_ARRAY_CRASH:
+            return (strstr(
+            mContext.getRenderer(), "Adreno") != NULL);
```
pixelflinger, line 1586: please implement this like the ones above, by creating a field in `mContext.bugs` and initializing it during construction.

filament/src/components/RenderableManager.cpp
```diff
@@ -404,6 +404,15 @@ void FRenderableManager::create(
                                 out, boneCount * sizeof(PerRenderableBoneUib::BoneData) }, 0);
                     }
                 }
+                else{
```
pixelflinger, line 407:
```suggestion
                } else {
```

filament/src/components/RenderableManager.cpp
```diff
@@ -436,6 +445,12 @@ void FRenderableManager::create(
                 morphTargets[i] = { upcast(morphing.buffer), (uint32_t)morphing.offset,
                                     (uint32_t)morphing.count };
             }
+            
+            // When targetCount equal 0, boneCount>0 in this case, do an initialization for the morphWeights uniform array to avoid crash on adreno gpu.
+            if (targetCount == 0 && driver.isWorkaroundNeeded(Workaround::ADRENO_UNIFORM_ARRAY_CRASH)) {
```
pixelflinger, line 450:
```suggestion
            if (UTILS_UNLIKELY(targetCount == 0 && driver.isWorkaroundNeeded(Workaround::ADRENO_UNIFORM_ARRAY_CRASH))) {
```

filament/src/components/RenderableManager.cpp
```diff
@@ -404,6 +404,15 @@ void FRenderableManager::create(
                                 out, boneCount * sizeof(PerRenderableBoneUib::BoneData) }, 0);
                     }
                 }
+                else{
+                    // When boneCount is 0, do an initialization for the bones uniform array to avoid crash on adreno gpu.
+                    if (driver.isWorkaroundNeeded(Workaround::ADRENO_UNIFORM_ARRAY_CRASH)) {
```
pixelflinger, line 409:
```suggestion
                    if (UTILS_UNLIKELY(driver.isWorkaroundNeeded(Workaround::ADRENO_UNIFORM_ARRAY_CRASH))) {
```

pixelflinger (changes requested):
Thanks for the changes, we just need to implement the workaround like the  other workarounds (see inline comments).

filament/backend/src/opengl/OpenGLContext.cpp
```diff
@@ -178,6 +181,9 @@ OpenGLContext::OpenGLContext() noexcept {
             // early exit condition is flattened in EASU code
             // (that should be regardless of ANGLE, but we should double check)
             bugs.split_easu = true;
+
+            // double check like split_easu above
+            bugs.enable_initialize_non_used_uniform_array = true;
```
pixelflinger, line 186: I think this should be `false` here, because this bug doesn't exist when using the Vulkan backend (Angle is using Vulkan).

pixelflinger, line 186:
```suggestion
```

filament/backend/src/vulkan/VulkanDriver.cpp
```diff
@@ -874,6 +874,8 @@ bool VulkanDriver::isWorkaroundNeeded(Workaround workaround) {
             return deviceProperties.vendorID == 0x5143; // Qualcomm
         case Workaround::ALLOW_READ_ONLY_ANCILLARY_FEEDBACK_LOOP:
             return true;
+        case Workaround::ADRENO_UNIFORM_ARRAY_CRASH:
+            return false;
```
pixelflinger, line 878: We need this in the Metal backend as well.

pixelflinger (changes requested):
Almost there! Thank you.

filament/backend/src/vulkan/VulkanDriver.cpp
hzzengxiaoqi: Hello, pixelflinger, I think I've added the same code in MetalDriver.mm.

dandingol03:
grea job

PR #5764, opened 2022-07-07
Camera API and documentation improvements

pixelflinger:
- getNear() and getCullingFar() now return doubles
- updated documentation
- all setProjection() calls can now throw (when enabled) and will
  do so if preconditions are not met (instead of setting a default
  projection).
- Frustum can now be logged on debug builds

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -4,6 +4,8 @@ This file contains one line summaries of commits that are worthy of mentioning i
 A new header is inserted each time a *tag* is created.
 
 ## main branch
+ 
+- engine: `Camera::getNear()` and `Camera::getCullingFar()` now return `doubles`
 
 ## v1.25.1
 
--- a/filament/src/Frustum.cpp
+++ b/filament/src/Frustum.cpp
@@ -19,6 +19,7 @@
 #include "Culler.h"
 
 #include <utils/compiler.h>
+#include <utils/Log.h>
 
 using namespace filament::math;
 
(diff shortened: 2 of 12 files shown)


filament/src/Frustum.cpp
```diff
@@ -98,3 +99,19 @@ float Frustum::contains(float3 p) const noexcept {
 }
 
 } // namespace filament
+
+#if !defined(NDEBUG)
+
+utils::io::ostream& operator<<(utils::io::ostream& out, filament::Frustum const& frustum) {
```
prideout, line 105: 😍

libs/math/include/math/mat4.h
```diff
@@ -491,11 +491,22 @@ template<typename T>
...
+    // 0
+    // 0
+
     m[2][0] = (right + left) / (right - left);
     m[2][1] = (top + bottom) / (top - bottom);
     m[2][2] = -(far + near) / (far - near);
     m[2][3] = -1;
+
+    // 0
+    // 0
```
prideout, line 509: Intentional?

prideout (changes requested):
Nice.  I think the `mat4.h` change might be accidental?

RELEASE_NOTES.md
```diff
@@ -8,6 +8,7 @@ A new header is inserted each time a *tag* is created.
 ## v1.25.1
 
 - engine: add support for automatic instancing. Must be enabled with `Engine::setAutomaticInstancingEnabled(bool)`
+- engine: `Camera::getNear()` and `Camera::getCullingFar()` now return `doubles`
```
bejado, line 11: Please update and put under `## main branch` heading

bejado (approved):

RELEASE_NOTES.md
pixelflinger: done

PR #5769, opened 2022-07-08
filamat: dictionaries now store each blob once, not twice. 

prideout:
BlobDictionary and LineDictionary were storing blobs as map keys to achieve simple compression, but they also stored duplicates in a vector for index-based lookup.

Now, the vector is storage and the map has keys that are string_view.

### DIFF
--- a/libs/filamat/src/eiff/BlobDictionary.h
+++ b/libs/filamat/src/eiff/BlobDictionary.h
@@ -17,9 +17,11 @@
 #ifndef TNT_FILAMAT_BLOBDICTIONARY_H
 #define TNT_FILAMAT_BLOBDICTIONARY_H
 
-#include <cassert>
+#include <memory>
 #include <string>
-#include <map>
+#include <string_view>
+#include <memory>
+#include <unordered_map>
 #include <vector>
 
 namespace filamat {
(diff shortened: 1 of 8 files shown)


bejado (approved):

libs/filamat/src/eiff/BlobDictionary.h
```diff
@@ -39,27 +42,17 @@ class BlobDictionary {
...
 
 private:
-
-    // Use an ordered map with a transparent comparator to allow lookups using
-    // std::string_view without need for construction of a std::string object.
-    std::map<std::string, size_t, std::less<>> mBlobIndices;
-    std::vector<std::string> mBlobs;
-    size_t mStorageSize;
+    std::unordered_map<std::string_view, size_t> mBlobIndices;
+    std::vector<std::unique_ptr<std::string>> mBlobs;
```
pixelflinger, line 55: is the unique_ptr needed here? the std::string_view is not a reference to the std::string, it's a reference to the std::string's data.

I'm assuming a unique_ptr is used here so that the string_view stay valid when mBlobs is modified, if that's the case, I don't think a heap allocation is needed.

pixelflinger (changes requested):
not technically asking for a change, just want to understand why a unique_ptr are needed.

libs/filamat/src/eiff/BlobDictionary.h
prideout: The blobs vector grows over time, which causes strings to be copied (not moved) during a growth event, and the original string data is freed.  I actually wasn't sure those wrappers were needed either, but I saw use-after-delete errors without them.

I think it must be possible to write a vector implementation that doesn't do this, disappointed by this STL behavior.

pixelflinger: ah right -- I guess if it worked like I said, it would prevent the small string optimization. never mind then!

pixelflinger (approved):

PR #5782, opened 2022-07-12
more Engine configuration cleanup

pixelflinger:
- add configuration for irradiance map to Config
  (currently, still unused)

- move all froxel configuration constants out of Engine.h, unlike
  the previous todo/comment, these shouldn't be part of Engine::Config.

### DIFF
--- a/filament/include/filament/Engine.h
+++ b/filament/include/filament/Engine.h
@@ -227,6 +227,7 @@ public:
          */
         uint32_t minCommandBufferSizeMB = FILAMENT_MIN_COMMAND_BUFFERS_SIZE_IN_MB;
 
+
         /**
          * Size in MiB of the per-frame high level command buffer.
          *
--- a/filament/src/Froxelizer.cpp
+++ b/filament/src/Froxelizer.cpp
@@ -44,6 +44,11 @@ namespace filament {
 
 using namespace backend;
 
+// TODO: these should come from a configuration object on View or Camera
+static constexpr size_t FROXEL_SLICE_COUNT = 16;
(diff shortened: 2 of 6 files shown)


prideout (approved):

filament/include/filament/Engine.h
```diff
@@ -239,6 +240,13 @@ class UTILS_PUBLIC Engine {
          * This value does not affect the application's memory usage.
          */
         uint32_t perFrameCommandsSizeMB = FILAMENT_PER_FRAME_COMMANDS_SIZE_IN_MB;
+
+        /**
+         * Whether to use an irradiance map for the environment lighting.
+         *
+         * NOTE: This is not currently supported. Always set to false.
+         */
+        bool useIrradianceMap = false;
```
bejado, line 249: I'm confused why this belongs under `Engine::Config`?

bejado: Sounds good. Seems like this belongs under `IndirectLight` or something, but maybe I'm not understanding its purpose.

pixelflinger: It's a precomputed texture for the diffuser of the ibl. Currently, we use the roughness=1 of the prefiltered cubemap, but this is technically both incorrect and low quality (which only matters for certain scenes).

To use the irradiance map, we'd need an extra sampler, first of all, but also, we'd need to either compute it from roughness=0, or, get it from the user (and we don't have an API for that). Precomputing it at runtime, would take a few hundred ms probably.

So it's one of those feature that we should really have, but is not critical.

PR #5783, opened 2022-07-13
Metal: add initial support for stencil buffers

bejado:
Still to come: MSAA support.

### DIFF
--- a/filament/backend/src/metal/MetalDriver.mm
+++ b/filament/backend/src/metal/MetalDriver.mm
@@ -336,6 +336,8 @@ void MetalDriver::createRenderTargetR(Handle<HwRenderTarget> rth,
         stencilTexture->updateLodRange(stencil.level);
         stencilAttachment = { stencilTexture, stencil.level, stencil.layer };
     }
+    ASSERT_POSTCONDITION(!stencil.handle || any(targetBufferFlags & TargetBufferFlags::STENCIL),
+            "The STENCIL flag was specified, but no stencil texture provided.");
 
     construct_handle<MetalRenderTarget>(rth, mContext, width, height, samples,
             colorAttachments, depthAttachment, stencilAttachment);
--- a/filament/backend/src/metal/MetalDriver.mm
+++ b/filament/backend/src/metal/MetalDriver.mm
@@ -328,13 +328,17 @@ void MetalDriver::createRenderTargetR(Handle<HwRenderTarget> rth,
     ASSERT_POSTCONDITION(!depth.handle || any(targetBufferFlags & TargetBufferFlags::DEPTH),
             "The DEPTH flag was specified, but no depth texture provided.");
 
-    construct_handle<MetalRenderTarget>(rth, mContext, width, height, samples,
-            colorAttachments, depthAttachment);
+    MetalRenderTarget::Attachment stencilAttachment = {};
+    if (stencil.handle) {
+        auto stencilTexture = handle_cast<MetalTexture>(stencil.handle);
+        ASSERT_PRECONDITION(stencilTexture->texture,
+                "Stencil texture passed to render target has no texture allocation.");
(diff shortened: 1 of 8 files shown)


prideout (approved):

filament/backend/src/metal/MetalDriver.mm
```diff
@@ -328,13 +328,17 @@
     ASSERT_POSTCONDITION(!depth.handle || any(targetBufferFlags & TargetBufferFlags::DEPTH),
             "The DEPTH flag was specified, but no depth texture provided.");
 
-    construct_handle<MetalRenderTarget>(rth, mContext, width, height, samples,
-            colorAttachments, depthAttachment);
+    MetalRenderTarget::Attachment stencilAttachment = {};
+    if (stencil.handle) {
```
pixelflinger, line 332: I think it's not enough to test `stencil.handle` here, you also need to test `targetBufferFlags`.

filament/backend/src/metal/MetalState.h
```diff
@@ -256,13 +258,23 @@ using PipelineStateCache = StateCache<MetalPipelineState, id<MTLRenderPipelineSt
...
-    char padding[7] = { 0 };                                            // 7 bytes
-
-    bool operator==(const DepthStencilState& rhs) const noexcept {
-        return this->compareFunction == rhs.compareFunction &&
-               this->depthWriteEnabled == rhs.depthWriteEnabled;
+    MTLCompareFunction depthCompare = MTLCompareFunctionAlways;                         // 8 bytes
+    MTLCompareFunction stencilCompare = MTLCompareFunctionAlways;                       // 8 bytes
+    MTLStencilOperation stencilOperationStencilFail = MTLStencilOperationKeep;          // 8 bytes
+    MTLStencilOperation stencilOperationDepthFail = MTLStencilOperationKeep;            // 8 bytes
+    MTLStencilOperation stencilOperationDepthStencilPass = MTLStencilOperationKeep;     // 8 bytes
```
pixelflinger, line 265: wow these are crazy expensive -- 8 bytes per constant!?!
might be worse encoding them with our internal constants and have getters.

bejado: Yeah :( Metal encodes them as an 8 byte unsigned long, even though there's only a handful of enumeration values. It would probably be safe to simply store them as `uint8_t`s.

filament/backend/src/metal/MetalHandles.mm
```diff
@@ -985,6 +1004,14 @@ BufferDescriptor b(nullptr, 0u, [](void* buffer, size_t size, void* user) {
     return result;
 }
 
+MetalRenderTarget::Attachment MetalRenderTarget::getStencilAttachment() {
+    Attachment result = stencil;
+    if (defaultRenderTarget) {
+        // TODO: do we want the default SwapChain to have a default stencil buffer?
```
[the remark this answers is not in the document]
bejado: I'll leave this as a TODO for now, but I'll think about this some more.

filament/backend/src/metal/MetalHandles.mm
```diff
@@ -870,6 +871,12 @@ BufferDescriptor b(nullptr, 0u, [](void* buffer, size_t size, void* user) {
         }
     }
 
+    if (stencilAttachment) {
+        stencil = stencilAttachment;
+
```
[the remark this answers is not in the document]
bejado: Yes good catch.

filament/backend/src/metal/MetalDriver.mm
bejado: Is it valid for `stencil.handle` to be a valid handle but `targetBufferFlags` not have the stencil bit set?

pixelflinger: I would say yes. It's really `targetBufferFlags` that tells you if you should use it or not.

PR #5784, opened 2022-07-13
new feature level API for backends

pixelflinger:
backend can now return a "feature level", each level corresponds to a
"bundle" of features.
Level1: ES3.0 capabilities
Level2: ES3.1 capabilities + 31 textures + cubemap arrays

Currently metal always returns level 1, GL and Vulkan return level 2
if 31 textures or more are supported.

### DIFF
--- a/android/filament-android/src/main/cpp/Engine.cpp
+++ b/android/filament-android/src/main/cpp/Engine.cpp
@@ -328,3 +328,24 @@ Java_com_google_android_filament_Engine_nIsAutomaticInstancingEnabled(JNIEnv*, j
     Engine* engine = (Engine*) nativeEngine;
     return (jboolean)engine->isAutomaticInstancingEnabled();
 }
+
+extern "C" JNIEXPORT jint JNICALL
+Java_com_google_android_filament_Engine_nGetSupportedFeatureLevel(JNIEnv *, jclass,
+        jlong nativeEngine) {
+    Engine* engine = (Engine*) nativeEngine;
+    return (jint)engine->getSupportedFeatureLevel();
+}
+
+extern "C" JNIEXPORT jint JNICALL
+Java_com_google_android_filament_Engine_nSetActiveFeatureLevel(JNIEnv *, jclass,
+        jlong nativeEngine, jint ordinal) {
+    Engine* engine = (Engine*) nativeEngine;
+    return (jint)engine->setActiveFeatureLevel((Engine::FeatureLevel)ordinal);
+}
+
+extern "C" JNIEXPORT jint JNICALL
+Java_com_google_android_filament_Engine_nGetActiveFeatureLevel(JNIEnv *, jclass,
+        jlong nativeEngine) {
+    Engine* engine = (Engine*) nativeEngine;
+    return (jint)engine->getActiveFeatureLevel();
+}
--- a/android/filament-android/src/main/java/com/google/android/filament/Engine.java
+++ b/android/filament-android/src/main/java/com/google/android/filament/Engine.java
@@ -47,7 +47,7 @@ import com.google.android.filament.proguard.UsedByReflection;
  * <pre>
  * import com.google.android.filament.*
  *
- * Engin engine         = Engine.create();
+ * Engine engine        = Engine.create();
  * SwapChain swapChain  = engine.createSwapChain(nativeWindow);
  * Renderer renderer    = engine.createRenderer();
  * Scene scene          = engine.createScene();
@@ -107,6 +107,7 @@ import com.google.android.filament.proguard.UsedByReflection;
  */
 public class Engine {
     private static final Backend[] sBackendValues = Backend.values();
+    private static final FeatureLevel[] sFeatureLevelValues = FeatureLevel.values();
 
     private long mNativeObject;
 
@@ -140,6 +141,18 @@ public class Engine {
          */
         NOOP,
     }
+
+    /**
+     * Defines the backend's feature levels.
+     */
+    public enum FeatureLevel {
+        /** Reserved, don't use */
+        FEATURE_LEVEL_0,
+        /** OpenGL ES 3.0 features (default) */
+        FEATURE_LEVEL_1,
+        /** OpenGL ES 3.1 features + 31 textures units + cubemap arrays */
+        FEATURE_LEVEL_2
+    };
 
     private Engine(long nativeEngine) {
         mNativeObject = nativeEngine;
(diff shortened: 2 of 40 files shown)


filament/backend/src/vulkan/VulkanDriver.h
```diff
@@ -153,6 +153,7 @@ class VulkanDriver final : public DriverBase {
     VulkanSamplerGroup* mSamplerBindings[VulkanPipelineCache::SAMPLER_BINDING_COUNT] = {};
     VkDebugReportCallbackEXT mDebugCallback = VK_NULL_HANDLE;
     VkDebugUtilsMessengerEXT mDebugMessenger = VK_NULL_HANDLE;
+    FeatureLevel mFeatureLevel = FeatureLevel::FEATURE_LEVEL_1;
```
prideout, line 156: NBD but I would not have bothered adding state here since we already persist the entire device caps and determining the answer to the query is a one liner.

pixelflinger: ok, Engine is going to cache the result anyways. so yeah, I can change that.

pixelflinger: done

filament/backend/include/backend/DriverEnums.h
```diff
@@ -59,6 +59,15 @@ static_assert(MAX_VERTEX_BUFFER_COUNT <= MAX_VERTEX_ATTRIBUTE_COUNT,
 
 static constexpr size_t CONFIG_BINDING_COUNT = 12;  // This is guaranteed by OpenGL ES.
 
+/**
+ * Defines the backend's feature levels
```
bejado, line 63:
```suggestion
 * Defines the backend's feature levels.
```

filament/backend/src/opengl/OpenGLContext.cpp
```diff
@@ -40,7 +40,7 @@ bool OpenGLContext::queryOpenGLVersion(GLint* major, GLint* minor) noexcept {
 OpenGLContext::OpenGLContext() noexcept {
     state.vao.p = &mDefaultVAO;
 
-    // There query work with all GL/GLES versions!
+    // These query work with all GL/GLES versions!
```
bejado, line 43:
```suggestion
    // These queries work with all GL/GLES versions!
```

bejado (approved):

pixelflinger:
reviewers, I really recommend to look at the CLs one by one.

android/filament-android/src/main/cpp/Engine.cpp
```diff
@@ -328,3 +328,24 @@ Java_com_google_android_filament_Engine_nIsAutomaticInstancingEnabled(JNIEnv*, j
     Engine* engine = (Engine*) nativeEngine;
     return (jboolean)engine->isAutomaticInstancingEnabled();
 }
+
+extern "C" JNIEXPORT jint JNICALL
+Java_com_google_android_filament_Engine_nGetSupportedFeatureLevel(JNIEnv *, jclass,
```
bejado, line 333:
```suggestion
Java_com_google_android_filament_Engine_nGetSupportedFeatureLevel(JNIEnv*, jclass,
```

android/filament-android/src/main/cpp/Engine.cpp
```diff
@@ -328,3 +328,24 @@ Java_com_google_android_filament_Engine_nIsAutomaticInstancingEnabled(JNIEnv*, j
...
+        jlong nativeEngine) {
+    Engine* engine = (Engine*) nativeEngine;
+    return (jint)engine->getSupportedFeatureLevel();
+}
+
+extern "C" JNIEXPORT jint JNICALL
+Java_com_google_android_filament_Engine_nSetActiveFeatureLevel(JNIEnv *, jclass,
```
bejado, line 340:
```suggestion
Java_com_google_android_filament_Engine_nSetActiveFeatureLevel(JNIEnv*, jclass,
```

android/filament-android/src/main/cpp/Engine.cpp
```diff
@@ -328,3 +328,24 @@ Java_com_google_android_filament_Engine_nIsAutomaticInstancingEnabled(JNIEnv*, j
...
+        jlong nativeEngine, jint ordinal) {
+    Engine* engine = (Engine*) nativeEngine;
+    return (jint)engine->setActiveFeatureLevel((Engine::FeatureLevel)ordinal);
+}
+
+extern "C" JNIEXPORT jint JNICALL
+Java_com_google_android_filament_Engine_nGetActiveFeatureLevel(JNIEnv *, jclass,
```
bejado, line 347:
```suggestion
Java_com_google_android_filament_Engine_nGetActiveFeatureLevel(JNIEnv*, jclass,
```

docs/Materials.md.html
```diff
@@ -932,6 +932,36 @@
...
+Type
+:    `number`
+
+Value
+:     An integer value, either 1 or 2. Defaults to 1.
+
+      Feature Level    |        Guaranteed features
+:----------------------|:---------------------------------
+1                      | 9 textures per material
+2                      | 12 textures per material, cubemap arrays, ESSL 3.10
```
bejado, line 946: How did you come up with the 9 and 12 texture limits?

filament/src/components/RenderableManager.cpp
```diff
@@ -253,6 +253,11 @@ RenderableManager::Builder::Result RenderableManager::Builder::build(Engine& eng
             continue;
         }
 
+        // we want a feature level violation to be a hard error (exception if enabled, or crash)
+        ASSERT_PRECONDITION(upcast(engine).hasFeatureLevel(material->getFeatureLevel()),
+                "Material \"%s\" has feature level %u which is not supported this Engine",
```
bejado, line 258: Maybe rephrase and add the current feature level of the engine:
```
Material xyz has feature level X which is not supported. This Engine instance supports feature level Y.
```
Same for `setMaterialInstanceAt`.

filament/src/details/Material.cpp
```diff
@@ -45,32 +45,35 @@ namespace filament {
 using namespace backend;
 
 static MaterialParser* createParser(Backend backend, const void* data, size_t size) {
-    MaterialParser* materialParser = new MaterialParser(backend, data, size);
+    // unique_ptr do we don't leak MaterialParser on failures below
```
bejado, line 48:
```suggestion
    // unique_ptr so we don't leak MaterialParser on failures below
```

bejado:
I think this is probably worth updating release notes?

docs/Materials.md.html
pixelflinger: 9 is what Impress is using now -- they use 9 and we're using 7. 
For feature level 2, I just picked 12 :-) I think we can revisit at some point.

bejado: Okay, sounds good. Seems like if the backend supports 31 then we should be able to expose a lot more to materials.

pixelflinger: Well, we need to keep some for us. And we're already technically "in debt". We could do 15 (users) + 16 (us). I think it's a limit we can raise later anyways, it wouldn't cause incompatibilities.

bejado: 👍

android/filament-android/src/main/java/com/google/android/filament/Engine.java
```diff
@@ -141,6 +142,18 @@ public enum Backend {
         NOOP,
     }
 
+    /**
+     * Defines the backend's feature levels.
+     */
+    public enum FeatureLevel {
+        /** Reserved, don't use */
+        FEATURE_LEVEL_0,
```
prideout, line 150: 😨

prideout (approved):
Feature Level 2 is documented in some places as "12 textures per material" and as "31 textures units" in other places. I get it, but do we need a comment somewhere (SibGenerator.h?) to explain the relationship, or to remind our future selves to update the docs when adding / removing samplers?

PR #5796, opened 2022-07-15
Add support for cubemap arrays in the backends

pixelflinger:
This is not intended to be used yet because we're not currently
checking that cubemap array are actually supported, however, if they
are, they should work.

### DIFF
--- a/filament/backend/src/metal/MetalHandles.mm
+++ b/filament/backend/src/metal/MetalHandles.mm
@@ -433,36 +433,27 @@ MetalTexture::MetalTexture(MetalContext& context, SamplerType target, uint8_t le
 
     const BOOL mipmapped = levels > 1;
     const BOOL multisampled = samples > 1;
-    const BOOL textureArray = target == SamplerType::SAMPLER_2D_ARRAY;
 
 #if defined(IOS)
+    const BOOL textureArray = target == SamplerType::SAMPLER_2D_ARRAY;
     ASSERT_PRECONDITION(!textureArray || !multisampled,
             "iOS does not support multisampled texture arrays.");
 #endif
 
-    const auto get2DTextureType = [](bool isArray, bool isMultisampled) {
-        uint8_t value = 0;
-        if (isMultisampled) {
-            value |= 0b10u;
+    const auto get2DTextureType = [](SamplerType target, bool isMultisampled) {
+        if (!isMultisampled) {
+            return getMetalType(target);
         }
-        if (isArray) {
-            value |= 0b01u;
-        }
-        switch (value) {
-            default:
-            case 0b00:
-                return MTLTextureType2D;
-            case 0b01:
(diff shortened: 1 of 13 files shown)


filament/backend/src/metal/MetalHandles.mm
```diff
@@ -433,46 +433,38 @@ BufferDescriptor b(nullptr, 0u, [](void* buffer, size_t size, void* user) {
...
-                return MTLTextureType2D;
-            case 0b01:
-                return MTLTextureType2DArray;
-            case 0b10:
-                return MTLTextureType2DMultisample;
-            case 0b11:
+    const auto get2DTextureType = [](SamplerType target, bool isMultisampled) {
+        if (!isMultisampled) {
+            return getMetalType(target);
+        } else {
```
bejado, line 445: You can remove the `else` statement here.

filament/backend/src/metal/MetalHandles.mm
```diff
@@ -433,46 +433,38 @@ BufferDescriptor b(nullptr, 0u, [](void* buffer, size_t size, void* user) {
...
+                    return getMetalType(target);
+            }
         }
     };
 
     MTLTextureDescriptor* descriptor;
     switch (target) {
         case SamplerType::SAMPLER_2D:
         case SamplerType::SAMPLER_2D_ARRAY:
+        case SamplerType::SAMPLER_CUBEMAP_ARRAY:
```
bejado, line 464: I think we can put this `case` statement next to the existing `case SamplerType::SAMPLER_CUBEMAP` on line 478 by copying down the `descriptor.arrayLength = depth` line.

filament/backend/src/metal/MetalHandles.mm
pixelflinger: done

filament/backend/src/metal/MetalHandles.mm
pixelflinger: done -- and also I think my version didn't work because the texture type was wrong.

bejado:
LGTM, but iOS is failing because it doesn't have 3.1 headers:

```
/Users/runner/work/filament/filament/filament/backend/src/opengl/OpenGLContext.h:408:14: error: use of undeclared identifier 'GL_TEXTURE_CUBE_MAP_ARRAY'
        case GL_TEXTURE_CUBE_MAP_ARRAY:         return 6;
```

I can help fix.

pixelflinger:
oh... I lost your fix. damn.

bejado:
> oh... I lost your fix. damn.

Here's the commit, you can cherry-pick it to your branch: https://github.com/google/filament/commit/66ab68aeb74dd15229feb43c4a269c06f341515e

bejado:
Took another look at the Metal changes, LGTM.

bejado (approved):

PR #5813, opened 2022-07-25
fix max mipmap levels calculation for 3D textures

pixelflinger:
the depth of the texture wasn't taken into account. In practice this
would restrict the number of mip levels of a 3d texture that would have
a larger dimension in depth.

### DIFF
--- a/filament/src/details/Texture.h
+++ b/filament/src/details/Texture.h
@@ -43,7 +43,6 @@ public:
     size_t getHeight(size_t level = 0) const noexcept;
     size_t getDepth(size_t level = 0) const noexcept;
     size_t getLevelCount() const noexcept { return mLevelCount; }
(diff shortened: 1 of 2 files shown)


filament/src/details/Texture.h
```diff
@@ -100,12 +99,13 @@ class FTexture : public Texture {
 
     // Returns the max number of levels for a texture of given max dimensions
     static inline uint8_t maxLevelCount(uint32_t maxDimension) noexcept {
```
bejado, line 101: Can these `maxLevelCount` functions be made `constexpr`?

bejado (approved):

filament/src/details/Texture.h
pixelflinger: I don't think so because `ilogbf` isn't `constexpr`.

PR #5821, opened 2022-07-26
fix a use after move

pixelflinger:


romainguy (approved):
Where is the user after move fix here? Those look like nice changes but they're all just casts. Was one of the missing casts somehow causing the issue?

PR #5827, opened 2022-07-27
rename `ShaderModel` enums

pixelflinger:
GL_ES_30   becomes MOBILE
GL_CORE_41 becomes DESKTOP

ShaderModel controls which flavor of GLSL (GL or ES) is used both when 
reading and outputting materials.

It's also used to set default quality settings and the default precision
of all fragment shaders.

Technically, Vulkan and Metal don't need this distinction, but the GL
backend does.

### DIFF
--- a/filament/backend/src/metal/MetalDriver.mm
+++ b/filament/backend/src/metal/MetalDriver.mm
@@ -550,9 +550,9 @@ void MetalDriver::terminate() {
 
 ShaderModel MetalDriver::getShaderModel() const noexcept {
 #if defined(IOS)
-    return ShaderModel::GL_ES_30;
+    return ShaderModel::MOBILE;
 #else
-    return ShaderModel::GL_CORE_41;
+    return ShaderModel::DESKTOP;
 #endif
 }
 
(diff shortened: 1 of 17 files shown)


prideout (changes requested):
We cast this enum to an int in `Material::Builder::build` and compare the result against a field that was extracted from the material package.  This implies that we need a material version bump, right?  Alternatively, maybe we could just start the enum at 1 instead of 0?

filament/backend/src/metal/MetalDriver.mm
```diff
@@ -552,7 +552,7 @@
 #if defined(IOS)
     return ShaderModel::GL_ES_30;
```
bejado, line 553: Please update this to `ShaderModel::MOBILE`.

filament/backend/src/noop/NoopDriver.cpp
```diff
@@ -35,7 +35,7 @@ ShaderModel NoopDriver::getShaderModel() const noexcept {
 #if defined(__ANDROID__) || defined(IOS) || defined(__EMSCRIPTEN__)
     return ShaderModel::GL_ES_30;
```
bejado, line 36: Please update this to `ShaderModel::MOBILE`.

bejado (changes requested):

filament/backend/src/metal/MetalDriver.mm
pixelflinger: done

filament/backend/src/noop/NoopDriver.cpp
pixelflinger: done

pixelflinger:
@prideout @bejado very good catches! thanks. all fixed.

prideout (approved):

bejado (approved):

PR #5829, opened 2022-07-28
Fix PlatformEGLHeadless build and add a build.sh option to build it

aeneev:
I have disabled building SDL with headless EGL, because SDL_config_minimal.h doesn't work with EGL, and I don't know how to implement a an SDL config that would work with EGL.

I have verified that this compiles, and verified that rendering works (with a 24-bit fallback for depth buffer) in a separate project.
```
$ ./build.sh -e release
$ find ./out/ -name "*EGL*"
./out/cmake-release/filament/backend/CMakeFiles/backend.dir/src/opengl/platforms/PlatformEGL.cpp.o
./out/cmake-release/filament/backend/CMakeFiles/backend.dir/src/opengl/platforms/PlatformEGLHeadless.cpp.o
./out/cmake-release/libs/bluegl/CMakeFiles/bluegl.dir/src/BlueGLLinuxEGL.cpp.o
```

About using char instead of GLubyte in bluegl. This is consistent with Khronos spec and doesn't depend gl headers: https://registry.khronos.org/EGL/sdk/docs/man/html/eglGetProcAddress.xhtml


build.sh
```diff
@@ -44,6 +44,8 @@ function print_help {
     echo "        Add iOS simulator support to the iOS build."
     echo "    -t"
     echo "        Enable SwiftShader support for Vulkan in desktop builds."
+    echo "    -e"
+    echo "        Enable EGL on linux support for in desktop builds."
```
bejado, line 48:
```suggestion
    echo "        Enable EGL on Linux support for desktop builds."
```

bejado (approved):
Thank you!

filament/backend/src/opengl/platforms/PlatformEGL.cpp
```diff
@@ -91,7 +91,11 @@ Driver* PlatformEGL::createDriver(void* sharedContext, const Platform::DriverCon
         return nullptr;
     }
 
+#if defined(__ANDROID__) || defined(FILAMENT_USE_EXTERNAL_GLES3) || defined(__EMSCRIPTEN__)
+    // PlatofrmEGL is used with and without GLES, but this function is only
+    // defined for the platforms above in gl_headers.h.
```
pixelflinger, line 96: Or slightly more accurately "but this function is only meaningful when GLES is used".

pixelflinger (approved):

filament/backend/src/opengl/platforms/PlatformEGL.cpp
aeneev: Done.

PR #5832, opened 2022-07-28
Headless EGL: Fallback to a 24-bit depth buffer

aeneev:


filament/backend/src/opengl/platforms/PlatformEGLHeadless.cpp
```diff
@@ -138,6 +138,17 @@ backend::Driver* PlatformEGLHeadless::createDriver(void* sharedContext) noexcept
         goto error;
     }
 
+    // fallback to a 24-bit depth buffer
+    if (mEGLConfig == nullptr) {
```
pixelflinger, line 142: I think it would be slightly better to check `configsCount==0` instead

pixelflinger (changes requested):

filament/backend/src/opengl/platforms/PlatformEGLHeadless.cpp
aeneev: Done.

bejado (approved):

pixelflinger (approved):

Issue #5828, opened 2022-07-27
Depth of Field on WebGL has artifacts


jfaust:
**Describe the bug**
Enabling depth of field in WebGL results in artifacts, depending on the settings.

**To Reproduce**
In `suzanne.html`, add:
```
this.view.setDepthOfFieldOptions({
  cocScale: 3.0,
  enabled: true,
});
this.camera.setFocusDistance(5);
this.camera.setExposure(8.0, 1 / 125.0, 100.0);
```
Around line 115.

**Expected behavior**
DoF should work similarly to on desktop.

**Screenshots**
![image](https://user-images.githubusercontent.com/1388377/181373242-0443102f-f535-4cab-ab3e-677912b082f2.png)

Depending on the settings, things can get worse, e.g.:
![image](https://user-images.githubusercontent.com/1388377/181373379-3ca18057-85a1-4ae0-b60f-4b433f3deebc.png)

**Desktop (please complete the following information):**
 - OS: Windows, Linux
 - GPU: NVIDIA GeForce RTX 3080 Ti
 - Backend: WebGL

prideout:
I'm guessing this is due to what Chrome might consider a "feedback loop", although nowadays it's quite permissive, e.g. we were able to remove our bloom workaround back in April.  This might be different since it involves depth.

Reproduces on macOS too btw.

jfaust:
I meant to mention - reproduces on Firefox on Windows too

PR #5833, opened 2022-07-28
Fix DoF on WebGL

pixelflinger:
The issue was that WebGL2.0 doesn't support texture swizzle.

Fixes: #5828


prideout (approved):
Thanks for investigating Mathias!

bejado (approved):

romainguy (approved):

PR #5834, opened 2022-07-28
Metal: implement MSAA stencil buffers

bejado:


pixelflinger (commented):
How is the "automatic resolve" of the stencil buffer done? is it like for depth? (nearest)?

bejado:
> How is the "automatic resolve" of the stencil buffer done? is it like for depth? (nearest)?

Metal has [two options](https://developer.apple.com/documentation/metal/mtlmultisamplestencilresolvefilter?language=objc):
- take sample 0 (default)
- use the same sample that the depth resolve would have chosen

pixelflinger:
> > How is the "automatic resolve" of the stencil buffer done? is it like for depth? (nearest)?
> 
> Metal has [two options](https://developer.apple.com/documentation/metal/mtlmultisamplestencilresolvefilter?language=objc):
> 
> * take sample 0 (default)
> * use the same sample that the depth resolve would have chosen (by default, this is also sample 0)

Oh nice, especially the 2nd option.

pixelflinger (approved):

PR #5853, opened 2022-07-31
1. const_cast 2. explicitly  both convert to uint8_t for comparison

jiapei100:


romainguy (commented):
Why are you changing MeshAssimp.cpp? Your change is breaking the build.

filament/backend/include/private/backend/MetalPlatform.h
```diff
@@ -20,7 +20,7 @@
 #include <backend/DriverEnums.h>
 #include <backend/Platform.h>
 
-#import <Metal/Metal.h>
+#include <Metal/Metal.h>
```
romainguy, line 23: #import is valid on macOS here, but #include is fine too.

bejado: I don't understand what this change has to do with the rest of the PR or why it's necessary. I'd prefer to keep the `#import` unless there's a specific reason.

filament/src/ToneMapper.cpp
```diff
@@ -238,7 +238,7 @@ DEFAULT_CONSTRUCTORS(DisplayRangeToneMapper)
 
 float3 DisplayRangeToneMapper::operator()(math::float3 c) const noexcept {
     // 16 debug colors + 1 duplicated at the end for easy indexing
-    constexpr float3 debugColors[17] = {
+    float3 debugColors[17] = {
```
pixelflinger, line 241: I tried this code with gcc 12.1 and it works.

In any case, try `static constexpr` instead. If it doesn't work, use `static const`.

filament/src/Froxelizer.cpp
```diff
@@ -679,10 +679,13 @@ void Froxelizer::froxelizeAssignRecordsCompress() noexcept {
...
-                .offset = offset,
-                .count = (uint8_t)std::min(size_t(255), b.lights.count()),
-        };
+        // FroxelEntry entry{
+        //         .offset = offset,
+        //         .count = (uint8_t)std::min(size_t(255), b.lights.count()),
+        // };
+        FroxelEntry entry;
+        entry.offset = offset;
+        entry.count = (uint8_t)std::min(size_t(255), b.lights.count());
```
pixelflinger, line 688: I tried this with gcc 12.1 and it works (compiling for C++17).

filament/src/Froxelizer.cpp
```diff
@@ -692,7 +695,11 @@ void Froxelizer::froxelizeAssignRecordsCompress() noexcept {
             // note: instead of dropping froxels we could look for similar records we've already
             // filed up.
             do {
-                froxels[i] = { .offset = 0, .count = allLightsCount };
+                FroxelEntry fe;
+                fe.offset = 0;
+                fe.count = allLightsCount;
+                froxels[i] = fe;
+                // froxels[i] = { .offset = 0, .count = allLightsCount };
```
pixelflinger, line 702: Please revert this as well, this seems to work with GCC 12.1.

filament/test/filament_test.cpp
```diff
@@ -66,7 +66,7 @@ static bool vec3eq(float3 a, float3 b) {
 }
 
 TEST(FilamentTest, AabbMath) {
-    constexpr Aabb aabb = {{4, 5, 6}, {12, 14, 11}};
+    Aabb aabb = {{4, 5, 6}, {12, 14, 11}};
```
pixelflinger, line 69: try `static constexpr`, if doesn't work use `static const`. Tried this with gcc 12.1 and it's working.

libs/filamentapp/src/MeshAssimp.cpp
```diff
@@ -51,9 +51,10 @@
 #include <assimp/postprocess.h>
 #include <assimp/cimport.h>
 #include <assimp/scene.h>
-#include <assimp/pbrmaterial.h>
+// #include <assimp/pbrmaterial.h>
+#include <assimp/GltfMaterial.h>
```
pixelflinger, line 55: The assimp update should be a separate PR.

pixelflinger (changes requested):
Most changes here seem to not be necessary with GCC 12.1. What version of GCC are you using?

The assimp update should be a separate PR.

pixelflinger:
I don't think any of these changes are warranted based on my tests with gcc. For assimp, we would consider updating as a separate PR.

PR #5862, opened 2022-08-01
MaterialInstance public API friendly to std::string_view

pixelflinger:
Internally we use std::string_view and the public API is augmented
with APIs that take a length for strings, which makes them useable
with string_view parameters.

Also fix exception specification for all setParameter methods, which
can throw if exceptions are enabled (or exit the program otherwise).


prideout (approved):
To be pedantic the docstrings could mention that `nameLength` should NOT include the terminating null, although that's pretty obvious.

PR #5863, opened 2022-08-01
get rid of `utils::StaticString`

pixelflinger:
In most places this is simply replaced by `std::string_view`.

We also change a few internal/private headers so they accept 
`std::string_view` instead of `utils::CString`.


prideout (approved):
Very nice.

PR #5869, opened 2022-08-02
make all camera APIs double

pixelflinger:
This was already mostly the case, but we had two stray APIs that were still `float3` at the public API level, but were already `double3` internally.

### DIFF
--- a/filament/include/filament/Camera.h
+++ b/filament/include/filament/Camera.h
@@ -356,17 +356,9 @@ public:
      * @param center    The point in world space the camera is looking at.
      * @param up        A unit vector denoting the camera's "up" direction.
      */
-    void lookAt(const math::float3& eye,
-                const math::float3& center,
-                const math::float3& up) noexcept;
-
-    /** Sets the camera's model matrix, assuming up is along the y axis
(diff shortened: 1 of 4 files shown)


filament/include/filament/Camera.h
```diff
@@ -356,17 +356,9 @@ class UTILS_PUBLIC Camera : public FilamentAPI {
...
-                const math::float3& up) noexcept;
-
-    /** Sets the camera's model matrix, assuming up is along the y axis
-     *
-     * @param eye       The position of the camera in world space.
-     * @param center    The point in world space the camera is looking at.
-     */
-    void lookAt(const math::float3& eye,
-                const math::float3& center) noexcept;
+    void lookAt(math::double3 eye,
```
bejado, line 359: Why not `const math::double3&`?

bejado (approved):

filament/include/filament/Camera.h
pixelflinger: yeah actually it's a good point, I will change it back. It looks better in more cases with references as seen in compiler explorer.

PR #5873, opened 2022-08-02
Move stencil state out of RasterState

bejado:
This allows us to grow the stencil state without having to pack additional bits into `RasterState`. In a separate PR I'll add support for separate front/back stencil states and masking.


pixelflinger (approved):

Benny6731:
hey

PR #5883, opened 2022-08-04
Fix PrimitiveInfo size

bejado:


pixelflinger (changes requested):
I would like Command to stay 64 bytes so it's exactly a cache line multiple.

pixelflinger (approved):

PR #5886, opened 2022-08-04
Add View API to enable the stencil buffer

bejado:
Will update RELEASE_NOTES when the Material API is added, otherwise at the moment there's no way to use the stencil buffer.

### DIFF
--- a/filament/include/filament/View.h
+++ b/filament/include/filament/View.h
@@ -637,6 +637,33 @@ public:
      */
     bool isFrontFaceWindingInverted() const noexcept;
 
+    /**
+     * Enables use of the stencil buffer. This API is currently a WIP and experimental.
+     *
+     * The stencil buffer is an 8-bit, per-fragment unsigned integer stored alongside the depth
+     * buffer. The stencil buffer is cleared at the beginning of a frame and discarded after the
+     * color pass.
+     *
+     * Each fragment's stencil value is set during rasterization by specifying stencil operations on
+     * a Material. The stencil buffer can be used as a mask for later rendering by setting a
+     * Material's stencil comparison function and reference value. Fragments that don't pass the
+     * stencil test are then discarded.
+     *
+     * Post-processing must be enabled in order to use the stencil buffer.
+     *
+     * A renderable's priority (see RenderableManager::setPriority) is useful to control the order
+     * in which primitives are drawn.
+     *
+     * @param enabled True to enable the stencil buffer, false disables it (default)
+     */
+    void setStencilBufferEnabled(bool enabled) noexcept;
+
+    /**
+     * Returns true if the stencil buffer is enabled.
+     * See setStencilBufferEnabled() for more information.
+     */
+    bool getStencilBufferEnabled() const noexcept;
+
     // for debugging...
 
     //! debugging: allows to entirely disable frustum culling. (culling enabled by default).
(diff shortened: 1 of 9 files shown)


filament/include/filament/View.h
```diff
@@ -637,6 +637,33 @@ class UTILS_PUBLIC View : public FilamentAPI {
...
+     * The stencil buffer is an 8-bit, per-fragment unsigned integer stored alongside the depth
+     * buffer. The stencil buffer is cleared at the beginning of a frame and discarded after the
+     * color pass.
+     *
+     * Each fragment's stencil value is set during rasterization by specifying stencil operations on
+     * a Material. The stencil buffer can be used as a mask for later rendering by setting a
+     * Material's stencil comparison function and reference value. Fragments that don't pass the
+     * stencil test are then discarded.
+     *
+     * Post-processing must be enabled in order to use the stencil buffer.
```
pixelflinger, line 652: why is that a requirement?

filament/src/RendererUtils.cpp
```diff
@@ -92,9 +94,15 @@ FrameGraphId<FrameGraphTexture> RendererUtils::colorPass(
...
+                    // clear newly allocated depth/stencil buffers, regardless of given clear flags
                     clearDepthFlags = TargetBufferFlags::DEPTH;
-                    data.depth = builder.createTexture("Depth Buffer", {
+                    clearStencilFlags = config.enabledStencilBuffer ?
+                            TargetBufferFlags::STENCIL : TargetBufferFlags::NONE;
+                    const char* const name = config.enabledStencilBuffer ?
+                             "Depth/Stencil Buffer" : "Depth Buffer";
+                    TextureFormat format = config.enabledStencilBuffer ?
+                            TextureFormat::DEPTH32F_STENCIL8 : TextureFormat::DEPTH32F;
+                    data.depth = builder.createTexture(name, {
```
pixelflinger, line 105: make sure this works with transmission, which reuses the depth buffer (also transmission + MSAA)

filament/src/View.cpp
```diff
@@ -275,6 +275,14 @@ bool View::isScreenSpaceRefractionEnabled() const noexcept {
     return upcast(this)->isScreenSpaceRefractionEnabled();
 }
 
+void View::setStencilBufferEnabled(bool enabled) noexcept {
+    upcast(this)->setStencilBufferEnabled(enabled);
+}
+
+bool View::getStencilBufferEnabled() const noexcept {
+    return upcast(this)->getStencilBufferEnabled();
+}
```
pixelflinger, line 284: don't forget the java and js bindings

pixelflinger (approved):
looks pretty good, just had a couple questions

filament/include/filament/View.h
bejado: At the moment, the default render target does not allocate a Stencil buffer. Maybe we should.
Technically the requirement is either post-processing OR MSAA must be enabled.

filament/src/RendererUtils.cpp
bejado: Good idea, I'll check

filament/include/filament/View.h
pixelflinger: Ah right. Maybe we should add a flag to the swapchain creation to specify it needs to support stencil? And say that it's a requirement if post-process is not used?

filament/src/View.cpp
bejado: Yes :) I'll add them in a separate PR once we have Material APIs to actually do something interesting with the stencil buffer.

filament/include/filament/View.h
bejado: Yeah that's a good idea

PR #5887, opened 2022-08-04
smalls improvement to GL backend stencil and raster state

pixelflinger:
reviewers: it'll much easier to review each CL separately. thanks.

### DIFF
--- a/filament/backend/include/backend/DriverEnums.h
+++ b/filament/backend/include/backend/DriverEnums.h
@@ -960,15 +960,6 @@ struct PolygonOffset {
 struct StencilState {
     using StencilFunction = SamplerCompareFunc;
 
-    StencilState() noexcept : referenceValue(0u), stencilWrite(false), padding(0u) {
-        static_assert(sizeof(StencilOperations) == sizeof(uint32_t), "StencilOperations size not what was intended");
-        static_assert(sizeof(StencilState) == 12u, "StencilState size not what was intended");
-        front.u = back.u = 0u;
-        front.stencilFunc = back.stencilFunc = StencilFunction::A;
-        front.readMask = back.readMask = 0xFF;
-        front.writeMask = back.writeMask = 0xFF;
-    }
-
     bool operator==(StencilState rhs) const noexcept {
         return front.u == rhs.front.u &&
                 back.u == rhs.back.u &&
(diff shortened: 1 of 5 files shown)


filament/backend/include/backend/DriverEnums.h
```diff
@@ -1002,29 +979,37 @@ struct StencilState {
...
             //! Masks the bits of the stencil values participating in the stencil comparison test.
-            uint8_t readMask                                : 8;                    // 24
+            uint8_t readMask;                                                       // 24
 
             //! Masks the bits of the stencil values updated by the stencil test.
-            uint8_t writeMask                               : 8;                    // 32
+            uint8_t writeMask;                                                      // 32
         };
         uint32_t u = 0;
     };
```
bejado, line 988: Do we even need this `union` anymore?

pixelflinger: that's a good point, let me play with it. probably not.

filament/backend/include/backend/DriverEnums.h
```diff
@@ -960,29 +960,6 @@ struct PolygonOffset {
 struct StencilState {
     using StencilFunction = SamplerCompareFunc;
 
-    StencilState() noexcept : referenceValue(0u), stencilWrite(false), padding(0u) {
-        static_assert(sizeof(StencilOperations) == sizeof(uint32_t), "StencilOperations size not what was intended");
-        static_assert(sizeof(StencilState) == 12u, "StencilState size not what was intended");
```
bejado, line 965: As long as we have padding bytes, it's probably a good idea to keep this assertions.

bejado: Oh just kidding, I see you put them below 😄

filament/backend/include/backend/DriverEnums.h
bejado: Note that this struct may grow—we can actually move the stencil reference value here, as it's also set separately per front/back facing polygons.

filament/backend/include/backend/DriverEnums.h
pixelflinger: I just moved it.

bejado (approved):

PR #5889, opened 2022-08-05
Stencil reference value can be separate for front and back

pixelflinger:


filament/backend/include/backend/DriverEnums.h
```diff
@@ -977,31 +977,29 @@ struct StencilState {
...
 
         //! Masks the bits of the stencil values updated by the stencil test.
-        uint8_t writeMask;                                                      // 32
+        uint8_t writeMask = 0xff;                                               // 40
     };
 
     //! Stencil operations for front-facing polygons
-    StencilOperations front = {
-            .stencilFunc = StencilFunction::A, .readMask = 0xff, .writeMask = 0xff };
+    StencilOperations front = { .stencilFunc = StencilFunction::A };
```
bejado, line 991: Could we put the `stencilFunc` initializer as part of the `StencilOperations` class and leave this as `= {}`?

bejado (approved):

filament/backend/include/backend/DriverEnums.h
pixelflinger: I don't think we can because it's a bitfield. Or we have to create a ctor, but then we can't use designated initializers.

Maybe it would be more logical to not do any default initialization at all in `StencilOperations` and leave it to `StencilStace`.

PR #5890, opened 2022-08-05
Ensure UTILS_DEPRECATED is used in the public API

Orctonk:
Renderer and Texture headers use the Attribute Specifier Sequence syntax instead of the UTILS_DEPRECATED macro which could cause issues when interfacing with compilers without support for this syntax. 

This merge request simply changes any occurrence of [[deprecated]] in the headers to UTILS_DEPRECATED


romainguy (approved):
Thank you! Please make sure to sign the CLA so we can merge your change.

pixelflinger (approved):

PR #5900, opened 2022-08-09
gltfio: add 'detach' methods to allow ownership transfer

prideout:
These new methods allow gltfio to be integrated into internal Google
libraries.

### DIFF
--- a/libs/gltfio/include/gltfio/FilamentAsset.h
+++ b/libs/gltfio/include/gltfio/FilamentAsset.h
@@ -334,6 +334,23 @@ public:
     void addEntitiesToScene(filament::Scene& targetScene, const Entity* entities, size_t count,
             SceneMask sceneFilter);
 
+    /**
+     * Releases ownership of entities and their Filament components.
+     *
+     * This makes the client take responsibility for destroying Filament
+     * components (e.g. Renderable, TransformManager component) as well as
+     * the underlying entities.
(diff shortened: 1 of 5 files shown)


prideout:
I'd like to cherry pick this to RC so I added it to section v1.25.5 in the release notes.

libs/gltfio/include/gltfio/FilamentAsset.h
```diff
@@ -334,6 +334,23 @@ class UTILS_PUBLIC FilamentAsset {
...
             SceneMask sceneFilter);
 
+    /**
+     * Releases ownership of entities and their Filament components.
+     *
+     * This makes the client take responsibility for destroying Filament
+     * components (e.g. Renderable, TransformManager component) as well as
+     * the underlying entities.
+     */
+    void detachFilamentComponents();
```
romainguy, line 344: Is the name "Filament" required here?

romainguy (approved):

libs/gltfio/include/gltfio/FilamentAsset.h
prideout: Imp uses our ECS, this makes it clear that Imp components are not being detached.

romainguy: 👍

PR #5908, opened 2022-08-11
gltfio: remove a memcpy via custom cgltf_file_options.

prideout:
Earlier versions of cgltf did not support file reader customization.
This was fixed back in Dec 2019 but at the time I did not notice, so
we never bothered cleaning up our usage.

In the future we would like WebGL + Android builds to permit texture
downloads to occur concurrently with the texture decoder jobs. This PR
is basically a preparatory refactoring, but with the nice side effect of
removing a memcpy.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -5,6 +5,8 @@ A new header is inserted each time a *tag* is created.
 
 ## main branch
 
+- gltfio: minor efficiency improvement for Android and WebGL builds.
+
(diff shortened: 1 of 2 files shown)


romainguy (approved):
Very nice. It reminds me that at some point we might want to help our Java APIs load stuff from `InputStream` because currently they either need a copy of the asset in memory, or use a file descriptor to mmap (which requires disabling compression in the APK).

Issue #5909, opened 2022-08-11
gltfio: allow texture decoding to start before everything is fetched


prideout:
Currently, `ResourceLoader` requires all necessary `addResourceData()` calls to occur before `asyncBeginLoad()`.

Load times on non-filesystem platforms (WebGL / Android) can be improved if we remove this requirement.

We intend to implement this feature in two stages:

Stage 1) Allow users to call `asyncBeginLoad` before all texture data has been provided, but after all vertex data has been provided.  This work is already in progress.

Stage 2) Allow users to call `asyncBeginLoad` before any data has been provided (both textures _and_ vertex data).  Note that Draco decoding and tangent computation jobs should start as early as possible.

prideout:
I am going to close this, it's not necessarily a crucial feature, also I recently wrote a 1-pager about Google3 integration, it's better to refer to that.

PR #5910, opened 2022-08-11
gltfio: concurrent texture downloading and decoding.

prideout:
This feature can improve load time when textures are downloaded from the
web on non-filesystem platforms like Android.

More specifically, this allows downloaded texture assets to arrive after
the user calls asyncBeginLoad(), which means that decoding and
downloading can occur concurrently.

Prior to this PR, we already used JobSystem for decoding, but we did not
kick off any jobs until after all assets were downloaded.

Still TBD: add this feature for external vertex data.

Partial fix for #5909.

### DIFF
--- a/libs/gltfio/src/ResourceLoader.cpp
+++ b/libs/gltfio/src/ResourceLoader.cpp
@@ -70,19 +70,20 @@ using UriDataCache = tsl::robin_map<std::string, gltfio::ResourceLoader::BufferD
 using TextureProviderList = tsl::robin_map<std::string, TextureProvider*>;
 
 struct ResourceLoader::Impl {
-    Impl(const ResourceConfiguration& config) {
-        mGltfPath = std::string(config.gltfPath ? config.gltfPath : "");
-        mEngine = config.engine;
-        mNormalizeSkinningWeights = config.normalizeSkinningWeights;
-        mRecomputeBoundingBoxes = config.recomputeBoundingBoxes;
-        mIgnoreBindTransform = config.ignoreBindTransform;
-    }
+    Impl(const ResourceConfiguration& config) :
+        mGltfPath(config.gltfPath ? config.gltfPath : ""),
+        mEngine(config.engine),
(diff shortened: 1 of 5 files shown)


libs/gltfio/src/ResourceLoader.cpp
```diff
@@ -302,20 +305,51 @@ ResourceLoader::~ResourceLoader() {
...
+    if (expr.length() >= ending.length()) {
+        return (expr.compare(expr.length() - ending.length(), ending.length(), ending) == 0);
+    }
+    return false;
+}
+
+// TODO: This is not a great way to determine if a resource is a texture, but we can remove it after
+// gltfio gains support for concurrent downloading of vertex data:
+// https://github.com/google/filament/issues/5909
+static bool isTexture(const char* uri) {
```
romainguy, line 321: What about .ktx?

web/filament-js/extensions.js
```diff
@@ -565,14 +565,28 @@ Filament.loadClassExtensions = function() {
...
+        let textureUris = new Set();
+        let bufferUris = new Set();
+        const absoluteToRelativeUri = {};
+        for (const relativeUri of this.getResourceUris()) {
+            const absoluteUri = '' + new URL(relativeUri, basePath);
+            absoluteToRelativeUri[absoluteUri] = relativeUri;
+            // TODO: This is not a great way to determine if a resource is a texture, but we can
+            // remove it after gltfio gains support for concurrent downloading of vertex data:
+            // https://github.com/google/filament/issues/5909
+            if (relativeUri.endsWith(".png")) {
```
romainguy, line 578: Call isTexture() here

romainguy (approved):

libs/gltfio/src/ResourceLoader.cpp
prideout: Gah, excellent catch, we support ktx2 containers in glTF so I need to check for that extension.  Thanks!  (hopefully will get rid of these checks entirely before too long)

romainguy (approved):

PR #5921, opened 2022-08-15
Do not assert for certain custom CONFIG values.

### LINKED ISSUE #5785: Part of mesh is only visible
I am trying to view my 3d model in mobile using android studio filament library 
After exporting my .blend file to .glb in Blender
The issue I faced was only certain parts of the mesh are visible , that is only the eyes and mouth of the character are visible in the modelviewer loadgltf , not the whole character mesh don't know why this happen.

this is the code snippet:

      private fun loadGlb(name: String) {
        Log.v("working","glb function called")
        val buffer = readAsset("models/${name}.glb")
        modelViewer.loadModelGlb(buffer)
        modelViewer.transformToUnitCube()

snippet by: https://medium.com/@philiprideout/getting-started-with-filament-on-android-d10b16f0ec67
png
only this part  of the character is visible 

I have checked the face orientation  of the 3d meshes and the normal maps. 

The exported  Gltf file (.gltf + .bin)  from blender(Checked on 2.83 , 2.92 , 3.1) versions  are not  loading  into the  model viewer
The exported  Gllb file  from blender(Checked on 2.83 , 2.92 , 3.1) versions  are  loading into the model viewer , but entire mesh is not visible, only part of the mesh is visible ,  and animation  is playing  .
Also please let me know the exact  glb , gltf format supported by filament.

[arya-1.zip](https://github.com/google/filament/files/9108077/arya-1.zip)

This is embedded glTF model. it has no .bin file in it .

 
[arya-1glb.zip](https://github.com/google/filament/files/9108098/arya-1glb.zip)

This is glb version of the same model

prideout:
Developers who know their users have powerful devices can exceed
minspec constraints by adjust CONFIG_MAX_BONE_COUNT and/or
CONFIG_MAX_MORPH_TARGET_COUNT. If they do this, they should not
hit a `static_asset`.

Fixes #5785.

### DIFF
--- a/libs/filabridge/include/private/filament/UibStructs.h
+++ b/libs/filabridge/include/private/filament/UibStructs.h
@@ -288,8 +288,11 @@ struct PerRenderableBoneUib { // NOLINT(cppcoreguidelines-pro-type-member-init)
     };
     BoneData bones[CONFIG_MAX_BONE_COUNT];
 };
-static_assert(sizeof(PerRenderableBoneUib) <= 16384,
-        "PerRenderableUibBone exceed max UBO size");
+
+// ES3.0 only guarantees a max UBO size of 16 KiB, but users may wish to exceed the minspec
+// by adjusting CONFIG_MAX_BONE_COUNT.
+static_assert(sizeof(PerRenderableBoneUib) == CONFIG_MAX_BONE_COUNT * 64,
+        "PerRenderableUibBone has unexpected size");
 
 // ------------------------------------------------------------------------------------------------
 // MARK: -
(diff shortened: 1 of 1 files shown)


libs/filabridge/include/private/filament/UibStructs.h
```diff
@@ -288,8 +288,11 @@ struct PerRenderableBoneUib { // NOLINT(cppcoreguidelines-pro-type-member-init)
     };
     BoneData bones[CONFIG_MAX_BONE_COUNT];
 };
-static_assert(sizeof(PerRenderableBoneUib) <= 16384,
-        "PerRenderableUibBone exceed max UBO size");
+
+// ES3.0 only guarantees a max UBO size of 16 KiB, but users may wish to exceed the minspec
+// by adjusting CONFIG_MAX_BONE_COUNT.
+static_assert(sizeof(PerRenderableBoneUib) == CONFIG_MAX_BONE_COUNT * 64,
```
bejado, line 294: Maybe replace 64 with `sizeof(BoneData)`? That makes me question if we really need this assertion now

Similar suggestion for the morph targets.

libs/filabridge/include/private/filament/UibStructs.h
```diff
@@ -288,8 +288,11 @@ struct PerRenderableBoneUib { // NOLINT(cppcoreguidelines-pro-type-member-init)
...
+        "PerRenderableUibBone has unexpected size");
```
pixelflinger, line 295: This assert becomes a bit useless because the structure itself is an array of `CONFIG_MAX_BONE_COUNT`.

The point of this assert was to make sure to not set a `CONFIG_MAX_BONE_COUNT` that would exceed the minspec we support.

So I think the better/correct solution is to introduce a `CONFIG_MINSPEC_UBO_SIZE` constant the developers can adjust. Same for `CONFIG_MAX_MORPH_TARGET_COUNT`

pixelflinger (changes requested):

Issue #5811, opened 2022-07-25
Fix level 3 MSVC warnings in public headers


palver123:
**Is your feature request related to a problem? Please describe.**
In our application using Filament, we use `/W3` (Warning level 3) and `/WX` (Treat warnings as errors) MSVC compiler options. This causes compiler warnings to pop up in the public headers of Filament and `/WX` promotes those warnings to compiler errors. 

**Describe the solution you'd like**
We have fixed all W3 level warnings in all public headers of Filament on our fork, and we'd like to upstream our changes. We are ready to open a PR which you can review thoroughly (only minor changes in a total of 16 files).

**Describe alternatives you've considered**
We considered including Filament headers as system headers, but our cross-platform build system doesn't fully support it (so `/external:I` flag is out of the question for us now).
We also considered fixing level 4 warnings too, but that needed more aggressive/arguable code changes, so we stopped at W3.

**OS and backend**
Windows specific (because of MSVC).

romainguy:
A PR would be welcome, thanks!

pixelflinger:
Or maybe the list of the warnings you get.

pixelflinger:
#5815 will undoubtedly address some of these warnings.

palver123:
Sorry, we got swamped with other urgent tasks so opening the PR got a bit delayed, but we will get to it soon (this week 🤞 )!

As for the exact list of warnings you can check the commit messages on [this branch](https://github.com/google/filament/compare/main...shapr3d:filament:shapr3d/fix_msvc_warnings). This is the branch we plan to open the PR from btw.

I checked #5815 and I think the warnings fixed there are separate from the MSVC warnings we found.

pixelflinger:
okay, a lot of these look totally reasonable. thanks!

PR #5926, opened 2022-08-16
Fix MSVC warnings

gaborvalasek:
As per https://github.com/google/filament/issues/5811 , this PR fixes `/W3` level warnings in Filament public headers on MSVC.


libs/math/include/math/mat2.h
```diff
@@ -243,7 +243,9 @@ class MATH_EMPTY_BASES TMat22 :
         uint64_t result = 0;
         // For some reason clang is not able to vectoize this loop when the number of iteration
         // is known and constant (!?!?!). Still this is better than operator==.
+#if defined(__clang__)
```
romainguy, line 246: Is there an MSVC version of this?

libs/utils/include/utils/Allocator.h
```diff
@@ -51,7 +51,7 @@ static inline P* align(P* p, size_t alignment) noexcept {
 template <typename P>
 static inline P* align(P* p, size_t alignment, size_t offset) noexcept {
     P* const r = align(add(p, offset), alignment);
-    assert(pointermath::add(r, -offset) >= p);
+    assert(r >= add(p, offset));
```
romainguy, line 54: What was the issue here?

libs/gltfio/include/gltfio/MaterialProvider.h
```diff
@@ -36,8 +36,13 @@ enum class AlphaMode : uint8_t {
 
 // The following struct gets hashed so all padding bits should be explicit.
 // Tell the compiler to emit a warning if it adds any padding.
+#if defined(__clang__)
 #pragma clang diagnostic push
 #pragma clang diagnostic warning "-Wpadded"
+#elif defined(_MSC_VER)
+#pragma warning( push )
+#pragma warning(disable : 4324)
```
romainguy, line 44: Since there are two instances of this, maybe define a UTILS_DISABLE_WARNING_PADDED (or a better name)?

libs/gltfio/include/gltfio/MaterialProvider.h
```diff
@@ -36,8 +36,13 @@ enum class AlphaMode : uint8_t {
 
 // The following struct gets hashed so all padding bits should be explicit.
 // Tell the compiler to emit a warning if it adds any padding.
+#if defined(__clang__)
 #pragma clang diagnostic push
 #pragma clang diagnostic warning "-Wpadded"
+#elif defined(_MSC_VER)
+#pragma warning( push )
```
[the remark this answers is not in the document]
pixelflinger: Yes, that would be better, I think it can be done with the `_Pragma` syntax.

libs/gltfio/include/gltfio/MaterialProvider.h
pixelflinger: I agree, if it's doable, it would probably be cleaner and less error prone in the future.

libs/utils/include/utils/algorithm.h
```diff
@@ -59,7 +59,7 @@ template<typename T, typename = std::enable_if_t<std::is_unsigned<T>::value>>
 constexpr inline T ctz(T x) noexcept {
     static_assert(sizeof(T) * CHAR_BIT <= 64, "details::ctz() only support up to 64 bits");
     T c = sizeof(T) * CHAR_BIT;
-    x &= -x;    // equivalent to x & (~x + 1)
+    x &= (~x + 1);
```
pixelflinger, line 62: I would like to keep the old code here because I checked that it generated a better sequence on ARM. So Maybe use a #if for msvc.

pixelflinger (changes requested):
thanks for all the fixes. We have just a few comments, see inline.

libs/uberz/include/uberz/ReadableArchive.h
```diff
@@ -30,8 +30,13 @@ namespace filament::uberz {
 // offset fields into pointers.
 void convertOffsetsToPointers(struct ReadableArchive* archive);
 
+#if defined(__clang__)
 #pragma clang diagnostic push
 #pragma clang diagnostic warning "-Wpadded"
+#elif defined(_MSC_VER)
+#pragma warning( push )
+#pragma warning(disable : 4324)
```
bejado, line 38: Add a comment describing the warning please.

libs/image/include/image/ColorTransform.h
```diff
@@ -231,8 +230,7 @@ inline std::unique_ptr<uint8_t[]> fromLinearToRGB_10_11_11_REV(const LinearImage
     using namespace filament::math;
     size_t w = image.getWidth();
     size_t h = image.getHeight();
-    UTILS_UNUSED_IN_RELEASE size_t channels = image.getChannels();
```
bejado, line 234: Does MSVC support  [`[[maybe unused]]`](https://en.cppreference.com/w/cpp/language/attributes/maybe_unused)? I don't think MSVC supports `__has_attribute` so we might have to define `UTILS_UNUSED_IN_RELEASE` specifically for MSVC.

libs/utils/include/utils/Allocator.h
palver123: `offset` is an unsigned integer type and compiler warning 4146 kicked in. Our initial solution was to pragma push disable the warning, but then we figured that the inequality could be rearranged.

Do you see a specific hazard in this change? I only considered integer overflow/underflow but the diff doesn't seem to worsen the chance of that. In fact if `add(p, offset)` overflowed, `align` would break (regardless of our changes).

libs/math/include/math/mat2.h
palver123: I think [Auto-Vectorization](https://docs.microsoft.com/en-us/cpp/parallel/auto-parallelization-and-auto-vectorization?view=msvc-170#auto-vectorizer) is the name of the similar MSVC feature. The exact SIMD solution it uses depends on `/arch` compiler flag (`/arch:x64` [defaults to SSE2](https://docs.microsoft.com/en-us/cpp/build/reference/arch-x64?view=msvc-170) instruction set). But it is not as configurable as in GCC or Clang.

[hint_parallel(n)](https://docs.microsoft.com/en-us/cpp/preprocessor/loop) seems to be the closest I could find, but that's only available with an opt-in compiler flag I have never heard of before (`/Qpar`). Plus it belongs to Auto-**Parallelizer** feature which parallelizes code via **threads**, not via vectorization.

So my short answer is probably no, there isn't an MSVC version 😞

libs/gltfio/include/gltfio/MaterialProvider.h
palver123: Sounds good, see `fb7c465c5d06411b2845e02397c98f8fa8493d74`. The only drawback is that if you compile with /W0 in MSVC, `#pragma warning(1: XXXX)` will do nothing 😞 . This is different in Clang, where even -Wno-everything is overridden by `clang diagnostic warning "-Wpadded"`. You can consider raising an error instead of a warning because both MSVC and Clang ignore compiler flags when a pragma requests an error for a warning.

Note: at first I misunderstood the clang code and disabled the warning on MSVC, instead of enabling it. Fixed that too

libs/utils/include/utils/algorithm.h
palver123: We were curious and tried to reproduce it on godbolt, but all compilers we tested generated the same code. But to keep all compilers happy I added the #if in 2944599e5974496b6da8964a1c058e80a6b1a41e

libs/image/include/image/ColorTransform.h
palver123: [It does](https://docs.microsoft.com/en-us/cpp/cpp/attributes?view=msvc-170) since VS 15.3 (and only if /std:c++17 (or later) compiler flag is used). This can be checked via [predefined macros](https://docs.microsoft.com/en-us/cpp/preprocessor/predefined-macros?view=msvc-170) of MSVC.
`_MSC_VER` can effectively identify the VS version (the corresponding value to VS 15.3 [is 1911](https://dev.to/yumetodo/list-of-mscver-and-mscfullver-8nd)) and the language version compiler flag can be tested via `_MSVC_LANG`.
See ac08c4018c9bfec4285ce2e084f4326b706631fe.

libs/uberz/include/uberz/ReadableArchive.h
palver123: Is this still applicable after fb7c465c5d06411b2845e02397c98f8fa8493d74?

libs/image/include/image/ColorTransform.h
palver123: Note that now `UTILS_UNUSED_IN_RELEASE` is back in ColorTransform.h it could still emit a warning for someone who uses a very old VS version or compiles Filament with /std:c++14 or lesser (although I am not even sure if Filament would compile that way)

libs/utils/include/utils/Allocator.h
pixelflinger: sounds good. thanks.

libs/utils/include/utils/algorithm.h
pixelflinger: okay good to know. thanks for looking into it.

libs/utils/include/utils/compiler.h
```diff
@@ -64,6 +64,12 @@
 #    define UTILS_PRIVATE
 #endif
 
+#define UTILS_HAS_CXX17 0
+#if __cplusplus >= 201703L  || (defined(_MSVC_LANG) && _MSVC_LANG >= 201703L)
+#undef UTILS_HAS_CXX17
+#define UTILS_HAS_CXX17 1
+#endif
```
pixelflinger, line 71: confused about this one because we require C++17. So this should always be true.

libs/math/include/math/mat2.h
pixelflinger: sounds good thanks!

pixelflinger (approved):

libs/utils/include/utils/compiler.h
bejado: I think the goal is for clients to be able to compile against Filament public headers with older versions of MSVC.

libs/uberz/include/uberz/ReadableArchive.h
bejado: No, looks good :)

libs/image/include/image/ColorTransform.h
bejado: Thanks! Older versions of MSVC I'm not too concerned about; like you said, not even sure if they would compile.

bejado (approved):

libs/utils/include/utils/compiler.h
palver123: I wanted to ask/check the required C++ language version in Filament; I am glad to hear it is C++17.

And @bejado is right about my original intention. However I checked the public headers of Filament and they contain C++17 features (nested namespace definition, `std::string_view`) so it is not a valid use case to compile against Filament public headers with `/std:c++14` (or lower). So I removed UTILS_HAS_CXX17  in f8c74de3507e1861f7433a392b25fad69c4cbc04.

palver123: Fun fact: _MaterialBuilder.h_ and _DriverEnums.h_ use designated initializers too ([which is a C++ 20 feature](https://en.cppreference.com/w/cpp/language/aggregate_initialization#Designated_initializers)).

libs/gltfio/include/gltfio/MaterialProvider.h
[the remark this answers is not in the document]
palver123: @romainguy I added **2 macros** in compiler.h but maybe I misunderstood and you meant to have **3 separate** macros like this?

```cpp
UTILS_WARNING_PUSH
UTILS_WARNING_ENABLE_PADDING
//...
UTILS_WARNING_POP
```

Is that right? Should I split `UTILS_PUSH_WARNING_ENABLE_PADDED` into 2 macros?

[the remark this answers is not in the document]
romainguy: Yes, I meant 3 separate macros.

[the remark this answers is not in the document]
palver123: Okay, rewrote the macros in 8e03640d5ac1c50fa9c86b95c8c4305cd85e10c5

romainguy (approved):

romainguy:
Thank you! I just started the presubmit runs, we should be able to merge your contribution as soon as they pass.

PR #5932, opened 2022-08-16
gltfio API change: assets are now always 'instanced' etc

prideout:
This change was motivated by some internal work at Google and has the benefit of simplifying the gltfio API and
implementation. There are 2 major API changes:

(1) Consolidate separate loader entry points for GLB and GLTF.

The distinction between GLB and GLTF can be made from the file content alone, because GLB has a 4-byte magic string in its header. There is no need for separate entry points.  Clients do not (and should not) need to check the file name extension.

(2) Remove the distinction between "instanced" and "non-instanced" glTF assets.

In the new scheme, all assets have at least 1 instance.

Broadly speaking, in gltfio an "asset" is a collection of Filament objects like textures and vertex buffers, while an "instance" is a collection of entities and components (e.g. the transform hierarchy).

This API change makes life easier for clients because they no longer need to decide a priori if they will ever need to add instances.

This change also moves some public-facing methods from `FilamentAsset` to `FilamentInstance`:

    - getSkinCount, getSkinNameAt
    - getJointCountAt, getJointsAt
    - attachSkin, detachSkin

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -6,6 +6,8 @@ A new header is inserted each time a *tag* is created.
 ## main branch
 
 - engine: new feature level APIs, see `Engine::getSupportedFeatureLevel()`
+- gltfio: add unified `AssetLoader::createAsset()` method [⚠️ **API Change**]
+- gltfio: all assets are now "instanced" [⚠️ **API Change**]
 
 ## v1.25.6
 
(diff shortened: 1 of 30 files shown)


pixelflinger (approved):

RELEASE_NOTES.md
```diff
@@ -5,6 +5,9 @@ A new header is inserted each time a *tag* is created.
 
 ## main branch
 
+- gltfio: add unified `AssetLoader::loadAsset()` method [⚠️ **API Change**]
```
bejado, line 8: Do you mean `AssetLoader::createAsset`?

bejado (approved):

RELEASE_NOTES.md
prideout: Great catch, fixed.

PR #5934, opened 2022-08-17
Organize Subprojects into Folders in the IDE

zalo:
Currently, subprojects are displayed alphabetically in a flat list, with libraries, samples, third party utilities, and core projects interleaved:
SubprojectsAlphabetically

This PR adds CMake attributes that help IDEs (at least, Visual Studio) sort the subprojects within filament so that they are easier to group together and reason about.

This is what it looks like after the PR:
SubprojectsInFolders

My hope with this PR is that filament will become less intimidating to newcomers approaching this project for the first time.


bejado:
Thanks! Overall looks good, though for a few third_party libraries I'm not sure the correct CMakeList file is updated. By convention, for 3P libraries we put our own (modified) CMakeLists.txt file under a `tnt/` directory. So, for example, the change to `third_party/hat-trie/CMakeLists.txt` should be reverted and made instead to `third_party/hat-trie/tnt/CMakeLists.txt`.

zalo:
@bejado  Ah; makes sense!   I just reverted the changes to the third_party libraries that use the `tnt` pattern. 👍

zalo:
(Also, unrelated to the PR, I was looking at your homepage @bejado and saw an ambient occlusion demo similar to one I contributed to `three.js`; you might find it fun: https://threejs.org/examples/#webgl_shadowmap_progressive )

filament/CMakeLists.txt
```diff
@@ -523,6 +523,9 @@ add_library(${TARGET} STATIC ${PRIVATE_HDRS} ${PUBLIC_HDRS} ${SRCS} ${DATA_BINS}
 # specify where the public headers of this library are
 target_include_directories(${TARGET} PUBLIC ${PUBLIC_HDR_DIR})
 
+# add this subproject to the Core folder
+set_target_properties(${TARGET} PROPERTIES FOLDER Core)
```
pixelflinger, line 527: What about we use "Filament" instead of "Core", so it mimics better the folder names?

bejado:
Looks like we have a build failure:

```
CMake Error at third_party/libsdl2/tnt/CMakeLists.txt:233 (set_target_properties):
  set_target_properties Can not find target to add properties to: sdl2main
```

bejado:
> (Also, unrelated to the PR, I was looking at your homepage @bejado and saw an ambient occlusion demo similar to one I contributed to `three.js`; you might find it fun: https://threejs.org/examples/#webgl_shadowmap_progressive )
> 
> ([and also the radiosity version](https://zalo.github.io/three.js/examples/#webgl_shadowmap_progressive))

Very cool 😎 always impressive to see what Three.js is capable of!

filament/CMakeLists.txt
zalo: Makes sense; should “shaders” and “vkShaders” go into a separate sub-folder?  They’re currently in the Core/Filament folder, but since they appear to be auto-generated from the .vs and .fs files, I’m not sure if they should be on the same categorical level…

romainguy: Any generated code should in a sub-folder called Generated if possible. It's stuff a developer is not supposed to look at (unless working on the code generator itself).

zalo: Alrighty; these are now in the Filament/Generated subfolder

shaders/CMakeLists.txt
```diff
@@ -108,6 +108,7 @@ endif()
 # ==================================================================================================
 add_library(${TARGET} STATIC ${RESGEN_SOURCE})
 target_include_directories(${TARGET} PUBLIC ${CMAKE_CURRENT_BINARY_DIR})
+set_target_properties(${TARGET} PROPERTIES FOLDER Filament/Generated)
```
romainguy, line 111: Those shaders are hand written and shouldn't go in Generated. They should go in Filament/Shaders

romainguy (changes requested):

shaders/CMakeLists.txt
zalo: Done!

romainguy (approved):

romainguy:
Thanks!

PR #5942, opened 2022-08-18
Set Frame Delay according to Monitor Refresh Rate

zalo:
This is a simple PR that changes FilamentApp to delay its rendering to be roughly in line with the monitor framerate.

This helps all the samples look temporally smooth on higher refresh rate monitors, while still leaving it clear how to decouple rendering from vsync for new programmers.

Alternative techniques (with finer than 1ms granularity) like using `SDL_GL_SetSwapInterval` or `SDL_RENDERER_PRESENTVSYNC` might be preferable, but I'm still too new to SDL to figure out how to integrate them.    Even with integer milliseconds, the beat-frequency (rate of frame stutters) will never exceed ~1hz~ 0.5hz, so the incremental improvement will be minimal.

### DIFF
--- a/libs/filamentapp/src/FilamentApp.cpp
+++ b/libs/filamentapp/src/FilamentApp.cpp
@@ -395,7 +395,7 @@ void FilamentApp::run(const Config& config, SetupCallback setupCallback,
         SDL_DisplayMode Mode;
         int refreshIntervalMS = (SDL_GetDesktopDisplayMode(
             SDL_GetWindowDisplayIndex(window->mWindow), &Mode) == 0 && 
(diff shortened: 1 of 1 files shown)


romainguy (approved):
Instead of using swap interval or a delay I'd rather do what we do on Android and use vsync events. We've always wanted to do it as a cross platform API inside filament but desktop isn't our priority.

bejado (approved):

zalo:
That makes sense; though I’m not sure what the best way to add them in SDL is.

It also just occurred to me that this change doesn’t take into account the time it takes for the application to run/render.  Might need to add some extra timing logic…

PR #5943, opened 2022-08-18
cleanup SamplerBindingMap usages

pixelflinger:
- remove unused SamplerBindingMap parameter 
  In fact, the SamplerBindingMap is used, but it is passed inside the
  MaterialInfo structure, so it doesn't need to be passed as parameter.

- Don't create unneeded SamplerBindingMap temporaries


prideout (approved):
Nice.  Also, it might be possible to consolidate the two `populate()` calls into `prepareToBuild`.

bejado (approved):

PR #5950, opened 2022-08-18
lots of cleanup around materials and uniforms

pixelflinger:
reviewers: another PR that will be a lot easier to review each CL separately

The biggest change is that we are now storing the uniform block bindings inside the material file, instead of hardcoding them in two places. This info is only needed for GLES3.0/GL4.1 though.

Another big change is that we keep only a single entry point for UniformInterfaceBlock::add


prideout (approved):
LGTM if moving the items in RELEASE_NOTES is intentional.  Also, is there a reason to call `operator=()` explicitly instead of just doing `a = b`?

Issue #5859, opened 2022-08-01
Lighting singularity in web viewer


echadwick-wayfair:
**Describe the bug**
Web viewer shows a strange lighting singularity.

**To Reproduce**
1. https://google.github.io/filament/viewer/
2. Drag-and-drop the Duck sample model (https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/1.0/Duck)

**Expected behavior**
The lighting should be smooth and continuous across the surface. However it is showing a singularity.

**Screenshots**
![image](https://user-images.githubusercontent.com/40177036/182210443-8ac279d4-57b6-4593-a3fe-a2dbfebd97d1.png)

**Desktop (please complete the following information):**
 - OS: Windows 10
 - GPU: NVIDIA RTX 2070 laptop
 - Browser: Chrome Version 103.0.5060.134 (Official Build) (64-bit)

romainguy:
See https://github.com/google/filament/discussions/5824

It looks like a Chrome bug with DirectX on Windows :(

romainguy:
See https://bugs.chromium.org/p/chromium/issues/detail?id=1348363

BStringhamVRSK:
@romainguy 
I'm looking for a possible workaround for this problem until it gets fixed.
Is it possible, from within our program, to switch the Chrome backend from DirectX to OpenGL?  Or, does that require the user to manually set it?  It is probably untenable to ask our users to do the manual switch.

prideout:
I don't think it's possible to programmatically switch Chrome's backend, especially since the GPU process is shared by all tabs.  There might some other workaround you could try though, e.g. with view settings or matc optimization flags.

prideout:
I can reproduce this even without using our glTF ubershaders, e.g. the parquet shader ball looks like this.

Next I'll try playing with the opt flags.

<img width="1024" alt="windows" src="https://user-images.githubusercontent.com/1288904/185994460-86a20bf3-9638-4c8c-b9ab-23bc56b968ff.png">

echadwick-wayfair:
Anisotropy for free?  ;)

PR #5975, opened 2022-08-23
Add CONFIG for MaxInstanceCount.

prideout:


prideout:
This can be used to work around #5859.

pixelflinger (approved):

PR #5978, opened 2022-08-23
record sampler informations in the material file

pixelflinger:
The main goal of this change is to have "ugly" code only in filamat.
In particular, the sampler binding information is now recorded
into the material file and used on the filament side to inform the
backend.

This means that on the filament side we only have generic code with
all the "uglyness" in filamat. In particular SamplerBindingMap and 
SibGenerator now only exist in filamat.


prideout (commented):
Does matdbg still work?  ><

RELEASE_NOTES.md
```diff
@@ -13,6 +13,7 @@ A new header is inserted each time a *tag* is created.
 - engine: `clipSpaceTransform` is now only available with `vertexDomain:Device` [⚠️ **API Change**]
 - gltfio: add unified `AssetLoader::createAsset()` method [⚠️ **API Change**]
 - gltfio: all assets are now "instanced" [⚠️ **API Change**]
+- engine: rework material/shader sampler binding code [⚠️ **Recompile Materials**]
```
bejado, line 16: Please put under "main branch"

filament/src/MaterialParser.cpp
```diff
@@ -486,4 +498,48 @@ bool ChunkUniformBlockBindings::unflatten(filaflat::Unflattener& unflattener,
...
 
 
+bool ChunkSamplerBlockBindings::unflatten(Unflattener& unflattener,
+        SamplerGroupBindingInfoList* pSamplerGroupBindingInfoList,
+        SamplerBindingToNameMap* pSamplerBindingToNameMap) {
+    assert_invariant(pSamplerGroupBindingInfoList && pSamplerBindingToNameMap);
+    SamplerGroupBindingInfoList& samplerGroupBindingInfoList = *pSamplerGroupBindingInfoList;
+    SamplerBindingToNameMap& samplerBindingToNameMap = *pSamplerBindingToNameMap;
+
+    uint8_t count;
```
bejado, line 508: Maybe rename to `groupCount` for readability (and to avoid confusion with `SamplerGroupBindingSize::count`). I would also use a separate `count` variable below instead of re-using this one.

bejado (approved):

pixelflinger:
> I think we had considered doing it this way long ago, but I can't remember why we didn't do this in the first place.

Actually it used to work more or less like that, and I'm the one who changed it :-( It was misguided.

> Anyway it LGTM if it works on all backends. :) It's nice that SibGenerator and UibGenerator can live in the same folder again.

And more importantly, they won't pollute filament's binary with extra code and data.

PR #5984, opened 2022-08-24
use separate enums for uniform and sampler binding points

pixelflinger:


bejado (approved):

pixelflinger:
reviewers, I updated the PR with changes to backends.

PR #5999, opened 2022-08-26
Implement setThreadName() on Windows

veghbalazs:
This PR completes a TODO, thread names are now set on Windows too (e.g. `FEngine::loop`, `JobSystem::loop`).

Windows API requires conversion of input string to UTF-16. I assumed that the input `const char*` may not only contain ASCII characters, but also can be a UTF-8 string.

filament-threadname


bejado (approved):

bejado:
Thanks!

romainguy (approved):

PR #6001, opened 2022-08-26
Specification constants support

pixelflinger:


filament/backend/include/backend/Program.h
```diff
@@ -84,6 +84,15 @@ class Program {
     Program& setSamplerGroup(size_t bindingPoint, ShaderStageFlags stageFlags,
             Sampler const* samplers, size_t count) noexcept;
 
+    struct SpecializationConstant {
+        uint32_t id;                                // id set in glsl, needed for vulkan
```
bejado, line 88: both Vulkan and Metal yes?

filament/backend/src/metal/MetalHandles.mm
```diff
@@ -372,7 +372,22 @@ BufferDescriptor b(nullptr, 0u, [](void* buffer, size_t size, void* user) {
...
+            const std::array<MTLDataType, 3> types{
+                MTLDataTypeInt, MTLDataTypeFloat, MTLDataTypeBool };
+            std::visit([&sc, constants, type = types[sc.value.index()]](auto&& arg) {
+                [constants setConstantValue:&arg
+                                       type:type
+                                    atIndex:sc.id];
+            }, sc.value);
+        }
+
+        *shaderFunctions[i] = [library newFunctionWithName:@"main0"
```
bejado, line 387: nit: align the `:` on each line

filament/backend/include/backend/Program.h
pixelflinger: yes thanks

prideout (approved):
Looks great.  Our only usage of std::variant, I think.

PR #6005, opened 2022-08-29
fix uninitialized variable access

pixelflinger:
this caused a vk validation error


romainguy:
Already fixed in https://github.com/google/filament/pull/6004/

PR #6007, opened 2022-08-29
Add JNI for TransformManager.getChildCount(), TransformManager.getChildren() and Scene.hasEntity()

ThomasGorisse:


android/filament-android/src/main/cpp/Scene.cpp
```diff
@@ -83,3 +83,10 @@ Java_com_google_android_filament_Scene_nGetLightCount(JNIEnv *env, jclass type,
     Scene* scene = (Scene*) nativeScene;
     return (jint) scene->getLightCount();
 }
+
+extern "C" JNIEXPORT jboolean JNICALL
+Java_com_google_android_filament_Scene_nHasEntity(JNIEnv *env, jclass type, jlong nativeScene,
+        jint entity) {
+    Scene* scene = (Scene*) nativeScene;
+    return (jboolean) scene->hasEntity((Entity &) entity);
```
romainguy, line 91: Why are you casting to a reference? Entities should always been values.

android/filament-android/src/main/cpp/TransformManager.cpp
```diff
@@ -110,6 +110,26 @@ Java_com_google_android_filament_TransformManager_nGetParent(JNIEnv*, jclass,
...
+    TransformManager* tm = (TransformManager*) nativeTransformManager;
+    return tm->getChildCount((TransformManager::Instance) i);
+}
+
+extern "C" JNIEXPORT void JNICALL
+Java_com_google_android_filament_TransformManager_nGetChildren(JNIEnv* env,
+        jclass, jlong nativeTransformManager, jint i,
+        jintArray outEntities_, jint count) {
+    TransformManager* tm = (TransformManager*) nativeTransformManager;
+    jint* entities = env->GetIntArrayElements(outEntities_, NULL);
```
romainguy, line 125: Use nullptr instead of NULL

android/filament-android/src/main/java/com/google/android/filament/TransformManager.java
```diff
@@ -199,6 +200,33 @@ public int getParent(@EntityInstance int i) {
...
+    /**
+     * Gets a list of children for a transform component.
+     *
+     * @param i the {@link EntityInstance} of the transform component to get the children from.
+     * @param count The maximum number of children to retrieve.
+     * @return Array of retrieved children {@link Entity}.
+     */
+    @Entity
+    @NonNull
+    public  int[] getChildren(@EntityInstance int i, @IntRange(from = 1) int count) {
```
romainguy, line 223: Nit: remove extra space after `public`.

There should also be an API that takes an int[] passed by the caller so allocations can be avoided. This method can call the other one then, passing a `new int[]`.

romainguy (changes requested):

android/filament-android/src/main/java/com/google/android/filament/TransformManager.java
```diff
@@ -199,6 +200,46 @@ public int getParent(@EntityInstance int i) {
...
+     * @param outEntities array to receive the result sized to the maximum number of children to
+     *                    retrieve
+     */
+    @Entity
+    @NonNull
+    public void getChildren(@EntityInstance int i, @Size(min = 1) int[] outEntities) {
+        if (outEntities.length < 1) throw new ArrayIndexOutOfBoundsException("outEntities must be at least 1");
+        nGetChildren(mNativeObject, i, outEntities, outEntities.length);
+    }
+
```
pixelflinger, line 242: I think these API should follow the pattern we've used elsewhere. That is, you only have the version that takes `outEntitites` and if it is null then you allocate it.

pixelflinger (changes requested):

android/filament-android/src/main/java/com/google/android/filament/TransformManager.java
ThomasGorisse:
```suggestion
    /**
     * Gets a list of children for a transform component.
     *
     * @param i the {@link EntityInstance} of the transform component to get the children from.
     * @return Array of retrieved children {@link Entity}.
     */
    public @Entity @NonNull int[] getChildren(@EntityInstance int i) {
        return getChildren(i, null);
    }

    /**
     * Gets a list of children for a transform component.
     *
     * @param i        the {@link EntityInstance} of the transform component to get the children
     *                 from.
     * @param entities array to receive the result sized to the maximum number of children to
     *                 retrieve.
     */
    public @Entity @NonNull int[] getChildren(@EntityInstance int i, @Nullable @Size(min = 1) int[] entities) {
        if (entities == null) {
            entities = new int[getChildCount(i)];
        }
        if (entities.length > 0) {
            nGetChildren(mNativeObject, i, entities, entities.length);
        }
        return entities;
    }

```

android/filament-android/src/main/java/com/google/android/filament/TransformManager.java
```diff
@@ -199,6 +200,45 @@ public int getParent(@EntityInstance int i) {
...
+    /**
+     * Gets a list of children for a transform component.
+     *
+     * @param i the {@link EntityInstance} of the transform component to get the children from.
+     * @return Array of retrieved children {@link Entity}.
+     */
+    public @Entity @NonNull int[] getChildren(@EntityInstance int i) {
+        return getChildren(i, null);
+    }
+
```
pixelflinger, line 223: I would just remove this version, just to match what we do elsewhere.

pixelflinger (changes requested):

pixelflinger (approved):

bejado (approved):

PR #6013, opened 2022-08-30
Fix filamat internal state tracking

pixelflinger:
in a few places filamat loses the value of targetApi and targetLanguage
and attempt to guess it from other values, this lead to inconsistencies.

we now keep both targetApi and targetLanguage and use them appropriately.

in particular we don't use the optimization level to "guess" what 
targetLangage we're on.

In the end this resolve a prior unit test failure and allows us to not
have to use "spirv" rules when generating only GLSL.


pixelflinger:
reviewers, the first commit is a reformat so it's easier to ignore it during the reivew.

prideout (approved):

PR #6016, opened 2022-08-30
The default arena sizes in the makefile didn't match the code

pixelflinger:
Also added some ascii art to explain better how the various arenas are
used.


romainguy (approved):

prideout (commented):
The new values seem to break with the some advice that says: "As a rule of thumb use the same value as FILAMENT_PER_FRRAME_COMMANDS_SIZE_IN_MB".   Is this advice still valid?

PR #6017, opened 2022-08-30
gltfio cleanup: NodeMap is now a FixedCapacityVector, etc.

prideout:


libs/gltfio/src/AssetLoader.cpp
```diff
@@ -1420,9 +1419,9 @@ void FAssetLoader::importSkins(FFilamentAsset* primary,
     const auto& nodeMap = instance->nodeMap;
     for (cgltf_size i = 0, len = gltf->nodes_count; i < len; ++i) {
         const cgltf_node& node = gltf->nodes[i];
+        Entity entity = nodeMap[i];
```
bejado, line 1422: You mention that `nodeMap` could be sparsely populated. Do you need to verify that `entity` is valid?

bejado (approved):

libs/gltfio/src/AssetLoader.cpp
prideout: Wow excellent catch, in some situations (e.g. multiple gltf scenes in one gltf file) this might be a zero entity, so we need an `if` here.  Will fix.

bejado: 👍

PR #6018, opened 2022-08-30
Add overdraw visualization to gltf_viewer

bejado:
Screen Shot 2022-08-30 at 6 01 03 PM

When enabled, overdraw fragments are colored:
- blue: 1 time overdrawn
- green: 2 times overdrawn
- magenta: 3 times overdrawn
- red: 4+ times overdrawn


pixelflinger (approved):

pixelflinger:
This doesn't seem to work with `-a opengl` nor `-a vulkan`. In the opengl case, the overdraw is not drawn at all, but the model works. In the vulkan case, nothing draws.

bejado:
> This doesn't seem to work with `-a opengl` nor `-a vulkan`. In the opengl case, the overdraw is not drawn at all, but the model works. In the vulkan case, nothing draws.

Hm, I'll take a look. I thought I tested with OpenGL.

PR #6019, opened 2022-08-30
cleanup Froxelizer comments

pixelflinger:
also fix typos, formatting.


filament/src/Froxelizer.cpp
```diff
@@ -49,41 +49,57 @@ static constexpr size_t FROXEL_SLICE_COUNT = 16;
...
 
+// Max number of froxels limited by:
+//   - max texture size [min 2048]
+//   - chosen texture width [64]
+//
+// Also, increasing the number of froxels adds more pressure on the "record buffer" which stores
+// the light indices per froxel. The record buffer is limited to min(16K[ubo], 64K[uint16]) entries,
+// so with 8192 froxels, we can store 2 lights per froxels assuming they're all used. In practice, some
+// froxels are not used, so we can store more.
+constexpr size_t FROXEL_BUFFER_ENTRY_COUNT = 8162;
```
bejado, line 60: Should this be `8192`?

filament/src/Froxelizer.cpp
```diff
@@ -49,41 +49,57 @@ static constexpr size_t FROXEL_SLICE_COUNT = 16;
...
 // number of lights processed by one group (e.g. 32)
 static constexpr size_t LIGHT_PER_GROUP = sizeof(Froxelizer::LightGroupType) * 8;
 
 // number of groups (i.e. jobs) to use for froxelization (e.g. 8)
 static constexpr size_t GROUP_COUNT =
         (CONFIG_MAX_LIGHT_COUNT + LIGHT_PER_GROUP - 1) / LIGHT_PER_GROUP;
 
+// This depends on the maximum number of lights (currently 255), and can't be more than 16 bits.
+static_assert(CONFIG_MAX_LIGHT_INDEX <= std::numeric_limits<uint8_t>::max(),
+        "can't have more than 256 lights");
```
bejado, line 89: Is the max 255 or 256?

filament/src/Froxelizer.cpp
pixelflinger: lol. yes. thanks.

filament/src/Froxelizer.cpp
pixelflinger: it was 256 with max index being 255. I fixed the comment.

PR #6022, opened 2022-08-31
graphvizify should only be implemented on debug builds

pixelflinger:


romainguy (approved):

filament/src/fg/PassNode.cpp
```diff
@@ -312,6 +313,9 @@ utils::CString RenderPassNode::graphvizify() const noexcept {
     s.append("]");
 
     return utils::CString{ s.c_str() };
+#else
+    retrun {};
```
bejado, line 317:
```suggestion
    return {};
```

PR #6030, opened 2022-09-01
check that user materials don't exceed their allowed features

pixelflinger:


filament/backend/include/backend/DriverEnums.h
```diff
@@ -48,11 +48,18 @@ static constexpr uint64_t SWAP_CHAIN_CONFIG_ENABLE_XCB          = 0x4;
 static constexpr uint64_t SWAP_CHAIN_CONFIG_APPLE_CVPIXELBUFFER = 0x8;
 
 static constexpr size_t MAX_VERTEX_ATTRIBUTE_COUNT  = 16;   // This is guaranteed by OpenGL ES.
-static constexpr size_t MAX_VERTEX_SAMPLER_COUNT    = 16;   // This is guaranteed by OpenGL ES.
-static constexpr size_t MAX_FRAGMENT_SAMPLER_COUNT  = 16;   // This is guaranteed by OpenGL ES.
 static constexpr size_t MAX_SAMPLER_COUNT           = 32;   // This is guaranteed by OpenGL ES.
 static constexpr size_t MAX_VERTEX_BUFFER_COUNT     = 16;   // Max number of bound buffer objects.
 
+static constexpr struct {
```
bejado, line 54: Maybe rename to`FEATURE_LEVEL_CAPS` and add a comment on how it should be indexed? This wasn't clear to me at first.

libs/filamat/src/MaterialBuilder.cpp
```diff
@@ -941,10 +947,11 @@ MaterialBuilder& MaterialBuilder::useLegacyMorphing() noexcept {
 }
 
 Package MaterialBuilder::build(JobSystem& jobSystem) noexcept {
+    bool success;
     if (materialBuilderClients == 0) {
-        utils::slog.e << "Error: MaterialBuilder::init() must be called before build()."
-            << utils::io::endl;
+        slog.e << "Error: MaterialBuilder::init() must be called before build()." << io::endl;
         // Return an empty package to signal a failure to build the material.
+error:
```
bejado, line 954: I think simply returning `Package::invalidPackage()` at each error is less bug-prone compared to a goto.

libs/filamat/src/MaterialBuilder.cpp
```diff
@@ -1006,6 +1027,69 @@ Package MaterialBuilder::build(JobSystem& jobSystem) noexcept {
...
+            slog.e << "\"" << sampler.name.c_str() << "\" "
+                    << Enums::toString(sampler.type).c_str() << " " << stage << '\n';
+        }
+        flush(slog.e);
+    };
+
+    const auto userSamplerCount = info.sib.getSize();
+    switch (info.featureLevel) {
+        case FeatureLevel::FEATURE_LEVEL_1: {
+            // TODO: we need constants somewhere for these values
```
bejado, line 1056: Could we use `PerViewSib::SAMPLER_COUNT` to derive this? Is this number still accurate if morphing is used?

filament/backend/src/metal/MetalDriver.mm
```diff
@@ -1354,15 +1354,15 @@
         return mContext->samplerStateCache.getOrCreateState(s);
     };
 
-    id<MTLTexture> texturesToBindVertex[MAX_VERTEX_SAMPLER_COUNT] = {};
-    id<MTLSamplerState> samplersToBindVertex[MAX_VERTEX_SAMPLER_COUNT] = {};
+    id<MTLTexture> texturesToBindVertex[FEATURE_CAPS[1].MAX_VERTEX_SAMPLER_COUNT] = {};
```
bejado, line 1357: Use `FEATURE_CAPS[+FeatureLevel::FEATURE_LEVEL_1]` to be consistent with SamplerBindingMap.cpp?

pixelflinger: done

libs/filamat/src/MaterialBuilder.cpp
pixelflinger: We could say that it's e.g.: `16 - "what we need in filament"` but I think it would be wrong to do it this way. I think it should be the hardcoded number that is in our (non existant) docs. Because we could make mistakes in `PerViewSib` for instance.

This number should be accurate "by definition". In practice it is because morphing doesn't count against the "fragment" sampler count.

PR #6044, opened 2022-09-06
gltfio: fix crash when material is  unspecified

prideout:
I moved the `inputMat = inputMat ? inputMat : &kDefaultMat`.  Wups.


prideout:
Wrong fix, closing.

PR #6045, opened 2022-09-06
gltfio: fix crash when material is unspecified

prideout:
The recent introduction of index-based mapping between cgltf objects and Filament objects was broken for the case where the model omits a material specification (which is a valid thing to do).

This fix also refactors some hodgepodge data & logic into a proper class called `MaterialInstanceCache` (which is internal only).


prideout:
I did not update the release notes because this will be cherry-picked into the RC branch, which is where the regression was introduced.

bejado (approved):
I'll cherry-pick this one too

PR #6046, opened 2022-09-07
gltfio: do not clear out the texture slots too early.

prideout:
This broke asyncGetLoadProgress() and caused WebGL to crash reliably because ResourceLoader got destroyed too soon.

Bug was introduced with de7dfc2ea6dce02f9c8d8002d04df72f39fb8e0d.

I intend to cherry pick this to rc/1.27.0, which is where it was introduced, so there's no need to update the release notes.


bejado (approved):

bejado:
I'll go ahead and cherry-pick to rc/1.27.0 :)

pixelflinger (approved):

PR #6050, opened 2022-09-07
gltfio MaterialProvider: add new method for getting `Material`

prideout:
gltfio can now ask the material provider plugin which `Material` would be used for a given set of requirements.  Prior to this change, gltfio could only create a new `MaterialInstance`.

This method is necessary to support an upcoming gltfio feature.

### DIFF
--- a/android/gltfio-android/src/main/cpp/AssetLoader.cpp
+++ b/android/gltfio-android/src/main/cpp/AssetLoader.cpp
@@ -42,6 +42,7 @@ class JavaMaterialProvider : public MaterialProvider {
 
     jmethodID mMaterialKeyConstructor;
     jmethodID mCreateMaterialInstance;
+    jmethodID mGetMaterial;
     jmethodID mGetMaterials;
     jmethodID mNeedsDummyData;
     jmethodID mDestroyMaterials;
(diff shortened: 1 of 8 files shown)


android/gltfio-android/src/main/cpp/AssetLoader.cpp
```diff
@@ -139,6 +144,49 @@ class JavaMaterialProvider : public MaterialProvider {
...
+        jobject javaKey = mEnv->NewObject(mMaterialKeyClass, mMaterialKeyConstructor);
+
+        auto& helper = MaterialKeyHelper::get();
+        helper.copy(mEnv, javaKey, *config);
+
+        // Convert the optional label into a Java string.
+        jstring stringLabel = label ? mEnv->NewStringUTF(label) : nullptr;
+
+        // Allocate space for the output argument.
+        jintArray uvMapArray = mEnv->NewIntArray(8);
```
bejado, line 158: Where does `8` come from? Should we check that `uvmap->size()` isn't greater?

prideout: Nice catch, the 8 is fixed but I should not use a magic number here.  Fixed.

bejado (approved):

PR #6051, opened 2022-09-07
gltfio: big simplification for textures + add sRGB related assert.

prideout:
`FilamentAsset` previously had three confusing fields related to textures: mTextures, mTextureSlots, and mTextureBindings.

These are now consolidated into a single field, which simply has one top level array item per `cgltf_texture`.

The memory footprint is smaller because we no longer bother to store `TextureSampler` objects, instead we simply defer their construction until calling setParameter.

Last but not least, we now assert in debug builds if the usages of a particular texture have inconsistent sRGB flags. In the past, inconsistent sRGB would be silently ignored.

In other words, we will now assert if you try to use the same texture for `baseColor` and `normalMap`, whereas in the past we would exhibit non-deterministic behavior in this situation (with respect to sRGB semantics).

### DIFF
--- a/libs/gltfio/src/AssetLoader.cpp
+++ b/libs/gltfio/src/AssetLoader.cpp
@@ -243,7 +243,7 @@ public:
     FNodeManager mNodeManager;
 
     // Transient state used only for the asset currently being loaded:
-    FFilamentAsset* mResult;
+    FFilamentAsset* mAsset;
     tsl::robin_map<cgltf_node*, SceneMask> mRootNodes;
     const char* mDefaultNodeName;
     bool mError = false;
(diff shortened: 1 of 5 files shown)


libs/gltfio/src/AssetLoader.cpp
```diff
@@ -1226,7 +1224,7 @@ MaterialInstance* FAssetLoader::createMaterialInstance(const cgltf_data* srcAsse
     }
 
     if (matkey.hasBaseColorTexture) {
-        addTextureBinding(mi, "baseColorMap", baseColorTexture.texture, true);
+        mResult->addTextureBinding(mi, "baseColorMap", baseColorTexture.texture, true);
```
pixelflinger, line 1227: `mResult` is kind of a weird name for a class attribute. `mAsset` maybe?

pixelflinger (approved):

libs/gltfio/src/AssetLoader.cpp
prideout: Yes good idea, added a commit.

PR #6054, opened 2022-09-09
Vk: do not provide descriptors for unused samplers.

prideout:


filament/backend/src/vulkan/VulkanBlitter.cpp
```diff
@@ -376,7 +376,9 @@ void VulkanBlitter::blitSlowDepth(VkImageAspectFlags aspect, VkFilter filter,
         .imageLayout = getDefaultImageLayout(TextureUsage::DEPTH_ATTACHMENT)
     };
 
-    mPipelineCache.bindSamplers(samplers);
+    VulkanPipelineCache::UsageFlags usageFlags;
+    VulkanPipelineCache::setUsageFlags(&usageFlags, 0, ShaderStageFlags::FRAGMENT);
```
pixelflinger, line 380: shouldn't that be called `getUsageFlags`?

filament/backend/src/vulkan/VulkanPipelineCache.cpp
```diff
@@ -35,7 +35,7 @@ namespace filament::backend {
 
 static VulkanPipelineCache::RasterState createDefaultRasterState();
 
-static VkShaderStageFlags getShaderStageFlags(utils::bitset64 key, uint16_t binding) {
+static VkShaderStageFlags getShaderStageFlags(VulkanPipelineCache::UsageFlags key, uint16_t binding) {
     VkShaderStageFlags flags = 0;
     if (key.test(binding * 2 + 0)) {
         flags |= VK_SHADER_STAGE_VERTEX_BIT;
```
pixelflinger, line 41: this code makes me nervous for when we'll need to add a COMPUTE stage. Is there a way to make this less easy to miss?

filament/backend/src/vulkan/VulkanPipelineCache.cpp
```diff
@@ -46,21 +46,13 @@ static VkShaderStageFlags getShaderStageFlags(utils::bitset64 key, uint16_t bind
...
-        const auto& stageFlags = samplerGroupInfo[binding].stageFlags;
-        for (const auto& sampler : samplerGroupInfo[binding].samplers) {
-            if (any(stageFlags & ShaderStageFlags::VERTEX)) {
-                key.set(sampler.binding * 2 + 0);
-            }
-            if (any(stageFlags & ShaderStageFlags::FRAGMENT)) {
-                key.set(sampler.binding * 2 + 1);
-            }
-        }
+void VulkanPipelineCache::setUsageFlags(UsageFlags* key, uint16_t binding, ShaderStageFlags flags) {
```
pixelflinger, line 49: I think it would be better to call it `getUsageFlags` and have it return a `UsageFlags` that you *or* with the one you already have.

filament/backend/src/vulkan/VulkanPipelineCache.cpp
prideout: Let me add an assert in `setUsageFlags` (soon to be renamed) if any flags get passed in that are not VERTEX or FRAGMENT.  I will also add an assert that looks at PIPELINE_STAGE_COUNT, near the definition of PipelineLayoutKey, in case we need to change the bitset64 into something else.

prideout: Ah, I can make this code less brittle if I used MAX_FRAGMENT_SAMPLER_COUNT and MAX_FRAGMENT_SAMPLER_COUNT.

filament/backend/src/vulkan/VulkanPipelineCache.cpp
prideout: Pushed a new commit for this.

pixelflinger (approved):

PR #6058, opened 2022-09-09
scissor improvements

pixelflinger:
- improve scissor documentation, specify coordinate system and clipping behavior
- scissor now takes the guard bands into account
- on the backend side the scissor is no longer clipped to the viewport
- clipping to the viewport is now done on the filament side
- don't set scissor when material instance doesn't specify one (common case)

### DIFF
--- a/filament/backend/src/metal/MetalDriver.mm
+++ b/filament/backend/src/metal/MetalDriver.mm
@@ -1321,18 +1321,12 @@ void MetalDriver::draw(PipelineState ps, Handle<HwRenderPrimitive> rph, uint32_t
     // Set scissor-rectangle.
     // In order to do this, we compute the intersection between:
     //  1. the scissor rectangle
-    //  2. the current viewport
-    //  3. the render target attachment dimensions (important, as the scissor can't be set larger)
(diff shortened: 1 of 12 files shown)


prideout (approved):

filament/backend/src/metal/MetalDriver.mm
```diff
@@ -1329,21 +1329,16 @@
...
     const auto attachmentSize = mContext->currentRenderTarget->getAttachmentSize();
     const float aleft = 0.0f, atop = 0.0f;
     const float aright = static_cast<float>(attachmentSize.x);
     const float abottom = static_cast<float>(attachmentSize.y);
 
-    const auto left   = std::fmax(std::fmax(sleft, vleft), aleft);
-    const auto right  = std::fmin(std::fmin(sright, vright), aright);
-    const auto top    = std::fmax(std::fmax(stop, vtop), atop);
-    const auto bottom = std::fmin(std::fmin(sbottom, vbottom), abottom);
+    const auto left   = std::fmax(sleft, aleft);
```
bejado, line 1338: Could you also update the comment on lines 1321-1327 to reflect this?

bejado (approved):

PR #6062, opened 2022-09-12
Add return case to fix G3 compiler error

bejado:
We need a return case here, otherwise the code implicitly falls-through in release builds which generates a compiler error in Google3.


prideout (approved):
Thanks Ben!

PR #6065, opened 2022-09-12
gltfio: fix regression when textures are "shared"

prideout:
Some (poorly authored) glTF assets have several `image` elements that all refer to the same URL or buffer view. When this occurs, we create only 1 Filament Texture.

These assets regressed after #6051, which consolidated the texture-related fields in `FilamentAsset`, but did not include an ownership flag in the new `TextureInfo` struct.

### DIFF
--- a/libs/gltfio/src/FilamentAsset.cpp
+++ b/libs/gltfio/src/FilamentAsset.cpp
@@ -84,7 +84,9 @@ FFilamentAsset::~FFilamentAsset() {
         mEngine->destroy(ib);
     }
     for (auto tx : mTextures) {
-        mEngine->destroy(tx.texture);
+        if (UTILS_LIKELY(tx.isOwner)) {
+            mEngine->destroy(tx.texture);
+        }
     }
     for (auto tb : mMorphTargetBuffers) {
         mEngine->destroy(tb);
--- a/libs/gltfio/src/ResourceLoader.cpp
+++ b/libs/gltfio/src/ResourceLoader.cpp
@@ -63,6 +63,13 @@ using FilepathTextureCache = tsl::robin_map<std::string, Texture*>;
 using UriDataCache = tsl::robin_map<std::string, gltfio::ResourceLoader::BufferDescriptor>;
 using TextureProviderList = tsl::robin_map<std::string, TextureProvider*>;
 
+enum class CacheResult {
+    ERROR,
+    NOT_READY,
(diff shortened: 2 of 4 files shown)


libs/gltfio/src/FilamentAsset.cpp
```diff
@@ -84,7 +84,9 @@ FFilamentAsset::~FFilamentAsset() {
         mEngine->destroy(ib);
     }
     for (auto tx : mTextures) {
-        mEngine->destroy(tx.texture);
+        if (UTILS_LIKELY(tx.shared == false)) {
+            mEngine->destroy(tx.texture);
+        }
```
pixelflinger, line 89: how do we know that other ones don't have `shared` to `true`?

libs/gltfio/src/ResourceLoader.cpp
```diff
@@ -669,22 +674,30 @@ void ResourceLoader::Impl::cancelTextureDecoding() {
...
-            if (Texture* texture = getOrCreateTexture(asset, textureIndex, info.flags)) {
-                asset->applyTextureBinding(textureIndex, slot);
+
+    // Create new texture objects if they are not cached and kick off decoding jobs.
+    for (size_t textureIndex = 0, n = asset->mTextures.size(); textureIndex < n; ++textureIndex) {
+        FFilamentAsset::TextureInfo& info = asset->mTextures[textureIndex];
+        auto [texture, cacheResult] = getOrCreateTexture(asset, textureIndex, info.flags);
+        if (texture == nullptr) {
+            if (cacheResult == CacheResult::NOT_READY) {
+                mRemainingTextureDownloads++;
```
pixelflinger, line 685: that's single threaded correct?

prideout: Yes all reads and writes to `mRemainingTextureDownloads` occur on the same thread.

libs/gltfio/src/FilamentAsset.cpp
prideout: Because we know a priori that it's set up that way.  I had considered calling this field `isOwner` maybe that's a better name?

prideout: I changed this field to `isOwner`, I think it's more clear now.

pixelflinger (approved):

PR #6073, opened 2022-09-15
Vk getFeatureLevel() should use `maxPerStageDescriptorSamplers`

prideout:


filament/backend/src/vulkan/VulkanDriver.cpp
```diff
@@ -885,19 +885,18 @@ bool VulkanDriver::isWorkaroundNeeded(Workaround workaround) {
...
-            return FeatureLevel::FEATURE_LEVEL_3;
-        }
+    const auto& fl2 = FEATURE_LEVEL_CAPS[+FeatureLevel::FEATURE_LEVEL_2];
+    if (fl2.MAX_VERTEX_SAMPLER_COUNT > limits.maxPerStageDescriptorSamplers ||
+        fl2.MAX_FRAGMENT_SAMPLER_COUNT > limits.maxPerStageDescriptorSamplers) {
         return FeatureLevel::FEATURE_LEVEL_2;
     }
-    return FeatureLevel::FEATURE_LEVEL_1;
+
+    return FeatureLevel::FEATURE_LEVEL_3;
```
pixelflinger, line 899: I don't think the condition for fl3 is correct. It never checks against the fl3 values.

pixelflinger (changes requested):

filament/backend/src/vulkan/VulkanDriver.cpp
prideout: Doh!  Fixed.

pixelflinger (approved):

PR #6074, opened 2022-09-15
basic backend support for compute

pixelflinger:
- added api to dispatch a compute shader
- added api to create and bind a ssbo
- added api to read back a buffer

Only implemented in the gl backend

### DIFF
--- a/filament/backend/CMakeLists.txt
+++ b/filament/backend/CMakeLists.txt
@@ -450,3 +450,23 @@ if (APPLE AND NOT IOS)
     set_target_properties(backend_test_mac PROPERTIES FOLDER Tests)
 endif()
 
+# ==================================================================================================
+# Compute tests
+
+if (NOT IOS AND NOT WEBGL)
+
+add_executable(compute_test
+        test/ComputeTest.cpp
+        test/Arguments.cpp
+        test/test_ComputeBasic.cpp
+        )
+
+target_link_libraries(compute_test PRIVATE
+        backend
+        getopt
+        gtest
+        )
+
+set_target_properties(compute_test PROPERTIES FOLDER Tests)
+
+endif()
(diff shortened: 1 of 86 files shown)


pixelflinger:
this is a FYI PR, but comments are welcome.

filament/backend/CMakeLists.txt
```diff
@@ -450,3 +450,19 @@ if (APPLE AND NOT IOS)
     set_target_properties(backend_test_mac PROPERTIES FOLDER Tests)
 endif()
 
+# ==================================================================================================
+# Compute tests
+
+add_executable(compute_test
```
bejado, line 456: Could we put these tests under the same `backend_test` target?

pixelflinger: they're more simple and have to run on Android -- backend_test seemed to do too much.

daemyung:
It's a so nice feature!

filament/backend/src/metal/MetalHandles.mm
```diff
@@ -424,6 +425,7 @@ BufferDescriptor b(nullptr, 0u, [](void* buffer, size_t size, void* user) {
             switch (type) {
                 case ShaderStage::VERTEX:    return vertexSamplerBlockInfo;
                 case ShaderStage::FRAGMENT:  return fragmentSamplerBlockInfo;
+                // FIXME: implement ShaderStage::COMPUTE
```
bejado, line 428: Let's add a `case` here, otherwise we generate a compiler error.

filament/backend/CMakeLists.txt
bejado, line 456: You may need to surround this with a `if (APPLE AND NOT IOS)` block to avoid building on iOS.

libs/filamat/src/GLSLPostProcessor.h
```diff
@@ -74,7 +75,8 @@ class GLSLPostProcessor {
         SpirvBlob* spirvOutput = nullptr;
         std::string* mslOutput = nullptr;
         EShLanguage shLang = EShLangFragment;
-        int langVersion = 0;
+        // use 100 for ES environment, 110 for desktop
+         int langVersion = 0;
```
bejado, line 79:
```suggestion
        int langVersion = 0;
```

bejado (approved):

PR #6087, opened 2022-09-19
Add `MetalRingBuffer` for future argument buffer management

bejado:
This helper class will be used soon to manage buffer allocations  for argument buffers.

### DIFF
--- a/filament/backend/src/metal/MetalBuffer.h
+++ b/filament/backend/src/metal/MetalBuffer.h
@@ -135,7 +135,10 @@ public:
         const auto occupiedSlots = mOccupiedSlots.load(std::memory_order_relaxed);
         assert_invariant(occupiedSlots <= mSlotCount);
         if (UTILS_UNLIKELY(occupiedSlots == mSlotCount)) {
-            // We don't have any room left, so we fall back to creating a one-off buffer.
+            // We don't have any room left, so we fall back to creating a one-off aux buffer.
+            // If we already have an aux buffer, it will get freed here, unless it has been retained
+            // by a MTLCommandBuffer. In that case, it will be freed when the command buffer
+            // finishes executing.
             mAuxBuffer = [mDevice newBufferWithLength:mSlotSizeBytes options:mBufferOptions];
             assert_invariant(mAuxBuffer);
             return {mAuxBuffer, 0};
--- a/filament/backend/src/metal/MetalBuffer.h
+++ b/filament/backend/src/metal/MetalBuffer.h
@@ -24,8 +24,11 @@
 
 #include <Metal/Metal.h>
 
-namespace filament {
-namespace backend {
+#include <utils/compiler.h>
+
+#include <tuple>
+
+namespace filament::backend {
 
 class MetalBuffer {
 public:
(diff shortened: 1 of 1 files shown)


filament/backend/src/metal/MetalBuffer.h
```diff
@@ -80,7 +83,105 @@ class MetalBuffer {
     MetalContext& mContext;
 };
 
-} // namespace backend
-} // namespace filament
+template <typename P>
```
pixelflinger, line 86: maybe rename `P` to `TYPE`

filament/backend/src/metal/MetalBuffer.h
```diff
@@ -80,7 +83,105 @@ class MetalBuffer {
...
+
+    /**
+     * Create a new allocation in the buffer.
+     * @param cmdBuffer When this command buffer has finished executing on the GPU, the previous
+     *                  ring buffer allocation will be freed.
+     * @return the id<MTLBuffer> and offset for the new allocation
+     */
+    std::tuple<id<MTLBuffer>, NSUInteger> createNewAllocation(id<MTLCommandBuffer> cmdBuffer) {
+        const auto occupiedSlots = mOccupiedSlots.load(std::memory_order_relaxed);
+        if (UTILS_UNLIKELY(occupiedSlots == mSlotCount)) {
```
pixelflinger, line 134: maybe use `>=` to be on the safe side (or assert we never are `>`)

filament/backend/src/metal/MetalBuffer.h
```diff
@@ -80,7 +83,105 @@ class MetalBuffer {
...
+            // We don't have any room left, so we fall back to creating a one-off buffer.
+            mAuxBuffer = [mDevice newBufferWithLength:mSlotSizeBytes options:mBufferOptions];
+            assert_invariant(mAuxBuffer);
+            return {mAuxBuffer, 0};
+        }
+        mCurrentSlot = (mCurrentSlot + 1) % mSlotCount;
+        mOccupiedSlots.fetch_add(1, std::memory_order_relaxed);
```
pixelflinger, line 141: I don't understand the concurrency logic here, it looks lije this method can be called from multiple threads, since `mOccupiedSlots` is atomic, but at the same time `mCurrentSlot` is used unprotected, so is `mAuxBuffer`.

bejado: `createNewAllocation` is only ever called from the driver thread. The completion handler though (on line 148) is invoked by Metal and called from an arbitrary thread.

pixelflinger: yes I saw that later :-)

filament/backend/src/metal/MetalBuffer.h
```diff
@@ -80,7 +83,105 @@ class MetalBuffer {
...
+     * Create a new allocation in the buffer.
+     * @param cmdBuffer When this command buffer has finished executing on the GPU, the previous
+     *                  ring buffer allocation will be freed.
+     * @return the id<MTLBuffer> and offset for the new allocation
+     */
+    std::tuple<id<MTLBuffer>, NSUInteger> createNewAllocation(id<MTLCommandBuffer> cmdBuffer) {
+        const auto occupiedSlots = mOccupiedSlots.load(std::memory_order_relaxed);
+        if (UTILS_UNLIKELY(occupiedSlots == mSlotCount)) {
+            // We don't have any room left, so we fall back to creating a one-off buffer.
+            mAuxBuffer = [mDevice newBufferWithLength:mSlotSizeBytes options:mBufferOptions];
```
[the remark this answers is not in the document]
bejado: Since it's a reference-counted Objective-C object, it can never be "leaked". When calling this method the second time, the previous `mAuxBuffer` will be freed, unless it has been retained by any Metal command buffers (which will then free it after they've finished executing).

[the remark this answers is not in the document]
pixelflinger: aahhhh. cool. I think this deserves a small comment.

pixelflinger (approved):

PR #6090, opened 2022-09-20
fix several OOB in filamesh and matc

pixelflinger:


romainguy (approved):

libs/filamat/src/Includes.cpp
```diff
@@ -261,10 +261,17 @@ std::vector<FoundInclude> parseForIncludes(const utils::CString& source) {
...
 
         // Increment until we reach the next "
-        while (sourceString[result] != '"') {
+        while (result < sourceString.length() && sourceString[result] != '"') {
             result++;
         }
 
+        // check we're not at the end of the line -- this would be a malformed include directive.
+        if (result >= sourceString.length()) {
+            result = sourceString.find("#include", result);
```
bejado, line 270: Do we need the call to `.find` here?

pixelflinger: I have no idea, I just copied how that code handles EOF above. It looks like it tries to ignore that include and find the next one?

PR #6093, opened 2022-09-21
Metal: use argument buffers for SamplerGroups

bejado:
This switches Metal to use [argument buffers](https://developer.apple.com/documentation/metal/buffers/about_argument_buffers?language=objc) for referencing samplers and textures in draw calls. Argument buffers are similar to Vulkan descriptors.

The primarily reason for doing this is to support more textures/samplers for feature level 3 on newer Metal devices. This also opens up some potential CPU performance improvements.

The logic is a bit tricky, as we don't necessarily know all the final textures at `updateSamplerGroup` time, so `MetalSamplerGroup` has some additional bookkeeping to do.

### DIFF
--- a/libs/filamat/src/shaders/CodeGenerator.h
+++ b/libs/filamat/src/shaders/CodeGenerator.h
@@ -161,6 +161,8 @@ public:
             std::string& shader, filament::SamplerInterfaceBlock const& sib) noexcept;
 
     // These constants must match the equivalent in MetalState.h.
+    // These values represent the starting index for uniform and sampler group [[buffer(n)]]
+    // bindings. See the chart at the top of MetalState.h.
     static constexpr uint32_t METAL_UNIFORM_BUFFER_BINDING_START = 17u;
     static constexpr uint32_t METAL_SAMPLER_GROUP_BINDING_START = 27u;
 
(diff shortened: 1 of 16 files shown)


libs/filamat/src/shaders/CodeGenerator.h
```diff
@@ -142,6 +142,10 @@ class UTILS_PRIVATE CodeGenerator {
     static void fixupExternalSamplers(
             std::string& shader, filament::SamplerInterfaceBlock const& sib) noexcept;
 
+    // These constants must match the equivalent in MetalState.h.
+    static constexpr uint32_t METAL_UNIFORM_BUFFER_BINDING_START = 17u;
+    static constexpr uint32_t METAL_SAMPLER_GROUP_BINDING_START = 27u;
```
pixelflinger, line 147: maybe a comment explaining how 17 and 27 where chosen.

libs/filamat/src/shaders/ShaderGenerator.cpp
```diff
@@ -278,15 +278,17 @@ std::string ShaderGenerator::createVertexProgram(ShaderModel shaderModel,
                 UibGenerator::getPerRenderableMorphingUib());
         cg.generateSamplers(vs,
                 material.samplerBindings.getBlockOffset(SamplerBindingPoints::PER_RENDERABLE_MORPHING),
-                SibGenerator::getPerRenderPrimitiveMorphingSib(variant));
+                SibGenerator::getPerRenderPrimitiveMorphingSib(variant),
+                SamplerBindingPoints::PER_RENDERABLE_MORPHING);
```
pixelflinger, line 282: not a big deal, but I would move `SamplerBindingPoints` as the 2nd parameter.

pixelflinger (approved):

Issue #6094, opened 2022-09-21
VulkanSwapChain::create sizing


jwinarske:
**Describe the bug**

Running on Linux Wayland with two different Vulkan drivers I am seeing VulkanSwapChain::create fail.

**NVIDIA RTX A2000 Laptop GPU**

Error
```
VULKAN ERROR: (VUID-VkSwapchainCreateInfoKHR-imageExtent-01274) Validation Error: [ VUID-VkSwapchainCreateInfoKHR-imageExtent-01274 ] Object 0: handle = 0x7fa4b499bcc0, type = VK_OBJECT_TYPE_DEVICE; | MessageID = 0x7cd0911d | vkCreateSwapchainKHR() called with imageExtent = (4294967295,4294967295), which is outside the bounds returned by vkGetPhysicalDeviceSurfaceCapabilitiesKHR(): currentExtent = (4294967295,4294967295), minImageExtent = (1,1), maxImageExtent = (16384,16384). The Vulkan spec states: imageExtent must be between minImageExtent and maxImageExtent, inclusive, where minImageExtent and maxImageExtent are members of the VkSurfaceCapabilitiesKHR structure returned by vkGetPhysicalDeviceSurfaceCapabilitiesKHR for the surface (https://vulkan.lunarg.com/doc/view/1.3.216.0/linux/1.3-extensions/vkspec.html#VUID-VkSwapchainCreateInfoKHR-imageExtent-01274)
```

**Intel(R) UHD Graphics (TGL GT1)**

Error
```
[...]

PR #6097, opened 2022-09-22
Wayland FilamentApp

jwinarske:
-move Wayland specific code into getNativeWindow
-make wayland struct static

Signed-off-by: Joel Winarske <joel.winarske@gmail.com>

### DIFF
--- a/libs/filamentapp/src/NativeWindowHelperLinux.cpp
+++ b/libs/filamentapp/src/NativeWindowHelperLinux.cpp
@@ -32,7 +32,17 @@ void* getNativeWindow(SDL_Window* sdlWindow) {
     }
     else if (wmi.subsystem == SDL_SYSWM_WAYLAND) {
 #if defined(FILAMENT_SUPPORTS_WAYLAND)
-        return wmi.info.wl.display;
+        // Static is used here to allocate the struct pointer for the lifetime of the program.
+        // Without static the valid struct quickyly goes out of scope, and ends with seemingly
+        // random segfaults.
+        static struct {
+            struct wl_display *display;
+            struct wl_surface *surface;
+        } wayland {
+            wmi.info.wl.display,
+            wmi.info.wl.surface,
+        };
+        return (void *) &wayland;
 #endif
     }
     return nullptr;
--- a/libs/filamentapp/src/FilamentApp.cpp
+++ b/libs/filamentapp/src/FilamentApp.cpp
@@ -18,11 +18,7 @@
 
 #include <filamentapp/FilamentApp.h>
 
-#if !defined(WIN32)
-#if defined(FILAMENT_SUPPORTS_WAYLAND)
-#    include <SDL_syswm.h>
-#endif
-#else
+#if defined(WIN32)
 #    include <SDL_syswm.h>
 #    include <utils/unwindows.h>
 #endif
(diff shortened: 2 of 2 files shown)


jwinarske:
Does not warrant release note addition.

prideout:
Thanks Joel, does this fix one of the issues that you filed?

jwinarske:
This PR is combination with proposed fix in  https://github.com/google/filament/issues/6094 enable Wayland, and the sample apps.

Happy to make it a WIP until we have agreed on the solution for the swap chain sizing issue.

prideout (approved):

libs/filamentapp/src/NativeWindowHelperLinux.cpp
```diff
@@ -32,7 +32,14 @@ void* getNativeWindow(SDL_Window* sdlWindow) {
     }
     else if (wmi.subsystem == SDL_SYSWM_WAYLAND) {
 #if defined(FILAMENT_SUPPORTS_WAYLAND)
-        return wmi.info.wl.display;
+        static struct {
```
romainguy, line 35: Will this ever work for multiple windows/multiple displays? It seems that returning a pointer to a static struct is a recipe for subtle bugs somewhere down the line.

jwinarske: As-is this supports multi-window/display scenarios.  The display and surface pointer stays constant during the lifecycle of the Wayland client session.

The easiest scenario to switch startup outputs for an existing Wayland based app is to set the `WAYLAND_DISPLAY` environmental variable to the desired display.  Then of course one can programmatically move/set windows to a desired output.  Wayland based compositors are very flexible in this regard.

So with the Filament demo apps I can launch on any output, or move them between outputs.

jwinarske: The other very important fact `static` is addressing a problem.  Thread 1 is the owner of the stack memory which the struct points to.  When the VK surface is created which consumes the pointer, it may be on Thread 73.  If static isn't used you will _most likely_ hit a segfault.

romainguy: Yes without static this would not work, even on a single thread.

pixelflinger: Okay I really want a comment here explaining everything that was said in this thread.

pixelflinger (changes requested):
very nice change. I just would like a comment explaining the `static` logic.

jwinarske:
@pixelflinger Thanks! and done.

pixelflinger (approved):

pixelflinger:
I'm not sure what happened here, but it looks like there are 20 commits in this branch?

jwinarske:
Updated branch

jwinarske:
"This branch is out-of-date with the base branch" -> "Update branch"
main rebase

pixelflinger:
It still says it's out of date for me (and I don't have a update button) and also says there are 20 commits, that seems wrong.

jwinarske:
@pixelflinger it's saying it's up to date now

pixelflinger:
@jwinarske could you sync your branch and rebase there? I can't commit this PR as it is.

jwinarske:
Sure.  I'll reset hard to main, then add single commit on top

jwinarske:
@pixelflinger should be good to go

PR #6098, opened 2022-09-22
gltfio API changes, parts 1 and 2.

prideout:
This moves `MaterialInstance` objects into the `FilamentInstance` objects that they apply to, rather than storing them in one monolithic list in `FilamentAsset`.

This also removes the "broadcast" animator from FilamentAsset.  Clients should simply use the per-instance animators.

In the near future, we will also move all the entity getter methods from `FilamentAsset` to `FilamentInstance`.

These changes are motivated by the need to allow instance creation from immutable assets, which is a feature request from Google.


bejado (approved):

romainguy (commented):
You also need to update the Java APIs (and JNI).

PR #6099, opened 2022-09-23
Update cgltf to version 1.13

daemyung:


daemyung:
I think this PR does not require to warrant a release note.

romainguy:
I’m not against upgrading but it’s a bit useless if we don’t implement support for the new extensions…

daemyung:
@romainguy Yes... Honestly this upgrade is needed for me. I need to re-write GLTF and GLB in some reasons and there is some bugs in current cgltf in filament. So I hope to accept this PR if there is no regression issues.

bejado (approved):

romainguy:
@daemyung Sure, but why are you using the version of cgltf that comes with Filament and not your own? `third_party` libraries are not part of our public API.

daemyung:
@romainguy Yes I can do that and technically you are write. I just don't want to increase more dependencies on the current project.

@bejado Thanks!

Issue #6091, opened 2022-09-20
Feature request: function to reset internal GL state tracking


jfaust:
**Is your feature request related to a problem? Please describe.**
In our project we're mixing Filament rendering with an external OpenGL-based renderer that handles the 2D UI. On non-web platforms we use shared contexts and everything works great. On the web though, there's no option to do this. This means we sometimes run into conflicts with Filament due to its internal state tracking. It shows up in unexpected ways - like textures not getting bound correctly, or wrong sampler settings getting applied - and always takes a bit to track down.

**Describe the solution you'd like**
I'd like a function to reset the internal state tracking. Something like `Engine::resetStateTracking()`, that we could call once a frame. On non-GL platforms (and maybe even on everything but WebGL specifically) it'd be a no-op.

**Describe alternatives you've considered**
Currently we reset state to valid defaults after the external GL library renders, but that is only a partial workaround. We've considered using `glGet`, but it's too slow.

**OS and backend**
Yes, this is GL-specific, as none of the other APIs have global state.

This is something I'd be happy to submit a PR for if it's likely to be accepted.

pixelflinger:
I'm very reluctant adding such an API because we're not actually tracking *all* the state that might exist.  To implement this correctly, we would have to:
- reset the state we track
- reset the state we set during init (i.e. constant state)
- reset the state we don't know we rely on (like all the GL defaults)
- reset the state we don't know about (i.e. you could change some extension state we're not using) but that has an impact on rendering or on our state.

This is also a lot of tedious work we don't need ourselves, so I think the best course of action would be to submit a PR adding this feature and we'll review it.

jfaust:
I think solving just the first one is at least a good start. We're already able to solve any problems caused by 2-4, but due to how the state tracking works (and the lack of ability to introspect it), there's often no guaranteed way to work around bugs caused by the state tracking itself.

Would a PR just doing (1) have a chance? Or should I do 1-3? (4) seems like it's out of scope.

pixelflinger:
Yes a PR just doing (1) would have a chance.

jfaust:
Resolved via #6105

PR #6105, opened 2022-09-27
Add a function to re-set internal GL state tracking

jfaust:
This makes it possible to integration with external renderers in WebGL without causing rendering artifacts.

See https://github.com/google/filament/issues/6091

### DIFF
--- a/filament/backend/src/opengl/OpenGLContext.cpp
+++ b/filament/backend/src/opengl/OpenGLContext.cpp
@@ -61,6 +61,8 @@ OpenGLContext::OpenGLContext() noexcept {
     glGetIntegerv(GL_MAX_DRAW_BUFFERS, &gets.max_draw_buffers);
     glGetIntegerv(GL_MAX_TEXTURE_IMAGE_UNITS, &gets.max_texture_image_units);
     glGetIntegerv(GL_MAX_COMBINED_TEXTURE_IMAGE_UNITS, &gets.max_combined_texture_image_units);
+    glGetIntegerv(GL_MAX_TRANSFORM_FEEDBACK_SEPARATE_ATTRIBS, &gets.max_transform_feedback_separate_attribs);
+    glGetIntegerv(GL_MAX_UNIFORM_BUFFER_BINDINGS, &gets.max_uniform_buffer_bindings);
     glGetIntegerv(GL_UNIFORM_BUFFER_OFFSET_ALIGNMENT, &gets.uniform_buffer_offset_alignment);
 
     constexpr auto const caps3 = FEATURE_LEVEL_CAPS[+FeatureLevel::FEATURE_LEVEL_3];
@@ -247,31 +249,7 @@ OpenGLContext::OpenGLContext() noexcept {
     flush(slog.v);
 #endif
 
-    /*
-     * Set our default state
-     */
-
-    // We need to make sure our internal state matches the GL state when we start.
-    // (some of these calls may be unneeded as they might be the gl defaults)
-    glDisable(GL_BLEND);
-    glDisable(GL_CULL_FACE);
-    glDisable(GL_SCISSOR_TEST);
-    glDisable(GL_DEPTH_TEST);
-    glDisable(GL_STENCIL_TEST);
-    glDisable(GL_DITHER);
-    glDisable(GL_SAMPLE_ALPHA_TO_COVERAGE);
-    glDisable(GL_SAMPLE_COVERAGE);
-    glDisable(GL_POLYGON_OFFSET_FILL);
-
-    // Point sprite size and seamless cubemap filtering are disabled by default in desktop GL.
-    // In OpenGL ES, these flags do not exist because they are always on.
-#if BACKEND_OPENGL_VERSION == BACKEND_OPENGL_VERSION_GL
-    enable(GL_PROGRAM_POINT_SIZE);
-#endif
-
-#ifdef GL_ARB_seamless_cube_map
-    enable(GL_TEXTURE_CUBE_MAP_SEAMLESS);
-#endif
+    setDefaultState();
 
 #ifdef GL_EXT_texture_filter_anisotropic
     if (ext.EXT_texture_filter_anisotropic) {
@@ -288,10 +266,6 @@ OpenGLContext::OpenGLContext() noexcept {
         }
         glDeleteSamplers(1, &s);
     }
-#endif
-
-#ifdef GL_FRAGMENT_SHADER_DERIVATIVE_HINT
-    glHint(GL_FRAGMENT_SHADER_DERIVATIVE_HINT, GL_NICEST);
 #endif
 
 #if !defined(NDEBUG) && defined(GL_KHR_debug)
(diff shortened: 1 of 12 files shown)


filament/include/filament/Engine.h
```diff
@@ -725,6 +725,20 @@ class UTILS_PUBLIC Engine {
       */
     utils::JobSystem& getJobSystem() noexcept;
 
+    /**
+      * Tells the driver to reset any internal state tracking if necessary.
```
romainguy, line 729: You should mention this is only useful when the Filament driver and the external driver share the same thread and context (ie WebGL). This makes me wonder if the implementation shouldn't be a no-op except in WebGL? Just to avoid extra costs to apps calling this unnecessarily. @pixelflinger thoughts?

filament/include/filament/Engine.h
```diff
@@ -725,6 +725,20 @@ class UTILS_PUBLIC Engine {
...
+      * This should only be used in WebGL, and only if you are integrating an external
+      * renderer with Filament. This can be useful because Filament keeps track of the GL
+      * state it has set (like which texture is bound), and does not re-set that state if
+      * it does not think it needs to. However, if an external renderer has set different
+      * state in the mean time, Filament will use that new state unknowingly.
+      * 
+      * If you are in this situation, call this function - ideally only once per frame, 
+      * immediately after the external renderer has finished.
+      */
+    void resetInternalStateTracking() noexcept;
```
romainguy, line 740: Maybe use a more specific name like resetDriverState() or resetBackendState()?

jfaust: If we restrict this to only existing in `WebGL`, I could go even more specific like `resetWebGLDriverState()`

pixelflinger: I think `resetBackendState()` sounds good. I would just `#if defined(__EMSCRIPTEN__)` it so it doesn't even appear on regular builds.

filament/backend/src/opengl/OpenGLContext.cpp
```diff
@@ -521,4 +521,21 @@ void OpenGLContext::deleteVertexArrays(GLsizei n, const GLuint* arrays) noexcept
...
 
+void OpenGLContext::resetStateTracking() noexcept {
+    // Store off strings instead of re-calling glGetString()
+    char const* vendor = state.vendor;
+    char const* renderer = state.renderer;
+    char const* version = state.version;
+    char const* shader = state.shader;
+
+    // Reset state to defaults
+    state = State {};
```
pixelflinger, line 532: I don't understand why this works. Nothing guarantees that the default state matches the current GL state left by the app.

I think, unfortunately, what you need to do is go through every bit of state in the `State` structure and forcibly set the corresponding GL state to that.

pixelflinger (changes requested):
I don't think the actual state reset works as it is written.

filament/backend/src/opengl/OpenGLContext.cpp
jfaust: Good point. This worked for the bits I had good repros for - bound textures, viewport, fbo, since they got set to null values & re-set. I should've realized that won't work for most of the rest of the state. Will fix.

jfaust: I think I've taken care of this, but I may have misunderstood some piece - `state.vao` and `state.textures` in particular I'm not completely confident about.

pixelflinger:
looks reasonable, will review in more details soon.

filament/backend/src/opengl/OpenGLContext.cpp
pixelflinger: I *think* technically, you don't need to reset the  `State` structure anymore, since you're re-syncing the gl state to it.

pixelflinger (changes requested):
I think we should make this work without reseting the `State` structure -- this should work because you're re-syncing the gl state to it.

filament/backend/src/opengl/OpenGLContext.cpp
jfaust: I think there are three exceptions right now:
1. I'm calling `bindVertexArray(nullptr);` - this looks fairly straightforward to change:
```cpp
glBindVertexArray(state.vao.p->vao);
if (UTILS_UNLIKELY(bugs.vao_doesnt_store_element_array_buffer_binding)) {
    glBindBuffer(GL_ELEMENT_ARRAY_BUFFER, state.vao.p->elementArray);
}
```

2. Resetting texture units is just binding 0 everywhere. If I'm understanding correctly, I'll need to lookup the unit for the call to `glBindSampler`: `state.textures.units[unit].sampler`? And then for `glBindTexture()`: ```
```cpp
int targets_index = getIndexForTextureTarget(target);
glBindTexture(target, state.textures.units[unit].targets[targets_index].texture_id`);
```

3. I broke out setting the initial `glDisable()/glEnable()`, `glHint` and `glClipControl` into `OpenGLContext::setDefaultState()`. I can move those back into the constructor & manually run through `state.enables`.  I assume I'll need to run through the values in `getIndexForCap()` and set those based on what's set in `state.enables.caps`?

I'd also then change `Driver::resetStateTracking()` to `Driver::resetState()`.

If those proposed changes look good I'll update the PR.

jfaust: @pixelflinger I'm attempting to make these changes - however, I'm running into issues with the texture state that I think is part of a different issue.

`OpenGLDriver` uses  `OpenGLContext::MAX_TEXTURE_UNIT_COUNT - 1` in a bunch of places when binding textures to upload them. However, that constant is larger than the allowed number of units available in WebGL, at least on the browsers I've tried on Windows. This causes warnings to be printed (Firefox): `WebGL warning: activeTexture: TEXTURE61 must be < MAX_COMBINED_TEXTURE_IMAGE_UNITS (32).`. I think this also means the state tracking ends up in an invalid state, because the active texture unit is failing to be set. I'll open a ticket about this.

pixelflinger: yes, we fixed that one recently. it was my bad.

jfaust: Ah, I should've synced. Thanks

jfaust:
I missed `state.buffers` somehow, so don't review this yet

jfaust:
OK, I think this works. I'm kind of shaky on whether the code handling `state.buffers` is totally correct, but it does prevent the errors we see in our use case. Given the complexity of that code and the need to keep it in sync I'm almost inclined to zero out `state.buffers` and just reset all the buffer bindings, but I'll leave that up to you.

pixelflinger:
> OK, I think this works. I'm kind of shaky on whether the code handling `state.buffers` is totally correct, but it does prevent the errors we see in our use case. Given the complexity of that code and the need to keep it in sync I'm almost inclined to zero out `state.buffers` and just reset all the buffer bindings, but I'll leave that up to you.

I think it's a good idea. Let's do that!

filament/backend/src/opengl/OpenGLContext.cpp
```diff
@@ -325,6 +298,49 @@ OpenGLContext::OpenGLContext() noexcept {
...
+        GL_DEPTH_TEST,
+        GL_STENCIL_TEST,
+        GL_DITHER,
+        GL_SAMPLE_ALPHA_TO_COVERAGE,
+        GL_SAMPLE_COVERAGE,
+        GL_POLYGON_OFFSET_FILL,  
+    };
+    size_t capsCount = sizeof(caps) / sizeof(caps[0]);
+
+    for (GLenum capi = 0; capi < capsCount; ++capi) {
```
pixelflinger, line 319:
```suggestion

    UTILS_NOUNROLL
    for (GLenum capi = 0; capi < capsCount; ++capi) {
```

jfaust:
> > OK, I think this works. I'm kind of shaky on whether the code handling `state.buffers` is totally correct, but it does prevent the errors we see in our use case. Given the complexity of that code and the need to keep it in sync I'm almost inclined to zero out `state.buffers` and just reset all the buffer bindings, but I'll leave that up to you.
> 
> I think it's a good idea. Let's do that!

Thoughts on if it's worth doing that for texture/sampler state too?

pixelflinger:
> > > OK, I think this works. I'm kind of shaky on whether the code handling `state.buffers` is totally correct, but it does prevent the errors we see in our use case. Given the complexity of that code and the need to keep it in sync I'm almost inclined to zero out `state.buffers` and just reset all the buffer bindings, but I'll leave that up to you.
> > 
> > 
> > I think it's a good idea. Let's do that!
> 
> Thoughts on if it's worth doing that for texture/sampler state too?

Yes I would, it should simplify the code a lot, and make it much less error prone, like you said. Thanks a lot!

jfaust:
> Yes I would, it should simplify the code a lot, and make it much less error prone, like you said. Thanks a lot!

OK, this seems gtg. Thanks for spending the time on this review!

pixelflinger (approved):

pixelflinger (approved):

pixelflinger:
@jfaust could you sync main into your branch and rebase and squash your changes on top (we don't do merges, only rebase), so we have only one CL in the end. Thanks.

jfaust:
@pixelflinger done

romainguy:
Thanks @jfaust!

PR #6107, opened 2022-09-27
Get the morph target buffer to the given primitive

daemyung:


pixelflinger (approved):

prideout (changes requested):
The doc comment should also say: "or null if it doesn't exist".

Also I think the method can be marked `noexcept`.

daemyung:
@prideout Yes it's done.

prideout (approved):

PR #6108, opened 2022-09-27
reduce setParameter(texture) calls

pixelflinger:


filament/src/PostProcessManager.cpp
```diff
@@ -1919,45 +1919,64 @@ PostProcessManager::BloomPassOutput PostProcessManager::bloomPass(FrameGraph& fg
...
 
-                mi->use(driver);
-                mi->setParameter("source", hwIn, {
+                auto const& material = getPostProcessMaterial("bloomDownsample");
+                auto const* ma = material.getMaterial(mEngine);
+
+                FMaterialInstance* mis[] = {
+                        ma->createInstance("bloomUpsample-ping"),
+                        ma->createInstance("bloomUpsample-pong"),
+                        ma->createInstance("bloomUpsample-first"),
```
bejado, line 1932: Should these be named `bloomDownsample-*`?

bejado (approved):
Nice!

filament/src/PostProcessManager.cpp
pixelflinger: ah good catch

PR #6125, opened 2022-09-29
Fixes for 1.27.2 release

bejado:
These are needed for G3.


romainguy (approved):
Thanks! We'll have to turn off the warning in AS then.

PR #6127, opened 2022-09-29
fix artifacts EVSM shadows and allow 32 bits texture and full precision EVSMs as options.

pixelflinger:


bejado:
Maybe worth a release note?

bejado (approved):

PR #6130, opened 2022-09-30
fix "stable" shadows

pixelflinger:
We were always adjusting the near/far of the view volume based on the scene content for shadowing, which defeated the "stable" shadows.

Also changed the default cascade splits from a linear split to a log2 split, because due to the perspective projection, the log2 split actually looks linear. Also intuitively, it makes more sense to give more resolution to the shadows close to the camera.

### DIFF
--- a/android/filament-android/src/main/java/com/google/android/filament/LightManager.java
+++ b/android/filament-android/src/main/java/com/google/android/filament/LightManager.java
@@ -240,7 +240,7 @@ public class LightManager {
          */
         @NonNull
         @Size(min = 3)
-        public float[] cascadeSplitPositions = { 0.25f, 0.50f, 0.75f };
+        public float[] cascadeSplitPositions = { 0.125f, 0.25f, 0.50f };
 
         /** Constant bias in world units (e.g. meters) by which shadows are moved away from the
          * light. 1mm by default.
--- a/filament/src/ShadowMap.cpp
+++ b/filament/src/ShadowMap.cpp
@@ -109,17 +109,19 @@ void ShadowMap::updateDirectional(const FScene::LightSoa& lightData, size_t inde
 #endif
 
     // Adjust the camera's projection for the light's shadowFar
-    mat4f cullingProjection(camera.cullingProjection);
-    if (params.options.shadowFar > 0.0f) {
-        float n = camera.zn;
-        float f = params.options.shadowFar;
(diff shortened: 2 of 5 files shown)


android/filament-android/src/main/java/com/google/android/filament/LightManager.java
```diff
@@ -240,7 +240,7 @@ public static class ShadowOptions {
          */
         @NonNull
         @Size(min = 3)
-        public float[] cascadeSplitPositions = { 0.25f, 0.50f, 0.75f };
+        public float[] cascadeSplitPositions = { 0.125f, 0.25f, 0.50f };
```
romainguy, line 243: 0.125f as the first split might be too close. In debug view I often need to move the first split farther away because with the default fov it’s outside of the view frustum often.

filament/src/ShadowMap.cpp
```diff
@@ -109,17 +109,19 @@ void ShadowMap::updateDirectional(const FScene::LightSoa& lightData, size_t inde
...
-    mat4f cullingProjection(camera.cullingProjection);
-    if (params.options.shadowFar > 0.0f) {
-        float n = camera.zn;
-        float f = params.options.shadowFar;
-        // orthographic projection
-        assert_invariant(std::abs(cullingProjection[2].w) <= std::numeric_limits<float>::epsilon());
-        cullingProjection[2].z = 2.0f / (n - f);
-        cullingProjection[3].z = (f + n) / (n - f);
-    }
+    const mat4f cullingProjection{ [&](auto p) {
```
romainguy, line 112: Why the lambda here? I don’t get what benefit it brings

pixelflinger: I just wanted `cullingProjection` to be `const` -- it's all inlined of course.

romainguy (approved):

android/filament-android/src/main/java/com/google/android/filament/LightManager.java
pixelflinger: It is quite close but still visible on our demos (e.g. sponza). It also works a lot better for "stable" mode, and provides a visually linear partitioning. In stable mode with the default far plane, the first split is at about 12m, it's pretty far. With the old default it would be 25m and so it looks like it's not even working.

PR #6149, opened 2022-10-05
Fix compilation for GCC

MrKepzie:
This PR fixes multiple issues that prevented GCC from being used. 
This addresses the following issues:
- C++ 20 designated initializers were used here and there, preventing compilation with C++17
- The anonymous struct in FrameGraphRenderPass was not compiling
- removed some noexcept which would fail the compilation
- ambiguous typedefs
- missing inludes
- symbols visibility at lnk time

This PR is a bunch of cherry picks in-between other feature commits, please make sure to review and test it on your side.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -228,16 +228,12 @@ if (CMAKE_C_COMPILER_ID MATCHES "Clang")
     if (CMAKE_C_COMPILER_VERSION VERSION_LESS MIN_CLANG_VERSION)
         message(FATAL_ERROR "Detected C compiler Clang ${CMAKE_C_COMPILER_VERSION} < ${MIN_CLANG_VERSION}")
     endif()
-elseif (NOT MSVC)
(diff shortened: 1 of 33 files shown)


romainguy:
Thank you for your contribution but our support of GCC is limited to public APIs and we will not maintain support to compile Filament itself with GCC.

MrKepzie:
Ah, took 1h to compile does commits, but should have asked first.
Well at least this fixes filamat/include/MaterialBuilder.h and the ambiguous enum typdefs in filament/include/View.h,Material.h,MaterialInstance.h

pixelflinger:
Too many changes we can't take or don't want to. 
For designated initializers, I think the best course of action is to enable C++20 when compiling with gcc (I think we even do that for MSVC). I'm not sure about the other issues, some need investigation, e.g. the noexcept ones.
Also we don't use, by convention, the <c...> includes and prefer the old <....h> versions (except for math.h), as it's generally faster to compile.

For the missing includes and ambiguous typedefs, please provide a separate PR.

thanks.

PR #6152, opened 2022-10-06
Add support for point light shadows

pixelflinger:


TheOriginalBytePlayer:
WOW... you guys obviously trained under Scotty... "Och, Captain, I dinna know if I can make that work."  two minutes later... :)

So where would I go to try this out?

shaders/src/light_punctual.fs
```diff
@@ -210,7 +209,20 @@ void evaluatePunctualLights(const MaterialInputs material,
...
             if (light.castsShadows) {
-                visibility = shadow(false, light_shadowMap, light.shadowLayer, light.shadowIndex, 0u);
+                uint layer = light.shadowLayer;
+                highp vec4 shadowPosition;
+                if (light.type == LIGHT_TYPE_POINT) {
+                    // point-light shadows are sampled from a direction
+                    highp vec3 r = getWorldPosition() - light.worldPosition;
+                    highp vec4 nf = shadowUniforms.shadows[light.shadowIndex].lightFromWorldZ;
+                    // getShadowPosition returns zLight which is needed for PCSS/DPCF
+                    shadowPosition = getShadowPosition(r, nf, layer, light.zLight);
```
bejado, line 219: Could we move the pointlight `light.zLight` calculation to the `getLight` function?

pixelflinger: We could but I didn't do it because this calculation can be skipped if there is no shadow or in the early exit case.

PR #6153, opened 2022-10-07
Add FrameGraph debugger lib: package structure + CmakeLists.txt

feresr:
Project structure for a FrameGraph debugger, should more or less mimic what `matdbg`.  cc: @pixelflinger


filament/src/details/Engine.cpp
```diff
@@ -396,6 +402,10 @@ void FEngine::shutdown() {
     ASSERT_PRECONDITION(ThreadUtils::isThisThread(mMainThreadId),
             "Engine::shutdown() called from the wrong thread!");
 
+#if FILAMENT_ENABLE_FGDBG
+    delete debug.fgdbg;
+#endif
```
feresr, line 407: `matdbg` does not `delete debug.server` here. But I think we should

libs/fgdbg/include/fgdbg/DebugServer.h
```diff
@@ -0,0 +1,31 @@
...
+ */
+
+#ifndef FGDBG_DEBUGSERVER_H
+#define FGDBG_DEBUGSERVER_H
+
+namespace filament {
+namespace fgdbg {
+
+class DebugServer {
+  public:
```
romainguy, line 24: We use 4 spaces indent.

libs/fgdbg/src/DebugServer.cpp
```diff
@@ -0,0 +1,29 @@
...
+ * limitations under the License.
+ */
+
+#include "fgdbg/DebugServer.h"
+#include "iostream"
+
+namespace filament::fgdbg {
+
+DebugServer::DebugServer() {
+  std::cout << "FrameGraph debug server created" << std::endl;
```
romainguy, line 23: Use slog instead of std::cout inside libfilament. std::cout is only used in host tools.

README.md
```diff
@@ -324,6 +324,7 @@ and tools.
   - `image`:                  Image filtering and simple transforms
   - `imageio`:                Image file reading / writing, only intended for internal use
   - `matdbg`:                 DebugServer for inspecting shaders at run-time (debug builds only)
+  - `fgdbg`:                  Frame Graph inspector and debugger (debug builds only)
```
romainguy, line 327: Keep the list alphabetized

filament/src/details/Engine.cpp
pixelflinger: @prideout ?

filament/src/details/Engine.h
```diff
@@ -57,6 +57,16 @@
...
 #include <filament/VertexBuffer.h>
 
+#if FILAMENT_ENABLE_FGDBG
+#include <fgdbg/DebugServer.h>
+#else
+namespace filament {
+namespace fgdbg {
+class DebugServer;
+} // namespace fgdbg
+} // namespace filament
```
pixelflinger, line 67:
```suggestion
namespace filament::fgdbg {
class DebugServer;
} // namespace filament::fgdbg
```

filament/src/details/Engine.h
```diff
@@ -516,6 +526,7 @@ class FEngine : public Engine {
             bool doFrameCapture = false;
         } renderer;
         matdbg::DebugServer* server = nullptr;
+        fgdbg::DebugServer* fgdbg = nullptr;
```
pixelflinger, line 529: you could also use a `std::unique_ptr<fgdbg::DebugServer>` to not have to worry about the `delete`.

libs/fgdbg/include/fgdbg/DebugServer.h
pixelflinger: Also we don't indent `public:` / `private:`

libs/fgdbg/src/DebugServer.cpp
pixelflinger: include `<utils/Log.h>` and use `utils::slog.d` instead of `std::cout`.

libs/fgdbg/src/DebugServer.cpp
```diff
@@ -0,0 +1,29 @@
...
+}
+
+DebugServer::~DebugServer(){
```
pixelflinger, line 26:
```suggestion
DebugServer::~DebugServer() {
```

libs/fgdbg/src/DebugServer.cpp
```diff
@@ -0,0 +1,29 @@
...
+  std::cout << "FrameGraph debug server destroyed" << std::endl;
```
pixelflinger, line 27:
```suggestion
    std::cout << "FrameGraph debug server destroyed" << std::endl;
```

pixelflinger (changes requested):
Very nice first PR!

CMakeLists.txt
```diff
@@ -685,6 +692,10 @@ if (FILAMENT_BUILD_FILAMAT OR IS_HOST_PLATFORM)
     if (FILAMENT_ENABLE_MATDBG OR IS_HOST_PLATFORM)
         add_subdirectory(${LIBRARIES}/matdbg)
     endif()
+
+    if (FILAMENT_ENABLE_FRAMEGRAPH OR IS_HOST_PLATFORM)
```
bejado, line 696: Where is `FILAMENT_ENABLE_FRAMEGRAPH` defined? Do you mean `FILAMENT_ENABLE_FGDBG`?

libs/fgdbg/include/fgdbg/DebugServer.h
feresr: Thanks, done!

filament/src/details/Engine.h
feresr: Cool I'll do that, didn't know where we stand on smart pointers :)

CMakeLists.txt
feresr: Good catch! Yes, I renamed the project and missed this one, thank you!

pixelflinger:
I think you need to sync main into your branch and rebase them on top.

feresr:
rebase done cc: @pixelflinger

filament/src/details/Engine.cpp
prideout: Why do you need to delete fgdbg is you're using a smart pointer?  Can you rename `server` to `matdbg` to help avoid confusion between the two debuggers?  You can also convert matdbg to use a smart pointer so that it is consistent with the new fgdbg stuff.

prideout (changes requested):
Looks good but for consistency can you do a few minor things: (1) rename `server` to `matdbg` and(2) change it over to be `unique_ptr`

filament/src/details/Engine.cpp
feresr: Done! 
Yes it used to be a raw pointer, this is outdated code

prideout (approved):

PR #6155, opened 2022-10-07
Install filagui, filamentapp, and depedencies

### LINKED ISSUE #6154: Install filamentapp and filagui
I want to create an application outside of filament and want to use filamentapp and filagui to accelerate building an application. But two libraries are not installed. So there is no way to do this.

Install filamentapp and filagui.

daemyung:
Fix #6154

### DIFF
--- a/libs/filagui/CMakeLists.txt
+++ b/libs/filagui/CMakeLists.txt
@@ -94,3 +94,11 @@ if (NOT MSVC)
             $<$<CONFIG:Release>:-ffast-math>
     )
 endif()
+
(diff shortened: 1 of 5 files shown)


daemyung:
This PR is not my favorite. Because third party libraries should be installed. I'm pretty sure you also don't like this. To hide third party dependencies on filamentapp and filagui, they need public headers and hide implementations. But theses needs a lot of changes. This PR is minimal modifications to link filagui and filmentapp.

romainguy:
See my comment on the issue :(

PR #6168, opened 2022-10-10
Add dummy DebugServer for fgdbg with support for Websockets

feresr:
Sets up a debug server, test in the browser by opening a console and running

```
var webSocket = new WebSocket('ws://localhost:8082');
webSocket.onmessage = function(data) { console.log(data.data); }
```


filament/src/details/Engine.cpp
```diff
@@ -646,6 +646,15 @@ int FEngine::loop() {
...
     while (true) {
+        #if FILAMENT_ENABLE_FGDBG
+              if (debug.fgdbg) {
+                static int fgdbgMessage = 0;
+                fgdbgMessage++;
+                if (fgdbgMessage % 100 == 0) {
+                  debug.fgdbg->sendMessage(fgdbgMessage/100);
+                }
+              }
+        #endif
```
feresr, line 657: Just for testing purposes, we can do without this. Let me know if should take this out before merge

PR #6169, opened 2022-10-10
Add DebugServer for fgdbg with support for Websockets

feresr:
Test this by running Filament, and opening a browser to `http://localhost:8082/`
console:
```
var webSocket = new WebSocket('ws://localhost:8082');
webSocket.onmessage = function(data) { console.log(data.data); }
```

### DIFF
--- a/filament/src/details/Engine.cpp
+++ b/filament/src/details/Engine.cpp
@@ -646,15 +646,16 @@ int FEngine::loop() {
     uint32_t id = std::thread::hardware_concurrency() - 1;
 
     while (true) {
-        #if FILAMENT_ENABLE_FGDBG
-              if (debug.fgdbg) {
-                static int fgdbgMessage = 0;
-                fgdbgMessage++;
-                if (fgdbgMessage % 100 == 0) {
-                  debug.fgdbg->sendMessage(fgdbgMessage/100);
-                }
-              }
-        #endif
+#if FILAMENT_ENABLE_FGDBG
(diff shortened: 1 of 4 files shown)


filament/src/details/Engine.cpp
```diff
@@ -646,6 +646,15 @@ int FEngine::loop() {
...
     while (true) {
+        #if FILAMENT_ENABLE_FGDBG
+              if (debug.fgdbg) {
+                static int fgdbgMessage = 0;
+                fgdbgMessage++;
+                if (fgdbgMessage % 100 == 0) {
+                  debug.fgdbg->sendMessage(fgdbgMessage/100);
+                }
+              }
+        #endif
```
feresr, line 657: This is only for testing/debug proposes

libs/fgdbg/src/DebugServer.cpp
```diff
@@ -0,0 +1,111 @@
...
+
+#include "fgdbg/DebugServer.h"
+#include <CivetServer.h>
+#include <utils/Log.h>
+#include <tsl/robin_set.h>
+
+using namespace utils;
+namespace filament::fgdbg {
+
+using namespace utils;
```
bejado, line 25: You have this `using` line twice

feresr: I **really** want to use it
haha good catch thanks

romainguy:
Can we merge this?

romainguy:
Thanks @feresr!

PR #6181, opened 2022-10-12
Metal: fix simulator regression due to arg buffers

### LINKED ISSUE #6180: Metal validateCommonDrawErrors triggered
With recent filament versions I am getting 

    -[MTLDebugRenderCommandEncoder validateCommonDrawErrors:]:5252: failed assertion `Draw Errors Validation
    Fragment Function(main0): the offset into the buffer spvDescriptorSet3 that is bound at buffer index 29 
    must be a multiple of 256 but was set to 16.

Only seems to be happening on the iOS simulator, but that may simply be a consequence of the particular screen sizes I have tested with so far. Works fine when building "For my mac". Will do more digging tomorrow to see where this started happening.

bejado:
Fixes #6180

### DIFF
--- a/filament/backend/src/metal/MetalBuffer.h
+++ b/filament/backend/src/metal/MetalBuffer.h
@@ -106,7 +106,12 @@ public:
     // address space. Constant buffers have specific alignment requirements when specifying an
     // offset.
 #if defined(IOS)
+#if TARGET_OS_SIMULATOR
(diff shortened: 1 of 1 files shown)


kpeeters:
That did the trick, thanks!

pixelflinger (approved):

PR #6189, opened 2022-10-14
Initial support for Metal compute shaders / SSBOs

bejado:
`ComputeTest.basic` passes. `readBufferSubData` is not yet implemented, so `ComputeTest.copy` still fails.

### DIFF
--- a/filament/backend/src/metal/MetalDriver.h
+++ b/filament/backend/src/metal/MetalDriver.h
@@ -39,7 +39,7 @@ class MetalTexture;
 struct MetalUniformBuffer;
 struct MetalContext;
 struct MetalProgram;
-struct UniformBufferState;
+struct BufferState;
 
 #ifndef FILAMENT_METAL_HANDLE_ARENA_SIZE_IN_MB
 #define FILAMENT_METAL_HANDLE_ARENA_SIZE_IN_MB 8
@@ -129,7 +129,9 @@ private:
             uint32_t offset, uint32_t minIndex, uint32_t maxIndex, uint32_t count);
 
     void finalizeSamplerGroup(MetalSamplerGroup* sg);
-    void enumerateBoundUniformBuffers(const std::function<void(const UniformBufferState&,
+    void enumerateBoundUniformBuffers(const std::function<void(const BufferState&,
+            MetalBuffer*, uint32_t)>& f);
+    void enumerateBoundSsbos(const std::function<void(const BufferState&,
             MetalBuffer*, uint32_t)>& f);
 
 };
--- a/filament/backend/src/metal/MetalDriver.mm
+++ b/filament/backend/src/metal/MetalDriver.mm
@@ -243,7 +243,7 @@ void MetalDriver::createIndexBufferR(Handle<HwIndexBuffer> ibh, ElementType elem
 
 void MetalDriver::createBufferObjectR(Handle<HwBufferObject> boh, uint32_t byteCount,
         BufferObjectBinding bindingType, BufferUsage usage) {
-    construct_handle<MetalBufferObject>(boh, *mContext, usage, byteCount);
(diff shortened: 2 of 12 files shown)


filament/backend/src/metal/MetalDriver.h
```diff
@@ -129,7 +129,9 @@ class MetalDriver final : public DriverBase {
             uint32_t offset, uint32_t minIndex, uint32_t maxIndex, uint32_t count);
 
     void finalizeSamplerGroup(MetalSamplerGroup* sg);
-    void enumerateBoundUniformBuffers(const std::function<void(const UniformBufferState&,
+    void enumerateBoundUniformBuffers(const std::function<void(const BufferState&,
+            MetalBuffer*, uint32_t)>& f);
+    void enumerateBoundSsbos(const std::function<void(const BufferState&,
```
pixelflinger, line 134: Should both method be replaced by one generic `enumerateBoundBuffers` that takes a `BufferObjectBinding` instead? I guess it's not relevant for "vertex" buffer, since they're not bound.

filament/backend/src/metal/MetalDriver.mm
```diff
@@ -1635,7 +1702,65 @@
...
+        ssbosToBind[index] = buffer;
+        ssboOffsets[index] = state.offset;
+    });
+    MetalBuffer::bindBuffers(getPendingCommandBuffer(mContext), computeEncoder, SSBO_BINDING_START,
+            MetalBuffer::Stage::COMPUTE, ssbosToBind, ssboOffsets, MAX_SSBO_COUNT);
+
+    [computeEncoder setComputePipelineState:computePipelineState];
+
+    MTLSize threadgroupsPerGrid = MTLSizeMake(workGroupCount.x, workGroupCount.y, workGroupCount.z);
+    // TODO: the threadgroup size should be specified in the Program
```
pixelflinger, line 1758:
```suggestion
    // FIXME: the threadgroup size should be specified in the Program
```

filament/backend/test/test_ComputeBasic.cpp
```diff
@@ -22,25 +22,29 @@ using namespace filament::backend;
 TEST_F(ComputeTest, basic) {
     auto& driver = getDriverApi();
 
-    std::string_view shader_gles310 = {R"(
+    std::string shader_gles310 = {R"(
```
pixelflinger, line 25: why the change to `string` from `string_view`?

filament/backend/test/test_ComputeBasic.cpp
```diff
@@ -92,11 +96,25 @@ void main() {
...
+    float elements[1];
+};
+struct Input_data {
+    float elements[1];
+};
+constant uint3 gl_WorkGroupSize [[maybe_unused]] = uint3(16u, 1u, 1u);
+kernel void main0(device Output_data& output_data [[buffer(0)]],
+        device Input_data& input_data [[buffer(1)]],
+        uint3 gl_GlobalInvocationID [[thread_position_in_grid]]) {
+    output_data.elements[gl_GlobalInvocationID.x] = input_data.elements[gl_GlobalInvocationID.x];
```
pixelflinger, line 113: Since these are backend tests, I don't think you should use the `gl_` nomenclature, it's confusing. You can use any name you want here.

pixelflinger (approved):
Only had very minor comments.

filament/backend/test/test_ComputeBasic.cpp
bejado: The shaders passed to backends need to have a null terminator, which `std::string_view` doesn't guarantee.

filament/backend/test/test_ComputeBasic.cpp
bejado: Yeah that's true. This was exactly the output from Spir-V cross, but I agree it doesn't matter in this case.

PR #6195, opened 2022-10-15
fix(ci): replace set-output to GITHUB_OUTPUT

### LINKED ISSUE #6194: fix(ci): set-output in Github Actions to be deprecated
Hi Team!

https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/
As per this URL, set-output of Github Actions is deprecated.

I would like to create a PR for a fix to this issue.
Thanks.

k1rnt:
resolved #6194

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -39,8 +39,8 @@ jobs:
         run: |
           REF=${RELEASE_TAG:-${GITHUB_REF}}
           TAG=${REF##*/}
-          echo ::set-output name=ref::${REF}
-          echo ::set-output name=tag::${TAG}
+          echo "ref=${REF}" >> $GITHUB_OUTPUT
+          echo "tag=${TAG}" >> $GITHUB_OUTPUT
       - uses: actions/checkout@v2.0.0
         with:
           ref: ${{ steps.git_ref.outputs.ref }}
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -171,8 +171,8 @@ jobs:
         run: |
           REF=${RELEASE_TAG:-${GITHUB_REF}}
           TAG=${REF##*/}
-          echo ::set-output name=ref::${REF}
-          echo ::set-output name=tag::${TAG}
+          echo "ref=${REF}" >> $GITHUB_OUTPUT
+          echo "tag=${TAG}" >> $GITHUB_OUTPUT
```
k1rnt, line 175: Perhaps you need to specify `shell: bash`.
See also : https://github.com/google/go-github/pull/2492#issuecomment-1276469770

.github/workflows/release.yml
```diff
@@ -39,8 +39,8 @@ jobs:
         run: |
           REF=${RELEASE_TAG:-${GITHUB_REF}}
           TAG=${REF##*/}
-          echo ::set-output name=ref::${REF}
-          echo ::set-output name=tag::${TAG}
+          echo "ref=${REF}" >> $GITHUB_OUTPUT
```
bejado, line 42: Can you help me understand what `GITHUB_OUTPUT` is? Is it a custom environment variable? I only see `GITHUB_ENV` listed in [the documentation](https://docs.github.com/en/actions/learn-github-actions/environment-variables).

k1rnt: @bejado Thanks for your comment!
The documentation for `GITHUB_OUTPUT` is a little further down that page here.
https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-output-parameter

k1rnt: This page is [linked](https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/) within the #6194 link It is located a little further down the page.

bejado (approved):
Thanks so much!

PR #6198, opened 2022-10-17
A basic POT 2D allocator

pixelflinger:
This 2D allocator only supports power-of-two, square allocations. It uses a quadtree to store the allocated regions.

The quadtree implementation currently has a fixed depth (templated).

The allocator does a best-fit match, but currently doesn't implement a free method.

### DIFF
--- a/filament/src/AtlasAllocator.cpp
+++ b/filament/src/AtlasAllocator.cpp
@@ -1,0 +1,161 @@
+/*
+ * Copyright (C) 2022 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "AtlasAllocator.h"
+
+#include <utils/debug.h>
+
+namespace filament {
+
+using namespace utils;
+
+static inline constexpr std::pair<uint8_t, uint8_t> unmorton(uint16_t m) noexcept {
+    uint32_t r = (m | (uint32_t(m) << 15u)) & 0x55555555u;
+    r = (r | (r >> 1u)) & 0x33333333u;
+    r = (r | (r >> 2u)) & 0x0f0f0f0fu;
+    r =  r | (r >> 4u);
+    return { uint8_t(r), uint8_t(r >> 16u) };
+}
+
+AtlasAllocator::AtlasAllocator(size_t maxTextureSize) noexcept {
+    // round to power-of-two immediately inferior or equal to the size specified.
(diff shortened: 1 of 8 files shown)


filament/src/AtlasAllocator.h
```diff
@@ -0,0 +1,93 @@
...
+        constexpr bool isAllocated() const noexcept { return allocated; }
+        // whether this node has children. Implies it's not allocated.
+        constexpr bool hasChildren() const noexcept { return children != 0; }
+        // whether this node has all four children. Implies hasChildren().
+        constexpr bool hasAllChildren() const noexcept { return children == 4; }
+        bool allocated      : 1;    // true / false
+        uint8_t children    : 3;    // 0, 1, 2, 3, 4
+    };
+
+    using QuadTree = utils::QuadTreeArray<Node, 4>;
```
bejado, line 55: Use a `constexpr` constant for 4?

filament/src/AtlasAllocator.h
```diff
@@ -0,0 +1,93 @@
...
+    using NodeId = QuadTree::NodeId;
+
+public:
+    /*
+     * Cretate allocator and specify the maximum texture size. Must be a power of two.
```
bejado, line 60:
```suggestion
     * Create allocator and specify the maximum texture size. Must be a power of two.
```

filament/src/AtlasAllocator.cpp
```diff
@@ -0,0 +1,161 @@
...
+static inline constexpr std::pair<uint8_t, uint8_t> unmorton(uint16_t m) noexcept {
+    uint32_t r = (m | (uint32_t(m) << 15u)) & 0x55555555u;
+    r = (r | (r >> 1u)) & 0x33333333u;
+    r = (r | (r >> 2u)) & 0x0f0f0f0fu;
+    r =  r | (r >> 4u);
+    return { uint8_t(r), uint8_t(r >> 16u) };
+}
+
+AtlasAllocator::AtlasAllocator(size_t maxTextureSize) noexcept {
+    // round to power-of-two immediately inferrior or equal to the size specified.
```
bejado, line 34:
```suggestion
    // round to power-of-two immediately inferior or equal to the size specified.
```

libs/utils/include/utils/QuadTree.h
```diff
@@ -0,0 +1,180 @@
...
+
+} // namespace QuadTreeUtils
+
+/**
+ * A Quad-tree encoded as an array.
+ * @tparam T        Type of the quad-tree nodes
+ * @tparam HEIGHT   Height of the quad-tree, muse be <= 4
+ */
+template<typename T, size_t HEIGHT>
+class QuadTreeArray : public std::array<T, QuadTreeUtils::size(HEIGHT)> {
```
bejado, line 93: Does `QuadTreeArray` really need to subclass from `std::array`? It has no member variables and only static methods.

bejado (approved):

libs/utils/include/utils/QuadTree.h
pixelflinger: it gets `operator[]` from it. I wanted to show that it's really just an array.

filament/src/AtlasAllocator.h
pixelflinger: done

filament/src/AtlasAllocator.cpp
pixelflinger: done

filament/src/AtlasAllocator.h
pixelflinger: done

PR #6210, opened 2022-10-20
fgdbg client project structure + simple server connection

feresr:
I've decided to use [compose-mpp](https://www.jetbrains.com/lp/compose-mpp/) mostly because I'm not great with Css/Html/Js and because I like working with Kotlin/Compose/Coroutines

Open to hear feedback about the pros and cons of this approach

Screenshot:
ws


feresr:
code in `libs/fgdbg/client/src/main/kotlin/Main.kt` is likely to change substantially. I'm not even sure we'll use sockets at this point. 
This PR is mostly to set up the client project structure, dependencies, etc. 
And to discuss if this is the right approach for this tool

romainguy:
As the manager of the Compose team I can't be against it 😁

feresr:
@romainguy haha happy to hear that!
Actively learning Compose, don't hold back on feedback 😀

libs/fgdbg/client/build.gradle.kts
```diff
@@ -0,0 +1,46 @@
+import org.jetbrains.compose.compose
+import org.jetbrains.compose.desktop.application.dsl.TargetFormat
+import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
+
+plugins {
+    kotlin("jvm") version "1.6.10"
+    id("org.jetbrains.compose") version "1.1.1"
+}
+
+group = "com.google.filament"
```
feresr, line 10: should I suffix`.fgdbg` here 🤔

libs/fgdbg/client/build.gradle.kts
```diff
@@ -0,0 +1,46 @@
...
+version = "1.0"
+val ktorVersion: String by project
+
+repositories {
+    google()
+    mavenCentral()
+    maven("https://maven.pkg.jetbrains.space/public/p/compose/dev")
+  maven {
```
romainguy, line 18: Indentation is incorrect on this maven block

libs/fgdbg/client/src/main/kotlin/Main.kt
```diff
@@ -0,0 +1,85 @@
+import androidx.compose.material.MaterialTheme
```
romainguy, line 1: Add a copyright/license block

libs/fgdbg/client/src/main/kotlin/Main.kt
```diff
@@ -0,0 +1,85 @@
...
+import io.ktor.client.*
+import io.ktor.client.plugins.websocket.*
+import io.ktor.http.*
+import io.ktor.websocket.Frame
+import io.ktor.websocket.readText
+import kotlinx.coroutines.delay
+
+private enum class ConnectionState { UNKNOWN, CONNECTED, CONNECTING, ERROR }
+
+private data class Connection(
```
romainguy, line 30: Our code style uses 4 spaces instead of 2.

romainguy (commented):
Just small nits

libs/fgdbg/client/src/main/kotlin/Main.kt
romainguy, line 1: More important: tools should go in tools/, not in libs/

romainguy (commented):
Added an important note. Also eventually we'll need a CMake target to build the tool.

libs/fgdbg/client/build.gradle.kts
romainguy: No the group is correct. If we wanted to publish this we would differentiate it by using a separate artifact id containing `fgfbg`.

libs/fgdbg/client/build.gradle.kts
feresr: good catch!

libs/fgdbg/client/src/main/kotlin/Main.kt
feresr: added

feresr: Makes sense, I'll do this on a followup PR to keep this one in scope. 
Should I move everything or just `fgdbg/client`?
I am mostly mirroring what `matdbg` does, should `matdbg/web` go inside `tools/` too?

feresr:
> we'll need a CMake target to build the tool.

This should be interesting. Do you mean have CMake call gradle under the hood?
Or can compose projects be built with cmake too?

bejado (approved):
Excited for this 😎

Issue #6142, opened 2022-10-03
Max morph targets is limited to 256


Obg1:
I cannot use models with higher morph targets.
Maybe you can add cpu to avoid this limit?

[Max morph targets 256 limit](https://github.com/google/filament/pull/5332)

romainguy:
There is no plan to support more morph targets at this point.

Obg1:
@romainguy looking at the code I see that each weight is passed using 4 floats and reading only the first one.
Looks like it can be changed to use all of the 4 floats to pass more data and increase the limit to 256*4 instead of 256.
What do you think? Am I missing something?
https://github.com/google/filament/blob/a64b1eccdf37dde0eb038606225cf69b13b8659d/shaders/src/getters.vs#L99

Obg1:
I've submitted a PR:
https://github.com/google/filament/pull/6214

PR #6214, opened 2022-10-22
Increase the number of morph targets to 1024 instead of 256

Obg1:
I'm using the fact that we are passing float4 so I can pass 4x data for the morph weights

### DIFF
--- a/shaders/src/getters.vs
+++ b/shaders/src/getters.vs
@@ -100,7 +100,9 @@ void morphPosition(inout vec4 p) {
     ivec3 texcoord = ivec3(getVertexIndex() % MAX_MORPH_TARGET_BUFFER_WIDTH, getVertexIndex() / MAX_MORPH_TARGET_BUFFER_WIDTH, 0);
     uint c = getObjectUniforms().morphTargetCount;
     for (uint i = 0u; i < c; ++i) {
-        float w = morphingUniforms.weights[i][0];
+        uint pos = i / 4;
+        uint idx_pos = i % 4;
+        float w = morphingUniforms.weights[pos][idx_pos];
         if (w != 0.0) {
             texcoord.z = int(i);
             p += w * texelFetch(morphTargetBuffer_positions, texcoord, 0);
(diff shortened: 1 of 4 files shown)


Obg1:
https://github.com/google/filament/issues/6142

romainguy:
Please read the comments next to the files you modified. We cannot guarantee this change will work on ES 3.0

romainguy:
Specifically in EngineEnums.h

Obg1:
@romainguy I've reverted it to back 256, but since it's 256xfloat4 I'm allowing to use 256x4 = 1024 morph target.
Let me know if I've made a mistake as it's my first time making a change in Filament code

pixelflinger:
Unfortunately this won't work on all implementations of GLES which has a limit to 256 layers for texture arrays.

Obg1:
@pixelflinger I'm still using 256 layers but using the entire float4 instead of the float[0]. Look at the code changes

shaders/src/getters.vs
```diff
@@ -100,7 +100,9 @@ void morphPosition(inout vec4 p) {
     ivec3 texcoord = ivec3(getVertexIndex() % MAX_MORPH_TARGET_BUFFER_WIDTH, getVertexIndex() / MAX_MORPH_TARGET_BUFFER_WIDTH, 0);
     uint c = getObjectUniforms().morphTargetCount;
     for (uint i = 0u; i < c; ++i) {
-        float w = morphingUniforms.weights[i][0];
+        uint pos = i / 4;
+        uint idx_pos = i % 4;
+        float w = morphingUniforms.weights[pos][idx_pos];
         if (w != 0.0) {
             texcoord.z = int(i);
```
pixelflinger, line 107: This is using more than 256 layers.

pixelflinger (changes requested):

pixelflinger:
> @pixelflinger I'm still using 256 layers but using the entire float4 instead of the float[0]. Look at the code changes

No, you are. I commented on the line in the shader that will access more than 256 layers.

shaders/src/getters.vs
Obg1: pos is devided by 4, so 1024 / 4 is 256 maximum

Obg1:
pos is devided by 4, so 1024 / 4 is 256 maximum

bejado:
> pos is devided by 4, so 1024 / 4 is 256 maximum

The problematic line is:
```
texcoord.z = int(i);
```

Obg1:
Closed as I'm still accessing more than 256 layers

PR #6220, opened 2022-10-24
Check GPU support for depth resolve on Metal

veghbalazs:
Resolving multisampled depth attachments requires GPU support on Metal. This PR adds the necessary checks.

For reference, this feature corresponds to the "MSAA depth resolve" capability in the [Metal Feature Set Tables](https://developer.apple.com/metal/Metal-Feature-Set-Tables.pdf). (Unfortunately the live version of this table has recently been turned incomplete, because macOS 13 dropped support for `.mac1` GPU family. Its capabilities can be checked in [this archived version of the document](https://web.archive.org/web/20220309044948/https://developer.apple.com/metal/Metal-Feature-Set-Tables.pdf)).

Based on our testing, depth resolve implies support for stencil resolve using the `.sample0` [resolve filter](https://developer.apple.com/documentation/metal/mtlmultisamplestencilresolvefilter). Other resolve filters however (like `.min` and `.max`) have stricter requirements - but Filament does not make use of those currently.

### DIFF
--- a/filament/backend/src/metal/MetalHandles.mm
+++ b/filament/backend/src/metal/MetalHandles.mm
@@ -1054,6 +1054,7 @@ void MetalRenderTarget::setUpRenderPassAttachments(MTLRenderPassDescriptor* desc
         descriptor.depthAttachment.slice = 0;
         const bool discard = any(discardFlags & TargetBufferFlags::DEPTH);
         if (!discard) {
+            assert_invariant(context->supportsAutoDepthResolve);
             descriptor.depthAttachment.resolveTexture = depthAttachment.getTexture();
             descriptor.depthAttachment.resolveLevel = depthAttachment.level;
             descriptor.depthAttachment.resolveSlice = depthAttachment.layer;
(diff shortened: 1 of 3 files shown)


filament/backend/src/metal/MetalHandles.mm
```diff
@@ -1086,10 +1087,12 @@ BufferDescriptor b(nullptr, 0u, [](void* buffer, size_t size, void* user) {
         descriptor.stencilAttachment.slice = 0;
         const bool discard = any(discardFlags & TargetBufferFlags::STENCIL);
         if (!discard) {
+            assert_invariant(context->supportsAutoDepthResolve);
             descriptor.stencilAttachment.resolveTexture = stencilAttachment.getTexture();
             descriptor.stencilAttachment.resolveLevel = stencilAttachment.level;
             descriptor.stencilAttachment.resolveSlice = stencilAttachment.layer;
             descriptor.stencilAttachment.storeAction = MTLStoreActionMultisampleResolve;
+            descriptor.stencilAttachment.stencilResolveFilter = MTLMultisampleStencilResolveFilterSample0;
```
bejado, line 1095: I think this will generate a compiler warning since `stencilResolveFilter` is only available on iOS 12+. (and we support iOS 11+). We may need to surround with `@available`.

bejado (changes requested):

filament/backend/src/metal/MetalHandles.mm
veghbalazs: That's right, fixed it in https://github.com/google/filament/pull/6220/commits/dd6f2f2dbfb25c4456c6a4e9b739ffac24a50fc4.

bejado (approved):
Thanks!

PR #6228, opened 2022-10-26
fix the `doubleSided` property documentation

pixelflinger:


bejado (approved):

romainguy:
The original doc was lost at some point. Here's the original:

```
### Rasterization: doubleSided

Type
:    `boolean`

Value
:     `true` or `false`. Defaults to `false`.

Description
:     Enables two-sided rendering and its capability to be toggled at run time. When set to `true`,
      `culling` is automatically set to `none`; if the triangle is back-facing, the triangle's
      normal is flipped to become front-facing. When explicitly set to `false`, this allows the
      double-sidedness to be toggled at run time.

~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ JSON
material {
    doubleSided : true
}
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

romainguy (approved):

PR #6230, opened 2022-10-27
skip primitives with FRONT_AND_BACK face culling

pixelflinger:


bejado (approved):

romainguy (approved):
Could this have unintended side effects though? Like could someone rely on the vertex shader execution?

pixelflinger:
> Could this have unintended side effects though? Like could someone rely on the vertex shader execution?

I was wondering about this. They could even rely on the stencil buffer even maybe?

However, I think this is still the right thing to do because in out API we're calling it "face culling", so if it's culled, it's not drawn. That seems reasonable to me.

wdyt?

romainguy:
Yeah that's reasonable.

PR #6232, opened 2022-10-27
Calculate primitive's AABB correctly

### LINKED ISSUE #6231: AABB is incorrect at some glTF models.
**Describe the bug**

The asset's AABB looks incorrect.

image

**To Reproduce**
1. Download [b.glb.zip](https://github.com/google/filament/files/9877382/b.glb.zip).
2. Unzip b.glb.zip
3. Execute gltf_viewer with b.glb
4. Hierarchy > Show bounds

* This model is optimized by meshoptimizer.

**Expected behavior**

AABB should be fit.

image

daemyung:
AABB of morph target is displacement data so it should be added to primitive's base AABB.

Fix #6231

### DIFF
--- a/libs/gltfio/src/AssetLoader.cpp
+++ b/libs/gltfio/src/AssetLoader.cpp
@@ -964,6 +964,7 @@ bool FAssetLoader::createPrimitive(const cgltf_primitive& inPrim, Primitive* out
         targetsCount = MAX_MORPH_TARGETS;
     }
 
+    const Aabb baseAabb(outPrim->aabb);
     for (cgltf_size targetIndex = 0; targetIndex < targetsCount; targetIndex++) {
         const cgltf_morph_target& morphTarget = inPrim.targets[targetIndex];
         for (cgltf_size aindex = 0; aindex < morphTarget.attributes_count; aindex++) {
@@ -982,10 +983,20 @@ bool FAssetLoader::createPrimitive(const cgltf_primitive& inPrim, Primitive* out
                 return false;
             }
 
+            if (!accessor->has_min || !accessor->has_max) {
+                continue;
+            }
+
+            Aabb targetAabb(baseAabb);
             const float* minp = &accessor->min[0];
             const float* maxp = &accessor->max[0];
-            outPrim->aabb.min = min(outPrim->aabb.min, float3(minp[0], minp[1], minp[2]));
-            outPrim->aabb.max = max(outPrim->aabb.max, float3(maxp[0], maxp[1], maxp[2]));
+
+            // We assume that the range of morph target weight is [-1,1].
+            targetAabb.min += float3(minp[0], minp[1], minp[2]);
+            targetAabb.max += float3(maxp[0], maxp[1], maxp[2]);
+
+            outPrim->aabb.min = min(outPrim->aabb.min, targetAabb.min);
+            outPrim->aabb.max = max(outPrim->aabb.max, targetAabb.max);
 
             VertexBuffer::AttributeType fatype;
             VertexBuffer::AttributeType actualType;
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -5,6 +5,8 @@ A new header is inserted each time a *tag* is created.
 
 ## main branch
 
+- gltfio: calculate primitive's AABB correctly. 
+
 ## v1.28.2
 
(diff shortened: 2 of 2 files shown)


daemyung:
You can get more context at [here](https://github.com/zeux/meshoptimizer/pull/504).

pixelflinger:
@daemyung so how do you solve the problem in the end?

daemyung:
> @daemyung so how do you solve the problem in the end?

@pixelflinger The solution is just using primitive's AABB. So I deleted code that merging morph targets' AABB.

pixelflinger:
> > @daemyung so how do you solve the problem in the end?
> 
> @pixelflinger The solution is just using primitive's AABB. So I deleted code that merging morph targets' AABB.

I guess what I don't understand is why the AABB are correct if you only use the primitive's AABB? we're saying that it is the responsibility of the asset creator to set  primitive AABB that will work with all weights (like we do for skinning)? I'm okay with that, just want to understand.

daemyung:
> I guess what I don't understand is why the AABB are correct if you only use the primitive's AABB? 

Technically it's wrong if we just use primitive's base AABB. Application needs to calculate AABB with morphing. But the problem is that glTF spec doesn't define the range of morph target weight so we can't get the size of model well. For example morph target weight could be 1.0f or 10000.0f. So filament needs to define the range of morph target weight then we can use morph target's AABB. For example [three.js](https://github.com/mrdoob/three.js/blob/868aee4c3bda29077ba53989cea07948a474782e/examples/js/loaders/GLTFLoader.js#L3905) defines this range from -1.0 to 1.0. But filament doesn't have any spec for this so I look into [recomputeBoundingBoxes](https://github.com/google/filament/blob/15ba54c4c51fcf1ff80790f4223984e5db0106cb/libs/gltfio/include/gltfio/FilamentInstance.h#L138) in gltfio and filament doesn't use morph target to calculate model's AABB. So I change code like this.

> it is the responsibility of the asset creator to set primitive AABB that will work with all weights (like we do for skinning)?

I think so too if animations for morphing is defined. Without animations for morphing we can't know the range of morph target weight.

daemyung:
@pixelflinger After talking with you, I have changed my mind. Updated code is calculating AABB with morph target's AABB.

image

pixelflinger (approved):

libs/gltfio/src/AssetLoader.cpp
```diff
@@ -982,10 +983,20 @@ bool FAssetLoader::createPrimitive(const cgltf_primitive& inPrim, Primitive* out
...
+            }
+
+            Aabb targetAabb(baseAabb);
             const float* minp = &accessor->min[0];
             const float* maxp = &accessor->max[0];
-            outPrim->aabb.min = min(outPrim->aabb.min, float3(minp[0], minp[1], minp[2]));
-            outPrim->aabb.max = max(outPrim->aabb.max, float3(maxp[0], maxp[1], maxp[2]));
+
+            // We assume that the range of morph target weight is [-1,1].
+            targetAabb.min += float3(minp[0], minp[1], minp[2]);
```
romainguy, line 995: Why are we adding the mins here?

pixelflinger:
> @pixelflinger After talking with you, I have changed my mind. Updated code is calculating AABB with morph target's AABB assuming range is -1 to 1. Below figure is the result.

I think this makes sense because you can always author the data so it does what you need with weights between -1 and 1.

We should probably document that somewhere. Like saying we allow weights out of the [-1,1]  range, but that bounding boxes in gltfio assume that.

libs/gltfio/src/AssetLoader.cpp
daemyung: I guess morph weight could be -1. So model size could increase opposite direction.

romainguy: I don't see how adding the mins makes sense. This can grow the aabb?

daemyung: I didn't check. But this AABB is displacement data (deltas) so it could be I guess. For example base primitive AABB min is [0, 0, 0] and target AABB(for delta) min is [-1, -1, -1] then final AABB min will be [-1, -1, -1].

romainguy: Ah If it's displacement then it makes more sense.

PR #6235, opened 2022-10-27
Add Vulkan GGP platform.

mattvera:
This PR adds support for a new Vulkan backend that enables the VK_GGP_stream_descriptor_surface vulkan extension and creates the surface using the vkCreateStreamDescriptorSurfaceGGP function provided by that extension.


romainguy:
Thanks for opening a PR. Could you describe what GGP is? And why a different backend is required?

prideout (approved):
Cool.  In our parlance, this adds a "platform" rather than a "backend".

PR #6240, opened 2022-10-28
Preserve iOS stack traces

palver123:
This PR follows up on #624 and adds back unwind tables and frame pointers to the generated code **on iOS**.

Our iOS app uses Filament for PBR rendering and uses Crashlytics for tracking crashes. We faced the problem of not seeing stack frames of Filament code in our crash reports because unwind tables and frame pointers were emitted from the Filament lib:
image

Adding them back fixed the issue without significantly increasing the app size. In our case, our Filament static lib grew from 137.1 MB to 139.5 MB. (The lib file in question includes  x86_64 and arm64 binaries of the following libraries and their dependencies: `backend`, `filabridge`, `filaflat`, `filament`, `ibl`, `image`, `utils`).

We tested these changes on the following platforms:
 - iPadOS (arm64 devices only)
 - iOS Simulator (x86_64 and arm64 host machines too)
 - Mac Catalyst app on macOS (x86_64 and arm64 machines too)

### Would it be worth adding them back on Android too?
We can't test Android and haven't investigated if similar problems would arise on WebGL. Since we are unfamiliar with these platforms we decided not to touch their build settings in Filament.
 
### What about Windows?
For the record, our crash reports on Windows didn't suffer from missing stack Filament stack frames. Either because the MSVC settings are different for Filament (didn't check) or because we use Sentry instead of Crashlytics.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -348,7 +348,12 @@ endif()
 # On Android RELEASE builds, we disable exceptions and RTTI to save some space (about 75 KiB
 # saved by -fno-exception and 10 KiB saved by -fno-rtti).
 if (ANDROID OR IOS OR WEBGL)
-    set(CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_RELEASE} -fno-exceptions -fno-unwind-tables -fno-asynchronous-unwind-tables -fno-rtti")
+    set(CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_RELEASE} -fno-exceptions -fno-rtti")
+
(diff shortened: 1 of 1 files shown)


pixelflinger (approved):
I'm sort of okay with this change, but remember that with `-fno-omit-frame-pointer` the compiler has one less register to work with, which can have a performance impact, especially on 32-bits platforms.

In release builds, the extra "bloat" is not serving the user, only the developper and I'm not sure if it's a good trade-off.

I wouldn't make this change for Android.

bejado (approved):

romainguy:
Filament supports all Android devices. Which is easy since there's no difference for us across those devices. The only requirement is OpenGL ES 3.0.

PR #6242, opened 2022-10-28
Metal: generate argument buffer structures ourselves

bejado:
We're using argument buffers for texture resources, however, we cannot rely on spirv-cross to generate the argument buffer definitions.

Consider a shader with 3 textures:
```
layout (set = 0, binding = 0) uniform sampler2D texture1;
layout (set = 0, binding = 1) uniform sampler2D texture2;
layout (set = 0, binding = 2) uniform sampler2D texture3;
```

If only texture1 and texture2 are used in the material, then texture3 will be optimized away.
This results in an argument buffer like the following:
```
struct spvDescriptorSetBuffer0 {
    texture2d<float> texture1 [[id(0)]];
    sampler texture1Smplr [[id(1)]];
    texture2d<float> texture2 [[id(2)]];
    sampler texture2Smplr [[id(3)]];
};
```

Note that this happens even if the "pad_argument_buffer_resources" and "force_active_argument_buffer_resources" options are true, because the optimization has already happened by the time the SPIR-V gets to spirv-cross.

This would be fine, except older Apple devices don't like it when the argument buffer in the shader doesn't precisely match the one generated at runtime.

So, this PR adds a`MetalArgumentBuffer ` class to replace spirv-cross' argument buffer definitions with our own that contain all the textures/samples, even those that would be optimized away.

### DIFF
--- a/libs/filamat/src/GLSLPostProcessor.cpp
+++ b/libs/filamat/src/GLSLPostProcessor.cpp
@@ -286,10 +286,10 @@ void GLSLPostProcessor::spirvToToMsl(const SpirvBlob *spirv, std::string *outMsl
     *outMsl = minifier.removeWhitespace(*outMsl);
 
     // Replace spirv-cross' generated argument buffers with our own.
-    for (const auto* argBuffer : argumentBuffers) {
+    for (auto* argBuffer : argumentBuffers) {
         auto argBufferMsl = argBuffer->getMsl();
         MetalArgumentBuffer::replaceInShader(*outMsl, argBuffer->getName(), argBufferMsl);
-        delete argBuffer;
+        MetalArgumentBuffer::destroy(&argBuffer);
     }
 }
 
--- a/libs/filamat/src/GLSLPostProcessor.cpp
+++ b/libs/filamat/src/GLSLPostProcessor.cpp
@@ -28,6 +28,8 @@
 #include "shaders/CodeGenerator.h"
 #include "shaders/MaterialInfo.h"
 
+#include "MetalArgumentBuffer.h"
+
 #include <filament/MaterialEnums.h>
 #include <private/filament/Variant.h>
 #include "SibGenerator.h"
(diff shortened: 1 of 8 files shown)


libs/filamat/src/MetalArgumentBuffer.cpp
```diff
@@ -0,0 +1,216 @@
...
+
+        // Theoretically we could continue to find and replace other occurrences, but there should
+        // only ever be a single definition of the argument buffer structure.
+        return true;
+    }
+
+    return false;
+}
+
+} // namespace filamat
```
pixelflinger, line 216: no newline at end of file

libs/filamat/src/MetalArgumentBuffer.h
```diff
@@ -0,0 +1,114 @@
...
+#include <string>
+#include <variant>
+#include <vector>
+
+#include <backend/DriverEnums.h>
+
+namespace filamat {
+
+class MetalArgumentBuffer {
+   public:
```
pixelflinger, line 30: don't indent public/protected/private

libs/filamat/src/MetalArgumentBuffer.h
```diff
@@ -0,0 +1,114 @@
...
+
+    class Builder {
+       public:
```
pixelflinger, line 33: don't indent

libs/filamat/src/MetalArgumentBuffer.h
```diff
@@ -0,0 +1,114 @@
...
+         * @param index the [[id(n)]] index of the texture argument
+         * @param name the name of the texture argument
+         */
+        Builder& sampler(size_t index, const std::string& name) noexcept;
+
+        MetalArgumentBuffer* build();
+
+        friend class MetalArgumentBuffer;
+
+       private:
```
pixelflinger, line 64: don't indent

libs/filamat/src/MetalArgumentBuffer.h
```diff
@@ -0,0 +1,114 @@
...
+     * Searches shader for the target argument buffer, and replaces it with the replacement string.
+     *
+     * @param shader the source MSL shader that contains the target argument buffer definition
+     * @param targetArgBufferName the name of the argument buffer definition to replace
+     * @param replacement the replacement argument buffer definition
+     * @return true if the target was found, false otherwise
+     */
+    static bool replaceInShader(std::string& shader, const std::string& targetArgBufferName,
+            const std::string& replacement) noexcept;
+   private:
```
pixelflinger, line 104: don't indent

pixelflinger (changes requested):

libs/filamat/src/GLSLPostProcessor.cpp
```diff
@@ -262,6 +284,13 @@ void GLSLPostProcessor::spirvToToMsl(const SpirvBlob *spirv, std::string *outMsl
 
     *outMsl = mslCompiler.compile();
     *outMsl = minifier.removeWhitespace(*outMsl);
+
+    // Replace spirv-cross' generated argument buffers with our own.
+    for (const auto* argBuffer : argumentBuffers) {
+        auto argBufferMsl = argBuffer->getMsl();
+        MetalArgumentBuffer::replaceInShader(*outMsl, argBuffer->getName(), argBufferMsl);
+        delete argBuffer;
```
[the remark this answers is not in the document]
bejado: Good idea

pixelflinger (approved):

PR #6243, opened 2022-10-29
Add MaterialInstance getters for overridable properties

pixelflinger:
In addition to the new getters, this change fixes the duplication of a MaterialInstance, which didn't carry along the following states:
  - alpha mask threshold
  - specular AA threshold
  - specular AA variance
  - double sidedness

TODO: stencil state, polygon offset and scissor are stil not queryable


romainguy (approved):
We also need the Java APIs :)

romainguy (approved):

PR #6244, opened 2022-10-29
Fix a build error on Android

daemyung:
The size of struct could be different on each architecture.


daemyung:
@prideout @pixelflinger Is it possible to gurantee same type size on arm7 and arm64?

pixelflinger:
Oh we can just remove this assert, it's not very useful.

pixelflinger (approved):

PR #6246, opened 2022-10-31
Recompute bounding boxes with morph targets

daemyung:
Before `recomputeBoundingBoxes` doesn't consider morph targets. You can see a mesh is bigger than its bounding box below figure.

image

With this PR, bounding boxes are calculated correctly because morph targets are included in recomputing.

image


prideout (commented):
Looks okay to me, especially since this feature is optional.  In practice the glTF author should set min/max appropriately.

One idea for reducing # of heap allocations is to move the `targetData` declaration out of the loop and do a pre-pass to determine maximum size.

daemyung:
>  In practice the glTF author should set min/max appropriately.

Right. Filament still read those data from a file. There is a bug and it's fixed by #6232.

> One idea for reducing # of heap allocations is to move the `targetData` declaration out of the loop and do a pre-pass to determine maximum size.

Thanks. It's very good point. I reuse already allocated buffer. I think it's fine because the base primitive and the morph target must have same count and dimension as far as I know.

PR #6252, opened 2022-10-31
Reduce the size of ShadowMap to 40 bytes from 2KB+

pixelflinger:
ShadowMap had recently inflated due to its internal use of  PerViewUniforms, which keeps a shadow copy of the uniform data. This copy is not needed for ShadowMaps.

### DIFF
--- a/filament/src/PerShadowMapUniforms.cpp
+++ b/filament/src/PerShadowMapUniforms.cpp
@@ -1,0 +1,169 @@
+/*
+ * Copyright (C) 2022 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "PerShadowMapUniforms.h"
+
+#include "ShadowMapManager.h"
+
+#include "details/Camera.h"
+#include "details/Engine.h"
+
+#include <private/filament/EngineEnums.h>
+
+namespace filament {
+
+using namespace backend;
+using namespace math;
+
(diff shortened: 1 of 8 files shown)


filament/src/PerShadowMapUniforms.cpp
```diff
@@ -0,0 +1,169 @@
+/*
+ * Copyright (C) 2021 The Android Open Source Project
```
bejado, line 2:
```suggestion
 * Copyright (C) 2022 The Android Open Source Project
```

bejado (approved):

filament/src/PerShadowMapUniforms.cpp
```diff
@@ -0,0 +1,169 @@
...
+        FEngine& engine,
+        float exposure,
+        float3 const& sceneSpaceDirection,
+        PerShadowMapUniforms::LightManagerInstance directionalLight) noexcept {
+    FLightManager& lcm = engine.getLightManager();
+    auto& s = edit(transaction);
+
+    const float3 l = -sceneSpaceDirection; // guaranteed normalized
+
+    if (directionalLight.isValid()) {
```
romainguy, line 111: Isn't that code duplicated from `PerViewUniforms.cpp` (same with the other `prepare*` methods)?

filament/src/PerShadowMapUniforms.cpp
```diff
@@ -0,0 +1,169 @@
...
+        s.lightColorIntensity = colorIntensity;
+        s.lightChannels = lcm.getLightChannels(directionalLight);
+
+        const bool isSun = lcm.isSunLight(directionalLight);
+        // The last parameter must be < 0.0f for regular directional lights
+        float4 sun{ 0.0f, 0.0f, 0.0f, -1.0f };
+        if (UTILS_UNLIKELY(isSun && colorIntensity.w > 0.0f)) {
+            // currently we have only a single directional light, so it's probably likely that it's
+            // also the Sun. However, conceptually, most directional lights won't be sun lights.
+            float radius = lcm.getSunAngularRadius(directionalLight);
```
romainguy, line 125: Lots of stuff here we technically don't need for shadow maps.

filament/src/PerShadowMapUniforms.cpp
pixelflinger: yes it's duplicated, but conceptually it could be different, so I kept it duplicated (it could even be a different UBO for shadow maps, in theory).

PR #6264, opened 2022-11-03
fix spotlight attenuation

pixelflinger:
Spotlight attenuation was computed only when shadowing was enabled.

### DIFF
--- a/shaders/src/light_punctual.fs
+++ b/shaders/src/light_punctual.fs
@@ -154,17 +154,19 @@ Light getLight(const uint lightIndex) {
     light.worldPosition = positionFalloff.xyz;
     light.channels = channels;
     light.contactShadows = bool(typeShadow & 0x10u);
-#if defined(VARIANT_HAS_SHADOWING)
 #if defined(VARIANT_HAS_DYNAMIC_LIGHTING)
+    light.type = (typeShadow & 0x1u);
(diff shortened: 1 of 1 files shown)


romainguy (approved):

shaders/src/light_punctual.fs
```diff
@@ -154,17 +154,19 @@ Light getLight(const uint lightIndex) {
...
     light.shadowIndex = (typeShadow >>  8u) & 0xFFu;
     light.shadowLayer = (typeShadow >> 16u) & 0xFFu;
     light.castsShadows   = bool(channels & 0x10000u);
-    light.type = (typeShadow & 0x1u);
     if (light.type == LIGHT_TYPE_SPOT) {
-        light.attenuation *= getAngleAttenuation(-direction, light.l, scaleOffset);
         light.zLight = dot(shadowUniforms.shadows[light.shadowIndex].lightFromWorldZ, vec4(worldPosition, 1.0));
     }
 #endif
+    if (light.type == LIGHT_TYPE_SPOT) {
```
romainguy, line 167: Kind of annoying to do the test twice. We could fix that.

pixelflinger: I'm hopping the optimizer will take care of it when both are happening.

PR #6269, opened 2022-11-03
Fix matc variantFilter

bejado:
I was seeing the VSM variant turned on for the Skybox, when it certainly doesn't need to be. I think the `UserVariantFilterBit` values are out of date. We may want to consider a better way of keeping them in sync with `Variant.h`.

### DIFF
--- a/libs/filabridge/include/filament/MaterialEnums.h
+++ b/libs/filabridge/include/filament/MaterialEnums.h
@@ -232,14 +232,15 @@ enum class Property : uint8_t {
     // when adding new Properties, make sure to update MATERIAL_PROPERTIES_COUNT
 };
 
+// These flags should match the equivalents in Variant.h.
 enum class UserVariantFilterBit : uint32_t {
     DIRECTIONAL_LIGHTING        = 0x01,
     DYNAMIC_LIGHTING            = 0x02,
     SHADOW_RECEIVER             = 0x04,
     SKINNING                    = 0x08,
-    FOG                         = 0x10,
-    VSM                         = 0x20,
-    SSR                         = 0x40,
+    FOG                         = 0x20,
+    VSM                         = 0x40,
+    SSR                         = VSM | SHADOW_RECEIVER,
 };
 
 using UserVariantFilterMask = uint32_t;


romainguy (approved):
Good catch, and yeah we should find a way to automate this.

pixelflinger (changes requested):
This change is definitely wrong.

The "UserVariant" are like virtual variant bits. They are not supposed to match the Variant bits. The User Variant" must always be single bits.

And they get "converted" to the more compressed Variant.

This is needed because Variant is no longer "a bit per feature", there are overlaps. But "user variant" is.


You can file a bug on me for the original problem you were debugging.

pixelflinger:
> I was seeing the VSM variant turned on for the Skybox, when it certainly doesn't need to be. I think the `UserVariantFilterBit` values are out of date. We may want to consider a better way of keeping them in sync with `Variant.h`.

When you say you're seeing the "VSM" variant, do you mean that the VSM bit is set, or that you are seeing `Variant::isVSMVariant()` return true? The `VSM` bit can be set in the case of `SSR` for instance (it'll be `SSR|SRE`).

romainguy:
Ugh, good to know. I wonder then if the problem is that our unlit doesn't filter out SSR variants by default?

bejado:
@pixelflinger Here's what I'm seeing:

```
GLSL shaders:
    #0   mobile vs 0x00                 isVSMVariant: false
    #1   mobile fs 0x00                 isVSMVariant: false
    #2   mobile vs 0x10   DEP           isVSMVariant: false
    #3   mobile fs 0x10   DEP           isVSMVariant: false
    #4   mobile fs 0x20   FOG           isVSMVariant: false
    #5   mobile fs 0x30   DEP|FOG       isVSMVariant: false
    #6   mobile fs 0x44   SRE|VSM       isVSMVariant: false
    #7   mobile vs 0x50   DEP|VSM       isVSMVariant: true
    #8   mobile fs 0x50   DEP|VSM       isVSMVariant: true
    #9   desktop vs 0x00                isVSMVariant: false
    #10  desktop fs 0x00                isVSMVariant: false
    #11  desktop vs 0x10   DEP          isVSMVariant: false
    #12  desktop fs 0x10   DEP          isVSMVariant: false
    #13  desktop fs 0x20   FOG          isVSMVariant: false
    #14  desktop fs 0x30   DEP|FOG      isVSMVariant: false
    #15  desktop fs 0x44   SRE|VSM      isVSMVariant: false
    #16  desktop vs 0x50   DEP|VSM      isVSMVariant: true
    #17  desktop fs 0x50   DEP|VSM      isVSMVariant: true
```

I'll investigate further and open a bug for you.

PR #6270, opened 2022-11-04
Add fgdbg::Session.h, Pass.h, Resource.h dtos/models

feresr:
Still in WIP, feel free to take a look and leave suggestions

### DIFF
--- a/filament/src/details/Engine.cpp
+++ b/filament/src/details/Engine.cpp
@@ -260,7 +260,8 @@ void FEngine::init() {
 
 #if FILAMENT_ENABLE_FGDBG
     if (!debug.fgdbg) {
-        debug.fgdbg = std::make_unique<fgdbg::DebugServer>();
+        auto server = std::make_shared<fgdbg::DebugServer>();
+        debug.fgdbg = std::make_unique<fgdbg::Session>("sample", server);
     }
 #endif
 
@@ -652,7 +653,7 @@ int FEngine::loop() {
             static int fgdbgMessage = 0;
             fgdbgMessage++;
             if (fgdbgMessage % 100 == 0) {
-                debug.fgdbg->sendMessage(fgdbgMessage/100);
+                debug.fgdbg->update();
             }
         }
 #endif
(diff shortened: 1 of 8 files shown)


libs/fgdbg/include/fgdbg/Resource.h
```diff
@@ -0,0 +1,30 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef FGDBG_RESOURCE_H
+#define FGDBG_RESOURCE_H
+
+#include <string>
+
+namespace filament::fgdbg {
```
bejado, line 22: Resource.h and Pass.h could probably be consolidated to a single header, unless you're planning on adding a lot more to them.

filament/src/details/Engine.cpp
```diff
@@ -260,7 +260,8 @@ void FEngine::init() {
 
 #if FILAMENT_ENABLE_FGDBG
     if (!debug.fgdbg) {
-        debug.fgdbg = std::make_unique<fgdbg::DebugServer>();
+        auto server = std::shared_ptr<fgdbg::DebugServer>();
+        debug.fgdbg = std::make_unique<fgdbg::Session>("sample", server);
```
feresr, line 264: `Server` is a shared resource among multiple `Sessions`
so it gets injected in the `Session` constructor

libs/fgdbg/include/fgdbg/Resource.h
feresr: I think that's a good idea! What do we usually call the enclosing .h file `models.h`?

filament/src/details/Engine.cpp
```diff
@@ -652,7 +653,7 @@ int FEngine::loop() {
             static int fgdbgMessage = 0;
             fgdbgMessage++;
             if (fgdbgMessage % 100 == 0) {
-                debug.fgdbg->sendMessage(fgdbgMessage/100);
```
feresr, line 655: `sendMessage()` is now private to `server`

pixelflinger (approved):

PR #6271, opened 2022-11-04
unify point and spotlight shadows

pixelflinger:
point light shadows are not sampled just like spotlight shadows, the only difference is that we're calculating the face first to access the corresponding shadowmap data (including the light transform).


shaders/src/light_punctual.fs
```diff
@@ -211,18 +211,15 @@ void evaluatePunctualLights(const MaterialInputs material,
...
             if (light.castsShadows) {
                 uint shadowIndex = light.shadowIndex;
-                highp vec4 shadowPosition;
                 if (light.type == LIGHT_TYPE_POINT) {
                     // point-light shadows are sampled from a direction
                     highp vec3 r = getWorldPosition() - light.worldPosition;
-                    highp uint face = 0u;
-                    // getShadowPosition returns zLight which is needed for PCSS/DPCF
-                    shadowPosition = getShadowPosition(r, shadowIndex, light.zLight, face);
+                    highp uint face = getPointLightFace(r);
```
romainguy, line 217: Why is the face `highp`? Won't it be `0..5`?

shaders/src/light_punctual.fs
```diff
@@ -211,18 +211,15 @@ void evaluatePunctualLights(const MaterialInputs material,
...
                     shadowIndex += face;
-                } else {
-                    // getShadowPosition needs zLight for applying the normal bias
-                    shadowPosition = getShadowPosition(false, shadowIndex, 0u, light.zLight);
+                    light.zLight = dot(shadowUniforms.shadows[shadowIndex].lightFromWorldZ,
+                            vec4(getWorldPosition(), 1.0));
                 }
+                highp vec4 shadowPosition = getShadowPosition(false, shadowIndex, 0u, light.zLight);
```
romainguy, line 222: `getShadowPosition` calls `getSpotLightSpacePosition`, should we rename the later to `getPointLightSpacePosition`?

shaders/src/light_punctual.fs
pixelflinger: good catch, it's an oversight.

shaders/src/light_punctual.fs
pixelflinger: well, in the shader there is no concept of "point light position". Once we have the proper "shadowIndex", it's just like a spotlight -- so it calls getSpotLightSpacePosition(). We could rename it to `getDynamicLightPosition()` (dynamic as opposed to directional).

filament/src/ShadowMap.cpp
```diff
@@ -413,41 +413,10 @@ ShadowMap::ShaderParameters ShadowMap::updateDirectional(FEngine& engine,
...
-    // if the scene was empty, near > far
-    mHasVisibleShadows = -sceneInfo.lsNearFar[0] < -sceneInfo.lsNearFar[1];
-
-    // FIXME: we need a configuration for minimum near plane (for now hardcoded to 1cm)
-    float nearPlane = std::max(0.01f, -sceneInfo.lsNearFar[0]);
-    float farPlane  = std::min(radius, -sceneInfo.lsNearFar[1]);
-
-    float outerConeAngleDegrees = outerConeAngle * f::RAD_TO_DEG;
-    const mat4f Mp = mat4f::perspective(outerConeAngleDegrees * 2.0f, 1.0f, nearPlane, farPlane);
+ShadowMap::ShaderParameters ShadowMap::updateSpotOrPoint(
```
bejado, line 416: Could we rename to `updatePunctual`?

bejado (approved):

PR #6280, opened 2022-11-09
Include the windows.h header after packing alignment

daemyung:
Fix winnt.h(2501): error C2118: negative subscript.

### DIFF
--- a/third_party/libsdl2/include/SDL_syswm.h
+++ b/third_party/libsdl2/include/SDL_syswm.h
@@ -33,12 +33,6 @@
 #include "SDL_video.h"
 #include "SDL_version.h"
 
-#include "begin_code.h"
-/* Set up for C function definitions, even when using C++ */
-#ifdef __cplusplus
-extern "C" {
-#endif
-
 /**
  *  \file SDL_syswm.h
  *
@@ -108,6 +102,12 @@ typedef void *EGLSurface;
 
 #if defined(SDL_VIDEO_DRIVER_VIVANTE)
 #include "SDL_egl.h"
+#endif
+
+#include "begin_code.h"
+/* Set up for C function definitions, even when using C++ */
+#ifdef __cplusplus
+extern "C" {
 #endif
 
 /**


daemyung:
I have build errors on the specific build environment and it's caused by SDL2. It's fixed by [it](https://github.com/libsdl-org/SDL/commit/c04dca0dad2696d7dcd51427c8183b19ebff8a60). I hope to accept upgrading SDL2 into Filament.

romainguy:
You are making massive changes to tnt/CMakeLists.txt and I’m not comfortable with that. I’d rather we took the opportunity to get rid of SDL2. We’d rather use something simpler and smaller like glfw or whatever else. It’s just for sample apps.

daemyung:
@romainguy I agree. What I done is that copying base CMakeLists.txt from SDL2 and apply some changes for Filament. Anyway if I cherry pick [this patch](https://github.com/libsdl-org/SDL/commit/c04dca0dad2696d7dcd51427c8183b19ebff8a60) on sdl2 in filament then will you comfortable more before switching to glfw or whatever else? Or switching glfw is a better option? Please let me know your idea. Thanks!

daemyung:
@romainguy Also I have tried to modify tnt/CMakeLists.txt but it's very hard. Because there are too many changes on SDL2. So please understand this situation.

romainguy:
I'd rather we cherry-picked the patch for now and tried to move to not SDL2 as the next step.

@pixelflinger any opinion?

pixelflinger:
Why do we need this update in the first place?

pixelflinger:
I agree with @romainguy it looks like the patch might be easy to cherry pick.

bejado:
Definitely in favor of taking the patch instead of upgrading all of SDL

daemyung:
> Why do we need this update in the first place?

I have to support x86 on Windows and SDL2 has some build errors. I know Filament doesn't support x86 on Windows. But these changes are regardless that.

daemyung:
Ok I will cherry-pick these changes.

pixelflinger (approved):

bejado (approved):

PR #6286, opened 2022-11-10
rework how we set the viewport for shadows

pixelflinger:
The viewport is now entirely handled by ShadowMap, that is, we're no longer setting a viewport in the renderpass, instead the ShadowMap uses a post projection transform to achieve the same effect. This will become useful when store shadow maps in an Atlas.

### DIFF
--- a/filament/src/ShadowMap.cpp
+++ b/filament/src/ShadowMap.cpp
@@ -368,8 +368,8 @@ ShadowMap::ShaderParameters ShadowMap::updateDirectional(FEngine& engine,
 
     // Computes St the transform to use in the shader to access the shadow map texture
     // i.e. it transforms a world-space vertex to a texture coordinate in the shadowmap
-    const mat4 MbMt = getTextureCoordsMapping(shadowMapInfo);
-    const mat4f St = mat4f(MbMt * S);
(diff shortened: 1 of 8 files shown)


filament/src/ShadowMap.cpp
```diff
@@ -620,22 +619,12 @@ mat4 ShadowMap::getTextureCoordsMapping(ShadowMapInfo const& info) noexcept {
...
-    // factor
-    const float v = float(info.textureDimension) / float(info.atlasDimension);
-    const mat4f Mv(mat4f::row_major_init{
-            v,    0.0f, 0.0f, 0.0f,
-            0.0f, v,    0.0f, 0.0f,
-            0.0f, 0.0f, 1.0f, 0.0f,
-            0.0f, 0.0f, 0.0f, 1.0f
-    });
-
     // apply the 1-texel border viewport transform
```
bejado, line 622: Is this comment still relevant?

bejado (approved):

filament/src/ShadowMap.cpp
pixelflinger: not really :-)

PR #6307, opened 2022-11-18
Handle shadow borders better

pixelflinger:
All shadow maps need a 1-px border for different reasons. The directional shadow needs a "not in shadow" border because it uses the intersection of receivers and casters in light-space. Other shadow types need a border because of bilinear access at the  edges. Until now, the spot/point shadow border was handled with the sampler's CLAMP.

Instead, now, we always generate a border and in one case we fill it with "not in shadow" (by not rendering into it), in the other cases we render the 2px larger shadowmap so we get completely correct bilinear  filter.

Additionally, for PCF, DPCF and PCSS we use a large filter kernel which accesses data outside the texture, until now this was handled with CLAMP, but that failed (2 of the edges were not handled in the same way) when  the shadowmap was smaller than the texture.
We fix this by clamping manually the texture coordinates.

The real motivation for these changes is to allow the use of a shadow Atlas later.

### DIFF
--- a/filament/src/ShadowMap.cpp
+++ b/filament/src/ShadowMap.cpp
@@ -368,7 +368,8 @@ ShadowMap::ShaderParameters ShadowMap::updateDirectional(FEngine& engine,
 
     // Computes St the transform to use in the shader to access the shadow map texture
     // i.e. it transforms a world-space vertex to a texture coordinate in the shadowmap
-    const auto [Mt, Mn] = ShadowMap::getTextureCoordsMapping(shadowMapInfo, getViewport());
+    const backend::Viewport viewport = getViewport();
+    const auto [Mt, Mn] = ShadowMap::getTextureCoordsMapping(shadowMapInfo, viewport);
     const mat4f St = math::highPrecisionMultiply(Mt, S);
 
     ShadowMap::ShaderParameters shaderParameters;
@@ -389,6 +390,8 @@ ShadowMap::ShaderParameters ShadowMap::updateDirectional(FEngine& engine,
     } else {
         shaderParameters.lightSpace = computeVsmLightSpaceMatrix(St, Mv, znear, zfar);
     }
+
+    shaderParameters.scissorNormalized = getViewportNormalized(shadowMapInfo);
 
     // We apply the constant bias in world space (as opposed to light-space) to account
     // for perspective and lispsm shadow maps. This also allows us to do this at zero-cost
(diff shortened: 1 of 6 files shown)


filament/src/ShadowMap.cpp
```diff
@@ -1197,11 +1205,34 @@ backend::Viewport ShadowMap::getViewport() const noexcept {
...
+        case ShadowType::DIRECTIONAL:
+            return { border, border, dim - 2u * border, dim - 2u * border };
+        case ShadowType::SPOT:
+        case ShadowType::POINT:
+        default:
+            return { 0, 0, dim, dim };
+    }
+}
+
+math::float4 ShadowMap::getViewportNormalized(ShadowMapInfo const& shadowMapInfo) const noexcept {
```
bejado, line 1228: Maybe we can think of name besides `viewport`? Since Filament treats a Viewport as {left, bottom, width, height} but this returns {left, bottom, right, top}.

filament/src/ShadowMap.cpp
```diff
@@ -1197,11 +1205,34 @@ backend::Viewport ShadowMap::getViewport() const noexcept {
     // We set a viewport with a 1-texel border for when we index outside the
     // texture. This can only happen for the directional light when "focus shadow casters is used".
     const uint32_t dim = mOptions->mapSize;
-    if (isDirectionalShadow()) {
-        return { 1, 1, dim - 2, dim - 2 };
-    } else {
-        return { 0, 0, dim, dim };
+    const uint16_t border = 1u;
+    return { border, border, dim - 2u * border, dim - 2u * border };
```
bejado, line 1209: Could we merge `getViewport` and `getScissor` into a single function? Do we really need a separate `getViewport` function?

shaders/src/shadowing.fs
```diff
@@ -443,13 +467,18 @@ float evaluateShadowVSM(const highp vec2 moments, const highp float depth) {
...
+        const highp vec4 scissorNormalized,
         const uint layer, const highp vec4 shadowPosition) {
 
     // note: shadowPosition.z is in linear light-space normalized to [0, 1]
     //  see: ShadowMap::computeVsmLightSpaceMatrix() in ShadowMap.cpp
     //  see: computeLightSpacePosition() in common_shadowing.fs
     highp vec3 position = vec3(shadowPosition.xy * (1.0 / shadowPosition.w), shadowPosition.z);
 
+    // Note: we don't need to clamp to `scissorNormalized` in the VSM case because this is only
+    // needed when the shadow casters and receivers are differents, which is never the case with VSM
```
bejado, line 479:
```suggestion
    // needed when the shadow casters and receivers are different, which is never the case with VSM
```

bejado (approved):

PR #6312, opened 2022-11-21
Box and sphere IBLs with tinting

gaborvalasek:
### Overview

This PR expands Filament IBLs in two ways:
* it adds IBL color tinting capabilities to alter the color of reflections
* it adds two types of local IBLs: spheres and axis-aligned boxes

### Tinting

Our application allows users to take screenshots with solid and gradient color backgrounds. We use custom IBLs and it was visually displeasing that the background color was not incorporated into the reflections.

Without IBL tinting, this is how two metallic objects would look like in front of a red background:
image

The following screenshot shows how the previous scene looks like using a full tint:
image

The weight of the tint can be customized (with a scaler in the range of [0,1], see the above screenshot). 

This does alter the intensity of the IBL but for our particular application, this proved to be a desired property. 

The tinting itself is an overlay blend using artist established constants for split-value. Tint weight is left as a user-adjustable parameter.

(While you are testing, please note that ACES does drastic hue shifts on fully saturated colors, so you may see a mismatch between the postprocessed background color and the reflected colors due to this.)

### Localized IBLs

Reflections on low roughness bodies with parallel planar faces look awkward with infinite IBLs, that is, what is currently available. For example, the bottom-left of the following screenshot is vague in terms of how many objects are there:
image

By using localized IBLs, this ambiguity may be resolved. If the IBL is mapped onto a spherical proxy geometry, we get
image

while an axis-aligned box provides reflections such as shown below:
image

This PR contains an optimized version of the sphere-ray intersection that only works correctly if the ray is inside the IBL sphere (which is a reasonable restriction for IBLs). In terms of AMD ISA assembly, it requires 34 full rate and 2 quarter rate (rcp and sqrt) instructions. The numerically stable way of computing both roots would require 42 full rate and 3 quarter rate in ISA. 

For rays outside the IBL sphere, the difference is shown below. The full computation of both roots and selecting the closest one along the ray looks as:
image

The optimized computation shown in the PR renders the following image:
image

In terms of performance, the current solution uses a runtime branch to determine the type of the IBL. Whether this is acceptable is up to the reviewers to decide.

### DIFF
--- a/filament/include/filament/IndirectLight.h
+++ b/filament/include/filament/IndirectLight.h
@@ -20,6 +20,7 @@
 #define TNT_FILAMENT_INDIRECTLIGHT_H
 
 #include <filament/FilamentAPI.h>
+#include <filament/Options.h>
 
 #include <utils/compiler.h>
 
@@ -342,6 +343,68 @@ public:
     /** @deprecated use static versions instead */
     UTILS_DEPRECATED
     math::float4 getColorEstimate(math::float3 direction) const noexcept;
+
+    /**
+     * Sets IBL lookup options.
+     *
+     * @param options Options for IBL lookup settings.
+     */
+    void setIblOptions(IblOptions const& options) noexcept;
+
+    /**
+     * Gets IBL lookup options.
+     *
+     * @return IBL lookup options.
+     */
+    IblOptions const& getIblOptions() const noexcept;
+
+    /**
+     * Sets IBL type.
+     *
+     * @param iblTechnique Determine the IBL type (infinite, or finite box/sphere)
+     */
+    void setIblTechnique(const IblOptions::IblTechnique iblTechnique) noexcept;
+
+    /**
+     * Gets IBL type.
+     */
+    IblOptions::IblTechnique getIblTechnique() const noexcept;
+
+    /**
+     * Sets the center of the finite IBL proxy geometry.
+     *
+     * @param iblCenter The center of the finite IBL proxy geometry
+     */
+    void setIblCenter(const math::float3& iblCenter) noexcept;
+
+    /**
+     * Gets IBL center.
+     */
+    const math::float3& getIblCenter() const noexcept;
+
+    /**
+     * Sets the half extents of the finite IBL proxy geometry.
+     *
+     * @param iblCenter The half extents of the finite IBL proxy geometry
+     */
+    void setIblHalfExtents(const math::float3& iblHalfExtents) noexcept;
+
+    /**
+     * Gets IBL half extents.
+     */
+    const math::float3& getIblHalfExtents() const noexcept;
+
+    /**
+     * Sets the IBL tint (.rgb) and blend intensity (.a)
+     *
+     * @param iblTintAndIntensity The tint color (.rgb) and blend weight between shaded and tinted (.a)
+     */
+    void setIblTintAndIntensity(const math::float4& iblTintAndIntensity) noexcept;
+
+    /**
+     * Gets IBL tint (.rgb) and blend intensity (.a).
+     */    
+    const math::float4& getIblTintAndIntensity() const noexcept;    
 };
 
 } // namespace filament
--- a/filament/include/filament/IndirectLight.h
+++ b/filament/include/filament/IndirectLight.h
@@ -349,62 +349,62 @@ public:
      *
      * @param options Options for IBL lookup settings.
      */
-    void setIblOptions(IblOptions const& options) noexcept;
+    void setOptions(IblOptions const& options) noexcept;
 
     /**
      * Gets IBL lookup options.
      *
      * @return IBL lookup options.
      */
-    IblOptions const& getIblOptions() const noexcept;
+    IblOptions const& getOptions() const noexcept;
 
     /**
      * Sets IBL type.
      *
      * @param iblTechnique Determine the IBL type (infinite, or finite box/sphere)
      */
-    void setIblTechnique(const IblOptions::IblTechnique iblTechnique) noexcept;
+    void setTechnique(const IblOptions::IblTechnique iblTechnique) noexcept;
 
     /**
      * Gets IBL type.
      */
-    IblOptions::IblTechnique getIblTechnique() const noexcept;
+    IblOptions::IblTechnique getTechnique() const noexcept;
 
     /**
      * Sets the center of the finite IBL proxy geometry.
      *
      * @param iblCenter The center of the finite IBL proxy geometry
      */
-    void setIblCenter(const math::float3& iblCenter) noexcept;
+    void setCenter(const math::float3& iblCenter) noexcept;
 
     /**
      * Gets IBL center.
      */
-    const math::float3& getIblCenter() const noexcept;
+    const math::float3& getCenter() const noexcept;
 
     /**
      * Sets the half extents of the finite IBL proxy geometry.
      *
      * @param iblCenter The half extents of the finite IBL proxy geometry
      */
-    void setIblHalfExtents(const math::float3& iblHalfExtents) noexcept;
+    void setHalfExtents(const math::float3& iblHalfExtents) noexcept;
 
     /**
      * Gets IBL half extents.
      */
-    const math::float3& getIblHalfExtents() const noexcept;
+    const math::float3& getHalfExtents() const noexcept;
 
     /**
      * Sets the IBL tint (.rgb) and blend intensity (.a)
      *
-     * @param iblTintAndIntensity The tint color (.rgb) and blend weight between shaded and tinted (.a)
+     * @param iblTintAndStrength The tint color (.rgb) and blend weight between shaded and tinted (.a)
      */
-    void setIblTintAndIntensity(const math::float4& iblTintAndIntensity) noexcept;
+    void setTintAndStrength(const math::float4& iblTintAndStrength) noexcept;
 
     /**
      * Gets IBL tint (.rgb) and blend intensity (.a).
      */    
-    const math::float4& getIblTintAndIntensity() const noexcept;    
+    const math::float4& getTintAndStrength() const noexcept;    
 };
 
 } // namespace filament
--- a/filament/include/filament/IndirectLight.h
+++ b/filament/include/filament/IndirectLight.h
@@ -96,6 +96,10 @@ class UTILS_PUBLIC IndirectLight : public FilamentAPI {
 
 public:
 
+    enum class Mode : uint32_t {
+        IBL_INFINITE, IBL_FINITE_SPHERE, IBL_FINITE_BOX
+    };
+
     //! Use Builder to construct an IndirectLight object instance
     class Builder : public BuilderBase<BuilderDetails> {
         friend struct BuilderDetails;
@@ -346,23 +350,9 @@ public:
     math::float4 getColorEstimate(math::float3 direction) const noexcept;
 
     /**
-     * Sets IBL lookup options.
-     *
-     * @param options Options for IBL lookup settings.
-     */
-    void setOptions(IblOptions const& options) noexcept;
-
-    /**
-     * Gets IBL lookup options.
-     *
-     * @return IBL lookup options.
-     */
-    IblOptions const& getOptions() const noexcept;
-
-    /**
      * Gets IBL type.
      */
-    IblOptions::IblTechnique getTechnique() const noexcept;
+    IndirectLight::Mode getTechnique() const noexcept;
 
     /**
      * Sets IBL type to sphere.
@@ -402,7 +392,7 @@ public:
      *
      * @param iblCenter The center of the finite IBL proxy geometry
      */
-    void setCenter(const math::float3& iblCenter) noexcept;
+    void setCenter(const math::float3& inIblCenter) noexcept;
 
     /**
      * Gets IBL center.
@@ -414,7 +404,7 @@ public:
      *
      * @param iblCenter The half extents of the finite IBL proxy geometry
      */
-    void setHalfExtents(const math::float3& iblHalfExtents) noexcept;
+    void setHalfExtents(const math::float3& inIblHalfExtents) noexcept;
 
     /**
      * Gets IBL half extents.
@@ -426,7 +416,7 @@ public:
      *
      * @param iblTintAndStrength The tint color (.rgb) and blend weight between shaded and tinted (.a)
      */
-    void setTintAndStrength(const math::float4& iblTintAndStrength) noexcept;
+    void setTintAndStrength(const math::float4& inIblTintAndStrength) noexcept;
 
     /**
      * Gets IBL tint (.rgb) and blend intensity (.a).
--- a/filament/include/filament/IndirectLight.h
+++ b/filament/include/filament/IndirectLight.h
@@ -19,6 +19,7 @@
 #ifndef TNT_FILAMENT_INDIRECTLIGHT_H
 #define TNT_FILAMENT_INDIRECTLIGHT_H
 
+#include <filament/Box.h>
 #include <filament/FilamentAPI.h>
 #include <filament/Options.h>
 
@@ -359,16 +360,42 @@ public:
     IblOptions const& getOptions() const noexcept;
 
     /**
-     * Sets IBL type.
-     *
-     * @param iblTechnique Determine the IBL type (infinite, or finite box/sphere)
-     */
-    void setTechnique(const IblOptions::IblTechnique iblTechnique) noexcept;
-
-    /**
      * Gets IBL type.
      */
     IblOptions::IblTechnique getTechnique() const noexcept;
+
+    /**
+     * Sets IBL type to sphere.
+     *
+     * @param The center (.xyz) and radius (.w) of the IBL sphere proxy geometry
+     */
+    void setSphereProxy(const filament::math::float4& sphere);
+
+    /**
+     * Gets the properties of sphere IBL.
+     *
+     * @return The center (.xyz) and radius (.w) of the IBL sphere proxy geometry
+     */
+    filament::math::float4 getSphereProxy() const;
+
+    /**
+     * Sets IBL type to box.
+     *
+     * @param The IBL box proxy geometry
+     */
+    void setBoxProxy(const filament::Box& box);
+
+    /**
+     * Gets the properties of box IBL.
+     *
+     * @return The filament::Box representing the the IBL box proxy geometry
+     */
+    filament::Box getBoxProxy() const;
+
+    /**
+     * Reverts back to infinite IBLs.
+     */
+    void clearProxy();
 
     /**
      * Sets the center of the finite IBL proxy geometry.
--- a/filament/include/filament/Options.h
+++ b/filament/include/filament/Options.h
@@ -451,7 +451,7 @@ struct IblOptions {
     /**
      * Center of the sphere or box IBL geometry.
      */
-    math::float3 iblCenter;
+    math::float3 iblCenter = math::float3(0.0f);
 
     /**
      * If iblTechnique is IBL_INFINITE, this property is unused.
@@ -462,7 +462,7 @@ struct IblOptions {
      * 
      * If iblTechnique is IBL_FINITE_BOX, then the half extents of the box along the X, Y, Z axes.
      */
-    math::float3 iblHalfExtents;
+    math::float3 iblHalfExtents = math::float3(100.0f);
 
     /**
      * The color (.rgb) and the intensity (.a) of the IBL tint.
--- a/filament/include/filament/Options.h
+++ b/filament/include/filament/Options.h
@@ -438,38 +438,6 @@ struct SoftShadowOptions {
     float penumbraRatioScale = 1.0f;
 };
 
-/**
- * View-level options for local IBLs.
- */
-struct IblOptions {    
-    enum class IblTechnique : uint32_t {
-        IBL_INFINITE, IBL_FINITE_SPHERE, IBL_FINITE_BOX
-    };
-
-    IblTechnique iblTechnique = IblTechnique::IBL_INFINITE;
-
-    /**
-     * Center of the sphere or box IBL geometry.
-     */
-    math::float3 iblCenter = math::float3(0.0f);
-
-    /**
-     * If iblTechnique is IBL_INFINITE, this property is unused.
-     * 
-     * If iblTechnique is IBL_FINITE_SPHERE, then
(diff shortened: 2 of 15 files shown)


filament/include/filament/IndirectLight.h
```diff
@@ -342,6 +343,68 @@ class UTILS_PUBLIC IndirectLight : public FilamentAPI {
...
+     * @return IBL lookup options.
+     */
+    IblOptions const& getIblOptions() const noexcept;
+
+    /**
+     * Sets IBL type.
+     *
+     * @param iblTechnique Determine the IBL type (infinite, or finite box/sphere)
+     */
+    void setIblTechnique(const IblOptions::IblTechnique iblTechnique) noexcept;
```
romainguy, line 366: If the `IblTechnique` type is part of `IblOptions`, why have separate getters/setters here? Actually why aren't all the new APIs part of `IblOptions`? That would be consistent with other similar APIs.

shaders/src/common_lighting.fs
```diff
@@ -71,6 +75,64 @@ float computeMicroShadowing(float NoL, float visibility) {
     return microShadow * microShadow;
 }
 
+/**
+ * IBL utilities
+ */
+
+vec2 IntersectAABB(vec3 rayOrigin, vec3 rayDir, vec3 boxMin, vec3 boxMax) {
```
romainguy, line 82: We don't use upper case as the first letter of functions in our code style.

shaders/src/light_indirect.fs
```diff
@@ -13,12 +13,31 @@
...
+    const float kT = 0.65;
+    const float kOneOverT = 1.0 / kT;
+    return (a < kT) ? a * kOneOverT * b : a - kT + b;
+}
+vec3 OverlayBlend(vec3 a, vec3 b) {
+    return vec3(OverlayBlend(a.x, b.x), OverlayBlend(a.y, b.y), OverlayBlend(a.z, b.z));
+}
+
+vec3 TintIbl(vec3 color) {
+    return mix(color, OverlayBlend(color, frameUniforms.iblTintAndIntensity.rgb), frameUniforms.iblTintAndIntensity.w);
```
romainguy, line 32: @pixelflinger Not fond of adding that cost unconditionally on mobile. It's not huge but it's not nothing either. Wdyt?

shaders/src/common_lighting.fs
```diff
@@ -71,6 +75,64 @@ float computeMicroShadowing(float NoL, float visibility) {
...
+        // Normalize sphere-space by scaling down positions by radius. We don't scale down ray direction to preserve 
+        // the convenient A = 1 in the quadratic formula. iblHalfExtents.y contains the reciprocal of the IBL sphere radius.
+        vec3 rayPosNormalized = rayPos * frameUniforms.iblHalfExtents.y;
+
+        float B = 2.0 * dot(rayPosNormalized, rayDir);
+        float C = dot(rayPosNormalized, rayPosNormalized) - 1.0; // 1.0 = r^2, as we are in normalized space
+
+        t0 = 0.5 * (-B + sqrt(B*B - 4.0 * C));
+        t0 *= frameUniforms.iblHalfExtents.x;
+    }
```
romainguy, line 124: } and else on the same line

shaders/src/common_lighting.fs
```diff
@@ -71,6 +75,64 @@ float computeMicroShadowing(float NoL, float visibility) {
...
+    
+    if (frameUniforms.iblTechnique == IBL_TECHNIQUE_FINITE_SPHERE) {
+        // Normalize sphere-space by scaling down positions by radius. We don't scale down ray direction to preserve 
+        // the convenient A = 1 in the quadratic formula. iblHalfExtents.y contains the reciprocal of the IBL sphere radius.
+        vec3 rayPosNormalized = rayPos * frameUniforms.iblHalfExtents.y;
+
+        float B = 2.0 * dot(rayPosNormalized, rayDir);
+        float C = dot(rayPosNormalized, rayPosNormalized) - 1.0; // 1.0 = r^2, as we are in normalized space
+
+        t0 = 0.5 * (-B + sqrt(B*B - 4.0 * C));
```
romainguy, line 122: Spaces around * (for B*B)

shaders/src/light_indirect.fs
```diff
@@ -13,12 +13,31 @@
...
+// IBL tinting helpers
+//------------------------------------------------------------------------------
+
+float OverlayBlend(float a, float b) {
+    // Visually almost-luminance-preserving formulation. This uses an artist-provided constant below
+    // that they've found the visually most pleasing under _most_ (but not all) circumstances.
+    const float kT = 0.65;
+    const float kOneOverT = 1.0 / kT;
+    return (a < kT) ? a * kOneOverT * b : a - kT + b;
+}
```
romainguy, line 26: Add empty line between functions

libs/filabridge/include/private/filament/UibStructs.h
```diff
@@ -161,8 +161,19 @@ struct PerViewUib { // NOLINT(cppcoreguidelines-pro-type-member-init)
     float ssrDistance;                  // ssr world raycast distance, 0 when ssr is off
     float ssrStride;                    // ssr texel stride, >= 1.0
 
+    // --------------------------------------------------------------------------------------------
+    // Localized, tinted IBL parameters
+    // --------------------------------------------------------------------------------------------
+    math::float3 iblCenter;           // center of the sphere or IBL AABB
```
romainguy, line 167: This change will require to bump the materials version number as it creates a binary incompatibility.

@bejado

filament/src/details/View.cpp
```diff
@@ -385,6 +385,7 @@ void FView::prepareLighting(FEngine& engine, FEngine::DriverApi& driver, ArenaSc
     }
 
     mPerViewUniforms.prepareAmbientLight(engine, *ibl, intensity, exposure);
+    mPerViewUniforms.prepareIblLight(ibl->getIblOptions());
```
romainguy, line 388: Could you fold those two functions instead? prepareAmbientLight() gets a reference to the IBL and has access to IblOptions. There's no need for a second, separate function.

filament/include/filament/IndirectLight.h
romainguy: Using only `IblOptions` would allow to get rid of all the APIs you add below, making the PR much simpler.

romainguy (commented):
Thank you for this contribution! I left a few comments, mostly stylistic nits, but the main one is about just using the `IblOptions` struct instead of adding many new methods.

romainguy:
One more note: can you update RELEASE_NOTES.md to describe your changes as well?

shaders/src/common_lighting.fs
pixelflinger, line 82: All these probably need to be `highp`. Also please use the `const` decorator where possible.

shaders/src/common_lighting.fs
```diff
@@ -71,6 +75,64 @@ float computeMicroShadowing(float NoL, float visibility) {
...
+    vec3 tMax = (boxMax - rayOrigin) / rayDir;
+    vec3 t1 = min(tMin, tMax);
+    vec3 t2 = max(tMin, tMax);
+    float tNear = max3(t1);
+    float tFar = min3(t2);
+    return vec2(tNear, tFar);
+}
+
+// Assume: a <= b
+float GetSmallestPositive(float a, float b) {
```
pixelflinger, line 93: I would use `#define` instead so that you don't have to deal with the `precision` :-(

shaders/src/common_lighting.fs
```diff
@@ -89,9 +151,9 @@ vec3 getReflectedVector(const PixelParams pixel, const vec3 v, const vec3 n) {
     float bendFactor          = abs(pixel.anisotropy) * saturate(5.0 * pixel.perceptualRoughness);
     vec3  bentNormal          = normalize(mix(n, anisotropicNormal, bendFactor));
 
-    vec3 r = reflect(-v, bentNormal);
+    vec3 r = GetAdjustedReflectedDirection(-v, bentNormal);
 #else
-    vec3 r = reflect(-v, n);
+    vec3 r = GetAdjustedReflectedDirection(-v, n);
```
pixelflinger, line 156: maybe update the comment of this function.

pixelflinger:
I like the IBL types change, but I'm not yet sure it's correct with respect to `shading_normal`. e.g. should we be just applying it there?  I need more time to think about it.

I'm less sure about the tinting change. Can you explain more why you need it? Why not just generate the correct (tinted) IBL?
My problem with it is that it's a feature that will virtually never be used and yet it always costs. I'm not sure how to avoid this though.


Could we separate these two changes into two distinct PRs?

filament/include/filament/IndirectLight.h
```diff
@@ -342,6 +343,68 @@ class UTILS_PUBLIC IndirectLight : public FilamentAPI {
...
+     * Sets the half extents of the finite IBL proxy geometry.
+     *
+     * @param iblCenter The half extents of the finite IBL proxy geometry
+     */
+    void setIblHalfExtents(const math::float3& iblHalfExtents) noexcept;
+
+    /**
+     * Gets IBL half extents.
+     */
+    const math::float3& getIblHalfExtents() const noexcept;
```
pixelflinger, line 395: Instead of `IblTechnique` I would simply use `IndirectLight::Mode`: `INFINITY`, `BOX`, `SPHERE`. 

I don't think we need a setter for the `Mode`, you could just have:

`void setSphereProxy(math::float4 sphere)`
`void setBoxProxy(Box const& box)`
`void clearProxy()`

`Mode getMode() const`
`math::float4 getSphereProxy()`
`Box const& getBoxProxy()`

filament/include/filament/IndirectLight.h
```diff
@@ -342,6 +343,68 @@ class UTILS_PUBLIC IndirectLight : public FilamentAPI {
...
+
+    /**
+     * Sets the IBL tint (.rgb) and blend intensity (.a)
+     *
+     * @param iblTintAndIntensity The tint color (.rgb) and blend weight between shaded and tinted (.a)
+     */
+    void setIblTintAndIntensity(const math::float4& iblTintAndIntensity) noexcept;
```
pixelflinger, line 402: Because this API is on `IndirectLight` you don't need `Ibl` in the name. I would just go with `setTint(float4)`.
I would not use the word "intensity" because it sounds like it's the intensity of the ibl. Maybe use `Strength`.

filament/include/filament/Options.h
```diff
@@ -438,6 +438,38 @@ struct SoftShadowOptions {
     float penumbraRatioScale = 1.0f;
 };
 
+/**
+ * View-level options for local IBLs.
+ */
+struct IblOptions {    
+    enum class IblTechnique : uint32_t {
+        IBL_INFINITE, IBL_FINITE_SPHERE, IBL_FINITE_BOX
```
pixelflinger, line 446: As implemented the mode "box" doesn't seem to work with `IndirectLight::setRotation`.

filament/src/details/View.cpp
pixelflinger: Agreed we only need `prepareAmbientLight`.

shaders/src/light_indirect.fs
```diff
@@ -59,6 +78,8 @@ vec3 Irradiance_SphericalHarmonics(const vec3 n) {
         + frameUniforms.iblSH[8] * (n.x * n.x - n.y * n.y)
 #endif
         , 0.0);
+
+    return TintIbl(rawResult);
```
pixelflinger, line 82: In theory we should be able to apply the tinting on the SH themselves.

shaders/src/light_indirect.fs
pixelflinger: Maybe we should make this a material option? I think that would make sense because it's going to be used to rarely.

pixelflinger (changes requested):

shaders/src/light_indirect.fs
gaborvalasek: Unfortunately, overlay blend is not linear (`OverlayBlend(sum_i coeff[i](**n**) * SH[i], c) != sum_i coeff[i](**n**) * OverlayBlend(SH[i], c)`), so we can't do that (theoretically speaking). I would have preferred a simple multiplicative blend for this very reason but we couldn't make ends meet visually with those, we had to let the bright parts keep their un-tinted color, which led to the nonlinear overlay blend. Now, in practice, it may not matter much, we can do a comparison if you'd like to.

In terms of always paying for this: in general, this might indeed be better as a material option/permutation.

shaders/src/common_lighting.fs
gaborvalasek: I'm amending mine (6dc6210f556cee2f8ce36cfd404588c5ddb3fa2d) but why are `Irradiance_SphericalHarmonics`, `Irradiance_RoughnessOne`, and `PrefilteredDFG_LUT` cased differently? Is it OK if I fix up those too to start with lower-case or camel-cased functions are to be handled and cased differently? If so, what does the prefix/postfix signify?

shaders/src/light_indirect.fs
gaborvalasek: WDYT about adding a material attribute that effectively corresponds to `kT` of `overlayBlend` (the highest intensity that we still overlay blend with the tint color)? I'd `#define` it and if that's zero I'm not tinting anything, otherwise, I'm tinting and now kT becomes a user adjustable parameter.

shaders/src/common_lighting.fs
gaborvalasek: Fixed in 6dc6210f556cee2f8ce36cfd404588c5ddb3fa2d

shaders/src/common_lighting.fs
gaborvalasek: Fixed in 6dc6210f556cee2f8ce36cfd404588c5ddb3fa2d

shaders/src/light_indirect.fs
gaborvalasek: Fixed in 6dc6210f556cee2f8ce36cfd404588c5ddb3fa2d

filament/src/details/View.cpp
gaborvalasek: Fixed in 4cf1c080a6ba99c26eccebabd0426674d0eff062

shaders/src/common_lighting.fs
gaborvalasek: Added these in 6dc6210f556cee2f8ce36cfd404588c5ddb3fa2d . Shall I `highp` up all local variables in these functions too? I'm less worried about the spherical IBLs as the intersection computations are normalized to [-1,1]^3 but the box IBL's computation domain can be arbitrarily large. Although if you are OK with me using another 3 floats in the UBO to supply the reciprocal half extents, then I can normalize there too. Does this sound reasonable?

shaders/src/common_lighting.fs
gaborvalasek: I've manually inlined it in 6dc6210f556cee2f8ce36cfd404588c5ddb3fa2d, as there was a single call site left for it.

shaders/src/common_lighting.fs
gaborvalasek: Added some clarification in 6dc6210f556cee2f8ce36cfd404588c5ddb3fa2d .

filament/include/filament/IndirectLight.h
gaborvalasek: Fixed in 82f894ddd62738097d91afc491a1ba7f5228564a

filament/include/filament/Options.h
gaborvalasek: I believe what you saw was the sphere's reciprocal radius overwriting the Y half extent of the box. 3adc8ec5279767b07c6389d192e69952a6d90b07 fixes this up.

pixelflinger:
We've fixed a CI issue, would you be able to rebase this branch? Thank you!

filament/include/filament/Options.h
pixelflinger: No I mean, that if we call setRotation() the IBL will be rotated but not the box (it's not a problem for the sphere).

filament/include/filament/IndirectLight.h
```diff
@@ -342,6 +343,68 @@ class UTILS_PUBLIC IndirectLight : public FilamentAPI {
...
+     * @param options Options for IBL lookup settings.
+     */
+    void setOptions(IblOptions const& options) noexcept;
+
+    /**
+     * Gets IBL lookup options.
+     *
+     * @return IBL lookup options.
+     */
+    IblOptions const& getOptions() const noexcept;
```
pixelflinger, line 359: We don't need set/getOptions

filament/include/filament/IndirectLight.h
gaborvalasek: Fixed in b0f2ef543036f99a374befdcb44e38fbfa2b8df5 and 901f2a025aa0c1be79f2c56d51f7833d533bb4e4

filament/include/filament/IndirectLight.h
gaborvalasek: Fixed in b0f2ef543036f99a374befdcb44e38fbfa2b8df5 and 901f2a025aa0c1be79f2c56d51f7833d533bb4e4

filament/include/filament/IndirectLight.h
gaborvalasek: Fixed in b0f2ef543036f99a374befdcb44e38fbfa2b8df5 and 901f2a025aa0c1be79f2c56d51f7833d533bb4e4 + rebased it.

filament/include/filament/IndirectLight.h
```diff
@@ -342,6 +348,80 @@ class UTILS_PUBLIC IndirectLight : public FilamentAPI {
     /** @deprecated use static versions instead */
     UTILS_DEPRECATED
     math::float4 getColorEstimate(math::float3 direction) const noexcept;
+
+    /**
+     * Gets IBL type.
+     */
+    IndirectLight::Mode getTechnique() const noexcept;
```
pixelflinger, line 355: either `Mode` should be called `Technique`, or this should `getMode`. I think I prefer `getMode()`. Also the comment should be more descriptive and not use the word "type". (technique vs mode vs type, it's confusing).

Actually, I wonder if we should just call this `ProxyType` and `getProxyType()`.

filament/include/filament/IndirectLight.h
```diff
@@ -342,6 +348,80 @@ class UTILS_PUBLIC IndirectLight : public FilamentAPI {
...
+     *
+     * @param iblCenter The half extents of the finite IBL proxy geometry
+     */
+    void setHalfExtents(const math::float3& inIblHalfExtents) noexcept;
+
+    /**
+     * Gets IBL half extents.
+     */
+    const math::float3& getHalfExtents() const noexcept;
+
```
pixelflinger, line 413: Remove these API, they're not needed anymore.
(you can keep the getters on the `FIndirectLight` side)

filament/src/details/IndirectLight.h
```diff
@@ -57,13 +58,51 @@ class FIndirectLight : public IndirectLight {
...
+     * If iblTechnique is IBL_FINITE_SPHERE, then
+     *      .x is the radius of the sphere
+     *      .y is the reciprocal of the radius of the sphere
+     *
+     * If iblTechnique is IBL_FINITE_BOX, then the half extents of the box along the X, Y, Z axes.
+     */
+    math::float3 iblHalfExtents = math::float3(100.0f);
+
+    // The color (.rgb) and the intensity (.a) of the IBL tint.
+    math::float4 iblTintAndStrength = math::float4(0.0f);
```
pixelflinger, line 104: Please use the proper naming convention (`m` prefix, camel case).
Also why is the default value of halfExtents 100?
Use this syntax:
  `math::type mName{ value };`

libs/filabridge/include/private/filament/UibStructs.h
```diff
@@ -161,8 +161,19 @@ struct PerViewUib { // NOLINT(cppcoreguidelines-pro-type-member-init)
...
+    uint32_t padding3;
+
+    math::float3 iblHalfExtents;      // sphere radius and inverse radius; box half extents along X, Y, Z axes
+    uint32_t iblTechnique;            // 0: infinite cubemap IBL, 1: (origin centered) finite sphere IBL, 2: (origin centered) finite cube IBL
```
pixelflinger, line 171: don't forget to rename to `proxyType` to match the cpu side.

shaders/src/common_lighting.fs
```diff
@@ -72,6 +76,59 @@ float computeMicroShadowing(float NoL, float visibility) {
...
+
+vec2 intersectAABB(const highp vec3 rayOrigin, const highp vec3 rayDir, const highp vec3 boxMin, const highp vec3 boxMax) {
+    vec3 tMin = (boxMin - rayOrigin) / rayDir;
+    vec3 tMax = (boxMax - rayOrigin) / rayDir;
+    vec3 t1 = min(tMin, tMax);
+    vec3 t2 = max(tMin, tMax);
+    float tNear = max3(t1);
+    float tFar = min3(t2);
+    return vec2(tNear, tFar);
+}
```
pixelflinger, line 91: I believe all this needs to be `highp`

shaders/src/common_lighting.fs
```diff
@@ -72,6 +76,59 @@ float computeMicroShadowing(float NoL, float visibility) {
...
+
+/**
+ * This function returns an IBL lookup direction, taking into account the current IBL type (e.g. infinite spherical, 
+ * finite/local sphere/box), an initial intended lookup direction (baseDir) and the particular normal we compute
+ * reflections against (e.g. either the interpolated surface or clearcoat normal).
+ */
+vec3 getAdjustedReflectedDirection(const highp vec3 baseDir, const highp vec3 normal) {
```
pixelflinger, line 98: all this probably needs to be `highp`.

pixelflinger (changes requested):

pixelflinger:
@gaborvalasek most builds are failing.

pixelflinger:
@gaborvalasek I still want to separate the 2 features in differents PR.

pixelflinger:
After more discussions internally, we'd like to close this PR as it is. We do like the new IBL modes but the API needs more work and it needs to take the ibl rotation into account. The IBL tinting's use is still unclear and certainly not a common case, so it might be more appropriate to have it as a amterial parameter.
Regardless these should all be split in separate PRs.
Thank you for contributing!

PR #6313, opened 2022-11-21
Use Frisvad's method to compute tangent space from normals-only mesh input

ViktorHeisenberger:
This PR introduces Frisvad's method for computing a tangent space from a normal vector for input meshes. It provides a smoother result, than the current one, having only one singularity for normals in the -Y direction.

Comparison of the current and new method:
TBN_rand_vs_frisvad


romainguy (approved):
Thank you, this looks great! Marking as approved, but I'd like confirmation from @pixelflinger and @bejado

pixelflinger (approved):

pixelflinger (changes requested):
Looks great. Thanks!  I think this one deserves updating the `RELEASE_NOTES.md` file.

ViktorHeisenberger:
> Looks great. Thanks! I think this one deserves updating the `RELEASE_NOTES.md` file.

Thanks! Updated in d315ec3

romainguy (approved):
Looks great, thank you so much!

PR #6316, opened 2022-11-22
gltfio: Store default transform for animated entities

### LINKED ISSUE #6311: gltf animation has wrong orientation if rotation is non zero and no rotation sampler is used
**Describe the bug**
object during scaling animation has wrong orientation if rotation is non zero and no rotation sampler is used. Setting a single rotation keyframe(resulting in a constant sampler) resolves the issue.

**To Reproduce**
1. load the file `keyframe_rotation_no.gltf`
[keyframe_rotation.zip](https://github.com/google/filament/files/10055922/keyframe_rotation.zip)

2. notice that when no animation is played, the beams are parallel
3. play the animation which uniformly scales one of the beams
4. the beams in the animation should be parallel but they are not. The animating beam now seems to have zero rotation.

**Expected behavior**
Since no rotation sampler is used, the rotation should remain unchanged between animating and not animating. This is the case in other gltf viewers such as donmccurdy's or babylonjs.
The file `keyframe_rotation_yes.gltf` has a constant rotation sampler and works as expected.

**Screenshots**
Screenshot 2022-11-21 at 11 58 08
Screenshot 2022-11-21 at 11 58 35

- tested on filament-1.26.0, filament-1.28.3, https://google.github.io/filament/viewer/
- model exported by Khronos glTF Blender I/O v3.3.32

**Additional context**
We encountered this issue when trying to optimise our animations using gltfpack. We assume it removes animation tracks only containing a single keyframe that has no effect on the already set rotation.

bejado:
If an entity has its scale animated to 0 its transformation matrix can't store rotation information, so it effectively "resets" its orientation. To fix, we store the original, unanimated transform matrix.

Fixes #6311

### DIFF
--- a/libs/gltfio/src/Animator.cpp
+++ b/libs/gltfio/src/Animator.cpp
@@ -57,6 +57,7 @@ struct Sampler {
 struct Channel {
     const Sampler* sourceData;
     Entity targetEntity;
+    math::mat4f restTransform;  // the default transform of targetEntity, with no animation applied
     enum { TRANSLATION, ROTATION, SCALE, WEIGHTS } transformType;
 };
 
@@ -417,9 +418,11 @@ void AnimatorImpl::addChannels(const FixedCapacityVector<Entity>& nodeMap,
             }
             continue;
         }
+        TransformManager::Instance node = transformManager->getInstance(targetEntity);
         Channel dstChannel;
         dstChannel.sourceData = samplers + (srcChannel.sampler - srcSamplers);
         dstChannel.targetEntity = targetEntity;
+        dstChannel.restTransform = transformManager->getTransform(node);
         setTransformType(srcChannel, dstChannel);
         dst.channels.push_back(dstChannel);
     }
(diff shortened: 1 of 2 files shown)


libs/gltfio/src/Animator.cpp
```diff
@@ -434,7 +437,7 @@ void AnimatorImpl::applyAnimation(const Channel& channel, float t, size_t prevIn
     // Perform the interpolation. This is a simple but inefficient implementation; Filament
     // stores transforms as mat4's but glTF animation is based on TRS (translation rotation
     // scale).
-    mat4f xform = transformManager->getTransform(node);
+    mat4f xform = channel.restTransform;
```
pixelflinger, line 440: I'm not sure if this is correct. This only works if a `Channel` maps to *one* Entity. But for example if 2 `Channel` can animate the same Entity (e.g. one animates the scale and one the rotation), this doesn't work.

On the other hand, if there is a one to one relation between `Channel` and `Entity`, then this code a way more complex and costly than it needs. We could instead store a union of T, R or S. We wouldn't need to do the `decomposeMatrix` (i.e. with this code we don't need it, since it's a constant).

So imho, either this fix is incorrect or the code is more complex than it needs.

pixelflinger (changes requested):
I think this is not correct if several channels animate the same transform, or the code is too complex if not.

pixelflinger:
I think the right way to do this is to create a TRS manager which is a "dump" component manager that just stores the TRS for an entity. Then in `applyTransform` we retrieve and animate/interpolate that, and then from that we construct the xform which we set in the regular transform manager.

So we don't have to decode the entity's transfrom into TRS each time.

libs/gltfio/src/Animator.cpp
bejado: Yeah, you're right. A `Channel` can affect more than one Entity. I'll have to think of a more robust solution.

PR #6320, opened 2022-11-25
Add RGTC + BPTC support

palver123:
Filament already supports S3TC. I assumed the naming conventions followed that of [EXT_texture_compression_s3tc](https://registry.khronos.org/OpenGL/extensions/EXT/EXT_texture_compression_s3tc.txt).

Based on the existing S3TC support we extended Filament with RGTC and BPTC support because we use Filament with BC textures on macOS and Windows. The relevant OpenGL extensions are [EXT_texture_compression_rgtc](https://registry.khronos.org/OpenGL/extensions/EXT/EXT_texture_compression_rgtc.txt) and [EXT_texture_compression_bptc](https://registry.khronos.org/OpenGL/extensions/EXT/EXT_texture_compression_bptc.txt).

We tested it with Metal backend (on Intel and ARM64 macs) and on Windows (with OpenGL backend). Didn't test the Vulkan backend.


libs/ktxreader/include/ktxreader/Ktx1Reader.h
```diff
@@ -51,7 +51,6 @@ namespace Ktx1Reader {
     PixelDataFormat toPixelDataFormat(const KtxInfo& info);
     bool isCompressed(const KtxInfo& info);
     TextureFormat toTextureFormat(const KtxInfo& info);
-    TextureFormat toSrgbTextureFormat(TextureFormat tex);
```
palver123, line 54: The original implementation of this function contained code that would have to be updated now, but I couldn't find the implementation anymore. So I deleted the declaration. Let me know if I missed sth!

docs/webgl/reference.html
```diff
@@ -1140,6 +1140,14 @@ <h2>enum <a id='CompressedPixelDataType' href='#CompressedPixelDataType'>Compres
 <li>DXT1_RGBA</li>
 <li>DXT3_RGBA</li>
 <li>DXT5_RGBA</li>
+<li>RED_RGTC1</li>
```
palver123, line 1143: I wasn't sure if this file has to be kept up-to-date. Let me know if I should revert

libs/ktxreader/src/Ktx2Reader.cpp
```diff
@@ -69,9 +69,6 @@ struct FinalFormatInfo {
 //     transcoder_texture_format::cTFPVRTC2_4_RGBA
 //     transcoder_texture_format::cTFPVRTC1_4_RGB
 //     transcoder_texture_format::cTFPVRTC1_4_RGBA
-//     transcoder_texture_format::cTFBC4_R
-//     transcoder_texture_format::cTFBC5_RG
-//     transcoder_texture_format::cTFBC7_RGBA (this format would add size bloat to the transcoder)
```
palver123, line 74: I am somewhat concerned about this comment. Should I not add BC7 to the supported formats here?

filament/backend/include/backend/DriverEnums.h
```diff
@@ -369,6 +369,18 @@ enum class CompressedPixelDataType : uint16_t {
...
+    SIGNED_RED_RGTC1,       // BC4 signed
+    RED_GREEN_RGTC2,        // BC5 unsigned
+    SIGNED_RED_GREEN_RGTC2, // BC5 signed
+
+    // BPTC formats available with a GLES extension
+    RGB_BPTC_SIGNED_FLOAT,  // BC6H signed
+    RGB_BPTC_UNSIGNED_FLOAT,// BC6H unsigned
+    RGBA_BPTC_UNORM,        // BC7
+    SRGB_ALPHA_BPTC_UNORM,  // BC7 sRGB
+
```
pixelflinger, line 383: Move these at the end of the list so that it doesn't change the enum values of existing formats.

filament/backend/include/backend/DriverEnums.h
```diff
@@ -523,6 +535,18 @@ enum class TextureFormat : uint16_t {
...
```
pixelflinger, line 549: - move at the end of the list
- update `Texture.java`

pixelflinger (changes requested):

filament/backend/include/backend/DriverEnums.h
palver123: Oops I've been excluding the `android` folder when searching the code because I didn't realize it contains code too (that's how I missed Texture.java). Checked the whole folder and only Texture.java needs a little update, done in 559ec06bb211a5c6eba48a09367b7364b05bf17e.

As for the enum order, see 1067be3b4f445feb370a1296e3f3d4a10306be8e

pixelflinger: These enums in java must match the C++ ones, so they also need to be moved at the end.

pixelflinger (changes requested):
java enums must match C++ enums

filament/backend/include/backend/DriverEnums.h
palver123: The enum values are the same in DriverEnums.h and Texture.java.. What sort of mismatch do you mean?

pixelflinger: Well I mean that they must be at the end below `SRGB8_ALPHA8_ASTC_12x12`. The java code uses the order of the enums to convert them to the C++ enum. That's a bit fragile, but that's how it's done currently.
Do I make sense?

palver123: 🤦 I only now realized I forgot to update `TextureFormat`. Sorry, fixed in a52753b6f73ed3ef3b98ec384de0139ce41dcc64

pixelflinger (approved):

filament/backend/src/opengl/GLUtils.h
```diff
@@ -472,6 +472,34 @@ constexpr /* inline */ GLenum getInternalFormat(TextureFormat format) noexcept {
             return 0;
 #endif
 
+#if defined(GL_EXT_texture_compression_rgtc)
+        case TextureFormat::RED_RGTC1:            return GL_COMPRESSED_RED_RGTC1_EXT;
```
romainguy, line 476: Small nitpick: the returns are misaligned here

libs/ktxreader/src/Ktx2Reader.cpp
romainguy: We should try and see how big the transcoder becomes with extra support for those formats. They are not helpful on mobile nor for WebGL so it can be a concern (our main priority remains mobile and binary size matters a lot there).

romainguy (approved):

libs/ktxreader/src/Ktx2Reader.cpp
palver123: I quickly checked the size of libktxreader.a on iOS, release build. It increased by 2408 bytes (from 88272). This seems acceptable to me but we are not so much pressured by app size right now, so my definition of 'acceptable' might be different.

I don't know how to check the binary size on WebGL or Android builds so I propose 3 options:
 - The above numbers on the iOS binary size is convincing enough to merge the PR as it is
 - I revert the changes in this file and someone else can add it if needed. We do use Basis Universal compressed textures in .ktx files but we use https://github.com/KhronosGroup/KTX-Software for that
 - I put the changes in this file behind a macro. If so, what is the appropriate macro? `#if defined(__EMSCRIPTEN__) || defined(ANDROID) || defined(IOS)`?

filament/backend/src/opengl/GLUtils.h
palver123: f18d72e34936c4f40df4d0ed09dd818017822c5e

PR #6328, opened 2022-11-30
delete list_licenses

ccworld1000:
delete list_licenses

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -554,7 +554,7 @@ function(list_licenses OUTPUT MODULES)
         endif()
     endforeach()
     configure_file(${FILAMENT}/build/licenses.inc.in ${OUTPUT})
-endfunction(list_licenses)
+endfunction()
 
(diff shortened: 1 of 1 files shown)


pixelflinger:
Can we get more context please?

ccworld1000:
syntax error @pixelflinger

romainguy:
It's not necessary anymore but perfectly valid even in CMake 3.25. https://cmake.org/cmake/help/latest/command/endfunction.html

We can take the change there's no downside to it though. It should trigger a syntax error though (and doesn't for us?)

pixelflinger:
@ccworld1000 you need to sign the CLA for us to be able to commit the change.

ccworld1000:
Due to some restrictions or special reasons, I cannot access the "CLA" association.
You can help me handle it manually or turn off the changes directly.

@pixelflinger

pixelflinger:
closing, will be handled by #6384

PR #6360, opened 2022-12-07
Make all OpenGLPlatform related interfaces public

pixelflinger:
- Remove DefaultPlatform from public headers
DefaultPlatform is now PlatormFactory and is just an implementation
detail.

- VulkanPlatform is now public
- Improvements to the OpenGLPlatform interface
- Cleanup PlatformEGL and PlatformEGLAndroid

Also:

- reworked how external image worked so that all the implementation
  is contained in the Platform and no dependency to the backend is
  needed.

- MetalPlatform is now entirely private and final

This is because MetalPlatform wasn't an interface. We probably need one
so it could be customized in the future. But as it stands now it's
better to be private.


romainguy (approved):

romainguy (approved):

filament/backend/include/private/backend/OpenGLPlatform.h
```diff
@@ -25,83 +25,222 @@ namespace filament::backend {
...
     virtual SwapChain* createSwapChain(void* nativeWindow, uint64_t& flags) noexcept = 0;
 
-    // headless swapchain
+    /**
+     * Called by the driver create a headless SwapChain.
+     *
+     * @param width     width of the buffer
+     * @param height    height of the buffer
+     * @param flags         extra flags used by the implementation, see filament::SwapChain
+     * @return          The driver's SwapChain object.
```
bejado:
```suggestion
     * @param width     width of the buffer
     * @param height    height of the buffer
     * @param flags     extra flags used by the implementation, see filament::SwapChain
     * @return          The driver's SwapChain object.
```

bejado (approved):

PR #6364, opened 2022-12-08
Add fallback return to uintToColorDebug()

veghbalazs:
If materials are compiled in debug (`-gd`) then the following Metal shader compilation warning appears at runtime:
```
program_source:552:1: warning: control may reach end of non-void function.
``` 
It originates from `common_graphics.fs` where `uintToColorDebug()` is missing a fallback case.


romainguy (approved):
Thank you!

PR #6368, opened 2022-12-09
Transparency in screen-space refraction

veghbalazs:
Hi! In this PR we extended lighting calculations with the ability to retain transparency in screen-space refractive materials. That is, objects with such a material now don't show up as fully opaque in the final image when the background is cleared to `[0,0,0,0]` and `View` is set to `BlendMode::TRANSLUCENT`. 

The use-case: our users can create models and assign materials from a library of fine-tuned presets - including refractive ones. We wanted to let them create snapshots exported as PNG that look reasonable when composited on a background of any color.

It's been quite challenging to figure out how an alpha value sampled from the background should change as the "transparent ray of light" passes through an object with a given base color, transmission etc. So we'd love to hear what you think, it's a rather unconventional extension to PBR.

### More details
We chose our formulas such that results remain unchanged when sampled alpha is `1.0`. We modified `evaluateRefraction()` to take the sampled alpha channel into consideration. Texture `light_ssr` is now read into a `vec4`. For this, we had to set the texture's format to `TextureFormat::RGBA16F` when `View::BlendMode` is `TRANSLUCENT`. 
​
After the background texture is sampled, the color components (`Ft`) of the background are shifted towards white based on if we have a non-opaque sampled alpha (`at`). You can think of it as turning "transparent light" into "white light", which helps in retaining the base color of the materials.

Subsequent steps that adjust `Ft` also update `at` (base color, Fresnel, absorption, transmission).

Finally, the fragment shader only outputs the computed alpha if `BLEND_MODE_OPAQUE` is true. Other blend modes are unsupported: computed alpha value should be stored in the render target directly, alpha-compositing on top of some existing color would be incorrect and pointless. (Notice that an opaque refractive material may "cut a transparent hole" in the render target if the transparent background shows up in the refracted contents).

### Example screenshot
This is how the same snapshot looks before and after these changes - as seen in an image editor (checkerboard), and in front of colorful backgrounds.
cylinders

The original PNGs can be downloaded from here:
[cylinders-transparency.zip](https://github.com/google/filament/files/10208304/cylinders-transparency.zip)

### DIFF
--- a/filament/src/details/Renderer.cpp
+++ b/filament/src/details/Renderer.cpp
@@ -842,7 +842,7 @@ void FRenderer::renderJob(ArenaScope& arena, FView& view) {
 
     PostProcessManager::ScreenSpaceRefConfig const ssrConfig = PostProcessManager::prepareMipmapSSR(
             fg, svp.width, svp.height,
-            ssReflectionsOptions.enabled ? TextureFormat::RGBA16F : TextureFormat::R11F_G11F_B10F,
+            (ssReflectionsOptions.enabled || view.getBlendMode() == View::BlendMode::TRANSLUCENT) ? TextureFormat::RGBA16F : TextureFormat::R11F_G11F_B10F,
             view.getCameraUser().getFieldOfView(Camera::Fov::VERTICAL), config.scale);
     config.ssrLodOffset = ssrConfig.lodOffset;
     blackboard["ssr"] = ssrConfig.ssr;
--- a/shaders/src/light_indirect.fs
+++ b/shaders/src/light_indirect.fs
@@ -472,7 +472,7 @@ void refractionThinSphere(const PixelParams pixel,
     ray.d = d;
 }
 
-vec3 evaluateRefraction(
+vec4 evaluateRefraction(
     const PixelParams pixel,
     const vec3 n0, vec3 E) {
 
@@ -515,9 +515,8 @@ vec3 evaluateRefraction(
     // when reading from the cubemap, we are not pre-exposed so we apply iblLuminance
     // which is not the case when we'll read from the screen-space buffer
     vec3 Ft = prefilteredRadiance(ray.direction, perceptualRoughness) * frameUniforms.iblLuminance;
+    float at = 1.0;
 #else
-    vec3 Ft;
-
     // compute the point where the ray exits the medium, if needed
     vec4 p = vec4(getClipFromWorldMatrix() * vec4(ray.position, 1.0));
     p.xy = uvToRenderTargetUV(p.xy * (0.5 / p.w) + 0.5);
@@ -525,25 +524,32 @@ vec3 evaluateRefraction(
     // distance to camera plane
     const float invLog2sqrt5 = 0.8614;
     float lod = max(0.0, (2.0f * log2(perceptualRoughness) + frameUniforms.refractionLodOffset) * invLog2sqrt5);
-    Ft = textureLod(light_ssr, vec3(p.xy, 0.0), lod).rgb;
+    vec4 Fat = textureLod(light_ssr, vec3(p.xy, 0.0), lod);
+
+    // if it's a transparent background, shift color components towards white
+    vec3 Ft = mix(vec3(1.0), Fat.rgb, Fat.a);
+    float at = Fat.a;
 #endif
 
     // base color changes the amount of light passing through the boundary
     Ft *= pixel.diffuseColor;
+    at += (1.0 - at) * (1.0 - luminance(pixel.diffuseColor.rgb));
 
     // fresnel from the first interface
     Ft *= 1.0 - E;
+    at += (1.0 - at) * luminance(E);
 
     // apply absorption
 #if defined(MATERIAL_HAS_ABSORPTION)
     Ft *= T;
+    at += (1.0 - at) * (1.0 - luminance(T));
 #endif
 
-    return Ft;
+    return vec4(Ft, at);
 }
 #endif
 
-void evaluateIBL(const MaterialInputs material, const PixelParams pixel, inout vec3 color) {
+void evaluateIBL(const MaterialInputs material, const PixelParams pixel, inout vec3 color, inout float alpha) {
     // specular layer
     vec3 Fr = vec3(0.0f);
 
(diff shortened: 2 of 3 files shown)


shaders/src/light_indirect.fs
```diff
@@ -515,35 +515,41 @@ vec3 evaluateRefraction(
...
     p.xy = uvToRenderTargetUV(p.xy * (0.5 / p.w) + 0.5);
 
     // distance to camera plane
     const float invLog2sqrt5 = 0.8614;
     float lod = max(0.0, (2.0f * log2(perceptualRoughness) + frameUniforms.refractionLodOffset) * invLog2sqrt5);
-    Ft = textureLod(light_ssr, vec3(p.xy, 0.0), lod).rgb;
+    vec4 Fat = textureLod(light_ssr, vec3(p.xy, 0.0), lod);
+
+    // if it's a transparent background, shift color components towards white
+    vec3 Ft = mix(vec3(1.0), Fat.rgb, Fat.a);
```
romainguy, line 530: I'm trying to understand whether this is correct. It seems to be the only reasonable solution but I'm not entirely comfortable with it yet.

shaders/src/light_indirect.fs
```diff
@@ -515,35 +515,41 @@ vec3 evaluateRefraction(
...
+    float at = Fat.a;
 #endif
 
     // base color changes the amount of light passing through the boundary
     Ft *= pixel.diffuseColor;
+    at += (1.0 - at) * (1.0 - luminance(pixel.diffuseColor.rgb));
```
romainguy, line 536: Why are you using the luminance here? also, this should only be done when necessary. There's no reason to compute this when the RT is opaque.

shaders/src/light_indirect.fs
```diff
@@ -515,35 +515,41 @@ vec3 evaluateRefraction(
...
 
     // fresnel from the first interface
     Ft *= 1.0 - E;
+    at += (1.0 - at) * luminance(E);
```
romainguy, line 540: Same for those 2 changes.

filament/src/details/Renderer.cpp
```diff
@@ -842,7 +842,7 @@ void FRenderer::renderJob(ArenaScope& arena, FView& view) {
 
     PostProcessManager::ScreenSpaceRefConfig const ssrConfig = PostProcessManager::prepareMipmapSSR(
             fg, svp.width, svp.height,
-            ssReflectionsOptions.enabled ? TextureFormat::RGBA16F : TextureFormat::R11F_G11F_B10F,
+            (ssReflectionsOptions.enabled || view.getBlendMode() == View::BlendMode::TRANSLUCENT) ? TextureFormat::RGBA16F : TextureFormat::R11F_G11F_B10F,
```
romainguy, line 845: @pixelflinger wdyt?

pixelflinger: looks good to me. I'm wondering if we should, in addition, have a setting in view for this feature? So it's still possible to do refraction in a blended view in 11_11_10 -- if you know you don't care about that?

romainguy:
I don't think it makes sense as a material property and prob not even as a View option. It's not going to break anything, and the PBR-ness can't really be addressed anyway since the concept of full transparency has not real-world meaning anyway.

shaders/src/light_indirect.fs
```diff
@@ -515,35 +515,41 @@ vec3 evaluateRefraction(
     // when reading from the cubemap, we are not pre-exposed so we apply iblLuminance
     // which is not the case when we'll read from the screen-space buffer
     vec3 Ft = prefilteredRadiance(ray.direction, perceptualRoughness) * frameUniforms.iblLuminance;
+    float at = 1.0;
```
pixelflinger, line 518: might be possible to make it `const` here.

pixelflinger:
I'm pretty sure this outputs invalid pixels, i.e. pixels that are not "premultiplied alpha", i.e. where RGB is guaranteed to be <= A. And I'm unsure about why we do this in BLEND_MODE_OPAQUE only: "computed alpha value should be stored in the render target directly" that seems wrong.

I'm really not sure, I need more time to think about it.

pixelflinger:
Do you have screenshots with different backgrounds and with and without this change?

veghbalazs:
> Do you have screenshots with different backgrounds and with and without this change?

Updated the description with before-after images.

shaders/src/light_indirect.fs
veghbalazs: 1. Here too is very difficult to reason about correctness due to the lack of real-world interpretation... So the idea is that because base color decreases the amount light passing through the boundary, it would also turn "transparent light" into "more opaque". The question is, by how much? Because opacity is a scalar, going with the luminance of RGB sounded reasonable, but I'm very much open to other ideas.

2. Shall I put them behind `if (frameUniforms.needsAlphaChannel)` checks, or can you identify other flags/defines where these can be skipped?

filament/src/details/Renderer.cpp
veghbalazs: What's your use-case for blended views? Maybe transparency in refractions would be beneficial there too.
(I'm fine with this being available either via a setting or automatically, I think I'll leave this for you to decide).

pixelflinger:
Sorry for taking so long to review. I really like this change but there are few things I'm still not convinced about. Like you said it's hard to reason about. I would like make a few experiments just to prove myself wrong :-)
Would you be able to share your model with us (the one in the screenshot)?

This is what bothers me:  It seems to me that a refractive object should essentially be blended with the background exactly the same way we normally blend (ONE / ONE_MINUS_SRC_ALPHA) -- the problem is that we don't have an alpha value, we have these transmission value instead. And it looks like you're essentially figuring out an alpha by using the luminance and then using some "weird" blending equation.

Anyways, I'd like to try a few other things.

veghbalazs:
No worries, we opened this PR knowing that it might be a niche feature, and we too want to learn from your feedback. 

Sure, I can share the model, attached it here: [Cylinders.shapr.zip](https://github.com/google/filament/files/10226544/Cylinders.shapr.zip). What would be your tool of choice to be experimenting in? 

The screenshots were exported from Shapr3D, and the model file is in our app's own format. You can install it on macOS, iPadOS or Windows [from here](https://www.shapr3d.com/). You'll have to complete a few minutes long intro tutorial and create an account before the app lets you import the model. The screenshots feature is only accessible to paid users though - please share your account email address and we can grant you a gift license. You'll be able to enter "Visualization" mode once the model is opened, and there you'll find several more refractive materials too under the Glass and Plastic categories. The transparent screenshot can be exported using the "Capture" button. Let me know if there's anything I can help you with.

As for other formats, unfortunately our export options don't yet support PBR materials. We have an experimental glTF export in the works, (currently inaccessible in the live app). I exported this model ([Cylinders.glb.zip](https://github.com/google/filament/files/10226546/Cylinders.glb.zip)), however when opened in the glTF viewer sample app, the materials end up very different.

pixelflinger:
@veghbalazs understood. I was hoping we could get a gltf. I think I'll try to write a small test app from our material_sandbox sample maybe.

veghbalazs:
Closing this PR for now. Although the solution in its current form is fine for our use case, we see that it needs to be more fleshed out to be ready for general use and to fit Filament's material system well. (This diff still can serve as starting point, in case later there's renewed interest in the topic of refractive objects on transparent backgrounds).

PR #6383, opened 2022-12-14
Workaround for a bug in Adreno's shader compiler

### LINKED ISSUE #6355: Shader link error on UBERARCHIVE_DEFAULT_DATA (Adreno Only)
**Describe the bug**
This bug only occurs in Adreno. 
Using uberarchive builded with disabling optimization about matc,  
create material with below code.
`gltfio::createUbershaderProvider(engine, UBERARCHIVE_DEFAULT_DATA, UBERARCHIVE_DEFAULT_SIZE)`
then shader link error occur in OpenGLProgram::checkProgramStatus. 

**To Reproduce**
1. Open filament/CMakeList.txt and edit so that code:set(MATC_OPT_FLAGS -gd) always works. like below.
`# Disable optimizations and enable debug info (preserves names in SPIR-V)
set(MATC_OPT_FLAGS -gd)`
 
2. Build android demo app with command: ./build.sh -p desktop,android -i release

3. Run sample-gltf-viewer

4. We can check this bug. object did not rendered, with shader-error.

**Expected behavior**
Object is rendered without shader-error on Adreno, with matc-option(-gd)

**Logs**
`E  KHR_debug ERROR: shader 0 is not a shader object
E  KHR_debug ERROR: program 58 is not a program object or shader 0 is not a shader object`

 - Device: [Pixel 3]
 - OS: [Android 12]
 - Backend: [OpenGL]
 - GPU: [Adreno]
 
image

**Additional context**
If instanced option on base.mat is true, it works well.

pixelflinger:
At least some adreno compilers don't like returning an element of a UBO array that is a structure in the vertex shader. To work this around we have to copy the each of the structure fields.

Fixes #6355

### DIFF
--- a/shaders/src/common_instancing.glsl
+++ b/shaders/src/common_instancing.glsl
@@ -1,0 +1,35 @@
+//------------------------------------------------------------------------------
+// Instancing
+// ------------------------------------------------------------------------------------------------
+
+PerRenderableData object_uniforms;
+
+void initObjectUniforms(out PerRenderableData p) {
+#if defined(MATERIAL_HAS_INSTANCES)
+    // the material manages instancing, all instances share the same uniform block.
+    p = objectUniforms.data[0];
+#else
+    // automatic instancing was used, each instance has its own uniform block.
+
+    // We're copying each field separately to workaround an issue in some Adreno drivers
+    // that fail on non-const array access in a UBO. Accessing the fields works however.
+    // e.g.: this fails `p = objectUniforms.data[instance_index];`
+    p.worldFromModelMatrix = objectUniforms.data[instance_index].worldFromModelMatrix;
+    p.worldFromModelNormalMatrix = objectUniforms.data[instance_index].worldFromModelNormalMatrix;
+    p.morphTargetCount = objectUniforms.data[instance_index].morphTargetCount;
+    p.flagsChannels = objectUniforms.data[instance_index].flagsChannels;
+    p.objectId = objectUniforms.data[instance_index].objectId;
+    p.userData = objectUniforms.data[instance_index].userData;
+#endif
+}
+
+//------------------------------------------------------------------------------
+// Instance access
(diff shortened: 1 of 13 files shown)


shaders/src/getters.fs
```diff
@@ -1,13 +1,11 @@
...
+// Uniforms access
 //------------------------------------------------------------------------------
 
-#if defined(MATERIAL_HAS_INSTANCES)
-/** @public-api */
-int getInstanceIndex() {
-    return instance_index;
+/** sort-of public */
+float getObjectUserData() {
+    return gObjectUniforms.userData;
```
romainguy, line 7: Note: we don't use the g prefix for other globals. But we used _, so like object_uniforms. Doesn't matter much though.

romainguy (approved):

shaders/src/common_instancing.glsl
```diff
@@ -0,0 +1,31 @@
...
+// ------------------------------------------------------------------------------------------------
+
+PerRenderableData gObjectUniforms;
+
+void initObjectUniforms(out PerRenderableData p) {
+#if defined(MATERIAL_HAS_INSTANCES)
+    // the material manages instancing, all instances share the same uniform block.
+    p = objectUniforms.data[0];
+#else
+    // automatic instancing was used, each instance has its own uniform block.
```
bejado, line 12: Maybe add a comment here that you're purposefully initializing each field separately.

bejado (approved):

shaders/src/getters.fs
pixelflinger: done

shaders/src/common_instancing.glsl
pixelflinger: good point. done.

daemyung:
@pixelflinger Really thanks!

PR #6388, opened 2022-12-16
expose joint inverse bind matrices via method on FilamentInstance

### LINKED ISSUE #5520: I want to get inverseBindMatrices to do dynamic bone animation.
I found that #5310 had already retrieved skin name and joints, but I need `inverseBindMatrices` to calculate `boneVector` for dynamic bone such as spring effect. Besides, for multi-instances rendering, I want to get skins at `FilamentInstance`.
screen

Add more queries for skin information to `FilamentAsset` and `FilamentInstance`. 

`SkinVector` is already in both `FilamentAsset` and `FilamentInstance`, so I think just add functions like #5310 does should be enough. Something like 
```
size_t getMatrixCountAt(size_t skinIndex) const noexcept;

const filament::math::mat4f* getInverseBindMatricesAt(size_t skinIndex) const noexcept;
```

nmfisher:
To set glTF bone transforms manually (i.e. other than using an Animator instance), users need access to the inverseBindMatrices for skin joints. This PR exposes these matrices via a method on FilamentInstance.

Fixes #5520.


libs/gltfio/include/gltfio/FilamentInstance.h
```diff
@@ -126,6 +126,13 @@ class UTILS_PUBLIC FilamentInstance {
      */
     void detachSkin(size_t skinIndex, utils::Entity target) noexcept;
 
+    /**
+     * Gets inverse bind matrices for all joints at the given skin index.
+     *
+     * See getJointCountAt for determining the number of matrices returned (i.e. the number of joints).
+     */
+    const math::mat4f* getInverseBindMatricesAt(size_t skinIndex) const noexcept;
```
pixelflinger, line 134:
```suggestion
   math::mat4f const* getInverseBindMatricesAt(size_t skinIndex) const;
```

pixelflinger (changes requested):

pixelflinger (approved):

Issue #6292, opened 2022-11-11
[vulkan - linux] multiple extensions of same name cause segfault


jwinarske:
**Describe the bug**

In this segment with kMaxEnabledLayersCount == 1, and when availableLayers has more than one entry of the same name a segfault happens.

```
    uint32_t layerCount;
    vkEnumerateInstanceLayerProperties(&layerCount, nullptr);
    FixedCapacityVector<VkLayerProperties> availableLayers(layerCount);
    vkEnumerateInstanceLayerProperties(&layerCount, availableLayers.data());
    auto enabledLayers = FixedCapacityVector<const char*>::with_capacity(kMaxEnabledLayersCount);
    for (const auto& desired : DESIRED_LAYERS) {
        for (const VkLayerProperties& layer : availableLayers) {
            const std::string_view availableLayer(layer.layerName);
            if (availableLayer == desired) {
                enabledLayers.push_back(desired.data());
            }
        }
    }
```

**To Reproduce**
Attempt to run on system that reports the same layer more than once.
 
**Expected behavior**
No segfault

**Logs**
```
__GI_raise 0x00007f83b72ac00b
__GI_abort 0x00007f83b728b859
<unknown> 0x00007f83b7481deb
<unknown> 0x00007f83b746a8f7
<unknown> 0x00007f83b7480fb3
std::terminate() 0x00007f83b7480f58
__clang_call_terminate 0x00007f8336baffeb
filament::backend::VulkanDriver::VulkanDriver(filament::backend::VulkanPlatform *, const char *const *, unsigned int, const filament::backend::Platform::DriverConfig &) VulkanDriver.cpp:318
filament::backend::VulkanDriver::create(filament::backend::VulkanPlatform *, const char *const *, unsigned int, const filament::backend::Platform::DriverConfig &) VulkanDriver.cpp:329
filament::backend::VulkanDriverFactory::create(filament::backend::VulkanPlatform *, const char *const *, unsigned int, const filament::backend::Platform::DriverConfig &) VulkanDriver.cpp:89
filament::backend::PlatformVkLinuxWayland::createDriver(void *, const filament::backend::Platform::DriverConfig &) PlatformVkLinuxWayland.cpp:37
filament::FEngine::loop() Engine.cpp:627
std::__invoke<…>(int (filament::FEngine::*&&)(), filament::FEngine *&&) type_traits:3635
std::__thread_execute<…>(std::tuple<…> &, std::__tuple_indices<…>) thread:280
[...]

Issue #6192, opened 2022-10-15
`gltf_viewer` segfault on Linux w/ Intel and Vulkan.


will-ca:
**Describe the bug**
`bin/gltf_viewer` from the official releases page segfaults on launch for me on Linux, and also with a fresh build.

https://github.com/google/filament/releases

**To Reproduce**
1. Download and extract `filament-v1.28.0-linux.tgz` OR build from source according to the instructions.
2. Try to run `bin/gltf_viewer` (`./out/cmake-release/samples/gltf_viewer`).
3. Observe window pop up, followed by segfault before any draw calls.

**Expected behavior**
The application `gltf_viewer` should run and not crash.

**Logs**

Terminal:

```Bash
$ ./bin/gltf_viewer 
FEngine (64 bits) created at 0x36ff8f0 (threading is enabled)
FEngine resolved backend: Vulkan
MESA-INTEL: warning: Performance support disabled, consider sysctl dev.i915.perf_stream_paranoid=0
Vulkan device driver: Intel open-source Mesa driver Mesa 22.1.7
Selected physical device 'Intel(R) UHD Graphics 620 (KBL GT2)' from 1 physical devices. (vendor 0x8086, device 0x5917, driver 0x5801007, api 1.3)
FEngine feature level: 1
The specified IBL path does not exist: /home/User/filament/bin/assets/ibl/lightroom_14b
vkCreateSwapchain: 1024x640, 50, 0, 4, 1

Segmentation fault (core dumped)
```

GDB output and backtrace, if that helps?

```GDB
$ gdb -ex=r --args ./bin/gltf_viewer 
GNU gdb (GDB) 12.1
Copyright (C) 2022 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
Type "show copying" and "show warranty" for details.
This GDB was configured as "x86_64-pc-linux-gnu".
Type "show configuration" for configuration details.
For bug reporting instructions, please see:
<https://www.gnu.org/software/gdb/bugs/>.
Find the GDB manual and other documentation resources online at:
    <http://www.gnu.org/software/gdb/documentation/>.

For help, type "help".
Type "apropos word" to search for commands related to "word"...
Reading symbols from ./bin/gltf_viewer...
(No debugging symbols found in ./bin/gltf_viewer)
[...]

PR #6395, opened 2022-12-19
[vulkan] Pop markers before commands flush

poweifeng:
The flush call to command buffer resets the mCurrent pointer in VulkanCommands such that popGroupMarker will actually trigger the creation of a new buffer.

Before we flush (and reset the client-side state), we should pop group markers if there is one.

This needs a follow-up where we need to make sure that the pop after flush should never happen.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -17,7 +17,7 @@ A new header is inserted each time a *tag* is created.
              normalized position and keeps z reversed [⚠️ **Recompile Materials**]
 - backend: workaround Adreno shader compiler bug (#6355) [⚠️ **Recompile Materials**]
 - geometry: change computing tangent basis from normal vector to use Frisvad's method
-- backend: workaround marker/label pop ordering bug (6292)
+- backend: workaround marker/label pop ordering bug (#6192)
 
 ## v1.30.0
 
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -17,6 +17,7 @@ A new header is inserted each time a *tag* is created.
              normalized position and keeps z reversed [⚠️ **Recompile Materials**]
 - backend: workaround Adreno shader compiler bug (#6355) [⚠️ **Recompile Materials**]
 - geometry: change computing tangent basis from normal vector to use Frisvad's method
+- backend: workaround marker/label pop ordering bug (6292)
 
 ## v1.30.0
 
--- a/filament/backend/src/vulkan/VulkanCommands.cpp
+++ b/filament/backend/src/vulkan/VulkanCommands.cpp
@@ -154,6 +154,17 @@ bool VulkanCommands::flush() {
     const int64_t index = mCurrent - &mStorage[0];
     VkSemaphore renderingFinished = mSubmissionSignals[index];
 
+    // VUID-vkEndCommandBuffer-commandBuffer-01815
+    // TODO: We need to order calls so that there is no outstanding label or markers.
+    if (!mCurrentDebugLabel.empty()) {
+      endDebugUtilsLabel();
+    }
+
+    // VUID-vkEndCommandBuffer-commandBuffer-00062
+    if (!mCurrentDebugMarker.empty()) {
+      endDebugMarker();
+    }
+
     vkEndCommandBuffer(mCurrent->cmdbuffer);
 
     // If the injected semaphore is an "image available" semaphore that has not yet been signaled,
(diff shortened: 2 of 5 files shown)


RELEASE_NOTES.md
```diff
@@ -17,6 +17,7 @@ A new header is inserted each time a *tag* is created.
              normalized position and keeps z reversed [⚠️ **Recompile Materials**]
 - backend: workaround Adreno shader compiler bug (#6355) [⚠️ **Recompile Materials**]
 - geometry: change computing tangent basis from normal vector to use Frisvad's method
+- backend: workaround marker/label pop ordering bug (6292)
```
pixelflinger, line 20:
```suggestion
- backend: workaround marker/label pop ordering bug (#6292)
```

filament/backend/src/vulkan/VulkanCommands.cpp
```diff
@@ -154,6 +154,17 @@ bool VulkanCommands::flush() {
     const int64_t index = mCurrent - &mStorage[0];
     VkSemaphore renderingFinished = mSubmissionSignals[index];
 
+    // VUID-vkEndCommandBuffer-commandBuffer-01815
+    // TODO: We need to order calls so that there is no outstanding label or markers.
+    if (!mCurrentDebugLabel.empty()) {
```
pixelflinger, line 159:
```suggestion
    if (UTILS_UNLIKELY(!mCurrentDebugLabel.empty())) {
```

filament/backend/src/vulkan/VulkanCommands.cpp
```diff
@@ -154,6 +154,17 @@ bool VulkanCommands::flush() {
...
+      endDebugUtilsLabel();
+    }
+
+    // VUID-vkEndCommandBuffer-commandBuffer-00062
+    if (!mCurrentDebugMarker.empty()) {
```
pixelflinger, line 164:
```suggestion
    if (UTILS_UNLIKELY(!mCurrentDebugMarker.empty())) {
```

pixelflinger (approved):
Approved, but I had two small comments.

pixelflinger:
Is this a fix for #6192 or #6292 ??

poweifeng:
#6192 actually.  I'll update the `RELEASE_NOTES.md`

filament/backend/src/vulkan/VulkanCommands.cpp
poweifeng: looks like `UTILS_UNLIKELY` is a compiler hint for branching, but we're actually going to hit this branch on almost every frame. Still add it?

filament/backend/src/vulkan/VulkanCommands.cpp
```diff
@@ -154,6 +154,17 @@ bool VulkanCommands::flush() {
     const int64_t index = mCurrent - &mStorage[0];
     VkSemaphore renderingFinished = mSubmissionSignals[index];
 
+    // VUID-vkEndCommandBuffer-commandBuffer-01815
+    // TODO: We need to order calls so that there is no outstanding label or markers.
+    if (!mCurrentDebugLabel.empty()) {
+      endDebugUtilsLabel();
```
bejado, line 160: 4 spaces indent, please

filament/backend/src/vulkan/VulkanCommands.cpp
```diff
@@ -154,6 +154,17 @@ bool VulkanCommands::flush() {
...
+    }
+
+    // VUID-vkEndCommandBuffer-commandBuffer-00062
+    if (!mCurrentDebugMarker.empty()) {
+      endDebugMarker();
```
bejado, line 165: 4 spaces indent

filament/backend/src/vulkan/VulkanCommands.cpp
```diff
@@ -271,6 +282,93 @@ void VulkanCommands::updateFences() {
...
+void VulkanCommands::endDebugUtilsLabel() {
+    // TODO: the following should be turned into asserts after refactoring the push/pop
+    // label flow.
+    if (!mCurrent) {
+	return;
+    }
+    if (!mCurrent->cmdbuffer) {
+	return;
+    }
+    if (!mCurrentDebugLabel.empty()) {
```
bejado, line 319: Should this be `if (mCurrentDebugLabel.empty())` ? Same question inside `endDebugMarker`.

filament/backend/src/vulkan/VulkanCommands.cpp
poweifeng: done

filament/backend/src/vulkan/VulkanCommands.cpp
poweifeng: done

filament/backend/src/vulkan/VulkanCommands.cpp
pixelflinger: Well, after understanding better that this is actually more a workaround, I think it doesn't matter. Eventually this code should go away (maybe an assert), once we figure out a better API for debug markers. So I'll leave up to you.

pixelflinger (approved):

poweifeng:
Abandoning PR to do a different workaround

PR #6398, opened 2022-12-20
[filamesh] Add option to skip secondary UV coords

poweifeng:
Also added error and exit for when there's a mismatch in the number of vertices vs. the number of UV coords (either primary or secondary).


poweifeng:
I added an option to skip secondary UVs. But ok to remove if that workaround is not necessary.

romainguy (approved):

pixelflinger (approved):

PR #6453, opened 2023-01-12
Ortho shading fixes

palver123:
This PR aims to improve on some artifacts we encountered while using an orthographic camera in a Filament scene. 

### `shading_view`
In shading_parameters.fs, `shading_view` was always calculated as the (normalized) vector from the fragment position to the camera eye position (in world coordinates). But ortho cameras don't have an eye position in a sense as perspective cameras do. With orthographic projection the view direction is uniform for every fragment. Therefore we passed the camera forward direction in `PerViewUniforms`.
Note that sometimes the unnormalized view direction was used too (in other words: `shading_view` with a magnitude). Thus `getWorldViewVectorWithMagnitude()` was born. It replaces code that was dependent on the shaky 'eye position' notion with more correct

### Skyboxes
We noticed that skybox looks weird in ortho view too. In the image below the background should be a vertical gradient but instead it's focused at the center:
image

Based on frame captures the `eye_position` varying was miscalculated in skybox.mat. I didn't quite follow what went wrong in the maths, but saw that `computeWorldPosition()` (called in `initMaterialVertex()`) correctly calculates the world position for DEVICE domain vertex shaders (such as skybox.mat) so I reused that and now skyboxes look correct.

### DIFF
--- a/docs/Materials.html
+++ b/docs/Materials.html
@@ -2708,7 +2708,6 @@ type aliases:
 
 </p><div class="table"><table class="table"><tbody><tr><th style="text-align:left"> Name </th><th style="text-align:center"> Type </th><th style="text-align:left"> Description </th></tr>
 <tr><td style="text-align:left"> <strong class="asterisk">getResolution()</strong> </td><td style="text-align:center"> float4 </td><td style="text-align:left"> Dimensions of the view's effective viewport in pixels: <code>width</code>, <code>height</code>, <code>1 / width</code>, <code>1 / height</code>. This might be different from <code>View::getViewport()</code> for instance because of added rendering guard-bands. This can be used in conjunction with <code>getNormalizedViewportCoord()</code> to generate pixel coordinates. </td></tr>
-<tr><td style="text-align:left"> <strong class="asterisk">getWorldCameraPosition()</strong> </td><td style="text-align:center"> float3 </td><td style="text-align:left"> Position of the camera/eye in world space </td></tr>
 <tr><td style="text-align:left"> <strong class="asterisk">getWorldOffset()</strong> </td><td style="text-align:center"> float3 </td><td style="text-align:left"> The shift required to obtain API-level world space </td></tr>
 <tr><td style="text-align:left"> <strong class="asterisk">getTime()</strong> </td><td style="text-align:center"> float </td><td style="text-align:left"> Current time as a remainder of 1 second. Yields a value between 0 and 1 </td></tr>
 <tr><td style="text-align:left"> <strong class="asterisk">getUserTime()</strong> </td><td style="text-align:center"> float4 </td><td style="text-align:left"> Current time in seconds: <code>time</code>, <code>(double)time - time</code>, <code>0</code>, <code>0</code> </td></tr>
@@ -2716,16 +2715,6 @@ type aliases:
 <tr><td style="text-align:left"> <strong class="asterisk">getExposure()</strong> </td><td style="text-align:center"> float </td><td style="text-align:left"> Photometric exposure of the camera </td></tr>
 <tr><td style="text-align:left"> <strong class="asterisk">getEV100()</strong> </td><td style="text-align:center"> float </td><td style="text-align:left"> <a href="https://en.wikipedia.org/wiki/Exposure_value">Exposure value at ISO 100</a> of the camera </td></tr>
 </tbody></table></div>
-
-<p></p><p>
-
-</p><div class="admonition tip"><div class="admonition-title"> world space</div>
-
-<p></p><p>
-
-    To achieve good precision, the “world space” in Filament's shading system does not necessarily
-    match the API-level world space. To obtain the position of the API-level camera, custom
-    materials can add <code>getWorldOffset()</code> to <code>getWorldCameraPosition()</code>.</p></div>
 
 <p></p>
 <a class="target" name="vertexonly">&nbsp;</a><a class="target" name="materialdefinitions/shaderpublicapis/vertexonly">&nbsp;</a><a class="target" name="toc4.5.5">&nbsp;</a><h3>Vertex only</h3>
@@ -2751,6 +2740,7 @@ The following APIs are only available from the fragment block:
 <tr><td style="text-align:left"> <strong class="asterisk">getWorldTangentFrame()</strong> </td><td style="text-align:center"> float3×3 </td><td style="text-align:left"> Matrix containing in each column the <code>tangent</code> (<code>frame[0]</code>), <code>bi-tangent</code> (<code>frame[1]</code>) and <code>normal</code> (<code>frame[2]</code>) of the vertex in world space. If the material does not compute a tangent space normal for bump mapping or if the shading is not anisotropic, only the <code>normal</code> is valid in this matrix. </td></tr>
 <tr><td style="text-align:left"> <strong class="asterisk">getWorldPosition()</strong> </td><td style="text-align:center"> float3 </td><td style="text-align:left"> Position of the fragment in world space (see note below about world-space) </td></tr>
 <tr><td style="text-align:left"> <strong class="asterisk">getWorldViewVector()</strong> </td><td style="text-align:center"> float3 </td><td style="text-align:left"> Normalized vector in world space from the fragment position to the eye </td></tr>
+<tr><td style="text-align:left"> <strong class="asterisk">getWorldViewVectorWithMagnitude()</strong> </td><td style="text-align:center"> float3 </td><td style="text-align:left"> Vector in world space from the fragment position to the eye </td></tr>
 <tr><td style="text-align:left"> <strong class="asterisk">getWorldNormalVector()</strong> </td><td style="text-align:center"> float3 </td><td style="text-align:left"> Normalized normal in world space, after bump mapping (must be used after <code>prepareMaterial()</code>) </td></tr>
 <tr><td style="text-align:left"> <strong class="asterisk">getWorldGeometricNormalVector()</strong> </td><td style="text-align:center"> float3 </td><td style="text-align:left"> Normalized normal in world space, before bump mapping (can be used before <code>prepareMaterial()</code>) </td></tr>
 <tr><td style="text-align:left"> <strong class="asterisk">getWorldReflectedVector()</strong> </td><td style="text-align:center"> float3 </td><td style="text-align:left"> Reflection of the view vector about the normal (must be used after <code>prepareMaterial()</code>) </td></tr>
@@ -2773,8 +2763,9 @@ The following APIs are only available from the fragment block:
 
 <p></p><p>
 
-    To obtain API-level world space coordinates, custom materials should add <code>getWorldOffset()</code> to
-    <code>getWorldPosition()</code> (et al).</p></div>
+    To achieve good precision, the “world space” in Filament's shading system does not necessarily
+    match the API-level world space. To obtain API-level world space coordinates, custom materials
+    should add <code>getWorldOffset()</code> to <code>getWorldPosition()</code> (et al).</p></div>
 
 <p></p><p>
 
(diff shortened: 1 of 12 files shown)


samples/sample_full_pbr.cpp
```diff
@@ -281,9 +281,7 @@ static void setup(Engine* engine, View* view, Scene* scene) {
             vec2 uvDy = dFdy(uv0);
 
             mat3 tangentFromWorld = transpose(getWorldTangentFrame());
-            vec3 tangentCameraPosition = tangentFromWorld * getWorldCameraPosition();
-            vec3 tangentFragPosition = tangentFromWorld * getWorldPosition();
-            vec3 v = normalize(tangentCameraPosition - tangentFragPosition);
+            vec3 v = tangentFromWorld * getWorldViewVector();
```
romainguy, line 284: Have you tested this change?

palver123: I tried testing it but this sample doesn't work for me with the Metal backend (not even on main branch). I tried using assets/models/monkey/monkey.obj (with a height map generated in GIMP to test the parallax mapping code path).

Firstly, the following line gave me a compiler error in `MaterialBuilder`:
`for (int i = 0; i < numLayers; i++) {`

```metal
ERROR: 0:24: '<' :  wrong operand types: no operation '<' exists that takes a left-hand operand of type ' temp mediump int' and a right operand of type ' temp mediump float' (or there is no acceptable conversion)
ERROR: 0:24: '' : compilation terminated 
```

Fixing the error got me past the Precondition panic but the monkey was rendered flat black and upon rotating the camera the render target content got stuck (wasn't cleared) so I gave up.

But here `tangentFromWorld` is [the usual TBN matrix](https://github.com/google/filament/blob/main/shaders/src/shading_parameters.fs#L31) containing only rotation and mirroring. Thus it is safe to use it to transform direction vectors (no need to transform eye/fragment pos to tangent space). It also keeps the vector's length, so the normalizing at the end is not needed (because `getWorldViewVector()` is already normalized)

filament/src/materials/skybox.mat
```diff
@@ -54,8 +54,6 @@ fragment {
 
 vertex {
     void materialVertex(inout MaterialVertexInputs material) {
-        float3 p = getPosition().xyz;
-        float3 unprojected = mulMat4x4Float3(getViewFromClipMatrix(), p).xyz;
-        material.eyeDirection.xyz = mulMat3x3Float3(getWorldFromViewMatrix(), unprojected);
+        material.eyeDirection.xyz = material.worldPosition.xyz;
```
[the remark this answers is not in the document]
palver123: I tested it in our app (that uses skybox and ortho camera too) and I tested it in gltf viewer sample (modified to use an ortho camera) with the default DamagedHelmet model. Both looked OK

pixelflinger:
I think this PR should be split in two separate PRs:

1. the ortho fixes
2. the skybox fix, which looks like it is completely independent

Also I don't think we need to change the shader public API at all. The user of the API should know if they are using an ortho or perspective projection and they can decide which vector do use, they should have all the info needed. In one case it's the camera-position to fragment, in the other it's the camera-plane to fragment.


The last problem is the "if ortho" in the shader, which I really dislike. I have to think more about it. It seems to me that the fragment could be projected on to the near plane, then the result of that to the fragment is the vector you want, and that should work with any projection.

docs/Materials.html
```diff
@@ -2708,7 +2708,6 @@
 
 </p><div class="table"><table class="table"><tbody><tr><th style="text-align:left"> Name </th><th style="text-align:center"> Type </th><th style="text-align:left"> Description </th></tr>
 <tr><td style="text-align:left"> <strong class="asterisk">getResolution()</strong> </td><td style="text-align:center"> float4 </td><td style="text-align:left"> Dimensions of the view's effective viewport in pixels: <code>width</code>, <code>height</code>, <code>1 / width</code>, <code>1 / height</code>. This might be different from <code>View::getViewport()</code> for instance because of added rendering guard-bands. This can be used in conjunction with <code>getNormalizedViewportCoord()</code> to generate pixel coordinates. </td></tr>
-<tr><td style="text-align:left"> <strong class="asterisk">getWorldCameraPosition()</strong> </td><td style="text-align:center"> float3 </td><td style="text-align:left"> Position of the camera/eye in world space </td></tr>
```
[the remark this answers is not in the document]
palver123: 20cb8d0e7502adcc8780d3611830cd890f76c9aa

docs/Materials.md.html
```diff
@@ -2298,7 +2292,7 @@
 **getWorldTangentFrame()**              | float3x3 |  Matrix containing in each column the `tangent` (`frame[0]`), `bi-tangent` (`frame[1]`) and `normal` (`frame[2]`) of the vertex in world space. If the material does not compute a tangent space normal for bump mapping or if the shading is not anisotropic, only the `normal` is valid in this matrix.
 **getWorldPosition()**                  | float3   |  Position of the fragment in world space (see note below about world-space)
 **getWorldViewVector()**                | float3   |  Normalized vector in world space from the fragment position to the eye
-**getWorldNormalVector()**              | float3   |  Normalized normal in world space, after bump mapping (must be used after `prepareMaterial()`)
+**getWorldViewVectorWithMagnitude()**   | float3   |  Vector in world space from the fragment position to the eye**getWorldNormalVector()**              | float3   |  Normalized normal in world space, after bump mapping (must be used after `prepareMaterial()`)
```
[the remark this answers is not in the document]
palver123: I fixed the formatting error in b1f62e05db8eab3e15ecde631ea17906d2f33de3
I will think about the name (or even the need the justification) of this new method on the public API 🤔

libs/filabridge/include/private/filament/UibStructs.h
```diff
@@ -67,7 +67,10 @@ struct PerViewUib { // NOLINT(cppcoreguidelines-pro-type-member-init)
 
     float lodBias;                  // load bias to apply to user materials
     float refractionLodOffset;
+    float padding0;
     float padding1;
+
+    math::float3 cameraForward;
```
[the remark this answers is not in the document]
palver123: Good point. 320c6e9268068f6f4a873c9ed40fc4154777b9e4

shaders/src/shading_parameters.fs
```diff
@@ -33,7 +33,10 @@ void computeShadingParams() {
 #endif
 
     shading_position = vertex_worldPosition.xyz;
-    shading_view = normalize(frameUniforms.cameraPosition - shading_position);
+
+    // Perspective or orthographic:
+    shading_view = isPerspectiveProjection() ? (frameUniforms.cameraPosition - shading_position) : -frameUniforms.cameraForward;
+    shading_view = normalize(shading_view);
```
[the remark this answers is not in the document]
palver123: Right, now I extract it from the matrix (320c6e9268068f6f4a873c9ed40fc4154777b9e4)

shaders/src/getters.fs
```diff
@@ -54,6 +54,23 @@ vec3 getWorldViewVector() {
     return shading_view;
 }
 
+bool isPerspectiveProjection() {
+    return frameUniforms.clipFromViewMatrix[2].w != 0.0;
+}
+
+/** @public-api */
+vec3 getWorldViewVectorWithMagnitude() {
```
pixelflinger, line 62: What about we call this simply `getWorldViewVector()` and we document that it's not normalized.

pixelflinger (changes requested):
I really like that it's fixing problems with ortho, but I'm not ready to accept this change as is, mainly because of the reliance on "is ortho". I don't have a better solution off the top of my head, I need to think more about it.

Thank you for this PR though, I just want to think about it more.


Meanwhile, I propose a better name for the new function.

palver123:
I extracted the skybox related changes to #6459 because while they are also relevant to how Filament behaves with an orthographic camera they can be easily separated. (I see it already got merged meanwhile)

> Also I don't think we need to change the shader public API at all.

Since both of you mentioned this I reverted the changes to the public API. For us only the calculation of `shading_view` is what's necessary (and with Romain's suggestion even the related Uib changes could be reverted). Fog, sample_full_pbr (height maps) and punctual lights will keep using `getWorldCameraPosition()`. We don't use them so it's not vital for us, we just intended to 'fix all occurrences'. Let me know if I shouldn't have reverted!

> The last problem is the "if ortho" in the shader, which I really dislike

You both mentioned this. My thinking was that branching happens based on an uniform variable so it is not a performanc-wrecking kind of shader code branching. Especially because both code branches are ~ the same ALU complexity, so even if both branches are executed in shader cores, register occupancy should be even for the 2 branches (I could see how uneven occupancy cause a register spill and cost us performance)

> It seems to me that the fragment could be projected on to the near plane...

Is it the possible perf. cost of branching in the shader what you dislike or is it the way we identify ortho projections? If I understand correctly projecting to near plane gets rid of the branching, but needs much more arithmetics than the current code in the PR. I expect the branching to outperform it.
As for the 'identification', it was taken from Camera.cpp where you already inspect the proj. matrix to identify ortho projections.

shaders/src/getters.fs
palver123: I ended up reverting the changes to the public API. With this I reverted our ortho fixes for fog and punctual light too

shaders/src/shading_parameters.fs
```diff
@@ -33,7 +33,10 @@ void computeShadingParams() {
 #endif
 
     shading_position = vertex_worldPosition.xyz;
-    shading_view = normalize(frameUniforms.cameraPosition - shading_position);
+
+    // Perspective or orthographic:
+    shading_view = isPerspectiveProjection() ? (frameUniforms.cameraPosition - shading_position) : -frameUniforms.cameraForward;
```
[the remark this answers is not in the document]
palver123: Please see my question at the end of [my comment below](https://github.com/google/filament/pull/6453#issuecomment-1384328268).

> Is it the possible perf. cost of branching in the shader what you dislike or is it the way we identify ortho projections?

palver123:
@pixelflinger , @romainguy Did you have time to my (not too short) comments? Our main goal is to bring our Filament fork closer to yours (to easy syncing). In the case of this PR I can achieve that by upstreaming much less from our fork and then adjust ours to yours. The minimal changes we need is in `shading_view` calculation and ortho skyboxes (already merged in another PR).

Do you think you can accept this minimal change set and let the other fixes go (or land in a new PR if you'd like)?

samples/sample_full_pbr.cpp
palver123: This part can be reverted too if you'd like. I figured this fixes a bug, without modifying the public API, that's why I kept this

shaders/src/shading_parameters.fs
```diff
@@ -33,7 +33,12 @@ void computeShadingParams() {
 #endif
 
     shading_position = vertex_worldPosition.xyz;
-    shading_view = normalize(frameUniforms.cameraPosition - shading_position);
+
+    // Perspective or orthographic:
+    shading_view = isPerspectiveProjection() ?
```
pixelflinger, line 38: I think I can live with this. Could you add a better comment that explains quickly why we need a different vector for ortho?

pixelflinger (changes requested):
I think I can live with the minimal change left, thanks.
@romainguy wdyt?

shaders/src/shading_parameters.fs
palver123: Sure, done in 24a67b7d4b644e9ebb724807be04268a91a86583

libs/filamat/src/UibGenerator.cpp
```diff
@@ -142,7 +142,7 @@ BufferInterfaceBlock const& UibGenerator::getPerViewUib() noexcept  {
             { "ssrStride",               0, Type::FLOAT                  },
 
             // bring PerViewUib to 2 KiB
-            { "reserved", sizeof(PerViewUib::reserved)/16, Type::FLOAT4 }
+            { "reserved", sizeof(PerViewUib::reserved)/16, Type::FLOAT4  }
```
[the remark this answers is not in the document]
palver123: f2486d0256b5105e93b41cb69e5d7b66a7381199

samples/sample_full_pbr.cpp
palver123: Reverted in cec57142d6c13d658c3f859648f0471291ae46cb.

palver123: https://github.com/google/filament/pull/6490

pixelflinger (approved):

poweifeng (approved):

Issue #6390, opened 2022-12-16
Applying rotation to projection matrix changes spotlight


kpeeters:
I am trying to rotate the final rendered image by pi/2 or -pi/2 by adjusting the projection matrix. That works fine, except that light produced by spotlights changes. This is what I do:

    math::mat4 rot;
    double angle = 3.1415/2;
    rot[0][0]= std::cos(angle);
    rot[0][1]=-std::sin(angle);
    rot[1][0]= std::sin(angle);
    rot[1][1]= std::cos(angle);
    rot[2][2]=1;
    rot[3][3]=1;

    auto projmat = camera->getProjectionMatrix();
    camera->setCustomProjection(rot * projmat, camera_near, camera_far);

Directional lights behave fine (i.e. their effect looks the same with or without the rotation), it's just the spots that change.

Am I doing this wrong?

romainguy:
Doing this probably impacts the froxels and how we find what part of the image are lit by each spotlight. @pixelflinger

kpeeters:
@romainguy Yes, it sounds it's something like that. In any case, I was expecting this kind of post-transformation to not influence anything in the final image, except a 90 degree rotation.

pixelflinger:
I'm not sure if this will affect your problem with the spotlights, but it's not correct to set a custom projection this way.

This is because you're setting the same projection for viewing and culling and that projection has an infinite fare (because you got it from `getProjectionMatrix()`. Instead, you should use the overload of `setCustomProjection` that takes two matrices and get them from `getProjectionMatrix()` and `getCullingProjectionMatrix()` and rotate those.

Try that and let us know if that changes anything  (probably not, I'm guessing).

I think we do have assumptions in a few places about the shape of the projection matrix, I would have to look into this in more details.

kpeeters:
Using the `setCustomProjection` that accepts two matrices does unfortunately not resolve the issue.

kpeeters:
If there is a better way to rotate the image by multiples of pi/2 then that would work too (in other words, a cleaner way to apply a 2d transformation to the final render). The problem I am trying to solve is to adjust for a rotated view on a mobile device. I have a view which has a fixed orientation (I want to keep it that way to avoid re-creating the swapchain on every orientation change) and thus need to rotate filament's output to keep it so that the top of the view points in the physical upwards direction. More in https://github.com/google/filament/discussions/6348.

pixelflinger:
Can you send a few screenshots of the working and not working case with spotlights?

[...]

PR #6455, opened 2023-01-12
fix dynamic lighting with custom projections

pixelflinger:
Forxelization made many assumptions about the shape of the projection
matrix, the gist of these fixes is to remove those assumptions.

Fix https://github.com/google/filament/issues/6390

### DIFF
--- a/filament/src/Froxelizer.cpp
+++ b/filament/src/Froxelizer.cpp
@@ -256,6 +256,66 @@ void Froxelizer::computeFroxelLayout(
 }
 
 UTILS_NOINLINE
+void Froxelizer::updateBoundingSpheres(
+        math::float4* const UTILS_RESTRICT boundingSpheres,
+        size_t froxelCountX, size_t froxelCountY, size_t froxelCountZ,
+        math::float4 const* UTILS_RESTRICT planesX,
+        math::float4 const* UTILS_RESTRICT planesY,
+        float const* UTILS_RESTRICT planesZ) noexcept {
+
+    SYSTRACE_CALL();
+
+    // TODO: this could potentially be parallel_for'ized
+
+    /*
+     * Now compute the bounding sphere of each froxel, which is needed for spotlights
+     * We intersect 3 planes of the frustum to find each 8 corners.
+     */
+
+    UTILS_ASSUME(froxelCountX > 0);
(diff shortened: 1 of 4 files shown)


filament/src/Froxelizer.cpp
```diff
@@ -793,40 +760,46 @@ void Froxelizer::froxelizePointAndSpotLight(
...
-    const size_t x0 = imin.first;
-    const size_t y0 = imin.second;
+    const auto [x0, y0] = clipToIndices(pmin);
     const size_t z0 = findSliceZ(znear);
 
-    const auto imax = clipToIndices(max(xyRightNear, xyRightFar));
-    const size_t x1 = imax.first  + 1;  // x1 points to 1 past the last value (like end() does
-    const size_t y1 = imax.second;      // y1 points to the last value
-    const size_t z1 = findSliceZ(zfar); // z1 points to the last value
+    const auto [x1, y1] = clipToIndices(pmax);
```
romainguy, line 799: x1 was `imax.first+1` and now it's just `imax.first`, is that expected?

filament/src/Froxelizer.cpp
```diff
@@ -838,43 +811,41 @@ void Froxelizer::froxelizePointAndSpotLight(
...
-                    // The reduced sphere from the previous stage intersects this horizontal plane
+                    // The reduced sphere from the previous stage intersects this horizontal plane,
                     // and we now have new smaller sphere centered on these two previous planes
                     size_t bx = std::numeric_limits<size_t>::max(); // horizontal begin index
                     size_t ex = 0; // horizontal end index
 
-                    // find the begin index (left side)
-                    for (size_t ix = x0; ix < x1; ++ix) {
+                    // find the "begin" index (left side)
+                    for (size_t ix = x0; ix < x1 + 1; ++ix) {
```
romainguy, line 843: I assume that's where the + 1 now is right?

romainguy (approved):

filament/src/Froxelizer.cpp
pixelflinger: yes

kpeeters:
The fix in ma/froxelizer does the trick, thanks a lot!

PR #6459, opened 2023-01-13
Fix skybox shader in orthographic view

palver123:
We noticed that skybox looks weird in ortho view too. In the image below the background should be a vertical gradient but instead it's focused at the center:
image

Based on frame captures the `eye_position` varying was miscalculated in skybox.mat. I didn't quite follow what went wrong in the maths, but saw that `computeWorldPosition()` (called in `initMaterialVertex()`) correctly calculates the world position for DEVICE domain vertex shaders (such as skybox.mat) so I reused that and now skyboxes look correct.

Here is a before/after from gltf viewer sample (modified to use an ORTHO camera):
image

### DIFF
--- a/filament/src/materials/skybox.mat
+++ b/filament/src/materials/skybox.mat
@@ -54,8 +54,6 @@ fragment {
 
 vertex {
     void materialVertex(inout MaterialVertexInputs material) {
-        float3 p = getPosition().xyz;
-        float3 unprojected = mulMat4x4Float3(getViewFromClipMatrix(), p).xyz;
-        material.eyeDirection.xyz = mulMat3x3Float3(getWorldFromViewMatrix(), unprojected);
+        material.eyeDirection.xyz = material.worldPosition.xyz;
     }
 }


filament/src/materials/skybox.mat
```diff
@@ -54,8 +54,6 @@ fragment {
 
 vertex {
     void materialVertex(inout MaterialVertexInputs material) {
-        float3 p = getPosition().xyz;
-        float3 unprojected = mulMat4x4Float3(getViewFromClipMatrix(), p).xyz;
-        material.eyeDirection.xyz = mulMat3x3Float3(getWorldFromViewMatrix(), unprojected);
+        material.eyeDirection.xyz = material.worldPosition.xyz;
```
pixelflinger, line 57: @romainguy this change looks good to me.
- there was definitely a bug, imho, in the current implementation which is that we forgot to do the `w` divide!
- and I think there is another bug due to our reversed z

I think this fix correct. wdyt?

pixelflinger (approved):

filament/src/materials/skybox.mat
romainguy: Yeah it looks correct to me as well but I wanted to make sure it had been tested. The perspective divide didn't matter since the skybox is always at infinite (z=1).

PR #6467, opened 2023-01-17
Fix enabledLayers

poweifeng:
instanceCreateInfo should not take stack-allocated data


filament/backend/src/vulkan/VulkanContext.cpp
```diff
@@ -144,17 +144,18 @@ void printDepthFormats(VkPhysicalDevice device) {
     }
 }
 
-bool enableLayers(VkInstanceCreateInfo* instanceCreateInfo) {
-    const std::string_view DESIRED_LAYERS[] = {
-        "VK_LAYER_KHRONOS_validation",
+// These strings need to be allocated outside a function stack
+const std::string_view DESIRED_LAYERS[] = {
```
pixelflinger, line 148: this should be `static`

pixelflinger (changes requested):

filament/backend/src/vulkan/VulkanContext.cpp
poweifeng: It's already in an anonymous namesapce (https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#sf22-use-an-unnamed-anonymous-namespace-for-all-internalnon-exported-entities)

Do you think a static is still necessary?

pixelflinger: oh I missed that. Thanks. No need then.

pixelflinger (approved):

Issue #6358, opened 2022-12-07
SurfaceOrientation doesn't duplicate vertices when generating flat normals


cpheinrich:
**Describe the bug**
I am having two (separate I believe) issues rendering some meshes in Filament. I'm only opening one issue because they are both related to the same mesh type. We generate many meshes of this type, and the issue is not confined to this specific mesh, but all meshes of this type. First, this is what the mesh looks like in scenekit:

![scenekit](https://user-images.githubusercontent.com/12475454/206257949-aba9adbf-0758-4abc-a0b8-3e39db1dec82.jpeg)

***Issue 1***

The Filament directional lights only illuminate parts of the scene. This is clear if I light the scene with only a directional light. This issue appears in our viewer as well as the sample-gltf iOS viewer. As you can see a large chunk of the ground plane is not illuminated by the light (maybe there is a setting to fix this, but I played with all that seemed relevant)

<img width="868" alt="filament_directional_only" src="https://user-images.githubusercontent.com/12475454/206258226-0b5b1a7a-b281-4873-a756-025e3b23eeba.png">

At first I thought it was maybe an issue on our side with the triangulation, but as you can see in meshlab the ground plane only has two triangles and appear to be unrelated to the illumination boundaries

![meshlab](https://user-images.githubusercontent.com/12475454/206258589-7adc4a5f-5ef7-4da7-b518-fd37acdc288b.png)

*** Issue 2 ***

The colors for the mesh are pretty far off from what they are in all other viewers (SceneKit, MeshLab, ThreeJS). In particular they appear to be much less saturated. I have tried adjusting the colors and intensity of the lights but this did not fix it. Here is what it looks like in Filament when I adjusted the directional and indirect lights to be as good as I could make them

<img width="809" alt="filament_directional_plus_indirect" src="https://user-images.githubusercontent.com/12475454/206259135-c4210b4e-c279-4eeb-aada-c495c9f304b5.png">

**To Reproduce**

Load this .glb in filament and observe the issues:

[house.glb.zip](https://github.com/google/filament/files/10178644/house.glb.zip)

These were my exact lighting settings:

```
    math::float3 harmonics[1];
    harmonics[0] = math::float3(1.0f, 1.0f, 1.0f);
    indirect_light_ = IndirectLight::Builder().irradiance(1, harmonics).intensity(24000).build(*engine_);
    if (indirect_light_ == nullptr) {
      throw std::runtime_error("Creating a filament indirect light failed!");
    }
[...]

PR #6476, opened 2023-01-19
geometry: Add TangentSpaceMesh class outline

poweifeng:
- Defined class with documentation in comments
 - Implemented algorithm selection
 - No actual algorithms written yet

Part of google/filament#6358

### DIFF
--- a/libs/geometry/include/geometry/TangentSpaceMesh.h
+++ b/libs/geometry/include/geometry/TangentSpaceMesh.h
@@ -285,7 +285,7 @@ public:
      *
      * @return The number of vertices
      */
-    size_t getTriangleCount() const;
+    size_t getTriangleCount() const noexcept;
 
     /**
      * Get output triangles.
--- a/libs/geometry/include/geometry/TangentSpaceMesh.h
+++ b/libs/geometry/include/geometry/TangentSpaceMesh.h
@@ -1,0 +1,188 @@
+/*
+ * Copyright (C) 2023 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef TNT_GEOMETRY_TANGENTSPACEMESH_H
+#define TNT_GEOMETRY_TANGENTSPACEMESH_H
+
+#include <math/quat.h>
+#include <math/vec3.h>
+#include <math/vec4.h>
+#include <utils/compiler.h>
+
+namespace filament {
+namespace geometry {
+
+struct TangentSpaceMeshInput;
+struct TangentSpaceMeshOutput;
+
+/**
+ * For building Filament-style TANGENTS buffers given an input mesh.
+ * Supercedes the implementation in SurfaceOrientation.h
+ */
+class UTILS_PUBLIC TangentSpaceMesh {
+public:
+    enum class AlgorithmHint : uint8_t {
+        /*
+         * Tries to select the best possible algorithm given the input. The corresponding algorithms
+         * are detailed in the corresponding enums.
+         *   INPUT                                  ALGORITHM
+         *   -----------------------------------------------------------
+         *   normals                                FRISVAD
+         *   positions + indices                    FLAT_SHADING
+         *   normals + tangents                     SIGN_OF_W
+         *   normals + uvs + positions + indices    MIKKTSPACE
+         */
+        DEFAULT = 0,
+
+        /*
+         * --- mikktspace ---
+         * Requires: normals + uvs + positions + indices
+         * Reference: Mikkelsen, M., 2008. Simulation of wrinkled surfaces revisited.
+         *            https://github.com/mmikk/MikkTSpace
+         *            https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#meshes-overview
+         * Note: Will remesh
+         */
+        MIKKTSPACE = 1,
+
+        /*
+         * --- Lengyel's method ---
+         * Requires: normals + uvs + positions + indices
+         * Reference: Lengyel, E., 2019. Foundations of Game Engine Development: Rendering. Terathon
+         *     Software LLC.. (Chapter 7)
+         */
+        LENGYEL = 2,
+
+        /*
+         * --- Hughes-Moller method ---
+         * Requires: normals
+         * Reference: Möller, T. and Hughes, J.F., 1999. Efficiently building a matrix to rotate one
+         *     vector to another. Journal of graphics tools, 4(4), pp.1-4.
+         */
+        HUGHES_MOLLER =3,
+
+        /*
+         * --- Frisvad's method ---
+         * Requires: normals
+         * Reference: Frisvad, J.R., 2012. Building an orthonormal basis from a 3D unit vector
+         *     without normalization. Journal of Graphics Tools, 16(3), pp.151-159.
+         *     http://people.compute.dtu.dk/jerf/code/hairy/
+         */
+        FRISVAD = 4,
+
+        /*
+         * --- Flat Shading ---
+         * Requires: positions + indices
+         * Note: Will remesh
+         */
+        FLAT_SHADING = 5,
+
+        /*
+         * --- Sign of W ---
+         * Requires: normals + tangents
+         * Note: The sign of W determines the orientation of the bitangent.
+         */
+        SIGN_OF_W = 6
+    };
+
+    /*
+     * Computation of the tangent space is intended to be synchronous (worked on the same thread).
+     * Client is expected to keep the input immutable and in a good state for the duration of both
+     * computation *and* query. That is, when querying the result of the tangent spaces, part of the
+     * result might depend on the input data.
+     */
+    class Builder {
+    public:
+        Builder() noexcept;
+        ~Builder() noexcept;
+        Builder(Builder&& that) noexcept;
+        Builder& operator=(Builder&& that) noexcept;
+
+        Builder& vertexCount(size_t vertexCount) noexcept;
+
+        Builder& normals(const filament::math::float3*, size_t stride = 0) noexcept;
+        Builder& tangents(const filament::math::float4*, size_t stride = 0) noexcept;
+        Builder& uvs(const filament::math::float2*, size_t stride = 0) noexcept;
+        Builder& positions(const filament::math::float3*, size_t stride = 0) noexcept;
+
+        Builder& triangleCount(size_t triangleCount) noexcept;
+        Builder& triangles(const filament::math::uint3*) noexcept;
+        Builder& triangles(const filament::math::ushort3*) noexcept;
+
+        Builder& algorithmHint(AlgorithmHint) noexcept;
+
+        TangentSpaceMesh* build();
+
+    private:
+        Builder(const Builder&) = delete;
+        Builder& operator=(const Builder&) = delete;
+
+        TangentSpaceMesh* mMesh;
+    };
+
+    ~TangentSpaceMesh() noexcept;
+    TangentSpaceMesh(TangentSpaceMesh&& that) noexcept;
+    TangentSpaceMesh& operator=(TangentSpaceMesh&& that) noexcept;
+
+    size_t getVertexCount() const noexcept;
+
+    /*
+     * The following methods for retrieving vertex-correlated data assumes that the *out* param
+     * provided by the client is at least of *getVertexCount()* length (accounting for stride).
+     *
+     * If the chosen algorithm did not remesh the input, the client is advised to just use the
+     * data they provided instead of querying.  For example, if the chosen method is FRISVAD, then
+     * the client should not need to call "getPositions". We will simply copy from the input
+     * *positions* in that case.
+     *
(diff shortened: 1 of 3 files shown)


poweifeng:
I'm breaking the SurfaceOrientation successor down into multiple commits. This is the first.  
Please take a look. Thank you!

libs/geometry/include/geometry/TangentSpaceMesh.h
```diff
@@ -0,0 +1,188 @@
...
+
+namespace filament {
+namespace geometry {
+
+struct TangentSpaceMeshInput;
+struct TangentSpaceMeshOutput;
+
+/**
+ * For building Filament-style TANGENTS buffers given an input mesh.
+ * Supercedes the implementation in SurfaceOrientation.h
```
bejado, line 33:
```suggestion
 * Supersedes the implementation in SurfaceOrientation.h
```

libs/geometry/include/geometry/TangentSpaceMesh.h
```diff
@@ -0,0 +1,188 @@
...
+         */
+        LENGYEL = 2,
+
+        /*
+         * --- Hughes-Moller method ---
+         * Requires: normals
+         * Reference: Möller, T. and Hughes, J.F., 1999. Efficiently building a matrix to rotate one
+         *     vector to another. Journal of graphics tools, 4(4), pp.1-4.
+         */
+        HUGHES_MOLLER =3,
```
bejado, line 74:
```suggestion
        HUGHES_MOLLER = 3,
```

libs/geometry/include/geometry/TangentSpaceMesh.h
```diff
@@ -0,0 +1,188 @@
...
+namespace geometry {
+
+struct TangentSpaceMeshInput;
+struct TangentSpaceMeshOutput;
+
+/**
+ * For building Filament-style TANGENTS buffers given an input mesh.
+ * Supercedes the implementation in SurfaceOrientation.h
+ */
+class UTILS_PUBLIC TangentSpaceMesh {
```
bejado, line 35: Since this is a public header file, maybe denote that this is still a WIP.

libs/geometry/include/geometry/TangentSpaceMesh.h
```diff
@@ -0,0 +1,188 @@
...
+     * result might depend on the input data.
+     */
+    class Builder {
+    public:
+        Builder() noexcept;
+        ~Builder() noexcept;
+        Builder(Builder&& that) noexcept;
+        Builder& operator=(Builder&& that) noexcept;
+
+        Builder& vertexCount(size_t vertexCount) noexcept;
```
bejado, line 113: Looks like the `vertexCount` implementation is missing.

libs/geometry/src/TangentSpaceMesh.cpp
```diff
@@ -0,0 +1,305 @@
...
+
+    bool foundHint = false;
+    AlgorithmHint outHint = AlgorithmHint::DEFAULT;
+    switch (input->hint) {
+        case AlgorithmHint::DEFAULT:
+            outHint = selectBestDefaultAlgorithm(inputType);
+            foundHint = true;
+            break;
+        case AlgorithmHint::MIKKTSPACE:
+            if (inputType & NORMALS_UVS_POSITIONS_INDICES) {
```
bejado, line 139: Is this meant to check if `inputType` has normals, uvs, positions, _AND_ indices? Because this will return true if any of them are set. I think you want something like:

```
if ( (inputType & NORMALS_UVS_POSITIONS_INDICES) == NORMALS_UVS_POSITIONS_INDICES) {
```

Same for the other conditions.

bejado (changes requested):

libs/geometry/include/geometry/TangentSpaceMesh.h
pixelflinger: Will `SurfaceOrientation` be re-implemented on top of this? 
That would be cool.

libs/geometry/include/geometry/TangentSpaceMesh.h
pixelflinger: Actually I would not make it public until it's ready.

libs/geometry/include/geometry/TangentSpaceMesh.h
```diff
@@ -0,0 +1,188 @@
...
+
+        Builder& algorithmHint(AlgorithmHint) noexcept;
+
+        TangentSpaceMesh* build();
+
+    private:
+        Builder(const Builder&) = delete;
+        Builder& operator=(const Builder&) = delete;
+
+        TangentSpaceMesh* mMesh;
```
pixelflinger, line 133:
```suggestion
        TangentSpaceMesh* mMesh = nullptr;
```

Just to be safe.

libs/geometry/include/geometry/TangentSpaceMesh.h
```diff
@@ -0,0 +1,188 @@
...
+        Builder& triangles(const filament::math::uint3*) noexcept;
+        Builder& triangles(const filament::math::ushort3*) noexcept;
+
+        Builder& algorithmHint(AlgorithmHint) noexcept;
+
+        TangentSpaceMesh* build();
+
+    private:
+        Builder(const Builder&) = delete;
+        Builder& operator=(const Builder&) = delete;
```
pixelflinger, line 131: By convention the `= delete` ctors should be `public` (we may not be consistant with that everywhere).

That being said, Ideally the `Builder` would be copyable. And it should definitely be movable (no reason not to).
I can see how making it copyable is tricky, so maybe not for this one. But let's make it movable.

libs/geometry/include/geometry/TangentSpaceMesh.h
```diff
@@ -0,0 +1,188 @@
...
+
+        TangentSpaceMesh* mMesh;
+    };
+
+    ~TangentSpaceMesh() noexcept;
+    TangentSpaceMesh(TangentSpaceMesh&& that) noexcept;
+    TangentSpaceMesh& operator=(TangentSpaceMesh&& that) noexcept;
+
+    size_t getVertexCount() const noexcept;
```
pixelflinger, line 140: document this

libs/geometry/include/geometry/TangentSpaceMesh.h
```diff
@@ -0,0 +1,188 @@
...
+     * provided by the client is at least of *getVertexCount()* length (accounting for stride).
+     *
+     * If the chosen algorithm did not remesh the input, the client is advised to just use the
+     * data they provided instead of querying.  For example, if the chosen method is FRISVAD, then
+     * the client should not need to call "getPositions". We will simply copy from the input
+     * *positions* in that case.
+     *
+     * If the client calls getPositions and positions were not provided as input, we will throw
+     * and exception. Similar behavior will apply to UVs.
+     */
```
pixelflinger, line 153: For public method, use the javadoc style (which we process with doxygen) comments. Most important is to start with `/**` and to document parameters with `@param`. Also use `@see` where appropriate.

libs/geometry/include/geometry/TangentSpaceMesh.h
poweifeng: Done

libs/geometry/include/geometry/TangentSpaceMesh.h
poweifeng: Done

libs/geometry/include/geometry/TangentSpaceMesh.h
```diff
@@ -0,0 +1,188 @@
...
+struct TangentSpaceMeshInput;
+struct TangentSpaceMeshOutput;
+
+/**
+ * For building Filament-style TANGENTS buffers given an input mesh.
+ * Supercedes the implementation in SurfaceOrientation.h
+ */
+class UTILS_PUBLIC TangentSpaceMesh {
+public:
+    enum class AlgorithmHint : uint8_t {
```
pixelflinger, line 38: rename this `Algorithm`

libs/geometry/src/TangentSpaceMesh.cpp
poweifeng: Done

libs/geometry/include/geometry/TangentSpaceMesh.h
poweifeng: Done

libs/geometry/include/geometry/TangentSpaceMesh.h
poweifeng: Done

libs/geometry/include/geometry/TangentSpaceMesh.h
```diff
@@ -0,0 +1,188 @@
...
+        Builder& normals(const filament::math::float3*, size_t stride = 0) noexcept;
+        Builder& tangents(const filament::math::float4*, size_t stride = 0) noexcept;
+        Builder& uvs(const filament::math::float2*, size_t stride = 0) noexcept;
+        Builder& positions(const filament::math::float3*, size_t stride = 0) noexcept;
+
+        Builder& triangleCount(size_t triangleCount) noexcept;
+        Builder& triangles(const filament::math::uint3*) noexcept;
+        Builder& triangles(const filament::math::ushort3*) noexcept;
+
+        Builder& algorithmHint(AlgorithmHint) noexcept;
```
pixelflinger, line 125: by convention we always name parameters in headers. This will be needed when adding the doxygen doc.

libs/geometry/include/geometry/TangentSpaceMesh.h
```diff
@@ -0,0 +1,188 @@
...
+    void getPositions(filament::math::float3* out, size_t stride = 0) const;
+    void getUVs(filament::math::float2* out, size_t stride = 0) const;
+    void getQuats(filament::math::quatf* out, size_t stride = 0) const noexcept;
+    void getQuats(filament::math::short4* out, size_t stride = 0) const noexcept;
+    void getQuats(filament::math::quath* out, size_t stride = 0) const noexcept;
+
+    size_t getTriangleCount() const;
+
+    /*
+     * The following methods for retrieving triangles assumes that the *out* param provided by the
```
pixelflinger, line 163: I think we need to add `getAlgorithm` to query which algorithm was actually chosen if `DEFAULT` was used.

bejado (approved):

libs/geometry/src/TangentSpaceMesh.cpp
```diff
@@ -0,0 +1,305 @@
...
+const uint8_t POSITIONS_BIT = 0x08;
+const uint8_t INDICES_BIT = 0x10;
+
+// Input types
+const uint8_t NORMALS = NORMALS_BIT;
+const uint8_t NORMALS_TANGENTS = NORMALS_BIT | TANGENTS_BIT;
+const uint8_t POSITIONS_INDICES = POSITIONS_BIT | INDICES_BIT;
+const uint8_t NORMALS_UVS_POSITIONS_INDICES = NORMALS_BIT | UVS_BIT | POSITIONS_BIT | INDICES_BIT;
+
+const char* HintToString(AlgorithmHint hint) {
```
pixelflinger, line 76: - return a `std::string_view`
- I think you should name this `getAlgorithmAsString` or even just `to_string` (one of the rare case we break our naming convention for mimicing the stl's).

pixelflinger (changes requested):

libs/geometry/include/geometry/TangentSpaceMesh.h
poweifeng: done

libs/geometry/include/geometry/TangentSpaceMesh.h
poweifeng: Already movable. Moved `=delete` copy constructors to `public`

libs/geometry/include/geometry/TangentSpaceMesh.h
poweifeng: done

libs/geometry/include/geometry/TangentSpaceMesh.h
poweifeng: done

libs/geometry/include/geometry/TangentSpaceMesh.h
poweifeng: done

libs/geometry/include/geometry/TangentSpaceMesh.h
poweifeng: I added it, but I wonder if a `isRemeshed` method is more specific. Since it's the only information that the client will need.

libs/geometry/include/geometry/TangentSpaceMesh.h
poweifeng: done

libs/geometry/include/geometry/TangentSpaceMesh.h
```diff
@@ -0,0 +1,328 @@
...
+     * provided as input, we will throw and exception.
+     *
+     * @param out Client's array for the output triangles in unsigned 16-bit indices.
+     */
+    void getTriangles(filament::math::ushort3* out) const;
+
+    /**
+     * @return The algorithm used to compute the output mesh.
+     */
+    Algorithm getAlgorithm() const;
```
pixelflinger, line 313:
```suggestion
    Algorithm getAlgorithm() const noexcept;
```

libs/geometry/include/geometry/TangentSpaceMesh.h
```diff
@@ -0,0 +1,328 @@
...
+
+    /**
+     * Get number of output triangles.
+     * The number of output triangles is the same as the number of input triangles. However, when a
+     * "remesh" is carried out the output triangles are not guarranteed to have any correlation with
+     * the input.
+     *
+     * @return The number of vertices
+     */
+    size_t getTriangleCount() const;
```
pixelflinger, line 288:
```suggestion
    size_t getTriangleCount() const noexcept;
```

pixelflinger (approved):

PR #6480, opened 2023-01-21
Add systrace support on Apple platforms

bejado:
This add supports for `SYSTRACE` macros on Apple platforms. Systrace intervals are recorded as signposts and can be viewed in Instruments.


libs/utils/include/utils/darwin/Systrace.h
```diff
@@ -0,0 +1,259 @@
...
+#define SYSTRACE_DISABLE()
+#define SYSTRACE_CONTEXT()
+#define SYSTRACE_NAME(name)
+#define SYSTRACE_NAME_BEGIN(name)
+#define SYSTRACE_NAME_END()
+#define SYSTRACE_CALL()
+#define SYSTRACE_ASYNC_BEGIN(name, cookie)
+#define SYSTRACE_ASYNC_END(name, cookie)
+#define SYSTRACE_VALUE32(name, val)
+#define SYSTRACE_VALUE64(name, val)
```
poweifeng, line 257: Seems like this is covered in `utils/Systrace.h`?

poweifeng (commented):
There seems to be quite a bit of code duplication between android and apple. Any thoughts on  de-duping?  Maybe abstract interfaces and implementation per platform?

libs/utils/include/utils/darwin/Systrace.h
bejado: Yes, good catch.

bejado:
@poweifeng I considered that, but I think performance is a concern here so I'm not sure if we want to introduce virtual calls. I'm sure Mathias will have some thoughts.

pixelflinger:
> @poweifeng I considered that, but I think performance is a concern here so I'm not sure if we want to introduce virtual calls. I'm sure Mathias will have some thoughts.

Yes, we want these to be usable in release builds.

pixelflinger (approved):

poweifeng (approved):

PR #6483, opened 2023-01-24
Add support for sRGB swapchains

pixelflinger:
This change adds a 'SRGB' config flag when creating a SwapChain that
enables linear to sRGB conversion on write.

When using this flag, the linear->srgb conversion in the color grading
post processing should be disabled (or, the whole post-processing
stage should be disabled).

There is also a new query to determine if this flag is supported by
the underlaying platform.

On Metal, this happens automatically when the underlaying layer is sRGB.


bejado (approved):
This is probably worth a release note, yeah?

poweifeng (approved):

pixelflinger:
> This is probably worth a release note, yeah?

Yes! thanks. Done and submitted.

PR #6492, opened 2023-01-25
Initialize object_uniforms in depth shader

### LINKED ISSUE #6487: Picking broken in 1.30.0
**Describe the bug**
asset->getName always returns null in 1.30.0

**To Reproduce**
1. Build 1.30.0
2. Run gltf_viewer 
  **Expected behavior** **Expected behavior**
asset->getName should return the name of the node 
**Screenshots**
image
Screen Shot 2023-01-24 at 12 51 13 PM
Screen Shot 2023-01-24 at 12 50 15 PM

**Logs**

 - OS: [e.g. iOS] MacOS
 - GPU: [e.g. NVIDIA GTX 1080] 
 - Backend: [OpenGL/Vulkan] Metal 

**Additional context**
You can see in the images that the gltf_viewer correctly displayed the name of the asset entity in the UI before 1.30.0 and no longer does in 1.30.0.  The listed commit is the one where we can verify it stopped working.

bejado:
Fixes #6487

### DIFF
--- a/shaders/src/depth_main.fs
+++ b/shaders/src/depth_main.fs
@@ -17,6 +17,8 @@ highp vec2 computeDepthMomentsVSM(const highp float depth);
 void main() {
     filament_lodBias = frameUniforms.lodBias;
 
+    initObjectUniforms(object_uniforms);
+
 #if defined(BLEND_MODE_MASKED) || ((defined(BLEND_MODE_TRANSPARENT) || defined(BLEND_MODE_FADE)) && defined(MATERIAL_HAS_TRANSPARENT_SHADOW))
     MaterialInputs inputs;
     initMaterial(inputs);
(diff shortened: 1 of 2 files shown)


poweifeng (approved):

shaders/src/depth_main.fs
```diff
@@ -47,6 +47,7 @@ void main() {
     fragColor.xy = computeDepthMomentsVSM(depth);
     fragColor.zw = computeDepthMomentsVSM(-1.0 / depth); // requires at least RGBA16F
 #elif defined(VARIANT_HAS_PICKING)
+    initObjectUniforms(object_uniforms); // only needed for picking
```
pixelflinger, line 50: I would be more confortable moving this just under the `filament_lodBias - ` line.

I think it's okay to protect it with `#ifdef VARIANT_HAS_PICKING` but with a comment stating that currently it's only used in that case. But I think it'll be safer to not even do that and let the compiler eliminate the unneeded code.

pixelflinger (changes requested):

PR #6498, opened 2023-01-27
Improve Session API, send formatted string to server

feresr:
Uploading [Pass] and [Resource] information to the server. 

Much of the code in this PR is prototype (No error handling, no optimisations) just to get something working quickly and iterate later. Don't usually code in cpp, feedback appreciated. 

Screenshot 2023-01-27 at 4 59 42 pm

### DIFF
--- a/libs/fgdbg/client/build.gradle.kts
+++ b/libs/fgdbg/client/build.gradle.kts
@@ -3,8 +3,9 @@ import org.jetbrains.compose.desktop.application.dsl.TargetFormat
 import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
 
 plugins {
-    kotlin("jvm") version "1.6.10"
-    id("org.jetbrains.compose") version "1.1.1"
+    kotlin("jvm") version "1.8.0"
+    id("org.jetbrains.compose") version "1.3.0"
+    kotlin("plugin.serialization") version "1.4.20"
 }
 
 group = "com.google.filament"
@@ -27,6 +28,7 @@ dependencies {
         implementation("io.ktor:ktor-client-core:$ktorVersion")
         implementation("io.ktor:ktor-client-cio:$ktorVersion")
         implementation("io.ktor:ktor-client-websockets:$ktorVersion")
+        implementation("com.charleskorn.kaml:kaml:0.51.0")
     }
 }
 
--- a/libs/fgdbg/client/src/main/kotlin/Main.kt
+++ b/libs/fgdbg/client/src/main/kotlin/Main.kt
@@ -16,15 +16,22 @@
 
 import androidx.compose.material.MaterialTheme
 import androidx.compose.desktop.ui.tooling.preview.Preview
+import androidx.compose.foundation.background
+import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.Row
+import androidx.compose.foundation.layout.Spacer
 import androidx.compose.foundation.layout.height
+import androidx.compose.foundation.layout.width
 import androidx.compose.material.Button
+import androidx.compose.material.ButtonColors
+import androidx.compose.material.ButtonDefaults
 import androidx.compose.material.Text
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.DisposableEffect
 import androidx.compose.runtime.LaunchedEffect
 import androidx.compose.runtime.MutableState
 import androidx.compose.runtime.getValue
+import androidx.compose.runtime.mutableStateListOf
 import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.remember
 import androidx.compose.runtime.setValue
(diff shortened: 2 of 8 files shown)


libs/fgdbg/include/fgdbg/Session.h
```diff
@@ -17,15 +29,29 @@ class Session {
 
 public:
     Session(const std::string& name,
-            std::shared_ptr<DebugServer> server) : name{ name }, server{ server } {}
+            std::shared_ptr<DebugServer> server) : name{ name }, server{ std::move(server) } {}
```
bejado, line 32: This `std::move` here is confusing and probably unnecessary. The `shared_ptr` is passed by value into the `Session` constructor, so it has already been copied. There's no need to `std::move` it.

libs/fgdbg/src/Session.cpp
```diff
@@ -15,15 +15,72 @@
...
-void Session::addResources(const std::vector<Resource>& resources) {}
+void Session::addResource(const Resource& resource) {
+    this->resources.push_back(resource);
+}
 
 void Session::update() const {
-    server->sendMessage(0);
+    // TODO(@raviola): Temporary format
+    //  <---------------passes------------------->|<----------resources----------------->
+    //  id:name:readId-readId:writeId-writeId, ...|id:name:size,id:name:size,id:name:size
```
bejado, line 43: Have you considered using JSON or some other interchange format?

pixelflinger: oh... that would be cool.

pixelflinger (commented):
a lot of this is a bit over my head :)
what I understand lgtm.

libs/fgdbg/include/fgdbg/Session.h
feresr: I'm not great with move-semantics, CLION likes `std::move`:
> Clang-Tidy: Parameter 'server' is passed by value and only copied once; consider moving it to avoid unnecessary copies

libs/fgdbg/src/Session.cpp
feresr: Defenitely!  Updated this logic to use YAML, let me know what you think

libs/fgdbg/client/build.gradle.kts
```diff
@@ -27,6 +28,7 @@ dependencies {
         implementation("io.ktor:ktor-client-core:$ktorVersion")
         implementation("io.ktor:ktor-client-cio:$ktorVersion")
         implementation("io.ktor:ktor-client-websockets:$ktorVersion")
+        implementation("com.charleskorn.kaml:kaml:0.51.0")
```
feresr, line 31: Adding a third party dep to parse YAML, is this OK?

libs/fgdbg/include/fgdbg/Session.h
bejado: I think the `std::move` does give a very small performance win here, but in general I tend to avoid moving things unless I have a strong reason as it can lead to subtle bugs. For example, if you tried to access the `server` parameter after the move, it would be null. I'm fine either way though.

feresr: Thanks for the clarification. I agree moving here is unsafe and copying a pointer is cheap. Done

libs/fgdbg/client/build.gradle.kts
pixelflinger: @romainguy what do you think?

romainguy: It's Apache 2, fine with me

libs/fgdbg/client/src/main/kotlin/Main.kt
```diff
@@ -28,77 +32,142 @@ import androidx.compose.runtime.getValue
...
-    val reconnectDelay: Int = 5,
+  val host: String = "localhost",
+  val port: Int = 8082,
+  val path: String = "/",
+  val state: MutableState<ConnectionState> = mutableStateOf(ConnectionState.UNKNOWN),
+  val reconnectDelay: Int = 5,
 )
 
+@kotlinx.serialization.Serializable
+data class Payload(
```
romainguy, line 60: Reminder that our code style uses 4 spaces, not 2

libs/fgdbg/client/src/main/kotlin/Main.kt
```diff
@@ -28,77 +32,142 @@ import androidx.compose.runtime.getValue
...
-    val state: MutableState<ConnectionState> = mutableStateOf(ConnectionState.UNKNOWN),
-    val reconnectDelay: Int = 5,
+  val host: String = "localhost",
+  val port: Int = 8082,
+  val path: String = "/",
+  val state: MutableState<ConnectionState> = mutableStateOf(ConnectionState.UNKNOWN),
+  val reconnectDelay: Int = 5,
 )
 
+@kotlinx.serialization.Serializable
```
romainguy, line 59: Why use a fully qualified name here?

romainguy (approved):
LGTM so far minus the code style issue, please address this before merging

Issue #4780, opened 2021-10-29
Build x86 filament on windows 10 use vs2019,generated a lot of error in bluegl.prox


MaybeWong:
Hello,I have to  use filament on windows x86.
But I bulid it  generated error like that.
E:\WY\filament\filament-main\libs\bluegl\src\BlueGLCoreWindowsImpl.S(22): error A2013: .MODEL must precede this directive
1>E:\WY\filament\filament-main\libs\bluegl\src\BlueGLCoreWindowsImpl.S(25): error A2034: must be in segment block : bluegl_glProgramUniform1dvEXT
1>E:\WY\filament\filament-main\libs\bluegl\src\BlueGLCoreWindowsImpl.S(26): error A2034: must be in segment block
1>E:\WY\filament\filament-main\libs\bluegl\src\BlueGLCoreWindowsImpl.S(27): error A2034: must be in segment block
1>E:\WY\filament\filament-main\libs\bluegl\src\BlueGLCoreWindowsImpl.S(28): fatal error A1010: unmatched block nesting : bluegl_glProgramUniform1dvEXT
Is there any suggestions？I look forward to receiving your reply，thank u！

MaybeWong:
@romainguy Any version is not supported anymore, including the previous version？

romainguy:
No we've never supported x86 on Windows, only on Android (and even there we plan on getting rid of it as soon as possible).

PR #6499, opened 2023-01-27
Support Windows32 on bluegl as C backend

### LINKED ISSUE #2773: Plan to support x86 on windows？
I want to use filament on windows x86.
But it only support x64.
Do you plan to support x86 on windows？

daemyung:
Now filament can run on Windows32 even this platform isn't supported officially.

Fix #2773, #4780


romainguy:
I'm not against adding a C backend as an *option* but not as a *replacement* of the ASM backend on platforms that we already support.

daemyung:
@romainguy Ok I've restored ASM backend. Could you give a reason keeping that? Is it performance related issue? I think that bluegl C backend can support all platforms with less code than before.

libs/bluegl/CMakeLists.txt
```diff
@@ -4,7 +4,7 @@ project(bluegl ASM)
 set(TARGET              bluegl)
 set(PUBLIC_HDR_DIR      include)
 
-if (WIN32)
+if (WIN32 AND CMAKE_SIZEOF_VOID_P EQUAL 8)
```
romainguy, line 7: Instead of the cryptic `CMAKE_SIZEOF_VOID_P EQUAL 8` is would be better to create an intermediate variable called IS_64_BIT.

libs/bluegl/CMakeLists.txt
```diff
@@ -19,7 +19,11 @@ set(SRCS
 
 if (WIN32)
     set(SRCS ${SRCS} src/BlueGLWindows.cpp)
-    set(SRCS ${SRCS} src/BlueGLCoreWindowsImpl.S)
+    if(${CMAKE_SIZEOF_VOID_P} STREQUAL 4)
```
romainguy, line 22: Same here, also why is the test string based whereas above it was integer based?

pixelflinger:
> @romainguy Ok I've restored ASM backend. Could you give a reason keeping that? Is it performance related issue? I think it's a good advantage that bluegl C backend can support all platforms. Less code and same functionality.

Yes, this avoids a lot of saving/restoring registers on the stack, as well as a "jump" back to the thunk, just to jump again to the caller.

libs/bluegl/CMakeLists.txt
daemyung: Thanks. I will update.

libs/bluegl/CMakeLists.txt
daemyung: Oh... it will be my mistake.

daemyung:
> > @romainguy Ok I've restored ASM backend. Could you give a reason keeping that? Is it performance related issue? I think it's a good advantage that bluegl C backend can support all platforms. Less code and same functionality.
> 
> Yes, this avoids a lot of saving/restoring registers on the stack, as well as a "jump" back to the thunk, just to jump again to the caller.

Thanks for your explanation. It's make sense. :)

romainguy (approved):
Thanks!

Issue #6481, opened 2023-01-21
getWorldTransformAccurate - Crash in Android


atul161:
Android Crash 
---------------------

Steps to reproduce:
 - https://github.com/atul161/multigltfpoc  
 -  In this demo project there is two screen. The first screen have one button . After
 clicking on button it forwarded to the another activity. In this acivity I have 2 glb. Once is box like cube glb
and another is avatar glb. I somehow managed to put my avatar glb inside that cube.

- When we go back to previous activity and again click on button continously.  Repeating the same steps multiple tiime can lead to crash.

`A  Fatal signal 11 (SIGSEGV), code 2 (SEGV_ACCERR), fault addr 0x7d0d228bc0 in tid 29960 (yo.multigltfpoc), pid 29960 (yo.multigltfpoc)
2023-01-21 16:20:06.448 30191-30191 DEBUG                   pid-30191                            A  Cmdline: com.yoyo.multigltfpoc
2023-01-21 16:20:06.448 30191-30191 DEBUG                   pid-30191                            A  pid: 29960, tid: 29960, name: yo.multigltfpoc  >>> com.yoyo.multigltfpoc <<<
2023-01-21 16:20:06.449 30191-30191 DEBUG                   pid-30191                            A        #00 pc 00000000000d083c  /data/app/~~kC-G1cM-PCl_B5uDSydXPQ==/com.yoyo.multigltfpoc-OToSl0qcBTNMdZqBIOswiw==/lib/arm64/libfilament-jni.so (filament::TransformManager::getWorldTransformAccurate(utils::EntityInstance<filament::TransformManager, false>) const+28) (BuildId: 8568a16823add60c2469a2c8bde87089e5ceca3e)`

- There  is one more bug here. If you see the the model viewer , i have commented out look at code. When you uncomment this and if you repeat the same steps like above then automatically the placement of objects changed. Note: You need to click multiple times.

-   https://drive.google.com/file/d/1zNxYfB5ldD_77D3i4qsiqdA-zg-oKk6w/view?usp=sharing   [Video Link]

**Redmi Note 10 *
 - OS: Android 12]
 - 4gb ram

Please have a look on this.  @prideout  @romainguy .
If possible can you explain what should be reason for this.

poweifeng:
I tried out your code.  The issue is passing an `Entity` into `getWorldTransform` when it expects an `Instance`.  Here's a diff that addresses your problem.

```
64a65,70
>     var loadedModel: Boolean = false
>         private set
> 
[...]

PR #6500, opened 2023-01-27
 Destroy FilamentInstance's mRoot in EntityManager

poweifeng:
I'd appreciate a second set (or more) eyes on this. I found that some entities remained in `EntityManager` even after all the assets have been released while looking into #6481. 

Found that the [leaked entity is created][AssetLoaderCreateInstance] and set to the instance's root, but is not destroyed from `EntityManager` in the instance's destructor nor in the destructor of `FilamentAsset`. 

[AssetLoaderCreateInstance]: https://github.com/google/filament/blob/v1.31.2/libs/gltfio/src/AssetLoader.cpp#L511

### DIFF
--- a/libs/gltfio/src/FilamentAsset.cpp
+++ b/libs/gltfio/src/FilamentAsset.cpp
@@ -38,8 +38,10 @@ FFilamentAsset::~FFilamentAsset() {
     releaseSourceData();
 
     // Destroy all instance objects. Instance entities / components are
-    // destroyed later in this method because they are owned by the asset.
+    // destroyed later in this method because they are owned by the asset
+    // (except for the root of the instance).
     for (FFilamentInstance* instance : mInstances) {
+        mEntityManager->destroy(instance->mRoot);
         delete instance;
     }
 


libs/gltfio/src/FilamentAsset.cpp
```diff
@@ -38,8 +38,10 @@ FFilamentAsset::~FFilamentAsset() {
     releaseSourceData();
 
     // Destroy all instance objects. Instance entities / components are
-    // destroyed later in this method because they are owned by the asset.
+    // destroyed later in this method because they are owned by the asset
+    // (except for the root of the instance).
     for (FFilamentInstance* instance : mInstances) {
+        mEntityManager->destroy(instance->mRoot);
```
romainguy, line 44: Can't multiple instances have the same root?

poweifeng: Each instance is created via a call to `createInstance`, so in that case, it seems no.

https://github.com/google/filament/blob/v1.31.2/libs/gltfio/src/AssetLoader.cpp#L488

romainguy: 👍

romainguy (approved):

PR #6502, opened 2023-01-30
Fix a build error

daemyung:


libs/filamat/include/filamat/MaterialBuilder.h
```diff
@@ -38,6 +38,7 @@
 #include <string>
 #include <utility>
 #include <vector>
+#include <memory>
```
romainguy, line 41: Please keep the includes in alphabetical order, thanks!

romainguy (changes requested):

pixelflinger (approved):

libs/filamat/include/filamat/MaterialBuilder.h
daemyung: Oh... I don't know about this. I made a #6507 for this.

PR #6503, opened 2023-01-30
Include compatible headers for C++

daemyung:


libs/filamat/include/filamat/MaterialBuilder.h
```diff
@@ -39,8 +39,8 @@
 #include <utility>
 #include <vector>
 
-#include <stddef.h>
-#include <stdint.h>
+#include <cstddef>
+#include <cstdint>
```
pixelflinger, line 43: We should not use STL include in public headers. Is this fixing an issue?

pixelflinger (changes requested):

libs/filamat/include/filamat/MaterialBuilder.h
pixelflinger: I realize this header has tons of STL headers, but I think that's not ideal. Anyways, let's not add more if it's not needed.

daemyung: > We should not use STL include in public headers. Is this fixing an issue?

No it's not fixing any issues. For C++ project, including compatible headers is the canonical way. Including C headers isn't recommended. So I replaced including headers. If including C headers is the coding style of filament, please let me know. I will close this PR.

pixelflinger: yes it's the coding style we are using so far. If we ever change it, we would update all headers at once.

daemyung: Understood I will close this PR. Thanks.

PR #6509, opened 2023-01-31
Add support for sRGB swapchains

pixelflinger:
This change adds a 'SRGB' config flag when creating a SwapChain that enables linear to sRGB conversion on write.

When using this flag, the linear->srgb conversion in the color grading post processing should be disabled (or, the whole post-processing stage should be disabled).

There is also a new query to determine if this flag is supported by the underlaying platform.

On Metal, this happens automatically when the underlaying layer is sRGB.

This reverts commit 3799e219fcbd391be8e08fbd808a7cdd4d3532f2.


android/filament-android/src/main/java/com/google/android/filament/SwapChain.java
```diff
@@ -91,11 +91,34 @@ public class SwapChain {
...
         mNativeObject = nativeSwapChain;
         mSurface = surface;
     }
 
+    /**
+     * Return whether createSwapChain supports the SWAP_CHAIN_CONFIG_SRGB_COLORSPACE flag.
+     * The default implementation returns false.
+     *
+     * @param engine A reference to the filament Engine
+     * @return true is SWAP_CHAIN_CONFIG_SRGB_COLORSPACE is supported, false otherwise.
```
romainguy, line 116: Is > if

filament/backend/include/backend/platforms/OpenGLPlatform.h
```diff
@@ -68,6 +68,14 @@ class OpenGLPlatform : public Platform {
      */
     virtual SwapChain* createSwapChain(void* nativeWindow, uint64_t flags) noexcept = 0;
 
+    /**
+     * Return whether createSwapChain supports the SWAP_CHAIN_CONFIG_SRGB_COLORSPACE flag.
+     * The default implementation returns false.
+     *
+     * @return true is SWAP_CHAIN_CONFIG_SRGB_COLORSPACE is supported, false otherwise.
```
romainguy, line 75: Is > if

romainguy (approved):

android/filament-android/src/main/java/com/google/android/filament/SwapChain.java
pixelflinger: good catch thanks

PR #6515, opened 2023-01-31
improve sample_full_pbr

pixelflinger:
- -m option now works with "directory_*.png" or just "*.png"
- "color" replaced by "albedo" to mach other places in the source tree
- fixed warnings
- clear the background when IBL is not used


samples/sample_full_pbr.cpp
```diff
@@ -76,13 +77,13 @@ struct PbrMap {
...
 
 static std::array<PbrMap, MAP_COUNT> g_maps = {
-    PbrMap { "_color",      "baseColorMap",  true,  nullptr },
-    PbrMap { "_ao",         "aoMap",         false, nullptr },
-    PbrMap { "_roughness",  "roughnessMap",  false, nullptr },
-    PbrMap { "_metallic",   "metallicMap",   false, nullptr },
-    PbrMap { "_normal",     "normalMap",     false, nullptr },
-    PbrMap { "_bentNormal", "bentNormalMap", false, nullptr },
-    PbrMap { "_height",     "heightMap",     false, nullptr },
+    PbrMap { "albedo",     "baseColorMap",  true,  nullptr },
```
romainguy, line 80: baseColor would be fine but I strongly dislike albedo as a term :/

romainguy: Also _color is what's used in the sample sets in third_party/textures/

pixelflinger: it looks like we're using different terms in different places :-(

romainguy: Where do we use albedo? I've always avoided that term. It's never mentioned in the docs, etc.

pixelflinger: okay, I've switched back to "color" and renamed monkey/albedo.png to monkey/color.png

samples/CMakeLists.txt
```diff
@@ -151,7 +151,7 @@ endfunction()
 add_mesh("assets/models/monkey/monkey.obj" "suzanne.filamesh")
 
 set (COMPRESSION "--compression=uastc")
-add_ktxfiles("assets/models/monkey/albedo.png" "albedo.ktx2" "${COMPRESSION}")
```
romainguy, line 154: prideout!

romainguy: This change means we need to change loading code somewhere?

romainguy (approved):

samples/CMakeLists.txt
pixelflinger: well that's why I kept the `albedo.ktx2`

pixelflinger: lol ah yes it is.

PR #6518, opened 2023-02-01
Data verification before server upload

feresr:
[Perform a hash check + Verify that the data is valid] before upload

### DIFF
--- a/libs/fgdbg/include/fgdbg/Session.h
+++ b/libs/fgdbg/include/fgdbg/Session.h
@@ -23,13 +23,14 @@
 #include <memory>
 #include <string>
 #include <utility>
+#include <unordered_map>
 
 namespace filament::fgdbg {
 class Session {
 
 public:
-    Session(const std::string& name,
-            std::shared_ptr<DebugServer> server) : name{ name }, server{ server } {}
+    Session(std::string name,
+            std::shared_ptr<DebugServer> server) : mName{ std::move(name) }, mServer{ server } {}
 
     void addPasses(const std::vector<Pass>& passes);
     void addPass(const Pass& pass);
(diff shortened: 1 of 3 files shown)


libs/fgdbg/src/Session.cpp
```diff
@@ -16,25 +16,39 @@
...
 namespace filament::fgdbg {
 
 void Session::addPasses(const std::vector<Pass>& passes) {
-    this->passes.insert(this->passes.end(), passes.begin(), passes.end());
+    for (auto& pass: passes) { addPass(pass); }
 }
 
 void Session::addPass(const Pass& pass) {
-    this->passes.push_back(pass);
+    this->passes.insert(std::pair(pass.id, pass));
```
pixelflinger, line 29: instead of the `std::pair(a, b)` syntax in a lot of places you could just use `{a, b}` which is less verbose.

libs/fgdbg/src/Session.cpp
```diff
@@ -16,25 +16,39 @@
...
 #include "fgdbg/Session.h"
 #include <sstream>
+#include <set>
+#include <iostream>
 
 namespace filament::fgdbg {
 
 void Session::addPasses(const std::vector<Pass>& passes) {
-    this->passes.insert(this->passes.end(), passes.begin(), passes.end());
+    for (auto& pass: passes) { addPass(pass); }
```
pixelflinger, line 25: put the content of the `for` on its own line, which is helps a lot when debugging (e.g. to set a breakpoint).

libs/fgdbg/include/fgdbg/Session.h
```diff
@@ -42,16 +43,24 @@ class Session {
...
 
 private:
     const std::string& name;
     std::shared_ptr<DebugServer> server;
 
-    // TODO(@raviola) These might need to be a set/map in order to guarantee there are no two [Pass]
-    //  or [Resource] with the same id. This will also allow for updates by key.
-    std::vector<Pass> passes{};
-    std::vector<Resource> resources{};
+    std::unordered_map<size_t, Pass> passes{};
```
pixelflinger, line 52: our code style uses the `m` prefix for private members. e.g. here `mPasses` (and elsewhere)

libs/fgdbg/include/fgdbg/Session.h
```diff
@@ -42,16 +43,24 @@ class Session {
     /**
      * Send session information to server
      */
-    void update() const;
+    void update();
 
 private:
     const std::string& name;
```
pixelflinger, line 49: that looks dangerous. A reference is just like a pointer in C++ there is no guarantee about the lifetime of the object. I would just use a `std::string` without `&`

pixelflinger (changes requested):

libs/fgdbg/include/fgdbg/Session.h
feresr: good catch, thanks!

bejado (approved):

PR #6530, opened 2023-02-03
geometry: Add TangentSpaceMesh implementations

poweifeng:
Added:
 - Hughes-Moller
 - Frisvad's method
 - Flat shading

### DIFF
--- a/libs/geometry/include/geometry/TangentSpaceMesh.h
+++ b/libs/geometry/include/geometry/TangentSpaceMesh.h
@@ -56,7 +56,6 @@ public:
          *   -----------------------------------------------------------
          *   normals                                FRISVAD
          *   positions + indices                    FLAT_SHADING
-         *   normals + tangents                     SIGN_OF_W
          *   normals + uvs + positions + indices    MIKKTSPACE
          * </pre>
          */
@@ -88,8 +87,14 @@ public:
          * Hughes-Moller method
          *
          * **Requires**: `normals` <br/>
-         * **Reference**: Möller, T. and Hughes, J.F., 1999. Efficiently building a matrix to rotate one
-         *     vector to another. Journal of graphics tools, 4(4), pp.1-4.
+         * **Reference**:
+         *     - Hughes, J.F. and Moller, T., 1999. Building an orthonormal basis from a unit
+         *       vector. journal of graphics tools, 4(4), pp.33-35.
+         *     - Parker, S.G., Bigler, J., Dietrich, A., Friedrich, H., Hoberock, J., Luebke, D.,
+         *       McAllister, D., McGuire, M., Morley, K., Robison, A. and Stich, M., 2010.
+         *       Optix: a general purpose ray tracing engine. Acm transactions on graphics (tog),
+         *       29(4), pp.1-13.
+         * **Note**: We implement the Optix variant, which is documented in the second reference above.
          */
         HUGHES_MOLLER = 3,
 
@@ -110,15 +115,7 @@ public:
          * **Requires**: `positions + indices` <br/>
          * **Note**: Will remesh
          */
-        FLAT_SHADING = 5,
-
-        /**
-         * Sign of W
-         *
-         * **Requires**: `normals + tangents` <br/>
-         * **Note**: The sign of W determines the orientation of the bitangent.
-         */
-        SIGN_OF_W = 6
+        FLAT_SHADING = 5
     };
 
     /**
--- a/libs/geometry/include/geometry/TangentSpaceMesh.h
+++ b/libs/geometry/include/geometry/TangentSpaceMesh.h
@@ -186,7 +186,12 @@ public:
         Builder& algorithm(Algorithm algorithm) noexcept;
 
         /**
-         * Computes the tangent space mesh.
+         * Computes the tangent space mesh. The resulting mesh object is owned by the callee. The
+         * callee must call TangentSpaceMesh::destroy on the object once they are finished with it.
+         *
+         * The state of the Builder will be reset after each call to build(). The client needs to
+         * populate the builder with parameters again if they choose to re-use it.
+         *
          * @return A TangentSpaceMesh
          */
         TangentSpaceMesh* build();
@@ -195,7 +200,11 @@ public:
         TangentSpaceMesh* mMesh = nullptr;
     };
 
-    ~TangentSpaceMesh() noexcept;
+    /**
+     * Destory the mesh object
+     * @param mesh A pointer to a TangentSpaceMesh ready to be destroyed
+     */
+     static void destroy(TangentSpaceMesh* mesh) noexcept;
 
     /**
      * Move constructor
@@ -310,6 +319,7 @@ public:
     Algorithm getAlgorithm() const noexcept;
 
 private:
+    ~TangentSpaceMesh() noexcept;
     TangentSpaceMesh() noexcept;
     TangentSpaceMesh(const TangentSpaceMesh&) = delete;
     TangentSpaceMesh& operator=(const TangentSpaceMesh&) = delete;
--- a/libs/geometry/include/geometry/TangentSpaceMesh.h
+++ b/libs/geometry/include/geometry/TangentSpaceMesh.h
@@ -216,6 +216,9 @@ public:
      */
     TangentSpaceMesh& operator=(TangentSpaceMesh&& that) noexcept;
 
+    TangentSpaceMesh(const TangentSpaceMesh&) = delete;
+    TangentSpaceMesh& operator=(const TangentSpaceMesh&) = delete;
+
     /**
      * Number of output vertices
      *
@@ -321,8 +324,6 @@ public:
 private:
     ~TangentSpaceMesh() noexcept;
     TangentSpaceMesh() noexcept;
-    TangentSpaceMesh(const TangentSpaceMesh&) = delete;
-    TangentSpaceMesh& operator=(const TangentSpaceMesh&) = delete;
     TangentSpaceMeshInput* mInput;
     TangentSpaceMeshOutput* mOutput;
 
--- a/libs/geometry/src/TangentSpaceMesh.cpp
+++ b/libs/geometry/src/TangentSpaceMesh.cpp
@@ -471,7 +471,7 @@ void TangentSpaceMesh::getTriangles(ushort3* out) const {
 
     const bool is16 = mOutput->triangles16 != nullptr;
     const size_t stride = sizeof(decltype(*out));
-    for (size_t i = 0; i < mOutput->triangleCount; ++i) {
+    for (size_t i = 0, c = mOutput->triangleCount; i < c; ++i) {
         if (is16) {
             *out = mOutput->triangles16[i];
         } else {
--- a/libs/geometry/src/TangentSpaceMesh.cpp
+++ b/libs/geometry/src/TangentSpaceMesh.cpp
@@ -16,8 +16,9 @@
 
 #include <geometry/TangentSpaceMesh.h>
 
-#include <math/vec3.h>
-#include <math/vec4.h>
+#include <math/mat3.h>
+#include <math/norm.h>
+
 #include <utils/Log.h>
 #include <utils/Panic.h>
 
@@ -27,18 +28,18 @@ namespace geometry {
 using namespace filament::math;
 using Builder = TangentSpaceMesh::Builder;
 using Algorithm = TangentSpaceMesh::Algorithm;
+using MethodPtr = void(*)(const TangentSpaceMeshInput*, TangentSpaceMeshOutput*);
+using NormalsOnlyKernelPtr = void(*)(const float3& N, float3& T, float3& B);
 
 struct TangentSpaceMeshInput {
     size_t vertexCount = 0;
     const float3* normals = nullptr;
-    const float4* tangents = nullptr;
     const float2* uvs = nullptr;
     const float3* positions = nullptr;
     const ushort3* triangles16 = nullptr;
     const uint3* triangles32 = nullptr;
 
     size_t normalStride = 0;
-    size_t tangentStride = 0;
     size_t uvStride = 0;
     size_t positionStride = 0;
     size_t triangleCount = 0;
@@ -52,24 +53,22 @@ struct TangentSpaceMeshOutput {
     size_t triangleCount = 0;
     size_t vertexCount = 0;
 
-    float4* tangents = nullptr;
-    float2* uvs = nullptr;
-    float3* positions = nullptr;
-    uint3* triangles32 = nullptr;
-    ushort3* triangles16 = nullptr;
+    const quatf* tangentSpace = nullptr;
+    const float2* uvs = nullptr;
+    const float3* positions = nullptr;
+    const uint3* triangles32 = nullptr;
+    const ushort3* triangles16 = nullptr;
 };
 
 namespace {
 
 const uint8_t NORMALS_BIT = 0x01;
-const uint8_t TANGENTS_BIT = 0x02;
-const uint8_t UVS_BIT = 0x04;
-const uint8_t POSITIONS_BIT = 0x08;
-const uint8_t INDICES_BIT = 0x10;
+const uint8_t UVS_BIT = 0x02;
+const uint8_t POSITIONS_BIT = 0x04;
+const uint8_t INDICES_BIT = 0x08;
 
 // Input types
 const uint8_t NORMALS = NORMALS_BIT;
-const uint8_t NORMALS_TANGENTS = NORMALS_BIT | TANGENTS_BIT;
 const uint8_t POSITIONS_INDICES = POSITIONS_BIT | INDICES_BIT;
 const uint8_t NORMALS_UVS_POSITIONS_INDICES = NORMALS_BIT | UVS_BIT | POSITIONS_BIT | INDICES_BIT;
 
@@ -87,21 +86,19 @@ std::string_view to_string(Algorithm algorithm) {
             return "FRISVAD";
         case Algorithm::FLAT_SHADING:
             return "FLAT_SHADING";
-        case Algorithm::SIGN_OF_W:
-            return "SIGN_OF_W";
         default:
             PANIC_POSTCONDITION("Unknown algorithm %u", static_cast<uint8_t>(algorithm));
     }
 }
 
+#define IS_INPUT_TYPE(inputType, TYPE) ((inputType & TYPE) == TYPE)
+
 Algorithm selectBestDefaultAlgorithm(uint8_t inputType) {
     Algorithm outAlgo;
-    if (inputType & NORMALS_UVS_POSITIONS_INDICES) {
+    if (IS_INPUT_TYPE(inputType, NORMALS_UVS_POSITIONS_INDICES)) {
         outAlgo = Algorithm::MIKKTSPACE;
-    } else if (inputType & POSITIONS_INDICES) {
+    } else if (IS_INPUT_TYPE(inputType, POSITIONS_INDICES)) {
         outAlgo = Algorithm::FLAT_SHADING;
-    } else if (inputType & NORMALS_TANGENTS) {
-        outAlgo = Algorithm::SIGN_OF_W;
     } else {
         ASSERT_PRECONDITION(inputType & NORMALS,
                 "Must at least have normals or (positions + indices) as input");
@@ -110,15 +107,10 @@ Algorithm selectBestDefaultAlgorithm(uint8_t inputType) {
     return outAlgo;
 }
 
-#define IS_INPUT_TYPE(inputType, TYPE) ((inputType & TYPE) == TYPE)
-
 Algorithm selectAlgorithm(TangentSpaceMeshInput *input) {
     uint8_t inputType = 0;
     if (input->normals) {
         inputType |= NORMALS_BIT;
-    }
-    if (input->tangents) {
-        inputType |= TANGENTS_BIT;
     }
     if (input->positions) {
         inputType |= POSITIONS_BIT;
@@ -167,12 +159,6 @@ Algorithm selectAlgorithm(TangentSpaceMeshInput *input) {
                 foundAlgo = true;
             }
             break;
-        case Algorithm::SIGN_OF_W:
-            if (IS_INPUT_TYPE(inputType, NORMALS_TANGENTS)) {
-                outAlgo = Algorithm::SIGN_OF_W;
-                foundAlgo = true;
-            }
-            break;
         default:
             PANIC_POSTCONDITION("Unknown algo %u", static_cast<uint8_t>(input->algorithm));
     }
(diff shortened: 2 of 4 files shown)


libs/geometry/src/TangentSpaceMesh.cpp
```diff
@@ -186,15 +172,130 @@ Algorithm selectAlgorithm(TangentSpaceMeshInput *input) {
...
+        kernel(n, t, b);
+        quats[qindex] = mat3f::packTangentFrame({t, b, n});
+        normal = (const float3*) (((const uint8_t*) normal) + nstride);
+    }
+    output->tangentSpace = quats;
+    output->vertexCount = input->vertexCount;\
+    output->triangleCount = input->triangleCount;\
+    output->uvs = input->uvs;\
+    output->positions = input->positions;\
+    output->triangles32 = input->triangles32;\
```
bejado, line 221: Remove the trailing `\` here

libs/geometry/src/TangentSpaceMesh.cpp
```diff
@@ -186,15 +172,130 @@ Algorithm selectAlgorithm(TangentSpaceMeshInput *input) {
...
+    output->triangles16 = input->triangles16;
+}
+
+#define STRIDE_HELPER(ptr, index, stride, cast_type)\
+    (*((cast_type*) (((const uint8_t*) ptr) + index * stride)))
+
+void flatShadingMethod(const TangentSpaceMeshInput* input, TangentSpaceMeshOutput* output) {
+    const size_t vertexCount = input->vertexCount;
```
bejado, line 229: I don't think `vertexCount` is used anywhere

libs/geometry/src/TangentSpaceMesh.cpp
```diff
@@ -186,15 +172,130 @@ Algorithm selectAlgorithm(TangentSpaceMeshInput *input) {
...
+
+    const float2* uvs = input->uvs;
+    const size_t uvstride = input->uvStride ? input->uvStride : sizeof(float2);
+
+    const bool isTriangle16 = input->triangles16 != nullptr;
+    const uint8_t* triangles = isTriangle16 ? (const uint8_t*) input->triangles16 :
+            (const uint8_t*) input->triangles32;
+
+    const size_t triangleCount = input->triangleCount;
+    const size_t tstride = input->triangles16 ? sizeof(ushort3) : sizeof(uint3);
```
bejado, line 242: You can use `isTriangle16` again here

libs/geometry/src/TangentSpaceMesh.cpp
```diff
@@ -255,16 +353,47 @@ Builder& Builder::algorithm(Algorithm algo) noexcept {
...
+#define CLEAN_HELPER(ptr, inputPtr) \
+   if (ptr && ptr != inputPtr) delete [] ptr;\
+   ptr = nullptr
+
+TangentSpaceMesh::~TangentSpaceMesh() noexcept {
+    CLEAN_HELPER(mOutput->tangentSpace, nullptr);
+    CLEAN_HELPER(mOutput->uvs, mInput->uvs);
+    CLEAN_HELPER(mOutput->positions, mInput->positions);
+    CLEAN_HELPER(mOutput->triangles16, mInput->triangles16);
+    CLEAN_HELPER(mOutput->triangles32, mInput->triangles32);
```
bejado, line 394: This makes me a little nervous. It's probably fine for now, but maybe consider using smart pointers or creating separate pointers in the `mOutput` structure to clearly demarcate between user-provided data and data this class is responsible for freeing.

bejado (approved):

libs/geometry/src/TangentSpaceMesh.cpp
```diff
@@ -52,24 +53,22 @@ struct TangentSpaceMeshOutput {
...
-    float4* tangents = nullptr;
-    float2* uvs = nullptr;
-    float3* positions = nullptr;
-    uint3* triangles32 = nullptr;
-    ushort3* triangles16 = nullptr;
+    const quatf* tangentSpace = nullptr;
+    const float2* uvs = nullptr;
+    const float3* positions = nullptr;
+    const uint3* triangles32 = nullptr;
+    const ushort3* triangles16 = nullptr;
```
pixelflinger, line 60: not a super big deal bu we prefer `type const *` rather than `const type *`

libs/geometry/src/TangentSpaceMesh.cpp
```diff
@@ -87,21 +86,19 @@ std::string_view to_string(Algorithm algorithm) {
...
         case Algorithm::FLAT_SHADING:
             return "FLAT_SHADING";
-        case Algorithm::SIGN_OF_W:
-            return "SIGN_OF_W";
         default:
             PANIC_POSTCONDITION("Unknown algorithm %u", static_cast<uint8_t>(algorithm));
     }
 }
 
+#define IS_INPUT_TYPE(inputType, TYPE) ((inputType & TYPE) == TYPE)
```
pixelflinger, line 94: I would just use an inline function.

libs/geometry/src/TangentSpaceMesh.cpp
```diff
@@ -186,15 +172,130 @@ Algorithm selectAlgorithm(TangentSpaceMeshInput *input) {
 
     return outAlgo;
 }
-
 #undef IS_INPUT_TYPE
 
+// The paper uses a Z-up world basis, which has been converted to Y-up here
+void frisvadKernel(const float3& N, float3& T, float3& B) {
```
pixelflinger, line 178: should all these function be `noexcept`?

libs/geometry/src/TangentSpaceMesh.cpp
```diff
@@ -186,15 +172,130 @@ Algorithm selectAlgorithm(TangentSpaceMeshInput *input) {
...
+    if (N.y < -1.0f + std::numeric_limits<float>::epsilon()) {
+        // Handle the singularity
+        T = float3{-1.0f, 0.0f, 0.0f};
+        B = float3{0.0f, 0.0f, -1.0f};
+        return;
+    }
+    const float a = 1.0f / (1.0f + N.y);
+    const float b = -N.z * N.x * a;
+    T = float3(b, -N.z, 1.0f - N.z * N.z * a);
+    B = float3(1.0f - N.x * N.x * a, -N.x, b);
```
pixelflinger, line 188: use `float3{}` instead of `float3()` syntax, to be consistant with other places (here and below)

libs/geometry/src/TangentSpaceMesh.cpp
```diff
@@ -186,15 +172,130 @@ Algorithm selectAlgorithm(TangentSpaceMeshInput *input) {
...
+    output->vertexCount = input->vertexCount;\
+    output->triangleCount = input->triangleCount;\
+    output->uvs = input->uvs;\
+    output->positions = input->positions;\
+    output->triangles32 = input->triangles32;\
+    output->triangles16 = input->triangles16;
+}
+
+#define STRIDE_HELPER(ptr, index, stride, cast_type)\
+    (*((cast_type*) (((const uint8_t*) ptr) + index * stride)))
```
pixelflinger, line 226: I would use an inline function and call it something like `pointerAdd`

libs/geometry/src/TangentSpaceMesh.cpp
pixelflinger, line 221: remove the `\` at the end of the line

libs/geometry/src/TangentSpaceMesh.cpp
```diff
@@ -255,16 +353,47 @@ Builder& Builder::algorithm(Algorithm algo) noexcept {
...
 }
 
 TangentSpaceMesh::TangentSpaceMesh() noexcept
         :mInput(new TangentSpaceMeshInput()), mOutput(new TangentSpaceMeshOutput()) {
 }
 
-TangentSpaceMesh::~TangentSpaceMesh() noexcept {}
+#define CLEAN_HELPER(ptr, inputPtr) \
+   if (ptr && ptr != inputPtr) delete [] ptr;\
+   ptr = nullptr
```
pixelflinger, line 387: I think this should be a function too. Generally we try to avoid macros.

libs/geometry/src/TangentSpaceMesh.cpp
```diff
@@ -255,16 +353,47 @@ Builder& Builder::algorithm(Algorithm algo) noexcept {
...
+        default:
+            break;
+    }
+    assert_invariant(method);
+    method(mMesh->mInput, mMesh->mOutput);
+    return mMesh;
 }
 
 TangentSpaceMesh::TangentSpaceMesh() noexcept
         :mInput(new TangentSpaceMeshInput()), mOutput(new TangentSpaceMeshOutput()) {
```
pixelflinger, line 382: do these need to be on the heap? Why not just make them members? Also I don't see where they're `delete`'ed?

libs/geometry/src/TangentSpaceMesh.cpp
```diff
@@ -278,37 +407,85 @@ TangentSpaceMesh& TangentSpaceMesh::operator=(TangentSpaceMesh&& that) noexcept
...
+            *out = ushort3(static_cast<uint16_t>(tri.x),
+                    static_cast<uint16_t>(tri.y),
+                    static_cast<uint16_t>(tri.z));
+        }
+        STRIDE_TO(out, stride);
+    }
 }
 
-void TangentSpaceMesh::getQuats(quatf*, size_t) const noexcept {
+#define GET_QUATS_IMPL(out, stride, conversion) \
```
pixelflinger, line 469: definitely make that an inline.

libs/geometry/src/TangentSpaceMesh.cpp
```diff
@@ -278,37 +407,85 @@ TangentSpaceMesh& TangentSpaceMesh::operator=(TangentSpaceMesh&& that) noexcept
...
+    stride = stride ? stride : sizeof(decltype(*out));\
+    for (size_t i = 0; i < mOutput->vertexCount; ++i) {\
```
pixelflinger, line 471:
```suggestion
    for (size_t i = 0, c = mOutput->vertexCount; i <c; ++i) {\
```

libs/geometry/src/TangentSpaceMesh.cpp
```diff
@@ -278,37 +407,85 @@ TangentSpaceMesh& TangentSpaceMesh::operator=(TangentSpaceMesh&& that) noexcept
...
+        *out = conversion(mOutput->tangentSpace[i].xyzw);\
```
pixelflinger, line 472: why do you need to specify `.xyzw`? also I would cache `mOutput->tangentSpace` outside of the loop, because otherwise the compile will have to reload `mOutput` at each iteration.

In addition if you used an inline function, you could mark the in and out pointers as `UTILS_RESTRICTED`.

pixelflinger (changes requested):

libs/geometry/src/TangentSpaceMesh.cpp
poweifeng: done

libs/geometry/src/TangentSpaceMesh.cpp
poweifeng: done

libs/geometry/src/TangentSpaceMesh.cpp
poweifeng: noted. Will switch to smart pointers in the next patch.

libs/geometry/src/TangentSpaceMesh.cpp
poweifeng: done. Does it also apply to function parameters?

libs/geometry/src/TangentSpaceMesh.cpp
poweifeng: done

libs/geometry/src/TangentSpaceMesh.cpp
poweifeng: done

libs/geometry/src/TangentSpaceMesh.cpp
poweifeng: done

libs/geometry/src/TangentSpaceMesh.cpp
poweifeng: done

libs/geometry/src/TangentSpaceMesh.cpp
poweifeng: Sorry, do you mean `mInput` `mOutput` ? We clear out any allocation in the destructor TangentSpaceMesh.  It's the client's responsibility to call delete on the mesh once it's computed.

libs/geometry/src/TangentSpaceMesh.cpp
poweifeng: done

libs/geometry/src/TangentSpaceMesh.cpp
poweifeng: done

libs/geometry/src/TangentSpaceMesh.cpp
poweifeng: done

libs/geometry/src/TangentSpaceMesh.cpp
poweifeng: done

libs/geometry/src/TangentSpaceMesh.cpp
poweifeng: done

libs/geometry/src/TangentSpaceMesh.cpp
poweifeng: Actually I see what you're saying.  On the heap because the `.h` only sees the forward declaration of these two types, and the actual declaration is in the anonymous namespace here.  

I've delete them in the destructor.

libs/geometry/src/TangentSpaceMesh.cpp
```diff
@@ -87,21 +86,21 @@ std::string_view to_string(Algorithm algorithm) {
             return "FRISVAD";
         case Algorithm::FLAT_SHADING:
             return "FLAT_SHADING";
-        case Algorithm::SIGN_OF_W:
-            return "SIGN_OF_W";
         default:
             PANIC_POSTCONDITION("Unknown algorithm %u", static_cast<uint8_t>(algorithm));
```
pixelflinger, line 90: if a method is `noexcept` it shouldn't have any `PANIC_` in it. Because PANIC_ actually throws. In this case it's impossible to hit 'default' because you're switching on an enum. In fact it's better to not have the `default` because the compiler will complain if you don't have all the `case`

libs/geometry/src/TangentSpaceMesh.cpp
```diff
@@ -87,21 +86,21 @@ std::string_view to_string(Algorithm algorithm) {
...
     }
 }
 
-Algorithm selectBestDefaultAlgorithm(uint8_t inputType) {
+inline bool isInputType(const uint8_t inputType, const uint8_t checkType) {
```
pixelflinger, line 94:
```suggestion
inline bool isInputType(const uint8_t inputType, const uint8_t checkType) noexcept {
```

libs/geometry/src/TangentSpaceMesh.cpp
```diff
@@ -87,21 +86,21 @@ std::string_view to_string(Algorithm algorithm) {
...
+    return ((inputType & checkType) == checkType);
+}
+
+Algorithm selectBestDefaultAlgorithm(uint8_t inputType) noexcept {
```
pixelflinger, line 98: can't be `noexcept` because you have `ASSERT_PRECONDITION`.
From a style standpoint we always try to have the precondition asserts be the first thing the method does -- or have them just before calling the method.

libs/geometry/src/TangentSpaceMesh.cpp
```diff
@@ -138,41 +132,35 @@ Algorithm selectAlgorithm(TangentSpaceMeshInput *input) {
...
             }
             break;
-        case Algorithm::SIGN_OF_W:
-            if (IS_INPUT_TYPE(inputType, NORMALS_TANGENTS)) {
-                outAlgo = Algorithm::SIGN_OF_W;
-                foundAlgo = true;
-            }
-            break;
         default:
             PANIC_POSTCONDITION("Unknown algo %u", static_cast<uint8_t>(input->algorithm));
```
pixelflinger, line 165: That's actually not a post condition failure.  a post condition failure would be something like running out of memory, like... there was no mistake when the method was called, but the method is not able to do what it is supposed to do.
But like before I would argue that the `default` case cannot happen because we're switching on an enum.

libs/geometry/src/TangentSpaceMesh.cpp
```diff
@@ -187,14 +175,155 @@ Algorithm selectAlgorithm(TangentSpaceMeshInput *input) {
...
+    output->uvs = input->uvs;
+    output->positions = input->positions;
+    output->triangles32 = input->triangles32;
+    output->triangles16 = input->triangles16;
+}
+
+template<typename CastType, typename InputType>
+inline CastType pointerAdd(const InputType* ptr, size_t index, size_t stride) {
+    return (*((CastType*) (((const uint8_t*) ptr) + index * stride)));
+}
```
pixelflinger, line 229: I would always have the InputType be the CastType and if you need to cast, just use static_cast<> after calling `pointerAdd`

libs/geometry/src/TangentSpaceMesh.cpp
```diff
@@ -187,14 +175,155 @@ Algorithm selectAlgorithm(TangentSpaceMeshInput *input) {
...
+
+    size_t vindex = 0;
+    for (size_t tindex = 0; tindex < triangleCount; ++tindex) {
+        uint3 tri = isTriangle16 ?
+                uint3(pointerAdd<ushort3, uint8_t>(triangles, tindex, tstride)) :
+                pointerAdd<uint3, uint8_t>(triangles, tindex, tstride);
+
+        const float3 pa = pointerAdd<float3, float3>(positions, tri.x, pstride);
+        const float3 pb = pointerAdd<float3, float3>(positions, tri.y, pstride);
+        const float3 pc = pointerAdd<float3, float3>(positions, tri.z, pstride);
```
pixelflinger, line 262: if you change `pointerAdd` so it takes only one template parameter then you won't need to specify it here, because it'll be deduced.

libs/geometry/src/TangentSpaceMesh.cpp
```diff
@@ -187,14 +175,155 @@ Algorithm selectAlgorithm(TangentSpaceMeshInput *input) {
...
+    }
+}
 
 } // anonymous namespace
 
 Builder::Builder() noexcept
         :mMesh(new TangentSpaceMesh()) {}
 
-Builder::~Builder() noexcept { delete mMesh; }
+Builder::~Builder() noexcept { }
```
pixelflinger, line 326: There is a leak here if the user never calls `build()`. `mMesh` won't be destroyed.

libs/geometry/src/TangentSpaceMesh.cpp
```diff
@@ -255,16 +381,45 @@ Builder& Builder::algorithm(Algorithm algo) noexcept {
...
+            break;
+        case Algorithm::FLAT_SHADING:
+            method = flatShadingMethod;
+            break;
+        default:
+            break;
+    }
+    assert_invariant(method);
+    method(mMesh->mInput, mMesh->mOutput);
+    return mMesh;
```
pixelflinger, line 406: I think this design is a bit problematic because a Builder can only be used once. e.g. this wouldn't work:

```
Builder builder;
builder.triangles();
auto a = builder.build();
auto b = builder.build();
```

So I think at least, we need to reset Builder to its initial state after `build()` and document it.

In addition, I don't like that *we* are allocating the mesh but we let the user destroy it. Instead, you should add a `static TangentSpaceMesh::destroy(TangentSpaceMesh*)` and make the dtor private.
This will allow us to use pools in the future for instance, and it'll protect us from binary compatibility issues too.

libs/geometry/src/TangentSpaceMesh.cpp
```diff
@@ -240,11 +363,14 @@ Builder& Builder::triangleCount(size_t triangleCount) noexcept {
...
 
 Builder& Builder::triangles(const uint3* triangle32) noexcept {
+    ASSERT_PRECONDITION(!mMesh->mInput->triangles16,
+            "Triangles already provided in unsigned shorts");
     mMesh->mInput->triangles32 = triangle32;
     return *this;
 }
 
 Builder& Builder::triangles(const ushort3* triangle16) noexcept {
+    ASSERT_PRECONDITION(!mMesh->mInput->triangles32, "Triangles already provided in unsigned ints");
```
pixelflinger, line 373: because this is noexcept you can't have a `ASSERT_PRECONDITON`.

The pattern that I used elsewhere is to not check for any errors in the builder's method, and do all the error checking in `build`, which is the only method to not be `noexcept`.

pixelflinger (changes requested):

libs/geometry/src/TangentSpaceMesh.cpp
poweifeng: done

libs/geometry/src/TangentSpaceMesh.cpp
poweifeng: done

libs/geometry/src/TangentSpaceMesh.cpp
poweifeng: done

libs/geometry/src/TangentSpaceMesh.cpp
poweifeng: done

libs/geometry/src/TangentSpaceMesh.cpp
poweifeng: done

libs/geometry/src/TangentSpaceMesh.cpp
poweifeng: done

libs/geometry/src/TangentSpaceMesh.cpp
poweifeng: done

libs/geometry/src/TangentSpaceMesh.cpp
poweifeng: done

libs/geometry/src/TangentSpaceMesh.cpp
pixelflinger: yes

libs/geometry/src/TangentSpaceMesh.cpp
```diff
@@ -187,14 +171,157 @@ Algorithm selectAlgorithm(TangentSpaceMeshInput *input) {
...
+    if (abs(N.x) > abs(N.z) + std::numeric_limits<float>::epsilon()) {
+        T = float3{-N.y, N.x, 0.0f};
+    } else {
+        T = float3{0.0f, -N.z, N.y};
+    }
+    T = normalize(T);
+    B = cross(N, T);
+}
+
+void normalsOnlyMethod(const TangentSpaceMeshInput* input, TangentSpaceMeshOutput* output,
```
pixelflinger, line 198: I'm wondering if we shouldn't pass `NormalsOnlyKernelPtr` as a template parameter instead so that we don't have a function call in the main loop. Ideally we want the main loop to be all inlined and possible vectorized.

libs/geometry/src/TangentSpaceMesh.cpp
```diff
@@ -187,14 +171,157 @@ Algorithm selectAlgorithm(TangentSpaceMeshInput *input) {
...
+    NormalsOnlyKernelPtr kernel) noexcept {
+    const size_t vertexCount = input->vertexCount;
+    quatf* quats = new quatf[vertexCount];
+
+    const float3* normal = input->normals;
```
pixelflinger, line 203: `normal` can be UTILS_RESTRICT I believe. And I would rename it to `normals`.

libs/geometry/src/TangentSpaceMesh.cpp
```diff
@@ -187,14 +171,157 @@ Algorithm selectAlgorithm(TangentSpaceMeshInput *input) {
     return outAlgo;
 }
 
-#undef IS_INPUT_TYPE
+// The paper uses a Z-up world basis, which has been converted to Y-up here
+void frisvadKernel(const float3& N, float3& T, float3& B) noexcept {
```
pixelflinger, line 175: We don't have an official rule/style for output parameters like here, but I've been starting to use the return-by-value pattern. in this case it would be like this:

```
std::pair<float3, float3> kernel(float3 const& N) noexcept {
    ...
    return {T, B};
}

// calling site becomes:

auto [t, b] = kernel(n);
```

I like how it looks better. But I'll let you decide if you want to do it this way.

libs/geometry/src/TangentSpaceMesh.cpp
```diff
@@ -187,14 +171,157 @@ Algorithm selectAlgorithm(TangentSpaceMeshInput *input) {
...
+
+    const float3* normal = input->normals;
+    size_t nstride = input->normalStride ? input->normalStride : sizeof(float3);
+
+    for (size_t qindex = 0; qindex < vertexCount; ++qindex) {
+        float3 n = *normal;
+        float3 b, t;
+        kernel(n, t, b);
+        quats[qindex] = mat3f::packTangentFrame({t, b, n});
+        normal = (const float3*) (((const uint8_t*) normal) + nstride);
```
pixelflinger, line 211: I think you can use the `pointerAdd` here.

libs/geometry/src/TangentSpaceMesh.cpp
```diff
@@ -187,14 +171,157 @@ Algorithm selectAlgorithm(TangentSpaceMeshInput *input) {
...
+inline void takeStride(DataType*& out, size_t stride) noexcept {
+    out = (DataType*) (((uint8_t*) out) + stride);
+}
+
+template <typename OutputType>
+using QuatConversionFunc = OutputType(*)(const quatf&);
+
+template<typename OutputType>
+inline void getQuatsImpl(OutputType* UTILS_RESTRICT out, const quatf* UTILS_RESTRICT tangentSpace,
+        size_t vertexCount, size_t stride, QuatConversionFunc<OutputType> conversion) noexcept {
```
pixelflinger, line 309: here also wondering if the conversion shouldn't be a template, so it's inlined in the tight loop.

pixelflinger (changes requested):
A few more changes. Also I think we need a unit test.

libs/geometry/src/TangentSpaceMesh.cpp
poweifeng: Done.  Folded all the function calls into the loop.

libs/geometry/src/TangentSpaceMesh.cpp
poweifeng: done

libs/geometry/src/TangentSpaceMesh.cpp
poweifeng: done

libs/geometry/src/TangentSpaceMesh.cpp
poweifeng: done.  I'm not sure I understand the reasoning for adding `UTILS_RESTRICT` here, but I did add it.

libs/geometry/src/TangentSpaceMesh.cpp
```diff
@@ -187,14 +181,170 @@ Algorithm selectAlgorithm(TangentSpaceMeshInput *input) {
...
+}
```
pixelflinger, line 338: Looks like this isn't used anymore?

libs/geometry/include/geometry/TangentSpaceMesh.h
```diff
@@ -313,6 +319,7 @@ class TangentSpaceMesh {
     Algorithm getAlgorithm() const noexcept;
 
 private:
+    ~TangentSpaceMesh() noexcept;
     TangentSpaceMesh() noexcept;
     TangentSpaceMesh(const TangentSpaceMesh&) = delete;
     TangentSpaceMesh& operator=(const TangentSpaceMesh&) = delete;
```
pixelflinger, line 325: by convention deleted methods should have `public` visibility.

pixelflinger (approved):
There is just two minor comments, otherwise LGTM. Thank you for the unit test!

libs/geometry/src/TangentSpaceMesh.cpp
poweifeng: done

libs/geometry/include/geometry/TangentSpaceMesh.h
poweifeng: done

PR #6539, opened 2023-02-08
Fix JNI after minification is applied

igokoro:
`UbershaderProvider.getNativeObject()` is accessed from `AssetLoader.cpp`, so it must be annotated with `@UsedByNative("AssetLoader.cpp")` to avoid runtime crashes when minification is applied.

I'm able to successfully run my application with minification enabled with this change.

Closes #5954


romainguy (approved):
Thanks!

PR #6546, opened 2023-02-10
The default render channel is now 2 instead of 0

pixelflinger:
This also fixes some potential issues with refraction when using render channels.

### DIFF
--- a/android/filament-android/src/main/java/com/google/android/filament/RenderableManager.java
+++ b/android/filament-android/src/main/java/com/google/android/filament/RenderableManager.java
@@ -307,11 +307,16 @@ public class RenderableManager {
 
         /**
          * Set the channel this renderable is associated to. There can be 4 channels.
-         * All renderables in a given channel are rendered together, regardless of anything else.
-         * They are sorted as usual withing a channel.
-         * Channels work similarly to priorities, except that they enforce the strongest ordering.
          *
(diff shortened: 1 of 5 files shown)


romainguy (approved):

android/filament-android/src/main/java/com/google/android/filament/RenderableManager.java
```diff
@@ -307,11 +307,16 @@ public Builder priority(@IntRange(from = 0, to = 7) int priority) {
 
         /**
          * Set the channel this renderable is associated to. There can be 4 channels.
-         * All renderables in a given channel are rendered together, regardless of anything else.
-         * They are sorted as usual withing a channel.
-         * Channels work similarly to priorities, except that they enforce the strongest ordering.
          *
-         * @param channel clamped to the range [0..3], defaults to 0.
+         * <p>All renderables in a given channel are rendered together, regardless of anything else.
+         * They are sorted as usual withing a channel.</p>
```
poweifeng, line 312: nit: within

poweifeng (approved):

android/filament-android/src/main/java/com/google/android/filament/RenderableManager.java
pixelflinger: thanks

PR #6547, opened 2023-02-10
-Werror for libfilament and backend

poweifeng:


bejado (approved):

filament/backend/src/opengl/platforms/PlatformCocoaTouchGL.mm
```diff
@@ -88,6 +88,9 @@
             pImpl->mGLContext, nullptr, &pImpl->mTextureCache);
     assert_invariant(success == kCVReturnSuccess);
 
+    // Necessary to avoid unused-variable warning in release build.
+    (void)success;
```
pixelflinger, line 92: you can use `UTILS_UNUSED_IN_RELEASE` instead. e.g.:

```
UTILS_UNUSED_IN_RELEASE bool success;
```

pixelflinger (changes requested):
I don't understand why the changes to `HwRenderPrimitiveFactory`, the comparators were written this way for better performance.

poweifeng:
> I don't understand why the changes to `HwRenderPrimitiveFactory`, the comparators were written this way for better performance.

It's because of this: https://github.com/google/filament/actions/runs/4148557313/jobs/7176658562#step:3:2325

Though I'm not sure if that's a bug in clang-8 or not, I'm unable to repro on my local machine of clang-14.

poweifeng:
> > I don't understand why the changes to `HwRenderPrimitiveFactory`, the comparators were written this way for better performance.
> 
> It's because of this: https://github.com/google/filament/actions/runs/4148557313/jobs/7176658562#step:3:2325
> 
> Though I'm not sure if that's a bug in clang-8 or not, I'm unable to repro on my local machine of clang-14.

I'll dig a little further

pixelflinger:
> > > I don't understand why the changes to `HwRenderPrimitiveFactory`, the comparators were written this way for better performance.
> > 
> > 
> > It's because of this: https://github.com/google/filament/actions/runs/4148557313/jobs/7176658562#step:3:2325
> > Though I'm not sure if that's a bug in clang-8 or not, I'm unable to repro on my local machine of clang-14.
> 
> I'll dig a little further

How did it work until now? Did it generate a warning only on that target?
We had many issues with that specific target. 
Maybe you can just disable that warning.

poweifeng:
> > > > I don't understand why the changes to `HwRenderPrimitiveFactory`, the comparators were written this way for better performance.
> > > 
> > > 
> > > It's because of this: https://github.com/google/filament/actions/runs/4148557313/jobs/7176658562#step:3:2325
> > > Though I'm not sure if that's a bug in clang-8 or not, I'm unable to repro on my local machine of clang-14.
> > 
> > 
> > I'll dig a little further
> 
> How did it work until now? Did it generate a warning only on that target? We had many issues with that specific target. Maybe you can just disable that warning.

It's been throwing a warning for previous builds: https://github.com/google/filament/actions/runs/4167262765/jobs/7212628562#step:3:2395

But it is a bug with clang: https://reviews.llvm.org/D60540

I think we might be able to resolve it if we update clang.

filament/backend/src/opengl/platforms/PlatformCocoaTouchGL.mm
poweifeng: done

pixelflinger (approved):

PR #6548, opened 2023-02-10
gltfio: Fix race condition affecting mobile and web

bejado:
When loading a glTF file on platforms without a filesystem, a client calls `addResourceData` to populate `ResourceLoader`'s cache with data. For example, a web client might make HTTP fetch requests to fill in buffer data. This internal cache of data is stored in `ResourceLoader::Impl::mUriDataCache`.

This works well, except the cache must persist until after the data has been uploaded to the GPU.

There was already a mechanism (see `struct UploadEvent ` and `uploadUserdata`) in place to ensure that glTF data persisted until after it had been uploaded to the GPU. However, this mechanism did not extend to client-provided data. Thus, a race occured between Filament's driver consuming the buffer and it getting freed.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -5,6 +5,8 @@ A new header is inserted each time a *tag* is created.
 
 ## main branch
 
+- gltfio: fix potential early freeing of data provided with `ResourceLoader::addResourceData`.
+
 ## v1.31.4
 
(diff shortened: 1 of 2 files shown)


poweifeng (approved):

pixelflinger (approved):
nice one!

bejado:
> did that fix a known github issue?

No GitHub issues AFAIK but an issue seen with the [model-viewer](https://github.com/google/) project.

PR #6559, opened 2023-02-14
Update linux-continuous.yml to use ubuntu-22.04

poweifeng:


bejado (approved):
Thanks for looking into this! Let's also update `.github/workflows/release.yml` and `.github/workflows/presubmit.yml` while we're at it.

poweifeng:
> Thanks for looking into this! Let's also update `.github/workflows/release.yml` and `.github/workflows/presubmit.yml` while we're at it.

done

PR #6564, opened 2023-02-15
optimizations for large scenes

pixelflinger:
[reviewing each change separately might be a lot easier]

The main aim of this changes is to improve CPU performance with very large scenes.
- added some systraces
- improved `StructureOfArray` resize
- made `FScene::prepare` multithreaded
- optimizations around `visitScene`: avoid calling when possible, and optimize when called (more to be done here)

All these optimization save 15 to 20ms on a 50ms frame (CPU time only) with ~20K renderables


filament/src/ShadowMap.cpp
```diff
@@ -1128,13 +1128,13 @@ void ShadowMap::visitScene(const FScene& scene, uint32_t visibleLayers,
     SYSTRACE_CALL();
 
     using State = FRenderableManager::Visibility;
-    FScene::RenderableSoa const& UTILS_RESTRICT soa = scene.getRenderableData();
```
romainguy, line 1131: Why remove the restricts here?

filament/src/ShadowMapManager.cpp
```diff
@@ -458,8 +459,18 @@ ShadowMapManager::ShadowTechnique ShadowMapManager::updateCascadeShadowMaps(FEng
         // entire camera frustum, as if we only had a single cascade.
         ShadowMap& shadowMap = *mCascadeShadowMaps[0];
 
+        const auto direction = lightData.elementAt<FScene::DIRECTION>(0);
```
romainguy, line 462: Shouldn't we pick the index of the main directional light? (i.e. the one of highest intensity)

filament/src/ShadowMap.cpp
pixelflinger: Ah yes, they were useless because we're only reading from these pointers. I don't know why I did that.

filament/src/ShadowMapManager.cpp
pixelflinger: `lightData` only store the main directional light, at index 0.  This is done in `FScene::prepare`

PR #6572, opened 2023-02-17
Add UV-based tangent space algorithms

poweifeng:
Added mikktspace as a third party lib


libs/geometry/src/TangentSpaceMesh.cpp
```diff
@@ -321,6 +323,252 @@ void flatShadingMethod(const TangentSpaceMeshInput* input, TangentSpaceMeshOutpu
     output->triangleCount = outTriangleCount;
 }
 
+class MikktspaceImpl {
```
pixelflinger, line 326: maybe MikktspaceImpl could be in its own cpp?

libs/geometry/src/TangentSpaceMesh.cpp
```diff
@@ -321,6 +323,252 @@ void flatShadingMethod(const TangentSpaceMeshInput* input, TangentSpaceMeshOutpu
...
+    }
+
+    inline const uint3 getTriangle(int triangleIndex) const noexcept {
+        const size_t tstride = mIsTriangle16 ? sizeof(ushort3) : sizeof(uint3);
+        return mIsTriangle16 ?
+               uint3(*(ushort3*)(pointerAdd(mTriangles, triangleIndex, tstride))) :
+               *(uint3*)(pointerAdd(mTriangles, triangleIndex, tstride));
+    }
+
+    TangentSpaceMeshOutput* mOutput;
```
bejado, line 460: I think `mOutput` is unused.

libs/geometry/src/TangentSpaceMesh.cpp
```diff
@@ -321,6 +323,252 @@ void flatShadingMethod(const TangentSpaceMeshInput* input, TangentSpaceMeshOutpu
...
+    size_t const uvStride = input->uvStride ? input->uvStride : sizeof(float2);
+    auto const* triangles16 = input->triangles16;
+    auto const* triangles32 = input->triangles32;
+    auto const* positions = input->positions;
+    auto const* uvs = input->uvs;
+    auto const* normals = input->normals;
+
+    std::vector<float3> tan1(vertexCount);
+    std::vector<float3> tan2(vertexCount);
+    memset(tan1.data(), 0, sizeof(float3) * vertexCount);
```
bejado, line 503: Do we need the `memset` here? Could we instead do something like:

```
    std::vector<float3> tan1(vertexCount, float3{0.0f});
    std::vector<float3> tan2(vertexCount, float3{0.0f});
```

libs/geometry/src/TangentSpaceMesh.cpp
poweifeng: done

libs/geometry/src/TangentSpaceMesh.cpp
poweifeng: done

libs/geometry/src/TangentSpaceMesh.cpp
poweifeng: done.  

Might be there for optimization (copied from SurfaceOrientation), but will remove since I'm not sure.

libs/geometry/CMakeLists.txt
```diff
@@ -26,7 +26,7 @@ include_directories(${PUBLIC_HDR_DIR})
 
 add_library(${TARGET} STATIC ${PUBLIC_HDRS} ${SRCS})
 
-target_link_libraries(${TARGET} PUBLIC math utils)
+target_link_libraries(${TARGET} PUBLIC math meshoptimizer mikktspace utils)
```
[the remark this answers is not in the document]
bejado: Yeah, I think in that scenario we'd get linker errors. I think it's worth it for now to reduce the number of individual libraries clients are required to link against.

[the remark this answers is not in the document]
pixelflinger: actually `meshoptimizer` and `mikktspace` shouldn't be `PUBLIC`.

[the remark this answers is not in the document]
poweifeng: @pixelflinger done.

bejado (approved):

pixelflinger (approved):

PR #6580, opened 2023-02-21
Fix build errors on macOS

daemyung:
Build errors are caused by deprecated enums.

PR #6581, opened 2023-02-21
Fix build errors caused by deprecated enums on macOS

daemyung:


daemyung:
@romainguy How's about this approch?

bejado:
What is the value of `CMAKE_OSX_DEPLOYMENT_TARGET` inside your `CMakeCache.txt`? It should be 10.15. I'm not sure how you'd be seeing the errors you mentioned in #6580 unless that was set to 13.0.

daemyung:
@bejado `CMAKE_OSX_DEPLOYMENT_TARGET` is empty.

```
//Minimum OS X version to target for deployment (at runtime); newer
// APIs weak linked. Set to empty string for default value.
CMAKE_OSX_DEPLOYMENT_TARGET:STRING=
``
`

bejado:
@daemyung Try setting it to `10.15`. It should be getting set to 10.15 in our [build.sh script here](https://github.com/google/filament/blob/29983772edfddf0ed2ebf29ab1c3e198c4a749e8/build.sh#L215-L223).

daemyung:
@bejado It's fine now. Do you want to close this PR? I think it's worth to merge.

bejado:
@daemyung I'd prefer not to have unnecessary`#if` checks.

PR #6582, opened 2023-02-21
Support ccache on Windows with ninja generator

ViktorHeisenberger:
Building Filament or its parts on Windows, using the `ninja` generator, with `ccache` installed, fails with the following message: 
`ninja: fatal: CreateProcess: %1 is not a valid Win32 application.`

This is due to the way `ccache` support is implemented in the root `CMakeLists.txt` script. Executing line 70, `cmake` will find `ccache` installed, and enter the `if` branch on line 71. It will then configure the `launch-c` and `launch-cxx` Bash scripts, and after failing the `Xcode` check, will set them as compiler launchers (lines 88-90). When `ninja` tries to run these scripts, it will fail with the message above.

This PR adds a condition to handle Windows, and sets `ccache` directly as compiler launcher. In order to support passing parameters to `ccache`, either a batch file, or a powershell script would be needed. I failed to properly create a batch script (it always wanted to open the command in a new command prompt window), and dismissed powershell, as it is disabled by default to run scripts that aren't digitally signed. The `CCACHE_CPP2=true` parameter is not needed, as it's `ccache`'s current default behavior. The `CCACHE_SLOPPINESS` parameter affects cache hit rate. Building did become faster without it, so I accepted the result.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -69,25 +69,30 @@ endif()
 # ==================================================================================================
 find_program(CCACHE_PROGRAM ccache)
 if (CCACHE_PROGRAM)
-    set(C_LAUNCHER   "${CCACHE_PROGRAM}")
-    set(CXX_LAUNCHER "${CCACHE_PROGRAM}")
-
-    configure_file(build/launch-c.in   launch-c)
-    configure_file(build/launch-cxx.in launch-cxx)
-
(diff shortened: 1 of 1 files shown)


romainguy:
Thank you!

romainguy (approved):

PR #6589, opened 2023-02-23
Cleanup how we get GL extensions entry-points

pixelflinger:
- iOS is treated the same way than Android now. The only difference is that iOS only provides prototypes and no typedef, whereas  Android only provides typedefs and no prototypes.

- Timer queries is core in GL and an extension on all versions of GLES. On iOS it's not available at the header level. An additional subtlety is that glGetObjectuiv is core in GLES 3.0, so it conflicts with the extension. So, now we do things correctly:
  - on desktop we use the core methods
  - on ios we ifdef out everything related to timer queries
  - on gles 2.0 and up we use only the extension entry-points

### DIFF
--- a/filament/backend/src/opengl/OpenGLDriver.cpp
+++ b/filament/backend/src/opengl/OpenGLDriver.cpp
@@ -69,7 +69,9 @@ using namespace utils;
 namespace filament::backend {
 
 Driver* OpenGLDriverFactory::create(
-        OpenGLPlatform* const platform, void* const sharedGLContext, const Platform::DriverConfig& driverConfig) noexcept {
+        OpenGLPlatform* const platform,
+        void* const sharedGLContext,
+        const Platform::DriverConfig& driverConfig) noexcept {
     return OpenGLDriver::create(platform, sharedGLContext, driverConfig);
 }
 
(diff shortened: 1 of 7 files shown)


filament/backend/src/opengl/OpenGLDriver.cpp
```diff
@@ -177,6 +179,10 @@ OpenGLDriver::OpenGLDriver(OpenGLPlatform* platform, const Platform::DriverConfi
     slog.i << "OS version: " << mPlatform.getOSVersion() << io::endl;
 #endif
 
+    // Timer queries are core in GL 3.3, otherwise we need EXT_disjoint_timer_query
+    // iOS headers don't define GL_EXT_disjoint_timer_query, so make aboslutely sure
```
bejado, line 183:
```suggestion
    // iOS headers don't define GL_EXT_disjoint_timer_query, so make absolutely sure
```

bejado (approved):

filament/backend/src/opengl/OpenGLDriver.cpp
pixelflinger: thank you ben. I will fix in a follow-up PR.

PR #6590, opened 2023-02-24
Use member init on BlitFunctionKey

daemyung:


romainguy:
Please explain the reason of your change.

bejado (changes requested):
We've run into issues initializing keys this way. The upper bits of a `bool` are undefined, which comes into play when `BlitFunctionKey` is hashed. We use `std::memset` to be safe.

daemyung:
I see. I just want to remove clang warnings and enhance readability.

PR #6593, opened 2023-02-24
vulkan: Make readPixels async

poweifeng:


filament/backend/src/vulkan/VulkanDriver.cpp
```diff
@@ -1323,150 +1326,11 @@ void VulkanDriver::stopCapture(int) {
...
-    // Transition the source image layout (which might be the swap chain)
-
-    const VkImageSubresourceRange srcRange = {
-        .aspectMask = VK_IMAGE_ASPECT_COLOR_BIT,
-        .baseMipLevel = srcAttachment.level,
-        .levelCount = 1,
-        .baseArrayLayer = srcAttachment.layer,
-        .layerCount = 1,
+    auto destroypbd = [this](PixelBufferDescriptor&& pbd) {
+        scheduleDestroy(std::move(pbd));
```
bejado, line 1330: Is `this` (the `VulkanDriver`) still guaranteed to be alive when this callback is called?

bejado (approved):

poweifeng:
Will attempt again using a single-threaded solution.

PR #6595, opened 2023-02-27
geometry: clean up TangentSpaceMesh

poweifeng:
- Reordered the 'const var' declarations to 'var const'
 - Use std::vector for data arrays instead of manually managing allocations

### DIFF
--- a/libs/geometry/src/TangentSpaceMeshInternal.h
+++ b/libs/geometry/src/TangentSpaceMeshInternal.h
@@ -22,6 +22,8 @@
 #include <math/mat3.h>
 #include <math/norm.h>
 
+#include <utils/Panic.h>
+
 #include <vector>
 
 namespace filament::geometry {
@@ -32,21 +34,24 @@ using Algorithm = TangentSpaceMesh::Algorithm;
 template<typename T>
 class InternalArray {
 public:
-    void borrow(const T* ptr) noexcept {
+    void borrow(T const* ptr) {
+        assert_invariant(mAllocated.empty());
         mBorrowed = ptr;
     }
 
-    T* allocate(size_t size) noexcept {
+    T* allocate(size_t size) {
+        assert_invariant(!mBorrowed);
         mAllocated.resize(size);
         mAllocated.shrink_to_fit();
         return mAllocated.data();
     }
 
     explicit operator bool() const noexcept {
-        return !mBorrowed && mAllocated.size() > 0;
+        return !mBorrowed && !mAllocated.empty();
     }
 
     const T* get() {
+        assert_invariant((bool) this);
         if (mBorrowed) {
             return mBorrowed;
         }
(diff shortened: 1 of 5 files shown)


libs/geometry/src/TangentSpaceMeshInternal.h
```diff
@@ -22,11 +22,42 @@
...
 namespace filament::geometry {
 
 using namespace filament::math;
 using Algorithm = TangentSpaceMesh::Algorithm;
 
+template<typename T>
+class InternalArray {
+public:
+    void borrow(const T* ptr) noexcept {
+        mBorrowed = ptr;
```
pixelflinger, line 36: Should it be an error to borrow a non-empty vector? Maybe we at least need an assert?

libs/geometry/src/TangentSpaceMeshInternal.h
```diff
@@ -22,11 +22,42 @@
...
+    }
+
+    T* allocate(size_t size) noexcept {
```
pixelflinger, line 39: Should it be an error to allocate() a borrowed vector? assert?

libs/geometry/src/TangentSpaceMeshInternal.h
```diff
@@ -22,11 +22,42 @@
...
+
+    explicit operator bool() const noexcept {
+        return !mBorrowed && mAllocated.size() > 0;
+    }
+
+    const T* get() {
+        if (mBorrowed) {
+            return mBorrowed;
+        }
+        return mAllocated.data();
```
pixelflinger, line 53: Here you should assert() that the vector is not empty.

libs/geometry/src/TangentSpaceMeshInternal.h
```diff
@@ -22,11 +22,42 @@
...
+    }
+
+private:
+    std::vector<T> mAllocated;
+    const T* mBorrowed = nullptr;
```
pixelflinger, line 58: `T const*` :-)

libs/geometry/src/TangentSpaceMeshInternal.h
poweifeng: done

libs/geometry/src/TangentSpaceMeshInternal.h
poweifeng: done

libs/geometry/src/TangentSpaceMeshInternal.h
poweifeng: done

libs/geometry/src/TangentSpaceMeshInternal.h
poweifeng: done

pixelflinger (approved):

PR #6601, opened 2023-02-28
rework again how we manage different versions of GL 

pixelflinger:
try to be more explicit about which configurations are supported,
and use the same pattern everywhere for checking the gl version at
either compile and runtime.

### DIFF
--- a/filament/backend/src/opengl/gl_headers.h
+++ b/filament/backend/src/opengl/gl_headers.h
@@ -89,7 +89,7 @@ extern PFNGLDEBUGMESSAGECALLBACKKHRPROC glDebugMessageCallbackKHR;
 extern PFNGLGETDEBUGMESSAGELOGKHRPROC glGetDebugMessageLogKHR;
 #endif
 #ifdef GL_EXT_clip_control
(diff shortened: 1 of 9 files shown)


filament/backend/src/opengl/gl_headers.h
```diff
@@ -46,12 +66,22 @@
...
 
+/* Validate the header configurations we aim to support */
 
-#if (!defined(GL_ES_VERSION_2_0) && !defined(GL_VERSION_4_1))
-#error "Minimum header version must be OpenGL ES 2.0 or OpenGL 4.1"
+#if defined(GL_VERSION_4_5)
+#elif defined(GL_ES_VERSION_3_1)
+#elif defined(GL_ES_VERSION_3_0)
+#   if !defined(IOS) && !defined(__EMSCRIPTEN__)
+#       error "GLES 3.0 headers only supported on iOS"
```
bejado, line 75: Should this error read "GLES 3.0 headers only supported on iOS and WebGL2"?

pixelflinger: yep. thanks.

PR #6618, opened 2023-03-07
Metal: fix memory leak in readPixels

ryumiel:
The PixelBufferDescriptor was not being deallocated properly, which resulted in the leak. This patch explicitly deletes the PixelBufferDescriptor at the end of the callback to prevent the leak .This is necessary as the move constructor does not automatically deallocate the existing PixelBufferDescriptor.


bejado (approved):
Nice catch, thanks!

PR #6620, opened 2023-03-08
gltfio: Release cancelled, decoded texture resources

poweifeng:
When we call TextureProvider::cancelDecoding, we should make sure that textures that have been decoded, but not yet used (popped) should be released (i.e. memory freed and the meta data marked appropriately.)


bejado (approved):

libs/gltfio/src/StbProvider.cpp
```diff
@@ -219,6 +219,18 @@ void StbProvider::cancelDecoding() {
...
+
+    // For cancelled jobs, we need to set the TextureInfo to the popped state and free the decoded
+    // data.
+    for (auto& info : mTextures) {
+        if (info->state != TextureState::DECODING) {
+            continue;
+        }
+        if (intptr_t data = info->decodedTexelsBaseMipmap.load()) {
+            delete [] (uint8_t*) data;
+        }
```
pixelflinger, line 231: Is there a risk that another thread deletes this `data` between the `load()` and the `delete[]`? Asking because an atomic is involved. Either way I think a comment about that would be useful here.

libs/ktxreader/include/ktxreader/Ktx2Reader.h
```diff
@@ -142,7 +142,7 @@ class Ktx2Reader {
 
         protected:
             Async() noexcept = default;
-            ~Async() = default;
+            virtual ~Async() = default;
```
pixelflinger, line 145: virtual method should always be defined in a cpp.

libs/gltfio/src/StbProvider.cpp
poweifeng: added comment 
https://github.com/google/filament/pull/6620/files#diff-22acc1b406850d2147dd6e5930f60ea0eb3e38cc31f1493ef5b63d9a5bfb02c5R229

libs/ktxreader/include/ktxreader/Ktx2Reader.h
poweifeng: done

romainguy (approved):

pixelflinger (approved):

PR #6629, opened 2023-03-10
third_party: fix unqualified call to 'std::move'

mackong:


bejado:
Thanks, but these changes should be made to the source-of-truth Khronos repository:

https://github.com/KhronosGroup/SPIRV-Cross

PR #6630, opened 2023-03-10
web: fix animation stopped premature

mackong:


bejado:
What's the reason for this change?

pixelflinger:
> What's the reason for this change?

What I'm more curious about is why was there a modulo before?

pixelflinger (commented):
This looks good to me but before accepting the change I want to understand why they was a modulo before.

mackong:
@bejado for glTF with animations duration long than 1s, modular operate will make these animations stop premature and loop again. 

This is a test glTF with long duration animations. [MSBanner.zip](https://github.com/google/filament/files/10947208/MSBanner.zip)

Following is result with modular:
https://user-images.githubusercontent.com/2212586/224456981-d244d7fb-f95a-4811-b31c-ec0d75cd5475.mp4

and without modular:
https://user-images.githubusercontent.com/2212586/224457002-fbb010cb-1452-46fc-84e2-1c7af54e0cba.mp4

bejado (approved):

PR #6631, opened 2023-03-10
gltfio: fix crash when custom attributes used

mackong:


bejado:
Thanks, this makes sense to me. Do you have a glTF file that uses custom attributes we can use to verify?

libs/gltfio/src/TangentsJob.cpp
```diff
@@ -49,7 +49,7 @@ void TangentsJob::run(Params* params) {
     std::unique_ptr<float3[]> morphDeltas;
 
     // Build a mapping from cgltf_attribute_type to cgltf_accessor.
-    const int NUM_ATTRIBUTES = 8;
+    const int NUM_ATTRIBUTES = cgltf_attribute_type_max_enum;
```
pixelflinger, line 52: Should it be `std::min(cgltf_attribute_type_max_enum, 8)` or have some assert if it's more than 8? @poweifeng ?

mackong:
[robot.zip](https://github.com/google/filament/files/10947140/robot.zip)

This is glTF file with custom attributes as follows:

image

**NOTE:** on linux or android, load this glTF will not make app to crash sometimes, seems caused by "undefined behaviour"; but on mac or iOS, app will crash by SIGABRT always.

poweifeng (approved):

PR #6633, opened 2023-03-11
Pick transparent renderables

### LINKED ISSUE #6614: Picking transparent renderables
Filament only supports picking opaque renderables, however, it may be helpful to pick transparent renderables, for example, a glass Stanford dragon or an Android View renderable in SceneView.

It's possible to move picking to a separate pass so that transparent renderables can participate in picking without affecting the structure buffer. I've already created a patch that implements that. It may be also useful to support excluding materials from picking.

Using AABB picking instead of Filament picking in SceneView, however, picking transparent renderables can be useful in other frameworks and applications.

**OS and backend**
The feature isn't OS or backend specific.

grassydragon:
Closes  #6614

### DIFF
--- a/filament/src/RenderPass.cpp
+++ b/filament/src/RenderPass.cpp
@@ -669,7 +669,9 @@ RenderPass::Command* RenderPass::generateCommandsImpl(uint32_t extraFlags,
 
                 // FIXME: should writeDepthForShadowCasters take precedence over mi->getDepthWrite()?
                 cmdDepth.primitive.rasterState.depthWrite = (1 // only keep bit 0
-                        & (mi->isDepthWriteEnabled() | (mode == TransparencyMode::TWO_PASSES_ONE_SIDE))
+                        & (mi->isDepthWriteEnabled()
+                            || (mode == TransparencyMode::TWO_PASSES_ONE_SIDE)
+                            || Variant::isPickingVariant(variant))
                         & !(filterTranslucentObjects & translucent)
                         & !(depthFilterAlphaMaskedObjects & rs.alphaToCoverage))
                             | writeDepthForShadowCasters;
--- a/filament/src/PostProcessManager.cpp
+++ b/filament/src/PostProcessManager.cpp
@@ -352,7 +352,6 @@ PostProcessManager::StructurePassOutput PostProcessManager::structure(FrameGraph
     // It consists of a mipmapped depth pass, tuned for SSAO
     struct StructurePassData {
         FrameGraphId<FrameGraphTexture> depth;
-        FrameGraphId<FrameGraphTexture> picking;
     };
 
     // sanitize a bit the user provided scaling factor
@@ -377,24 +376,14 @@ PostProcessManager::StructurePassOutput PostProcessManager::structure(FrameGraph
                 data.depth = builder.write(data.depth,
                         FrameGraphTexture::Usage::DEPTH_ATTACHMENT | FrameGraphTexture::Usage::SAMPLEABLE);
 
-                if (config.picking) {
-                    data.picking = builder.createTexture("Picking Buffer", {
-                            .width = width, .height = height,
-                            .format = TextureFormat::RG32UI });
-
-                    data.picking = builder.write(data.picking,
-                            FrameGraphTexture::Usage::COLOR_ATTACHMENT);
-                }
-
                 builder.declareRenderPass("Structure Target", {
-                        .attachments = { .color = { data.picking }, .depth = data.depth },
-                        .clearFlags = TargetBufferFlags::COLOR0 | TargetBufferFlags::DEPTH
+                        .attachments = { .depth = data.depth },
+                        .clearFlags = TargetBufferFlags::DEPTH
                 });
             },
             [=, renderPass = pass](FrameGraphResources const& resources,
                     auto const& data, DriverApi& driver) mutable {
                 Variant structureVariant(Variant::DEPTH_VARIANT);
-                structureVariant.setPicking(config.picking);
 
                 auto out = resources.getRenderPassInfo();
                 renderPass.setRenderFlags(structureRenderFlags);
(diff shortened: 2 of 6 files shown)


filament/src/RenderPass.cpp
```diff
@@ -669,7 +669,9 @@ RenderPass::Command* RenderPass::generateCommandsImpl(uint32_t extraFlags,
 
                 // FIXME: should writeDepthForShadowCasters take precedence over mi->getDepthWrite()?
                 cmdDepth.primitive.rasterState.depthWrite = (1 // only keep bit 0
-                        & (mi->isDepthWriteEnabled() | (mode == TransparencyMode::TWO_PASSES_ONE_SIDE))
+                        & (mi->isDepthWriteEnabled()
+                            || (mode == TransparencyMode::TWO_PASSES_ONE_SIDE)
+                            || Variant::isPickingVariant(variant))
```
grassydragon, line 674: Should I add a boolean material property to support excluding a material from picking?

romainguy: @pixelflinger can explain this in more detail but this function is in the deep inner loop of Filament and written to be branchless. You should be using long circuit operators instead of short circuit. I'm also not sure if we want to test for picking here.

grassydragon:
> How are transparents handled with this change? Do they take precedence over opaques behind them?

I believe that the whole scene is rendered to the picking buffer based on the fragment depth. Yes, a renderable is picked if it is the closest to the camera.

filament/src/RenderPass.cpp
grassydragon: > You should be using long circuit operators instead of short circuit.

Then I can cast the boolean value to an integer to avoid the compilation error.

> I'm also not sure if we want to test for picking here.

It is required to render the transparent objects in the picking pass.

pixelflinger:
Thanks for the PR I need to maul over it for a while. My areas of concerns:

- we have an extra pass (in the case of SSAO enabled, we would reuse the structure pass).
- the picking pass is now full resolution, so 4x more costly
- me might want a way to exclude the transparents (?)

Maybe a picking pass is the right way to do it, this would allow to control what goes into that pass.  Maybe we need a way to select the resolution, because 1/4 res picking might still be useful at 1/4 the price.
Also wondering if there is a clever way to use the color/depth write modes to be able to generate the picking buffer in the same pass as the structure buffer as before.

Lots of things to think about...

grassydragon:
> me might want a way to exclude the transparents (?)

Yes, I think it would be great to use a boolean material property for that.

grassydragon:
Hi @pixelflinger!
What are your thoughts on this pull request? I'm ready to attempt adding more changes if needed.

pixelflinger:
> Hi @pixelflinger! What are your thoughts on this pull request? I'm ready to attempt adding more changes if needed.

We're very sorry for the delay, we've been, and still are, very busy. This will take a bit longer.

dandingol03:
@grassydragon  How is going about this pr. I also met the same problem which transparent cannot be picking in `post-process`

grassydragon:
> @grassydragon @pixelflinger  How is going about this pr. I also met the same problem which transparent cannot be picking in `post-process`

I guess that everyone at Google is now preparing for I/O.

hshapley:
With I/O behind us, any chance the team could take another look at this? We have a need for picking transparent renderables, so it'd be great to have a solution for this. Would supporting customization of the picking pass (ability to exclude transparents and specify resolution) be sufficient to get this approved? Or is the current approach sufficient for now? Thanks.

Precha24 (approved):
Good, thanks

grassydragon:
Hi @Precha24!
Do you need any help with resolving the issues in the pull request?

Precha24:
Yes I do

Precha24 (commented):
Good, thanks

grassydragon:
Hi @pixelflinger!
I'm ready to address any issues in the pull request, however, we need to agree on how this feature should be implemented. Does the Filament team have a complete vision on that?

pixelflinger:
@grassydragon sorry for the long time getting back to this PR.

This is what I would like to see:
- have an option to include or not the transparents. I think there is value to not include the transparents in picking.
- reuse the structure pass in the case transparents are not included (each extra pass costs about the same as a shadow pass, it's not insignificant). Thanks to the framegraph design this can be done pretty easily: keep the old code, keep the new code and just select which "picking" resource you want to use based on wether transparents should be included -- the framegraph will take care of removing the pass that's not needed.

mbalajee:
@grassydragon Thanks for fixing this.  We're also [blocked](https://github.com/google/filament/issues/7950) because of this issue.  
I am not super familiar with the changes in the PR but I can attempt to fix the conflict @grassydragon lemme know, thanks 🙏 @pixelflinger

grassydragon:
@mbalajee, to be honest, I've been planning to close this pull request for a while since we've gone back to using a collision system for picking 3D objects in Kotlin instead of doing that on the GPU. However, you can try implementing the suggestions given by @pixelflinger and resolving the conflicts.

pixelflinger:
Yes, I kept it open because it's an interesting feature and definitely something that would be good to fix. 
We do not have any urgency on it though.

mbalajee:
> @mbalajee, to be honest, I've been planning to close this pull request for a while since we've gone back to using a collision system for picking 3D objects in Kotlin instead of doing that on the GPU. However, you can try implementing the suggestions given by @pixelflinger and resolving the conflicts.

I see, thanks for the update.  Is your code for  "collision system for picking 3D objects in Kotlin" available publicly?  If not any lead would be helpful, would like to see which works best for us

grassydragon:
> Is your code for  "collision system for picking 3D objects in Kotlin" available publicly?

Yes, it's a part of the Sceneview framework but originally comes from Sceneform: https://github.com/SceneView/sceneview-android

mbalajee:
> > Is your code for  "collision system for picking 3D objects in Kotlin" available publicly?
> 
> Yes, it's a part of the Sceneview framework but originally comes from Sceneform: https://github.com/SceneView/sceneview-android

Thanks I'll take a look at that.  If I get some time will try to address the comments in this PR.

PR #6637, opened 2023-03-13
vulkan: fix readPixels

poweifeng:
- Need to explicitly clean-up ReadPixels and not use its destructor
 - Wait on previously created fences (for the current frame) before submitting the commands for reading pixels
 - Fix passing block-scoped var into Vk

### DIFF
--- a/filament/backend/src/vulkan/VulkanCommands.cpp
+++ b/filament/backend/src/vulkan/VulkanCommands.cpp
@@ -271,7 +271,42 @@ void VulkanCommands::updateFences() {
     }
 }
 
-} // namespace filament::backend
+bool VulkanCommands::isValidFence(VkFence inFence) const {
+    for (auto& wrapper: mStorage) {
+        if (wrapper.cmdbuffer == VK_NULL_HANDLE) {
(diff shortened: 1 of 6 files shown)


filament/backend/src/vulkan/VulkanCommands.cpp
```diff
@@ -271,7 +271,42 @@ void VulkanCommands::updateFences() {
...
+        VkResult const result = fence->status.load(std::memory_order_relaxed);
+        if (fence->fence == inFence && (result == VK_INCOMPLETE || result == VK_NOT_READY)) {
+            return true;
+        }
+    }
+    return false;
+}
+
+std::vector<VkFence> VulkanCommands::getFences() const {
+    std::vector<VkFence> fences;
```
pixelflinger, line 292: maybe you can reserve mStorage.size() to avoid multiple allocations at push_back? Also considere using a FixedCapacityVector.

pixelflinger (approved):

poweifeng:
Obsolete since this is an outdated attempt

PR #6639, opened 2023-03-14
Vulkan: fix stack use-after-free

bejado:
`clearValues` needs to live past the call to `vkCmdBeginRenderPass`.


poweifeng (approved):

pixelflinger (approved):
I suppose we should use Rust :-)

PR #6646, opened 2023-03-17
Update Materials.md.html

Bolloxim:
Fixed typo in the source documentation : getUserTimeMode(float m) which should be getUserTimeMod(float m)


bejado (approved):
Thanks!

PR #6652, opened 2023-03-20
Expose specialization constants to materials as constant parameters

bejado:
Still TODO:

- [ ] JavaScript/TS bindings
- [ ] Java bindings
- [ ] Switch internal Filament materials to use constant parameters


filament/src/details/Material.cpp
```diff
@@ -179,6 +194,45 @@ FMaterial::FMaterial(FEngine& engine, const Material::Builder& builder)
...
     }
 
+    utils::FixedCapacityVector<MaterialConstant> constants;
+    // Older materials won't have a constants chunk, but that's okay.
+    parser->getConstants(&constants);
+
+    // Verify that all the constant specializations exist in the material and that their types match.
+    // The first two specialization constants are defined internally by Filament.
+    // Constants #2 onwards are user-defined in the material.
+    mSpecializationConstants.reserve(constants.size() + 2);
```
pixelflinger, line 204: should we keep some slots as reserved? Thinking 8 maybe?

filament/src/details/Material.cpp
```diff
@@ -179,6 +194,45 @@ FMaterial::FMaterial(FEngine& engine, const Material::Builder& builder)
...
+                ASSERT_PRECONDITION(std::holds_alternative<float>(value), errorMessage,
+                        name, mName.c_str_safe(), "float", types[value.index()]);
+                break;
+            case ConstantType::BOOL:
+                ASSERT_PRECONDITION(std::holds_alternative<bool>(value), errorMessage,
+                        name, mName.c_str_safe(), "bool", types[value.index()]);
+                break;
+        }
+        uint32_t index = std::distance(constants.begin(), found);
+        mSpecializationConstants.push_back({index + 2, value});
```
pixelflinger, line 233: also maybe don't hardcode `2` since it's used more than once.

libs/filamat/src/Enums.cpp
```diff
@@ -160,4 +160,15 @@ std::unordered_map<std::string, SamplerFormat>& Enums::getMap<SamplerFormat>() n
     return mStringToSamplerFormat;
 };
 
+std::unordered_map<std::string, ConstantType > Enums::mStringToConstantType = {
+        { "int",   ConstantType ::INT },
+        { "float", ConstantType ::FLOAT },
+        { "bool",  ConstantType ::BOOL },
```
pixelflinger, line 166: extra space before `::`

filament/src/details/Material.h
```diff
@@ -27,11 +27,13 @@
 #include <private/filament/SamplerInterfaceBlock.h>
 #include <private/filament/SubpassInfo.h>
 #include <private/filament/Variant.h>
+#include <private/filament/ConstantInfo.h>
 
 #include <utils/compiler.h>
 #include <utils/Mutex.h>
 
 #include <atomic>
+#include <unordered_map>
```
poweifeng, line 36: not used?

filament/src/details/Material.cpp
bejado: Yup, good idea.

filament/src/details/Material.cpp
```diff
@@ -71,6 +74,7 @@ struct Material::BuilderDetails {
     size_t mSize = 0;
     MaterialParser* mMaterialParser = nullptr;
     bool mDefaultMaterial = false;
+    std::unordered_map<const char*, std::variant<int32_t, float, bool>> mConstantSpecializations;
```
pixelflinger, line 77: if we assume "not many" spec constants, we could just use a vector here and do a linear search. Also could we use `std::string_view` instead of `const char*`, this might save some calls to `strlen`.

filament/src/details/Material.cpp
```diff
@@ -91,6 +95,17 @@ Material::Builder& Material::Builder::package(const void* payload, size_t size)
     return *this;
 }
 
+template<typename T, typename>
+Material::Builder& Material::Builder::constant(const char* name, T value) {
+    ASSERT_PRECONDITION(name != nullptr, "name cannot be null");
+    mImpl->mConstantSpecializations[name] = value;
```
pixelflinger, line 101: I see why a map is convenient :-)

filament/src/details/Material.cpp
```diff
@@ -91,6 +95,17 @@ Material::Builder& Material::Builder::package(const void* payload, size_t size)
...
+    return *this;
+}
+
+template Material::Builder& Material::Builder::constant<int32_t>(const char* name, int32_t value);
+template Material::Builder& Material::Builder::constant<float>(const char* name, float value);
+template Material::Builder& Material::Builder::constant<bool>(const char* name, bool value);
```
pixelflinger, line 107: look at `MaterialInstance::setParameter`, these method should take a `const char* name, size_t len` to be compatible with `std::string_view`. It might be good to have a version that takes a `StringLiteral` also.

libs/filamat/src/MaterialBuilder.cpp
```diff
@@ -212,6 +213,53 @@ MaterialBuilder& MaterialBuilder::parameter(const char* name, SamplerType sample
     return parameter(name, samplerType, SamplerFormat::FLOAT, precision);
 }
 
+template<typename T, typename>
+MaterialBuilder& MaterialBuilder::constant(const char* name, ConstantType type, T defaultValue) {
```
pixelflinger, line 217: `MaterialBuilder.h` already make plenty of uses of the STL, so I would just use `std::string_view` for the `name` parameter.

pixelflinger (changes requested):
Only comment (sorry for not catching this before) is that everywhere (pretty-much) we pass a spec-constant name, we should use a std::string_view (for internal stuff) and `const char*, size_t` explicitly for public APIs.

filament/src/details/Material.cpp
bejado: I just realized something. I shouldn't be using a `const char*` here, because the user could release the string after calling `.constant`. We need to persist the string until the material is actually compiled. Likewise I can't use `string_view` either. I should be using `std::string` or `CString` instead.

libs/filamat/src/MaterialBuilder.cpp
bejado: I'm not sure it makes sense to use `string_view` here either. We need to make a copy of the string anyway, plus all the other methods on `MaterialBuilder` are using a `const char*`.

filament/src/details/Material.cpp
```diff
@@ -71,6 +74,7 @@ struct Material::BuilderDetails {
     size_t mSize = 0;
     MaterialParser* mMaterialParser = nullptr;
     bool mDefaultMaterial = false;
+    std::unordered_map<std::string, std::variant<int32_t, float, bool>> mConstantSpecializations;
```
pixelflinger, line 77: You could use a `CString` here which tends to generate less code.

I guess it was buggy in the previous version because it used `const char*` without a copy.

filament/src/details/Material.cpp
```diff
@@ -179,6 +194,45 @@ FMaterial::FMaterial(FEngine& engine, const Material::Builder& builder)
...
+
+    // Verify that all the constant specializations exist in the material and that their types match.
+    // The first specialization constants are defined internally by Filament.
+    // The subsequent constants are user-defined in the material.
+    mSpecializationConstants.reserve(constants.size() + CONFIG_MAX_RESERVED_SPEC_CONSTANTS);
+    mSpecializationConstants.push_back({0, (int)mEngine.getSupportedFeatureLevel()});
+    mSpecializationConstants.push_back({1, (int)CONFIG_MAX_INSTANCES});
+    for (const auto& [name, value] : builder->mConstantSpecializations) {
+        auto found = std::find_if(
+                constants.begin(), constants.end(), [name = name](const auto& constant) {
```
pixelflinger, line 209: why not use `[&name]` here?

pixelflinger: Also if `MaterialConstant` used a `std::string_view` you could use `operator==` instead of `strncmp`, which may be a bit cleaner.

pixelflinger (approved):
LGTM, just check my comments to see if you want to change something.

bejado:
> LGTM, just check my comments to see if you want to change something.

Good suggestions, I'll make adjustments in a future PR.

PR #6655, opened 2023-03-20
Add sun size and halo properties to glTF viewer controls.

romainguy:
This change also simplifies libviewer by removing duplicated code.


pixelflinger (commented):
aren't there some more changes needed for this to work with the remote ui?

I think you need to run beamsplitter, right?

romainguy:
beamsplitter is for when we change `filament/include/Options.h`, which isn't what I'm doing here.

pixelflinger (approved):

PR #6656, opened 2023-03-21
Fix a bug in matdbg where it could reorder the shaders in the package

pixelflinger:
This happened because the code iterated over the keys of a hashmap, which obviously were not guaranteed to be in the same order as those entries where added to the hashmap.

We fix this by adding a "visitor" to the materialchunk, so we can iterate through it in order and retried the info we need.


romainguy (approved):

libs/filaflat/src/MaterialChunk.cpp
```diff
@@ -175,5 +175,34 @@ bool MaterialChunk::getShader(ShaderContent& shaderContent,
...
 }
 
+void MaterialChunk::visitTextShaders(
+        utils::Invocable<void(uint8_t, Variant::type_t, uint8_t)>&& visitor) const {
+    assert_invariant(mMaterialTag != filamat::ChunkType::MaterialSpirv);
+
+    Unflattener unflattener{ mUnflattener }; // make a copy
+
+    // read() calls below cannot fail by construction, because we've already run through them
+    // in the constructore.
```
poweifeng, line 185: nit: constructor

pixelflinger, line 185:
```suggestion
    // in the constructor.
```

PR #6668, opened 2023-03-23
spirv-cross: use a reasonable float to string conversion

pixelflinger:
We only need 9 digits, not 34, to represent floats. This reduces material packages size.

### DIFF
--- a/third_party/spirv-cross/spirv_common.hpp
+++ b/third_party/spirv-cross/spirv_common.hpp
@@ -241,8 +241,29 @@ static inline void fixup_radix_point(char *str, char radix_point)
 	}
 }
 
+static inline std::string convert_to_smallest_string(float f, char locale_radix_point) {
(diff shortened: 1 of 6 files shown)


romainguy (approved):

poweifeng (approved):

third_party/spirv-cross/spirv_common.hpp
```diff
@@ -241,8 +241,29 @@ static inline void fixup_radix_point(char *str, char radix_point)
...
+    if (!strchr(buf, '.') && !strchr(buf, 'e'))
+        strcat(buf, ".0");
+
+    return { buf };
+}
+
 inline std::string convert_to_string(float t, char locale_radix_point)
 {
+    return convert_to_smallest_string(t, locale_radix_point);
+
```
poweifeng, line 266: this reads a little strange with the original code below.  Maybe we can add a comment for posterity?

poweifeng (approved):

PR #6677, opened 2023-03-25
vulkan: prefer discrete GPU over integrated

poweifeng:


pixelflinger (approved):

romainguy (approved):
At some point we may want to make this some kind of option when creating the engine.

PR #6686, opened 2023-03-27
better selection of opengl context version on desktop

### LINKED ISSUE #6682: Incorrect OpenGL version detected on desktop
**Describe the bug**

I was trying to figure out why I couldn't use feature level > 1 and I think I found some hints.
queryOpenGLVersion returns 4.1 if WGL_CONTEXT_MINOR_VERSION_ARB is set to 1 in PlatformWGL.cpp (wglCreateContextAttribs call), and 4.6 if I set it to 6.

I got the idea to try this after I found a similar issue in another code base : https://github.com/mpv-player/mpv/commit/0e76372e86d92982e17219fda6597f92ca67e7e8

/*
     * Nvidia drivers will not provide the highest supported version
     * when 320 core is requested. Instead, it just returns 3.2. This
     * would be bad, as we actually want compute shaders that require
     * 4.2, so we have to request a sufficiently high version. We use
     * 440 to maximise driver compatibility as we don't need anything
     * from newer versions.
     */

Looking at how they initialize GL, it seems that they try higher versions first :

https://github.com/mpv-player/mpv/blob/0d44ae319dfd5209eb0733caafdcbcf67d228cf3/video/out/opengl/context_glx.c

Maybe it's an idea for a proper fix ?

 - OS: Windows
 - GPU: NVIDIA GTX 1080
 - Backend: OpenGL

pixelflinger:
fixes #6682

### DIFF
--- a/filament/backend/src/opengl/platforms/PlatformWGL.cpp
+++ b/filament/backend/src/opengl/platforms/PlatformWGL.cpp
@@ -99,12 +99,6 @@ Driver* PlatformWGL::createDriver(void* const sharedGLContext,
         0, 0, 0
     };
 
-    int attribs[] = {
-        WGL_CONTEXT_MAJOR_VERSION_ARB, 4,
-        WGL_CONTEXT_MINOR_VERSION_ARB, 1,
-        0
-    };
-
     HGLRC tempContext = NULL;
 
     mHWnd = CreateWindowA("STATIC", "dummy", 0, 0, 0, 1, 1, NULL, NULL, NULL, NULL);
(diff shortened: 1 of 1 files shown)


romainguy (approved):

filament/backend/src/opengl/platforms/PlatformWGL.cpp
```diff
@@ -127,7 +121,22 @@ Driver* PlatformWGL::createDriver(void* const sharedGLContext,
...
+
+    int attribs[] = {
+            WGL_CONTEXT_MAJOR_VERSION_ARB, 4,
+            WGL_CONTEXT_MINOR_VERSION_ARB, 1,
+            0
+    };
+
+    // try all versions down, from GL 4.5 to 4.1
+    for (int minor = 5; minor >= 1; minor--) {
+        attribs[3] = minor;
```
poweifeng, line 133: attribs[1] here?

romainguy: Oh yeah, good catch!

pixelflinger: uh, I'm confused why? `attribs[1]` is for `WGL_CONTEXT_MAJOR_VERSION_ARB` and contains `4`.

poweifeng: haha my bad. Misread it.

romainguy: Oh yeah^2, you're right, I always forget the keys are in the array :)

PR #6689, opened 2023-03-28
Support the external image on macOS

daemyung:
Implement CocoaExternalImage.


filament/backend/src/opengl/platforms/CocoaExternalImage.h
```diff
@@ -0,0 +1,94 @@
...
+
+    CocoaExternalImage(const CVOpenGLTextureCacheRef textureCache,
+            const SharedGl& sharedGl) noexcept;
+    ~CocoaExternalImage() noexcept;
+
+    /**
+     * Set this external image to the passed-in CVPixelBuffer.
+     * Afterwards, calling glGetTexture returns the GL texture name backed by the CVPixelBuffer.
+     *
+     * Calling set with a BGRA image performs a blit image from BGRA to RGBA.
```
pixelflinger, line 54: Could you not use texture swizzling instead?

It also should say that BGRA is the only supported format -- apparently it doesn't support anything else below?

filament/backend/src/opengl/platforms/CocoaExternalImage.mm
```diff
@@ -0,0 +1,236 @@
...
+    // Release references to a previous external image, if we're holding any.
+    release();
+
+    if (!image) {
+        return false;
+    }
+
+    OSType formatType = CVPixelBufferGetPixelFormatType(image);
+    ASSERT_POSTCONDITION(formatType == kCVPixelFormatType_32BGRA,
+            "macOS external images must be in either 32BGRA format.");
```
pixelflinger, line 116: It looks like the ONLY format supported is 32BGRA, why does it say "either"?

filament/backend/src/opengl/platforms/CocoaExternalImage.mm
```diff
@@ -0,0 +1,236 @@
...
+    CVOpenGLTextureRef texture = nullptr;
+    UTILS_UNUSED_IN_RELEASE CVReturn success =
+            CVOpenGLTextureCacheCreateTextureFromImage(kCFAllocatorDefault,
+            mTextureCache, image, nil, &texture);
+    assert_invariant(success == kCVReturnSuccess);
+
+    return texture;
+}
+
+GLuint CocoaExternalImage::encodeBlitTexture(GLuint srcTexture, size_t width, size_t height) noexcept {
```
pixelflinger, line 174: if I understand correctly, this is used mainly to convert from GL_TEXTURE_RECTANGLE target to GL_TEXTURE_2D, right? it also does the swizzling from BGRA. If so, can you add a comment about that?

filament/backend/src/opengl/platforms/PlatformCocoaGL.mm
```diff
@@ -256,6 +269,9 @@
 
 void PlatformCocoaGL::commit(Platform::SwapChain* swapChain) noexcept {
     [pImpl->mGLContext flushBuffer];
+
+    // This needs to be done periodically.
+    CVOpenGLTextureCacheFlush(pImpl->mTextureCache, 0);
```
pixelflinger, line 274: again I don't want to do this if there is no need. It should only be done if there are external images active.
also can you comment what this does and why this is needed periodically?

filament/backend/src/opengl/platforms/PlatformCocoaGL.mm
```diff
@@ -266,6 +282,41 @@
...
+    auto* p = static_cast<CocoaExternalImage*>(texture);
+    delete p;
+}
+
+void PlatformCocoaGL::retainExternalImage(void* externalImage) noexcept {
+    // Take ownership of the passed in buffer. It will be released the next time
+    // setExternalImage is called, or when the texture is destroyed.
+    // CVPixelBufferRef pixelBuffer = (CVPixelBufferRef) externalImage;
+    // CVPixelBufferRetain(pixelBuffer);
+}
```
pixelflinger, line 305: why is this comment out?

filament/backend/src/opengl/platforms/PlatformCocoaGL.mm
```diff
@@ -266,6 +282,41 @@
...
+
+bool PlatformCocoaGL::setExternalImage(void* externalImage, ExternalTexture* texture) noexcept {
```
pixelflinger, line 307: I think you need to update the documentation in `Texture` to explain what `externalImage` is.

pixelflinger (changes requested):

filament/backend/src/opengl/platforms/PlatformCocoaGL.mm
daemyung: Maybe it's my mistake.

filament/backend/src/opengl/platforms/CocoaExternalImage.h
daemyung: Maybe it can support other formats. But I don't have other use cases now. We can implement other cases later. I will delete this comments.

filament/backend/src/opengl/platforms/CocoaExternalImage.mm
daemyung: It's my mistake. I copy source code from CocoaTouchExternalImage.mm and modify for macOS. "either" is coming along from copy and paste.

filament/backend/src/opengl/platforms/PlatformCocoaGL.mm
daemyung: It's already explained at OpenGLPlatform.h

filament/backend/src/opengl/platforms/PlatformCocoaGL.mm
daemyung: This code is coming from [here](https://github.com/google/filament/blob/d00223332ea71322bb2eb6a0e79100da79bd83ff/filament/backend/src/opengl/platforms/PlatformCocoaTouchGL.mm#L172). Removing this flush is fine on my case. Please let me know do you want to remove this then I will remove it.

filament/backend/src/opengl/platforms/CocoaExternalImage.mm
daemyung: Instead of adding comment I have changed a function name.

filament/backend/src/opengl/platforms/PlatformCocoaGL.mm
```diff
@@ -159,6 +162,14 @@
...
     int result = bluegl::bind();
     ASSERT_POSTCONDITION(!result, "Unable to load OpenGL entry points.");
+
+    UTILS_UNUSED_IN_RELEASE CVReturn success = CVOpenGLTextureCacheCreate(kCFAllocatorDefault, nullptr,
+            [pImpl->mGLContext CGLContextObj], [pImpl->mGLContext.pixelFormat CGLPixelFormatObj], nullptr,
+            &pImpl->mTextureCache);
+    assert_invariant(success == kCVReturnSuccess);
+
+    pImpl->mExternalImageSharedGl = new CocoaExternalImage::SharedGl();
+
```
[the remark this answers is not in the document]
daemyung: It's updated. This pattern is coming from [here](https://github.com/google/filament/blob/d00223332ea71322bb2eb6a0e79100da79bd83ff/filament/backend/src/opengl/platforms/PlatformCocoaTouchGL.mm#L91).

filament/backend/src/opengl/platforms/CocoaExternalImage.mm
```diff
@@ -0,0 +1,236 @@
...
+    mTexture = createTextureFromImage(image);
+    mRgbaTexture = encodeCopyRectangleToTexture2D(CVOpenGLTextureGetName(mTexture),
+            CVPixelBufferGetWidth(image), CVPixelBufferGetHeight(image));
+    CHECK_GL_ERROR(utils::slog.e)
+
+    return true;
+}
+
+GLuint CocoaExternalImage::getGlTexture() const noexcept {
+    if (mEncodedToRgba) {
```
bejado, line 132: Do we even need the `mEncodedToRgba` flag? Seems that after `set` is called, `mRgbaTexture` will be valid. Before set is called, neither `mRgbaTexture` nor `mTexture` are valid, so maybe we should return 0 or throw an exception in that case.

daemyung: Yes, that makes sense. Your feedback will be taken into account.

bejado (approved):

daemyung:
@pixelflinger @bejado Could you merge it?

pixelflinger (approved):

PR #6694, opened 2023-03-29
fix `-ffast-math` and `-fno-finite-math-only`

pixelflinger:
- remove `-ffast-math` in place where it didn't seem too useful
- added `-fno-finite-math-only` everywhere we use `-ffast-math` so that `isinf`/`isnan` work.


poweifeng (approved):

libs/geometry/CMakeLists.txt
```diff
@@ -42,9 +42,7 @@ set_target_properties(${TARGET} PROPERTIES FOLDER Libs)
 # Compiler flags
 # ==================================================================================================
 if (MSVC)
-    target_compile_options(${TARGET} PRIVATE $<$<CONFIG:Release>:/fp:fast>)
 else()
-    target_compile_options(${TARGET} PRIVATE $<$<CONFIG:Release>:-ffast-math>)
```
romainguy, line 47: Might be helpful here for geometry processing (computing tangents etc.)

romainguy (approved):

PR #6697, opened 2023-03-31
TangentSpaceMesh fixes

bejado:
This PR fixes a bug in the Lengyel method to respect the normal stride.

When using the Lengyel method, there's another bug where `getTriangles(ushort3*)` crashes. Even though 16 bit indices were provided, `is16` is false. I think this is due to a problem with `InternalArray::operator bool`. If the array has a valid "borrowed" pointer, then we should return true.

Note that the Lengyel method doesn't actually remesh, but `getTriangles` should still work.

### DIFF
--- a/libs/geometry/tests/test_tangent_space_mesh.cpp
+++ b/libs/geometry/tests/test_tangent_space_mesh.cpp
@@ -347,6 +347,11 @@ TEST_F(TangentSpaceMeshTest, Lengyel) {
     size_t const vertexCount = mesh->getVertexCount();
     std::vector<quatf> quats(vertexCount);
     mesh->getQuats(quats.data());
+
+    ASSERT_EQ(mesh->getTriangleCount(), CUBE_TRIANGLES.size());
+    std::vector<ushort3> triangles(mesh->getTriangleCount());
+    mesh->getTriangles(triangles.data());
+
     for (size_t i = 0; i < vertexCount; ++i) {
         float3 const n = quats[i] * NORMAL_AXIS;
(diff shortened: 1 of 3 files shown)


libs/geometry/tests/test_tangent_space_mesh.cpp
```diff
@@ -347,6 +347,11 @@ TEST_F(TangentSpaceMeshTest, Lengyel) {
     size_t const vertexCount = mesh->getVertexCount();
     std::vector<quatf> quats(vertexCount);
     mesh->getQuats(quats.data());
+
+    ASSERT_EQ(mesh->getTriangleCount(), CUBE_TRIANGLES.size());
+    std::vector<ushort3> triangles(mesh->getTriangleCount());
+    mesh->getTriangles(triangles.data());
```
poweifeng, line 353: If the mesh wasn't remeshed, then the caller shouldn't have a reason to call getTriangle, correct? 
Maybe we'd throw an exception in that case actually.

poweifeng: actually, I misread the issue.  You are correct!

poweifeng (approved):

pixelflinger (approved):

libs/geometry/tests/test_tangent_space_mesh.cpp
bejado: Yup, there's no real reason to call `getTriangle`, it felt wrong to just crash though.

PR #6699, opened 2023-03-31
skinning now works when the model is far from the origin

pixelflinger:
We are doing skinning computations effectively in world space, so
some of the math must be done in high-precision.

### DIFF
--- a/samples/gltf_viewer.cpp
+++ b/samples/gltf_viewer.cpp
@@ -84,6 +84,7 @@ struct App {
     ViewerGui* viewer;
     Config config;
     Camera* mainCamera;
+    Entity rootTransformEntity;
 
     AssetLoader* assetLoader;
     FilamentAsset* asset = nullptr;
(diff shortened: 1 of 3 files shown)


romainguy (approved):

samples/gltf_viewer.cpp
```diff
@@ -907,6 +915,13 @@ int main(int argc, char** argv) {
         Skybox* skybox = scene->getSkybox();
         applySettings(engine, app.viewer->getSettings().viewer, &camera, skybox, renderer);
 
+        // technically we don't need to do this each frame
+        auto& tcm = engine->getTransformManager();
+        TransformManager::Instance const& root = tcm.getInstance(app.rootTransformEntity);
+        tcm.setParent(tcm.getInstance(camera.getEntity()), root);
```
bejado, line 921: Does this work if the user selects a glTF camera?

bejado (approved):

samples/gltf_viewer.cpp
pixelflinger: yeah it works.

PR #6708, opened 2023-04-04
Update libz and libpng to fix compilation warnings

romainguy:


romainguy:
Our libpng compiles against our libz and libpng does a version check on libz, which is why both are updated together here.

poweifeng (approved):

poweifeng (approved):

PR #6717, opened 2023-04-05
vulkan: Fix validation errors

poweifeng:
- Depth attachment layout has generated a lot of error due to it being read-only. But the store-ops for the attachment during the renderpass are all write ops. We set the depth attachment layout as VK_IMAGE_LAYOUT_DEPTH_STENCIL_ATTACHMENT_OPTIMAL
 - Enable extra blitting step for SSAO because of the above layout conundrum.
 - Remove `assert_invariant(utils::popcount(sampleCount) == 1);` from `reduceSampleCount`. This assert fails when enabling the duplicate pass for SSAO.


filament/backend/src/vulkan/VulkanDriver.cpp
```diff
@@ -652,7 +652,26 @@ bool VulkanDriver::isWorkaroundNeeded(Workaround workaround) {
...
             return deviceProperties.vendorID == 0x5143; // Qualcomm
         case Workaround::ALLOW_READ_ONLY_ANCILLARY_FEEDBACK_LOOP:
-            return true;
+            // Supporting depth attachment as both sampler and attachment is only possible if we set
+            // the depth attachment as read-only (e.g. during SSAO pass), however note that the
+            // store-ops for attachments wrt VkRenderPass only has VK_ATTACHMENT_STORE_OP_DONT_CARE
+            // and VK_ATTACHMENT_STORE_OP_STORE for versions below 1.3. Only at 1.3 and above do we
+            // have a true read-only choice VK_ATTACHMENT_STORE_OP_NONE. That means for < 1.3, we
+            // will trigger a validation sync error if we use the depth attachment also as a
+            // sampler. See full error below:
```
pixelflinger, line 661: Amazing...

pixelflinger (changes requested):
Could you separate "Index buffers did not have a pipeline barriers after loading them." into a different PR (as it seems unrelated to the depth issue).

pixelflinger (approved):

PR #6719, opened 2023-04-06
improvements to material code and material creation

pixelflinger:
- `ProgramBuilder` is a relic, rename to `Program`
- use the enum instead of ints everywhere for `ShaderModel` and `ShaderStage`
- Improve how we cache shared shaders:

Some shaders can be shared across all materials (e.g. the depth 
shaders). We use the filament default material as the "source" of
the cache, but until now we relied on an a priori knowledge of which
variants were present in the default material.

With this change, we now query once the list of variants (of interest)
in the default material and reuse that list for caching these variants
later.

This is better because the cached variants are now entirely driven by
the default material (which they depend on anyways). This is also faster
because we don't need to query which variant we need each time we create
a material.

### DIFF
--- a/filament/src/details/Material.cpp
+++ b/filament/src/details/Material.cpp
@@ -281,18 +281,28 @@ FMaterial::FMaterial(FEngine& engine, const Material::Builder& builder)
     parser->hasCustomDepthShader(&mHasCustomDepthShader);
     mIsDefaultMaterial = builder->mDefaultMaterial;
 
-    // pre-cache the shared variants -- these variants are shared with the default material.
-    // (note: the default material is unlit, so only unlit variants can be shared)
-    if (UTILS_UNLIKELY(!mIsDefaultMaterial && !mHasCustomDepthShader)) {
-        FMaterial const* const pMaterial = engine.getDefaultMaterial();
-        auto& cachedPrograms = mCachedPrograms;
-        for (Variant::type_t k = 0, n = VARIANT_COUNT; k < n; ++k) {
-            const Variant variant(k);
-            if (Variant::isValidDepthVariant(variant) &&
-                (Variant::filterVariant(variant, false) == variant)) {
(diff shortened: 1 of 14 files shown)


romainguy (approved):

filament/src/details/Material.cpp
```diff
@@ -358,6 +368,23 @@ BufferInterfaceBlock::FieldInfo const* FMaterial::reflect(
...
+    filaflat::MaterialChunk const& materialChunk{ mMaterialParser->getMaterialChunk() };
+
+    variants.reserve(materialChunk.getShaderCount());
+    materialChunk.visitShaders([&variants](
+            ShaderModel model, Variant variant, ShaderStage stage) {
+        variants.push_back(variant);
+    });
+    std::sort(variants.begin(), variants.end(),
+            [](Variant lhs, Variant rhs) { return lhs.key < rhs.key; });
+    std::unique(variants.begin(), variants.end());
```
bejado, line 383: Do we need to resize `variants` after calling `std::unique`?

filament/src/details/Material.cpp
```diff
@@ -358,6 +368,23 @@ BufferInterfaceBlock::FieldInfo const* FMaterial::reflect(
     return mUniformInterfaceBlock.getFieldInfo(name);
 }
 
+utils::FixedCapacityVector<Variant> FMaterial::getVariants() const noexcept {
```
bejado, line 371: Are we using this anywhere?

filament/src/details/Material.cpp
pixelflinger: uh. yes, I forgot to upload it. thanks!

pixelflinger: shoot I forgot to remove that code. thanks!

poweifeng (approved):

PR #6721, opened 2023-04-07
Improve size optimizations when compiling material

romainguy:
This changes the behavior of the size optimizer in matc (-S), but only for GLSL and MSL. With this change we gain a ~65% size reduction on a lit material compiled for OpenGL. To get those gains we generate extra SPIRV debug information to preserve variable names and better utilize the line dictionary. Unfortunately this break the SPIRV optimizer so we skip it and instead rely on a simple DCE pass provided by glslang. We also enhance the whitespace removal pass of the GLSL minifier to move lone { and } to the previous line, which avoids generating an extra index in each shader variant. Each index being at least as big as the character itself, this is quite wasteful.

When generating SPIRV for Vulkan, we rely on spirv-opt for size optimizations as before.

### DIFF
--- a/NEW_RELEASE_NOTES.md
+++ b/NEW_RELEASE_NOTES.md
@@ -9,4 +9,4 @@ appropriate header in [RELEASE_NOTES.md](./RELEASE_NOTES.md).
 ## Release notes for next branch cut
 
 - materials: improved size reduction of OpenGL/Metal shaders by ~65% when compiling materials with
-             size optimizations (`matc -S`)
+             size optimizations (`matc -S`) [⚠️ **Recompile Materials**]
--- a/NEW_RELEASE_NOTES.md
+++ b/NEW_RELEASE_NOTES.md
@@ -8,3 +8,5 @@ appropriate header in [RELEASE_NOTES.md](./RELEASE_NOTES.md).
 
 ## Release notes for next branch cut
 
+- materials: improved size reduction of OpenGL/Metal shaders by ~65% when compiling materials with
+             size optimizations (`matc -S`)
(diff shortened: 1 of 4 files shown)


libs/filamat/src/ShaderMinifier.cpp
```diff
@@ -137,7 +137,7 @@ namespace {
  * - Remove leading white spaces at the beginning of each line
  * - Remove empty lines
  */
-std::string ShaderMinifier::removeWhitespace(const std::string& s) const {
+std::string ShaderMinifier::removeWhitespace(const std::string& s, bool mergeBraces) const {
```
pixelflinger, line 140: why not doing this regardless of the optimization mode?

NEW_RELEASE_NOTES.md
```diff
@@ -8,3 +8,5 @@ appropriate header in [RELEASE_NOTES.md](./RELEASE_NOTES.md).
 
 ## Release notes for next branch cut
 
+- materials: improved size reduction of OpenGL/Metal shaders by ~65% when compiling materials with
+             size optimizations (`matc -S`)
```
pixelflinger, line 12: kind of obvious, but you could add a `[recompile materials]`

pixelflinger (approved):

libs/filamat/src/ShaderMinifier.cpp
romainguy: Because it actually makes things worse for the other cases.

NEW_RELEASE_NOTES.md
romainguy: Sure

PR #6750, opened 2023-04-20
Fix Android samples so they handle surface resizes properly

pixelflinger:


android/samples/sample-hello-camera/src/main/java/com/google/android/filament/hellocam/MainActivity.kt
```diff
@@ -401,9 +401,20 @@ class MainActivity : Activity(), ActivityCompat.OnRequestPermissionsResultCallba
             camera.setProjection(45.0, aspect, 0.1, 20.0, Camera.Fov.VERTICAL)
 
             view.viewport = Viewport(0, 0, width, height)
+
+            synchronizePendingFrames(engine)
         }
     }
 
+    private fun synchronizePendingFrames(engine: Engine) {
```
romainguy, line 409: Why don't we do this automatically in UiHelper after we call the app's onResized()?

pixelflinger: `UiHelper` doesn't have access to the `Engine`. We kept `UiHelper` independent from filament.

romainguy: Ah right. Maybe we could provide synchronizePendingFrames(Engine) as an API somewhere then?

pixelflinger: I thought of that, but this would add a dependency on `Engine.java` in `UiHelper` and I know this will break google3.  We could provide it in another library that all our samples use, but the goal of the samples was to be standalone. I couldn't find a better way.

romainguy: Why would that break google3? UiHelper is part of the Filament JAR which contains the Engine API. The new method doesn't have to be in UiHelper but it could be a in a new class.

romainguy: No because we can put that class where we have UiHelper, and the other helpers: https://github.com/google/filament/tree/main/android/filament-android/src/main/java/com/google/android/filament/android

pixelflinger: Ah I think that would work and shouldn't impact google3 since it's a completely new class. I don't think they're building the samples (@bejado ?) so we wouldn't have to update the build rules there.

PR #6762, opened 2023-04-24
Add support for supplying instance local transforms

bejado:
Clients can now supply a local transform for each GPU instance. Still working through some TODOs, putting this up for early feedback.

### DIFF
--- a/filament/include/filament/InstanceBuffer.h
+++ b/filament/include/filament/InstanceBuffer.h
@@ -63,7 +63,7 @@ public:
          * @param localTransforms an array of math::mat4f with length instanceCount, must remain
          *                        valid until after build() is called
          */
-        Builder& localTransforms(math::mat4f const* localTransforms);
+        Builder& localTransforms(math::mat4f const* localTransforms) noexcept;
 
         /**
          * Creates the InstanceBuffer object and returns a pointer to it.
--- a/filament/include/filament/InstanceBuffer.h
+++ b/filament/include/filament/InstanceBuffer.h
@@ -1,0 +1,90 @@
+/*
+* Copyright (C) 2023 The Android Open Source Project
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*
+*      http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+
+#ifndef TNT_FILAMENT_INSTANCEBUFFER_H
+#define TNT_FILAMENT_INSTANCEBUFFER_H
+
+#include <filament/FilamentAPI.h>
+
+#include <filament/Engine.h>
+
+#include <math/mathfwd.h>
+
+namespace filament {
+
+/**
+ * InstanceBuffer holds draw (GPU) instance transforms. These can be provided to a renderable to
+ * "offset" each draw instance.
+ *
+ * @see RenderableManager::Builder::instances(size_t, InstanceBuffer*)
+ */
+class UTILS_PUBLIC InstanceBuffer : public FilamentAPI {
+    struct BuilderDetails;
+
+public:
+    class Builder : public BuilderBase<BuilderDetails> {
+        friend struct BuilderDetails;
+
+    public:
+
+        /**
+         * @param instanceCount the number of instances this InstanceBuffer will support, must be
+         *                      >= 1 and <= \c CONFIG_MAX_INSTANCES
+         */
+        Builder(size_t instanceCount) noexcept;
+
+        Builder(Builder const& rhs) noexcept;
+        Builder(Builder&& rhs) noexcept;
+        ~Builder() noexcept;
+        Builder& operator=(Builder const& rhs) noexcept;
+        Builder& operator=(Builder&& rhs) noexcept;
+
+        /**
+         * Provide an initial local transform for each instance. The local transform is
+         * relative to the renderable's transform, so adjusting the renderable's transform will
+         * affect all instances.
+         *
+         * The array of math::mat4f must have length instanceCount, provided when constructing this
+         * Builder.
+         *
+         * @param localTransforms an array of math::mat4f with length instanceCount, must remain
+         *                        valid until after build() is called
+         */
+        Builder& localTransforms(math::mat4f const* localTransforms);
+
+        /**
+         * Creates the InstanceBuffer object and returns a pointer to it.
+         */
+        InstanceBuffer* build(Engine& engine);
+
+    private:
+        friend class FInstanceBuffer;
+    };
+
+    /**
+     * Sets the local transform for each instance. The local transform is relative to the
+     * renderable's transform.
+     *
+     * @param localTransforms an array of math::mat4f with length count, need not outlive this call
+     * @param count the number of local transforms
+     * @param offset index of the first instance to set local transforms
+     */
+    void setLocalTransforms(math::mat4f const* localTransforms, size_t count, size_t offset = 0);
+};
+
+} // namespace filament
+
+#endif //TNT_FILAMENT_INSTANCEBUFFER_H
--- a/filament/include/filament/InstanceBuffer.h
+++ b/filament/include/filament/InstanceBuffer.h
@@ -53,9 +53,10 @@ public:
         Builder& operator=(Builder&& rhs) noexcept;
 
         /**
-         * Provide an initial local transform for each instance. The local transform is
-         * relative to the renderable's transform, so adjusting the renderable's transform will
-         * affect all instances.
+         * Provide an initial local transform for each instance. Each local transform is relative to
+         * the transform of the associated renderable. This forms a parent-child relationship
+         * between the renderable and its instances, so adjusting the renderable's transform will
+-        * affect all instances.
          *
          * The array of math::mat4f must have length instanceCount, provided when constructing this
          * Builder.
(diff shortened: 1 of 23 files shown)


filament/src/components/RenderableManager.cpp
```diff
@@ -301,10 +304,49 @@ RenderableManager::Builder::Result RenderableManager::Builder::build(Engine& eng
...
 
 RenderableManager::Builder& RenderableManager::Builder::instances(size_t instanceCount) noexcept {
+    ASSERT_PRECONDITION(mImpl->mInstanceTransforms == nullptr && mImpl->mInstanceBuffer == nullptr,
+            "cannot call instances(size_t) override, instance local transforms already provided");
     mImpl->mInstanceCount = clamp((unsigned int)instanceCount, 1u, 32767u);
     return *this;
 }
 
+RenderableManager::Builder& RenderableManager::Builder::instances(size_t instanceCount,
+        math::mat4f const* localTransforms) noexcept {
```
pixelflinger, line 314: not sure if we should have both APIs. I think I prefer we be opinionated and pick one.

filament/src/components/RenderableManager.cpp
```diff
@@ -301,10 +304,49 @@ RenderableManager::Builder::Result RenderableManager::Builder::build(Engine& eng
...
+    ASSERT_PRECONDITION(instanceCount >= 1, "instanceCount must be >= 1.");
```
pixelflinger, line 315: typically, we do all the validation in `build`, not in the individual builder methods.

the idea is that there is a single method that "is heavy" and "can fail".

filament/src/components/RenderableManager.h
```diff
@@ -201,15 +206,26 @@ class FRenderableManager : public RenderableManager {
     };
     static_assert(sizeof(MorphWeights) == 8);
 
+    struct Instances {
+        union {
+            FInstanceBuffer* buffer;
+            uint64_t padding;          // ensures the pointer is 64 bits on all archs
+        };
+        uint16_t count;
+        bool shouldDestroyBuffer; // if true, buffer was created by Filament and should be destroyed
```
pixelflinger, line 215: I think I prefer "ownBuffer" instead of "shouldDestroy"

pixelflinger: But this would go away if we only supported that API with a buffer.

filament/src/components/RenderableManager.h
```diff
@@ -422,6 +444,20 @@ utils::Slice<FRenderableManager::MorphTargets>& FRenderableManager::getMorphTarg
     return mManager[instance].morphTargets;
 }
 
+void FRenderableManager::setInstanceTransforms(Instance instance,
+        math::mat4f const* localTransforms, size_t count, size_t offset) noexcept {
+    Instances& instances = mManager[instance].instances;
+    ASSERT_PRECONDITION(instances.buffer != nullptr,
```
pixelflinger, line 450: if method has an "ASSERT_*" it cannot be `noexcept`.

filament/src/details/Scene.cpp
```diff
@@ -347,6 +351,12 @@ void FScene::prepareVisibleRenderables(Range<uint32_t> visibleRenderables) noexc
         uboData.userData = sceneData.elementAt<USER_DATA>(i);
 
         mHasContactShadows = mHasContactShadows || visibility.screenSpaceContactShadows;
+
+        // Prepare the InstanceBuffer for this Renderable.
+        // TODO: maybe do this inside updateUBOs? And avoid updating the actual UBO?
+        if (UTILS_UNLIKELY(instanceBuffer)) {
+            instanceBuffer->prepare(mEngine, model, uboData);
+        }
```
pixelflinger, line 359: this should be done after culling otherwise we're doing thing for everything in the scene, even if it's not visible.

libs/math/include/math/mat3.h
```diff
@@ -289,6 +289,14 @@ class MATH_EMPTY_BASES TMat33 :
         return matrix::cof(m);
     }
 
+    /**
+     * Pre-scale a matrix m by the inverse of the largest scale factor to avoid large post-transform
+     * magnitudes in the shader. This is useful for normal transformations, to avoid large
+     * post-transform magnitudes in the shader, especially in the fragment shader, where we use
+     * medium precision.
+     */
+    static constexpr TMat33<T> prescaleForNormals(const TMat33& m) noexcept;
```
pixelflinger, line 298: I don't think this should be on `mat3`. It can be a helper method somewhere though.

pixelflinger (changes requested):

filament/src/components/RenderableManager.cpp
bejado: The main reason why I think the "manual" instancing case is useful is that it allows way more instances. Maybe we should go ahead and implement the "instanced chopping" we discussed to allow a large number of instances even with the hybrid version.

filament/src/details/Scene.cpp
bejado: Doesn't `prepareVisibleRenderables` happen after culling already?

pixelflinger: oh yes it does, sorry I misread the code

filament/src/details/InstanceBuffer.cpp
```diff
@@ -0,0 +1,127 @@
...
+    // TODO: keep track of dirtiness. If the rootTransform and localTransforms have not changed,
+    // there's no need to update the buffer.
+
+    // TODO: allocate this staging buffer from a pool.
+    uint32_t stagingBufferSize = sizeof(PerRenderableUib);
+    PerRenderableData* stagingBuffer = (PerRenderableData*)::malloc(stagingBufferSize);
+    // TODO: consider using JobSystem to parallelize this.
+    for (size_t i = 0; i < mInstanceCount; i++) {
+        stagingBuffer[i] = ubo;
+        math::mat4f model = rootTransform * mLocalTransforms[i];
```
pixelflinger, line 107: Are you sure it's not `mLocalTransforms[i] * rootTransform` ? Shouldn't we apply the rootTransform first, and then the local transform?

filament/src/details/InstanceBuffer.cpp
```diff
@@ -0,0 +1,127 @@
...
+    DriverApi& driver = engine.getDriverApi();
+
+    // TODO: keep track of dirtiness. If the rootTransform and localTransforms have not changed,
+    // there's no need to update the buffer.
+
+    // TODO: allocate this staging buffer from a pool.
+    uint32_t stagingBufferSize = sizeof(PerRenderableUib);
+    PerRenderableData* stagingBuffer = (PerRenderableData*)::malloc(stagingBufferSize);
+    // TODO: consider using JobSystem to parallelize this.
+    for (size_t i = 0; i < mInstanceCount; i++) {
```
pixelflinger, line 105:
```suggestion
    for (size_t i = 0, c = mInstanceCount; i < c; i++) {
```

filament/src/details/Scene.cpp
```diff
@@ -347,6 +351,12 @@ void FScene::prepareVisibleRenderables(Range<uint32_t> visibleRenderables) noexc
         uboData.userData = sceneData.elementAt<USER_DATA>(i);
 
         mHasContactShadows = mHasContactShadows || visibility.screenSpaceContactShadows;
+
+        // Prepare the InstanceBuffer for this Renderable.
+        // TODO: maybe do this inside updateUBOs? And avoid updating the actual UBO?
```
pixelflinger, line 356: if possible this would be more consistant

filament/include/filament/InstanceBuffer.h
```diff
@@ -0,0 +1,90 @@
...
+
+    private:
+        friend class FInstanceBuffer;
+    };
+
+    /**
+     * Sets the local transform for each instance. The local transform is relative to the
+     * renderable's transform.
+     *
+     * @param localTransforms an array of math::mat4f with length count, need not outlive this call
```
pixelflinger, line 81: maybe we need a better explanation of what that transform is? i.e. it transforms from where to where?

filament/src/details/InstanceBuffer.cpp
bejado: I'm fairly certain.

Say the user wants to create a forrest of instanced trees, each with a random translation and scale. Each tree’s “local transform” could be computed like:

```
mat4f localTransform =
    mat4f::translation(float3{rand(), 0.0f, rand()}) *          // call this T
    mat4f::rotation(rand(), float3{0.0f, 1.0f, 0.0f});          // call this R
```

To move the forest as a whole, the user can simply adjust the rootTransform (renderable's transform). Assume they set it to a translation T'.

If the computation was `mLocalTransforms[i] * rootTransform`, then we'd end up with:

```
worldSpace = T * R * T' * vertex
```

That isn't correct, or at least very unintuitive. This is also complicated by the fact that T' can contain the world offset transformation.

Instead, we want `rootTransform * mLocalTransforms[i]`:

```
worldSpace = T' * T * R * vertex
```

pixelflinger: you're right, I got myself confused.

filament/src/details/Engine.cpp
```diff
@@ -703,6 +703,10 @@ FMorphTargetBuffer* FEngine::createMorphTargetBuffer(const MorphTargetBuffer::Bu
     return create(mMorphTargetBuffers, builder);
 }
 
+FInstanceBuffer* FEngine::createInstanceBuffer(const InstanceBuffer::Builder& builder) noexcept {
+    return create(mInstanceBuffers, builder);
+}
```
pixelflinger, line 708: I think you're missing:

```
    cleanupResourceList(std::move(mInstanceBuffers));
```

in `Engine::shutdown`

filament/src/RenderPass.h
```diff
@@ -234,21 +234,22 @@ class RenderPass {
         backend::Handle<backend::HwBufferObject> skinningHandle;        // 4 bytes
         backend::Handle<backend::HwBufferObject> morphWeightBuffer;     // 4 bytes
         backend::Handle<backend::HwSamplerGroup> morphTargetBuffer;     // 4 bytes
+        backend::Handle<backend::HwBufferObject> instanceBufferHandle;  // 4 bytes
```
pixelflinger, line 237: I'm starting to think that we might want to have an indirection here to store all three `Handle` related to skinning/morphing (it's going to be four soon).

pixelflinger (changes requested):

filament/src/RenderPass.h
bejado: Yeah that would be nice. Let's do it in a separate PR?

filament/include/filament/InstanceBuffer.h
bejado: I rephrased this for clarification, let me know if you have any suggestions.

filament/src/RenderPass.h
pixelflinger: definitely, and in fact let's do this after the >4 bones PR.

PR #6769, opened 2023-04-26
cleanup PlatformEGL a bit.

pixelflinger:
- add a helper Config class to more easily manage EGL's attribute pairs.

- consolidate config selection code

### DIFF
--- a/filament/backend/src/opengl/platforms/PlatformEGL.cpp
+++ b/filament/backend/src/opengl/platforms/PlatformEGL.cpp
@@ -45,25 +45,31 @@ using namespace glext;
 // ---------------------------------------------------------------------------------------------
 
 void PlatformEGL::logEglError(const char* name) noexcept {
-    const char* err;
-    switch (eglGetError()) {
-        case EGL_NOT_INITIALIZED:       err = "EGL_NOT_INITIALIZED";    break;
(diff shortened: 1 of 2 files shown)


filament/backend/src/opengl/platforms/PlatformEGL.cpp
```diff
@@ -110,36 +116,24 @@ Driver* PlatformEGL::createDriver(void* sharedContext, const Platform::DriverCon
...
-    };
-
-    EGLint contextAttribs[] = {
-            EGL_CONTEXT_CLIENT_VERSION, 3,
-            EGL_NONE, EGL_NONE, // reserved for EGL_CONTEXT_OPENGL_NO_ERROR_KHR below
-            EGL_NONE
-    };
-
-    EGLint pbufferAttribs[] = {
+    EGLint const pbufferAttribs[] = {
```
poweifeng, line 119: why not make this a Config for consistency?

poweifeng (approved):

filament/backend/src/opengl/platforms/PlatformEGL.cpp
pixelflinger: it didn't need it because it's `const`. trying to avoid some heap allocations.

poweifeng: ack

PR #6770, opened 2023-04-26
basic ES2 support for filament

pixelflinger:
Limitations include (but are not limited to):
- no post processing
- no lighting (only `unlit` materials are supported)
- no skinning or morphing
- no instancing
- all ES2 shader limitation apply
- all ES2 texture formats apply
- materials must be written specifically for ES2 and work only on an ES2 context

Fog is supported. 

Some extensions are mandatory:
OES_depth_texture
OES_depth24
OES_packed_depth_stencil
OES_rgb8_rgba8
OES_standard_derivatives
OES_texture_npot
OES_vertex_array_object

### DIFF
--- a/filament/backend/include/backend/Program.h
+++ b/filament/backend/include/backend/Program.h
@@ -90,6 +90,10 @@ public:
     // the program everything it needs to know about the uniforms at a given binding
     Program& uniforms(uint32_t index, UniformInfo const& uniforms) noexcept;
 
+    // Note: This is only needed for GLES2.0.
+    Program& attributes(
+            utils::FixedCapacityVector<std::pair<utils::CString, uint8_t>> attributes) noexcept;
+
     // sets the 'bindingPoint' sampler group descriptor for this program.
     // 'samplers' can be destroyed after this call.
     // This effectively associates a set of (BindingPoints, index) to a texture unit in the shader.
@@ -118,6 +122,9 @@ public:
     auto const& getBindingUniformInfo() const { return mBindingUniformInfo; }
     auto& getBindingUniformInfo() { return mBindingUniformInfo; }
 
+    auto const& getAttributes() const { return mAttributes; }
+    auto& getAttributes() { return mAttributes; }
+
     utils::CString const& getName() const noexcept { return mName; }
     utils::CString& getName() noexcept { return mName; }
 
@@ -137,6 +144,7 @@ private:
     utils::CString mName;
     utils::Invocable<utils::io::ostream&(utils::io::ostream& out)> mLogger;
     utils::FixedCapacityVector<SpecializationConstant> mSpecializationConstants;
+    utils::FixedCapacityVector<std::pair<utils::CString, uint8_t>> mAttributes;
     std::array<UniformInfo, Program::UNIFORM_BINDING_COUNT> mBindingUniformInfo;
 };
 
--- a/filament/backend/include/backend/Program.h
+++ b/filament/backend/include/backend/Program.h
@@ -48,7 +48,15 @@ public:
         ShaderStageFlags stageFlags = ShaderStageFlags::ALL_SHADER_STAGE_FLAGS;
     };
 
+    struct Uniform {
+        utils::CString name;    // full qualified name of the uniform field
+        uint16_t offset;        // offset in 'uint32_t' into the uniform buffer
+        uint8_t size;           // >1 for arrays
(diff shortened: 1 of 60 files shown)


filament/backend/src/opengl/OpenGLContext.h
```diff
@@ -568,12 +610,19 @@ void OpenGLContext::bindVertexArray(RenderPrimitive const* p) noexcept {
 
 void OpenGLContext::bindBufferRange(GLenum target, GLuint index, GLuint buffer,
         GLintptr offset, GLsizeiptr size) noexcept {
+    assert_invariant(state.major > 2);
+
+#ifndef FILAMENT_SILENCE_NOT_SUPPORTED_BY_ES2
+    assert_invariant(true
```
bejado, line 616: I think you meant `false` here?

filament/backend/include/backend/Program.h
```diff
@@ -90,6 +90,10 @@ class Program {
     // the program everything it needs to know about the uniforms at a given binding
     Program& uniforms(uint32_t index, UniformInfo const& uniforms) noexcept;
 
+    // Note: This is only needed for GLES2.0.
+    Program& attributes(
+            utils::FixedCapacityVector<std::pair<utils::CString, uint8_t>> attributes) noexcept;
```
bejado, line 95: Make `attributes` a `const&` then simply assign to `mAttributes` instead of moving? This is consistent with the other methods of `Program`.

shaders/src/common_instancing.glsl
```diff
@@ -20,16 +21,26 @@ void initObjectUniforms(out PerRenderableData p) {
     p.flagsChannels = objectUniforms.data[instance_index].flagsChannels;
     p.objectId = objectUniforms.data[instance_index].objectId;
     p.userData = objectUniforms.data[instance_index].userData;
+#   else
+    p.worldFromModelMatrix = objectUniforms.data[0].worldFromModelMatrix;
```
bejado, line 25: In this case could we do `p = objectUniforms.data[0]`?

pixelflinger: no because that's not supported in ES2!!!

filament/src/details/View.cpp
```diff
@@ -973,19 +973,38 @@ void FView::executePickingQueries(backend::DriverApi& driver,
...
-        });
+
+        if (UTILS_UNLIKELY(driver.getFeatureLevel() == FeatureLevel::FEATURE_LEVEL_0)) {
+            driver.readPixels(handle, x, y, 1, 1, {
+                    &pQuery->result.reserved1, 4u, // 4
+                    backend::PixelDataFormat::RGBA, backend::PixelDataType::UBYTE,
+                    pQuery->handler, [](void*, size_t, void* user) {
+                        FPickingQuery* pQuery = static_cast<FPickingQuery*>(user);
+                        int32_t const identity = int32_t(pQuery->result.reserved1 >> 8);
+                        float const depth = float(pQuery->result.reserved1 & 0xFF) / 255.0f;
```
poweifeng, line 984: Looks like the alpha bit is the depth, but in the shader it reads like the r channel is the depth?

pixelflinger: no because actually RGBA8 / UBYTE ends up being R,G,B,A in memory which is 0xABGR when loaded as a uint32_t. But this is a good point though that I'm relying on little-endian. I should fix/clarify that.

filament/backend/src/opengl/gl_headers.h
```diff
@@ -42,10 +42,15 @@
     #if defined(__EMSCRIPTEN__)
     #   include <GLES3/gl3.h>
     #else
+//    #   include <GLES2/gl2.h>
     #   include <GLES3/gl31.h>
     #endif
     #include <GLES2/gl2ext.h>
 
+    #if defined(GL_ES_VERSION_2_0) && !defined(GL_ES_VERSION_3_0)
```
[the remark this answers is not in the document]
pixelflinger: done

filament/src/details/View.cpp
pixelflinger: made it endian proof

filament/backend/src/opengl/OpenGLContext.h
pixelflinger: fixed

poweifeng (approved):

PR #6786, opened 2023-05-03
ES2: add support for devices without VAOs

pixelflinger:


filament/backend/src/opengl/OpenGLContext.h
```diff
@@ -38,7 +38,7 @@ class OpenGLPlatform;
 class OpenGLContext {
 public:
     static constexpr const size_t MAX_TEXTURE_UNIT_COUNT = MAX_SAMPLER_COUNT;
-    static constexpr const size_t DUMMY_TEXTURE_BINDING = 31; // highest binding less than 32
+    static constexpr const size_t DUMMY_TEXTURE_BINDING = 7; // highest binding guaranteed to work with ES2
```
bejado, line 41: This won't conflict with other texture bindings?

bejado (approved):

poweifeng (approved):

filament/backend/src/opengl/OpenGLContext.h
pixelflinger: The conflict doesn't matter, I picked 31 initially because it was the least likely to be used already and cause extra gl calls. It's just that we can cause some extra gl calls, but I don't think it really matters.

PR #6789, opened 2023-05-04
gltfio patches

bjovanovic84:
the changes here expand gltfio to support operation in certain low resource scenarios (i.e. rendering on a small screen mobile device):
- asset loader config is expanded to support skipping of ambient occlusion, normal and metallic roughness maps (useful when loading photogrammetry models)
- texture provider now has support for defining max texture size and downsampling it if needed (no point loading 8k textures on small screens)


libs/gltfio/src/FilamentAsset.cpp
```diff
@@ -81,7 +81,7 @@ FFilamentAsset::~FFilamentAsset() {
         mEngine->destroy(ib);
     }
     for (auto tx : mTextures) {
-        if (UTILS_LIKELY(tx.isOwner)) {
+        if (UTILS_LIKELY(tx.isOwner) && tx.texture != nullptr) {
```
romainguy, line 84: This check is not necessary.

libs/gltfio/include/gltfio/AssetLoader.h
```diff
@@ -62,6 +62,15 @@ struct AssetConfiguration {
 
     //! Optional default node name for anonymous nodes
     char* defaultNodeName = nullptr;
+
+    //! Whether to skip ambient occlusion maps. Generally not needed for photogrammetry assets
+    bool skipAmbientOcclusionMaps = false;
```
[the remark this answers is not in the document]
bjovanovic84: this is just an optimization for different rendering scenarios, i.e. when just viewing a single pre-lit model (photogrammetry capture) those are not necessary and just waste resources, even though they are included in the gltf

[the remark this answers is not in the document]
pixelflinger: why would they be included in the gltf if they serve no purpose?

libs/gltfio/include/gltfio/TextureProvider.h
```diff
@@ -171,7 +173,7 @@ class UTILS_PUBLIC TextureProvider {
  * Creates a simple decoder based on stb_image that can handle "image/png" and "image/jpeg".
  * This works only if your build configuration includes STB.
  */
-TextureProvider* createStbProvider(filament::Engine* engine);
+TextureProvider* createStbProvider(filament::Engine* engine, const std::optional<unsigned int> &maxTextureSize = std::optional<unsigned int>());
```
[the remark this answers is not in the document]
omri-polycam: Just a more clear and elegant way of expressing that it isn't in use.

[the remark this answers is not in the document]
pixelflinger: We don't use the STL in public APIs.

Could this be an API on `TextureProvider` instead? Maybe `setMaxTextureSize(size_t)`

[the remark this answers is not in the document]
romainguy: And it adds STL dependencies to a "public" header that's not necessary. It also doesn't convey what the default max size might be.

libs/gltfio/src/FilamentAsset.cpp
pixelflinger: yes  not necessary because `Engine::destroy` handles `nullptr`.

libs/gltfio/src/ResourceLoader.cpp
```diff
@@ -767,6 +767,9 @@ void ResourceLoader::Impl::createTextures(FFilamentAsset* asset, bool async) {
     // Create new texture objects if they are not cached and kick off decoding jobs.
     for (size_t textureIndex = 0, n = asset->mTextures.size(); textureIndex < n; ++textureIndex) {
         FFilamentAsset::TextureInfo& info = asset->mTextures[textureIndex];
+        if(info.bindings.empty()) {
```
pixelflinger, line 770:
```suggestion
        if (info.bindings.empty()) {
```

libs/gltfio/src/StbProvider.cpp
```diff
@@ -25,6 +25,10 @@
 #include <filament/Engine.h>
 #include <filament/Texture.h>
 
+#include <image/LinearImage.h>
+#include <image/ColorTransform.h>
+#include <image/ImageSampler.h>
```
pixelflinger, line 30:
```suggestion
#include <image/ColorTransform.h>
#include <image/ImageSampler.h>
#include <image/LinearImage.h>
```

libs/gltfio/src/StbProvider.cpp
```diff
@@ -84,15 +90,35 @@ class StbProvider final : public TextureProvider {
     std::string mRecentPushMessage;
     std::string mRecentPopMessage;
     Engine* const mEngine;
+    const std::optional<unsigned int> mMaxTextureSize;
 };
 
+static int scaleDownDimension(int dimension, float scaleDownFactor) {
+    assert_invariant(dimension > 0);
+    const auto scaledDimension = std::max(1, static_cast<int>(std::round(static_cast<float>(dimension) / scaleDownFactor)));
+    return (scaledDimension);
```
pixelflinger, line 99:
```suggestion
    return scaledDimension;
```

libs/gltfio/src/StbProvider.cpp
```diff
@@ -84,15 +90,35 @@ class StbProvider final : public TextureProvider {
     std::string mRecentPushMessage;
     std::string mRecentPopMessage;
     Engine* const mEngine;
+    const std::optional<unsigned int> mMaxTextureSize;
 };
 
+static int scaleDownDimension(int dimension, float scaleDownFactor) {
```
pixelflinger, line 96: parameters should be `uint32_t` because negative values are not accepted.

libs/gltfio/src/StbProvider.cpp
```diff
@@ -84,15 +90,35 @@ class StbProvider final : public TextureProvider {
...
+    assert_invariant(dimension > 0);
+    const auto scaledDimension = std::max(1, static_cast<int>(std::round(static_cast<float>(dimension) / scaleDownFactor)));
+    return (scaledDimension);
+}
+
+static void scaleDownDimensionsIfNeeded(int &width, int &height, unsigned int maxTextureSize) {
```
pixelflinger, line 102: for this, I prefer something like:
```
std::pair<uint32_t, uint32_t> scaleDownDimensionsIfNeeded(uint32_t width, uint32_t height,, uint32_t maxSize);
```

rather than using in/out parameters.

Type should not be signed, and also we're using sized types in our APIs.

libs/gltfio/src/StbProvider.cpp
```diff
@@ -84,15 +90,35 @@ class StbProvider final : public TextureProvider {
...
+  const auto largerDimension = std::max(width, height);
+  if(largerDimension <= maxTextureSize) {
```
pixelflinger, line 104:
```suggestion
  if (largerDimension <= maxTextureSize) {
```

libs/gltfio/src/StbProvider.cpp
```diff
@@ -84,15 +90,35 @@ class StbProvider final : public TextureProvider {
...
+}
+
 Texture* StbProvider::pushTexture(const uint8_t* data, size_t byteCount,
             const char* mimeType, TextureFlags flags) {
     int width, height, numComponents;
     if (!stbi_info_from_memory(data, byteCount, &width, &height, &numComponents)) {
         mRecentPushMessage = std::string("Unable to parse texture: ") + stbi_failure_reason();
         return nullptr;
     }
+    if(mMaxTextureSize.has_value()) {
```
pixelflinger:
```suggestion
    if (mMaxTextureSize.has_value()) {
```

libs/gltfio/src/StbProvider.cpp
```diff
@@ -84,15 +90,35 @@ class StbProvider final : public TextureProvider {
...
+
+static void scaleDownDimensionsIfNeeded(int &width, int &height, unsigned int maxTextureSize) {
+  const auto largerDimension = std::max(width, height);
+  if(largerDimension <= maxTextureSize) {
+    return;
+  }
+  const auto scaleDownFactor = static_cast<float>(largerDimension) / static_cast<float>(maxTextureSize);
+  width = scaleDownDimension(width, scaleDownFactor);
+  height = scaleDownDimension(height, scaleDownFactor);
+}
```
pixelflinger, line 110: indentation is 4 spaces

libs/gltfio/src/StbProvider.cpp
```diff
@@ -84,15 +90,35 @@ class StbProvider final : public TextureProvider {
...
+
 Texture* StbProvider::pushTexture(const uint8_t* data, size_t byteCount,
             const char* mimeType, TextureFlags flags) {
     int width, height, numComponents;
     if (!stbi_info_from_memory(data, byteCount, &width, &height, &numComponents)) {
         mRecentPushMessage = std::string("Unable to parse texture: ") + stbi_failure_reason();
         return nullptr;
     }
+    if(mMaxTextureSize.has_value()) {
+      scaleDownDimensionsIfNeeded(width, height, *mMaxTextureSize);
```
pixelflinger: indentation 4 spaces

libs/gltfio/src/StbProvider.cpp
```diff
@@ -183,9 +197,11 @@ void StbProvider::updateQueue() {
...
                 continue;
             }
-            Texture::PixelBufferDescriptor pbd((uint8_t*) data,
-                    texture->getWidth() * texture->getHeight() * 4, Texture::Format::RGBA,
-                    Texture::Type::UBYTE, [](void* mem, size_t, void*) { stbi_image_free(mem); });
+            const auto textureSize = texture->getWidth() * texture->getHeight() * 4;
+            auto pbd = info->resized ? Texture::PixelBufferDescriptor((uint8_t*)data, textureSize, Texture::Format::RGBA, Texture::Type::UBYTE,
+                                                                      [](void* mem, size_t, void*) { delete[] ((uint8_t*)mem); })
+                                     : Texture::PixelBufferDescriptor((uint8_t*)data, textureSize, Texture::Format::RGBA, Texture::Type::UBYTE,
+                                                                      [](void* mem, size_t, void*) { stbi_image_free(mem); });
```
pixelflinger, line 203: 100 columns max, continuation is 8 spaces.

libs/gltfio/src/StbProvider.cpp
```diff
@@ -245,23 +261,47 @@ const char* StbProvider::getPopMessage() const {
     return mRecentPopMessage.empty() ? nullptr : mRecentPopMessage.c_str();
 }
 
+void StbProvider::decode(TextureInfo& info) {
```
pixelflinger, line 263: formating for this method:
 100 columns, continuation 8 spaces

libs/gltfio/include/gltfio/TextureProvider.h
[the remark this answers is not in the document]
pixelflinger: Document that specifying a maximum texture size could cause a CPU-side resizing of the texture (hence incurring a runtime cost and modification of the input data. You should document how the resizing is done; is it nearest or bilinear, or other?). Also document that images are converted to RGBA if they're not already.

libs/image/include/image/ImageSampler.h
```diff
@@ -20,6 +20,8 @@
 #include <image/LinearImage.h>
 
 #include <utils/compiler.h>
+#include <memory>
+#include <cstdint>
```
pixelflinger, line 24: we don't expose the STL in public headers. use `stdint.h` instead and don't use `memory` in public APIs.

libs/image/include/image/ImageSampler.h
```diff
@@ -159,6 +161,20 @@ uint32_t getMipmapCount(const LinearImage& source);
...
+
+/**
+ * Scales down a single byte per component RGBA image where each destination pixel corresponds to
+ * a whole number of source pixels on each dimension. canSimpleScaleDown(sourceWidth, sourceHeight, destinationWidth, destinationHeight)
+ * must return true as a precondition.
+ */
+std::unique_ptr<uint8_t[]> simpleScaleDownRgba(
+    const uint8_t *source, uint32_t sourceWidth, uint32_t sourceHeight, uint32_t sourceLineStride,
+    uint32_t destinationWidth, uint32_t destinationHeight, uint32_t destinationLineStride);
+
```
pixelflinger, line 177: don't put this in public headers, this is only internal machinery.

libs/image/src/ImageSampler.cpp
```diff
@@ -379,4 +379,64 @@ Filter filterFromString(const char* rawname) {
...
+          sum += (uint32_t)source[componentSource + sourceComponentOffsets[sourceComponentOffset]];
+        }
+        destination[pixelDestination + component] = sum / sourcePixelsPerDestinationPixel;
+      }
+      pixelSource += pixelSourceIncrement;
+    }
+  }
+
+  return(destination);
+}
```
pixelflinger, line 434: indentation 4 spaces, 100 columns

libs/image/src/ImageSampler.cpp
```diff
@@ -379,4 +379,64 @@ Filter filterFromString(const char* rawname) {
...
+
+std::unique_ptr<uint8_t[]> simpleScaleDownRgba(
+    const uint8_t *source, uint32_t sourceWidth, uint32_t sourceHeight, uint32_t sourceLineStride,
+    uint32_t destinationWidth, uint32_t destinationHeight, uint32_t destinationLineStride) {
+  return(simpleScaleDown<4u>(source, sourceWidth, sourceHeight, sourceLineStride, destinationWidth, destinationHeight, destinationLineStride));
+}
```
pixelflinger, line 440: indentation 4 spaces, 100 columns

pixelflinger (changes requested):
lots of issues with formatting.

libs/gltfio/include/gltfio/AssetLoader.h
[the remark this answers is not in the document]
bjovanovic84: one example would be object capture API, we're using it for some 3d reconstructions and it produces a model with all of those included. we do not want to strip them as they may be useful (now or in the future). one example is putting the reconstructed model in an AR scene, we obviously want ambient occlusion maps in such scenario. but when iterating on the model (i.e. cropping it, resizing etc.) those just inflate the memory used w/o any benefit, the difference is substantial and ranges typically 2-3x

libs/gltfio/src/StbProvider.cpp
bjovanovic84: ended up calling this `getFinalDimensions` with the signature you suggested

libs/image/include/image/ImageSampler.h
bjovanovic84: moved these to the `StbProvider`, not exposed publicly anymore

libs/gltfio/src/StbProvider.cpp
```diff
@@ -183,9 +198,20 @@ void StbProvider::updateQueue() {
...
+                            textureSize,
+                            Texture::Format::RGBA,
+                            Texture::Type::UBYTE,
+                            [](void* mem, size_t, void*) { delete[] ((uint8_t*)mem); })
+                            : Texture::PixelBufferDescriptor(
+                                    (uint8_t*)data,
+                                    textureSize,
+                                    Texture::Format::RGBA,
+                                    Texture::Type::UBYTE,
+                                    [](void* mem, size_t, void*) { stbi_image_free(mem); });
```
pixelflinger, line 214:
```suggestion
            const auto textureSize = texture->getWidth() * texture->getHeight() * 4;
            auto pbd = info->resized
                    ? Texture::PixelBufferDescriptor(
                            (uint8_t*)data,
                            textureSize,
                            Texture::Format::RGBA,
                            Texture::Type::UBYTE,
                            [](void* mem, size_t, void*) { delete[] ((uint8_t*)mem); })
                    : Texture::PixelBufferDescriptor(
                            (uint8_t*)data,
                            textureSize,
                            Texture::Format::RGBA,
                            Texture::Type::UBYTE,
                            [](void* mem, size_t, void*) { stbi_image_free(mem); });
```

libs/gltfio/src/StbProvider.cpp
```diff
@@ -245,23 +271,123 @@ const char* StbProvider::getPopMessage() const {
...
+        return;
+    }
+    info.resized = ((width != info.texture->getWidth()) || (height != info.texture->getHeight()));
+    if (info.resized) {
+        if (canSimpleScaleDown(width, height,
+                info.texture->getWidth(), info.texture->getHeight())) {
+            auto scaled_down_texels = simpleScaleDownRgba(
+                    texels, width, height, width * sizeof(std::uint8_t) * 4u,
+                    info.texture->getWidth(), info.texture->getHeight(),
+                    info.texture->getWidth() * sizeof(std::uint8_t) * 4u);
```
pixelflinger, line 364: why are you using SimpleScaleDownRgba instead of always using image::resampleImage?

Is it for performance? Does it matter? What are the odds that given a max size, an image can be resized by a integer amount?

Also, when you take the canSimpleScalDown path, you're not first transforming to linear, so your averaging is wrong. At any rate, both algorithm wouldn't produce the same result and I think they should.

pixelflinger (changes requested):

libs/gltfio/src/StbProvider.cpp
omri-polycam: Hi pixelflinger. Yes, this is for performance reasons which we are pretty sensitive to in our application. Typically this will be used to take our power of 2 dimension textures and scale them down by a factor of 2 or 4 (for example 2048 to 1024 or maybe 512).
So this basically averages 4 source texels or 16 source texels to determine the value of the destination pixel. You are correct that this computation is done in sRGB and not linear scale so the answer will not be 100% precise, however I don't think the difference would be discernible to most users.
A conversion to linear, running `SimpleScaleDownRgba()` followed by a conversion back to sRGB is doable, but would be considerably more computationally expensive. Alternatively, if you are pretty confident the performance difference between this and `image::resampleImage()` would be negligible even in the dimension special case needed for `SimpleScaleDownRgba()`, we could simply use `image::resampleImage()` in all cases and remove `SimpleScaleDownRgba()`.

romainguy: > A conversion to linear, running SimpleScaleDownRgba() followed by a conversion back to sRGB is doable, but would be considerably more computationally expensive.

It is, but it's also the correct thing to do for color data. Filament is properly color managed everywhere, and we need to preserve that.

bjovanovic84: I just did some profiling w/o the simple scale down method (using resample) and the results for scaling down an 8k example texture down to 4k are:
- ios: simple scale down ~100ms, rescale >2m
- android: simple scale down ~200ms, rescale >5m
- wasm: simple scale down ~250ms, rescale >5m

admittedly the textures we're working with are always a power of 2, so simple scale down works on the fly, but using rescale is prohibitively expensive. note that all of those numbers were using release builds, I can provide the exact texture if you'd like to test it yourself.

even just conversions from/to sRBG/linar take considerably long time
```
toLinearWithAlpha took 747 ms
fromLinearTosRGB took 164 ms
```

do those numbers look correct to you or are we doing something wrong here?

pixelflinger: Given those numbers we really need to fix `image::resampleImage`, because if someone uses this new API and ends up in the slow path, they'll think the program is hung. That just seems unacceptable.

Given this situation, I think it would be better for the resizing to be done by the app, so that it's not our (filament's) responsibility how the scaling is done.

I think, instead of specifying a max texture size, we could pass a texture transform callback (use utils::Invocable), if the callback is provided, gltfio will call it for all textures. The callback can decide if it wants to resize the texture and do so if needed.

I realize this would change this PR significantly, but I think it's a better solution. @romainguy what do you think?

bjovanovic84: that sounds reasonable to me, that gives us enough flexibility, we can keep this downsampling logic on our side. will make those changes tomorrow

pixelflinger:
@bjovanovic84 there has not been any activity here in a while, should we close this PR?

PR #6790, opened 2023-05-04
iOS: make sure to save/restore the framebuffer binding

pixelflinger:
this is needed because filament does state tracking.


bejado (approved):

bejado:
Verified on an iOS device.

PR #6791, opened 2023-05-04
Use a single material for separable gaussian blur

bejado:


pixelflinger:
PR looks good, but I'm worried that this means that all cases will "cost" 4 components worth of math.

I don't think the driver can optimize this out, unless it recompiles the shader based on the attachments.

So, I'm wondering if we could at least use the spec-constant to clear the channels we don't need at the end. for instance, we could do this:
```
if (numComponents == 1) {
 out.gba = vec3(0);
}
if (numComponents == 2) {
 out.ba = vec2(0);
}
if (numComponents == 3) {
 out.a = 0.0;
}
```

this is not guaranteed to work, but at least we give a change to the compiler to optimize out all the math leading to the unused components.

one way to guarantee this will happen (at least with spirv-opt) would be to rewrite the code using scalars instead of vectors.

bejado:
> PR looks good, but I'm worried that this means that all cases will "cost" 4 components worth of math.
> 
> I don't think the driver can optimize this out, unless it recompiles the shader based on the attachments.
> 
> So, I'm wondering if we could at least use the spec-constant to clear the channels we don't need at the end. for instance, we could do this:
> 
> ```
> if (numComponents == 1) {
>  out.gba = vec3(0);
> }
> if (numComponents == 2) {
>  out.ba = vec2(0);
> }
> if (numComponents == 3) {
>  out.a = 0.0;
> }
> ```
> 
> this is not guaranteed to work, but at least we give a change to the compiler to optimize out all the math leading to the unused components.
> 
> one way to guarantee this will happen (at least with spirv-opt) would be to rewrite the code using scalars instead of vectors.

Sounds good.

pixelflinger (approved):

PR #6792, opened 2023-05-04
Fix Sampler External texture build

dchen23:
SAMPLER_EXTERNAL is always false at texture build, cause related samples (sample-hello-camera, sample-stream-test) crash at native.

### DIFF
--- a/filament/src/details/Texture.cpp
+++ b/filament/src/details/Texture.cpp
@@ -137,12 +137,11 @@ Texture* Texture::Builder::build(Engine& engine) {
         switch (sampler) {
             case SamplerType::SAMPLER_2D:
             case SamplerType::SAMPLER_CUBEMAP:
+            case SamplerType::SAMPLER_EXTERNAL:
                 return true;
             case SamplerType::SAMPLER_3D:
             case SamplerType::SAMPLER_2D_ARRAY:
                 return engine.hasFeatureLevel(FeatureLevel::FEATURE_LEVEL_1);
-            case SamplerType::SAMPLER_EXTERNAL:
-                return false;
             case SamplerType::SAMPLER_CUBEMAP_ARRAY:
                 return engine.hasFeatureLevel(FeatureLevel::FEATURE_LEVEL_2);
         }


romainguy:
Those apps don't crash? I just tested them. What crash do you see?

dchen23:
> Those apps don't crash? I just tested them. What crash do you see?

```
2023-05-04 16:35:16.108 11634-11634 Filament                com...e.android.filament.streamtest  E  Panic
                                                                                                    in build:153
                                                                                                    reason: SamplerType 3 not support at feature level 1
2023-05-04 16:35:16.108 11634-11634 Filament                com...e.android.filament.streamtest  E  
2023-05-04 16:35:16.108 11634-11634 libc                    com...e.android.filament.streamtest  A  Fatal signal 6 (SIGABRT), code -1 (SI_QUEUE) in tid 11634 (ment.streamtest), pid 11634 (ment.streamtest)
2023-05-04 16:35:16.277 21077-21077 DEBUG                   pid-21077                            A  Process name is com.google.android.filament.streamtest, not key_process
2023-05-04 16:35:16.278 21077-21077 DEBUG                   pid-21077                            A  Cmdline: com.google.android.filament.streamtest
```
Tested latest `main` branch.

romainguy:
What do you test on?

dchen23:
Oppo Findx5 pro and OnePlus 7 pro

pixelflinger (approved):
thanks!

pixelflinger:
> I'm confused how this ever worked..

I think I broke this recently.

dchen23:
> I'm confused how this ever worked..

I think the issue is that `validateTarget` is moved from `setImage` to `build`. Before modification, the SAMPLER_EXTERNAL case under `setImage` returns false because this type don't need to set image, and right now it has code to deal with that:
```
if (UTILS_UNLIKELY(!engine.hasFeatureLevel(FeatureLevel::FEATURE_LEVEL_1))) {
        ASSERT_PRECONDITION(buffer.stride == 0 || buffer.stride == width,
                "PixelBufferDescriptor stride must be 0 (or width) at FEATURE_LEVEL_0");
    }
```
So the SAMPLER_EXTERNAL case under `build` should return true to allow this type to create.

---
By the way, though the native crash fixed for **sample-hello-camera** and **sample-stream-test**, the result still not right for **sample-hello-camera**, it won't stream preview. I have located the issue is caused by the change under "filament/backend/src/opengl/OpenGLContext.h" (commt:c67d9c).  The `DUMMY_TEXTURE_BINDING` changes to `7` which can be compatible with es2 `MaxCombinedTextureImageUnits` which is `8`. I can't figure the reason, the logic seems all right, but just won't work, and only `7` not work. So the workaround is set `DUMMY_TEXTURE_BINDING` to `0`\~`6` for es2 and `8`\~`31` additional for es3. It's pretty strange, so I didn't make a pull.

PR #6797, opened 2023-05-09
vulkan: refactor VulkanPlatform implementations

poweifeng:
Fold all the platform code into a single concrete implementation named PlatformVulkan.


bejado:
LGTM, but is there a technical reason for this change? Personally I prefer the implementations in separate platform files, at least perhaps separating out Linux-based / Windows / macOS.

pixelflinger (approved):

poweifeng:
> LGTM, but is there a technical reason for this change? Personally I prefer the implementations in separate platform files, at least perhaps separating out Linux-based / Windows / macOS.

We need a generic implementation that the client can subclass from. I suppose we can expose different VulkanPlatform subclasses, but they'll mostly be the same. It's easier to manage, IMO, if we isolate the platform specifics into just a few places and reduce code duplication.  Yes, it's mostly a preference thing.

bejado (approved):

PR #6810, opened 2023-05-16
vulkan: VulkanContext/Platform refactor

poweifeng:
We need to keep the context handles as part of the platform class so that we can implement the new swapchain API based on them.

 - Move creation of "context" handles from Context to Platform
 - VulkanContext contains immutable data
 - Change constructor of classes that depended on VulkanContext
 - Move timer query logic from VulkanDriver to VulkanTimerQuery and VulkanTimestamps


filament/backend/include/backend/platforms/VulkanPlatform.h
```diff
@@ -27,14 +32,80 @@ namespace filament::backend {
...
 
     int getOSVersion() const noexcept override { return 0; }
 
-   ~VulkanPlatform() override;
+
+    /**
+     * Clean up any resources owned by the Platform. For example, if the Vulkan instance handle was
+     * generated by the platform, we need to clean it up in this method.
+     */
+    virtual void cleanup();
```
pixelflinger, line 47: I generally use `terminate()` in filament.

pixelflinger (changes requested):
terminate instead of cleanup to be consistant with other apis?

filament/backend/include/backend/platforms/VulkanPlatform.h
poweifeng: renamed.

pixelflinger (approved):

PR #6811, opened 2023-05-16
New RenderableManager API to opt-out of fog

pixelflinger:
Fog can now be opted-out on a per renderable basis. When fog is disabled on a renderable it removes the requirement that this renderable's  materials have the FOG variant.

### DIFF
--- a/filament/src/components/RenderableManager.h
+++ b/filament/src/components/RenderableManager.h
@@ -64,6 +64,7 @@ public:
         bool morphing                   : 1;
         bool screenSpaceContactShadows  : 1;
         bool reversedWindingOrder       : 1;
+        bool fog                        : 1;
     };
 
     static_assert(sizeof(Visibility) == sizeof(uint16_t), "Visibility should be 16 bits");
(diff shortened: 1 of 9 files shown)


filament/src/components/RenderableManager.h
```diff
@@ -114,6 +115,8 @@ class FRenderableManager : public RenderableManager {
     inline void setReceiveShadows(Instance instance, bool enable) noexcept;
     inline void setScreenSpaceContactShadows(Instance instance, bool enable) noexcept;
     inline void setCulling(Instance instance, bool enable) noexcept;
+    inline void setFogEnabled(Instance instance, bool enable) noexcept;
+    inline bool getFogEnabled(RenderableManager::Instance instance) const noexcept;
```
poweifeng, line 119: Why qualify Instance as `RenderableManager::Instance` here and not in `setFogEnabled`?

poweifeng (approved):

filament/src/components/RenderableManager.h
pixelflinger: oh it's just a mistake. thanks.

PR #6813, opened 2023-05-17
vulkan: fix depth+stencil image aspect

poweifeng:


filament/backend/src/vulkan/VulkanTexture.cpp
```diff
@@ -383,8 +383,13 @@ VkImageView VulkanTexture::getImageView(VkImageSubresourceRange range, VkImageVi
 }
 
 VkImageAspectFlags VulkanTexture::getImageAspect() const {
-    return isDepthFormat(mVkFormat) ? VK_IMAGE_ASPECT_DEPTH_BIT :
-            VK_IMAGE_ASPECT_COLOR_BIT;
+    if (isDepthStencilFormat(mVkFormat)) {
```
pixelflinger, line 386: there is never a case where it's stencil only?

pixelflinger (approved):

filament/backend/src/vulkan/VulkanTexture.cpp
poweifeng: Good point. Pulled possible options into a general method for aspect bit.

PR #6827, opened 2023-05-23
Support for custom fog-color texture

pixelflinger:
Fix 282968398

### DIFF
--- a/filament/include/filament/Options.h
+++ b/filament/include/filament/Options.h
@@ -237,9 +237,31 @@ struct FogOptions {
     /**
      * The fog color will be sampled from the IBL in the view direction and tinted by `color`.
      * Depending on the scene this can produce very convincing results.
-     * This simulate a more anisotropic phase-function.
+     *
+     * This simulates a more anisotropic phase-function.
+     *
+     * `fogColorFromIbl` is ignored when skyTexture is specified.
+     *
+     * @see skyColor
      */
     bool fogColorFromIbl = false;
+
+    /**
+     * skyTexture must be a mipmapped cubemap. When provided, the fog color will be sampled from
+     * this texture, higher resolution mip levels will be used for objects at the far clip plane,
+     * and lower resolution mip levels for objects closer to the camera. The skyTexture should
+     * typically be heavily blurred; a typical way to produce this texture is to blur the base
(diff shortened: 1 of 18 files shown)


filament/include/filament/Options.h
```diff
@@ -237,10 +237,32 @@ struct FogOptions {
...
+     * levels as usual. How blurred the base level is somewhat of an artistic decision.
+     *
+     * This simulates a more anisotropic phase-function.
+     *
+     * `fogColorFromIbl` is ignored when skyTexture is specified.
+     *
+     * @see Texture
+     * @see fogColorFromIbl
+     */
+    Texture* skyColor = nullptr;    //!< %codegen_skip_json% %codegen_skip_javascript%
```
romainguy, line 264: Should we leave a TODO to add JS support later?

shaders/src/fog.fs
```diff
@@ -42,15 +42,15 @@ vec4 fog(vec4 color, highp vec3 view) {
 
 #if MATERIAL_FEATURE_LEVEL > 0
     if (frameUniforms.fogColorFromIbl > 0.0) {
-        // get fog color from envmap
-        // TODO: use a lower resolution mip as we get further (problem we don't have mips!)
-        float lod = frameUniforms.iblRoughnessOneLevel;
+        float normalizedDepth = d * frameUniforms.fogOneOverFarMinusNear - frameUniforms.fogNearOverFarMinusNear;
+        lowp vec2 minMaxMip = unpackHalf2x16(frameUniforms.fogMinMaxMip);
```
romainguy, line 46: lowp isn't needed here, since we default to mediump, therefore fp16.

shaders/src/fog.fs
```diff
@@ -42,15 +42,15 @@ vec4 fog(vec4 color, highp vec3 view) {
...
+        lowp float lod = mix(minMaxMip.y, minMaxMip.x, saturate(normalizedDepth));
```
romainguy, line 47: Same here

romainguy (approved):

filament/include/filament/Options.h
pixelflinger: We could, but we already have other places like this (e.g. for the bloom dirt texture). I think the `%codegen_skip_json%` is like a TODO basically.

PR #6828, opened 2023-05-23
IBLPrefilter can now generate the irradiance texture

pixelflinger:


libs/iblprefilter/src/materials/iblprefilter.mat
```diff
@@ -139,15 +146,30 @@ void postProcess(inout PostProcessInputs postProcess) {
...
-        // { L, lod }, with L.z == NoL
-        mediump vec4 entry = texelFetch(materialParams_kernel, ivec2(materialParams.attachmentLevel, i), 0);
-        if (entry.z > 0.0) {
-            float l = entry.w + materialParams.lodOffset;// we don't need to clamp, the h/w does it for us
-            Lx += sampleEnvironment(materialParams_environment, Tx * entry.xyz, l) * entry.z;
-            Ly += sampleEnvironment(materialParams_environment, Ty * entry.xyz, l) * entry.z;
-            Lz += sampleEnvironment(materialParams_environment, Tz * entry.xyz, l) * entry.z;
-            kernelWeight += entry.z;
+
+    if (materialConstants_irradiance) {
```
romainguy, line 150: The only difference between the loops is the ivec2() passed to texelFetch and how kernelWeight is computed. Seems like you could share the loop easily. The x param of the ivec2 can be a variable set outside the loop, and you can keep kernelWeight+=entry.z in the loop but override the kernelWeight when materialConstants_irradiance is true.

romainguy (approved):

libs/iblprefilter/src/materials/iblprefilter.mat
pixelflinger: in the 2nd loop the Lx, Ly, Lz (which are each vec3s) are multiplied by `entry.z`, not in the first loop. I could move the spec-constant check inside the loop, but it seemed clearer to keep them separate.

romainguy: Sorry I missed that bit. Yeah makes sense to keep them separate.

poweifeng (approved):

PR #6835, opened 2023-05-24
add support for custom fog color in gltf_viewer

pixelflinger:
this is implemented by here by using the skybox texture and blurring it with the irradiance filter + mipmapping. This only creates a subtle anisotropic phase-function effect.

### DIFF
--- a/libs/filamentapp/src/IBL.cpp
+++ b/libs/filamentapp/src/IBL.cpp
@@ -52,6 +52,7 @@ IBL::~IBL() {
     mEngine.destroy(mTexture);
     mEngine.destroy(mSkybox);
     mEngine.destroy(mSkyboxTexture);
+    mEngine.destroy(mFogTexture);
 }
 
 bool IBL::loadFromEquirect(Path const& path) {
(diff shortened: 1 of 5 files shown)


libs/filamentapp/src/IBL.cpp
```diff
@@ -136,6 +141,15 @@ bool IBL::loadFromKtx(const std::string& prefix) {
     mSkyboxTexture = Ktx1Reader::createTexture(&mEngine, skyKtx, false);
     mTexture = Ktx1Reader::createTexture(&mEngine, iblKtx, false);
 
+    // TODO: create the fog texture, it's a bit complicated because IBLPrefilter requires
+    //       the source image to have miplevels, and it's not guaranteed here and also
+    //       not guaranteed we can generate them (e.g. texture could be compressed)
+    //IBLPrefilterContext context(mEngine);
```
romainguy, line 147: We could check the texture format? Also for gltf_viewer, I don't think we have a way to load a compressed texture really.

romainguy: I would uncomment that code :)

romainguy (approved):

libs/filamentapp/src/IBL.cpp
pixelflinger: okay, I'll get back to it

PR #6852, opened 2023-05-31
Update MATERIAL_VERSION to 38

poweifeng:


libs/filabridge/include/filament/MaterialEnums.h
```diff
@@ -27,7 +27,7 @@
 namespace filament {
 
 // update this when a new version of filament wouldn't work with older materials
-static constexpr size_t MATERIAL_VERSION = 35;
+static constexpr size_t MATERIAL_VERSION = 38;
```
pixelflinger, line 30: do we need to skip several versions?

poweifeng: I thought it's suppose to mirror the middle release number (i.e. 1.37.0)
But I see it actually went backwards here?  https://github.com/google/filament/compare/v1.36.0...release

Maybe I'm not understanding it correctly.

pixelflinger (approved):

PR #6859, opened 2023-05-31
release: adjust version-bump script for linux

poweifeng:


bejado (approved):
😑 wow.. Thanks for investigating this!

PR #6864, opened 2023-06-01
vulkan: add option to choose gpu in platform

poweifeng:
- Added a optional flag to `gltf_viewer` to demonstrate this feature.

### DIFF
--- a/libs/filamentapp/src/FilamentApp.cpp
+++ b/libs/filamentapp/src/FilamentApp.cpp
@@ -71,9 +71,7 @@ public:
             return;
         }
         // Check to see if it is an integer, if so turn it into an index.
-        if (std::find_if(gpuHint.begin(), gpuHint.end(), [](unsigned char c) {
-                return !std::isdigit(c);
-            }) == gpuHint.end()) {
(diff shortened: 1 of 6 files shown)


libs/filamentapp/src/FilamentApp.cpp
```diff
@@ -58,6 +60,36 @@ using namespace filagui;
...
+using namespace filament::backend;
+
+class FilamentAppVulkanPlatform : public VulkanPlatform {
+public:
+    FilamentAppVulkanPlatform(std::string const& gpuHint) {
+        if (gpuHint.empty()) {
+            return;
+        }
+        // Check to see if it is an integer, if so turn it into an index.
+        if (std::find_if(gpuHint.begin(), gpuHint.end(), [](unsigned char c) {
```
bejado, line 74: Maybe rewrite as
```
std::all_of(gpuHint.begin(), gpuHint.end(), ::isdigit)
```
for readability

bejado (approved):

libs/filamentapp/src/FilamentApp.cpp
poweifeng: nice, thanks for the tip!

PR #6867, opened 2023-06-05
Fix strict weak ordering in VulkanPlatform.cpp

danlark1:


poweifeng (commented):
Just out of curiosity.  How did this affect how GPU was chosen for your machine?

pixelflinger (approved):

PR #6869, opened 2023-06-05
vulkan: fix context sharing

poweifeng:
- Make sure terminate() doesn't free shared context
 - Assume no extensions have been enabled.


filament/backend/src/vulkan/platform/VulkanPlatform.cpp
```diff
@@ -552,12 +556,18 @@ Driver* VulkanPlatform::createDriver(void* sharedContext,
...
+
+        mImpl->mSharedContext = true;
     }
 
     VulkanContext context;
 
-    auto instExts = getInstanceExtensions();
-    instExts.merge(getRequiredInstanceExtensions());
+    ExtensionSet instExts;
+    // If using a shared context, we do not assume any extensions.
```
pixelflinger, line 566: should we getting the extensions that are available from the shared context instead? does it matter?

pixelflinger (approved):

PR #6887, opened 2023-06-11
Increase max morph targets from 256 to 512

Obg1:
Using two textures instead of one and a modified shader to increase support for 512 max morph targets


pixelflinger:
I don't think this is a feature we want right now. The added complexity is not worth it.
If we wanted to support more than 256 morph targets we would add a query API of some sort to check if the h/w can support it (i.e. support texture arrays larger than 256 layers) and only allow it on this type of h/w. This wouldn't require changes to the shaders.

Obg1:
Since the model I'm using has more than 256 blend shapes and needs to run on GL ES  it a working fix.
But if that's not a common scenario I can understand your point.

PR #6892, opened 2023-06-13
align uniform buffer to 16 bytes

pixelflinger:
this is needed on armv7 because we use alignas to get strcture-alignment, but that also implies (to the compiler) that the structure itself is aligned properly.


poweifeng (approved):

bejado (approved):
Maybe add a comment explaining the 16, but can be done after merging

PR #6904, opened 2023-06-14
vulkan: address debugUtils bug

poweifeng:
vkCmdEndDebugUtilsLabelEXT expects that a label was "pushed" onto the command queue (as described in the spec). It is possible to push labels across command buffers, but the pushed label must still be in the queue (unexecuted) when End is called. This implies that we need to make sure the labels are in a good state (all popped) when vkQueueSubmit is called.

- We add a stack to carry the labels across vkQueueSubmit.
- Also add CPU time durations between push and pop to provide rough CPU execution times (in debug).
- Add systrace markers for Android systrace

### DIFF
--- a/filament/backend/src/vulkan/VulkanCommands.cpp
+++ b/filament/backend/src/vulkan/VulkanCommands.cpp
@@ -77,27 +77,27 @@ void VulkanGroupMarkers::push(std::string const& marker, Timestamp start) noexce
 #endif
 }
 
-std::tuple<std::string, Timestamp> VulkanGroupMarkers::pop() noexcept {
+std::pair<std::string, Timestamp> VulkanGroupMarkers::pop() noexcept {
     auto const marker = mMarkers.top();
     mMarkers.pop();
 
 #if FILAMENT_VULKAN_VERBOSE
     auto const topTimestamp = mTimestamps.top();
     mTimestamps.pop();
-    return std::make_tuple(marker, topTimestamp);
+    return std::make_pair(marker, topTimestamp);
 #else
-    return std::make_tuple(marker, Timestamp{});
+    return std::make_pair(marker, Timestamp{});
 #endif
 }
 
-std::tuple<std::string, Timestamp> VulkanGroupMarkers::top() const {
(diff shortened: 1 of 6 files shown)


filament/backend/src/vulkan/VulkanCommands.cpp
```diff
@@ -64,10 +68,51 @@ static VkCommandPool createPool(VkDevice device, uint32_t queueFamilyIndex) {
...
+void VulkanGroupMarkers::push(std::string const& marker, Timestamp start) noexcept {
+    mMarkers.push(marker);
+#if FILAMENT_VULKAN_VERBOSE
+    mTimestamps.push(start.time_since_epoch().count() > 0.0
+                                  ? start
+                                  : std::chrono::high_resolution_clock::now());
+#endif
+}
+
+std::tuple<std::string, Timestamp> VulkanGroupMarkers::pop() noexcept {
```
pixelflinger, line 80: Why not a std::pair instead?

filament/backend/src/vulkan/VulkanCommands.cpp
```diff
@@ -270,7 +333,94 @@ void VulkanCommands::updateFences() {
     }
 }
 
-} // namespace filament::backend
+void VulkanCommands::pushGroupMarker(char const* str, VulkanGroupMarkers::Timestamp timestamp) {
+#if FILAMENT_VULKAN_VERBOSE
+    // If the timestamp is not 0, then we are carrying over a marker across buffer submits.
+    // If it is 0, then this is a normal marker push and we should just print debug line as usual.
+    if (timestamp.time_since_epoch().count() == 0.0) {
+        utils::slog.d << "----> " << str << "\n" << utils::io::flush;
```
pixelflinger, line 341: You can use utils::io::endl instead of \n + flush

pixelflinger (approved):

filament/backend/src/vulkan/VulkanCommands.cpp
poweifeng: done

pixelflinger (approved):

PR #6911, opened 2023-06-19
Fix texture binding when active texture unit changed

awkr:
1. Fix a bug: should rebind texture when active texture unit changed. ( A method should only do one thing )
2. Add a feature: expose state updating result.

### DIFF
--- a/filament/backend/src/opengl/OpenGLContext.h
+++ b/filament/backend/src/opengl/OpenGLContext.h
@@ -110,7 +110,7 @@ public:
     inline void useProgram(GLuint program) noexcept;
 
           void pixelStore(GLenum, GLint) noexcept;
-    inline void activeTexture(GLuint unit) noexcept;
+    inline bool activeTexture(GLuint unit) noexcept;
     inline void bindTexture(GLuint unit, GLuint target, GLuint texId, size_t targetIndex) noexcept;
     inline void bindTexture(GLuint unit, GLuint target, GLuint texId) noexcept;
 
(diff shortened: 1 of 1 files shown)


filament/backend/src/opengl/OpenGLContext.h
```diff
@@ -674,7 +676,7 @@ void OpenGLContext::bindTexture(GLuint unit, GLuint target, GLuint texId, size_t
     update_state(state.textures.units[unit].targets[targetIndex].texture_id, texId, [&]() {
         activeTexture(unit);
         glBindTexture(target, texId);
-    }, (target == GL_TEXTURE_EXTERNAL_OES) && bugs.texture_external_needs_rebind);
+    }, ((target == GL_TEXTURE_EXTERNAL_OES) && bugs.texture_external_needs_rebind) || activeTexture(unit));
```
romainguy, line 679: activeTexture() is now called twice. Also changing texture unit doesn’t mean you must rebind the texture.

romainguy (commented):
See comment, also please explain in more detail what issue you are fixing here.

filament/backend/src/opengl/OpenGLContext.h
awkr: Hi, I am using lib `backend` directly in my project, excluding other functionalities of filament, such as `Engine`, and `Frame graph`. 

1. Usually, I may create a texture in the Application layer via `glGenTextures`, and pass it to the `backend` via `importTexture`. In short, for the `backend`, since an external texture can be used, it also means that other GL states can be modified, such as by calling `glActiveTexture` from the Application layer, in these cases `backend` may not be notified.

2. And one recommendation for coding is that a function preferably does only one thing. But here `bindTexture`, we made a check by the rule of binding state yet do other things like `activeTexture` which itself has state checking already.

romainguy: libbackend is not designed to be used in an environment where you make your own GL calls. It needs to do its tracking.

awkr: @romainguy Okay I see. besides the design purpose, let's just think about the implementation that `bindTexture` is combining target binding with unit activating. Of the two, if either one is changed, the other should be updated. so the pseudocode can be:

``` C++
if (unit-changed or target-changed) {
  update-unit;
  update-target;
}
```

PR #6914, opened 2023-06-21
Make initializeGlExtensions protected

SahilMadan:
This function is used in subclass PlatformEGLHeadless (see [here](https://github.com/google/filament/blob/1edad92d7341cb3f8dee59ef532aa6f13a411b61/filament/backend/src/opengl/platforms/PlatformEGLHeadless.cpp#L196)) resulting in build errors.

### DIFF
--- a/filament/backend/include/backend/platforms/PlatformEGL.h
+++ b/filament/backend/include/backend/platforms/PlatformEGL.h
@@ -142,8 +142,9 @@ protected:
         } egl;
     } ext;
 
-private:
     void initializeGlExtensions() noexcept;
+
+private:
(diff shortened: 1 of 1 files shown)


bejado:
LGTM. Can you please accept the CLA?

SahilMadan:
CLA has been accepted. Is it possible for me to apply the label "internal" or would somebody else need to do that?

poweifeng:
> CLA has been accepted. Is it possible for me to apply the label "internal" or would somebody else need to do that?

Applied it for you.  thanks.

bejado (approved):

PR #6918, opened 2023-06-23
refine: bind to default VAO when finished drawing

awkr:
Bind back to the default VAO when finished drawing, the purpose is to keep GL state management clean.


awkr:
Not sure whether to raise this pr, but my goal is to make the backend as general as possible ☺

romainguy:
Thanks for your contribution but this is unnecessary. The goal of state tracking in the backend is precisely so this kind of cleanup isn't needed.

PR #6925, opened 2023-06-27
Disable shader precompilation on WebGL

bejado:


pixelflinger:
Just to be clear, this is only a problem because Chrome advertises KHR_parallel_shader_compile BUT doesn't honor it on all platforms :⁠-(

imho this is another bug we should file against them

On Windows, this should work though (ie we should be able to display something while other shaders are compiling).


A better fix, if possible, would be to disable this at runtime only when this extension is buggy, using the "workaround framework".

bejado:
> On Windows, this should work though (ie we should be able to display something while other shaders are compiling).

Can you elaborate? The original discussion mentions Windows specifically: https://github.com/google/filament/discussions/6924

poweifeng (approved):

pixelflinger:
It was my understanding that on Windows, Chrome did implement the extension correctly, and in windows only.

This can be tested on the Khronos page for the extension.

But it's possible that it also depends on the graphic card?

Either way, it's lame that they advertise an extension and then don't implement it.

I'm okay with the fix. It's just that it is due to another Chrome bug.

PR #6938, opened 2023-07-06
fix: 64-Bits Windows, Path::exists() fails for files over 2GB.

emotiog:
## Summary
I have modified the code to address the incorrect handling of files larger than 2GB in a 64-bit Windows environment.

## Problem
I discovered that when using `utils::Path` to input a file path of more than 2GB and calling `utils::Path::exists()`, it throws an error indicating that the file does not exist.

example
``` C
    utils::Path src = utils::Path("big_file");
    bool exists = src.exists();
    // 'exists' would be false when the 'big_file' size is over 2 GB
```

## Environment
- OS: Windows 11 Enterprise 21H2 (x64 processor)
- Build: 22000,2057
- CPU: 12th Gen Intel(R) Core(TM) i9-12900K
- RAM: 128 GB

**(Mac and Linux(Ubuntu) have not occurred this error)**

## Test
- I have tested various types of pre-defined syntax, but I have confirmed that only those two statements resulted in this error on my computer.

### Test Code
``` C
#include <iostream>
#include <string>
#include <filesystem>

void stat_test(const std::string& url) noexcept
{
    {
        struct stat file;
        int i = stat(url.c_str(), &file);
        std::cout << "1) stat \tcase: " << (i == 0 ? "true" : "false") << " \tsize: " << file.st_size << std::endl;
    }

    {
        struct _stat64 file;
        int i = _stat64(url.c_str(), &file);
        std::cout << "2) _stat64 \tcase: " << (i == 0 ? "true" : "false") << " \tsize: " << file.st_size << std::endl << std::endl;
    }
}

int main(int argc, char** argv)
{
    std::cout << "stat test" << std::endl << std::endl;
    std::string url = "../big_file.ply"; // test file whose size is larger than 2GB.

    std::filesystem::path p = std::filesystem::path(url);
    if (!std::filesystem::exists(p))
    {
        std::cout << "file-path is invalid" << std::endl;
        return 0;
    }

#if defined(__LP64__) 
    std::cout << "__LP64__" << std::endl;
#endif
#if defined(_WIN64) 
    std::cout << "_WIN64" << std::endl;
    stat_test(url);
#endif
   
#if defined(__x86_64__) && !defined(__ILP32__)
    std::cout << "__x86_64__&& __ILP32__" << std::endl;
#endif

#if defined(_M_X64)
    std::cout << "_M_X64" << std::endl;
    stat_test(url);
#endif

#if defined(__ia64) 
    std::cout << "__ia64" << std::endl;
#endif
#if defined (_M_IA64)
    std::cout << "_M_IA64" << std::endl;
#endif
#if defined(__aarch64__)
    std::cout << "__aarch64__" << std::endl;
#endif
#if defined(__powerpc64__)
    std::cout << "__powerpc64__" << std::endl;
#endif

    return 1;
}
```

### Result
- This result shows you which predefined syntaxes would be compiled and shown the wrong result at `utils::Path::exists()`.

``` text
stat test
_WIN64
1) stat         case: false     size: 0
2) _stat64      case: true      size: 2699676395

_M_X64
1) stat         case: false     size: 0
2) _stat64      case: true      size: 2699676395
```

## Detail
`utils::Path::exists() const` is defined as follows:
``` C
bool Path::exists() const {
    struct stat file;
    return stat(c_str(), &file) == 0;
}
```

Here, the `struct stat` is mapped to `struct _stat64i32` in the Windows environment, and this structure supports 32-bit file formats.
> https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/stat-functions?view=msvc-170#time-type-and-file-length-type-variations-of-_stat

### Conclusion

To address the mentioned issue, I have added code that can choose the `_stat64` or `stat` based on the pre-defined conditions.

I chose the conditions in the "Test Code" section.

This modification allows for distinguishing file sizes larger than 2GB.

### DIFF
--- a/NEW_RELEASE_NOTES.md
+++ b/NEW_RELEASE_NOTES.md
@@ -8,3 +8,4 @@ appropriate header in [RELEASE_NOTES.md](./RELEASE_NOTES.md).
 
 ## Release notes for next branch cut
 
+- Java/Kotlin: user callbacks were not invoked on successful texture upload
--- a/android/filament-android/src/main/cpp/Texture.cpp
+++ b/android/filament-android/src/main/cpp/Texture.cpp
@@ -511,9 +511,7 @@ public:
 
 private:
     void* mData = nullptr;
-    jobject mBitmap = nullptr;
-    jobject mHandler = nullptr;
-    jobject mCallback = nullptr;
+    jobject mBitmap{};
     AndroidBitmapInfo mInfo{};
 };
 
--- a/android/samples/sample-textured-object/src/main/java/com/google/android/filament/textured/MainActivity.kt
+++ b/android/samples/sample-textured-object/src/main/java/com/google/android/filament/textured/MainActivity.kt
@@ -231,7 +231,7 @@ class MainActivity : Activity() {
         animator.repeatCount = ValueAnimator.INFINITE
         animator.addUpdateListener { a ->
             val v = (a.animatedValue as Float)
-            camera.lookAt(cos(v) * 4.5, 1.5, sin(v) * 4.5, 0.0, 0.0, 0.0, 0.0, 1.0, 0.0)
+            camera.lookAt(cos(v) * 5.5, 1.5, sin(v) * 5.5, 0.0, 0.0, 0.0, 0.0, 1.0, 0.0)
         }
         animator.start()
     }
@@ -253,7 +253,7 @@ class MainActivity : Activity() {
 
         // Stop the animation and any pending frame
         choreographer.removeFrameCallback(frameScheduler)
-        animator.cancel();
+        animator.cancel()
 
         // Always detach the surface before destroying the engine
         uiHelper.detach()
--- a/build.sh
+++ b/build.sh
@@ -90,8 +90,8 @@ function print_help {
     echo "    Desktop matc target, release build:"
     echo "        \$ ./$self_name release matc"
     echo ""
-    echo "    Build gltf_viewer then immediately run it with no arguments:"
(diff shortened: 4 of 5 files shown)


romainguy:
Thanks for updating the release notes but more importantly you need to sign the CLA.

emotiog:
> Thanks for updating the release notes but more importantly you need to sign the CLA.

Thank you for letting me know. 
I have already signed the CLA as follows. 

Agreement | Name | Date Signed | Manage
-- | -- | -- | --
Google Individual CLA | Junsang Lee | Jul 06, 2023 18:53 PDT | Edit Contact Information

Please let me know if there is anything else I need to consider.

romainguy:
That's odd it still doesn't show up as a valid check.

emotiog:
> That's odd it still doesn't show up as a valid check.

I have just confirmed that my email address was set to private. 
Now, I have changed the email address permission to public!

romainguy:
It still fails, can you sign the CLA again please?

romainguy:
Looks like the issue is the email listed in the PR:

> The following contributors were found for this pull request:
> 
> ✅ https://github.com/google/filament/commit/f703c672ed9c8bc7048b3177598ca1142c97b3ae PR Opener: @emotiog
> ❌ https://github.com/google/filament/commit/f703c672ed9c8bc7048b3177598ca1142c97b3ae Author: <em***g​@gmail.com>
>

emotiog:
I apologize for the mistake. 
There was a typo in the email during the commit. It has been corrected now.

romainguy:
Thank you so much!

PR #6940, opened 2023-07-06
Fix dynamic resolution with quality > low and translucency

### LINKED ISSUE #6927: When setting a transparent background, an incorrect background was rendered
**Describe the bug**
When I ran the sample-gltf-viewer demo and rendered the following Ford car model with transparent surface, the rendered result was wrong as follows:

Sometimes this situation also occurs when rendering the model in sample-gltf-viewer demo.
But the continuous display will return to right after a period of time
[ford_car.zip](https://github.com/google/filament/files/11893680/ford_car.zip)

**To Reproduce**
1. setup the sample-gltf-viewer project and 
2. set transparent background
```
val uiHelper = UiHelper(UiHelper.ContextErrorPolicy.DONT_CHECK)
        uiHelper.isOpaque = false
        modelViewer = MyModelViewer(surfaceView,uiHelper = uiHelper)

val options = modelViewer.renderer.clearOptions
        options.clear = true
        modelViewer.renderer.clearOptions = options

view.blendMode = View.BlendMode.TRANSLUCENT

```
3. render  [Ford car model](https://github.com/draty01/filerepo/blob/main/ford_car.7z)

**Expected behavior**
render the model with clear and  transparent background

**Screenshots**
img_v2_8f948ff9-405e-4a74-9879-1427ecd0e92g
the orange background is Activity backgroud

img_v2_dc3cb326-63f3-450d-b0d0-827b08e2306g

**Logs**
[log.txt](https://github.com/google/filament/files/11893668/log.txt)

 - OS: 
 - GPU: 
 - Backend: [OpenGL]

 - Device: [ xiaomi 10 ]
 - OS:     【 Android 13】【 miui14】
 
**Additional context**
It seems that models with large size are prone to appear

romainguy:
The scene graph was using the wrong boolean to decide whether to fallback to low quality upscaling when the render target is translucent. It was instead only looking at the view's blending mode. We need to check both, as color grading does in the same function.

Fixes #6927

### DIFF
--- a/NEW_RELEASE_NOTES.md
+++ b/NEW_RELEASE_NOTES.md
@@ -8,3 +8,4 @@ appropriate header in [RELEASE_NOTES.md](./RELEASE_NOTES.md).
 
 ## Release notes for next branch cut
 
+- rendering: dynamic resolution would not work with a translucent render target and quality > low
--- a/android/samples/sample-transparent-view/src/main/java/com/google/android/filament/transparentrendering/MainActivity.kt
+++ b/android/samples/sample-transparent-view/src/main/java/com/google/android/filament/transparentrendering/MainActivity.kt
@@ -140,9 +140,9 @@ class MainActivity : Activity() {
         camera = engine.createCamera(engine.entityManager.create())
 
         // clear the swapchain with transparent pixels
-        val options = renderer.clearOptions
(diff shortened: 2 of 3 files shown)


poweifeng:
belated lgtm

Issue #6888, opened 2023-06-12
rendering issue on gltf_viewer when flare enabled with some focal length and focus distance in arch linux


arduinocc04:
**Describe the bug**
Small black boxes attached to reflected sunlight appear on gltf_viewer screen when flare enabled with some focal length and focus distance in arch linux with Vulkan

**To Reproduce**
1. `./gltf_viewer`
2. Enable Flare.
3. Change the Focal length of camera sufficiently.
4. Enable DoF
5. Change the Focus distance of DoF sufficiently.

However, changing Aperture of camera, Speed of camera, ISO of camera may make easier to reproduce this bug.

**Expected behavior**
The black boxes do not appear.

**Screenshots**
Aperture, Speed, ISO, Focal length, Focus distance changed
![image](https://github.com/google/filament/assets/27909137/25467a07-f183-428a-a327-6633291e3a2d)
![image](https://github.com/google/filament/assets/27909137/c8422040-b6cd-4be2-b81c-f6f343e9c02d)
![image](https://github.com/google/filament/assets/27909137/0c62201e-dcf5-4df3-990e-e0869bd2e80d)

Aperture, Focal length, Focus distance changed
![image](https://github.com/google/filament/assets/27909137/d842e30b-0cc9-453f-9584-88edf4291679)

Focal length, Focus distance changed
![image](https://github.com/google/filament/assets/27909137/c3057c6b-3fc9-47c5-9e8e-797083168261)

**Logs**
FEngine (64 bits) created at 0x55eda2730700 (threading is enabled)
FEngine resolved backend: Vulkan
Vulkan device driver: NVIDIA 530.41.03
Selected physical device 'NVIDIA GeForce RTX 3060' from 1 physical devices. (vendor 0x10de, device 0x2504, driver 0x848a40c0, api 1.3)
Backend feature level: 1
FEngine feature level: 1
vkCreateSwapchain: 1024x640, 44, 0, 3, 1

**Desktop (please complete the following information):**
 - OS: Arch Linux - 6.3.6-arch1-1
 - GPU: NVIDIA RTX 3060 (12GB) - NVIDIA 530.41.03
 - Backend: Vulkan

poweifeng:
Are you able to reproduce this in the opengl backend?  (Add `-a opengl`) to your `gltf_viewer` command. 

Also, what version of Filament are you working with?

arduinocc04:
With opengl, I failed to reproduce that bug. And I used v1.38.0

[...]

PR #6943, opened 2023-07-07
Fix rendering issue when using DoF

arduinocc04:
Fix rendering issue when using DoF.  
This solves #6888.

### DIFF
--- a/filament/src/materials/dof/dof.mat
+++ b/filament/src/materials/dof/dof.mat
@@ -477,6 +477,7 @@ void fastTile(inout vec4 color, inout float alpha,
     }
     alpha = cocToAlpha(kernelSize);
     color *= rcp(sampleCount(ringCountFast));
+    color = min(vec4(MEDIUMP_FLT_MAX), color);
 }
 
 void foregroundTile(inout vec4 foreground, inout float fgOpacity,
@@ -658,6 +659,7 @@ void postProcess(inout PostProcessInputs postProcess) {
     // the downside of doing this is that we couldn't use a different upscaler for each
     // layer, but this is a lot less costly
     postProcess.color = foreground + (1.0 - fgOpacity) * background;
+    postProcess.color = min(vec4(MEDIUMP_FLT_MAX), postProcess.color);
     postProcess.alpha = fgOpacity  + (1.0 - fgOpacity) * bgOpacity;
 }
 


filament/src/materials/dof/dof.mat
```diff
@@ -477,6 +477,7 @@ void fastTile(inout vec4 color, inout float alpha,
     }
     alpha = cocToAlpha(kernelSize);
     color *= rcp(sampleCount(ringCountFast));
+    color = saturate(color);
```
romainguy, line 480: Doesn't the DoF work on HDR values? If so, `saturate()` is the wrong thing to do. `saturateMediump()` might be what you want though. Also, is it because values go too high/to infinity or because we have `NaN`s?

romainguy (changes requested):

filament/src/materials/dof/dof.mat
romainguy: Based on the screenshots in the bug, it seems like what's needed is a `max(0.0f, color)` instead?

arduinocc04: `saturateMediump()` didn't work.
image
image

Also, `max(0.0f, color)` didn't work.
image

And, I think it's because values go too high/to infinity because `min(1.0f, color);` works.

romainguy: Ah right you are on desktop so `saturateMediump()` is a no-op. Again, `min(1.0, color)` is wrong if the data is in HDR, which it may not be if DoF works on pre-tonemapped data.

arduinocc04: I found implementation of `saturateMediump()` [here](https://gist.github.com/romainguy/a2e9208f14cae37c579448be99f78f25?permalink_comment_id=2194308#gistcomment-2194308).  
And `min(65504.0f, color)` looks working for me. Is it okay to replace `saturate(color)` with `min(65004.0f, color)`?

arduinocc04: And does it imply we can use `saturateMediump()` for all device not only mobile?

filament/src/materials/dof/dof.mat
```diff
@@ -477,6 +477,7 @@ void fastTile(inout vec4 color, inout float alpha,
     }
     alpha = cocToAlpha(kernelSize);
     color *= rcp(sampleCount(ringCountFast));
+    color = min(vec4(65504.0f), color);
```
pixelflinger, line 480: can you use `MEDIUMP_FLT_MAX` instead?

filament/src/materials/dof/dof.mat
```diff
@@ -658,6 +659,7 @@ void postProcess(inout PostProcessInputs postProcess) {
     // the downside of doing this is that we couldn't use a different upscaler for each
     // layer, but this is a lot less costly
     postProcess.color = foreground + (1.0 - fgOpacity) * background;
+    postProcess.color = min(vec4(65504.0f), postProcess.color);
```
pixelflinger, line 662: Did you check that doing the `min` here is not enough? I wonder if the one on line 480 is needed if we have this one.

arduinocc04: Oh sorry, I didn't noticed your comment. Do you mean that I have to use `max(vec4(MEDIUMP_FLT_MIN, x)` also?  
And line 480 is necessary. If I omit it, black squares are shown inside the rings.

pixelflinger: yes, use `MEDIUMP_FLT_MIN`. thanks.

filament/src/materials/dof/dof.mat
```diff
@@ -658,6 +659,8 @@ void postProcess(inout PostProcessInputs postProcess) {
     // the downside of doing this is that we couldn't use a different upscaler for each
     // layer, but this is a lot less costly
     postProcess.color = foreground + (1.0 - fgOpacity) * background;
+    postProcess.color = min(vec4(MEDIUMP_FLT_MAX), postProcess.color);
+    postProcess.color = max(vec4(MEDIUMP_FLT_MIN), postProcess.color);
```
pixelflinger, line 663: no we don't need `max(MEDIUMP_FLT_MIN` that would prevent the color to go to 0.

pixelflinger (approved):

PR #6946, opened 2023-07-10
Fixing PPM export and windows build errors with VS 22

dabeschte:
This small PR implements all fixes that I needed to make this project compile under Windows 10 with Visual Studio 2022.

- added vulkan: fixed missing "bluevk" library
- set c++ standard to 20. Did not compile with 17, and defaulted at least on my system to latest experimental which did not compile.
- fixed PPM export writing additional bytes which break the file format. This happened, because the file was not opened explicitly as a binary

(reupload because of some CLA/email conflicts)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -166,7 +166,6 @@ if (WIN32)
     # "VCRUNTIME140.dll not found. Try reinstalling the app.", but give users
     # a choice to opt for the shared runtime if they want.
     option(USE_STATIC_CRT "Link against the static runtime libraries." ON)
-    set(CMAKE_CXX_STANDARD 20)
 
     # On Windows we need to instruct cmake to generate the .def in order to get the .lib required
     # when linking against dlls. CL.EXE will not generate .lib without .def file (or without pragma
@@ -283,11 +282,11 @@ endif()
 # ==================================================================================================
 set(CXX_STANDARD "-std=c++17")
 if (WIN32)
-    set(CXX_STANDARD "/std:c++17")
+    set(CXX_STANDARD "/std:c++20")
 endif()
 
 if (MSVC)
-    set(CXX_STANDARD "/std:c++latest")
+    set(CXX_STANDARD "/std:c++20")
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${CXX_STANDARD} /W0 /Zc:__cplusplus")
 else()
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${CXX_STANDARD} -fstrict-aliasing -Wno-unknown-pragmas -Wno-unused-function -Wno-deprecated-declarations")
@@ -453,7 +452,6 @@ endif()
 
 # By default, build with Vulkan support on desktop platforms, although clients must request to use
 # it at run time.
-# if (WIN32 OR WEBGL OR IOS)
 if (WEBGL OR IOS)
     option(FILAMENT_SUPPORTS_VULKAN "Include the Vulkan backend" OFF)
 else()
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -166,6 +166,7 @@ if (WIN32)
     # "VCRUNTIME140.dll not found. Try reinstalling the app.", but give users
     # a choice to opt for the shared runtime if they want.
     option(USE_STATIC_CRT "Link against the static runtime libraries." ON)
+    set(CMAKE_CXX_STANDARD 20)
```
romainguy, line 169: What doesn't compile with C++17 for you?

dabeschte: I get error `C7555` in [DriverEnums.h line 1109 and 1113](https://github.com/google/filament/blob/45f5a07989861b9b01782c4a8ba188a6ea107f55/filament/backend/include/backend/DriverEnums.h#L1109)

`use of designated initializers requires at least '/std:c++20'`

This also happends in other files like [here](https://github.com/google/filament/blob/45f5a07989861b9b01782c4a8ba188a6ea107f55/filament/include/filament/Box.h#L180) or [here](https://github.com/google/filament/blob/45f5a07989861b9b01782c4a8ba188a6ea107f55/libs/viewer/src/Settings.cpp#L587)

pixelflinger: We can fix these two lines in `DriverEnums.h` easily so the don't use designated initializers.

Regardless, I thing this is a problem is other places and I thought we were compiling with C++20 already on windows, specifically because of that?

bejado: My previous comment only meant that we shouldn't use C++20 globally across the project, since 1P clients are limited to C++1 7 at the moment. On Windows we have more flexibility.

On Windows, we're currently using `/std:c++latest` for MSVC, but maybe we need to explicitly use `/std:c++20` (apparently those two flags have slightly different meanings). @dabeschte can you try replacing this line with `/std:c++20` and seeing if it fixes the designated initializer error?

https://github.com/google/filament/blob/626577fe3d3697d3a5f3cca691b24e0e57f159aa/CMakeLists.txt#L285

dabeschte: No that does not work, since the if (MSVC) 3 lines below overrides that. This results in MSVC using the latest preview, not the latest stable compiler version.
image

The error that I get in this case is 
```
Error	C3878	syntax error: unexpected token 'this' following 'expression'	backend	<filament>\libs\utils\include\utils\Range.h	37	
```

Caused by this piece of code that actually looks fine to me, but since we are using preview features, I assume the compiler might have a bug here?

```
    bool overlaps(const Range<T>& that) const noexcept {
        return that.first < this->last && that.last > this->first;
    }
```

Anyways, changing line 289 in CMakeLists.txt to "/std:c++20" solves the problem for Windows.
That said, MSVC is actually right about the invalid usage of C++20 features in multiple files - not only in DriverEnums.h 
[G++ just doesn't show a warning by default](https://stackoverflow.com/questions/69009389/how-to-get-away-with-using-designated-initializers-in-c17-or-why-is-it-seemi#:~:text=understanding%20is%20backward.-,C%2B%2B17%20does%20not%20support%20designated%20initialisers%20at%20all,acceptance%20in%20the%20standardisation%20process.)

pixelflinger: `C3878` has just been fixed.

bejado (approved):

pixelflinger (approved):

CMakeLists.txt
z3moon: I'm currently building Filament using VS22 now with no issue. So maybe the latest preview has been fixed for the issue?

poweifeng: Can you try changing the line and see if it still compiles?

```
std::ofstream ppmStream(out, std::ios_base::binary);
```

poweifeng:
So we do have vs22 build running now ([ref](https://github.com/google/filament/blob/main/build/windows/build-github.bat#L53)).  Doesn't seem like we need what's mentioned here.  Please re-open if you see different.

PR #6952, opened 2023-07-13
web: remove const qualifier for getMaterialInstances

mackong:
With the *const* qualifier on the returned MaterialInstances, we cannot call setter functions like *setDepthWrite* etc. from the object, caused by
```
BindingError: Cannot convert argument of type MaterialInstance const* to parameter type MaterialInstance*
```

### DIFF
--- a/web/filament-js/jsbindings.cpp
+++ b/web/filament-js/jsbindings.cpp
@@ -368,7 +368,7 @@ using EntityVector = std::vector<utils::Entity>;
 register_vector<std::string>("RegistryKeys");
 register_vector<utils::Entity>("EntityVector");
 register_vector<FilamentInstance*>("AssetInstanceVector");
-register_vector<const MaterialInstance*>("MaterialInstanceVector");
+register_vector<MaterialInstance*>("MaterialInstanceVector");
 
 // CORE FILAMENT CLASSES
 // ---------------------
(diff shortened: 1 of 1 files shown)


web/filament-js/filament.d.ts
```diff
@@ -624,7 +624,7 @@ export class gltfio$FilamentInstance {
     public getSkinNames(): Vector<string>;
     public attachSkin(skinIndex: number, entity: Entity): void;
     public detachSkin(skinIndex: number, entity: Entity): void;
-    public getMaterialInstances(): Vector<MaterialInstance>;
```
romainguy, line 627: Why this change here? The other getters return a Vector<>.

mackong: *getMaterialVariantNames* is wrapper by *vectorToArray*,  which is convenient when iterate, so make this change to *getMaterialInstances* too.
What's the convention (vectorToArray or not) here, should all interfaces defined in *filament.d.ts* use Vector?

bejado: I think we should keep this consistent with the other APIs and continue to return a `Vector<MaterialInstance>`.

bejado (changes requested):

web/filament-js/filament.d.ts
mackong: Sure, reverted.

PR #6955, opened 2023-07-13
Removed Code Duplication and Improved Readability

abhi27033:
Normalize the plane vectors outside the loop: Instead of normalizing each plane vector individually, you can normalize them all at once after the loop that calculates them. This reduces redundant calculations and improves performance.

Use a loop for copying planes: Instead of manually copying each plane to the output array, you can use a loop to copy them. This simplifies the code and makes it more concise.

Use dot product function: Instead of explicitly calculating the dot product using dot(mPlanes[i].xyz, p), you can use the dot function from the Filament math library. 
This enhances readability and reduces code duplication.

### DIFF
--- a/filament/src/Frustum.cpp
+++ b/filament/src/Frustum.cpp
@@ -28,9 +28,18 @@ namespace filament {
 Frustum::Frustum(const mat4f& pv) {
     Frustum::setProjection(pv);
 }
+// NOTE: if we don't specify noinline here, LLVM inlines this huge function into
+// two (?!) version of the Frustum(const mat4f& pv) constructor!
 
 UTILS_NOINLINE
 void Frustum::setProjection(const mat4f& pv) {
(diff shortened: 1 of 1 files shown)


abhi27033 (commented):
Normalize the plane vectors outside the loop: Instead of normalizing each plane vector individually, you can normalize them all at once after the loop that calculates them. This reduces redundant calculations and improves performance.

Use a loop for copying planes: Instead of manually copying each plane to the output array, you can use a loop to copy them. This simplifies the code and makes it more concise.

Use dot product function: Instead of explicitly calculating the dot product using dot(mPlanes[i].xyz, p), you can use the dot function from the Filament math library.
This enhances readability and reduces code duplication.

romainguy:
This code was written explicitly this way, and checked against the compiled assembly code to make sure we get proper SIMD output from the compiler. Thank you for your submission but we'd rather keep the code as-is, especially since the proposal doesn't make the code easier to read/maintain (there are only ever 6 planes anyway and this won't change).

PR #6965, opened 2023-07-19
Support EXT_clip_cull_distance for future use

bejado:
This PR sets up the ability for shaders to use `gl_ClipDistance`, which will be needed in the future. Desktop GL supports this natively. OpenGL ES requires the EXT_clip_cull_distance extension.

Unfortunately glslang does not support this extension, so we have to employ a workaround for mobile when going through glslang. We instead write to `filament_gl_ClipDistance`, and then modify the SPIR-V to decorate this as `gl_ClipDistance`. See the comment in SpirvFixup.h.

Note this PR does not actually use `gl_ClipDistance` yet, so there should be no change to shaders.

### DIFF
--- a/filament/backend/src/opengl/OpenGLContext.cpp
+++ b/filament/backend/src/opengl/OpenGLContext.cpp
@@ -453,11 +453,7 @@ void OpenGLContext::setDefaultState() noexcept {
     }
 
     if (ext.EXT_clip_cull_distance) {
-#if defined(BACKEND_OPENGL_VERSION_GL)
         glEnable(GL_CLIP_DISTANCE0);
-#elif defined(GL_CLIP_DISTANCE0_EXT)
-        glEnable(GL_CLIP_DISTANCE0_EXT);
-#endif
     }
 }
 
--- a/filament/backend/src/opengl/OpenGLContext.cpp
+++ b/filament/backend/src/opengl/OpenGLContext.cpp
@@ -451,6 +451,14 @@ void OpenGLContext::setDefaultState() noexcept {
         glClipControlEXT(GL_LOWER_LEFT_EXT, GL_ZERO_TO_ONE_EXT);
 #endif
     }
+
+    if (ext.EXT_clip_cull_distance) {
+#if defined(BACKEND_OPENGL_VERSION_GL)
+        glEnable(GL_CLIP_DISTANCE0);
+#elif defined(GL_CLIP_DISTANCE0_EXT)
+        glEnable(GL_CLIP_DISTANCE0_EXT);
+#endif
+    }
 }
 
 #ifdef BACKEND_OPENGL_VERSION_GLES
@@ -469,6 +477,7 @@ void OpenGLContext::initExtensionsGLES() noexcept {
     using namespace std::literals;
     ext.APPLE_color_buffer_packed_float = exts.has("GL_APPLE_color_buffer_packed_float"sv);
     ext.EXT_clip_control = exts.has("GL_EXT_clip_control"sv);
+    ext.EXT_clip_cull_distance = exts.has("GL_EXT_clip_cull_distance"sv);
     ext.EXT_color_buffer_float = exts.has("GL_EXT_color_buffer_float"sv);
     ext.EXT_color_buffer_half_float = exts.has("GL_EXT_color_buffer_half_float"sv);
     ext.EXT_debug_marker = exts.has("GL_EXT_debug_marker"sv);
@@ -537,6 +546,7 @@ void OpenGLContext::initExtensionsGL() noexcept {
     ext.ARB_shading_language_packing = exts.has("GL_ARB_shading_language_packing"sv);
     ext.EXT_color_buffer_float = true;  // Assumes core profile.
     ext.EXT_color_buffer_half_float = true;  // Assumes core profile.
+    ext.EXT_clip_cull_distance = true;
     ext.EXT_debug_marker = exts.has("GL_EXT_debug_marker"sv);
     ext.EXT_discard_framebuffer = false;
     ext.EXT_disjoint_timer_query = true;
--- a/libs/filamat/src/GLSLPostProcessor.cpp
+++ b/libs/filamat/src/GLSLPostProcessor.cpp
@@ -616,9 +616,11 @@ void GLSLPostProcessor::fixupClipDistance(
     if (!config.usesClipDistance) {
         return;
     }
+    // This should match the version of SPIR-V used in GLSLTools::prepareShaderParser.
     SpirvTools tools(SPV_ENV_UNIVERSAL_1_3);
     std::string disassembly;
-    tools.Disassemble(spirv, &disassembly);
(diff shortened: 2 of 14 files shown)


poweifeng (approved):

romainguy (approved):

pixelflinger:
@bejado can we ask glslang for support?

filament/backend/src/opengl/OpenGLContext.cpp
```diff
@@ -449,6 +449,14 @@ void OpenGLContext::setDefaultState() noexcept {
...
 #elif defined(GL_EXT_clip_control)
         glClipControlEXT(GL_LOWER_LEFT_EXT, GL_ZERO_TO_ONE_EXT);
+#endif
+    }
+
+    if (ext.EXT_clip_cull_distance) {
+#if defined(BACKEND_OPENGL_VERSION_GL)
+        glEnable(GL_CLIP_DISTANCE0);
+#elif defined(GL_CLIP_DISTANCE0_EXT)
+        glEnable(GL_CLIP_DISTANCE0_EXT);
```
pixelflinger, line 459: I like to  `#define` the core defines to the ext ones instead. See, `gl_headers.h` where you'd do:

```#define GL_CLIP_DISTANCE0 GL_CLIP_DISTANCE0_EXT```

That's how I did it for other extensions.

File: libs/filamat/src/GLSLPostProcessor.cpp [hunk 2]
```diff
@@ -331,6 +333,10 @@ bool GLSLPostProcessor::process(const std::string& inputShader, Config const& co
     const char* shaderCString = inputShader.c_str();
     tShader.setStrings(&shaderCString, 1);
 
+    // This allows shaders to query if they will be run through glslang.
+    // OpenGL shaders without optimization, for example, won't have this define.
+    tShader.setPreamble("#define FILAMENT_GLSLANG 1\n");
```
pixelflinger, line 338: It should probable not be defined to a value, since it should be tested with #ifdef / #ifndef

bejado:
> @bejado can we ask glslang for support?

There's already an issue open for this: https://github.com/KhronosGroup/glslang/issues/566. It's been inactive for quite awhile though.

libs/filamat/src/shaders/CodeGenerator.cpp
```diff
@@ -40,7 +40,7 @@ io::sstream& CodeGenerator::generateSeparator(io::sstream& out) {
 }
 
 utils::io::sstream& CodeGenerator::generateProlog(utils::io::sstream& out, ShaderStage stage,
-        MaterialInfo const& material) const {
+        MaterialInfo const& material, filament::Variant v) const {
```
pixelflinger, line 43: Should we pass the variant here, or should it be a field in `MaterialInfo`?

shaders/src/varyings.glsl
```diff
@@ -32,3 +32,17 @@ LAYOUT_LOCATION(11) VARYING highp vec4 vertex_lightSpacePosition;
...
 
 // Note that fragColor is an output and is not declared here; see main.fs and depth_main.fs
+
+#if 0 // only enable for variants that need gl_ClipDistance
+#if defined(GL_ES) && defined(FILAMENT_GLSLANG)
+// On ES, gl_ClipDistance is not a built-in, so we have to rely on EXT_clip_cull_distance
+// However, this extension is not supported by glslang, so we instead write to
+// filament_gl_ClipDistance, which will get decorated at the SPIR-V stage to refer to the built-in.
+// The location here does not matter, so long as it doesn't conflict with others.
+LAYOUT_LOCATION(100) out float filament_gl_ClipDistance[1];
```
pixelflinger, line 42: why `100` ?

pixelflinger (changes requested):

libs/filamat/src/shaders/CodeGenerator.cpp
bejado: I thought about that, but `MaterialInfo` contains information about the material as a whole, while use of `gl_ClipDistance` will depend on the specific variant.

shaders/src/varyings.glsl
bejado: It's arbitrary and will be replaced at the SPIR-V stage. I just picked a large number.

pixelflinger:
> > @bejado can we ask glslang for support?
> 
> There's already an issue open for this: [KhronosGroup/glslang#566](https://github.com/KhronosGroup/glslang/issues/566). It's been inactive for quite awhile though.

@bejado would you try to see if it would be hard to implement, maybe we can contribute? Maybe reach out to John Kessenich on github to see if he can give some pointers? 
I feel like the implementation must be there for desktop, it's probably just a matter of activating it when the extension is used?

Also, a question I have is how is this going to work on a device that doesn't have the extension? Is the idea that there will be a query and filament won't use that variant if not supported?

bejado:
> > > @bejado can we ask glslang for support?
> > 
> > 
> > There's already an issue open for this: [KhronosGroup/glslang#566](https://github.com/KhronosGroup/glslang/issues/566). It's been inactive for quite awhile though.
> 
> @bejado would you try to see if it would be hard to implement, maybe we can contribute? Maybe reach out to John Kessenich on github to see if he can give some pointers? I feel like the implementation must be there for desktop, it's probably just a matter of activating it when the extension is used?
> 
> Also, a question I have is how is this going to work on a device that doesn't have the extension? Is the idea that there will be a query and filament won't use that variant if not supported?

I can reach out to Khronos to see if we can contribute, but I prefer we merge this workaround now to unblock our 1P clients. If we get the extension implemented, then we can revert this.

OpenGL ES Devices that don't have the extension won't be able to use features that depend on `gl_ClipSpace` (namely, stereo rendering).

libs/filamat/src/GLSLPostProcessor.cpp
```diff
@@ -590,6 +611,21 @@ void GLSLPostProcessor::optimizeSpirv(OptimizerPtr optimizer, SpirvBlob& spirv)
...
 }
 
+void GLSLPostProcessor::fixupClipDistance(
+        SpirvBlob& spirv, GLSLPostProcessor::Config const& config) const {
+    if (!config.usesClipDistance) {
+        return;
+    }
+    SpirvTools tools(SPV_ENV_UNIVERSAL_1_3);
+    std::string disassembly;
+    tools.Disassemble(spirv, &disassembly);
```
pixelflinger, line 621: can this fails? (e.g. if in the future UNIVERSAL_1_3 is not the right constant).

pixelflinger (approved):

libs/filamat/src/GLSLPostProcessor.cpp
bejado: It can, I'll add an assertion and comment that it should match the version we use here:

https://github.com/google/filament/blob/731dd761d9fb6b97c557d4ba6c8e77a9b0babca8/libs/filamat/src/sca/GLSLTools.cpp#L443

PR #6967, opened 2023-07-19
Implement preliminary support for instanced stereo

bejado:
This PR introduces side-by-side instanced stereo rendering. The changes include the addition of a new stereo material variant and a new `setStereoscopicOptions` API on `View`. Additionally, new APIs have been added to `Camera` to position the eyes and configure their projections.

This feature is still in development; post-processing, picking, shadowing, and punctual lighting are not functional.


filament/include/filament/Camera.h
```diff
@@ -140,6 +143,18 @@ namespace filament {
...
  *
+ * Stereoscopic rendering
+ * ======================
+ *
+ * The Camera's transform (as set by setModelMatrix or via TransformManager) defines a "head" space,
+ * which typically corresponds to the location of the viewer's head. Each eye's transform is set
+ * relative to this head space by setEyeModelMatrix.
+ *
+ * Each eye also maintains its own projection matrix. These can be set with setCustomEyeProjection.
+ * Care must taken to correctly set the projectionForCulling matrix, as well as its corresponding
```
pixelflinger, line 154:
```suggestion
 * Care must be taken to correctly set the projectionForCulling matrix, as well as its corresponding
```

filament/include/filament/Camera.h
```diff
@@ -350,6 +378,25 @@ class UTILS_PUBLIC Camera : public FilamentAPI {
...
+     * eyes[1] = mat4::translation(double3{ 0.03, 0.0, 0.0});
+     * camera.setEyeModelMatrix(eyes);
+     * ~~~~~~~~~~~
+     *
+     * This method is not intended to be called every frame. Instead, to update the position of the
+     * head, use Camera::setModelMatrix.
+     *
+     * @param model an array of two eye model matrices, the first, Eye 0, the second, Eye 1.
+     */
+    void setEyeModelMatrix(math::mat4 const (&model)[2]) noexcept;
```
pixelflinger, line 398: I would prefer an API that takes an "eye id" instead of an array. e.g.:

`setEyeModelMatrix(uint8_t eyeId, mat4& model)`

filament/include/filament/Camera.h
```diff
@@ -234,6 +249,19 @@ class UTILS_PUBLIC Camera : public FilamentAPI {
...
+     *
+     * The projectionForCulling, near, and far parameters establish a "culling frustum" which must
+     * encompass anything either eye can see.
+     *
+     * @param projectionForCulling custom projection matrix for culling, must encompass both eyes
+     * @param near distance in world units from the camera to the culling near plane. \p near > 0.
+     * @param far distance in world units from the camera to the culling far plane. \p far > \p near.
+     * @see setCustomProjection
+     */
+    void setCustomEyeProjection(math::mat4 const (&projection)[2],
```
pixelflinger, line 262: Not a fan of the reference to an array, Instead I think it would be better to pass a `mat4 const*` and a count. You can even say that "count *must* be 2".

filament/src/RenderPass.cpp
```diff
@@ -522,6 +523,13 @@ RenderPass::Command* RenderPass::generateCommandsImpl(uint32_t extraFlags,
                 soaInstanceInfo[i].count | PrimitiveInfo::USER_INSTANCE_MASK;
         cmdColor.primitive.instanceBufferHandle = soaInstanceInfo[i].handle;
 
+        // soaInstanceInfo[i].count is the number of instances the user has requested, either for
+        // manual or hybrid instancing. Instanced stereo doubles the number of instances.
+        if (hasInstancedStereo) {
```
pixelflinger, line 528:
```suggestion
        if (UTILS_UNLIKELY(hasInstancedStereo)) {
```

filament/src/RenderPass.cpp
```diff
@@ -541,6 +549,11 @@ RenderPass::Command* RenderPass::generateCommandsImpl(uint32_t extraFlags,
             cmdDepth.primitive.instanceBufferHandle = soaInstanceInfo[i].handle;
             cmdDepth.primitive.materialVariant.setSkinning(hasSkinningOrMorphing);
             cmdDepth.primitive.rasterState.inverseFrontFaces = inverseFrontFaces;
+
+            if (hasInstancedStereo) {
```
pixelflinger, line 553:
```suggestion
            if (UTILS_UNLIKELY(hasInstancedStereo)) {
```

filament/src/details/Camera.cpp
```diff
@@ -142,7 +152,7 @@ void UTILS_NOINLINE FCamera::setProjection(Camera::Projection projection,
     FCamera::setCustomProjection(p, c, near, far);
 }
 
-math::mat4 FCamera::getProjectionMatrix() const noexcept {
+math::mat4 FCamera::getProjectionMatrix(int eye) const noexcept {
```
pixelflinger, line 155: `eye` should definitely be unsigned. Use `uint8_t` or `size_t`. Also `assert_invariant` that it's 0 or 1.

filament/src/details/Camera.cpp
```diff
@@ -177,6 +187,11 @@ void UTILS_NOINLINE FCamera::setModelMatrix(const mat4& modelMatrix) noexcept {
     transformManager.setTransform(transformManager.getInstance(mEntity), modelMatrix);
 }
 
+void UTILS_NOINLINE FCamera::setEyeModelMatrix(math::mat4 const (&model)[2]) noexcept {
```
pixelflinger, line 190: Wondering if this should be `setEyeViewMatrix` instead (and we don't inverse()), since that's what we need? Not sure.

filament/src/details/Camera.cpp
```diff
@@ -204,7 +219,7 @@ void FCamera::setExposure(float aperture, float shutterSpeed, float sensitivity)
 }
 
 double FCamera::getFocalLength() const noexcept {
-    return (FCamera::SENSOR_SIZE * mProjection[1][1]) * 0.5;
+    return (FCamera::SENSOR_SIZE * mEyeProjection[0][1][1]) * 0.5;
```
pixelflinger, line 222: this deserves a comment and I would use a local variable for projection[0]. Probably need to update the documentation of getFocalLength.

filament/src/details/Camera.cpp
```diff
@@ -260,7 +275,11 @@ math::details::TMat44<T> inverseProjection(const math::details::TMat44<T>& p) no
 // ------------------------------------------------------------------------------------------------
 
 CameraInfo::CameraInfo(FCamera const& camera) noexcept {
-    projection         = mat4f{ camera.getProjectionMatrix() };
+    projection         = mat4f{ camera.getProjectionMatrix(0) };
+    eyeProjection[0]   = mat4f{ camera.getProjectionMatrix(0) };
```
pixelflinger, line 279: do we need both `projection` and `eyeProjection[0]` ? that's going to be confusing. Maybe we need to bite the bullet and change `projection` everywhere?

filament/src/details/Camera.h
```diff
@@ -73,13 +78,15 @@ class FCamera : public Camera {
...
+    math::mat4 getProjectionMatrix(int eye = 0) const noexcept;
 
     // culling the projection matrix to be used for culling, contains scaling/shift
     math::mat4 getCullingProjectionMatrix() const noexcept;
 
+    math::mat4 getEyeFromViewMatrix(int eye) const noexcept { return mEyeFromView[eye]; }
+
     // viewing projection matrix set by the user
-    math::mat4 getUserProjectionMatrix() const noexcept { return mProjection; }
+    math::mat4 getUserProjectionMatrix() const noexcept { return mEyeProjection[0]; }
```
pixelflinger, line 89: That seems wrong. Now that the user can specify 2 projection matrices, this should be able to return both. This method is used in the public API, that's what `getProjectionMatrix` calls.

filament/src/details/Camera.h
```diff
@@ -205,10 +215,19 @@ struct CameraInfo {
     explicit CameraInfo(FCamera const& camera) noexcept;
     CameraInfo(FCamera const& camera, const math::mat4& worldOriginCamera) noexcept;
 
-    math::mat4f projection;         // projection matrix for drawing (infinite zfar)
+    int eyeCount = 1;
```
pixelflinger, line 218: should be unsigned. Maybe move it to the end of the structure, to try to avoid padding.

filament/src/details/Camera.h
```diff
@@ -205,10 +215,19 @@ struct CameraInfo {
...
+
+    // projection matrix for drawing (infinite zfar)
+    // for monoscopic rendering
+    math::mat4f projection;
+
+    // for stereo rendering, one matrix per eye
+    math::mat4f eyeProjection[2]  = {};
```
pixelflinger, line 225: don't think we should have both. that's a wasted mat4f

filament/src/materials/defaultMaterial0.mat
```diff
@@ -1,7 +1,8 @@
 material {
     name : "Filament Default Material",
     shadingModel : unlit,
-    featureLevel: 0
+    featureLevel: 0,
+    variantFilter : [ stereo ]
```
pixelflinger, line 5: I'm starting to wonder if we shouldn't have a `variant` parameter that **adds** a variant. we could make `stereo` off by default.  otherwise I feel like 99% of the apps will need this filter.

libs/filabridge/include/filament/MaterialEnums.h
```diff
@@ -243,7 +243,8 @@ enum class UserVariantFilterBit : UserVariantFilterMask {
     FOG                         = 0x10,
     VSM                         = 0x20,
     SSR                         = 0x40,
-    ALL                         = 0x7F,
+    STE                         = 0x80,
```
pixelflinger, line 246: ouch. but yeah, can't think of a better way.

libs/filabridge/include/private/filament/UibStructs.h
```diff
@@ -81,6 +81,15 @@ struct PerViewUib { // NOLINT(cppcoreguidelines-pro-type-member-init)
     math::mat4f userWorldFromWorldMatrix;   // userWorld <- world
     math::float4 clipTransform;             // [sx, sy, tx, ty] only used by VERTEX_DOMAIN_DEVICE
 
+    // --------------------------------------------------------------------------------------------
+    // Stereoscopic rendering
+    // --------------------------------------------------------------------------------------------
+
+    math::mat4f eyeFromWorldMatrix[4];
+    math::mat4f clipFromEyeMatrix[4];
```
pixelflinger, line 89: why 4 matrices, it seems like the filament side API is limited to 2?

libs/filabridge/include/private/filament/Variant.h
```diff
@@ -50,32 +50,34 @@ struct Variant {
     // FOG: Fog
     // PCK: Picking (depth variant only)
     // VSM: Variance shadow maps
+    // STE: Instanced stereo rendering
     //
     //   X: either 1 or 0
     //                               1     0    0      0     1     0     1
     //                      +-----+-----+-----+-----+-----+-----+-----+-----+
-    // Variant              |  0  | VSM | FOG | DEP | SKN | SRE | DYN | DIR |   128
+    // Variant              | STE | VSM | FOG | DEP | SKN | SRE | DYN | DIR |   128
```
pixelflinger, line 58:
```suggestion
    // Variant              | STE | VSM | FOG | DEP | SKN | SRE | DYN | DIR |   256
```

shaders/src/common_getters.glsl
```diff
@@ -24,7 +24,12 @@ highp mat4 getViewFromClipMatrix() {
 
 /** @public-api */
 highp mat4 getClipFromWorldMatrix() {
+#if defined(VARIANT_HAS_INSTANCED_STEREO)
+    int eye = instance_index % 2;
+    return frameUniforms.clipFromEyeMatrix[eye] * frameUniforms.eyeFromWorldMatrix[eye];
```
pixelflinger, line 29: I'm assuming we do need both matrices in the shader and that's why we can pre-concatenate them (unless we create even more matrices).

libs/filabridge/include/private/filament/Variant.h
```diff
@@ -50,32 +50,34 @@ struct Variant {
...
     //                      +-----+-----+-----+-----+-----+-----+-----+-----+
     //                                    PCK
     //
     // Standard variants:
     //                      +-----+-----+-----+-----+-----+-----+-----+-----+
-    //                      |  0  | VSM | FOG |  0  | SKN | SRE | DYN | DIR |    64 - 22 = 42
+    //                      | STE | VSM | FOG |  0  | SKN | SRE | DYN | DIR |    64 - 22 = 42
```
pixelflinger, line 64:
```suggestion
    //                      | STE | VSM | FOG |  0  | SKN | SRE | DYN | DIR |    128 - 44 = 84
```

libs/filabridge/include/private/filament/Variant.h
```diff
@@ -50,32 +50,34 @@ struct Variant {
...
     //                      +-----+-----+-----+-----+-----+-----+-----+-----+
-    //                      |  0  | VSM | PCK |  1  | SKN |  0  |  0  |  0  |   8 - 2 = 6
+    //                      | STE | VSM | PCK |  1  | SKN |  0  |  0  |  0  |   8 - 2 = 6
     //                      +-----+-----+-----+-----+-----+-----+-----+-----+
-    //       Vertex depth            X     0     1     X     0     0     0
-    //     Fragment depth            X     X     1     0     0     0     0
-    //           Reserved            1     1     1     X     0     0     0     [ -2]
+    //       Vertex depth      X     X     0     1     X     0     0     0
+    //     Fragment depth      0     X     X     1     0     0     0     0
+    //           Reserved      X     1     1     1     X     0     0     0     [  -8]
```
pixelflinger, line 79:
```suggestion
    //           Reserved      X     1     1     1     X     0     0     0     [  -4]
```

libs/filabridge/include/private/filament/Variant.h
```diff
@@ -50,32 +50,34 @@ struct Variant {
...
+    //           Reserved      X     X     X     1     X     X     X     X     [-128]
```
pixelflinger, line 80:
```suggestion
```

libs/filabridge/include/private/filament/Variant.h
```diff
@@ -50,32 +50,34 @@ struct Variant {
...
+    //    Fragment shader      0     X     X     0     0     X     X     X
+    //       Fragment SSR      0     1     0     0     0     1     0     0
+    //           Reserved      X     1     1     0     X     1     0     0      [ -4]
+    //           Reserved      X     0     X     0     X     1     0     0      [ -8]
+    //           Reserved      X     1     X     0     X     0     X     X      [-32]
     //
     // Depth variants:
     //                      +-----+-----+-----+-----+-----+-----+-----+-----+
-    //                      |  0  | VSM | PCK |  1  | SKN |  0  |  0  |  0  |   8 - 2 = 6
+    //                      | STE | VSM | PCK |  1  | SKN |  0  |  0  |  0  |   8 - 2 = 6
```
pixelflinger, line 75:
```suggestion
    //                      | STE | VSM | PCK |  1  | SKN |  0  |  0  |  0  |   16 - 4 = 12
```

pixelflinger (changes requested):

libs/filabridge/include/private/filament/UibStructs.h
bejado: Yup, I think this was a typo on my end.

filament/include/filament/Camera.h
bejado: I think if you prefer an "eye id" API, then we can do something similar to `setEyeModelMatrix`:

```
    void setCustomEyeProjection(uint8_t eyeId, math::mat4& projection),
```

filament/src/details/Camera.cpp
bejado: We could have both?

filament/src/materials/defaultMaterial0.mat
bejado: Let me experiment with that and see how feasible it would be.

shaders/src/common_getters.glsl
bejado: I think we eventually want both `clipFromEyeMatrix` and `eyeFromWorldMatrix`. This way, a user can get "eye" space through the `eyeFromWorldMatrix`.

filament/include/filament/Camera.h
pixelflinger: yes I'd like that, but this method also has parameters that are for head, so not related to a given eye. if we can resolve that it would be nice.

filament/src/materials/defaultMaterial0.mat
bejado: I think we can add a new `variantInclude` parameter.

Also, I realized I need to update MaterialBuilder to filter out the stereo variant automatically for feature level 0. So this change isn't strictly necessary.

filament/include/filament/Camera.h
bejado: Ah right, which is I think why I wrote it this way. I'll switch to `mat4 const*` for now.

filament/src/details/Camera.cpp
bejado: Actually, I think it makes sense to only offer `setEyeModelMatrix`. We don't let users set the regular camera view matrix directly either.

filament/src/details/Camera.cpp
```diff
@@ -274,7 +300,11 @@ CameraInfo::CameraInfo(FCamera const& camera) noexcept {
 
 CameraInfo::CameraInfo(FCamera const& camera, const math::mat4& worldOriginCamera) noexcept {
     const mat4 modelMatrix{ worldOriginCamera * camera.getModelMatrix() };
-    projection         = mat4f{ camera.getProjectionMatrix() };
+    for (uint8_t i = 0; i < CONFIG_STEREOSCOPIC_EYES; i++) {
+        eyeProjection[i]   = mat4f{ camera.getProjectionMatrix(i) };
+    }
+    eyeFromView[0]     = mat4f{ camera.getEyeFromViewMatrix(0) };
+    eyeFromView[1]     = mat4f{ camera.getEyeFromViewMatrix(1) };
```
pixelflinger, line 307: Why not put this in the loop too?

filament/src/details/Camera.h
```diff
@@ -73,13 +81,19 @@ class FCamera : public Camera {
 
     // viewing the projection matrix to be used for rendering, contains scaling/shift and possibly
     // other transforms needed by the shaders
-    math::mat4 getProjectionMatrix() const noexcept;
+    math::mat4 getProjectionMatrix(uint8_t eye = 0) const noexcept;
 
     // culling the projection matrix to be used for culling, contains scaling/shift
     math::mat4 getCullingProjectionMatrix() const noexcept;
 
+    math::mat4 getEyeFromViewMatrix(int eye) const noexcept { return mEyeFromView[eye]; }
```
pixelflinger, line 89: sometimes `eye` is a `uint8_t` and sometimes it's a `int`. We should be consistant. I think it should be unsigned at least.

filament/src/details/Camera.h
```diff
@@ -73,13 +81,19 @@ class FCamera : public Camera {
...
+
     // viewing projection matrix set by the user
-    math::mat4 getUserProjectionMatrix() const noexcept { return mProjection; }
+    math::mat4 getUserProjectionMatrix(uint8_t eyeId) const {
```
pixelflinger, line 92: I don't know why it wasn't before, but the return value should be a `const&`

libs/filabridge/src/Variant.cpp
```diff
@@ -161,9 +164,9 @@ constexpr inline size_t fragment_variant_count() noexcept {
 }
 
 static_assert(reserved_is_not_valid());
-static_assert(reserved_variant_count() == 80);
-static_assert(valid_variant_count() == 48);
-static_assert(vertex_variant_count() == 16 - (2 + 0) + 4 - 0);        // 18
+static_assert(reserved_variant_count() == 160);
+static_assert(valid_variant_count() == 96);
+static_assert(vertex_variant_count() == 32 - (4 + 0) + 8 - 0);        // 36
```
pixelflinger, line 169: I think something might be wrong here. Only the number of vertex variants should have doubled. The total number of variants shouldn't be double because STE doesn't affect fragment variant count.
`filterVariantFragment` must be updated; STE is similar to SKN in that way.

pixelflinger (changes requested):

libs/filabridge/src/Variant.cpp
bejado: All of this variant logic is very confusing to me, I don't understand where a lot of these numbers come from or how they relate to the ASCII chart in Variant.h.

For instance, why is `valid_variant_count()` 48, but `vertex_variant_count() + fragment_variant_count()` only 42?

It seems like `filterVariantFragment` should _not_ be updated, because it's `&`-ing with bits that _should_ affect the fragment shader.

bejado: Here's what the code does currently, which seems correct to me:

**No stereo**, 42 total variants:
- 18 vertex (4 DEP)
- 24 fragment (3 DEP)

**With stereo**, 60 total variants:
- 36 vertex (8 DEP)
- 24 fragment (3 DEP)

bejado: It makes sense that valid variant count doubled (which, as we discussed, is not to be confused with the total number of generated shaders), because STE is completely orthogonal to any of the other variant bits.

pixelflinger (approved):

PR #6968, opened 2023-07-20
Fix ineffective FOG variant filter

bejado:
The FOG variant filter has no effect because `UserVariantFilterBit::FOG != FOG`. Thus, this has no effect:
```
            variant.key &= ~(filterMask & FOG);
```

I'm not sure why we needed to bitwise AND the `filterMask` with the variant bit. Maybe I'm missing something.

### DIFF
--- a/NEW_RELEASE_NOTES.md
+++ b/NEW_RELEASE_NOTES.md
@@ -7,3 +7,5 @@ for next branch cut* header.
 appropriate header in [RELEASE_NOTES.md](./RELEASE_NOTES.md).
 
 ## Release notes for next branch cut
+
+- matc: fix buggy `variant-filter` flag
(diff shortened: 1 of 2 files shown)


pixelflinger (approved):

pixelflinger:
> The FOG variant filter has no effect because `UserVariantFilterBit::FOG != FOG`. Thus, this has no effect:
> 
> ```
>             variant.key &= ~(filterMask & FOG);
> ```
> 
> I'm not sure why we needed to bitwise AND the `filterMask` with the variant bit. Maybe I'm missing something.

It was just stupid from my part. This is because I was thinking of doing something like `variant.key &= ~filterMask` and because the bits of the two enums don't match, I thought of "keeping" only the good ones -- but obviously this was completely wrong.

Good catch!

PR #6980, opened 2023-07-21
vulkan: fix subpass validation

### LINKED ISSUE #6513: Vulkan debug report : Input attachment descriptor image view is not a subpass input attachment
This error validation  occurs when calling vkCmdDrawIndexed function in VulkanDriver::draw:

VULKAN ERROR: (Validation) Validation Error: [ UNASSIGNED-input-attachment-descriptor-not-in-subpass ] Object 0:
 handle = 0x980f360000000011, type = VK_OBJECT_TYPE_IMAGE_VIEW; | MessageID = 0xe897be5f | Input attachment descriptor image view is not a subpass input attachment

poweifeng:
- Before, we supposed that the maximum number of input attachment should match the maximum number of color attachments. But in reality, we've only used one input attachment for the second subpass.
 - The problem with the above supposition is that the descriptor set layout for the input attachment descriptor set must have the exact number of input attachment specified in the shader. If the *layout* has more input attachment slots than specified in the shader, then we'd run into a validation error.
 - In this patch, we fix the number of max input attachment in the descriptor set layout to 1, since we ever only make use of one.

Fixes #6513

### DIFF
--- a/filament/backend/src/vulkan/VulkanPipelineCache.h
+++ b/filament/backend/src/vulkan/VulkanPipelineCache.h
@@ -58,7 +58,11 @@ public:
 
     static constexpr uint32_t UBUFFER_BINDING_COUNT = Program::UNIFORM_BINDING_COUNT;
     static constexpr uint32_t SAMPLER_BINDING_COUNT = MAX_SAMPLER_COUNT;
-    static constexpr uint32_t TARGET_BINDING_COUNT = MRT::MAX_SUPPORTED_RENDER_TARGET_COUNT;
+
+    // We assume only one possible input attachment between two subpasses. See also the subpasses
+    // definition in VulkanFboCache.
+    static constexpr uint32_t INPUT_ATTACHMENT_COUNT = 1;
+
     static constexpr uint32_t SHADER_MODULE_COUNT = 2;
     static constexpr uint32_t VERTEX_ATTRIBUTE_COUNT = MAX_VERTEX_ATTRIBUTE_COUNT;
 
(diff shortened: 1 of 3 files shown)


filament/backend/src/vulkan/VulkanPipelineCache.h
```diff
@@ -58,7 +58,11 @@ class VulkanPipelineCache : public CommandBufferObserver {
 
     static constexpr uint32_t UBUFFER_BINDING_COUNT = Program::UNIFORM_BINDING_COUNT;
     static constexpr uint32_t SAMPLER_BINDING_COUNT = MAX_SAMPLER_COUNT;
-    static constexpr uint32_t TARGET_BINDING_COUNT = MRT::MAX_SUPPORTED_RENDER_TARGET_COUNT;
+
+    // We assume only one possible input attachment between two subpasses. See also the subpasses
+    // definition in VulkanFboCache.
+    static constexpr uint32_t INPUT_ATTACHMENT_COUNT = 1;
+
```
pixelflinger, line 65: what would need to change so that we break that assumption? I suppose the material knows how many inputs it has, so that should be passed to the renderpass or something like that?

pixelflinger (approved):

filament/backend/src/vulkan/VulkanPipelineCache.h
poweifeng: In the really general case, we'd need to know how many subpasses, and the attachment -> input mapping between consecutive subpasses at the start of the renderpass. 

right now we assume we have at most one additional subpass, and the sole input attachment to the second subpass is always the first color attachment.

PR #6988, opened 2023-07-24
Fix several issues with parallel material compilation

pixelflinger:
- fix an issue causing callbacks to be called too late
- fix cleanup of material parallel compilation
- Add a way to query the validity of filament objects

### DIFF
--- a/filament/src/details/Engine.cpp
+++ b/filament/src/details/Engine.cpp
@@ -465,7 +465,7 @@ void FEngine::shutdown() {
         getDriverApi().terminate();
     } else {
         mDriverThread.join();
-
+        // Driver::terminate() has been called here.
     }
 
     // Finally, call user callbacks that might have been scheduled.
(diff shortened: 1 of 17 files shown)


filament/src/details/Engine.cpp
```diff
@@ -852,6 +852,12 @@ void FEngine::cleanupResourceListLocked(Lock& lock, ResourceList<T>&& list) {
 
 // -----------------------------------------------------------------------------------------------
 
+template<typename T>
+UTILS_ALWAYS_INLINE
+inline bool FEngine::isValid(const T* ptr, ResourceList<T>& list) {
+    return (ptr == nullptr) || (list.find(ptr) != list.end());
```
poweifeng, line 858: a null pointer is a valid resource?

poweifeng (approved):

filament/src/details/Engine.cpp
pixelflinger: > a null pointer is a valid resource?

good catch, I wrongly reused the code for `destroy` (where nullptr is considered ok, like `free()`)

bejado (approved):

PR #6996, opened 2023-07-26
Fix broken links in web tutorials

pulkitgarg784:


poweifeng (commented):
I have no problem navigating the pages here: https://google.github.io/filament/webgl/tutorial_redball.html

Why this change?

PR #7028, opened 2023-08-02
vulkan: fix TSAN in readpixels (#7023)

poweifeng:


poweifeng:
cherry-picking to rc/1.40.5

Issue #6933, opened 2023-07-01
Android ModelViewer Crash On Quick Creation and Detach


JustinBis:
**Describe the bug**
Filament's `ModelViewer` helper for Android will crash if you detach the connected view quickly after creation.

I narrowed this down to the engine destruction within the detach handler used in ModelViewer:

```
// In filament-android's ModelViewer.kt
    private fun addDetachListener(view: android.view.View) {
        view.addOnAttachStateChangeListener(object : android.view.View.OnAttachStateChangeListener {
            override fun onViewAttachedToWindow(v: android.view.View) {}
            override fun onViewDetachedFromWindow(v: android.view.View) {
                uiHelper.detach()
                ...
                // This is the line that ultimately causes the crash
                engine.destroy()
            }
        })
    }
```

**To Reproduce**
1. Initialize a ModelViewer with a SurfaceView in some Fragment that you can quickly destroy, such as a bottom sheet
2. Quickly destroy the Fragment (e.g. dismiss the bottom sheet)
3. Observe the crash below

**Expected behavior**
Creating and immediately removing the view associated with a ModelViewer should not crash.

**Logs**
```
2023-07-01 13:15:12.526 25027-25027 libc                    ai.nokto.wire                        A  Fatal signal 6 (SIGABRT), code -1 (SI_QUEUE) in tid 25027 (ai.nokto.wire), pid 25027 (ai.nokto.wire)
2023-07-01 13:15:12.527  1018-24974 ExynosVide...Enc-Vp8Enc sam...hardware.media.c2@1.0-service  E  [updateAverageQpInfoToParam] Get_AverageQp() is failed
2023-07-01 13:15:12.545  1018-24974 ExynosVide...Enc-Vp8Enc sam...hardware.media.c2@1.0-service  E  [updateAverageQpInfoToParam] Get_AverageQp() is failed
---------------------------- PROCESS STARTED (25123) for package ai.nokto.wire ----------------------------
2023-07-01 13:15:12.834 25121-25121 DEBUG                   crash_dump64                         A  *** *** *** *** *** *** *** *** *** *** *** *** *** *** *** ***
2023-07-01 13:15:12.834 25121-25121 DEBUG                   crash_dump64                         A  Build fingerprint: 'google/bluejay/bluejay:13/TQ3A.230605.010/10121037:user/release-keys'
2023-07-01 13:15:12.834 25121-25121 DEBUG                   crash_dump64                         A  Revision: 'MP1.0'
[...]

PR #7036, opened 2023-08-03
fix a crash when shutting the engine down

pixelflinger:
We were breaking the promise of pending shader compilation jobs by  destroying the corresponding std::promise embedded in the job queue.  In practice there was no danger of a deadlock by construction, but std::promise throws an exception in that case. On builds without exception enabled, this we be turned into an abort().

We fix this by using our own mechanism for signaling instead of std::promise. This ends up be more lightweight anyways.

Fix: #6933


bejado (approved):

filament/backend/src/opengl/ShaderCompilerService.cpp
```diff
@@ -346,10 +378,9 @@ ShaderCompilerService::program_token_t ShaderCompilerService::createProgram(
                     // TODO: see if we could completely eliminate this callback here
                     //       and instead just rely on token->gl.program being atomically
                     //       set by the compiler thread.
-                    assert_invariant(token->binary.valid());
                     // we're using the compiler thread, check if the program is ready, no-op if not.
                     using namespace std::chrono_literals;
```
poweifeng, line 382: unused?

poweifeng (approved):

PR #7062, opened 2023-08-10
support PowerVR GPUs better

pixelflinger:
This PR contains mostly workarounds for PowerVR compiler bugs but also some improvements to `ShaderCompilerService`.

### DIFF
--- a/shaders/src/common_instancing.glsl
+++ b/shaders/src/common_instancing.glsl
@@ -9,6 +9,10 @@ highp int object_uniforms_flagsChannels;                   // see packFlags() be
 highp int object_uniforms_objectId;                        // used for picking
 highp float object_uniforms_userData;   // TODO: We need a better solution, this currently holds the average local scale for the renderable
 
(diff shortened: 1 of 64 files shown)


shaders/src/common_instancing.glsl
```diff
@@ -17,45 +21,35 @@ void initObjectUniforms() {
...
+    // instancing handled by the material
+    if ((objectUniforms.data[0].flagsChannels & FILAMENT_OBJECT_INSTANCE_BUFFER_BIT) != 0) {
+        // hybrid instancing, we have a instance buffer per object
+        i = instance_index;
+    } else {
+        // fully manual instancing
+        i = 0;
     }
+#   else
+    // automatic instancing
```
bejado, line 35: This could be hybrid-instancing too, if the material doesn't specify `instanced: true`.

bejado (approved):

poweifeng (approved):

PR #7075, opened 2023-08-16
rework ShaderCompilerService to improve performance

pixelflinger:
- remove support for non-shared contextes parallel compilation. this wasn't used. we can always revive it later if we need to.

- rework how callbacks work so that we don't have to use a work list executed at each tick() in the shared context case (common case). this improves performance significantly on low-end devices, by not having to go through the list to check if all programs are compiled, multiple times per frame.

The new CallbackManager handles scheduling the callbacks after all previous programs are compiled.

### DIFF
--- a/filament/backend/src/opengl/CallbackManager.cpp
+++ b/filament/backend/src/opengl/CallbackManager.cpp
@@ -1,0 +1,69 @@
+/*
+ * Copyright (C) 2023 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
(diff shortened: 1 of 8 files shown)


filament/backend/src/opengl/CallbackManager.cpp
```diff
@@ -0,0 +1,67 @@
...
+CallbackManager::CallbackManager(DriverBase& driver) noexcept
+    : mDriver(driver), mCallbacks(1) {
+}
+
+CallbackManager::~CallbackManager() noexcept = default;
+
+void CallbackManager::terminate() noexcept {
+    for (auto&& item: mCallbacks) {
+        if (item.callback.func) {
+            mDriver.scheduleCallback(
```
bejado, line 32: Does `terminate` unconditionally call all callbacks, even if their conditions haven't been met? Maybe document that behavior

filament/backend/src/opengl/CallbackManager.h
```diff
@@ -0,0 +1,99 @@
...
+    struct Callback {
+        CallbackHandler* handler = nullptr;
+        CallbackHandler::Callback func = {};
+        void* user = nullptr;
+    };
+
+    struct CallbackList {
+        mutable std::atomic_int count{};
+        Callback callback;
+    };
```
bejado, line 50: It's a bit confusing that this is called `CallbackList`, when this is only a single callback. Can you keep the `std::atomic_int` as part of `Callback`?

pixelflinger: oh yeah, it's more the callback's list (of condition to meet) -- which is implemented as a counter here.

PR #7085, opened 2023-08-17
rework how we initialize the gl context

pixelflinger:
- early initialization is now implemented with static methods so that it's very clear which state they need.

- the version number is no longer used outside of initialization, instead we use the feature level.

- ES3.0 Adreno devices are downgraded to feature level 0

### DIFF
--- a/filament/backend/src/opengl/OpenGLContext.cpp
+++ b/filament/backend/src/opengl/OpenGLContext.cpp
@@ -49,6 +49,7 @@ bool OpenGLContext::queryOpenGLVersion(GLint* major, GLint* minor) noexcept {
 }
 
 OpenGLContext::OpenGLContext() noexcept {
+
     state.vao.p = &mDefaultVAO;
 
     // These queries work with all GL/GLES versions!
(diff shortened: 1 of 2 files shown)


romainguy (approved):

filament/backend/src/opengl/OpenGLContext.cpp
```diff
@@ -458,9 +268,287 @@ void OpenGLContext::setDefaultState() noexcept {
...
+        // Runtime OpenGL version is ES 2.x
+#       if defined(BACKEND_OPENGL_LEVEL_GLES30)
+        // mandatory extensions (all supported by Mali-400 and Adreno 304)
+        assert_invariant(exts.OES_depth_texture);
+        assert_invariant(exts.OES_depth24);
+        assert_invariant(exts.OES_packed_depth_stencil);
+        assert_invariant(exts.OES_rgb8_rgba8);
+        assert_invariant(exts.OES_standard_derivatives);
+        assert_invariant(exts.OES_texture_npot);
+#    endif
```
poweifeng, line 520:
```suggestion
#       endif
```

poweifeng (approved):

PR #7086, opened 2023-08-17
Add Engine query for stereoscopic support

bejado:
Also don't precompile STE variants when `Material::compile` is called if stereo is not supported.


filament/src/details/Renderer.cpp
```diff
@@ -417,6 +417,9 @@ void FRenderer::render(FView const* view) {
 
     assert_invariant(mSwapChain);
 
+    ASSERT_PRECONDITION(!view->hasInstancedStereo() || mEngine.getDriverApi().isStereoSupported(),
+            "Stereo rendering is not supported.");
+
```
pixelflinger, line 422: this should be checked in `setStereoscopicOptions` instead.

pixelflinger (changes requested):

filament/src/details/Renderer.cpp
bejado: I thought about that, but all of our `View.setXYZOptions` methods are `noexcept`. Are we okay breaking that convention for `setStereoscopicOptions`?

pixelflinger: yes, the noexcept predated the feature levels. I think that's the best place for these checks.

The original idea was that these setters would never fail because they would just "clamp" the parameters to something that works. But for feature levels I think we should assert/throw.

poweifeng (approved):
lgtm outside of noexcept comment.

filament/src/details/View.cpp
```diff
@@ -56,7 +56,8 @@ static constexpr float PID_CONTROLLER_Ki = 0.002f;
 static constexpr float PID_CONTROLLER_Kd = 0.0f;
 
 FView::FView(FEngine& engine)
-        : mFroxelizer(engine),
+        : mEngine(engine),
```
pixelflinger, line 59: I have resisted all this time storing an `Engine` in `View`, I don't want to add that dependency (in fact, I wish I had been more diligent about this in other classes).
So instead, you can just store the "isStereoSupported" in a `bool:1`.

pixelflinger (changes requested):

PR #7103, opened 2023-08-23
Transition setFrameCompletedCallback to take a CallbackHandler

bejado:
`setFrameCompletedCallback` should use the same `CallbackHandler` mechanism as the rest of Filament. This is particularly useful for clients to control which thread the frame completed callback is called on.


poweifeng (approved):

pixelflinger (approved):
should we do it for `setFrameScheduledCallback` as well?

bejado:
> should we do it for `setFrameScheduledCallback` as well?

Yes, I'll do it in a separate PR.

PR #7111, opened 2023-08-25
Revert "workaround another PowerVR compiler bug "

pixelflinger:
This reverts commit 58f96be2c439c0e057bfce335667958e44e2ae07.

This caused material files to increase in size significantly. It turns out that glslang has to generate a copy for each parameter that is passed to a function as a non-const parameter.

This revert will break IMG devices again, but that should be the case only on debug builds. Release builds lose the const qualifier by  virtue of going through spirv. We'll try to address this some other  way later.


bejado (approved):
Cherry-pick to rc/1.42.0?

PR #7122, opened 2023-08-29
Properly apply emissive to masked materials

romainguy:
The emissive property should not be multiplied by the color alpha in masked materials. The alpha is treated as a coverage value in that case, not an opacity value.

Before:

Screenshot 2023-08-29 at 4 54 17 PM

After:

Screenshot 2023-08-29 at 4 21 52 PM

### DIFF
--- a/NEW_RELEASE_NOTES.md
+++ b/NEW_RELEASE_NOTES.md
@@ -9,3 +9,5 @@ appropriate header in [RELEASE_NOTES.md](./RELEASE_NOTES.md).
 ## Release notes for next branch cut
 
 - `setFrameCompletedCallback` now takes a `backend::CallbackHandler`.
+- The `emissive` property was not applied properly to `MASKED` materials, and could cause
+  dark fringes to appear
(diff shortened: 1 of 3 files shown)


romainguy:
Note: other glTF renderers don't have this issue because they set `baseColor.a` to `1.0` when a fragment passes the alpha mask test. We don't do this because we want to keep an alpha value as coverage to smooth the edges. This means we can't multiply the emissive color by that alpha when we do this (but we save an instruction, so eh… that's a win).

NEW_RELEASE_NOTES.md
```diff
@@ -9,3 +9,5 @@ appropriate header in [RELEASE_NOTES.md](./RELEASE_NOTES.md).
 ## Release notes for next branch cut
 
 - `setFrameCompletedCallback` now takes a `backend::CallbackHandler`.
+- The `emissive` property was not applied properly to `MASKED` materials, and could cause
+  dark fringes to appear
```
poweifeng, line 13: recompile materials?

poweifeng (approved):

PR #7129, opened 2023-08-31
Add support for 2 additional material variables

bejado:


pixelflinger:
FYI: all ES3.0 devices have at least 15 varying vectors. 38.4% have exactly 15.
However, almost 30% of ES2 devices have 12 or less. 6% have only 8.

pixelflinger (changes requested):
limit the 2 extra variables to FEATURE_LEVEL_1 and up (i.e. not on ES2).

bejado:
As discussed, clsoing this for now

PR #7130, opened 2023-08-31
Fix possible change of scale sign when decomposing matrix

mackong:
When animate for rotation or scale path, decompose the transform matrix may change the sign of scale, and then compose the matrix back with scale or rotation changed by animation, the matrix may be wrong.

This PR save the origin scale value of node from gltf data, and then use sign of this value as new decomposed scale's sign. If no origin scale vlaue provided, still use determinant's sign.

According to gltf spec, when a node is targeted for animation (referenced by an animation.channel.target), matrix **MUST NOT** be present, so the origin scale value is always available without decompose for a animation node.

Here is a test gltf:
[animation.glb.tar.gz](https://github.com/google/filament/files/12488413/animation.glb.tar.gz)

Before:
image

After:
image

### DIFF
--- a/filament/include/filament/TransformManager.h
+++ b/filament/include/filament/TransformManager.h
@@ -223,6 +223,23 @@ public:
     children_iterator getChildrenEnd(Instance parent) const noexcept;
 
     /**
+     * Sets a scale of a transform component.
+     * This scale is preserved for sign check when decomposing transform matrix.
+     * @param ci     The instance of the transform component to set the scale to.
+     * @param scale  The preserved scale.
+     * @see getScale()
+     */
+    void setScale(Instance ci, const math::float3& scale) noexcept;
+
+    /**
+     * Returns the scale of a transform component.
+     * @param ci The instance of the transform component to query the scale.
+     * @return The scale of the component.
+     * @see setTransform()
+     */
+    const math::float3& getScale(Instance ci) const noexcept;
+
+    /**
      * Sets a local transform of a transform component.
      * @param ci              The instance of the transform component to set the local transform to.
      * @param localTransform  The local transform (i.e. relative to the parent).
--- a/filament/src/components/TransformManager.h
+++ b/filament/src/components/TransformManager.h
@@ -27,6 +27,7 @@
 #include <utils/Slice.h>
 
 #include <math/mat4.h>
+#include "math/vec3.h"
 
 namespace filament {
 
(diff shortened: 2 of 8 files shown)


filament/include/filament/TransformManager.h
```diff
@@ -222,6 +222,23 @@ class UTILS_PUBLIC TransformManager : public FilamentAPI {
...
     children_iterator getChildrenEnd(Instance parent) const noexcept;
 
+    /**
+     * Sets a scale of a transform component.
+     * This scale is preserved for sign check when decomposing transform matrix.
+     * @param ci     The instance of the transform component to set the scale to.
+     * @param scale  The preserved scale.
+     * @see getScale()
+     */
+    void setScale(Instance ci, const math::float3& scale) noexcept;
```
romainguy, line 232: I'll let @pixelflinger do the full review, but I'm pretty sure we do not want something like this on `TransformManager`.

filament/src/components/TransformManager.h
```diff
@@ -142,6 +149,7 @@ class UTILS_PRIVATE FTransformManager : public TransformManager {
         WORLD,          // world transform
         LOCAL_LO,       // accurate local translation
         WORLD_LO,       // accurate world translation
+        SCALE,          // preserved scaling for sign check when decompose matrix
```
romainguy, line 152: You are storing extra data in the ECS that doesn't seem necessary.

libs/gltfio/include/gltfio/math.h
```diff
@@ -66,8 +67,17 @@ UTILS_PUBLIC inline void decomposeMatrix(const filament::math::mat4f& mat, filam
     float scaley = length(float3({d, e, f}));
     float scalez = length(float3({g, h, i}));
     float3 s = { scalex, scaley, scalez };
-    if (det < 0) {
-        s = -s;
+    if (preservedScale != float3{}) {
+        const float signx = (preservedScale.x < 0) ? -1 : 1;
```
romainguy, line 71: You are using an entire float3 to store 3 bits of information. All you need is whether each of x/y/z was negative or not.

libs/math/include/math/TQuatHelpers.h
```diff
@@ -244,6 +244,10 @@ class TQuatFunctions {
 
     friend inline
     QUATERNION<T> MATH_PURE slerp(const QUATERNION<T>& p, const QUATERNION<T>& q, T t) {
+        if (p == q) {
```
romainguy, line 247: This is unrelated to the rest of this PR

romainguy (changes requested):

libs/math/include/math/TQuatHelpers.h
mackong: Deleted

pixelflinger:
@mackong I understand the problem you're trying to solve, but this is not the correct fix. We do not want to add a partial TRS representation to `TransformManager`. The correct solution is for *you* to write your own "TRS" transform manager, it could for instance store the translation, rotation and scaling of your transform, using the same Entity of your object.
All you'd have to do is use your new TRS manager to calculate the final mat4 and pass that to the transform manager. You could do that just before rendering, once a frame.

mackong:
Oh, i see, i'll try the "TRS" transform manager solution.

PR #7131, opened 2023-09-01
Use flatmat for mat parameters

mackong:


web/filament-js/jsbindings.cpp
```diff
@@ -1337,11 +1337,11 @@ class_<MaterialInstance>("MaterialInstance")
             (MaterialInstance* self, std::string name, filament::math::float4 value), {
         self->setParameter(name.c_str(), value); }), allow_raw_pointers())
     .function("setMat3Parameter", EMBIND_LAMBDA(void,
-            (MaterialInstance* self, std::string name, filament::math::mat3f value), {
-        self->setParameter(name.c_str(), value); }), allow_raw_pointers())
+            (MaterialInstance* self, std::string name, flatmat3 value), {
```
romainguy, line 1340: This is an API change, can you explain why you are making that change? I know we use flatmat in another places, but maybe we should have 2 APIs then, the old one for compatibility and the new one you are proposing.

mackong: Mainly because we only bound flatmat3/4 type, if use filament::math::mat3/4f, unbound type error will be occur.
 
Also, the API defined in filament.d.ts is:
```ts
export type mat3 = glm.mat3|number[];
export type mat4 = glm.mat4|number[];

export class MaterialInstance {
    public setMat3Parameter(name: string, value: mat4): void;
    public setMat4Parameter(name: string, value: mat3): void;
}
```
use flatmat3/4 is conform to this declaration. seems not a API change.

romainguy (approved):

PR #7138, opened 2023-09-04
Fix possible change of scale sign when decomposing matrix

mackong:
When animate for rotation or scale path, decompose the transform matrix may change the sign of scale, and then compose the matrix back with scale or rotation changed by animation, the matrix may be wrong.

This PR add a new TrsTransformManager to save the TRS value of node from gltf data, and then compose the TRS and assign to TransformManager's transform.

Here is a test gltf:
[animation.glb.tar.gz](https://github.com/google/filament/files/12488413/animation.glb.tar.gz)

Before:
image

After:
image

### DIFF
--- a/libs/gltfio/include/gltfio/TrsTransformManager.h
+++ b/libs/gltfio/include/gltfio/TrsTransformManager.h
@@ -1,0 +1,114 @@
+/*
+ * Copyright (C) 2023 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef GLTFIO_TRSTRANSFORMMANAGER_H
+#define GLTFIO_TRSTRANSFORMMANAGER_H
+
+#include <filament/FilamentAPI.h>
+
+#include <utils/compiler.h>
+#include <utils/EntityInstance.h>
+#include <math/quat.h>
+#include <math/vec3.h>
+#include <math/mat4.h>
+
+using namespace filament::math;
+
+namespace utils {
+class Entity;
+} // namespace utils
+
+namespace filament::gltfio {
+
+class FTrsTransformManager;
+
+/**
+ * TrsTransformManager is used to add entities with glTF-specific trs information.
+ *
+ * Trs information here just used for Animation, DON'T use for transform.
+ */
+class UTILS_PUBLIC TrsTransformManager {
+public:
+    using Instance = utils::EntityInstance<TrsTransformManager>;
+    using Entity = utils::Entity;
+
+    /**
+     * Returns whether a particular Entity is associated with a component of this TrsTransformManager
+     * @param e An Entity.
+     * @return true if this Entity has a component associated with this manager.
+     */
+    bool hasComponent(Entity e) const noexcept;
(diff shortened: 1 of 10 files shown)


libs/gltfio/include/gltfio/TrsTransformManager.h
```diff
@@ -0,0 +1,114 @@
...
+     * @param scale             The scale to initialize the trs transform component with.
+     *
+     * If this component already exists on the given entity, it is first destroyed as if
+     * destroy(Entity e) was called.
+     *
+     * @see destroy()
+     */
+    void create(Entity entity);
+    void create(Entity entity, const float3& translation,
+                const quatf& rotation, const float3& scale); //!< \overload
```
bejado, line 78:
```suggestion
    void create(Entity entity, const float3& translation, const quatf& rotation,
            const float3& scale);  //!< \overload
```

libs/gltfio/include/gltfio/TrsTransformManager.h
```diff
@@ -0,0 +1,114 @@
...
+    const float3& getTranslation(Instance ci) const noexcept;
+
+    void setRotation(Instance ci, const quatf& rotation) noexcept;
+    const quatf& getRotation(Instance ci) const noexcept;
+
+    void setScale(Instance ci, const float3& scale) noexcept;
+    const float3& getScale(Instance ci) const noexcept;
+
+    void setTrs(Instance ci, const float3& translation,
+                const quatf& rotation, const float3& scale) noexcept;
```
bejado, line 98:
```suggestion
    void setTrs(Instance ci, const float3& translation, const quatf& rotation,
            const float3& scale) noexcept;
```

libs/gltfio/src/AssetLoader.cpp
```diff
@@ -400,7 +402,8 @@ void FAssetLoader::createRootAsset(const cgltf_data* srcAsset) {
     #endif
 
     mDummyBufferObject = nullptr;
-    mAsset = new FFilamentAsset(&mEngine, mNameManager, &mEntityManager, &mNodeManager, srcAsset);
+    mAsset = new FFilamentAsset(&mEngine, mNameManager, &mEntityManager,
+                                &mNodeManager, &mTrsTransformManager, srcAsset);
```
bejado, line 406:
```suggestion
    mAsset = new FFilamentAsset(&mEngine, mNameManager, &mEntityManager, &mNodeManager,
            &mTrsTransformManager, srcAsset);
```

bejado (approved):
LGTM, just a few formatting fixes

libs/gltfio/include/gltfio/TrsTransformManager.h
mackong: DONE

libs/gltfio/include/gltfio/TrsTransformManager.h
mackong: DONE

libs/gltfio/src/AssetLoader.cpp
mackong: DONE

poweifeng (approved):

libs/gltfio/src/AssetLoader.cpp
```diff
@@ -564,6 +567,8 @@ void FAssetLoader::recurseEntities(const cgltf_data* srcAsset, const cgltf_node*
         float3* scale = (float3*) &node->scale[0];
         float3* translation = (float3*) &node->translation[0];
         localTransform = composeMatrix(*translation, *rotation, *scale);
```
pixelflinger, line 569: we should remove localTransform here, and use the new Trs transform manager to calculate it, so that it's done the same way everywhere.

pixelflinger (changes requested):
Thank you for this PR, looks very good. My only comment is to always use the TrsTransformManager to calculate the real transform.

libs/gltfio/src/AssetLoader.cpp
mackong: FIXED

pixelflinger (approved):

PR #7147, opened 2023-09-06
Fix race condition in ShaderCompilerService

bejado:


pixelflinger (approved):

filament/backend/src/opengl/ShaderCompilerService.cpp
```diff
@@ -338,6 +343,12 @@ GLuint ShaderCompilerService::getProgram(ShaderCompilerService::program_token_t&
 
     token->compiler.cancelTickOp(token);
 
+    auto job = token->compiler.mCompilerThreadPool.dequeue(token);
```
pixelflinger, line 346: this should be conditional to whether the thread pool is used or not.

pixelflinger (changes requested):

pixelflinger (approved):

PR #7150, opened 2023-09-08
Fix incorrect SamplerParams operators

bejado:
This was causing the OpenGL backend to assert on x86_64 Macs:

```
assert_invariant(mSamplerMap.find(params) == mSamplerMap.end());
```

### DIFF
--- a/filament/backend/include/backend/DriverEnums.h
+++ b/filament/backend/include/backend/DriverEnums.h
@@ -813,22 +813,22 @@ struct SamplerParams { // NOLINT
     struct Hasher {
         size_t operator()(SamplerParams p) const noexcept {
             // we don't use std::hash<> here, so we don't have to include <functional>
-            return *reinterpret_cast<uint64_t const*>(reinterpret_cast<char const*>(&p));
+            return *reinterpret_cast<uint32_t const*>(reinterpret_cast<char const*>(&p));
         }
     };
 
     struct EqualTo {
         bool operator()(SamplerParams lhs, SamplerParams rhs) const noexcept {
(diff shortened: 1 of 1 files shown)


filament/backend/include/backend/DriverEnums.h
```diff
@@ -838,6 +838,7 @@ struct SamplerParams { // NOLINT
         return SamplerParams::LessThan{}(lhs, rhs);
     }
 };
+static_assert(sizeof(SamplerParams) == 4);
```
romainguy, line 841: 👍

romainguy (approved):

pixelflinger (approved):
Thanks!

filament/backend/include/backend/DriverEnums.h
```diff
@@ -813,22 +813,22 @@ struct SamplerParams { // NOLINT
     struct Hasher {
         size_t operator()(SamplerParams p) const noexcept {
             // we don't use std::hash<> here, so we don't have to include <functional>
-            return *reinterpret_cast<uint64_t const*>(reinterpret_cast<char const*>(&p));
+            return *reinterpret_cast<uint32_t const*>(reinterpret_cast<char const*>(&p));
```
suzp1984, line 816: Is there any conflict with #7104 ?

`This was causing the OpenGL backend to assert on x86_64 Macs:`
It seems it's not just an `OpenGL` backend issue.

bejado: https://github.com/google/filament/pull/7104 was the cause of the bug, yes. And certainly this could affect backends other than OpenGL, but the assert that was firing was specific to the OpenGL backend.

PR #7155, opened 2023-09-11
Improvements to shadowing

pixelflinger:
- use the geometric normal to apply the shadow bias. This affects
  cascades > 0 and spot/point lights.

- use the scene's origin as a reference point for stabilizing the
  shadowmap, this is more robust.

- clamp directional shadowmap correctly to the 1-texel border, which
  needs to be reachable, as it is a valid value.

- don't snap the shadowmap to texel boundaries if stable mode is not
  active (before we only didn't do it based on lispsm). Stable mode can
  make the shadow unstable when both the camera and the scene move 
  together, so it's better to have a more predictable API where
  "stable" mode means that the snapping occurs and doesn't otherwise.

- add "far origin" distance slider to the debug ui

FIXES=[299310624]

### DIFF
--- a/samples/gltf_viewer.cpp
+++ b/samples/gltf_viewer.cpp
@@ -101,6 +101,7 @@ struct App {
 
     bool actualSize = false;
     bool originIsFarAway = false;
+    float originDistance = 6378137; // Earth's radius in [m]
 
     struct Scene {
         Entity groundPlane;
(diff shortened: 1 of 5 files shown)


samples/gltf_viewer.cpp
```diff
@@ -101,6 +101,7 @@ struct App {
 
     bool actualSize = false;
     bool originIsFarAway = false;
+    float originDistance = 6378137;
```
romainguy, line 104: Comment to explain this value?

romainguy (approved):

samples/gltf_viewer.cpp
pixelflinger: isn't it obvious it's the radius of Earth in meters? :-)

PR #7165, opened 2023-09-12
Improve bloom

pixelflinger:
- add a quality option
- remove the ping-pong code, we'll disable for GPU that don't work instead.
- improve quality by doing a better first downscale (using a 5x5 gaussian).
- improve performance by using a 9 tap filter instead of 13 for in most cases
- fix usages of setMinMaxLevels as it resets the base level to "min"


filament/src/PostProcessManager.cpp
```diff
@@ -1893,38 +1931,73 @@ PostProcessManager::BloomPassOutput PostProcessManager::bloomPass(FrameGraph& fg
...
+
+    while (2 * bloomWidth < float(desc.width) || 2 * bloomHeight < float(desc.height)) {
+        if (inoutBloomOptions.quality == QualityLevel::LOW ||
+            inoutBloomOptions.quality == QualityLevel::MEDIUM) {
+            input = downscalePass(fg, input, {
+                            .width  = (desc.width  = std::max(1u, desc.width  / 2)),
+                            .height = (desc.height = std::max(1u, desc.height / 2)),
+                            .format = outFormat
+                    },
+                    threshold, inoutBloomOptions.highlight, threshold);
```
romainguy, line 1966: Did you intend to use `threshold` as the value for the `fireflies` param?

filament/src/PostProcessManager.cpp
```diff
@@ -1893,38 +1931,73 @@ PostProcessManager::BloomPassOutput PostProcessManager::bloomPass(FrameGraph& fg
...
+        }
     }
 
+    // convert back to integer width/height
+    uint32_t const width  = std::max(1u, uint32_t(std::floor(bloomWidth)));
+    uint32_t const height = std::max(1u, uint32_t(std::floor(bloomHeight)));
+
+    input = downscalePass(fg, input,
+            { .width = width, .height = height, .format = outFormat },
+            threshold, inoutBloomOptions.highlight, threshold);
```
romainguy, line 1988: Same here

filament/src/PostProcessManager.cpp
pixelflinger: yes, they're always happening together

poweifeng (approved):

PR #7171, opened 2023-09-14
Fix the output of Windows release workflow

poweifeng:
This probably broke due to a change from old windows command prompt to powershell. The way to refer to env variables in powershell is `$Env:var`


romainguy (approved):

pixelflinger (approved):

bejado (approved):
Nice, thank you! I had been meaning to look into that..

PR #7177, opened 2023-09-15
improve/simplify "Focus on shadow casters" algorithm

pixelflinger:
Functionally this shouldn't be too different, but we have some improvements:
- better detection of "no shadows" cases
- more computations done in light-space, which should result in better light frustum.
- spit the code into several static functions

### DIFF
--- a/filament/src/ShadowMap.h
+++ b/filament/src/ShadowMap.h
@@ -26,6 +26,7 @@
 
 #include <backend/DriverApiForward.h>
 
+#include <filament/Box.h>
 #include <filament/Viewport.h>
 
 #include <math/mat4.h>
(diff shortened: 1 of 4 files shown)


filament/src/ShadowMap.h
```diff
@@ -215,17 +219,50 @@ class ShadowMap {
             const ShadowMapInfo& shadowMapInfo,
             const FLightManager::ShadowParams& params) noexcept;
 
+    struct DirectionalShadowBounds {
+        math::mat4f Mv;
+        float zNear;
+        float zFar;
+        FrustumBoxIntersection lsClippedShadowVolume;
+        size_t vertexCount;
+        bool visibleShadows = false;
```
bejado, line 228: Should the other members have default initializers as well?

bejado (approved):

poweifeng (approved):

PR #7181, opened 2023-09-18
vulkan: cache vertex buffer info

poweifeng:
To reduce CPU on draw(), we move the VertexBuffer related metadata out of draw into a cache. We store the cache info outside of the actual VulkanVertexBuffer class since VulkanVertexBuffer subclass HwVertexBuffer, which is a Handle meant to be minimal.

This  means that we need to cache on the heap, but it should be ok since the caching is only for scene set-up and not per-frame.


filament/backend/src/vulkan/VulkanPipelineInfoCache.h
```diff
@@ -0,0 +1,79 @@
...
+    void increaseCapacity() {
+        mBacking.emplace_back(std::make_unique<VertexBufferInfoArray>());
+        auto& array = *(*(mBacking.end() - 1));
+        for (auto& info: array) {
+            mBuffers.push_back(&info);
+        }
+    }
+
+    using VertexBufferInfoArray = std::array<VertexBufferInfo, CACHE_INCREMENT>;
+    std::list<VertexBufferInfo*> mBuffers;
```
pixelflinger, line 73: `std::list` is pretty bad, it has 16-bytes overhead + it fragments the heap a lot, does lots of pointer chasing, etc... it's bad for the  cache. I wouldn't use. Honestly I would just use a `std::vector<VertexBufferInfo*>` instead. Also for something like this I wouldn't use `unique_ptr`, you should be able to do all the management yourself in the cache and avoid the overhead of unique_ptr.

You could also use a `std::deque` for `mBacking`, because objects don't move on insertion, so you could "simply" store a pointer in the `Handle<>`. But `deque` is really not efficient on some platforms (on windows).

I wonder if you couldn't do something more simple where you'd store a pointer in the `Handle<>` and it could point to a custom "pool" allocator. The size of the pool would have to be determined at compile time, but you can implement a "fallback" to malloc when it runs out of space + a log.

At the very least I would not use `std::list`, if insert/deletion is rare (which I assume it is), a simple `std::vector` will do. At least it'll fragment the heap much less.

pixelflinger (changes requested):

filament/backend/src/vulkan/VulkanPipelineInfoCache.h
poweifeng: thanks for the heads-up on std::list.  I opted to just go with allocating on the heap for the struct, maybe we can revisit the pool in the future. Let me know if that's ok.

filament/backend/src/vulkan/VulkanHandles.h
```diff
@@ -94,11 +94,46 @@ struct VulkanVertexBuffer : public HwVertexBuffer, VulkanResource {
...
+              bufferDesc(size, VkVertexInputBindingDescription{}),
+              vkbuffers(size, VK_NULL_HANDLE),
+              offsets(size, 0),
+              attribToBufferIndex(size, -1) {}
+
+        Array<VkVertexInputAttributeDescription> attribDesc;
+        Array<VkVertexInputBindingDescription> bufferDesc;
+        Array<VkBuffer> vkbuffers;
+        Array<VkDeviceSize> offsets;
+        Array<int8_t> attribToBufferIndex;
```
pixelflinger, line 133: all of this actually looks like a `utils::StructureOfArray<>` -- if you used it, it would be a single allocation for all arrays which would be a bit more efficient. Might maybe be easier to use too, since it makes things appear as a struct.

pixelflinger (approved):

bejado (approved):

filament/backend/src/vulkan/VulkanHandles.h
poweifeng: Switched to SoA.

PR #7183, opened 2023-09-18
new API to return a Material's supported variants

pixelflinger:
FIXES=[297456590]


poweifeng (approved):

bejado (commented):
If a client uses this API with a material compiled for feature level 0, what's the expected behavior? Should it return `false` for variants like `SSR` and `VSM`, because these are filtered out [_after_](https://github.com/google/filament/blob/main/libs/filamat/src/MaterialBuilder.cpp#L1155-L1164) `writeCommonChunks`.

pixelflinger:
> If a client uses this API with a material compiled for feature level 0, what's the expected behavior? Should it return `false` for variants like `SSR` and `VSM`, because these are filtered out [_after_](https://github.com/google/filament/blob/main/libs/filamat/src/MaterialBuilder.cpp#L1155-L1164) `writeCommonChunks`.

oh, good catch. I think we want to filter them out first.

PR #7190, opened 2023-09-21
Fix: remove duplicated links inside ios samples.

suzp1984:
There are some duplicated library linkers inside ios samples Build Settings.


bejado:
Thanks, but the Xcode projects are generated using [XcodeGen](https://github.com/google/filament/blob/main/ios/samples/README.md#xcodegen). The [yml](https://github.com/google/filament/blob/main/ios/samples/app-template.yml) project files should be updated then the projects re-generated.

suzp1984:
Done, also remove the duplicated libs from the [gltf-viewer | hello-ar | hello-gltf | hello-pbr]/project.xml

bejado (approved):

PR #7207, opened 2023-09-26
vk: remove unnecessary flush and gc

poweifeng:


filament/backend/src/vulkan/VulkanDriver.cpp
```diff
@@ -1370,10 +1372,6 @@ void VulkanDriver::commit(Handle<HwSwapChain> sch) {
 
     VulkanSwapChain* swapChain = mResourceAllocator.handle_cast<VulkanSwapChain*>(sch);
 
-    if (mCommands->flush()) {
-        collectGarbage();
-    }
```
pixelflinger, line 1375: wouldn't make more sense to flush here as opposed to in endFrame? e.g. on GL eglSwapBuffers() makes an implicit flush.

filament/backend/src/vulkan/VulkanSwapChain.cpp
```diff
@@ -88,8 +88,8 @@ void VulkanSwapChain::present() {
                 .layerCount = 1,
         };
         mColors[mCurrentSwapIndex]->transitionLayout(cmdbuf, subresources, VulkanLayout::PRESENT);
-        mCommands->flush();
     }
+    mCommands->flush();
```
pixelflinger, line 92: oh there is a flush in present. so do we need the one in endFrame()?

filament/backend/src/vulkan/VulkanDriver.cpp
poweifeng: yes, it's been moved inside present

filament/backend/src/vulkan/VulkanSwapChain.cpp
poweifeng: for reasons I don't understand yet - Yes, we do need it in `endFrame()`.  I see an increase in `cpu` usage when it's not there, on both MoltenVK and adreno. I plan to turn on a performance validation on an nvidia card to see if that tells me anything.

pixelflinger (approved):

PR #7215, opened 2023-09-28
Fix masked materials

romainguy:
When a material using the MASKED blend mode, we sharpen the alpha channel to avoid banding and unwanted translucency with MSAA. To do this properly however, we must output the *sharpened* alpha value, and not the *original* alpha value. This change moves the alpha masking test to right after we initialize the material inputs with the user code so that the rest of our shaders will user the correct alpha value. The line that was causing problems in our case was the last line in evaluateLights() which returns a vec4 containing an alpha computed from the input's alpha instead of the sharpened alpha.

This change has the added benefit of unifying the alpha masking code between lit and unlit shaders, which contained duplicated code for not good reason. Note that moving the call to the applyAlphaMask() function has no effect of *when* a discard happens in the flow of our shaders.

Before (GL backend, MSAA on):
Screenshot 2023-09-27 at 6 46 32 PM

After:
Screenshot 2023-09-27 at 6 46 39 PM

### DIFF
--- a/NEW_RELEASE_NOTES.md
+++ b/NEW_RELEASE_NOTES.md
@@ -14,3 +14,4 @@ appropriate header in [RELEASE_NOTES.md](./RELEASE_NOTES.md).
 - build: fix emscripten-1.3.46 build
 - engine: materials built for feature level 0 can now also be loaded in higher feature levels [⚠️
   **New Material Version**]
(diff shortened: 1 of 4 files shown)


romainguy:
One more fix needed for unlit

poweifeng (approved):

PR #7218, opened 2023-09-29
(Partially) revert "Remove now-redundant feature level 0 materials"

elizagamedev:
This reverts most of commit 9a6b8bf24e729c95b057aab0582310636e996cbe. The hello triangle sample remains unreverted.

The original commit inadvertently broke screen space reflections, and perhaps other features when the default material was used. The source of the issue is that MaterialBuilder.cpp (correctly) filters out variants that aren't supported in feature level 0 materials, including screen space reflections.

Unfortunately, while the "feature level 0 compatibility" feature itself was intended to make creating duplicate materials like this redundant in client code, unfortunately, it seems the best solution for resolving this issue is to simply keep these redundant materials in the core.

To elaborate: clients should expect that feature level 0 materials that they create work on /all/ feature levels /exactly/ or /close to exactly/ identically. This includes restricting more advanced features that theoretically could be available on a higher feature level, like SSR. It's already true that if a user would like to optionally opt-in to a more advanced material which takes advantage of more advanced features, they would have to maintain two separate versions of that material: one for feature level 3 and one for feature level 1. It should be no different in this case.

However, the materials built into the engine core are an exception to this expectation. Given that feature level 0 was tacked on after the fact with fewer features, there must /by necessity/ have been a new material introduced for both the default material and the default skybox specifically for feature level 0 with fewer features than extant client apps expected to be included by default. I imagine if filament were to be rebuilt from the ground up, this exception wouldn't exist. However, the end result is this somewhat messy redundancy.

### DIFF
--- a/filament/CMakeLists.txt
+++ b/filament/CMakeLists.txt
@@ -245,6 +245,11 @@ set(MATERIAL_SRCS
         src/materials/vsmMipmap.mat
 )
 
+set(MATERIAL_ES2_SRCS
+        src/materials/defaultMaterial0.mat
+        src/materials/skybox0.mat
+)
+
 # Embed the binary resource blob for materials.
 get_resgen_vars(${RESOURCE_DIR} materials)
 list(APPEND PRIVATE_HDRS ${RESGEN_HEADER})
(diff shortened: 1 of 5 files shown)


elizagamedev:
> Please add a release note line to [NEW_RELEASE_NOTES.md](https://github.com/google/filament/edit/exv/fix-ssr/NEW_RELEASE_NOTES.md). If this PR does not warrant a release note, add the 'internal' label to this PR.

Done.

NEW_RELEASE_NOTES.md
```diff
@@ -10,3 +10,4 @@ appropriate header in [RELEASE_NOTES.md](./RELEASE_NOTES.md).
 - materials: fix alpha masked materials when MSAA is turned on [⚠️ **Recompile materials**]
 - materials: better support materials with custom depth [**Recompile Materials**]
 - engine: fade shadows at shadowFar distance instead of hard cutoff [⚠️ **New Material Version**]
+- engine: fix critical regression where enabling screen-space reflections would crash
```
bejado, line 13: No release note needed since the problematic change ("Remove now-redundant feature level 0 materials") has not been released yet.

bejado (approved):
Please merge into main and cherry-pick to rc/1.44.0

pixelflinger:
Would it be possible to filter out the variants only when generating the essl 1.0 version of the material?

I can't help thinking that both "redundant" materials are actually identical.

PR #7222, opened 2023-10-02
Fix build failures due to filamat lite removal

bejado:


romainguy (approved):
This will also change the names of the artifacts for Maven, but I don't think there are more files to update for this.

poweifeng (approved):

pixelflinger (approved):

PR #7224, opened 2023-10-03
Fix build failures due to filamat lite removal #2

poweifeng:


bejado (approved):
Didn't realize we also got rid of gltfio full/lite, thanks!

PR #7223, opened 2023-10-03
engine: add job system thread count configuration

poweifeng:
- Plumb Engine::Config in Java
 - Add Engine::Builder for Java
 - Add jobSystemThreadCount to Engine::Config

BUG=303129581

PR #7231, opened 2023-10-04
fix java Engine.Builder method access modifiers.

suzp1984:
# 1. Fix Java Engine.Builder methods access modifiers.

The default java class method access modifier is `protected`, The `Engine.Builder` is a static inner class of `Engine`, so the `Builder`'s methods, except the `public constructor`, are not accessible outside of `Engine` namespace.
And the `Engine.create` methods are announced `deprecated`. So I believe the `Engine.Builder`'s build methods should be publicly access from outside.

# 2. Fix warning that still use the `deprecated` `Engine.create` api.

### DIFF
--- a/android/filament-android/src/main/java/com/google/android/filament/Engine.java
+++ b/android/filament-android/src/main/java/com/google/android/filament/Engine.java
@@ -173,7 +173,7 @@ public class Engine {
          * @param backend Driver backend to use
          * @return A reference to this Builder for chaining calls.
          */
-        Builder backend(Backend backend) {
+        public Builder backend(Backend backend) {
             nSetBuilderBackend(mNativeBuilder, backend.ordinal());
             return this;
         }
@@ -186,7 +186,7 @@ public class Engine {
          *                       <b>must be</b> an instance of {@link android.opengl.EGLContext}.
          * @return A reference to this Builder for chaining calls.
          */
-        Builder sharedContext(Object sharedContext) {
+        public Builder sharedContext(Object sharedContext) {
             if (Platform.get().validateSharedContext(sharedContext)) {
                 nSetBuilderSharedContext(mNativeBuilder,
                         Platform.get().getSharedContextNativeHandle(sharedContext));
@@ -201,7 +201,7 @@ public class Engine {
          * @param config A {@link Config} object
          * @return A reference to this Builder for chaining calls.
          */
-        Builder config(Config config) {
+        public Builder config(Config config) {
             nSetBuilderConfig(mNativeBuilder, config.commandBufferSizeMB,
                     config.perRenderPassArenaSizeMB, config.driverHandleArenaSizeMB,
                     config.minCommandBufferSizeMB, config.perFrameCommandsSizeMB,
@@ -219,7 +219,7 @@ public class Engine {
          * @exception IllegalStateException can be thrown if there isn't enough memory to
          * allocate the command buffer.
          */
-        Engine build() {
+        public Engine build() {
             long nativeEngine = nBuilderBuild(mNativeBuilder);
             if (nativeEngine == 0) throw new IllegalStateException("Couldn't create Engine");
             return new Engine(nativeEngine);
@@ -352,9 +352,7 @@ public class Engine {
      * @exception IllegalStateException can be thrown if there isn't enough memory to
      * allocate the command buffer.
      *
-     * @deprecated use {@link Builder}
      */
-    @Deprecated
     @NonNull
     public static Engine create() {
         return new Builder().build();
@@ -374,9 +372,7 @@ public class Engine {
      * @exception IllegalStateException can be thrown if there isn't enough memory to
      * allocate the command buffer.
      *
-     * @deprecated use {@link Builder}
      */
-    @Deprecated
     @NonNull
     public static Engine create(@NonNull Backend backend) {
         return new Builder()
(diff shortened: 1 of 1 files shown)


romainguy:
The default visibility is `package private` not `protected`, but yes those methods need a `public` access modifier. `Engine.create()` is not deprecated, there's no reason to stop using it.

android/filament-android/src/main/java/com/google/android/filament/Engine.java
```diff
@@ -32,7 +32,7 @@
  * <pre>
  * import com.google.android.filament.*
  *
- * Engine engine = Engine.create();
```
romainguy, line 35: Don't change this.

android/filament-android/src/main/java/com/google/android/filament/Engine.java
```diff
@@ -47,7 +47,7 @@
  * <pre>
  * import com.google.android.filament.*
  *
- * Engine engine        = Engine.create();
```
romainguy, line 50: Don't change this.

android/filament-android/src/main/java/com/google/android/filament/Engine.java
```diff
@@ -432,7 +432,7 @@ public boolean isValid() {
      * <li>Terminate the rendering engine's thread.</li>
      *
      * <pre>
-     * Engine engine = Engine.create();
```
romainguy, line 435: Don't change this.

android/filament-utils-android/src/main/java/com/google/android/filament/utils/ModelViewer.kt
```diff
@@ -146,7 +146,7 @@ class ModelViewer(
 
     constructor(
             surfaceView: SurfaceView,
-            engine: Engine = Engine.create(),
+            engine: Engine = Engine.Builder().build(),
```
romainguy, line 149: Don't change this.

android/filament-utils-android/src/main/java/com/google/android/filament/utils/ModelViewer.kt
```diff
@@ -166,7 +166,7 @@ class ModelViewer(
     @Suppress("unused")
     constructor(
             textureView: TextureView,
-            engine: Engine = Engine.create(),
+            engine: Engine = Engine.Builder().build(),
```
romainguy, line 169: Don't change this.

android/samples/sample-stream-test/src/main/java/com/google/android/filament/streamtest/MainActivity.kt
```diff
@@ -132,7 +132,9 @@ class MainActivity : Activity(), ActivityCompat.OnRequestPermissionsResultCallba
 
     private fun setupFilament() {
         val eglContext = createEGLContext()
-        engine = Engine.create(eglContext)
+        engine = Engine.Builder()
```
romainguy, line 135: Don't change this.

android/samples/sample-textured-object/src/main/java/com/google/android/filament/textured/MainActivity.kt
```diff
@@ -114,7 +114,7 @@ class MainActivity : Activity() {
     }
 
     private fun setupFilament() {
-        engine = Engine.create()
+        engine = Engine.Builder().build()
```
romainguy, line 117: Don't change this.

romainguy (changes requested):

android/filament-utils-android/src/main/java/com/google/android/filament/utils/ModelViewer.kt
suzp1984: But there are warnings about `deprecated` api.
So, should the `@Deprecated` annotation be removed from `Engine.create`?

romainguy: Apparently that `@Deprecated` was recently added, that's why I didn't see it in my source tree. It seems heavy-handed. @poweifeng are we absolutely sure we need to deprecate those APIs? Now all Java apps/samples will have a deprecation warning at compile time for no good reason.

poweifeng: I suppose it's ok to *not deprecate*.  I'm kind of following the reasoning in this comment: https://github.com/google/filament/blob/main/filament/include/filament/Engine.h#L371

Though the comment itself does not indicate a deprecation.

suzp1984: Okay, I deleted the `@Deprecated` annotation from `Engine.create`.
But I prefer to keep `@Deprecated`, it's the `Java`'s way in my mind.

It will revert part of #7223 .

android/filament-utils-android/src/main/java/com/google/android/filament/utils/ModelViewer.kt
suzp1984: done, revert the code.

poweifeng (approved):
Thanks for the fix!

pixelflinger (approved):

android/filament-utils-android/src/main/java/com/google/android/filament/utils/ModelViewer.kt
suzp1984: Screen Shot 2023-10-06 at 10 50 12 AM

I forgot to revert the `NEW_RELEASE_NOTES.md`, part of #7223  , @poweifeng Do you want me to do it in another PR? or do you want to do it by yourself?

`engine: In Java, decpreate Engine.create() and recommend Engine.Builder instead` ->
`engine: In Java, add Engine.Builder to create Engine instance.`

poweifeng: I can make a change. Thanks!

PR #7236, opened 2023-10-05
Add AgX tonemapper

bejado:
AgX:
Screenshot 2023-10-05 at 11 39 32 AM

ACES (legacy):
Screenshot 2023-10-05 at 11 39 24 AM


filament/src/ToneMapper.cpp
```diff
@@ -230,6 +230,73 @@ float3 FilmicToneMapper::operator()(math::float3 x) const noexcept {
...
+            - 126.7     * x4 * x
+            + 92.06     * x4
+            - 28.72     * x2 * x
+            + 4.361     * x2
+            - 0.1718    * x
+            + 0.002857;
+}
+
+float3 AgxToneMapper::operator()(float3 v) const noexcept {
+    // TODO: It's unclear if we need to temporarily transform to 709 primaries again.
```
romainguy, line 273: If you don't change ColorGrading.cpp the values you'll get here are in Rec2020. See selectColorGradingTransformIn()/selectColorGradingTransformOut(). Seems like if we don't do anything we'll match Blender? Might be worth checking Blender's implementation to see what they do.

filament/include/filament/ToneMapper.h
```diff
@@ -115,6 +115,13 @@ struct UTILS_PUBLIC FilmicToneMapper final : public ToneMapper {
     math::float3 operator()(math::float3 x) const noexcept override;
 };
 
+struct UTILS_PUBLIC AgxToneMapper final : public ToneMapper {
```
romainguy, line 118: Don't forget the Java APIs

filament/src/ToneMapper.cpp
```diff
@@ -230,6 +230,73 @@ float3 FilmicToneMapper::operator()(math::float3 x) const noexcept {
...
+
+    // Log2 encoding
+    v = max(v, 1E-10); // avoid 0 or negative numbers for log2
+    v = log2(v);
+    v = (v - AgxMinEv) / (AgxMaxEv - AgxMinEv);
+
+    v = clamp(v, 0, 1);
+
+    // Apply sigmoid
+    v = agxDefaultContrastApprox(v);
```
romainguy, line 288: Did you try with the agxLook() step as well?

romainguy (commented):
How well does this interact with luminance scaling? Can you try? If it doesn't behave well (since AgX kind of does it already) we may want to no-op luminanceScaling with AgX.

filament/src/ToneMapper.cpp
bejado: Right, which I why I originally converted back to sRGB with the commented-out: `v = Rec2020_to_sRGB * v;`. The way I have it matches Blender (and looks good to me), but I'm not sure if it's _right_, as Sobotka's original AgX seems to keep everything in 709.

filament/src/ToneMapper.cpp
bejado: Not yet, but I'll give it a shot

filament/src/ToneMapper.cpp
romainguy: I can put you in touch with Troy on Discord if you want. Does Blender maybe use a different inset/outset matrix? The idea is to just slightly expand/contract the gamut.

bejado: Technically yes, Blender's uses different `primaries_rotate` and `primaries_scale` parameters, resulting in a different inset/outset matrices. I'm not sure if they adjusted the parameters for aesthetic reasons or to somehow compensate for Rec2020.

romainguy: I would expect it was made to map better to Rec2020. That said we don't render in 2020, we only use 2020 during color grading.

bejado:
> How well does this interact with luminance scaling? Can you try? If it doesn't behave well (since AgX kind of does it already) we may want to no-op luminanceScaling with AgX.

https://github.com/google/filament/assets/5298046/837e23fb-cdbc-4090-935c-8280394a105c

I wouldn't say it breaks it, but I do I think I prefer AgX without luminance scaling.

romainguy:
Agreed, it's better without, but prob not worth no-oping luminance scaling when AgX is selected.

NEW_RELEASE_NOTES.md
```diff
@@ -8,5 +8,6 @@ appropriate header in [RELEASE_NOTES.md](./RELEASE_NOTES.md).
 
 ## Release notes for next branch cut
 
-engine: Added parameter for configuring JobSystem thread count
-engine: In Java, decpreate Engine.create() and recommend Engine.Builder instead
\ No newline at end of file
+- engine: Added parameter for configuring JobSystem thread count
+- engine: In Java, decpreate Engine.create() and recommend Engine.Builder instead
```
romainguy, line 12: Typo from previous PR dcpreate -> deprecate

romainguy (commented):
LGTM so far, let me know what agxLook() does. Might be something optional driven by a parameter?

pixelflinger (approved):

bejado:
> LGTM so far, let me know what agxLook() does. Might be something optional driven by a parameter?

FYI

https://github.com/google/filament/assets/5298046/2e8f6d3f-be5c-4710-a484-58b05c475039

filament/src/ToneMapper.cpp
```diff
@@ -230,6 +230,106 @@ float3 FilmicToneMapper::operator()(math::float3 x) const noexcept {
...
+        sat = 1.3;
+    }
+    if (look == AgxToneMapper::AgxLook::PUNCHY) {
+        slope = float3(1.0);
+        power = float3(1.35, 1.35, 1.35);
+        sat = 1.4;
+    }
+
+    // ASC CDL
+    val = pow(val * slope + offset, power);
```
romainguy, line 302: Note: our existing color grading APIs let you apply a CDL.

romainguy (approved):

PR #7244, opened 2023-10-09
fix ubershader index for transmission&volume material

mackong:
Spec features for volume "contains" features for transmission, so transmission's ubershader index should place before volume's, otherwise volume will be picked for materials where only KHR_materials_transmission extension used.


pixelflinger (approved):

pixelflinger:
@romainguy does that look right?

PR #7250, opened 2023-10-09
Create use-after-free detector for Metal textures

bejado:


filament/backend/include/private/backend/DriverAPI.inc
```diff
@@ -219,7 +219,7 @@ DECL_DRIVER_API_R_N(backend::TextureHandle, importTexture,
         backend::TextureUsage, usage)
 
 DECL_DRIVER_API_R_N(backend::SamplerGroupHandle, createSamplerGroup,
-        uint32_t, size)
+        uint32_t, size, const char*, name)
```
pixelflinger, line 222: Using a raw pointer here is dangerous. I'm *okay* with it if we document/make sure that on the caller side we use string literals. And I think it's the right thing to do. I don't want to add heap allocations for a debug feature like this.

Given that, we don't need to use a `CString` on the backend side (which will heap allocate). Just store the raw pointer there too.

filament/backend/src/metal/MetalContext.h
```diff
@@ -111,6 +117,13 @@ struct MetalContext {
     tsl::robin_set<MetalSamplerGroup*> samplerGroups;
     tsl::robin_set<MetalTexture*> textures;
 
+    // This deque implements delayed destruction for Metal texture handles. It keeps a handle to
+    // the most recent kMetalFreedTextureListSize texture handles. When we're asked to destroy a
+    // texture handle, we free its texture memory, but keep the MetalTexture object alive, marking
+    // it as "terminated". If we later are asked to use that texture, we can check its terminated
+    // status and throw an error instead of crashing.
+    std::deque<Handle<HwTexture>> texturesToDestroy;
```
pixelflinger, line 125: `deque` is very inefficient in terms of code size and heap allocations. I would prefer we use a bespoke, fixed-size, circular buffer. Especially for textures.

filament/backend/src/metal/MetalDriver.mm
```diff
@@ -530,8 +530,19 @@
...
+    // Free memory from the texture and mark it as freed.
+    metalTexture->terminate();
+
+    // Delay the destruction of this texture handle.
+    mContext->texturesToDestroy.push_back(th);
+    if (mContext->texturesToDestroy.size() > kMetalFreedTextureListSize) {
+        Handle<HwTexture> handleToFree = mContext->texturesToDestroy.front();
+        mContext->texturesToDestroy.pop_front();
+        destruct_handle<MetalTexture>(handleToFree);
+    }
```
pixelflinger, line 545: the delayed free should be protected by a debug option. Either set at runtime or compile time, but we need a way to turn it off.

filament/backend/src/metal/MetalDriver.mm
```diff
@@ -854,22 +871,36 @@
...
+            NSString* reason = [NSString stringWithFormat:
+                    @"Filament Metal texture use after free, sampler "
+                    @"group = %s, texture index = %zu",
+                    sb->debugName.c_str_safe(), s];
+            NSException* useAfterFreeException =
+                    [NSException exceptionWithName:@"MetalTextureUseAfterFree"
+                                            reason:reason
+                                          userInfo:nil];
+            [useAfterFreeException raise];
+        }
```
pixelflinger, line 893: this code looks identical to the one in `finalizeSamplerGroup` ca we move it in a "checkUseAfterFree" method?

filament/backend/src/metal/MetalHandles.h
```diff
@@ -322,6 +328,7 @@ class MetalSamplerGroup : public HwSamplerGroup {
     void useResources(id<MTLRenderCommandEncoder> renderPassEncoder);
 
     size_t size;
+    utils::CString debugName;
```
pixelflinger, line 331: because we used a `const char*` u p to this point, and it wasn't safe anyways, there is no point using a CString here. we should document that the name needs to be a string literal.

filament/src/details/MaterialInstance.cpp
```diff
@@ -73,7 +73,7 @@ FMaterialInstance::FMaterialInstance(FEngine& engine,
 
     if (!material->getSamplerInterfaceBlock().isEmpty()) {
         mSamplers = other->getSamplerGroup();
-        mSbHandle = driver.createSamplerGroup(mSamplers.getSize());
+        mSbHandle = driver.createSamplerGroup(mSamplers.getSize(), mMaterial->getName().c_str());
```
pixelflinger, line 76: that would be extremely unsafe because the material could be destroyed before the backend gets the pointer.

pixelflinger (changes requested):

filament/src/details/MaterialInstance.cpp
bejado: I can replace this with something generic like "User material", but I think it would be super helpful if our error message was able to print out the name of the material. But making that safe requires a string copy and heap allocation. What do you think?

filament/backend/src/metal/MetalDriver.mm
bejado: What about a new option in `DriverConfig`?

```cpp
    struct DriverConfig {
        /*
         * size of handle arena in bytes. Setting to 0 indicates default value is to be used.
         * Driver clamps to valid values.
         */
        size_t handleArenaSize = 0;
        
        size_t textureUseAfterFreePoolSize = 64;
    };

```

Setting to 0 could be the default and disable it.

filament/src/details/MaterialInstance.cpp
pixelflinger: What about we use a fixed size for the name, like 32 bytes (which is the size of a std::string). You can then use a `std::array<char*, 32>` 
You could even create a `FixedSizeString<size_t>` in libutils if you wanted to, to make it much more conveniant.

pixelflinger (changes requested):

libs/utils/include/utils/FixedCircularBuffer.h
```diff
@@ -0,0 +1,60 @@
...
+        mSize--;
+        return result;
+    }
+
+private:
+    T mData[N];
+
+    size_t mBegin = 0;
+    size_t mEnd = 0;
+    size_t mSize = 0;
```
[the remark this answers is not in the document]
bejado: Actually yes we do, otherwise we can't distinguish between an empty buffer and a full buffer. In both cases `mBegin == mEnd`. Alternatively we could allocate +1 slots (but keep the same capacity) so `mEnd` never reaches `mBegin`.

filament/backend/src/metal/MetalDriver.mm
```diff
@@ -854,22 +870,26 @@
     assert_invariant(sb->size == data.size / sizeof(SamplerDescriptor));
     auto const* const samplers = (SamplerDescriptor const*) data.buffer;
 
-#ifndef NDEBUG
-    // In debug builds, verify that all the textures in the sampler group are still alive.
+    // Verify that all the textures in the sampler group are still alive.
```
pixelflinger, line 873: unfortunate that this whole loop is technically not needed.

pixelflinger (approved):

libs/utils/include/utils/FixedCircularBuffer.h
```diff
@@ -0,0 +1,77 @@
...
+    bool empty() const noexcept { return mSize == 0; }
+
+    /**
+     * Push v into the buffer. If the buffer is already full, removes the oldest item and returns
+     * it. If this buffer has no capacity, simply returns v.
+     * @param v the new value to push into the buffer
+     * @return if the buffer was full, the oldest value which was displaced
+     */
+    std::optional<T> push(T v) noexcept {
+        if (mCapacity == 0) {
```
poweifeng, line 48: just curious, why would you ever have a mCapacity== 0, also I think it'll segfault when you do `% mCapacity`?

bejado: `mCapacity` is set to 0 in the case where the use-after-free detection is turned off. In that case, `mSize` is also 0, and `% 0` should never happen.

poweifeng: ack

PR #7257, opened 2023-10-10
matinfo: add support for viewing ESSL1 code

elizagamedev:


libs/matdbg/src/ShaderInfo.cpp
```diff
@@ -94,12 +94,12 @@ bool getMetalShaderInfo(const ChunkContainer& container, ShaderInfo* info) {
     return true;
 }
 
-bool getGlShaderInfo(const ChunkContainer& container, ShaderInfo* info) {
-    if (!container.hasChunk(ChunkType::MaterialGlsl)) {
+bool getGlShaderInfo(const ChunkContainer& container, ShaderInfo* info, ChunkType chunkType) {
```
poweifeng, line 97: Just roughly looking the three methods: `getGlShaderInfo` `getMetalShaderInfo`, `getVkShaderInfo`, looks like they are all the same except for the chunktype?  Maybe consolidate?

libs/matdbg/src/TextWriter.cpp
```diff
@@ -402,13 +402,23 @@ static void printShaderInfo(ostream& text, const vector<ShaderInfo>& info,
     text << endl;
 }
 
-static bool printGlslInfo(ostream& text, const ChunkContainer& container) {
+static bool printGlslInfo(ostream& text, const ChunkContainer& container, ChunkType chunkType) {
     vector<ShaderInfo> info;
     info.resize(getShaderCount(container, ChunkType::MaterialGlsl));
-    if (!getGlShaderInfo(container, info.data())) {
+    if (!getGlShaderInfo(container, info.data(), ChunkType::MaterialGlsl)) {
```
poweifeng, line 408: So it's not `chunkType` here and the line above?  

Also, maybe now that we're passing chunkType to `get*ShaderInfo`, maybe we can further consolidate these methods?  (If the `get*ShaderInfo` can be consolidated into one).

pixelflinger (approved):

libs/matdbg/src/ShaderInfo.cpp
elizagamedev: Done.

libs/matdbg/src/TextWriter.cpp
elizagamedev: Oops, missed this.

bejado (approved):

poweifeng (approved):

PR #7259, opened 2023-10-11
samples: support apply all animations in gltf_viewer

mackong:
For gltf models with more than one animations, we now can only check one by one, but sometimes we need to check the whole effect of all animations applied simultaneously.

This patch add a new "Apply All Animations" button for gltf_viewer, for check all animations quickly. 

Now for _mCurrentAnimation_:
_-1_: means not playing animation
_0 ~ numAnimations_: means playing the selected animation
_numAnimations_: means playing all animations

### DIFF
--- a/libs/viewer/include/viewer/ViewerGui.h
+++ b/libs/viewer/include/viewer/ViewerGui.h
@@ -227,7 +227,7 @@ public:
      */
     Settings& getSettings() { return mSettings; }
 
-    void stopAnimation() { mCurrentAnimation = 0; }
+    void stopAnimation() { mCurrentAnimation = -1; }
 
     int getCurrentCamera() const { return mCurrentCamera; }
 
@@ -256,7 +256,7 @@ private:
     std::function<void()> mCustomUI;
 
     // Properties that can be changed from the UI.
-    int mCurrentAnimation = 1; // It is a 1-based index and 0 means not playing animation
+    int mCurrentAnimation = 0; // -1 means not playing animation and count means plays all of them (0-based index)
     int mCurrentVariant = 0;
     bool mEnableWireframe = false;
     int mVsmMsaaSamplesLog2 = 1;
@@ -276,7 +276,7 @@ private:
 
     // Cross fade animation parameters.
     float mCrossFadeDuration = 0.5f;  // number of seconds to transition between animations
-    int mPreviousAnimation = 0;       // one-based index of the previous animation
+    int mPreviousAnimation = -1;      // zero-based index of the previous animation
     double mCurrentStartTime = 0.0f;  // start time of most recent cross-fade (seconds)
     double mPreviousStartTime = 0.0f; // start time of previous cross-fade (seconds)
     bool mResetAnimation = true;      // set when building ImGui widgets, honored in applyAnimation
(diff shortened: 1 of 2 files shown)


NEW_RELEASE_NOTES.md
```diff
@@ -10,3 +10,4 @@ appropriate header in [RELEASE_NOTES.md](./RELEASE_NOTES.md).
 
 - engine: Added parameter for configuring JobSystem thread count
 - engine: In Java, introduce Engine.Builder
+- samples: support apply all animations in gltf_viewer
```
romainguy, line 13: We don't add release notes about samples.

romainguy (commented):
Can you please explain the purpose of this change?

libs/viewer/src/ViewerGui.cpp
```diff
@@ -1098,18 +1104,20 @@ void ViewerGui::updateUserInterface() {
         }
 
         Animator& animator = *mInstance->getAnimator();
-        if (animator.getAnimationCount() > 0 && ImGui::CollapsingHeader("Animation")) {
+        const size_t numAnimations = animator.getAnimationCount();
```
pixelflinger, line 1107:
```suggestion
        const size_t animationCount = animator.getAnimationCount();
```

libs/viewer/src/ViewerGui.cpp
```diff
@@ -523,12 +523,18 @@ void ViewerGui::applyAnimation(double currentTime, FilamentInstance* instance) {
         mResetAnimation = false;
     }
     const double elapsedSeconds = currentTime - mCurrentStartTime;
-    if (numAnimations > 0 && mCurrentAnimation > 0) {
-        animator.applyAnimation(mCurrentAnimation - 1, elapsedSeconds);
-        if (elapsedSeconds < mCrossFadeDuration && mPreviousAnimation > 0) {
+    if (numAnimations > 0 && mCurrentAnimation >= 0) {
```
pixelflinger, line 526:
```suggestion
    if (animationCount > 0 && mCurrentAnimation >= 0) {
```

I know the existing code is use `numAnimation` but that was wrong. Please update the name as part of the PR since you're modifying this code. Thank you!

libs/viewer/include/viewer/ViewerGui.h
```diff
@@ -256,7 +256,7 @@ class UTILS_PUBLIC ViewerGui {
     std::function<void()> mCustomUI;
 
     // Properties that can be changed from the UI.
-    int mCurrentAnimation = 1; // It is a 1-based index and 0 means not playing animation
+    int mCurrentAnimation = 0; // It is a 0-based index and -1 means not playing animation
```
pixelflinger, line 259: document that when set to the animation count it plays all of them (there are 2 special values, -1 and count).

pixelflinger (changes requested):

libs/viewer/src/ViewerGui.cpp
mackong: Updated.

libs/viewer/include/viewer/ViewerGui.h
mackong: Updated.

Issue #7256, opened 2023-10-10
Materials compiled with --optimize-size crashes on Android 


poweifeng:
Repro:
 - Change default [here] to `Optimization::SIZE`
 - Build release build and android/samples/sample-gltf-viewer
 - Push apk

App crashes with log:
```
10-10 16:23:16.952 16102 16160 E Filament: Compilation error in fragment shader "sao":
10-10 16:23:16.952 16102 16160 E Filament: "3:217: L0002: Undeclared variable 'mp_copy_uv'
10-10 16:23:16.952 16102 16160 E Filament: 3:219: L0002: Undeclared variable 'arg'
10-10 16:23:16.952 16102 16160 E Filament: 3:220: L0002: Undeclared variable 'arg'
10-10 16:23:16.952 16102 16160 E Filament: 3:222: L0002: Undeclared variable 'param_3'
10-10 16:23:16.952 16102 16160 E Filament: 3:223: L0002: Undeclared variable 'v'
10-10 16:23:16.952 16102 16160 E Filament: 3:224: L0002: Undeclared variable 'v'
10-10 16:23:16.952 16102 16160 E Filament: 3:225: L0002: Undeclared variable 'vv'
10-10 16:23:16.952 16102 16160 E Filament: 3:226: L0002: Undeclared variable 'param_6'
10-10 16:23:16.952 16102 16160 E Filament: 3:227: L0002: Undeclared variable 'w'
10-10 16:23:16.952 16102 16160 E Filament: 3:228: L0002: Undeclared variable 'vn'
10-10 16:23:16.952 16102 16160 E Filament: 3:229: L0002: Undeclared variable '_1087'
10-10 16:23:16.952 16102 16160 E Filament: 3:230: L0002: Undeclared variable 'w'"
10-10 16:23:16.952 16102 16160 E Filament: Link error in "sao":
10-10 16:23:16.952 16102 16160 E Filament: "Link failed because of invalid fragment shader."
```

Tested on: Pixel4XL, Pixel 6pro

[here]: https://github.com/google/filament/blob/eeb53606c81ec8c933de9e13a96d9bf4c9b75bfe/tools/matc/src/matc/Config.h#L144

PR #7260, opened 2023-10-11
Remove problematic GlslangToSpv option: emitNonSemanticShaderDebugInfo

bejado:
The `emitNonSemanticShaderDebugInfo` was causing an assertion to fire in debug builds of matc and also generating problematic code that crashed on Android.

Removing it also seems to reduce lit material sizes by around 10%:

```
matc --api opengl --platform mobile --optimize-size -o out.cmat samples/materials/texturedLit.mat
```

```
Before: 154291 bytes
After: 138673 bytes
```

Fixes https://github.com/google/filament/issues/7256


poweifeng (approved):

poweifeng:
Maybe add 

Fixes #7256

romainguy:
So we don't run the DCE at all anymore for optimize-size?

bejado:
> So we don't run the DCE at all anymore for optimize-size?

Correct

PR #7263, opened 2023-10-12
matdbg: change from websocket to GET

poweifeng:
- Use a hanging-GET approach to reduce dependency on websockets.
 - Also add mutex to protect access to MaterialRecords, which is written to/read from from multiple threads.

### DIFF
--- a/libs/matdbg/include/matdbg/DebugServer.h
+++ b/libs/matdbg/include/matdbg/DebugServer.h
@@ -24,6 +24,7 @@
 #include <private/filament/Variant.h>
 
 #include <tsl/robin_map.h>
+#include <utils/Mutex.h>
 
 class CivetServer;
 
@@ -43,9 +44,8 @@ struct MaterialRecord {
 /**
  * Server-side material debugger.
  *
- * This class manages an HTTP server and a WebSockets server that listen on a secondary thread. It
- * receives material packages from the Filament C++ engine or from a standalone tool such as
- * matinfo.
+ * This class manages an HTTP server. It receives material packages from the Filament C++ engine or
+ * from a standalone tool such as matinfo.
  */
 class DebugServer {
 public:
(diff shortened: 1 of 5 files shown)


libs/matdbg/include/matdbg/DebugServer.h
```diff
@@ -99,7 +99,10 @@ class DebugServer {
     const backend::Backend mBackend;
 
     CivetServer* mServer;
+
     tsl::robin_map<MaterialKey, MaterialRecord> mMaterialRecords;
+    mutable utils::Mutex mMaterialRecordsMutex;
```
bejado, line 104: Do we need to hold the lock inside `~DebugServer`?

libs/matdbg/src/ApiHandler.cpp
```diff
@@ -212,6 +212,34 @@ bool ApiHandler::handleGetApiShader(struct mg_connection* conn,
...
+void ApiHandler::addMaterial(MaterialRecord const* material) {
+    std::unique_lock<utils::Mutex> lock(mStatusMutex);
+    snprintf(statusMaterialId, sizeof(statusMaterialId), "%8.8x", material->key);
+    mStatusCondition.notify_all();
+}
+
+bool ApiHandler::handleGetStatus(struct mg_connection* conn,
+        struct mg_request_info const* request) {
+    char const* qstr = request->query_string;
+    char buf[1] = {};
```
bejado, line 224: Remove?

bejado (approved):
LGTM, but what's the advantage with hanging-GET as opposed to websockets?

libs/matdbg/include/matdbg/DebugServer.h
poweifeng: Yes, thanks.  I also rearranged it a little so that the server terminates before free-ing the records.

poweifeng:
> LGTM, but what's the advantage with hanging-GET as opposed to websockets?

Just a bit cleaner on the client and server. Websocket doesn't seem necessary since communication between server/client is unidirectional for the types of request using the socket (i.e. only edit request without needing a reply etc or only material update messages from server).

libs/matdbg/src/ApiHandler.cpp
poweifeng: done

PR #7272, opened 2023-10-17
add Renderer::getClearOptions

pixelflinger:
FIXES=[243846268]

### DIFF
--- a/web/filament-js/jsbindings.cpp
+++ b/web/filament-js/jsbindings.cpp
@@ -612,6 +612,7 @@ class_<Renderer>("Renderer")
         engine->execute();
     }), allow_raw_pointers())
     .function("_setClearOptions", &Renderer::setClearOptions, allow_raw_pointers())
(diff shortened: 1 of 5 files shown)


web/filament-js/jsbindings.cpp
```diff
@@ -612,6 +612,7 @@ class_<Renderer>("Renderer")
         engine->execute();
     }), allow_raw_pointers())
     .function("_setClearOptions", &Renderer::setClearOptions, allow_raw_pointers())
+    .function("getClearOptions", &Renderer::getClearOptions)
```
romainguy, line 615: No Java version? :D

romainguy (approved):

web/filament-js/jsbindings.cpp
pixelflinger: It already existed!

poweifeng (approved):

PR #7274, opened 2023-10-18
dynamic shadowmap visualization 

pixelflinger:
The directional shadowmap visualizer is implemented behind a 
specialization constant. Add the DebugRegistry infrastructure to be
able to update the spec-constant at runtime and have a subset of 
all materials invalidated.

This allows to toggle the visualization at runtime using a debug
property.

This is also a proof of concept that we can update spec-constants
at runtime; we could probably leverage this work for engine-wide
shader configurations.

Screenshot 2023-10-18 at 12 06 35

### DIFF
--- a/filament/src/details/Material.cpp
+++ b/filament/src/details/Material.cpp
@@ -247,6 +247,9 @@ FMaterial::FMaterial(FEngine& engine, const Material::Builder& builder)
                     +ReservedSpecializationConstants::CONFIG_FROXEL_BUFFER_HEIGHT,
                     (int)maxFroxelBufferHeight });
     mSpecializationConstants.push_back({
+                    +ReservedSpecializationConstants::CONFIG_DEBUG_DIRECTIONAL_SHADOWMAP,
+                    (bool)engine.debug.shadowmap.debug_directional_shadowmap });
+    mSpecializationConstants.push_back({
                     +ReservedSpecializationConstants::CONFIG_STATIC_TEXTURE_TARGET_WORKAROUND,
                     (bool)staticTextureWorkaround });
     mSpecializationConstants.push_back({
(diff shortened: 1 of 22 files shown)


filament/src/details/Material.cpp
```diff
@@ -447,6 +450,48 @@ FMaterial::~FMaterial() noexcept {
...
+                return item.id == ReservedSpecializationConstants::CONFIG_DEBUG_DIRECTIONAL_SHADOWMAP;
+            });
+    if (pos != mSpecializationConstants.end()) {
+        pos->value = mEngine.debug.shadowmap.debug_directional_shadowmap;
+    }
+
+    DriverApi& driverApi = mEngine.getDriverApi();
+    auto& cachedPrograms = mCachedPrograms;
+    for (size_t k = 0, n = VARIANT_COUNT; k < n; ++k) {
+        const Variant variant(k);
```
poweifeng, line 467:
```suggestion
        Variant const variant(k);
```

shaders/src/main.fs
```diff
@@ -66,6 +66,26 @@ void main() {
...
+#if defined(VARIANT_HAS_SHADOWING) && defined(VARIANT_HAS_DIRECTIONAL_LIGHTING)
+    if (CONFIG_DEBUG_DIRECTIONAL_SHADOWMAP) {
+        float a = fragColor.a;
+        highp vec4 p = getShadowPosition(getShadowCascade());
+        p.xy = p.xy * (1.0 / p.w);
+        if (p.xy != saturate(p.xy)) {
+            vec4 c = vec4(1.0, 0, 1.0, 1.0) * a;
+            fragColor = mix(fragColor, c, 0.2);
+        } else {
+            p.xy = saturate(p.xy);
```
poweifeng, line 78: isn't this already true if we're in this conditional block?

poweifeng (approved):

shaders/src/main.fs
pixelflinger: yes! good catch.

PR #7276, opened 2023-10-18
Allow explicitly initializing at feature level 0

elizagamedev:
This change does three main things. First, it adds an option to the Engine Builder to pick the feature level at which to instantiate Filament. The only real practical purpose of allowing this is to be able to instantiate at feature level 0. Secondly, it allows feature level 0 to properly work on non-ES2 devices. Thirdly, it changes both Android and desktop hellotriangle samples to explicitly opt-in to feature level 0.

Unfortunately, feature levels are used in two different, somewhat contradictory ways presently in Filament, which can make reasoning about this change a bit confusing. From a client perspective, feature levels refer to buckets of capabilities which are guaranteed to be supported. Internally, there is a separate "feature level" stored internally at the Driver subclass level which generally corresponds to the maximum supported feature level, but is also referenced when activating workarounds for limited devices. For example, Uniform Buffer Objects are not supported in ES2, however, Filament supports emulating them such that the client does not need to care at all; a supported feature is a supported feature. But internally, Filament uses this "Driver" feature level to determine whether or not a given workaround is needed. There were several cases where the "active feature level" was being examined in order to activate these workarounds rather than the "driver feature level", which was incorrect.

Why should non-ES2-only devices want to activate feature level 0? Allowing this behavior 1. makes feature level 0 more consistent with the behavior of other feature levels and 2. allows clients a layer of validation that their software will work on all devices supported by Filament if they explicitly opt into it.

Consistency: Filament guarantees that any given device which supports a given feature level will also support running on every feature level below, except for feature level 0. This change removes that exception.

Validation: It's not perfect, and there will likely be bugs and unexpected differences in behavior between ES2 and non-ES2 devices that crop up in the future between two devices running on the same feature level. However, it's at least a basic high level layer of validation that enables more rapid testing workflows directly via desktop versions of Filament rather than having to fiddle with something like ANGLE to get perfect GLES 2.0 compliance. Additionally, it expands options for automated testing (with the same caveats).

This change has been tested on both the desktop and Android versions of hellotriangle.

### DIFF
--- a/android/samples/sample-hello-triangle/src/main/java/com/google/android/filament/hellotriangle/MainActivity.kt
+++ b/android/samples/sample-hello-triangle/src/main/java/com/google/android/filament/hellotriangle/MainActivity.kt
@@ -110,7 +110,7 @@ class MainActivity : Activity() {
     }
 
     private fun setupFilament() {
-        engine = Engine.create()
+        engine = Engine.Builder().featureLevel(Engine.FeatureLevel.FEATURE_LEVEL_0).build()
         renderer = engine.createRenderer()
         scene = engine.createScene()
         view = engine.createView()
@@ -120,13 +120,8 @@ class MainActivity : Activity() {
     private fun setupView() {
         scene.skybox = Skybox.Builder().color(0.035f, 0.035f, 0.035f, 1.0f).build(engine)
 
-        if (engine.activeFeatureLevel == Engine.FeatureLevel.FEATURE_LEVEL_0) {
-            // post-processing is not supported at feature level 0
-            view.isPostProcessingEnabled = false
-        } else {
-            // NOTE: Try to disable post-processing (tone-mapping, etc.) to see the difference
-            // view.isPostProcessingEnabled = false
-        }
+        // post-processing is not supported at feature level 0
+        view.isPostProcessingEnabled = false
 
         // Tell the view which camera we want to use
         view.camera = camera
--- a/filament/src/Froxelizer.cpp
+++ b/filament/src/Froxelizer.cpp
@@ -104,11 +104,11 @@ Froxelizer::Froxelizer(FEngine& engine)
     static_assert(std::is_same_v<RecordBufferType, uint8_t>,
             "Record Buffer must use bytes");
 
-    if (UTILS_UNLIKELY(engine.getActiveFeatureLevel() == FeatureLevel::FEATURE_LEVEL_0)) {
+    DriverApi& driverApi = engine.getDriverApi();
+
+    if (UTILS_UNLIKELY(driverApi.getFeatureLevel() == FeatureLevel::FEATURE_LEVEL_0)) {
         return;
     }
-
-    DriverApi& driverApi = engine.getDriverApi();
 
     mFroxelBufferEntryCount = std::min(
             FROXEL_BUFFER_MAX_ENTRY_COUNT,
--- a/filament/src/details/Engine.cpp
+++ b/filament/src/details/Engine.cpp
@@ -1170,6 +1170,8 @@ Engine::FeatureLevel FEngine::getSupportedFeatureLevel() const noexcept {
 Engine::FeatureLevel FEngine::setActiveFeatureLevel(FeatureLevel featureLevel) {
     ASSERT_PRECONDITION(featureLevel <= getSupportedFeatureLevel(),
             "Feature level %u not supported", (unsigned)featureLevel);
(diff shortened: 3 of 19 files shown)


android/samples/sample-hello-triangle/src/main/java/com/google/android/filament/hellotriangle/MainActivity.kt
```diff
@@ -120,13 +120,8 @@ class MainActivity : Activity() {
...
         scene.skybox = Skybox.Builder().color(0.035f, 0.035f, 0.035f, 1.0f).build(engine)
 
-        if (engine.activeFeatureLevel == Engine.FeatureLevel.FEATURE_LEVEL_0) {
-            // post-processing is not supported at feature level 0
-            view.isPostProcessingEnabled = false
-        } else {
-            // NOTE: Try to disable post-processing (tone-mapping, etc.) to see the difference
-            // view.isPostProcessingEnabled = false
-        }
+        // post-processing is not supported at feature level 0
```
pixelflinger, line 123: I think this is okay -- but it was nice to have a sample that would adjust itself to the supported feature level automatically; like what a real app would do.

filament/src/Froxelizer.cpp
```diff
@@ -104,12 +104,12 @@ Froxelizer::Froxelizer(FEngine& engine)
     static_assert(std::is_same_v<RecordBufferType, uint8_t>,
             "Record Buffer must use bytes");
 
-    if (UTILS_UNLIKELY(engine.getActiveFeatureLevel() == FeatureLevel::FEATURE_LEVEL_0)) {
+    DriverApi& driverApi = engine.getDriverApi();
+
+    if (UTILS_UNLIKELY(driverApi.getFeatureLevel() == FeatureLevel::FEATURE_LEVEL_0)) {
```
pixelflinger, line 109: Yes! Do we agree that we have an invariant that driver feature level 0 *implies* engine feature level 0, but the reverse is no longer true (it was before because it wasn't possible to lower the feature level, which started at least at 1 on non ES2 devices).

pixelflinger (approved):

filament/src/Froxelizer.cpp
elizagamedev: Yes, this seems to me like the correct way to think about the state of the engine now.

android/samples/sample-hello-triangle/src/main/java/com/google/android/filament/hellotriangle/MainActivity.kt
elizagamedev: I could always leave this bit in, even though it doesn't effectively do anything, just to demonstrate how you'd implement it.

elizagamedev:
I pushed another commit to explicitly prevent upgrading from feature level 0 at runtime, since level 0 is initialized in such a way that upgrading during runtime would break things. Unfortunately this means that feature level 0 is still slightly idiosyncratic compared to other feature levels.

bejado (approved):

pixelflinger:
> I pushed another commit to explicitly prevent upgrading from feature level 0 at runtime, since level 0 is initialized in such a way that upgrading during runtime would break things. Unfortunately this means that feature level 0 is still slightly idiosyncratic compared to other feature levels.

I think that's okay because FL0 is just here to test our code and in real life we won't never upgrade to FL1+.

filament/src/details/Engine.cpp
```diff
@@ -1145,6 +1145,11 @@ Engine::FeatureLevel FEngine::getSupportedFeatureLevel() const noexcept {
 Engine::FeatureLevel FEngine::setActiveFeatureLevel(FeatureLevel featureLevel) {
     ASSERT_PRECONDITION(featureLevel <= getSupportedFeatureLevel(),
             "Feature level %u not supported", (unsigned)featureLevel);
+    if (mActiveFeatureLevel <= FeatureLevel::FEATURE_LEVEL_0) {
+        // Prevent upgrading from feature level 0, as it is initialized in an incompatible way from
+        // higher feature levels.
+        return FeatureLevel::FEATURE_LEVEL_0;
+    }
```
pixelflinger, line 1152: I think this should be a `ASSERT_POSTCONDITION` (or precondition if we document it).

pixelflinger (changes requested):

filament/src/details/Engine.cpp
elizagamedev: Done, including revising the documentation comments.

pixelflinger (approved):

PR #7277, opened 2023-10-18
vk: support stencil format in swapchain

### LINKED ISSUE #7233: Vulkan: Implement stencil buffer support for the default render target
https://github.com/google/filament/pull/7227 allocates a default stencil buffer for Metal for the default render target if the `SwapChain::CONFIG_HAS_STENCIL_BUFFER` flag is set on the `SwapChain`. Other backends should behave similarly.

poweifeng:
Fixes #7233
FIXES=302197523

### DIFF
--- a/filament/backend/src/vulkan/platform/VulkanPlatformSwapChainImpl.cpp
+++ b/filament/backend/src/vulkan/platform/VulkanPlatformSwapChainImpl.cpp
@@ -30,7 +30,7 @@ namespace {
 
 std::tuple<VkImage, VkDeviceMemory> createImageAndMemory(VulkanContext const& context,
         VkDevice device, VkExtent2D extent, VkFormat format) {
-    bool const isDepth = isDepthFormat(format);
+    bool const isDepth = isVkDepthFormat(format);
     // Filament expects blit() to work with any texture, so we almost always set these usage flags.
     // TODO: investigate performance implications of setting these flags.
     VkImageUsageFlags const blittable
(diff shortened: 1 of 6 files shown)


filament/backend/src/vulkan/platform/VulkanPlatformSwapChainImpl.cpp
```diff
@@ -152,10 +161,15 @@ VkResult VulkanPlatformSurfaceSwapChain::create() {
     // Find a suitable surface format.
     FixedCapacityVector<VkSurfaceFormatKHR> const surfaceFormats
             = enumerate(vkGetPhysicalDeviceSurfaceFormatsKHR, mPhysicalDevice, mSurface);
-    FixedCapacityVector<VkFormat> expectedFormats
-            = {VK_FORMAT_R8G8B8A8_UNORM, VK_FORMAT_B8G8R8A8_UNORM};
+    FixedCapacityVector<VkFormat> expectedFormats = {
```
pixelflinger, line 164: why not use a `std::array<>` here? `FixedCapacityVector` does heap allocation which I don't think we should do for 2 elements.

pixelflinger (approved):
lgtm except maybe the use of FixedCapacityVector vs std::array?

filament/backend/src/vulkan/platform/VulkanPlatformSwapChainImpl.cpp
poweifeng: done.  good point.

PR #7285, opened 2023-10-20
Fix the dangling pointer to the destroyed sampler group

awkr:
This PR aims to fix the two scenarios:

1. The dangling pointer to the destroyed sampler group, which may cause a crash when the program updates samplers;
2. Besides scenario 1, there may be such a requirement, that is, although the sampler is defined in the shader, there is no need to bind any sampler during rendering due to business needs.

Hope it makes sense :)

### DIFF
--- a/filament/backend/src/opengl/OpenGLProgram.cpp
+++ b/filament/backend/src/opengl/OpenGLProgram.cpp
@@ -211,6 +211,9 @@ void OpenGLProgram::updateSamplers(OpenGLDriver* const gld) const noexcept {
         size_t const binding = usedBindingPoints[i];
         assert_invariant(binding < Program::SAMPLER_BINDING_COUNT);
         auto const * const sb = samplerBindings[binding];
+        if (!sb) { // A sampler may not need to be used, although it's declared in the shader
+            continue;
+        }
         assert_invariant(sb);
         for (uint8_t j = 0, m = sb->textureUnitEntries.size(); j < m; ++j, ++tmu) { // "<=" on purpose here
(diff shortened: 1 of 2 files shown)


filament/backend/src/opengl/OpenGLProgram.cpp
```diff
@@ -211,6 +211,9 @@ void OpenGLProgram::updateSamplers(OpenGLDriver* const gld) const noexcept {
         size_t const binding = usedBindingPoints[i];
         assert_invariant(binding < Program::SAMPLER_BINDING_COUNT);
         auto const * const sb = samplerBindings[binding];
+        if (!sb) { // A sampler may not need to be used, although it's declared in the shader
+            continue;
+        }
```
pixelflinger, line 216: We should never allow this because it's not allowed in Vulkan and Metal. Also there is an assert for this just below.
If there are scenarios where this happens, we should fix those.

pixelflinger (changes requested):

filament/backend/src/opengl/OpenGLProgram.cpp
awkr: Ok I see, but it may happen in OpenGL.

pixelflinger: If it happens we need to understand why and not allow it to happen.

pixelflinger (changes requested):

pixelflinger:
Thank you for this contribution. I'm closing it because I have included this fix in an upcoming PR.

Issue #7269, opened 2023-10-16
gltfio: Can't create multiple asset instances


ThomasGorisse:
Hi,

I have an issue with `AssetLoader.createInstance(asset)`:
This works as intended:
```kotlin
val asset1 = assetLoader.createAsset(assetBuffer1)
val instance1 = assetLoader.createInstance(asset1)
val instance2 = assetLoader.createInstance(asset1)
```
This generate an AssertionError:
```kotlin
val asset1 = assetLoader.createAsset(assetBuffer1)
val instance1 = assetLoader.createInstance(asset1)
val asset2 = assetLoader.createAsset(assetBuffer2)
val instance2 = assetLoader.createInstance(asset1)
```
Thrown error:
```
Fatal signal 11 (SIGSEGV), code 2 (SEGV_ACCERR), fault addr 0x7a3a23b22c in tid 29038 (e.armodelviewer), pid 29038 (e.armodelviewer)
2023-10-16 14:47:48.721 29310-29310 DEBUG                   pid-29310                            A  Cmdline: io.github.sceneview.sample.armodelviewer
2023-10-16 14:47:48.721 29310-29310 DEBUG                   pid-29310                            A  pid: 29038, tid: 29038, name: e.armodelviewer  >>> io.github.sceneview.sample.armodelviewer <<<
2023-10-16 14:47:48.722 29310-29310 DEBUG                   pid-29310                            A        #00 pc 000000000024c38c  /data/app/~~2dYpOZ4DiMHGPPMEwyHZMw==/io.github.sceneview.sample.armodelviewer-Oo7jCdrsbYxdtx3mcNWFDg==/base.apk!libgltfio-jni.so (offset 0x2c69000) (BuildId: f801755d82cf91f2da2b6ebd6f98f3df02d4c58d)
```
Precision: I never call `asset1.releaseSourceData()` and well load resources (no other issue with the instance rendering)

PR #7288, opened 2023-10-23
Fix creating multiple asset instances #7269

z3moon:


bejado (approved):

libs/gltfio/src/AssetLoader.cpp
```diff
@@ -343,95 +341,130 @@ FFilamentAsset* FAssetLoader::createInstancedAsset(const uint8_t* bytes, uint32_
...
         return nullptr;
     }
 
-    createRootAsset(sourceAsset);
+    FFilamentAsset* fAsset = createRootAsset(sourceAsset);
     if (mError) {
-        delete mAsset;
-        mAsset = nullptr;
+        delete fAsset;
+        fAsset = nullptr;
```
poweifeng, line 355: nit: can be omitted IMO.  Same for similar lines below.

poweifeng (approved):

libs/gltfio/src/AssetLoader.cpp
z3moon: Thanks for the suggestion! I agree, we can improve the way of handling errors during asset loading in a separate commit.

PR #7296, opened 2023-10-24
Miscellaneous feature level 0 fixes

elizagamedev:
Fix edge case where an empty struct could be generated in an ESSL 1.0 shader.

Include _maskThreshold and _doubleSided in ESSL 1.0 shaders.

Add GL_OES_standard_derivatives extension to ESSL 1.0 shaders. According to gpuinfo.org, this has 96% device coverage and supports both Mali-400 and Adreno (TM) 304.

Remove 3D sampler support from ESSL 1.0 shaders. This extension is only supported by 62% of devices.

Change filagui material to a FL0 material.

### DIFF
--- a/filament/src/materials/blitLow.mat
+++ b/filament/src/materials/blitLow.mat
@@ -23,7 +23,6 @@ material {
     depthWrite : false,
     depthCulling : false,
     domain: postprocess,
-    shadingModel : unlit,
     featureLevel : 0
 }
 
--- a/filament/src/materials/blitLow.mat
+++ b/filament/src/materials/blitLow.mat
@@ -22,7 +22,9 @@ material {
     ],
     depthWrite : false,
     depthCulling : false,
-    domain: postprocess
+    domain: postprocess,
+    shadingModel : unlit,
+    featureLevel : 0
 }
 
 vertex {
@@ -34,7 +36,10 @@ vertex {
 
 fragment {
     void postProcess(inout PostProcessInputs postProcess) {
+#if __VERSION__ == 100
+        postProcess.color = texture2D(materialParams_color, variable_vertex.xy);
+#else
         postProcess.color = textureLod(materialParams_color, variable_vertex.xy, 0.0);
+#endif
     }
 }
-
--- a/NEW_RELEASE_NOTES.md
+++ b/NEW_RELEASE_NOTES.md
@@ -12,4 +12,6 @@ appropriate header in [RELEASE_NOTES.md](./RELEASE_NOTES.md).
 - matc: Enable `GL_OES_standard_derivatives` extension in ESSL 1.0 shaders
 - matc: Fix code generation of double sided and masked materials in ESSL 1.0 shaders
 - filagui: Add support for feature level 0
-- mtac: Add support for post-process materials in feature level 0
+- matc: Add support for post-process materials in feature level 0
+- engine: Add `Material::getFeatureLevel()`
+- engine: Add missing `Material::getReflectionMode()` method in Java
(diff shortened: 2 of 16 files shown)


elizagamedev:
I pushed a change which allows post-processing materials to be built for feature level 0; see the commit message for details.

filament/src/materials/blitLow.mat
```diff
@@ -22,7 +22,9 @@ material {
     ],
     depthWrite : false,
     depthCulling : false,
-    domain: postprocess
+    domain: postprocess,
+    shadingModel : unlit,
```
pixelflinger, line 26: ShadingModel should be irrelevant for a `postprocess` domain material. We shouldn't have to specify it. I think internally we can just make it `unlit` automatically.

pixelflinger (changes requested):

filament/src/materials/blitLow.mat
elizagamedev: Alright, I'll take care of this tomorrow.

elizagamedev: Done in the latest commit.

elizagamedev:
I pushed another commit. Sorry to keep introducing more and more for this PR, but this should be the last, since I have Filament running on ES2 now with some basic postprocessing.

I wrote basically an entire essay in the commit description, so I'm replicating it below:

First, this commit introduces some very simple bugfixes regarding ES2
compatibility related to postprocessing.

Second, this commit adds support for creating textures specified as R8, SRGB8,
and SRGB8_A8 in ES2. R8 is trivial: just use GL_LUMINANCE instead. The sRGB
formats, however, are maybe a bit more controversial. As implemented, they
instead just use the equivalent non-sRGB formats. This is of course technically
incorrect. There are a few approaches to how to add sRGB compatibility for ES2
that I can think of.

1. Do a bunch of complex shader nonsense in matc. Maybe even traversing the AST
and ensuring any texture lookup of a texture flagged as sRGB uses some
compatibility function. This would require static analysis to track if samplers
are reassigned to another variable, for example. This of course also breaks down
if you don't know at compile time if the shader will receive an RGB or an sRGB
sampler, or if the shader should be able to support both RGB or sRGB samplers.
Really only worth mentioning here for the sake of completion.

2. You could also generate simple compatibility functions to look up each
sampler, which would only apply to FL0 materials.

First, we would have to extend the material format to be able to explicitly
"color" a sampler as sRGB or not, like:

```
parameters : [
    {
        type : sampler2d,
        name : albedo,
        precision : medium,
        colorSpace : srgb,
    },
    {
        type : sampler2d,
        name : normal,
        precision : medium,
        colorSpace : linear,
    }
],
```

Then, the following GLSL code would be generated.

```glsl
\#if __VERSION__ == 100
vec4 texture_albedo(vec2 position) {
  return sRGBtoLinear(texture2D(materialParams_albedo, position));
}
vec4 texture_normal(vec2 position) {
  return texture2D(materialParams_albedo, position);
}
\#else
vec4 texture_albedo(vec2 position) {
  return texture(materialParams_albedo, position);
}
vec4 texture_normal(vec2 normal) {
  return texture(materialParams_normal, position);
}
\#endif
```

Finally, at runtime, if a sampler is "colored" one way or the other, we would
verify that only the appropriate kinds of samplers are bound.

I'm actually very partial to this solution. Since sRGB compatibility is only a
concern on ES2, we can generate this code only for FL0 shaders, which already
require GLSL shader authors to care about ESSL 1.0 compatibility by calling the
appropriate `textureXX` functions. Additionally, it provides a layer of
high-level validation that texture lookups are correct, even if a real ES2
context is not available on the device being tested.

3. Leave it entirely up to the client. (What this commit does.) This leaves
client code ripe for making mistakes, but luckily, we can go back and do
solution 2 whenever. If specifying a color space for a sampler remains optional,
then if this feature is retrofitted in the future, client code will continue to
compile.

bejado (approved):

PR #7298, opened 2023-10-24
vk: remove unused platform GGP

poweifeng:


bejado:
I'm fine with removing, but this was added by a 3P contributor, FYI (#6235). Is it causing a maintenance burden?

poweifeng:
> I'm fine with removing, but this was added by a 3P contributor, FYI (#6235). Is it causing a maintenance burden?

It's been removed internally, therefore removing it here as well.

pixelflinger (approved):

PR #7301, opened 2023-10-25
matdbg: UI refresh

poweifeng:


libs/matdbg/src/DebugServer.cpp
```diff
@@ -50,13 +50,28 @@ using utils::FixedCapacityVector;
...
-#if !SERVE_FROM_SOURCE_TREE
 #include "matdbg_resources.h"
-#endif
+#include <unordered_map>
+
+namespace {
+
+struct Asset {
+    utils::CString mime;
+    utils::CString data;
```
bejado, line 68: I wonder if we could use `std::string_view`s here instead of `CString`, since all of this data is constant. This way we avoid copying the data in the `CString` constructor.

bejado (approved):

libs/matdbg/src/DebugServer.cpp
poweifeng: good call, done

pixelflinger (approved):

PR #7325, opened 2023-11-01
matdbg: fix matinfo use case

poweifeng:
- Make sure matinfo works by selecting a default backend in the absence of activeShaders.
 - Add options to select backend in matinfo mode.
 - Workaround cursor misplacement for monaco
 - Refactor menu sections into a common element.


bejado (approved):
Thank you for the fix!

PR #7328, opened 2023-11-01
Support up to 4 side-by-side stereoscopic eyes

bejado:
This change allows clients to render up to `CONFIG_MAX_STEREOSCOPIC_EYES` side-by-side stereoscopic eyes. This maximum is currently 4, though this can be raised even higher in the future. The number of eyes rendered is configured in `Engine::Config::stereoscopicEyeCount` when the Filament `Engine` is initialized. There is no restriction on what the eyes are used for, however, the intended use-case is to support foveated rendering with 4 eyes.

To test, run `gltf_viewer --eyes=4`.

Screenshot 2023-11-01 at 4 34 08 PM


samples/gltf_viewer.cpp
```diff
@@ -946,25 +965,27 @@ int main(int argc, char** argv) {
...
-            mat4 projections[2];
-            // Use an aspect ratio of 1.0. The viewport will be taken into account in
-            // FilamentApp.cpp.
-            projections[0] = mat4::perspective(70.0, 1.0, .1, 10.0);
-            projections[1] = mat4::perspective(70.0, 1.0, .1, 10.0);
-            c.setCustomEyeProjection(projections, 2, projections[0], .1, 10.0);
-            // FIXME: the aspect ratio will be incorrect until configureCamerasForWindow is
-            // triggered, which will happen the next time the window is resized.
+            const mat4 modelMatrices[4] = { rightEye, leftEye, rightEye, leftEye };
+            for (int i = 0; i < std::min(app.config.stereoscopicEyeCount, 4); i++) {
```
poweifeng, line 976: nit: there's a mismatch here with `CONFIG_MAX_STEREOSCOPIC_EYES` if say `CONFIG_MAX_STEREOSCOPIC_EYES` needs to raised to 6 in the future.  Maybe we should just use `4` in the `std::cerr` above?

poweifeng (approved):

filament/backend/src/opengl/OpenGLContext.cpp
```diff
@@ -284,6 +284,7 @@ void OpenGLContext::setDefaultState() noexcept {
 
     if (ext.EXT_clip_cull_distance) {
         glEnable(GL_CLIP_DISTANCE0);
+        glEnable(GL_CLIP_DISTANCE1);
     }
```
pixelflinger, line 288: I'm a little concerned having the clip planes always enabled even when not needed; could they have an impact on performance or even cause bugs (in the vast majority of cases where we don't use stereo).
Maybe we should formally add a enable/disable for clip planes in the backend API? I guess that will depend on how does this work on Metal or Vulkan (we can do that in another PR).

filament/include/filament/Engine.h
```diff
@@ -291,6 +291,14 @@ class UTILS_PUBLIC Engine {
...
          */
         size_t textureUseAfterFreePoolSize = 0;
+
+        /*
+         * The number of eyes to render when stereoscopic rendering is enabled. Supported values are
+         * between 1 and CONFIG_MAX_STEREOSCOPIC_EYES (inclusive).
+         *
+         * @see View::setStereoscopicOptions
+         */
+        uint8_t stereoscopicEyeCount = 2;
```
pixelflinger, line 301: I checked `OVR_multiview` and it indeed needs to know the number of views in vertex shaders, so we do need to know this value when we create materials if we want to support it someday. This is unfortunate because if it wasn't for that, this setting could be in `View` (where it would make much more sense).
I think there should be a comment about that somewhere (in a cpp).

filament/src/details/Camera.cpp
```diff
@@ -96,11 +96,11 @@ void UTILS_NOINLINE FCamera::setCustomProjection(mat4 const& p,
...
 void UTILS_NOINLINE FCamera::setCustomEyeProjection(math::mat4 const* projection, size_t count,
         math::mat4 const& projectionForCulling, double near, double far) {
-    ASSERT_PRECONDITION(count >= CONFIG_STEREOSCOPIC_EYES,
+    const Engine::Config& config = mEngine.getConfig();
+    ASSERT_PRECONDITION(count >= config.stereoscopicEyeCount,
             "All eye projections must be supplied together, count must be >= "
-            "CONFIG_STEREOSCOPIC_EYES(%d)",
-            CONFIG_STEREOSCOPIC_EYES);
-    for (uint8_t i = 0; i < CONFIG_STEREOSCOPIC_EYES; i++) {
+            "config.stereoscopicEyeCount (%d)", config.stereoscopicEyeCount);
```
pixelflinger, line 102: I think `count` should be either exactly `1` or `config.stereoscopicEyeCount` (and also documented), because it's an `Engine` wide setting and with `OVR_multiview` shaders won't be able to do less.

filament/src/details/Camera.cpp
```diff
@@ -165,7 +165,8 @@ void UTILS_NOINLINE FCamera::setProjection(Camera::Projection projection,
 }
 
 math::mat4 FCamera::getProjectionMatrix(uint8_t eye) const noexcept {
-    assert_invariant(eye < CONFIG_STEREOSCOPIC_EYES);
+    UTILS_UNUSED_IN_RELEASE const Engine::Config& config = mEngine.getConfig();
```
pixelflinger, line 168: I'm starting to think that `Camera` should be created with an eye count, 1 by default. It would be cleaner if Camera knew how many eyes it has, instead of having to ask the engine. Also we could validate things better, for example that the camera matches the view's stereo mode.

filament/src/fg/FrameGraph.h
```diff
@@ -527,7 +527,7 @@ class FrameGraph {
 
 template<typename Data, typename Setup, typename Execute>
 FrameGraphPass<Data>& FrameGraph::addPass(char const* name, Setup setup, Execute&& execute) {
-    static_assert(sizeof(Execute) < 1024, "Execute() lambda is capturing too much data.");
+    static_assert(sizeof(Execute) < 2048, "Execute() lambda is capturing too much data.");
```
pixelflinger, line 530: ouch. why was this needed? Which pass has the issue?

samples/gltf_viewer.cpp
pixelflinger, line 976: probably `Engine` needs a query for the max number of eyes supported.

pixelflinger (changes requested):

z3moon (approved):

samples/gltf_viewer.cpp
bejado: Yeah, good point. In fact I can simply make "even" eyes be the right eye and "odd" eyes be the left eye. That way, this will always work no matter the eye count.

filament/src/fg/FrameGraph.h
bejado: It's the `prepareShadowPass`, which now requires 1064 bytes. Maybe we can raise it to something like 1128.

pixelflinger: Ah yes because CameraInfo is huge now. Okay. I'm fine with 2048. This was there to catch unintended captures, at the time 1KiB seemed reasonable. Because this is deferred we need to capture by copy.

filament/src/details/Camera.cpp
bejado: Yeah I think I agree. Let's do this in a separate PR, since that will be another API change?

filament/src/details/Camera.cpp
bejado: Why 1? This API method requires all eyes to be set at the same time, so I think we can require it to be exactly `config.stereoscopicEyeCount`.

filament/include/filament/Engine.h
bejado: Could we not use a spec constant with `OVR_multiview`?

filament/backend/src/opengl/OpenGLContext.cpp
bejado: Yeah sounds good. Metal doesn't need to do anything to enable them. Not sure about Vulkan.

filament/include/filament/Engine.h
pixelflinger: So we could use a spec-constant, but that needs to be known at material creation time because we can't update spec-constant afterwards (technically we can by invalidating the cache, but we need to flush this out more).
There is also a small hurdle which is that GLSL accepts const expressions, but ESSL doesn't :-(  So for GLES we'll have to hack a #define.

filament/src/details/Camera.cpp
pixelflinger: okay

filament/src/details/Camera.cpp
pixelflinger: sgtm

pixelflinger (approved):

filament/include/filament/Engine.h
bejado: > so we do need to know this value when we create materials if we want to support it someday

Do you mean the material author needs to know this value, or Filament needs to know this value when a material is built?

PR #7331, opened 2023-11-02
engine: move setFrontFaceWindingInverted from View to MaterialInstance

poweifeng:
Moving setFrontFaceWindingInverted to MaterialInstance will enable finer control over face inversion and aligns better with Vulkan's pipeline definition (see VkGraphicsPipelineCreateInfo).


filament/src/RenderPass.cpp
```diff
@@ -579,10 +576,12 @@ RenderPass::Command* RenderPass::generateCommandsImpl(uint32_t extraFlags,
             auto const& morphTargets = morphing.targets[pi];
             FMaterialInstance const* const mi = primitive.getMaterialInstance();
             FMaterial const* const ma = mi->getMaterial();
+            bool const invertedFrontFaces =
+                    mi->isFrontFaceWindingInverted() ^ soaVisibility[i].reversedWindingOrder;
```
pixelflinger, line 580: `soaVisibility[i]` can be hoisted outside the loop.

web/filament-js/jsbindings.cpp
```diff
@@ -1411,7 +1411,10 @@ class_<MaterialInstance>("MaterialInstance")
     .function("setStencilWriteMask", EMBIND_LAMBDA(void,
             (MaterialInstance* self, uint8_t writeMask), {
                 self->setStencilWriteMask(writeMask, backend::StencilFace::FRONT_AND_BACK);
-            }), allow_raw_pointers());
+            }), allow_raw_pointers())
+    .function("setFrontFaceWindingInverted", &MaterialInstance::setFrontFaceWindingInverted)
+    .function("isFrontFaceWindingInverted", &MaterialInstance::isFrontFaceWindingInverted);
```
pixelflinger, line 1416: we don't need to remove the View's version? were they missing?

pixelflinger (changes requested):
In this particular loop I like to be explicit about the dereference that can be hoisted outside it.

filament/src/RenderPass.cpp
poweifeng: done

web/filament-js/jsbindings.cpp
poweifeng: yes, the View methods were missing.

pixelflinger (approved):

PR #7332, opened 2023-11-02
Cleanup RenderPassNode::declareRenderTarget

z3moon:
There's no functional change. Remove an unused local variable `outgoingEdges` to save CPU resource. Tidy up an usage of local variable to slightly improve readability.

### DIFF
--- a/filament/src/fg/PassNode.cpp
+++ b/filament/src/fg/PassNode.cpp
@@ -90,18 +90,18 @@ uint32_t RenderPassNode::declareRenderTarget(FrameGraph& fg, FrameGraph::Builder
     auto incomingEdges = dependencyGraph.getIncomingEdges(this);
 
     for (size_t i = 0; i < RenderPassData::ATTACHMENT_COUNT; i++) {
-        FrameGraphId<FrameGraphTexture> const& cur_attachment_handle =
+        FrameGraphId<FrameGraphTexture> const& handle =
                 data.descriptor.attachments.array[i];
-        if (cur_attachment_handle) {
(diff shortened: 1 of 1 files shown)


filament/src/fg/PassNode.cpp
```diff
@@ -82,26 +82,26 @@ uint32_t RenderPassNode::declareRenderTarget(FrameGraph& fg, FrameGraph::Builder
...
     // to compute the discard flags.
 
     DependencyGraph const& dependencyGraph = fg.getGraph();
     auto incomingEdges = dependencyGraph.getIncomingEdges(this);
-    auto outgoingEdges = dependencyGraph.getOutgoingEdges(this);
 
     for (size_t i = 0; i < RenderPassData::ATTACHMENT_COUNT; i++) {
-        if (descriptor.attachments.array[i]) {
-            data.attachmentInfo[i] = attachments.array[i];
+        FrameGraphId<FrameGraphTexture> const& cur_attachment_handle =
```
poweifeng, line 93: nit: prefer camelCase for most names in filament.

pixelflinger: yes you can just call it `handle`.

pixelflinger (changes requested):
just rename `cur_attachment_handle` to just `handle`.

filament/src/fg/PassNode.cpp
z3moon: Oh, you're right. I'm still getting used to it. Let me fix it. :)

poweifeng (approved):

PR #7341, opened 2023-11-09
bugfix: fix the invalid pointer cast.

HandsomeXi:
According to my understanding, The BufferDescriptor::make func 's parameter T* should be a pointer to T, not a pointer to a 
 T pointer. So the static_cast in the lambda should cast to T*.  This error can lead to memory errors that are difficult to troubleshoot.


pixelflinger:
Very good catch! I will close this PR and upload a new one that also fixes all the other places where this bug exists (there are a few other), and also cleanup some areas that don't have this bug bug where the code is misleading.

Thanks a lot!

PR #7347, opened 2023-11-10
Move stereoscopic debug options to Settings framework

### LINKED ISSUE #7330: stereo option in ViewerGui doesn't use the Settings framework
This code

```
        StereoscopicOptions stereoOptions = mView->getStereoscopicOptions();
        ImGui::Checkbox("Instanced stereo", &stereoOptions.enabled);
        if (stereoOptions.enabled) {
            ImGui::SliderFloat("Ocular distance", &mOcularDistance, 0.0f, 10.0f);

        }
        mView->setStereoscopicOptions(stereoOptions);
```

in `ViewerGui.cpp` is wrong because it will activate stereo on the UI context (when running on Web with remote ui). Like with all other settings, stereo must use the `Settings` framework or be moved to  the `Debug` tab. However, for testing it would be much better that we can use this on devices.

bejado:
This also adds a public `Engine::getConfig` API, which helps simplify things.

Fixes #7330

### DIFF
--- a/libs/viewer/src/Settings.cpp
+++ b/libs/viewer/src/Settings.cpp
@@ -302,6 +302,8 @@ static int parse(jsmntok_t const* tokens, int i, const char* jsonChunk, ViewSett
             i = parse(tokens, i + 1, jsonChunk, &out->vsmShadowOptions);
         } else if (compare(tok, jsonChunk, "postProcessingEnabled") == 0) {
             i = parse(tokens, i + 1, jsonChunk, &out->postProcessingEnabled);
+        } else if (compare(tok, jsonChunk, "stereoscopicOptions") == 0) {
+            i = parse(tokens, i + 1, jsonChunk, &out->stereoscopicOptions);
         } else {
             slog.w << "Invalid view setting key: '" << STR(tok, jsonChunk) << "'" << io::endl;
             i = parse(tokens, i + 1);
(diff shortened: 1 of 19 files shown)


libs/viewer/src/Settings.cpp
```diff
@@ -646,6 +651,20 @@ void applySettings(Engine* engine, const ViewerOptions& settings, Camera* camera
         camera->setFocusDistance(settings.cameraFocusDistance);
     }
     engine->setAutomaticInstancingEnabled(settings.autoInstancingEnabled);
+
+    // Eyes are rendered from left-to-right, i.e., eye 0 is rendered to the left side of the
+    // window.
+    // For testing, we want to render a side-by-side layout so users can view with
+    // "cross-eyed" stereo.
+    // For cross-eyed stereo, Eye 0 is really the RIGHT eye, while Eye 1 is the LEFT eye.
```
z3moon, line 659: Just curious, is there any reason why we didn't match the order of the eyes on the screen with the order of the array?

pixelflinger (changes requested):
don't forget the java and js bindings.

pixelflinger (approved):
LGTM but needs java and js bindings.

poweifeng (approved):

libs/viewer/src/Settings.cpp
bejado: So in fact we do, and maybe my comment is a little confusing. But eye 0 is always rendered to the left side of the screen, and eye 1 is rendered to the right side:

```
0 | 1
```

What this comment alludes to, is the fact that to render [cross-eyed stereo](https://www.kula3d.com/how-to-use-the-cross-eyed-method), the left and right eyes need to be flipped.

PR #7358, opened 2023-11-13
Enable optimizations for ESSL 1.0 code

elizagamedev:
The CL introducing the ESSL 1.0 chunk in materials inadvertently disabled optimizations for said code. This commit reintroduces those optimizations and fixes associated bugs which manifested. In particular, spirv-cross was generating uints for bools; this has been fixed with a hack. Additionally, spirv-cross is now compiled with exceptions enabled so that matc can gracefully fail and show the code which failed to compile rather than abruptly aborting.

[Upstream issue for spirv-cross.](https://github.com/KhronosGroup/SPIRV-Cross/issues/2223)

[Upstream PR for spirv-cross.](https://github.com/KhronosGroup/SPIRV-Cross/pull/2227)

### DIFF
--- a/libs/filamat/src/shaders/CodeGenerator.cpp
+++ b/libs/filamat/src/shaders/CodeGenerator.cpp
@@ -301,8 +301,9 @@ utils::io::sstream& CodeGenerator::generateProlog(utils::io::sstream& out, Shade
     out << '\n';
     out << SHADERS_COMMON_DEFINES_GLSL_DATA;
 
-    if (mFeatureLevel > FeatureLevel::FEATURE_LEVEL_0 &&
-            material.featureLevel == FeatureLevel::FEATURE_LEVEL_0) {
+    if (material.featureLevel == FeatureLevel::FEATURE_LEVEL_0 &&
+            (mFeatureLevel > FeatureLevel::FEATURE_LEVEL_0
+                    || mTargetLanguage == TargetLanguage::SPIRV)) {
         // Insert compatibility definitions for ESSL 1.0 functions which were removed in ESSL 3.0.
 
         // This is the minimum required value according to the OpenGL ES Shading Language Version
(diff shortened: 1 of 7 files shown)


filament/src/materials/blitLow.mat
```diff
@@ -35,7 +35,7 @@ vertex {
 
 fragment {
     void postProcess(inout PostProcessInputs postProcess) {
-#if __VERSION__ == 100
+#if EFFECTIVE_VERSION == 100
```
pixelflinger, line 38: maybe rename to `FILAMENT_EFFECTIVE_VERSION` we have been pretty bad with namespacing, but I've been trying to remedy it.

libs/filamat/src/shaders/CodeGenerator.cpp
```diff
@@ -151,6 +151,36 @@ utils::io::sstream& CodeGenerator::generateProlog(utils::io::sstream& out, Shade
...
+                    effective_version = "450";
+                } else {
+                    effective_version = "410";
+                }
+                break;
+            default:
+                assert(false);
+        }
+    }
+    generateDefine(out, "EFFECTIVE_VERSION", effective_version);
```
bejado, line 182: Prefix with `FILAMENT_`?

third_party/spirv-cross/spirv_glsl.cpp
```diff
@@ -14977,7 +14977,11 @@ string CompilerGLSL::flags_to_qualifiers_glsl(const SPIRType &type, const Bitset
 	{
 		auto &execution = get_entry_point();
 
-		if (flags.get(DecorationRelaxedPrecision))
+		if (type.basetype == SPIRType::UInt && is_legacy()) {
```
bejado, line 14980: Are these changes submitted to spirv-cross upstream? Otherwise, we'll lose these patches when building inside of G3

libs/filamat/src/shaders/CodeGenerator.cpp
```diff
@@ -460,19 +491,22 @@ io::sstream& CodeGenerator::generateOutput(io::sstream& out, ShaderStage type,
     const char* materialTypeString = getOutputTypeName(materialOutputType);
     const char* typeString = getOutputTypeName(outputType);
 
+    bool generate_essl3_code = mTargetLanguage == TargetLanguage::SPIRV
```
pixelflinger, line 494: why can't we generate essl1 with target lang spirv?

pixelflinger (changes requested):

third_party/spirv-cross/spirv_glsl.cpp
elizagamedev: Submitted an upstream PR, but I'm not sure they'll like it since it doesn't fix the root of the problem.

libs/filamat/src/shaders/CodeGenerator.cpp
elizagamedev: I described it a bit above in a comment elsewhere in this file:

    // During compilation and optimization, __VERSION__ reflects the shader language version of the
    // intermediate code, not the version of the final code. spirv-cross automatically adapts
    // certain language features (e.g. fragment output) but leaves others untouched (e.g. sampler
    // functions, bit shift operations). Client code may have to make decisions based on this
    // information, so define a FILAMENT_EFFECTIVE_VERSION constant.

Since the intermediate code is always ESSL 3.0 even if it eventually becomes ESSL 1.0, we have to sometimes generate ESSL 3.0 stuff like this.

pixelflinger (approved):

PR #7359, opened 2023-11-13
Fix some ES2 issues

elizagamedev:
Fix specification of mipmaps when generating textures.

Fix oversight where emulated UBOs would not replace uniforms when swapped out for another.

### DIFF
--- a/filament/backend/src/opengl/OpenGLDriver.cpp
+++ b/filament/backend/src/opengl/OpenGLDriver.cpp
@@ -172,7 +172,7 @@ OpenGLDriver::OpenGLDriver(OpenGLPlatform* platform, const Platform::DriverConfi
           mShaderCompilerService(*this),
           mHandleAllocator("Handles", driverConfig.handleArenaSize),
           mSamplerMap(32) {
-  
+
     std::fill(mSamplerBindings.begin(), mSamplerBindings.end(), nullptr);
 
     // set a reasonable default value for our stream array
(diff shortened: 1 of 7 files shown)


bejado (approved):

filament/backend/src/opengl/OpenGLDriver.cpp
```diff
@@ -595,7 +596,7 @@ void OpenGLDriver::textureStorage(OpenGLDriver::GLTexture* t,
                         }
                     } else {
                         glTexImage2D(t->gl.target, level, GLint(t->gl.internalFormat),
-                                GLsizei(width), GLsizei(height), 0,
+                                GLsizei(width >> level), GLsizei(height >> level), 0,
```
pixelflinger, line 599: you need to `max(1, x)` here so that the width or height can't end-up being 0

filament/backend/src/opengl/OpenGLDriver.h
```diff
@@ -363,7 +361,8 @@ class OpenGLDriver final : public DriverBase {
     void setScissor(Viewport const& scissor) noexcept;
 
     // ES2 only. Uniform buffer emulation binding points
-    std::array<std::pair<void const*, uint16_t>, Program::UNIFORM_BINDING_COUNT> mUniformBindings = {};
+    GLuint lastAssignedEmulatedUboId = 0;
```
pixelflinger, line 364: `m` prefix: `mLastAssignedEmulatedUboId`

filament/backend/src/opengl/OpenGLDriver.h
```diff
@@ -83,11 +83,9 @@ class OpenGLDriver final : public DriverBase {
         }
 
         struct {
+            GLuint id;
```
pixelflinger, line 86: That's very unfortunate, it'll double the size of this structure for the regular case. I can't think of a better way though.

pixelflinger (changes requested):

pixelflinger (approved):

elizagamedev:
I made a slight change. I added a fix to depth_main.fs to remove the ESSL 3.0+ specific syntax with ESSL 1.0 compatible syntax which was in #7358 . That PR depends on upstream changes so I agreed with @poweifeng to delay it for now, but this fix is still critical for FL0 to function.

PR #7366, opened 2023-11-15
Add FILAMENT_ENABLE_FEATURE_LEVEL_0 option

elizagamedev:
This allows clients to selectively exclude feature level 0 support.


poweifeng (approved):

elizagamedev:
Further comment on the stuff added to CMakeLists.txt: one of those `foreach` blocks has been resurrected and tweaked from a previous commit to target `FILAMENT_ENABLE_FEATURE_LEVEL_0` instead of `IS_MOBILE_TARGET` like it traditionally had. The others are hacks which will be eliminated following a change to matc to give it the ability to validate against FL0 without actually building ESSL 1.0 shaders, and as such they'll be removed along with the duplicated materials by the time the next release rolls around.

bejado (approved):

Issue #7386, opened 2023-11-28
GCC Compilation Errors


errissa:
**Describe the bug**
As described in #4990 there are some using declarations that trigger a GCC error (-fpermissive). The PR that closed #4990 addressed some of the issues but compilation errors remain in `filament/include/filament/View.h` and `filament/include/filament/Material.h`. 

**To Reproduce**
The following minimal program should trigger the error when compiled with GCC 9.4.
```
#include <filament/Material.h>
#include <filament/View.h>

int main() {
    return 0;
}
```

**Expected behavior**
Program that includes the referenced Filament headers compiles successfully with gcc 9.4.

**Screenshots**
_NA_

**Logs**
_NA_

**Desktop (please complete the following information):**
 - OS: Ubuntu 20.04
 - GPU: NVIDIA GeForce RTX 3080 Ti Laptop GPU/PCIe/SSE2
 - Backend: OpenGL and Vulkan

**Smartphone (please complete the following information):**
_NA_

**Additional context**

romainguy:
Unfortunately supporting GCC isn't one of our priorities.

pixelflinger:
So this is interesting. According to GCC our code is actually wrong, but clang doesn't catch this case, meaning even if we fix it, we are bound to recreate that problem down the line.

gcc will only emit a warning when using `-fpermissive` is that not a solution? gcc13 has a new `-Wchanges-meaning` which can be used instead of `-fpermissive`

rsepulveda-biocorellc:
@pixelflinger Thanks for accepting and merging #7387. For some `-fpermissive` could be a solution, but for Open3D we build with `-Werror` so we'd still get a compilation failure with `-fpermissive`.

errissa:
@pixelflinger Thanks for accepting and merging #7387. For some, `-fpermissive` would get prevent the compilation failure but we build Open3D with `-Werror` so we'd still get a compilation error.

PR #7387, opened 2023-11-28
Fix compilation errors when compiling with gcc on Linux

errissa:
Addresses issue #7386.

### DIFF
--- a/filament/include/filament/Material.h
+++ b/filament/include/filament/Material.h
@@ -50,11 +50,11 @@ class UTILS_PUBLIC Material : public FilamentAPI {
     struct BuilderDetails;
 
 public:
-    using BlendingMode = BlendingMode;
-    using Shading = Shading;
-    using Interpolation = Interpolation;
-    using VertexDomain = VertexDomain;
-    using TransparencyMode = TransparencyMode;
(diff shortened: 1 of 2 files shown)


errissa:
> lgtm but this seems weird. is that a compiler bug?

I have tried to figure that out but I can't tell. Here's a sample error message:
```
error: declaration of ‘using QualityLevel = enum class filament::QualityLevel’ changes meaning of ‘QualityLevel’ [-fpermissive]
```
According to the gcc documentation the `-fpermissive` option does the following:
```
-fpermissive
    Downgrade some diagnostics about nonconformant code from errors to warnings. Thus, using -fpermissive will allow some nonconforming code to compile.
```
So, gcc thinks the using declarations I changed are "non-conforming." Are they really and `clang` is allowing them? I'm not enough of a C++ spec nerd to answer that definitively. I can say, however, that `QualityLevel` is defined twice - once in `Options.h` and once in `ColorGrading.h`. Perhaps that creates enough ambiguity to make `using QualityLevel = QualityLevel` "non-conforming."

pixelflinger:
@errissa more info here #7386

PR #7389, opened 2023-11-28
better handle invalid programs in release builds

pixelflinger:
Until now invalid program would basically be undefined behavior, which in practice was a crash via a null pointer dereference. With this change, invalid programs cause drawing ops to become no-ops.

Additionally fixed an unsynchronized access of a the variable containing the program id. I don't think it would have caused issues though.

FIXES=[311775564]

### DIFF
--- a/filament/backend/src/opengl/OpenGLDriver.cpp
+++ b/filament/backend/src/opengl/OpenGLDriver.cpp
@@ -3547,7 +3547,7 @@ void OpenGLDriver::draw(PipelineState state, Handle<HwRenderPrimitive> rph, uint
         return;
     }
 
-    GLRenderPrimitive* const rp = handle_cast<GLRenderPrimitive *>(rph);
+    GLRenderPrimitive* const rp = handle_cast<GLRenderPrimitive*>(rph);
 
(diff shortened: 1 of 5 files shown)


filament/backend/src/opengl/OpenGLDriver.cpp
```diff
@@ -3531,18 +3537,17 @@ void OpenGLDriver::draw(PipelineState state, Handle<HwRenderPrimitive> rph, uint
...
+        // Avoid fatal (or cascading) errors that can occur during the draw call when the program
+        // is invalid. The shader compile error has already been dumped to the console at this
+        // point, so it's fine to simply return early.
         return;
     }
 
-    useProgram(p);
-
-    GLRenderPrimitive* rp = handle_cast<GLRenderPrimitive *>(rph);
+    GLRenderPrimitive* const rp = handle_cast<GLRenderPrimitive *>(rph);
```
poweifeng, line 3550:
```suggestion
    GLRenderPrimitive* const rp = handle_cast<GLRenderPrimitive*>(rph);
```

poweifeng (approved):

PR #7396, opened 2023-11-30
Fix ESSL 1.0 external samplers

elizagamedev:
matc was re-introducing the incorrect extension for external samplers in ESSL 1.0 code after spirv optimizations.


pixelflinger (approved):
good one!

PR #7399, opened 2023-11-30
vk: spec const workaround through runtime spirv modifications

poweifeng:
In cases where the driver behaves poorly with specialization const, we manually change the spec const into a constant.

FIXES=310603393

### DIFF
--- a/NEW_RELEASE_NOTES.md
+++ b/NEW_RELEASE_NOTES.md
@@ -8,4 +8,3 @@ appropriate header in [RELEASE_NOTES.md](./RELEASE_NOTES.md).
 
 ## Release notes for next branch cut
 
-- matc: Fix ESSL 1.0 codegen when using external samplers [⚠️ **Recompile materials**]
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ repositories {
 }
 
 dependencies {
-    implementation 'com.google.android.filament:filament-android:1.47.0'
+    implementation 'com.google.android.filament:filament-android:1.48.0'
 }
 ```
 
@@ -51,7 +51,7 @@ Here are all the libraries available in the group `com.google.android.filament`:
 iOS projects can use CocoaPods to install the latest release:
 
 ```shell
-pod 'Filament', '~> 1.47.0'
+pod 'Filament', '~> 1.48.0'
 ```
 
 ### Snapshots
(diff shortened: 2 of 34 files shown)


filament/backend/CMakeLists.txt
```diff
@@ -313,6 +315,7 @@ endif()
 
 if (FILAMENT_SUPPORTS_VULKAN)
     target_link_libraries(${TARGET} PUBLIC bluevk vkmemalloc vkshaders smol-v)
+    target_link_libraries(${TARGET} PRIVATE SPIRV-Tools)
```
pixelflinger, line 318: Oh I don't think we can do that, at least not in release. This will add tons of code to filament.

pixelflinger (changes requested):
I love the idea and amazed it works, however we need a version that works on binary.

filament/backend/CMakeLists.txt
poweifeng: ok, noted.  I'll revise.

poweifeng: done

filament/backend/src/vulkan/platform/VulkanPlatform.cpp
```diff
@@ -331,6 +331,24 @@ VkDevice createLogicalDevice(VkPhysicalDevice physicalDevice,
...
+            .sType = VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_DRIVER_PROPERTIES,
+    };
+    VkPhysicalDeviceProperties2 physicalDeviceProperties2 = {
+            .sType = VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_PROPERTIES_2,
+            .pNext = &driverProperties,
+    };
+    vkGetPhysicalDeviceProperties2KHR(device, &physicalDeviceProperties2);
+    std::string_view driverName { driverProperties.driverName };
+    if (driverName.find("AMDVLK for Stadia") != std::string_view::npos) {
+        workarounds.convertSpecConstToConst = true;
```
pixelflinger, line 346: why not always do this? this would solve the problem where we cannot use spec constants for sizing ubos.

poweifeng: I have not tested this with other drivers, only desktop nvidia and AMDVLK so far.  
IIRC sizing ubos was a pixel4 (i.e. adreno) issue?

poweifeng: I've changed it to always do it. PTAL.

pixelflinger: no sizing UBOs is actually  a SPIRV limitation!

pixelflinger (approved):

PR #7401, opened 2023-12-02
Add spirv-headers as a separate third_party repo

poweifeng:
Previously, we pulled in spirv-headers as part of spirv-tools. We still keep this behavior but move the content of the repo to third_party.

We introduce a script for updating spirv-tools, update the patch file. The same script will also pull in a dependent spirv-headers.


pixelflinger (approved):

third_party/spirv-tools/filament-update.sh
```diff
@@ -0,0 +1,51 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+# This script is used to update SPIRV-Tools from source.
+# This script takes in a git commit hash as an argument.
```
bejado, line 16: Nit: add shebang
```
#!/usr/bin/env bash
```

bejado (approved):

third_party/spirv-tools/filament-update.sh
poweifeng: done

PR #7405, opened 2023-12-05
on ANGLE we now use a thread pool for parallel shader compilation

pixelflinger:
in general we now prefer using a thread pool instead of the KHR extension, because we have less control over how the queue is managed by the driver.

ANGLE supports many threads very well, so we use cpu_threads/2 for the pool size, at background priority.

### DIFF
--- a/filament/backend/src/opengl/ShaderCompilerService.h
+++ b/filament/backend/src/opengl/ShaderCompilerService.h
@@ -95,19 +95,42 @@ public:
             CallbackHandler* handler, CallbackHandler::Callback callback, void* user);
 
 private:
+    struct Job {
+        template<typename FUNC>
+        Job(FUNC&& fn) : fn(std::forward<FUNC>(fn)) {}
+        Job(std::function<bool(Job const& job)> fn,
+                CallbackHandler* handler, void* user, CallbackHandler::Callback callback)
+                : fn(std::move(fn)), handler(handler), user(user), callback(callback) {
+        }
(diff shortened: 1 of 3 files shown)


filament/backend/src/opengl/ShaderCompilerService.h
```diff
@@ -95,19 +95,42 @@ class ShaderCompilerService {
...
+        }
+        std::function<bool(Job const& job)> fn;
+        CallbackHandler* handler = nullptr;
+        void* user = nullptr;
+        CallbackHandler::Callback callback{};
+    };
+
+    enum class Mode {
+        UNDEFINED,      // init() has not beed called yet.
+        SYNCHRONOUS,    // synchrnous shader compilation
```
bejado, line 113:
```suggestion
        SYNCHRONOUS,    // synchronous shader compilation
```

filament/backend/src/opengl/ShaderCompilerService.h
```diff
@@ -95,19 +95,42 @@ class ShaderCompilerService {
...
+        THREAD_POOL,    // asynchronous shader compilation using a thread-pool (most common)
+        ASYNCHRONOUS    // asynchrnous shader compilation using KHR_parallel_shader_compile
```
bejado, line 115:
```suggestion
        ASYNCHRONOUS    // asynchronous shader compilation using KHR_parallel_shader_compile
```

bejado (approved):

PR #7408, opened 2023-12-06
Added an option to shorten compile time in Visual Studio

mdagois:
The option is `FILAMENT_SHORTEN_MSVC_COMPILATION` and is ON by default (except for CI).
It enables multi-processor compilation and removes run-time STL checks to improve the performance of tools that runs at compile time (e.g. `matc`).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -231,6 +231,16 @@ if (WIN32)
             string(REPLACE "/INCREMENTAL" "/INCREMENTAL:NO" ${LinkerFlag} ${${LinkerFlag}})
         endforeach()
     endif()
+
+    if (MSVC)
+        option(FILAMENT_SHORTEN_MSVC_COMPILATION "Shorten compile-time in Visual Studio" ON)
+        if (FILAMENT_SHORTEN_MSVC_COMPILATION)
+            # enable multi-processor compilation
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -231,6 +231,16 @@ if (WIN32)
             string(REPLACE "/INCREMENTAL" "/INCREMENTAL:NO" ${LinkerFlag} ${${LinkerFlag}})
         endforeach()
     endif()
+
+    if (MSVC)
+        option(FILAMENT_SHORTEN_MSVC_COMPILATION "Shorten compile-time in Visual Studio" ON)
```
poweifeng, line 236: would it make sense to always do this for MSVC? In which case we wouldn't need the extra option.

bejado, line 236: I think we need to do one of two things: turn this `OFF` by default, or turn it `OFF` for CI builds (see [build-github.bat](https://github.com/google/filament/blob/main/build/windows/build-github.bat) and the `FILAMENT_WINDOWS_CI_BUILD` flag).

IIRC, we used to have `/MP` turned on for CI builds, but it led to instability due to limited memory on CI machines.

pixelflinger: Can we a a set of flags for CI  and one for local builds. This way, when/if we get better machines for CI et can easily add the option?

pixelflinger (approved):

PR #7419, opened 2023-12-11
Fixed feature level 0 material build rules

mdagois:
Changed the feature level 0 materials custom commands so that materials for all APIs are built.
This allows the following:
* Samples using the default material (e.g. hellotriangle) can be run in Vulkan (and potentially other APIs that are not OpenGL).
* The OpenGL backend can be disabled using `-DFILAMENT_SUPPORTS_OPENGL=OFF`. Previously, turning off the OpenGL backend was leading to a compilation error related to missing level 0 materials.


elizagamedev:
Looks good to me, thanks for the contribution!

PR #7428, opened 2023-12-13
Fix UiHelper

### LINKED ISSUE #7424: Android: Models disappear on 1.49.0
**Describe the bug**
With new version 1.49.0 for Android the 3D models disappear after onPause - onResume. 
I am using compose, inspired by https://gist.github.com/romainguy/b3e283aefc5a5590073764d2dc0bd797 
Everything works well but once I lock the phone and open it again, the 3D models disappear and not load again.

This was working well before with 1.45.1 where the models would be again visible.

Thank you!

**To Reproduce**
1. Use Android app similar to https://gist.github.com/romainguy/b3e283aefc5a5590073764d2dc0bd797
2. Lock the phone and unlock it
3. 3D models dissapear

**Expected behavior**
Models should stay on screen as before.

**Screenshots**
NA

**Logs**
NA

 - Device: Xiaomi Poco F3
 - OS: Android 13

**Additional context**
I see an important improvement in the speed on loading the models with 1.49.0 compared to 1.45.1 but I cant figure out why they disappear with latest changes.

pixelflinger:
We shouldn't remove the listener callbacks when a surface is destroyed by the system (e.g. screen off/on) because then we won't know when it comes back.

But we still need to do this when the user calls UiHelper.detach() or when we are attaching to a new surface.

Fixes #7424

### DIFF
--- a/android/filament-android/src/main/java/com/google/android/filament/android/UiHelper.java
+++ b/android/filament-android/src/main/java/com/google/android/filament/android/UiHelper.java
@@ -329,7 +329,6 @@ public class UiHelper {
         @Override
         public void detach() {
             mTextureView.setSurfaceTextureListener(null);
-            setSurface(null);
         }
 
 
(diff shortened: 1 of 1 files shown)


romainguy:
LGTM

poweifeng (approved):

bejado (approved):

PR #7431, opened 2023-12-14
add a way to set a log consumer 

pixelflinger:
this is a private API to capture filament's logs.

### DIFF
--- a/libs/utils/src/ostream.cpp
+++ b/libs/utils/src/ostream.cpp
@@ -24,9 +24,12 @@
 #include <utils/PrivateImplementation-impl.h>
 
 #include <algorithm>
+#include <mutex>
 #include <string>
 #include <string_view>
+#include <utility>
 
+#include <stdlib.h>
 #include <stdarg.h>
 
 template class utils::PrivateImplementation<utils::io::ostream_>;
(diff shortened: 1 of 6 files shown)


poweifeng (approved):

libs/utils/src/ostream.cpp
```diff
@@ -35,6 +38,31 @@ namespace utils::io {
...
+    pImpl->mConsumer = { consumer, user };
+}
+
+ostream& flush(ostream& s) noexcept {
+    auto* const pImpl = s.mImpl;
+    pImpl->mLock.lock();
+    auto const callback = pImpl->mConsumer;
+    if (UTILS_UNLIKELY(callback.first)) {
+        auto& buf = s.getBuffer();
+        char* const str = strdup(buf.get());
```
selcott, line 53: Can buf.get() be null?

pixelflinger: ah good remark. I think it can technically be `null` if the log has never been written to, should be rare, but still.

pixelflinger: okay, fixed a couple nullptr issues.

PR #7436, opened 2023-12-15
don't use eglGetProcAddress on emscripten

pixelflinger:
In practice none of the extensions we would use eglGetProcAddress for are supported. And we had a case where an extension was reported supported but eglGetProcAddress didn't return the corresponding entry point.

update web demos remote ui.

FIXES=[315033914]


kainino0x:
FYI I think I forgot to mention this on the bug - Emscripten changed to disable GetProcAddress by default in [3.1.51](https://github.com/emscripten-core/emscripten/releases/tag/3.1.51):
https://github.com/emscripten-core/emscripten/pull/20802

romainguy (approved):

kainino0x:
> FYI I think I forgot to mention this on the bug - Emscripten changed to disable GetProcAddress by default in [3.1.51](https://github.com/emscripten-core/emscripten/releases/tag/3.1.51): [emscripten-core/emscripten#20802](https://github.com/emscripten-core/emscripten/pull/20802)

It's going to change back to enabled-by-default, at least for now, to prevent breaking people's builds.

PR #7440, opened 2023-12-18
Make`applyMaterialVariant` Method Public in FilamentInstance Class

princeku07:
The current implementation restricts access to this method, and making it public will enable external classes or components to utilise this functionality


romainguy (approved):

romainguy:
Thanks for your contribution, could you please sign the CLA so we can accept your patch?

PR #7448, opened 2023-12-21
Basic support for MS samplers and vk/metal depth resolve update

pixelflinger:
This PR adds basic support for multi sampled samplers in user materials. This is currently not advertised because this is still a bit work-in-progress, but is enough to fix an issue with depth resolve (see below).

In theory, multisample samplers should only be available at feature-level 2+, but we currently don't enforce it, and it turns out that glslang will happily compile `sampler2DMS` with ESSL 300. This actually serves us because we now *need* this for both vulkan and metal at all feature levels, but Vulkan can theoretically return feature-level 1 (I think we should fix this and then enable feature level 2 in the resolve material).

We also now do the depth resolve from the filament side instead of the backend, if the backend doesn't support it. Currently both metal and vulkan backends don't.

This PR actually fixes an issue with depth resolve on Vulkan which didn't work, the problem could be visible with MSAA+TAA (the history reprojection didn't work). This removes a fair amount of code in the vk/mtl backends.


bejado:
Metal changes LGTM

PR #7457, opened 2023-12-26
Fix a couple race conditions in Metal backend

bejado:
1. MetalFence is constructed from two threads:
- the main thread, from `MetalDriver::createFenceS`
- the driver thread, from `MetalTimerQueryFence::beginTimeElapsedQuery`

Because the `MetalFence` constructor reads and writes to `context.signalId`, it needs to be made atomic.

2. `MetalBufferPool::mCurrentframe` needs to be made atomic, as it is accessed by two threads:
- the driver thread, from `MetalBufferPool::gc`, read/writes to `mCurrentFrame`
- a completion handler thread, from `MetalBufferPool::releaseBuffer`, reads `mCurrentFrame`

This also adds support for a new CMake flag to enable TSAN.

b/314267870
b/314265331

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -41,6 +41,8 @@ option(FILAMENT_LINUX_IS_MOBILE "Treat Linux as Mobile" OFF)
 
 option(FILAMENT_ENABLE_ASAN_UBSAN "Enable Address and Undefined Behavior Sanitizers" OFF)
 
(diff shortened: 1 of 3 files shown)


elizagamedev:
LGTM, maybe copy some of the context of the PR description to a comment beside the variable though

elizagamedev (approved):

PR #7475, opened 2024-01-05
Generate dummy stereo variants for FL0 mats

elizagamedev:
See #7415 for a more detailed description of why this change is necessary.

The remaining variants which are filtered from FL0 materials are all related to lighting, so further hacks like this won't be necessary.

Future work involves properly supporting differing sets of variants based on shader language.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -9,6 +9,7 @@ Instead, if you are authoring a PR for the main branch, add your release note to
 
 ## v1.49.3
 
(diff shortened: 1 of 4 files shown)


bejado (approved):
Does this need to go out with the 1.49.3 release on Monday? If so, we can merge directly into rc/1.49.3 (then cherry-pick to main) and add the release note directly to the RELEASE_NOTES.md file.

elizagamedev:
Alright, made the necessary changes; is this still good to go?

bejado:
Looks good. @poweifeng FYI

PR #7479, opened 2024-01-09
bloom: disable fireflies reduction when using TAA

pixelflinger:
TAA already does a fireflies reduction pass, so it's not needed when applying bloom.


romainguy (approved):
I assume all the #includes were missing?

PR #7483, opened 2024-01-09
geometry: allow additional attributes in TangentSpaceMesh

poweifeng:
- Add methods for adding attributes to the input mesh
     - Add method in TangentSpaceMesh for when user provides the
       tangents
     - Separate client-side Algorithm enum from implementation algorithm
       (AlgorithmImpl)
     - Fix CMake config for combining static libs


libs/geometry/include/geometry/TangentSpaceMesh.h
```diff
@@ -106,16 +108,24 @@ class TangentSpaceMesh {
...
-         * Flat Shading
-         *
-         * **Requires**: `positions + indices` <br/>
-         * **Note**: Will remesh
-         */
-        FLAT_SHADING = 5
+    /**
+     * This enum specifies the auxiliary attributes of each vertex that can be provided as input.
+     * These attributes do not affect the computation of the tangent space, but they will properly
+     * mapped when a remeshing is carried out.
```
bejado, line 116:
```suggestion
     * This enum specifies the auxiliary attributes of each vertex that can be provided as input.
     * These attributes do not affect the computation of the tangent space, but they will be
     * properly mapped when a remeshing is carried out.
```

libs/geometry/include/geometry/TangentSpaceMesh.h
```diff
@@ -284,6 +324,66 @@ class TangentSpaceMesh {
...
+    /**
+     * Get output auxiliary attributes.
+     * Assumes the `out` param is at least of getVertexCount() length (while accounting for
+     * `stride`).
+     *
+     * @param  out    Client-allocated array that will be used for copying out attribute as two
+     * floats.
+     * @param  stride Stride for iterating through `out`
+     */
+    void getAux(AuxAttribute attribute, filament::math::float2* out,
```
bejado, line 336: Could we make `getAux` a template method and pre-instantiate all the specializations?

libs/geometry/src/TangentSpaceMeshInternal.h
```diff
@@ -50,57 +117,258 @@ class InternalArray {
...
+
+    // Pass back the std::variant instead of the content.
+    InData data(AttributeImpl attribute) const {
+        auto res = attributeData.find(attribute);
+        assert_invariant(res != attributeData.end());
+        return res->second.data;
+    }
+
+    template<typename DataType>
+    inline DataType const* aux(AttributeImpl attribute) const {
```
bejado, line 261: It was a bit confusing to me that this `aux` method can access both auxiliary data _and_ non-auxiliary data.

libs/geometry/include/geometry/TangentSpaceMesh.h
```diff
@@ -170,17 +180,47 @@ class TangentSpaceMesh {
...
 
+        /**
+         * Sets "auxiliary" attributes that will be properly mapped when remeshed.
+         *
+         * @param attribute The attribute of the data to be stored
+         * @param data The data to be store
+         * @param stride The stride for iterating through `attribute`
+         * @return Builder
+         */
+        Builder& aux(AuxAttribute attribute, InData data, size_t stride = 0) noexcept;
```
bejado, line 191: What if I mismatch the auxiliary attribute type with a pointer to the wrong data type?

```
filament::math::float3 uvs;
...
TangentSpaceMesh::Builder()
    .aux(AuxAttribute::UV1, &uvs)
```

I think this will still compile and run without error, but inside `TangetSpaceMeshOutput::passthrough`, we assume a different type (UV1 should be `float2`).

libs/geometry/src/TangentSpaceMeshInternal.h
poweifeng: Sorry, renamed to data() for consistency

libs/geometry/include/geometry/TangentSpaceMesh.h
poweifeng: Good point.  I added an assert in `build()` to validate the input type of the attributes.

libs/geometry/include/geometry/TangentSpaceMesh.h
poweifeng: Done

libs/geometry/include/geometry/TangentSpaceMesh.h
poweifeng: Not without exposing implementation details unfortunately.  Or can you think of a way around that?

bejado: You can do something like this:

```cpp
// In TangentSpaceMesh.h:
template<typename T>
void getAux(AuxAttribute attribute, T* out, size_t stride = 0) const noexcept;
```

```cpp
// In TangentSpaceMesh.cpp:
template void TangentSpaceMesh::getAux<filament::math::float2>(AuxAttribute attribute, filament::math::float2* out, size_t stride) const noexcept;
template void TangentSpaceMesh::getAux<filament::math::float3>(AuxAttribute attribute, filament::math::float3* out, size_t stride) const noexcept;
// etc

template<typename T>
void TangentSpaceMesh::getAux(AuxAttribute attribute, T* out, size_t stride) const noexcept {
    // ...
}
```

We do something similar inside [MaterialInstance.cpp](https://github.com/google/filament/blob/fc07e79c0d0882300634d3143488f4caa157ae5e/filament/src/MaterialInstance.cpp#L109-L123) for `setParameter`.

poweifeng: nice! done

bejado (approved):

PR #7495, opened 2024-01-16
Remove spirv-opt size optimizations for MSL

bejado:
For generating MSL, turning off `--optimize-size` optimization passes (still keeping the ones necessary for fp16), reduces sizes of lit materials by nearly 400 KiB:
- sandboxLit.mat: 376.6 KiB savings
- gltfio lit_opaque.mat: 395.7 KiB savings
- bakedColor.mat: 1.799 KiB increase

Note there is a slight increase with unlit materials, as seen above with bakedColor.mat.

Removing these optimizations helps the line dictionary compress better. It's likely the Metal compiler itself does a lot of these optimizations.

An added benefit is this makes MSL shaders much easier to debug.

### DIFF
--- a/libs/filamat/src/GLSLPostProcessor.cpp
+++ b/libs/filamat/src/GLSLPostProcessor.cpp
@@ -589,7 +589,13 @@ std::shared_ptr<spvtools::Optimizer> GLSLPostProcessor::createOptimizer(
     });
 
     if (optimization == MaterialBuilder::Optimization::SIZE) {
-        registerSizePasses(*optimizer, config);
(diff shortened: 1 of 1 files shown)


pixelflinger:
we're not taking the same path that with glsl (i.e. not calling size optimization at all) because, I'm assuming, we need the fp16 transformation?  I think there should be a comment explaining why this is different from glsl.

bejado:
Putting this on hold for a bit because I found a bug in spirv-opt that causes the SPIR-V to fail validation if we run it through the fp16 transformation pass: https://github.com/KhronosGroup/SPIRV-Tools/pull/5536

PR #7506, opened 2024-01-18
fix IBLPrefilterContext::EquirectangularToCubemap

pixelflinger:
it was incorrectly mapping the equirect image to a cubemap due to a typo in our overload of atan2 which was swapping its parameters.

atan2 is now removed, and we use atan(y,x) instead. Also modified the code slightly so it matches almost exactly cmgen's.

FIXES=[320856413]

### DIFF
--- a/libs/iblprefilter/src/materials/equirectToCube.mat
+++ b/libs/iblprefilter/src/materials/equirectToCube.mat
@@ -48,7 +48,7 @@ fragment {
 void dummy() {}
 
 highp vec2 toEquirect(const highp vec3 s) {
-    highp float xf = atan2(s.x, s.z) * (1.0 / PI);  // range [-1.0, 1.0]
+    highp float xf = atan(s.x, s.z) * (1.0 / PI);   // range [-1.0, 1.0]
     highp float yf = asin(s.y) * (2.0 / PI);        // range [-1.0, 1.0]
     xf = (xf + 1.0) * 0.5;                          // range [0, 1.0]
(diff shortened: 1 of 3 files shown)


bejado (approved):

shaders/src/common_math.glsl
```diff
@@ -19,7 +19,6 @@
 #endif
 
 #define saturate(x)        clamp(x, 0.0, 1.0)
-#define atan2(x, y)        atan(y, x)
```
romainguy, line 22: Afaict there's no typo here, it matches HLSL's atan2 behavior which is atan2(x, y) == GLSL's atan(y, x)

romainguy (approved):

libs/iblprefilter/src/materials/equirectToCube.mat
```diff
@@ -62,17 +62,22 @@ mediump vec3 sampleEquirect(mediump sampler2D equirect, const highp vec3 r) {
 
 void postProcess(inout PostProcessInputs postProcess) {
     highp vec2 uv = variable_vertex.xy; // interpolated at pixel's center
-    highp vec2 p = uv * 2.0 - 1.0;
+    highp vec2 p = vec2(
+        uv.x * 2.0 - 1.0,
+        1.0 - uv.y * 2.0);
+
     float side = materialParams.side;
+    float l = 1.0 / sqrt(p.x * p.x + p.y * p.y + 1.0);
```
romainguy, line 70: Use invsqrt() instead

PR #7511, opened 2024-01-20
Add CMake option FILAMENT_USE_MULTIVIEW_STEREO

z3moon:
Passing -DFILAMENT_USE_MULTIVIEW_STEREO allows the generated project files to contain the preprocessor FILAMENT_USE_MULTIVIEW_STEREO. This will be referenced to determine which implementation for stereo rendering should be used at compile time.

The value of this parameter defaults to NO, which sticks to the original behavior.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -23,6 +23,8 @@ option(FILAMENT_USE_EXTERNAL_GLES3 "Experimental: Compile Filament against OpenG
 
 option(FILAMENT_USE_SWIFTSHADER "Compile Filament against SwiftShader" OFF)
 
+option(FILAMENT_USE_MULTIVIEW_STEREO "Compile Filament against multiview extension for stereo rendering; This disables instanced stereo" OFF)
+
 option(FILAMENT_ENABLE_LTO "Enable link-time optimizations if supported by the compiler" OFF)
 
 option(FILAMENT_SKIP_SAMPLES "Don't build samples" OFF)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -23,6 +23,8 @@ option(FILAMENT_USE_EXTERNAL_GLES3 "Experimental: Compile Filament against OpenG
 
 option(FILAMENT_USE_SWIFTSHADER "Compile Filament against SwiftShader" OFF)
 
+option(FILAMENT_USE_MULTIVIEW_STEREO "Compile Filament against multiview extension for stereo rendering; This disables instanced stereo" OFF)
```
bejado, line 26: Let's clarify that this only affects which stereo mode our samples use (i.e., gltf_viewer). I think this should go in our samples CMakeLists.txt (samples/CMakeLists.txt).

Can we also make this a string instead of a bool? Maybe `FILAMENT_SAMPLES_STEREO_TYPE` and let it be either "instanced" or "multiview".

CMakeLists.txt
```diff
@@ -531,6 +533,11 @@ else()
     option(FILAMENT_DISABLE_MATOPT "Disable material optimizations" ON)
 endif()
 
+# Build Filament with multiview extension for stereo rendering. This disables the instanced stereo implementation.
+if (FILAMENT_USE_MULTIVIEW_STEREO)
+  add_definitions(-DFILAMENT_USE_MULTIVIEW_STEREO)
```
bejado, line 538: I don't think we need to add a C++ define for this. Filament itself shouldn't need to know which type of stereo our samples use.

bejado (changes requested):

PR #7515, opened 2024-01-23
Fix missing/extra includes in all public headers

pixelflinger:


bejado (approved):
LGTM, I created a CL to check that this doesn't break G3: cl/600628952. Let's make sure that passes first.

filament/include/filament/Box.h
```diff
@@ -21,10 +21,14 @@
...
-#include <limits>
-
+#include <math/mat3.h>
 #include <math/mat4.h>
 #include <math/vec3.h>
+#include <math/vec4.h>
+
+#include <float.h>
+
+#include <stddef.h>
```
poweifeng, line 31: Does Filament prefer c-style headers over c++ (i.e. <cstddef> as a convention?

poweifeng (approved):

z3moon (approved):

filament/include/filament/Box.h
pixelflinger: Yes. I'm trying to avoid the use of the STL in public headers as much as possible, so that we don't create unnecessary dependencies and reduce compilation times.

PR #7517, opened 2024-01-23
Shadowmap texture vizualizer

pixelflinger:
The vizualizer is implemented inside filament itself and activated using the debug registry. This intended for filament development use.
shadow-debug

### DIFF
--- a/filament/CMakeLists.txt
+++ b/filament/CMakeLists.txt
@@ -238,6 +238,7 @@ set(MATERIAL_SRCS
         src/materials/resolveDepth.mat
         src/materials/separableGaussianBlur.mat
         src/materials/skybox.mat
+        src/materials/shadowmap.mat
         src/materials/ssao/bilateralBlur.mat
         src/materials/ssao/bilateralBlurBentNormals.mat
         src/materials/ssao/mipmapDepth.mat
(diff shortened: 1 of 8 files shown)


bejado (approved):

filament/CMakeLists.txt
```diff
@@ -238,6 +238,7 @@ set(MATERIAL_SRCS
         src/materials/resolveDepth.mat
         src/materials/separableGaussianBlur.mat
         src/materials/skybox.mat
+        src/materials/shadowmap.mat
```
poweifeng, line 241: won't we be increasing the size of the lib for a debug feature?

poweifeng (approved):

filament/CMakeLists.txt
pixelflinger: yeah, a bit. I was wondering if it should be there in DEBUG only. not sure. I think it's probably small enough?

romainguy: You should check its size in the generated folders. mat files can quickly grow :/

pixelflinger: Yeah but this one has just one variant.

PR #7523, opened 2024-01-25
support execinfo.h on Android 33 and above

pixelflinger:
Note: this currently requires to compile filament with API_LEVEL 33.

### DIFF
--- a/libs/utils/src/CallStack.cpp
+++ b/libs/utils/src/CallStack.cpp
@@ -15,19 +15,31 @@
  */
 
 #include <utils/CallStack.h>
-#include <utils/Log.h>
+#include <utils/CString.h>
+#include <utils/compiler.h>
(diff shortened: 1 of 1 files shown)


libs/utils/src/CallStack.cpp
```diff
@@ -15,21 +15,36 @@
...
 
+// execinfo.h is available as of Android 33
+#if defined(__ANDROID__) && (__ANDROID_API__ >= 33)
+#include <execinfo.h>
+#undef HAS_EXECINFO
+#define HAS_EXECINFO 1
+#endif
+
+#if defined(UNIX)
+#endif
```
poweifeng, line 46: Is this suppose to be empty?

pixelflinger: oh good catch, forgot to remove it.

Issue #6188, opened 2022-10-14
Imported GL_TEXTURE_EXTERNAL_OES remains black on Mali


kpeeters:
Having converted my code away from the stream APIs which were removed in v1.21.3, in favour of `import`, I am now running into an issue with `GL_TEXTURE_EXTERNAL_OES` camera textures remaining black on Mali GPUs. This looks similar to #2831, but my texture is definitely `CLAMP_TO_EDGE` and in any case Filament is enforcing that now. It also looks similar to #1448, which I believe pointed at some `updateTexImage` missing for some people, but I am not sure whether that is the underlying cause here. Does `updateTexImage` need to be called on imported textures to make it conform to GL specifications?

Import of course works fine on other GPUs that I have been able to test against, and it also works on the emulator, but given the Mali market share, it would still be nice to have a workaround or fix.

kpeeters:
I do not know if it is related, but on the emulator, the `sample-hello-camera` shows a black camera feed for me as well.

pixelflinger:
Yes, Thank you this is useful.

In fact, we don't even know that the import() api ever worked on Mali, right?

kpeeters:
I have tried going back further but have not found a version that worked on Mali (though the search was far from complete).

kpeeters:
@pixelflinger Do you have any idea what may be the difference between the old method and the new import-based method, as far as this bug is concerned? I'd like to spend some time to investigate this, so if there's anything you have already figured out, please let me know.

pixelflinger:
@kpeeters Do you have a repro test case for this?

kpeeters:
No, not yet, it's all still part of a project that's too large as a test case. I can try to find some time to fit it onto `sample-hello-camera`; probably not bad to have as a test there anyway.

pixelflinger:
fixed in #7532

PR #7532, opened 2024-01-26
fix external images

pixelflinger:
It turns out that many drivers require to rebind the external texture after its content has changed, so we now always do that.

Bug #6188


romainguy (approved):
Wtf

kpeeters:
I will give this a spin asap, thanks a lot!

pixelflinger:
> I will give this a spin asap, thanks a lot!

This fixed the Pixel 8, Pixel 4 and emulator for me.

kpeeters:
Great stuff, all works fine now (will test some more devices but so far looking good). Thanks!

PR #7535, opened 2024-01-29
Metal: schedule PresentDrawable for destruction on the main thread

bejado:
Internal crash reports are showing crashes within `[CAMetalDrawable dealloc]`, which is currently being called on the driver thread (to match the same thread where it is acquired). It is suspected this is a poorly behaved object, which must be dealloced on the main thread.

This PR adds a new flag, `FILAMENT_RELEASE_PRESENT_DRAWABLE_MAIN_THREAD`, on by default, which changes the behavior to release the `CAMetalDrawable` on the main thread.

### DIFF
--- a/filament/backend/src/metal/MetalHandles.mm
+++ b/filament/backend/src/metal/MetalHandles.mm
@@ -245,30 +245,53 @@ void MetalSwapChain::present() {
     }
 }
 
-struct PresentDrawableData {
-    void* drawable = nullptr;
-    MetalDriver* driver = nullptr;
+#ifndef FILAMENT_RELEASE_PRESENT_DRAWABLE_MAIN_THREAD
+#define FILAMENT_RELEASE_PRESENT_DRAWABLE_MAIN_THREAD 1
+#endif
+
+class PresentDrawableData {
+public:
+    PresentDrawableData() = delete;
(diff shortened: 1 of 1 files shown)


poweifeng (approved):

filament/backend/src/metal/MetalHandles.mm
```diff
@@ -245,30 +245,53 @@ static inline MTLTextureUsage getMetalTextureUsage(TextureUsage usage) {
...
+    PresentDrawableData() = delete;
+    PresentDrawableData(const PresentDrawableData&) = delete;
+    PresentDrawableData& operator=(const PresentDrawableData&) = delete;
+
+    static PresentDrawableData* create(id<CAMetalDrawable> drawable, MetalDriver* driver) {
+        assert_invariant(driver);
+        return new PresentDrawableData(drawable, driver);
+    }
+
+    void maybePresentAndCleanupAsync(bool shouldPresent) {
```
pixelflinger, line 263: I would prefer this method to be static and take a PresentDrawableData* (or &), because when it returns the object is effectively destroyed. That would be clearer. Also I would rename to `maybePresentAndDestroyAsync`.

pixelflinger (changes requested):

pixelflinger (approved):

PR #7540, opened 2024-02-01
Improve memory allocations

pixelflinger:
This set of 4 CLs (probably better reviewed independently), improve several aspects of our Arena-based allocations:

- CommandStream is now automatically flushed when needed during command encoding
- CircularBuffer API and implementation is improved
- We fix a couple threading issues 
- RenderPass is rewritten so that it is now immutable and constructed from a builder
- Finally we fix the long standing bug where we would abort if running out of space in the high or low level command arena

Together this fixes crashes/aborts when running out of space in the per-frame Arena.

FIXES=[277115740]


bejado (approved):

filament/src/RenderPass.h
```diff
@@ -424,50 +406,129 @@ class RenderPass {
     static void updateSummedPrimitiveCounts(
             FScene::RenderableSoa& renderableData, utils::Range<uint32_t> vr) noexcept;
 
-    // a reference to the Engine, mostly to get to things like JobSystem
 
     // Arena where all Commands are allocated. The Arena owns the commands.
```
poweifeng, line 410:
```suggestion
```

libs/utils/include/utils/Allocator.h
```diff
@@ -229,35 +276,35 @@ class AtomicFreeList {
     AtomicFreeList& operator=(const FreeList& rhs) = delete;
 
     void* pop() noexcept {
-        Node* const storage = mStorage;
+        Node* const pStorage = mStorage;
 
         HeadPtr currentHead = mHead.load();
         while (currentHead.offset >= 0) {
             // The value of "next" we load here might already contain application data if another
```
poweifeng, line 283:
```suggestion
            // The value of "pNext" we load here might already contain application data if another
```

poweifeng (approved):

PR #7545, opened 2024-02-03
OpenGL: release the resources allocated by the shader compiler

andreperezmaselco:
Add an OpenGL hint to release the resources allocated by the shader compiler.

### DIFF
--- a/filament/backend/src/opengl/ShaderCompilerService.cpp
+++ b/filament/backend/src/opengl/ShaderCompilerService.cpp
@@ -584,6 +584,9 @@ void ShaderCompilerService::compileShaders(OpenGLContext& context,
             outShaders[i] = shaderId;
         }
     }
+
+    // Hint to release the resources allocated by the shader compiler.
+    glReleaseShaderCompiler();
 }
 
 // If usages of the Google-style line directive are present, remove them, as some


filament/backend/src/opengl/ShaderCompilerService.cpp
```diff
@@ -584,6 +584,9 @@ void ShaderCompilerService::compileShaders(OpenGLContext& context,
             outShaders[i] = shaderId;
         }
     }
+
+    // Hint to release the resources allocated by the shader compiler.
+    glReleaseShaderCompiler();
```
romainguy, line 589: What are you attempting to solve here? More shaders may need compilation, and this would create unnecessary churn.

andreperezmaselco: Okay, other shaders will be compiled.
Check the OpenGL specification.
This hint does not prevent later use of the shader compiler.
The idea is to deallocate specific resources from previous compilations, not common resources.
Obviously this will depend on the quality of the OpenGL implementation.

romainguy: I know it does not prevent future compilations, but is this actually fixing or improving anything?

andreperezmaselco: Yes, it is improving memory usage.

romainguy: I understand what it's *supposed* to do but have you measured its effect? We would also need to test this on all targets.

andreperezmaselco: I've only tested it on a few Filament samples on Linux.
I understand if you prefer not to proceed with the change due to the uncertainty of different OpenGL implementations on different devices.

pixelflinger (commented):
I think we want to ere on the side of performance, instead if memory usage. Also this worries me in the context of parallel shader compilation, because we've seen driver bugs in that area.

If it was improving things specifically on Android, without much performance impact, I would be more inclined to take this change.


I think what would make more sense, would be to have a filament API an app can call to request filament to free as much memory as it can, for low memory situations, and we could use this API then.

Overall, I don't think the risk outweighs the benefits here.

andreperezmaselco:
Yes. What I think is that, in general, the application will not keep compiling shaders after the rendering loop starts. Therefore, I believe that at this point only what is **needed** for renderings should be allocated.

PR #7547, opened 2024-02-04
Syntax and semantics improvement in GLUtils.{h,cpp}

andreperezmaselco:
• Use the `const` type qualifier.
• Early return in `switch` statement.

### DIFF
--- a/filament/backend/src/opengl/GLUtils.cpp
+++ b/filament/backend/src/opengl/GLUtils.cpp
@@ -28,38 +28,30 @@ using namespace utils;
 namespace GLUtils {
 
 UTILS_NOINLINE
-const char* getGLError(GLenum error) noexcept {
-    const char* string = "unknown";
+const char* getGLError(const GLenum error) noexcept {
     switch (error) {
         case GL_NO_ERROR:
-            string = "GL_NO_ERROR";
-            break;
+            return "GL_NO_ERROR";
         case GL_INVALID_ENUM:
-            string = "GL_INVALID_ENUM";
-            break;
+            return "GL_INVALID_ENUM";
         case GL_INVALID_VALUE:
-            string = "GL_INVALID_VALUE";
-            break;
+            return "GL_INVALID_VALUE";
         case GL_INVALID_OPERATION:
-            string = "GL_INVALID_OPERATION";
-            break;
+            return "GL_INVALID_OPERATION";
         case GL_INVALID_FRAMEBUFFER_OPERATION:
-            string = "GL_INVALID_FRAMEBUFFER_OPERATION";
-            break;
+            return "GL_INVALID_FRAMEBUFFER_OPERATION";
         case GL_OUT_OF_MEMORY:
-            string = "GL_OUT_OF_MEMORY";
-            break;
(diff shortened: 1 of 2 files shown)


filament/backend/src/opengl/GLUtils.cpp
```diff
@@ -28,84 +28,68 @@ using namespace utils;
 namespace GLUtils {
 
 UTILS_NOINLINE
-const char* getGLError(GLenum error) noexcept {
-    const char* string = "unknown";
+const char* getGLError(const GLenum error) noexcept {
```
pixelflinger, line 31: I prefer this method to have a single `return` statement so it's possible to set a breakpoint on it.

filament/backend/src/opengl/GLUtils.cpp
```diff
@@ -28,84 +28,68 @@ using namespace utils;
...
     GLenum const err = checkGLError(out, function, line);
     if (err != GL_NO_ERROR) {
         debug_trap();
     }
 }
 
 UTILS_NOINLINE
-const char* getFramebufferStatus(GLenum status) noexcept {
-    const char* string = "unknown";
+const char* getFramebufferStatus(const GLenum status) noexcept {
```
pixelflinger, line 70: I prefer this method to have a single `return` statement so it's possible to set a breakpoint on it.

filament/backend/src/opengl/GLUtils.cpp
```diff
@@ -28,84 +28,68 @@ using namespace utils;
...
         default:
-            break;
+            return "unknown";
     }
-    return string;
 }
 
 UTILS_NOINLINE
-GLenum checkFramebufferStatus(io::ostream& out, GLenum target, const char* function, size_t line) noexcept {
+GLenum checkFramebufferStatus(io::ostream& out, const GLenum target, const char *const function, const size_t line) noexcept {
```
pixelflinger, line 92: please use `type const name` syntax, which is what we've been starting use everywhere in newer code.

pixelflinger (changes requested):
see notes inline.

I'm not opposed to adding `const` everywhere, but I think it creates unnecessary source changes -- which interfere with `git blame` for instance, for no great reason in this case. Especially that we generally don't do that throughout filament (maybe we should, but that case we should do it globally).

I'm inclined to not take this PR, I don't think it ads great value.

However, adding `const` to pointer types (both for the pointer and pointee when appropriate) is at least somewhat useful. So I'd take those changes.

filament/backend/src/opengl/GLUtils.cpp
andreperezmaselco: That's fine, but then there's no standard across similar functions.
Other functions are using early return in switch statements.

filament/backend/src/opengl/GLUtils.cpp
andreperezmaselco: I repeat the previous comment.

filament/backend/src/opengl/GLUtils.cpp
andreperezmaselco: OK. It has the same meaning.

PR #7553, opened 2024-02-05
Refactor the mesh reader

andreperezmaselco:
- Check if an error occurred when opening the mesh file.
- Use the `const` type qualifier.
- Convert to void expression when not using the function return.
- Remove unused variable.
- Fix pointer dereferencing for material count.
- Use array instead of `std::vector`.

### DIFF
--- a/libs/filameshio/src/MeshReader.cpp
+++ b/libs/filameshio/src/MeshReader.cpp
@@ -159,30 +159,32 @@ static size_t fileSize(int fd) {
 
 namespace filamesh {
 
-MeshReader::Mesh MeshReader::loadMeshFromFile(filament::Engine* engine, const utils::Path& path,
-        MaterialRegistry& materials) {
-
+MeshReader::Mesh MeshReader::loadMeshFromFile(filament::Engine *const engine, const utils::Path& path,
+        MaterialRegistry& materialRegistry) {
     Mesh mesh;
 
-    int fd = open(path.c_str(), O_RDONLY);
+    /* Open the mesh file for reading only. */
+    const int fd = open(path.c_str(), O_RDONLY);
 
-    size_t size = fileSize(fd);
-    char* data = (char*) malloc(size);
-    read(fd, data, size);
+    /* Check if an error occurred when opening the file. */
+    if (fd == -1) return mesh;
+
+    const size_t size = fileSize(fd);
+    const char *const data = (char *)malloc(size);
 
     if (data) {
-        char *p = data;
-        char *magic = p;
-        p += sizeof(MAGICID);
+        (void)read(fd, (void *)data, size);
 
-        if (!strncmp(MAGICID, magic, 8)) {
-            mesh = loadMeshFromBuffer(engine, data, nullptr, nullptr, materials);
+        if (strncmp(MAGICID, data, 8) == 0) {
+            mesh = loadMeshFromBuffer(engine, data, nullptr, nullptr, materialRegistry);
         }
 
         Fence::waitAndDestroy(engine->createFence());
-        free(data);
+        free((void *)data);
     }
-    close(fd);
+
+    /* Close the file. */
+    (void)close(fd);
 
     return mesh;
 }
@@ -197,7 +199,7 @@ MeshReader::Mesh MeshReader::loadMeshFromBuffer(filament::Engine* engine,
 
 MeshReader::Mesh MeshReader::loadMeshFromBuffer(filament::Engine* engine,
         void const* data, Callback destructor, void* user,
-        MaterialRegistry& materials) {
+        MaterialRegistry& materialRegistry) {
     const uint8_t* p = (const uint8_t*) data;
     if (strncmp(MAGICID, (const char *) p, 8)) {
         utils::slog.e << "Magic string not found." << utils::io::endl;
@@ -217,14 +219,14 @@ MeshReader::Mesh MeshReader::loadMeshFromBuffer(filament::Engine* engine,
     Part* parts = (Part*) p;
     p += header->parts * sizeof(Part);
 
-    uint32_t materialCount = (uint32_t) *p;
+    const uint32_t materialCount = *(const uint32_t *)p;
     p += sizeof(uint32_t);
 
-    std::vector<std::string> partsMaterial(materialCount);
+    std::string materials[materialCount];
     for (size_t i = 0; i < materialCount; i++) {
         uint32_t nameLength = (uint32_t) *p;
         p += sizeof(uint32_t);
-        partsMaterial[i] = (const char*) p;
+        materials[i] = (const char*) p;
         p += nameLength + 1; // null terminated
     }
 
@@ -354,7 +356,7 @@ MeshReader::Mesh MeshReader::loadMeshFromBuffer(filament::Engine* engine,
     RenderableManager::Builder builder(header->parts);
     builder.boundingBox(header->aabb);
 
-    const auto defaultmi = materials.getMaterialInstance(utils::CString(DEFAULT_MATERIAL));
+    const auto defaultmi = materialRegistry.getMaterialInstance(utils::CString(DEFAULT_MATERIAL));
     for (size_t i = 0; i < header->parts; i++) {
         builder.geometry(i, RenderableManager::PrimitiveType::TRIANGLES,
                 mesh.vertexBuffer, mesh.indexBuffer, parts[i].offset,
@@ -363,18 +365,18 @@ MeshReader::Mesh MeshReader::loadMeshFromBuffer(filament::Engine* engine,
         // It may happen that there are more parts than materials
         // therefore we have to use Part::material instead of i.
         uint32_t materialIndex = parts[i].material;
-        if (materialIndex >= partsMaterial.size()) {
+        if (materialIndex >= materialCount) {
             utils::slog.e << "Material index (" << materialIndex << ") of mesh part ("
-                    << i << ") is out of bounds (" << partsMaterial.size() << ")" << utils::io::endl;
+                    << i << ") is out of bounds (" << materialCount << ")" << utils::io::endl;
             continue;
         }
 
         const utils::CString materialName(
-                partsMaterial[materialIndex].c_str(), partsMaterial[materialIndex].size());
-        const auto mat = materials.getMaterialInstance(materialName);
+                materials[materialIndex].c_str(), materials[materialIndex].size());
+        const auto mat = materialRegistry.getMaterialInstance(materialName);
         if (mat == nullptr) {
             builder.material(i, defaultmi);
-            materials.registerMaterialInstance(materialName, defaultmi);
+            materialRegistry.registerMaterialInstance(materialName, defaultmi);
         } else {
             builder.material(i, mat);
         }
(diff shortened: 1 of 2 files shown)


libs/filameshio/src/MeshReader.cpp
```diff
@@ -159,30 +159,32 @@ static size_t fileSize(int fd) {
...
-    char* data = (char*) malloc(size);
-    read(fd, data, size);
+    const size_t size = fileSize(fd);
+    const char *const data = (char *)malloc(size);
 
     if (data) {
-        char *p = data;
-        char *magic = p;
-        p += sizeof(MAGICID);
+        (void)read(fd, (void *)data, size);
```
pixelflinger, line 176: That's not a style we're using anywhere. Why not use the return value and do the `strncmp` only if successful? Or, at least `assert()` if not.

libs/filameshio/src/MeshReader.cpp
```diff
@@ -159,30 +159,32 @@ static size_t fileSize(int fd) {
...
         }
 
         Fence::waitAndDestroy(engine->createFence());
-        free(data);
+        free((void *)data);
     }
-    close(fd);
+
+    /* Close the file. */
+    (void)close(fd);
```
pixelflinger, line 187: don't do that.

libs/filameshio/src/MeshReader.cpp
```diff
@@ -217,14 +219,14 @@ MeshReader::Mesh MeshReader::loadMeshFromBuffer(filament::Engine* engine,
     Part* parts = (Part*) p;
     p += header->parts * sizeof(Part);
 
-    uint32_t materialCount = (uint32_t) *p;
+    const uint32_t materialCount = *(const uint32_t *)p;
```
pixelflinger, line 222: `uint32_t const materialCount = *(uint32_t const *)p;`

libs/filameshio/src/MeshReader.cpp
```diff
@@ -217,14 +219,14 @@ MeshReader::Mesh MeshReader::loadMeshFromBuffer(filament::Engine* engine,
...
     p += sizeof(uint32_t);
 
-    std::vector<std::string> partsMaterial(materialCount);
+    std::string materials[materialCount];
```
pixelflinger, line 225: You can't do that. Variable size static arrays is a compiler extension, it's not standard.

pixelflinger (changes requested):
see inline. the main issue is that you can't use variable size C arrays, it's a compiler extension and not part of the standard.

libs/filameshio/src/MeshReader.cpp
andreperezmaselco: Why?

libs/filameshio/src/MeshReader.cpp
andreperezmaselco: OK.

libs/filameshio/src/MeshReader.cpp
```diff
@@ -159,30 +159,32 @@ static size_t fileSize(int fd) {
...
 
-MeshReader::Mesh MeshReader::loadMeshFromFile(filament::Engine* engine, const utils::Path& path,
-        MaterialRegistry& materials) {
-
+MeshReader::Mesh MeshReader::loadMeshFromFile(filament::Engine *const engine, const utils::Path& path,
+        MaterialRegistry& materialRegistry) {
     Mesh mesh;
 
-    int fd = open(path.c_str(), O_RDONLY);
+    /* Open the mesh file for reading only. */
```
romainguy, line 166: This comment isn't very useful

libs/filameshio/src/MeshReader.cpp
```diff
@@ -159,30 +159,32 @@ static size_t fileSize(int fd) {
...
+    const int fd = open(path.c_str(), O_RDONLY);
+
+    /* Check if an error occurred when opening the file. */
```
romainguy, line 169: Same here, this comment doesn't add anything useful

libs/filameshio/src/MeshReader.cpp
```diff
@@ -159,30 +159,32 @@ static size_t fileSize(int fd) {
...
+            mesh = loadMeshFromBuffer(engine, data, nullptr, nullptr, materialRegistry);
         }
 
         Fence::waitAndDestroy(engine->createFence());
-        free(data);
+        free((void *)data);
     }
-    close(fd);
+
+    /* Close the file. */
```
romainguy, line 186: Again, this comment is not useful

libs/filameshio/src/MeshReader.cpp
```diff
@@ -159,30 +159,32 @@ static size_t fileSize(int fd) {
...
 
-        if (!strncmp(MAGICID, magic, 8)) {
-            mesh = loadMeshFromBuffer(engine, data, nullptr, nullptr, materials);
+        if (strncmp(MAGICID, data, 8) == 0) {
+            mesh = loadMeshFromBuffer(engine, data, nullptr, nullptr, materialRegistry);
         }
 
         Fence::waitAndDestroy(engine->createFence());
-        free(data);
+        free((void *)data);
```
romainguy, line 183: Why the cast here?

libs/filameshio/src/MeshReader.cpp
```diff
@@ -159,30 +159,32 @@ static size_t fileSize(int fd) {
...
 
     if (data) {
-        char *p = data;
-        char *magic = p;
-        p += sizeof(MAGICID);
+        (void)read(fd, (void *)data, size);
 
-        if (!strncmp(MAGICID, magic, 8)) {
-            mesh = loadMeshFromBuffer(engine, data, nullptr, nullptr, materials);
+        if (strncmp(MAGICID, data, 8) == 0) {
```
romainguy, line 178: Don't make unnecessary changes please.

libs/filameshio/src/MeshReader.cpp
romainguy: The difference really won't matter. Keep as-is.

romainguy:
Note: this is not a refactoring as the title of the PR implies

libs/filameshio/src/MeshReader.cpp
pixelflinger: it's not a style we're using anywhere else in filament. I don't want to start diverging in style, without a strong benefit.

libs/filameshio/src/MeshReader.cpp
pixelflinger: I know, but that's the style we strive use elsewhere (even if we're not always consistant, that's the style we should be using).

libs/filameshio/src/MeshReader.cpp
```diff
@@ -159,30 +159,32 @@ static size_t fileSize(int fd) {
...
+    const int fd = open(path.c_str(), O_RDONLY);
+
+    /* Check if an error occurred when opening the file. */
+    if (fd == -1) return mesh;
 
-    size_t size = fileSize(fd);
-    char* data = (char*) malloc(size);
-    read(fd, data, size);
+    const size_t size = fileSize(fd);
+    const char *const data = (char *)malloc(size);
```
romainguy, line 173: @pixelflinger We should decide what style we want for pointers declarations. It seems most of our casts are `(char*) foo` (not space before *, one space after the cast), but not always.

pixelflinger: it seems to me the cast should follow the same style as a variable declaration. Since we use:

`char *value;` and not `char* value;` in most of our code, I think a cast should be `(char *)` as well.  It seems to me we usually don't put a space after the cast (this file did, but that seems to be an outlier, I think).

so, I think `(char *)value` is probably the style I would go with.

libs/filameshio/src/MeshReader.cpp
andreperezmaselco: Because `data` is not of type `void *`.
It is a syntax error if the cast is not done.

andreperezmaselco:
> Note: this is not a refactoring as the title of the PR implies

No? What is it then?

libs/filameshio/src/MeshReader.cpp
andreperezmaselco: But the prerequisites for building Filament say clang 14.0 (or more recent) according to the BUILDING.md file.
This compiler allows such allocation.

romainguy:
Refactoring would imply modifying the structure (classes/methods/functions/etc.).

libs/filameshio/src/MeshReader.cpp
andreperezmaselco: OK. It's a choice.

libs/filameshio/src/MeshReader.cpp
andreperezmaselco: OK. It's a choice.

andreperezmaselco:
> Refactoring would imply modifying the structure (classes/methods/functions/etc.).

It is a possible definition, but it is not the only one.

libs/filameshio/src/MeshReader.cpp
romainguy: It doesn't matter, we should stick to the standard. std::vector works just fine here, there's no reason to change this.

libs/filameshio/src/MeshReader.cpp
andreperezmaselco: OK, but the style regarding this type of sentence is not clear.

libs/filameshio/src/MeshReader.cpp
andreperezmaselco: OK. It's a choice.

libs/filameshio/src/MeshReader.cpp
andreperezmaselco: OK. It's a choice.

libs/filameshio/src/MeshReader.cpp
andreperezmaselco: OK. It's a choice.

libs/filameshio/src/MeshReader.cpp
andreperezmaselco: OK. It's a choice.

libs/filameshio/src/MeshReader.cpp
andreperezmaselco: It's useful for someone who is reading the code and doesn't know that -1 is returned when there is an error.
Being useful or not is a relative matter.
You should think about considering the subjectivity of code contributors.

libs/filameshio/src/MeshReader.cpp
andreperezmaselco: Not very useful for whom?

libs/filameshio/src/MeshReader.cpp
andreperezmaselco: It matters if you don't want to waste memory.

libs/filameshio/src/MeshReader.cpp
andreperezmaselco: The same here.

libs/filameshio/src/MeshReader.cpp
romainguy: The comment provides no useful information that can't be deducted from reading the code or from basic knowledge of common APIs. It's the equivalent of commenting `getAlpha()` with "Returns the alpha value.". Comments are great if they provide insights into what the code does and why (and to be fair, I'm sure we have many comments in Filament that aren't very useful).

andreperezmaselco: It is not the same thing.
For example, if a person has generally worked with more abstract functions like `fopen`, then what reason does one have to think that `open` will specifically open a file?

PR #7556, opened 2024-02-06
Metal: track buffer allocations

bejado:


filament/backend/include/backend/Platform.h
```diff
@@ -188,9 +188,28 @@ class UTILS_PUBLIC Platform {
...
+    using DebugLogFunc = utils::Invocable<void(const char* UTILS_NONNULL str, size_t len)>;
+
+    /**
+     * Sets the callback function that the backend can use to log backend-specific debug
+     * information.
+     *
+     * @param str A null-terminated c-string containing debug log information.
+     * @param len The length of str, not including the null terminator.
+     */
+    void setDebugLogFunc(DebugLogFunc&& debugLog) noexcept;
```
pixelflinger, line 200: maybe this comment should be on `DebugLogFunc`. Also talk about the threading expectations (i.e. should the callback implementation be thread-safe).

filament/backend/include/backend/Platform.h
```diff
@@ -188,9 +188,28 @@ class UTILS_PUBLIC Platform {
...
+
+    /**
+     * @return true if debugLog is valid.
+     */
+    bool hasDebugLogFunc() const noexcept;
+
+    void debugLog(const char* UTILS_NONNULL str, size_t len);
```
pixelflinger, line 207: This needs a comment that explains that this is to be used by the backend.

pixelflinger (approved):

PR #7557, opened 2024-02-06
rework how we size the HandleAllocator's pools

pixelflinger:
- update the pools sizes for metal and vulkan, which were very outdated.
- add debug code on all backends to print the size of each handle  (with a compile time switch)

The most important change is that now the 3 pools of HandleAllocator are sized so that each can accommodate about the same amount of handles. This makes it easier to reason about. The total amount of handles is three times that, since there are 3 pools. 
We also try to allocate the buckets so that handles are evenly distributed, however, that's very hand wavy.

With the current setup the number of handles per pool is as follows:
- GL : 3240 / pool / MiB
- VK : 1820 / pool / MiB
- MTL: 1310 / pool / MiB

### DIFF
--- a/filament/backend/include/private/backend/HandleAllocator.h
+++ b/filament/backend/include/private/backend/HandleAllocator.h
@@ -40,18 +40,22 @@
 #   define HANDLE_TYPE_SAFETY 0
 #endif
 
-#define HandleAllocatorGL  HandleAllocator<16, 64, 208>
-#define HandleAllocatorVK  HandleAllocator<16, 64, 880>
-#define HandleAllocatorMTL HandleAllocator<16, 64, 584>
+#define HandleAllocatorGL  HandleAllocator<16,  64, 208>    // ~3640 / pool / MiB
+#define HandleAllocatorVK  HandleAllocator<80, 176, 320>    // ~1820 / pool / MiB
+#define HandleAllocatorMTL HandleAllocator<48, 160, 592>    // ~1310 / pool / MiB
 
 namespace filament::backend {
 
 /*
  * A utility class to efficiently allocate and manage Handle<>
  */
-template <size_t P0, size_t P1, size_t P2>
+template<size_t P0, size_t P1, size_t P2>
 class HandleAllocator {
 public:
+
+    static_assert(P0 % 16 == 0, "HandleAllocator Pools must be multiple of 16 bytes");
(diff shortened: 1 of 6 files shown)


pixelflinger:
@bejado MetalRenderPrimitive is an very big object, which is unfortunate because we expect to have tons of those. On the GL side this corresponds to a VAO and is a very small object. I was wondering if there was a way to reduce its size significantly (not by allocating stuff in the heap, that doesn't really helps). If not, maybe we should rethink our backend API, if possible, to be more metal friendly.

poweifeng (approved):

bejado:
> MetalRenderPrimitive is an very big object, which is unfortunate because we expect to have tons of those. On the GL side this corresponds to a VAO and is a very small object. I was wondering if there was a way to reduce its size significantly (not by allocating stuff in the heap, that doesn't really helps). If not, maybe we should rethink our backend API, if possible, to be more metal friendly. 

[`VertexDescription`](https://github.com/google/filament/blob/af792e3d184775784b431536b3f76bbd5b6492c4/filament/backend/src/metal/MetalState.h#L85) is the main contributor to its size. Without using dynamic memory I'm not sure how to make it smaller, but we could consider making it a separate driver object, maybe `HwVertexAssembly` or something -- that way render primitives can share vertex assembly data. I assume this would be helpful for Vulkan as well.

pixelflinger:
> > MetalRenderPrimitive is an very big object, which is unfortunate because we expect to have tons of those. On the GL side this corresponds to a VAO and is a very small object. I was wondering if there was a way to reduce its size significantly (not by allocating stuff in the heap, that doesn't really helps). If not, maybe we should rethink our backend API, if possible, to be more metal friendly.
> 
> [`VertexDescription`](https://github.com/google/filament/blob/af792e3d184775784b431536b3f76bbd5b6492c4/filament/backend/src/metal/MetalState.h#L85) is the main contributor to its size. Without using dynamic memory I'm not sure how to make it smaller, but we could consider making it a separate driver object, maybe `HwVertexAssembly` or something -- that way render primitives can share vertex assembly data. I assume this would be helpful for Vulkan as well.

This is confusing to me because a `RenderPrimitive` is essentially just a reference to a VertexBuffer and IndexBuffer, so why is all that state not stored (or storable) there.

PR #7559, opened 2024-02-06
[release] update `base64` command

poweifeng:
Seems like a `-i` is now necessary for the command. Note that we recently startede using mac-mx machines.


bejado (approved):
Thanks for fixing!

pixelflinger (approved):

PR #7561, opened 2024-02-07
Better handle collisions and use-after-free detection

pixelflinger:
- each handle now has a 4-bits "age", meaning that handles are recycled only after 16 alloc/free cycles. This is used to detect double-free and use-after free. This should also allow us to compare handles, because freeing and reallocating an object, won't produce the same Handle (at least for 16 rounds).

- removed "type safety" checks because it's almost impossible to get it wrong thanks to our compile time type safety checks. This didn't provide a useful value added.

- This feature is built on top of being able to set/get a 8 bits tag associated with the memory block returned by the pool allocator. We use the "extra" parameter of the allocator to allocate a "hidden" structure containing the age of that memory block.

- Also we don't allow to compare Handle<> of different types

### DIFF
--- a/filament/backend/include/private/backend/HandleAllocator.h
+++ b/filament/backend/include/private/backend/HandleAllocator.h
@@ -24,21 +24,18 @@
 #include <utils/compiler.h>
 #include <utils/debug.h>
 #include <utils/ostream.h>
+#include <utils/Panic.h>
 
 #include <tsl/robin_map.h>
 
+#include <cstddef>
 #include <exception>
 #include <type_traits>
 #include <unordered_map>
+#include <utility>
 
 #include <stddef.h>
 #include <stdint.h>
-
-#if !defined(NDEBUG) && UTILS_HAS_RTTI
-#   define HANDLE_TYPE_SAFETY 1
-#else
-#   define HANDLE_TYPE_SAFETY 0
-#endif
 
 #define HandleAllocatorGL  HandleAllocator<16,  64, 208>    // ~3640 / pool / MiB
 #define HandleAllocatorVK  HandleAllocator<80, 176, 320>    // ~1820 / pool / MiB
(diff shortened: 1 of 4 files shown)


filament/backend/include/private/backend/HandleAllocator.h
```diff
@@ -24,22 +24,19 @@
 #include <utils/compiler.h>
 #include <utils/debug.h>
 #include <utils/ostream.h>
+#include <utils/Panic.h>
 
 #include <tsl/robin_map.h>
 
+#include <cstddef>
```
poweifeng, line 31: is this needed even with `<stddef.h>` below?

filament/backend/include/private/backend/HandleAllocator.h
```diff
@@ -305,42 +292,51 @@ class HandleAllocator {
...
-    inline void* handleToPointer(HandleBase::HandleId id) const noexcept {
+    inline std::pair<void*, uint32_t> handleToPointer(HandleBase::HandleId id) const noexcept {
         // note: the null handle will end-up returning nullptr b/c it'll be handled as
         // a non-pool handle.
         if (UTILS_LIKELY(isPoolHandle(id))) {
             char* const base = (char*)mHandleArena.getArea().begin();
-            size_t offset = id << Allocator::MIN_ALIGNMENT_SHIFT;
-            return static_cast<void*>(base + offset);
+            uint32_t const tag = id & HANDLE_TAG_MASK;
+            size_t offset = (id & HANDLE_INDEX_MASK) * Allocator::getAlignment();
```
poweifeng, line 324:
```suggestion
            size_t const offset = (id & HANDLE_INDEX_MASK) * Allocator::getAlignment();
```

filament/backend/include/private/backend/HandleAllocator.h
```diff
@@ -223,29 +190,55 @@ class HandleAllocator {
...
         }
 
         // this is in fact always called with a constexpr size argument
-        inline void free(void* p, size_t size) noexcept {
+        inline void free(void* p, size_t size, uint8_t age) noexcept {
             assert_invariant(p >= mArea.begin() && (char*)p + size <= (char*)mArea.end());
+
+            // check for double-free
+            Node* const pNode = static_cast<Node*>(p);
+            uint8_t& expectedAge = pNode[-1].age;
```
poweifeng, line 237: would you mind a little explanation on this line?  if p == mArea.begin(), are we going out of bounds?

poweifeng (approved):

filament/backend/include/private/backend/HandleAllocator.h
pixelflinger: yes needed for `std::max_alight_t`

filament/backend/include/private/backend/HandleAllocator.h
pixelflinger: okay. the way it works is that when you specify an "Extra" value to the allocator, it always allocates at least "Extra" bytes before the allocation.

PR #7566, opened 2024-02-09
gltfio: use tangent space mesh

### LINKED ISSUE #6358: SurfaceOrientation doesn't duplicate vertices when generating flat normals

### LINKED ISSUE #7444: render flaw in lastest filament version

**Describe the bug**
I find render flaw with my glb model using lastest filament on desktop， here is contrast：
in filament:
image
image

in website  https://gltf-viewer.donmccurdy.com/
image
image

I guess whether if some flaw in TBN algorithm, most method show TBN should be calculated in vs or ps， below is my test glb model.
[up.zip](https://github.com/google/filament/files/13722078/up.zip)

**To Reproduce**
1. ...

**Expected behavior**

**Screenshots**

**Logs**
If applicable, copy **full** logs from your console here. Please *do not*
use screenshots of logs, copy them as text, use gist or attach an *uncompressed* file.

 - OS: windows 10
 - GPU: amd
 - Backend: opengl

 - Device: [e.g. Pixel 2]
 - OS: [e.g. Android Pie 9.0]

**Additional context**

poweifeng:
- Fix missing attributes in TangentSpaceMesh
     - Fix missing reference in MikktspaceImpl.cpp
     - Add gltfio/src/extended to implement an alternate loader for
       primitives. This is largely based on the implementation in
       AssetLoader/ResourceLoader
     - Pull common methods between the two implementations into a
       utility namespace.
     - Switch gltf_viewer to use new gltf loader (extended loader).
     - Able to correctly produce flat shading from gltf that only have
       vertex positions and indices.
     - Able to run mikktspace to generate tangent spaces as per gltf
       spec
     - Note that java/webgl samples have not been ported to this
       framework due to difficulty in maintaining interaction between
       AssetLoader/ResourceLoader
    
    Fixes #7444; Fixes #6358

### DIFF
--- a/libs/gltfio/include/gltfio/AssetLoader.h
+++ b/libs/gltfio/include/gltfio/AssetLoader.h
@@ -38,6 +38,18 @@ namespace filament::gltfio {
 
 class NodeManager;
 
+// Use this struct to enable mikktspace-based tangent-space computation.
+/**
+ * \struct AssetConfigurationExtended AssetLoader.h gltfio/AssetLoader.h
+ * \brief extends struct AssetConfiguration
+ * Useful if client needs mikktspace tangents.
+ */
+struct AssetConfigurationExtended {
+    //! Optional The same parameter as provided to \struct ResourceConfiguration ResourceLoader.h
+    //! gltfio/ResourceLoader.h
+    char const* gltfPath;
+};
+
 /**
  * \struct AssetConfiguration AssetLoader.h gltfio/AssetLoader.h
  * \brief Construction parameters for AssetLoader.
@@ -62,6 +74,9 @@ struct AssetConfiguration {
 
     //! Optional default node name for anonymous nodes
     char* defaultNodeName = nullptr;
+
+    //! Optional to enable mikktspace tangents.
+    AssetConfigurationExtended* ext = nullptr;
 };
 
 /**
--- a/NEW_RELEASE_NOTES.md
+++ b/NEW_RELEASE_NOTES.md
@@ -7,3 +7,5 @@ for next branch cut* header.
 appropriate header in [RELEASE_NOTES.md](./RELEASE_NOTES.md).
 
 ## Release notes for next branch cut
+
+ - gltfio: support for fallback flat-shading and mikktspace tangents
--- a/android/gltfio-android/CMakeLists.txt
+++ b/android/gltfio-android/CMakeLists.txt
@@ -81,9 +81,17 @@ set(GLTFIO_SRCS
         ${GLTFIO_DIR}/src/TangentsJob.cpp
         ${GLTFIO_DIR}/src/TangentsJob.h
         ${GLTFIO_DIR}/src/UbershaderProvider.cpp
+        ${GLTFIO_DIR}/src/Utility.cpp
+        ${GLTFIO_DIR}/src/Utility.h
         ${GLTFIO_DIR}/src/Wireframe.cpp
         ${GLTFIO_DIR}/src/Wireframe.h
         ${GLTFIO_DIR}/src/downcast.h
+        ${GLTFIO_DIR}/src/extended/AssetLoaderExtended.cpp
+        ${GLTFIO_DIR}/src/extended/AssetLoaderExtended.h
+        ${GLTFIO_DIR}/src/extended/ResourceLoaderExtended.cpp
+        ${GLTFIO_DIR}/src/extended/ResourceLoaderExtended.h
+        ${GLTFIO_DIR}/src/extended/TangentsJobExtended.cpp
+        ${GLTFIO_DIR}/src/extended/TangentsJobExtended.h
 
         src/main/cpp/Animator.cpp
         src/main/cpp/AssetLoader.cpp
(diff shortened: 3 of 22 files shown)


poweifeng:
Please take a look.  If the approach looks ok, I'll add the java/js bindings.

libs/gltfio/include/gltfio/AssetLoader.h
```diff
@@ -38,6 +38,18 @@ namespace filament::gltfio {
...
+// Use this struct to enable mikktspace-based tangent-space computation.
+/**
+ * \struct AssetConfigurationExtended AssetLoader.h gltfio/AssetLoader.h
+ * \brief extends struct AssetConfiguration
+ * Useful if client needs mikktspace tangents.
+ */
+struct AssetConfigurationExtended {
+    //! Optional The same parameter as provided to \struct ResourceConfiguration ResourceLoader.h
+    //! gltfio/ResourceLoader.h
+    char const* gltfPath;
```
bejado, line 50: I think this is a significant philosophical shift in API. With this change, now the `AssetLoader` is locked to a specific glTF path. Before, this was limited to `ResourceLoader`. I don't have a good solution off the top of my head, but I would be in favor of removing the concept of the "file system" from gltfio. Instead, we could require the client to search the file system and provide all the resource data through `addResourceData`. In fact, this is what iOS, Android, and Web already do.

poweifeng: This is an interesting point. I didn't really understand the need to separate AssetLoader and ResourceLoader, and your explanation makes that much clearer.  

I think, longterm, if I'm able to, I'd like to move toward the future you described where filesystem is abstracted out. And then we wouldn't have to have two API entrance points for loading gltf. For now, this is kind of just hacking to make things work on top of the existing two classes. 

Still, I'd like to introduce this change, which will admittedly only benefit desktops and only when all the data are on disk, but it'll definitely make the desktop gtlf_viewer nicer.

poweifeng:
I'm breaking this into several changes, so marking as draft

libs/gltfio/include/gltfio/AssetLoader.h
Nolram12345: I'd like to add on to this that the ability to load data from *anywhere*, not just the filesystem, is really quite important, whether it's loading from a network, or a compressed archive of some kind. Only being able to load from regular filesystem files isn't flexible enough.

poweifeng: Noted. I'll likely add to this with a more general loading mechanism in a follow-up.  Thank you for the comment.

poweifeng:
implemented in:
 - gltfio: refactor for clarity (#7652)
 - gltfio: add extended tangents job (#7666)
 - gltfio: add Asset/Resource extended implementations (7678) 
 - gltfio: enable extended implementation (7776)

PR #7569, opened 2024-02-09
Add OpenGL extension for multiview

z3moon:
This extension is going to be used for multiview implementation in OpenGL.


filament/backend/src/opengl/OpenGLDriver.cpp
```diff
@@ -1900,11 +1900,12 @@ bool OpenGLDriver::isSRGBSwapChainSupported() {
...
 
 bool OpenGLDriver::isStereoSupported() {
-    // Stereo requires instancing and EXT_clip_cull_distance.
+    // Instanced-stereo requires instancing and EXT_clip_cull_distance.
+    // Multiview-stereo requires ES 3.0 and OVR_multiview2.
     if (UTILS_UNLIKELY(mContext.isES2())) {
         return false;
     }
-    return mContext.ext.EXT_clip_cull_distance;
+    return mContext.ext.EXT_clip_cull_distance || mContext.ext.OVR_multiview2;
```
z3moon, line 1908: I'm not quite satisfied with this as it could mean only one of them is available. If we can use a preprocessor, something like BACKEND_OPENGL_MULTIVIEW_STEREO and BACKEND_OPENGL_INSTANCED_STEREO, the semantics of this function will become more specific. Also it does mean the engine only has one algorithm baked at compile time, which I preferred.

bejado: I think Filament will want the ability to query what type(s) of stereo each backend supports. Could we change the signature to `bool OpenGLDriver::isStereoSupported(StereoType type)`?

z3moon: I update the changes as suggested.

filament/backend/src/metal/MetalDriver.mm
```diff
@@ -754,8 +754,15 @@
...
 
-bool MetalDriver::isStereoSupported() {
-    return true;
+bool MetalDriver::isStereoSupported(backend::StereoscopicType stereoscopicType) {
+    switch (stereoscopicType) {
+    case backend::StereoscopicType::INSTANCED:
+        return true;
+    case backend::StereoscopicType::MULTIVIEW:
+        // TODO: implement multiview feature in Metal.
+    default:
```
pixelflinger, line 763: personally I prefer not to have a `default` because if we ever add an enum type, the compiler will be able to warn us that not all cases are handled. This saved me before :-)

pixelflinger (approved):

filament/backend/src/metal/MetalDriver.mm
z3moon: Ah that's actually better. Let me remove it.

PR #7573, opened 2024-02-13
Modernize HwRenderPrimitive and draw() APIs

pixelflinger:
HwRenderPrimitive doesn't need to know about the index offset and index count, these parameters are only needed when drawing. draw() is updated consequently.

This is a first step towards being able to lower the overhead of draw similar draw calls. 

A side effect of this is that the HwRenderPrimitiveFactory now will cache buffers regardless of their index count & offset.


poweifeng (approved):
vulkan lgtfm

bejado (approved):

PR #7574, opened 2024-02-13
Add stereoscopic type to Engine::Config

z3moon:
This new type value will determine the algorithm used when stereoscopic rendering is enabled.

### DIFF
--- a/android/filament-android/src/main/java/com/google/android/filament/Engine.java
+++ b/android/filament-android/src/main/java/com/google/android/filament/Engine.java
@@ -364,7 +364,7 @@ public class Engine {
          * The type of technique for stereoscopic rendering.
          *
          * This setting determines the algorithm used when stereoscopic rendering is enabled. This
-         * decision applies the entire Engine during the lifetime of the Engine. E.g., multiple
+         * decision applies to the entire Engine for the lifetime of the Engine. E.g., multiple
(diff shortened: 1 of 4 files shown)


filament/backend/include/backend/DriverEnums.h
```diff
@@ -1212,6 +1212,14 @@ enum class Workaround : uint16_t {
     DISABLE_THREAD_AFFINITY
 };
 
+//! The type of technique for stereoscopic rendering
+enum class StereoscopicType : uint8_t {
```
pixelflinger, line 1216: I think we might need a `NONE` enum?

android/filament-android/src/main/java/com/google/android/filament/Engine.java
```diff
@@ -349,6 +360,19 @@ public static class Config {
          */
         public long jobSystemThreadCount = 0;
 
+        /**
+         * The type of technique for stereoscopic rendering.
+         *
+         * This setting determines the algorithm used when stereoscopic rendering is enabled. This
+         * decision applies the entire Engine during the lifetime of the Engine. E.g., multiple
```
bejado, line 367:
```suggestion
         * decision applies to the entire Engine for the lifetime of the Engine. E.g., multiple
```

bejado (approved):

PR #7581, opened 2024-02-20
Metal: throw an NSException when attempting to draw with invalid program

bejado:
Throw an NSException when a program fails to compile and then is used for drawing; this helps aid debugging compiler errors in production, where stdout logs are not available.

b/325321035

### DIFF
--- a/filament/backend/src/metal/MetalShaderCompiler.h
+++ b/filament/backend/src/metal/MetalShaderCompiler.h
@@ -45,7 +45,7 @@ public:
     public:
         using Raster = std::tuple<id<MTLFunction>, id<MTLFunction>>;
         using Compute = id<MTLFunction>;
-        using Error = NSString*;
+        using Error = std::tuple<NSString*, NSString*>; // error message, Program name
         struct None {};
 
         MetalFunctionBundle() : mPrograms{None{}} {}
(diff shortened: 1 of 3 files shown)


filament/backend/src/metal/MetalShaderCompiler.h
```diff
@@ -41,36 +43,74 @@ class MetalShaderCompiler {
...
+            NSString* reason =
+                    [NSString stringWithFormat:
+                            @"Attempting to draw with an id<MTLFunction> that "
+                            @"failed to compile. Compiler error: %@", errorMessage];
+            NSException* compilationFailureException =
+                    [NSException exceptionWithName:@"MetalCompilationFailure"
+                                            reason:reason
+                                          userInfo:nil];
+            [compilationFailureException raise];
+        }
```
pixelflinger, line 74: The "slow" path of vlaidate() should be in the cpp file, so it's not inlined. It should probably be marked as UTILS_NO_INLINE too.

pixelflinger (changes requested):
I think this deserves a better description in the checking comment. Also is this closing a specific bug? (does it need an entry in readme or a FIXES tag?).

filament/backend/src/metal/MetalShaderCompiler.mm
```diff
@@ -225,4 +229,22 @@ bool isReady() const noexcept {
     mCallbackManager.setCallback(handler, callback, user);
 }
 
+UTILS_NOINLINE
+void MetalShaderCompiler::MetalFunctionBundle::validate() const {
+    if (UTILS_LIKELY(!std::holds_alternative<Error>(mPrograms))) {
```
poweifeng, line 234: couldn't this just make use of `isValid`?  Should `None` also throw something?

bejado: Good point about `None`, I updated the code to also throw an exception when it encounters None.

poweifeng (approved):

pixelflinger (approved):

PR #7584, opened 2024-02-21
feat: Add `-w` parameter to `build.sh` to disable Filamat build

mrousavy:
Adds the `-w` option to the script, which excludes filamat from the CMake build (`-DFILAMENT_BUILD_FILAMAT=OFF`).

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -23,6 +23,8 @@ function print_help {
     echo "        This is sometimes needed instead of -c (which still misses some clean steps)."
     echo "    -d"
     echo "        Enable matdbg."
+    echo "    -w"
+    echo "        Exclude Filamat from the build."
     echo "    -f"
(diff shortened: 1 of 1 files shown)


romainguy (commented):
This is very specific, can you explain why you need this in build.sh?

mrousavy:
Hey @romainguy - I think I might've missed something when creating this PR, I thought filamat was fully modular and could easily be decoupled if needed (to shave off the MBs of the binary) - is that not how the Android Maven artefact is set up as well, with filamat being an optional extra dependency?

romainguy:
Filament itself never links against `filamat`. Filamat is however used by `matc` to compile materials, so it's always built for the host anyway. This compile-time flag is only used to not build the Android versions of `filamat`. But like you said, it's a separate AAR anyway, so if you never use it, you'll never pay for its cost.

PR #7590, opened 2024-02-22
Fix broken Android build

poweifeng:
The math for PrimitiveInfo's size is incorrect. We correct the padding.

See commit 11a3c06418dfbfa5a5cfac8248b26b6af48633b1 to see where the padding error originated.


filament/src/RenderPass.h
```diff
@@ -253,6 +253,7 @@ class RenderPass {
         uint16_t instanceCount;                                         // 2 bytes [MSb: user]
         Variant materialVariant;                                        // 1 byte
         uint8_t rfu1;                                                   // 1 byte
+        uint32_t rfu2;                                                  // 4 byte
```
z3moon, line 256: Great, I also found this while correcting the debug issue. Could you correct the comment below within the Command struct regarding the size of this struct?

z3moon (approved):

filament/src/RenderPass.h
poweifeng: done

bejado (approved):

PR #7591, opened 2024-02-22
Add missing changes for NDK version update

z3moon:
These are missing parts from the commit
111ad96134dc171a9ca225821ed4f304ff9e1572.


poweifeng:
I think you'll also want to change `android/build.gradle` and `android/samples/sample-gltf-viewer/build.gradle`; not sure if i missed anywhere else.

poweifeng:
So the previous attempt to bump to 21 is here: https://github.com/google/filament/commit/3d53ae583fa86a65c7732a7011839682861149d3

and the revert is here: 
https://github.com/google/filament/commit/2d43ad59ec2fb27b1a1beb4ebbacee47e5bce95e

Not sure if we'll run into issues in the update again.

z3moon:
Nice catch and thanks for the context. Done.

android/samples/sample-gltf-viewer/build.gradle
```diff
@@ -30,7 +30,7 @@ android {
     compileSdkVersion versions.compileSdk
     defaultConfig {
         applicationId "com.google.android.filament.gltf"
-        minSdkVersion 19
+        minSdkVersion 21
```
poweifeng, line 33: we can probably replace this with `versions.minSdk` like other samples: https://github.com/google/filament/blob/main/android/samples/sample-material-builder/build.gradle#L31C1-L32C1

z3moon: Good point. Done.

poweifeng (approved):

PR #7592, opened 2024-02-23
vk: remove subpasses to simplify descriptor set refactor

poweifeng:
- Also did some clean up of debug code


filament/backend/src/vulkan/VulkanDriver.cpp
```diff
@@ -1415,40 +1415,13 @@ void VulkanDriver::endRenderPass(int) {
                 0, 1, &barrier, 0, nullptr, 0, nullptr);
     }
 
-    if (mCurrentRenderPass.currentSubpass > 0) {
-        for (uint32_t i = 0; i < VulkanPipelineCache::INPUT_ATTACHMENT_COUNT; i++) {
-            mPipelineCache.bindInputAttachment(i, {});
-        }
-        mCurrentRenderPass.currentSubpass = 0;
```
bejado, line 1422: Do we still need the `currentSubpass` field inside `VulkanRenderPass`?

poweifeng: good catch.  removed.

bejado (approved):

pixelflinger (approved):

PR #7601, opened 2024-02-25
Update imgui to v1.90.4

WookeyBiscotti:
The new version of ImGui has support for the sdl3 backend, and there are also several new useful widgets.


bejado:
Thank you for your PR. Unfortunately at the moment we're locked into an older version of imgui due an internal Google dependency, so we would first need to wait for that to be updated.

pixelflinger:
Converting to draft until we are in sync with google's internal repo.

WookeyBiscotti:
@bejado Thanks for the info! Could you please tell me which version filament is locked into?

pixelflinger:
@WookeyBiscotti I am going to close this PR because internally, we're now using version 1.91.0 WIP. The official version is 1.92.3. I think that means we could update filament to 1.90.9.
Of course we would very much appreciate it if you submitted a PR, but if not, we will probably do this ourselves soon.

Issue #7539, opened 2024-01-31
setFrameRateOptions doesn't work


JiangnanFu:
**Describe the bug**
Hello，
I'm want 60fps on android phone with 120Hz refresh rate, and I find setFrameRateOptions API. when set FrameRateOptions.interval = 1 / 60.0f, it doesn't work and I'm still stuck at 120 fps.

filament version 1.43.0

**To Reproduce**
1. renderer set like this 
![image](https://github.com/google/filament/assets/52485635/c7ac28a7-0ed9-4a6d-8759-fa16493d9248)

2. doFrame function like this
![image](https://github.com/google/filament/assets/52485635/2f6406d5-944b-44c2-8971-63dd0c959c7f)

**Expected behavior**
60fps

**Smartphone (please complete the following information):**
 - Device: xiaomi Poco F3
 - OS: Android 13
 - GPU: Andreno 650
 - Backend: OpenGL

pixelflinger:
You should set `interval` to 2 in that case (the unit is "multiples of 1/120" in your case).

However, this will actually not solve your problem because `FrameRateOptions` only controls the dynamic scaling. It's not for setting the desired frame rate (I admit the documentation is not clear).

Currently, it is up to you do to set the pace, `uiHelper` currently doesn't support adjusting the framerate, so you will have to implement your own `doFrame` and call render() every other frame.

JiangnanFu:
I get it, Thanks for your answer!

PR #7604, opened 2024-02-26
Fix Renderer FrameRateOptions interval default

yschimke:
From https://github.com/google/filament/issues/7539

And I got to the same conclusion based on https://github.com/google/filament/blob/f0f7e299d2ce1a5f3f62821b058353fe00c2015f/filament/src/details/View.cpp#L191


pixelflinger (approved):
good catch

PR #7606, opened 2024-02-27
docs: Fix version in README

mrousavy:
Version 1.50.3 is not published to maven. 1.49.3 is the latest.


poweifeng:
Sorry for the inconsistency, but the maven update will be forthcoming (eta next week) so I'd prefer not to change the documentation just yet.

PR #7608, opened 2024-02-27
Rename InstancedStereo as Stereo

z3moon:
We're going to add a new implementation of stereoscopic rendering using multiview. Thus we want to remove the word `Instanced` from all methods and properties.

### DIFF
--- a/libs/filamat/src/MaterialBuilder.cpp
+++ b/libs/filamat/src/MaterialBuilder.cpp
@@ -894,7 +894,7 @@ bool MaterialBuilder::generateShaders(JobSystem& jobSystem, const std::vector<Va
                         .domain = mMaterialDomain,
                         .materialInfo = &info,
                         .hasFramebufferFetch = mEnableFramebufferFetch,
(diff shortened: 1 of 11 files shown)


libs/filamat/src/MaterialBuilder.cpp
```diff
@@ -894,7 +894,7 @@ bool MaterialBuilder::generateShaders(JobSystem& jobSystem, const std::vector<Va
                         .domain = mMaterialDomain,
                         .materialInfo = &info,
                         .hasFramebufferFetch = mEnableFramebufferFetch,
-                        .usesClipDistance = v.variant.hasInstancedStereo(),
+                        .usesClipDistance = v.variant.hasStereo(),
```
poweifeng, line 897: how will you distinguish between the two here, since multiview doesn't need clipDistance.

z3moon: I'm going to add a new param to material later, so the MaterialInfo structure above will include a type of stereoscopic as a member variable.

poweifeng (approved):

PR #7613, opened 2024-02-28
Add a new material param, stereoscopicType

z3moon:
This new parameter allows us to specify which implementation of stereoscopic rendering Filament uses for the material.

This change just includes material parameter addition and shader code changes, so it doesn't affect the current rendering behavior.

These changes will follow as separate commits.
- render pipeline changes
- material parameter override via matc parameter
- material document update

### DIFF
--- a/libs/filamat/src/MaterialBuilder.cpp
+++ b/libs/filamat/src/MaterialBuilder.cpp
@@ -498,6 +498,12 @@ MaterialBuilder& MaterialBuilder::transparencyMode(TransparencyMode mode) noexce
     return *this;
 }
 
+MaterialBuilder& MaterialBuilder::stereoscopicType(StereoscopicType stereoscopicType) noexcept
+{
+    mStereoscopicType = stereoscopicType;
+    return *this;
+}
+
 MaterialBuilder& MaterialBuilder::reflectionMode(ReflectionMode mode) noexcept {
     mReflectionMode = mode;
     return *this;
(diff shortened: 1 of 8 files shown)


libs/filamat/src/MaterialBuilder.cpp
```diff
@@ -498,6 +498,12 @@ MaterialBuilder& MaterialBuilder::transparencyMode(TransparencyMode mode) noexce
     return *this;
 }
 
+MaterialBuilder& MaterialBuilder::stereoscopicType(StereoscopicType stereoscopicType) noexcept
+{
```
pixelflinger, line 502:
```suggestion
MaterialBuilder& MaterialBuilder::stereoscopicType(StereoscopicType stereoscopicType) noexcept {
```

libs/filamat/src/shaders/CodeGenerator.cpp
```diff
@@ -181,6 +181,14 @@ utils::io::sstream& CodeGenerator::generateProlog(utils::io::sstream& out, Shade
     }
     generateDefine(out, "FILAMENT_EFFECTIVE_VERSION", effective_version);
 
+    if (material.stereoscopicType == StereoscopicType::INSTANCED) {
+        generateDefine(out, "FILAMENT_STEREO_INSTANCED", true);
+    } else if (material.stereoscopicType == StereoscopicType::MULTIVIEW) {
+        generateDefine(out, "FILAMENT_STEREO_MULTIVIEW", true);
+    } else {
+        assert(false);
+    }
```
pixelflinger, line 190: if you used a switch/case here you wouldn't need the assert

pixelflinger (approved):

libs/filamat/src/shaders/CodeGenerator.cpp
z3moon: Ah makes sense. Let me change it to switch.

PR #7616, opened 2024-02-29
Update code generator for multiview

z3moon:
It generates required shader code for multiview based on parameters.

### DIFF
--- a/libs/filamat/include/filamat/MaterialBuilder.h
+++ b/libs/filamat/include/filamat/MaterialBuilder.h
@@ -525,7 +525,7 @@ public:
     //! Specify the stereoscopic type (default is INSTANCED)
     MaterialBuilder& stereoscopicType(StereoscopicType stereoscopicType) noexcept;
 
-    //! Specify the number of eye count for stereoscopic rendering
+    //! Specify the number of eyes for stereoscopic rendering
     MaterialBuilder& stereoscopicEyeCount(uint8_t eyeCount) noexcept;
 
     /**
--- a/libs/filamat/include/filamat/MaterialBuilder.h
+++ b/libs/filamat/include/filamat/MaterialBuilder.h
@@ -525,6 +525,9 @@ public:
     //! Specify the stereoscopic type (default is INSTANCED)
     MaterialBuilder& stereoscopicType(StereoscopicType stereoscopicType) noexcept;
 
+    //! Specify the number of eye count for stereoscopic rendering
+    MaterialBuilder& stereoscopicEyeCount(uint8_t eyeCount) noexcept;
+
     /**
      * Enable / disable custom surface shading. Custom surface shading requires the LIT
      * shading model. In addition, the following function must be defined in the fragment
@@ -834,6 +837,7 @@ private:
     VertexDomain mVertexDomain = VertexDomain::OBJECT;
     TransparencyMode mTransparencyMode = TransparencyMode::DEFAULT;
     StereoscopicType mStereoscopicType = StereoscopicType::INSTANCED;
+    uint8_t mStereoscopicEyeCount = 2;
 
     filament::AttributeBitset mRequiredAttributes;
 
--- a/libs/filamat/src/GLSLPostProcessor.cpp
+++ b/libs/filamat/src/GLSLPostProcessor.cpp
@@ -536,13 +536,18 @@ bool GLSLPostProcessor::fullOptimization(const TShader& tShader,
             glslOptions.emit_uniform_buffer_as_plain_uniforms = true;
         }
 
-        // For stereo variants using multiview feature, this generates the shader code below.
-        //   #extension GL_OVR_multiview2 : require
-        //   layout(num_views = 2) in;
-        if (config.variant.hasStereo() &&
-            config.shaderType == ShaderStage::VERTEX &&
-            config.materialInfo->stereoscopicType == StereoscopicType::MULTIVIEW) {
-            glslOptions.ovr_multiview_view_count = config.materialInfo->stereoscopicEyeCount;
+        if (config.variant.hasStereo() && config.shaderType == ShaderStage::VERTEX) {
+            switch (config.materialInfo->stereoscopicType) {
+            case StereoscopicType::INSTANCED:
+                // Nothing to generate
+                break;
+            case StereoscopicType::MULTIVIEW:
+                // For stereo variants using multiview feature, this generates the shader code below.
+                //   #extension GL_OVR_multiview2 : require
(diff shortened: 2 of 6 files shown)


libs/filamat/src/shaders/CodeGenerator.cpp
```diff
@@ -83,6 +90,16 @@ utils::io::sstream& CodeGenerator::generateProlog(utils::io::sstream& out, Shade
                 out << "#version 410 core\n\n";
                 out << "#extension GL_ARB_shading_language_packing : enable\n\n";
             }
+            if (v.hasStereo() && stage == ShaderStage::VERTEX) {
```
poweifeng, line 93: how about just `if (v.hasStereo() && stage == ShaderStage::VERTEX && material.stereoscopicType == StereoscopicType::MULTIVIEW)` ? 

Simliar below?

z3moon: Yeah that was my first code, but I actually changed it in favor of the warning detection in switch. Let me change GLSLPostProcessor.cpp in the same way.

libs/filamat/src/shaders/CodeGenerator.cpp
```diff
@@ -94,6 +111,17 @@ utils::io::sstream& CodeGenerator::generateProlog(utils::io::sstream& out, Shade
     // #included code. This way, glslang reports errors more accurately.
     out << "#extension GL_GOOGLE_cpp_style_line_directive : enable\n\n";
 
+    if (v.hasStereo() && stage == ShaderStage::VERTEX) {
+        switch (material.stereoscopicType) {
+        case StereoscopicType::INSTANCED:
+            // Nothing to generate
+            break;
+        case StereoscopicType::MULTIVIEW:
+            out << "layout(num_views = " << material.stereoscopicEyeCount << ") in;\n";
```
poweifeng, line 120: shouldn't this be `CONFIG_STEREO_EYE_COUNT` ?

libs/filamat/src/GLSLPostProcessor.cpp
```diff
@@ -536,6 +536,20 @@ bool GLSLPostProcessor::fullOptimization(const TShader& tShader,
...
+        if (config.variant.hasStereo() && config.shaderType == ShaderStage::VERTEX) {
+            switch (config.materialInfo->stereoscopicType) {
+            case StereoscopicType::INSTANCED:
+                // Nothing to generate
+                break;
+            case StereoscopicType::MULTIVIEW:
+                // For stereo variants using multiview feature, this generates the shader code below.
+                //   #extension GL_OVR_multiview2 : require
+                //   layout(num_views = 2) in;
+                glslOptions.ovr_multiview_view_count = config.materialInfo->stereoscopicEyeCount;
```
poweifeng, line 548: Is this meant for the backend to consume?  If so, how will you keep it in sync with the spec constant value?  I wonder how the metal backend is using the spec constant.

libs/filamat/src/shaders/CodeGenerator.cpp
z3moon: No, it should be the number of currently configured number.

poweifeng: but then how is the spec constant used in the shader?

libs/filamat/src/GLSLPostProcessor.cpp
z3moon: This is for spirv-cross to generate glsl code when we pass  the optimize (-S) param or target Metal/Vulkan to matc. Normally this pass will be not hit unless we generate glsl code using matinfo. The spec constant is updated with the same value.

libs/filamat/src/shaders/CodeGenerator.cpp
z3moon: That spec constant, CONFIG_STEREO_EYE_COUNT, is only being referenced for instanced stereo. And this code generation is only for multiview stereo.

poweifeng: So it's *not* possible to change the number of views dynamically via spec constant?

poweifeng: hmm, is that a limitation of glsl? (where you can't put a `#define` constant in `layout()` or something)?  @pixelflinger , would you mind weighing in on your original comment concerning spec const?

libs/filamat/include/filamat/MaterialBuilder.h
```diff
@@ -525,6 +525,9 @@ class UTILS_PUBLIC MaterialBuilder : public MaterialBuilderBase {
     //! Specify the stereoscopic type (default is INSTANCED)
     MaterialBuilder& stereoscopicType(StereoscopicType stereoscopicType) noexcept;
 
+    //! Specify the number of eye count for stereoscopic rendering
```
bejado, line 528: nit:
```suggestion
    //! Specify the number of eyes for stereoscopic rendering
```

Maybe also document the default number of eyes.

bejado (approved):

libs/filamat/src/shaders/CodeGenerator.cpp
z3moon: AFAICT, this is where spec constant is generated https://github.com/google/filament/blob/d21613a5e666e5e28280d8f19b9a191d23453083/libs/filamat/src/shaders/CodeGenerator.cpp#L862-L879

And this is where the constant is added. https://github.com/google/filament/blob/d21613a5e666e5e28280d8f19b9a191d23453083/libs/filamat/src/shaders/CodeGenerator.cpp#L295

In the function, it uses #define for non-spirv, in which case it is possible to synthesize something like "layout(view_nums=MY_DEFINE)". However, it's not possible in the other case like "layout(view_nums=g_eyecnt_variable)".

There are many places where it directly generates "layout" in CodeGenerator.cpp. One example is https://github.com/google/filament/blob/d21613a5e666e5e28280d8f19b9a191d23453083/libs/filamat/src/shaders/CodeGenerator.cpp#L98

poweifeng: I see so the problem is the spirv codegen.  In that case, I would just generate `"layout(num_views = " << material.stereoscopicEyeCount << ")` for spirv and use the `#define` for gl/metal. Since the vk backend will have to do a non-trivial substitution for this regardless.

poweifeng (approved):
lgtm given offline discussion

poweifeng (approved):

pixelflinger (approved):

PR #7620, opened 2024-02-29
vk: Initial draft for descriptor set refactoring

poweifeng:
- Add cache for ds layouts and ds
 - Abstract descriptors API into VulkanDescriptorSetManager
 - Note that this is just a draft and not hooked into the current implementation.

### DIFF
--- a/filament/backend/src/vulkan/VulkanUtility.h
+++ b/filament/backend/src/vulkan/VulkanUtility.h
@@ -23,6 +23,8 @@
 
 #include <bluevk/BlueVK.h>
 
+#include <utility>
+
 namespace filament::backend {
 
 VkFormat getVkFormat(ElementType type, bool normalized, bool integer);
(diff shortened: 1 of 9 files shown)


filament/backend/src/vulkan/VulkanUtility.h
```diff
@@ -88,10 +88,18 @@ utils::FixedCapacityVector<OutType> enumerate(
 #undef EXPAND_ENUM_NO_ARGS
 #undef EXPAND_ENUM_ARGS
 
-
 // Useful shorthands
 using VkFormatList = utils::FixedCapacityVector<VkFormat>;
 
+// Helpful fixed size array
+template<typename T, size_t capacity>
+class FillArray : public std::vector<T> {
```
z3moon, line 96: In general, we'd want to avoid inheriting from std containers because of the slicing issue. Would FixedCapacityVector work instead?

poweifeng: FixedCapacityV might work with a bit of work, but what's the slicing issue you're referring to?  I won't be slicing the data in this array.

z3moon: https://en.wikipedia.org/wiki/Object_slicing
In addition, std containers don't have virtual destructors, so it could potentially cause memory leaks for resources in subclasses.
Although I don't see both issues in your class for now, I'd rather avoid anti patterns. For example, someone may pick up your class and want to add more features on it.

poweifeng: ok, sounds good.  I've refactored an existing array with similar intent for re-use.  PTAL.  thanks.

z3moon (approved):

pixelflinger (approved):
that looks reasonable. I wonder if a one-pager style design doc would be helpful when reviewing these changes?

filament/backend/src/vulkan/caching/VulkanDescriptorSet.cpp
```diff
@@ -0,0 +1,646 @@
...
+PORT_CONSTANT(VulkanDescriptorSetManager, MAX_SUPPORTED_SHADER_STAGE)
+PORT_CONSTANT(VulkanDescriptorSetManager, VERTEX_STAGE)
+PORT_CONSTANT(VulkanDescriptorSetManager, FRAGMENT_STAGE)
+
+PORT_CONSTANT(Program, UNIFORM_BINDING_COUNT)
+PORT_CONSTANT(Program, SAMPLER_BINDING_COUNT)
+
+#undef PORT_CONSTANT
+
+uint8_t BIT_COUNT[256] = {0, 1, 1, 2, 1, 2, 2, 3, 1, 2, 2, 3, 2, 3, 3, 4, 1, 2, 2, 3, 2, 3, 3, 4, 2,
```
bejado, line 47: I wonder if we could generate this with a `constexpr` function

filament/backend/src/vulkan/caching/VulkanDescriptorSet.cpp
```diff
@@ -0,0 +1,646 @@
...
+    using Cache = std::unordered_map<Key, VulkanDescriptorSet*, HashFn, Equal>;
+};
+
+constexpr uint8_t MAX_BINDINGS = UNIFORM_BINDING_COUNT * MAX_SUPPORTED_SHADER_STAGE;
+static_assert(MAX_BINDINGS >= SAMPLER_BINDING_COUNT * MAX_SUPPORTED_SHADER_STAGE);
+
+template<typename MaskType>
+inline MaskType genMask(uint8_t stage, MaskType bit) {
+    return static_cast<MaskType>(bit << (sizeof(MaskType) * 4 * (stage - 1)));
+}
```
bejado, line 129: This logic is a bit confusing to me. Maybe document what the `4` represents and what a "mask" is?

poweifeng: This function is actually no longer necessary.  removed.

filament/backend/src/vulkan/caching/VulkanDescriptorSet.cpp
poweifeng: yes!  thanks for the suggestion.  it reads cleaner.

filament/backend/src/vulkan/VulkanUtility.h
```diff
@@ -88,10 +90,100 @@ utils::FixedCapacityVector<OutType> enumerate(
...
 #undef EXPAND_ENUM_ARGS
 
-
 // Useful shorthands
 using VkFormatList = utils::FixedCapacityVector<VkFormat>;
 
+// An Array that will be fixed capacity, but the "size" (as in user added elements) is variable.
+// Note that this class is movable.
+template<typename T, uint16_t CAPACITY>
+class CappedArray {
```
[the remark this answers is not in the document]
poweifeng: Is there a use for it outside of the vk backend?  I"ll probably save that for a follow-up (if it's necessary).

poweifeng:
> that looks reasonable. I wonder if a one-pager style design doc would be helpful when reviewing these changes?

yeah, I'll do a follow-up one-pager.

bejado (approved):

PR #7626, opened 2024-03-05
Fixed validation error VUID-vkAcquireNextImageKHR-semaphore-01779

mdagois:
The validation error triggers even on **hellotriangle** using AMD (desktop), QCOM (mobile) and Mali (mobile) GPUs.

Prior to this PR, only a single semaphore object was used to synchronize all the calls to **vkAcquireNextImage** (signal) and **vkQueueSubmit** (wait).
The issue is that by the time **vkQueueSubmit** returns, the semaphore is not necessarily reset.
When multiple frames are in flight, calls to **vkAcquireNextImage** might try to reuse the semaphore while it is still in the **wait** status, because the semaphore is reset at a driver/hardware-dependent timing that's likely linked to the GPU queue execution.

The solution proposed by this PR is to use a pool of semaphores big enough to cover all queue submissions even if they were run concurrently.

### DIFF
--- a/NEW_RELEASE_NOTES.md
+++ b/NEW_RELEASE_NOTES.md
@@ -9,3 +9,4 @@ appropriate header in [RELEASE_NOTES.md](./RELEASE_NOTES.md).
 ## Release notes for next branch cut
 
 - Add new API `SwapChain::getFrameScheduledCallback`
+- Fixed Vulkan validation error VUID-vkAcquireNextImageKHR-semaphore-01779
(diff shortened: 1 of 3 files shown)


NEW_RELEASE_NOTES.md
```diff
@@ -9,3 +9,4 @@ appropriate header in [RELEASE_NOTES.md](./RELEASE_NOTES.md).
 ## Release notes for next branch cut
 
 - Add new API `SwapChain::getFrameScheduledCallback`
+- Fixed Vulkan validation error VUID-vkAcquireNextImageKHR-semaphore-01779
```
poweifeng, line 12:
```suggestion
- vulkan: fixed validation error VUID-vkAcquireNextImageKHR-semaphore-01779
```

poweifeng (approved):
Thank you!

PR #7629, opened 2024-03-05
Add support for protected contexts

pixelflinger:
Protected contexts are now supported by the OpenGLPlatform interface and implemented in EGLPlatform.

Protected contexts can read from regular and protected resources but can only write to protected resources (e.g. protected swap chains or textures backed by protected memory. These can be created on Android via AHardwareBuffer and EGLImage for instance).

The underlaying EGL implementation must support protected contexts.

Switching to a protected context is achieved by using a protected-content SwapChain in Renderer::beginFrame(). A protected-content SwapChain can be created using the new CONFIG_PROTECTED_CONTENT flag at creation time.

The OpenGL backend implementation will then use a protected context for rendering until an unprotected SwapChain is used again.

The crux of this implementation is to use different VAOs in the different contexts, because those can't be shared between contexts. We also need to synchronize the state with our state cache and ensure VAOs objects are destructed properly in the right context.

### DIFF
--- a/filament/backend/src/opengl/OpenGLContext.h
+++ b/filament/backend/src/opengl/OpenGLContext.h
@@ -29,9 +29,11 @@
 #include "GLUtils.h"
 
 #include <array>
+#include <functional>
 #include <set>
 #include <tuple>
 #include <utility>
+#include <vector>
 
 namespace filament::backend {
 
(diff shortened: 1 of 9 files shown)


filament/backend/src/opengl/OpenGLContext.h
```diff
@@ -305,8 +319,19 @@ class OpenGLContext final : public TimerQueryFactoryInterface {
 
     FeatureLevel getFeatureLevel() const noexcept { return mFeatureLevel; }
 
+    // This is the index of the context in use. Must be 0 or 1. This is used to maange the
```
poweifeng, line 322:
```suggestion
    // This is the index of the context in use. Must be 0 or 1. This is used to manage the
```

poweifeng (approved):

filament/backend/src/opengl/OpenGLContext.h
patka00: 1

PR #7635, opened 2024-03-06
vk: clean up debug tools

poweifeng:
- Fix broken resource leak print out.
 - Add sampler name to debugUtils when enabled


filament/backend/src/vulkan/VulkanCommands.cpp
```diff
@@ -436,7 +436,7 @@ void VulkanCommands::popGroupMarker() {
         auto const endTime = std::chrono::high_resolution_clock::now();
         std::chrono::duration<double> diff = endTime - startTime;
         utils::slog.d << "<---- " << marker << " elapsed: " << (diff.count() * 1000) << " ms\n"
-                      << utils::io::flush;
+                      << utils::io::endl;
```
pixelflinger, line 439: you might want to remove the last `\n` ?

pixelflinger (approved):

PR #7637, opened 2024-03-06
Try fixing windows artifact output again

poweifeng:
The [previous] change assumed that the shell is powershell, but the shell is actually commands (cmd). 

The [previous] change assumed we're in the root directory.  This assumption is probably correct [ref]. So we keep that change.

[ref]:  https://github.com/google/filament/blob/main/build/windows/build-github.bat#L134
[previous]: 373c5710b11bec4ee27e094b255e4eb4c4ed9915


bejado (approved):
Thanks Powei!

PR #7642, opened 2024-03-08
add post-lighting mix factor support

pixelflinger:
This adds a new material property (float postLightingMixFactor) which is used to mix the original color with the post-lighting blended color. The default value is 1.0, which keeps the current behavior.

FIXES=[328498606]


libs/filamat/include/filamat/MaterialBuilder.h
```diff
@@ -410,7 +410,8 @@ class UTILS_PUBLIC MaterialBuilder : public MaterialBuilderBase {
     /**
      * Set the blending mode of the post-lighting color for this material.
      * Only OPAQUE, TRANSPARENT and ADD are supported, the default is TRANSPARENT.
-     * This setting requires the material property "postLightingColor" to be set.
+     * This setting requires the material properties "postLightingColor" and
+     * "postLightingMixFactor" to be set.
```
romainguy, line 414: Since the default is 1.0, the mix factor doesn't need to be set.

romainguy (approved):
You should also update Materials.md.html and .html

bejado (approved):

PR #7643, opened 2024-03-08
vk: add /usr/local/lib to rpath on macos

poweifeng:
For reasons unknown, after upgrading to XCode 15.3, dlopen can no longer find libvulkan.1.dylib. We fix it by explicitly adding /usr/local/lib to rpath for macos.


poweifeng:
hmm, made a mistake.  this doesn't fix the issue.

pixelflinger (approved):
thanks!

PR #7644, opened 2024-03-08
Add multiview filamat for default materials

z3moon:
Add prebuilt materials for the engine default materials. They'll be selected for multiview stereoscopic implementation.

### DIFF
--- a/libs/filamat/src/GLSLPostProcessor.cpp
+++ b/libs/filamat/src/GLSLPostProcessor.cpp
@@ -169,6 +169,7 @@ void GLSLPostProcessor::spirvToMsl(const SpirvBlob *spirv, std::string *outMsl,
     }
 
     mslOptions.argument_buffers = true;
+    mslOptions.ios_support_base_vertex_instance = true;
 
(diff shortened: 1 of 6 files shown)


libs/filamat/src/GLSLPostProcessor.cpp
```diff
@@ -169,6 +169,7 @@ void GLSLPostProcessor::spirvToMsl(const SpirvBlob *spirv, std::string *outMsl,
     }
 
     mslOptions.argument_buffers = true;
+    mslOptions.ios_support_base_vertex_instance = true;
```
bejado, line 172: Just curious, why is this required?

bejado (approved):

libs/filamat/src/GLSLPostProcessor.cpp
z3moon: Without this, the multiview shader doesn't compile for iOS, complaining `BaseInstance requires Metal 1.1 and Mac or Apple A9+ hardware` from spirv-cross.

https://github.com/KhronosGroup/SPIRV-Cross/issues/1612

PR #7653, opened 2024-03-11
assert when swapchain creation fails

pixelflinger:
This is better than risking a null-pointer dereference later.


filament/backend/src/opengl/OpenGLDriver.cpp
```diff
@@ -1418,6 +1418,8 @@ void OpenGLDriver::createSwapChainR(Handle<HwSwapChain> sch, void* nativeWindow,
 
     GLSwapChain* sc = handle_cast<GLSwapChain*>(sch);
     sc->swapChain = mPlatform.createSwapChain(nativeWindow, flags);
+    PANIC_POSTCONDITION("createSwapChain(%p, 0x%lx) failed. See logs for details.",
```
poweifeng, line 1421: should these be `ASSERT_POSTCONDITION` on `sc-swapChain`?

pixelflinger: yes! thanks!

PR #7654, opened 2024-03-12
Move SwapChain flags into separate file

bejado:


z3moon (approved):

android/filament-android/src/main/java/com/google/android/filament/SwapChainFlags.java
```diff
@@ -0,0 +1,87 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.android.filament;
+
+public final class SwapChainFlags {
```
poweifeng, line 19: I don't know how, but maybe a word about the reason for this separation?  So that we don't forget and undo this later?

poweifeng (approved):

android/filament-android/src/main/java/com/google/android/filament/SwapChainFlags.java
bejado: Yeah good idea

PR #7655, opened 2024-03-12
More Platform improvements for protected contexts 

pixelflinger:
- assert when swapchain creation fails
- handle non-0 default FBO with protected contexts
- More Platform improvements for protected contexts
  - `isSwapChainProtected()` is now virtual
  - `createDefaultSwapChain()` is renamed to `getDefaultSwapChain()`
  - new `getCurrentContextType()` returns the current context type
  - `makeCurrent` now takes an additional `ContextType`
  - `PlatformEGL::getContextForType()` to retrieve the `EGLContext` for
     a given type.
  - `PlatformEGL::makeCurrent` non-virtual utilities to set only the
    context or swapchains.

### DIFF
--- a/filament/backend/src/opengl/OpenGLContext.cpp
+++ b/filament/backend/src/opengl/OpenGLContext.cpp
@@ -49,7 +49,8 @@ bool OpenGLContext::queryOpenGLVersion(GLint* major, GLint* minor) noexcept {
 #endif
 }
 
-OpenGLContext::OpenGLContext(OpenGLPlatform& platform) noexcept {
+OpenGLContext::OpenGLContext(OpenGLPlatform& platform) noexcept
+        : mPlatform(platform) {
 
     state.vao.p = &mDefaultVAO;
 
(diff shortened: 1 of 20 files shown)


filament/backend/src/opengl/OpenGLContext.cpp
```diff
@@ -751,6 +755,50 @@ void OpenGLContext::initExtensionsGL(Extensions* ext, GLint major, GLint minor)
 
 #endif // BACKEND_OPENGL_VERSION_GL
 
+
+void OpenGLContext::bindFramebuffer(GLenum target, GLuint buffer) noexcept {
+    if (UTILS_UNLIKELY(buffer == 0)) {
+        // we're binding the default frame buffer, resolve it's actual name
```
poweifeng, line 761:
```suggestion
        // we're binding the default frame buffer, resolve its actual name
```

poweifeng (approved):

PR #7658, opened 2024-03-12
Add option FILAMENT_SAMPLES_STEREO_TYPE for samples/gltfio

z3moon:
This option can be either "instanced" or "multiview", indicating what stereoscopic rendering type shaders in samples/gltfio should be built for.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -531,6 +531,16 @@ else()
     option(FILAMENT_DISABLE_MATOPT "Disable material optimizations" ON)
 endif()
 
+# This only affects the prebuilt shader files in gltfio and samples, not filament library.
+# The value can be either "instanced" or "multiview".
+set(FILAMENT_SAMPLES_STEREO_TYPE "instanced" CACHE STRING
+    "Stereoscopic type that shader files in gltfio and samples are built for."
+)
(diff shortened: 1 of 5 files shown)


CMakeLists.txt
```diff
@@ -531,6 +531,16 @@ else()
     option(FILAMENT_DISABLE_MATOPT "Disable material optimizations" ON)
 endif()
 
+# This only affects the prebuilt shader files in gltfio and samples, not filament library.
+# The value can be either "instanced" or "multiview".
+set(FILAMENT_SAMPLES_STEREO_TYPE "instanced" CACHE STRING
```
bejado, line 536: I think we would be okay with including both types of stereo materials in our samples, since for those we're not concerned about binary size. We could set the stereo type with a runtime command-line flag or an environment variable. That way we could more easily compare the two.

Just something to think about, no action required.

bejado (approved):

CMakeLists.txt
z3moon: Yeah good point. Can we have gltfio(UberShaderProvider) include both materials as well? gltf_viewer depends on it, so gltfio should have both to facilitate the commandline approach.

bejado: I think that would be okay. Maybe we can compare the binary size between a gltfio library with/without both stereo types. G3 doesn't use gltfio, so it would just be a potential concern be for us and 3P clients.

PR #7660, opened 2024-03-12
Add layerCount to createRenderTarget

z3moon:
This new parameter indicates whether the render target will be created for multiview.

If the value is greater than 1, it tells the render target should be created for multiview. Otherwise, 1 or 0, it creates a single layer render target.

### DIFF
--- a/filament/backend/include/backend/TargetBufferInfo.h
+++ b/filament/backend/include/backend/TargetBufferInfo.h
@@ -83,8 +83,8 @@ public:
     }
 
     // this is here for backward compatibility
-    MRT(Handle<HwTexture> handle, uint8_t layerCount, uint8_t level, uint16_t layer) noexcept
-            : mInfos{{ handle, layerCount, level, layer }} {
+    MRT(Handle<HwTexture> handle, uint8_t level, uint16_t layer) noexcept
+            : mInfos{{ handle, 0, level, layer }} {
     }
 };
 
--- a/filament/backend/include/backend/TargetBufferInfo.h
+++ b/filament/backend/include/backend/TargetBufferInfo.h
@@ -32,6 +32,10 @@ struct TargetBufferInfo {
     // texture to be used as render target
     Handle<HwTexture> handle;
 
+    // starting layer index for multiview. This value is only used when the `layerCount` for the
+    // render target is greater than 1.
+    uint8_t baseViewIndex = 0;
+
     // level to be used
     uint8_t level = 0;
 
@@ -79,8 +83,8 @@ public:
     }
 
     // this is here for backward compatibility
-    MRT(Handle<HwTexture> handle, uint8_t level, uint16_t layer) noexcept
-            : mInfos{{ handle, level, layer }} {
+    MRT(Handle<HwTexture> handle, uint8_t layerCount, uint8_t level, uint16_t layer) noexcept
+            : mInfos{{ handle, layerCount, level, layer }} {
     }
 };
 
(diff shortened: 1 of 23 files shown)


filament/src/RendererUtils.cpp
```diff
@@ -150,7 +151,17 @@ FrameGraphId<FrameGraphTexture> RendererUtils::colorPass(
                 data.depth = builder.read(data.depth, FrameGraphTexture::Usage::DEPTH_ATTACHMENT);
 
                 data.color = builder.write(data.color, FrameGraphTexture::Usage::COLOR_ATTACHMENT);
-                data.depth = builder.write(data.depth, FrameGraphTexture::Usage::DEPTH_ATTACHMENT);
+                if (engine.getConfig().stereoscopicType == StereoscopicType::MULTIVIEW) {
+                    // Add sampleable usage flag for depth in multiview rendering, othewise it's
+                    // treated renderbuffer in the backend and crashed.
+                    data.depth = builder.write(data.depth,
+                        FrameGraphTexture::Usage::DEPTH_ATTACHMENT |
+                        FrameGraphTexture::Usage::SAMPLEABLE);
```
pixelflinger, line 159: this is fine for now -- but I think this is needed because of a bug on OpenGLDriver.cpp.  Can you please file a bug for me?  The issue is that if a texture has more than one layer, we shouldn't use a renderbuffer even if it doesn't have the sampleable usage flag.

filament/src/RendererUtils.cpp
```diff
@@ -150,7 +151,17 @@ FrameGraphId<FrameGraphTexture> RendererUtils::colorPass(
...
+                    layerCount = engine.getConfig().stereoscopicEyeCount;
```
pixelflinger, line 160: I think here we need an `assert_invariant()` that data.depth and data.color have at least `layerCount` layers.

Also on line 115 `data.depth` should be created with `.depth` of `layerCount`. And `colorBufferDesc` must also have `layerCount` `.depth`.


I *think* if post-processing is disabled, this will work anyways because `data.depth` and `data.color` will be "magically" replaced by the "real" swapchain, which, in theory should be set-up properly.

I am okay with making this assumption for now, but we absolutely need a giant comment/TODO/FIXME here.

pixelflinger (changes requested):

filament/backend/include/backend/TargetBufferInfo.h
```diff
@@ -79,8 +83,8 @@ class MRT {
     }
 
     // this is here for backward compatibility
-    MRT(Handle<HwTexture> handle, uint8_t level, uint16_t layer) noexcept
-            : mInfos{{ handle, level, layer }} {
+    MRT(Handle<HwTexture> handle, uint8_t layerCount, uint8_t level, uint16_t layer) noexcept
+            : mInfos{{ handle, layerCount, level, layer }} {
```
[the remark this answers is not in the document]
z3moon: Ah you're right. Let me remove it, so that all test code can be reverted.

filament/src/RendererUtils.cpp
z3moon: Yeah, I created a bug for this.  b/329491941

filament/src/RendererUtils.cpp
z3moon: Yeah as we said offline, the rest (setting the number of depth, etc) is in workspace. I'll soon make a following PR that handles what you said here.

pixelflinger (approved):

PR #7666, opened 2024-03-13
gltfio: add extended tangents job

poweifeng:
- TangentsJobExtended extracts data from cgltf accessor and runs geometry::TangentSpaceMesh on the attributes and computes the tangent space.
 - The /extended folder is meant for running this process. Note that this API might remesh the input and will require corresponding changes that might break previous assumptions.
 - The general flow of the code is modeled after src/TangentsJob.h
 - This is not hooked into current code and should have no practical effect on gltfio.

### DIFF
--- a/libs/gltfio/src/extended/TangentSpaceMeshWrapper.cpp
+++ b/libs/gltfio/src/extended/TangentSpaceMeshWrapper.cpp
@@ -1,0 +1,330 @@
+/*
+ * Copyright (C) 2024 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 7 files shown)


libs/gltfio/src/extended/TangentSpaceMeshWrapper.cpp
```diff
@@ -0,0 +1,330 @@
...
+    } while (0)
+
+#define DO_MESH_RET_IMPL(METHOD)                                                                   \
+    do {                                                                                           \
+        if (mTsMesh) {                                                                             \
+            return mTsMesh->METHOD();                                                              \
+        } else {                                                                                   \
+            return mPassthroughMesh->METHOD();                                                     \
+        }                                                                                          \
+    } while (0)
```
pixelflinger, line 137: is this to avoid using the interface? just curious.

pixelflinger (approved):

libs/gltfio/src/extended/TangentSpaceMeshWrapper.cpp
poweifeng: Yeah, just to avoid interfaces and virtual calls. But maybe it's ok here since it's on being called every frame or anything like that.

PR #7671, opened 2024-03-15
Connect multiview components together

z3moon:
Plumb through multiview configurations to the pipeline so that the engine draws scenes using multiview extension. Users need to prepare shaders compiled with the `multiview` param and set the `stereoscopicType` flag to MULTIVIEW in the Engine::Config to enable multiview feature.

In this change, postprocessings for multiview are not yet supported. So we all disable them until they're supported.

The debug option `combineMultiviewImages` combines layers as one image, which allows us to check the final result.


filament/src/RenderPass.cpp
```diff
@@ -591,9 +593,16 @@ RenderPass::Command* RenderPass::generateCommandsImpl(RenderPass::CommandTypeFla
         // manual or hybrid instancing. Instanced stereo multiplies the number of instances by the
         // eye count.
         if (UTILS_UNLIKELY(hasStereo)) {
-            cmdColor.primitive.instanceCount =
-                    (soaInstanceInfo[i].count * stereoEyeCount) |
-                    PrimitiveInfo::USER_INSTANCE_MASK;
+            switch (stereoscopicType) {
```
pixelflinger, line 596: instead of passing `stereoscopicType`, just change the `hasStereo` to `hasInstancedStereo` and rename the `IS_STEREOSCOPIC` flag to `IS_INSTANCED_STEREOSCOPY`. This way we can remove the extra test and simplify this change a lot.

android/filament-android/src/main/cpp/View.cpp
```diff
@@ -483,10 +483,11 @@ Java_com_google_android_filament_View_nIsStencilBufferEnabled(JNIEnv *, jclass,
 extern "C"
 JNIEXPORT void JNICALL
 Java_com_google_android_filament_View_nSetStereoscopicOptions(JNIEnv *, jclass, jlong nativeView,
-        jboolean enabled) {
+        jboolean enabled, jboolean combineMultiviewImages) {
     View* view = (View*) nativeView;
     View::StereoscopicOptions options {
-        .enabled = (bool) enabled
+        .enabled = (bool) enabled,
+        .combineMultiviewImages = (bool) combineMultiviewImages,
```
pixelflinger, line 490: instead of using a boolean here, use the Engine's debugging properties facility:

In `Engine.h` add a field to the `debug` structure:

```
struct {
  (...)
  struct {
     bool combine_multiview_images = false; // or true if you want it to be the defulat
  } stereo;
} debug;
```

Then in `Renderer.cpp` at the top add something like:

```
    debugRegistry.registerProperty("d.stereo.combine_multiview_images",
            &engine.debug.stereo.combine_multiview_images);
```
And finally, you can just test against `engine.debug.stereo.combine_multiview_images`.


An app can set this property like this:
```
auto& debug = engine->getDebugRegistry();
bool* combineMultiViewEnabled = debug.getPropertyAddress<bool>("d.stereo.combine_multiview_images");
*combineMultiViewEnabled = true;
```

pixelflinger (changes requested):

filament/src/RenderPass.cpp
z3moon: Makes sense. Let me update it that way.

android/filament-android/src/main/cpp/View.cpp
z3moon: Great! let me apply this. Thanks!

samples/gltf_viewer.cpp
```diff
@@ -893,6 +893,8 @@ int main(int argc, char** argv) {
                                     "d.shadowmap.display_shadow_texture_channel"), 0, 3);
                     ImGui::Unindent();
                 }
+                ImGui::Checkbox("Combine Multiview Images",
```
poweifeng, line 896: What happens when this is unchecked?  Is it just a black screen?

Maybe the flow could be that if multiview is enabled, then this is always on for gltf_viewer. wdyt?

z3moon: It just displays the left view, not crashing.

The flag is enabled by default for now, so gltf_viewer shows side by side image. However, the default will be set to false once they can take multi-layer swapchain.

z3moon: Also I think it's still good to let users switch this flag because they can realize the final image is combined by the debug flag, not a single image originally.

I think this option should be visible only when the multiview is enabled. Let me update it.

filament/src/RenderPass.cpp
```diff
@@ -590,10 +589,10 @@ RenderPass::Command* RenderPass::generateCommandsImpl(RenderPass::CommandTypeFla
         // soaInstanceInfo[i].count is the number of instances the user has requested, either for
         // manual or hybrid instancing. Instanced stereo multiplies the number of instances by the
         // eye count.
-        if (UTILS_UNLIKELY(hasStereo)) {
+        if (UTILS_UNLIKELY(hasInstancedStereo)) {
             cmdColor.primitive.instanceCount =
-                    (soaInstanceInfo[i].count * stereoEyeCount) |
-                    PrimitiveInfo::USER_INSTANCE_MASK;
+                (soaInstanceInfo[i].count * stereoEyeCount) |
+                PrimitiveInfo::USER_INSTANCE_MASK;
```
pixelflinger, line 595: nit: we use a continuation indent of 8 spaces

filament/src/RenderPass.cpp
```diff
@@ -624,10 +623,10 @@ RenderPass::Command* RenderPass::generateCommandsImpl(RenderPass::CommandTypeFla
             cmdDepth.primitive.skinningTexture = skinning.handleSampler;
             cmdDepth.primitive.morphWeightBuffer = morphing.handle;
 
-            if (UTILS_UNLIKELY(hasStereo)) {
+            if (UTILS_UNLIKELY(hasInstancedStereo)) {
                 cmdColor.primitive.instanceCount =
-                        (soaInstanceInfo[i].count * stereoEyeCount) |
-                        PrimitiveInfo::USER_INSTANCE_MASK;
+                    (soaInstanceInfo[i].count * stereoEyeCount) |
+                    PrimitiveInfo::USER_INSTANCE_MASK;
```
pixelflinger, line 629: continuation is 8 spaces

filament/src/details/Renderer.cpp
```diff
@@ -700,7 +711,9 @@ void FRenderer::renderJob(RootArenaScope& rootArenaScope, FView& view) {
     RenderPass::RenderFlags renderFlags = 0;
     if (view.hasShadowing())                renderFlags |= RenderPass::HAS_SHADOWING;
     if (view.isFrontFaceWindingInverted())  renderFlags |= RenderPass::HAS_INVERSE_FRONT_FACES;
-    if (view.hasStereo())                   renderFlags |= RenderPass::IS_STEREOSCOPIC;
+    if (view.hasStereo() &&
+        engine.getConfig().stereoscopicType == backend::StereoscopicType::INSTANCED)
+        renderFlags |= RenderPass::IS_INSTANCED_STEREOSCOPIC;
```
pixelflinger, line 716:
```suggestion
    if (view.hasStereo() &&
            engine.getConfig().stereoscopicType == backend::StereoscopicType::INSTANCED) {
        renderFlags |= RenderPass::IS_INSTANCED_STEREOSCOPIC;
    }
```

filament/src/details/Renderer.cpp
```diff
@@ -1152,6 +1171,15 @@ void FRenderer::renderJob(RootArenaScope& rootArenaScope, FView& view) {
         }
     }
 
+    // Debug: combine the array texture for multiview into a single image.
+    if (isRenderingMultiview &&
+        engine.debug.stereo.combine_multiview_images) {
```
pixelflinger, line 1176:
```suggestion
    if (UTILS_UNLIKELY(isRenderingMultiview &&
        engine.debug.stereo.combine_multiview_images)) {
```

pixelflinger (changes requested):

filament/src/RenderPass.cpp
z3moon: Done

filament/src/RenderPass.cpp
z3moon: Done

pixelflinger (approved):
nice!

PR #7675, opened 2024-03-15
Allow rendering thread to pause

elizagamedev:
This PR adds a new `pause()` option to the `Engine` `Builder` and a new function `setPaused()` to the `Engine`. While paused, the rendering thread will pause indefinitely for commands as if none are available. As soon as the rendering thread is unpaused, the commands are immediately executed.

### DIFF
--- a/android/filament-android/src/main/java/com/google/android/filament/Engine.java
+++ b/android/filament-android/src/main/java/com/google/android/filament/Engine.java
@@ -240,6 +240,18 @@ public class Engine {
         }
 
         /**
+         * Sets the initial paused state of the rendering thread.
+         *
+         * @param paused Whether to start the rendering thread paused.
+         * @return A reference to this Builder for chaining calls.
+         * @warning Experimental.
+         */
+        public Builder paused(boolean paused) {
+            nSetBuilderPaused(mNativeBuilder, paused);
+            return this;
+        }
+
+        /**
          * Creates an instance of Engine
          *
          * @return A newly created <code>Engine</code>, or <code>null</code> if the GPU driver couldn't
@@ -1189,6 +1201,13 @@ public class Engine {
         nFlush(getNativeObject());
     }
 
+    /**
+     * Pause or resume the rendering thread.
+     * @warning Experimental.
+     */
+    public void setPaused(boolean paused) {
+        nSetPaused(getNativeObject(), paused);
+    }
 
     @UsedByReflection("TextureHelper.java")
     public long getNativeObject() {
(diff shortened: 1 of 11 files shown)


filament/backend/src/CommandBufferQueue.cpp
```diff
@@ -56,6 +57,15 @@ void CommandBufferQueue::requestExit() {
...
 }
 
+void CommandBufferQueue::setPaused(bool paused) {
+    if (paused) {
+        mPaused = true;
+    } else {
+        mPaused = false;
+        mCondition.notify_one();
+    }
+}
```
pixelflinger, line 67: You need to take the lock while updating mPaused otherwise nothing guarantees that the change will be "seen" in the other thread (e.g. because of the caches).

filament/backend/src/CommandBufferQueue.cpp
```diff
@@ -127,7 +137,7 @@ std::vector<CommandBufferQueue::Range> CommandBufferQueue::waitForCommands() con
         return std::move(mCommandBuffersToExecute);
     }
     std::unique_lock<utils::Mutex> lock(mLock);
-    while (mCommandBuffersToExecute.empty() && !mExitRequested) {
+    while ((mCommandBuffersToExecute.empty() || mPaused) && !mExitRequested) {
```
pixelflinger, line 140: maybe make paused `false` by default, which I think is the common case.

pixelflinger (changes requested):
Nice solution!

android/filament-android/src/main/java/com/google/android/filament/Engine.java
```diff
@@ -239,6 +239,18 @@ public Builder featureLevel(FeatureLevel featureLevel) {
             return this;
         }
 
+        /**
+         * Sets the initial paused state of the rendering thread.
```
pixelflinger, line 243: Here we need to document the caveats of this API:
- "buffer" callbacks won't be called until unpaused. So, in particular, if the user is relying on the callbacks to unpause, that would create a deadlock that would be undetectable by us.
- if the user is calling too many APIs that fill up the queue, it will eventually block and since we're not draining it, this will also deadlock. But we can detect that one and assert.

pixelflinger (changes requested):
We need to document the caveats and detect cases that will cause a deadlock. See inline.

elizagamedev:
Okay, everything is addressed properly, I think.

android/filament-android/src/main/java/com/google/android/filament/Engine.java
```diff
@@ -1189,6 +1203,22 @@ public void flush() {
...
+    /**
+     * Pause or resume the rendering thread.
+     *
+     * <p>Warning: This is an experimental API. In particular, note the following caveats.
+     *
+     * <ul><li>
+     * Buffer callbacks will never be called as long as the rendering thread is paused.
+     * Do not rely on a buffer callback to unpause the thread.
+     * </li><li>
+     * While the rendering thread is paused, rendering commands will continued to be queued
```
pixelflinger, line 1215:
```suggestion
     * While the rendering thread is paused, rendering commands will continue to be queued
```

pixelflinger (approved):

PR #7680, opened 2024-03-19
vk: fix broken gltf renderings

### LINKED ISSUE #7674: Some GLTF files do not render on Vulkan
**Describe the bug**
Some GLTF and especially GLB model files cause intense glitches when rendered with Filament on the Vulkan backend.
This happens on a variety of GLB/GLTF models (about 25% of models tested).

**To Reproduce**
1. Download this example GLB by James Phan: [Sketchfab Link](https://sketchfab.com/3d-models/subway-car-abc21eee368448a4884376ef1fc444fa#download)
2. Load the model into filament's GLTF viewer. The HDRI is irrelevant, but in my testing I used [this free HDRI from PolyHaven](https://polyhaven.com/a/industrial_sunset_puresky).

**Expected behavior**
The model should render correctly, as it does on the OpenGL backend.

**Screenshots**
Here is the visual output on Vulkan:
image

For reference, here is the exact same GLB loaded up in OpenGL:
image

**Logs**
The logs do not show anything extraordinary.
```
FEngine (64 bits) created at 000001C3D410F9F0 (threading is enabled)
Vulkan device driver: NVIDIA 551.61
Selected physical device 'NVIDIA GeForce RTX 4080' from 1 physical devices. (vendor 0x10de, device 0x2704, driver 0x89cf4000, api 1.3)
Backend feature level: 1
FEngine feature level: 1
vkCreateSwapchain: 1024x640, 44, 0, swapchain-size=3, identity-transform=true, depth=130
vkCreateSwapchain: 1920x1017, 44, 0, swapchain-size=3, identity-transform=true, depth=130
```

 - OS: Windows 10 Pro 22H2
 - GPU: NVIDIA RTX 4080 OC
 - Backend: Vulkan
 
 **Extra notes**
 According to the Khronos GLTF validator, the model file is indeed a valid GLTF/GLB file without errors.

poweifeng:
This is due to color attachments being set to store=discard when they are multi-sampled. It's unclear why that condition exists. For now, removing it will fix the rendering issues with transparent object + MSAA. We'll keep it as such until an issue surfaces.

Fixes #7674

### DIFF
--- a/filament/backend/src/vulkan/VulkanFboCache.cpp
+++ b/filament/backend/src/vulkan/VulkanFboCache.cpp
@@ -239,7 +239,7 @@ VkRenderPass VulkanFboCache::getRenderPass(RenderPassKey config) noexcept {
             .format = config.colorFormat[i],
             .samples = (VkSampleCountFlagBits) config.samples,
             .loadOp = clear ? kClear : (discard ? kDontCare : kKeep),
-            .storeOp = config.samples == 1 ? kEnableStore : kDisableStore,
+            .storeOp = kEnableStore,
             .stencilLoadOp = kDontCare,
             .stencilStoreOp = kDisableStore,
             .initialLayout = ((!discard && config.initialColorLayoutMask & (1 << i)) || clear)


bejado (approved):
At least with Metal, for MSAA passes we use "memoryless" textures (which don't store their contents). These then get resolved to a "real" non-MSAA texture. Maybe Vulkan was doing the same?

poweifeng:
> At least with Metal, for MSAA passes we use "memoryless" textures (which don't store their contents). These then get resolved to a "real" non-MSAA texture. Maybe Vulkan was doing the same?

Thanks for the useful background.  I think that was the intention here, but wasn't implemented completed (as per [ARM's example]).  I'll file an issue for tracking implementing it correctly. 


[ARM's example]: https://arm-software.github.io/vulkan-sdk/multisampling.html

z3moon (approved):

PR #7685, opened 2024-03-19
Improve Skybox eyeDirection precision

### LINKED ISSUE #7151: skybox is different performance in debug and release environments.
**Describe the bug**
in demo gltf-viewer(filament 1.41.0)

choose: Metal Debug:
IMG_0466

choose: Metal Release:
IMG_0467

How can I get the same result？

bejado:
Fixes #7151

See my comment here: https://github.com/google/filament/issues/7151#issuecomment-1871432263

### DIFF
--- a/filament/src/materials/skybox.mat
+++ b/filament/src/materials/skybox.mat
@@ -58,6 +58,14 @@ fragment {
 
 vertex {
     void materialVertex(inout MaterialVertexInputs material) {
-        material.eyeDirection.xyz = material.worldPosition.xyz;
+        vec4 p = getPosition();
+        // GL convention to inverted DX convention
+        p.z = p.z * -0.5 + 0.5;
+        vec4 worldPosition = getWorldFromClipMatrix() * p;
+        // Getting the true world position would require dividing by w, but since this is a skybox
+        // at inifinity, this results in very large numbers for material.eyeDirection.
+        // Since the eyeDirection is only used as a direction vector in the fragment shader, we can
+        // skip that step to improve precision.
+        material.eyeDirection.xyz = worldPosition.xyz;
     }
 }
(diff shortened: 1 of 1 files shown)


filament/src/materials/skybox.mat
```diff
@@ -58,6 +58,14 @@ fragment {
...
-        material.eyeDirection.xyz = material.worldPosition.xyz;
+        vec4 p = getPosition();
+        // GL convention to inverted DX convention
+        p.z = p.z * -0.5 + 0.5;
+        vec4 worldPosition = getWorldFromClipMatrix() * p;
+        // Getting the true world position would require dividing by w, but since this is a skybox
+        // at inifinity, this results in very large numbers for material.eyeDirection.
+        // Since the eyeDirection is only used as a direction vector in the fragment shader, we can
+        // skip that step to improve precision.
+        material.eyeDirection.xyz = worldPosition.xyz;
```
pixelflinger, line 69: I'm even wondering if we shouldn't `normalize()` here instead of in the fragment shader (when the sun disk is enabled) -- that would make it even clearer that the w-divide is not needed.

Can you add a comment that this code is taken from `computeWorldPosition` and assumes `VERTEX_DOMAIN_DEVICE`.

pixelflinger (changes requested):

filament/src/materials/skybox.mat
bejado: We can't `normalize()` here because `eyeDirection` still needs to be interpolated. See the difference:

https://github.com/google/filament/assets/5298046/1b9a573a-aeb7-4eaf-b06f-078a277268d5

pixelflinger (approved):

Issue #7683, opened 2024-03-19
Get parameter value of material from gltfio


hannojg:
**Is your feature request related to a problem? Please describe.**

(I am kinda new to filament, so not sure if I am missing anything.)

Right now I am rendering from glb files using filament. That means I don't control the materials.
I now want to make changes to the materials. For example a material's blend mode might be "opaque" and it has no transparency, but I want to change it to something else (e.g. to support transparency).
However, I want to make that change "at runtime" when the material instance has already been created.
I feel like with the current API thats not possible? I can only change parameters of a material that are already part of the material.

**Describe the solution you'd like**

My use case really is to add "transparency support for entities". Meaning I want to control the alpha value of a whole entity (for which I need to control the transparency alpha on the materials if I understand correctly). 

**Describe alternatives you've considered**

If you control the material creation it seems possible, but if you don't it looks like you don't really have any options?

Note: I feel like it has been indirectly mentioned before in this issue: https://github.com/google/filament/issues/1865

**OS and backend**

Not OS or backend specific.

hannojg:
After looking into this more detailed, and for the specific use case of controlling the opacity I found this discussion:

- https://github.com/google/filament/discussions/4504

It seems like we'd be able to control the opacity / alpha of an material by calling e.g.

```
materialInstance->setParameter("baseColorFactor", math::float4({1,1,1, 0.4}));
```

The only issue is that we would also override the rgb values.

So a solution would be to get the value of a parameter as well, which seems to be missing right now in gltfio.

hannojg:
When you say "material needs to be marked transparent", you mean that it needs to have a blend mode of `transparent` or `fade`, right?

It seems like the blending mode can't be changed at runtime, which is understandable. However it would be nice if we could set the blending mode during asset / material instantiation from the public API.
You wrote that it should be possible here:

- https://github.com/google/filament/issues/5736#issuecomment-1166242445

But I haven't found a way to make that work (given that I only import a glb file and don't control the material itself)

hannojg:
I just opened a proposal draft PR for getting the parameters from a `MaterialInstance`:

- https://github.com/google/filament/pull/7686

Let me know what you think!

pixelflinger:
I've seen the PR we need more time to get to it. Please bare with us! thanks!

Issue #5736, opened 2022-06-25
Alpha Value Not Working Properly In Scenefrom GLB file


arun4fms:
am using sceneform with modelrenderable where GLB is loaded from an external URL. I tried to change the opacity of the GLB with the given below method, but there is no change in alpha or opacity.

private void addHighlightToNode(Node node, ModelRenderable andyRenderablee) {
        MaterialInstance[] materialInstances = filamentAsset.getMaterialInstances();

        TextureSampler textureSampler = new TextureSampler();
        for(MaterialInstance materialInstance: materialInstances){
                    materialInstance.setParameter("baseColorFactor", Colors.RgbaType.SRGB, 0.0f,0.0f,0.0f,0.0f);
       }

arun4fms:
@romainguy  Please do have a look into this. I just saw your few comments regarding blending mode but for materialInstances I haven't seen any option to set Bleningmode

romainguy:
You need your gltf material to be defined as alpha blended. You can't override this at runtime after loading the gltf. You can however provide an override during loading.

arun4fms:
@romainguy  how can i provide an override during loading

Issue #1865, opened 2019-11-11
I'm giving up with Google Filament for the web


Antonio-Sorrentini:
I'm giving up with Google Filament for the web.
I wanted to create A 3D online editor but after almost 3 weeks of tries I understand now that Filament is not made for this kind of uses. The javascript interfaces only offers "set" methods, which in itself would not be a showstopper but it becomes when you load things with gltf. Yes entities are there and you can add them to the scene, but it's almost just that, you cannot get the geometries and change them, you cannot get the materials and change them. Yes set methods are there, but not knowing the initial values is like being blind, and not only. You cannot choose which ubershader gltfio uses to import materials which means that if this ubershader does not offer you (as in fact it does not) the ability to customize parameters like reflectance or clearCoat (which in my opinion together with automatic shadows are the real strengths of Filament when compared to threejs, babylonjs, etc.) you cannot change those values even if you wanted to do that blindly without knowing the initial values. One could argue that the gltf ubershader does not handle those other paramenters because they are not communicable with gltft anyway, but that is not a good reason, of course after the import you could make changes if the ubershader allowed them. On the other side one could say that nothing prevents you from building the materials on your own and apply those to the imported entity. Yes it is true, but it's like rewriting the gltf importer on your own and at that point one wonders if it's easier doing that or building reflections and shadows on your own in other engines like threejs and the like and go with those other engine. I know it's not even remotely the same, otherwise I would have not started with Filament to begin with, but for now I'm forced to fallback to this other path, I need to have a prototype of what I'm trying to achieve as soon as possible, this is mandatory for me.
There is also a bug somewhere in Filament which sometimes causes the browser to consume 100% GPU (not CPU) forever even if nothing at all happens in the scene. I've tried all my best to make this reproducible but had no success, it seems to happens really randomly, and in the end I can't be sure if it's a Filament bug or a chrome one.
Anyway I like a lot Filament, and I'll continue to follow the project. I really hope that in the future I can switch back to it. In the meantime I'm strongly grateful for all the very quick answers I got from the people involved in this project. This has been very appreciated and gives me the confidence that the project will always move in the right directions.
Good luck and see you soon.

romainguy:
Thank you for your feedback! I'm sorry to hear that it didn't work out for you. Just so that we understand where the limitations are, your main issue is really related to the `gltfio` library right? It is true that this library has been mostly designed — so far — to load glTF files and not edit them, which explains the lack of "getters" you mentioned.

[...]

PR #7686, opened 2024-03-20
feat: support getParameter from MaterialInstance

hannojg:
It has been a common user request to add the ability to read values from a `MaterialInstance`:

- https://github.com/google/filament/discussions/4504#discussioncomment-1208080
- https://github.com/google/filament/issues/1865
- https://github.com/google/filament/issues/7683#issuecomment-2009215476
- https://github.com/google/filament/issues/5736

This PR is an attempt to add support for this. It works for my use case, where I want to change the alpha of a material (to control an entities opacity), e.g.:

```cpp
math::float4 rgba = materialInstance->getParameter<math::float4>("baseColorFactor");
float newAlpha = 0.5;
materialInstance->setParameter("baseColorFactor", math::float4({rgba.r, rgba.g, rgba.b, newAlpha}));
```

**Changed**

I basically added a `getParameter` function that you can pass a type to.

-> **Note:** I am probably missing use cases. Please let me know where I am missing something or need to watch out for. Also happy to add tests if necessary 😊 !

Fixes: https://github.com/google/filament/issues/7683


filament/src/MaterialInstance.cpp
```diff
@@ -183,6 +183,34 @@ template UTILS_PUBLIC void MaterialInstance::setParameter<mat4f>   (const char*
 
 // ------------------------------------------------------------------------------------------------
 
+template<typename T>
+T FMaterialInstance::getParameterImpl(std::string_view name) const noexcept {
+    ssize_t offset = mMaterial->getUniformInterfaceBlock().getFieldOffset(name, 0);
+    return downcast(this)->getUniformBuffer().getUniform<T>(offset);
```
hannojg, line 189: I think we might want to add a check here if the offset value is correct? 
What error would we throw in case of an invalid offset?

pixelflinger: `getFieldInfo()` already validates that `name` exists and throws an exception if it doesn't. So you need to make this function not `noexcept`. I think you can just add an `assert_invariant(offset>=0)` because I don't think it's possible to get there with a bad offset.

filament/include/filament/MaterialInstance.h
```diff
@@ -234,6 +234,28 @@ class UTILS_PUBLIC MaterialInstance : public FilamentAPI {
         setParameter(name, strlen(name), type, color);
     }
 
+    /**
+     * Gets the value of a parameter by name.
+     * 
+     * @param name          Name of the parameter as defined by Material. Cannot be nullptr.
+     * @param nameLength    Length in `char` of the name parameter.
+     * @throws utils::PreConditionPanic if name doesn't exist or no-op if exceptions are disabled.
+     */
```
pixelflinger, line 243: you could also add one can check the paramter exists with `Material::hasParameter` and add a `@see` tag.

pixelflinger (changes requested):

filament/include/filament/MaterialInstance.h
hannojg: Sure! Added the `@see` tag , but what do you mean by "can check"?

`getFieldInfo(name)` already has a `ASSERT_PRECONDITION` that checks if the info with the given name was found, so I thought that'd be sufficient.

Where would you like me to add the `hasParameter` check?

filament/src/MaterialInstance.cpp
hannojg: added!

filament/include/filament/MaterialInstance.h
pixelflinger: I just meant we can add in the documentation that it's possible to check the that parameter exists using `hasParameter`. For some applications, it's not acceptable to fail, so they can check up-front.

pixelflinger (approved):

filament/src/MaterialInstance.cpp
```diff
@@ -183,6 +183,35 @@ template UTILS_PUBLIC void MaterialInstance::setParameter<mat4f>   (const char*
...
+}
+
+template<typename T>
+T MaterialInstance::getParameter(const char* name, size_t nameLength) const {
+    return downcast(this)->getParameterImpl<T>({ name, nameLength });
+}
+
+// explicit template instantiation of our supported types
+template UTILS_PUBLIC float MaterialInstance::getParameter<float>        (const char* name, size_t nameLength) const;
```
poweifeng, line 199: nit: why all the spaces here?

poweifeng (commented):
What happens when you try to get a texture.  Or when you try to get a color, it doesn't return whether it's Linear or sRGB?  If those aren't supported, maybe we need to be explicit what is supported and what is not?

hannojg:
@poweifeng you're right they are currently not supported.
I will look into either adding support for those, or clearly documentation what's supported and what's not.
Either on the weekend or early next week!

filament/src/MaterialInstance.cpp
hannojg: Thanks for catching that, thats definitely not on purpose 😅 will clean that up soon!

pixelflinger:
> @poweifeng you're right they are currently not supported yet. I will look into either adding support for those, or clearly documentation what's supported and what's not. Either on the weekend or early next week!

I think it's fine to document it only works for  non-texture parameters for now.  We can always have another PR later that adds support for that.

pixelflinger:
@hannojg would you be able to rebase/update this branch? thanks!

hannojg:
Sorry for letting this go stale!
Updated it + added the documentation comments.

bejado (approved):

poweifeng (approved):

PR #7689, opened 2024-03-20
attempt to detect buffer overflows in Texture::setImage()

pixelflinger:
We verify that the buffer given to setImage() is at least as large as needed for the given region to transfer; at least based on the size given.

This might help catch b/330407429.


poweifeng:
> We verify that the buffer given to setImage() is at least as large as needed for the given region to transfer; at least based on the size given.
> 
> This might help catch b/330407429.

You might want to add in the commit message
```BUGS=330407429```
so that it'll post to the bug (but not close it).

poweifeng (approved):

PR #7696, opened 2024-03-21
Replace value of num_views with engine's eye count

z3moon:
For OpenGL multiview, it honors the qualifier `layout(num_views = X)` specified in shader files to determine the number of views for multiview.

We cannot recompile materials everytime the value changes. So replace the value of num_views with the engine's eye count when shaders compile.

### DIFF
--- a/filament/backend/src/opengl/ShaderCompilerService.cpp
+++ b/filament/backend/src/opengl/ShaderCompilerService.cpp
@@ -30,6 +30,7 @@
 #include <utils/Log.h>
 #include <utils/Systrace.h>
 
+#include <cctype>
 #include <chrono>
 #include <string>
 #include <string_view>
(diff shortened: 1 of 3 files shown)


poweifeng (commented):
Why do this in materialparser as oppose to in the GL backend (like a spec constant replacement)? This will have impact the performance for even non-GL backends.

z3moon:
> Why do this in materialparser as oppose to in the GL backend (like a spec constant replacement)? This will have impact the performance for even non-GL backends.

Actually I'm amending this PR as this is not the right place to modify as you mentioned. Let me update this as draft until it's ready.

filament/backend/src/opengl/ShaderCompilerService.cpp
```diff
@@ -597,6 +612,51 @@ std::string_view ShaderCompilerService::process_GOOGLE_cpp_style_line_directive(
     return { source, len };
 }
 
+// Look up the `source` to replace the number of eyes for multiview with the given
+// number. This is necessary for OpenGL because OpenGL relies on the number
+// specified in shader files to determine the number of views for multiview.
+// This funciton assumes the number of views is a single digit.
+std::string_view ShaderCompilerService::process_OVR_multiview2(OpenGLContext& context,
+        char* source, size_t len, int32_t eyeCount) noexcept {
+    if (context.ext.OVR_multiview2) {
```
pixelflinger, line 621: it looks like this will run pretty much all the time. Is there a way we could skip this for the common case, which is when stereo is not used?

pixelflinger (approved):

filament/backend/src/opengl/ShaderCompilerService.cpp
z3moon: It's possible If we store a metadata indicating STE or multiview in Program class. Variant is in filabridge, so it seems not preferable to store it in Program class (filament/backend) as we discussed yesterday. However, we could use a boolean or more generic variable instead.

PR #7706, opened 2024-03-25
Improve multiview shader replacement

z3moon:
Replace the num_views for OpenGL multiviwe only when
- The engine is initialized with multiview stereo
- The variant for the material contains STE flag
- The program is for surface
- It's vertex shader (this is already in)

### DIFF
--- a/filament/backend/include/backend/Program.h
+++ b/filament/backend/include/backend/Program.h
@@ -145,7 +145,7 @@ public:
 
     uint64_t getCacheId() const noexcept { return mCacheId; }
 
-    bool getMultiview() const noexcept { return mMultiview; }
+    bool isMultiview() const noexcept { return mMultiview; }
 
     CompilerPriorityQueue getPriorityQueue() const noexcept { return mPriorityQueue; }
 
--- a/filament/backend/include/backend/Program.h
+++ b/filament/backend/include/backend/Program.h
@@ -116,6 +116,8 @@ public:
 
     Program& cacheId(uint64_t cacheId) noexcept;
 
+    Program& multiview(bool multiview) noexcept;
+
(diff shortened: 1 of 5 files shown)


filament/backend/src/opengl/ShaderCompilerService.cpp
```diff
@@ -263,8 +263,7 @@ ShaderCompilerService::program_token_t ShaderCompilerService::createProgram(
                     // compile the shaders
                     std::array<GLuint, Program::SHADER_TYPE_COUNT> shaders{};
                     compileShaders(gl,
-                            std::move(program.getShadersSource()),
-                            program.getSpecializationConstants(),
+                            program,
```
pixelflinger, line 266: I would like to keep the old code -- I didn't want to pass `program` directly into `compileShaders` because it's unclear what gets `std::move`'d at the call site.

pixelflinger (changes requested):

filament/backend/src/opengl/ShaderCompilerService.cpp
z3moon: I see. Let me change it back.

filament/backend/include/backend/Program.h
```diff
@@ -143,6 +145,8 @@ class Program {
 
     uint64_t getCacheId() const noexcept { return mCacheId; }
 
+    bool getMultiview() const noexcept { return mMultiview; }
```
poweifeng, line 148: nit: suggest `isMultiview`

poweifeng (approved):

filament/backend/include/backend/Program.h
z3moon: Makes sense, done!

pixelflinger (approved):

PR #7707, opened 2024-03-25
Add FILAMENT_ENABLE_MULTIVIEW option

z3moon:
This allows the engine to include multiview shader code for default materials.

### DIFF
--- a/filament/src/details/Engine.cpp
+++ b/filament/src/details/Engine.cpp
@@ -356,8 +356,12 @@ void FEngine::init() {
                     MATERIALS_DEFAULTMATERIAL_DATA, MATERIALS_DEFAULTMATERIAL_SIZE);
                 break;
             case StereoscopicType::MULTIVIEW:
+#ifdef FILAMENT_ENABLE_MULTIVIEW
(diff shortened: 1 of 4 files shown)


filament/src/details/Engine.cpp
```diff
@@ -356,8 +356,12 @@ void FEngine::init() {
                     MATERIALS_DEFAULTMATERIAL_DATA, MATERIALS_DEFAULTMATERIAL_SIZE);
                 break;
             case StereoscopicType::MULTIVIEW:
+#ifdef FILAMENT_ENABLE_MULTIVIEW
                 defaultMaterialBuilder.package(
                     MATERIALS_DEFAULTMATERIAL_MULTIVIEW_DATA, MATERIALS_DEFAULTMATERIAL_MULTIVIEW_SIZE);
+#else
+                assert_invariant(false);
```
poweifeng, line 363: Maybe add a sentence on the "why" of the assert?
Like 

```
assert_invariant(false && "Engine has been configured for multiview but Filament was not built with multiview support")
```

poweifeng (approved):

PR #7711, opened 2024-03-26
vk: delete instead of ref-count EmptyTexture

poweifeng:


poweifeng:
Merging due to time sensitivity

PR #7720, opened 2024-03-28
Throw an NSException on Apple platforms when panicking

bejado:
On Apple platforms, we should be throwing an `NSException`, which gives us better crash reporting.

### DIFF
--- a/libs/utils/src/Panic.cpp
+++ b/libs/utils/src/Panic.cpp
@@ -16,6 +16,10 @@
 
 #include <utils/Panic.h>
 
+#if defined(__APPLE__)
+#include <utils/darwin/Exception.h>
+#endif
+
 #include <atomic>
 #include <cstdarg>
 #include <cstdlib>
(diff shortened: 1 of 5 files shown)


libs/utils/src/Panic.cpp
```diff
@@ -139,6 +152,11 @@ void TPanic<T>::panic(char const* function, char const* file, int line, const ch
     va_end(args);
     T e(function, formatFile(file), line, reason);
     e.log();
+
+#if defined(__APPLE__)
+    details::throwDarwinException(e);
```
bejado, line 157: Actually, we may want to give users the option of which type of exception they want thrown, maybe through a compile-time flag. @pixelflinger thoughts?

pixelflinger: I'm thinking we should throw the C++ exception first if enabled. If not it'll revert to the darwinException.
If the user wants a darwin exception that can always throw one from their C++ exception handler.

pixelflinger (changes requested):
give priority to the C++ exception

PR #7721, opened 2024-03-28
Fix WebGL build with newer Emscripten

bejado:
Newer Emscripten toolchains require us to wrap raw pointers in `allow_raw_pointer`.


poweifeng (approved):
which version of emscripten does this work with?  Do we need to update the build script for web?

bejado:
> which version of emscripten does this work with? Do we need to update the build script for web?

I'm not sure when the Emscripten change was made, but this should be backwards compatible. That is, we should have been doing this all along, but it only recently became a failure. So no, I don't think we need to update any build scripts.

PR #7722, opened 2024-03-29
the user can now set a panic handler callback

pixelflinger:
BUGS=[331457244]

### DIFF
--- a/libs/utils/include/utils/Panic.h
+++ b/libs/utils/include/utils/Panic.h
@@ -250,42 +250,71 @@ namespace utils {
  */
 class UTILS_PUBLIC Panic {
 public:
+
+    using PanicHandlerCallback = void(*)(void* user, Panic const& panic);
+
+    /**
+     * Sets a user-defined handler for the Panic. If exceptions are enabled, the concrete Panic
+     * object will be thrown upon return; moreover it is acceptable to throw from the provided
+     * callback, but it is unsafe to throw the Panic object itself, since it's just an interface.
+     * It is also acceptable to abort from the callback. If exceptions are not enabled, std::abort()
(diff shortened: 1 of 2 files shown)


libs/utils/include/utils/Panic.h
```diff
@@ -250,42 +250,69 @@ namespace utils {
...
 class UTILS_PUBLIC Panic {
 public:
+
+    using PanicHandlerCallback = void(*)(void* user, Panic const& panic);
+
+    /**
+     * Sets a user-defined handler for the Panic. If exceptions are enabled, the concrete Panic
+     * object will be thrown upon return; moreover it is acceptable to throw from the provided
+     * callback, but it is unsafe to throw the Panic object itself. It is also acceptable
+     * to abort / terminate from the callback. If exception are not enabled, std::abort() will
```
poweifeng, line 260:
```suggestion
     * to abort / terminate from the callback. If exceptions are not enabled, std::abort() will
```

poweifeng (approved):

libs/utils/include/utils/Panic.h
```diff
@@ -250,42 +250,69 @@ namespace utils {
...
+     * be automatically called upon return.
+     *
+     * Caveat: this API can misbehave if is used as a static library in multiple translation units,
+     * some of these translation units might not see the callback.
+     *
+     * @param handler pointer to the user defined handler for the Panic
+     * @param user  user pointer given back to the callback
+     */
+    static void setPanicHandler(PanicHandlerCallback handler, void* user) noexcept;
```
bejado, line 269: Maybe add a note about which thread the handler gets called on

bejado (approved):

PR #7724, opened 2024-03-29
GL backend should only allocate renderbuffers when possible

pixelflinger:
There was several cases where the the gl backend would wrongly  use a  renderbuffer instead of a texture.

BUGS=[329491941]


z3moon (approved):

z3moon:
Confirmed it worked well. Thanks!

PR #7737, opened 2024-04-04
wip: new descriptor set backend API

pixelflinger:


filament/backend/src/vulkan/caching/VulkanDescriptorSetManager.cpp
```diff
@@ -374,632 +246,208 @@ struct Equal {
...
-        }
-        commands->acquire(set);
 
-        if (mBoundState.vkSets[UBO_SET_ID] != vkSet) {
-            vkCmdBindDescriptorSets(mBoundState.cmdbuf, VK_PIPELINE_BIND_POINT_GRAPHICS,
-                    mBoundState.pipelineLayout, 0, 1, &vkSet, 0, nullptr);
-            mBoundState.vkSets[UBO_SET_ID] = vkSet;
+        for (uint8_t i = 0; i < UNIQUE_DESCRIPTOR_SET_COUNT; ++i) {
+            auto& itr = iterators[i];
+            if (itr != mHistory.end()) {
```
pixelflinger, line 407: it seems to me that all this code could be replaced by a 4 bits bitset with `forEachBitSet()`. All this does above is figure which of the 4 set needs a call to `bind()`, so this should just be 4 bits.

This code is part of the most inner loop we have, it should be optimized to death!

There should also be a fast path for "just changing the offsets". The way I did this in the gl backend is that I keep two sets of bitset and I can very quickly know which sets (1) don't change, (2) are only a bind offset update or  (3) need a real rebind.

pixelflinger:
see ma/descriptor-sets-1.54.3

PR #7755, opened 2024-04-13
Set VkApplicationInfo.pEngineName to Filament

pippy360:
Setting the pEngineName helps with the collection of metrics and is generally good practice.


poweifeng:
superceded by: https://github.com/google/filament/pull/7756

PR #7759, opened 2024-04-15
Introduce new matedit tool

bejado:
matedit is a new tool that can "edit" compiled material files. Initially, it will support a single command, "external-compile", which will be used to precompile MSL (Metal Shading Language) shaders into Metal shader libraries. This command will accept a path to an external, user-provided script which will perform the actual compilation.

BUGS=[333547148]


tools/matedit/src/main.cpp
```diff
@@ -0,0 +1,170 @@
...
+
+    const std::string command = argv[optionIndex++];
+    if (command != "external-compile") {
+        std::cerr << "Unrecognized command: '" << command << "'. Must be 'external-compile'"
+                  << std::endl;
+        return 1;
+    }
+
+    for (int i = optionIndex; i < argc; ++i) {
+        if (strcmp(argv[i], "--") == 0) {
```
poweifeng, line 156: nit: with this, technically you could have
```external-compile my-script.sh -- -my-script-option=1```
which I don't know is expected to be valid.

poweifeng (approved):

tools/matedit/src/main.cpp
bejado: Hmm, good point. Maybe we should only ignore "--" args between "external-compile" and the path to the script.

PR #7763, opened 2024-04-16
Fix broken dynamic resolution

poweifeng:
Previous PR #7758 accidentally removed the setting of 'resolution' for shaders other than 'blitLow'.  This PR undoes that.


pixelflinger (approved):
good catch. I remember now that I had kept the parameter so that all blitter use the same "interface".

z3moon (approved):

poweifeng:
> good catch. I remember now that I had kept the parameter so that all blitter use the same "interface".

Make sense. Feel free to let me know if you'd like to go back to that.

PR #7766, opened 2024-04-16
don't crash if we don't have a Camera set on View

pixelflinger:
- also add a method to query if a camera was set

### DIFF
--- a/android/filament-android/src/main/java/com/google/android/filament/View.java
+++ b/android/filament-android/src/main/java/com/google/android/filament/View.java
@@ -242,6 +242,15 @@ public class View {
     }
 
     /**
+     * Query whether a camera is set.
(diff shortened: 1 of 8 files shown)


android/filament-android/src/main/java/com/google/android/filament/View.java
```diff
@@ -1238,6 +1247,7 @@ void clearNativeObject() {
     private static native void nSetName(long nativeView, String name);
     private static native void nSetScene(long nativeView, long nativeScene);
     private static native void nSetCamera(long nativeView, long nativeCamera);
+    private static native void nHasCamera(long nativeView);
```
poweifeng, line 1250:
```suggestion
    private static native boolean nHasCamera(long nativeView);
```

poweifeng (approved):

PR #7768, opened 2024-04-16
filagui: Add support for external textures

elizagamedev:
We have an instance of a client trying to render a GL_TEXTURE_EXTERNAL_OES texture with filagui, which currently fails.


poweifeng (approved):

libs/filagui/src/materials/uiBlitExternal.mat
```diff
@@ -0,0 +1,29 @@
+material {
+    name : uiBlitExternal,
+    parameters : [
+        {
+            type : samplerExternal,
+            name : albedo
```
pixelflinger, line 6: What you could have done is using the same material + a bool spec constant to chose which sampler is used at the call site. on the c++ side you'd still need to create 2 materials, but from the same package. It might be easier to maintain?

elizagamedev: I agree. I pushed a new commit to do this.

pixelflinger (approved):

PR #7776, opened 2024-04-18
gltfio: enable extended implementation

### LINKED ISSUE #6358: SurfaceOrientation doesn't duplicate vertices when generating flat normals

poweifeng:
This change will enable proper flat-shading and MikkTSpace.

Fixes #6358, #7444

## Flat Shading comparison (model from #6358)

Before:
image

After: 
image

## Tangent space comparison (model from #7444)

Before:
image

After: 
image


samples/gltf_viewer.cpp
```diff
@@ -747,14 +747,25 @@ int main(int argc, char** argv) {
                 createJitShaderProvider(engine, OPTIMIZE_MATERIALS) :
                 createUbershaderProvider(engine, UBERARCHIVE_DEFAULT_DATA, UBERARCHIVE_DEFAULT_SIZE);
 
-        app.assetLoader = AssetLoader::create({engine, app.materials, app.names });
+        AssetConfigurationExtended extConfig = {};
```
romainguy, line 750: You're not doing anything with `extConfig`. You only set its gltfPath but you don't pass the config to anything.

libs/gltfio/include/gltfio/AssetLoader.h
```diff
@@ -38,6 +38,18 @@ namespace filament::gltfio {
...
+// Use this struct to enable mikktspace-based tangent-space computation.
+/**
+ * \struct AssetConfigurationExtended AssetLoader.h gltfio/AssetLoader.h
+ * \brief extends struct AssetConfiguration
+ * Useful if client needs mikktspace tangents.
+ */
+struct AssetConfigurationExtended {
+    //! Optional The same parameter as provided to \struct ResourceConfiguration ResourceLoader.h
+    //! gltfio/ResourceLoader.h
+    char const* gltfPath;
```
romainguy, line 50: I don't understand why this is realted to mikkt space? Why would setting a glTF path in a separate struct change the tangent space computations? This doesn't seem like a user-friendly API (and does it mean it only works with local files?)

libs/gltfio/include/gltfio/AssetLoader.h
```diff
@@ -62,6 +74,9 @@ struct AssetConfiguration {
 
     //! Optional default node name for anonymous nodes
     char* defaultNodeName = nullptr;
+
+    //! Optional to enable mikktspace tangents.
+    AssetConfigurationExtended* ext = nullptr;
```
bejado, line 79: Maybe comment on the required lifetime for this pointer? Based on the code it seems like it only needs to exist during `AssetLoader::create`

samples/gltf_viewer.cpp
poweifeng: sorry, I did not mean to commit gltf_viewer changes.  That was just for testing.  I undid the changes

libs/gltfio/include/gltfio/AssetLoader.h
poweifeng: Yes, deally, we'd just add support for mikktspace without changing the existing API, but that's not possible with how AssetLoader/ResourceLoader is structured.  I opted to introduce minimal changes to existing API to address the opened bugs.  So that any use case with the right criteria (desktop and only disk resources) can make use of it. But, the real, long-term solution should be a new loader API (separate from AssetLoader/ResourceLoader) that would better fit the assumption of running a remeshing algo like mikktspace.

libs/gltfio/include/gltfio/AssetLoader.h
poweifeng: Good point.  Comment added.

libs/gltfio/include/gltfio/AssetLoader.h
poweifeng: I added some comments in the header and in the commit to be more explicit about this approach's limitation.

pixelflinger: ah.. we should probably have a bug open then.

poweifeng: good point, I filed #7782 for tracking.

libs/gltfio/src/AssetLoader.cpp
```diff
@@ -202,14 +205,22 @@ class MaterialInstanceCache {
...
             mNameManager(config.names),
             mTransformManager(config.engine->getTransformManager()),
             mMaterials(*config.materials),
             mEngine(*config.engine),
-            mDefaultNodeName(config.defaultNodeName) {}
+            mDefaultNodeName(config.defaultNodeName) {
+        if (config.ext) {
+#if defined(__ANDROID__) || defined(IOS) || defined(__EMSCRIPTEN__)
+            PANIC_PRECONDITION("Extend asset loading is not supported on this platform");
+#endif
```
pixelflinger, line 219: Is there a way to query if it's supported?

pixelflinger (approved):

libs/gltfio/src/AssetLoader.cpp
poweifeng: moved this to a static method so that client can use the method to check for support.

PR #7777, opened 2024-04-18
add support for ADPF

pixelflinger:
- we add a new `utils::PerformanceHintManager` class that is only available on Android
- we also make use `utils::PerformanceHintManager` in `PlatformEGLAndroid`

TODOs: 
- add support in the Vulkan backend
- add support for the main thread and thread pools


filament/src/details/Renderer.cpp
```diff
@@ -298,7 +298,10 @@ bool FRenderer::beginFrame(FSwapChain* swapChain, uint64_t vsyncSteadyClockTimeN
         FEngine& engine = mEngine;
         FEngine::DriverApi& driver = engine.getDriverApi();
 
-        driver.beginFrame(appVsync.time_since_epoch().count(), mFrameId);
+        driver.beginFrame(
+                appVsync.time_since_epoch().count(),
+                int64_t(1'000'000'000.0 / mDisplayInfo.refreshRate),
```
poweifeng, line 303: [comment] mentions it that `refreshRate` is `0` for offscreen, is the comment outdated? 

[comment]: https://github.com/google/filament/blob/main/filament/include/filament/Renderer.h#L77

filament/backend/src/opengl/platforms/PlatformEGLAndroid.cpp
```diff
@@ -112,8 +117,34 @@ void PlatformEGLAndroid::terminate() noexcept {
...
+    PlatformEGL::beginFrame(monotonic_clock_ns, refreshIntervalNs, frameId);
+}
+
+void backend::PlatformEGLAndroid::preCommit() noexcept {
+    if (mPerformanceHintSession.isValid()) {
+        auto const actualWorkDuration = std::chrono::duration_cast<std::chrono::nanoseconds>(
+                clock::now() - mStartTimeOfActualWork);
+        mPerformanceHintSession.reportActualWorkDuration(actualWorkDuration.count());
+    }
+    OpenGLPlatform::preCommit();
```
poweifeng, line 137: `PlatformEGL` here?

filament/src/details/Renderer.cpp
pixelflinger: ah good catch. I should probably pass 0 in that case.

filament/backend/src/opengl/platforms/PlatformEGLAndroid.cpp
pixelflinger: uh. yes thanks!

poweifeng (approved):

bejado (approved):

PR #7781, opened 2024-04-19
don't rely on FMaterialInstance having a default ctor

pixelflinger:
FMaterialInstance needed a default ctor because it is a field of FMaterial but cannot be initialized before FMaterial itself is initialized. So we had a defautl ctor and we'd finish the initialization later. Conceptually the default material instance should have been  new'ed and a pointer to it stored instead. 

That's basically what we do now, but to avoid the extra allocation, we in-place new and delete the default material instance into an aligned_storage inside FMaterial.

### DIFF
--- a/filament/src/details/Material.h
+++ b/filament/src/details/Material.h
@@ -290,7 +290,7 @@ private:
     bool mIsDefaultMaterial = false;
     bool mSpecularAntiAliasing = false;
 
-    // reserve some sopce to construct the default material instance
+    // reserve some space to construct the default material instance
     std::aligned_storage<sizeof(FMaterialInstance), alignof(FMaterialInstance)>::type mDefaultInstanceStorage;
     static_assert(sizeof(mDefaultInstanceStorage) >= sizeof(mDefaultInstanceStorage));
 
--- a/filament/src/details/Material.h
+++ b/filament/src/details/Material.h
@@ -48,6 +48,7 @@
 #include <atomic>
 #include <memory>
 #include <mutex>
+#include <new>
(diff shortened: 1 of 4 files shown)


filament/src/details/Material.h
```diff
@@ -284,7 +290,10 @@ class FMaterial : public Material {
     bool mIsDefaultMaterial = false;
     bool mSpecularAntiAliasing = false;
 
-    FMaterialInstance mDefaultInstance;
+    // reserve some sopce to construct the default material instance
```
bejado, line 293: nit
```suggestion
    // reserve some space to construct the default material instance
```

bejado (approved):

filament/src/details/Material.h
```diff
@@ -284,7 +290,10 @@ class FMaterial : public Material {
...
+    std::aligned_storage<sizeof(FMaterialInstance), alignof(FMaterialInstance)>::type mDefaultInstanceStorage;
+    static_assert(sizeof(mDefaultInstanceStorage) >= sizeof(mDefaultInstanceStorage));
```
poweifeng, line 295: nit: may I ask for a little explanation on this assert?

poweifeng (approved):

filament/src/details/Material.h
pixelflinger: it's really not needed, I have it because if you forget the `::type` to the right of the expression, it compiles and you really don't get what you want (I did this before!), and that would catch it.  It's really just a way to be extra sure that I have enough space.

PR #7789, opened 2024-04-22
vk: Fix unsupported depth blitting

poweifeng:
On certain hardware (pixel 4 for example), blitting of depth texture is not supported as an "optimalTilingFeature".  In these cases, we'd would need to do a shader-based blit. We
 - Add the shader blit in PostProcessingManager
 - Add a driver API to check for support for blitting depthStencil attachments.
 - Fix some debugging ifdefs in vk backend.

The validation fixed is:
`[ VUID-vkCmdBlitImage-dstImage-02000 ] Object 0: handle = 0xb400007c300701d0, type = VK_OBJECT_TYPE_COMMAND_BUFFER; Object 1: handle = 0xf2039b0000000771, type = VK_OBJECT_TYPE_IMAGE; | MessageID = 0x86bc2a78 | In vkCmdBlitImage, VkFormatFeatureFlags (0x1c601) does not support required feature VK_FORMAT_FEATURE_2_BLIT_DST_BIT for format 126 used by VkImage 0xf2039b0000000771[] with tiling VK_IMAGE_TILING_OPTIMAL. The Vulkan spec states: The format features of dstImage must contain VK_FORMAT_FEATURE_BLIT_DST_BIT`


filament/backend/include/private/backend/DriverAPI.inc
```diff
@@ -303,6 +303,7 @@ DECL_DRIVER_API_SYNCHRONOUS_0(bool, isProtectedContentSupported)
 DECL_DRIVER_API_SYNCHRONOUS_N(bool, isStereoSupported, backend::StereoscopicType, stereoscopicType)
 DECL_DRIVER_API_SYNCHRONOUS_0(bool, isParallelShaderCompileSupported)
 DECL_DRIVER_API_SYNCHRONOUS_0(bool, isDepthStencilResolveSupported)
+DECL_DRIVER_API_SYNCHRONOUS_N(bool, isDepthStencilBlitSupported, backend::TextureFormat, format)
```
pixelflinger, line 306: so depth blitting support is not all or nothing? it depends on the specific format?

filament/src/PostProcessManager.cpp
```diff
@@ -3087,6 +3095,57 @@ FrameGraphId<FrameGraphTexture> PostProcessManager::blit(FrameGraph& fg, bool tr
...
+                mi->setParameter("depth", depth, {.filterMag = filterMag, .filterMin = filterMin});
+                mi->setParameter("viewport",
+                        float4{float(vp.left) / inputDesc.width,
+                            float(vp.bottom) / inputDesc.height, float(vp.width) / inputDesc.width,
+                            float(vp.height) / inputDesc.height});
+                mi->commit(driver);
+                mi->use(driver);
+
+                auto pipeline = material.getPipelineState(mEngine);
+                render(out, pipeline, driver);
```
pixelflinger, line 3143: you should be able to use `commitAndRender` here

filament/src/PostProcessManager.cpp
```diff
@@ -723,41 +724,48 @@ FrameGraphId<FrameGraphTexture> PostProcessManager::screenSpaceAmbientOcclusion(
...
+                    auto const& src = resources.getTexture(data.input);
+                    auto const& dst = resources.getTexture(data.output);
+                    auto const& srcSubDesc = resources.getSubResourceDescriptor(data.input);
+                    auto const& dstSubDesc = resources.getSubResourceDescriptor(data.output);
+                    auto const& desc = resources.getDescriptor(data.output);
+                    assert_invariant(desc.samples == resources.getDescriptor(data.input).samples);
+                    // here we can guarantee that src and dst format and size match, by
+                    // construction.
+                    driver.blit(dst, dstSubDesc.level, dstSubDesc.layer, {0, 0}, src,
+                            srcSubDesc.level, srcSubDesc.layer, {0, 0}, {desc.width, desc.height});
```
pixelflinger, line 761: could we keep the original formatting here? it makes it easier to read the src/dst parameters

filament/src/PostProcessManager.cpp
```diff
@@ -723,41 +724,48 @@ FrameGraphId<FrameGraphTexture> PostProcessManager::screenSpaceAmbientOcclusion(
      * The pass below that does this is automatically culled if not needed, which is decided by
      * each backend.
      */
+    auto const& depthDesc = fg.getDescriptor(depth);
+    FrameGraphId<FrameGraphTexture> duplicateDepthOutput;
+    if (mEngine.getDriverApi().isDepthStencilBlitSupported(depthDesc.format)) {
```
pixelflinger, line 729: At this point it might be better to either:
- move the DuplicateDepthPass into PostProcessManager (so that we get a simple `if{}else{}` here), or
- have a single blitDepth() method in PostProcessManager that picks the right implementation

pixelflinger (changes requested):

filament/backend/include/private/backend/DriverAPI.inc
poweifeng: In practice, it's probably all or nothing. 
But in Vulkan, for the general use case of "can you do x with texture", it's  a check against the format of the texture.

filament/src/PostProcessManager.cpp

filament/src/PostProcessManager.cpp
poweifeng: moved the conditional to `blitDepth`.

pixelflinger (approved):

filament/src/PostProcessManager.cpp

filament/backend/src/vulkan/platform/VulkanPlatform.cpp
```diff
@@ -524,6 +524,28 @@ VkFormatList findAttachmentDepthFormats(VkPhysicalDevice device) {
...
+    for (VkFormat format = (VkFormat) 1;;) {
+        if (isVkDepthFormat(format)) {
+            VkFormatProperties props;
+            vkGetPhysicalDeviceFormatProperties(device, format, &props);
+            if ((props.optimalTilingFeatures & required) == required) {
+                selectedFormats.push_back(format);
+            }
+        }
+        format = (VkFormat) (1 + (int) format);
+        if (format == VK_FORMAT_ASTC_12x12_SRGB_BLOCK) {
```
bejado, line 540: Maybe comment that this is the last format provided by VK_VERSION_1_0? And do we want to exclude `VK_FORMAT_ASTC_12x12_SRGB_BLOCK` from this check (maybe this if needs to be before line 539)?

poweifeng: yeah, this could definitely be cleaner.  I'll put up a follow-up to clean this up.

PR #7791, opened 2024-04-23
small cleanups and fixes

pixelflinger:
- add missing includes in PerShadowMapUninforms and FRenderableManager
- fix StructureOfArray::push_back(&&) which did a copy instead of a move
- give proper move semantics to Handle<>

### DIFF
--- a/filament/backend/include/backend/Handle.h
+++ b/filament/backend/include/backend/Handle.h
@@ -75,6 +75,19 @@ protected:
     HandleBase(HandleBase const& rhs) noexcept = default;
     HandleBase& operator=(HandleBase const& rhs) noexcept = default;
 
+    HandleBase(HandleBase&& rhs) noexcept
+            : object(rhs.object) {
+        rhs.object = nullid;
+    }
+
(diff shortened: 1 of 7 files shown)


filament/backend/include/backend/Handle.h
```diff
@@ -75,6 +75,18 @@ class HandleBase {
...
     HandleBase& operator=(HandleBase const& rhs) noexcept = default;
 
+    HandleBase(HandleBase&& rhs) noexcept
+            : object(rhs.object) {
+        rhs.object = nullid;
+    }
+
+    HandleBase& operator=(HandleBase&& rhs) noexcept {
+        auto t = rhs.object;
+        rhs.object = object;
```
z3moon, line 85: This looks like a swap semantic. Do we need to assign the old value to rhs?

z3moon (approved):

filament/backend/include/backend/Handle.h
pixelflinger: Actually come to think about it, you're right. That's not a good implementation. Will fix.

PR #7792, opened 2024-04-23
Switch setFrameScheduledCallback to use utils::Invocable

bejado:
This allows clients to pass capturing lambdas, which simplifies memory management.

BUGS=[335465900]


filament/backend/include/backend/DriverEnums.h
```diff
@@ -1224,7 +1225,7 @@ static_assert(sizeof(StencilState::StencilOperations) == 5u,
 static_assert(sizeof(StencilState) == 12u,
         "StencilState size not what was intended");
 
-using FrameScheduledCallback = void(*)(PresentCallable callable, void* user);
+using FrameScheduledCallback = utils::Invocable<void(backend::PresentCallable)>;
```
pixelflinger, line 1228: once downside to remember is that `Invocable` always ends-up doing a heap allocation

pixelflinger (approved):

PR #7796, opened 2024-04-24
Metal: track types of buffers

bejado:
BUGS=[336819840]

### DIFF
--- a/filament/backend/src/metal/MetalBuffer.h
+++ b/filament/backend/src/metal/MetalBuffer.h
@@ -35,18 +35,32 @@ class TrackedMetalBuffer {
 public:
 
     enum class Type {
-        NONE,
-        GENERIC,
-        RING,
-        STAGING,
+        NONE = 0,
+        GENERIC = 1,
+        RING = 2,
+        STAGING = 3,
     };
+    static constexpr size_t TypeCount = 3;
+
+    static constexpr auto toIndex(Type t) {
+        assert_invariant(t != Type::NONE);
+        switch (t) {
+            case Type::NONE:
+            case Type::GENERIC:
+                return 0;
+            case Type::RING:
+                return 1;
+            case Type::STAGING:
+                return 2;
+        }
+    }
 
     TrackedMetalBuffer() noexcept : mBuffer(nil) {}
     TrackedMetalBuffer(nullptr_t) noexcept : mBuffer(nil) {}
     TrackedMetalBuffer(id<MTLBuffer> buffer, Type type) : mBuffer(buffer), mType(type) {
         assert_invariant(type != Type::NONE);
         if (buffer) {
-            aliveBuffers[type]++;
+            aliveBuffers[toIndex(type)]++;
             mType = type;
         }
     }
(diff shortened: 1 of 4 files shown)


poweifeng (approved):

filament/backend/src/metal/MetalBuffer.h
```diff
@@ -57,18 +70,29 @@ class TrackedMetalBuffer {
...
-        other.mBuffer = temp;
+        std::swap(mBuffer, other.mBuffer);
+        std::swap(mType, other.mType);
     }
 
     id<MTLBuffer> mBuffer;
+    Type mType = Type::NONE;
 
-    static std::atomic<uint64_t> aliveBuffers;
+    static std::unordered_map<TrackedMetalBuffer::Type, std::atomic<uint64_t>> aliveBuffers;
```
pixelflinger, line 95: it's really not ideal to have a hashmap look-up each time. Please considere putting all this behind a compile or runtime flag so we can disable it later, once we figured out the issues. It's probably worth filing a bug so we don't forget.

pixelflinger (approved):

pixelflinger (changes requested):
looks like there could be a race condition on accessing `aliveBuffers` ?

filament/backend/src/metal/MetalBuffer.h
bejado: Great points-- for this iteration, I think I can avoid the hash map and do the "lookup" at compile time. And good idea, I'll put this behind a flag.

filament/backend/src/metal/MetalBuffer.h
```diff
@@ -57,18 +84,29 @@ class TrackedMetalBuffer {
...
-        other.mBuffer = temp;
+        std::swap(mBuffer, other.mBuffer);
+        std::swap(mType, other.mType);
     }
 
     id<MTLBuffer> mBuffer;
+    Type mType = Type::NONE;
 
-    static std::atomic<uint64_t> aliveBuffers;
+    static std::array<uint64_t, TypeCount> aliveBuffers;
```
pixelflinger, line 109: should probably be `mAliveBuffers`

filament/backend/src/metal/MetalBuffer.mm
```diff
@@ -22,7 +22,7 @@
 namespace filament {
 namespace backend {
 
-std::atomic<uint64_t> TrackedMetalBuffer::aliveBuffers = 0;
+std::array<uint64_t, TrackedMetalBuffer::TypeCount> TrackedMetalBuffer::aliveBuffers = { 0 };
```
pixelflinger, line 25: so you sure this is never accessed from multiple threads?

pixelflinger (approved):

filament/backend/src/metal/MetalBuffer.h
bejado: Leaving off the `m` prefix since it's `static`

filament/backend/src/metal/MetalBuffer.mm
bejado: Yes, this is always accessed from the driver thread

PR #7800, opened 2024-04-25
clean-ups to prepare for descriptor set changes

pixelflinger:
The main goal of this change is to avoid having to select the 
"per renderable" UBO at execution time. Instead we want the
PrimitiveInfo structure to already know which UBO will be used.
Concretely, this means that this determination must be done when
the RenderPass is created.

When automatic instancing is used, the RenderPass creates a temporary
UBO to store the instance info. This UBO's life-time is dictated by both
the life-time of the RenderPass and the Executors that where created
from it. For this reason we introduce SharedHandle<> to correctly 
account for the owner's lifetime. This fixes a potential bugs where
that handle could have been destroyed and used later; in practice this
bug didn't happen however.

A couple other changes:

- RenderPass has a bunch of fields that were actually temporary, so we
removed those.

- The canonical "per-renderable" UBO was owned by View but accessed 
through Scene. This was confusing, it's now accessed through View.

### DIFF
--- a/filament/src/RenderPass.h
+++ b/filament/src/RenderPass.h
@@ -19,16 +19,19 @@
 
 #include "Allocators.h"
 
+#include "SharedHandle.h"
+
 #include "details/Camera.h"
 #include "details/Scene.h"
 
 #include "private/filament/Variant.h"
-#include "utils/BitmaskEnum.h"
 
+#include <backend/DriverApiForward.h>
 #include <backend/DriverEnums.h>
 #include <backend/Handle.h>
 
 #include <utils/Allocator.h>
+#include <utils/BitmaskEnum.h>
 #include <utils/Range.h>
 #include <utils/Slice.h>
 #include <utils/architecture.h>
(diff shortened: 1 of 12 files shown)


filament/src/RenderPass.h
```diff
@@ -257,16 +261,13 @@ class RenderPass {
         bool hasMorphing : 1;                                           //              1 bit
         bool hasHybridInstancing : 1;                                   //              1 bit
 
-        uint64_t rfu[2];                                                // 16 bytes
-
-        static const uint16_t USER_INSTANCE_MASK = 0x8000u;
-        static const uint16_t INSTANCE_COUNT_MASK = 0x7fffu;
+        uint32_t rfu[3];                                                // 16 bytes
```
poweifeng, line 264:
```suggestion
        uint32_t rfu[3];                                                // 12 bytes
```

poweifeng (approved):

filament/src/RenderPass.h
pixelflinger: good catch, I'll fix in another CL since it's just a comment. thanks!

PR #7802, opened 2024-04-26
Remove bitcode from iOS builds

vpopovyc:
With the release of Xcode 14, bitcode was deprecated.

Reference: https://developer.apple.com/documentation/xcode-release-notes/xcode-14-release-notes#Deprecations

> Xcode no longer builds bitcode by default and generates a warning message if a project explicitly enables bitcode: “Building with bitcode is deprecated. Please update your project and/or target settings to disable bitcode.”

It’s important to note that any app submitted to the AppStore with a bitcode will be rejected, triggering the [ITMS-90482](https://www.reddit.com/r/iOSProgramming/comments/11diyye/apple_insisting_project_contains_bitcode/) error.

RN, the latest Filament binary for iOS, contains the bitcode.
Therefore, with this change, I propose to allow the optional exclusion of the bitcode. 

I’ve followed https://github.com/google/filament/pull/2708 to revert code blocks that were affected.

Sidenote: 
It may be reasonable to assume that a bitcode can be turned off by default for iOS builds.


bejado:
LGTM, but since it's deprecated now, we can probably just remove bitcode altogether. I think I originally added bitcode support because it was a requirement for Cocoapods. If that's no longer the case, then let's just remove it. Clients are expected to use a recent Xcode anyway.

vpopovyc:
Ok, I'll update the PR to remove it.

bejado (approved):

PR #7808, opened 2024-04-26
vk: finer-grain barrier at renderpass end

poweifeng:
- At the end of a renderpass we use a more fine-grained barrier for each of the attachments in the render target
 - Make sure that buffer update are barrier'd from previous reads
 - Remove previous Mali workaround barriers.  Seems to be fine without them on pixels + Mali.


pixelflinger (approved):
does this translate into measurable changes? e.g. can you "see" something different in AGI for instance?

poweifeng:
> does this translate into measurable changes? e.g. can you "see" something different in AGI for instance?

Unfortunately, no visible improvements from AGI. This is a part of process of elimination, I suppose, to identify the cause of bubbles that we see in AGI.

PR #7809, opened 2024-04-30
fix ES2 support

pixelflinger:
- make sure we can build with ES2 headers only
- make sure to not use ES3 features when in ES2 mode (post processing was accidentally creating an R8 texture,  which is not supported)


elizagamedev:
LGTM. I remember running into problems with 3D textures on ES2.

elizagamedev (approved):

poweifeng (approved):

PR #7816, opened 2024-05-01
Add preferredShaderLanguage option to Engine::Config

bejado:


filament/include/filament/Engine.h
```diff
@@ -28,6 +28,8 @@
 #include <stdint.h>
 #include <stddef.h>
 
+#include <optional>
```
pixelflinger, line 31: We shouldn't use the STL in public headers.

filament/include/filament/Engine.h
```diff
@@ -340,6 +343,25 @@ class UTILS_PUBLIC Engine {
...
+         * shaders if present within a material, falling back to MSL. Setting
+         * preferredShaderLanguage to ShaderLanguage::MSL will instead instruct Filament to check
+         * for the presence of MSL in a material first, falling back to METAL_LIBRARY if MSL is not
+         * present.
+         *
+         * Currently, the only supported values are ShaderLanguage::MSL or
+         * ShaderLanguage::METAL_LIBRARY. When using a non-Metal backend, setting this has no
+         * effect.
+         */
+        std::optional<ShaderLanguage> preferredShaderLanguage;
```
pixelflinger, line 364: Instead of an optional, I would just have an enum that contains a DEFAULT value.

pixelflinger (changes requested):

pixelflinger (approved):

PR #7817, opened 2024-05-01
Add push constants

poweifeng:
- Push constants is a small set of bytes that can be recorded directly on the command buffer.
 - Introduce its use for internal purposes.
 - Implemented it for the vulkan backend.


filament/backend/include/private/backend/DriverAPI.inc
```diff
@@ -434,6 +434,10 @@ DECL_DRIVER_API_N(bindSamplers,
         uint32_t, index,
         backend::SamplerGroupHandle, sbh)
 
+DECL_DRIVER_API_N(writePushConstants,
+        backend::ShaderStage, stage,
+        backend::PushConstantArray, pushConstants)
```
pixelflinger, line 439: I would prefer an API like `setPushConstant(stage, index, value)` so that it's very efficient to set just a couple ones. Otherwise we'll have to send 128 bytes each time (possible per draw call) in the command stream.

filament/backend/include/backend/Program.h
```diff
@@ -117,6 +120,8 @@ class Program {
     Program& specializationConstants(
             utils::FixedCapacityVector<SpecializationConstant> specConstants) noexcept;
 
+    Program& pushConstants(PushConstantStruct const& pushConstants) noexcept;
```
pixelflinger, line 123: I would prefer a FixedCapacityVector based API, passed by value (so we can std::move), like the other APIs here.

filament/backend/include/backend/DriverEnums.h
```diff
@@ -1219,6 +1222,20 @@ struct StencilState {
     uint8_t padding = 0;
 };
 
+struct PushConstant {
+    using Variant = std::variant<int, float, bool>;
+    char const* name = nullptr;
```
pixelflinger, line 1227: In this header the push constant should only be a value, it shouldn't need to have a name.  The name is needed in Program though, so that the gl backend can find it in the shader.

filament/backend/include/backend/DriverEnums.h
```diff
@@ -1219,6 +1222,20 @@ struct StencilState {
...
+    Variant value;
+};
+
+using PushConstantArray = std::array<PushConstant, MAX_PUSH_CONSTANT_COUNT>;
+
+struct PushConstantStruct {
```
pixelflinger, line 1233: This shouldn't be needed in this header. Probably just in Program.

filament/backend/include/backend/Program.h
pixelflinger, line 123: Here we need to describe the push constant buffer, so we shouldn't pass any "value".

I think this should be a vector of names (this will associate the name to an index basically, and it'll only be needed for the gl driver in theory, and maybe the vk driver to allocate the buffer size?).

poweifeng: In the definition of the standalone uniform struct for GL:
```
struct MyConstants {
    float t;
};
uniform MyConstants k;
```
Is it ok for `k` remain be the same across all shaders?

filament/backend/include/private/backend/DriverAPI.inc
poweifeng: So should Material do the name to index mapping?

filament/backend/include/backend/Program.h
pixelflinger: Yes, I think so.

In gl, you can look up the names of each field of the structure, it will return a location you can use with glUniform to set it.

So what we need, is for Program to be given the name of each field of the structure. I think I'm this example it would be "k.foo", along with an index to be used when setting the value with the driver API (bit I think I'm this case the index should be implicit, based on the order on the vector, which should match the order in the struct{}).

The implementation of program in each backend will then maintain a mapping of its choosing to implement the setPushConstant.

I think on vk/mtl it'll be just an offset into the buffer (like index * 4). In gl it will be mapping from index to location + a call to glUniform.

I hope this makes sense.

poweifeng: sure, but one last question.  For `RenderPass` or `PostProcessManager` when they call `driver.setPushConstant(stage, index, value)`, who do they ask for the index (I assume they would only know the name of the variable within the struct), is it `FMaterial`?

filament/backend/include/private/backend/DriverAPI.inc
poweifeng: done

filament/backend/include/backend/DriverEnums.h
poweifeng: removed.

filament/backend/include/backend/DriverEnums.h
poweifeng: removed.

filament/backend/include/backend/Program.h
pixelflinger: What I have envisioned for now is that the `index` would be hardcoded inside `EngineEnums.h`. We'd probably have something like:

```
enum class PushConstantId {
    MORPHING_BUFFER_OFFSET = 0,   // uint32
    SRGB_CONVERSION = 1,                   // bool
};
```
In `libfilamat` we would output the push constant block, something like this:

```
struct {
 int morphingBufferOffset;
 bool srgbConversion;
} pushConstants;
```

`MaterialBuilder` would gain a way to define the push constants e.g.:
```
MaterialBuilder::pushConstant(name, type, index);
```
This info will be used to generate the `pushConstants` struct above.

And I think for now the calls to `MaterialBuilder` would just be hardcoded, e.g.:
```
builder.pushConstant("morphingBufferOffset", Type::INT, PushConstantId::MORPHING_BUFFER_OFFSET);
builder.pushConstant("srgbConversion", Type::BOOL, PushConstantId::SRGB_CONVERSION);
```


I *think* the `stage` parameter should only appear when setting the push constant in the driver API, either that, or we assume that they're always set in both -- I don't think we should have a stage per push constant.  In GL they will always be set in both (so they cannot share an id).

poweifeng: done

filament/backend/include/backend/Program.h
```diff
@@ -117,6 +117,9 @@ class Program {
     Program& specializationConstants(
             utils::FixedCapacityVector<SpecializationConstant> specConstants) noexcept;
 
+    Program& pushConstants(ShaderStage stage,
+            utils::FixedCapacityVector<char const*> constants) noexcept;
```
pixelflinger, line 121: I think we might have to use `CString` here because this goes through the command stream and it would be easu to get dangling pointers.

pixelflinger: Do we need the type of each constants here?

filament/backend/include/private/backend/DriverAPI.inc
```diff
@@ -434,6 +434,11 @@ DECL_DRIVER_API_N(bindSamplers,
         uint32_t, index,
         backend::SamplerGroupHandle, sbh)
 
+DECL_DRIVER_API_N(setPushConstant,
+        backend::ShaderStage, stage,
+        uint8_t, index,
+        backend::PushConstantVariant, value)
```
pixelflinger, line 440: if the program doesn't know the type from `Program::pushConstants()` it won't be able to validate that the value is correct. I think that's probably okay for an internal API though.

filament/backend/src/Program.cpp
```diff
@@ -91,6 +91,12 @@ Program& Program::specializationConstants(
     return *this;
 }
 
+Program& Program::pushConstants(ShaderStage stage,
+        utils::FixedCapacityVector<char const*> constants) noexcept {
+    mPushConstants[static_cast<uint8_t>(stage)] = constants;
```
pixelflinger, line 96: you need to `std::move()` or `std::swap()` here

filament/backend/src/opengl/OpenGLDriver.cpp
```diff
@@ -285,6 +285,10 @@ void OpenGLDriver::bindSampler(GLuint unit, GLuint sampler) noexcept {
     mContext.bindSampler(unit, sampler);
 }
 
+void OpenGLDriver::setPushConstant(backend::ShaderStage stage, uint8_t index,
+        backend::PushConstantVariant value) {
```
pixelflinger, line 289: The implementation here would be to call `glUniform{i|f}()` with the correct location from the index.
That location would have been pre-cached during init from the name. Easy.

filament/backend/src/vulkan/VulkanHandles.cpp
```diff
@@ -112,27 +112,69 @@ inline VkDescriptorSetLayout createDescriptorSetLayout(VkDevice device,
...
+        binaryValue = *reinterpret_cast<uint32_t const*>(&ival);
+    }
+    vkCmdPushConstants(cmdbuf->buffer(), layout, getVkStage(stage), index * 4, 4, &binaryValue);
+}
+
 VulkanProgram::VulkanProgram(VkDevice device, Program const& builder) noexcept
     : HwProgram(builder.getName()),
       VulkanResource(VulkanResourceType::PROGRAM),
-      mInfo(new PipelineInfo()),
+      mInfo(new PipelineInfo(builder)),
```
pixelflinger, line 177: new(std::nothrow)

libs/filabridge/include/private/filament/EngineEnums.h
```diff
@@ -71,6 +72,23 @@ enum class ReservedSpecializationConstants : uint8_t {
...
+};
+const utils::FixedCapacityVector<char const*> PUSH_CONSTANT_NAMES = {
+    "morphingBufferOffset",
+};
+const utils::FixedCapacityVector<PushConstantType> PUSH_CONSTANT_TYPES = {
+    PushConstantType::INT,
+};
+
+constexpr char const* PUSH_CONSTANT_STRUCT_VAR_NAME = "pushConstants";
+
```
pixelflinger, line 91: These should not appear in EngineEnums.h because filament itself will never use them. This should be hidden in some other private header. It's only needed right now because this info is not stored into the material, but it should be. You can add a header next to EngineEnums.h for now (same location).

pixelflinger (changes requested):
- use std::move()/std::swap() when setting the constants in Program
- use CString instead of char*
- remove conceptually private stuff from EngineEnums.h

filament/backend/src/vulkan/VulkanHandles.cpp
```diff
@@ -112,27 +112,69 @@ inline VkDescriptorSetLayout createDescriptorSetLayout(VkDevice device,
...
+    mRangeCount = 0;
+    for (auto stage : { ShaderStage::VERTEX, ShaderStage::FRAGMENT, ShaderStage::COMPUTE }) {
+        auto const& names = program.getPushConstants(stage);
+        if (names.empty()) {
+            continue;
+        }
+        mRanges[mRangeCount++] = {
+            .stageFlags = getVkStage(stage),
+            .offset = 0,
+            .size = (uint32_t) names.size() * 4,
```
bejado, line 152: `sizeof(uint32_t)` instead of `4`?

filament/backend/src/vulkan/VulkanHandles.cpp
```diff
@@ -112,27 +112,69 @@ inline VkDescriptorSetLayout createDescriptorSetLayout(VkDevice device,
...
+        bool const bval = std::get<bool>(value);
+        binaryValue = static_cast<uint32_t const>(bval ? VK_TRUE : VK_FALSE);
+    } else if (std::holds_alternative<float>(value)) {
+        float const fval = std::get<float>(value);
+        binaryValue = *reinterpret_cast<uint32_t const*>(&fval);
+    } else {
+        int const ival = std::get<int>(value);
+        binaryValue = *reinterpret_cast<uint32_t const*>(&ival);
+    }
+    vkCmdPushConstants(cmdbuf->buffer(), layout, getVkStage(stage), index * 4, 4, &binaryValue);
```
bejado, line 171: Same here

filament/backend/src/vulkan/caching/VulkanPipelineLayoutCache.h
```diff
@@ -35,13 +35,27 @@ class VulkanPipelineLayoutCache {
...
+        // Note that there is also an offset parameter for push constants, but
+        // we always assume our update range will have the offset 0.
+    };
+
+    struct PipelineLayoutKey {
+        using DescriptorSetLayoutArray = std::array<VkDescriptorSetLayout,
+                VulkanDescriptorSetLayout::UNIQUE_DESCRIPTOR_SET_COUNT>;
+        DescriptorSetLayoutArray descSetLayouts = {};
+        PushConstantKey pushConstant[Program::SHADER_TYPE_COUNT] = {};
+    };
```
bejado, line 50: Maybe assert the size of `PipelineLayoutKey` is what we expect to ensure no padding bytes are inserted by the compiler.

filament/backend/include/private/backend/DriverAPI.inc
poweifeng: done

filament/backend/src/opengl/OpenGLDriver.cpp
poweifeng: done

filament/backend/src/vulkan/caching/VulkanPipelineLayoutCache.h
poweifeng: done

filament/backend/include/backend/Program.h
poweifeng: done

filament/backend/include/backend/Program.h
poweifeng: done

filament/backend/src/Program.cpp
poweifeng: done

filament/backend/src/vulkan/VulkanHandles.cpp
poweifeng: done

libs/filabridge/include/private/filament/EngineEnums.h
poweifeng: So the "reserved" spec consts are also in this file.  Should they move as well?

filament/backend/src/vulkan/VulkanHandles.cpp
poweifeng: done

filament/backend/src/vulkan/VulkanHandles.cpp
poweifeng: done

filament/backend/src/opengl/OpenGLProgram.cpp
```diff
@@ -30,6 +30,7 @@
 #include <utils/Systrace.h>
 
 #include <array>
+#include <sstream>
```
pixelflinger, line 33: I really don't want to include <sstream> because it can bring in a lot of code. I just saw it's used in `Material.cpp`, but that's a mistake, I'll fix it.

filament/backend/src/opengl/OpenGLProgram.cpp
```diff
@@ -62,6 +64,11 @@ OpenGLProgram::OpenGLProgram(OpenGLDriver& gld, Program&& program) noexcept
         lazyInitializationData->uniformBlockInfo = std::move(program.getUniformBlockBindings());
     }
 
+    // We only keep the push constants for vertex stage because we'd like to keep this class a
+    // certain size.
+    lazyInitializationData->pushConstants =
+            std::move(program.getPushConstants(ShaderStage::VERTEX));
+
```
pixelflinger, line 71: I don't understand this comment, this will probably break if we don't have the two sets?

filament/backend/src/opengl/OpenGLProgram.cpp
```diff
@@ -203,6 +210,29 @@ void OpenGLProgram::initializeProgramState(OpenGLContext& context, GLuint progra
...
+    auto& constants = lazyInitializationData.pushConstants;
+    if (!constants.empty()) {
+        mVertexPushConstants.reserve(constants.size());
+        mVertexPushConstants.resize(constants.size());
+        // The constants are defined in a struct of name PUSH_CONSTANT_STRUCT_VAR_NAME. We prepend
+        // the prefix here to avoid string manipulation in a the draw loop.
+        std::stringbuf constantPrefix;
+        constantPrefix.sputn(Program::PUSH_CONSTANT_STRUCT_VAR_NAME,
+                strlen(Program::PUSH_CONSTANT_STRUCT_VAR_NAME));
+        constantPrefix.sputn(".", 1);
```
pixelflinger, line 224: you can probably implement this with just a `snprintf()`. Try to avoid the STL for string manipulation as it brings it a lot of code.

filament/backend/src/opengl/OpenGLProgram.cpp
```diff
@@ -203,6 +210,29 @@ void OpenGLProgram::initializeProgramState(OpenGLContext& context, GLuint progra
...
+
+        std::transform(constants.begin(), constants.end(), mVertexPushConstants.begin(),
+                [&constantPrefix, program](Program::PushConstant& constant) -> PushConstantBundle {
+                    constant.name.insert(0, utils::CString(constantPrefix.str().c_str()));
+                    auto const loc = glGetUniformLocation(program, constant.name.c_str());
+                    return {
```
pixelflinger, line 230: you need to check if loc is < 0, that would happen if somehow the uniform wasn't found. I think that can happen if the uniform is not used by the shader.  So this is not an error. It just means that setting the push constant will be a no-op.

filament/backend/src/opengl/OpenGLProgram.cpp
```diff
@@ -203,6 +210,29 @@ void OpenGLProgram::initializeProgramState(OpenGLContext& context, GLuint progra
         }
     }
     mUsedBindingsCount = usedBindingCount;
+
+    // Push constant initialization
+    auto& constants = lazyInitializationData.pushConstants;
+    if (!constants.empty()) {
+        mVertexPushConstants.reserve(constants.size());
+        mVertexPushConstants.resize(constants.size());
+        // The constants are defined in a struct of name PUSH_CONSTANT_STRUCT_VAR_NAME. We prepend
```
pixelflinger, line 219: it would be much better if the struct name was given through `Program` so we don't hardcode this in the backend.

filament/backend/src/opengl/OpenGLProgram.h
```diff
@@ -78,6 +85,13 @@ class OpenGLProgram : public HwProgram {
         GLuint program = 0;
     } gl;                                               // 4 bytes
 
+    utils::FixedCapacityVector<PushConstantBundle> const* getPushConstants() {
```
pixelflinger, line 88: that should be getPushConstant(stage). I think let's bite the bullet and implement the stages here.

filament/backend/src/opengl/OpenGLProgram.h
```diff
@@ -108,10 +121,12 @@ class OpenGLProgram : public HwProgram {
         mutable uint16_t age = std::numeric_limits<uint16_t>::max();
     };
     UniformsRecord const* mUniformsRecords = nullptr;
+
+    utils::FixedCapacityVector<PushConstantBundle> mVertexPushConstants;
 };
 
 // if OpenGLProgram is larger tha 64 bytes, it'll fall in a larger Handle bucket.
-static_assert(sizeof(OpenGLProgram) <= 64); // currently 48 bytes
+static_assert(sizeof(OpenGLProgram) <= 64); // currently 54 bytes
```
pixelflinger, line 129: I think you can keep the size below 64 bytes even if you handle both stages, by storing the info for both stages in the same vector, you just need a uint8_t to store the offset (we have 3 bytes padding available) and change the API a bit to return a range (you can use utils::Slice) when getting the constants.

libs/filabridge/include/private/filament/EngineEnums.h
```diff
@@ -71,6 +72,23 @@ enum class ReservedSpecializationConstants : uint8_t {
...
+};
+
+using PushConstantType = backend::ConstantType;
+
+const utils::FixedCapacityVector<char const*> PUSH_CONSTANT_NAMES = {
+    "morphingBufferOffset",
+};
+const utils::FixedCapacityVector<PushConstantType> PUSH_CONSTANT_TYPES = {
+    PushConstantType::INT,
+};
```
pixelflinger, line 90: I would prefer a single vector of pair<name, type>. And in fact maybe we need two such vector, one for each stage.

pixelflinger (changes requested):

filament/backend/src/opengl/OpenGLProgram.cpp
poweifeng: removed.

filament/backend/src/opengl/OpenGLProgram.cpp
poweifeng: done

filament/backend/src/opengl/OpenGLProgram.cpp
poweifeng: done in OpenGLDriver

filament/backend/src/opengl/OpenGLProgram.cpp
poweifeng: done.

filament/backend/src/opengl/OpenGLProgram.h
poweifeng: done in OpenGLDriver.  I encapsulated the offset+array in a new struct `PushConstantBundle`, which can be stashed in OpenGLDriver when pipeline is bound, and queried for constant when `setPushConstant` is called.

filament/backend/src/opengl/OpenGLProgram.h
poweifeng: done, but didn't use `utils::Slice` and just did simple arithmetic

libs/filabridge/include/private/filament/EngineEnums.h
poweifeng: done

libs/filabridge/include/private/filament/EngineEnums.h
pixelflinger: I guess my point is that this should be written into the material file as a Chunk, if we did that, we wouldn't need these constants in a header shared between filamat and filament (i.e. a filabridge header).

I would prefer that you implement that. But, if not, the next best thing is to put the constants in their own header to make it more clear that they shouldn't be part of EngineEnums.h.

I think I agree the spec constant should work like that as well.

poweifeng: I see.  I can put these and push constant into their own headers in this PR, and do a follow-up to put them in Chunks.

poweifeng: Actually, looking at the code, it looks like the reserved spec constants are different in that they have default values that need to be set from the filament side (so semantics do have to be shared between filamat and filament). 

I'll update this PR to put just the push constants in a Chunk.

filament/backend/src/opengl/OpenGLDriver.cpp
```diff
@@ -3802,6 +3833,12 @@ void OpenGLDriver::bindPipeline(PipelineState state) {
     gl.polygonOffset(state.polygonOffset.slope, state.polygonOffset.constant);
     OpenGLProgram* const p = handle_cast<OpenGLProgram*>(state.program);
     mValidProgram = useProgram(p);
+
+    if (!mCurrentPushConstants) {
+        mCurrentPushConstants = new (std::nothrow) PushConstantBundle{p->getPushConstants()};
+    } else {
+        (*mCurrentPushConstants) = p->getPushConstants();
+    }
```
pixelflinger, line 3841: Just use a value instead of a pointer since it will always be allocated, this will avoid a `if` here the first time.

filament/backend/src/opengl/OpenGLProgram.cpp
```diff
@@ -46,14 +46,14 @@ struct OpenGLProgram::LazyInitializationData {
     Program::UniformBlockInfo uniformBlockInfo;
     Program::SamplerGroupInfo samplerGroupInfo;
     std::array<Program::UniformInfo, Program::UNIFORM_BINDING_COUNT> bindingUniformInfo;
+    utils::FixedCapacityVector<Program::PushConstant> pushConstants;
+    uint8_t pushConstantFragmentStageOffset;
```
pixelflinger, line 50: I think it's fine to do it this way, but I would prefer that lazyInitializationData just store both vectors (it's temporary, and efficient because you can std::move), and instead have the logic of packing them into a single vector done during the actual program initialization.

filament/backend/src/opengl/OpenGLProgram.h
```diff
@@ -38,6 +38,24 @@ namespace filament::backend {
 
 class OpenGLDriver;
 
+struct PushConstantBundle {
+    uint8_t fragmentStageOffset = 0;
+    utils::FixedCapacityVector<std::pair<GLint, ConstantType>> const* constants = nullptr;
```
pixelflinger, line 43: I think you can even make this a reference

filament/backend/src/opengl/OpenGLProgram.h
```diff
@@ -78,6 +96,13 @@ class OpenGLProgram : public HwProgram {
         GLuint program = 0;
     } gl;                                               // 4 bytes
 
+    PushConstantBundle getPushConstants() {
```
pixelflinger, line 99: what you could do instead, is have a `getPushConstants(ShaderStage)` and have it return a `utils::Slice< pair<GLint, ConstantType>>`, Slice is basically the same thing as `std::range` and it can be used like a vector, it's very convenient.  You would initialize it with `{ mPushConstants.data() + stageOffset, stageSize }` This essentially what you're doing already, except manually.
But the benefit would be that once you get the slice for a stage, there is no more overhead accessing a particular entry.

filament/backend/src/opengl/OpenGLProgram.h
```diff
@@ -95,23 +120,29 @@ class OpenGLProgram : public HwProgram {
...
         Program::UniformInfo uniforms;
         LocationInfo locations;
         mutable GLuint id = 0;
         mutable uint16_t age = std::numeric_limits<uint16_t>::max();
     };
-    UniformsRecord const* mUniformsRecords = nullptr;
+    UniformsRecord const* mUniformsRecords = nullptr;               // 8 bytes
+
+    // Store [location, type] pairs.
+    utils::FixedCapacityVector<std::pair<GLint, ConstantType>> mPushConstants;  // 16 bytes
```
pixelflinger, line 141: not asking to do it in the PR, but technically we could save 8 bytes here by just using a raw pointer and storing the size in one of the reserved uint8_t. Might be worth commenting in case we need the 8 bytes some day. It doesn't really need to be a vector because we're never doing vector ops on it, once it's created.

pixelflinger (changes requested):

filament/backend/src/opengl/OpenGLProgram.h
poweifeng: The problem is I don't know which stage will be queried when the pipeline is bound (only when `setPushConstant` is called).  I could store both vertex and fragment slices in PushConstantBundle if you're ok with that.

filament/backend/src/opengl/OpenGLDriver.cpp
poweifeng: It's ok to pull in `OpenGLProgram.h` in `OpenGLDriver.h`?

filament/backend/src/opengl/OpenGLProgram.h
poweifeng: used two slices instead.

filament/backend/src/opengl/OpenGLProgram.h
poweifeng: stored both slices in the bundle

filament/backend/src/opengl/OpenGLProgram.h
poweifeng: added a comment on this.

libs/filabridge/include/private/filament/EngineEnums.h
poweifeng: done

filament/backend/src/opengl/OpenGLDriver.cpp
pixelflinger: Yes that should be okay.

pixelflinger (approved):

PR #7821, opened 2024-05-03
Update filament for web viewer

poweifeng:


pixelflinger (approved):

pixelflinger (commented):
you could update docs/remote too (there is a script there)

poweifeng:
> you could update docs/remote too (there is a script there)

Done

PR #7829, opened 2024-05-07
Add an Engine debug setting to force GLES 2.0

pixelflinger:
This setting is only meaningful on GLES backends, it's otherwise ignored. When set to true, the backend will try to force a ES2 context if supported. If not supported by the platform,
the backend will pretend it's a ES2 context.

This setting is currently only taken into account by the EGL platform.

### DIFF
--- a/filament/backend/include/backend/Platform.h
+++ b/filament/backend/include/backend/Platform.h
@@ -65,6 +65,12 @@ public:
          * Disable backend handles use-after-free checks.
          */
         bool disableHandleUseAfterFreeCheck = false;
+
+        /**
+         * Force GLES2 context if supported, or pretend the context is ES2. Only meaningful on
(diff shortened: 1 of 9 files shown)


filament/backend/include/backend/Platform.h
```diff
@@ -65,6 +65,12 @@ class UTILS_PUBLIC Platform {
          * Disable backend handles use-after-free checks.
          */
         bool disableHandleUseAfterFreeCheck = false;
+
+        /**
+         * Force GLES2 context if supported, or pretend the context is ES2. Only meaningful on
+         * GLES 3.x backendds
```
poweifeng, line 71:
```suggestion
         * GLES 3.x backends.
```

poweifeng (approved):

PR #7834, opened 2024-05-07
Metal: log slow buffer allocation times

bejado:


poweifeng (approved):

filament/backend/src/metal/MetalBuffer.h
```diff
@@ -29,9 +29,42 @@
...
+#include <chrono>
 
 namespace filament::backend {
 
+class ScopedAllocationTimer {
+public:
+    ScopedAllocationTimer(const char* name) : mBeginning(clock_t::now()), mName(name) {}
+    ~ScopedAllocationTimer() {
+        using namespace std::literals::chrono_literals;
+        static constexpr auto LONG_TIME_THRESHOLD = 10s;
```
pixelflinger, line 41: using `auto` here makes me very nervous.

filament/backend/src/metal/MetalBuffer.h
```diff
@@ -29,9 +29,42 @@
...
+                platform->debugUpdateStat(
+                        buffer, static_cast<uint64_t>(allocationTimeMicroseconds.count()));
+            }
+        }
+    }
+
+    static void setPlatform(MetalPlatform* p) { platform = p; }
+
+private:
+    typedef std::chrono::high_resolution_clock clock_t;
```
pixelflinger, line 60: I would use monotonic clock here, because you can't be guaranteed that high_Resolution_clock is monotonic.

pixelflinger (approved):
lgtm-ish, would prefer the use of monotonic clock.

PR #7839, opened 2024-05-09
Metal: implement more accurate buffer tracking

bejado:
This PR updates our Metal buffer tracking logic to use a `NSHashTable` to track alive buffers. `NSHashTable` is configured to hold weak references to the buffers we add to it. When a buffer is deallocated, it is automatically removed from the hash table. By counting the number of buffers in the hash table, we get an accurate count of the number of alive buffers.

The previous implementation had the limitation that once we pass a buffer off to Metal (when a command buffer becomes the sole owner of it, for example), we assume Metal will eventually release it and and decrement our count. In reality, that buffer could stay alive for a frame or two while the command buffer executes. Using the hash table should give us a more accurate tally of the alive buffer count.

### DIFF
--- a/filament/backend/src/metal/MetalBuffer.h
+++ b/filament/backend/src/metal/MetalBuffer.h
@@ -65,9 +65,12 @@ private:
     const char* mName;
 };
 
-class TrackedMetalBuffer {
-public:
+#ifndef FILAMENT_METAL_BUFFER_TRACKING
+#define FILAMENT_METAL_BUFFER_TRACKING 0
+#endif
 
+class MetalBufferTracking {
+public:
     static constexpr size_t EXCESS_BUFFER_COUNT = 30000;
 
     enum class Type {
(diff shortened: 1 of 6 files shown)


filament/backend/src/metal/MetalBuffer.h
```diff
@@ -91,66 +94,58 @@ class TrackedMetalBuffer {
...
     }
 
-    id<MTLBuffer> get() const noexcept { return mBuffer; }
-    operator bool() const noexcept { return bool(mBuffer); }
-
     static uint64_t getAliveBuffers() {
         uint64_t sum = 0;
-        for (const auto& v : aliveBuffers) {
-            sum += v;
+        for (size_t i = 1; i < TypeCount; i++) {
```
pixelflinger, line 123: why start at i=1 ?

filament/backend/src/metal/MetalBuffer.h
```diff
@@ -91,66 +94,58 @@ class TrackedMetalBuffer {
...
-
-    TrackedMetalBuffer(TrackedMetalBuffer&&) = delete;
-    TrackedMetalBuffer(TrackedMetalBuffer const&) = delete;
-    TrackedMetalBuffer& operator=(TrackedMetalBuffer const&) = delete;
+    static void setPlatform(MetalPlatform* p) { platform = p; }
 
-    TrackedMetalBuffer& operator=(TrackedMetalBuffer&& rhs) noexcept {
-        swap(rhs);
-        return *this;
+    static id<MTLBuffer> track(id<MTLBuffer> buffer, Type type) {
```
pixelflinger, line 109: you're never using the return value of track() so I would make it `void` (or use it).

filament/backend/src/metal/MetalBuffer.h
```diff
@@ -91,66 +94,58 @@ class TrackedMetalBuffer {
...
         return sum;
     }
 
     static uint64_t getAliveBuffers(Type type) {
         assert_invariant(type != Type::NONE);
-        return aliveBuffers[toIndex(type)];
+        NSHashTable* hashTable = aliveBuffers[toIndex(type)];
+        // Caution! We can't simply use hashTable.count here, which is inaccurate.
+        // See http://cocoamine.net/blog/2013/12/13/nsmaptable-and-zeroing-weak-references/
+        return hashTable.objectEnumerator.allObjects.count;
```
pixelflinger, line 134: everything hinges on this working, which looks like magic to me. do you have high confidence this is indeed working?

pixelflinger (changes requested):

filament/backend/src/metal/MetalBuffer.h
bejado: 0 is `Type::NONE`.

filament/backend/src/metal/MetalBuffer.h
bejado: I do. Objective-C guarantees that `weak` pointers are set to `nil` automatically when whatever they're referencing is deallocated. So essentially the `hashTable` simply needs to tally which pointers aren't `nil`.

pixelflinger (approved):

PR #7845, opened 2024-05-12
fix typo checking FILAMENT_ENABLE_MATDBG

pixelflinger:
this macro must be checked with #if not #ifdef


poweifeng (approved):
that's interesting.  Is there a preferred convention for filament between using the value vs. checking for the existent of the #define?

pixelflinger:
> that's interesting. Is there a preferred convention for filament between using the value vs. checking for the existent of the #define?

While fixing this I realized we do not have a well established convention; but it looks like everything that's set by cmake ends up being defined to 0 or 1, so #ifdef doesn't work.

PR #7847, opened 2024-05-13
morphing API changes

pixelflinger:
- we add support of the `offset` parameter when setting a morph target. This is made possible by using the new push-constants feature
- we deprecate the old morphing api that allowed a morph target buffer per primitive. the new API is currently implemented on top of the old one.


poweifeng (approved):

pixelflinger:
I'm converting this to draft because this breaks morphing on metal because it looks like we're running out of argument buffers.

bejado (approved):

PR #7856, opened 2024-05-14
Metal, fix callbacks being called only once

### LINKED ISSUE #7826: `setFrameCompletedCallback` is only called once, then it crashes
**Describe the bug**

We are trying to render into an `AVAssetWriter`, for that we need to know when the frame is completed.
We are trying to use `swapChain->setFrameCompletedCallback` for that purpose.
The frame callback is called once, but after that we get an assertion error:

```
Assertion failed: (mInvoker && mInvocable), function operator(), file Invocable.h, line 134.
```

**To Reproduce**

1. Set a frame completed callback using the metal backend:
```cpp
auto swapChain = _engine->createSwapChain(nativeWindow, SwapChain::CONFIG_APPLE_CVPIXELBUFFER);
  swapChain->setFrameCompletedCallback(
                                       nullptr, [](SwapChain* UTILS_NONNULL swapChain) {
                                           // note: we pass a completely empty callback, and it still crashes
                                       });
```

**Expected behavior**

The callback gets called continuously while rendering.

**Screenshots**

Screenshot 2024-05-06 at 13 30 13

**Logs**
If applicable, copy **full** logs from your console here. Please *do not*
use screenshots of logs, copy them as text, use gist or attach an *uncompressed* file.

 - OS: iOS
 - GPU: Apple GPU
 - Backend: Metal

 - Device: iPhone 13 Pro
 - OS: iOS 17.4

**Additional context**

n/a

bejado:
There was a bug where the `frameScheduledCallback` and `frameCompletedCallback` needed to be set before every frame, otherwise they would either be called only once or crash in debug builds.

Fixes #7826

### DIFF
--- a/filament/backend/CMakeLists.txt
+++ b/filament/backend/CMakeLists.txt
@@ -417,6 +417,7 @@ if (APPLE OR LINUX)
         test/test_MissingRequiredAttributes.cpp
         test/test_ReadPixels.cpp
         test/test_BufferUpdates.cpp
+        test/test_Callbacks.cpp
         test/test_MRT.cpp
         test/test_LoadImage.cpp
         test/test_StencilBuffer.cpp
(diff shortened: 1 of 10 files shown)


pixelflinger (approved):

poweifeng (approved):

mrousavy:
nice!!

PR #7857, opened 2024-05-14
Refactor stereo build flags and configs

poweifeng:
- Add option to build.sh to build for paritcular stereo techniques (default to NONE). Only applies to samples.
 - Consoldiate viewer checkbox for debugging stereo rendering
 - Add DriverConfig flag for stereoscopic type so that it can be used to determine availability of the feature and (to be completed) enable corresponding GPU features.


bejado (approved):

android/filament-android/src/main/java/com/google/android/filament/Engine.java
```diff
@@ -162,6 +162,8 @@ public enum FeatureLevel {
      * The type of technique for stereoscopic rendering
      */
     public enum StereoscopicType {
+        /** No stereoscopic rendering. */
+        NONE,
         /** Stereoscopic rendering is performed using instanced rendering technique. */
         INSTANCED,
         /** Stereoscopic rendering is performed using the multiview feature from the graphics backend. */
```
pixelflinger, line 169: there should be a not here that this value needs to match the materials used.

build.sh
```diff
@@ -64,6 +64,8 @@ function print_help {
     echo "        enabling debug paths in the backend from the build script. For example, make a"
     echo "        systrace-enabled build without directly changing #defines. Remember to add -f when"
     echo "        changing this option."
+    echo "    -S type"
+    echo "        Enable stereoscopic rendering where type is one of [instanced|multiview]."
```
pixelflinger, line 68: specify that it's for samples only

filament/backend/include/backend/Platform.h
```diff
@@ -41,6 +41,24 @@ class UTILS_PUBLIC Platform {
     struct Fence {};
     struct Stream {};
 
+    /**
+     * The type of technique for stereoscopic rendering
+     */
```
pixelflinger, line 46: a note that it needs to match the materials

filament/backend/src/metal/MetalDriver.h
```diff
@@ -17,6 +17,7 @@
 #ifndef TNT_FILAMENT_DRIVER_METALDRIVER_H
 #define TNT_FILAMENT_DRIVER_METALDRIVER_H
 
+#include "backend/DriverEnums.h"
```
pixelflinger, line 20:
```suggestion
#include <backend/DriverEnums.h>
```

pixelflinger (changes requested):

android/filament-android/src/main/java/com/google/android/filament/Engine.java
poweifeng: done

build.sh
poweifeng: done

filament/backend/include/backend/Platform.h
poweifeng: done

android/filament-android/src/main/java/com/google/android/filament/Engine.java
poweifeng: done.

filament/src/details/Skybox.cpp
```diff
@@ -142,6 +142,8 @@ FMaterial const* FSkybox::createMaterial(FEngine& engine) {
                 assert_invariant(false);
 #endif
                 break;
+            case Engine::StereoscopicType::NONE:
```
z3moon, line 145: Can a normal code flow hit here? I think this shouldn't be hit because it'll include no skybox material otherwise. Can we add an assert here? But I feel like NONE doesn't make sense in this case.

z3moon (changes requested):

filament/src/details/Skybox.cpp
poweifeng: done

pixelflinger (approved):

filament/src/details/Engine.cpp
```diff
@@ -364,6 +365,8 @@ void FEngine::init() {
                 assert_invariant(false);
 #endif
                 break;
+            case StereoscopicType::NONE:
```
z3moon, line 368: Here's one more place to handle the default materials. Please do the same approach as did for skybox here before submitting.

z3moon (approved):

filament/src/details/Engine.cpp
poweifeng: done

PR #7858, opened 2024-05-15
Implement push constants for Metal

bejado:


pixelflinger (commented):
Why write the push constant in draw2 instead of doing it right when updatePushConstant is called?

filament/backend/src/metal/MetalContext.mm
```diff
@@ -153,5 +153,61 @@ bool isInRenderPass(MetalContext* context) {
...
+    }
+    constexpr size_t PUSH_CONSTANT_SIZE_BYTES = 4;
+    constexpr size_t PUSH_CONSTANT_BUFFER_INDEX = 26;
+
+    size_t bufferSize = PUSH_CONSTANT_SIZE_BYTES * mPushConstants.size();
+    char* buffer = static_cast<char*>(malloc(bufferSize));
+    for (size_t i = 0; i < mPushConstants.size(); i++) {
+        const auto& constant = mPushConstants[i];
+        std::visit([buffer, i](auto arg) {
+                    *(decltype(arg)*)(buffer + PUSH_CONSTANT_SIZE_BYTES * i) = arg;
```
poweifeng, line 180: how do metal programs interpret bools?  This'll set a 1 in the lowest-word of the 4-bytes?

filament/backend/src/metal/MetalContext.mm
```diff
@@ -153,5 +153,61 @@ bool isInRenderPass(MetalContext* context) {
...
+void MetalPushConstantBuffer::setBytesIfDirty(
+        id<MTLCommandEncoder> encoder, ShaderStage stage) {
+    if (!mDirty) {
+        return;
+    }
+    constexpr size_t PUSH_CONSTANT_SIZE_BYTES = 4;
+    constexpr size_t PUSH_CONSTANT_BUFFER_INDEX = 26;
+
+    size_t bufferSize = PUSH_CONSTANT_SIZE_BYTES * mPushConstants.size();
+    char* buffer = static_cast<char*>(malloc(bufferSize));
```
poweifeng, line 176: can this be  a scratch buffer pre-allocated the driver/global space?

filament/backend/src/metal/MetalContext.mm
bejado: Good question, let me investigate. Maybe it's a good idea to explicitly write all the bits, setting the constant to 0x00000000 or 0x00000001 depending on the bool.

filament/backend/src/metal/MetalContext.mm
bejado: Yeah, we can probably allocate a static buffer and only use malloc if we need more space.

poweifeng: note that the max size of number of push constant if 32 (or 128 bytes) due to vk spec limit.

bejado: Ah okay that's good to know! In that case I can definitely allocate static storage space. I also just saw the `MAX_PUSH_CONSTANT_COUNT` that you added, so I'll use that.

filament/backend/src/metal/MetalContext.mm
bejado: So turns out that spirv-cross will convert the push constant `bool`s to `uint`s (probably for alignment purposes). So in that case, I definitely need to ensure I'm writing all the bytes.

pixelflinger:
> > Why write the push constant in draw2 instead of doing it right when updatePushConstant is called?
> 
> So that if you update multiple push constants before drawing, we only have to call `set{Fragment,Vertex}Bytes` once. Maybe this is premature optimization, since there's only a single push constant at the moment.
> 
> Edit: I'll also make note that Metal's API requires you to update the _entire_ buffer each time you call `set{Fragment,Vertex}Bytes`, hence the need to keep track of the push constants and "batch" updates.

Ok, it's good for now. I think we'll have to think about a better API that doesn't force us to do a check in draw2(). I really want to get to the point where draw2() is just that: a draw call.  We can revisit after my descriptor-set changes.

filament/backend/src/metal/MetalContext.mm
```diff
@@ -153,5 +153,61 @@ bool isInRenderPass(MetalContext* context) {
...
+                                                           atIndex:PUSH_CONSTANT_BUFFER_INDEX];
+            break;
+        case ShaderStage::COMPUTE:
+            [(id<MTLComputeCommandEncoder>)encoder setBytes:buffer
+                                                     length:bufferSize
+                                                    atIndex:PUSH_CONSTANT_BUFFER_INDEX];
+            break;
+    }
+
+    free(buffer);
```
pixelflinger, line 203: If the lifetime of the buffer doesn't need to extend the call to setVertexBytes and friends, then it should just be allocated on the stack, since we know it's small.

filament/backend/src/metal/MetalDriver.mm
```diff
@@ -1865,6 +1876,12 @@
             UNIFORM_BUFFER_BINDING_START, MetalBuffer::Stage::VERTEX | MetalBuffer::Stage::FRAGMENT,
             uniformsToBind, offsets, Program::UNIFORM_BINDING_COUNT);
 
+    // Update push constants.
+    for (size_t i = 0; i < Program::SHADER_TYPE_COUNT; i++) {
+        auto& pushConstants = mContext->currentPushConstants[i];
+        pushConstants.setBytesIfDirty(mContext->currentRenderPassEncoder, static_cast<ShaderStage>(i));
+    }
```
pixelflinger, line 1883: I would have a unlikley check here to skip the loop if no constant is dirty.

pixelflinger (changes requested):

pixelflinger (approved):

Issue #7622, opened 2024-03-01
Copy/Re-use animation from one instance on another one with the same skeleton


hannojg:
**Is your feature request related to a problem? Please describe.**

In my use-case I have a main character model. This model contains all animations for the character.
I now have many other models that are clothes or accessories that have the same skeleton as the character, but don't contain any animations.
When the character equips for example a hat and we run the idle animation of the character, we copy over the idle animation to the hat model. Playing both simultaneously gives the impression of the hat fitting on the character's hat, as it's following all of its animations.

The current app code is written in threejs but we want to migrate to filament, and we're wondering if we can use the same approach with filament.
Here is an example code snippet of how to do this in threejs:

```ts
// this._character being the main model with all animations and
// this._accessoryObject being the accessory model we want to copy the animations to
this._character.copyAnimationsTo(
	this._accessoryObject.animations,
	this._accessoryObject.mixer,
);

copyAnimationsTo(animationsMap: { [key: string]: THREE.AnimationAction }, mixer: THREE.AnimationMixer) {
	const animations = this._character.animations
	animations.forEach(animation => {
		const animationAction = mixer.clipAction(animation)
		animationsMap[animation.name] = animationAction
	})
}
```

**Describe the solution you'd like**

I haven't thought of any exact solution yet, as I feel it might introduce a new API (if the feature doesn't already exist?) and I don't have any specific API in mind.

Maybe something like:

```cpp
animator->addInstance(otherAsset->getInstance())
```

**Describe alternatives you've considered**

I see that the `Animator` has (internal) code i.e. `addInstance`, however thats internal and only usable with an FFilamentInstance. 

**OS and backend**

I think it's not.

hannojg:
I got it working by making a new public constructor for the Animator:

```cpp
// Animator.h
Animator(FilamentAsset *asset, FilamentInstance *instance) : Animator(downcast(asset), downcast(instance))
{
}
```

and then instantiating an animator on my own like this:

```cpp
Animator* animator = new gltfio::Animator(mainCharacterAsset, hatAsset->getInstance());
```

And then using this animator instead of `hatAsset->getInstance()->getAnimator()` to drive the animation.

I am sure the constructor was private for a reason and shouldn't be exposed 😅  But it proves it's possible with filament, which is awesome! Are you open to discuss an API to enable this feature?

hannojg:
As a side note: I tried to use `animator->addInstance` with a FilamentInstance (I changed the code for that).
[...]

PR #7859, opened 2024-05-15
feat: updateBoneMatricesForInstance

hannojg:
Added a method to control to which instance to apply the bone matrices updates of an animator.

This is particularly useful if you want to apply the animation from one master asset to other instances that are different assets (have different meshes) but share the same skeleton as the master asset. A common example for this is e.g. clothing.

In our use case we have one main character which contains all the animations. We then have clothings that are rigged to the same skeleton of the main character.
We are then able to apply the transformations of the character to the clothing assets while playing an animation.

The initial discussion has been started here:

- https://github.com/google/filament/issues/7622


libs/gltfio/include/gltfio/Animator.h
```diff
@@ -56,6 +56,18 @@ class UTILS_PUBLIC Animator {
...
+     * Updates the bone matrices of the specified instance using the state of this animatior.
+     * This is useful if you have another instance that has the same skeleton as the asset of this animator,
+     * and you wish to apply the same animation to those instances (e.g. clothing).
+     * 
+     * NOTE: In most cases, you only need to use the updateBoneMatrices() method. This method is necessary
+     * only when you need to synchronize animations across multiple instances with the same skeleton.
+     * 
+     * @param instance The instance to update.
+     */
+    void updateBoneMatricesForInstance(FilamentInstance* instance);
```
poweifeng, line 69: I feel like maybe we can formalize `addInstance` below instead of adding this method?  

We can do some validation of the the animation associated with the "new", added instance vs the one that's used to build this animator.

hannojg: I thought so as well first. However, the "problem" I found with that is that we are again creating all those channels and data for the new instance.
With the solution I have right now we don't need to recreate any data (however I admit that my solution is maybe a bit fragile as it relies on the name of the entities to synchronise entity transforms).

ninjz:
Wasn't able to get this solution working on Android. Are there any additional instructions you can provide on how this should be used?

hannojg:
> Wasn't able to get this solution working on Android. Are there any additional instructions you can provide on how this should be used?

I provided instructions here:
- https://github.com/google/filament/issues/7622#issuecomment-2138069662

hannojg:
Hey, any chance to get this one reviewed? Is there any way how i can help?

PR #7862, opened 2024-05-16
add getReasonLiteral() on TPanic

pixelflinger:
currently it only returns the format string.


libs/utils/include/utils/Panic.h
```diff
@@ -286,6 +286,11 @@ class UTILS_PUBLIC Panic {
      */
     virtual const char* getReason() const noexcept = 0;
 
+    /**
+     * You know who you are.
+     */
```
bejado, line 291: 😆 maybe mark this for "internal use only"?

PR #7869, opened 2024-05-17
A new stream-based Panic API

pixelflinger:
going forward, instead of using the printf style syntax for panics we use the c++ stream syntax

The new macros that replace ASSERT_*CONDITON are

FILAMENT_CHECK_PRECONDITON
FILAMENT_CHECK_POSTCONDITION
FILAMENT_CHECK_ARITIHMETIC

Example usage:

FILAMENT_CHECK_PRECONDITON(condition) << "Message";

The effect is exactly the same as before, but it'll let clients use Abseil's logging facilities.

### DIFF
--- a/filament/backend/include/private/backend/HandleAllocator.h
+++ b/filament/backend/include/private/backend/HandleAllocator.h
@@ -173,8 +173,8 @@ public:
                 uint8_t const age = (tag & HANDLE_AGE_MASK) >> HANDLE_AGE_SHIFT;
                 auto const pNode = static_cast<typename Allocator::Node*>(p);
                 uint8_t const expectedAge = pNode[-1].age;
-                ASSERT_POSTCONDITION(expectedAge == age,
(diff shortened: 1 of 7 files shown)


bejado (approved):
LGTM. Just an observation: if a user does this:

```
FILAMENT_CHECK_POSTCONDITION(false);
```

Then we get an empty "reason" string. I wonder if there's a way we can require users to provide a message (if we even want to require it) at compile time. Or, maybe we could provide a default message.

poweifeng (approved):

pixelflinger:
> LGTM. Just an observation: if a user does this:
> 
> ```
> FILAMENT_CHECK_POSTCONDITION(false);
> ```
> 
> Then we get an empty "reason" string. I wonder if there's a way we can require users to provide a message (if we even want to require it) at compile time. Or, maybe we could provide a default message.

interesting idea. I think we probably can provide the stringified condition. I will give it a shot.

PR #7888, opened 2024-05-29
Acquire a mutex before releasing CAMetalDrawables on main thread.

ullerrm:
PresentDrawable was moved to main thread by default in google#7535 and stopped most crashes when a drawable is released. But there still appears to be crashes if a drawable is released on main thread at the same time that -nextDrawable is called from the Filament render thread. (It's likely that the drawable pool in CAMetalLayer is completely non-thread-safe.)

So, add a mutex to the swapchain and always acquire it before creating or releasing a CAMetalDrawable.


filament/backend/src/metal/MetalHandles.mm
```diff
@@ -256,9 +268,10 @@ static inline MTLTextureUsage getMetalTextureUsage(TextureUsage usage) {
     PresentDrawableData(const PresentDrawableData&) = delete;
     PresentDrawableData& operator=(const PresentDrawableData&) = delete;
 
-    static PresentDrawableData* create(id<CAMetalDrawable> drawable, MetalDriver* driver) {
+    static PresentDrawableData* create(id<CAMetalDrawable> drawable,
+            std::mutex* drawableMutex, MetalDriver* driver) {
```
pixelflinger, line 272: I really don't like that, passing the lock in here. but I'll let @bejado decide.

filament/backend/src/metal/MetalHandles.mm
```diff
@@ -143,6 +143,10 @@ static inline MTLTextureUsage getMetalTextureUsage(TextureUsage usage) {
     return (NSUInteger) layer.drawableSize.height;
 }
 
+#ifndef FILAMENT_LOCK_METAL_DRAWABLE_POOL
+#define FILAMENT_LOCK_METAL_DRAWABLE_POOL 1
+#endif
```
pixelflinger, line 148: do we need to make this optional?

filament/backend/src/metal/MetalHandles.mm
```diff
@@ -277,16 +290,27 @@ static void maybePresentAndDestroyAsync(PresentDrawableData* that, bool shouldPr
...
 
 private:
-    PresentDrawableData(id<CAMetalDrawable> drawable, MetalDriver* driver)
-        : mDrawable(drawable), mDriver(driver) {}
+    PresentDrawableData(id<CAMetalDrawable> drawable, std::mutex* drawableMutex,
+            MetalDriver* driver)
+        : mDrawable(drawable), mDrawableMutex(drawableMutex), mDriver(driver) {}
 
     static void cleanupAndDestroy(PresentDrawableData *that) {
+        std::optional<std::lock_guard<std::mutex>> opt_lock;
```
pixelflinger, line 298: wrong naming convention, use `optionalLock`

pixelflinger (changes requested):

filament/backend/src/metal/MetalHandles.mm
ullerrm: Not a huge fan of it either, but I didn't see a cleaner way. PresentCallables can potentially stay alive for up to a frame, if the customer's using it to synchronize a Filament frame with a UIKit CATransaction.

One alternative would be making mDrawableMutex a `std::shared_ptr<std::mutex>` and passing it in that form -- that would divorce lifetimes and allow a PresentCallable to cleanly outlive MetalSwapChain if needed. But it seemed wasteful to me to move that mutex into a separate heap allocation and do a bunch of atomics just to get access to another atomic.

(There's also the option of cutting the Gordian Knot, by dispatch_sync()ing from render thread to main thread in order to get the next drawable. But that seems even worse.)

filament/backend/src/metal/MetalHandles.mm
ullerrm: Was just following the same pattern as the release-on-main feature. It doesn't strictly need to be optional, I think. Will remove.

filament/backend/src/metal/MetalHandles.mm
ullerrm: Done. (If we're not making this an optional feature, then the optional isn't even needed.)

bejado (approved):

pixelflinger (approved):

PR #7892, opened 2024-05-30
Multiview support for vulkan

GrantComm:
FIXES=[332425392]

Adds multiview support for vulkan. This is done by adding a layerCount to the renderTarget, which is used to determine if multiview is available and being used in the current renderpass.


filament/backend/src/vulkan/VulkanContext.cpp
```diff
@@ -73,7 +73,7 @@ VkImageSubresourceRange VulkanAttachment::getSubresourceRange() const {
             .baseMipLevel = uint32_t(level),
             .levelCount = 1,
             .baseArrayLayer = uint32_t(layer),
-            .layerCount = 1,
+            .layerCount = uint32_t(viewCount) > 0 ? uint32_t(viewCount) : texture->getLayerCount() > 1 ? texture->getLayerCount() : 1,
```
phoenixxxx, line 76: It's a neat pick but I don't think the first cast is necessary `viewCount > 0 ?`. I also think that it would be cleaner and more readable if we just moved the whole logic for layerCount comewhere else and then here just do `.layerCount = layerCount`

filament/backend/src/vulkan/VulkanDriver.cpp
```diff
@@ -907,7 +913,7 @@ bool VulkanDriver::isStereoSupported() {
         case backend::StereoscopicType::INSTANCED:
             return mContext.isClipDistanceSupported();
         case backend::StereoscopicType::MULTIVIEW:
-            // TODO: implement multiview feature in Vulkan.
+            return true;
```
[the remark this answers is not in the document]
phoenixxxx: Also generally we should follow the convention from the file, like do something like:

`bool VulkanDriver::isMultiviewSupported() {`

build.sh
```diff
@@ -168,6 +168,8 @@ VULKAN_ANDROID_GRADLE_OPTION=""
 
 EGL_ON_LINUX_OPTION="-DFILAMENT_SUPPORTS_EGL_ON_LINUX=OFF"
 
+MULTIVIEW_OPTION="-DFILAMENT_SAMPLES_STEREO_TYPE=multiview"
```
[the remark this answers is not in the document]
pixelflinger: yes that's for our samples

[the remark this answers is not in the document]
mdagois: Should it be part of the PR?

filament/backend/src/vulkan/VulkanDriver.cpp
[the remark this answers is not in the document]
mdagois: > Why would performance be impacted if we don't use the extension? 

I am pretty sure I read that it could affect shader compilation as a whole, but, of course, I cannot find the reference anymore... I am asking Vulkan buffs right now.

> I don't I want a compile-time switch for stuff like this.

Ok, it is a misunderstanding then. I thought that's what the Filament team wanted.
It is easier for us to just activate the extension all the time (when it is present that is).
Assuming we find it has a performance impact even if not used, is there some kind of runtime config you'd want us to implement to turn multiview off?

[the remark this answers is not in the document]
poweifeng: Actually, we do want it to be based on availability and configuration.  The config, however, is not a build flag but [this](https://github.com/google/filament/blob/main/filament/include/filament/Engine.h#L318).  
See my attempt here to enable clip space only when necessary: https://github.com/google/filament/commit/450644ccd557bae704fb99e4ae31bb919b65c701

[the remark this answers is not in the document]
mdagois: Got it, thanks for the reference links!

BTW, it seems the extension is enabled in Angle unconditionally, so it should not have any negative impact when unused.

filament/backend/src/vulkan/VulkanTexture.cpp
```diff
@@ -387,8 +390,10 @@ void VulkanTexture::setPrimaryRange(uint32_t minMiplevel, uint32_t maxMiplevel)
 }
 
 VkImageView VulkanTexture::getAttachmentView(VkImageSubresourceRange range) {
```
poweifeng, line 392: We should probably add getMultiviewAttachmentView instead of overloading this one.

filament/backend/src/vulkan/VulkanContext.h
```diff
@@ -43,6 +43,7 @@ struct VulkanCommandBuffer;
 struct VulkanAttachment {
     VulkanTexture* texture = nullptr;
     uint8_t level = 0;
+    uint8_t baseViewIndex = 0;
     uint16_t layer = 0;
```
poweifeng, line 48: we do need a layerCount here unfortunately.  I might think about how to refactor this later, but for now, we need it to know the correct VkImageSubresourceRange

filament/backend/src/vulkan/VulkanFboCache.cpp
```diff
@@ -288,16 +311,23 @@ VkRenderPass VulkanFboCache::getRenderPass(RenderPassKey config) noexcept {
...
         const bool discardStart = any(config.discardStart & TargetBufferFlags::DEPTH);
         const bool discardEnd = any(config.discardEnd & TargetBufferFlags::DEPTH);
+        const VkAttachmentLoadOp loadOp = clear ? kClear : (discardStart ? kDontCare : kKeep);
         depthAttachmentRef.layout = imgutil::getVkLayout(VulkanLayout::DEPTH_ATTACHMENT);
         depthAttachmentRef.attachment = attachmentIndex;
+        VkImageLayout initialLayout = imgutil::getVkLayout(config.initialDepthLayout);
+        
+        if (loadOp == VK_ATTACHMENT_LOAD_OP_LOAD && imgutil::getVkLayout(config.initialDepthLayout) == VK_IMAGE_LAYOUT_UNDEFINED) {
+            initialLayout = VK_IMAGE_LAYOUT_DEPTH_STENCIL_ATTACHMENT_OPTIMAL;
+        }
```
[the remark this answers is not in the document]
GrantComm: http://b/347363913

filament/backend/src/vulkan/VulkanDriver.cpp
```diff
@@ -1277,10 +1281,14 @@ void VulkanDriver::beginRenderPass(Handle<HwRenderTarget> rth, const RenderPassP
...
         .samples = rt->getSamples(),
         .subpassMask = uint8_t(params.subpassMask),
+        .viewCount = rt->getLayerCount(),
     };
+
     for (int i = 0; i < MRT::MAX_SUPPORTED_RENDER_TARGET_COUNT; i++) {
         const VulkanAttachment& info = rt->getColor(i);
+
         if (info.texture) {
+            rpkey.viewCount = info.texture->getLayerCount();
```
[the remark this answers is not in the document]
GrantComm: This is needed because there are times when the renderTarget's layer count is 2. But the underlying vulkan texture's layer count is 1. If there is a vulkan texture, then the viewcount needs to be set to the value from the texture. Without it validation layers errors occur:

VULKAN ERROR: (Validation) Validation Error: [ VUID-VkFramebufferCreateInfo-renderPass-04536 ] | MessageID = 0x5e2eb360 | vkCreateFramebuffer(): pCreateInfo->pAttachments[1] has a layer count (1) less than or equal to the highest bit in the view mask (1) of subpass 0. The Vulkan spec states: If renderPass was specified with non-zero view masks, each element of pAttachments that is used as an input, color, resolve, or depth/stencil attachment by renderPass must have a layerCount greater than the index of the most significant bit set in any of those view masks (https://vulkan.lunarg.com/doc/view/1.3.268.0/linux/1.3-extensions/vkspec.html#VUID-VkFramebufferCreateInfo-renderPass-04536)

and a segmentation fault crashes the program.

One caveat: If there were ever a case when there is a texture of samplerType::SAMPLER_2D_ARRAY that is not being used in a multiview configuration, it would trigger the multiview structure being added to the renderpass. To get around that. I could pass the value of mContext.isMultiviewEnabled() as part of the RenderPassKey to add an additional check before chaining multiviewCreateInfo to the renderpass.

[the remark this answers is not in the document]
poweifeng: the texture should not have a layer count. But, for now, I propose both RenderTarget and Attachment will both have layerCounts.  You can do an assert here to verify that the render target and attachment counts are the same.

poweifeng:
> FIXES=[332425392]

Please add more detail to the "commit message".  You can modify your commit message with `git commit --amend`

filament/backend/src/vulkan/VulkanDriver.cpp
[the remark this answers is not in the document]
GrantComm: Sorry, not vulkan texture, but underlying vulkan image in the texture. When the gtlf_viewer application is not in stereo mode, the layercount of the underlying vulkan image is 1 but the layercount for the vulkan attachment and vulkan rendertarget are both 2. All three (vulkan texture's vkimage, vulkan attachment, and vulkan rendertarget) are all 2, only when the application is in stereo mode. 

Without the current use of the layercount from the vulkan texture's vkImage, the validation error listed above is raised and a crash occurs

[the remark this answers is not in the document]
poweifeng: I think there might be a bug here.  If the rendertarget has 2 layers, then the backing image of the attachments should also have at least 2 layers.  If you see the rendertarget somehow has greater number of layers than any of the attachments, then it's a bug IIUC.

filament/backend/src/vulkan/VulkanDriver.cpp
```diff
@@ -595,6 +595,8 @@ void VulkanDriver::createRenderTargetR(Handle<HwRenderTarget> rth,
             colorTargets[i] = {
```
poweifeng, line 595: can you add a 
`assert_invariant(color[i].layerCount == layerCount);` here?

filament/backend/src/vulkan/VulkanDriver.cpp
```diff
@@ -609,6 +611,8 @@ void VulkanDriver::createRenderTargetR(Handle<HwRenderTarget> rth,
         depthStencil[0] = {
             .texture = mResourceAllocator.handle_cast<VulkanTexture*>(depth.handle),
             .level = depth.level,
+            .baseViewIndex = depth.baseViewIndex,
+            .layerCount = layerCount,
```
poweifeng, line 615: Let's also add an `assert_invariant` here to make sure that layerCount matches `depth.layerCount`

filament/backend/src/vulkan/VulkanDriver.cpp
```diff
@@ -621,6 +625,8 @@ void VulkanDriver::createRenderTargetR(Handle<HwRenderTarget> rth,
         depthStencil[1] = {
             .texture = mResourceAllocator.handle_cast<VulkanTexture*>(stencil.handle),
             .level = stencil.level,
+            .baseViewIndex = stencil.baseViewIndex,
```
poweifeng, line 628: same here.

filament/backend/src/vulkan/VulkanFboCache.cpp
```diff
@@ -183,9 +184,28 @@ VkRenderPass VulkanFboCache::getRenderPass(RenderPassKey config) noexcept {
         .subpassCount = hasSubpasses ? 2u : 1u,
         .pSubpasses = subpasses,
         .dependencyCount = hasSubpasses ? 1u : 0u,
-        .pDependencies = dependencies
+        .pDependencies = dependencies,
     };
 
+    VkRenderPassMultiviewCreateInfo multiviewCreateInfo = {};
+    uint32_t subpassViewMask = (1 << config.viewCount) - 1;
```
poweifeng, line 191:
```suggestion
    uint32_t const subpassViewMask = (1 << config.viewCount) - 1;
```

filament/backend/src/vulkan/VulkanFboCache.cpp
```diff
@@ -183,9 +184,28 @@ VkRenderPass VulkanFboCache::getRenderPass(RenderPassKey config) noexcept {
...
+    // Prepare a view mask array for the maximum number of subpasses. All subpasses have all views
+    // activated.
+    uint32_t viewMasks[2] = {subpassViewMask, subpassViewMask};
```
poweifeng, line 194:
```suggestion
    uint32_t const viewMasks[2] = {subpassViewMask, subpassViewMask};
```

poweifeng (commented):
this looks really good and really close.  Just some leftover nits.

filament/backend/src/vulkan/VulkanDriver.cpp
GrantComm: color[i] is an MRT. depth and stencil are TargetBufferInfo. Neither currently have "layerCount" as a member. Would you want those classes to have a layerCount as well?

poweifeng: ah, no I misread it. It's fine without the assert.  Same with the others.

filament/backend/src/vulkan/VulkanTexture.cpp
```diff
@@ -66,8 +66,8 @@ VulkanTexture::VulkanTexture(VkDevice device, VkPhysicalDevice physicalDevice,
       mStagePool(stagePool),
       mDevice(device),
       mAllocator(allocator),
-      mCommands(commands) {
-
+      mCommands(commands)
+      {
```
poweifeng, line 70:
```suggestion
      mCommands(commands) {
```

filament/backend/src/vulkan/VulkanFboCache.cpp
```diff
@@ -183,9 +184,28 @@ VkRenderPass VulkanFboCache::getRenderPass(RenderPassKey config) noexcept {
...
+      multiviewCreateInfo.sType = VK_STRUCTURE_TYPE_RENDER_PASS_MULTIVIEW_CREATE_INFO;
+      multiviewCreateInfo.pNext = nullptr;
+      multiviewCreateInfo.subpassCount = hasSubpasses ? 2u : 1u;
+      multiviewCreateInfo.pViewMasks = viewMasks;
+      multiviewCreateInfo.dependencyCount = 0;
+      multiviewCreateInfo.pViewOffsets = nullptr;
+      multiviewCreateInfo.correlationMaskCount = 1;
+      multiviewCreateInfo.pCorrelationMasks = &subpassViewMask;
+
+      renderPassInfo.pNext = &multiviewCreateInfo;
```
poweifeng, line 206: indent 4 here

filament/backend/src/vulkan/VulkanFboCache.cpp
```diff
@@ -288,16 +308,19 @@ VkRenderPass VulkanFboCache::getRenderPass(RenderPassKey config) noexcept {
         const bool clear = any(config.clear & TargetBufferFlags::DEPTH);
         const bool discardStart = any(config.discardStart & TargetBufferFlags::DEPTH);
         const bool discardEnd = any(config.discardEnd & TargetBufferFlags::DEPTH);
+        const VkAttachmentLoadOp loadOp = clear ? kClear : (discardStart ? kDontCare : kKeep);
         depthAttachmentRef.layout = imgutil::getVkLayout(VulkanLayout::DEPTH_ATTACHMENT);
         depthAttachmentRef.attachment = attachmentIndex;
+        VkImageLayout initialLayout = imgutil::getVkLayout(config.initialDepthLayout);
```
poweifeng, line 314: why make these changes here? no logic seems to have changed.

GrantComm: Removed. Sorry, was part of an earlier  revision.

filament/backend/src/vulkan/VulkanDriver.cpp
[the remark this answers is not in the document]
GrantComm: Yes, bug fix is in this pr: https://github.com/google/filament/pull/7987

filament/backend/src/vulkan/VulkanHandles.h
```diff
@@ -319,6 +319,7 @@ struct VulkanRenderTarget : private HwRenderTarget, VulkanResource {
     VulkanAttachment& getMsaaDepth();
     uint8_t getColorTargetCount(const VulkanRenderPass& pass) const;
     uint8_t getSamples() const { return mSamples; }
+    uint8_t getLayerCount() { return mLayerCount; }
```
poweifeng, line 322:
```suggestion
    uint8_t getLayerCount() const { return mLayerCount; }
```

filament/backend/src/vulkan/VulkanDriver.cpp
```diff
@@ -1267,6 +1273,8 @@ void VulkanDriver::beginRenderPass(Handle<HwRenderTarget> rth, const RenderPassP
         }
     }
 
+    uint8_t renderTargetLayerCount = rt->getLayerCount();
```
poweifeng, line 1276:
```suggestion
    uint8_t const renderTargetLayerCount = rt->getLayerCount();
```

poweifeng (approved):
LGTM except for two minor things. 

Can you also make sure @z3moon signs off on the changes to
filament/src/RendererUtils.cpp
filament/src/fg/FrameGraphRenderPass.h

? 

Thank you.

z3moon (approved):

PR #7894, opened 2024-06-01
Attach the current thread to the JVM before calling jni routines.

vauradkar:


romainguy:
Any code path using JNI should already be calling getEnvironment(). What issue are you trying to fix here?

vauradkar:
On android, I am using filament along with another Google internal library to build headless view. The view is not a part of android app. This causes a segv during "Setup" that looks something like 

```
JavaBBinder::onTransact(unsigned int, android::Parcel const&, android::Parcel*, unsigned int)+591)
android::BBinder::transact(unsigned int, android::Parcel const&, android::Parcel*, unsigned int)+238)
android::os::BpHintManager::getHintSessionPreferredRate(long*)+471)
APerformanceHintManager::create(android::sp<android::os::IHintManager>)+269)
APerformanceHintManager::getInstance()+184)
utils::PerformanceHintManager::PerformanceHintManager()+29)
filament::backend::PlatformEGLAndroid::PlatformEGLAndroid()+60)
filament::backend::PlatformFactory::create(filament::backend::Backend*)+122)
filament::FEngine::loop()+27)
```

romainguy:
The proper fix is to have the performance manager get the environment to make sure the thread is properly setup.

pixelflinger:
I'm closing this PR because it is replaced with https://github.com/google/filament/pull/7896

PR #7899, opened 2024-06-03
keep the normal transforms in float during skinning

pixelflinger:


bejado (approved):
Bump material version?

poweifeng (approved):

PR #7902, opened 2024-06-04
Update Android dependencies to latest

romainguy:


poweifeng (approved):

hannojg:
Does this force us to build our apps with the latest NDK version v27 when building our apps?

PR #7913, opened 2024-06-07
add debug option to disable sub-passes

pixelflinger:


filament/src/details/Renderer.cpp
```diff
@@ -596,7 +598,8 @@ void FRenderer::renderJob(RootArenaScope& rootArenaScope, FView& view) {
                     hasColorGrading &&
                     msaaSampleCount <= 1 &&
                     !bloomOptions.enabled && !dofOptions.enabled && !taaOptions.enabled &&
-                    driver.isFrameBufferFetchSupported(),
+                    driver.isFrameBufferFetchSupported() &&
+                    !engine.debug.renderer.disable_subpasses,
```
poweifeng, line 602: I think `isFrameBufferFetchMultiSampleSupported` below is also related

poweifeng (approved):

filament/src/details/Renderer.cpp
pixelflinger: done

PR #7919, opened 2024-06-12
Add ability to play animations on a separate instance

mentholyspirit:
related to https://github.com/google/filament/issues/7622

### DIFF
--- a/android/gltfio-android/src/main/java/com/google/android/filament/gltfio/Animator.java
+++ b/android/gltfio-android/src/main/java/com/google/android/filament/gltfio/Animator.java
@@ -37,28 +37,26 @@ import java.nio.Buffer;
  */
 public class Animator {
     private long mNativeObject;
-    private Boolean mIsOwner = false;
+    private boolean mIsOwner = false;
 
     Animator(long nativeObject) {
         mNativeObject = nativeObject;
     }
 
+    /**
+     * Creates an Animator that can manipulate animations in the provided FilamentAsset.
+     * <p>
+     * <strong>Important:</strong> This Animator manages native resources that must be 
+     * explicitly freed by calling {@link #destroy()} when no longer needed. Failing to 
+     * call {@link #destroy()} will result in native memory leaks.
+     * </p>
+     *
+     * @param asset The FilamentAsset containing the animations
+     * @param instance The FilamentInstance to animate
+     */
     public Animator(FilamentAsset asset, FilamentInstance instance) {
         mNativeObject = nCreateAnimatorFromAssetAndInstance(asset.getNativeObject(), instance.getNativeObject());
         mIsOwner = true;
-    }
-    
-    @Override
-    public void finalize() {
-        try {
-            super.finalize();
-        } catch (Throwable t) { // Ignore
-        } finally {
-            if (mIsOwner) {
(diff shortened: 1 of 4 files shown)


pixelflinger:
@mentholyspirit could sigh the CLA so we could submit the change eventually?

mentholyspirit:
Signed the CLA

libs/gltfio/include/gltfio/Animator.h
```diff
@@ -106,13 +106,17 @@ class UTILS_PUBLIC Animator {
...
     Animator(const Animator& animator) = delete;
     Animator(Animator&& animator) = delete;
     Animator& operator=(const Animator&) = delete;
 
     AnimatorImpl* mImpl;
+public:
+    Animator(FilamentAsset *asset, FilamentInstance *instance) : Animator(reinterpret_cast<FFilamentAsset*>(asset), reinterpret_cast<FFilamentInstance*>(instance))
+    {
+    }
+    ~Animator();
```
pixelflinger, line 119: move this to the existing `public` section and do not inline the implementation. You also need to document what this ctor does, how to use it and explain that in this case the caller must `delete` it. You also need to explain the expected lifetime of the `asset` and the `instance` with respect to the lifetime of the Animator.

Because this constructor will be confusing you *must* explain in the documentation that the normal way of getting an Animator is to call `FilamentInstance::getAnimator()` and that this new ctor is for a specific use case.

Documentation is everything here.

android/gltfio-android/src/main/java/com/google/android/filament/gltfio/Animator.java
```diff
@@ -37,11 +37,30 @@
...
+        try {
+            super.finalize();
+        } catch (Throwable t) { // Ignore
+        } finally {
+            if (mIsOwner) {
+                nDestroyAnimator(mNativeObject);
+                mNativeObject = 0;
+            }
+        }
+    }
```
pixelflinger, line 62: I don't like this -- I think we can't rely on finalizers from being called. I'm not sure what the correct pattern would be here. @romainguy what do you think?

pixelflinger (changes requested):
new ctor should not be inline and should be clearly documented.
lifetime management of the Animator on the java side seems sketchy to me, I'll defer to @romainguy here.

android/gltfio-android/src/main/java/com/google/android/filament/gltfio/Animator.java
```diff
@@ -37,11 +37,30 @@
  */
 public class Animator {
     private long mNativeObject;
+    private Boolean mIsOwner = false;
```
romainguy, line 40: This should be lower case `boolean`, otherwise it's an object.

android/gltfio-android/src/main/java/com/google/android/filament/gltfio/Animator.java
romainguy: Finalizers are unfortunately not reliable (and at best take a lot of time to be invoked). We did use them in a couple of places but it's better to have a manual destroy().

romainguy (changes requested):

poweifeng (approved):

PR #7926, opened 2024-06-13
Metal: use a bump allocator for vertex and index uploads

bejado:
This switches staging buffer allocation for vertex and index data to a linear bump allocation scheme instead of using a pool of buffers. The idea is that for certain apps, vertex uploads tend to be smaller and more spiky. Batching the uploads into a larger shared buffer reduces the number and frequency of buffers we need to request from Metal.

### DIFF
--- a/filament/backend/src/metal/MetalBufferPool.mm
+++ b/filament/backend/src/metal/MetalBufferPool.mm
@@ -116,5 +116,38 @@ void MetalBufferPool::reset() noexcept {
     mFreeStages.clear();
 }
 
+MetalStagingAllocator::MetalStagingAllocator(id<MTLDevice> device, size_t capacity)
+    : mDevice(device), mCapacity(capacity) {
+    if (mCapacity > 0) {
+        mCurrentUploadBuffer = { [device newBufferWithLength:capacity options:MTLStorageModeShared],
+            TrackedMetalBuffer::Type::STAGING };
+    }
+}
+
+std::pair<id<MTLBuffer>, size_t> MetalStagingAllocator::allocateStagingArea(size_t size) {
+    if (size == 0) {
+        return { nil, 0 };
+    }
+    if (size > mCapacity) {
+        return { [mDevice newBufferWithLength:size options:MTLStorageModeShared], 0 };
+    }
+    assert_invariant(mCurrentUploadBuffer);
+
+    // Align the head to a 4-byte boundary.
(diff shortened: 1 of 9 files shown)


filament/backend/src/metal/MetalBufferPool.mm
```diff
@@ -116,5 +116,38 @@
...
+
+    if (UTILS_LIKELY(mHead + size <= mCapacity)) {
+        const size_t oldHead = mHead;
+        mHead += size;
+        return { mCurrentUploadBuffer.get(), oldHead };
+    }
+
+    // We're finished with the current allocation.
+    mCurrentUploadBuffer = { [mDevice newBufferWithLength:mCapacity options:MTLStorageModeShared],
+        TrackedMetalBuffer::Type::STAGING };
```
pixelflinger, line 147: question... instead of systematically allocating a new buffer, could we reuse a previous one once it's freed?

- you start with a buffer
- when you need a new one you check if there is one avail, if not you allocate a new one
- when a buffer loses all its reference you either keep it for reuse, or, if there is already one you delete it

it's basically a pool of 1 buffer. But it might reduce the churn a lot.

pixelflinger (changes requested):
mostly minor comments and questions

filament/backend/src/metal/MetalBufferPool.mm
bejado: I thought about this, and while I agree it might be smarter, it also requires more machinery to keep track of when a buffer is no longer in use (probably requiring us to use fences or hold a lock). I can explore this more in the future, but I like the simplicity of the current approach at least for now.

filament/backend/src/metal/MetalBufferPool.mm
```diff
@@ -116,5 +116,38 @@
...
+        return { [mDevice newBufferWithLength:size options:MTLStorageModeShared], 0 };
+    }
+    assert_invariant(mCurrentUploadBuffer);
+
+    // Align the head to a 4-byte boundary.
+    mHead = (mHead + 3) & ~3;
+
+    if (UTILS_LIKELY(mHead + size <= mCapacity)) {
+        const size_t oldHead = mHead;
+        mHead += size;
```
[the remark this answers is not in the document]
bejado: Yes, `copyIntoBuffer` is always called from the driver thread.

pixelflinger (approved):

PR #7928, opened 2024-06-15
Add ability to disable panning in the OrbitManipulator

mentholyspirit:
It's useful when you want to have the ability to rotate the camera, but want to limit panning.

### DIFF
--- a/android/filament-utils-android/src/main/java/com/google/android/filament/utils/Manipulator.java
+++ b/android/filament-utils-android/src/main/java/com/google/android/filament/utils/Manipulator.java
@@ -280,8 +280,8 @@ public class Manipulator {
          * @return this <code>Builder</code> object for chaining calls
          */
         @NonNull
-        public Builder panDisabled(Boolean enabled) {
-            nBuilderpanDisabled(mNativeBuilder, enabled);
+        public Builder panning(Boolean enabled) {
+            nBuilderpanning(mNativeBuilder, enabled);
             return this;
         }
 
@@ -494,7 +494,7 @@ public class Manipulator {
     private static native void nBuilderFlightPanSpeed(long nativeBuilder, float x, float y);
     private static native void nBuilderFlightMoveDamping(long nativeBuilder, float damping);
     private static native void nBuilderGroundPlane(long nativeBuilder, float a, float b, float c, float d);
-    private static native void nBuilderpanDisabled(long nativeBuilder, Boolean enabled);
+    private static native void nBuilderPanning(long nativeBuilder, Boolean enabled);
     private static native long nBuilderBuild(long nativeBuilder, int mode);
 
     private static native void nDestroyManipulator(long nativeManip);
--- a/android/filament-utils-android/src/main/java/com/google/android/filament/utils/Manipulator.java
+++ b/android/filament-utils-android/src/main/java/com/google/android/filament/utils/Manipulator.java
@@ -281,7 +281,7 @@ public class Manipulator {
          */
         @NonNull
         public Builder panning(Boolean enabled) {
-            nBuilderpanning(mNativeBuilder, enabled);
+            nBuilderPanning(mNativeBuilder, enabled);
             return this;
         }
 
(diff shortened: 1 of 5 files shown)


libs/camutils/include/camutils/Manipulator.h
```diff
@@ -143,6 +144,8 @@ class CAMUTILS_PUBLIC Manipulator {
         Builder& groundPlane(FLOAT a, FLOAT b, FLOAT c, FLOAT d);  //! Plane equation used as a raycast fallback
         Builder& raycastCallback(RayCallback cb, void* userdata);  //! Raycast function for accurate grab-and-pan
 
+        Builder& panDisabled(bool enabled);  //! Sets whether panning with two fingers is enabled
```
poweifeng, line 147: I'd remove the "two fingers" part, which only applies to mobile.

poweifeng (approved):

libs/camutils/include/camutils/Manipulator.h
```diff
@@ -143,6 +144,8 @@ class CAMUTILS_PUBLIC Manipulator {
         Builder& groundPlane(FLOAT a, FLOAT b, FLOAT c, FLOAT d);  //! Plane equation used as a raycast fallback
         Builder& raycastCallback(RayCallback cb, void* userdata);  //! Raycast function for accurate grab-and-pan
 
+        Builder& panDisabled(bool enabled);  //! Sets whether panning is enabled
```
pixelflinger, line 147: rename to `panning`

android/filament-utils-android/src/main/java/com/google/android/filament/utils/Manipulator.java
```diff
@@ -274,6 +274,17 @@ public Builder groundPlane(float a, float b, float c, float d) {
             return this;
         }
 
+        /**
+         * Sets whether panning is enabled in the manipulator.
+         *
+         * @return this <code>Builder</code> object for chaining calls
+         */
+        @NonNull
+        public Builder panDisabled(Boolean enabled) {
```
pixelflinger, line 283: rename to `panning`

libs/camutils/include/camutils/Manipulator.h
```diff
@@ -109,6 +109,7 @@ class CAMUTILS_PUBLIC Manipulator {
         vec4 groundPlane;
         RayCallback raycastCallback;
         void* raycastUserdata;
+        bool panDisabled;
```
pixelflinger, line 112: rename to `panning`

libs/camutils/src/Manipulator.cpp
```diff
@@ -144,6 +144,14 @@ Manipulator<FLOAT>::Builder& Manipulator<FLOAT>::Builder::raycastCallback(RayCal
     return *this;
 }
 
+
+
+template <typename FLOAT> typename
+Manipulator<FLOAT>::Builder& Manipulator<FLOAT>::Builder::panDisabled(bool enabled) {
+    details.panDisabled = enabled;
```
pixelflinger, line 151: this is *very* confusing. It should all be done in terms of "panning being enabled or not".

pixelflinger (changes requested):
rename `panDisabled`  to `panning(bool) where `true` means panning enabled and `false` means panning disabled. Make the default be `true`.

mentholyspirit:
@pixelflinger The problem then comes when someone tries to use it like this:
`Config config = {};
config.upVector = vec2(1.f, 0.f);
mainpulator.setProperties(config)`
Then `panning` will be set to false, which is not the expected behaviour.

pixelflinger:
> @pixelflinger The problem then comes when someone tries to use it like this: `Config config = {}; config.upVector = vec2(1.f, 0.f); mainpulator.setProperties(config)` Then `panning` will be set to false, which is not the expected behaviour.

Why can't we set `panning` to `true` by default in `Config`?

pixelflinger (approved):

pixelflinger:
@mentholyspirit the Android presubmit is failing.

PR #7933, opened 2024-06-20
Fix memory leak of asset loading for gltf_viewer

z3moon:
When a user drags and drops a gltf file to the gltf_viewer window, it loads the new asset.

While this happens, the function `checkAsset` tries parsing the gltf file, but it doesn't free it. Fix this.

### DIFF
--- a/samples/gltf_viewer.cpp
+++ b/samples/gltf_viewer.cpp
@@ -549,8 +549,9 @@ static bool checkAsset(const utils::Path& filename) {
 
     // Parse the glTF file and create Filament entities.
     cgltf_options options{};
-    cgltf_data* sourceAsset;
+    cgltf_data* sourceAsset = nullptr;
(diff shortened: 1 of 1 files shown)


samples/gltf_viewer.cpp
```diff
@@ -549,8 +549,9 @@ static bool checkAsset(const utils::Path& filename) {
 
     // Parse the glTF file and create Filament entities.
```
pixelflinger, line 550: looks like this comment is wrong too, right? It's not actually creating Filament entities here.

pixelflinger (changes requested):
looks good, but maybe we should also fix the comment that seems wrong to me.

samples/gltf_viewer.cpp
z3moon: Yeah, it seems like this function used to behave like what the comment says and was split into pieces later. Let me correct the comment.

pixelflinger (approved):

PR #7935, opened 2024-06-20
a convenience to get the current time

pixelflinger:
BUGS=[343764098]


pixelflinger:
this actually doesn't address the issue of b/343764098 -- closing.

PR #7937, opened 2024-06-20
add minimal support for Texture Views in the backend

pixelflinger:
Texture Views are currently limited to the min/max lod, and it's not allowed to have several views of the same texture in a single shader. Only sampleable textures can have a texture view.

In this CL:

metal: skeleton implementation

vulkan: skeleton implementation

opengl: implementation should be functional. the min/max level is now
        set when we bind a sampler group.
        When we create a view we simply duplicate the texture handle
        and update its min/max lod.  We also now keep a ref count of
        the open gl texture name so that we only destroy it when
        no more references (i.e. views) are present. The counter is
        implemented as a Handle<> so it can be shared efficiently.


filament/src/details/Texture.cpp
```diff
@@ -465,7 +472,44 @@ void FTexture::generateMipmaps(FEngine& engine) const noexcept {
...
+    // TODO: this will eventually include swizzling
+    return any(mUsage & Usage::SAMPLEABLE) && mLevelCount > 1;
+}
+
+void FTexture::updateLodRange(DriverApi& driver, uint8_t baseLevel, uint8_t levelCount) const noexcept {
+    if (any(mUsage & Usage::SAMPLEABLE) && mLevelCount > 1) {
+        auto& range = mLodRange;
+        uint8_t const last = int8_t(baseLevel + levelCount);
+        if (range.first > baseLevel || range.last < last) {
+            range.first = std::min(range.first, baseLevel);
```
bejado, line 490: I'm confused by the logic here. If `range` starts out as `{0,0}` (denoting all LODs available), won't `range.first` always be 0?

pixelflinger: yes good catch, that's a bug.

pixelflinger:
I created a new PR with ma/descriptor-sets as the base branch

PR #7939, opened 2024-06-22
fix use-after-free dereference risk

pixelflinger:
Texture handles were resolved to pointers when updating a SamplerGroup, as that point the handle was checked for use-after-free. However, the texture could be destroyed later while still active in the SamplerGroup, this would result in using the pointer which now contains garbage.

We now keep the handle and resolve the texture when binding samplers to the program; which will also perform the use-after-free check.

### DIFF
--- a/filament/backend/src/opengl/OpenGLDriver.h
+++ b/filament/backend/src/opengl/OpenGLDriver.h
@@ -126,7 +126,7 @@ public:
     struct GLSamplerGroup : public HwSamplerGroup {
         using HwSamplerGroup::HwSamplerGroup;
         struct Entry {
-            GLTexture const* texture = nullptr;
+            Handle<HwTexture> th;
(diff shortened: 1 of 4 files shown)


filament/backend/src/opengl/OpenGLDriver.h
```diff
@@ -256,6 +256,11 @@ class OpenGLDriver final : public DriverBase {
         return mHandleAllocator.handle_cast<Dp, B>(handle);
     }
 
+    template<typename B>
+    bool is_valid(Handle<B>& handle) {
```
bejado, line 260: Is this getting called anywhere?

pixelflinger: so it's not, but it probably will in the texture view PR.

PR #7942, opened 2024-06-26
Renames Vulkan platform's destroy function to destroySwapChain

GrantComm:
FIXES=[349602983]
Renames Vulkan platform's "destroy" function, which destroys the SwapChain, to "destroySwapChain" to mimic the other platforms.


pixelflinger (commented):
Won't this break existing client code? (in particular in google3). As for the comment in b/349602983 you could just call `destroySwapChain()` from `destroy()` if that's the behavior you want.

I don't see a lot of value in that change vs. problems in can introduce.

PR #7944, opened 2024-06-26
Exposes createInstance and createLogicalDevice functions

GrantComm:
FIXES=[349603734]

Exposes createInstance and createLogicalDevice as protected member functions so that they can be overridden.


poweifeng:
I've commented on b/349603734 . Let's continue the discussion there.

PR #7946, opened 2024-06-26
Texture View support

pixelflinger:
Texture Views are currently limited to the min/max lod, and it's not
allowed to have several views of the same texture in a single shader.
Only sampleable textures can have a texture view.


pixelflinger:
this has been added to ma/descriptor-sets

PR #7947, opened 2024-06-27
IBL drag & drop support for desktop gltf_viewer

z3moon:
Users now can drag and drop IBL files for desktop gltf_viewer.

It attempts to load gltf files first then tries to load IBL files.

### DIFF
--- a/samples/gltf_viewer.cpp
+++ b/samples/gltf_viewer.cpp
@@ -58,6 +58,7 @@
 
 #include <cgltf.h>
 
+#include <algorithm>
 #include <array>
 #include <fstream>
 #include <iostream>
@@ -520,14 +521,8 @@ static utils::Path getPathForIBLAsset(std::string_view string) {
     }
 
     if (filename.isDirectory()) {
-        bool found = false;
-        for (const auto& file: filename.listContents()) {
-            if (isIBL(file)) {
-                found = true;
-                break;
-            }
-        }
-        if (!found) {
+        std::vector<Path> files = filename.listContents();
+        if (std::none_of(files.cbegin(), files.cend(), isIBL)) {
             return {};
         }
     } else if (!isIBL(filename)) {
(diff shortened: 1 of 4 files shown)


samples/gltf_viewer.cpp
```diff
@@ -508,30 +508,64 @@ static void onClick(App& app, View* view, ImVec2 pos) {
...
     if (!filename.exists()) {
         std::cerr << "file " << filename << " not found!" << std::endl;
         return {};
     }
+
     if (filename.isDirectory()) {
-        auto files = filename.listContents();
-        for (const auto& file: files) {
-            if (file.getExtension() == "gltf" || file.getExtension() == "glb") {
+        bool found = false;
```
poweifeng, line 523: use `std::find` here?

samples/gltf_viewer.cpp
```diff
@@ -508,30 +508,64 @@ static void onClick(App& app, View* view, ImVec2 pos) {
...
+            return {};
+        }
+    } else if (!isIBL(filename)) {
+        return {};
+    }
+
+    return filename;
+}
+
+static utils::Path getPathForGLTFAsset(std::string_view string) {
```
poweifeng, line 540: I feel like this and `getPathForIBLAsset` are pretty much the same except for `isGLTF` and `isIBL`.  Maybe we can pull that into a paramter and just have a single method?

pixelflinger (approved):
LGTM after addressing powei's comments.

samples/gltf_viewer.cpp
z3moon: good point. let me update it.

samples/gltf_viewer.cpp
z3moon: Yeah I actually considered unifying them at first. Though it's possible to do that, it's quite tricky to distinguish the slightly different but critical behaviors. So it's likely to end up with convoluted code.

poweifeng (approved):

PR #7949, opened 2024-06-27
Add saveRawVariants flag to matc

bejado:
This is useful for debugging shaders and the spirv pipeline.


libs/filamat/include/filamat/MaterialBuilder.h
```diff
@@ -586,6 +587,14 @@ class UTILS_PUBLIC MaterialBuilder : public MaterialBuilderBase {
     //! If true, will output the generated GLSL shader code to stdout.
     MaterialBuilder& printShaders(bool printShaders) noexcept;
 
+    /**
+     * If true, will will write the raw generated GLSL for each variant to a text file in the
```
pixelflinger, line 591:
```suggestion
     * If true, this will write the raw generated GLSL for each variant to a text file in the
```

pixelflinger (approved):

PR #7957, opened 2024-07-02
Fix a misusage of bindBufferRange

z3moon:


pixelflinger (changes requested):
No the code is actually correct. We can't bind *less* than the size declared in the shader.

The shader declares `PerRenderableData[64]` so we *must* bind that size.  If you look how the buffer is allocated, it has an extra `PerRenderableUib` bytes at the end so we can specifically do that.

z3moon:
Thanks for correcting me. Let me close this PR.

PR #7964, opened 2024-07-06
Fix precondition check when creating static geometry.

adamwych:
These 2 preconditions checks are currently incorrect and cause a crash when trying to create static geometry through `RenderableManager`:

https://github.com/google/filament/blob/ef9bbece2af9730af633f002a58680033b0ab410/filament/src/components/RenderableManager.cpp#L446-L450

Since skinning and morphing is not possible, they should ensure that counts are 0, not higher than 0.


pixelflinger (approved):

pixelflinger:
@adamwych please sing the CLA so we can submit your PR. Thanks.

PR #7978, opened 2024-07-18
Minor fixes in VulkanPipelineLayoutCache

mdagois:
Made a couple of minor fixes in the VulkanPipelineLayoutCache class.
- Fixed a typo
- Reused an iterator to avoid a redundant map retrieval.

### DIFF
--- a/filament/backend/src/vulkan/caching/VulkanPipelineLayoutCache.cpp
+++ b/filament/backend/src/vulkan/caching/VulkanPipelineLayoutCache.cpp
@@ -33,11 +33,11 @@ VkPipelineLayout VulkanPipelineLayoutCache::getLayout(
 
     // build the push constant layout key
     uint32_t pushConstantRangeCount = program->getPushConstantRangeCount();
-    auto const& pushCostantRanges = program->getPushConstantRanges();    
+    auto const& pushConstantRanges = program->getPushConstantRanges();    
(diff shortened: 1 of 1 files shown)


filament/backend/src/vulkan/caching/VulkanPipelineLayoutCache.cpp
```diff
@@ -52,9 +52,8 @@ VkPipelineLayout VulkanPipelineLayoutCache::getLayout(
         }
     }
 
-    auto iter = mPipelineLayouts.find(key);
-    if (iter != mPipelineLayouts.end()) {
-        PipelineLayoutCacheEntry& entry = mPipelineLayouts[key];
+    if (PipelineLayoutMap::iterator iter = mPipelineLayouts.find(key); iter != mPipelineLayouts.end()) {
```
poweifeng, line 55: nit: this line seems > 100 chars?

poweifeng (approved):
Thanks for the clean up!

filament/backend/src/vulkan/caching/VulkanPipelineLayoutCache.cpp
mdagois: Oops, I didn't know there was a 100 char rules. I will have to change that file again in the near future, so I can fix that at that time.

PR #7979, opened 2024-07-18
Fixed a barrier in the VulkanBuffer class.

mdagois:
The barrier that prevents the GPU from processing two copies to the same buffer at the same time was not taking the byte offset into account. The barrier was always using an offset of zero. This PR fixes the issue.


poweifeng (approved):
Thank you!

For posterity record, this was missing from #7818.

PR #7981, opened 2024-07-18
New external image API

bejado:
For descriptor sets, we want a 1:1 mapping between external image and texture handle.

### DIFF
--- a/filament/src/details/Texture.h
+++ b/filament/src/details/Texture.h
@@ -134,7 +134,6 @@ private:
     void updateLodRange(uint8_t baseLevel, uint8_t levelCount) noexcept;
 
     backend::Handle<backend::HwTexture> mHandle;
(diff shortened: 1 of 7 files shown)


filament/src/details/Texture.h
```diff
@@ -134,6 +134,7 @@ class FTexture : public Texture {
     void updateLodRange(uint8_t baseLevel, uint8_t levelCount) noexcept;
 
     backend::Handle<backend::HwTexture> mHandle;
+    backend::Handle<backend::HwTexture> mExternalImageHandle;
```
pixelflinger, line 137: why can't we just use `mHandle`?

bejado: Hmm, I think we can actually!

pixelflinger (approved):

PR #7983, opened 2024-07-19
RenderPass batch size was too small

pixelflinger:
The batch size was calculated incorrectly, which in theory could lead to command buffer overflows (albeit unlikely).

### DIFF
--- a/filament/src/RenderPass.cpp
+++ b/filament/src/RenderPass.cpp
@@ -38,6 +38,7 @@
 #include <backend/PipelineState.h>
 
 #include "private/backend/CircularBuffer.h"
+#include "private/backend/CommandStream.h"
 
 #include <utils/compiler.h>
 #include <utils/debug.h>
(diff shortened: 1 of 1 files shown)


filament/src/RenderPass.cpp
```diff
@@ -885,14 +885,31 @@ void RenderPass::Executor::execute(FEngine& engine,
...
-        // the safer side.
-        size_t const maxCommandSizeInBytes = 256;
+        // CommandStream protocol. Currently, the maximum is 320 bytes.
+        // The batch size is calculated by adding the size of all commands that can possibly be
+        // emitted per draw call:
+        constexpr size_t const maxCommandSizeInBytes =
+                sizeof(CustomCommand) +
+                sizeof(COMMAND_TYPE(scissor)) +
+                sizeof(COMMAND_TYPE(bindUniformBuffer)) +
+                sizeof(COMMAND_TYPE(bindSamplers)) +
```
bejado, line 895: Why the 6 calls to `bindSamplers`?

pixelflinger: if you look at the rendering loop, the worst case scenario is that we're doing that many calls to bindSamplers. The first one happens in `MaterialInstance::use().`

bejado: Got it

bejado (approved):

PR #7989, opened 2024-07-23
gltfio: enable escaped unicode for node name

### LINKED ISSUE #7846: Treatment of escaped vs unescaped Unicode character in JSON strings
**Describe the bug**
Escaped unicode characters in glTF are treated differently than unescaped ones.

```
"name": "\u4f60\u597d"         // escaped
"name": "你好"                 // unescaped
```

See also [2.7 JSON Encoding](https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#json-encoding) in the glTF spec: "Non-ASCII characters stored in glTF JSON MAY be escaped. [...] [Two examples differing only in string escaping] represent the same glTF JSON data."

**To Reproduce**
1. Open [EscapedUnicodeTest.glb.zip](https://github.com/google/filament/files/15285624/EscapedUnicodeTest.glb.zip) with `bin/gltf_viewer` from the latest release. Note that the node name is shown as `\u4f60\u597d`.
2. Do the same with [UnicodeTest.glb.zip](https://github.com/google/filament/files/15285629/UnicodeTest.glb.zip). Note that the node name is shown as `??`. (Presumably this is because the font doesn't have the correct glyphs. This issue is just about the discrepancy though, not the font.)

**Expected behavior**
No difference between these persists beyond the JSON loading phase.

**Screenshots**
With `"\u4f60\u597d"`

With `"你好"`

 - OS: Linux
 - GPU: 
 - Backend: 

**Additional context**
The Blender glTF exporter currently writes the escaped form (it's the default for Python's builtin JSON library). A user presumably noticed the difference in Filament and reported a bug to Blender: https://projects.blender.org/blender/blender/issues/121692

poweifeng:
Fixes #7846

### DIFF
--- a/libs/gltfio/src/AssetLoader.cpp
+++ b/libs/gltfio/src/AssetLoader.cpp
@@ -58,6 +58,8 @@
 
 #include "downcast.h"
 
+#include <codecvt>
+#include <locale>
 #include <memory>
 
 using namespace filament;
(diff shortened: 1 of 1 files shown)


pixelflinger (approved):

libs/gltfio/src/AssetLoader.cpp
```diff
@@ -88,12 +90,38 @@ static constexpr cgltf_material kDefaultMat = {
...
+    std::string sImpl(getNameImpl());
+
+    // We handle the potential case of escaped characters in the JSON which should be properly
+    // interpreted as unicode. See spec:
+    // https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#json-encoding
+    std::string sEscapedUnicode;
+    size_t cur = 0, idx = 0;
+    std::wstring_convert<std::codecvt_utf8<char32_t>, char32_t> conv;
+    while ((idx = sImpl.find("\\u", cur)) != std::string::npos) {
+        auto endIdx = sImpl.find_first_not_of("0123456789abcdefABCDEF", idx + 2);
```
bejado, line 111: I think we need to limit this to only 6 characters. Otherwise we could have input like this which causes a `std::range_error: wstring_convert: to_bytes error`:

```
\u6211\u53EBBenjamin
```

(should be 我叫Benjamin)

> If the character is in the Basic
   Multilingual Plane (U+0000 through U+FFFF), then it may be
   represented as a six-character sequence: a reverse solidus, followed
   by the lowercase letter u, followed by four hexadecimal digits that
   encode the character's code point.

See https://www.ietf.org/rfc/rfc4627.txt

poweifeng: Thanks for the spec link.  This makes things a lot easier to understand.

poweifeng: Addressed.

bejado (approved):

libs/gltfio/src/AssetLoader.cpp
bejado: LGTM. My other recommendation is that `to_bytes` can throw a `std::range_error` error if the unicode point is not valid, but not sure if we care about handling that case since that would mean the string is malformed.

poweifeng: I feel that's more of the user's responsibility.  Let's discuss again if/when someone files a bug on it.

PR #7990, opened 2024-07-23
vk: use std::unordered_map in PipelineLayoutCache

poweifeng:
This is to address a weird map.find() miss that only happens on ARM in release mode.

### DIFF
--- a/filament/backend/src/vulkan/caching/VulkanPipelineLayoutCache.h
+++ b/filament/backend/src/vulkan/caching/VulkanPipelineLayoutCache.h
@@ -22,7 +22,7 @@
 
 #include <utils/Hash.h>
 
-#include <tsl/robin_map.h>
+#include <unordered_map>
 
 namespace filament::backend {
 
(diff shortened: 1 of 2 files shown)


pixelflinger (approved):

filament/backend/src/vulkan/caching/VulkanPipelineLayoutCache.h
```diff
@@ -36,8 +36,8 @@ class VulkanPipelineLayoutCache {
     void terminate() noexcept;
 
     struct PushConstantKey {
-        uint8_t stage;// We have one set of push constant per shader stage (fragment, vertex, etc).
-        uint8_t size;
+        uint8_t stage = 0;// We have one set of push constant per shader stage (fragment, vertex, etc).
+        uint8_t size = 0;
```
pixelflinger, line 40: couldn't the problem come from that?

bejado (approved):
LGTM, but if there's a bug with `tsl::robin_map`, we should investigate because we use that all over the place. In my experience, bugs have been related to padding bytes.

filament/backend/src/vulkan/caching/VulkanPipelineLayoutCache.h
poweifeng: that's what i thought. But just making this change does not fix the problem unfortunately.

poweifeng:
> LGTM, but if there's a bug with `tsl::robin_map`, we should investigate because we use that all over the place. In my experience, bugs have been related to padding bytes.

Agreed.  Let me post to our chat more about this.  I experimented with various things, and the cause still eludes me.  AFAICT there are no padding issues in this key.

poweifeng:
Internal bug for tracking has been created.  Landing this just to unblock usage on certain devcies.

PR #7991, opened 2024-07-23
Move the ResourceAllocator from Engine to Renderer

pixelflinger:
The ResourceAllocator used to be global and owned by the Engine, this was causing some issues when using several Renderers because each one could cause the eviction of cache data for another.

We now have a ResourceAllocator per Renderer, which makes more sense because most resources are allocated by the FrameGraph. However, there are some textures temporarily owned by Views in their  FrameHistory (e.g.: with SSR or TAA) and because Views can be used by  multiple Renderers, this means that FrameHistory entries can have  resources owned by different ResourceAllocators. 
So we need to handle these cases:

- a Renderer is destroyed before a View that rendered into it: The view won't be able to destroy the resources later, so they are now destroyed when the Renderer is destroyed, effectively making the  handles stale.

- a View is destroyed and has FrameHistory state that belongs to  multiple Renderer and therefore need to be returned to the corresponding ResourceManager

Both cases are handled by storing a weak pointer to the correct ResourceAllocator which each FrameHistory entry. This allows the view to return the resources when it is destroyed AND this allows all parties to check that the FrameEntry is valid and its handles are  usable. In particular Renderer now checks the validity of the history before using its handles (an Entry could have resources belonging to another, destroyed, Renderer).

Finally we add a View::clearFrameHistory() method which can be used when a view is moved from one Renderer to another, or when it is  appropriate to ignore the frame history (e.g.: cut-scene). Generally, it is not a good idea to use the same View with multiple Renderers as it will most certainly mess-up the frame history (At least not if  temporal effects are used).

### DIFF
--- a/android/filament-android/src/main/cpp/View.cpp
+++ b/android/filament-android/src/main/cpp/View.cpp
@@ -531,3 +531,11 @@ Java_com_google_android_filament_View_nGetFogEntity(JNIEnv *env, jclass clazz,
     View *view = (View *) nativeView;
     return (jint)view->getFogEntity().getId();
 }
+
+extern "C"
+JNIEXPORT void JNICALL
+Java_com_google_android_filament_View_nClearFrameHistory(JNIEnv *env, jclass clazz,
+        jlong nativeView) {
+    View *view = (View *) nativeView;
+    view->clearFrameHistory();
+}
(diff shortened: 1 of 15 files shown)


pixelflinger:
I have possibly a better idea to implement this, so closing for now.

PR #7993, opened 2024-07-23
Move the ResourceAllocator from Engine to Renderer

pixelflinger:
The ResourceAllocator used to be global and owned by the Engine, this was causing some issues when using several Renderers because each one could cause the eviction of cache data for another.

We now have a ResourceAllocator per Renderer, which makes more sense because most resources are allocated by the FrameGraph.

We also introduce a ResourceAllocatorDisposer class, which is used for checking in and out a texture from the cache, and destroy the texture when it's checked-out. That objet is still global.

### DIFF
--- a/android/filament-android/src/main/java/com/google/android/filament/View.java
+++ b/android/filament-android/src/main/java/com/google/android/filament/View.java
@@ -1233,6 +1233,18 @@ public class View {
         return nGetFogEntity(getNativeObject());
     }
 
+    /**
+     * When certain temporal features are used (e.g.: TAA or Screen-space reflections), the view
+     * keeps a history of previous frame renders associated with the Renderer the view was last
+     * used with. When switching Renderer, it may be necessary to clear that history by calling
+     * this method. Similarly, if the whole content of the screen change, like when a cut-scene
+     * starts, clearing the history might be needed to avoid artifacts due to the previous frame
+     * being very different.
+     */
+    public void clearFrameHistory(Engine engine) {
+        nClearFrameHistory(getNativeObject(), engine.getNativeObject());
+    }
+
     public long getNativeObject() {
         if (mNativeObject == 0) {
             throw new IllegalStateException("Calling method on destroyed View");
@@ -1294,7 +1306,7 @@ public class View {
     private static native void nSetMaterialGlobal(long nativeView, int index, float x, float y, float z, float w);
     private static native void nGetMaterialGlobal(long nativeView, int index, float[] out);
     private static native int nGetFogEntity(long nativeView);
-
+    private static native void nClearFrameHistory(long nativeView, long nativeEngine);
 
     /**
      * List of available ambient occlusion techniques.
--- a/filament/include/filament/View.h
+++ b/filament/include/filament/View.h
@@ -40,6 +40,7 @@ class CallbackHandler;
 
 class Camera;
 class ColorGrading;
+class Engine;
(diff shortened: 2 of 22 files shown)


android/filament-android/src/main/java/com/google/android/filament/View.java
```diff
@@ -1233,6 +1233,18 @@ public int getFogEntity() {
         return nGetFogEntity(getNativeObject());
     }
 
+    /**
+     * When certain temporal features are used (e.g.: TAA or Screen-space reflections), the view
+     * keeps an history of previous frame renders associated with the Renderer the view was last
```
poweifeng, line 1238:
```suggestion
     * keeps a history of previous frame renders associated with the Renderer the view was last
```

filament/include/filament/View.h
```diff
@@ -878,6 +879,17 @@ class UTILS_PUBLIC View : public FilamentAPI {
      */
     utils::Entity getFogEntity() const noexcept;
 
+
+    /**
+     * When certain temporal features are used (e.g.: TAA or Screen-space reflections), the view
+     * keeps an history of previous frame renders associated with the Renderer the view was last
```
poweifeng, line 885:
```suggestion
     * keeps a history of previous frame renders associated with the Renderer the view was last
```

filament/src/ResourceAllocator.cpp
```diff
@@ -254,4 +265,45 @@ void ResourceAllocator::purge(
...
+
+std::optional<ResourceAllocator::TextureKey> ResourceAllocatorDisposer::checkin(
+        backend::TextureHandle handle) {
+    std::optional<ResourceAllocator::TextureKey> result;
+    // find the texture in the in-use list (it must be there!)
+    auto it = mInUseTextures.find(handle);
+    assert_invariant(it != mInUseTextures.end());
+    if (it == mInUseTextures.end()) {
+        return result;
+    }
```
poweifeng, line 302: I'm a bit confused by this. It reads like in debug, we expect the handle to be in the in-use list, but in release, we don't?

filament/test/filament_framegraph_test.cpp
```diff
@@ -34,6 +34,10 @@ using namespace backend;
 
 class MockResourceAllocator : public ResourceAllocatorInterface {
     uint32_t handle = 0;
+    struct MokeDisposer : public ResourceAllocatorDisposerInterface {
```
poweifeng, line 37:
```suggestion
    struct MockDisposer : public ResourceAllocatorDisposerInterface {
```

filament/src/ResourceAllocator.cpp
pixelflinger: In debug we assert if we don't find the item in the list, and in release we return `result` which is a `std::optional<>` with no value set. But I will change it to `std::nullopt` to be clearer.

poweifeng: sorry, to be clearer, why do we expect different behavior for debug and release?

pixelflinger: Well in both case we expect it to be in the list. But in DEBUG I assert (so we catch it) and in RELEASE I return an error so that we don't corrupt the state. I could remove the 
```
    if (it == mInUseTextures.end()) {
        return std::nullopt;
    }
```
because it should never happen.

poweifeng (approved):

Issue #7997, opened 2024-07-26
[Android] Crash when using extended asset loader


mbalajee:
⚠️ **Issues not using this template will be systematically closed.**

**Describe the bug**
When using the extended asset loader the app crashes at [this](https://github.com/google/filament/blob/b54a2046cb12f47c34e6e45ffda298c3ad34033f/libs/gltfio/src/extended/TangentsJobExtended.cpp#L201) line in `TangentJobExtended`,
```cpp
if (inStride == outStride) {
   std::memcpy(out, in, data::byteCount<InDataType>() * count);
} 
```

**To Reproduce**
1. Run the sample gltf viewer app by passing an empty string while creating Asset to opt in for extended asset loader
2. From activity 1 with just one button launch MainActivity which loads the model
3. Go back to activity 1 and click button again to launch MainActivity
4. Repeat the step 3 a few times
5. App crashes with the above message

**Log**
```
libc                    com.google.android.filament.gltf     A  Fatal signal 11 (SIGSEGV), code 1 (SEGV_MAPERR), fault addr 0x0 in tid 19346 (Thread-6), pid 17322 (d.filament.gltf)
DEBUG                   pid-19402                            A  Cmdline: com.google.android.filament.gltf
DEBUG                   pid-19402                            A  pid: 17322, tid: 19346, name: Thread-6  >>> com.google.android.filament.gltf <<<
DEBUG                   pid-19402                            A        #01 pc 0000000001336730  /data/app/~~ygrXgdh3prb7VdGAGSFBrg==/com.google.android.filament.gltf-rTXoEEZ-eKX05S5qnSodEw==/lib/arm64/libgltfio-jni.so (void (anonymous namespace)::data::copy<filament::math::details::TVec3<float>*, filament::math::details::TVec3<float>*>(filament::math::details::TVec3<float>*, unsigned long, filament::math::details::TVec3<float>*, unsigned long, unsigned long)+244) (BuildId: c57cac424fa8a4538dda3465fa66b08785fac5b7)
[...]

PR #7999, opened 2024-07-26
Set mCgltfBuffersLoaded in constructor

mbalajee:
Value of `mCgltfBuffersLoaded` is sometimes retained across creation of FAssetLoader which skips loading the buffer in AssetLoaderExtended#createPrimitive leading to null pointer crash.

I also tried `(*loader)->mLoaderExtended.reset()` in `AssetLoader::destroy` but that doesn't seem to reset the value of `mCgltfBuffersLoaded`

Fixes the issue, https://github.com/google/filament/issues/7997

### DIFF
--- a/libs/gltfio/src/extended/AssetLoaderExtended.h
+++ b/libs/gltfio/src/extended/AssetLoaderExtended.h
@@ -70,7 +70,8 @@ struct AssetLoaderExtended {
         : mEngine(engine),
           mGltfPath(config.gltfPath),
           mMaterials(materials),
-          mUriDataCache(std::make_shared<UriDataCache>()) {}
+          mUriDataCache(std::make_shared<UriDataCache>()),
+          mCgltfBuffersLoaded(false) {}
 
     ~AssetLoaderExtended() = default;
 


mbalajee:
@pirlande-fw

libs/gltfio/src/extended/AssetLoaderExtended.h
```diff
@@ -70,7 +70,9 @@ struct AssetLoaderExtended {
         : mEngine(engine),
           mGltfPath(config.gltfPath),
           mMaterials(materials),
-          mUriDataCache(std::make_shared<UriDataCache>()) {}
+          mUriDataCache(std::make_shared<UriDataCache>()) {
+        mCgltfBuffersLoaded = false;
```
poweifeng, line 74: please move this up to the constructor member initializer list.

mbalajee: Done

mbalajee:
@poweifeng Cleaned up commits.  Should be ready for review now.

poweifeng (approved):

PR #8001, opened 2024-07-29
Fix a crash for IBL resource loading

z3moon:
This fixes a crash introduced by a8ace2891d2d83e8b9dd7d06a01591d3ff53c764

The refactored FrameInfoManager is now required to be called properly within the renderStandaloneView method, which is mainly being called by IBLPrefilterContext.

Also fix a null pointer reference bug for OpenGLTimer::State, which happenes when the renderer for IBLPrefilterContext is destroyed.

### DIFF
--- a/filament/src/FrameInfo.h
+++ b/filament/src/FrameInfo.h
@@ -163,7 +163,7 @@ public:
     // call this immediately before "swap buffers"
     void endFrame(backend::DriverApi& driver) noexcept;
 
-    details::FrameInfo const getLastFrameInfo() const noexcept {
+    details::FrameInfo getLastFrameInfo() const noexcept {
         // if pFront is not set yet, return FrameInfo(). But the `valid` field will be false in this case.
         return pFront ? *pFront : details::FrameInfo();
     }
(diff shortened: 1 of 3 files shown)


bejado (approved):

filament/src/FrameInfo.h
```diff
@@ -163,9 +163,9 @@ class FrameInfoManager {
     // call this immediately before "swap buffers"
     void endFrame(backend::DriverApi& driver) noexcept;
 
-    details::FrameInfo const& getLastFrameInfo() const noexcept {
-        // if pFront is not set yet, return front() but in this case front().valid will be false
-        return pFront ? *pFront : mFrameTimeHistory.front();
+    details::FrameInfo const getLastFrameInfo() const noexcept {
```
pixelflinger, line 166:
```suggestion
    details::FrameInfo getLastFrameInfo() const noexcept {
```

filament/src/FrameInfo.h
```diff
@@ -163,9 +163,9 @@ class FrameInfoManager {
...
+        // if pFront is not set yet, return FrameInfo(). But the `valid` field will be false in this case.
+        return pFront ? *pFront : details::FrameInfo();
```
pixelflinger, line 168:
```suggestion
        return pFront ? *pFront : details::FrameInfo{};
```

pixelflinger (approved):
just to minor suggestions.

PR #8005, opened 2024-07-31
Fix ES2/anisotropy bug in updateSamplerGroup

elizagamedev:
This function attempts to set texture parameters in these two cases, but the texture is not guaranteed to be bound. Perhaps it once was, but the assumption broke at some point.

### DIFF
--- a/filament/backend/src/opengl/OpenGLDriver.cpp
+++ b/filament/backend/src/opengl/OpenGLDriver.cpp
@@ -2341,6 +2341,15 @@ void OpenGLDriver::updateSamplerGroup(Handle<HwSamplerGroup> sbh,
             GLTexture const* const t = handle_cast<const GLTexture*>(th);
             assert_invariant(t);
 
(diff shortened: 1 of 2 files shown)


pixelflinger (approved):

filament/backend/src/opengl/OpenGLDriver.cpp
```diff
@@ -2341,6 +2341,15 @@ void OpenGLDriver::updateSamplerGroup(Handle<HwSamplerGroup> sbh,
             GLTexture const* const t = handle_cast<const GLTexture*>(th);
             assert_invariant(t);
 
+            if (UTILS_UNLIKELY(es2
```
pixelflinger, line 2344: looks like UTILS_UNLIKELY with the #if inside doesn't work with MSVC

pixelflinger (changes requested):
windows build broke

pixelflinger (approved):

PR #8021, opened 2024-08-07
Support for transient attachments

mdagois:
Changes
- Added support for transient attachment in the VulkanTexture class.
- Used transient attachment for the MSAA sidecar texture.

Testing on a Pixel 8 Pro, AGI traces show a difference of roughly 100MB in GPU memory consumption.
The device was rebooted after each test to reduce the noise.

Figure 1: Transient disabled
transient_disabled

Figure 2: Transient enabled
transient_enabled

The actual difference is likely smaller.
The resolution is 1008x1920 with 4 samples and a 32-bit format (VK_FORMAT_B10G11R11_UFLOAT_PACK32).
In that configuration, the reduction in memory consumption should be around 30MB.

For reference, here are the perfetto traces: [perfetto_traces.zip](https://github.com/user-attachments/files/16520374/perfetto_traces.zip).


mdagois:
As an additional note, I used the `vkGetDeviceMemoryCommitment` function to make sure the memory was not committed when using the transient attachment.
The results are that, for the allocation size was 30,965,760 bytes, the committed size is zero.
This is the expected behavior.
Unfortunately, the `vkGetDeviceMemoryCommitment` function cannot be used for non-lazily allocated memory, so we cannot do a apple-to-apple comparison.

poweifeng (commented):
I think this looks good, but it might be missing a bit, which is the part that was undone in #7680.  See also the "// DONT_CARE is critical here" comment in this document: https://arm-software.github.io/vulkan-sdk/multisampling.html

mdagois:
> I think this looks good, but it might be missing a bit, which is the part that was undone in #7680. See also the "// DONT_CARE is critical here" comment in this document: https://arm-software.github.io/vulkan-sdk/multisampling.html

That means the number I got showing a 100MB saving is probably just happenstance then. :)
Let me try with the method you told me about offline (`adb shell dumpsys meminfo app.name`).

mdagois:
> I think this looks good, but it might be missing a bit, which is the part that was undone in #7680. See also the "// DONT_CARE is critical here" comment in this document: https://arm-software.github.io/vulkan-sdk/multisampling.html

I fixed the issue by setting the loadOp based on whether the attachement uses lazily allocated memory.
Please have a look.

mdagois:
With the *discard* storeop, I can see the committed memory stays at zero for the MSAA attachement for as long as the application runs.
When switching to a *store* storeop, I can see the committed memory turning to 30MB.
So, the transient system seems to work as intended.

poweifeng (approved):
awesome work!

filament/backend/src/vulkan/VulkanDriver.cpp
```diff
@@ -1293,8 +1295,16 @@ void VulkanDriver::beginRenderPass(Handle<HwRenderTarget> rth, const RenderPassP
...
             rpkey.initialColorLayoutMask |= 1 << i;
             rpkey.colorFormat[i] = info.getFormat();
-            if (rpkey.samples > 1 && info.texture->samples == 1) {
-                rpkey.needsResolveMask |= (1 << i);
+            if (rpkey.samples > 1) {
+                const VulkanTexture* sidecar = info.texture->getSidecar();
+                assert_invariant(sidecar);
+                if (sidecar && sidecar->isTransientAttachment()) {
+                    rpkey.usesLazilyAllocatedMemory |= (1 << i);
+                    assert_invariant(sidecar->samples > 1);
```
pixelflinger, line 1303: move this assert just bellow the `sidecar` assert.

filament/backend/src/vulkan/VulkanTexture.cpp
```diff
@@ -97,12 +97,31 @@ VulkanTexture::VulkanTexture(VkDevice device, VkPhysicalDevice physicalDevice,
...
     }
 
+    // Determine if we can use the transient usage flag combined with lazily allocated memory.
+    const bool useTransientAttachment =
+        // Transient attachment is requested.
+        preferTransientAttachment &&
+        // Lazily allocated memory is available.
+        context.isLazilyAllocatedMemorySupported() &&
+        // Usage consists of attachment flags only.
+        !any(tusage & ~TextureUsage::ALL_ATTACHMENTS) &&
```
pixelflinger, line 107: use `none()` instead of `!any()`

filament/backend/src/vulkan/VulkanHandles.cpp
```diff
@@ -352,11 +352,17 @@ VulkanRenderTarget::VulkanRenderTarget(VkDevice device, VkPhysicalDevice physica
         if (texture && texture->samples == 1) {
             auto msTexture = texture->getSidecar();
             if (UTILS_UNLIKELY(!msTexture)) {
+                // Clear all usage flags that are not related to attachments, so that we can
+                // use the transient usage flag.
+                const TextureUsage usage = texture->usage & TextureUsage::ALL_ATTACHMENTS;
```
pixelflinger, line 357: Is this correct? what if there was some other usage bits set? Shouldn't we instead passing the `usage` as is and let the VulkanTexture decide if it can use it?

filament/backend/src/vulkan/VulkanTexture.cpp
```diff
@@ -97,12 +97,31 @@ VulkanTexture::VulkanTexture(VkDevice device, VkPhysicalDevice physicalDevice,
...
     // Filament expects blit() to work with any texture, so we almost always set these usage flags.
+    // We do not add the flags for transient attachments, because that would prevent the usage of
+    // the VK_IMAGE_USAGE_TRANSIENT_ATTACHMENT_BIT flag.
     // TODO: investigate performance implications of setting these flags.
-    constexpr VkImageUsageFlags blittable = VK_IMAGE_USAGE_TRANSFER_DST_BIT |
-            VK_IMAGE_USAGE_TRANSFER_SRC_BIT;
+    const VkImageUsageFlags blittable =
+       useTransientAttachment ?
+       0U :
+       VK_IMAGE_USAGE_TRANSFER_DST_BIT | VK_IMAGE_USAGE_TRANSFER_SRC_BIT;
```
pixelflinger, line 119: We should probably revisit why blitable usages are systematically added, regardless of the `BLIT_` flags -- we shouldn't have to do these shenanigans with `blitable`.

pixelflinger (changes requested):

filament/backend/src/vulkan/VulkanDriver.cpp
mdagois: Fixed.

filament/backend/src/vulkan/VulkanTexture.cpp
mdagois: Fixed.

filament/backend/src/vulkan/VulkanTexture.cpp
mdagois: Totally agree. I think forcing everything to be blittable sliently is not good, because of the potential use cases, like transient, where it gets in the way.
It is a broader change though, unrelated with transient support perse. Should we just add a separate todo?

filament/backend/src/vulkan/VulkanHandles.cpp
mdagois: As far as I understand, `texture->usage` is the `usage` of the resolved texture.
We cannot use that for MSAA, as the requirement are not as strict.

In this particular case, the `SAMPLEABLE` flag is raised for the original texture, but it is not necessary nor desirable for the MSAA texture, and it prevents us from setting the attachment as transient. We cannot have any flags besides the four ***_ATTACHMENT_BIT for transient attachments.

mdagois: Thinking about it... we could eventually remove the `preferTransientAttachment` parameter from the `VulkanTexture`. I don't think there is a case where you would not want to create a transient attachment if it is possible to do so.
What do you think?

pixelflinger: > As far as I understand, `texture->usage` is the `usage` of the resolved texture. We cannot use that for MSAA, as the requirements are stricter for transient attachments.
> 
> In this particular case, the `SAMPLEABLE` flag is raised for the original texture, but it is not necessary nor desirable for the MSAA texture, and it prevents us from setting the attachment as transient. We cannot have any flags besides the four ***_ATTACHMENT_BIT for transient attachments.

mh... I want to understand why the SAMPLEABLE flag is set then. Did the framegraph set it? if it did, it's probably that it needs it?

filament/backend/src/vulkan/VulkanTexture.cpp
pixelflinger: yes, let's add a todo, and maybe you can have a look next? (or we will). you can even file a github issue.

mdagois: I can definitely have a look. I think some usage flags that are silently added might also be getting in the way of protected content.

filament/backend/src/vulkan/VulkanHandles.cpp
mdagois: I don't know the frame graph well enough to confirm that. I am not sure if that helps, but by the time `ResourceAllocator::createTexture` is called, the `SAMPLEABLE` flag is already there.

However, I think it is not relevant for MSAA. Whether the resolved texture is sampled or not should not impact the MSAA texture usage in any way. But, we still need to use the right attachment flags, hence the mask with `TextureUsage::ALL_ATTACHMENTS`.

mdagois: As discussed offline, I checked the way filament handles the MSAA resolve.
The `VK_IMAGE_USAGE_TRANSFER_SRC_BIT` flag is only needed when using `vkCmdResolveImage`. Filament uses the more efficient `VkSubpassDescription::pResolveAttachments` for the sidecar textures, so we are safe masking the flags unrelated to attachments.

pixelflinger:
@mdagois could you rebase the branch and resolve the conflicts? thanks!

mdagois:
> @mdagois could you rebase the branch and resolve the conflicts? thanks!

I rebased the branch.
I also simplified the code in the constructor (like removing the `preferTransient` boolean) since we don't have the blittable flags raised by default anymore.

mdagois:
@pixelflinger @poweifeng I rebased the branch and tested it again. It still work as intended. :)
Only minor changes were necessary (around the new texture state). Please check the PR again.

poweifeng (approved):

pixelflinger (approved):

PR #8030, opened 2024-08-09
Re-enable missing packing/unpacking functions in ESSL, fix IMG shader compilation

bejado:


filament/backend/src/opengl/ShaderCompilerService.cpp
```diff
@@ -590,24 +590,48 @@ void ShaderCompilerService::compileShaders(OpenGLContext& context,
...
                     (GLint)prolog.length(),
                     (GLint)specializationConstantString.length(),
                     (GLint)packingFunctions.length(),
                     (GLint)body.length() - 1 // null terminated
             };
 
             GLuint const shaderId = glCreateShader(glShaderType);
-            glShaderSource(shaderId, sources.size(), sources.data(), lengths.data());
+
+            if (context.bugs.concatenate_shader_strings) {
```
pixelflinger, line 618:
```suggestion
            if (UTILS_UNLIKELY(context.bugs.concatenate_shader_strings)) {
```

pixelflinger (changes requested):
just use the UNLIKELY for the workaround version

pixelflinger (approved):

PR #8031, opened 2024-08-09
Add warning for the incompatibility of stereo type

z3moon:
Print a warning in case the stereoscopic type in a compiled package is different than what's in the engine's setting. The application may proceed, but it could end up visual glitches when enabling stereoscopic rendering.

This requires the stereoscopic type to be written into the package, which needs a material version bump.

### DIFF
--- a/filament/src/details/Material.cpp
+++ b/filament/src/details/Material.cpp
@@ -192,6 +192,22 @@ Material* Material::Builder::build(Engine& engine) {
         return nullptr;
     }
 
+    MaterialDomain materialDomain;
+    materialParser->getMaterialDomain(&materialDomain);
+    if (materialDomain == MaterialDomain::SURFACE) {
+        StereoscopicType const engineStereoscopicType = engine.getConfig().stereoscopicType;
+        StereoscopicType materialStereoscopicType = StereoscopicType::NONE;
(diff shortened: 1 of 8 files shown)


pixelflinger (approved):

filament/src/details/Material.cpp
```diff
@@ -192,6 +192,22 @@ Material* Material::Builder::build(Engine& engine) {
         return nullptr;
     }
 
+    MaterialDomain materialDomain;
+    materialParser->getMaterialDomain(&materialDomain);
+    if (materialDomain == MaterialDomain::SURFACE) {
+        StereoscopicType const engineStereoscopicType = engine.getConfig().stereoscopicType;
+        StereoscopicType materialStereoscopicType = StereoscopicType::NONE;
+        materialParser->getStereoscopicType(&materialStereoscopicType);
```
poweifeng, line 200: Does the bool returned here indicate whether getting the chunk was successful or not? Should we only do the following conditional only if the chunk was successfully parsed? 

I think this'll happen if the material is of an older version than filament.

z3moon: Yeah I also realized it returns a boolean but found other usages don't check that either. Maybe it'd nice to handle them together later.

PR #8038, opened 2024-08-12
Support tagging driver handles with a name 

bejado:
BUGS=[349683761]

### DIFF
--- a/filament/backend/include/private/backend/DriverAPI.inc
+++ b/filament/backend/include/private/backend/DriverAPI.inc
@@ -162,6 +162,10 @@ DECL_DRIVER_API_0(finish)
 // reset state tracking, if the driver does any state tracking (e.g. GL)
 DECL_DRIVER_API_0(resetState)
 
+DECL_DRIVER_API_N(setDebugTag,
+        backend::HandleBase::HandleId, handleId,
+        std::string&&, tag)
+
 /*
  * Creating driver objects
  * -----------------------
--- a/filament/backend/include/private/backend/DriverAPI.inc
+++ b/filament/backend/include/private/backend/DriverAPI.inc
@@ -164,7 +164,7 @@ DECL_DRIVER_API_0(resetState)
 
 DECL_DRIVER_API_N(setDebugTag,
         backend::HandleBase::HandleId, handleId,
-        std::string&&, tag)
+        utils::CString, tag)
 
 /*
  * Creating driver objects
--- a/filament/backend/include/private/backend/HandleAllocator.h
+++ b/filament/backend/include/private/backend/HandleAllocator.h
@@ -174,10 +174,10 @@ public:
                 uint8_t const age = (tag & HANDLE_AGE_MASK) >> HANDLE_AGE_SHIFT;
                 auto const pNode = static_cast<typename Allocator::Node*>(p);
                 uint8_t const expectedAge = pNode[-1].age;
-                utils::CString const tag = get_handle_tag(handle.getId());
+                // get_handle_tag() is only called if the check fails.
                 FILAMENT_CHECK_POSTCONDITION(expectedAge == age)
                         << "use-after-free of Handle with id=" << handle.getId()
-                        << ", tag=" << tag.c_str_safe();
+                        << ", tag=" << get_handle_tag(handle.getId()).c_str_safe();
             }
         }
 
--- a/filament/backend/include/private/backend/HandleAllocator.h
+++ b/filament/backend/include/private/backend/HandleAllocator.h
@@ -20,18 +20,17 @@
 #include <backend/Handle.h>
 
 #include <utils/Allocator.h>
+#include <utils/CString.h>
 #include <utils/Log.h>
+#include <utils/Panic.h>
 #include <utils/compiler.h>
 #include <utils/debug.h>
 #include <utils/ostream.h>
-#include <utils/Panic.h>
 
 #include <tsl/robin_map.h>
 
 #include <cstddef>
 #include <exception>
-#include <optional>
-#include <string>
 #include <type_traits>
 #include <unordered_map>
 #include <utility>
@@ -175,10 +174,10 @@ public:
                 uint8_t const age = (tag & HANDLE_AGE_MASK) >> HANDLE_AGE_SHIFT;
                 auto const pNode = static_cast<typename Allocator::Node*>(p);
                 uint8_t const expectedAge = pNode[-1].age;
-                std::optional<std::string_view> const maybeTag = get_handle_tag(handle.getId());
+                utils::CString const tag = get_handle_tag(handle.getId());
                 FILAMENT_CHECK_POSTCONDITION(expectedAge == age)
                         << "use-after-free of Handle with id=" << handle.getId()
-                        << ", tag=" << maybeTag.value_or("(no tag)");
+                        << ", tag=" << tag.c_str_safe();
             }
         }
 
@@ -205,7 +204,7 @@ public:
         return handle_cast<Dp>(const_cast<Handle<B>&>(handle));
     }
 
-    void associate_handle(HandleBase::HandleId id, std::string&& tag) noexcept {
+    void associate_handle(HandleBase::HandleId id, utils::CString&& tag) noexcept {
         // TODO: for now, only pool handles check for use-after-free, so we only keep tags for
         // those
         if (isPoolHandle(id)) {
@@ -221,14 +220,14 @@ public:
         }
     }
 
-    std::optional<std::string_view> get_handle_tag(HandleBase::HandleId id) const noexcept {
+    utils::CString get_handle_tag(HandleBase::HandleId id) const noexcept {
         if (!isPoolHandle(id)) {
-            return {};
+            return "(no tag)";
         }
         if (auto pos = mDebugTags.find(id); pos != mDebugTags.end()) {
             return pos->second;
         }
-        return {};
+        return "(no tag)";
     }
 
 private:
@@ -393,7 +392,7 @@ private:
     // Below is only used when running out of space in the HandleArena
     mutable utils::Mutex mLock;
     tsl::robin_map<HandleBase::HandleId, void*> mOverflowMap;
-    tsl::robin_map<HandleBase::HandleId, std::string> mDebugTags;
+    tsl::robin_map<HandleBase::HandleId, utils::CString> mDebugTags;
     HandleBase::HandleId mId = 0;
     bool mUseAfterFreeCheckDisabled = false;
 };
--- a/filament/backend/include/private/backend/HandleAllocator.h
+++ b/filament/backend/include/private/backend/HandleAllocator.h
@@ -174,10 +174,10 @@ public:
                 uint8_t const age = (tag & HANDLE_AGE_MASK) >> HANDLE_AGE_SHIFT;
                 auto const pNode = static_cast<typename Allocator::Node*>(p);
                 uint8_t const expectedAge = pNode[-1].age;
-                // get_handle_tag() is only called if the check fails.
+                // getHandleTag() is only called if the check fails.
                 FILAMENT_CHECK_POSTCONDITION(expectedAge == age)
                         << "use-after-free of Handle with id=" << handle.getId()
-                        << ", tag=" << get_handle_tag(handle.getId()).c_str_safe();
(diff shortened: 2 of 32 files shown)


filament/backend/include/private/backend/DriverAPI.inc
```diff
@@ -162,6 +162,10 @@ DECL_DRIVER_API_0(finish)
 // reset state tracking, if the driver does any state tracking (e.g. GL)
 DECL_DRIVER_API_0(resetState)
 
+DECL_DRIVER_API_N(setDebugTag,
+        backend::HandleBase::HandleId, handleId,
+        std::string&&, tag)
```
pixelflinger, line 167: we shouldn't use the STL here, ideally. Can we use `utils::CString` instead?

filament/backend/include/private/backend/HandleAllocator.h
```diff
@@ -173,8 +175,10 @@ class HandleAllocator {
                 uint8_t const age = (tag & HANDLE_AGE_MASK) >> HANDLE_AGE_SHIFT;
                 auto const pNode = static_cast<typename Allocator::Node*>(p);
                 uint8_t const expectedAge = pNode[-1].age;
-                FILAMENT_CHECK_POSTCONDITION(expectedAge == age) <<
-                        "use-after-free of Handle with id=" << handle.getId();
+                std::optional<std::string_view> const maybeTag = get_handle_tag(handle.getId());
```
pixelflinger, line 178: that's no good because we'll pay the cost of the string allocation and hashmap lookup everytime, but we really only need it when the check fails. I think we can have get_handle_tag() always return a `CString`;  it can return "unknown" or whatever itself, instead of using an optional<>.

filament/backend/include/private/backend/HandleAllocator.h
```diff
@@ -201,6 +205,32 @@ class HandleAllocator {
         return handle_cast<Dp>(const_cast<Handle<B>&>(handle));
     }
 
+    void associate_handle(HandleBase::HandleId id, std::string&& tag) noexcept {
+        // TODO: for now, only pool handles check for use-after-free, so we only keep tags for
+        // those
+        if (isPoolHandle(id)) {
+            // Truncate the tag's age to N bits.
+            constexpr uint8_t N = 2; // support a history of 4 tags
```
pixelflinger, line 213: so that's just so we don't use too much memory per handle? i.e. we keep 4 instead of 16?

filament/backend/include/private/backend/HandleAllocator.h
```diff
@@ -363,6 +393,7 @@ class HandleAllocator {
     // Below is only used when running out of space in the HandleArena
     mutable utils::Mutex mLock;
     tsl::robin_map<HandleBase::HandleId, void*> mOverflowMap;
+    tsl::robin_map<HandleBase::HandleId, std::string> mDebugTags;
```
pixelflinger, line 396: yeah I Would use CString for this, it has less overhead because it's not trying to do the small string optimization, which we don't need here.

filament/backend/include/private/backend/DriverAPI.inc
pixelflinger: also, I *think* you could use a by-value parameter here. This will leave the choice to the caller.
i.e. the caller can do std::move() if they wish and the cost will be identical. On the other hand, if they don't want to move the string, they also can.

filament/include/filament/Texture.h
```diff
@@ -202,6 +202,17 @@ class UTILS_PUBLIC Texture : public FilamentAPI {
...
+        /**
+         * Associate an optional tag with this Texture for debugging purposes.
+         *
+         * name will show in error messages and should be kept as short as possible.
+         * The name string must persist until build() has been called on this Builder.
+         *
+         * @param name A short string to identify this Texture
+         * @param len Length name tag, or 0 to compute the length if name is null-terminated
+         */
+        Builder& name(const char* UTILS_NONNULL name, size_t len = 0) noexcept;
```
pixelflinger, line 214: Also because we will make a copy anyways, there is no benefit in having the restriction that the name must be valid until build() is called. I think we just document that we are making a copy and internally it's just a CString.

With this API in the future we can imagine we do something more efficient that doesn't even need a heap allocation (e.g. we could pass the name in the command stream, and we could copy it into an internal pool).

I think we should restrict the size of `name` to a maximum size. Maybe 128 bytes?

filament/backend/include/private/backend/HandleAllocator.h
```diff
@@ -201,6 +205,32 @@ class HandleAllocator {
...
+        if (isPoolHandle(id)) {
+            // Truncate the tag's age to N bits.
+            constexpr uint8_t N = 2; // support a history of 4 tags
+            constexpr uint8_t mask = (1 << N) - 1;
+
+            uint8_t const age = (id & HANDLE_AGE_MASK) >> HANDLE_AGE_SHIFT;
+            uint8_t const newAge = age & mask;
+            uint32_t const key = (id & ~HANDLE_AGE_MASK) | (newAge << HANDLE_AGE_SHIFT);
+
+            mDebugTags[key] = std::move(tag);
```
pixelflinger, line 220: This line is the costly one and it will potentially be done for each handle created. Maybe we can make a note of that here. It's important that it doesn't allocate on the heap too often; so maybe we do a reserve for like 500 or 1000 entries in the ctor?
We could have a note that we could potentially use a custom allocator.

pixelflinger (changes requested):

filament/backend/include/private/backend/HandleAllocator.h
bejado: Yeah exactly. Not sure if we want to make this configurable by the user.

filament/backend/include/private/backend/DriverAPI.inc
bejado: Yeah, let me see if `utils::CString` will work.

> the caller can do std::move() if they wish and the cost will be identical

This may be pedantic but a by-value parameter means the compiler still has to invoke the string's move constructor (though it's probably cheap), whereas passing by rvalue reference does not. Probably a wash in terms of performance though.

filament/backend/include/private/backend/HandleAllocator.h
```diff
@@ -173,8 +174,10 @@ class HandleAllocator {
                 uint8_t const age = (tag & HANDLE_AGE_MASK) >> HANDLE_AGE_SHIFT;
                 auto const pNode = static_cast<typename Allocator::Node*>(p);
                 uint8_t const expectedAge = pNode[-1].age;
-                FILAMENT_CHECK_POSTCONDITION(expectedAge == age) <<
-                        "use-after-free of Handle with id=" << handle.getId();
+                // get_handle_tag() is only called if the check fails.
+                FILAMENT_CHECK_POSTCONDITION(expectedAge == age)
+                        << "use-after-free of Handle with id=" << handle.getId()
+                        << ", tag=" << get_handle_tag(handle.getId()).c_str_safe();
```
pixelflinger, line 180: I should have caught this before (sorry), but this should be `getHandleTag()`.

filament/backend/include/private/backend/HandleAllocator.h
```diff
@@ -201,6 +204,34 @@ class HandleAllocator {
         return handle_cast<Dp>(const_cast<Handle<B>&>(handle));
     }
 
+    void associate_handle(HandleBase::HandleId id, utils::CString&& tag) noexcept {
```
pixelflinger, line 207: probably `associateTagToHandle()` ?

filament/src/details/BufferObject.cpp
```diff
@@ -45,6 +50,15 @@ BufferObject::Builder& BufferObject::Builder::bindingType(BindingType bindingTyp
     return *this;
 }
 
+BufferObject::Builder& BufferObject::Builder::name(const char* name, size_t len) noexcept {
+    if (!name) {
+        return *this;
+    }
+    size_t const length = std::min(len == 0 ? strlen(name) : len, size_t { 128u });
```
pixelflinger, line 57: Not sure it's safe to use zero as meaning strlen() because that's a legitimate value that a string_view could return.
So either you have an override that doesn't take a length and then you know you need to do a strlen(), or we need to always require the length to be passed.

pixelflinger (changes requested):

filament/src/details/BufferObject.cpp
bejado: For now, let's just always require it to be passed

filament/backend/include/private/backend/HandleAllocator.h
bejado: Some of the other methods are snake case, so wasn't sure :)

pixelflinger (changes requested):
I would like to factorize the implementation of `name()` somehow. I don't like having this code repeated so many times.

filament/src/FilamentAPI-impl.h
```diff
@@ -26,4 +26,16 @@
...
 
+#include <algorithm>
+
+template <typename Builder>
+Builder& filament::NameMixin<Builder>::name(const char* _Nonnull name, size_t len) noexcept {
+    if (!name) {
+        return static_cast<Builder&>(*this);
+    }
+    size_t const length = std::min(len, size_t { 128u });
+    mName = utils::CString(name, length);
```
pixelflinger, line 37: none of this needs to know about `Builder`, so this can be in a cpp file into a static function outside of NameMixIn -- with the current code, all this will be duplicated and increase code size.

e.g. could be:
```
::filament::builderMakeName(&mName, name, len);
return static_cast<Builder&>(*this);
```

pixelflinger (changes requested):

filament/include/filament/FilamentAPI.h
```diff
@@ -54,6 +56,18 @@ class UTILS_PUBLIC FilamentAPI {
...
 
+template <typename Builder>
+class UTILS_PUBLIC NameMixin {
+public:
+    Builder& name(const char* name, size_t len) noexcept;
+
+    std::optional<utils::CString> getName() const noexcept { return mName; }
+
+private:
+    std::optional<utils::CString> mName;
```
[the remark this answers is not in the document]
bejado: I was trying to distinguish between a tag that isn't set versus one with empty length. But yeah, it's probably fine to say that empty tag is the same as not setting a tag.

filament/include/filament/Texture.h
```diff
@@ -215,7 +215,7 @@ class UTILS_PUBLIC Texture : public FilamentAPI {
          * @param len Length of name, should be less than or equal to 128
          * @return This Builder, for chaining calls.
          */
-        Builder& name(const char* UTILS_NONNULL name, size_t len) noexcept;
+         // Builder& name(const char* UTILS_NONNULL name, size_t len) noexcept;
```
pixelflinger, line 218: is this here on purpose for documentation, or a left over?

filament/include/filament/FilamentAPI.h
```diff
@@ -54,6 +55,22 @@ class UTILS_PUBLIC FilamentAPI {
...
+
+template <typename Builder>
+class UTILS_PUBLIC BuilderNameMixin {
+public:
+    Builder& name(const char* name, size_t len) noexcept {
+        builderMakeName(mName, name, len);
+        return static_cast<Builder&>(*this);
+    }
+
+    utils::CString getName() const noexcept { return mName; }
```
pixelflinger, line 68: `CString const&` ?

pixelflinger: if you wanted to be fancy you could even have a version like this:

`utils::CString&& getName() const noexcept & { return mName; }`

that makes the function returns a movable reference if `this` is a temporary :-)

filament/src/FilamentBuilder.cpp
```diff
@@ -0,0 +1,32 @@
...
+
+namespace filament {
+
+void builderMakeName(utils::CString& outName, const char* name, size_t len) noexcept {
+    if (!name) {
+        return;
+    }
+    size_t const length = std::min(len, size_t { 128u });
+    outName = utils::CString(name, length);
+    return;
```
pixelflinger, line 29: don't need `return` here.

pixelflinger (changes requested):

PR #8045, opened 2024-08-15
fix several issue with the debug datasource in View

pixelflinger:
- the last View created was always overriding previous View's datasource
- because of lazy registering of the data source it was possible that the registering lambda was called after the view was destroyed, leading to crashes
- all view would share the same PID parameters and these would be initialized to default value instead of the user provided value. so debug build would behave differently.

With this change we improve things:
- now only the first view gets to publish its data source. it's still not ideal, but works for our use case with gltf_Viewer
- the view can now unregister itself when it's destroyed
- only the view that successfully registered uses the debug PID values and publishes its data source.
- the normal parameters are used until we query the datasource (from imgui), so by default the behavior is now identical to release builds

This fixes a crash in gltf_viewer when opening the Debug panel.

### DIFF
--- a/filament/src/details/View.h
+++ b/filament/src/details/View.h
@@ -576,7 +576,12 @@ private:
                                                     }};
 
 #ifndef NDEBUG
-    std::unique_ptr<std::array<DebugRegistry::FrameHistory, 5*60>> mDebugFrameHistory;
+    struct DebugState {
+        std::unique_ptr<std::array<DebugRegistry::FrameHistory, 5*60>> debugFrameHistory{};
+        bool owner = false;
+        bool active = false;
+    };
+    std::shared_ptr<DebugState> mDebugState{ new DebugState };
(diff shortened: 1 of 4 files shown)


bejado (approved):

filament/src/details/View.h
```diff
@@ -576,7 +576,12 @@ class FView : public View {
                                                     }};
 
 #ifndef NDEBUG
-    std::unique_ptr<std::array<DebugRegistry::FrameHistory, 5*60>> mDebugFrameHistory;
+    struct DebugState {
+        std::unique_ptr<std::array<DebugRegistry::FrameHistory, 5*60>> debugFrameHistory{};
```
bejado, line 580: Maybe use a constant for `5*60` since it's also used in View.cpp

poweifeng (approved):

PR #8046, opened 2024-08-15
vk: fix missing read-after-write barrier

poweifeng:
We relied on layout transition to emit a barrier, but sometimes
it's skipped. In that case, we explicitly emit a barrier for the
transition from attachment -> sampler.

Without this, we'll hit read-after-write sync validation error
and produce various artifacts (e.g a fade to white effect) on
different platforms.


pixelflinger (changes requested):
Please add a better description for the change and/or link to a bug that explain what the problem was and how it is fixed.

poweifeng:
> Please add a better description for the change and/or link to a bug that explain what the problem was and how it is fixed.

done

pixelflinger (approved):

PR #8055, opened 2024-08-19
improve FilamentApp split-view mode

pixelflinger:
Screenshot 2024-08-18 at 23 57 34


filament/src/ShadowMapManager.cpp
```diff
@@ -1078,4 +1078,15 @@ ShadowMapManager::CascadeSplits::CascadeSplits(Params const& params) noexcept
     }
 }
 
+utils::FixedCapacityVector<Camera const*>
+ShadowMapManager::getDirectionalShadowCamera() const noexcept {
```
bejado, line 1082: `getDirectionalShadowCameras`?

bejado (approved):

PR #8071, opened 2024-08-21
decoding meshoptimized gltf just once

mbalajee:
Decoding the mesh optimized gltf in a loop slows down the extended algo drastically & crashes the app with larger primitives

### DIFF
--- a/libs/gltfio/src/extended/AssetLoaderExtended.cpp
+++ b/libs/gltfio/src/extended/AssetLoaderExtended.cpp
@@ -527,8 +527,10 @@ bool AssetLoaderExtended::createPrimitive(Input* input, Output* out,
 
     if (!mCgltfBuffersLoaded) {
         mCgltfBuffersLoaded = utility::loadCgltfBuffers(gltf, mGltfPath.c_str(), mUriDataCache);
+        if (!mCgltfBuffersLoaded) {
+            return false;
+        }
         utility::decodeMeshoptCompression(gltf);
-        if (!mCgltfBuffersLoaded) return false;
     }
 
     utility::decodeDracoMeshes(gltf, prim, input->dracoCache);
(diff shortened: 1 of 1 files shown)


mbalajee:
@pirlande-fw

libs/gltfio/src/extended/AssetLoaderExtended.cpp
```diff
@@ -527,11 +527,11 @@ bool AssetLoaderExtended::createPrimitive(Input* input, Output* out,
 
     if (!mCgltfBuffersLoaded) {
         mCgltfBuffersLoaded = utility::loadCgltfBuffers(gltf, mGltfPath.c_str(), mUriDataCache);
+        utility::decodeMeshoptCompression(gltf);
         if (!mCgltfBuffersLoaded) return false;
```
poweifeng, line 531:
```suggestion
        if (!mCgltfBuffersLoaded) {
            return false;
        }
        utility::decodeMeshoptCompression(gltf);
```

libs/gltfio/src/extended/AssetLoaderExtended.cpp
```diff
@@ -527,11 +527,11 @@ bool AssetLoaderExtended::createPrimitive(Input* input, Output* out,
 
     if (!mCgltfBuffersLoaded) {
         mCgltfBuffersLoaded = utility::loadCgltfBuffers(gltf, mGltfPath.c_str(), mUriDataCache);
+        utility::decodeMeshoptCompression(gltf);
```
pixelflinger, line 530: What if this function is called with a different `input->gltf`?

mbalajee: that would be true for line 529 too right? this function is called from only one place, at-least with the current implementation,  `FAssetLoader::createPrimitives` which passes the `gltf` from root asset.  May be do both `loadCgltfBuffers` & `decodeMeshoptCompression` at the calling site (FAssetLoader::createPrimitives) similar to `ResourceLoader::loadResources`?

poweifeng: yeah, I need to look at this closer I think.  For now, this patch is good to go.

poweifeng (approved):

pixelflinger (approved):

PR #8073, opened 2024-08-21
Fix, assertion when an external image is used but not yet set

bejado:


filament/src/details/Texture.cpp
```diff
@@ -528,20 +516,53 @@ void FTexture::updateLodRange(uint8_t baseLevel, uint8_t levelCount) noexcept {
     }
 }
 
+void FTexture::setHandles(backend::Handle<backend::HwTexture> handle) noexcept {
+    assert_invariant(!(mHandle && !mHandleForSampling));
```
pixelflinger, line 520: `(!mHandle || mHandleForSampling)` is maybe easier to understand?

filament/src/details/Texture.cpp
```diff
@@ -528,20 +516,53 @@ void FTexture::updateLodRange(uint8_t baseLevel, uint8_t levelCount) noexcept {
...
+    driver.update3DImage(handle, 0, 0, 0, 0, 1, 1, 1,
+            { (char*)&pixels[0], sizeof(pixels),
+                    Texture::PixelBufferDescriptor::PixelDataFormat::RGBA,
+                    Texture::PixelBufferDescriptor::PixelDataType::UBYTE });
+    return handle;
+}
+
 backend::Handle<backend::HwTexture> FTexture::getHwHandleForSampling() const noexcept {
+    if (UTILS_UNLIKELY(mTarget == SamplerType::SAMPLER_EXTERNAL && !mHandleForSampling)) {
+        setHandleForSampling(createPlaceholderTexture(*mDriver));
```
pixelflinger, line 553: setHandleForSampling() could return `mHandleForSampling`, so here you could `return setHandleForSampling()`

pixelflinger (approved):

PR #8074, opened 2024-08-22
Refactored the handling of usage flags in the VulkanTexture constructor

mdagois:
- Removed the `VK_IMAGE_USAGE_TRANSFER_SRC_BIT` and `VK_IMAGE_USAGE_TRANSFER_DST_BIT` flags that were systematically added to all color and depth attachments.
- Modified the behavior of `TextureUsage::BLIT_SRC`, `TextureUsage::BLIT_DST`, and `TextureUsage::UPLOADABLE`, so that they only raise the corresponding `VkImage` usage flags.
- Note that I shied away from removing `VK_IMAGE_USAGE_SAMPLED_BIT` from depth texture with more than 1 sample, as I am not familiar enough with the custom shader-based resolve. I believe we could move the flag to where the depth sidecar is created (using `TextureUsage::SAMPLEABLE`), but I am not sure 100%.

### DIFF
--- a/filament/backend/src/vulkan/VulkanHandles.cpp
+++ b/filament/backend/src/vulkan/VulkanHandles.cpp
@@ -16,7 +16,6 @@
 
 #include "VulkanHandles.h"
 
-#include "VulkanDriver.h"
 #include "VulkanConstants.h"
 #include "VulkanDriver.h"
 #include "VulkanMemory.h"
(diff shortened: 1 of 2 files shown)


mdagois:
NOTE: this PR is meant as a preliminary cleanup for the [Support for transient attachments](https://github.com/google/filament/pull/8021).

pixelflinger (approved):
looks good but would like @poweifeng to review as well

poweifeng (approved):

PR #8080, opened 2024-08-24
fix screen-space reflection when sub-pass colorgrading is used 

pixelflinger:
The problem was that in that case, SSR was using the colorgraded
color buffer as history.

### DIFF
--- a/filament/src/RendererUtils.cpp
+++ b/filament/src/RendererUtils.cpp
@@ -153,6 +153,8 @@ FrameGraphId<FrameGraphTexture> RendererUtils::colorPass(
                 }
 
                 if (colorGradingConfig.asSubpass) {
+                    assert_invariant(config.msaa <= 1);
+                    assert_invariant(colorBufferDesc.samples <= 1);
                     data.output = builder.createTexture("Tonemapped Buffer", {
                             .width = colorBufferDesc.width,
                             .height = colorBufferDesc.height,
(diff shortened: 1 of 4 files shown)


pixelflinger:
note: it might be easier to review each CL separately. The first one fixes the problem, the 2nd one cleans up.

filament/src/RendererUtils.cpp
```diff
@@ -292,57 +290,60 @@ std::pair<FrameGraphId<FrameGraphTexture>, bool> RendererUtils::refractionPass(
...
+                ssrConfig.refraction,
+                true, ssrConfig);
 
         // Now we're doing the refraction pass proper.
-        // This uses the same framebuffer (color and depth) used by the opaque pass. This happens
-        // automatically because these are set in the Blackboard (they were set by the opaque
-        // pass). For this reason, `desc` below is only used in colorPass() for the width and
-        // height.
+        // This uses the same framebuffer (color and depth) used by the opaque pass.
+        // For this reason, `desc` below is only used in colorPass() for the width and height.
```
poweifeng, line 318: "`desc`" seems like it's out of date comment?

poweifeng (approved):

filament/src/RendererUtils.cpp
pixelflinger: fixed

PR #8082, opened 2024-08-26
Metal: annotate memory allocation failures with buffer tag

bejado:
BUGS=[349683761, 348452642]


filament/backend/src/metal/MetalDriver.mm
```diff
@@ -2073,19 +2099,17 @@
 }
 
 void MetalDriver::setDebugTag(HandleBase::HandleId handleId, utils::CString tag) {
+    utils::slog.d << "Setting debug tag of " << handleId << " to " << tag.c_str_safe() << utils::io::endl;
```
pixelflinger, line 2102: remove this

filament/backend/src/metal/MetalDriver.mm
```diff
@@ -885,6 +908,9 @@
 void MetalDriver::updateIndexBuffer(Handle<HwIndexBuffer> ibh, BufferDescriptor&& data,
         uint32_t byteOffset) {
     auto* ib = handle_cast<MetalIndexBuffer>(ibh);
+
+    mHandleAllocator.getHandleTag(ibh.getId());
+
```
pixelflinger, line 913: what is this for? it's not using the return value?

pixelflinger (changes requested):

pixelflinger (approved):

PR #8098, opened 2024-08-30
repair the "no buffer padding case"

pixelflinger:
with a previous change we were too aggressive in falling back to  "no buffer padding", we need to do this only in the case where "as subpass" would have been used if supported by the h/w.

### DIFF
--- a/filament/src/details/Renderer.cpp
+++ b/filament/src/details/Renderer.cpp
@@ -651,14 +651,19 @@ void FRenderer::renderJob(RootArenaScope& rootArenaScope, FView& view) {
 
     const bool isProtectedContent =  mSwapChain && mSwapChain->isProtected();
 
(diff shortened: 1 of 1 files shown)


filament/src/details/Renderer.cpp
```diff
@@ -651,14 +651,19 @@ void FRenderer::renderJob(RootArenaScope& rootArenaScope, FView& view) {
 
     const bool isProtectedContent =  mSwapChain && mSwapChain->isProtected();
 
+    // Conditions to meet to be able to use the sub-pass rendering path. This is regardless of
+    // whether the backend supports subpasses (or if they are disabled but the debugRegistry)
```
poweifeng, line 655:
```suggestion
    // whether the backend supports subpasses (or if they are disabled in the debugRegistry).
```

poweifeng (approved):

PR #8100, opened 2024-08-30
Enable mmap on iOS

bejado:
Don't remember why this was ever disabled, seems supported on iOS.

BUGS=[362965958]


pixelflinger (approved):

pixelflinger (changes requested):
please include the bug number in the description

pixelflinger (approved):

PR #8103, opened 2024-09-03
github: Split Android CI to per ABI builds

poweifeng:
- Pulled the android continuous workflow into an action
 - Split the android continuous build into 3 ABIs armv7, armv8a,
   and x86_64 (note that x86 is not present).
 - Remove the upload artifacts step from previous workflow.
   Unclear why that was present.
 - Split the android continous into debug build and a release
   build commands, enabling deletion of intermediate output
   directory.

### DIFF
--- a/.github/actions/android-continuous/action.yml
+++ b/.github/actions/android-continuous/action.yml
@@ -1,0 +1,17 @@
+name: 'Android Continuous'
+inputs:
+  build-abi:
+    description: 'The target platform ABI'
+    required: true
+    default: 'armeabi-v7a'
+runs:
+  using: "composite"
+  steps:
+    - uses: actions/setup-java@v3
+      with:
+        distribution: 'temurin'
+        java-version: '17'
+    - name: Run build script
+      run: |
+        cd build/android && printf "y" | ./build.sh continuous ${{ inputs.build-abi }}
+      shell: bash
--- a/.github/workflows/android-continuous.yml
+++ b/.github/workflows/android-continuous.yml
@@ -8,32 +8,35 @@ on:
       - rc/**
 
 jobs:
-  build-android:
-    name: build-android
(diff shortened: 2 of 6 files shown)


poweifeng:
Please don't review yet, I'm running this to get the workflows to run on presubmit as a test.

poweifeng:
> Please don't review yet, I'm running this to get the workflows to run on presubmit as a test.

Now ready for review

pixelflinger:
can you add a detailed description of what this does? and/or link to bug in buganizer with the details. thanks!

poweifeng:
> can you add a detailed description of what this does? and/or link to bug in buganizer with the details. thanks!

Done. Description was in the commit, but not sure why it didn't make it to the PR description.

bejado (approved):

bejado:
> Remove the upload artifacts step from previous workflow. Unclear why that was present.

This was to allow clients to use continuous builds:

https://github.com/google/filament/blob/cdb539b3cf13774c0deb9a091ed734e72bf78a7d/README.md?plain=1#L57-L65

I'm fine with getting rid of these if it makes our GitHub Action code easier to mantain. But we should also probably remove this section from our readme.

poweifeng:
> > Remove the upload artifacts step from previous workflow. Unclear why that was present.
> 
> This was to allow clients to use continuous builds:
> 
> https://github.com/google/filament/blob/cdb539b3cf13774c0deb9a091ed734e72bf78a7d/README.md?plain=1#L57-L65
> 
> I'm fine with getting rid of these if it makes our GitHub Action code easier to maintain. But we should also probably remove this section from our readme.

I see.  Thanks for the background.  Yeah, I'm going to remove them for now and we can revisit later.  (Because a typical `.aar` will need all abis I think, which is not possible with how I'm splitting them).

PR #8106, opened 2024-09-05
Expose getShadowType on View.

voithos:
While this method exists on FView, it is missing from View and thus not accessible for filament users. This adds a pass-through method on View.


bejado (approved):
Thanks!

PR #8110, opened 2024-09-09
fix FrameSkipper

pixelflinger:
FrameSkipper was recently broken because of a typo resulting in  the frame latency being always 3 instead of the default of 2.

This change also makes the maximum latency 2 instead of 3, because on ANDROID, 3 can cause CPU throttling at seemingly random places on the GL thread.

### DIFF
--- a/filament/src/FrameSkipper.h
+++ b/filament/src/FrameSkipper.h
@@ -34,7 +34,7 @@ namespace filament {
 class FrameSkipper {
     /*
      * The maximum frame latency acceptable on ANDROID is 2 because higher latencies will be
-     * throttled anyway is BufferQueueProducer::dequeueBuffer(), because ANDROID is generally
+     * throttled anyway in BufferQueueProducer::dequeueBuffer(), because ANDROID is generally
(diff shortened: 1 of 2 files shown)


filament/src/FrameSkipper.h
```diff
@@ -32,7 +32,18 @@ namespace filament {
  * outrun the GPU.
  */
 class FrameSkipper {
-    static constexpr size_t MAX_FRAME_LATENCY = 3;
+    /*
+     * The maximum frame latency acceptable on ANDROID is 2 because higher latencies will be
+     * throttled anyway is BufferQueueProducer::dequeueBuffer(), because ANDROID is generally
```
poweifeng, line 37: "throttled away in" ?

poweifeng (approved):

filament/src/FrameSkipper.h
pixelflinger, line 37:
```suggestion
     * throttled anyway in BufferQueueProducer::dequeueBuffer(), because ANDROID is generally
```

PR #8115, opened 2024-09-10
Check for invalid texture inside MaterialInstance

bejado:
@pixelflinger would you prefer if `FMaterialInstance:` holds on to a `FEngine&` reference instead of passing it each time to `commit`?


pixelflinger:
> @pixelflinger would you prefer if `FMaterialInstance:` holds on to a `FEngine&` reference instead of passing it each time to `commit`?

No, I prefer to pass it each time.

pixelflinger (approved):

pixelflinger:
@bejado I have modified the PR to take the `FEngine` from the `FMaterial`, so no need to pass it around everywhere.

z3moon (approved):

poweifeng (approved):

PR #8118, opened 2024-09-11
Fix a crash for color grading

z3moon:
Set postprocess descriptor set for color grading


bejado (approved):

pixelflinger (approved):
I don't quite understand why this is necessary because `bindPostProcessDescriptorSet()` is already called in `colorGradingPrepareSubpass()`.

Let's check this in to unblock things right now, but I will take a look later.

z3moon:
It's called from Renderer.cpp https://github.com/google/filament/blob/main/filament/src/details/Renderer.cpp#L1056

PR #8122, opened 2024-09-12
vk: cache descriptor set layout

poweifeng:
We need to cache descriptor set layouts because Filament creates new layout objects for the same layout content. Adding to this is that descriptor set's lifetime is tied to the layout that it is created from, and we are trying to recycle used descriptor sets wrt layout. This implies that we need to (unfortunately) separate the vk layout lifetime from the filament layout lifetime.

In this commit, we use a cache so that the same layout content will always map to one vk layout.

Also fixed bugs on bit-shifting and counting.


poweifeng:
Closing due to wrong branch.

PR #8126, opened 2024-09-13
vk: workaround a renderStandaloneView issue

poweifeng:
Found through testing that renderStandaloneView+vk+swiftshader seems to cause synchronization issues, which results in incorrect rendering. Here we workaround the issue by forcibly flush and wait per renderStandaloneView call.

BUG=361822355

### DIFF
--- a/filament/src/details/Renderer.cpp
+++ b/filament/src/details/Renderer.cpp
@@ -536,6 +536,12 @@ void FRenderer::renderStandaloneView(FView const* view) {
         renderInternal(view);
 
         driver.endFrame(mFrameId);
+
+        // This is a workaround for internal bug b/361822355.
+        // TODO: properly address the bug and remove this workaround.
+        if (engine.getBackend() == backend::Backend::VULKAN) {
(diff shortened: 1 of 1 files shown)


z3moon (approved):

filament/src/details/Renderer.cpp
```diff
@@ -536,6 +536,12 @@ void FRenderer::renderStandaloneView(FView const* view) {
         renderInternal(view);
 
         driver.endFrame(mFrameId);
+
+        // This is a workaround for internal bug b/361822355.
+        // TODO: properly address the bug and remove this workaround.
+        if (engine.getBackend() == backend::Backend::VULKAN) {
+            engine.flushAndWait();
```
pixelflinger, line 543: I think that's too agressive. You should be able to just do a `driver.flush()` before the `renderInternal()` call.

pixelflinger (changes requested):

pixelflinger (approved):

PR #8130, opened 2024-09-13
Add support for exr files in gltf_Vvewer

pixelflinger:


libs/filamentapp/src/IBL.cpp
```diff
@@ -37,6 +39,8 @@
 
 #include <string.h>
 
+#include <utils/Log.h>
```
poweifeng, line 42: not used?

poweifeng (approved):

PR #8138, opened 2024-09-18
Do not fix missing samplers for textures with mutable handles

poweifeng:
Because these textures are not going to be present when the sampler "fixup" occurs.


bejado (approved):

poweifeng:
Superceded by #8144

PR #8140, opened 2024-09-19
fix fragment/vertex shader descriptor set layout mismatch

pixelflinger:
The SSR variant uses a custom fragment shader but the generic vertex shader. For this reason their descriptor set layout for the "per view" set end-up being different (the fragment shader one has less descriptors in it and the descriptor itself is create from the SSR variant data).  In the end, we end up with a descriptor set that's too small for the vertex shader and Metal was rightly complaining.

We fix this by:
- removing the descriptors unused in the vertex shader in the generic case
- adding the remaining descriptors in the SSR version of the layout

In practice this only adds one UBO (for shadows). That UBO is actually not even used in the SSR variants, but the (generic) vertex shader  layout has it, so it must be provided. WE do this by providing a dummy UBO, which will never be used anyways.

In addition, we add a check to matc which will make sure such mismatch won't happen again.

### DIFF
--- a/filament/src/ds/SsrPassDescriptorSet.cpp
+++ b/filament/src/ds/SsrPassDescriptorSet.cpp
@@ -59,7 +59,7 @@ void SsrPassDescriptorSet::setFrameUniforms(TypedUniformBuffer<PerViewUib>& unif
             uniforms.getUboHandle(), 0, uniforms.getSize());
 
     // This is actually not used for the SSR variants, but the descriptor-set layout needs
-    // to have this UBO because the fragment's shader used is the "generic" one. Both Metal
+    // to have this UBO because the fragment shader used is the "generic" one. Both Metal
     // and GL would be okay without this, but Vulkan's validation layer would complain.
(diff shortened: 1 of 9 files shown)


filament/src/ds/SsrPassDescriptorSet.cpp
```diff
@@ -42,17 +42,27 @@ SsrPassDescriptorSet::SsrPassDescriptorSet() noexcept = default;
...
+    driver.destroyBufferObject(mShadowUbh);
 }
 
 void SsrPassDescriptorSet::setFrameUniforms(TypedUniformBuffer<PerViewUib>& uniforms) noexcept {
     // initialize the descriptor-set
     mDescriptorSet.setBuffer(+PerViewBindingPoints::FRAME_UNIFORMS,
             uniforms.getUboHandle(), 0, uniforms.getSize());
 
+    // This is actually not used for the SSR variants, but the descriptor-set layout needs
+    // to have this UBO because the fragment's shader used is the "generic" one. Both Metal
```
poweifeng, line 62:
```suggestion
    // to have this UBO because the fragment shader used is the "generic" one. Both Metal
```

poweifeng (approved):

PR #8143, opened 2024-09-19
github: make windows build use all available cores

poweifeng:
- Also used a smaller runner as the gains from the 32-core was
      not efficient when comparing output times.
    - Clean-up:
        - Rename the android continuous to a proper name
        - set 'echo on' for the Windows release build so we'll know
          why the output asset does not get "moved" correclty.


poweifeng:
ready for review

bejado (approved):

PR #8151, opened 2024-09-24
Fix broken layerCount for some types of textures

z3moon:
Currently mLayerCount for RenderTarget is always updated to the number of depth for attachments, which incurs unintended behaviors for some types of textures. i.e., 2d array, cubemap array, and 3d textures.

Fix this by updating mLayerCount only for multiview use case.

BUGS=[369165616]

### DIFF
--- a/filament/src/details/RenderTarget.cpp
+++ b/filament/src/details/RenderTarget.cpp
@@ -44,7 +44,11 @@ struct RenderTarget::BuilderDetails {
     uint32_t mWidth{};
     uint32_t mHeight{};
     uint8_t mSamples = 1;   // currently not settable in the public facing API
+    // The number of layers for the render target. The value should be 1 except for multiview.
+    // If `mMultiview` is true, this value is appropriately updated based on the depth of textures.
+    // Hence, #>1 means using multiview
     uint8_t mLayerCount = 1;
+    bool mMultiview = false;
 };
 
 using BuilderType = RenderTarget;
@@ -74,6 +78,13 @@ RenderTarget::Builder& RenderTarget::Builder::layer(AttachmentPoint pt, uint32_t
     mImpl->mAttachments[(size_t)pt].layer = layer;
     return *this;
 }
+
+RenderTarget::Builder& RenderTarget::Builder::multiview(AttachmentPoint pt, uint8_t baseLayer/*= 0*/) noexcept {
+    mImpl->mAttachments[(size_t)pt].layer = baseLayer;
+    mImpl->mMultiview = true;
+    return *this;
+}
+
 
 RenderTarget* RenderTarget::Builder::build(Engine& engine) {
     using backend::TextureUsage;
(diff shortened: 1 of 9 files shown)


filament/src/details/RenderTarget.cpp
```diff
@@ -105,28 +116,42 @@ RenderTarget* RenderTarget::Builder::build(Engine& engine) {
...
+                        << "Texture sampler must be of 2d array for multiview";
+                FILAMENT_CHECK_PRECONDITION(attachment.layer + attachment.layerCount <= d)
+                        << "For multiview, layer + layerCount cannot exceed the number of depth";
+                if (layerCountMultiview) {
+                    FILAMENT_CHECK_PRECONDITION(layerCountMultiview == attachment.layerCount)
+                            << "layerCount for multiview should match";
+                } else {
+                    layerCountMultiview = attachment.layerCount;
+                }
+            }
```
pixelflinger, line 137: instead of doing this, I would calculate the min and max layerCount and do the checks outside of the loop:

CHECK( min == max ) // all counts must match
CHECK( min > 1 && SAMPLER_2D || min <= 1 ) // multiview only with sampler 2D

The (attachment.layer + attachment.layerCount <= d) check can be always done (even when attachment.layerCount ==0).

pixelflinger (changes requested):

pixelflinger (approved):

filament/src/details/RenderTarget.cpp
```diff
@@ -105,28 +116,40 @@ RenderTarget* RenderTarget::Builder::build(Engine& engine) {
...
-    uint32_t minDepth = std::numeric_limits<uint32_t>::max();
-    uint32_t maxDepth = 0;
+    uint32_t minLayerCount = std::numeric_limits<uint32_t>::max();
+    uint32_t maxLayerCount = 0;
     for (auto const& attachment : mImpl->mAttachments) {
         if (attachment.texture) {
             const uint32_t w = attachment.texture->getWidth(attachment.mipLevel);
             const uint32_t h = attachment.texture->getHeight(attachment.mipLevel);
             const uint32_t d = attachment.texture->getDepth(attachment.mipLevel);
+            const uint32_t l = attachment.layerCount;
```
poweifeng, line 126: technically speaking, shouldn't we always have at least one layer?  Should this be `layer` here instead of `layerCount`?

poweifeng: sorry I disabled auto-merge while I still have this one question.

z3moon: layerCount here means the number of layers used for multiview. This variable indicates whether we use multiview . If this value is 0, then it means we don't use multiview for this attachment.

poweifeng: why not just make it the same as `mLayerCount` in `BuilderDetails` so that `> 1` means that multiview is enabled?  So we don't have two different interpretations of the same variable name.

poweifeng (approved):

filament/src/details/RenderTarget.cpp
z3moon: That was my original implementation. But the new interface `.multiview(attachpoint, layerCount, startingLayer)` requires each attachment point to have a layerCount.

PR #8161, opened 2024-09-25
gl: add PlatformOSMesa as an offscreen context

poweifeng:
For GL+Linux, PlatformGLX will try to open an X11 window regardless of whether we are doing headless/offscreen rendering or not.

Here we add an OSMesa platform, which will allow us to avoid opening any window on Linux. This is particularly useful for situation where a display is not available, like for CI.

One important detail is that even though we are displaying through a window, we keep the SDL2 dependency in tact for gltf_viewer. This is due to the fact that gltf_viewer is built upon FilamentApp, which is heavily integrated with SDL2. This is mostly ok since we won't be hitting any path for opening a window due to gltf_viewer's existing support for headless mode.


build.sh
```diff
@@ -64,6 +64,9 @@ function print_help {
     echo "        enabling debug paths in the backend from the build script. For example, make a"
     echo "        systrace-enabled build without directly changing #defines. Remember to add -f when"
     echo "        changing this option."
+    echo "    -X osmesa_path"
+    echo "        Indicates a path to the path of a completed OSMesa build. OSMesa is used to create an"
```
pixelflinger, line 68: "a path to the path" ?

pixelflinger (approved):

build.sh
poweifeng: done

PR #8163, opened 2024-09-26
Remove textureUseAfterFreePoolSize and texture use-after-free

bejado:
Remove `textureUseAfterFreePoolSize`, we'll switch to a more generic use-after-free detection for "heap" handles.


pixelflinger (changes requested):
instead of doing this -- it would be better to add generic use-after-free detection support for all "heap" handles.
this is also a rare case

pixelflinger (approved):

PR #8171, opened 2024-10-02
Work around client descriptor set issues

poweifeng:
- We change GLDescriptorSet::Buffer default constructor to workaround a client's compiler set up issue.
- We removed the assert_invariant that checks that ubo/samplers are not changed after committed in DescriptorSet. This caused an existing client's build to crash.


pixelflinger (approved):
lgtm but let's make sure to come back to this. in particular the extra `bool` in DescriptorSet will increase the size of descriptor sets by 8 bytes.

PR #8182, opened 2024-10-07
Check for use-after-free for heap handles

bejado:


filament/backend/include/private/backend/HandleAllocator.h
```diff
@@ -179,21 +179,40 @@ class HandleAllocator {
...
+            }
         }
 
         return static_cast<Dp>(p);
     }
 
     template<typename B>
     bool is_valid(Handle<B>& handle) {
-        if (handle && isPoolHandle(handle.getId())) {
+        if (!handle) {
```
pixelflinger, line 205: I think a null handle is considered valid (at least it was before). Maybe we should keep that behavior for now?

filament/backend/include/private/backend/HandleAllocator.h
```diff
@@ -179,21 +179,40 @@ class HandleAllocator {
...
+        }
+        if (isPoolHandle(handle.getId())) {
             auto [p, tag] = handleToPointer(handle.getId());
             uint8_t const age = (tag & HANDLE_AGE_MASK) >> HANDLE_AGE_SHIFT;
             auto const pNode = static_cast<typename Allocator::Node*>(p);
             uint8_t const expectedAge = pNode[-1].age;
             return expectedAge == age;
         }
-        return true;
+        return handleToPointer(handle.getId()).first != nullptr;
```
pixelflinger, line 215: we can move the `handleToPointer()` call outside the `if`.

pixelflinger (changes requested):

filament/backend/include/private/backend/HandleAllocator.h
bejado: Hm, it seems strange to me that a null handle is considered valid, but you're right, that's the way it was before.

pixelflinger (approved):

PR #8184, opened 2024-10-07
Make sure color attachment textures have copy-able usage set

poweifeng:
To ensure the source of readPixels() is properly copy-able, we need to make sure textures that are created as color attachments have the right copy usage bit (BLIT_SRC) set.


filament/src/ResourceAllocator.cpp
```diff
@@ -162,6 +162,12 @@ backend::TextureHandle ResourceAllocator::createTexture(const char* name,
         uint32_t width, uint32_t height, uint32_t depth,
         std::array<backend::TextureSwizzle, 4> swizzle,
         backend::TextureUsage usage) noexcept {
+    // If this texture is a color attachment, then make sure it can be the source of a copy for
+    // readPixels().
+    if (any(usage & backend::TextureUsage::COLOR_ATTACHMENT)) {
+        usage |= backend::TextureUsage::BLIT_SRC;
+    }
+
```
pixelflinger, line 170: we shouldn't do this here. for the framegraph, we should use the correct usage bits. also I don't think any of these textures are ever used with readPixels(), but if they are, the pass doing that needs to call `read()` with the proper usage bit.

filament/src/details/Texture.cpp
```diff
@@ -183,6 +183,12 @@ Texture* Texture::Builder::build(Engine& engine) {
         }
     }
 
+    // If this texture is a color attachment, then make sure it can be the source of a copy for
+    // readPixels().
+    if (any(mImpl->mUsage & TextureUsage::COLOR_ATTACHMENT)) {
+        mImpl->mUsage |= TextureUsage::BLIT_SRC;
+    }
```
pixelflinger, line 190: I think we should have a bug for this and have a plan to fix all the call sites and eventually remove this. I guess this is an API change: we need to document that `readPixels()` (and maybe others) need the `BLIT_SRC` usage bit. And we also need to have a PRECONDITON check in readPixels() so that it crashes early instead of in the backends.

pixelflinger (changes requested):

filament/src/details/Texture.cpp
poweifeng: Maybe we don't even need this?  It looks like `readPixels` only works on the default rendertarget: 
https://github.com/google/filament/blob/main/filament/src/details/Renderer.cpp#L451

filament/src/ResourceAllocator.cpp
poweifeng: moved to the right place (on the picking texture).

filament/src/details/Texture.cpp
poweifeng: Updated with warning message and added API comment.

bejado (approved):

pixelflinger (approved):

PR #8187, opened 2024-10-09
Update hardware Handle

z3moon:
In certain compilers, the assignment operators defined as default
doesn't automatically make a call to the parent's method if it's
user-defined.

Make this behavior explicit to avoid this edge case.

BUGS=[371980551]

### DIFF
--- a/filament/backend/include/backend/Handle.h
+++ b/filament/backend/include/backend/Handle.h
@@ -108,7 +108,9 @@ struct Handle : public HandleBase {
     Handle(Handle&& rhs) noexcept = default;
 
     // Explicitly redefine copy/move assignment operators rather than just using default here.
-    // Because it doesn't make a call to the parent's method automatically in certain compilers.
+    // Because it doesn't make a call to the parent's method automatically during the std::move
+    // function call(https://en.cppreference.com/w/cpp/algorithm/move) in certain compilers like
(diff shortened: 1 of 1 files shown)


filament/backend/include/backend/Handle.h
```diff
@@ -106,8 +107,16 @@ struct Handle : public HandleBase {
     Handle(Handle const& rhs) noexcept = default;
     Handle(Handle&& rhs) noexcept = default;
 
-    Handle& operator=(Handle const& rhs) noexcept = default;
-    Handle& operator=(Handle&& rhs) noexcept = default;
+    // Explicitly redefine copy/move assignment operators rather than just using default here.
+    // Because it doesn't make a call to the parent's method automatically in certain compilers.
```
bejado, line 111: That's extremely strange. What compiler(s) are affected? What's the effect? Is `HandleId object` not copied correctly?

z3moon: This issue occurs when `std::move(fences.begin() + 1, fences.end(), fences.begin());` is called (FrameSkipper.cpp:69) for specific NDK versions (<= 25.1.8937393, works fine from 26). Let me update the descriptions and comments.

pixelflinger (approved):

PR #8189, opened 2024-10-09
gles: add a flag to enable the validation of the nativewindow

pixelflinger:
We are seeing a cluster of crashes that could be due to using an EGLSurface whose ANativeWindow has become invalid. This could happen if we continued to use (i.e. draw with) an EGLSurface after  SurfaceHolder::onSurfaceDestroyed() has returned.

This new flag enables an assertion that the native window is valid at the time of makeCurrent(), which happens early in the frame.

BUG=[330392256]

### DIFF
--- a/filament/backend/include/backend/platforms/PlatformEGLAndroid.h
+++ b/filament/backend/include/backend/platforms/PlatformEGLAndroid.h
@@ -89,6 +89,11 @@ protected:
      */
     AcquiredImage transformAcquiredImage(AcquiredImage source) noexcept override;
 
+protected:
+    bool makeCurrent(ContextType type,
+            SwapChain* drawSwapChain,
+            SwapChain* readSwapChain) noexcept override;
+
 private:
     struct InitializeJvmForPerformanceManagerIfNeeded {
         InitializeJvmForPerformanceManagerIfNeeded();
@@ -102,6 +107,10 @@ private:
 
     using clock = std::chrono::high_resolution_clock;
     clock::time_point mStartTimeOfActualWork;
+
+    void* mNativeWindowLib = nullptr;
+    int32_t (*ANativeWindow_getBuffersDefaultDataSpace)(ANativeWindow* window) = nullptr;
+    bool mAssertNativeWindowIsValid = false;
 };
 
 } // namespace filament::backend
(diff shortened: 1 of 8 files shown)


bejado (approved):

filament/backend/include/backend/platforms/PlatformEGLAndroid.h
```diff
@@ -102,6 +107,10 @@ class PlatformEGLAndroid : public PlatformEGL {
 
     using clock = std::chrono::high_resolution_clock;
     clock::time_point mStartTimeOfActualWork;
+
+    void* mNativeWindowLib = nullptr;
+    int32_t (*ANativeWindow_getBuffersDefaultDataSpace)(ANativeWindow* window);
```
poweifeng, line 112: `= nullptr` here to make sure it's initialized?

pixelflinger: good point, done.

pixelflinger:
re-request review because I added a code path for when we don't have API 34.

filament/include/filament/Engine.h
```diff
@@ -380,8 +380,14 @@ class UTILS_PUBLIC Engine {
          * it's a GLES2 context. Ignored on other backends.
          */
         bool forceGLES2Context = false;
-    };
 
+        /**
+         * Assert the native window associated to a SwapChain is valid when calling makeCurrent().
+         * This is only supported for:
+         *      - EGLPlatformAndroid with at least API 34
```
poweifeng, line 387: outdated comment?

filament/include/filament/Engine.h
```diff
@@ -380,8 +380,14 @@ class UTILS_PUBLIC Engine {
...
+         */
+        bool assertNativeWindowIsValid = false;
```
poweifeng, line 389: is this needed in java as well?

poweifeng (approved):

filament/include/filament/Engine.h
pixelflinger: ah yes. probably. thanks.

filament/include/filament/Engine.h
pixelflinger: yep. thanks.

poweifeng (approved):

PR #8193, opened 2024-10-10
vk: remove history from DescriptorSetManager

poweifeng:
"history" is a map from a DescriptorSet pointer to a set of bookkeeping values (we delay "binding" until "commit" so need to keep values until then). Instead of using a map, we can store these values in the DescriptorSet itself so that we save on a map look-up.

### DIFF
--- a/filament/backend/src/vulkan/VulkanHandles.h
+++ b/filament/backend/src/vulkan/VulkanHandles.h
@@ -135,9 +135,13 @@ public:
     using OnRecycle = std::function<void(VulkanDescriptorSet*)>;
 
     VulkanDescriptorSet(VulkanResourceAllocator* allocator, VkDescriptorSet rawSet,
+            UniformBufferBitmask const& dynamicUboMask,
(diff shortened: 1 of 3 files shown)


filament/backend/src/vulkan/VulkanHandles.h
```diff
@@ -147,13 +151,24 @@ struct VulkanDescriptorSet : public VulkanResource, HwDescriptorSet {
         }
     }
 
+    void setOffsets(backend::DescriptorSetOffsetArray&& offsets) noexcept {
+        mOffsets = std::move(offsets);
+    }
+
+    backend::DescriptorSetOffsetArray* getOffsets() {
```
pixelflinger, line 158: shouldn't this be `const* getOffsets() const`

pixelflinger (approved):

filament/backend/src/vulkan/VulkanHandles.h
poweifeng: good catch. done

PR #8194, opened 2024-10-10
Metal: add better logging for invalid index element size

bejado:
BUGS=[371010314]

### DIFF
--- a/filament/backend/src/metal/MetalDriver.mm
+++ b/filament/backend/src/metal/MetalDriver.mm
@@ -416,12 +416,28 @@ void MetalDriver::createVertexBufferR(Handle<HwVertexBuffer> vbh,
 
 void MetalDriver::createIndexBufferR(Handle<HwIndexBuffer> ibh, ElementType elementType,
         uint32_t indexCount, BufferUsage usage) {
+    // Errors are postponed until the next tick, to give Filament a chance to call setDebugTag on
+    // the handle; this way we get nicer error messages.
+
+    // Check that the elementType is valid.
+    const bool isSupportedElementType =
+            elementType == ElementType::USHORT || elementType == ElementType::UINT;
+    if (UTILS_UNLIKELY(!isSupportedElementType)) {
+        runAtNextTick([this, ibh, isSupportedElementType, elementType]() {
+            FILAMENT_CHECK_PRECONDITION(isSupportedElementType)
+                    << "Invalid elementType (" << static_cast<int>(elementType)
+                    << "). Only ElementType::USHORT and ElementType::UINT supported"
+                    << ", tag=" << mHandleAllocator.getHandleTag(ibh.getId()).c_str_safe();
+        });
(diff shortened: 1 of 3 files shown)


filament/backend/src/metal/MetalDriver.mm
```diff
@@ -416,12 +416,28 @@
...
 void MetalDriver::createIndexBufferR(Handle<HwIndexBuffer> ibh, ElementType elementType,
         uint32_t indexCount, BufferUsage usage) {
+    // Errors are postponed until the next tick, to give Filament a chance to call setDebugTag on
+    // the handle; this way we get nicer error messages.
+
+    // Check that the elementType is valid.
+    const bool isSupportedElementType =
+            elementType == ElementType::USHORT || elementType == ElementType::UINT;
+    if (UTILS_UNLIKELY(!isSupportedElementType)) {
+        runAtNextTick([this, ibh, isSupportedElementType, elementType]() {
```
poweifeng, line 426: nit: `isSupportedElementType` feels redundant since we're sure it's false at this point.

poweifeng (approved):

filament/backend/src/metal/MetalEnums.h
```diff
@@ -71,7 +71,7 @@ constexpr inline MTLIndexType getIndexType(size_t elementSize) noexcept {
     } else if (elementSize == 4) {
         return MTLIndexTypeUInt32;
     }
-    FILAMENT_CHECK_POSTCONDITION(false) << "Index element size not supported.";
+    FILAMENT_CHECK_POSTCONDITION(false) << "Index element size " << elementSize << " not supported.";
```
pixelflinger, line 74: we really shouldn't have a `FILAMENT_CHECK` in a method like this. This type of bugs should be handled on the filament side, and just for safety, we can have an assert_invariant() here (DEBUG only, really just for to catch really serious bugs).

pixelflinger (changes requested):

filament/backend/src/metal/MetalDriver.mm
bejado: Good catch, so my thinking here was to make the Panic's literal more helpful. Otherwise is just has "false" which isn't very descriptive.

filament/backend/src/metal/MetalEnums.h
bejado: Agreed. I'll switch it to a `assert_invariant()`.

filament/backend/src/metal/MetalDriver.mm
pixelflinger: We should catch these errors on the filament side. There are very few places where we  call createIndexBuffer (probably only one). If done on the filament side, it should never ever happen to be wrong on the backend side.

pixelflinger (changes requested):

bejado:
Let me rethink this a bit and try to implement something at the Filament level.

poweifeng (approved):

pixelflinger (approved):

PR #8199, opened 2024-10-12
Add support for compiling under c++20 on clang-based compilers.

emezeske:
The warnings that this change ignores are tricky to fix in the code in a way that is backwards compatible with c++17, but it is possible to do and it would allow these warnings to be reinstated. But for now it seems reasonable to just ignore the warnings if building under c++20, since it compiles and works just fine.

This is useful, for example, if Filament is being used as a CMake submodule inside a project that compiles under c++20.

### DIFF
--- a/filament/CMakeLists.txt
+++ b/filament/CMakeLists.txt
@@ -607,6 +607,13 @@ else()
             -Wover-aligned
             -Werror
     )
+    if (CMAKE_CXX_STANDARD EQUAL 20)
+        # The lambdas for passes in PostProcessManager.cpp capture this
+        # implicitly in a way that's deprecated in c++20, but can't easily be
+        # fixed in a way that's backwards compatible with c++17:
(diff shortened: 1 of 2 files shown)


bejado (approved):
Thanks, we're limited to C++17 at the moment, but this change seems reasonable. My only concern is when we do switch to C+20 we may forget about these warnings, but that's an action item on us.

emezeske:
> Thanks, we're limited to C++17 at the moment, but this change seems reasonable. My only concern is when we do switch to C+20 we may forget about these warnings, but that's an action item on us.

Yep, that's what I figured, on both counts! The good news is that at least on MSVC and AppleClang these warnings were the only thing that hard-prevented compilation, so when the day comes you won't have a ton of work to do. (I think there were a few other warning nits but everything seems to work great.)

pixelflinger (approved):

emezeske:
Thanks, folks!

PR #8201, opened 2024-10-16
no need to copy the PostProcessMaterial mConstants field

pixelflinger:
It always references static data. Additionnaly, we don't need to use a vector to store the specialization constants, because it's also all static data.

And finally, we don't need a boolean to know the state of the PostProcessMaterial, the mSize field can encode the same information.

### DIFF
--- a/filament/src/PostProcessManager.h
+++ b/filament/src/PostProcessManager.h
@@ -21,10 +21,11 @@
 
 #include "FrameHistory.h"
 
-#include "ds/DescriptorSetLayout.h"
 #include "ds/PostProcessDescriptorSet.h"
 #include "ds/SsrPassDescriptorSet.h"
 #include "ds/TypedUniformBuffer.h"
+
+#include <private/filament/Variant.h>
 
 #include <fg/FrameGraphId.h>
 #include <fg/FrameGraphResources.h>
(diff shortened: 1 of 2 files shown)


filament/src/PostProcessManager.h
```diff
@@ -71,16 +68,19 @@ struct CameraInfo;
...
+            std::variant<int32_t, float, bool> value;
+        };
         std::string_view name;
         uint8_t const* data;
-        int size;
-        utils::FixedCapacityVector<ConstantInfo> constants = {};
+        size_t size;
+        // the life-time of objects pointed to by this initializer_list<> is extended to the
+        // life-time of the initializer_list
+        std::initializer_list<ConstantInfo> constants;
```
bejado, line 83: What's the advantage of using `std::initializer_list` here instead of `utils::FixedCapacityVector`?

bejado (approved):

filament/src/PostProcessManager.h
pixelflinger: good question! This reduces significantly (I should say entirely) the code generated to create the vector. Instead, all the data becomes fully static -- no code generated at all. With the vector, it has to loop through all the items in the list and essentially do a `push_back`.

PR #8205, opened 2024-10-16
add missing includes

nim65s:
Hi,

Here are some trivial fixes which seems required to build with clang 18 from nixpkgs.

### DIFF
--- a/libs/utils/include/utils/memalign.h
+++ b/libs/utils/include/utils/memalign.h
@@ -40,7 +40,7 @@ inline void* aligned_alloc(size_t size, size_t align) noexcept {
 #if defined(WIN32)
     p = ::_aligned_malloc(size, align);
 #else
-    ::posix_memalign(&p, align, size);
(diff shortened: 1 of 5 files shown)


libs/utils/include/utils/memalign.h
```diff
@@ -17,6 +17,7 @@
 #ifndef TNT_UTILS_MEMALIGN_H
 #define TNT_UTILS_MEMALIGN_H
 
+#include <tuple>
```
pixelflinger, line 20: why is this one needed?

nim65s: For the `std::ignore`, which itself avoid a warn_unused_result

pixelflinger: okay let's just do `(void)` I don't want to pull a whole header just for that.

nim65s: Ok, done

pixelflinger (approved):

PR #8210, opened 2024-10-17
Add a warning for FRenderableManager

z3moon:
In some cases, users set materials first without providing render primitives, which has incurred the attribute mismatching warning. This isn't helpful because users don't know what action they should take to remove the warning.

Emit the warning only when the primitive handle is initialized so the AttributeBitset is properly populated.

BUGS=[372755205]


pixelflinger (changes requested):
Okay I understand the problem.  I don't think we should have a warning when the handle is undefined -- that's an okay situation.

Instead, I think we should only emit the original "attribute mismatch" warning when the handle is defined. Otherwise we simply can't tell, so let's not print anything.

z3moon:
That also works. Let me update the change.

pixelflinger (approved):

poweifeng (approved):

PR #8214, opened 2024-10-18
Fix a bug in the OpenGL backend that causes win32 errors not to be logged

emezeske:
As you might imagine, I am submitting this change because I'm debugging a bug in the openGL backend, specifically one that causes PlatformWGL::makeCurrent() to fail if the swap chain is removed and created too many times. I hope that I can figure out the issue there, but if someone more familiar with this code could figure it out, that would be really cool... (If anyone is curious, wglMakeCurrent() is returning win32 error code 6.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ repositories {
 }
 
 dependencies {
-    implementation 'com.google.android.filament:filament-android:1.54.5'
+    implementation 'com.google.android.filament:filament-android:1.55.0'
 }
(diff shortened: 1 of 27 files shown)


emezeske:
Oops, this pull request is on a branch where I did some merge commits... let me fix that and reopen.

emezeske:
Fixed PR: https://github.com/google/filament/pull/8216

PR #8218, opened 2024-10-22
Trace

mdagois:
This PR introduces a profiling mode for both GLES and Vulkan backends on Android.
The profiling mode relies on systrace to record very few backend events (right now, _beginFrame_ and _endFrame_ only). The name of the events is kept consistant across backend to allow automatic gathering of data through Perfetto metrics.
The profiling mode is made incompatible with other debug features on purpose, to avoid any errors (e.g. enabling debug features with heavy overhead) when profiling the backend performance.

### DIFF
--- a/filament/backend/src/SystraceProfile.h
+++ b/filament/backend/src/SystraceProfile.h
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2023 The Android Open Source Project
+ * Copyright (C) 2024 The Android Open Source Project
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/filament/backend/src/SystraceProfile.h
+++ b/filament/backend/src/SystraceProfile.h
@@ -1,0 +1,34 @@
+/*
+ * Copyright (C) 2023 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef TNT_FILAMENT_BACKEND_SYSTRACEPROFILE_H
+#define TNT_FILAMENT_BACKEND_SYSTRACEPROFILE_H
+
+#ifdef NDEBUG
+#define ENABLE_BACKEND_PROFILE
+#endif
+
+#if defined(ENABLE_BACKEND_PROFILE)
+#   define PROFILE_SCOPE(marker)   SYSTRACE_NAME(marker)
+#else
+#   define PROFILE_SCOPE(marker)
+#endif
+
+#define PROFILE_BEGINFRAME()    PROFILE_SCOPE("backend::beginFrame")
+#define PROFILE_ENDFRAME()      PROFILE_SCOPE("backend::endFrame")
+
+#endif // TNT_FILAMENT_BACKEND_SYSTRACEPROFILE_H
+
--- a/filament/backend/src/SystraceProfile.h
+++ b/filament/backend/src/SystraceProfile.h
@@ -17,18 +17,12 @@
 #ifndef TNT_FILAMENT_BACKEND_SYSTRACEPROFILE_H
 #define TNT_FILAMENT_BACKEND_SYSTRACEPROFILE_H
 
-#ifdef NDEBUG
-#define ENABLE_BACKEND_PROFILE
-#endif
+#include <utils/Systrace.h>
 
-#if defined(ENABLE_BACKEND_PROFILE)
-#   define PROFILE_SCOPE(marker)   SYSTRACE_NAME(marker)
-#else
-#   define PROFILE_SCOPE(marker)
-#endif
+#define PROFILE_SCOPE(marker)       SYSTRACE_NAME(marker)
 
-#define PROFILE_BEGINFRAME()    PROFILE_SCOPE("backend::beginFrame")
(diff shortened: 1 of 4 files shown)


bejado (approved):

filament/backend/src/opengl/OpenGLDriver.cpp
```diff
@@ -106,7 +106,8 @@
 
 #if DEBUG_MARKER_LEVEL > DEBUG_MARKER_NONE
 #   define DEBUG_MARKER() \
-        DebugMarker _debug_marker(*this, __func__);
+        const std::string _debug_marker_name = std::string("backend::") + __func__; \
+        DebugMarker _debug_marker(*this, _debug_marker_name.c_str());
```
pixelflinger, line 110: I don't what std::string operations (which have heap allocation) when tracing.  We need a solution that only relies on the preprocessor.

pixelflinger (changes requested):

filament/backend/src/opengl/OpenGLDriver.cpp
mdagois: I changed the GL implementation to use a stack-based char array instead of a std::string. Is that acceptable?
If yes, I'll go ahead and change the Vulkan backend too.
If not, I will just hardcore the marker names with litterals.

filament/backend/src/opengl/OpenGLDriver.cpp
```diff
@@ -106,7 +106,13 @@
 
 #if DEBUG_MARKER_LEVEL > DEBUG_MARKER_NONE
 #   define DEBUG_MARKER() \
-        DebugMarker _debug_marker(*this, __func__);
+        const char* _debug_marker_header = "backend::"; \
+        const size_t _debug_marker_header_len = strlen(_debug_marker_header); \
+        char _debug_marker_name[256] = {0}; \
+        constexpr size_t _debug_marker_name_max_len = sizeof(_debug_marker_name); \
+        strncpy(_debug_marker_name, _debug_marker_header, _debug_marker_name_max_len - 1); \
```
poweifeng, line 113: should it be `_debug_marker_header_len` here instead of `_debug_marker_name_max_len - 1`

filament/backend/src/opengl/OpenGLDriver.cpp
```diff
@@ -106,7 +106,13 @@
 
 #if DEBUG_MARKER_LEVEL > DEBUG_MARKER_NONE
 #   define DEBUG_MARKER() \
-        DebugMarker _debug_marker(*this, __func__);
+        const char* _debug_marker_header = "backend::"; \
+        const size_t _debug_marker_header_len = strlen(_debug_marker_header); \
+        char _debug_marker_name[256] = {0}; \
+        constexpr size_t _debug_marker_name_max_len = sizeof(_debug_marker_name); \
```
poweifeng, line 112: should it be `sizeof(_debug_marker_name) - 1` to keep the last char 0?

filament/backend/src/opengl/OpenGLDriver.cpp
```diff
@@ -106,7 +106,13 @@
...
+        strncpy(_debug_marker_name, _debug_marker_header, _debug_marker_name_max_len - 1); \
+        strncpy(_debug_marker_name + _debug_marker_header_len, __func__, _debug_marker_name_max_len - _debug_marker_header_len - 1); \
```
poweifeng, line 114: `_debug_marker_name_max_len - _debug_marker_header_len - 1`
can just be 
`_debug_marker_name_max_len - _debug_marker_header_len` I think

(If the comment on `_debug_marker_name_max_len` is taken).

poweifeng (commented):
typically camelCase is preferred across Filament.

mdagois:
@pixelflinger @poweifeng Please re-check this one. I completely overhauled it after a discussion with @pixelflinger.

filament/backend/src/SystraceProfile.h
```diff
@@ -0,0 +1,28 @@
+/*
+ * Copyright (C) 2023 The Android Open Source Project
```
poweifeng, line 2:
```suggestion
 * Copyright (C) 2024 The Android Open Source Project
```

poweifeng (approved):
vk lgtm

pixelflinger (approved):
LGTM for the GL backend

mdagois:
> LGTM for the GL backend

FYI, I disabled the group markers by default in the GL backend. I also made sure they cannot be used with profiling. This will align the behavior with the Vulkan backend.

pixelflinger:
> > LGTM for the GL backend
> 
> FYI, I disabled the group markers by default in the GL backend. I also made sure they cannot be used with profiling. This will align the behavior with the Vulkan backend.

Oh I missed that!  I actually want them on by default. It's been very useful. I have them only in places that the cost is not overwhelming (like I don't have them in draw).

filament/backend/src/opengl/OpenGLDriver.cpp
```diff
@@ -91,24 +92,35 @@
...
 #define DEBUG_MARKER_NONE             0x00    // no debug marker
 #define DEBUG_MARKER_OPENGL           0x01    // markers in the gl command queue (req. driver support)
 #define DEBUG_MARKER_BACKEND          0x02    // markers on the backend side (systrace)
-#define DEBUG_MARKER_ALL              0x03    // all markers
+#define DEBUG_MARKER_PROFILE          0x04    // profiling on the backend side (systrace)
+#define DEBUG_MARKER_ALL              (0xFF & ~DEBUG_MARKER_PROFILE) // all markers
 
 // set to the desired debug marker level (for user markers [default: All])
-#define DEBUG_GROUP_MARKER_LEVEL      DEBUG_GROUP_MARKER_ALL
+#define DEBUG_GROUP_MARKER_LEVEL      DEBUG_GROUP_MARKER_NONE
```
pixelflinger, line 104: I would like to keep the debug markers enabled by default. These are used by the framegraph and are very useful.

pixelflinger (changes requested):

filament/backend/src/opengl/OpenGLDriver.cpp
mdagois: I put them back in!

PR #8220, opened 2024-10-22
vk: add resource ref-counting alternative

poweifeng:
We add a set of classes to the vk backend that will enable cleaner ref-counting.

In particular, all allocation from the HandleAllocator will be wrapped within a resource_ptr<> smart pointer. This struct will maintain a count that will increment/decrement with respect to references (similar to std::shared_ptr).

This commit only introduces the new classes/structs and does not actually make any functional difference. A follow-up commit will make the switch to use the smart pointer.

We also put VulkanFence and VulkanTimerQuery in a different header because we will need to depend on them separately in the follow-up (reason: they are accessed on the backend thread but allocated on "sync"/filament thread).


filament/backend/src/vulkan/memory/ResourceManager.h
```diff
@@ -0,0 +1,149 @@
...
+        std::unique_lock<utils::Mutex> lock(s->mGcListMutex);
+        s->mGcList.push_back({type, id});
+    }
+
+    void gcImpl();
+    void printImpl() const;
+    void trackIncrement(ResourceType type, HandleId id);
+
+    AllocatorImpl mHandleAllocatorImpl;
+    static ResourceManager* sSingleton;
```
pixelflinger, line 133: is this really what you want? What if a process has to instances of filament? Do you want them to share this? what would the benefit be?

filament/backend/src/vulkan/memory/ResourceCounter.h
```diff
@@ -0,0 +1,337 @@
...
+        return mTsCounter == rhs.mTsCounter && mCounter == rhs.mCounter;
+    }
+
+    inline bool operator!=(RefCounter const& rhs) const {
+        return !(*this == rhs);
+    }
+
+private:
+    TSCounterPtr mTsCounter = nullptr;
+    CounterPtr mCounter = nullptr;
```
pixelflinger, line 291: Why can't the RefCounter be templated? This is weird that would have both pointers? This doubles its size for not reason?

filament/backend/src/vulkan/memory/ResourceCounter.cpp
```diff
@@ -0,0 +1,59 @@
...
+template void CounterPoolImpl<false>::createCounter(CounterIndex freeInd, ResourceType type,
+        HandleId id);
+
+// We define this functions in the cpp because they require knowing the destruction methods from
+// ResourceManager.
+template<bool THREAD_SAFE>
+void CounterPoolImpl<THREAD_SAFE>::createCounter(CounterIndex freeInd, ResourceType type,
+        HandleId id) {
+    auto const impl = [&]() {
+        Counter* ret = &mCounters[freeInd];
```
pixelflinger, line 48: what prevents `freeInd` to be out of bound?

filament/backend/src/vulkan/memory/ResourceCounter.h
```diff
@@ -0,0 +1,337 @@
...
+    // To keep down the size of this class, we only allocate utils::Mutex if necessary.
+    std::unique_ptr<utils::Mutex> mMutex; // 8
+};
+
+static_assert(sizeof(CounterImpl<true>) == 24);
+
+// This encapsulates a std::vector of counters and a list of free counters. The use of a vector (as
+// oppose to dynamically allocating/deallocating counters) is meant to ensure memory locality.
+template<bool THREAD_SAFE>
+struct CounterPoolImpl {
```
pixelflinger, line 121: I think you don't need a CounterPool and can simplify all of that by storing the "count" inside the handle itself.  You can for instance have all your handles derive from something like `RefBase` or `ThreadSafeRefBase`.
Then you make your ref_ptr<> undertand `RefBase` objects.

pixelflinger (changes requested):

filament/backend/src/vulkan/memory/ResourceManager.cpp
```diff
@@ -0,0 +1,111 @@
...
+        case ResourceType::UNDEFINED_TYPE:
+            break;
+    }
+#if FVK_ENABLED(FVK_DEBUG_ALLOCATION)
+    COUNTER[(size_t) type]--;
+#endif
+}
+
+void ResourceManager::trackIncrement(ResourceType type, HandleId id) {
+#if FVK_ENABLED(FVK_DEBUG_ALLOCATION)
```
bejado, line 96: Should we check that `type` is not `ResourceType::UNDEFINED_TYPE` here? Otherwise we'll write past `COUNTER`.

filament/backend/src/vulkan/memory/Resource.h
```diff
@@ -0,0 +1,60 @@
...
+    RENDER_PRIMITIVE = 5,
+    TEXTURE = 6,
+    TEXTURE_STATE = 7,
+    TIMER_QUERY = 8,
+    VERTEX_BUFFER = 9,
+    VERTEX_BUFFER_INFO = 10,
+    DESCRIPTOR_SET_LAYOUT = 11,
+    DESCRIPTOR_SET = 12,
+    FENCE = 13,
+    UNDEFINED_TYPE = 14,
```
bejado, line 49: Maybe comment that `UNDEFINED_TYPE` must be the last enum value.

filament/backend/src/vulkan/memory/ResourceCounter.h
```diff
@@ -0,0 +1,337 @@
...
+private:
+    void destroyNow(ResourceType type, HandleId id);
+    void destroyLater(ResourceType type, HandleId id);
+
+    bool mDestroyed = false;  // 1
+    uint16_t padding = 0;     // 2
+    uint32_t mCount;          // 4
+
+    // To keep down the size of this class, we only allocate utils::Mutex if necessary.
+    std::unique_ptr<utils::Mutex> mMutex; // 8
```
bejado, line 113: Could you do something like:

```
    typename std::conditional_t<THREAD_SAFE, utils::Mutex, std::monostate> mMutex;
```

I _think_ that only adds the `mMutex` member variable if `THREAD_SAFE` is `true`.

poweifeng: this is the c++ magic i was looking for:) (but couldnt' find).

filament/backend/src/vulkan/memory/ResourceManager.h
poweifeng: removed singleton

filament/backend/src/vulkan/memory/ResourceCounter.cpp
poweifeng: no more counter/pool

filament/backend/src/vulkan/memory/ResourceCounter.h
poweifeng: done

filament/backend/src/vulkan/memory/Resource.h
poweifeng: done

filament/backend/src/vulkan/memory/ResourceCounter.h
poweifeng: done

filament/backend/src/vulkan/memory/Resource.h
```diff
@@ -0,0 +1,138 @@
...
+    template <typename T>
+    inline void init(HandleId id, ResourceManager* resManager) {
+        this->id = id;
+        this->resManager = resManager;
+        this->restype = getTypeEnum<T>();
+    }
+
+    void destroy(ResourceType type, HandleId id);
+
+    HandleId id;                 // 8
```
pixelflinger, line 119: `HandleId` is a `uint32_t`

filament/backend/src/vulkan/memory/Resource.h
```diff
@@ -0,0 +1,138 @@
...
+    ResourceManager* resManager; // 8
+    ResourceType restype : 7;
+    bool mDestroyed      : 1;    // 1 - only for THREAD_SAFE
+    uint32_t mCount;             // 4
```
pixelflinger, line 123: could we use a 24-bits `mCount` so that we can reduce the overhead to just 4 bytes?

filament/backend/src/vulkan/memory/Resource.h
```diff
@@ -0,0 +1,138 @@
...
+ResourceType getTypeEnum() noexcept;
+
+std::string getTypeStr(ResourceType type);
+
+inline bool isThreadSafeType(ResourceType type) {
+    return type == ResourceType::FENCE || type == ResourceType::TIMER_QUERY;
+}
+
+template<bool THREAD_SAFE = false>
+struct ResourceImpl {
```
pixelflinger, line 65: is it all backend handles that now will inherit from `ResourceImpl` or only those that need reference counting? (Asking because the size overhead is not trivial).

filament/backend/src/vulkan/memory/Resource.h
pixelflinger: Also I wonder why the object needs its own handle?

filament/backend/src/vulkan/memory/ResourceCounter.h
poweifeng: done

filament/backend/src/vulkan/memory/Resource.h
poweifeng: yes, for simplicity all backend handles will inherit from `ResourceImpl`. With the previous change where I moved bits from VulkanRenderTarget (the largest handle) to the heap, we might be ok with this increase.

filament/backend/src/vulkan/memory/Resource.h
poweifeng: done

filament/backend/src/vulkan/memory/ResourceManager.cpp
poweifeng: done

filament/backend/src/vulkan/memory/Resource.h
pixelflinger: Okay, it's probably okay to keep it where you have it, because due to padding, you'll always have space for it I think.

filament/backend/src/vulkan/memory/Resource.h
pixelflinger: but that means that all handles will increase by at least 16 bytes. I think 8 bytes is the minimum (the cound + some meta data). What is really sad is to have to store a pointer. Would it be too painful to pass it around everywhere it's needed? Or maybe we can think of doing that in another change?

poweifeng: Yeah, I think for the pointer to the manager, we can definitely revisit.

filament/backend/src/vulkan/memory/Resource.h
```diff
@@ -0,0 +1,138 @@
...
+
+    void destroy(ResourceType type, HandleId id);
+
+    HandleId id;                 // 4
+    ResourceManager* resManager; // 8
+    ResourceType restype : 7;
+    bool mDestroyed      : 1;    // only for thread-safe
+    uint32_t mCount      : 24;   // restype + mDestroyed + mCount is 4 bytes.
+
+    typename std::conditional_t<THREAD_SAFE, utils::Mutex, std::monostate> mMutex;
```
pixelflinger, line 125: So I tested this and `std::monostate` will add padding (in this case probably 8 bytes) to all objects. I think we need `ResourceImpl` to be specialized.

pixelflinger (changes requested):

filament/backend/src/vulkan/memory/Resource.h
poweifeng: done

filament/backend/src/vulkan/memory/ResourcePointer.h
```diff
@@ -0,0 +1,196 @@
...
+
+private:
+    // inc() and dec() For tracking ref-count with respect to create/destroy backend APIs. They can
+    // only be used from VulkanDriver.
+    inline void dec() {
+        mRef->dec();
+    }
+
+    inline void inc() {
+        mRef->inc();
```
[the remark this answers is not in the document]
poweifeng: done

filament/backend/src/vulkan/memory/ResourcePointer.h
```diff
@@ -0,0 +1,196 @@
...
+    inline D* get() const {
+        assert_invariant(mRef);
+        return mRef;
+    }
+
+private:
+    // inc() and dec() For tracking ref-count with respect to create/destroy backend APIs. They can
+    // only be used from VulkanDriver.
+    inline void dec() {
+        mRef->dec();
```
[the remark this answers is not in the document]
poweifeng: done

filament/backend/src/vulkan/memory/Resource.h
```diff
@@ -0,0 +1,138 @@
...
+        : id(HandleBase::nullid),
+          resManager(nullptr),
+          mDestroyed(false),
+          mCount(0) {}
+
+private:
+    inline void inc() noexcept {
+        if UTILS_UNLIKELY (mDestroyed) {
+            return;
+        }
```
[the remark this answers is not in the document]
poweifeng: done

filament/backend/src/vulkan/memory/ResourcePointer.h
```diff
@@ -0,0 +1,196 @@
...
+    }
+
+    // move operator
+    inline resource_ptr<D>& operator=(resource_ptr<D> && rhs) {
+        if (mRef && mRef != rhs.mRef) {
+            mRef->dec();
+        }
+        mRef = rhs.mRef; // There should be no change in the reference count.
+        rhs.mRef = nullptr;
+        return *this;
```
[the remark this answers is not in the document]
poweifeng: done

filament/backend/src/vulkan/memory/Resource.h
```diff
@@ -0,0 +1,138 @@
...
+    inline void init(HandleId id, ResourceManager* resManager) {
+        this->id = id;
+        this->resManager = resManager;
+        this->restype = getTypeEnum<T>();
+    }
+
+    void destroy(ResourceType type, HandleId id);
+
+    HandleId id;                 // 4
+    ResourceManager* resManager; // 8
```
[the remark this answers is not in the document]
poweifeng: done

filament/backend/src/vulkan/memory/Resource.h
```diff
@@ -0,0 +1,154 @@
...
+        this->resManager = resManager;
+        this->restype = getTypeEnum<T>();
+    }
+
+    void destroy(ResourceType type, HandleId id);
+
+    ResourceManager* resManager; // 8
+    HandleId id;                 // 4
+    ResourceType restype : 5;
+    bool mDestroyed      : 1;    // only for thread-safe
```
pixelflinger, line 141: `mDestroyed` is not needed

filament/backend/src/vulkan/memory/Resource.h
```diff
@@ -0,0 +1,154 @@
...
+    uint32_t mCount      : 24;   // restype + mDestroyed + mCount is 4 bytes.
```
pixelflinger, line 142: this could be an `std::atomic<uint32_t>`

filament/backend/src/vulkan/memory/Resource.h
```diff
@@ -0,0 +1,154 @@
...
+        : resManager(nullptr),
+          id(HandleBase::nullid),
+          mDestroyed(false),
+          mCount(0) {}
+
+private:
+    inline void inc() noexcept {
+        assert_invariant(!mDestroyed);
+        std::unique_lock<utils::Mutex> lock(mMutex);
+        mCount++;
```
pixelflinger, line 113: replace with `mCount.fetch_add(1, std::memory_order_relaxed)`

filament/backend/src/vulkan/memory/Resource.h
```diff
@@ -0,0 +1,154 @@
...
+    }
+
+    inline void dec() noexcept {
+        assert_invariant(!mDestroyed);
+        bool doDestroy = false;
+        {
+            std::unique_lock<utils::Mutex> lock(mMutex);
+            doDestroy = --mCount == 0;
+        }
+        if (doDestroy) {
```
pixelflinger, line 123: replace with

 ```
 if (mCount.fetch_sub(1, std::memory_order_acq_rel) == 1) {
 ```

pixelflinger (changes requested):

filament/backend/src/vulkan/memory/Resource.h
poweifeng: done

filament/backend/src/vulkan/memory/Resource.h
poweifeng: done

filament/backend/src/vulkan/memory/Resource.h
poweifeng: done

filament/backend/src/vulkan/memory/Resource.h
poweifeng: done

pixelflinger (approved):

PR #8237, opened 2024-10-28
Metal: unbind descriptor set with null handle

bejado:


filament/backend/src/metal/MetalDriver.mm
```diff
@@ -1794,6 +1794,16 @@
         backend::DescriptorSetHandle dsh,
         backend::descriptor_set_t set,
         backend::DescriptorSetOffsetArray&& offsets) {
+
+    if (!dsh) {
```
pixelflinger, line 1798:
```suggestion
    if (UTILS_UNLIKELY(!dsh)) {
```

pixelflinger (changes requested):

pixelflinger (approved):

poweifeng (approved):

PR #8239, opened 2024-10-29
descriptor layouts distinguish samplers and external samplers

pixelflinger:
BUGS=[376089915]


poweifeng (approved):

bejado (approved):
LGTM, but we'll probably need to bump the material version for this?

pixelflinger:
> LGTM, but we'll probably need to bump the material version for this?

yes, because materials will now have a constant that won't be recognized by older versions of filament.

pixelflinger:
Ready for review. I have tested on h/w with camera and added the release notes.

PR #8241, opened 2024-10-30
Vk protected content command buffer

phoenixxxx:
This is the second phase of the protected content work. This includes the protected command buffer, and the associated scheduling and submission work. We also introduce a *getProtected* method in the VulkanCommands object for specific cases where protection level is activated.


filament/backend/src/vulkan/VulkanCommands.cpp
```diff
@@ -33,453 +33,504 @@ using namespace utils;
 
 namespace filament::backend {
 
+    namespace {
```
poweifeng, line 36: The spacing is different for this file.  

the base classes and namespace should have no spacing in front.

filament/backend/src/vulkan/VulkanDriver.h
```diff
@@ -111,6 +111,15 @@ class VulkanDriver final : public DriverBase {
 
 private:
     void collectGarbage();
+    inline VulkanCommandBuffer& getCommandBuffer() {
```
poweifeng, line 114: let's remove this and also `mIsRenderPassProtected` and `mIsContentProtected`

filament/backend/src/vulkan/VulkanHandles.h
```diff
@@ -185,8 +185,9 @@ struct PushConstantDescription {
 
     uint32_t getVkRangeCount() const noexcept { return mRangeCount; }
 
-    void write(VulkanCommands* commands, VkPipelineLayout layout, backend::ShaderStage stage,
-            uint8_t index, backend::PushConstantVariant const& value);
+    void write(VulkanCommands* commands, bool isProtected, VkPipelineLayout layout,
```
poweifeng, line 188: let's not change VulkanHandles in this PR

filament/backend/src/vulkan/platform/VulkanPlatformSwapChainImpl.cpp
```diff
@@ -29,7 +29,7 @@ namespace filament::backend {
 namespace {
 
 std::tuple<VkImage, VkDeviceMemory> createImageAndMemory(VulkanContext const& context,
-        VkDevice device, VkExtent2D extent, VkFormat format) {
+        VkDevice device, VkExtent2D extent, VkFormat format, bool isProtected) {
```
poweifeng, line 32: Can you do this in a separate PR?

PR #8243, opened 2024-10-31
Metal: remove cached pipeline states with deleted programs

bejado:
BUGS=[376480110]

### DIFF
--- a/filament/backend/src/metal/MetalState.h
+++ b/filament/backend/src/metal/MetalState.h
@@ -28,7 +28,9 @@
 
 #include <memory>
 #include <tsl/robin_map.h>
+
 #include <utils/Hash.h>
+#include <utils/Invocable.h>
 
 namespace filament {
 namespace backend {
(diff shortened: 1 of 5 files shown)


filament/backend/src/metal/MetalState.h
```diff
@@ -169,6 +171,15 @@ class StateCache {
 
     void setDevice(id<MTLDevice> device) noexcept { mDevice = device; }
 
+    void removeIf(utils::Invocable<bool(const StateType&)> fn) noexcept {
+        for (auto it = mStateCache.begin(); it != mStateCache.end(); it++) {
+            const auto& [key, _] = *it;
+            if (UTILS_UNLIKELY(fn(key))) {
+                mStateCache.erase(it);
```
poweifeng, line 178: Is this safe?  Iterating through the cache while removing key-value pairs?

bejado: I'm glad you mentioned that, because I double checked and turns out `tsl::robin_map` treats iterators differently than `std::unordered_map` (`robin_map` invalidates _all_ iterators upon `.erase()`). So I need to figure something else out.

pixelflinger: I wonder if you could use `utils/Bimap.h` instead.

```
 * A semi-generic custom bimap. This bimap stores a key/value pair and can retrieve
 * the value from the key and the key from the value.
 * It is optimized for large keys and small values.  The keys are stored out-of-line and are
 * never moved.
```

I don't know if it fits your usecase?

bejado: Not quite. I need to be able to match all keys that fit a given criteria.

poweifeng (approved):

PR #8247, opened 2024-11-02
a new Feature Flag API

pixelflinger:
Feature flags are intended to be used when a new feature is added to  filament and have generally two purposes:

1) during feature development, the feature can be implemented but
   disabled which can help developing large features. This way the
   feature can be tested by stakeholders while it's being developed
   without impacting other clients.

2) once a feature is ready, its flag can be enabled by default, but 
   in case the feature breaks something or has unintended consequence,
   clients have the option to turn it off.

Feature flags are intended to have a relatively short life span, i.e. once a feature is stable, the flag fill be removed.

There two types of feature flags. Constant feature flags can only be set during Engine initialization via Engine::Builder. Non-constant feature flags can be set at any time. 

Feature flags SHOULD NOT be used as configuration or settings.

Feature flags are designed with a few ideas in mind:
- they are very cheap to check inside the engine
- non-constant flags can easily be toggled using ImGUI

### DIFF
--- a/filament/src/details/Engine.cpp
+++ b/filament/src/details/Engine.cpp
@@ -46,6 +46,7 @@
 
 #include <utils/compiler.h>
 #include <utils/debug.h>
+#include <utils/Invocable.h>
 #include <utils/Log.h>
 #include <utils/Panic.h>
 #include <utils/PrivateImplementation-impl.h>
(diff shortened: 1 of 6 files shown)


z3moon (approved):

filament/src/details/Engine.cpp
```diff
@@ -219,7 +230,34 @@ FEngine::FEngine(Engine::Builder const& builder) :
         mMainThreadId(ThreadUtils::getThreadId()),
         mConfig(builder->mConfig)
 {
-    // we're assuming we're on the main thread here.
+    // update a feature flag from Engine::Config if the flag is not specified in the Builder
+    auto const featureFlagsBackwardCompatibility =
+            [this, &builder](std::string_view name, bool value) {
+        if (builder->mFeatureFlags.find(name) != builder->mFeatureFlags.end()) {
```
poweifeng, line 236: do we want `==` here?  (if the flag is *not* specified in the builder).

poweifeng (approved):

PR #8255, opened 2024-11-07
renderdiff: Use single lib for osmesa

poweifeng:
Previously, we linked against libOSMesa through the linker and then used dlopen to link against libGL. This is not how libOSMesa is intended to be used.  Instead, we use dlopen on libOSMesa (via bluegl), which then will map the correct GL methods for us, and for the OSMesa functions, we also map those functions from libOSMesa (instead of relying on compile-time linker).


libs/bluegl/src/BlueGLLinuxOSMesa.cpp
```diff
@@ -0,0 +1,59 @@
...
+namespace {
+using ProcAddressFunc = void*(*)(char const* funcName);
+}
+
+struct Driver {
+    ProcAddressFunc OSMesaGetProcAddress;
+    void* library;
+} g_driver = {nullptr, nullptr};
+
+bool initBinder() {
```
bejado, line 32: Where do these functions get used?

poweifeng: these are the GL api functions. See the parallel for regular linux - [BlueGLLinux][link].

[link]: https://github.com/google/filament/blob/main/libs/bluegl/src/BlueGLLinux.cpp

bejado (approved):

PR #8265, opened 2024-11-10
fix: conditional cstdio include in vk_mem_alloc

aminya:
Based on https://github.com/GPUOpen-LibrariesAndSDKs/VulkanMemoryAllocator/blob/05973d8aeb1a4d12f59aadfb86d20decadba82d1/include/vk_mem_alloc.h#L2842


poweifeng:
Is this meant to reduce the size of the lib?

poweifeng:
> > Is this meant to reduce the size of the lib?
> 
> It's something I noticed when I was comparing the upstream code vs the vendored third party. It's mostly for consistency, but it could benefit the lib size.
> 
> https://github.com/GPUOpen-LibrariesAndSDKs/VulkanMemoryAllocator/blob/05973d8aeb1a4d12f59aadfb86d20decadba82d1/include/vk_mem_alloc.h#L2842

I think the proper way to address this is actually to update our `third_party` vendor lib.

poweifeng (approved):

poweifeng:
let's worry about a proper update later

beschulz:
@poweifeng  @aminya This commit causes trouble with older clang versions. the error is "error: templates must have C++ linkage". This is because the include was changed from ```#include <stdio.h>``` to ```#include <cstdio>``` and the block in question is using c linkage via ```extern "C"```. The [reference code](https://github.com/GPUOpen-LibrariesAndSDKs/VulkanMemoryAllocator/blob/05973d8aeb1a4d12f59aadfb86d20decadba82d1/include/vk_mem_alloc.h#L2798) closes the c linkage before doing the include.

The fix is to wither include stdio.h or to close the c linkage like the reference does.

Would you be so kind to do either of those? it's a "five minute" fix. Otherwise pls let me know which one you prefer and I can create a PR and we can go through the bureaucracy of it all - no problem :)

aminya:
> Would you be so kind to do either of those? it's a "five minute" fix. Otherwise pls let me know which one you prefer and I can create a PR and we can go through the bureaucracy of it all - no problem :)

I'm not aware of the minimum Clang version requirements for Filament. If it's causing issues, it might be better that you create a PR with more details.

beschulz:
> > Would you be so kind to do either of those? it's a "five minute" fix. Otherwise pls let me know which one you prefer and I can create a PR and we can go through the bureaucracy of it all - no problem :)
> 
> I'm not aware of the minimum Clang version requirements for Filament. If it's causing issues, it might be better that you create a PR with more details.

Thanks for your quick reply.

I created the following PR: https://github.com/google/filament/pull/9562 - the change is extremely trivial and simply gets the header file closer to the upstream reference. The explanation of the issue should be easy to follow. I have signed the CLA. Thank you!

PR #8245, opened 2024-11-01
feat: add options to disable tests/benchmarks

aminya:
When building Filament from the source for production from a package manager like vcpkg, these options allow turning off the tests and benchmarks that are not needed for using the library. This significantly reduces compilation time.

Note that I didn't indent the `if`s for a smaller diff to make it easier to review, but I can do that if needed.

This is cherry-picked from my PR for allowing filament to be built via vcpkg:
https://github.com/microsoft/vcpkg/pull/41916

PR #8266, opened 2024-11-10
feat: support optional finding of externals instead of subdirectories

### LINKED ISSUE #5957: filament conflicts with other packages. please allow to use external dependencies
**Describe the bug**
I can't install filament:
```
$ make install
===>  Installing for filament-1.25.5
===>  Checking if filament is already installed
===>  Switching to root credentials for 'install' target
===>   Registering installation for filament-1.25.5
Installing filament-1.25.5...
pkg-static: filament-1.25.5 conflicts with zstd-1.5.2 (installs files into the same place).  Problematic file: /usr/local/lib/libzstd.a
*** Error code 1
```

It conflicts with the zstd-1.5.2 package that is used by a lot of other important packages,

Please allow to use externally installed zstd.

The same situation is with meshoptimizer and potentially with some other packages.

**To Reproduce**
n/a

**Expected behavior**
n/a

**Screenshots**
/a

**Logs**
n/a

aminya:
Fixes #5957

This adds an option to allow usage of external targets that are found via `find_package` and `find_path` instead of adding them via `add_subdirectory`. The implementation will fallback to the directories if a target/header cannot be found.

This is cherry-picked from my PR for allowing filament to be built via vcpkg:
https://github.com/microsoft/vcpkg/pull/41916

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -19,6 +19,8 @@ project(TNT)
 # ==================================================================================================
 # Options
 # ==================================================================================================
+option(FILAMENT_PREFER_EXTERNAL "Prefer the dependencies found externally via CMake" OFF)
+
 option(FILAMENT_USE_EXTERNAL_GLES3 "Experimental: Compile Filament against OpenGL ES 3" OFF)
 
 option(FILAMENT_ENABLE_LTO "Enable link-time optimizations if supported by the compiler" OFF)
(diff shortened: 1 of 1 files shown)


poweifeng:
Closing due to this PR having been idle for a while.  If you'd like to continue, would you mind first opening a feature request issue and we can have a discussion there to better understand the need? 
(related PRs #8267 #8266 #8245)

aminya:
@poweifeng The PR was ready to be reviewed. There are multiple issues requesting this already. Could you reopen it? I'll make sure all the CI failures are addressed.

patrikhuber:
Big thumbs-up from me for this PR - it looks like the best solution allowing to keep the modules in add_subdirectory for the people who want and use filament as a submodule/subdirectory, but will also allow a much smoother, better CMake experience for everyone else using find_package and/or a package manager.

PR #8267, opened 2024-11-10
feat: support installing and finding filament via find_package

aminya:
This allows installing the targets of Filament and finding it later via `find_package(filament CONFIG)`. This makes usage of Filament very easy from other CMake projects.

For this to work, I had to differentiate the public include paths during build vs being used when filament is installed.

This is cherry-picked from my PR for allowing filament to be built via vcpkg:
https://github.com/microsoft/vcpkg/pull/41916

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -14,7 +14,7 @@ endif()
 # ==================================================================================================
 # Project declaration
 # ==================================================================================================
-project(TNT)
+project(filament VERSION 1.56.0)
 
 # ==================================================================================================
 # Options
@@ -46,6 +46,8 @@ option(FILAMENT_ENABLE_FEATURE_LEVEL_0 "Enable Feature Level 0" ON)
 option(FILAMENT_ENABLE_MULTIVIEW "Enable multiview for Filament" OFF)
 
 option(FILAMENT_SUPPORTS_OSMESA "Enable OSMesa (headless GL context) for Filament" OFF)
+
+option(FILAMENT_INSTALL_RULES "Enable Filament's install rules" ON)
 
 set(FILAMENT_NDK_VERSION "" CACHE STRING
     "Android NDK version or version prefix to be used when building for Android."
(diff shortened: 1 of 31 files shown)


poweifeng:
Closing due to this PR having been idle for a while.  If you'd like to continue, would you mind first opening a feature request issue and we can have a discussion there to better understand the need? 
(related PRs #8267 #8266 #8245)

patrikhuber:
Hi @poweifeng,

The need is pretty clear and simple: Just having a proper filament-config/export.cmake file for the filament library. This is standard nowadays for all proper CMake/C++ libraries, and it makes it so much easier to use the library (filament in this case) in downstream projects. Not having it makes it a real hassle to integrate Filament into projects. Could this please be added?

Thanks a lot.

poweifeng:
> Hi @poweifeng,
> 
> The need is pretty clear and simple: Just having a proper filament-config/export.cmake file for the filament library. This is standard nowadays for all proper CMake/C++ libraries, and it makes it so much easier to use the library (filament in this case) in downstream projects. Not having it makes it a real hassle to integrate Filament into projects. Could this please be added?
> 
> Thanks a lot.

Thanks for the clarification.  This a valid request. I think because it was packaged as part of a set of making vkpkg to work that we weren't sure how to proceed.

patrikhuber:
> > Hi @poweifeng,
> > The need is pretty clear and simple: Just having a proper filament-config/export.cmake file for the filament library. This is standard nowadays for all proper CMake/C++ libraries, and it makes it so much easier to use the library (filament in this case) in downstream projects. Not having it makes it a real hassle to integrate Filament into projects. Could this please be added?
> > Thanks a lot.
> 
> Thanks for the clarification. This a valid request. I think because it was introduced as part of a set of making vkpkg to work that we weren't sure how to proceed.

Thanks a lot for the reply and for being open to the improvements. I can see now how it wasn't "clear and simple", depending on the context one's in. Sorry if it came across as presumptuous.

I've had a closer look at the changes in this PR: I'm not a super-expert on CMake export targets but it looks to me like it's doing the minimum needed and it's doing it well following modern CMake best practices. This will be useful for anyone using Filament with CMake using `find_package(filament CONFIG REQUIRED)` which is sort of the best practice nowadays for most setups/people (I think filament is meant to be used as a submodule but this is not ideal for many setups). As a side-effect, it'll also help integration into vcpkg - but this PR is completely vcpkg-independent.

Btw I don't know the author of this PR, I'm just a "third party" who also had a painful experience integrating Filament into their CMake build in the past, and who would really like to see this improvement land, one way or another. :)

And thank you for all your work on this @aminya!

PR #8273, opened 2024-11-14
vk: Use-after-free coverage

poweifeng:
To allow for use-after-free discovery with ref-counting, we keep a bit to indicate whether the handle is considered destroyed by Filament. We assert against "cast"-ing a destroyed handle.

### DIFF
--- a/filament/backend/src/vulkan/memory/ResourcePointer.h
+++ b/filament/backend/src/vulkan/memory/ResourcePointer.h
@@ -21,6 +21,7 @@
 #include "vulkan/memory/ResourceManager.h"
 
 #include <backend/Handle.h>
+#include <utils/compiler.h>
 
 #include <utility>
 
@@ -59,6 +60,11 @@ public:
     static enabled_resource_ptr<B> cast(ResourceManager* resManager,
             Handle<B> const& handle) noexcept {
         D* ptr = resManager->handle_cast<D*, B>(handle);
+        if (UTILS_UNLIKELY(ptr->isHandleConsideredDestroyed())) {
+            FILAMENT_CHECK_PRECONDITION(!ptr->isHandleConsideredDestroyed())
+                    << "Handle id=" << ptr->id << " (" << getTypeStr(ptr->restype)
+                    << ") is being used after it has been freed";
+        }
         return {ptr};
     }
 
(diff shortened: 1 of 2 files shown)


filament/backend/src/vulkan/memory/ResourcePointer.h
```diff
@@ -59,6 +60,11 @@ struct resource_ptr {
     static enabled_resource_ptr<B> cast(ResourceManager* resManager,
             Handle<B> const& handle) noexcept {
         D* ptr = resManager->handle_cast<D*, B>(handle);
+        if (UTILS_UNLIKELY(ptr->isHandleConsideredDestroyed())) {
+            FILAMENT_CHECK_PRECONDITION(!ptr->isHandleConsideredDestroyed())
+                    << "Handle id=" << ptr->id << " (" << getTypeStr(ptr->restype)
+                    << ") is being used after it has been freed";
+        }
```
pixelflinger, line 67: you don't need the `if` here, `FILAMENT_CHECK_PRECONDITION` already does all the `likely/unlikely` shenanigans

filament/backend/src/vulkan/memory/ResourcePointer.h
```diff
@@ -156,6 +162,7 @@ struct resource_ptr {
     // only be used from VulkanDriver.
     inline void dec() {
         assert_invariant(mRef);
+        mRef->setHandleConsiderDestroyed();
```
pixelflinger, line 165: should we at least `assert_invariant()` here that the handle is not already in the considered destroyed state? (I think this should never happen right?).

pixelflinger (changes requested):

filament/backend/src/vulkan/memory/ResourcePointer.h
poweifeng: done

filament/backend/src/vulkan/memory/ResourcePointer.h
poweifeng: I was trying to avoid hitting the switch in `getTypeStr` unless necessary, hence the roundabout logic here.
Do you think it's worth it or not? (or maybe it just going to be speculatively executed anyways?).

poweifeng:
Accidentally closed.

filament/backend/src/vulkan/memory/ResourcePointer.h
pixelflinger: Actually, `FILAMENT_CHECK_PRECONDITION` is pretty sick. If the condition is false, it *will not* execute at all what comes after the `<<`. I know it's magic.

poweifeng: done

pixelflinger (approved):

PR #8275, opened 2024-11-14
material: fix depth variant leak

poweifeng:
The default material does not cover all of the depth variants, and so for the client material's depth variants (with no custom depth shader), we need to check if the program is allocated for the material or if it is actually part of the default material.


poweifeng:
Repro step: 
 - build with `./build.sh -b debug gltf_viewer`
 - Run gltf_viewer with `./out/cmake-debug/samples/gltf_viewer -a opengl --batch=test.json -e third_party/models/lucy/lucy.glb --headless`

Download 
[test.json](https://github.com/user-attachments/files/17759054/test.json)

pixelflinger (changes requested):
I think this will break with a destroy-after-free in the case where the user didn't destroy materials and the Engine then does the cleanup in shutdown(), because the default material is destroyed first (it has to, otherwise it will be detected as leaked).

PR #8276, opened 2024-11-16
fix a HwProgram leak related to depth variant pre-caching

pixelflinger:
Materials that don't use a custom depth can share the same program for all depth variants. We take advantage of that by pre-caching the depths variants when a program is loaded. The depth variants themselves are owned by the default material which is created first. When a material is destroyed it skips deletion of those variants which will be truly destroyed when the default material is destroyed.

Unfortunately, the default material doesn't contain all depth variants, in particular, it doesn't have the VSM ones (because it's an unlit material).  On top of that, the way it pre-cached depth variant had a bug that made it miss some picking variants.

Because of that, these variants were skipped during material destruction but never actually destroyed, since the default material didn't know anything about them.

This PR fixes this by making the pre-caching completely lazy. When any material needs such a "precacheable" variant, it first checks if the default material has it. If not, it creates the variant, caches it locally and forwards it to the default material, which implicitly becomes the owner. Just like before newly created material precache everything the default material already has.

With this mechanism, it's impossible to have a depth variant without it also existing in the default material.

This also simplify the non-public invalidate() method, which doesn't need to populate the pre-cached programs, since this will happen naturally when needed.

There was also another issue where post-process materials were handled like regular materials. This didn't cause an problem but was inefficient since these only have a handful of variants.

### DIFF
--- a/filament/src/details/Material.cpp
+++ b/filament/src/details/Material.cpp
@@ -765,7 +765,7 @@ void FMaterial::destroyPrograms(FEngine& engine,
                 // we should not free them now.
 
                 // During Engine::shutdown(), auto-cleanup destroys the default material first,
-                // so this can be null, but yhis is only used for debugging.
+                // so this can be null, but this is only used for debugging.
                 UTILS_UNUSED_IN_RELEASE
                 auto UTILS_NULLABLE pDefaultMaterial = engine.getDefaultMaterial();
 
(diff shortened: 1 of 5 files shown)


filament/src/details/Material.cpp
```diff
@@ -734,22 +741,76 @@ void FMaterial::onQueryCallback(void* userdata, VariantList* pVariants) {
...
+                            driverApi.destroyProgram(std::move(cachedPrograms[k]));
+                        }
+                    }
+                }
+            } else {
+                // The depth variants may be shared with the default material, in which case
+                // we should not free them now.
+
+                // During Engine::shutdown(), auto-cleanup destroys the default material first,
+                // so this can be null, but yhis is only used for debugging.
```
poweifeng, line 768:
```suggestion
                // so this can be null, but this is only used for debugging.
```

PR #8280, opened 2024-11-18
vk: refactor utils into its own namespace

poweifeng:
We move the code that was contained within two "utils" files into separate utils sources in the vulkan/utils directory.

### DIFF
--- a/filament/backend/CMakeLists.txt
+++ b/filament/backend/CMakeLists.txt
@@ -185,8 +185,15 @@ if (FILAMENT_SUPPORTS_VULKAN)
             src/vulkan/platform/VulkanPlatform.cpp
             src/vulkan/platform/VulkanPlatformSwapChainImpl.cpp
             src/vulkan/platform/VulkanPlatformSwapChainImpl.h
-            src/vulkan/spirv/VulkanSpirvUtils.cpp
-            src/vulkan/spirv/VulkanSpirvUtils.h
+            src/vulkan/utils/Conversion.cpp
+            src/vulkan/utils/Conversion.h
+            src/vulkan/utils/Definitions.h
+            src/vulkan/utils/Helper.h
+            src/vulkan/utils/Image.h
+            src/vulkan/utils/Image.cpp
+            src/vulkan/utils/Spirv.h
+            src/vulkan/utils/Spirv.cpp
+            src/vulkan/utils/StaticVector.h
             src/vulkan/VulkanAsyncHandles.cpp
             src/vulkan/VulkanAsyncHandles.h
             src/vulkan/VulkanBlitter.cpp
@@ -205,8 +212,6 @@ if (FILAMENT_SUPPORTS_VULKAN)
             src/vulkan/VulkanFboCache.h
             src/vulkan/VulkanHandles.cpp
             src/vulkan/VulkanHandles.h
-            src/vulkan/VulkanImageUtility.cpp
-            src/vulkan/VulkanImageUtility.h
             src/vulkan/VulkanMemory.h
             src/vulkan/VulkanMemory.cpp
             src/vulkan/VulkanPipelineCache.cpp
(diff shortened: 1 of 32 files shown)


filament/backend/src/vulkan/utils/CappedArray.h
```diff
@@ -0,0 +1,151 @@
...
+    }
+
+private:
+    void swap(CappedArray& rhs) {
+        std::swap(mArray, rhs.mArray);
+        std::swap(mInd, rhs.mInd);
+    }
+
+    FixedSizeArray mArray;
+    uint32_t mInd = 0;
```
pixelflinger, line 146: I would rename `mInd` to `mSize`. Also this is uint32, but capacity is uint16.

filament/backend/src/vulkan/utils/CappedArray.h
```diff
@@ -0,0 +1,151 @@
...
+    }
+
+    inline ~CappedArray() {
+        clear();
+    }
+
+    inline const_iterator begin() const {
+        if (mInd == 0) {
+            return mArray.cend();
+        }
```
pixelflinger, line 60: I'm not understanding this?

filament/backend/src/vulkan/utils/CappedArray.h
```diff
@@ -0,0 +1,151 @@
...
+        return mArray.cbegin();
+    }
+
+    inline const_iterator end() const {
+        if (mInd > 0 && mInd < CAPACITY) {
```
pixelflinger, line 65: `mInd` should never be equal to `CAPACITY`, so that should be an assert.

filament/backend/src/vulkan/utils/CappedArray.h
```diff
@@ -0,0 +1,151 @@
...
+            return mArray.begin() + mInd;
+        }
+        return mArray.cend();
```
pixelflinger, line 68: I don't understand why not always returning `begin()+mInd`

filament/backend/src/vulkan/utils/CappedArray.h
```diff
@@ -0,0 +1,151 @@
...
+
+    inline void insert(T item) {
+        assert_invariant(mInd < CAPACITY);
+        mArray[mInd++] = item;
+    }
+
+    inline void erase(T item) {
+        PANIC_PRECONDITION("CappedArray::erase should not be called");
+    }
+
```
pixelflinger, line 107: just remove the function then, it will catch the problem at compile time.

filament/backend/src/vulkan/utils/CappedArray.h
```diff
@@ -0,0 +1,151 @@
...
+    inline void clear() {
+        if (mInd == 0) {
+            return;
+        }
```
pixelflinger, line 111: not needed

filament/backend/src/vulkan/utils/CappedArray.h
```diff
@@ -0,0 +1,151 @@
...
+        mInd = 0;
+    }
+
+    inline T& operator[](uint16_t ind) {
```
pixelflinger, line 115: use size_t here, rename `ind` to `index` and assert that it's smaller than mInd

pixelflinger (changes requested):

filament/backend/src/vulkan/utils/CappedArray.h
poweifeng: done

filament/backend/src/vulkan/utils/CappedArray.h
poweifeng: this is some hack leftover from before when I was using this in ref counting.

filament/backend/src/vulkan/utils/CappedArray.h
poweifeng: Added an assert.

filament/backend/src/vulkan/utils/CappedArray.h
poweifeng: leftover hack.  removed.

filament/backend/src/vulkan/utils/CappedArray.h
poweifeng: removed function.

filament/backend/src/vulkan/utils/CappedArray.h
poweifeng: removed.

filament/backend/src/vulkan/utils/CappedArray.h
poweifeng: done

bejado (approved):

pixelflinger (approved):

PR #8287, opened 2024-11-21
Bugfix to allow animations to use a keyframe at exactly time=duration

emezeske:
Without this change, trying to animate to the exact last keyframe wraps back to the first, which only works for animations that loop.


poweifeng (approved):

pixelflinger (approved):

emezeske:
Thanks, y'all!

Issue #8166, opened 2024-09-29
If the width of the image is not a multiple of 4, the rendering result by image_view will be skewed.


seker:
⚠️ **Issues not using this template will be systematically closed.**

**Describe the bug**
If the width of the image is not a multiple of 4, the rendering result by image_view will be skewed.

**To Reproduce**
1. /home/ros/git/github/filament/cmake-build-debug/samples/image_viewer -a vulkan /home/ros/git/momenta/avm/resource/avm_render/texture/rvc_top_view.png

**Expected behavior**
the rendering result by image_view will be skewed.

**Screenshots**
![截图 2024-09-29 14-22-04](https://github.com/user-attachments/assets/b2bf3c0e-f796-48ae-9154-45187efa2f14)

**Logs**
/home/ros/git/github/filament/cmake-build-debug/samples/image_viewer -a vulkan /home/ros/git/momenta/avm/resource/avm_render/texture/rvc_top_view.png
FEngine (64 bits) created at 0x6353dcd3bbe0 (threading is enabled)
Vulkan device driver: NVIDIA 550.107.02
Selected physical device 'NVIDIA GeForce MX450' from 3 physical devices. (vendor 0x10de, device 0x1f9c, driver 0x899ac080, api 1.3)
Backend feature level: 3
FEngine feature level: 3
vkCreateSwapchain: 1024x640, VK_FORMAT_B8G8R8A8_UNORM, VK_COLOR_SPACE_SRGB_NONLINEAR_KHR, swapchain-size=3, identity-transform=true, depth=VK_FORMAT_D32_SFLOAT_S8_UINT

**Desktop (please complete the following information):**
 - OS: Ubuntu 22.04.5 LTS
 - GPU:NVIDIA Corporation TU117M [GeForce MX450]
 - Backend: Vulkan

**Smartphone (please complete the following information):**
 - Device: -
 - OS:-

**Additional context**
![rvc_top_view](https://github.com/user-attachments/assets/ed25e708-3ef1-428d-bec7-7c2efe3f73c7)

PR #8312, opened 2024-12-13
Fix If the width of the image is not a multiple of 4, the rendering r…

seker:
Fix If the width of the image is not a multiple of 4, the rendering result by image_view will be skewed. (#8166)

### DIFF
--- a/samples/image_viewer.cpp
+++ b/samples/image_viewer.cpp
@@ -246,7 +246,7 @@ static void loadImage(App& app, Engine* engine, const Path& filename) {
             .height(h)
             .levels(0xff)
             .format(channels == 3 ?
-                    Texture::InternalFormat::RGB16F : Texture::InternalFormat::RGBA16F)
+                    Texture::InternalFormat::RGB16F : Texture::InternalFormat::RGBA32F)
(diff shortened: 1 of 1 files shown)


samples/image_viewer.cpp
```diff
@@ -246,7 +246,7 @@ static void loadImage(App& app, Engine* engine, const Path& filename) {
             .height(h)
             .levels(0xff)
             .format(channels == 3 ?
-                    Texture::InternalFormat::RGB16F : Texture::InternalFormat::RGBA16F)
+                    Texture::InternalFormat::RGB16F : Texture::InternalFormat::RGBA32F)
```
pixelflinger, line 249: I don't understand this change. It seems like this might be an issue with a particular backend? Why is RGBA16F not working?

romainguy: Definitely seems like a backend issue. Actually I would expect RGB16F to be the issue, not RGBA16F.

poweifeng:
let's hold off on this one. This is due to a bug in the vk backend, but said bug is not simple to address.

poweifeng:
Closing this for a proper fix of #8166

PR #8315, opened 2024-12-13
Add a way to set the shadow sampling quality

pixelflinger:
FIXES=[379754325]


libs/filamat/src/shaders/CodeGenerator.cpp
```diff
@@ -337,6 +337,9 @@ utils::io::sstream& CodeGenerator::generateProlog(utils::io::sstream& out, Shade
     generateSpecializationConstant(out, "CONFIG_SH_BANDS_COUNT",
             +ReservedSpecializationConstants::CONFIG_SH_BANDS_COUNT, 3);
 
+    generateSpecializationConstant(out, "CONFIG_SHADOW_SAMPLING_METHOD",
+            +ReservedSpecializationConstants::CONFIG_SHADOW_SAMPLING_METHOD, 1);
```
poweifeng, line 341: nit: would add a comment that 1 is `PCF_LOW`

poweifeng (approved):

PR #8317, opened 2024-12-17
Fix compilation material failure log

poweifeng:


bejado (approved):

pixelflinger (approved):
LGTM. I've been using the `+value` syntax instead of `(int)value`; it's the same, but a bit more compact.

poweifeng:
> LGTM. I've been using the `+value` syntax instead of `(int)value`; it's the same, but a bit more compact.

updated

PR #8322, opened 2024-12-18
matdbg: Fix material collision

poweifeng:
Fix 3 bugs wrt matdbg:

1. The numbering scheme for materials with same name was buggy
2. The shared depth variants of the default material should also show up as active material variants.
3. When two instantiation of FMaterial point to the exact same bits, we still need to treat them as two materials. (Otherwise we wouldn't be able to modify one of them in matdbg).

### DIFF
--- a/libs/matdbg/src/DebugServer.cpp
+++ b/libs/matdbg/src/DebugServer.cpp
@@ -193,8 +193,17 @@ DebugServer::addMaterial(const CString& name, const void* data, size_t size, voi
         return {};
     }
 
-    const uint32_t seed = 42;
-    const MaterialKey key = utils::hash::murmurSlow((const uint8_t*) data, size, seed);
+    // Note that it's possible to have two materials with the exact same content (however wasteful),
+    // but they refer to different instantiation of FMaterial. For that case, we need to make sure
+    // that the key for the hash is different by relying on the user provided pointer. Hence we run
+    // the hash function twice.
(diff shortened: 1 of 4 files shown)


libs/matdbg/src/DebugServer.cpp
```diff
@@ -193,8 +193,17 @@ DebugServer::addMaterial(const CString& name, const void* data, size_t size, voi
...
-    const uint32_t seed = 42;
-    const MaterialKey key = utils::hash::murmurSlow((const uint8_t*) data, size, seed);
+    // Note that it's possible to have two materials with the exact same content (however wasteful),
+    // but they refer to different instantiation of FMaterial. For that case, we need to make sure
+    // that the key for the hash is different by relying on the user provided pointer. Hence we run
+    // the hash function twice.
+    constexpr uint32_t seed = 42;
+    uint64_t appendedData[2] = {
+        utils::hash::murmurSlow((const uint8_t*) data, size, seed),
+        (uint64_t) userdata,
```
bejado, line 203: Would it be sufficient to only hash the userdata pointer here?

bejado (approved):

libs/matdbg/src/DebugServer.cpp
poweifeng: good point, but it seems `murmur` hashes to a fix value for <= 8 bytes. I used a 16 bytes value instead.

PR #8323, opened 2024-12-19
external textures are no longer restricted to SAMPLER_EXTERNAL

pixelflinger:


pixelflinger:
Converting this PR to draft because a few things are missing:
- the java/js bindings
- the gl backend needs to use the `imported` property (instead of the target) to force a rebind, in order to latch content changes

poweifeng (approved):

PR #8335, opened 2025-01-03
develop DdsProvider, support 'MSFT_texture_dds' extension.

seker:
develop DdsProvider, support 'MSFT_texture_dds' extension.


seker:
@poweifeng

poweifeng:
@seker , after discussion with the team, we feel that it's probably best to keep this code at the client level, where you'd implement the dds provider in your code and give it to gltfio. 

The reasoning is that Filament is a fairly small team, and we cannot afford to take on additional maintenance at will (like adding dep like gli).  Additionally, there is already a mechanism for client's to plug in their image format encoder/decoder in gltfio.  So weighing the trade-offs here we feel this work can be deferred to the client. 

We appreciate your contribution, and please continue to do so.  Thank you!

PR #8338, opened 2025-01-06
matdbg: enable source formatting

poweifeng:
We use clang-format to format the shader files. Since we assume clang is necessary for compilation, installing clang-format should not be too much of an overhead in terms of adding dependency.

Currently, we only support posix systems (macOS, linux).

### DIFF
--- a/libs/matdbg/src/SourceFormatter.cpp
+++ b/libs/matdbg/src/SourceFormatter.cpp
@@ -1,0 +1,80 @@
+/*
+ * Copyright (C) 2025 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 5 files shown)


libs/matdbg/src/SourceFormatter.cpp
```diff
@@ -0,0 +1,80 @@
...
+    pclose(fp);
+
+    std::string const CLANG_FORMAT_OPTIONS =
+            "-style='{"
+            "BasedOnStyle: Google, "
+            "IndentWidth: 4, "
+            "MaxEmptyLinesToKeep: 2"
+            "}'";
+
+    fp = popen(("clang-format " + CLANG_FORMAT_OPTIONS + "< " + TMP_FILENAME).c_str(), "r");
```
z3moon, line 47: nit: if you use a pipe, I think you don't need to create a temporary file so a bit more efficient.

e.g., echo "int main() {return 0;}" | clang-format -style=Google

but this already looks good, so I'll leave it up to your decision.

z3moon (approved):
Great change, thanks!

pixelflinger (approved):

PR #8366, opened 2025-01-17
validate MaterialInstance references when destroyed

pixelflinger:
With this change we now enforce two things:
- All MaterialInstance of a Material must be destroyed when destroying said Material. This has always been a documented  requirement of the public API, but wasn't enforced (only a warning was printed). This new assertion is unconditional.

- A MaterialInstance, when destroyed is no longer in use by any Renderable.

So before destroying a MaterialInstance, the user of API needs to  ensure that either all Renderable using that MaterialInstance in one of their Render Primitives are  destroyed, or, that these Renderable using that MaterialInstance are reset to another one or to null.

There is a new RenderableManager::clearMaterialInstanceAt() that can be used to clear a MaterialInstance on a Render Primitive.
  
Additionally, a Render Primitive with a null MaterialInstance is now silently skipped during rendering, instead of a null-dereference.

Finally, that second assert is protected by a new feature flag:  "features.engine.debug.assert_material_instance_in_use". This flag is enabled on DEBUG builds and disabled on RELEASE builds by default. The flag can be changed at any time using `Engine::setFeatureFlag()`.

BUGS=[333907416]

### DIFF
--- a/filament/src/components/RenderableManager.cpp
+++ b/filament/src/components/RenderableManager.cpp
@@ -820,7 +820,7 @@ void FRenderableManager::setMaterialInstanceAt(Instance instance, uint8_t level,
 }
 
 void FRenderableManager::clearMaterialInstanceAt(Instance instance, uint8_t level,
-    size_t primitiveIndex) {
+        size_t primitiveIndex) {
     if (instance) {
         Slice<FRenderPrimitive>& primitives = getRenderPrimitives(instance, level);
         if (primitiveIndex < primitives.size()) {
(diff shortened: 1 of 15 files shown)


filament/src/components/RenderableManager.cpp
```diff
@@ -819,6 +819,16 @@ void FRenderableManager::setMaterialInstanceAt(Instance instance, uint8_t level,
     }
 }
 
+void FRenderableManager::clearMaterialInstanceAt(Instance instance, uint8_t level,
+    size_t primitiveIndex) {
```
poweifeng, line 823:
```suggestion
        size_t primitiveIndex) {
```

poweifeng (approved):

PR #8382, opened 2025-01-23
add a version of Engine::flushAndWait that takes a timeout

pixelflinger:
The old version that doesn't take a timeout now always waits forever; before it waited for a few seconds on Android debug builds.

The new version has an explicit timeout that works on all platforms and returns success status.

FIXES=[384043020]

### DIFF
--- a/filament/src/details/Engine.cpp
+++ b/filament/src/details/Engine.cpp
@@ -665,50 +665,28 @@ void FEngine::flush() {
 }
 
 void FEngine::flushAndWait() {
-    FILAMENT_CHECK_PRECONDITION(!mCommandBufferQueue.isPaused())
-            << "Cannot call flushAndWait() when rendering thread is paused!";
+    flushAndWait(FENCE_WAIT_FOR_EVER);
+}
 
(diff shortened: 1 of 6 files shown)


filament/src/details/Engine.cpp
```diff
@@ -665,50 +665,28 @@ void FEngine::flush() {
...
-#else
-
-    FFence::waitAndDestroy(createFence(), FFence::Mode::FLUSH);
-
-#endif
-
     // finally, execute callbacks that might have been scheduled
     getDriver().purge();
+
+    return status == FenceStatus::CONDITION_SATISFIED;
```
poweifeng, line 689: I think there's also `FenceStatus::ERROR`? Here it'll get translated to timeout.

poweifeng (approved):
lgtm except one question

filament/src/details/Engine.cpp
pixelflinger: yeah, the return value is more success/fail -- we'll see how it goes, I didn't want to expose the 3-states.

PR #8387, opened 2025-01-24
Combine and install Abseil

bejado:
This will bundle all of Abseil into a single `libabseil.a` library which will get installed alongside Filament libraries for clients to link against.


pixelflinger (approved):

bejado:
FYI this is on hold while I fix the Windows build.

emezeske:
This PR appears to break projects that depend on Abseil themselves, see #8611 -- do you have any guidance on how to handle this breaking change?

PR #8395, opened 2025-01-28
Improve external texture image API

pixelflinger:
The current API used a void* in Texture::setExternalImage to pass the platform specific image. This was a problem because on some  platform like Android, there could be several types (EGLImageKHR and  AHardwareBuffer), and with the void* there was no way to distinguish them on the backend / platform side. 

We now have a new Texture API which takes a 
Platform::ExternalImageHandle, which is a reference-counted handle to a Platform::ExternalImage. Platform::ExternalImage is an opaque type only known by the platform.

This type must be created using a concrete Platform specific API, such as:

PlatformEGL::createExternalImage
PlatformEGLAndroid::createExternalImage
PlatformCocoaGL::createExternalImage
PlatformCocoaTouchGL::createExternalImage

New types can easily be added in the future.

There is no "destroy" call because the Platform::ExternalImageHandle is reference-counted and the client can let go of it as soon as it doesn't need it any longer (typically after calling  Texture::setExternalImage). Platform::ExternalImage is a small structure that just anonymously wraps the concrete types.

New internal driver APIs have been added to handle this functionality.

### DIFF
--- a/filament/backend/src/Platform.cpp
+++ b/filament/backend/src/Platform.cpp
@@ -16,7 +16,92 @@
 
 #include <backend/Platform.h>
 
+#include <utils/compiler.h>
+
+#include <atomic>
+#include <utility>
+
+#include <stddef.h>
+#include <stdint.h>
+
 namespace filament::backend {
+
+void Platform::ExternalImageHandle::incref(ExternalImage* p) noexcept {
+    if (p) {
+        // incrementing the ref-count doesn't acquire or release anything
+        p->mRefCount.fetch_add(1, std::memory_order_relaxed);
+    }
+}
+
+void Platform::ExternalImageHandle::decref(ExternalImage* p) noexcept {
+    if (p) {
+        // When decrementing the ref-count, unless it reaches zero, there is no need to acquire data; we need to
+        // release all previous writes though so they can be visible to the thread that will actually delete the
+        // object.
+        if (p->mRefCount.fetch_sub(1, std::memory_order_release) == 1) {
+            // if we reach zero, we're about to delete the object, we need to acquire all previous writes from other
+            // threads (i.e.: the memory from other threads prior to the decref() need to be visible now.
+            std::atomic_thread_fence(std::memory_order_acquire);
+            delete p;
+        }
+    }
+}
+
+Platform::ExternalImageHandle::ExternalImageHandle() noexcept = default;
+
+Platform::ExternalImageHandle::~ExternalImageHandle() noexcept {
+    decref(mTarget);
+}
+
+Platform::ExternalImageHandle::ExternalImageHandle(ExternalImage* p) noexcept
+        : mTarget(p) {
+    incref(mTarget);
+}
+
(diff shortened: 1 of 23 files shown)


filament/backend/src/Platform.cpp
```diff
@@ -16,8 +16,80 @@
...
+        decref(mTarget);
+        mTarget = rhs.mTarget;
+    }
+    return *this;
+}
+
+Platform::ExternalImageHandle& Platform::ExternalImageHandle::operator=(ExternalImageHandle&& rhs) noexcept {
+    auto* const temp = mTarget;
+    mTarget = rhs.mTarget;
+    rhs.mTarget = temp;
```
z3moon, line 72: this looks like a swap logic, maybe you meant something like
```
if (this != &rhs) {
    decref(mTarget);
    mTarget = rhs.mTarget;
    rhs.mTarget = nullptr;
}
```

pixelflinger: The swap logic should work because `rhs` is a temporary, right? So it's dtor is going to be called and the ref count dropped; and then we can omit the if.  wdyt?

z3moon: Well, the end result may be the same in most cases, but the process is different than typical algorithms and it can lead to unexpected behaviors. E.g.,
```
void Func() {
    ExternalImageHandle a = { .. inited with value A .. };
    ExternalImageHandle b = { .. inited with value B .. };
    a = std::move(b);
    // at this point, a's holding B and b's holding A. Instead a should hold B and b should hold null even though both are going to be destroyed at the end of this function scope.
}
```
also if `a` and `b` are both member variables of a class, which I think is very less likely but still possible, the values will remained swapped until explicitly destroyed.

filament/include/filament/Texture.h
```diff
@@ -406,6 +409,26 @@ class UTILS_PUBLIC Texture : public FilamentAPI {
...
+     * @param engine        Engine this texture is associated to.
+     * @param image         An opaque handle to a platform specific image. It must be created using Platform
+     *                      specific APIs. For example PlatformEGL::createExternalImage(EGLImageKHR eglImage)
+     *
+     * @see PlatformEGL::createExternalImage
+     * @see PlatformEGLAndroid::createExternalImage
+     * @see PlatformCocoaGL::createExternalImage
+     * @see PlatformCocoaTouchGL::createExternalImage
+     */
+    void setExternalImage(Engine& engine, ExternalImageHandleRef image) noexcept;
```
bejado, line 430: Is the idea that clients call the platform-specific `createExternalImage` and then pass that ref here?

bejado (approved):

filament/backend/src/Platform.cpp
pixelflinger: okay I buy it :-)

filament/include/filament/Texture.h
pixelflinger: yes

z3moon (approved):

PR #8397, opened 2025-01-29
Metal: add config to abandon frame if drawable cannot be acquired

bejado:
Give clients the option to abandon a frame if we cannot obtain a `CAMetalDrawable` from the swap chain.

BUGS = [387888833]

### DIFF
--- a/filament/backend/CMakeLists.txt
+++ b/filament/backend/CMakeLists.txt
@@ -139,11 +139,11 @@ if (FILAMENT_SUPPORTS_METAL)
             src/metal/MetalEnums.mm
             src/metal/MetalExternalImage.mm
             src/metal/MetalHandles.mm
-            src/metal/MetalPlatform.mm
             src/metal/MetalShaderCompiler.mm
             src/metal/MetalState.mm
             src/metal/MetalTimerQuery.mm
             src/metal/MetalUtils.mm
+            src/metal/PlatformMetal.mm
     )
 
     set(METAL_CPP_SRCS
--- a/filament/backend/include/backend/platforms/PlatformMetal.h
+++ b/filament/backend/include/backend/platforms/PlatformMetal.h
@@ -14,8 +14,8 @@
  * limitations under the License.
  */
 
-#ifndef TNT_FILAMENT_BACKEND_PRIVATE_METALPLATFORM_H
-#define TNT_FILAMENT_BACKEND_PRIVATE_METALPLATFORM_H
+#ifndef TNT_FILAMENT_BACKEND_PRIVATE_PLATFORMMETAL_H
+#define TNT_FILAMENT_BACKEND_PRIVATE_PLATFORMMETAL_H
 
 #include <backend/DriverEnums.h>
 #include <backend/Platform.h>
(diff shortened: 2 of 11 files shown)


filament/backend/include/backend/Platform.h
```diff
@@ -70,6 +88,13 @@ class UTILS_PUBLIC Platform {
 
         size_t metalUploadBufferSizeBytes = 512 * 1024;
 
+        /**
+         * The action to take if the Metal backend cannot acquire a Drawable for a frame. Only
+         * affects the Metal backend.
+         */
+        MetalDrawableFailureBehavior metalDrawableFailureBehavior =
+                MetalDrawableFailureBehavior::PANIC;
```
pixelflinger, line 96: there shouldn't be any metal specific stuff in Platform

filament/backend/include/backend/DriverEnums.h
```diff
@@ -1311,6 +1311,7 @@ enum class Workaround : uint16_t {
 };
 
 using StereoscopicType = backend::Platform::StereoscopicType;
+using MetalDrawableFailureBehavior = backend::Platform::MetalDrawableFailureBehavior;
```
pixelflinger, line 1314: should be any metal specific stuff in DriverEnums.h

filament/backend/include/backend/Platform.h
```diff
@@ -61,6 +61,24 @@ class UTILS_PUBLIC Platform {
         MULTIVIEW,
     };
 
+    /**
+     * The action the Metal backend should take if a Drawable cannot be acquired.
+     *
+     * Each frame rendered requires a CAMetalDrawable texture, which is presented on-screen at the
+     * completion of each frame. These are limited and provided round-robin style by the system.
+     */
+    enum class MetalDrawableFailureBehavior : uint8_t {
```
pixelflinger, line 70: this should be in a public `MetalPlatform` header that the concret  metal implementation derives from. just like we did for `OpenGLPlatform`

filament/include/filament/Engine.h
```diff
@@ -380,6 +381,21 @@ class UTILS_PUBLIC Engine {
...
+         * Each frame rendered requires a CAMetalDrawable texture, which is presented on-screen at
+         * the completion of each frame. These are limited and provided round-robin style by the
+         * system.
+         *
+         * - PANIC: Terminates the application and reports an error message (default).
+         * - ABORT_FRAME: Aborts execution of the current frame. The Metal backend will attempt to
+         * acquire a new drawable at the next frame.
+         */
+        MetalDrawableFailureBehavior metalDrawableFailureBehavior =
+                MetalDrawableFailureBehavior::PANIC;
```
pixelflinger, line 397: I would argue that this should be a setter in `PlatformMetal`. The app should be able to cast `Platform*` to `PlatformMetal*` when the backend is `Backend::METAL`. I know we already put some backend specific things in DriverConfig; so maybe we can do that here, but we should clean this up in the future.

pixelflinger (changes requested):

pixelflinger (approved):

emezeske:
Just out of curiosity: do you think this should default to ABORT_FRAME instead of PANIC? Just found that my app was crashing in the wild due to this and was pretty surprised at the default value.

PR #8399, opened 2025-01-29
Flips v coordinates for vulkan usage

GrantComm:
Flips the v coordinates for the hellostereo sample when using vulkan. 

Image before flipping coordinates:
Screenshot 2025-01-17 at 10 16 33 AM

Image after flipping coordinates:
Screenshot 2025-01-29 at 3 27 27 PM

### DIFF
--- a/samples/hellostereo.cpp
+++ b/samples/hellostereo.cpp
@@ -261,12 +261,20 @@ int main(int argc, char** argv) {
         float3 quadNormal = normalize(float3 {0, 0, 1});
         float3 u = normalize(cross(quadNormal, upVector));
         float3 v = cross(quadNormal, u);
-        static Vertex quadVertices[4] = {
-                {{quadCenter - u - v}, {1, 0}},
-                {{quadCenter + u - v}, {0, 0}},
-                {{quadCenter - u + v}, {1, 1}},
-                {{quadCenter + u + v}, {0, 1}}
-        };
+        static Vertex quadVertices[4];
+        
+        // When using Vulkan, the v coordinates need to be flipped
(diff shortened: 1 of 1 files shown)


samples/hellostereo.cpp
```diff
@@ -261,12 +261,20 @@ int main(int argc, char** argv) {
...
+            quadVertices[0] = {{quadCenter - u - v}, {1, 1}};
+            quadVertices[1] = {{quadCenter + u - v}, {0, 1}};
+            quadVertices[2] = {{quadCenter - u + v}, {1, 0}};
+            quadVertices[3] = {{quadCenter + u + v}, {0, 0}};
+        } else {
+            quadVertices[0] = {{quadCenter - u - v}, {1, 0}};
+            quadVertices[1] = {{quadCenter + u - v}, {0, 0}};
+            quadVertices[2] = {{quadCenter - u + v}, {1, 1}};
+            quadVertices[3] = {{quadCenter + u + v}, {0, 1}};
+        }
```
pixelflinger, line 277: I need to understand this better; this kind of shenanigans shouldn't be exposed to the user. The material should handle it.

pixelflinger (changes requested):

samples/hellostereo.cpp
z3moon: Agree, this should not be exposed to users. In other words, users don't have to worry about this. If you run the sample `rendertarget` it works well in Vulkan.

pixelflinger:
@GrantComm could you file a github issue instead? thanks.

GrantComm:
Moved to issue #8420

PR #8400, opened 2025-01-30
Update multiview shader code generator for Vulkan

rafadevai:
When targeting Vulkan with multiview, the shader
code generator was using the OpenGL extension
and built-in variables, which are not supported on Vulkan.

Changed it to use GL_EXT_multiview instead of
GL_OVR_multiview2 when the target API is Vulkan.


z3moon (approved):

pixelflinger:
@rafadevai please sign the CLA so we can commit the change.

pixelflinger (approved):

PR #8403, opened 2025-01-30
matdbg: add support for essl1 shader debugging

### LINKED ISSUE #8261: matdbg crashes when editing featureLevel: 0 materials on a FEATURE_LEVEL_0 device
This is happening because matdbg doesn't understand the Essl1 chunk and we end-up replacing the wrong one, but there are probably other issues; it seems that it's also not selecting the right compilation options.

This can be reproduced with the helloTriangle sample in the emulator by using the `config.forceGLES2Context` option, or using a real ES2 device.

elizagamedev:
Fixes #8261

Note that I haven't tested this on macOS yet.


pixelflinger (approved):

elizagamedev:
Confirmed working fine on macOS so merging!

PR #8409, opened 2025-02-03
java: must use return var in Engine

poweifeng:


bejado (approved):

android/filament-android/src/main/java/com/google/android/filament/Engine.java
```diff
@@ -1271,7 +1271,7 @@ public EntityManager getEntityManager() {
      * {@link  android.view.SurfaceHolder.Callback#surfaceDestroyed surfaceDestroyed}.</p>
      */
     public void flushAndWait() {
-        flushAndWait(Fence.WAIT_FOR_EVER);
+        boolean unused = flushAndWait(Fence.WAIT_FOR_EVER);
```
pixelflinger, line 1274:
```suggestion
        std::ignore = flushAndWait(Fence.WAIT_FOR_EVER);
```

poweifeng: this is java though?

z3moon (approved):

PR #8411, opened 2025-02-04
Add feature flag for material/instance destroy assert

poweifeng:
This will allow a more lenient rollout of the assertion, which otherwise will increase crashes for current clients.

We also disable features.engine.debug.assert_material_instance_in_use by default (it was enabled for debug builds).


pixelflinger (approved):
LGTM, however note that this precondition has always been documented.

PR #8413, opened 2025-02-04
Stop precaching stereo depth variants on non-supported platforms

bejado:
BUGS = [376125497]


poweifeng (approved):
still a mystery why we'd have stereo variants compiled for this particular platform?

bejado:
Filament is getting compiled with support for stereo, which means the default material also has stereo variants.

poweifeng:
> Filament is getting compiled with support for stereo, which means the default material also has stereo variants.

ah yes, interesting...

z3moon (approved):

pixelflinger (approved):

pixelflinger:
FYI, `FMaterial::compile` already did something like that.

PR #8415, opened 2025-02-05
Extension support in Vulkan Context

GrantComm:
Ensures extension support is stored in the VulkanContext when using a shared context. Currently, when a shared context is used the device and instance extensions are never checked, which results in the VulkanContext's mDebugUtilsSupported, mDebugMarkersSupported, and mMultiviewEnabled not being set properly.


filament/backend/src/vulkan/platform/VulkanPlatform.cpp
```diff
@@ -675,13 +675,13 @@ Driver* VulkanPlatform::createDriver(void* sharedContext,
 
     VulkanContext context;
 
-    ExtensionSet instExts;
-    // If using a shared context, we do not assume any extensions.
+    // This constains instance extensions that are required for the platform, which includes
+    // swapchain surface extensions.
+    auto const& swapchainExts = getSwapchainInstanceExtensions();
```
poweifeng, line 680: I think the proper way to do this is actually modify the api so that you'd pass the extensions enabled from the shared context.

GrantComm: Does this update make sense to you? Instead of passing the extension strings directly in the shared context, the booleans representing the supported extensions would be passed through the shared context. These booleans would be set during creation of the instance and device that are also passed in the shared context.

filament/backend/include/backend/platforms/VulkanPlatform.h
```diff
@@ -71,6 +71,9 @@ class VulkanPlatform : public Platform, utils::PrivateImplementation<VulkanPlatf
         // where the gpu only has one graphics queue. Then the client needs to ensure that no
         // concurrent access can occur.
         uint32_t graphicsQueueIndex = 0xFFFFFFFF;
+                bool debugUtilsSupported = false;
```
poweifeng, line 74: spacing here

filament/backend/include/backend/platforms/VulkanPlatform.h
```diff
@@ -71,6 +71,9 @@ class VulkanPlatform : public Platform, utils::PrivateImplementation<VulkanPlatf
...
+        bool debugMarkersSupported = false;
+        bool multiviewSupported = false;
```
poweifeng, line 76: I don't really think this scales well, but to unblock your work, we can do this for now, but note that I might change this in the future.

filament/backend/src/vulkan/platform/VulkanPlatform.cpp
```diff
@@ -620,6 +620,9 @@ struct VulkanPlatformPrivate {
     uint32_t mProtectedGraphicsQueueFamilyIndex = INVALID_VK_INDEX;
     uint32_t mProtectedGraphicsQueueIndex = INVALID_VK_INDEX;
     VkQueue mProtectedGraphicsQueue = VK_NULL_HANDLE;
+    bool mDebugMarkersSupported = false;
+    bool mDebugUtilsSupported = false;
+    bool mMultiviewSupported = false;
```
poweifeng, line 625: I don't think these need to be in the impl.  You can just use the values directly from the shared context and store them in VulkanContext.

filament/backend/src/vulkan/platform/VulkanPlatform.cpp
```diff
@@ -673,12 +675,13 @@ Driver* VulkanPlatform::createDriver(void* sharedContext,
         mImpl->mDevice = scontext->logicalDevice;
         mImpl->mGraphicsQueueFamilyIndex = scontext->graphicsQueueFamilyIndex;
         mImpl->mGraphicsQueueIndex = scontext->graphicsQueueIndex;
+        context.mDebugUtilsSupported = scontext->debugUtilsSupported;
+        context.mDebugMarkersSupported = scontext->debugMarkersSupported;
+        context.mMultiviewEnabled = scontext->multiviewSupported;
```
poweifeng, line 680: Can you move these lines next to the `!mImpl->mSharedContext` case?

GrantComm: Below the declaration of ExtensionSet instExts? I kept it in the if (sharedContext){} block since that's the only place where VulkanSharedContext is

poweifeng: I mean moving this to around 834 so that the two ways of setting these booleans are side-by-side.

GrantComm: Ah ok I understand now

poweifeng (approved):

PR #8417, opened 2025-02-06
Add clearEntities API in Scene.

HanYunChenLuo:
The API works for Scene to clear all entities. Sometimes we need clear all entities from Scene.

### DIFF
--- a/filament/include/filament/Scene.h
+++ b/filament/include/filament/Scene.h
@@ -142,9 +142,9 @@ public:
     void removeEntities(const utils::Entity* UTILS_NONNULL entities, size_t count);
 
     /**
-     * Clear all entities to the Scene.
+     * Remove all entities to the Scene.
      */
-    void clearEntities();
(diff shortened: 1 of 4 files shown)


filament/include/filament/Scene.h
```diff
@@ -141,6 +141,11 @@ class UTILS_PUBLIC Scene : public FilamentAPI {
      */
     void removeEntities(const utils::Entity* UTILS_NONNULL entities, size_t count);
 
+    /**
+     * Clear all entities to the Scene.
+     */
+    void clearEntities();
```
pixelflinger, line 147:
```suggestion
    void removeAllEntities();
```

pixelflinger (changes requested):
rename to `removeAllEntities()` and make it `noexcept`.

HanYunChenLuo:
Patch has been modified

pixelflinger (approved):

PR #8419, opened 2025-02-06
OpenGL: handle glFenceSync returning null

bejado:
BUGS = [394324805]

### DIFF
--- a/filament/backend/src/opengl/OpenGLDriver.cpp
+++ b/filament/backend/src/opengl/OpenGLDriver.cpp
@@ -3489,6 +3489,14 @@ void OpenGLDriver::whenFrameComplete(const std::function<void()>& fn) noexcept {
 
 void OpenGLDriver::whenGpuCommandsComplete(const std::function<void()>& fn) noexcept {
     GLsync sync = glFenceSync(GL_SYNC_GPU_COMMANDS_COMPLETE, 0);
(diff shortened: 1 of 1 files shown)


filament/backend/src/opengl/OpenGLDriver.cpp
```diff
@@ -3489,6 +3489,14 @@ void OpenGLDriver::whenFrameComplete(const std::function<void()>& fn) noexcept {
 
 void OpenGLDriver::whenGpuCommandsComplete(const std::function<void()>& fn) noexcept {
     GLsync sync = glFenceSync(GL_SYNC_GPU_COMMANDS_COMPLETE, 0);
+    // glFenceSync can return 0 in certain situations. For example, we've seen this happen on Web
+    // when the GL context has been lost.
+    // Unfortunately, there's not much we can do here. The fn callback will never execute, which
+    // could cause issues. If we've lost the WebGL context though, Filament is most likely shutting
```
pixelflinger, line 3495: It's very dangerous. e.g. the callback could be releasing a condition variable or memory.
I think I would prefer to do this only on Web (with a log); and assert in the other cases.

pixelflinger (changes requested):

filament/backend/src/opengl/OpenGLDriver.cpp
bejado: Yeah that sounds reasonable.

PR #8425, opened 2025-02-07
Use CAMetalLayer to create MVK surfaces to avoid potential deadlocks.

GavinKG:
On macOS using Vulkan backend w/ MoltenVK, filament currently passes down an NSView instance from main thread to create the surface on render thread (by calling the vkCreateMacOSSurfaceMVK). Older versions of MVK simply warn you on this kind of usage:

> [mvk-info] vkCreateMacOSSurfaceMVK(): You are not calling this function from the main thread. NSView should only be accessed from the main thread. When using this function outside the main thread, consider passing the CAMetalLayer itself in VkMacOSSurfaceCreateInfoMVK::pView, instead of the NSView.

Newer versions of MVK will try to post it back to its "main thread" utilizing dispatch_sync (see `MVKOSExtensions.mm`). But if there are any pending waits or fences e.g. `flushAndWait()`, `Fence::waitAndDestroy(mEngine->createFence())`, a deadlock will be introduced, as seen on the screenshots:

This can be reproduced by adding `engine->flushAndWait();` at the end of the setup functor on any samples like hellopbr:

So i'm thinking if we can just unify the logic to use the underlying CAMetalLayer on both Metal backend and Vulkan with MVK.

### DIFF
--- a/libs/filamentapp/src/FilamentApp.cpp
+++ b/libs/filamentapp/src/FilamentApp.cpp
@@ -704,18 +704,11 @@ FilamentApp::Window::Window(FilamentApp* filamentApp,
         ::prepareNativeWindow(mWindow);
 
         void* metalLayer = nullptr;
-        if (config.backend == filament::Engine::Backend::METAL) {
+        if (config.backend == filament::Engine::Backend::METAL || config.backend == filament::Engine::Backend::VULKAN) {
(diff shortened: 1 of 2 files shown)


poweifeng:
I verified this fix.  Thanks for this!

poweifeng (approved):

PR #8437, opened 2025-02-11
froxel: make froxel buffer size consistent

poweifeng:
The definition of the froxel buffer sizes were spread across multiple places and for different use cases (including allocating the buffer, setting to a descriptor set, and adjusting the relevant spec constant). In this commit, we try to unify the size definitions into Froxelizer::getFroxelBufferByteCount().

We also remove the vulkan-only special case for froxel buffer UBO. This case was introduced because spec constant couldn't be used to size arrays for some drivers in the past, but we introduced a workaround that replaced spec constants with constants.


pixelflinger (approved):

pixelflinger:
Does this fix the lighting issue?

poweifeng:
> Does this fix the lighting issue?

No, that's fixed by [this](https://github.com/google/filament/commit/117978f0925427c4c4554c010c7a275538cc9cc6).  

This is just a clean-up

PR #8438, opened 2025-02-11
Fix TSAN violation in HandleAllocator

bejado:
BUGS = [378375177]


pixelflinger (changes requested):
~I prefer to fix this using mLock instead of making it an atomic.~ Actually I changed my mind. It's fine.

pixelflinger (approved):

PR #8451, opened 2025-02-18
fix build when fgviewer is manually enabled

pixelflinger:


show50726:
We may also want to uncomment the lines around 541?

pixelflinger:
> We may also want to uncomment the lines around 541?

I think that one is not strictly necessary right? That one just makes it the default when DEBUG.

poweifeng:
> > We may also want to uncomment the lines around 541?
> 
> I think that one is not strictly necessary right? That one just makes it the default when DEBUG.

I think it is necessary because one could run `cmake` outside of `build.sh`, in which case, we'd get an undefined option for debug builds (I think).

show50726:
> > > We may also want to uncomment the lines around 541?
> > 
> > 
> > I think that one is not strictly necessary right? That one just makes it the default when DEBUG.
> 
> I think it is necessary because one could run `cmake` outside of `build.sh`, in which case, we'd get an undefined option for debug builds (I think).

I think there would not be error if an option is undefined (however the behavior might not always be as expected).

BTW, I think if we want to make users able to build and use fgviewer, I opened another PR with all necessary changes: #8453

PR #8463, opened 2025-02-20
add a feature flag to control handle tags for heap handles

pixelflinger:
heap allocated handle tags are kept forever and we suspect this is causing problems. the new feature flag "backend.disable_heap_handle_tags" is enabled by default and controls if we store these tags at all or not.

BUGS=[397766275]


filament/backend/include/private/backend/HandleAllocator.h
```diff
@@ -67,7 +68,8 @@ class DebugTag {
 template<size_t P0, size_t P1, size_t P2>
 class HandleAllocator : public DebugTag {
 public:
-    HandleAllocator(const char* name, size_t size, bool disableUseAfterFreeCheck) noexcept;
+    HandleAllocator(const char* name, size_t size,
+            bool disableUseAfterFreeCheck, bool disableHeapHandleTags = false) noexcept;
```
poweifeng, line 72: nit: why the `= false` default?

poweifeng (approved):

filament/backend/include/private/backend/HandleAllocator.h
pixelflinger: I added a default because otherwise there was tons of places that needed to be updated; I think it was in tests. I think `false` was the proper default?

poweifeng: got it, but maybe need a TODO here too since we have now two different defaults and already a bit confusing.

poweifeng (approved):
I think you'll need to undo the workaround that's currently on tot.

PR #8468, opened 2025-02-21
Texture: prevent SIGILL in sampler validator.

KostyaAtG:
Lambda's switch handled all _valid_ enum values without consideration of malicious input.
However, if input is maliciously crafted, then it will lead to undefined behavior (function shall return something, but code does not define what has to be returned if switch body is skipped).

Moved validator close to the beginning of the method to validate before use and added missing return.

### DIFF
--- a/filament/src/details/Texture.cpp
+++ b/filament/src/details/Texture.cpp
@@ -164,6 +164,26 @@ Texture* Texture::Builder::build(Engine& engine) {
             (isProtectedTexturesSupported && useProtectedMemory) || !useProtectedMemory)
             << "Texture is PROTECTED but protected textures are not supported";
 
+    const auto validateSamplerType = [&engine = downcast(engine)](SamplerType const sampler) -> bool {
+        switch (sampler) {
+            case SamplerType::SAMPLER_2D:
+            case SamplerType::SAMPLER_CUBEMAP:
+            case SamplerType::SAMPLER_EXTERNAL:
+                return true;
+            case SamplerType::SAMPLER_3D:
+            case SamplerType::SAMPLER_2D_ARRAY:
+                return engine.hasFeatureLevel(FeatureLevel::FEATURE_LEVEL_1);
+            case SamplerType::SAMPLER_CUBEMAP_ARRAY:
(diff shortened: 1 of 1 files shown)


KostyaAtG:
@pixelflinger 
Quick review identified few other places with similar patterns (switch over scoped enums, no return if switch body is skipped). To name a few:

- DriverEnums.h [1](https://github.com/google/filament/blob/3fae07f23cc1b6bc505173f329890927af34fb59/filament/backend/include/backend/DriverEnums.h#L184-L197), [2](https://github.com/google/filament/blob/3fae07f23cc1b6bc505173f329890927af34fb59/filament/backend/include/backend/DriverEnums.h#L214-L223)
- Camera.h [1](https://github.com/google/filament/blob/3fae07f23cc1b6bc505173f329890927af34fb59/filament/src/details/Camera.h#L138-L147)

Identification of all such places is tedious job, so I would recommend identifying such places automatically and patching them up to prevent undefined behavior in case of malicious input.

Clang18+ can be set up to detect missing "default" branches (`-Wswitch-default`), but even trunk version of clang does not detect absent return value in case of enums (`-Wreturn-type`).
https://godbolt.org/z/TKW9rE155

MSVC does better job and will report a warning `C4715: not all control paths return a value` in case of enums too.
https://godbolt.org/z/WzKGee8a7

pixelflinger:
I don't want to add default to switch/case because then, if we add an enum, we don't get the warning/error in all the places where we might need to handle it.

Ideally we would never have a "default" in any switch/case, but instead always handle all the enum values

KostyaAtG:
I am not rooting for having defaults everywhere, I am rooting to have a function to return something if switch body is skipped at least in user input validation cases.

poweifeng:
> I am not rooting for having defaults everywhere, I am rooting to have a function to return something if switch body is skipped at least in user input validation cases.

Another possibility is to call `FILAMENT_CHECK_POSTCONDITION` outside of the enum.  That way we wouldn't have to add a default (satisfying Mathia's point at compile time). And might be a way for a validation-only mechanism to intercept the assert (though I don't know about the customizability of `FILAMENT_CHECK_POSTCONDITION` today).

pixelflinger (approved):

PR #8470, opened 2025-02-24
android: prevent leak of cubemap for IndirectLight/Skybox

### LINKED ISSUE #6412: Memory leaked with Skybox and IndirectLight on Android
image

As the picture showing, Creating `Skybox` or `IndirectLight` by `KTXLoader.createXXX`  will create `cubemap` Texture.
Normally, we may use `Engine.destroy` to release the `Skybox` or `IndirectLight`, but the `cubemap` Texture will not be destroyed unless the engine is destroyed or destroying the `cubemap` Texture explicitly(it is very oddly because nobody can realize it).

poweifeng:
We break existing  API by returning both the IndirectLight and the texture. Same thing with Skybox.

Fix one bug where the float array in getSphericalHarmonics wasn't getting written out.

Fixes #6412

### DIFF
--- a/NEW_RELEASE_NOTES.md
+++ b/NEW_RELEASE_NOTES.md
@@ -7,3 +7,5 @@ for next branch cut* header.
 appropriate header in [RELEASE_NOTES.md](./RELEASE_NOTES.md).
 
 ## Release notes for next branch cut
+
+- android: breaking changes to API KTX1Loader::createIndirectLight and KTX1Loader::createSkybox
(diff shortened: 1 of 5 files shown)


pixelflinger (commented):
Was this solely a bug on the kotlin side? Or is there a more fundamental bug in the java implementation of IndirectLight?

poweifeng:
> Was this solely a bug on the kotlin side? Or is there a more fundamental bug in the java implementation of IndirectLight?

These are two utils functions (in Ktx1Loader) that exist only for Java/kotlin. It doesn't affect the java impl of IndirectLight.

romainguy (approved):

pixelflinger (approved):

PR #8471, opened 2025-02-26
Compile webgpu library and stub implementation.

sidreesshah:
These changes add Dawn as a webgpu backend dependency to filament and adds stub implementation to link against it.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -195,6 +195,7 @@ VULKAN_ANDROID_OPTION="-DFILAMENT_SUPPORTS_VULKAN=ON"
 VULKAN_ANDROID_GRADLE_OPTION=""
 
 WEBGPU_OPTION="-DFILAMENT_SUPPORTS_WEBGPU=OFF"
+WEBGPU_ANDROID_GRADLE_OPTION=""
 
 EGL_ON_LINUX_OPTION="-DFILAMENT_SUPPORTS_EGL_ON_LINUX=OFF"
 
@@ -530,6 +531,7 @@ function build_android {
             -Pcom.google.android.filament.dist-dir=../out/android-debug/filament \
             -Pcom.google.android.filament.abis=${ABI_GRADLE_OPTION} \
             ${VULKAN_ANDROID_GRADLE_OPTION} \
+            ${WEBGPU_ANDROID_GRADLE_OPTION} \
             ${MATDBG_GRADLE_OPTION} \
             ${MATOPT_GRADLE_OPTION} \
             :filament-android:assembleDebug \
@@ -579,6 +581,7 @@ function build_android {
             -Pcom.google.android.filament.dist-dir=../out/android-release/filament \
             -Pcom.google.android.filament.abis=${ABI_GRADLE_OPTION} \
             ${VULKAN_ANDROID_GRADLE_OPTION} \
+            ${WEBGPU_ANDROID_GRADLE_OPTION} \
             ${MATDBG_GRADLE_OPTION} \
             ${MATOPT_GRADLE_OPTION} \
             :filament-android:assembleRelease \
@@ -959,6 +962,7 @@ while getopts ":hacCfgimp:q:uvWslwedtk:bx:S:X:" opt; do
             ;;
         W)
             WEBGPU_OPTION="-DFILAMENT_SUPPORTS_WEBGPU=ON"
+            WEBGPU_ANDROID_GRADLE_OPTION="-Pcom.google.android.filament.include-webgpu"
             echo "Enable support for WebGPU(Experimental) in the core Filament library."
             ;;
         s)
--- a/build.sh
+++ b/build.sh
@@ -829,7 +829,7 @@ function check_debug_release_build {
 
 pushd "$(dirname "$0")" > /dev/null
 
-while getopts ":hacCfgijmp:q:uvslwedtk:bx:S:X:" opt; do
+while getopts ":hacCfgimp:q:uvslwedtk:bx:S:X:" opt; do
     case ${opt} in
         h)
             print_help
--- a/build.sh
+++ b/build.sh
@@ -44,6 +44,8 @@ function print_help {
     echo "        Run all unit tests, will trigger a debug build if needed."
     echo "    -v"
     echo "        Exclude Vulkan support from the Android build."
+    echo "    -W"
+    echo "        Include WebGPU support for the target platform. (NOT functional atm)."
     echo "    -s"
     echo "        Add iOS simulator support to the iOS build."
     echo "    -e"
@@ -192,6 +194,8 @@ INSTALL_COMMAND=
 VULKAN_ANDROID_OPTION="-DFILAMENT_SUPPORTS_VULKAN=ON"
 VULKAN_ANDROID_GRADLE_OPTION=""
 
+WEBGPU_OPTION="-DFILAMENT_SUPPORTS_WEBGPU=OFF"
+
 EGL_ON_LINUX_OPTION="-DFILAMENT_SUPPORTS_EGL_ON_LINUX=OFF"
 
 MATDBG_OPTION="-DFILAMENT_ENABLE_MATDBG=OFF"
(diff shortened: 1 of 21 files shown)


third_party/dawn/tnt/README.md
```diff
@@ -0,0 +1,26 @@
...
+mv dawn_copy dawn
+patch -p2 < dawn/tnt/001-dawn-static-lib.patch.
+git add dawn
+<may need to add following separately>
+git add add dawn/third_party/dxc/ dawn/third_party/vulkan-loader/src/ dawn/third_party/spirv-tools/src/ dawn/third_party/glslang/
+```
+
+Please be sure to test Filament before uploading your CL.
+Make sure to fix any new issues that are caused by this update. dawn/tnt/CMakeLists.txt may need to be edited for
+an update.
```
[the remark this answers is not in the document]
bridgewaterrobbie: +1, I think it will suffice to mention something along the lines of "For each matched dependency (Known: absiel, spirv-tools, spirv-cross, spirv-headers, glslang) follow the relevant instructions in the dependency `tnt` folder and update it to the version in the new Dawn commit"

filament/backend/CMakeLists.txt
```diff
@@ -237,6 +237,11 @@ if (FILAMENT_SUPPORTS_VULKAN)
     endif()
 endif()
 
+if (FILAMENT_SUPPORTS_WEBGPU)
```
bridgewaterrobbie, line 240: I may be wrong but I would not expect this to be necessary. The webgpu srcs should be compiled as part of the linked library, not as part of the backend srcs.

filament/src/details/Renderer.cpp
```diff
@@ -554,6 +554,8 @@ void FRenderer::renderStandaloneView(FView const* view) {
         if (engine.getBackend() == Backend::VULKAN) {
             engine.flushAndWait();
         }
+
+        // May need to do something similar for DAWN
```
bridgewaterrobbie, line 558: Nit: Add a TODO label so this comment does not get lost.

filament/src/fsr.cpp
```diff
@@ -48,6 +48,7 @@ void FSR_ScalingSetup(FSRUniforms* outUniforms, FSRScalingConfig config) noexcep
 
     auto yoffset = config.input.bottom;
     if (config.backend == backend::Backend::METAL ||
+        // config.backend == backend::Backend::DAWN ||
```
bridgewaterrobbie, line 51: Avoid merging commented code (Not sure if removal or uncommenting is best at this stage, add TODO if appropriate)

libs/filamentapp/src/FilamentApp.cpp
```diff
@@ -662,6 +662,7 @@ FilamentApp::Window::Window(FilamentApp* filamentApp,
         engineConfig.stereoscopicType = Engine::StereoscopicType::NONE;
 #endif
 
+        // Should we add DAWN here or separate path for Dawn?
```
bridgewaterrobbie, line 665: For this and line 890, add `TODO` to comment if we keep it.

libs/matdbg/src/ShaderReplacer.cpp
```diff
@@ -219,6 +219,7 @@ bool ShaderReplacer::replaceSpirv(ShaderModel shaderModel, Variant variant,
     };
 
     MaterialBuilder::TargetApi const targetApi = targetApiFromBackend(mBackend);
+    // Add DAWN
```
bridgewaterrobbie, line 222: Add TODO to comment

filament/backend/src/webgpu/WebGPUDriver.cpp
```diff
@@ -0,0 +1,402 @@
...
+ShaderModel WebGPUDriver::getShaderModel() const noexcept {
+#if defined(__ANDROID__) || defined(IOS)
+    return ShaderModel::MOBILE;
+#else
+    return ShaderModel::DESKTOP;
+#endif
+}
+
+ShaderLanguage WebGPUDriver::getShaderLanguage() const noexcept {
+    return ShaderLanguage::ESSL3;
```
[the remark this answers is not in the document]
sidreesshah: I wanted to keep the changes for introducing WebGPU flag only. I can WGSL to this change and update.

filament/backend/src/webgpu/WebGPUPlatform.h
```diff
@@ -0,0 +1,42 @@
...
+    final : public Platform {
+public:
+
+    int getOSVersion() const noexcept final { return 0; }
+
+    ~WebGPUPlatform() noexcept override = default;
+
+protected:
+
+    Driver* createDriver(void* sharedContext, const Platform::DriverConfig& driverConfig) noexcept override;
```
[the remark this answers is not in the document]
sidreesshah: sharedContext is a way for applications to share webgpu related info with the backend. I am being careful here with webgpu related info as i am not sure whether webgpu does have a webgpu context but webgpu instance could be one of the thing. If sharedContext does not make sense for WebGPU later on we can always disable it with void* /*sharedContext*/ as some other platforms have done.

filament/src/details/Renderer.cpp
sidreesshah: These were to be clean comments, i will put them for a subsequent ticket.

filament/backend/CMakeLists.txt
sidreesshah: These are not webgpu sources but the files we added to backend which will use webgpu api's and be built as part of backend.

build.sh
```diff
@@ -950,6 +958,10 @@ while getopts ":hacCfgimp:q:uvslwedtk:bx:S:X:" opt; do
             echo "Disabling support for Vulkan in the core Filament library."
             echo "Consider using -c after changing this option to clear the Gradle cache."
             ;;
+        W)
+            WEBGPU_OPTION="-DFILAMENT_SUPPORTS_WEBGPU=ON"
+            echo "Enable support for WebGPU(Experimental) in the core Filament library."
```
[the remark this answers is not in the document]
sidreesshah: Reason why i didn't go for stronger or more targeted statement was because i feel that we should keep the statement generic and not go into details whether its functional or not. So it can be experimental for now and then once it ready to announce, we can drop the experimental. Otherwise we will be reviewing and updating this statement after every few patches.

filament/backend/CMakeLists.txt
bridgewaterrobbie: Ah my bad, this makes sense.

filament/backend/src/webgpu/WebGPUDriver.cpp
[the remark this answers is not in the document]
sidreesshah: I didn't push WGSL enum with this patch. There were many TODO's that were required. I guess, we can do it in a separate patch with a bit more cases to cover.

filament/backend/src/webgpu/WebGPUPlatform.h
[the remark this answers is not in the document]
sidreesshah: Generally, i agree but Its related to code style. I would keep it the way, its defined in NoopPlatform.*. We are going to  decide on them in future patches. For now i am leaving it as is.

bejado:
Could we get rid of the `third_party/dawn/test` folder? We don't need to build Dawn tests. Usually we don't care about extraneous files, but this PR is already at around 100k new files.

sidreesshah:
> Could we get rid of the `third_party/dawn/test` folder? We don't need to build Dawn tests. Usually we don't care about extraneous files, but this PR is already at around 100k new files.

Done and cleaned some other folder  too. Pushed dawn in a separate PR to keep this PR light for github.

poweifeng (approved):
this lgtm! But please wait for Ben to approve as well. Thank you!

filament/backend/CMakeLists.txt
```diff
@@ -237,6 +237,11 @@ if (FILAMENT_SUPPORTS_VULKAN)
...
+    file(GLOB_RECURSE WEBGPU_SOURCES RELATIVE ${PROJECT_SOURCE_DIR} "src/webgpu/*.cpp" "src/webgpu/*.h")
```
bejado, line 241: Our convention is to not use `GLOB_RECURSE`, but instead list out files individually.

bejado (approved):
LGTM, one small comment

filament/backend/CMakeLists.txt
sidreesshah: I did that before but after some suggestions used GLOB_RECURSE. Reverted that now.

PR #8472, opened 2025-02-26
viewer: add option to export TIFF in automation

poweifeng:
PPM does not store alpha channel, but TIFF does.  We add a method to export RGBA to a TIFF file without compression.

We add the corresponding options to gltf_viewer and AutomationEngine.


libs/viewer/include/viewer/AutomationEngine.h
```diff
@@ -82,6 +82,11 @@ class UTILS_PUBLIC AutomationEngine {
          * If true, the tick function writes out a settings JSON file before advancing.
          */
         bool exportSettings = false;
+
+        /**
+         * If true, export the screenshot as TIFF (the default is PPM).
+         */
+        bool exportTIFF = false;
```
romainguy, line 89: This should probably be an enum instead

samples/gltf_viewer.cpp
```diff
@@ -232,6 +236,7 @@ static int handleCommandLineArguments(int argc, char* argv[], App* app) {
         { "settings",        required_argument,    nullptr, 't' },
         { "split-view",      no_argument,          nullptr, 'v' },
         { "vulkan-gpu-hint", required_argument,    nullptr, 'g' },
+        { "screenshot-as-tiff", no_argument,       nullptr, 'd' },
```
romainguy, line 239: You should preserve the alignement here, also: any reason to not make TIFF the default? PPM was chosen just because it's easy to output I think, but I doubt there's a good reason to keep both formats?

libs/viewer/include/viewer/AutomationEngine.h
poweifeng: done

pixelflinger (approved):

PR #8477, opened 2025-02-27
Initial tint transpilation support for WebGPU

bridgewaterrobbie:
BUGS =  [398273927]

### DIFF
--- a/libs/filamat/src/GLSLPostProcessor.cpp
+++ b/libs/filamat/src/GLSLPostProcessor.cpp
@@ -519,7 +519,7 @@ void GLSLPostProcessor::spirvToMsl(const SpirvBlob* spirv, std::string* outMsl,
 }
 
 bool GLSLPostProcessor::spirvToWgsl(const SpirvBlob *spirv, std::string *outWsl) {
-#ifdef FILAMENT_SUPPORTS_WEBGPU
+#if FILAMENT_SUPPORTS_WEBGPU
     //Currently no options we want to use
     const tint::spirv::reader::Options readerOpts{};
     tint::wgsl::writer::Options writerOpts{};
@@ -547,7 +547,7 @@ bool GLSLPostProcessor::spirvToWgsl(const SpirvBlob *spirv, std::string *outWsl)
     *outWsl = wgslOut->wgsl;
     return true;
 #else
-    slog.i << "Trying to emit WGSL without including WebGPU dependencies, please set CMake arg FILAMENT_SUPPORTS_WEBGPU" << io::endl;
+    slog.i << "Trying to emit WGSL without including WebGPU dependencies, please set CMake arg FILAMENT_SUPPORTS_WEBGPU and FILAMENT_SUPPORTS_WEBGPU" << io::endl;
     return false;
 #endif
 
--- a/libs/filamat/src/GLSLPostProcessor.cpp
+++ b/libs/filamat/src/GLSLPostProcessor.cpp
@@ -44,6 +44,10 @@
 #include <unordered_map>
 #include <vector>
 
+#ifdef FILAMENT_SUPPORTS_WEBGPU
+#include <tint/tint.h>
+#endif
+
 using namespace glslang;
 using namespace spirv_cross;
 using namespace spvtools;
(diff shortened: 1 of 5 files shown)


libs/filamat/src/GLSLPostProcessor.cpp
```diff
@@ -514,8 +518,43 @@ void GLSLPostProcessor::spirvToMsl(const SpirvBlob* spirv, std::string* outMsl,
...
+        // let it slide until combined image sampler is resolved
+        // return false;
+    }
+
+    tint::Result<tint::wgsl::writer::Output> wgslOut = tint::wgsl::writer::Generate(tintRead,writerOpts);
+    /// An instance of SuccessType that can be used to check a tint Result.
+    tint::SuccessType tintSuccess;
+
+    if (wgslOut != tintSuccess) {
+        slog.i << "This tint writer error is currently ignored during WebGPU bringup:" << wgslOut.Failure().reason.Str() << io::endl;
```
kpiascik, line 542: same here, this should be a warning if not error

libs/filamat/src/GLSLPostProcessor.cpp
```diff
@@ -514,8 +518,43 @@ void GLSLPostProcessor::spirvToMsl(const SpirvBlob* spirv, std::string* outMsl,
...
+bool GLSLPostProcessor::spirvToWgsl(const SpirvBlob *spirv, std::string *outWsl) {
+#ifdef FILAMENT_SUPPORTS_WEBGPU
+    //Currently no options we want to use
+    const tint::spirv::reader::Options readerOpts{};
+    tint::wgsl::writer::Options writerOpts{};
+
+    tint::Program tintRead = tint::spirv::reader::Read(*spirv, readerOpts);
+
+    if (tintRead.Diagnostics().ContainsErrors()) {
+        slog.i << "This tint reader error is currently ignored during WebGPU bringup:" << tintRead.Diagnostics().Str() << io::endl;
```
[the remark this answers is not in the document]
bridgewaterrobbie: Fixed, thanks

libs/filamat/src/GLSLPostProcessor.cpp
bridgewaterrobbie: Fixed, thanks. My thought is it should be a warning right now, but it can become an error once we actually have it working reliably.

libs/filamat/src/GLSLPostProcessor.cpp
```diff
@@ -514,8 +518,43 @@ void GLSLPostProcessor::spirvToMsl(const SpirvBlob* spirv, std::string* outMsl,
...
+    if (wgslOut != tintSuccess) {
+        slog.w << "This tint writer error is currently ignored during WebGPU bringup:" << wgslOut.Failure().reason.Str() << io::endl;
+        //TODO: We should actually return false here, but for initial debugging
+        // let it slide until combined image sampler is resolved
+        // return false;
+    }
+    *outWsl = wgslOut->wgsl;
+    return true;
+#else
+    slog.i << "Trying to emit WGSL without including WebGPU dependencies, please set CMake arg FILAMENT_SUPPORTS_WEBGPU and TINT_BUILD_SPV_READER" << io::endl;
```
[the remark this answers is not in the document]
bridgewaterrobbie: Hrm. A bit debatable, but either way functionally it will fail out due to the `return false` below it. It could also arguably be `slog.e`. 

I'd like to leave it at the moment unless others clarify the common patterns expected in filament.

libs/filamat/src/GLSLPostProcessor.cpp
```diff
@@ -44,6 +44,10 @@
 #include <unordered_map>
 #include <vector>
 
+#if TINT_BUILD_SPV_READER
```
jc3265, line 47: more of a question but
why in some cases we are using `#if defined`/`#ifdef` and here just `#if`

bridgewaterrobbie: Preprocessor macros are a key-value system essentially. `ifdef` is checking if the key exists regardless of the value, while `if` checks for if it is true.

Dawn/Tint always set a value for `TINT_BUILD_SPV_READER` of either "ON" or "OFF", so IFDEF is always true. Hence, `if` to do the check we really want.

More in fo at https://stackoverflow.com/questions/3802988/difference-between-preprocessor-directives-if-and-ifdef

poweifeng, line 47: why use `TINT_BUILD_SPV_READER` here instead of `FILAMENT_SUPPORTS_WEBGPU`?

third_party/dawn/tnt/CMakeLists.txt
```diff
@@ -12,5 +12,7 @@ set(DAWN_ABSEIL_DIR ${EXTERNAL}/abseil)
 set(DAWN_SPIRV_TOOLS_DIR ${EXTERNAL}/spirv-tools/)
 set(DAWN_SPIRV_HEADERS_DIR ${EXTERNAL}/spirv-headers)
 set(DAWN_GLSLANG_DIR ${EXTERNAL}/glslang/)
-
+if(NOT ANDROID)
```
poweifeng, line 15: maybe use `IS_HOST_PLATFORM` instead?

libs/filamat/src/GLSLPostProcessor.cpp
bridgewaterrobbie: The spirv reader is an optional feature of Tint, and defining things this way lets us exclude the spirv reader even while still supporting the rest of WebGPU (essentially, being able to run WGSL but not transpile to it at runtime)

This seemed a good way to reduce the eventual binary size on mobile devices, under the assumption that the materials are compiled with the app.

third_party/dawn/tnt/CMakeLists.txt
bridgewaterrobbie: I'm not familiar with that one, but I'd assume that is functionally "is not a mobile device"?

If so I can swap to that for sure.

bridgewaterrobbie: Adjusted in latest changelist

libs/filamat/src/GLSLPostProcessor.cpp
poweifeng: Filamat is typically not used during runtime. The exceptions are tests and JitShaderProvider, which don't need to worry about in terms of binary size.  The references from the backend to filamat structures are through filabridge for that very reason (to avoid adding material compilation code to any app binary). 

Imagine, otherwise, including spirv-opt into every app that uses filament.

bridgewaterrobbie: Right, and I wrote this code to try to enable that efficently. Tint is always a part of Dawn, and Dawn is a part of any WebGPU enabled Filament. That tint library though can be either with or without SPV_READER, and so the default of this PR makes it so that on mobile devices the filimat will not use the version of tint that contains the SPV reader.

These IF blocks are to be safe- if you compile for mobile and thus don't have the SPV reader, you will always get an error if you do the unexpected and try to convert to WGSL at runtime.

I had not yet looked into filabridge much yet- if you want this separation to be defined there instead I can look into it, but it should have the same ultimate result.

poweifeng (approved):

PR #8478, opened 2025-02-28
more towards using only static strings for debug tags

pixelflinger:


libs/utils/include/utils/StaticString.h
```diff
@@ -0,0 +1,140 @@
...
+
+namespace utils {
+
+/**
+ * @brief A lightweight string class that stores a pointer to a string literal and its size, without dynamic allocation.
+ *
+ * This class is designed to efficiently represent string literals. It does not allocate any memory
+ * and instead relies on the compiler to manage the memory of the string literals.
+ */
+class StaticString {
```
poweifeng, line 34: `UTILS_PUBLIC ` here

poweifeng (commented):
I think you'll also need to add the header to `DIST_HDRS` of the utils Cmakelists.txt

libs/utils/include/utils/StaticString.h
```diff
@@ -0,0 +1,140 @@
...
+        }
+        return strncmp(mData, other.mData, mSize);
+    }
+
+private:
+    const char* mData = nullptr;
+    size_type mSize = 0;
+};
+
+constexpr bool operator==(const StaticString& lhs, const StaticString& rhs) noexcept {
```
poweifeng, line 116: also `UTILS_PUBLIC` for these operator overloads?  Not sure...

poweifeng (approved):

PR #8482, opened 2025-03-03
Add dawn to third_party from https://github.com/google/dawn

sidreesshah:
Current copy is cloned using hash: 51694fd33cdaaec5bdd7ab3dc4f488f5818fd870

.git and .gitignore have been cleanup. tests,docs,webgpu-cts folders have been deleted.

For more details look at third_party/dawn/tnt/README.md file, which will be pushed in subsequent patch.


poweifeng (approved):
I'd remove `.github/workflows` as well.

sidreesshah:
> I'd remove `.github/workflows` as well.

Thanks for the review. I can see more .github/workflows in filament/third_party folder. They all can be cleaned in a separate patch.

PR #8484, opened 2025-03-04
Fix cleanup logic in VulkanStagePool::terminate

haneulk730:
Updated the `terminate` function to correct resource cleanup. Changed `mUsedStages` to `mUsedImages` and `mFreeStages` to `mFreeImages` to ensure proper memory management during termination.


poweifeng (approved):
thank you!

PR #8487, opened 2025-03-04
Add .clang-format to project root

poweifeng:


bridgewaterrobbie:
@poweifeng  do you intend to merge this? That would be nice as IDEs can automatically pick it up then.

poweifeng:
> @poweifeng do you intend to merge this? That would be nice as IDEs can automatically pick it up then.

Yes, but would like a filament team member to confirm.

bejado (approved):

PR #8488, opened 2025-03-04
External Texture implementation for Vulkan

rafadevai:
Add support for the new createTextureExternalImage2R API instead of createTextureExternalImageR.

Add the support to import AHardwareBuffer objects
into Vulkan in Android.


filament/backend/src/vulkan/platform/VulkanPlatformAndroid.cpp
```diff
@@ -197,20 +322,26 @@ VulkanPlatform::ExternalImageMetadata VulkanPlatform::getExternalImageMetadataIm
...
     return metadata;
 }
 
-VulkanPlatform::ImageData VulkanPlatform::createExternalImageImpl(void* externalImage,
-        VkDevice device, const VkAllocationCallbacks* allocator,
-        const ExternalImageMetadata& metadata) {
-    ImageData data = allocateExternalImage(externalImage, device, allocator, metadata);
+VulkanPlatform::ImageData VulkanPlatform::createExternalImageDataImpl(
+        ExternalImageHandleRef externalImage, VkDevice device,
+        const VkAllocationCallbacks* allocator, const ExternalImageMetadata& metadata) {
```
poweifeng, line 340: can you remove the `allocator` param?  Sorry I missed this when it was added.  In general, we just pass VKALLOC, which is just null, to calls that has an allocator param.

filament/backend/src/vulkan/platform/VulkanPlatformAndroid.cpp
```diff
@@ -103,11 +128,85 @@ void getVKFormatAndUsage(const AHardwareBuffer_Desc& desc, VkFormat& format,
...
+        case AHARDWAREBUFFER_FORMAT_S8_UINT:
+            return TextureFormat::STENCIL8;
+        case AHARDWAREBUFFER_FORMAT_R8_UNORM:
+            return TextureFormat::R8;
+        default:
+            return TextureFormat::UNUSED;
+    }
+}
+
+TextureUsage mapToFilamentUsage(unsigned int usage, TextureFormat format) noexcept {
```
poweifeng, line 172: Are you going to wait for the consolidation of these methods for GL/Vulkan?

filament/backend/src/vulkan/platform/VulkanPlatformAndroid.cpp
```diff
@@ -29,8 +32,28 @@ using namespace bluevk;
...
+            format = VK_FORMAT_R8G8B8_SRGB;
+            break;
+        default:
+            break;
+    }
+
+    return format;
+}
+
 void getVKFormatAndUsage(const AHardwareBuffer_Desc& desc, VkFormat& format,
```
poweifeng, line 55: now that we're passing in a non-reference param at the end, maybe it's a good time to refactor.   Would you mind having this return a `std::pair<VkFormat, VkImageUsageFlags>`?

filament/backend/include/backend/platforms/VulkanPlatformAndroid.h
```diff
@@ -0,0 +1,43 @@
...
+ */
+
+#ifndef TNT_FILAMENT_BACKEND_PLATFORMS_VULKAN_PLATFORM_ANDROID_H
+#define TNT_FILAMENT_BACKEND_PLATFORMS_VULKAN_PLATFORM_ANDROID_H
+
+#include <backend/Platform.h>
+
+#include <android/hardware_buffer.h>
+
+namespace filament::backend::vulkan::android {
```
poweifeng, line 24: can you change `vulkan::android` to `fvkandroid` ? It's continuing an established convention.

filament/backend/src/vulkan/platform/VulkanPlatformAndroid.cpp
alizahlalani: see https://github.com/google/filament/pull/8491

filament/backend/src/vulkan/platform/VulkanPlatformAndroid.cpp
rafadevai: Done

filament/backend/include/backend/platforms/VulkanPlatformAndroid.h
rafadevai: Done

filament/backend/include/backend/platforms/VulkanPlatformAndroid.h
```diff
@@ -0,0 +1,43 @@
...
+ */
+
+#ifndef TNT_FILAMENT_BACKEND_PLATFORMS_VULKAN_PLATFORM_ANDROID_H
+#define TNT_FILAMENT_BACKEND_PLATFORMS_VULKAN_PLATFORM_ANDROID_H
+
+#include <backend/Platform.h>
+
+#include <android/hardware_buffer.h>
+
+namespace filament::backend::fvkandroid {
```
poweifeng, line 24: nit: why do we need another namespace at all?

Is it to avoid `createExternalImage` conflicting? 

And why is `createExternalImage` `UTILS_PUBLIC` ?

poweifeng (approved):
lgtm in general.  Just need to rebase on top of the utils function on main.

Alway have two lingering questions.

Thanks!

filament/backend/include/backend/platforms/VulkanPlatformAndroid.h
rafadevai: The namespace is to avoid the conflict

The other one, in my head to be exposed to impress since it will be an public API, but not 100% sure if its needed or not. What do you think?

poweifeng: I think we can just follow the pattern of GL [here](https://github.com/google/filament/blob/main/filament/backend/include/backend/platforms/PlatformEGLAndroid.h#L63)

with that, i don't think we need the new namespace or the `UTILS_PUBLIC`.

PR #8489, opened 2025-03-05
Add a way to query the transform matrix of a surface texture from the driver

ajmalk:
From: https://developer.android.com/reference/android/graphics/SurfaceTexture

When sampling from the texture one should first transform the texture coordinates using the matrix queried via getTransformMatrix(float[]). The transform matrix may change each time updateTexImage() is called, so it should be re-queried each time the texture image is updated.

Exposing the matrix to the driver is the first step to being able to apply it to the uniform in the shader.

### DIFF
--- a/filament/backend/include/backend/platforms/OpenGLPlatform.h
+++ b/filament/backend/include/backend/platforms/OpenGLPlatform.h
@@ -323,11 +323,9 @@ public:
     /**
      * Returns the transform matrix of the texture attached to the stream.
      * @param stream Stream to get the transform matrix from
-     * @param uvTransform Output parameter: Transform matrix of the image bound to the texture.
+     * @param uvTransform Output parameter: Transform matrix of the image bound to the texture. Returns identity if not supported.
      */
-    virtual void getTransformMatrix(
-            Stream* UTILS_NONNULL stream,
-            math::mat3f* UTILS_NONNULL uvTransform) noexcept;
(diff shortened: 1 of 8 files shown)


filament/backend/include/backend/platforms/OpenGLPlatform.h
```diff
@@ -319,6 +320,15 @@ class OpenGLPlatform : public Platform {
...
             int64_t* UTILS_NONNULL timestamp) noexcept;
 
+    /**
+     * Returns the transform matrix of the texture attached to the stream.
+     * @param stream Stream to get the transform matrix from
+     * @param uvTransform Output parameter: Transform matrix of the image bound to the texture.
+     */
+    virtual void getTransformMatrix(
+            Stream* UTILS_NONNULL stream,
+            math::mat3f* UTILS_NONNULL uvTransform) noexcept;
```
pixelflinger, line 330: return by value in stead of output parameter. document it returns identity by default

pixelflinger (changes requested):
Thanks looks pretty good. I just would like to return by value instead of output parameter

pixelflinger (approved):

filament/backend/include/backend/platforms/OpenGLPlatform.h
ajmalk: Done

z3moon (approved):

ajmalk:
Can we please merge this in now?

PR #8490, opened 2025-03-05
Add a way to specify an additional uniform name along with a sampler param for the associated transform matrix

ajmalk:
Followup to https://github.com/google/filament/pull/8489

From: https://developer.android.com/reference/android/graphics/SurfaceTexture

When sampling from the texture one should first transform the texture coordinates using the matrix queried via getTransformMatrix(float[]). The transform matrix may change each time updateTexImage() is called, so it should be re-queried each time the texture image is updated.

Adding a way to specify the transform name in the material is the second step to being able to apply it to the uniform in the shader.

### DIFF
--- a/docs/Materials.md.html
+++ b/docs/Materials.md.html
@@ -994,9 +994,10 @@ Value
       name must be a valid GLSL identifier. Entries have an optional `precision`, which can be
       one of `default` (best precision for the platform, typically `high` on desktop, `medium` on
       mobile), `low`, `medium`, `high`. The type must be one of the types described in
-      table [materialParamsTypes]. For external textures, entries also have an option 
-      transformUniformName parameter to specify the name of the shader parameter that will be 
-      used to expose the transform matrix associated with the external sampler.  
+      table [materialParamsTypes]. For Android external textures, entries also have an optional 
+      transformName parameter to specify the name of the material parameter that will be 
+      used to expose the transform matrix associated with the external sampler. In iOS and Vulkan,
+      this will always be identity. 
 
          Type          |            Description
 :----------------------|:---------------------------------
--- a/docs/Materials.md.html
+++ b/docs/Materials.md.html
@@ -991,10 +991,12 @@ Type
 
 Value
 :     Each entry is an object with the properties `name` and `type`, both of `string` type. The
-      name must be a valid GLSL identifier. Entries also have an optional `precision`, which can be
+      name must be a valid GLSL identifier. Entries have an optional `precision`, which can be
       one of `default` (best precision for the platform, typically `high` on desktop, `medium` on
       mobile), `low`, `medium`, `high`. The type must be one of the types described in
-      table [materialParamsTypes].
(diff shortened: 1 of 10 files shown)


ajmalk:
> Maybe we could add a new associate method to BufferInterfaceBlock::Builder? It would take a uniform name + sampler binding?

I can make this change but this would still add the field to all the BufferInterfaceBlocks after they are built iiuc.

ajmalk:
Alternatively I could add a constructor with default values to the InterfaceBlockEntry that I think would be better since you can still pass in list list before.

ajmalk:
Done

pixelflinger (commented):
Does it make sense for the `transformName` to be set for non external samplers?  

Also maybe update docs/Materials.md.html around line 1025 to document the new parameter.

ajmalk:
Probably not. I'll add a check.

ajmalk:
done

docs/Materials.md.html
```diff
@@ -991,10 +991,12 @@
 
 Value
 :     Each entry is an object with the properties `name` and `type`, both of `string` type. The
-      name must be a valid GLSL identifier. Entries also have an optional `precision`, which can be
+      name must be a valid GLSL identifier. Entries have an optional `precision`, which can be
       one of `default` (best precision for the platform, typically `high` on desktop, `medium` on
       mobile), `low`, `medium`, `high`. The type must be one of the types described in
-      table [materialParamsTypes].
+      table [materialParamsTypes]. For external textures, entries also have an option
```
bejado, line 997:
```suggestion
      table [materialParamsTypes]. For external textures, entries also have an optional 
```

docs/Materials.md.html
```diff
@@ -991,10 +991,12 @@
 
 Value
 :     Each entry is an object with the properties `name` and `type`, both of `string` type. The
-      name must be a valid GLSL identifier. Entries also have an optional `precision`, which can be
+      name must be a valid GLSL identifier. Entries have an optional `precision`, which can be
       one of `default` (best precision for the platform, typically `high` on desktop, `medium` on
       mobile), `low`, `medium`, `high`. The type must be one of the types described in
-      table [materialParamsTypes].
+      table [materialParamsTypes]. For external textures, entries also have an option 
+      transformUniformName parameter to specify the name of the shader parameter that will be
```
bejado, line 998: I think this should just be `transformName`? Also, we should use "material parameter" instead of "shader parameter."

bejado (commented):
So a new material parameter will be set for each external image by Filament based on the return value of `getTransformMatrix`? Note that this is something specific to Android, we should make it clear that iOS will always be the identity matrix.

docs/Materials.md.html
ajmalk: Done

docs/Materials.md.html
ajmalk: Done

ajmalk:
Added the additional context for iOS and Vulkan.

pixelflinger (approved):

PR #8491, opened 2025-03-05
Add ahb mapping util functions to map ahb usage and format to TextureUsage and TextureFormat

alizahlalani:


filament/backend/src/BackendUtils.cpp
```diff
@@ -536,6 +541,83 @@ bool reshape(const PixelBufferDescriptor& data, PixelBufferDescriptor& reshaped)
...
+        usageFlags |= TextureUsage::PROTECTED;
+    }
+
+    if (usageFlags == TextureUsage::NONE) {
+        usageFlags = TextureUsage::COLOR_ATTACHMENT | TextureUsage::SAMPLEABLE;
+    }
+
+    return usageFlags;
+}
+#endif
```
poweifeng, line 619:
```suggestion
#endif //__ANDROID__
```

poweifeng (approved):

filament/backend/include/backend/DriverEnums.h
```diff
@@ -804,6 +804,33 @@ static constexpr bool isStencilFormat(TextureFormat format) noexcept {
     }
 }
 
+static constexpr bool isColorFormat(TextureFormat format) noexcept {
```
pixelflinger, line 807:
```suggestion
inline constexpr bool isColorFormat(TextureFormat format) noexcept {
```

filament/backend/include/backend/DriverEnums.h
```diff
@@ -804,6 +804,33 @@ static constexpr bool isStencilFormat(TextureFormat format) noexcept {
...
+        case TextureFormat::SRGB8:
+        case TextureFormat::SRGB8_A8:
+        case TextureFormat::RGB8:
+        case TextureFormat::RGB565:
+        case TextureFormat::RGB5_A1:
+        case TextureFormat::RGBA4:
+            return true;
+
+        default:
+            return false;
```
pixelflinger, line 830: don't use a `default:` clause, instead just `return false;` outside of the switch-case.

pixelflinger (changes requested):
I would prefer to move all the android specific code in a new `BackendUtilsAndroid.h`, or even better, if possible, just have this somewhere in Platform? Why can't it be in PlatformEGLAndroid?

filament/backend/include/backend/DriverEnums.h
alizahlalani: I can make a new `BackendUtilsAndroid.h` if that is the preference - it shouldn't be in PlatformEGLAndroid since it's also needed for Vulkan and would lead to repetitive code

filament/backend/include/backend/DriverEnums.h
alizahlalani: This leads to build errors along the lines of "error: 92 enumeration values not handled in switch: 'R8_SNORM', 'R8UI', 'R8I'... [-Werror,-Wswitch]" - I believe the theres a compiler flag cmake triggering this error

filament/backend/include/backend/DriverEnums.h
alizahlalani: Made new BackendUtilsAndroid files

filament/backend/CMakeLists.txt
```diff
@@ -104,6 +104,7 @@ if (FILAMENT_SUPPORTS_OPENGL AND NOT FILAMENT_USE_EXTERNAL_GLES3)
     if (ANDROID)
         list(APPEND SRCS src/opengl/platforms/ExternalStreamManagerAndroid.cpp)
         list(APPEND SRCS src/opengl/platforms/PlatformEGLAndroid.cpp)
+        list(APPEND SRCS src/BackendUtilsAndroid.cpp)
```
poweifeng, line 107: this'll need to be outside of the condition `if (FILAMENT_SUPPORTS_OPENGL AND NOT FILAMENT_USE_EXTERNAL_GLES3)`

filament/backend/include/private/backend/BackendUtilsAndroid.h
```diff
@@ -0,0 +1,63 @@
+/*
+ * Copyright (C) 2019 The Android Open Source Project
```
poweifeng, line 2:
```suggestion
 * Copyright (C) 2025 The Android Open Source Project
```

filament/backend/include/private/backend/BackendUtilsAndroid.h
```diff
@@ -0,0 +1,63 @@
...
+ * Maps AHardwareBuffer format to TextureFormat.
+ */
+TextureFormat mapToFilamentFormat(unsigned int format, bool isSrgbTransfer) noexcept;
+
+/**
+ * Maps AHardwareBuffer usage to TextureUsage.
+ */
+TextureUsage mapToFilamentUsage(unsigned int usage, TextureFormat format) noexcept;
+
+inline constexpr bool isColorFormat(TextureFormat format) noexcept {
```
poweifeng, line 34: I think this can actually stay in `DriverEnums.h`

Or if you think only BackendUtilsAndroid will use it, then you can move it to BackendUtilsAndroid.cpp either as a static method or in anonymous namespace.

filament/backend/src/BackendUtilsAndroid.cpp
```diff
@@ -0,0 +1,98 @@
+/*
+ * Copyright (C) 2019 The Android Open Source Project
```
poweifeng, line 2:
```suggestion
 * Copyright (C) 2025 The Android Open Source Project
```

filament/backend/CMakeLists.txt
alizahlalani: Done

filament/backend/include/private/backend/BackendUtilsAndroid.h
alizahlalani: Done

filament/backend/src/BackendUtilsAndroid.cpp
alizahlalani: Done

filament/backend/include/private/backend/BackendUtilsAndroid.h
alizahlalani: Moved backed to DriverEnums.h - seems correct to leave it where isDepth and isStencil live

filament/backend/include/backend/DriverEnums.h
```diff
@@ -807,6 +807,33 @@ static constexpr bool isStencilFormat(TextureFormat format) noexcept {
     }
 }
 
+static constexpr bool isColorFormat(TextureFormat format) noexcept {
```
poweifeng, line 810: still make it inline here.

pixelflinger (approved):

PR #8494, opened 2025-03-05
Ebridgewater/tint conversion validation

bridgewaterrobbie:
Enable the tint conversion codepath to be run when passed via matc.

Add unit test for tint conversion.

Run this new unit test as part of presubmit.

### DIFF
--- a/filament/backend/include/backend/DriverEnums.h
+++ b/filament/backend/include/backend/DriverEnums.h
@@ -182,6 +182,7 @@ enum class ShaderLanguage {
     SPIRV = 2,
     MSL = 3,
     METAL_LIBRARY = 4,
+    WGSL = 5,
 };
 
 static constexpr const char* shaderLanguageToString(ShaderLanguage shaderLanguage) {
@@ -196,6 +197,8 @@ static constexpr const char* shaderLanguageToString(ShaderLanguage shaderLanguag
             return "MSL";
         case ShaderLanguage::METAL_LIBRARY:
             return "Metal precompiled library";
+        case ShaderLanguage::WGSL:
+            return "WGSL";
     }
 }
 
--- a/libs/filamat/tests/test_filamat.cpp
+++ b/libs/filamat/tests/test_filamat.cpp
@@ -891,6 +891,72 @@ TEST_F(MaterialCompiler, FeatureLevel0Ess3CallFails) {
   EXPECT_FALSE(result.isValid());
 }
 
+#if TINT_BUILD_SPV_READER
+TEST_F(MaterialCompiler, WgslConversionBakedColor) {
+    std::string bakedColorCodeFrag(R"(
+        void material(inout MaterialInputs material) {
+            prepareMaterial(material);
+            material.baseColor = getColor();
+        }
+    )");
+    filamat::MaterialBuilder builder;
+    builder.targetApi(filamat::MaterialBuilderBase::TargetApi::WEBGPU);
+    builder.material(bakedColorCodeFrag.c_str());
+    builder.shading(filamat::MaterialBuilder::Shading::UNLIT);
+    builder.name("BakedColor");
+    builder.culling(CullingMode::NONE);
+    builder.featureLevel(FeatureLevel::FEATURE_LEVEL_0);
+    builder.require(filament::VertexAttribute::COLOR);
+    builder.variantFilter(static_cast<filament::UserVariantFilterMask>(
+            filament::UserVariantFilterBit::SKINNING | filament::UserVariantFilterBit::STE));
+    filamat::Package result = builder.build(*jobSystem);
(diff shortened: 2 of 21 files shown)


filament/backend/include/backend/DriverEnums.h
```diff
@@ -196,6 +197,11 @@ static constexpr const char* shaderLanguageToString(ShaderLanguage shaderLanguag
             return "MSL";
         case ShaderLanguage::METAL_LIBRARY:
             return "Metal precompiled library";
+        case ShaderLanguage::WGSL:
+            return "WebGPU";
```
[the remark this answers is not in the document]
bridgewaterrobbie: Agreed, will fix

libs/matdbg/src/TextWriter.cpp
```diff
@@ -429,7 +429,10 @@ static bool printShaderInfo(ostream& text, const ChunkContainer& container, Chun
             break;
         case ChunkType::MaterialMetalLibrary:
             text << "Metal precompiled shader libraries:" << endl;
-            break;
+        break;
+        case ChunkType::MaterialWgsl:
+            text << "Wgsl precompiled shader libraries:" << endl;
```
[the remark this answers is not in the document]
bridgewaterrobbie: Not sure what you mean here- the case sensitivity?

filament/backend/include/backend/DriverEnums.h
```diff
@@ -196,6 +197,11 @@ static constexpr const char* shaderLanguageToString(ShaderLanguage shaderLanguag
             return "MSL";
         case ShaderLanguage::METAL_LIBRARY:
             return "Metal precompiled library";
+        case ShaderLanguage::WGSL:
+            return "WGSL";
+        // TODO Should we add this?
+        default:
+            return "Unknown";
```
poweifeng, line 204: no defaults in general.  The compiler will catch missing cases.

libs/filamat/tests/test_filamat.cpp
```diff
@@ -891,7 +891,76 @@ TEST_F(MaterialCompiler, FeatureLevel0Ess3CallFails) {
...
+    )");
+    filamat::MaterialBuilder builder;
+    builder.targetApi(filamat::MaterialBuilderBase::TargetApi::WEBGPU);
+    builder.material(bakedColorCodeFrag.c_str());
+    builder.shading(filamat::MaterialBuilder::Shading::UNLIT);
+    builder.name("BakedColor");
+    builder.culling(CullingMode::NONE);
+    builder.featureLevel(FeatureLevel::FEATURE_LEVEL_0);
+    builder.require(filament::VertexAttribute::COLOR);
+    builder.variantFilter(static_cast<filament::UserVariantFilterMask>(filament::UserVariantFilterBit::SKINNING | filament::UserVariantFilterBit::STE));
```
poweifeng, line 910: nit: line > 100 chars

libs/filamat/tests/test_filamat.cpp
```diff
@@ -891,7 +891,76 @@ TEST_F(MaterialCompiler, FeatureLevel0Ess3CallFails) {
...
+#endif
+
 int main(int argc, char** argv) {
     ::testing::InitGoogleTest(&argc, argv);
-    return RUN_ALL_TESTS();
+    const int rv = RUN_ALL_TESTS();
+    if (testing::UnitTest::GetInstance()->test_to_run_count() == 0) {
+        //If you run a test filter that contains 0 tests that was likely not intentional. Fail in that scenario.
+        return 1;
+    }
```
poweifeng, line 964: this seems ok, but also not the default behavior of tests on google infra in general.

filament/backend/include/backend/DriverEnums.h
bridgewaterrobbie: Fixed

libs/filamat/tests/test_filamat.cpp
bridgewaterrobbie: Fixed

libs/filamat/tests/test_filamat.cpp
bridgewaterrobbie: Yes, this is really just to handle the fact that for now these tests are protected by the default-off `FILAMENT_SUPPORTS_WEBGPU`. I did not want to make the default "Run all tests" run this and fail when WebGPU is not supported, but I also did not want to accidentally pass things if we somehow had an issue building the WebGPU support. 

Assuming you are fine with this for now, we can leave it and remove it later. Alternatively, we can remove it and assume the risk of getting false-pass if the `-W` somehow stops building WebGPU support

poweifeng: sure, lgtm as is

poweifeng (approved):

PR #8496, opened 2025-03-06
Add a way to pass a transform matrix along with Stream::setAcquiredImage

ajmalk:
Followup to https://github.com/google/filament/pull/8493

Streams in OpenGL are either backed by SurfaceTexture (Native in filament) or ImageReader (Acquired in filament). SurfaceTextures have a transform attached to the image that's updated every frame but have to be read from the GL thread. The ImageReader path has the same transform but is read from the main thread. The transform in the SurfaceTexture path can be optionally requested in the shader (see https://github.com/google/filament/pull/8490). This PR adds a way for the  ImageReader path to also update this transform.


filament/backend/include/backend/AcquiredImage.h
```diff
@@ -18,6 +18,7 @@
 #define TNT_FILAMENT_BACKEND_PRIVATE_ACQUIREDIMAGE_H
 
 #include <backend/DriverEnums.h>
+#include "math/mat3.h"
```
pixelflinger, line 21: use <>

filament/backend/src/DriverBase.h
```diff
@@ -176,7 +176,7 @@ class DriverBase : public Driver {
         CallbackData(CallbackData&&) = delete;
         CallbackData& operator=(CallbackData const &) = delete;
         CallbackData& operator=(CallbackData&&) = delete;
-        void* storage[8] = {};
+        void* storage[13] = {};
```
pixelflinger, line 179: you can't change that, it was very much chosen to be 64 bytes. there should have been a comment.

filament/backend/src/opengl/OpenGLDriver.cpp
```diff
@@ -2087,7 +2087,7 @@ Handle<HwStream> OpenGLDriver::createStreamAcquired() {
 // called only once per frame. If the user pushes images to the same stream multiple times in a
 // single frame, we emit a warning and honor only the final image, but still invoke all callbacks.
 void OpenGLDriver::setAcquiredImage(Handle<HwStream> sh, void* hwbuffer,
-        CallbackHandler* handler, StreamCallback cb, void* userData) {
+        CallbackHandler* handler, StreamCallback cb, void* userData, math::mat3f transform) {
```
pixelflinger, line 2090: since we have access to the stream here, why do we need to receive the transform? can't we just ask the stream?

pixelflinger (changes requested):

filament/backend/src/opengl/OpenGLDriver.cpp
ajmalk: My understanding is that there's two options for working with streams in android and in the SurfaceTexture path, filament manages the SurfaceTexture (and thus calling updateTexImage and calling the transform) and with ImageReader, impress manages the ImageReader and pulling the frames and feeding them them to filament one by one. Filament doesn't have access to the ImageReader object in the latter path and thus can't get the transform from it. But because the frames are used from another thread, we still need to pass it all the way to filament for this to work.

filament/backend/include/backend/AcquiredImage.h
ajmalk: Done

filament/backend/src/opengl/OpenGLDriver.cpp
ajmalk: The way the "ACQUIRED" path works is this function is called repeatedly by impress, feeding it new frames and the user_thread struct in Stream acts as a 1 frame queue. So the "current" frame is the user_thread.aquired and the transform that needs to be used in the current frame is the one associated with that frame.

filament/backend/src/metal/MetalDriver.mm
```diff
@@ -985,7 +985,7 @@
 }
 
 void MetalDriver::setAcquiredImage(Handle<HwStream> sh, void* image,
-        CallbackHandler* handler, StreamCallback cb, void* userData) {
+        CallbackHandler* handler, StreamCallback cb, void* userData, math::mat3f transform) {
```
pixelflinger, line 988: out of curiosity why is this not passed by `const&` ?

filament/backend/include/private/backend/DriverAPI.inc
```diff
@@ -339,7 +339,7 @@ DECL_DRIVER_API_N(destroyDescriptorSet,         backend::DescriptorSetHandle, ds
 DECL_DRIVER_API_SYNCHRONOUS_0(void, terminate)
 DECL_DRIVER_API_SYNCHRONOUS_N(backend::StreamHandle, createStreamNative, void*, stream)
 DECL_DRIVER_API_SYNCHRONOUS_0(backend::StreamHandle, createStreamAcquired)
-DECL_DRIVER_API_SYNCHRONOUS_N(void, setAcquiredImage, backend::StreamHandle, stream, void*, image, backend::CallbackHandler*, handler, backend::StreamCallback, cb, void*, userData)
+DECL_DRIVER_API_SYNCHRONOUS_N(void, setAcquiredImage, backend::StreamHandle, stream, void*, image, backend::CallbackHandler*, handler, backend::StreamCallback, cb, void*, userData, math::mat3f, transform)
```
pixelflinger, line 342: I'm sorry I didnt' catch this before , but I would like the order of the parameters to be:

stream, image, transform, handler, callback, userdata

ie keep the image and transform together and keep the callback at the end.

filament/backend/src/opengl/OpenGLDriver.cpp
```diff
@@ -2167,8 +2167,12 @@ int64_t OpenGLDriver::getStreamTimestamp(Handle<HwStream> sh) {
 
 math::mat3f OpenGLDriver::getStreamTransformMatrix(Handle<HwStream> sh) {
     if (sh) {
-        GLStream* s = handle_cast<GLStream*>(sh);        
-        return mPlatform.getTransformMatrix(s->stream);
+        GLStream* s = handle_cast<GLStream*>(sh);
+        if (s->streamType == StreamType::NATIVE) {
+            return mPlatform.getTransformMatrix(s->stream);
+        } else {
+            return s->user_thread.acquired.transform;
```
pixelflinger, line 2174: I think the threading model is wrong here.  getStreamTransformMatrix() is always called from the backend thread, but `user_thread` is updated from the `setAcquiredImage` which is a synchronous call (user thread).

So there is a race.

I think the access in beginFrame (`user_thread.timestamp`) is actually a bug.

pixelflinger (changes requested):

filament/backend/src/opengl/OpenGLDriver.cpp
ajmalk: I moved it to another struct that's only access from the gl thread.

filament/backend/include/private/backend/DriverAPI.inc
ajmalk: Done

filament/backend/src/metal/MetalDriver.mm
ajmalk: You're right. Fixed.

filament/backend/src/opengl/OpenGLDriver.h
```diff
@@ -161,7 +161,15 @@ class OpenGLDriver final : public OpenGLDriverBase {
...
             AcquiredImage pending;
+            math::mat3f transform;
         } user_thread;
+
+        /*
+         * The fields below are accessed from the GL thread
+         */
+         struct {
+            math::mat3f transform;
+         } gl_thread;
```
pixelflinger, line 172: That's "fine" but it's not necessary because the default assumption is that Handle<> fields are accessed from the gl thread.

filament/backend/src/vulkan/VulkanDriver.cpp
```diff
@@ -26,6 +26,7 @@
 #include "VulkanHandles.h"
 #include "VulkanMemory.h"
 #include "VulkanTexture.h"
+#include "math/mat3.h"
```
pixelflinger, line 29: That should be moved below the <utils/> includes and use `<>` syntax

filament/include/filament/Stream.h
```diff
@@ -25,6 +25,7 @@
 #include <utils/compiler.h>
 
 #include <stdint.h>
+#include <math/mat3.h>
```
pixelflinger, line 28: move below <utils/> and above c and std headers

filament/src/details/Stream.h
```diff
@@ -36,8 +38,8 @@ class FStream : public Stream {
 
     backend::Handle<backend::HwStream> getHandle() const noexcept { return mStreamHandle; }
 
-    void setAcquiredImage(void* image, Callback callback, void* userdata) noexcept;
-    void setAcquiredImage(void* image, backend::CallbackHandler* handler, Callback callback, void* userdata) noexcept;
+    void setAcquiredImage(void* image, Callback callback, void* userdata, math::mat3f const & transform) noexcept;
+    void setAcquiredImage(void* image, backend::CallbackHandler* handler, Callback callback, void* userdata, math::mat3f const & transform) noexcept;
```
pixelflinger, line 42: no space after `const`

pixelflinger (commented):
We're only left with minor style issues. Implementation looks good.

filament/backend/src/opengl/OpenGLDriver.h
ajmalk: Just made it a member var.

filament/src/details/Stream.h
ajmalk: Done

pixelflinger (approved):

filament/src/details/Stream.cpp
```diff
@@ -95,13 +95,13 @@ void FStream::terminate(FEngine& engine) noexcept {
 }
 
 void FStream::setAcquiredImage(void* image,
-        Callback const callback, void* userdata) noexcept {
-    mEngine.getDriverApi().setAcquiredImage(mStreamHandle, image, nullptr, callback, userdata);
+    Callback const callback, void* userdata, math::mat3f const& transform) noexcept {
+mEngine.getDriverApi().setAcquiredImage(mStreamHandle, image, transform, nullptr, callback, userdata);
```
poweifeng, line 99: nit: indent

filament/src/details/Stream.cpp
```diff
@@ -95,13 +95,13 @@ void FStream::terminate(FEngine& engine) noexcept {
...
 }
 
 void FStream::setAcquiredImage(void* image,
-        CallbackHandler* handler, Callback const callback, void* userdata) noexcept {
-    mEngine.getDriverApi().setAcquiredImage(mStreamHandle, image, handler, callback, userdata);
+    CallbackHandler* handler, Callback const callback, void* userdata, math::mat3f const& transform) noexcept {
+mEngine.getDriverApi().setAcquiredImage(mStreamHandle, image, transform, handler, callback, userdata);
```
poweifeng, line 104: nit: indent

poweifeng (approved):

filament/src/details/Stream.cpp
ajmalk: done

filament/src/details/Stream.cpp
ajmalk: done

PR #8505, opened 2025-03-07
Add helpers for backend tests (cleanup+render loops).

mhoff12358:
Both for cleaning up graphics resources at the end of a test and for starting/stopping render loops.

### DIFF
--- a/filament/backend/test/Lifetimes.cpp
+++ b/filament/backend/test/Lifetimes.cpp
@@ -1,0 +1,92 @@
+/*
+ * Copyright (C) 2025 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "Lifetimes.h"
+
+#include "backend/Platform.h"
+#include "private/backend/DriverApi.h"
+
+RenderFrame::RenderFrame(filament::backend::DriverApi& api): mApi(api) {
+    mApi.beginFrame(0, 0, 0);
+}
+
+RenderFrame::~RenderFrame() {
+    mApi.endFrame(0);
+}
+
+Cleanup::Cleanup(filament::backend::DriverApi& driverApi) : mDriverApi(driverApi) {}
+
+Cleanup::~Cleanup() {
+    for (const auto swapChain: mSwapChains) {
+        mDriverApi.destroySwapChain(swapChain);
+    }
+    for (const auto descriptorSet : mDescriptorSets) {
+        mDriverApi.destroyDescriptorSet(descriptorSet);
+    }
+    for (const auto descriptorSetLayout : mDescriptorSetLayouts) {
+        mDriverApi.destroyDescriptorSetLayout(descriptorSetLayout);
+    }
+    for (const auto bufferObject : mBufferObjects) {
+        mDriverApi.destroyBufferObject(bufferObject);
+    }
+    for (const auto program : mPrograms) {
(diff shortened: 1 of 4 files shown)


mhoff12358:
This is missing a FIXES, waiting on getting the correct bug ID.

filament/backend/test/Lifetimes.h
```diff
@@ -0,0 +1,73 @@
...
+    ~RenderFrame();
+
+    RenderFrame(RenderFrame const& rhs) noexcept = delete;
+    RenderFrame& operator=(RenderFrame const& rhs) noexcept = delete;
+    RenderFrame(RenderFrame const&& rhs) noexcept = delete;
+    RenderFrame& operator=(RenderFrame const&& rhs) noexcept = delete;
+
+private:
+    filament::backend::DriverApi& api_;
+    filament::backend::SwapChainHandle swapChain_;
```
bejado, line 40: Nit: one convention in our codebase is to prefix member variables with `m`, so these would be `mApi` and `mSwapChain`.

filament/backend/test/Lifetimes.cpp
```diff
@@ -0,0 +1,86 @@
...
+}
+
+RenderFrame::~RenderFrame() {
+    api_.endFrame(0);
+}
+
+Cleanup::Cleanup(const test::BackendTest& testFixture) : testFixture_(testFixture) {}
+
+Cleanup::~Cleanup() {
+    auto& api = testFixture_.getDriverApi();
```
bejado, line 33: Will this compile? `testFixture` is marked const, but the `getDriverApi()` method is non-const.

filament/backend/test/Lifetimes.cpp
```diff
@@ -0,0 +1,86 @@
...
+    for (const auto renderTarget : renderTargets_) {
+        api.destroyRenderTarget(renderTarget);
+    }
+
+    for (const auto& callback : callbacks_) {
+        callback();
+    }
+}
+
+void Cleanup::AddSwapChain(filament::backend::SwapChainHandle swapChain) {
```
bejado, line 62: Nit: Our convention within Filament is to use lowerCamelCase for method names: `addSwapChain`, `addDescriptorSet`, etc.

filament/backend/test/Lifetimes.h
```diff
@@ -0,0 +1,73 @@
...
+};
+
+class Cleanup {
+public:
+    explicit Cleanup(const test::BackendTest& testFixture);
+    ~Cleanup();
+
+    // These methods can't be overloaded because of a compiler error caused by the type safe casting
+    // constructor needing to be resolved for each option.
+    void AddSwapChain(filament::backend::SwapChainHandle swapChain);
```
bejado, line 50: I had a couple optional ideas here. Maybe name all of these methods simply `add`, and overload based on the argument type. Would save some verbosity since the type of the handle will be easy to infer based on context. For example:

```
Handle<HwTexture> const srcTexture = api.createTexture(...);
cleanup.add(srcTexture);
```

You could also return the handle from the `add` method so we can do things like this:

```
Handle<HwTexture> const srcTexture = cleanup.add(api.createTexture(...));
```

bejado (changes requested):

filament/backend/test/Lifetimes.h
mhoff12358: I initially wrote this as a templated `add` method but IIRC there was some weird templating issue. Due to all the random github delays it's been like 3 weeks since I wrote this so I don't remember what the issue was though, I'll change back to that and either it'll work or I'll remember what the problem was.

filament/backend/test/Lifetimes.cpp
mhoff12358: Whoops, this change was sliced off of a larger branch and that snuck through somehow sorry. That's what I get for trying to rush it out on a friday evening.

filament/backend/test/Lifetimes.h
bejado: Sounds good! We can also address these improvements in follow-up PRs.

mhoff12358: 1) Haha, I had left a comment with the issue. It was that method *overloading* doesn't work because there's an implicit constructor from `Handle<A>` to `Handle<B>` if A is a base of B and that throws off the overloading.
It seems template specialization doesn't have that issue though.
2) Both were pretty easy changes to throw in.

PR #8511, opened 2025-03-10
[draft] fix webgpu build

sidreesshah:
Android build for webgpu is broken because of spir-v reader.

### DIFF
--- a/libs/filamat/src/GLSLPostProcessor.cpp
+++ b/libs/filamat/src/GLSLPostProcessor.cpp
@@ -44,7 +44,7 @@
 #include <unordered_map>
 #include <vector>
 
-#ifdef FILAMENT_SUPPORTS_WEBGPU
+#ifdef TINT_BUILD_SPV_READER
 #include <tint/tint.h>
 #endif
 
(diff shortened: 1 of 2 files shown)


libs/filamat/src/GLSLPostProcessor.cpp
```diff
@@ -44,7 +44,7 @@
 #include <unordered_map>
 #include <vector>
 
-#ifdef FILAMENT_SUPPORTS_WEBGPU
+#ifdef TINT_BUILD_SPV_READER
```
bridgewaterrobbie, line 47: Should be `If` not ifdef

Not a huge issue necessarily but optimization

bridgewaterrobbie:
@sidreesshah  is this PR still relevant, or did others supplant it?

libs/filamat/src/GLSLPostProcessor.cpp
kpiascik: actually this makes a big difference.  If it's defined and set to `OFF`, which this patch does , then this header will still be included

bridgewaterrobbie: Good catch, agreed. I didn’t notice the cmake change in this context originally when I commented this

PR #8512, opened 2025-03-10
Refactored external image handling in PlatformEGLAndroid

alizahlalani:


filament/backend/src/opengl/platforms/PlatformEGLAndroid.cpp
```diff
@@ -56,8 +57,8 @@
 // We require filament to be built with an API 19 toolchain, before that, OpenGLES 3.0 didn't exist
 // Actually, OpenGL ES 3.0 was added to API 18, but API 19 is the better target and
 // the minimum for Jetpack at the time of this comment.
-#if __ANDROID_API__ < 19
-#   error "__ANDROID_API__ must be at least 19"
+#if __ANDROID_API__ < 21
+#error "__ANDROID_API__ must be at least 21"
```
poweifeng, line 61: why is this necessary?

filament/backend/src/opengl/platforms/PlatformEGLAndroid.cpp
```diff
@@ -209,20 +208,22 @@ Driver* PlatformEGLAndroid::createDriver(void* sharedContext,
...
-                "eglPresentationTimeANDROID");
+        eglPresentationTimeANDROID =
+                (PFNEGLPRESENTATIONTIMEANDROIDPROC) eglGetProcAddress("eglPresentationTimeANDROID");
     }
 
     if (extensions.has("EGL_ANDROID_get_frame_timestamps")) {
-        eglGetCompositorTimingSupportedANDROID = (PFNEGLGETCOMPOSITORTIMINGSUPPORTEDANDROIDPROC)eglGetProcAddress(
-                "eglGetCompositorTimingSupportedANDROID");
-        eglGetCompositorTimingANDROID = (PFNEGLGETCOMPOSITORTIMINGANDROIDPROC)eglGetProcAddress(
+        eglGetCompositorTimingSupportedANDROID =
```
poweifeng, line 216: nit: why do all these reformatting? suggest that if you're changing the logic, then leave the formatting as is (unless it's cleaning up like breaking up a too-long line).

filament/backend/src/opengl/platforms/PlatformEGLAndroid.cpp
```diff
@@ -233,31 +234,100 @@ Driver* PlatformEGLAndroid::createDriver(void* sharedContext,
...
+}
+
+bool PlatformEGLAndroid::setImage(ExternalImageEGLAndroid const* eglExternalImage,
+        UTILS_UNUSED_IN_RELEASE ExternalTexture* texture) noexcept {
+    AHardwareBuffer* hardwareBuffer = eglExternalImage->aHardwareBuffer;
+
+    // Get the EGL client buffer from AHardwareBuffer
+    EGLClientBuffer clientBuffer = eglGetNativeClientBufferANDROID(hardwareBuffer);
+    // Questions around attributes with isSrgbTransfer and protected content
+    EGLint imageAttrs[] = { EGL_IMAGE_PRESERVED_KHR, EGL_TRUE, EGL_NONE, EGL_NONE, EGL_NONE,
```
poweifeng, line 286: order by n x 2 by formatting to match field and value.

filament/src/details/Texture.cpp
```diff
@@ -187,6 +184,9 @@ Texture* Texture::Builder::build(Engine& engine) {
     // SAMPLER_EXTERNAL implies imported.
     if (mImpl->mTarget == SamplerType::SAMPLER_EXTERNAL) {
         mImpl->mExternal = true;
+    } else {
```
poweifeng, line 187: nit: keep this up top and just condition on non-external, but maybe the real fix.

Also I wonder if the format here has to match the format we'd get from ahardwarebuffer.

filament/backend/src/opengl/platforms/PlatformEGLAndroid.cpp
alizahlalani: The formatting is all due to clang-format - https://github.com/google/filament/pull/8487

For logic, commit 1 might make this easier to review

filament/backend/src/opengl/platforms/PlatformEGLAndroid.cpp
poweifeng: It looks like it changed from 19 to 21?

filament/backend/src/opengl/platforms/PlatformEGLAndroid.cpp
poweifeng: Yes, but only run clang-format on the lines you changed.  Some non-clang-format looking lines are intentional for readability (and personal preferences).

filament/backend/src/opengl/platforms/PlatformEGLAndroid.cpp
alizahlalani: good catch that was an error when I reverted - fixed

filament/backend/src/opengl/platforms/PlatformEGLAndroid.cpp
alizahlalani: done.

filament/backend/src/opengl/platforms/PlatformEGLAndroid.cpp
alizahlalani: done

filament/src/details/Texture.cpp
alizahlalani: done re: nit

re: " if the format here has to match the format we'd get from ahardwarebuffer" - the format that was passed in via impress was directly derived from the hardwarebuffer via PlatformEGLAndroid::createExternalImage via ExternalImageEGLAndroid

poweifeng: yes, but I mean in the general case; not every client is impress.

alizahlalani: Given that this whole model relies on the client registering an image or ahardwarebuffer by platform directly, I'm not sure an extra check is needed, but I will defer to FIlament team on if you want this added. We don't currently retain MBuffer on engine in this implementation, but I'm sure there's some way to derive it from the image handle.

poweifeng: yeah, i'm not sure about this.  This code has existed before for other (non AHB) external samplers.  I wonder if skipping it entirely is ok. I'll leave this one to Mathias to comment.

poweifeng (approved):

filament/backend/src/opengl/platforms/PlatformEGLAndroid.cpp
```diff
@@ -233,24 +236,98 @@ Driver* PlatformEGLAndroid::createDriver(void* sharedContext,
...
+        UTILS_UNUSED_IN_RELEASE ExternalTexture* texture) noexcept {
+    AHardwareBuffer* hardwareBuffer = eglExternalImage->aHardwareBuffer;
+
+    // Get the EGL client buffer from AHardwareBuffer
+    EGLClientBuffer clientBuffer = eglGetNativeClientBufferANDROID(hardwareBuffer);
+    EGLint imageAttrs[] = {
+        EGL_IMAGE_PRESERVED_KHR, EGL_TRUE,
+        EGL_NONE, EGL_NONE,
+        EGL_NONE, EGL_NONE,
+        EGL_NONE, EGL_NONE
```
poweifeng, line 291: one more nit: looks like attrIndex is at most 6?  this array is 8 elements.

alizahlalani: great catch - fixed.

filament/backend/src/opengl/platforms/PlatformEGLAndroid.cpp
```diff
@@ -233,24 +236,98 @@ Driver* PlatformEGLAndroid::createDriver(void* sharedContext,
...
+            EGL_NATIVE_BUFFER_ANDROID, clientBuffer, imageAttrs);
+    if (eglImage == EGL_NO_IMAGE_KHR) {
+        // Handle error
+        slog.e << "Failed to create EGL image" << io::endl;
+        return false;
+    }
+
+    // Create and bind the OpenGL texture
+    glActiveTexture(GL_TEXTURE0);
+    glBindTexture(texture->target, texture->id);
```
pixelflinger, line 314: If you change ANY GL state, you must restore it upon returning -- because the backend caches the state and the cache would become out of date.

pixelflinger (changes requested):

filament/backend/src/opengl/platforms/PlatformEGLAndroid.cpp
alizahlalani: Trying to understand the lifecycle of this - where would I restore the state? In PlatformEGLAndroid?

filament/backend/include/backend/platforms/PlatformEGLAndroid.h
```diff
@@ -39,12 +40,26 @@ class ExternalStreamManagerAndroid;
...
+    struct ExternalImageEGLAndroid : public ExternalImageEGL {
+        AHardwareBuffer* aHardwareBuffer = nullptr;
+        bool sRGB = false;
+        unsigned int width;  // Texture width
+        unsigned int height; // Texture height
+        TextureFormat format;// Texture format
+        TextureUsage usage;  // Texture usage flags
+    protected:
+        ~ExternalImageEGLAndroid() override;
+    };
```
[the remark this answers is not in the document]
alizahlalani: resolved as discussed offline - I left the fields in ExternalImageEGLAndroid and made an extra metadata struct thats public to avoid additional AHardwareBufferDescribe calls in PlatformEGLAndroid::setImage which relies on the correct texture format and usage

filament/backend/include/backend/platforms/PlatformEGLAndroid.h
```diff
@@ -39,12 +40,26 @@ class ExternalStreamManagerAndroid;
...
+    /**
+     * Creates an ExternalImage from a EGLImageKHR
+     */
+    ExternalImageHandle createExternalImage(AHardwareBuffer const* buffer, bool sRGB) noexcept;
+
+    struct ExternalImageEGLAndroid : public ExternalImageEGL {
+        AHardwareBuffer* aHardwareBuffer = nullptr;
+        bool sRGB = false;
+        unsigned int width;  // Texture width
+        unsigned int height; // Texture height
```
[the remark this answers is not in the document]
alizahlalani: done

filament/backend/include/backend/platforms/PlatformEGLAndroid.h
```diff
@@ -39,12 +40,26 @@ class ExternalStreamManagerAndroid;
...
 
-protected:
+    /**
+     * Creates an ExternalImage from a EGLImageKHR
+     */
+    ExternalImageHandle createExternalImage(AHardwareBuffer const* buffer, bool sRGB) noexcept;
+
+    struct ExternalImageEGLAndroid : public ExternalImageEGL {
+        AHardwareBuffer* aHardwareBuffer = nullptr;
+        bool sRGB = false;
```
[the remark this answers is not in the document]
alizahlalani: done

filament/backend/src/opengl/platforms/PlatformEGLAndroid.cpp
alizahlalani: resolved as discussed offline

filament/backend/src/opengl/platforms/PlatformEGLAndroid.cpp
```diff
@@ -337,8 +341,12 @@ bool PlatformEGLAndroid::setImage(ExternalImageEGLAndroid const* eglExternalImag
     error = glGetError();
     if (error != GL_NO_ERROR) {
         slog.e << "Error after glEGLImageTargetTexture2DOES: " << error << io::endl;
+        glDeleteTextures(1, &texture->id);
+        eglDestroyImageKHR(eglGetCurrentDisplay(), eglImage);
         return false;
     }
+    glBindTexture(GL_TEXTURE_2D, prevTexture);
```
rafadevai, line 348: IRC the orders of the calls have to be in the opposite order, glActiveTexture first
Second, better restore the state before checking glEGLImageTargetTexture2DOES error so its properly restored also in the case of an error.

alizahlalani: done

filament/backend/include/backend/platforms/PlatformEGLAndroid.h
```diff
@@ -49,17 +49,15 @@ class PlatformEGLAndroid : public PlatformEGL {
      */
     ExternalImageHandle createExternalImage(AHardwareBuffer const* buffer, bool sRGB) noexcept;
 
-    struct ExternalImageEGLAndroid : public ExternalImageEGL {
-        AHardwareBuffer* aHardwareBuffer = nullptr;
-        bool sRGB = false;
-        unsigned int width;  // Texture width
-        unsigned int height; // Texture height
+    struct ExternalImageAndroidMetadata {
```
pixelflinger, line 52: Rename `ExternalImageDescAndroid` (to mimic AHB).

filament/backend/include/backend/platforms/PlatformEGLAndroid.h
```diff
@@ -49,17 +49,15 @@ class PlatformEGLAndroid : public PlatformEGL {
...
+        uint32_t width;      // Texture width
+        uint32_t height;     // Texture height
         TextureFormat format;// Texture format
         TextureUsage usage;  // Texture usage flags
-    protected:
-        ~ExternalImageEGLAndroid() override;
     };
 
+    ExternalImageAndroidMetadata getImageMetadata(ExternalImageHandle externalImage) noexcept;
```
pixelflinger, line 59: Rename `getExternalImageDesc`.

filament/backend/src/opengl/platforms/PlatformEGLAndroid.cpp
```diff
@@ -233,24 +236,124 @@ Driver* PlatformEGLAndroid::createDriver(void* sharedContext,
...
+    }
+
+    // Create and bind the OpenGL texture
+    GLint prevActiveTexture, prevTexture;
+    glGetIntegerv(GL_ACTIVE_TEXTURE, &prevActiveTexture);
+    glGetIntegerv(GL_TEXTURE_BINDING_2D, &prevTexture);
+    glActiveTexture(GL_TEXTURE0);
+    glBindTexture(texture->target, texture->id);
+    GLenum error = glGetError();
+    if (error != GL_NO_ERROR) {
```
pixelflinger, line 334:
```suggestion
    if (UTILS_UNLIKELY(error != GL_NO_ERROR)) {
```

filament/backend/src/opengl/platforms/PlatformEGLAndroid.cpp
```diff
@@ -233,24 +236,124 @@ Driver* PlatformEGLAndroid::createDriver(void* sharedContext,
...
+        slog.e << "Error after glBindTexture: " << error << io::endl;
+        glDeleteTextures(1, &texture->id);
+        eglDestroyImageKHR(eglGetCurrentDisplay(), eglImage);
+        glActiveTexture(prevActiveTexture);
+        glBindTexture(GL_TEXTURE_2D, prevTexture);
+        return false;
+    }
+    glEGLImageTargetTexture2DOES(texture->target, static_cast<GLeglImageOES>(eglImage));
+    error = glGetError();
+    if (error != GL_NO_ERROR) {
```
pixelflinger, line 344:
```suggestion
    if (UTILS_UNLIKELY(error != GL_NO_ERROR)) {
```

filament/backend/src/opengl/platforms/PlatformEGLAndroid.cpp
```diff
@@ -233,24 +236,124 @@ Driver* PlatformEGLAndroid::createDriver(void* sharedContext,
...
-Platform::ExternalImageHandle PlatformEGLAndroid::createExternalImage(AHardwareBuffer const* buffer, bool sRGB) noexcept {
-    auto* const p = new(std::nothrow) ExternalImageEGLAndroid;
-    p->aHardwareBuffer = const_cast<AHardwareBuffer*>(buffer);
-    p->sRGB = sRGB;
-    return ExternalImageHandle{ p };
+Platform::ExternalImageHandle PlatformEGLAndroid::createExternalImage(AHardwareBuffer const* buffer,
+        bool sRGB) noexcept {
+    if (__builtin_available(android 26, *)) {
+        auto* const p = new (std::nothrow) ExternalImageEGLAndroid;
+        auto hardwareBuffer = const_cast<AHardwareBuffer*>(buffer);
```
pixelflinger, line 243: I think you need to call `AHardwareBuffer_acquire` here, to keep a reference to it.

filament/backend/src/opengl/platforms/PlatformEGLAndroid.cpp
```diff
@@ -233,24 +236,124 @@ Driver* PlatformEGLAndroid::createDriver(void* sharedContext,
 
 PlatformEGLAndroid::ExternalImageEGLAndroid::~ExternalImageEGLAndroid() = default;
```
pixelflinger, line 237: This will need to call `AHardwareBuffer_release` (to balance the `AHardwareBuffer_acquire` we're adding in create)

pixelflinger (changes requested):

filament/backend/include/backend/platforms/PlatformEGLAndroid.h
alizahlalani: done

filament/backend/include/backend/platforms/PlatformEGLAndroid.h
alizahlalani: done

filament/backend/src/opengl/platforms/PlatformEGLAndroid.cpp
alizahlalani: done

filament/backend/src/opengl/platforms/PlatformEGLAndroid.cpp
alizahlalani: done

filament/backend/src/opengl/platforms/PlatformEGLAndroid.cpp
alizahlalani: done

pixelflinger (approved):

PR #8514, opened 2025-03-10
webgpu: fix filamat android build

poweifeng:
Also fix a change in backend API.

### DIFF
--- a/android/filamat-android/CMakeLists.txt
+++ b/android/filamat-android/CMakeLists.txt
@@ -11,6 +11,13 @@ if (FILAMENT_SUPPORTS_VULKAN)
     message("Library filamat ignores Vulkan settings")
 endif()
 
+if (FILAMENT_SUPPORTS_WEBGPU)
+    add_library(webgpu_dawn STATIC IMPORTED)
+    set_target_properties(webgpu_dawn PROPERTIES IMPORTED_LOCATION
+            ${FILAMENT_DIR}/lib/${ANDROID_ABI}/libwebgpu_dawn.a)
+endif()
+
+
 add_library(${FILAMAT_FLAVOR} STATIC IMPORTED)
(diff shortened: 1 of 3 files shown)


android/filamat-android/CMakeLists.txt
```diff
@@ -11,6 +11,13 @@ if (FILAMENT_SUPPORTS_VULKAN)
     message("Library filamat ignores Vulkan settings")
 endif()
 
+if (FILAMENT_SUPPORTS_WEBGPU)
+    add_library(webgpu_dawn STATIC IMPORTED)
+    set_target_properties(webgpu_dawn PROPERTIES IMPORTED_LOCATION
+            ${FILAMENT_DIR}/lib/${ANDROID_ABI}/libwebgpu_dawn.a)
+endif()
+
+
```
sidreesshah, line 20: nit: extra line

sidreesshah (approved):

android/filamat-android/CMakeLists.txt
```diff
@@ -50,4 +57,5 @@ target_link_libraries(filamat-jni
         utils
         log
         smol-v
+        $<$<STREQUAL:${FILAMENT_SUPPORTS_WEBGPU},ON>:webgpu_dawn> # for tint
```
bridgewaterrobbie, line 60: cmake style nit: Usually with CMake I have seen a list built up based on `if`, then the resulting list passed as an arg here rather than a dependent line like this.

If it works there isn't anything wrong with it though.

android/filamat-android/CMakeLists.txt
poweifeng: will resolve in next patch

android/filamat-android/CMakeLists.txt
poweifeng: also used [here](https://github.com/google/filament/blob/main/android/filament-android/CMakeLists.txt)

PR #8516, opened 2025-03-11
webgpu: enable unpacking WGSL code in filamat

rqreyes:
in filamat, enable unpacking WGSL code

BUGS = [397672412]

### DIFF
--- a/libs/filamat/src/MaterialBuilder.cpp
+++ b/libs/filamat/src/MaterialBuilder.cpp
@@ -1049,7 +1049,6 @@ bool MaterialBuilder::generateShaders(JobSystem& jobSystem, const std::vector<Va
 
 
                 switch (targetApi) {
-                    // TODO: Handle webgpu here
                     case TargetApi::WEBGPU:
                         assert(!spirv.empty());
                         assert(wgsl.length() > 0);
--- a/tools/matinfo/src/main.cpp
+++ b/tools/matinfo/src/main.cpp
@@ -50,10 +50,12 @@ struct Config {
     bool printESSL1 = false;
     bool printSPIRV = false;
     bool printMetal = false;
+    bool printWGSL = false;
     bool printDictionaryGLSL = false;
     bool printDictionaryESSL1 = false;
     bool printDictionarySPIRV = false;
     bool printDictionaryMetal = false;
+    bool printDictionaryWGSL = false;
     bool transpile = false;
     bool binary = false;
     bool analyze = false;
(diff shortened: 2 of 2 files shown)


tools/matinfo/src/main.cpp
```diff
@@ -50,10 +50,12 @@ struct Config {
     bool printESSL1 = false;
     bool printSPIRV = false;
     bool printMetal = false;
+    bool printWGSL = false;
     bool printDictionaryGLSL = false;
     bool printDictionaryESSL1 = false;
     bool printDictionarySPIRV = false;
     bool printDictionaryMetal = false;
+    bool printDictionaryWGSL = false;
```
bridgewaterrobbie, line 58: Mentioned in prior meeting, but "DictionaryWGSL" may not be relevant

filament/backend/include/backend/DriverEnums.h
```diff
@@ -196,6 +197,11 @@ static constexpr const char* shaderLanguageToString(ShaderLanguage shaderLanguag
             return "MSL";
         case ShaderLanguage::METAL_LIBRARY:
             return "Metal precompiled library";
+        case ShaderLanguage::WGSL:
+            return "WebGPU";
+        // TODO Should we add this?
```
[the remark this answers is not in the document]
rqreyes: https://github.com/google/filament/pull/8494#discussion_r1987593890

looks like according to powei, we should not have defaults in general and let the compiler catch missing cases

this TODO comment will be removed

tools/matinfo/src/main.cpp
rqreyes: it looks like `printDictionaryWGSL` does produce output, but i don't understand its relevance. can i get more context on what dictionary is and why all the other shader languages have dictionaries?

libs/filamat/src/MaterialBuilder.cpp
```diff
@@ -1038,6 +1051,12 @@ bool MaterialBuilder::generateShaders(JobSystem& jobSystem, const std::vector<Va
                 switch (targetApi) {
                     // TODO: Handle webgpu here
```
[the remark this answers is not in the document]
rqreyes: yup! will remove

tools/matinfo/src/main.cpp
rqreyes: looks like **print dictionary** displays all of the shaders for a target language per material, so i'll keep `printDictionaryWGSL` in

jc3265 (approved):

tools/matinfo/src/main.cpp
```diff
@@ -568,18 +582,45 @@ static bool parseChunks(Config config, void* data, size_t size) {
...
+            std::cout << (const char*) content.data();
+
             return true;
         }
     }
 
     TextWriter writer;
 
-    if (config.printDictionaryGLSL || config.printDictionaryESSL1 || config.printDictionarySPIRV || config.printDictionaryMetal) {
+    if (config.printDictionaryGLSL || config.printDictionaryESSL1 || config.printDictionarySPIRV || config.printDictionaryMetal || config.printDictionaryWGSL) {
```
poweifeng, line 617: make sure line width <= 100 chars

rqreyes: fixed! thanks

poweifeng (approved):

PR #8518, opened 2025-03-11
Add helper for tests to expect image results.

mhoff12358:
Currently still based around using image hashes but will be improved later.

BUGS=[402474472]


mhoff12358:
@bejado I haven't gone through and updated the style comments from the last PR for this one, and there's an error with gmock on windows, so it's still a draft. But PTAL at the actual EXPECT_IMAGE calls in test_Blit.cpp and lemme know if you have any thoughts.

I'm considering whether it's worth bundling the `flushAndWait();` into `ImageExpectations`'s destructor.

Also, how do you feel about the technically redundant {} wrapping the `ImageExpectations` scope given that it's at the end of the test. I tend to like it because it makes it clearer that there's an object whose destruction matters. And it means that if a test needs two you don't wind up with some in enclosed scopes and some not.

bejado:
> But PTAL at the actual EXPECT_IMAGE calls in test_Blit.cpp and lemme know if you have any thoughts.

LGTM! I’m guessing the pixel hash argument passed to `ScreenShotParams` is just to bridge over the current behavior for now, yes? I imagine once we have golden image comparison, the hash will not be needed—instead `EXPECT_IMAGE` will do the golden image comparison instead.

> I'm considering whether it's worth bundling the flushAndWait(); into ImageExpectations's destructor.

Hmm, I don’t have a strong opinion on this, but I think it makes sense to keep it separate. The `flushAndWait` is really only needed once at the end of a test. We could have multiple `EXPECT_IMAGE` checks but we only need to call `flushAndWait` once.

> Also, how do you feel about the technically redundant {} wrapping the ImageExpectations scope given that it's at the end of the test.

Yes, I’m also a fan of making the destruction explicit.

Let me know once the style issues and gmock is addressed, and I'll take a closer look. But everything LGTM so far.

mhoff12358:
@bejado Alright, this should be good to look at.

bejado (approved):
Just one nit: please keep line lengths <100. Otherwise LGTM.

PR #8521, opened 2025-03-13
update createDriver api to remove const weirdness

sidreesshah:


sidreesshah:
> is this addressing a real problem?

Yes, it is a concern imo. const here is a `const pointer to non const data` rather than `pointer to a const data`. At best, it could be ignored, as it does not do anything useful. At worst, it can give someone, who is looking at definition to misjudge it as data being const. Why would we want different signatures for declaration and definition?

jc3265 (approved):

Issue #8515, opened 2025-03-11
on linux, if both GCC and Clang are installed, build fails because CMake picks GCC


dneto0:
⚠️ **Issues not using this template will be systematically closed.**

**Describe the bug**
On Linux a default build will fail because CMake selects GCC instead of Clang

**To Reproduce**
1. On linux, where both GCC and Clang are installed, run `./build.sh release`

**Expected behavior**
When clang is available, Filament should be built.

**Screenshots**

**Logs**
If applicable, copy **full** logs from your console here. Please *do not*
use screenshots of logs, copy them as text, use gist or attach an *uncompressed* file.

```
$ ./build.sh release
Building release in out/cmake-release...
-- The C compiler identification is GNU 14.2.0
-- The CXX compiler identification is GNU 14.2.0
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /usr/bin/cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /usr/bin/c++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
CMake Error at CMakeLists.txt:286 (message):
  Detected C compiler GNU is unsupported


-- Configuring incomplete, errors occurred!
```

**Desktop (please complete the following information):**
 - OS: Linux
 - GPU: n/a
 - Backend: n/a

**Smartphone (please complete the following information):**
 - Device: n/a
 - OS: n/a

**Additional context**

dneto0:
I made a fresh docker image based on Ubuntu Noble 24.04, installed cmake, ninja-build, gcc, and clang.

Then specifying the CC and CXX and CXXFLAGS as suggested also fails:


```shell
# CC=/usr/bin/clang CXX=/usr/bin/clang++ CXXFLAGS=-stdlib=libc++ ./build.sh -c release
Cleaning build directories...
Building release in out/cmake-release...
-- The C compiler identification is Clang 18.1.3
-- The CXX compiler identification is Clang 18.1.3
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /usr/bin/clang - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - failed
[...]

PR #8522, opened 2025-03-13
build.sh: On Linux force use of latest Clang

dneto0:
CMake will default to GCC if it's available.
Force use of the latest Clang instead.
This can be overridden with CC, CXX, and CXXFLAGS

Fixed: #8515


build.sh
```diff
@@ -224,6 +224,17 @@ BUILD_CUSTOM_TARGETS=
 UNAME=$(uname)
 LC_UNAME=$(echo "${UNAME}" | tr '[:upper:]' '[:lower:]')
 
+if [[ "$LC_UNAME" == "linux" ]]; then
+    # On Linux both GCC and Clang are commonly available, but CMake
```
poweifeng, line 228: maybe exiting when clang is not available would be helpful as well.  Otherwise, we'd get a cmake error that's less scrutable (?).

poweifeng: actually, looks like this will break our github workflows since we're not using the latest clang version on system but a fixed one.

dneto0: I'm still struggling to build anyway.    E.g. the SDL2 build isn't picking up stdint.h typedefs, and I am not chasing everything down.  So this is getting opportunistic for me.

poweifeng:
closing this for clean-up.  I think there is something we can do better here with the build script. We'll keep #8515 open to address the issues.

PR #8523, opened 2025-03-13
APIs to query the texture limit sizes.

pixelflinger:
Assert textures dimensions are in range when creating a Texture object.

FIXES=[398901038]


pixelflinger:
@poweifeng @bejado I didn't do the vk/metal side of this (just hardcoded), FYI.

poweifeng (approved):

PR #8526, opened 2025-03-14
Generate 1D LUTs for color grading when possible

elizagamedev:
This introduces two new methods to `ToneMapper`: `isOneDimensional()` and `isLDR()`. `ColorGrading` references these values along with other parameters passed to the builder to determine if we can get away with only generating a one-dimensional LUT. Meanwhile, `PostProcessManager` takes care of setting the new spec constants and uniforms for `colorGrading.mat` and `colorGradingAsSubpass.mat`.

FIXES=[372037228]

### DIFF
--- a/filament/src/PostProcessManager.cpp
+++ b/filament/src/PostProcessManager.cpp
@@ -2649,7 +2649,7 @@ void PostProcessManager::configureTemporalAntiAliasingMaterial(
     setConstantParameter(ma, "varianceGamma", taaOptions.varianceGamma);
     if (dirty) {
         ma->invalidate();
-        // TODO: call Material::compile(), we can't si that now because it works only
+        // TODO: call Material::compile(), we can't do that now because it works only
         //       with surface materials
     }
 }
@@ -2676,7 +2676,7 @@ FMaterialInstance* PostProcessManager::configureColorGradingMaterial(
 
     if (dirty) {
         ma->invalidate();
-        // TODO: call Material::compile(), we can't si that now because it works only
+        // TODO: call Material::compile(), we can't do that now because it works only
         //       with surface materials
     }
 
@@ -2693,13 +2693,8 @@ FMaterialInstance* PostProcessManager::configureColorGradingMaterial(
             .wrapR = SamplerWrapMode::CLAMP_TO_EDGE,
             .anisotropyLog2 = 0
     };
-    if (colorGrading->isOneDimensional()) {
-        mi->setParameter("lut1d", colorGrading->getHwHandle(), params);
-        mi->setParameter("lut", mEngine.getZeroTexture(), params);
-    } else {
-        mi->setParameter("lut", colorGrading->getHwHandle(), params);
-        mi->setParameter("lut1d", mEngine.getZeroTexture(), params);
-    }
+
+    mi->setParameter("lut", colorGrading->getHwHandle(), params);
 
     const float lutDimension = float(colorGrading->getDimension());
     mi->setParameter("lutSize", float2{
--- a/filament/src/PostProcessManager.cpp
+++ b/filament/src/PostProcessManager.cpp
@@ -2306,36 +2306,11 @@ void PostProcessManager::colorGradingPrepareSubpass(DriverApi& driver,
         const FColorGrading* colorGrading, ColorGradingConfig const& colorGradingConfig,
         VignetteOptions const& vignetteOptions, uint32_t const width, uint32_t const height) noexcept {
 
-    float4 const vignetteParameters = getVignetteParameters(vignetteOptions, width, height);
-
-    auto const& material = getPostProcessMaterial("colorGradingAsSubpass");
-    FMaterialInstance* const mi = PostProcessMaterial::getMaterialInstance(mEngine, material);
-
-    mi->setParameter("lut", colorGrading->getHwHandle(), {
-            .filterMag = SamplerMagFilter::LINEAR,
-            .filterMin = SamplerMinFilter::LINEAR,
-            .wrapS = SamplerWrapMode::CLAMP_TO_EDGE,
-            .wrapT = SamplerWrapMode::CLAMP_TO_EDGE,
-            .wrapR = SamplerWrapMode::CLAMP_TO_EDGE,
-            .anisotropyLog2 = 0
-    });
-    const float lutDimension = float(colorGrading->getDimension());
-    mi->setParameter("lutSize", float2{
-        0.5f / lutDimension, (lutDimension - 1.0f) / lutDimension,
-    });
-
-    const float temporalNoise = mUniformDistribution(mEngine.getRandomEngine());
-
-    mi->setParameter("vignette", vignetteParameters);
-    mi->setParameter("vignetteColor", vignetteOptions.color);
-    mi->setParameter("dithering", colorGradingConfig.dithering);
-    mi->setParameter("outputLuminance", colorGradingConfig.outputLuminance);
-    mi->setParameter("temporalNoise", temporalNoise);
+    auto& material = getPostProcessMaterial("colorGradingAsSubpass");
+    FMaterialInstance* const mi =
+            configureColorGradingMaterial(material, colorGrading, colorGradingConfig,
+                    vignetteOptions, width, height);
     mi->commit(driver);
-
-    // load both variants
-    material.getMaterial(mEngine, PostProcessVariant::OPAQUE);
-    material.getMaterial(mEngine, PostProcessVariant::TRANSLUCENT);
 }
 
 void PostProcessManager::colorGradingSubpass(DriverApi& driver,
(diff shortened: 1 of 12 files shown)


filament/src/PostProcessManager.cpp
```diff
@@ -2697,6 +2654,70 @@ void PostProcessManager::configureTemporalAntiAliasingMaterial(
...
+            }
+        }
+    };
+
+    setConstantParameter(ma, "isOneDimensional", colorGrading->isOneDimensional());
+    setConstantParameter(ma, "isLDR", colorGrading->isLDR());
+
+    if (dirty) {
+        ma->invalidate();
+        // TODO: call Material::compile(), we can't si that now because it works only
```
pixelflinger, line 2679: "can't si" ?

filament/src/materials/colorGrading/colorGradingAsSubpass.mat
```diff
@@ -6,6 +6,11 @@ material {
             name : lut,
             precision: medium
         },
+        {
+            type : sampler2d,
+            name : lut1d,
+            precision: medium
+        },
```
pixelflinger, line 13: I would prefer that you still use a sampler3d (so we don't have to have two samplers), just create a Nx1x1 texture. Wouldn't that make things simpler?

filament/src/details/ColorGrading.cpp
```diff
@@ -581,8 +580,11 @@ static float3 luminanceScaling(float3 x,
 // Quality
 //------------------------------------------------------------------------------
 
-static std::tuple<TextureFormat, PixelDataFormat, PixelDataType>
-        selectLutTextureParams(ColorGrading::LutFormat const lutFormat) noexcept {
+static std::tuple<TextureFormat, PixelDataFormat, PixelDataType> selectLutTextureParams(
+        ColorGrading::LutFormat const lutFormat, const bool isOneDimensional) noexcept {
+    if (isOneDimensional) {
+        return { TextureFormat::R8, PixelDataFormat::R, PixelDataType::UBYTE };
```
pixelflinger, line 586: Why not use a fp16 texture? 8 bits is generally not enough to encode the sRGB<>Linear transform.

filament/src/details/ColorGrading.cpp
```diff
@@ -658,27 +660,28 @@ FColorGrading::FColorGrading(FEngine& engine, const Builder& builder) {
...
-    {
-        std::lock_guard<Mutex> const lock(configLock);
-        c.lutDimension          = builder->dimension;
-        c.adaptationTransform   = adaptationTransform(builder->whiteBalance);
-        c.colorGradingIn        = selectColorGradingTransformIn(builder->toneMapping);
-        c.colorGradingOut       = selectColorGradingTransformOut(builder->toneMapping);
-        c.colorGradingLuminance = selectColorGradingLuminance(builder->toneMapping);
-        c.oetf                  = selectOETF(builder->outputColorSpace);
-    }
+    const Config config = {
```
pixelflinger, line 663: so I see you removed the shenanigan with the Mutex. I agree with you that it shouldn't be needed. But I think ASAN was complaining and/or we ran into problems, and we never understood what it was. So for now, I would suggest to keep the code the way it was and we can make that specific change in a separate PR.

filament/src/details/ColorGrading.cpp
```diff
@@ -658,27 +660,28 @@ FColorGrading::FColorGrading(FEngine& engine, const Builder& builder) {
...
+            adaptationTransform(builder->whiteBalance),
+            selectColorGradingTransformIn(builder->toneMapping),
+            selectColorGradingTransformOut(builder->toneMapping),
+            selectColorGradingLuminance(builder->toneMapping),
+            selectOETF(builder->outputColorSpace),
+    };
 
-    mDimension = c.lutDimension;
+    mDimension = config.lutDimension;
+    mIsOneDimensional = !builder->hasAdjustments && !builder->luminanceScaling
```
pixelflinger, line 673: maybe add a comment here that this is true because we don't handle the case where the input and output spaces are different. but when we do, we'll have to check that too.

filament/src/details/ColorGrading.cpp
```diff
@@ -658,27 +660,28 @@ FColorGrading::FColorGrading(FEngine& engine, const Builder& builder) {
...
+            && builder->toneMapper->isOneDimensional();
+    mIsLDR = mIsOneDimensional && builder->toneMapper->isLDR();
 
-    size_t lutElementCount = c.lutDimension * c.lutDimension * c.lutDimension;
-    size_t elementSize = sizeof(half4);
+    size_t lutElementCount = mIsOneDimensional
+            ? config.lutDimension
```
pixelflinger, line 678: should we just use a hardcoded size lut of the 1D case? I feel that the defaults 16, 32, 64 are way too low for the 1D case. we can easily do 256 or 512 for max quality.

filament/src/details/ColorGrading.cpp
```diff
@@ -658,27 +660,28 @@ FColorGrading::FColorGrading(FEngine& engine, const Builder& builder) {
...
+            : (config.lutDimension * config.lutDimension * config.lutDimension);
+    size_t elementSize = mIsOneDimensional ? sizeof(uint8_t) : sizeof(half4);
```
pixelflinger, line 680: fp16 instead of u8?

filament/src/details/ColorGrading.cpp
```diff
@@ -688,170 +691,206 @@ FColorGrading::FColorGrading(FEngine& engine, const Builder& builder) {
...
 
-                    *p++ = half4{v, 0.0f};
-                }
+    //auto now = std::chrono::steady_clock::now();
+
+    if (mIsOneDimensional) {
+        uint8_t* UTILS_RESTRICT p = (uint8_t*) data;
+        for (size_t rgb = 0; rgb < config.lutDimension; rgb++) {
+            float s;
+            if (mIsLDR) {
```
pixelflinger, line 790: move this out of the loop, or even just do two loops?

filament/src/details/ColorGrading.cpp
```diff
@@ -688,170 +691,206 @@ FColorGrading::FColorGrading(FEngine& engine, const Builder& builder) {
...
+                float3 v = float3(rgb) * (1.0f / float(config.lutDimension - 1u));
+
+                v = (*builder->toneMapper)(float3(v));
+
+                // TODO: We should convert to the output color space if we use a working
+                //       color space that's not sRGB
+                // TODO: Allow the user to customize the output color space
+
```
pixelflinger, line 798: I think these comments are not valid in the 1D case

filament/src/details/ColorGrading.cpp
```diff
@@ -688,170 +691,206 @@ FColorGrading::FColorGrading(FEngine& engine, const Builder& builder) {
...
+                v = saturate(v);
+
+                // Apply OETF
+                v = config.oetf(v);
+
+                s = v.r;
+            } else {
+                s = hdrColorAt(rgb, rgb, rgb).r;
             }
+            *p++ = uint8_t(std::floor(s * 255.0f + 0.5f));
```
pixelflinger, line 809: fp16?

filament/src/details/ColorGrading.cpp
```diff
@@ -688,170 +691,206 @@ FColorGrading::FColorGrading(FEngine& engine, const Builder& builder) {
...
+        }
+    } else {
+        // Multithreadedly generate the tone mapping 3D look-up table using 32 jobs
+        // Slices are 8 KiB (128 cache lines) apart.
+        // This takes about 3-6ms on Android in Release
+        JobSystem& js = engine.getJobSystem();
+        auto *slices = js.createJob();
+        for (size_t b = 0; b < config.lutDimension; b++) {
+            auto* job = js.createJob(slices,
+                    [data, converted, b, &config, builder, &hdrColorAt](
```
pixelflinger, line 819: nice!

pixelflinger (changes requested):

filament/src/PostProcessManager.cpp
elizagamedev: I copied this from the above comment, but fixed

filament/src/details/ColorGrading.cpp
elizagamedev: Done.

filament/src/details/ColorGrading.cpp
elizagamedev: Is it not resolvable by having config be declared const? The issue seemed to me that ASAN thought multiple threads might be writing to ColorGrading.cpp. I guess it's not that simple...?

I'll ask offline more about this.

filament/src/details/ColorGrading.cpp
elizagamedev: Done.

filament/src/details/ColorGrading.cpp
elizagamedev: Done.

filament/src/details/ColorGrading.cpp
elizagamedev: Done.

filament/src/details/ColorGrading.cpp
elizagamedev: I was thinking the same thing after I went home for the day

filament/src/details/ColorGrading.cpp
elizagamedev: Done.

filament/src/details/ColorGrading.cpp
elizagamedev: Done.

filament/src/details/ColorGrading.cpp
elizagamedev: :)

pixelflinger (approved):

pixelflinger:
@elizagamedev make sure the PR comment has the FIXES=[]  line.

bejado (approved):

elizagamedev:
Seeing some scuba failures on my test CL so disabling merge for now. Will investigate tomorrow.

PR #8527, opened 2025-03-14
Adjust matc targets and flags for WebGPU

bridgewaterrobbie:
We may need to improve the modularity of `matc` invocations as a custom tool, as this logic fully changes how matc works if WebGPU is enabled, even if someone theoretically wants to test other targets. Until we are enabled by default though that is a minor problem

### DIFF
--- a/libs/filamat/src/GLSLPostProcessor.cpp
+++ b/libs/filamat/src/GLSLPostProcessor.cpp
@@ -518,8 +518,13 @@ void GLSLPostProcessor::spirvToMsl(const SpirvBlob* spirv, std::string* outMsl,
     }
 }
 
-bool GLSLPostProcessor::spirvToWgsl(const SpirvBlob *spirv, std::string *outWsl) {
+bool GLSLPostProcessor::spirvToWgsl(SpirvBlob *spirv, std::string *outWsl) {
 #if FILAMENT_SUPPORTS_WEBGPU
+    // We need to remove dead code for our variant-filter workaround for WebGPU to work
+    // This is especially relevant for removing the push constants that morphing uses when not disabled
+    spv::spirvbin_t remapper(0);
+    remapper.remap(*spirv, spv::spirvbin_base_t::DCE_ALL);
+
     //Currently no options we want to use
     const tint::spirv::reader::Options readerOpts{};
     tint::wgsl::writer::Options writerOpts{};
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -554,6 +554,10 @@ if (CMAKE_BUILD_TYPE MATCHES Release)
 else()
     option(FILAMENT_DISABLE_MATOPT "Disable material optimizations" ON)
 endif()
+#Removing optimizations results in push constants not being optimized away, causing tint errors.
+if (FILAMENT_SUPPORTS_WEBGPU)
+    set(FILAMENT_DISABLE_MATOPT OFF)
+endif()
```
poweifeng, line 560: instead of this, would it be possible to just do a single pass of dead code elimination (DCE) for webgpu, regardless of opt on/off?  Would that solve the issue?

bridgewaterrobbie: Yes, that does seem to work. I'll upload changes running the necessary optimizations specifically within the WGSL generation if that's fine (Which risks minor redundancy in the full case, but shouldn't be significant I think)

bridgewaterrobbie: Adjusted, thanks for the suggestion! I'd been looking for something like that, but the acronym "DCE" helped me find what I needed.

libs/filamat/src/GLSLPostProcessor.cpp
```diff
@@ -518,8 +518,12 @@ void GLSLPostProcessor::spirvToMsl(const SpirvBlob* spirv, std::string* outMsl,
     }
 }
 
-bool GLSLPostProcessor::spirvToWgsl(const SpirvBlob *spirv, std::string *outWsl) {
+bool GLSLPostProcessor::spirvToWgsl(SpirvBlob *spirv, std::string *outWsl) {
 #if FILAMENT_SUPPORTS_WEBGPU
+    // We need to remove dead code for our variant-filter workaround for WebGPU to work
+    spv::spirvbin_t remapper(0);
```
poweifeng, line 524: would you mind adding a note here about eliminating push constants because it's not supported yet?

bridgewaterrobbie: Sure, I'll elaborate in the comment.

bridgewaterrobbie: Added another line so it reads 
```
// We need to remove dead code for our variant-filter workaround for WebGPU to work
    // This is especially relevant for removing the push constants that morphing uses when not disabled
```

poweifeng (approved):

PR #8534, opened 2025-03-17
vk: disable transient texture for depth resolve

poweifeng:
Depth resolve is handled via a custom shader and therefore cannot use the transient texture path.

This is fixes the black screen bug when MSAA is turned on and when using the latest MoltenVK (1.2.11).


filament/backend/src/vulkan/VulkanTexture.cpp
```diff
@@ -249,12 +249,17 @@ VulkanTexture::VulkanTexture(VkDevice device, VkPhysicalDevice physicalDevice,
...
+            // Lazily allocated memory is available.
+            context.isLazilyAllocatedMemorySupported() &&
+            // Usage consists of attachment flags only.
+            none(tusage & ~TextureUsage::ALL_ATTACHMENTS) &&
+            // Usage contains at least one attachment flag.
+            any(tusage & TextureUsage::ALL_ATTACHMENTS) &&
+            // Depth resolve cannot use transient attachment because it uses a custom shader.
+            // TODO: see VulkanDriver::isDepthStencilResolveSupported() to know when to remove this
+            // restricion.
+            !(any(usage & TextureUsage::DEPTH_ATTACHMENT) && samples > 1);
```
z3moon, line 261: nit: do we want to check STENCIL_ATTACHMENT as well?

z3moon (approved):

filament/backend/src/vulkan/VulkanTexture.cpp
poweifeng: that's a good point.  So stencil resolve is not really covered by the shader.  I think if there's a multi-sample stencil buffer, then we'll definitely hit a bug with no good resolution for now. I added a comment about this, and how to address it for vk.

PR #8538, opened 2025-03-18
Refactor Vulkan android into its own class

rafadevai:


poweifeng (approved):
good stuff.  thank you!

PR #8546, opened 2025-03-19
External sampler frontend

phoenixxxx:
Looking at option 1 for ext sampler implementation.

This one makes the user set everything explicitely ahead of material construction time.


filament/src/ds/DescriptorSet.cpp
```diff
@@ -90,7 +90,16 @@ void DescriptorSet::commitSlow(DescriptorSetLayout const& layout,
             driver.updateDescriptorSetTexture(dsh, binding,
                     descriptors[binding].texture.th,
                     descriptors[binding].texture.params);
-        } else {
+        } else if (layout.isSamplerExternal(binding)) {
+            auto data = layout.getConstantSamplerData(binding);
+
+            // Not sure what to do here.... Should we validate or should we
```
phoenixxxx, line 96: @pixelflinger any thoughts on this? Basically at this point we have all the data in the layout, and everything was created ahea of time. 
So we do not requite "internalFormat" "ChromaConversion" and "SamplerStates" but we are able to validate at this point that they match the constant states that were set a creation time

poweifeng:
closing because no longer required

PR #8547, opened 2025-03-19
make createExternalImage UTILS_PUBLIC across platforms for future proofing

alizahlalani:
createExternalImagedepends on the user calling into this function directly - by adding UTILS_PUBLIC, we future proof setExternalImage functionality for clients who don't have innate awareness of Filament


pixelflinger (approved):

pixelflinger:
@alizahlalani looks like there is a conflict, could you resolve it?

alizahlalani:
> @alizahlalani looks like there is a conflict, could you resolve it?

done - @rafadevai will make the changes for Vulkan with https://github.com/google/filament/pull/8551

PR #8563, opened 2025-03-25
Dispose debug.server/debug.fgViewerServer when Engine destroyed 

nmfisher:
The material/framegraph debug servers currently aren't destroyed when an Engine is destroyed. This causes a panic when an Engine is created a second time (e.g. when hot restarting via Flutter), because the port is already taken.


show50726:
Thanks for catching this!

filament/src/details/Engine.cpp
```diff
@@ -508,6 +508,17 @@ FEngine::~FEngine() noexcept {
     if (mOwnPlatform) {
         PlatformFactory::destroy(&mPlatform);
     }
+
```
poweifeng, line 511: Thanks for doing this!

For clarity, I think it might be better to move this destruction to the `loop()` method after the `while(true)`

show50726: If we do this, would it cause a leak when deleting the engine directly? (Though I think it is rarely happened)

poweifeng: if `Engine::destroy()` is not called for destroying the engine, then we'll be leaking other things, so that's an incorrect API usage.

nmfisher: > Thanks for doing this!
> 
> For clarity, I think it might be better to move this destruction to the `loop()` method after the `while(true)`

Updated to move to `loop()`.

PR #8569, opened 2025-03-25
Fix test breaking changes with workarounds

poweifeng:
- Some tests require textures larger than 2048. Change the minimum to 4096 for now, and file proper bugs for tests.
- The introduction of uint8_t to DescriptorSetLayoutBinding caused unintended padding with using this struct as a hash key. We comment out the relevent code for now.


poweifeng:
Bypassing merge requirements for urgency.

PR #8571, opened 2025-03-26
[webgpu] Create WebGPUSurface Class

jneljneljnel:
Class for a WebGPU Surafce to configure a created surface.

### DIFF
--- a/filament/backend/src/webgpu/WebGPUDriver.cpp
+++ b/filament/backend/src/webgpu/WebGPUDriver.cpp
@@ -35,6 +35,7 @@
 #include <algorithm>
 #include <cstddef>
 #include <cstdint>
+#include <memory>
 #include <sstream>
 #include <string_view>
 #include <utility>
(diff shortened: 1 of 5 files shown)


filament/backend/src/webgpu/WebGPUSwapChain.cpp
```diff
@@ -0,0 +1,279 @@
...
+        printSurfaceCapabilitiesDetails(capabilities);
+#endif
+    }
+    const bool useSRGBColorSpace = (flags & SWAP_CHAIN_CONFIG_SRGB_COLORSPACE) != 0;
+    initConfig(mConfig, device, capabilities, useSRGBColorSpace);
+}
+
+WebGPUSwapChain::~WebGPUSwapChain() {
+    if (mConfigured) {
+        mSurface.Unconfigure();
```
poweifeng, line 225: would it crash if it was unconfigured?

jneljneljnel: No crash, but it gives an Error: [Surface "metal_surface"] is not configured.

poweifeng: I see. I'm wondering this can be inferred from the state of mConfig.  But as is, this is ok.

poweifeng (commented):
looks like you're going through revisions.  Feel free to let me know when I should look again.  Thanks.

filament/backend/src/webgpu/WebGPUDriver.cpp
```diff
@@ -342,6 +305,13 @@ void WebGPUDriver::destroyRenderTarget(Handle<HwRenderTarget> rth) {
 }
 
 void WebGPUDriver::destroySwapChain(Handle<HwSwapChain> sch) {
+    mSwapChain = nullptr;
+    // TODO:  use webgpu handle allocator from
+    //        https://github.com/google/filament/pull/8566
+//    if (sch) {
+//        HwSwapChain* hwSwapChain = handleCast<HwSwapChain*>(sch);
+//        destruct(sch, hwSwapChain);
+//    }
```
poweifeng, line 314: I think the allocator pr has landed, are you going to use it?

poweifeng (approved):

PR #8575, opened 2025-03-27
fix SGSR upscaling when a subpass is used

pixelflinger:
SGSR upscaling didn't take the viewport into account.

### DIFF
--- a/filament/src/materials/sgsr/sgsr1.mat
+++ b/filament/src/materials/sgsr/sgsr1.mat
@@ -10,6 +10,11 @@ material {
             type : float4,
             name : viewport,
             precision: high
+        },
+        {
+            type : float4,
+            name : viewportInfo,
+            precision: high
         }
     ],
     variables : [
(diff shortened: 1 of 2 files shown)


filament/src/materials/sgsr/sgsr1.mat
```diff
@@ -22,8 +27,8 @@ material {
 
 vertex {
     void postProcessVertex(inout PostProcessVertexInputs postProcess) {
-        postProcess.normalizedUV = uvToRenderTargetUV(postProcess.normalizedUV);
```
poweifeng, line 25: I don't know enough to say for sure, but just want to check that this is no longer necessary?

I see the same transform in many other material files.

pixelflinger: yeah I double checked, this is correct.  The postfx framework calculates` normalizedUV` for us as a convenience. Then we use it to calculate our variable, and in most case we need the `uvToRenderTargetUV` call. In the previous code we were using `normalizedUV` as a temporary, but this variable is only for our benefit.

PR #8576, opened 2025-03-27
Fix INVALID_OPERATION error for WebGL

z3moon:
Mozilla sometimes clears the generic buffer binding when another buffer is destroyed, if that other buffer is bound on an indexed buffer binding.

The fix makes sure the generic buffer binding is unbound from the context when being destroyed.

BUGS=[376125497]

### DIFF
--- a/filament/backend/src/opengl/OpenGLContext.cpp
+++ b/filament/backend/src/opengl/OpenGLContext.cpp
@@ -570,22 +570,13 @@ void OpenGLContext::initBugs(Bugs* bugs, Extensions const& exts,
             bugs->disable_invalidate_framebuffer = true;
         }
 
-        if (strstr(vendor, "Mesa") ||
-            strstr(vendor, "Mozilla")) {
+        if (strstr(vendor, "Mesa")) {
             // Seen on
             //  [Mesa],
             //  [llvmpipe (LLVM 17.0.6, 256 bits)],
             //  [4.5 (Core Profile) Mesa 24.0.6-1],
             //  [4.50]
-            // or
-            //  [Mozilla],
-            //  [GeForce GTX 980, or similar],
-            //  [OpenGL ES 3.0 (WebGL 2.0)],
-            //  [OpenGL ES GLSL ES 3.00 (WebGL GLSL ES 3.00)]
-            // not known which version are affected.
-            // For Mozilla, the issue seems to be observed regardless of which
-            // renderer it comes with.
-
+            // not known which version are affected
             bugs->rebind_buffer_after_deletion = true;
         }
     } else {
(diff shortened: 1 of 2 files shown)


filament/backend/src/opengl/OpenGLContext.cpp
```diff
@@ -566,16 +566,27 @@ void OpenGLContext::initBugs(Bugs* bugs, Extensions const& exts,
         } else if (strstr(renderer, "AMD") ||
                    strstr(renderer, "ATI")) {
             // AMD/ATI GPU
-        } else if (strstr(vendor, "Mesa")) {
+        } else if (strstr(renderer, "Mozilla")) {
+            bugs->disable_invalidate_framebuffer = true;
+        }
+
+        if (strstr(vendor, "Mesa") ||
+            strstr(vendor, "Mozilla")) {
```
poweifeng, line 574: Looks like the convention is use GPU vendors/drivers as the top-level category.

Why not just follow that? (Instead of having two GPUs point to the same logic).

z3moon: The existing logic was already under the if/else statement change, and I found the previous check `} else if (strstr(vendor, "Mesa")) {` is not hit if the any prior checks are met. So, in this change, I pull them out of the existing if/else chain, which is supposed to only check renderer string, and made it as a discrete if/else chain.

However, I found this didn't work at home anyway, where my FireFox uses ANGLE.
```
[ANGLE (NVIDIA, NVIDIA GeForce GTX 980 Direct3D11 vs_5_0 ps_5_0), or similar]
```
So I made a followup commit that actually pulls the new check as the top-level category. Please see the second commit.

I don't know if I should keep the existing check for Mesa to be still under the `isAngle == true` condition. What do you think?

filament/backend/src/opengl/OpenGLContext.cpp
```diff
@@ -587,6 +589,19 @@ void OpenGLContext::initBugs(Bugs* bugs, Extensions const& exts,
         }
     }
 
+    if (strstr(vendor, "Mozilla")) {
+        // Seen on
+        //  [Mozilla],
+        //  [GeForce GTX 980, or similar]
+        //    or [ANGLE (NVIDIA, NVIDIA GeForce GTX 980 Direct3D11 vs_5_0 ps_5_0), or similar]
```
poweifeng, line 596: So Firefox is using Angle on windows, and you're still seeing the bug without this workaround?

poweifeng: In fact, maybe we should only do this for the web build?

z3moon: Yeah I see the bug without it on my windows. I think we may as well make it universally as there's another check for Mozilla.

poweifeng (approved):
lgtm. but I think @pixelflinger should take a look before you commit.

pixelflinger (approved):

PR #8584, opened 2025-03-28
webgpu: select color format, present mode, and composite alpha mode from surface capabilities

AndyHovingh:


jneljneljnel (commented):
lgtm!

PR #8592, opened 2025-04-03
Fix Windows webGPU build

kpiascik:


filament/backend/CMakeLists.txt
```diff
@@ -306,6 +306,11 @@ target_include_directories(${TARGET} PUBLIC ${PUBLIC_HDR_DIR})
 # add this subproject to the Filament folder
 set_target_properties(${TARGET} PROPERTIES FOLDER Filament)
 
+# we need to export the headers properly for backend.lib on windows
+if (WIN32 AND FILAMENT_SUPPORTS_WEBGPU)
```
bridgewaterrobbie, line 310: I'd wonder if we actually want a windows guard on this. It wouldn't surprise me if we find this to be necessary on Android.

I don't see in the Dawn codebase exactly what this is expected to do.

bridgewaterrobbie (approved):

filament/backend/CMakeLists.txt
kpiascik: Doesn't WIN32 already guard this for Windows only?

bridgewaterrobbie: @kpiascik yes, and I’m questioning if maybe we want this in general. I’m not fully clear on what exactly this does to make windows behave though.

kpiascik: @bridgewaterrobbie there's a generated header `webgpu.h` that's part of the dawn generated headers.  It has the following `#define` block at the top:
```
#if defined(WGPU_SHARED_LIBRARY)
#    if defined(_WIN32)
#        if defined(WGPU_IMPLEMENTATION)
#            define WGPU_EXPORT __declspec(dllexport)
#        else
#            define WGPU_EXPORT __declspec(dllimport)
#        endif
#    else  // defined(_WIN32)
#        if defined(WGPU_IMPLEMENTATION)
#            define WGPU_EXPORT __attribute__((visibility("default")))
#        else
#            define WGPU_EXPORT
#        endif
#    endif  // defined(_WIN32)
#else       // defined(WGPU_SHARED_LIBRARY)
#    define WGPU_EXPORT
#endif  // defined(WGPU_SHARED_LIBRARY)
```
There is a link error when we try to link backend.lib (filament's backend) and it fails to resolve since the compile flag for these `cpp` files doesn't have `WGPU_IMPLEMENTATION` defined.

This is very much Windows only problem since we have no such problem on Linux or Mac.  If we discover there are issues with this approach where other platforms are impacted then we can revisit.

bridgewaterrobbie: Thanks!

I think it is fine as is (And approved) though I might soft-vote to still set the flag independently of Windows. If this is related to our intention to use it as a static library, it is possible that it will be appropriate out of windows in the future depending on Dawn changes. Our intention to use htis as a static library is unlikely to change.

There isn't much of a reason it can't wait until the need arises, other than needing to remember this issue when the time comes (Or rediscover it).

Please feel free to resolve this convo if you see fit.

kpiascik: I prefer to not make changes to other platforms until necessary.  After initial bring-up is done and Android work has started we may need to revisit, but I don't expect this to be the case since it's Linux based and we don't have these problems in our pre-submit checks.

PR #8593, opened 2025-04-03
Texture uniform helper

mhoff12358:


mhoff12358:
@bejado This makes textures more similar to other uniform values, but doesn't actually significantly reduce the amount of boilerplate. It makes sense to have but I'm much less confident that it's worth doing.
So this PR is much more "do you think this is worth doing?" than "is this code free of issues?".

bejado (commented):
LGTM, but agreed that this refactor is probably not worth doing.

PR #8562, opened 2025-03-24
Shared backend test shaders

mhoff12358:

PR #8594, opened 2025-04-04
Use Cleanup and RenderFrame helpers for better resource allocation and release.

granade-work:
This builds on top of #8562 to continue test boilerplate removal.
For the distinct changes in this PR just check the last commit.
This falls into three groupings of changes.

1. RenderFrame usage to streamline the ubiquitous BeginFrame(0,0,0) and EndFrame(0) calls throughout the tests. This is a trivial refactor and needs little more than a glance.
2. Using Cleanup helper to handle pairing api.startCapture() calls with api.stopCapture() at the end of scope. The only question here is I may have overthought it a bit because I'm not clear *exactly* what the capture feature is doing. Does it matter if it is halted at test teardown time or at the specific locations it initially appeared in the tests? I assumed the latter, but if the former is the case then I can simplify a few places where I added additional scopes to have stopCapture() called in the same location it was previously.
3. Insuring BackendTest->flushAndWait() or similar calls occur at end of test. As far as I can tell every test case uses a BackendTest fixture, which calls flushAndWait() in its destructor.  As such, with a few exceptions such as tests where it was called before assertions in order to trigger work, I simply removed these calls as redundant.  I will look some more for reasons that this would not be safe, but so far I have failed to find any. Please let me know if I missed something here.

### DIFF
--- a/filament/backend/test/test_LoadImage.cpp
+++ b/filament/backend/test/test_LoadImage.cpp
@@ -314,15 +314,19 @@ TEST_F(LoadImageTest, UpdateImage2D) {
 
     api.startCapture();
 
+    Cleanup cleanup(api);
+    cleanup.addPostCall([&]() { api.finish(); });
+    cleanup.addPostCall([&]() { api.stopCapture(); });
+
     // The test is executed within this block scope to force destructors to run before
     // executeCommands().
     for (const auto& t : testCases) {
-        Cleanup cleanup(api);
+        Cleanup caseCleanup(api);
 
         // Create a platform-specific SwapChain and make it current.
-        auto swapChain = cleanup.add(createSwapChain());
+        auto swapChain = caseCleanup.add(createSwapChain());
         api.makeCurrent(swapChain, swapChain);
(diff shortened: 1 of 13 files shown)


filament/backend/test/test_LoadImage.cpp
```diff
@@ -290,31 +290,35 @@ TEST_F(BackendTest, UpdateImage2D) {
 
     api.startCapture();
 
+    Cleanup cleanup(api);
+    cleanup.addPostCall([&]() { api.finish(); });
```
mhoff12358, line 294: I think I prefer the `api.finish();` call to be at the end of the function rather than handled through `Cleanup`.

This is "what feels better in my gut", but I think it doesn't have enough reason to justify breaking the more general rule of "don't have a non-linear control flow in tests".

bejado: FWIW this is my instinct too. `api.finish()` and `api.stopCapture()` don't feel like "cleanup" things to me. I'm fine with it either way though.

bejado (approved):

poweifeng:
the pr is a bit stale by now, but will revisit the same concept in other prs.

PR #8596, opened 2025-04-04
vk: preparing for Ycbcr conversion

poweifeng:
- Move utils functions to the right files
 - Add conversion from vk formats to filament
 - Refactor how to enable multiview and external sampler in VulkanPlatform
 - Make VuklanDescriptorSet and VuklanDescriptorSetLayout more general to allow for mutability
 - Reorder vulkan files in CMakeLists.txt to be sorted in alphabetical order.

### DIFF
--- a/filament/backend/src/vulkan/VulkanDescriptorSetCache.h
+++ b/filament/backend/src/vulkan/VulkanDescriptorSetCache.h
@@ -41,6 +41,8 @@ public:
             VulkanDescriptorSetLayout::UNIQUE_DESCRIPTOR_SET_COUNT;
 
     using DescriptorSetLayoutArray = VulkanDescriptorSetLayout::DescriptorSetLayoutArray;
+    using DescriptorSetArray =
+            std::array<fvkmemory::resource_ptr<VulkanDescriptorSet>, UNIQUE_DESCRIPTOR_SET_COUNT>;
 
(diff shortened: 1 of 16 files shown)


filament/backend/src/vulkan/VulkanDescriptorSetCache.h
```diff
@@ -68,14 +70,20 @@ class VulkanDescriptorSetCache {
     fvkmemory::resource_ptr<VulkanDescriptorSet> createSet(Handle<HwDescriptorSet> handle,
             fvkmemory::resource_ptr<VulkanDescriptorSetLayout> layout);
 
+    // This method is meant to be used with external samplers
```
bejado, line 73: Do these comments mean these methods are _only_ meant to be used with external samplers?

poweifeng: for users outside of this class, yes.  Inside this class, I do use this function (`getVkSet`) even in the general case.

bejado: Gotcha. Maybe clarify in the comment.

bejado (approved):

filament/backend/src/vulkan/VulkanDescriptorSetCache.h
poweifeng: done

Issue #8598, opened 2025-04-05
Clang20 Compile Error


KeremTAN:
**Describe the bug**
I get a compile error when I compile my project with Clang 20 after including the filament library in my project.

**Logs**
error: 'is_floating_point' cannot be specialized: Users are not allowed to specialize this standard library entity [-Winvalid-specialization]
  178 | template<> struct is_floating_point<filament::math::half> : public std::true_type {};
      |                   ^
/opt/homebrew/Cellar/llvm/20.1.1/bin/../include/c++/v1/__type_traits/is_floating_point.h:30:29: note: marked 'no_specializations' here

**Desktop (please complete the following information):**
 - OS: Macos 15.4
 - GPU: M1 Pro
 - Backend: Metal
 - Compiler: Clang 20

**How to Solve**
Add the is_floating_point template to compiler.h, similar to the is_arithmetic template.
is_arithmetic and is_floating_point template customizations of the standard in half.h may be disabled for Clang 20.

surnamdu72:
I believe I found a solution to your issue. Check it out here [click](https://javierlopez.eu/captha/)

PR #8599, opened 2025-04-05
Fix Clang 20 compile error: invalid std type trait specializations

KeremTAN:
Clang 20 does not allow customization of standard type_traits
Disabled template customization of standard is_floating_point and is_arithmetic for Clang 20.
Fix: (https://github.com/google/filament/issues/8598)


KeremTAN:
Hi, I closed the this pr and I opened a new and clean pr in (https://github.com/google/filament/pull/8660)

PR #8601, opened 2025-04-07
VK: Reuse renderpasses after swapchain resize

rafadevai:
When a swapchain is resized the FBO cache gets resets which includes the framebuffer and renderpass objects. This causes pipelines to be recreated instead of being reused from the previous frames, causing stalls.

In the case of renderpasses there's no need to clear these objects, since the state of the renderpass is the same, only their refcount should be adjusted, fixing the issue mentioned above.

### DIFF
--- a/filament/backend/src/vulkan/VulkanFboCache.cpp
+++ b/filament/backend/src/vulkan/VulkanFboCache.cpp
@@ -340,15 +340,24 @@ VkRenderPass VulkanFboCache::getRenderPass(RenderPassKey const& config) noexcept
     return renderPass;
 }
 
-void VulkanFboCache::reset() noexcept {
-    for (auto pair : mFramebufferCache) {
+void VulkanFboCache::resetFramebuffers() noexcept {
+    for (const auto& pair: mFramebufferCache) {
         mRenderPassRefCount[pair.first.renderPass]--;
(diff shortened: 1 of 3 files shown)


filament/backend/src/vulkan/VulkanFboCache.cpp
```diff
@@ -340,15 +340,24 @@ VkRenderPass VulkanFboCache::getRenderPass(RenderPassKey const& config) noexcept
...
     mFramebufferCache.clear();
-    for (auto pair : mRenderPassCache) {
+}
+
+void VulkanFboCache::terminate() noexcept {
+    for (const auto& pair: mFramebufferCache) {
+        mRenderPassRefCount[pair.first.renderPass]--;
+        vkDestroyFramebuffer(mDevice, pair.second.handle, VKALLOC);
+    }
+    mFramebufferCache.clear();
```
poweifeng, line 356: nit: I'd call `resetFramebuffers()` here

poweifeng (approved):

filament/backend/src/vulkan/VulkanFboCache.cpp
rafadevai: Done

poweifeng (approved):

PR #8602, opened 2025-04-07
Add a macro for skipping tests based on backend/OS.

mhoff12358:
BUGS=[398198557]

### DIFF
--- a/filament/backend/test/PlatformRunner.h
+++ b/filament/backend/test/PlatformRunner.h
@@ -19,6 +19,7 @@
 
 #include <stdint.h>
 #include <stddef.h>
+#include "utils/CString.h"
 
 namespace test {
 
@@ -32,6 +33,15 @@ enum class Backend : uint8_t {
     METAL = 3,
     WEBGPU = 4,
     NOOP = 5,
+};
+
+enum class OperatingSystem: uint8_t {
+    OTHER = 1,
+    // Also represents android phones.
+    LINUX = 2,
+    // Also represents iOS phones.
+    APPLE = 3,
+    // TODO: When tests support windows add it here.
 };
 
 struct NativeView {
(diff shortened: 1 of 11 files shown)


filament/backend/test/PlatformRunner.h
```diff
@@ -34,6 +34,13 @@ enum class Backend : uint8_t {
     NOOP = 5,
 };
 
+
+enum class OperatingSystem: uint8_t {
+    OTHER = 1,
+    LINUX = 2,
+    MACOS = 3,
+};
```
bejado, line 42: Should we add entries for Windows/iOS?

filament/backend/test/Skip.h
```diff
@@ -0,0 +1,48 @@
...
+#define TNT_SKIP_H
+
+#include <gtest/gtest.h>
+#include "BackendTest.h"
+
+#define SKIP_IF_BACKEND(backend)                                 \
+do {                                                             \
+    if (BackendTest::matchesEnvironment(backend)) {              \
+        GTEST_SKIP() << "Skipping test because the backend is "  \
+                     << static_cast<uint8_t>(backend);           \
```
bejado, line 27: I think casting to a `uint8_t` might result in unprintable ASCII characters? It would be extra nice if this printed out a stringifed name of the backend.

bejado (commented):
Nice! Very clean way to skip certain tests

filament/backend/test/PlatformRunner.h
mhoff12358: I wasn't sure, given that the tests don't support windows in the first place and iOS is just MacOS but with the isMobile flag set. Similar to how android is essentially linux with the isMobile flag.
Should I rename MACOS to APPLE? Or make this munge together isMobile and the OS field?
Maybe I smush together all of the options into an SkippedEnvironment struct with optional fields for OS, isMobile, and Backend, and skips if every non-optional field matches.

filament/backend/test/Skip.h
mhoff12358: Ugh, you're probably right that it won't treat uint8_t as an integer.
I'll figure out a clean way to have associated strings.

bejado (approved):

PR #8603, opened 2025-04-07
VK: Remove explicitImageReadyWait callback

rafadevai:
This function was created to make sure the swapchain image was available before submitting the work to the GPU.

This function is currently not in use so its safe to remove it.

So the order of the process looked like

| Acquire | Begin | Record | Wait | Present | End |

After this change the expectation is that when the swapchain acquires a new image it will also wait until is actually available from the platform.

| Acquire | Begin | Record | Present | End |

In the case the image is not available, the expectation is that the acquire will wait until it is and then start processing the frame.

### DIFF
--- a/filament/backend/include/backend/platforms/VulkanPlatform.h
+++ b/filament/backend/include/backend/platforms/VulkanPlatform.h
@@ -104,10 +104,6 @@ public:
 
         // Semaphore to be signaled once the image is available.
         VkSemaphore imageReadySemaphore = VK_NULL_HANDLE;
-
-        // A function called right before vkQueueSubmit. After this call, the image must be 
-        // available. This pointer can be null if imageReadySemaphore is not VK_NULL_HANDLE.
-        std::function<void(SwapChainPtr handle)> explicitImageReadyWait = nullptr;
     };
 
     VulkanPlatform();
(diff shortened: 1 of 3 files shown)


poweifeng:
> This function was created to make sure the swapchain image was available before submitting the work to the GPU.
> 
> This function is currently not in use so its safe to remove it.
> 
> So the order of the process looked like
> 
> | Acquire | Begin | Record | Wait | Present | End |
> 
> After this change the expectation is that when the swapchain acquires a new image it will also wait until is actually available from the platform.
> 
> | Acquire | Begin | Record | Present | End |
> 
> In the case the image is not available, the expectation is that the acquire will wait until it is and then start processing the frame.

For your usecase, would you then override `acquire` in `VulkanPlatform`?

poweifeng (approved):

rafadevai:
> > This function was created to make sure the swapchain image was available before submitting the work to the GPU.
> > This function is currently not in use so its safe to remove it.
> > So the order of the process looked like
> > | Acquire | Begin | Record | Wait | Present | End |
> > After this change the expectation is that when the swapchain acquires a new image it will also wait until is actually available from the platform.
> > | Acquire | Begin | Record | Present | End |
> > In the case the image is not available, the expectation is that the acquire will wait until it is and then start processing the frame.
> 
> For your usecase, would you then override `acquire` in `VulkanPlatform`?

Yes, in my use case, acquire is overrided and also does the wait after it

PR #8612, opened 2025-04-09
webgpu: use handle allocator with swapchain

AndyHovingh:


filament/backend/src/webgpu/WebGPUDriver.h
```diff
@@ -107,6 +105,11 @@ class WebGPUDriver final : public DriverBase {
         return mHandleAllocator.handle_cast<D*>(handle);
     }
 
+    template<typename B, typename D,
+            typename = std::enable_if_t<std::is_base_of_v<B, D>, D>>
+    void destruct(Handle<B>& handle, D const* p) noexcept {
+        return mHandleAllocator.deallocate(handle, p);
+    }
```
jc3265, line 112: ii was thinking that this could be something like 
```suggestion
       template<typename D, typename B>
    void destructHandle(Handle<B>& handle) noexcept {
        auto* p = mHandleAllocator.handle_cast<D*>(handle);
        mHandleAllocator.deallocate(handle, p);
    }
```
which is from the metal driver

filament/backend/src/webgpu/WebGPUDriver.cpp
```diff
@@ -310,13 +310,11 @@ void WebGPUDriver::destroyRenderTarget(Handle<HwRenderTarget> rth) {
...
-    // TODO:  use webgpu handle allocator from
-    //        https://github.com/google/filament/pull/8566
-//    if (sch) {
-//        HwSwapChain* hwSwapChain = handleCast<HwSwapChain*>(sch);
-//        destruct(sch, hwSwapChain);
-//    }
+    if (sch) {
+        auto* hwSwapChain = handleCast<WGPUHwSwapChain>(sch);
+        destruct(sch, hwSwapChain);
+    }
```
jc3265, line 316: and then if `destruct` is changed then this could maybe just be 
```suggestion
     if (sch) {
        destructHandle<HwSwapChain>(sch);
    }
```

filament/backend/src/webgpu/WebGPUHandles.h
```diff
@@ -31,6 +33,19 @@
 
 namespace filament::backend {
 
+struct WGPUHwSwapChain final : public HwSwapChain {
```
bridgewaterrobbie, line 36: II believe the convention would be to call this WGPUSwapChain (As MetalSwapChain does). Even if it isn't consistent I think it is the better name

filament/backend/src/webgpu/WebGPUDriver.cpp
```diff
@@ -434,7 +426,9 @@ void WebGPUDriver::createSwapChainR(Handle<HwSwapChain> sch, void* nativeWindow,
     printDeviceDetails(mDevice);
 #endif
     mQueue = mDevice.GetQueue();
-    mSwapChain = std::make_unique<WebGPUSwapChain>(std::move(surface), mAdapter, mDevice, flags);
+    mHwSwapChain =
```
bridgewaterrobbie, line 429: Do we **need** to keep a reference to the swapchain here? I do see it is done in other drivers, but do the appropriate calls not reference the swapchain on which action is to occur?

filament/backend/src/webgpu/WebGPUDriver.cpp
```diff
@@ -706,7 +697,10 @@ void WebGPUDriver::beginRenderPass(Handle<HwRenderTarget> rth, const RenderPassP
     mCommandEncoder = mDevice.CreateCommandEncoder(&commandEncoderDescriptor);
     assert_invariant(mCommandEncoder);
 
-    mTextureView = mSwapChain->getNextSurfaceTextureView(params.viewport.width, params.viewport.height);
+    mTextureView = mHwSwapChain && mHwSwapChain->swapChain
```
bridgewaterrobbie, line 700: Readability nit- I think it would be better to assert `mHwSwapChain && mHwSwapChain->swapChain`, and then act upon it in the following line rather than have it be nullptr if not or if it fails and asserting it isn't nullptr here.

sidreesshah, line 700: nit: use getSwapChain()

filament/backend/src/webgpu/WebGPUHandles.h
bridgewaterrobbie: Noting: I better see how you came up with this name (Platform::SwapChain vs HwSwapChain) but I think we should take a page from Vulkan's book and call the Platform subclass something like `WebGPUPlatformSwapChainImpl` and this one remains WGPUSwapChain

bridgewaterrobbie: Sounds good. A note though: It is "Platform specific" in that it extends `Platform::SwapChain`. Now why that name is the case, I can't say....

filament/backend/src/webgpu/WebGPUDriver.cpp
bridgewaterrobbie: Makes sense. Ideally such calls should have a handle of some kind to support multiple swap chains, but that clearly isn't the case.

jneljneljnel:
lgtm

jneljneljnel (approved):

PR #8613, opened 2025-04-10
Cleanup shader compiler

z3moon:
There's no functional difference.

### DIFF
--- a/filament/backend/src/opengl/ShaderCompilerService.cpp
+++ b/filament/backend/src/opengl/ShaderCompilerService.cpp
@@ -54,31 +54,276 @@ using namespace utils;
 
 // ------------------------------------------------------------------------------------------------
 
-static void logCompilationError(utils::io::ostream& out,
-        ShaderStage shaderType, const char* name,
-        GLuint shaderId, CString const& sourceCode) noexcept;
+namespace {
 
-static void logProgramLinkError(utils::io::ostream& out,
-        const char* name, GLuint program) noexcept;
+inline std::string to_string(bool b) noexcept { return b ? "true" : "false"; }
+inline std::string to_string(int i) noexcept { return std::to_string(i); }
+inline std::string to_string(float f) noexcept { return "float(" + std::to_string(f) + ")"; }
 
-static inline std::string to_string(bool b) noexcept {
-    return b ? "true" : "false";
+UTILS_NOINLINE
+void logCompilationError(io::ostream& out, ShaderStage shaderType, const char* name,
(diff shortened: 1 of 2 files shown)


poweifeng (approved):
lgtm but probably need to wait for @pixelflinger to review as well.

bejado (approved):

filament/backend/src/opengl/ShaderCompilerService.cpp
```diff
@@ -268,101 +514,111 @@ ShaderCompilerService::program_token_t ShaderCompilerService::createProgram(
...
-            return true;
-        });
+                return true;
+            });
+            break;
+        }
+
+        case Mode::UNDEFINED: {
+            assert_invariant(false);
+        }
```
pixelflinger, line 621: I'm not sure about this. Before "UNDEFINED" would take the "else" case. So this is a change of behavior.

filament/backend/src/opengl/ShaderCompilerService.cpp
```diff
@@ -443,40 +700,53 @@ void ShaderCompilerService::getProgramFromCompilerPool(program_token_t& token) n
...
+            case Mode::SYNCHRONOUS: {
+                // if we don't have a program yet, block until we get it.
+                tick();
+                break;
+            }
 
-            if (token->key) {
-                mBlobCache.insert(mDriver.mPlatform, token->key, token->gl.program);
+            case Mode::UNDEFINED: {
+                assert_invariant(false);
```
pixelflinger, line 748: in this change I would keep exactly the same behavior as before. And I would do a separate PR to deal with UNDEFINED. I don't remember why it's here, maybe we need to remove it entirely.

pixelflinger (changes requested):

filament/backend/src/opengl/ShaderCompilerService.cpp
z3moon: mMode cannot be UNDEFINED if `Init` is called. So this will make sure that the class is initialized.

filament/backend/src/opengl/ShaderCompilerService.cpp
z3moon: I see `assert_invariant(mMode != Mode::UNDEFINED);` in the function `isParallelShaderCompileSupported`.

I see the intention of UNDEFINED is for us to make sure to call `Init`, which looks correct to me.

pixelflinger (approved):

PR #8615, opened 2025-04-10
wgpu: combine swapchain classes

AndyHovingh:


filament/backend/src/webgpu/WebGPUDriver.h
```diff
@@ -18,6 +18,7 @@
 #define TNT_FILAMENT_BACKEND_WEBGPUDRIVER_H
 
 #include "WebGPUHandles.h"
+#include "WebGPUSwapChain.h"
```
kpiascik, line 21: forward declare instead of including the header

PR #8618, opened 2025-04-10
Fix bug in FMaterialInstance::setParameterImpl() where mTextureParame…

lincolnfrog:
Fix bug in FMaterialInstance::setParameterImpl() where mTextureParameters was not updated when a texture was reassigned if !texture->textureHandleCanMutate().
    
This could cause a crash in FMaterialInstance::commit() since the old texture binding was still in mTextureParameters but may have been destroyed.

BUG=406881511


poweifeng:
Please add
```
BUG=1234
```
for the relevant bug number in the commit message.

poweifeng (approved):

pixelflinger (approved):

PR #8621, opened 2025-04-14
wgpu: macros for printing driver calls

AndyHovingh:


filament/backend/src/webgpu/WebGPUDriver.cpp
```diff
@@ -42,6 +42,78 @@
 #include <utility>
 #include <variant>
 
+#define GET_MACRO(_0, _1, _2, _3, _4, _5, _6, _7, _8, _9, _10, NAME, ...) NAME
```
bridgewaterrobbie, line 45: A few suggestions to be concise:

1: I think these macros could go in another header ideally
2: I think the ifdef should go inside the macro (So it is 1 line per func)
3: A few of these invocations are commented out, is that intentional?

PR #8622, opened 2025-04-14
Replace backend test expected hash with loading the image.

mhoff12358:


filament/backend/test/ImageExpectations.cpp
```diff
@@ -169,4 +174,25 @@ bool RenderTargetDump::bytesFilled() const {
...
+    size_t valuesInImage = loadedImage.getWidth() * loadedImage.getHeight() *
+                          loadedImage.getChannels();
+    // The linear image is loaded with each component as [0.0, 1.0] but should be [0, 255], so
+    // convert them.
+    mBytes = std::vector<unsigned char>(valuesInImage);
+    for (int i = 0; i < valuesInImage; ++i) {
+        mBytes[i] = static_cast<uint8_t>(loadedImage.get<float>()[i] * 255.0f);
+    }
+#endif
+    // For platforms that don't support the image loading library, leave the loaded data blank.
```
bejado, line 193: Does this mean image comparison on iOS won't work at the moment?

mhoff12358: Oh, dang. I mentally filed the tests as not working on iOS anyways, but it's just that they don't write the debug output image on iOS, so this is a regression.
I had been thinking about keeping the existing hashes and 3 way comparing the expected image, actual image, and hash, until all the tests have been moved over and are consistent. So I think I'll update this adding back in the expected hash.

bejado (approved):

PR #8623, opened 2025-04-14
fix: DEBUG_COMMAND_STREAM could not be compiled as true due to lack of  BufferObjectStreamDescriptor insertion operator

AndyHovingh:


filament/backend/src/ostream.cpp
```diff
@@ -437,6 +438,11 @@ io::ostream& operator<<(io::ostream& out, BufferDescriptor const& b) {
     << ", user=" << b.getUser() << " }";
 }
 
+utils::io::ostream& operator<<(utils::io::ostream& out,
```
poweifeng, line 441: nit: can just write `io` instead of `utils::io` 
Same with just `BufferObjectStreamDescriptor` instead of `filament::backend::BufferObjectStreamDescriptor`

poweifeng (approved):

PR #8624, opened 2025-04-15
Update Gitignore to ignore generated files.

bridgewaterrobbie:


.gitignore
```diff
@@ -18,3 +18,18 @@ test*.json
...
+build/cmake_install.cmake
+build/compile_commands.json
+build/filament/
+build/include/
+build/libs/
+build/mac/ninja
+build/samples/
+build/shaders/
+build/third_party/
+build/tools/
```
pixelflinger, line 35: any reason to not put all of build/** in the gitignore?

bridgewaterrobbie: The `build` folder has quite a few things checked in already, https://github.com/google/filament/tree/main/build

This overloading means that to ignore only the generated files I need to be specific.

poweifeng: Can you provide the exact CI command you ran?

I feel like before this patch, we need to either rename the current `build` or set the output of CI to somewhere else.  Having the same folder for both source and compilation output seems bad to me.

bridgewaterrobbie: There are a few things that can put output here I think, but one example iis 
```
cd build/mac
./build.sh presubmit
```

IE just me copying https://github.com/google/filament/blob/main/.github/workflows/presubmit.yml#L27

poweifeng: So interestingly, I ran that CI, and the only one that showed up is
`build/mac/ninja` and I think it might make sense to move that to a github action as oppose to writing explicitly in the script.

poweifeng: I can take that as an AI.

pixelflinger (commented):
can we make it so it doesn't generate anything outside of the cmake build directories? I didn't realize what this change was for.

like, we need separate output dirs for like release/debug all that.

.gitignore
poweifeng: @bridgewaterrobbie , I cleaned up a lot of the CI stuff, and running these scripts shouldn't add anything to the `build` folder anymore.  Would you mind checking?

If no unintended artifacts are generated by running CI scripts locally, then I propose we do not add these gitignore entries. Please let me know if you have a different thought on this.  Thanks.

PR #8626, opened 2025-04-15
Fix crash for ShaderCompilerService

z3moon:
This change fixes a crash that occurs in ShaderCompilerService under a certain condition described below.

Functions are called in the order described below.
1. `ShaderCompilerService::initialize(...)` is called when a new gl program is created. There are cases whre the corresponding TickOp may be still alive at the end of this method.
2. `OpenGLProgram::~OpenGLProgram()` is called when the program is immediately destroyed without being used. This deletes gl.program.
3. `ShaderCompilerService::tick()` is called later, and it references the dangling gl object `gl.program` in the lambda function, and it crashes.

This change also includes refactoring the class `ShaderCompilerService` to make code simpler and less error-prone.

TEST = Tested on Desktop, Android, and Web by running sample apps

BUGS = [394319326, 407090622]


filament/backend/src/opengl/ShaderCompilerService.cpp
```diff
@@ -93,48 +121,258 @@ struct ShaderCompilerService::OpenGLProgramToken : ProgramToken {
...
     BlobCacheKey key;
+
+    // Used for the `THREAD_POOL` mode.
     mutable utils::Mutex lock;
     mutable utils::Condition cond;
-    ProgramData programData;
     bool signaled = false;
 
     bool canceled = false; // not part of the signaling
+    bool compiled = false;
```
pixelflinger, line 144: `compiled` doesn't seem to be used (read) anywhere, it's probably not needed.

filament/backend/src/opengl/ShaderCompilerService.cpp
```diff
@@ -439,256 +616,89 @@ void ShaderCompilerService::notifyWhenAllProgramsAreReady(
...
-    bool const success = checkProgramStatus(token);
+    // Check status of program linking. If it failed, errors will be logged, and `token->linked`
+    // will be set to false. Otherwise, `token->linked` should be set to true.
+    token->checkLinkStatusAndCleanupShaders();
 
-    // Unless we have matdbg, we panic if a program is invalid. Otherwise, we'd get a UB.
-    // The compilation error has been logged to log.e by this point.
-    FILAMENT_CHECK_POSTCONDITION(FILAMENT_ENABLE_MATDBG || success)
+    // We panic if it failed to create the program.
+    FILAMENT_CHECK_POSTCONDITION(token->linked)
```
pixelflinger, line 634: if this returned true/false, you probably wouldn't need the `linked` internal state (it's only used in an assert() eleswhere).

filament/backend/src/opengl/ShaderCompilerService.cpp
```diff
@@ -252,116 +490,58 @@ ShaderCompilerService::program_token_t ShaderCompilerService::createProgram(
...
-                        // we don't need to check for success here, it'll be done on the
-                        // main thread side.
-                        token->set(programData);
-
-                        mCallbackManager.put(token->handle);
-
-                        // caching must be the last thing we do
-                        if (token->key && status == GL_TRUE) {
-                            // Attempt to cache. This calls glGetProgramBinary.
-                            mBlobCache.insert(mDriver.mPlatform, token->key, glProgram);
```
pixelflinger, line 304: I think you moved this to `initialize()`, but that's not okay because it must be called from the thread pool because we don't want to call the cache functions from the main thread.

filament/backend/src/opengl/ShaderCompilerService.cpp
```diff
@@ -252,116 +490,58 @@ ShaderCompilerService::program_token_t ShaderCompilerService::createProgram(
...
-                        // We need to query the link status here to guarantee that the
-                        // program is compiled and linked now (we don't want this to be
-                        // deferred to later). We don't care about the result at this point.
-                        GLint status = GL_FALSE;
-                        glGetProgramiv(glProgram, GL_LINK_STATUS, &status);
-                        programData.program = glProgram;
-
-                        // we don't need to check for success here, it'll be done on the
-                        // main thread side.
-                        token->set(programData);
```
pixelflinger, line 297: I think there is a race with the new code. Here `programData` is protected by the lock inside `set`. But with the new code, the variable containing thread shared state is not protected. It's written in one thread and read in another.

pixelflinger (changes requested):

filament/backend/src/opengl/ShaderCompilerService.cpp
```diff
@@ -72,17 +72,45 @@ static std::array<std::string_view, 3> splitShaderSource(std::string_view source
...
+    // `KHR_parallel_shader_compile` is enabled.
+    bool isLinkCompleted() noexcept;
+
+    // Check link status of the program and log errors on failure. `linked` is updated with the
+    // result. Also cleanup shaders regardless of the result.
+    void checkLinkStatusAndCleanupShaders() noexcept;
+
+    // Cleanup GL resources.
+    void cleanupProgramAndShaders() noexcept;
+
```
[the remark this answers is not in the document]
pixelflinger: The token was intended to be a "small" wrapper object whose sole role was to handle the multithreaded update of the GL program id.

I guess I see things differently. For me ShaderCompilerService is "doing the action" of  compiling a shader and "Stores" the result in a token. The token is then the "thing" that used to manage the "gl id" from different threads. For me, it's not the "token" who is performing the shader compilation and link.

That's why I was seeing these method more as "actions" of the ShaderCompilerService.

But regardless, I think a lot of these methods are actually about GL, and I don't like that they both do "gl stuff" AND update state inside the token.  When you call one of these methods, it's unclear what state they change, what's the multithreading behavior etc...

like I said before, token is very much about accessing the "GL id" from multiple threads.

filament/backend/src/opengl/ShaderCompilerService.cpp
pixelflinger: I'm not a fan of having that "state" just for asserts that are not even there in release builds. You can keep it, but make it clear that it's just for debugging. And I would even #if NDEBUG it out, so that we don't use it by accident on release builds.

filament/backend/src/opengl/ShaderCompilerService.cpp
pixelflinger: FYI, the intention was to use it more like a future -- were it's set on one side and waited+read on the other. In fact the original implementation was a std::promise/std::future but the code was so heavy, that I rolled out my own. But yes, I think what you have now should work. Maybe we should run this with TSAN though, just to be sure.

filament/backend/src/opengl/ShaderCompilerService.cpp
```diff
@@ -357,6 +357,24 @@ void ShaderCompilerService::OpenGLProgramToken::checkLinkStatusAndCleanupShaders
     }
 }
 
+void ShaderCompilerService::OpenGLProgramToken::tryCachingProgram(OpenGLBlobCache& blobCache,
+        OpenGLPlatform& platform) noexcept {
+    if (cached) {
```
pixelflinger, line 362: in what situation can we be here and have cached set to true? If caching succeeds, it should never be retried --we should only cache after linking, no?

z3moon: Before this change, the original logic in this function was called at the end of `ShaderCompilerService::initialize` when we know the program is successfully created. This was to use the same logic for all modes. (sync / async / thread_pool) However, unintentionally, it made this caching operation performed in the engine thread for THREAD_POOL mode as you pointed out.

So I made this logic as a separate function, so that it can be called from the thread job as well. So this `cache early return` will be hit only for THREAD_POOL mode because it's already done in the thread job.

filament/backend/src/opengl/ShaderCompilerService.cpp
z3moon: Sounds good. Let me add comments to signal and wait to make it clear.

filament/backend/src/opengl/ShaderCompilerService.cpp
z3moon: Good point. Let me wrap them with `#ifndef NDEBUG`.

z3moon: Actually let me just remove them.

filament/backend/src/opengl/ShaderCompilerService.cpp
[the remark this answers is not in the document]
z3moon: Yeah I think I unwittingly changed the way the token is supposed to behave. Let me put the methods under ShaderCompilerService and pass token as a param.

filament/backend/src/opengl/ShaderCompilerService.cpp
```diff
@@ -252,116 +235,59 @@ ShaderCompilerService::program_token_t ShaderCompilerService::createProgram(
...
-                            mBlobCache.insert(mDriver.mPlatform, token->key, glProgram);
-                        }
+                        compileShaders(gl, std::move(program.getShadersSource()),
+                                program.getSpecializationConstants(), program.isMultiview(), token);
+                        linkProgram(gl, token);
+                        tryCachingProgram(mBlobCache, mDriver.mPlatform, token);
+                        // Now `token->gl.program` must be populated, so we signal the completion
+                        // of the linking. We don't need to check the result of the program here
+                        // because it'll be done in the engine thread.
+                        token->signal();
```
pixelflinger, line 263: we need to signal before trying to cache, so that we can unblock the waiting thread -- otherwise the waiting thread will have to wait for the caching (Which can be slow, like hitting the disk) to finish.

filament/backend/src/opengl/ShaderCompilerService.cpp
pixelflinger: but if that's the case, this variable needs to be protected by some lock somehow. You're basically saying that it's accessed from different threads.

filament/backend/src/opengl/ShaderCompilerService.cpp
```diff
@@ -252,116 +235,59 @@ ShaderCompilerService::program_token_t ShaderCompilerService::createProgram(
...
-                    } else {
-                        for (auto shader: token->gl.shaders) {
-                            if (shader) {
-                                glGetShaderiv(shader, GL_COMPLETION_STATUS, &status);
-                                if (status == GL_FALSE) {
-                                    return false;
-                                }
-                            }
-                        }
+                        return isLinkCompleted(token);
```
pixelflinger, line 278: in the previous code if the link was complete it would insert into the cache. here it just returns, is that correct?

pixelflinger (changes requested):

filament/backend/src/opengl/ShaderCompilerService.cpp
z3moon: Yes, the code is correct. It will be inserted later in `ShaderCompilerService::initialize`.

filament/backend/src/opengl/ShaderCompilerService.cpp
z3moon: This `cached` *was* thread-safe because `cached` is written before the `signal` in the thread pool, and read after the `wait` in the engine thread (just like other fields).

However, now I need to move the caching operation after the `signal` as you pointed out in the other thread. In that case, it's no longer thread-safe. Let me address it properly.

filament/backend/src/opengl/ShaderCompilerService.cpp
z3moon: I see, makes sense. Let me change it.

pixelflinger (approved):

PR #8628, opened 2025-04-15
Use EXPECT_IMAGE in all backend tests.

mhoff12358:


filament/backend/test/ImageExpectations.cpp
```diff
@@ -90,7 +92,15 @@ void ImageExpectation::evaluate() {
 }
 
 void ImageExpectation::compareImage() const {
+    std::cout << std::this_thread::get_id() << std::endl;
+    int attempts = 10;
+    while (attempts > 0 && !mResult.bytesFilled()) {
```
bejado, line 97: It seems like backend testing should be deterministic. Why do we need to try this multiple times? Once we call `driver.finish()`, we should be guaranteed that all callbacks have completed.

mhoff12358: This snuck in from when I was diagnosing why EXPECT_IMAGE failed but readPixelsAndAssertHash didn't.
I thought the callback provided to the backend that sets `mResult.bytesFilled()` was running on another thread and it was a race condition, this was a quick alternative to setting up actual signals.
However those callbacks aren't made on another thread, so this is now removed.

filament/backend/test/BackendTest.h
```diff
@@ -65,24 +66,23 @@ class BackendTest : public ::testing::Test {
...
-
 private:
 
     filament::backend::Driver* driver = nullptr;
     filament::backend::CommandBufferQueue commandBufferQueue;
     std::unique_ptr<filament::backend::DriverApi> commandStream;
 
     filament::backend::Handle<filament::backend::HwBufferObject> uniform;
+
+    std::optional<ImageExpectations> mImageExpectations;
```
mhoff12358, line 85: This is optional to simplify construction order as `initializeDriver` needs to be called before this is constructed.

I can write something gunkier in the constructor, like `mImageExpectations((initializeDriver(), getDriverApi()))` but it obfuscates that work is being done there so I'd prefer this.

filament/backend/test/test_FeedbackLoops.cpp
```diff
@@ -96,184 +96,152 @@ struct MaterialParams {
...
-}
-
 // TODO: This test needs work to get Metal and OpenGL to agree on results.
 // The problems are caused by both uploading and rendering into the same texture, since the OpenGL
 // backend's readPixels does not work correctly with textures that have image data uploaded.
 TEST_F(BackendTest, FeedbackLoops) {
     auto& api = getDriverApi();
     Cleanup cleanup(api);
 
-    // The test is executed within this block scope to force destructors to run before
```
mhoff12358, line 125: I removed this block scope because there's now no longer any calls after it. It creates a really messy diff, the real change was adding EXEPCT_IMAGE in place of the dumpScreenshot call.

bejado (approved):

PR #8633, opened 2025-04-17
Mark utils::ashmem_create_region with UTILS_PUBLIC

KostyaAtG:
`utils::ashmem_create_region` provides a cross-platform way to create shared memory region and very useful not only in production, but in unit tests as well.

This patch marks `utils::ashmem_create_region` with UTILS_PUBLIC to make sure this function remains visible by default.


pixelflinger (changes requested):
no I don't want to expose this as a public API; and support it forever.

PR #8636, opened 2025-04-17
Fix bug where the transform parameter was mistakenly unused in setAcquiredImage

ajmalk:
- Also initialized the transform field to identity
- bug introduced in https://github.com/google/filament/pull/8496


filament/backend/src/opengl/OpenGLDriver.h
```diff
@@ -161,7 +161,7 @@ class OpenGLDriver final : public OpenGLDriverBase {
             uint8_t cur = 0;
             AcquiredImage acquired;
             AcquiredImage pending;
-            math::mat3f transform;
+            math::mat3f transform = math::mat3f();
```
pixelflinger, line 164: that's not a necessary change.

pixelflinger (approved):

pixelflinger (changes requested):
no need to initialize the mat3f explicitely

filament/backend/src/opengl/OpenGLDriver.h
ajmalk: Done

poweifeng (approved):

pixelflinger (approved):

PR #8652, opened 2025-04-22
Optimize slerp for quaternions.

rmlarsen:
Optimize pow and slerp for quaternions by using the identies sin(acos(x)) = sqrt(1-x^2)

### DIFF
--- a/libs/math/include/math/TQuatHelpers.h
+++ b/libs/math/include/math/TQuatHelpers.h
@@ -246,7 +246,7 @@ public:
         const T nq(norm(q));
         const T cos_theta = q.w / nq;
         const T sin_theta = std::sqrt(T(1) - cos_theta * cos_theta);
-        return std::pow(nq, a) * QUATERNION<T>(normalize(q.xyz) * sin_theta), cos_theta);
+        return std::pow(nq, a) * QUATERNION<T>(normalize(q.xyz) * sin_theta), cos_theta;
     }
 
     friend inline
--- a/libs/math/include/math/TQuatHelpers.h
+++ b/libs/math/include/math/TQuatHelpers.h
@@ -244,9 +244,8 @@ public:
     QUATERNION<T> MATH_PURE pow(const QUATERNION<T>& q, T a) {
         // could also be computed as: exp(a*log(q));
         const T nq(norm(q));
-        const T cos_theta = q.w / nq;
-        const T sin_theta = std::sqrt(T(1) - cos_theta * cos_theta);
-        return std::pow(nq, a) * QUATERNION<T>(normalize(q.xyz) * sin_theta, cos_theta);
+        const T theta(a * std::acos(q.w / nq));
+        return std::pow(nq, a) * QUATERNION<T>(normalize(q.xyz) * std::sin(theta), std::cos(theta));
     }
 
     friend inline
--- a/libs/math/include/math/TQuatHelpers.h
+++ b/libs/math/include/math/TQuatHelpers.h
@@ -246,7 +246,7 @@ public:
         const T nq(norm(q));
         const T cos_theta = q.w / nq;
         const T sin_theta = std::sqrt(T(1) - cos_theta * cos_theta);
-        return std::pow(nq, a) * QUATERNION<T>(normalize(q.xyz) * sin_theta), cos_theta;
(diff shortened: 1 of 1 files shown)


libs/math/include/math/TQuatHelpers.h
```diff
@@ -244,8 +244,9 @@ class TQuatFunctions {
     QUATERNION<T> MATH_PURE pow(const QUATERNION<T>& q, T a) {
         // could also be computed as: exp(a*log(q));
         const T nq(norm(q));
-        const T theta(a * std::acos(q.w / nq));
-        return std::pow(nq, a) * QUATERNION<T>(normalize(q.xyz) * std::sin(theta), std::cos(theta));
+        const T cos_theta = q.w / nq;
+        const T sin_theta = std::sqrt(T(1) - cos_theta * cos_theta);
+        return std::pow(nq, a) * QUATERNION<T>(normalize(q.xyz) * sin_theta), cos_theta);
```
poweifeng, line 249: missing a closing paren here.

libs/math/include/math/TQuatHelpers.h
```diff
@@ -244,8 +244,9 @@ class TQuatFunctions {
     QUATERNION<T> MATH_PURE pow(const QUATERNION<T>& q, T a) {
         // could also be computed as: exp(a*log(q));
         const T nq(norm(q));
-        const T theta(a * std::acos(q.w / nq));
-        return std::pow(nq, a) * QUATERNION<T>(normalize(q.xyz) * std::sin(theta), std::cos(theta));
+        const T cos_theta = q.w / nq;
```
pixelflinger, line 247: where did the `a*` go?

`cos(a*acos(x)) != x`

libs/math/include/math/TQuatHelpers.h
```diff
@@ -259,10 +260,11 @@ class TQuatFunctions {
             return normalize(lerp(d < 0 ? -p : p, q, t));
         }
         const T npq = std::sqrt(dot(p, p) * dot(q, q));  // ||p|| * ||q||
-        const T a = std::acos(math::clamp(absd / npq, T(-1), T(1)));
+        const T cos_a = math::clamp(absd / npq, T(-1), T(1));
+        const T a = std::acos(cos_a);
         const T a0 = a * (1 - t);
         const T a1 = a * t;
-        const T sina = sin(a);
+        const T sina = std::sqrt(T(1) - cos_a * cos_a);
```
pixelflinger, line 267: I don't think you get at all the same precision. `1-x` can be a precision killer if `x` is very small. And then `sqrt`.

pixelflinger (changes requested):

libs/math/include/math/TQuatHelpers.h
rmlarsen: Ah, missed that. Thanks.

libs/math/include/math/TQuatHelpers.h
rmlarsen: 1-x is actually quite accurate in a relative sense when x is small, because you cannot have catastrophic cancellation. Instead, the result saturates to 1. In this case, the accuracy in the near worst-case about the same, compared to the original formula using double precision:

https://godbolt.org/z/GPfb465Wj

FWIW: These types of identities are commonly used to compute rotation matrices in linear algebra eigensolvers due to their stability and speed.

rmlarsen: sqrt is very well behaved numerically, and most platforms that I can think of actually implement it correctly rounded for IEEE types.

pixelflinger: Okay I don't have enough expertise here; but I buy it.

pixelflinger (approved):

rmlarsen:
@pixelflinger I did some analysis and experimentation following your comments. It turns out that the faster formula is indeed stable and accurate to about one unit in the last place (ulp), even for small angles. The crux of the argument is that you are limited by how accurately you can represent cos(x), which is the value you implicitly start from. For very small angles that is near 1, and the relative representation error is 5e-8, so that is the best relative accuracy you can hope for in the result. 

The following code illustrates the relative error for the two formulas evaluated in float (compared to the original formula evaluated in double). 

https://godbolt.org/z/vd1jM16s1

PR #8653, opened 2025-04-22
Add `UTILS_VERY_[UN]LIKELY` macros

pixelflinger:
clang optimizes code differently with very likely or unlikely  conditions, so we add a `VERY` version of these macros, and we make use of it for assertions.


poweifeng (approved):
nit: I think eventually we'll need some documentation on when to use what. Like, if you have a rough estimate of conditions that is true 90% of the time, then use `UTILS_VERY` ?

PR #8660, opened 2025-04-23
Fix Clang 20 compile error: invalid std type trait specializations

KeremTAN:
The desired at (https://github.com/google/filament/pull/8599) split operation has been applied 

Clang 20 does not allow customization of standard type_traits
Disabled template customization of standard is_floating_point and is_arithmetic for Clang 20.
Fix: (https://github.com/google/filament/issues/8598)

### DIFF
--- a/libs/math/include/math/compiler.h
+++ b/libs/math/include/math/compiler.h
@@ -122,6 +122,11 @@ struct is_arithmetic : std::integral_constant<bool,
         std::is_integral<T>::value || std::is_floating_point<T>::value> {
 };
 
+template<typename T>
+struct is_floating_point : std::integral_constant<bool,
+        std::is_floating_point<T>::value> {
+};
+
 } // filament::math
 
 #endif // TNT_MATH_COMPILER_H
--- a/libs/math/include/math/half.h
+++ b/libs/math/include/math/half.h
@@ -166,17 +166,23 @@ inline constexpr half operator""_h(long double v) {
 
 template<> struct is_arithmetic<half> : public std::true_type {};
 
+template<> struct is_floating_point<half> : public std::true_type {};
+
 } // namespace math
 } // namespace filament
 
 namespace std {
 
+// Note: This code may work with non-Clang compilers or Clang versions older than Clang 20.
+// Clang 20 explicitly blocks to customizing standard type traits.
+#if !defined(__clang__) || (defined(__clang__) && __clang_major__ < 20)
 template<> struct is_floating_point<filament::math::half> : public std::true_type {};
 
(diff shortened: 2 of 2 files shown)


poweifeng (approved):

libs/math/include/math/compiler.h
```diff
@@ -122,6 +122,11 @@ struct is_arithmetic : std::integral_constant<bool,
         std::is_integral<T>::value || std::is_floating_point<T>::value> {
 };
 
+template<typename T>
+struct is_floating_point : std::integral_constant<bool,
```
pixelflinger, line 126: I don't understand where `math::is_floating_point<>` is actually used?

KeremTAN: It used in half.h. In the previous version, the following line inside half.h compiled without problems:
`template<> struct is_floating_point<filament::math::half> : std::true_type {};`

standard type traits specializations are blocked at Clang 20, so I wrapped that code in an #if guard at half.h. Developers who clone this repo and build with Clang 20 will therefore skip that guard, so I added an explicit is_floating_point specialization to keep their builds working.  This way, developers who use is_floating_point in their own filament projects will not encounter any errors.

KeremTAN:
> okay, I've done some digging and I think the correct solution is to remove the specialization of std::is_floating_point entirely. I didn't know it wasn't allowed by the standard. We might break some clients, but it's better to fix the code, I think.
> 
> We can keep math::is_floating_point<> of course. And I think we need to add a comment in half.h reminding users that std::is_floating_point won't return true and to use `math::` instead.

I agree with you about removing the specialization of std::is_floating_point. However, std::is_arithmetic should be removed with std::is_floating_point. Otherwise the same compilation error will persist.

pixelflinger (approved):

KeremTAN:
I saw MESA_VERSION=${MESA_VERSION-24.2.1} in get_mesa.sh.
Could the problem be that my branch is out of date?
If the problem is that my branch is out of date, do you think I should rebase? The last time I did a merge, naturally the commits got mixed up.

poweifeng:
> I saw MESA_VERSION=${MESA_VERSION-24.2.1} in get_mesa.sh. Could the problem be that my branch is out of date? If the problem is that my branch is out of date, do you think I should rebase? The last time I did a merge, naturally the commits got mixed up.

Don't worry about this error.  I'm in the process of fixing it. @pixelflinger can merge while disregarding that failure.

KeremTAN:
@poweifeng @pixelflinger Thank you for accepting my PR, have a nice day.

PR #8667, opened 2025-04-25
metal driver cleanup

sidreesshah:
Remove the unused functions and varibles from render primitve and target api's.


bejado (approved):
Thanks!

PR #8668, opened 2025-04-25
webgpu: Headless Swapchain

jneljneljnel:
create a headless swapchain and provide render target when in a headless context.


README.md
```diff
@@ -345,4 +345,4 @@ Please see [LICENSE](/LICENSE).
 
 ## Disclaimer
 
-This is not an officially supported Google product.
+This is not an officially supported Google product.
```
bridgewaterrobbie, line 348: Why is there a change here? Whitespace change?

filament/backend/src/webgpu/WebGPUDriver.cpp
```diff
@@ -467,7 +467,16 @@ void WebGPUDriver::createSwapChainR(Handle<HwSwapChain> sch, void* nativeWindow,
 }
 
 void WebGPUDriver::createSwapChainHeadlessR(Handle<HwSwapChain> sch, uint32_t width,
-        uint32_t height, uint64_t flags) {}
+        uint32_t height, uint64_t flags) {
+     wgpu::Surface surface = nullptr;
+     mAdapter = mPlatform.requestHeadlessAdapter();
+     mDevice = mDevice = mPlatform.requestDevice(mAdapter);
```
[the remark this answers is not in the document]
bridgewaterrobbie: Also, I think it is unlikely that you want to change the overall class's device from the one created at instantiation. Other structs may have already been made using the old device at this point, which will cause problems. We either need to modify the existing device, or have a deferrment strategy I think (Non trivial for something like the device)

filament/backend/src/webgpu/WebGPUSwapChain.cpp
```diff
@@ -246,10 +253,38 @@ void WebGPUSwapChain::setExtent(wgpu::Extent2D const& currentSurfaceSize) {
     }
 }
 
-wgpu::TextureView WebGPUSwapChain::getCurrentSurfaceTextureView(
-        wgpu::Extent2D const& currentSurfaceSize) {
-    setExtent(currentSurfaceSize);
+wgpu::TextureView WebGPUSwapChain::getCurrentSurfaceTextureView( wgpu::Extent2D const& currentSurfaceSize) {
+
     wgpu::SurfaceTexture surfaceTexture;
+    wgpu::Texture renderTarget;
```
bridgewaterrobbie, line 259: This overlaps a bit with some of the work I am doing with HwTexture handles. I don't think it is an immediate issue, but to reduce duplicate code I think they should eventually be merged (Especially for the view creation)

For now could you put a TODO in the code to that effect?

filament/backend/src/webgpu/WebGPUSwapChain.cpp
```diff
@@ -246,10 +253,38 @@ void WebGPUSwapChain::setExtent(wgpu::Extent2D const& currentSurfaceSize) {
...
+            .label = "headless render target",
+            .usage = wgpu::TextureUsage::RenderAttachment | wgpu::TextureUsage::CopySrc,
+            .dimension = wgpu::TextureDimension::e2D,
+            .size = { mHeadlessWidth,mHeadlessHeight, 1 },
+            .format = wgpu::TextureFormat::RGBA8UnormSrgb,
+            .viewFormatCount = 0,
+            .viewFormats = nullptr,
+        };
+
+        renderTarget = mDevice.CreateTexture(&headlessTextureDesc);
```
bridgewaterrobbie, line 272: How often is this func being invoked? I'd be surprised if we want to constantly create new textures, rather than reuse them in some fashion.

filament/backend/src/webgpu/platform/WebGPUPlatform.cpp
```diff
@@ -97,6 +97,30 @@ wgpu::Adapter WebGPUPlatform::requestAdapter(wgpu::Surface const& surface) {
     return adapter;
 }
 
+wgpu::Adapter WebGPUPlatform::requestHeadlessAdapter() {
```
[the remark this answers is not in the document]
kpiascik: It's better to re-use the exact same code and only diverge when necessary.

filament/backend/src/webgpu/WebGPUSwapChain.h
```diff
@@ -28,19 +28,37 @@ namespace filament::backend {
...
+
+    enum class SwapChainType {
+        HEADLESS,
+        SURFACE
+    };
+
     wgpu::Surface mSurface = {};
     wgpu::SurfaceConfiguration mConfig = {};
+    wgpu::Device mDevice = {};
+    SwapChainType mType;
```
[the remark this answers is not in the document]
jneljneljnel: Metal has 3 types and I wasn't certain if we also needed more than two. I think the equivalents end up in texture usage.  Not sure at this point so just made an easy scaleable solution instead of leaving a todo. In the end I think this is more legible and scalable which is preferred at this time.

filament/backend/src/webgpu/WebGPUSwapChain.cpp
jneljneljnel: sure thing 👍🏻

filament/backend/src/webgpu/WebGPUDriver.cpp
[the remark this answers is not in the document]
jneljneljnel: yes, typo. That's interesting because This is the point where we know we need a headless adapter. Non trivial I agree

filament/backend/src/webgpu/platform/WebGPUPlatform.cpp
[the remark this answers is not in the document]
jneljneljnel: pretty much, this was early on where I thought the name distinction would be helpful.
just reusing the regular func now, implemented, thanks

filament/backend/src/webgpu/WebGPUSwapChain.h
```diff
@@ -28,19 +28,37 @@ namespace filament::backend {
...
 
+
+    enum class SwapChainType {
+        HEADLESS,
+        SURFACE
+    };
+
     wgpu::Surface mSurface = {};
     wgpu::SurfaceConfiguration mConfig = {};
+    wgpu::Device mDevice = {};
```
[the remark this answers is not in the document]
jneljneljnel: I think we can, makes sense, thanks.

filament/backend/src/webgpu/WebGPUDriver.cpp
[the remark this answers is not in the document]
jneljneljnel: On second thought, I don't think we need a new device or adapter at all, we just don't ever create a surface and configure it with a device.

filament/backend/src/webgpu/WebGPUDriver.cpp
```diff
@@ -447,27 +447,19 @@ void WebGPUDriver::createSwapChainR(Handle<HwSwapChain> sch, void* nativeWindow,
...
-                  "about the selected graphics device, surface, etc. To see this try "
-                  "rebuilding Filament with that flag, e.g. ./build.sh -x "
-               << FWGPU_PRINT_SYSTEM << " ..." << utils::io::endl;
-#endif
 }
 
 void WebGPUDriver::createSwapChainHeadlessR(Handle<HwSwapChain> sch, uint32_t width,
-        uint32_t height, uint64_t flags) {}
+        uint32_t height, uint64_t flags) {
+     wgpu::Extent2D extent = { width, height};
```
jc3265, line 458: just a question, why are we calling it `extent` and not something like `surfaceSize`?

filament/backend/src/webgpu/WebGPUSwapChain.cpp
```diff
@@ -246,10 +254,39 @@ void WebGPUSwapChain::setExtent(wgpu::Extent2D const& currentSurfaceSize) {
     }
 }
 
-wgpu::TextureView WebGPUSwapChain::getCurrentSurfaceTextureView(
-        wgpu::Extent2D const& currentSurfaceSize) {
-    setExtent(currentSurfaceSize);
+wgpu::TextureView WebGPUSwapChain::getCurrentSurfaceTextureView( wgpu::Extent2D const& currentSurfaceSize, wgpu::Device& device) {
```
jc3265, line 257: nit: you can prob `const` the `device`

filament/backend/src/webgpu/WebGPUDriver.cpp
jneljneljnel: So, in the case of webgpu, when rendering in a headless context there simply is not surface, we don't use one or configure it, so calling it surfaceSize would be misleading. That does beg the question though if getCurrentSurfaceTextureView should be renamed.

filament/backend/src/webgpu/WebGPUSwapChain.cpp
jneljneljnel: Addressed by creating 3 on init and reusing them, left a todo for 2 vs 3 buffering,  Thanks!

PR #8671, opened 2025-04-27
webgpu: initial render pipeline creation

AndyHovingh:
⚠️  Note: at this stage we get a series of device errors about invalid index buffer writes, invalid texture sample type, invalid bind group creation, and thus invalid pipeline layout creation. But, in the interest of quickly integrating code and rapid iteration, we can merge asap and refactor/fix as needed. We still see color being written to the screen in blinking red, but only because I commented out the part where we actually set the (invalid) pipeline in the render pass encoder. Without commenting out that line we no longer see the blinking red. For Mac I see solid pink and Android I just see black. I believe this is due to undefined behavior when setting an invalid pipeline. This also needs render target information for getting the number of color targets and multisample count.

BUGS = [410598756]

### DIFF
--- a/filament/backend/src/webgpu/WebGPUDriver.cpp
+++ b/filament/backend/src/webgpu/WebGPUDriver.cpp
@@ -16,6 +16,7 @@
 
 #include "webgpu/WebGPUDriver.h"
 
+#include "WebGPUPipelineCreation.h"
 #include "WebGPUSwapChain.h"
 #include <backend/platforms/WebGPUPlatform.h>
 
@@ -24,18 +25,20 @@
 #include "private/backend/Dispatcher.h"
 #include <backend/DriverEnums.h>
 #include <backend/Handle.h>
+#include <backend/TargetBufferInfo.h>
 
 #include <math/mat3.h>
 #include <utils/CString.h>
+#include <utils/Panic.h>
 #include <utils/ostream.h>
 
 #include <dawn/webgpu_cpp_print.h>
 #include <webgpu/webgpu_cpp.h>
 
 #include <algorithm>
+#include <array>
 #include <cstddef>
 #include <cstdint>
-#include <memory>
 #include <sstream>
 #include <string_view>
 #include <utility>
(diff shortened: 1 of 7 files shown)


jc3265 (approved):
looks good to me! 
:shipit:

jneljneljnel (approved):
👍

filament/backend/src/webgpu/WebGPUDriver.cpp
```diff
@@ -859,6 +862,44 @@ void WebGPUDriver::blit(
 }
 
 void WebGPUDriver::bindPipeline(PipelineState const& pipelineState) {
+    const auto* program = handleCast<WGPUProgram>(pipelineState.program);
+    assert_invariant(program);
+    FILAMENT_CHECK_POSTCONDITION(program->computeShaderModule == nullptr)
+            << "WebGPU backend does not (yet) support compute pipelines.";
+    FILAMENT_CHECK_POSTCONDITION(program->vertexShaderModule)
+            << "WebGPU backend requires a vertex shader module for a render pipeline";
+    std::array<wgpu::BindGroupLayout, MAX_DESCRIPTOR_SET_COUNT> bindGroupLayouts{};
```
sidreesshah, line 871: nit: better to use std::vector here. bindGroupLayoutCount won't be need in that case. bindGroupLayouts.size() will do the job.

filament/backend/src/webgpu/WebGPUPipelineCreation.cpp
```diff
@@ -0,0 +1,254 @@
...
+        for (size_t targetIndex = 0; targetIndex < fragmentState.targetCount; targetIndex++) {
+            auto& colorTarget = colorTargets[targetIndex];
+            colorTarget.format = colorFormat;
+            colorTarget.blend = rasterState.hasBlending() ? &blendState : nullptr;
+            colorTarget.writeMask =
+                    rasterState.colorWrite ? wgpu::ColorWriteMask::All : wgpu::ColorWriteMask::None;
+        }
+        pipelineDescriptor.fragment = &fragmentState;
+    }
+    const wgpu::RenderPipeline pipeline = device.CreateRenderPipeline(&pipelineDescriptor);
```
sidreesshah, line 248: nit: const is not needed.

filament/backend/src/webgpu/WebGPUPipelineCreation.cpp
```diff
@@ -0,0 +1,254 @@
...
+    pipelineLabelStream << program.name.c_str() << " pipeline";
+    const auto pipelineLabel = pipelineLabelStream.str();
+    wgpu::RenderPipelineDescriptor pipelineDescriptor{
+        .label = wgpu::StringView(pipelineLabel),
+        .layout = layout,
+        .vertex = {
+            .module = program.vertexShaderModule,
+            .entryPoint = "main",
+            .constantCount = program.constants.size(),
+            .constants = program.constants.data(),
```
poweifeng, line 193: this maybe requires some investigation. For spec constants in vulkan, I had to transform c++ bool into a VK_TRUE/FALSE, which are 32-bit.

poweifeng (approved):
lgtm with some minor comments.

PR #8683, opened 2025-04-30
webgpu: fix command buffer (+encoder) life-cycle (avoids breaks on multiple render passes)

AndyHovingh:


bridgewaterrobbie (approved):
lgtm

PR #8686, opened 2025-05-01
Add flag to disable depth precaching

z3moon:
Depth precaching can cause initial startup stalls, particularly on platforms such as WebGL on FireFox/Mac.

This option allows us to disable depth precaching, which can be useful for eliminating the initial startup hiccups or stalls due to the upfront precaching of depth variants.

BUG=[392917621]


z3moon:
I tried adding it to Engine::features, but I found it didn't work because the checking logic hits first before we can alter the flag. (after the engine is created)

pixelflinger (approved):

PR #8687, opened 2025-05-01
Add debug messaging during OpenGL init phase for non-debug build

dkasuga:
Currently, in Filament's optimized builds, error-checking macros like `CHECK_GL_ERROR` are disabled, likely for performance or code size reasons. However, this poses a problem for developers who use Filament as a pre-built library. They find it difficult to detect initialization errors originating within Filament, even if their own application is a debug build.

This change proposes the addition of functionality to output warning messages when OpenGL errors occur during initialization, even in optimized (non-debug) builds. 

FIXES=[333130292]

### DIFF
--- a/filament/backend/src/opengl/GLUtils.h
+++ b/filament/backend/src/opengl/GLUtils.h
@@ -50,6 +50,12 @@ void assertFramebufferStatus(utils::io::ostream& out, GLenum target, const char*
 #   define CHECK_GL_FRAMEBUFFER_STATUS(out, target) { GLUtils::checkFramebufferStatus(out, target, __func__, __LINE__); }
 #endif
 
+#ifdef defined(NDEBUG) && defined(FILAMENT_ENABLE_INIT_GL_WARNINGS_FOR_OPTIMIZED_BUILD)
+#   define CHECK_GL_INIT_ERROR_FOR_OPTIMIZED_BUILD(out) { GLUtils::assertGLError(out, __func__, __LINE__);}
+#else
+#   define CHECK_GL_INIT_ERROR_FOR_OPTIMIZED_BUILD(out)
+#endif
+
 constexpr GLuint getComponentCount(ElementType const type) noexcept {
     using ElementType = ElementType;
     switch (type) {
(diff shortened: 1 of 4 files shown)


filament/backend/src/opengl/GLUtils.h
```diff
@@ -50,6 +50,12 @@ void assertFramebufferStatus(utils::io::ostream& out, GLenum target, const char*
 #   define CHECK_GL_FRAMEBUFFER_STATUS(out, target) { GLUtils::checkFramebufferStatus(out, target, __func__, __LINE__); }
 #endif
 
+#if defined(NDEBUG) && defined(FILAMENT_ENABLE_INIT_GL_WARNINGS_FOR_OPTIMIZED_BUILD)
+#   define CHECK_GL_INIT_ERROR_FOR_OPTIMIZED_BUILD(out) { GLUtils::assertGLError(out, __func__, __LINE__);}
```
pixelflinger, line 54: This will actually terminate the program, so it's more than just a "warning".

I think this is how this should work:
- on debug builds, we always call `assertGLError`
- on release builds, by default we do nothing unless `FILAMENT_ENABLE_INIT_GL_WARNINGS_FOR_OPTIMIZED_BUILD` is set, in which case we call `checkGLError` (i.e. we don't crash in release builds, just log).

pixelflinger (changes requested):

filament/backend/src/opengl/GLUtils.h
z3moon: I see that the new CHECK_GL_INIT_ERROR_FOR_OPTIMIZED_BUILD macro has a specific purpose and isn't intended as a direct replacement for CHECK_GL_ERROR.

I guess mathias meant to keep our current idiom, assert for debug, assuming this new macro can be a replacement for CHECK_GL_ERROR. However, if your intention is for this new macro to be an additional check used alongside CHECK_GL_ERROR, meaning users might need to call both, that could be unintuitive without detailed comments. So comprehensive comments detailing the specific use case may be needed.

Either way, I would consider making the new macro sound more generic, renaming it something like CHECK_GL_ERROR_FOR_RELEASE, indicating this macro is to govern the behavior in release mode.

pixelflinger:
this has been hanging for a while, I'm not sure we got consensus on how to do this. I'm closing this for now.

PR #8689, opened 2025-05-01
renderdiff: add golden repo support

poweifeng:
- Add GoldenManager to manage access to the repo containing the goldens
- Add tif comparison code
- Enable comparison by default for actual test

### DIFF
--- a/test/renderdiff/src/golden_manager.py
+++ b/test/renderdiff/src/golden_manager.py
@@ -1,0 +1,80 @@
+import os
+import shutil
+
+from utils import execute, ArgParseImpl
+
+GOLDENS_DIR = 'renderdiff'
+
+class GoldenManager:
+  def __init__(self, working_dir, access_token=None):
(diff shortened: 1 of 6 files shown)


test/renderdiff/src/golden_manager.py
```diff
@@ -0,0 +1,84 @@
...
+  def update(self):
+    self._git_exec('fetch')
+    self._git_exec('checkout main')
+    self._git_exec('rebase')
+
+  def _git_exec(self, cmd):
+    execute(f'git {cmd}', cwd=self._assets_dir(), capture_output=False)
+
+  # static method
+  def get_branch_name(ident):
```
bejado, line 40: Is this method used anywhere?

bejado (approved):

test/renderdiff/src/golden_manager.py
poweifeng: good catch. removed.

PR #8691, opened 2025-05-02
Switch filament and backend to Abseil logging

bejado:
This switches everything under `filament/` to Abseil logging. The following mapping between `utils::slog` and Abseil severity is used:

```
slog.d  ->  DLOG(INFO)
slog.i  ->  LOG(INFO)
slog.v  ->  LOG(INFO)
slog.w  ->  LOG(WARNING)
slog.e  ->  LOG(ERROR)
```


filament/backend/src/metal/MetalDriver.mm
```diff
@@ -78,20 +79,18 @@
...
+    DLOG(INFO) << "MetalVertexBuffer: " << sizeof(MetalVertexBuffer);
+    DLOG(INFO) << "MetalVertexBufferInfo: " << sizeof(MetalVertexBufferInfo);
+    DLOG(INFO) << "MetalIndexBuffer: " << sizeof(MetalIndexBuffer);
+    DLOG(INFO) << "MetalRenderPrimitive: " << sizeof(MetalRenderPrimitive);
+    DLOG(INFO) << "MetalTexture: " << sizeof(MetalTexture);
+    DLOG(INFO) << "MetalTimerQuery: " << sizeof(MetalTimerQuery);
+    DLOG(INFO) << "HwStream: " << sizeof(HwStream);
+    DLOG(INFO) << "MetalRenderTarget: " << sizeof(MetalRenderTarget);
+    DLOG(INFO) << "MetalFence: " << sizeof(MetalFence);
+    DLOG(INFO) << "MetalProgram: " << sizeof(MetalProgram);
```
pixelflinger, line 93: it wasn't possible to keep the line-breaks? Just curious?

pixelflinger:
Will you be able to check how this affect the binary size on a release build?

filament/backend/src/metal/MetalDriver.mm
bejado: Each `DLOG` creates a new line break automatically. If we add the `\n` ourselves, then only the first line will have the Abseil log metadata, which can look strange.

bejado:
> Will you be able to check how this affect the binary size on a release build?

Yes, I'm planning on check that. Unfortunately I think Abseil log uses std iostream, so we may see a binary size increase.

filament/backend/src/opengl/ShaderCompilerService.cpp
```diff
@@ -772,26 +774,25 @@ UTILS_NOINLINE
         } else {
             line = shader.substr(start, end - start);
         }
-        out << lc++ << ":   " << line.c_str() << '\n';
+        LOG(ERROR) << lc++ << ":   " << line.c_str();
```
poweifeng, line 777: I wonder if this should be a utils/std::stream, and then output to LOG(ERROR) at the end.

filament/src/ResourceList.cpp
```diff
@@ -32,7 +32,7 @@ ResourceListBase::ResourceListBase(const char* typeName)
 ResourceListBase::~ResourceListBase() noexcept {
 #ifndef NDEBUG
```
poweifeng, line 33: ok to remove `#ifndef` ?

filament/src/components/CameraManager.cpp
```diff
@@ -37,8 +38,7 @@ void FCameraManager::terminate(FEngine& engine) noexcept {
     auto& manager = mManager;
     if (!manager.empty()) {
 #ifndef NDEBUG
```
poweifeng, line 40: ok to remove `#ifndef`?

filament/src/components/RenderableManager.cpp
```diff
@@ -727,8 +731,8 @@ void FRenderableManager::terminate() noexcept {
     auto& manager = mManager;
     if (!manager.empty()) {
 #ifndef NDEBUG
```
poweifeng, line 733: remove `#ifndef` here?

filament/backend/src/opengl/ShaderCompilerService.cpp
bejado: Yeah, I thought about that. The difference is that with `LOG(ERROR)`, Abseil prepends metadata for each line. I guess we have to decide which style we prefer.

bejado:
Unfortunately with Abseil, I'm seeing a **1.23 MiB** binary size increase on Android. I'll need to investigate this some more to see what's going on.

 jni/arm64-v8a/libfilament-jni.so:

| Configuration | Size (bytes) |
|---------------|--------------|
| main branch   | 2380232    |
| With Abseil   | 3673832    |

libs/utils/include/utils/Logger.h
```diff
@@ -0,0 +1,118 @@
...
+public:
+    explicit LogLine(Stream& stream)
+        : mStream(stream) {}
+
+    LogLine(const LogLine&) = delete;
+    LogLine(LogLine&&) = delete;
+    LogLine& operator=(const LogLine&) = delete;
+    LogLine& operator=(LogLine&&) = delete;
+
+    ~LogLine() noexcept { mStream << utils::io::endl; }
```
poweifeng, line 70: The semantics here is a little different from the old since it'll be flushed when the LogLine goes out of scope, as oppose to when `<<utils::io::endl` is used in the code.

poweifeng: maybe that's ok?

bejado: Yeah, the semantics are indeed different. I did my best to find all the places we _weren't_ using `utils::io::endl` and instead use a `stringstream` first to concatenate. IMO the way we did it explicitly with `utils::io::endl` is a better API, but this is just how Abseil works. I really wish Abseil had a `LOG_NOENDLINE`-style macro, but it doesn't.

poweifeng (approved):
lgtm but web build seems to be failing

PR #8694, opened 2025-05-02
Update release.yml to include native android build files

loganmlawrence:
Bundle the native android build files and publish them with releases, so they can be used within native android applications.


poweifeng:
I think this is ok if it works, but we don't have any documentation on building with native libs and ndk afaict.

poweifeng (approved):

poweifeng:
@bejado do you have thoughts on this PR?

bejado:
@poweifeng I'm okay with it, though maintenance of these build artifacts should be best effort. I also wonder how large they are and if GitHub will complain about us uploading too much.

bejado (approved):

loganmlawrence:
Would this be considered an 'internal' PR? Or should I add a release note for it?

PR #8695, opened 2025-05-02
Backend test python script updated to display side by side images.

mhoff12358:


filament/backend/test/BackendTest.h
```diff
@@ -73,8 +76,13 @@ class BackendTest : public ::testing::Test {
 
     static bool matchesEnvironment(Backend backend);
     static bool matchesEnvironment(OperatingSystem operatingSystem);
-    static bool matchesEnvironment(OperatingSystem operatingSystem, Backend backend);
 private:
+    // Adds all the images that failed an ImageExpectation to the XML metadata for the current tests
+    // case. Add --gtest_output=xml as a command line argument to generate a test_detail.xml file in
```
bejado, line 81: Just curious, why XML and not JSON?

bejado (approved):

filament/backend/test/BackendTest.h
mhoff12358: IIUC XML is the default, you can get JSON but XML is the first class citizen. Which lines up with what gitlab's CI expected from unit tests. And parsing XML versus JSON in python is pretty much the same amount of work as the standard libraries do all the heavy lifting anyways.

That said, it's a very low-investment decision so if there's other context (such as other filament tools being very JSON based) it would be not much work to go the other way.

PR #8699, opened 2025-05-05
Add getter functions for settings to build ColorGrading object

dkasuga:
Currently, ColorGrading is created using a builder API, but there's no way to retrieve the settings that were used to build it after the fact. As the most simple and robust solution, this change make the `ColorGrading` object storing its creation parameters internally and providing their getter functions.

This change also involves changing changing the interface of the existing `ToneMapper` setter of the builder so that it takes tone mapper as `shared_ptr` instead of the raw pointer. For JNI function impl, to avoid the risk of the double free between C++ codes and JVM GC, the `ToneMapper` instance is just copied and passed.

FIXES=[364901658]

### DIFF
--- a/filament/src/details/ColorGrading.h
+++ b/filament/src/details/ColorGrading.h
@@ -51,6 +51,8 @@ private:
     uint32_t mDimension;
     bool mIsOneDimensional;
     bool mIsLDR;
+
+    void initializeSettings(Settings& settings, const Builder& builder) noexcept;
 };
 
 FILAMENT_DOWNCAST(ColorGrading)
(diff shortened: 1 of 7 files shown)


filament/src/details/ColorGrading.h
```diff
@@ -51,6 +51,8 @@ class FColorGrading : public ColorGrading {
     uint32_t mDimension;
     bool mIsOneDimensional;
     bool mIsLDR;
+
+    void initializeSettings(Settings& settings, const Builder& builder) noexcept;
```
poweifeng, line 55: static method? 

nit: can you make settings the second param? Typically the "output" is at the end of the list of parameters.

poweifeng (approved):
minor comments

filament/src/details/ColorGrading.h
dkasuga: Thanks for your suggestion, and I now incorporated that to my change.

PR #8702, opened 2025-05-06
webgpu: Webgpu/hello triangle v1

bridgewaterrobbie:
Currently contains the hacky commits to get HelloTriangle on WebGPU

### DIFF
--- a/filament/backend/include/backend/DriverEnums.h
+++ b/filament/backend/include/backend/DriverEnums.h
@@ -30,6 +30,7 @@
 #include <utils/compiler.h>
 #include <utils/debug.h>
 #include <utils/ostream.h>
+#include <utils/StaticString.h>
 
 #include <math/vec4.h>
 
@@ -1139,6 +1140,7 @@ struct ExternalSamplerDatum {
 static_assert(sizeof(ExternalSamplerDatum) == 12);
 
 struct DescriptorSetLayout {
+    std::variant<utils::StaticString, utils::CString, std::monostate> label;
     utils::FixedCapacityVector<DescriptorSetLayoutBinding> bindings;
 
 //  TODO: uncomment when needed
--- a/filament/backend/src/webgpu/WebGPUDriver.cpp
+++ b/filament/backend/src/webgpu/WebGPUDriver.cpp
@@ -13,7 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-
+#include <utils/Hash.h>
 #include "webgpu/WebGPUDriver.h"
 
 #include "WebGPUPipelineCreation.h"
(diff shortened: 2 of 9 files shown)


filament/src/details/Material.cpp
```diff
@@ -1140,13 +1140,30 @@ void FMaterial::precacheDepthVariants(FEngine& engine) {
...
     std::array<backend::DescriptorSetLayout, 2> descriptorSetLayout;
     success = parser->getDescriptorSetLayout(&descriptorSetLayout);
     assert_invariant(success);
 
+    // Setup Labels for debugging
+    utils::CString namestr;
+    utils::StaticString perViewStr = "_perView";
+    utils::StaticString singleStr = "_single";
+
+    parser->getName(&namestr);
```
bejado, line 1156: You should be able to use `mName` here, which is initialized inside the `FMaterial` constructor.

filament/src/details/Material.cpp
```diff
@@ -1140,13 +1140,30 @@ void FMaterial::precacheDepthVariants(FEngine& engine) {
...
+
+    utils::CString singleLabel(namestr.c_str(), namestr.length() + singleStr.length());
+    memccpy(singleLabel.c_str() + namestr.length(), singleStr.c_str(), '\0', singleStr.length());
```
bejado, line 1159: We should really define a `utils::CString::append` method in terms of `utils::CString::insert`. That would make this cleaner.

pixelflinger: I'd be happy if you did it first as a separate PR :-)

libs/filabridge/src/DescriptorSets.cpp
```diff
@@ -33,11 +33,13 @@ namespace filament::descriptor_sets {
 
 using namespace backend;
 
-static DescriptorSetLayout const postProcessDescriptorSetLayout{{
+static DescriptorSetLayout const postProcessDescriptorSetLayout{
+    utils::StaticString("postProcess"),{
```
pixelflinger, line 37: There shouldn't be a need to specify `utils::StaticString` here? It should implicitly be able to take an actual static string. This is actually an important feature. Is it not working? Let me know.

bridgewaterrobbie: I think it’s because I made the parameter a variant- a simple quoted string could be either a static string or a CString so the ambiguity needs to be solved.

I did it this way to try and be as “static as possible”, and even have the option of making the dynamic strings be debug-only without losing access to the strings that can be static.

filament/src/details/Material.cpp
```diff
@@ -1140,13 +1140,30 @@ void FMaterial::precacheDepthVariants(FEngine& engine) {
...
+
+    utils::CString perViewLabel(namestr.c_str(), namestr.length() + perViewStr.length());
+    memccpy(perViewLabel.c_str() + namestr.length(), perViewStr.c_str(), '\0', perViewStr.length());
+
+    descriptorSetLayout[0].label = singleLabel;
+    descriptorSetLayout[1].label = perViewLabel;
```
poweifeng, line 1165: I think if you rebased on top of main, the perView layouts are moved out of Material.

filament/backend/src/webgpu/WebGPUDriver.cpp
```diff
@@ -952,6 +960,7 @@ void WebGPUDriver::bindPipeline(PipelineState const& pipelineState) {
             *vertexBufferInfo, layout, pipelineState.rasterState, pipelineState.stencilState,
             pipelineState.polygonOffset, pipelineState.primitiveType, mSwapChain->getColorFormat(),
             mSwapChain->getDepthFormat());
+    mPipelineMap[hash] = pipeline;
     // TODO: uncomment once we have a valid pipeline to set
```
sidreesshah, line 964: this comment should be removed.

filament/src/details/Material.cpp
bridgewaterrobbie: Done in https://github.com/google/filament/pull/8707

filament/backend/include/backend/DriverEnums.h
```diff
@@ -39,6 +39,7 @@
 
 #include <stddef.h>
 #include <stdint.h>
+#include <utils/StaticString.h>
```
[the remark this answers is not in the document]
bridgewaterrobbie: Done

bejado (approved):

PR #8704, opened 2025-05-06
VK: Add a new platform method to check UMA

rafadevai:
This heuristic should work on almost all android
devices, can be extended later as needed.

On non android platforms the default is always false.

### DIFF
--- a/filament/backend/src/vulkan/platform/VulkanPlatform.cpp
+++ b/filament/backend/src/vulkan/platform/VulkanPlatform.cpp
@@ -634,10 +634,7 @@ fvkutils::VkFormatList findBlittableDepthStencilFormats(VkPhysicalDevice device)
 /**
  *  Check if the GPU has a unified memory architecture.
  */
-bool hasUnifiedMemoryArchitecture(VkPhysicalDevice device) noexcept {
-    VkPhysicalDeviceMemoryProperties memoryProperties;
-    vkGetPhysicalDeviceMemoryProperties(device, &memoryProperties);
-
+bool hasUnifiedMemoryArchitecture(VkPhysicalDeviceMemoryProperties memoryProperties) noexcept {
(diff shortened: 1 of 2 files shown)


filament/backend/src/vulkan/platform/VulkanPlatformAndroid.cpp
```diff
@@ -165,6 +165,21 @@ VulkanPlatformAndroid::ExternalImageVulkanAndroid::~ExternalImageVulkanAndroid()
     }
 }
 
+bool VulkanPlatformAndroid::isUnifiedMemoryArchitecture() const noexcept {
```
poweifeng, line 168: Might be better to store this as a boolean in the `VulkanContext`.  I'm just afraid future people will call it too often. 

And then `isUnifiedMemoryArchitecture` can be protected.

poweifeng: yes exactly

rafadevai: So I changed it to be more generic and query it for any platform.

poweifeng (approved):

filament/backend/src/vulkan/platform/VulkanPlatform.cpp
```diff
@@ -631,6 +631,24 @@ fvkutils::VkFormatList findBlittableDepthStencilFormats(VkPhysicalDevice device)
     return ret;
 }
 
+/**
+ *  Check if the GPU has a unified memory architecture.
+ */
+bool hasUnifiedMemoryArchitecture(VkPhysicalDevice device) noexcept {
+    VkPhysicalDeviceMemoryProperties memoryProperties;
```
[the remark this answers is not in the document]
rafadevai: Done

phoenixxxx (approved):

PR #8705, opened 2025-05-07
add sampler type in DescriptorSetLayout

pixelflinger:


filament/backend/include/backend/DriverEnums.h
```diff
@@ -258,29 +258,91 @@ enum class TextureType : uint8_t {
...
-    SAMPLER_INT,
-    SAMPLER_UINT,
-    SAMPLER_DEPTH,
-    SAMPLER_EXTERNAL,
-    UNIFORM_BUFFER,
-    SHADER_STORAGE_BUFFER,
-    INPUT_ATTACHMENT,
-};
+ enum class DescriptorType : uint8_t {
+     SAMPLER_2D_FLOAT,
```
bridgewaterrobbie, line 262: Would a “2D_BEGIN/END” style enum marker for these categories be reasonable? We could do it as a follow up from us, but it would make the code consuming this a tad nicer (though definitely not required)

pixelflinger: I don't want to add enums that are not "real", it makes the API weird, especially with IDEs, but I did add a `isSampler()` helper.

filament/backend/include/backend/DriverEnums.h
```diff
@@ -233,14 +237,114 @@ static inline constexpr bool hasShaderType(ShaderStageFlags flags, ShaderStage t
...
+enum class TextureType : uint8_t {
+    FLOAT,
+    INT,
+    UINT,
+    DEPTH,
+    STENCIL,
+    DEPTH_STENCIL
 };
 
+ constexpr std::string_view to_string(TextureType type) noexcept {
```
bejado, line 249: nit
```suggestion
constexpr std::string_view to_string(TextureType type) noexcept {
```
A bunch of `constexpr` functions in this file have a preceding space

filament/backend/include/backend/DriverEnums.h
```diff
@@ -171,9 +172,11 @@ static constexpr const char* backendToString(Backend backend) {
             return "Metal";
         case Backend::WEBGPU:
             return "WebGPU";
-        default:
-            return "Unknown";
+        case Backend::DEFAULT:
+            return "Default";
+            break;
```
bejado, line 177:
```suggestion
```

filament/backend/src/metal/MetalHandles.mm
```diff
@@ -1472,7 +1490,25 @@ static void func(void* user) {
...
+            case DescriptorType::SAMPLER_CUBE_INT:
+            case DescriptorType::SAMPLER_CUBE_UINT:
+            case DescriptorType::SAMPLER_CUBE_DEPTH:
+            case DescriptorType::SAMPLER_CUBE_ARRAY_FLOAT:
+            case DescriptorType::SAMPLER_CUBE_ARRAY_INT:
+            case DescriptorType::SAMPLER_CUBE_ARRAY_UINT:
+            case DescriptorType::SAMPLER_CUBE_ARRAY_DEPTH:
+            case DescriptorType::SAMPLER_3D_FLOAT:
+            case DescriptorType::SAMPLER_3D_INT:
+            case DescriptorType::SAMPLER_3D_UINT:
```
bejado, line 1511: Metal also needs to know whether or not the descriptor is a multisampled sampler (to distinguish between `MTLTextureType2D` and `MTLTextureType2DMultisample`). Is that another field we could add to `DescriptorSetLayoutBinding`? Perhaps a new flag in `DescriptorFlags`?

filament/backend/include/backend/DriverEnums.h
```diff
@@ -233,14 +237,114 @@ static inline constexpr bool hasShaderType(ShaderStageFlags flags, ShaderStage t
...
+         case DescriptorType::SAMPLER_2D_ARRAY_DEPTH:
+         case DescriptorType::SAMPLER_CUBE_DEPTH:
+         case DescriptorType::SAMPLER_CUBE_ARRAY_DEPTH:
+             return true;
+         default: ;
+     }
+     return false;
+ }
+
+ constexpr bool isFloatDescriptor(DescriptorType const type) noexcept {
```
jc3265, line 304: would it be helpful to add others like `isIntDescriptor` or `isUIntDescriptor`. We could also add it in a subsequent PR. it would help to do things like 
```
if is Int Descriptor
sampletype wgpu::TextureSampleType::Sint
fi
```

filament/backend/src/metal/MetalHandles.mm
bridgewaterrobbie: WebGPU also needs this bit too, sorry I was forgetting on initial review.

pixelflinger: that's currently not a problem because we don't support MS textures since it's not supported in GLES3.0.  I don't remember if the materials can generate them. Regardless good feedback, but for another PR. Thanks.

filament/backend/include/backend/DriverEnums.h
pixelflinger: I think it's fine to add those when we need them. Feel free to do that if/when you need it.

filament/src/details/Texture.cpp
```diff
@@ -274,7 +274,9 @@ Texture* Texture::Builder::build(Engine& engine) {
 
 // ------------------------------------------------------------------------------------------------
 
-FTexture::FTexture(FEngine& engine, const Builder& builder) {
+FTexture::FTexture(FEngine& engine, const Builder& builder)
+    : mHasBlitSrc(false),
+      mExternal(false) {
```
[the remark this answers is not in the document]
pixelflinger: just a habit because those are bitfield and can't have default initializer in the header.

libs/filabridge/src/DescriptorSets.cpp
```diff
@@ -51,32 +58,88 @@ static constexpr std::initializer_list<DescriptorSetLayoutBinding> depthVariantD
...
-    { DescriptorType::SAMPLER,                                   ShaderStageFlags::FRAGMENT,  +PerViewBindingPoints::FOG            },
+    { DescriptorType::SAMPLER_2D_FLOAT,                          ShaderStageFlags::FRAGMENT,  +PerViewBindingPoints::STRUCTURE      },
+    { DescriptorType::SAMPLER_2D_ARRAY_DEPTH,                    ShaderStageFlags::FRAGMENT,  +PerViewBindingPoints::SHADOW_MAP     },
+    { DescriptorType::SAMPLER_2D_FLOAT,                          ShaderStageFlags::FRAGMENT,  +PerViewBindingPoints::IBL_DFG_LUT    },
+    { DescriptorType::SAMPLER_CUBE_FLOAT,                        ShaderStageFlags::FRAGMENT,  +PerViewBindingPoints::IBL_SPECULAR   },
+    { DescriptorType::SAMPLER_2D_ARRAY_FLOAT,                    ShaderStageFlags::FRAGMENT,  +PerViewBindingPoints::SSAO           },
+    { DescriptorType::SAMPLER_2D_ARRAY_FLOAT,                    ShaderStageFlags::FRAGMENT,  +PerViewBindingPoints::SSR            },
+    { DescriptorType::SAMPLER_2D_FLOAT,                          ShaderStageFlags::FRAGMENT,  +PerViewBindingPoints::FOG            },
```
bridgewaterrobbie, line 92: In my testing with hello triangle, this needs to be a `CUBE_FLOAT`. I'm not sure if the true fix is that simple, or if something more dynamic is necessary.

poweifeng (approved):

PR #8708, opened 2025-05-07
Add initialize method to PlatformMetal

bejado:
BUGS = [362991544]

### DIFF
--- a/filament/backend/include/backend/platforms/PlatformMetal.h
+++ b/filament/backend/include/backend/platforms/PlatformMetal.h
@@ -40,6 +40,22 @@ public:
     int getOSVersion() const noexcept override { return 0; }
 
     /**
+     * Optionally initializes the Metal platform by acquiring resources necessary for rendering.
+     *
+     * This method attempts to acquire a Metal device and command queue, returning true if both are
+     * successfully obtained, or false otherwise. Typically, these objects are acquired when
+     * the Metal backend is initialized. This method allows clients to check for their availability
+     * earlier.
+     *
+     * Calling initialize() is optional and safe to do so multiple times. After initialize() returns
+     * true, subsequent calls will continue to return true but have no effect.
+     *
+     * @returns true if the device and command queue have been successfully obtained; false
+     * otherwise.
+     */
+    bool initialize() noexcept;
+
+    /**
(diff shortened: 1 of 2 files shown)


filament/backend/include/backend/platforms/PlatformMetal.h
```diff
@@ -39,6 +39,22 @@ class PlatformMetal final : public Platform {
...
+     * the Metal backend is initialized. This method allows clients to check for their availability
+     * earlier.
+     *
+     * Calling initialize() is optional and safe to do so multiple times. After initialize() returns
+     * true, subsequent calls will continue to return true but have no effect.
+     *
+     * @returns true if the device and command queue have been successfully obtained; false
+     * otherwise.
+     */
+    bool initialize() noexcept;
```
poweifeng, line 56: which thread do you expect the user to call this from? 

For vk, the platform methods are meant to be called from the backend thread.  But this seems like a client-facing API?

pixelflinger: yes, and on top of that the method doesn't look thread safe, so we should definitely specify the threading model for this method in the doc.

filament/backend/src/metal/PlatformMetal.mm
```diff
@@ -48,7 +49,29 @@
...
+    MetalDevice device{};
+    createDevice(device);
+    if (device.device == nil) {
+        return false;
+    }
+
+    MetalCommandQueue commandQueue{};
+    createCommandQueue(device, commandQueue);
+    if (commandQueue.commandQueue == nil) {
+        return false;
```
pixelflinger, line 63: wouldn't we leak the MetalDevice here?

bejado: No, the device is stored inside `pImpl->mDevice`. It will be released when `PlatformMetal` is destroyed.

filament/backend/include/backend/platforms/PlatformMetal.h
bejado: Yeah, great points, definitely missed this.

poweifeng (approved):

PR #8709, opened 2025-05-07
renderdiff: script for updating golden images

poweifeng:
Adding a python script to enable updating new goldens into a staging branch in the golden repo (filament-assets).

The same script can be used in github workflow to automatically create a golden staging branch. This will be useful for users without access to a mac (the only platform for generating goldens as of now).

### DIFF
--- a/test/renderdiff/src/golden_manager.py
+++ b/test/renderdiff/src/golden_manager.py
@@ -1,14 +1,42 @@
+# Copyright (C) 2025 The Android Open Source Project
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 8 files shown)


test/renderdiff/src/golden_manager.py
```diff
@@ -41,29 +89,44 @@ def merge_to_main(self, branch, push_to_remote=False):
...
     # Force create the branch (note will overwrite the old branch)
     self._git_exec(f'switch -C {branch}')
     rdiff_dir = os.path.join(assets_dir, GOLDENS_DIR)
-    execute(f'rm -rf {rdiff_dir}')
-    execute(f'mkdir -p {rdiff_dir}')
-    shutil.copytree(src_dir, rdiff_dir, dirs_exist_ok=True)
-    self._git_exec(f'add {GOLDENS_DIR}')
+    if len(updates) == 0 and len(deletes) == 0:
+      execute(f'rm -rf {rdiff_dir}')
+      execute(f'mkdir -p {rdiff_dir}')
```
bejado, line 106: One suggestion is to use `pathlib` so this isn't platform-dependent.

PR #8712, opened 2025-05-08
webgpu: debug descriptor sets option/functionality

AndyHovingh:


AndyHovingh:
At this point, this is stale enough and not in demand that I will close it. If we need it in the future, we will re-think it then (if ever).

PR #8713, opened 2025-05-08
github: fix release mac build

poweifeng:


poweifeng:
Bypassing checks to unblock release.

PR #8715, opened 2025-05-08
webgpu: TextureView Aspect

jneljneljnel:
Picks an appropriate aspect based on format and usage.

### DIFF
--- a/filament/backend/src/webgpu/WebGPUHandles.cpp
+++ b/filament/backend/src/webgpu/WebGPUHandles.cpp
@@ -857,32 +857,31 @@ wgpu::TextureFormat WGPUTexture::fToWGPUTextureFormat(const TextureFormat& fUsag
 wgpu::TextureAspect WGPUTexture::inferWGPUTextureViewAspect(const wgpu::TextureFormat& format,
         const wgpu::TextureUsage& usage) {
 
-    if (format == wgpu::TextureFormat::Depth32Float ||
-        format == wgpu::TextureFormat::Depth24Plus ||
-        format == wgpu::TextureFormat::Depth16Unorm) {
-        if (static_cast<uint32_t>(usage) & static_cast<uint32_t>(wgpu::TextureUsage::RenderAttachment) ||
-            static_cast<uint32_t>(usage) & static_cast<uint32_t>(wgpu::TextureUsage::TextureBinding) ||
-            static_cast<uint32_t>(usage) & static_cast<uint32_t>(wgpu::TextureUsage::StorageBinding)) {
+    if (format == wgpu::TextureFormat::Depth32Float || format == wgpu::TextureFormat::Depth24Plus ||
+            format == wgpu::TextureFormat::Depth16Unorm) {
+        if (static_cast<uint32_t>(usage) &
+                        static_cast<uint32_t>(wgpu::TextureUsage::RenderAttachment) ||
+                static_cast<uint32_t>(usage) &
+                        static_cast<uint32_t>(wgpu::TextureUsage::TextureBinding) ||
+                static_cast<uint32_t>(usage) &
+                        static_cast<uint32_t>(wgpu::TextureUsage::StorageBinding)) {
             return wgpu::TextureAspect::DepthOnly;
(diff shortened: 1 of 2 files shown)


filament/backend/src/webgpu/WebGPUHandles.cpp
```diff
@@ -854,24 +854,153 @@ wgpu::TextureFormat WGPUTexture::fToWGPUTextureFormat(const TextureFormat& fUsag
     }
 }
 
+wgpu::TextureAspect WGPUTexture::inferWGPUTextureViewAspect(const wgpu::TextureFormat& format,
```
bridgewaterrobbie, line 857: IMO for brevity this should be some if-statements to determine if we are returning anything other than `ALL`, and a base return of ALL.

I'm not sure if a switch statement here is really ideal either

jneljneljnel: Debated this as well, happy to change it.

filament/backend/src/webgpu/WebGPUHandles.cpp
```diff
@@ -854,24 +854,58 @@ wgpu::TextureFormat WGPUTexture::fToWGPUTextureFormat(const TextureFormat& fUsag
...
+        }
+        return wgpu::TextureAspect::StencilOnly;
+    }
+
+    if (format == wgpu::TextureFormat::Depth24PlusStencil8 ||
+        format == wgpu::TextureFormat::Depth32FloatStencil8) {
+        if (static_cast<uint32_t>(usage) & static_cast<uint32_t>(wgpu::TextureUsage::TextureBinding)) {
+            return wgpu::TextureAspect::DepthOnly;
+        }
+        return wgpu::TextureAspect::All;
```
bridgewaterrobbie, line 888: No need for the repeated `return wgpu::TextureAspect::All` statements for most(maybe all) of these, especially the ones with multiple other ifs to get t here. If you don't enter one of the internal `if` blocks returning something else,  you won't enter any of the other `format == wgpu::TextureFormat` blocks and hit the bottom one instead.

filament/backend/src/webgpu/WebGPUHandles.cpp
```diff
@@ -854,24 +854,58 @@ wgpu::TextureFormat WGPUTexture::fToWGPUTextureFormat(const TextureFormat& fUsag
...
+        const wgpu::TextureUsage& usage) {
+
+    if (format == wgpu::TextureFormat::Depth32Float ||
+        format == wgpu::TextureFormat::Depth24Plus ||
+        format == wgpu::TextureFormat::Depth16Unorm) {
+        if (static_cast<uint32_t>(usage) & static_cast<uint32_t>(wgpu::TextureUsage::RenderAttachment) ||
+            static_cast<uint32_t>(usage) & static_cast<uint32_t>(wgpu::TextureUsage::TextureBinding) ||
+            static_cast<uint32_t>(usage) & static_cast<uint32_t>(wgpu::TextureUsage::StorageBinding)) {
+            return wgpu::TextureAspect::DepthOnly;
+        } else if (static_cast<uint32_t>(usage) & (static_cast<uint32_t>(wgpu::TextureUsage::CopySrc) | static_cast<uint32_t>(wgpu::TextureUsage::CopyDst))) {
```
bridgewaterrobbie, line 867: I think this line is too long, did you run formatting on the overall codeblock?

filament/backend/src/webgpu/WebGPUHandles.cpp
jneljneljnel: I felt like highlighting the special cases opposed to leaving a comment. I'm a little worried how much we're deferring to All and a case that's beyond our current understanding that needs to be handled will be hard to find.... but it's probably fine.

filament/backend/src/webgpu/WebGPUHandles.cpp
```diff
@@ -854,24 +854,56 @@ wgpu::TextureFormat WGPUTexture::fToWGPUTextureFormat(const TextureFormat& fUsag
     }
 }
 
+wgpu::TextureAspect WGPUTexture::inferWGPUTextureViewAspect(const wgpu::TextureFormat& format,
```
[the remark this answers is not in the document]
jneljneljnel: I wanted to make this public and use wgpu values as an easy way to get it to the headless textures, There seems to be more info from filament. Thanks

PR #8723, opened 2025-05-12
Add defines to strip unused default materials

z3moon:
Add defines for trimming default materials. These enable binary size reduction by omitting them. Referencing the trimmed material will cause associated renderables to be skipped from rendering.

BUG=[415840223]

### DIFF
--- a/filament/src/components/RenderableManager.cpp
+++ b/filament/src/components/RenderableManager.cpp
@@ -468,12 +468,16 @@ RenderableManager::Builder::Result RenderableManager::Builder::build(Engine& eng
         auto& entry = mImpl->mEntries[i];
 
         // entry.materialInstance must be set to something even if indices/vertices are null
-        FMaterial const* material;
-        if (!entry.materialInstance) {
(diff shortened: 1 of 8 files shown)


filament/src/components/RenderableManager.cpp
```diff
@@ -468,12 +468,16 @@ RenderableManager::Builder::Result RenderableManager::Builder::build(Engine& eng
         auto& entry = mImpl->mEntries[i];
 
         // entry.materialInstance must be set to something even if indices/vertices are null
```
bejado, line 470: Is this comment still valid? If `FILAMENT_TRIM_DEFAULT_MATERIAL` is set then `entry.materialInstance` could end up `null`. Will that cause a crash later on?

z3moon: I believe this will be handled in FRenderPrimitive::init in filament/src/RenderPrimitive.cpp. (above)

But the comment needs to be updated accordingly.

PR #8724, opened 2025-05-12
vk: Clean up image layout - transitions and definitions

poweifeng:
We rename and modify the mapping form the backend layouts to VK version of the layouts.  We fix some of the barrier definitions which were inaccurate.

Also fix and clarify some of the VulkanBuffer barrier related details.


phoenixxxx (approved):
This looks good to me (the logic makes sense, and I also say that several changes were related to the layout of the text), with the caveat that I just recently started looking deeper into the staging and memory architecture of the VK backend.

PR #8728, opened 2025-05-13
re-do "Use the Perfetto SDK instead of ATRACE" (#8701)

pixelflinger:
re-do "Use the Perfetto SDK instead of ATRACE" (#8701)

This time we create an entirely new private header: Tracing.h which
uses the perfetto SDK instead of ATRACE. The old Systrace.h is
unchanged to presever backward compatibility but is essentially
deprecated and no longer used within the filament repo.

The new TRACING_ macros use an explicit CATEGORY parameter, which is
declared in Tracing.h.

Moreover, tracing can be compiled out by defining 
FILAMENT_TRACING_ENABLED to false before including Tracing.h

iOS tracing is still supported and still controlled via 
FILAMENT_APPLE_SYSTRACE.

There are three perfetto categories defined:

- "filament/filament"
- "filament/jobsystem"
- "filament/gltfio"

The "filament/jobsystem" category is compiled out by default.

And they can be enabled in AGI / perfetto by adding:

```
data_sources {
  config {
    name: "track_event"
    track_event_config {
      disabled_categories: "*"
      enabled_categories: "filament/filament"
      enabled_categories: "filament/jobsystem"
      enabled_categories: "filament/gltfio"
    }
  }
}
```


libs/utils/include/private/utils/Tracing.h
```diff
@@ -0,0 +1,52 @@
+/*
+ * Copyright (C) 2016 The Android Open Source Project
```
poweifeng, line 2:
```suggestion
 * Copyright (C) 2025 The Android Open Source Project
```

libs/utils/src/android/Tracing.cpp
```diff
@@ -0,0 +1,39 @@
+/*
+ * Copyright (C) 2012 The Android Open Source Project
```
poweifeng, line 2:
```suggestion
 * Copyright (C) 2025 The Android Open Source Project
```

filament/backend/src/CommandStream.cpp
```diff
@@ -79,12 +80,12 @@ CommandStream::CommandStream(Driver& driver, CircularBuffer& buffer) noexcept
...
 
 void CommandStream::execute(void* buffer) {
-    // NOTE: we can't use SYSTRACE_CALL() or similar here because, execute() below, also
+    // NOTE: we can't use FILAMENT_TRACING_CALL() or similar here because, execute() below, also
     // uses systrace BEGIN/END and the END is not guaranteed to be happening in this scope.
 
     Profiler profiler;
 
-    if (SYSTRACE_TAG) {
+    if constexpr (SYSTRACE_TAG) {
```
poweifeng, line 88: maybe we don't use `SYSTRACE_TAG` here to avoid mixing of tracing APIs?

poweifeng: and below as well.

poweifeng (commented):
If you decide to avoid `SYSTRACE_TAG`, I think there's one more reference in `filament/CMakeLists.txt` and `filament/backend/CMakeLists.txt`

libs/utils/src/JobSystem.cpp
```diff
@@ -14,23 +14,21 @@
  * limitations under the License.
  */
 
-// Note: The overhead of SYSTRACE_TAG_JOBSYSTEM is not negligible especially with parallel_for().
-#ifndef SYSTRACE_TAG
-//#define SYSTRACE_TAG SYSTRACE_TAG_JOBSYSTEM
-#define SYSTRACE_TAG SYSTRACE_TAG_NEVER
-#endif
+// Note: The overhead of TRACING is not negligible especially with parallel_for().
+#define FILAMENT_TRACING_ENABLED false
```
poweifeng, line 18: just curious, the choice of `false` over `0` is intentional?

poweifeng (approved):

libs/utils/src/JobSystem.cpp
pixelflinger: not really, does it matter?

PR #8729, opened 2025-05-13
Move calls to SPV remapper to another .cpp file

dneto0:


dneto0:
This is required before updating `third_party/glslang`.

* Glslang recently changed from using spirv.hpp to spirv.hpp11, and SPVRemap.h exposes that in its interface.
* SPIRV-Cross exposes spirv.hpp from its spirv_glsl.hpp
* Filaments libs/filamat/src/GLSLPostprocess.cpp includes both
* But those two spirv headers can't be compiled together: they use the same include guard, and have incompatible definitions of enums.

Separating the call to SPVRemap prevents both spirv headers from colliding.

Tracking with internal bug crbug.com/417365277

poweifeng (approved):

pixelflinger (approved):

PR #8735, opened 2025-05-14
Webgpu: Use WebGPU target rather than rely on Vulkan similarities

bridgewaterrobbie:
BUGS=[2F417598567]

### DIFF
--- a/filament/backend/test/ShaderGenerator.cpp
+++ b/filament/backend/test/ShaderGenerator.cpp
@@ -118,7 +118,7 @@ ShaderGenerator::Blob ShaderGenerator::transpileShader(ShaderStage stage, std::s
     } else if (backend == Backend::VULKAN) {
         shader.insert(pos, "#define TARGET_VULKAN_ENVIRONMENT\n");
     } else if (backend == Backend::WEBGPU) {
-        shader.insert(pos, "#define TARGET_VULKAN_ENVIRONMENT\n");
+        shader.insert(pos, "#define TARGET_WEBGPU_ENVIRONMENT\n");
     }
 
     const char* shaderCString = shader.c_str();
--- a/filament/backend/test/test_LoadImage.cpp
+++ b/filament/backend/test/test_LoadImage.cpp
@@ -52,7 +52,7 @@ layout(location = 0, set = 0) uniform {samplerType} test_tex;
 void main() {
     vec2 fbsize = vec2(512);
     vec2 uv = gl_FragCoord.xy / fbsize;
-#if defined(TARGET_METAL_ENVIRONMENT) || defined(TARGET_VULKAN_ENVIRONMENT)
+#if defined(TARGET_METAL_ENVIRONMENT) || defined(TARGET_VULKAN_ENVIRONMENT) || defined(TARGET_WEBGPU_ENVIRONMENT)
     uv.y = 1.0 - uv.y;
 #endif
     fragColor = vec4(texture(test_tex, uv).rgb, 1.0f);
@@ -73,7 +73,7 @@ float getLayer(in sampler2DArray s) { return 2.0f; }
 void main() {
     vec2 fbsize = vec2(512);
     vec2 uv = gl_FragCoord.xy / fbsize;
-#if defined(TARGET_METAL_ENVIRONMENT) || defined(TARGET_VULKAN_ENVIRONMENT)
+#if defined(TARGET_METAL_ENVIRONMENT) || defined(TARGET_VULKAN_ENVIRONMENT) || defined(TARGET_WEBGPU_ENVIRONMENT)
     uv.y = 1.0 - uv.y;
 #endif
     fragColor = vec4(texture(test_tex, vec3(uv, getLayer(test_tex))).rgb, 1.0f);
@@ -91,7 +91,7 @@ layout(location = 0, set = 0) uniform sampler2D test_tex;
 void main() {
     vec2 fbsize = vec2(512);
     vec2 uv = gl_FragCoord.xy / fbsize;
-#if defined(TARGET_METAL_ENVIRONMENT) || defined(TARGET_VULKAN_ENVIRONMENT)
+#if defined(TARGET_METAL_ENVIRONMENT) || defined(TARGET_VULKAN_ENVIRONMENT) || defined(TARGET_WEBGPU_ENVIRONMENT)
     uv.y = 1.0 - uv.y;
 #endif
     fragColor = vec4(textureLod(test_tex, uv, 1.0f).rgb, 1.0f);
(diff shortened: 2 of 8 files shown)


libs/filamat/src/shaders/CodeGenerator.cpp
```diff
@@ -177,7 +177,6 @@ utils::io::sstream& CodeGenerator::generateCommonProlog(utils::io::sstream& out,
         case TargetApi::WEBGPU:
             //For now, no differences so inherit the same changes.
             // TODO Define a separte environment, OR relevant checks
```
kpiascik, line 179: remove the TODO

filament/backend/test/ShaderGenerator.cpp
```diff
@@ -118,7 +118,7 @@ ShaderGenerator::Blob ShaderGenerator::transpileShader(ShaderStage stage, std::s
     } else if (backend == Backend::VULKAN) {
         shader.insert(pos, "#define TARGET_VULKAN_ENVIRONMENT\n");
     } else if (backend == Backend::WEBGPU) {
-        shader.insert(pos, "#define TARGET_VULKAN_ENVIRONMENT\n");
+        shader.insert(pos, "#define TARGET_WEBGPU_ENVIRONMENT\n");
```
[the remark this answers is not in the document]
bridgewaterrobbie: Well I'm the one who did that hack >.> So understandable. To a large degree this is me cleaning up months old bringup-TODOs

filament/backend/test/test_LoadImage.cpp
```diff
@@ -52,7 +52,7 @@ layout(location = 0, set = 0) uniform {samplerType} test_tex;
 void main() {
     vec2 fbsize = vec2(512);
     vec2 uv = gl_FragCoord.xy / fbsize;
-#if defined(TARGET_METAL_ENVIRONMENT) || defined(TARGET_VULKAN_ENVIRONMENT)
+#if defined(TARGET_METAL_ENVIRONMENT) || defined(TARGET_VULKAN_ENVIRONMENT) || defined(TARGET_WEBGPU_ENVIRONMENT)
```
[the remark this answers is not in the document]
bridgewaterrobbie: Yeah I had to do a quick bit of research to ensure we are the same here, but it seems accurate

libs/filamat/src/shaders/CodeGenerator.cpp
```diff
@@ -176,8 +176,6 @@ utils::io::sstream& CodeGenerator::generateCommonProlog(utils::io::sstream& out,
         // TODO: Handle webgpu here
```
kpiascik, line 176: what about this TODO?

bridgewaterrobbie: This is a mostly tangential TODO- This PR is meant to tackle the shader generation defines of which platform. Given that this is a copy-paste of the metal and vulkan case, I presume it could be folded together but was branched out in case we learn differences as we get things up and running?

@jc3265 it looks like you authored this line but it was a few months ago. Was there a deeper motivation here?

kpiascik: yes

shaders/src/surface_getters.vs
```diff
@@ -24,7 +24,7 @@ float getObjectUserData() {
 #if __VERSION__ >= 300
 /** @public-api */
 int getVertexIndex() {
-#if defined(TARGET_METAL_ENVIRONMENT) || defined(TARGET_VULKAN_ENVIRONMENT)
+#if defined(TARGET_METAL_ENVIRONMENT) || defined(TARGET_VULKAN_ENVIRONMENT) || defined(TARGET_WEBGPU_ENVIRONMENT)
```
poweifeng, line 27: recommend not adding it because I don't know if this is supported in wgsl, and which would actually be the correct call.

shaders/src/surface_main.vs
```diff
@@ -246,8 +246,8 @@ void main() {
     position.y = -position.y;
 #endif
 
-#if !defined(TARGET_VULKAN_ENVIRONMENT) && !defined(TARGET_METAL_ENVIRONMENT)
-    // This is not needed in Vulkan or Metal because clipControl is always (1, 0)
+#if !defined(TARGET_VULKAN_ENVIRONMENT) && !defined(TARGET_METAL_ENVIRONMENT) && !defined(TARGET_WEBGPU_ENVIRONMENT)
```
poweifeng, line 249: recommend we delay adding this until we understand webgpu's clip control feature better (since it's not always available?).

shaders/src/surface_main.vs
```diff
@@ -8,7 +8,7 @@
 
 void main() {
 #if defined(FILAMENT_HAS_FEATURE_INSTANCING)
-#   if defined(TARGET_METAL_ENVIRONMENT) || defined(TARGET_VULKAN_ENVIRONMENT)
+#   if defined(TARGET_METAL_ENVIRONMENT) || defined(TARGET_VULKAN_ENVIRONMENT) || defined(TARGET_WEBGPU_ENVIRONMENT)
```
poweifeng, line 11: are we sure about wgsl preferring `gl_InstanceIndex` vs. `gl_InstanceID`?

bridgewaterrobbie: I don't know enough to say for sure, but the `else` block in the current code notes that it is a workaround, which I think is less likely desired for WebGPU (Especially as it will be processed into WGSL by Tint). It is definitely an assumption to revisit once we have enough other bits up and running to test

poweifeng: My thinking is that the addition of something implies confidence, whereas absence (if wrong) is just oversight or being uncertain.  I think we're in the latter case here wrt this part (and some of the others I mentioned).

bridgewaterrobbie: Removing this also causes Tint to fail, so it is at least "More correct" for the WGSL conversion.

poweifeng: sounds good.  I'm ok with these addition as is.  Thank you for trying these combinations.

poweifeng (approved):
lgtm after resolution of clipControl question.

Please also wait for Ben or Mathias to approve.  Thank you!

libs/filamat/src/shaders/CodeGenerator.cpp
jc3265: I dont think so, I think the TODO was for us to revisit at a later date and make sure that we were not missing any cases? 
I believe we put it here back when we were first trying to get `matc` to work

bejado (approved):
This LGTM. FYI, I made a [diagram on coordinate spaces](https://github.com/google/filament/blob/main/art/diagrams/NDC.png) awhile back that might be helpful. Eventually we should probably update it for WebGPU.

PR #8739, opened 2025-05-15
Make sure that all descriptors are always initialized

pixelflinger:


bejado (approved):

poweifeng (approved):

filament/src/details/Engine.cpp
```diff
@@ -664,8 +672,12 @@ void FEngine::prepare() {
 
     for (auto& materialInstanceList: mMaterialInstances) {
         materialInstanceList.second.forEach([&driver](FMaterialInstance* item) {
-            item->commitStreamUniformAssociations(driver);
-            item->commit(driver);
+            // post-process materials isntances must be commited explicitly because their
```
poweifeng, line 675:
```suggestion
            // post-process materials instances must be commited explicitly because their
```

poweifeng (approved):

PR #8742, opened 2025-05-15
Switch to block-based stage-pool for Vulkan

anishmgoyal:
Instead of allocating a staging buffer every time one is needed, allocate a large (8mb) block of memory, and divvy it up as needed. We will make this configurable in the future, to allow for tuning for different apps as needed.


filament/backend/src/vulkan/VulkanCommands.h
```diff
@@ -76,6 +78,12 @@ struct VulkanCommandBuffer {
         mResources.push_back(resource);
     }
 
+    inline void trackStageBlock(std::unique_ptr<VulkanStage::Block> stageBlock) {
```
poweifeng, line 81: If you can make `VulkanStage::Block` a `fvkmemory::Resource`, then you can essentially just use `acquire` from the line above.

filament/backend/src/vulkan/VulkanStagePool.h
```diff
@@ -28,12 +28,80 @@ namespace filament::backend {
...
+        const uint32_t mCapacity;
+        const uint32_t mOffset;
+    };
+
+    inline VmaAllocation memory() const { return mMemory; }
+    inline VkBuffer buffer() const { return mBuffer; }
+    inline uint32_t capacity() const { return mCapacity; }
+    inline void* mapping() const { return mMapping; }
+
+    inline uint32_t& currentOffset() { return mCurrentOffset; }
```
poweifeng, line 80: why not just make this `uint32_t` and modify `mCurrentOffset` below in `acquireBlock`?

filament/backend/src/vulkan/VulkanStagePool.cpp
```diff
@@ -22,52 +22,88 @@
 #include "vulkan/utils/Conversion.h"
 #include "vulkan/utils/Image.h"
 
+#include <optional>
```
poweifeng, line 25: is this used?

filament/backend/src/vulkan/VulkanStagePool.cpp
```diff
@@ -22,52 +22,88 @@
...
 #include <utils/Panic.h>
 
 static constexpr uint32_t TIME_BEFORE_EVICTION = 3;
 
 namespace filament::backend {
 
+namespace {
+
+// Note: these are temporary values, they will be configurable.
+static constexpr uint32_t MAX_EMPTY_BLOCKS_TO_RETAIN = 1;
```
phoenixxxx, line 35: In "VulkanStagePool.cpp" we do "static constexpr uint32_t TIME_BEFORE_EVICTION = 3;" maybe we can keep these values here @poweifeng ?

filament/backend/src/vulkan/VulkanStagePool.h
phoenixxxx: yeah I would agree with that. Save the getter for larger (our external) functions that need "mCurrentOffset" like in "acquireStage" but maybe in "acquireBlock" just access "mCurrentOffset" directly?

filament/backend/src/vulkan/VulkanStagePool.h
```diff
@@ -51,9 +119,15 @@ class VulkanStagePool {
...
     VulkanStagePool(VmaAllocator allocator, VulkanCommands* commands);
 
-    // Finds or creates a stage whose capacity is at least the given number of bytes.
-    // The stage is automatically released back to the pool after TIME_BEFORE_EVICTION frames.
-    VulkanStage const* acquireStage(uint32_t numBytes);
+    // Finds or creates a stage block whose capacity is at least the given
+    // number of bytes. Internally, creates and manages and subdivides large
+    // buffers so that we have less objects around that we have to keep track
+    // of.
+    // This function is NOT thread-safe.
```
phoenixxxx, line 126: Do we need to worry about thread safety here? Is the backend ran on other threads? I'm not sure to be honest.

filament/backend/src/vulkan/VulkanStagePool.h
```diff
@@ -28,12 +28,80 @@ namespace filament::backend {
...
+        Block& operator=(const Block& other) = delete;
+        Block& operator=(Block&& other) = delete;
+
+        inline VulkanStage* parentStage() const { return mParentStage; }
+        inline VkBuffer buffer() const { return parentStage()->buffer(); }
+        inline VmaAllocation memory() const { return parentStage()->memory(); }
+        inline uint32_t capacity() const { return mCapacity; }
+        inline uint32_t offset() const { return mOffset; }
+
+        inline void* mapping() {
```
phoenixxxx, line 64: This is a neat pick but mapping to a uint8_t* ptr might be better (we don't need to cast the stage mapping to char*) and since your use mapping to std::memcpy you do not need to cast down to void* (I could be wrong on this?) but it makes the overall close clearner to my eye.

phoenixxxx (commented):
As we discussed offline, there will be a technical implementation document inbound for this feature for future reference.

filament/backend/src/vulkan/VulkanStagePool.h
```diff
@@ -28,12 +28,80 @@ namespace filament::backend {
...
+// into smaller buffers as needed.
+class VulkanStage {
+public:
+    VulkanStage(VmaAllocation memory, VkBuffer buffer, uint32_t capacity, void* mapping)
+        : mMemory(memory),
+          mBuffer(buffer),
+          mCapacity(capacity),
+          mMapping(mapping) {}
+
+    class Block {
```
rafadevai, line 41: To me I when I think about Block, I think about a huge chunk allocation that will be subdivided and Stage would be the smaller subdivision inside a block.

Should we change both names?

filament/backend/src/vulkan/VulkanStagePool.h
anishmgoyal: We don't, but the comment is here to be explicit about this assumption. I'd much prefer to have assumptions documented than not.

filament/backend/src/vulkan/VulkanStagePool.h
```diff
@@ -70,11 +144,21 @@ class VulkanStagePool {
...
+    // unmapping memory, freeing the allocation, and deleting the VulkanStage
+    // object. Note: takes an r-value because after this call, `stage` won't
+    // exist.
+    void destroyStage(VulkanStage const*&& stage);
+
+    // Use an ordered multimap for quick (capacity => stage) lookups using lower_bound().
+    std::multimap<uint32_t, VulkanStage*> mStages;
 
     std::unordered_set<VulkanStageImage const*> mFreeImages;
     std::vector<VulkanStageImage const*> mUsedImages;
```
rafadevai, line 164: Another thing we need to change is this, we can also tie this resource to the command buffer and then go through the same process as the buffers to be returned back to the pool.

it can be its own pr if you want

filament/backend/src/vulkan/VulkanStagePool.cpp
```diff
@@ -22,52 +22,88 @@
...
+constexpr uint32_t STAGE_SIZE = 1048576;
+
+}// namespace
+
 VulkanStagePool::VulkanStagePool(VmaAllocator allocator, VulkanCommands* commands)
     : mAllocator(allocator),
       mCommands(commands) {}
```
rafadevai, line 42: I wonder if we should preallocate blocks at the start, thoughts?

filament/backend/src/vulkan/VulkanStagePool.h
anishmgoyal: i'm a bit on the fence about this - i don't want to call the parent class "VulkanBlock", since that's not descriptive, and "VulkanStageBlock::Stage" feels rather redundant. i'd rather brainstorm a different name for Block - maybe Segment? "VulkanStage::Segment"?

filament/backend/src/vulkan/VulkanCommands.h
anishmgoyal: Updated.

filament/backend/src/vulkan/VulkanStagePool.h
anishmgoyal: Updated

filament/backend/src/vulkan/VulkanStagePool.h
anishmgoyal: Changed Block to Segment

filament/backend/src/vulkan/VulkanDriver.cpp
```diff
@@ -1223,7 +1223,7 @@ void VulkanDriver::updateIndexBuffer(Handle<HwIndexBuffer> ibh, BufferDescriptor
     VulkanCommandBuffer& commands = mCommands.get();
     auto ib = resource_ptr<VulkanIndexBuffer>::cast(&mResourceManager, ibh);
     commands.acquire(ib);
```
poweifeng, line 1225: nit: I think you can move the `acquire` to inside `loadFromCpu` to `acquire(this)`.

filament/backend/src/vulkan/VulkanStagePool.cpp
```diff
@@ -141,27 +194,34 @@ void VulkanStagePool::gc() noexcept {
     }
     const uint64_t evictionTime = mCurrentFrame - TIME_BEFORE_EVICTION;
 
-    // Destroy buffers that have not been used for several frames.
-    decltype(mFreeStages) freeStages;
-    freeStages.swap(mFreeStages);
+    decltype(mStages) freeStages;
+    freeStages.swap(mStages);
+    uint8_t freeStageCount = 0;  // Assuming we'll never have > 255 free stages
     for (auto pair : freeStages) {
```
poweifeng, line 200: If you want to move `pair.second` don't you want `auto&` here instead?  Otherwise you're moving a copy?

poweifeng: I think there's another place where `destroyStage` is called and also has the same pattern.

filament/backend/src/vulkan/VulkanStagePool.h
```diff
@@ -28,12 +31,97 @@ namespace filament::backend {
...
+
+        inline void* mapping() const {
+            return reinterpret_cast<void*>(
+                    reinterpret_cast<char*>(mParentStage->mapping()) + offset());
+        }
+
+    private:
+        // When the application terminates, the recycle function is no longer
+        // useful, and should be discarded to avoid potential memory faults.
+        inline void clearRecycleFn() const { mOnRecycleFn = nullptr; }
```
rafadevai, line 87: So I dont think we need to have this at all. 

When the program terminates everything should be destroyed properly without trying to access dangling pointers.

anishmgoyal: i saw segfaults without this

filament/backend/src/vulkan/VulkanStagePool.h
```diff
@@ -28,12 +31,97 @@ namespace filament::backend {
...
+
+        // Ensure parent class can access the terminate method.
+        friend class VulkanStage;
+
+        VulkanStage* const mParentStage;
+        const uint32_t mCapacity;
+        const uint32_t mOffset;
+        mutable OnRecycle mOnRecycleFn;
```
rafadevai, line 95: No need for mutable

anishmgoyal: right now, segments are marked as const within the parent stage, which i think is somewhat correct semantically, but i can change them to non-const and remove mutable here.

filament/backend/src/vulkan/VulkanStagePool.h
rafadevai: You need to move the stagePool terminate after the resourceManager terminate, that should fix it because all the resource needs to be destroyed first before destroy the pool objects.

anishmgoyal: Let's loop @poweifeng into this conversation. This was the solution suggested after conversation with @poweifeng.

poweifeng: So tldr; either option seems fine. Following is my opinion.

I'm not oppose to the reshuffle either, but that termination ordering is getting to be pretty complex because of the interdependencies between components.  If we can safely remove the dependence between components on shutdown, that's what I'd suggest.

rafadevai: To me the clean the recycle functions looks like big hack and it shows because your getting a segfault without it which is not what we expect to happen if everyhing is being reclaimed properly

rafadevai: So about the dependencies, if the class expect some Resources to be destroyed they need to terminated after the ResourceManager there's no other way around it

poweifeng: sure, I think that's a reasonable way to look at it.

Another is that these Segments don't actually hold anything but are really pointers to actual allocations.  The VulkanStagePool holds the allocations. So clearing these pointers before or after StagePool seems moot to me. 

Again, I'm ok either way, but maybe your interpretation is more logical and you can even assert that all the segments have been returned.

filament/backend/src/vulkan/VulkanStagePool.cpp
```diff
@@ -28,46 +28,99 @@ static constexpr uint32_t TIME_BEFORE_EVICTION = 3;
...
-VulkanStage const* VulkanStagePool::acquireStage(uint32_t numBytes) {
-    // First check if a stage exists whose capacity is greater than or equal to the requested size.
-    auto iter = mFreeStages.lower_bound(numBytes);
-    if (iter != mFreeStages.end()) {
-        auto stage = iter->second;
-        mFreeStages.erase(iter);
-        stage->lastAccessed = mCurrentFrame;
-        mUsedStages.push_back(stage);
-        return stage;
+fvkmemory::resource_ptr<VulkanStage::Segment> VulkanStagePool::acquireStage(uint32_t numBytes) {
```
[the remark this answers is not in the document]
phoenixxxx: inline VkPhysicalDeviceLimits const& getPhysicalDeviceLimits() const noexcept {
      return mPhysicalDeviceProperties.properties.limits;
  }

is available from driver::VulkanContext

filament/backend/src/vulkan/VulkanStagePool.cpp
anishmgoyal: oops - yeah, these should all be refs. updated

filament/backend/src/vulkan/VulkanStagePool.h
anishmgoyal: i think, after looking into this, i'm fine with one of a couple of solutions:

1. keeping the recycle fn
2. change the 'recycle function' to be a flag that marks the segments as 'invalid' when the parent goes out of scope
3. adding a lifecycle hook to resources that can be called when the ref count drops to zero (separate from the destructor), that can optionally be implemented - so long as it's clear that it's only to be used for low-overhead operations

i feel very resistant to the idea of reorganizing the terminate functions, as that would cause hard-to-debug issues if we ever try to hold resources within the staging buffer pool. while this seems unlikely, i'd like to guard against that, as it seems like there's a strong intent in the code for resources to be cleaned up last.

@rafadevai if you still feel very strongly that we should reorder the terminate functions, though, then we can do that. but i did want to voice my concern about potential future issues / lack of clarity before going that route.

filament/backend/src/vulkan/VulkanStagePool.cpp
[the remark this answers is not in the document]
anishmgoyal: it does sound like we'll likely want to align suballocation blocks to nonCoherentAtomSize. maybe it makes sense for me to add that in a follow-up pr? or i could add it to this one

filament/backend/src/vulkan/VulkanDriver.cpp
anishmgoyal: looking at it really quickly, it seems the object loadFromCpu is being called from isn't the resource_ptr, so i don't know that it would be correct to do it that way? unless the buffer has access to its own resource_ptr somewhere, and i haven't noticed

filament/backend/src/vulkan/VulkanTexture.cpp
```diff
@@ -480,31 +480,29 @@ void VulkanTexture::updateImage(const PixelBufferDescriptor& data, uint32_t widt
...
-    VulkanStage const* stage = mState->mStagePool.acquireStage(hostData->size);
-    assert_invariant(stage->memory);
-    vmaMapMemory(mState->mAllocator, stage->memory, &mapped);
-    memcpy(mapped, hostData->buffer, hostData->size);
-    vmaUnmapMemory(mState->mAllocator, stage->memory);
-    vmaFlushAllocation(mState->mAllocator, stage->memory, 0, hostData->size);
+    // Note: the following stageBlock must be stored within the command buffer
+    // before going out of scope, to ensure proper bookkeeping within the
+    // staging buffer pool.
+    fvkmemory::resource_ptr<VulkanStage::Segment> stage =
```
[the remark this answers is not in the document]
anishmgoyal: i went with stageSegment

filament/backend/src/vulkan/VulkanDriver.cpp
poweifeng: hmm, yeah, I think you're right.  ok as is.

filament/backend/src/vulkan/VulkanStagePool.h
rafadevai: Yes please change the order of the terminate approach. If you dont do that, there can be issues by not returning all the objects back to the pool. This can be easily checked that after terminate no resource has been leaked, so we can guarantee that all segments returned to the pool.

Then after that we are good to go to destroy the blocks.

filament/backend/src/vulkan/VulkanStagePool.cpp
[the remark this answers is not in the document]
rafadevai: Lets go with aligning the suballocation to the nonCoherentAtomSize
Add it do this one.

filament/backend/src/vulkan/VulkanStagePool.h
anishmgoyal: kk. i'll also add a warning comment above the terminate function that the stagepool should not be holding any resources by the time terminate is called, to avoid someone adding resources to stagepool in the future (which would cause segfaults).

filament/backend/src/vulkan/VulkanStagePool.cpp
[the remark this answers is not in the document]
anishmgoyal: Added.

poweifeng (approved):

phoenixxxx (approved):

PR #8744, opened 2025-05-15
webgpu: Ebridgewater/fix dawn hack

bridgewaterrobbie:
BUGS=["416091729"]
Revert our Dawn hack, and both handle sizes that are different than Dawn enforces, but also enforce offset rules which don't have a workaround (At least for now).

### DIFF
--- a/filament/backend/src/webgpu/WebGPUDriver.cpp
+++ b/filament/backend/src/webgpu/WebGPUDriver.cpp
@@ -721,17 +721,20 @@ size_t WebGPUDriver::getMaxArrayTextureLayers() {
 
 void WebGPUDriver::updateIndexBuffer(Handle<HwIndexBuffer> ibh, BufferDescriptor&& p,
         uint32_t byteOffset) {
-    updateGPUBuffer(handleCast<WGPUIndexBuffer>(ibh), std::move(p), byteOffset);
+    handleCast<WGPUIndexBuffer>(ibh)->updateGPUBuffer(p, byteOffset, mQueue);
+    scheduleDestroy(std::move(p));
 }
 
 void WebGPUDriver::updateBufferObject(Handle<HwBufferObject> ibh, BufferDescriptor&& p,
         uint32_t byteOffset) {
-    updateGPUBuffer(handleCast<WGPUBufferObject>(ibh), std::move(p), byteOffset);
+    handleCast<WGPUBufferObject>(ibh)->updateGPUBuffer(p, byteOffset, mQueue);
+    scheduleDestroy(std::move(p));
 }
 
 void WebGPUDriver::updateBufferObjectUnsynchronized(Handle<HwBufferObject> ibh,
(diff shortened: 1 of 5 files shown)


filament/backend/src/webgpu/WebGPUHandles.cpp
```diff
@@ -162,6 +153,49 @@ wgpu::StringView getUserTextureViewLabel(filament::backend::SamplerType target)
 
 namespace filament::backend {
 
+void WGPUBufferBase::createBuffer(const wgpu::Device& device, wgpu::BufferUsage usage,
+        uint32_t size, const char* label) {
+    // Write size must be divisible by 4. If the whole buffer is written to as is common, so must
+    // the buffer size.
```
jc3265, line 159: I think this comment needs to be updated

filament/backend/src/webgpu/WebGPUHandles.h
```diff
@@ -44,7 +44,7 @@ class WGPUProgram final : public HwProgram {
     std::vector<wgpu::ConstantEntry> constants;
 };
 
-struct WGPUBufferObject;
+class WGPUBufferObject;
```
jc3265, line 47: Whats the point of forward declaring this here?, we define it in line 110

filament/backend/src/webgpu/WebGPUHandles.cpp
bridgewaterrobbie: How so?

filament/backend/src/webgpu/WebGPUHandles.h
bridgewaterrobbie: It was there because WGPUBufferObject used to be defined in the driver, which I changed. You are right that it isn't necessary now, fixed and thank you

filament/backend/src/webgpu/WebGPUHandles.cpp
jc3265: nvm at first it didnt made sense but im assuming that is the same as 
`...if the whole buffer is written to (as is common) so must the buffer size.`
but you are right

filament/backend/src/webgpu/WebGPUDriver.cpp
```diff
@@ -1012,8 +1018,11 @@ void WebGPUDriver::updateDescriptorSetBuffer(Handle<HwDescriptorSet> dsh,
     auto buffer = handleCast<WGPUBufferObject>(boh);
     if (!bindGroup->getIsLocked()) {
         // TODO making assumptions that size and offset mean the same thing here.
+        FILAMENT_CHECK_PRECONDITION(offset % mMinUniformBufferOffsetAlignment == 0)
+                << "Binding offset must be multiple of " << mMinUniformBufferOffsetAlignment
```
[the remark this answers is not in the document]
bridgewaterrobbie: Will refine t his when I loop back next week for shadow copies (As needed)

PR #8749, opened 2025-05-16
enforce offsets to be a multiple of 4 bytes

pixelflinger:
this affects VerteBuffer, BufferObject and IndexBuffer, this restriction already existed in the Metal backend, it is now promoted to all backends.


poweifeng (approved):
Maybe also remove the now redundant log as you pointed out? ([ref](https://github.com/google/filament/blob/7638b3ddd9fe28527eab5fe283876e16739f7e72/filament/src/details/VertexBuffer.cpp#L105))

PR #8751, opened 2025-05-16
Add helpers for creating RenderPipelineParams for backend tests

mhoff12358:
BUGS=[398199600]

### DIFF
--- a/shaders/src/surface_light_punctual.fs
+++ b/shaders/src/surface_light_punctual.fs
@@ -155,15 +155,15 @@ Light getLight(const uint lightIndex) {
     light.channels = int(channels);
     light.contactShadows = bool(typeShadow & 0x10u);
 #if defined(VARIANT_HAS_DYNAMIC_LIGHTING)
-    light.type = (typeShadow & 0x1u);
+    light.lightType = (typeShadow & 0x1u);
 #if defined(VARIANT_HAS_SHADOWING)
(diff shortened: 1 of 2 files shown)


filament/backend/test/BackendTest.cpp
```diff
@@ -114,31 +115,38 @@ PipelineState BackendTest::getColorWritePipelineState() {
...
-    viewport.bottom = 0;
-    viewport.width = view.width;
-    viewport.height = view.height;
+filament::backend::RenderPassParams BackendTest::getClearColorRenderPass(
+        std::optional<float4> color) {
+    RenderPassParams params = {};
+    params.flags.clear = TargetBufferFlags::COLOR;
+    params.flags.discardStart = TargetBufferFlags::ALL;
+    params.flags.discardEnd = TargetBufferFlags::NONE;
+    params.clearColor = color.value_or(float4(0, 1, 1, 1));
```
bejado, line 134: Just curious, is there an advantage to using a `std::optional` here as opposed to just a default argument? It might be better to use a default argument so the default value is clearer to the users of the API.

bejado (approved):

filament/backend/test/BackendTest.cpp
mhoff12358: It's a rust pattern leaking in. The downside to a default value is that now your header contains an implementation detail. Which is pretty minor, you're correct that just using a default value is the right approach here.

PR #8757, opened 2025-05-19
VK: Introduce VulkanMemoryPool

rafadevai:
This class will allow better tracking of memory
allocations and recycling of buffers. Currently
only the uniform buffers are recycled.

It will eventually allow us to dynamically change
the underlying GPU buffer of a VulkanBuffer when
updating an UBO for uniforms and also keep track
which of those buffers are still inflight and which ones are ready to be reuse for an UBO.

Its the first step on moving towards by passing
the staging buffer in UMA.

### DIFF
--- a/filament/backend/src/vulkan/VulkanBuffer.cpp
+++ b/filament/backend/src/vulkan/VulkanBuffer.cpp
@@ -17,37 +17,18 @@
 #include "VulkanBuffer.h"
 #include "VulkanMemory.h"
 
-#include <utils/Panic.h>
-
 using namespace bluevk;
 
 namespace filament::backend {
 
 VulkanBuffer::VulkanBuffer(VmaAllocator allocator, VulkanStagePool& stagePool,
-        VkBufferUsageFlags usage, uint32_t numBytes)
+        VulkanMemoryPool& memoryPool, VulkanBufferUsage usage, uint32_t numBytes)
     : mAllocator(allocator),
       mStagePool(stagePool),
-      mUsage(usage),
+      mMemoryPool(memoryPool),
+      mGpuMemory(mMemoryPool.acquire(usage, numBytes)),
       mUpdatedOffset(0),
-      mUpdatedBytes(0) {
-    // for now make sure that only 1 bit is set in usage
-    // (because loadFromCpu() assumes that somewhat)
-    assert_invariant(usage && !(usage & (usage - 1)));
-
-    // Create the VkBuffer.
-    VkBufferCreateInfo bufferInfo {
-        .sType = VK_STRUCTURE_TYPE_BUFFER_CREATE_INFO,
-        .size = numBytes,
-        .usage = usage | VK_BUFFER_USAGE_TRANSFER_DST_BIT
-    };
-
-    VmaAllocationCreateInfo allocInfo { .usage = VMA_MEMORY_USAGE_GPU_ONLY };
-    vmaCreateBuffer(mAllocator, &bufferInfo, &allocInfo, &mGpuBuffer, &mGpuMemory, nullptr);
-}
-
-VulkanBuffer::~VulkanBuffer() {
-    vmaDestroyBuffer(mAllocator, mGpuBuffer, mGpuMemory);
-}
+      mUpdatedBytes(0) {}
 
 void VulkanBuffer::loadFromCpu(VkCommandBuffer cmdbuf, const void* cpuData, uint32_t byteOffset,
         uint32_t numBytes) {
@@ -64,13 +45,13 @@ void VulkanBuffer::loadFromCpu(VkCommandBuffer cmdbuf, const void* cpuData, uint
             (byteOffset >= mUpdatedOffset && byteOffset <= (mUpdatedOffset + mUpdatedBytes))) {
         VkAccessFlags srcAccess = 0;
         VkPipelineStageFlags srcStage = 0;
-        if (mUsage & VK_BUFFER_USAGE_UNIFORM_BUFFER_BIT) {
+        if (getUsage() == VulkanBufferUsage::UNIFORM) {
             srcAccess = VK_ACCESS_SHADER_READ_BIT;
             srcStage = VK_PIPELINE_STAGE_VERTEX_SHADER_BIT | VK_PIPELINE_STAGE_FRAGMENT_SHADER_BIT;
-        } else if (mUsage & VK_BUFFER_USAGE_VERTEX_BUFFER_BIT) {
+        } else if (getUsage() == VulkanBufferUsage::VERTEX) {
             srcAccess = VK_ACCESS_VERTEX_ATTRIBUTE_READ_BIT;
             srcStage = VK_PIPELINE_STAGE_VERTEX_INPUT_BIT;
-        } else if (mUsage & VK_BUFFER_USAGE_INDEX_BUFFER_BIT) {
+        } else if (getUsage() == VulkanBufferUsage::INDEX) {
             srcAccess = VK_ACCESS_INDEX_READ_BIT;
             srcStage = VK_PIPELINE_STAGE_VERTEX_INPUT_BIT;
         }
@@ -81,7 +62,7 @@ void VulkanBuffer::loadFromCpu(VkCommandBuffer cmdbuf, const void* cpuData, uint
             .dstAccessMask = VK_ACCESS_TRANSFER_WRITE_BIT,
             .srcQueueFamilyIndex = VK_QUEUE_FAMILY_IGNORED,
             .dstQueueFamilyIndex = VK_QUEUE_FAMILY_IGNORED,
-            .buffer = mGpuBuffer,
+            .buffer = getGpuBuffer(),
             .offset = byteOffset,
             .size = numBytes,
         };
@@ -94,7 +75,7 @@ void VulkanBuffer::loadFromCpu(VkCommandBuffer cmdbuf, const void* cpuData, uint
         .dstOffset = byteOffset,
         .size = numBytes,
     };
-    vkCmdCopyBuffer(cmdbuf, stage->buffer, mGpuBuffer, 1, &region);
+    vkCmdCopyBuffer(cmdbuf, stage->buffer, getGpuBuffer(), 1, &region);
 
     mUpdatedOffset = byteOffset;
     mUpdatedBytes = numBytes;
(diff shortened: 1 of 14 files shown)


filament/backend/src/vulkan/VulkanMemory.h
```diff
@@ -32,9 +32,55 @@
...
+
+#include <functional>
 
 VK_DEFINE_HANDLE(VmaAllocator)
 VK_DEFINE_HANDLE(VmaAllocation)
 VK_DEFINE_HANDLE(VmaPool)
 
+namespace filament::backend {
+
+enum class VulkanBufferUsage : uint8_t {
```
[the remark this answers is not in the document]
rafadevai: So the current idiom seems to be the usage is for only one thing, VERTEX, INDEX, ... and not mix them like (VERTEX and INDEX) , so im trying to make that more explicit.

I thought about using BufferObjectBinding but it doesnt have INDEX , any suggestion are welcome

filament/backend/src/vulkan/VulkanMemoryPool.cpp
```diff
@@ -0,0 +1,193 @@
...
+    // If this is one of the first few frames, return early to avoid wrapping unsigned integers.
+    constexpr uint32_t TIME_BEFORE_EVICTION = 3;
+    if (++mCurrentFrame <= TIME_BEFORE_EVICTION) {
+        return;
+    }
+    const uint64_t evictionTime = mCurrentFrame - TIME_BEFORE_EVICTION;
+
+    // Destroy buffers that have not been used for several frames.
+    for (auto iter = mFreeUniformAllocations.begin(); iter != mFreeUniformAllocations.end();) {
+        if (iter->second.lastAccessed < evictionTime) {
```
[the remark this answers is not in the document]
rafadevai: So this is to recycle, lets see we allocate a new VkBuffer that later gets destroyed with an ubo, a new ubo could reuse it that's what is this for

The buffers in this list are already free and not used anywhere

filament/backend/src/vulkan/VulkanMemoryPool.h
```diff
@@ -0,0 +1,76 @@
...
+#include "VulkanContext.h"
+#include "VulkanMemory.h"
+#include "memory/Resource.h"
+#include "memory/ResourceManager.h"
+
+#include <map>
+
+namespace filament::backend {
+
+class VulkanMemoryPool {
```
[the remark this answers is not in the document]
rafadevai: Updated the names, let me know what you think

poweifeng (approved):
some minor comments

filament/backend/src/vulkan/VulkanGpuBufferCache.cpp
```diff
@@ -0,0 +1,208 @@
...
+    VkBufferCreateInfo const bufferInfo{
+        .sType = VK_STRUCTURE_TYPE_BUFFER_CREATE_INFO,
+        .size = numBytes,
+        // `VK_BUFFER_USAGE_TRANSFER_DST_BIT` is needed to allow updating the buffer through
+        // a staging using `vkCmdCopyBuffer`.
+        .usage = getVkBufferUsage(usage) | VK_BUFFER_USAGE_TRANSFER_DST_BIT,
+    };
+
+    VmaAllocationCreateFlags vmaFlags = 0;
+    if (usage == VulkanBufferUsage::UNIFORM) {
```
phoenixxxx, line 142: As discussed offline I think this should rely on BufferUsage::DYNAMIC

filament/backend/src/vulkan/VulkanBuffer.cpp
```diff
@@ -17,37 +17,18 @@
...
 namespace filament::backend {
 
 VulkanBuffer::VulkanBuffer(VmaAllocator allocator, VulkanStagePool& stagePool,
-        VkBufferUsageFlags usage, uint32_t numBytes)
+        VulkanGpuBufferCache& gpuBufferCache, VulkanBufferUsage usage, uint32_t numBytes)
     : mAllocator(allocator),
       mStagePool(stagePool),
-      mUsage(usage),
+      mGpuBufferCache(gpuBufferCache),
+      mGpuBufferHolder(mGpuBufferCache.acquire(usage, numBytes)),
```
phoenixxxx, line 29: See later comments (this is for all buffers)

filament/backend/src/vulkan/VulkanGpuBufferCache.cpp
```diff
@@ -0,0 +1,208 @@
...
+        // In the case of UMA, the uniform buffers will always be mappable
+        if (mContext.isUnifiedMemoryArchitecture()) {
+            vmaFlags |= VMA_ALLOCATION_CREATE_MAPPED_BIT |
```
phoenixxxx, line 145: I also would make is explicit that his will not work for dynamic vertex buffers on UMA for example. Which is something we seem to imply we will support, since on the basis of VulkanBuffer::VulkanBuffer(...) we are using this method for all buffers.

Suggestions:
- Rework the logic
- Put an @todo for the dynamic non-Uniform buffer case

filament/backend/src/vulkan/VulkanGpuBufferCache.cpp
```diff
@@ -0,0 +1,208 @@
...
+    // recycled after being yielded.
+    VulkanGpuBuffer const* gpuBuffer = allocate(usage, numBytes);
+    return fvkmemory::resource_ptr<VulkanGpuBufferHolder>::construct(&mResourceManager, gpuBuffer,
+            [this](VulkanGpuBuffer const* gpuBuffer) { this->yield(gpuBuffer); });
+}
+
+void VulkanGpuBufferCache::yield(VulkanGpuBuffer const* gpuBuffer) noexcept {
+    if (!gpuBuffer) {
+        return;
+    }
```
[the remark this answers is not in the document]
phoenixxxx: agreed, I think an assert is more useful than a silent failure. Unless the logic expects gpuBuffer to ever be null.

filament/backend/src/vulkan/VulkanGpuBufferCache.cpp
```diff
@@ -0,0 +1,208 @@
...
+}
+
+VulkanGpuBuffer const* VulkanGpuBufferCache::allocate(VulkanBufferUsage usage,
+        uint32_t numBytes) noexcept {
+    VkBufferCreateInfo const bufferInfo{
+        .sType = VK_STRUCTURE_TYPE_BUFFER_CREATE_INFO,
+        .size = numBytes,
+        // `VK_BUFFER_USAGE_TRANSFER_DST_BIT` is needed to allow updating the buffer through
+        // a staging using `vkCmdCopyBuffer`.
+        .usage = getVkBufferUsage(usage) | VK_BUFFER_USAGE_TRANSFER_DST_BIT,
```
phoenixxxx, line 138: another issue here, for the case where we are not using the vkCmdCopyBuffer (UMA + GPU/GPU access).

filament/backend/src/vulkan/VulkanGpuBufferCache.cpp
```diff
@@ -0,0 +1,208 @@
...
+    }
+
+    // We were not able to find a sufficiently large allocation, so create a new one that is
+    // recycled after being yielded.
+    VulkanGpuBuffer const* gpuBuffer = allocate(usage, numBytes);
+    return fvkmemory::resource_ptr<VulkanGpuBufferHolder>::construct(&mResourceManager, gpuBuffer,
+            [this](VulkanGpuBuffer const* gpuBuffer) { this->yield(gpuBuffer); });
+}
+
+void VulkanGpuBufferCache::yield(VulkanGpuBuffer const* gpuBuffer) noexcept {
```
phoenixxxx, line 78: nit: In this code base I didn't see many instances of the acquire/yield naming convention. It's fine to do it here, but I would consider acquire/release.

filament/backend/src/vulkan/VulkanMemory.h
```diff
@@ -32,9 +32,55 @@
...
+
+struct VulkanGpuBuffer {
+    VkBuffer vkbuffer = VK_NULL_HANDLE;
+    VmaAllocation vmaAllocation = VK_NULL_HANDLE;
+    VmaAllocationInfo allocationInfo;
+    uint32_t numBytes = 0;
+    VulkanBufferUsage usage = VulkanBufferUsage::UNKNOWN;
+};
+
+struct VulkanGpuBufferHolder : public fvkmemory::Resource {
```
phoenixxxx, line 61: nite: Class? There is logic and public/private sections

filament/backend/src/vulkan/VulkanGpuBufferCache.cpp
rafadevai: I added a comment on the definition that in UMA uniform buffers (doesnt matter if its dynamic or not) will be host visible + device, everything else its device only.

So I dont see why we need to change anything here

filament/backend/src/vulkan/VulkanGpuBufferCache.cpp
rafadevai: done

filament/backend/src/vulkan/VulkanGpuBufferCache.cpp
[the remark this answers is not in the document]
rafadevai: Done

filament/backend/src/vulkan/VulkanGpuBufferCache.h
```diff
@@ -0,0 +1,83 @@
...
+    VulkanGpuBufferCache(const VulkanGpuBufferCache&) = delete;
+    VulkanGpuBufferCache& operator=(const VulkanGpuBufferCache&) = delete;
+
+    // Allocates or reuse a new VkBuffer that is device local.
+    // In the case of Unified memory architecture, uniform buffers are also host visible.
+    fvkmemory::resource_ptr<VulkanGpuBufferHolder> acquire(VulkanBufferUsage usage,
+            uint32_t numBytes) noexcept;
+
+    // Return a `VulkanGpuBuffer` back to its corresponding pool
+    void yield(VulkanGpuBuffer const* gpuBuffer) noexcept;
```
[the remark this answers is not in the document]
rafadevai: Yes, done

filament/backend/src/vulkan/memory/Resource.cpp
```diff
@@ -80,6 +81,9 @@ ResourceType getTypeEnum() noexcept {
     if constexpr (std::is_same_v<D, VulkanFence>) {
         return ResourceType::FENCE;
     }
+    if constexpr (std::is_same_v<D, VulkanGpuBufferHolder>) {
+        return ResourceType::VULKAN_GPU_BUFFER;
```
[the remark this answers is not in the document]
rafadevai: Done

phoenixxxx:
We should rename this PR now. We are not dealing with "memory" but "buffers" (as opposed to images) and we seem to have moved away from "pool" to "cache"?

rafadevai:
> We should rename this PR now. We are not dealing with "memory" but "buffers" (as opposed to images) and we seem to have moved away from "pool" to "cache"?

Done

phoenixxxx (approved):

PR #8760, opened 2025-05-20
webgpu: depthstencil fix for android

jneljneljnel:
Setting stencil loadop and storeop to undefined if stencil aspect is nonexistent on a depthStencilAttachment to adhear to a Dawn specification.

BUGS=[417598236]

### DIFF
--- a/filament/backend/src/webgpu/WebGPUDriver.cpp
+++ b/filament/backend/src/webgpu/WebGPUDriver.cpp
@@ -854,18 +854,31 @@ void WebGPUDriver::beginRenderPass(Handle<HwRenderTarget> rth, RenderPassParams 
     //     FWGPU_LOGW << "Non Default render target"
     //                << utils::io::endl;
     // }
+    wgpu::TextureView depthTextureView = mSwapChain->getDepthTextureView();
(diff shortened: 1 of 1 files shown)


filament/backend/src/webgpu/WebGPUDriver.cpp
```diff
@@ -792,26 +792,40 @@ void WebGPUDriver::beginRenderPass(Handle<HwRenderTarget> rth, RenderPassParams
     //     FWGPU_LOGW << "Non Default render target"
     //                << utils::io::endl;
     // }
+    wgpu::TextureView depthTextureView = mSwapChain->getDepthTextureView();
+    wgpu::TextureFormat depthTextureFormat = mSwapChain->getDepthFormat();
+    wgpu::LoadOp stencilLoadOpValue;
+    wgpu::StoreOp stencilStoreOpValue;
+
+    if (depthTextureFormat == wgpu::TextureFormat::Depth24PlusStencil8 ||
```
jc3265, line 800: the PR Description has the reasoning on why this is needed BUT it may be helpful to also have a comment on the code

PR #8765, opened 2025-05-20
In backend tests replace calls to draw with draw2.

mhoff12358:
BUGS=[398199600]

### DIFF
--- a/filament/backend/test/test_LoadImage.cpp
+++ b/filament/backend/test/test_LoadImage.cpp
@@ -25,6 +25,7 @@
 #include <backend/DriverEnums.h>
 #include <backend/Handle.h>
 
+#include "TrianglePrimitive.h"
 #include "private/filament/SamplerInterfaceBlock.h"
 
 #include <vector>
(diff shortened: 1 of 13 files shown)


filament/backend/test/test_LoadImage.cpp
```diff
@@ -546,8 +581,18 @@ TEST_F(LoadImageTest, UpdateImage3D) {
 
         api.bindDescriptorSet(descriptorSet, 0, {});
 
-        renderTriangle({ { shader.getDescriptorSetLayout() } },
-                defaultRenderTarget, swapChain, shader.getProgram());
+        RenderPassParams params = getClearColorRenderPass();
+        params.viewport.width = 512;
```
poweifeng, line 585: nit: I'd add a constant for 512.

poweifeng (approved):
Thank you!

filament/backend/test/test_LoadImage.cpp
mhoff12358: The tests are absolutely littered with 512s, I've actually got another note in my notebook to clean them all up.
So I'll take care of that next.

PR #8767, opened 2025-05-21
Add multiview MSAA support for custom swapchain

z3moon:
BUGS=[417311684]

### DIFF
--- a/filament/include/filament/Texture.h
+++ b/filament/include/filament/Texture.h
@@ -174,7 +174,13 @@ public:
 
         /**
          * Specifies the numbers of samples used for MSAA (Multisample Anti-Aliasing).
-         * If this is invoked for array textures, we assume this is used for multiview.
+         *
+         * Calling this method implicitly indicates the texture is used as a render target. Hence,
+         * this method should not be used in conjunction with other methods that are semantically
+         * conflicting like `setImage`.
+         *
+         * If this is invoked for array textures, it means this texture is used for multiview.
+         *
          * @param samples Number of samples for this texture.
          * @return This Builder, for chaining calls.
          */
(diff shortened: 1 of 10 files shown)


filament/include/filament/Texture.h
```diff
@@ -172,6 +172,14 @@ class UTILS_PUBLIC Texture : public FilamentAPI {
          */
         Builder& levels(uint8_t levels) noexcept;
 
+        /**
+         * Specifies the numbers of samples used for MSAA (Multisample Anti-Aliasing).
```
bejado, line 176: I think we need a bit more documentation here. For example, do we expect MSAA textures to work with `setImage`? Can these textures be bound to a material parameter? I think we should inform clients that MSAA textures are to only be used as render target attachments.

z3moon: Good point. Let me update this.

bejado (approved):
LGTM. I think we could consider checking programmatically that MSAA textures are only used as render targets, but that can wait.

filament/include/filament/Texture.h
```diff
@@ -172,6 +172,20 @@ class UTILS_PUBLIC Texture : public FilamentAPI {
...
+         *
+         * Calling this method implicitly indicates the texture is used as a render target. Hence,
+         * this method should not be used in conjunction with other methods that are semantically
+         * conflicting like `setImage`.
```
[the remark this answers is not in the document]
z3moon: I found it is already being checked in `setImage`. https://github.com/google/filament/blob/169c8d57ffdc6bb008231b1d509bfc09a47f0a91/filament/src/details/Texture.cpp#L379

PR #8782, opened 2025-05-27
Add checks for MSAA

z3moon:


filament/src/details/Texture.cpp
```diff
@@ -376,6 +383,9 @@ void FTexture::setImage(FEngine& engine, size_t const level,
     FILAMENT_CHECK_PRECONDITION(!mExternal)
             << "External Texture not supported for this operation.";
 
+    FILAMENT_CHECK_PRECONDITION(any(mUsage & Texture::Usage::UPLOADABLE))
+            << "Texture is not uploadable.";
```
poweifeng, line 387: Does this have to be true?  Could you not have a texture that solely serves as the render target (and never have content uploaded from the CPU)?

z3moon: This method `setImage` uploads data from CPU, so I think this check is supposed to be here.

poweifeng: ah, yes, ack

poweifeng (approved):
pls wait for @pixelflinger to approve as well.

PR #8788, opened 2025-05-29
fix a buffer overflow during init

pixelflinger:
the default cubemap has RGBA pixels (4 bytes per pixel).

### DIFF
--- a/filament/src/details/Engine.cpp
+++ b/filament/src/details/Engine.cpp
@@ -370,13 +370,16 @@ void FEngine::init() {
             .sampler(Texture::Sampler::SAMPLER_CUBEMAP)
             .build(*this));
 
-    static constexpr std::array<uint8_t, 6> zeroCubemap{};
-    static constexpr std::array<uint8_t, 4> zeroRGBA{};
-    static constexpr std::array<uint8_t, 4> oneRGBA{ 0xff, 0xff, 0xff, 0xff };
-    static constexpr std::array<float, 1> oneFloat{ 1.0f };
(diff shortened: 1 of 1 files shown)


poweifeng (approved):

filament/src/details/Engine.cpp
```diff
@@ -370,13 +370,13 @@ void FEngine::init() {
...
-    static constexpr std::array<uint8_t, 4> oneRGBA{ 0xff, 0xff, 0xff, 0xff };
-    static constexpr std::array<float, 1> oneFloat{ 1.0f };
+    static constexpr std::array<uint32_t, 6> zeroCubemap{};
+    static constexpr std::array<uint32_t, 1> zeroRGBA{};
+    static constexpr std::array<uint32_t, 1> oneRGBA{ 0xffffffff };
+    static constexpr std::array<float   , 1> oneFloat{ 1.0f };
 
     driverApi.update3DImage(mDefaultIblTexture->getHwHandle(), 0, 0, 0, 0, 1, 1, 6,
-            { zeroCubemap.data(), 6, Texture::Format::RGBA, Texture::Type::UBYTE });
+            { zeroCubemap.data(), zeroCubemap.size(), Texture::Format::RGBA, Texture::Type::UBYTE });
```
poweifeng, line 379: should be 24 instead of 6?

poweifeng (changes requested):

filament/src/details/Engine.cpp
pixelflinger: fudge lol. you're right. good catch.

poweifeng (approved):

PR #8790, opened 2025-05-29
webgpu: use params in draw call

jneljneljnel:
BUGS=[397432947]


filament/backend/src/webgpu/WebGPUDriver.cpp
```diff
@@ -864,8 +864,13 @@ void WebGPUDriver::draw2(uint32_t indexOffset, uint32_t indexCount, uint32_t ins
     mRenderPassEncoder.DrawIndexed(indexCount, instanceCount, indexOffset, 0, 0);
 }
 
-void WebGPUDriver::draw(PipelineState, Handle<HwRenderPrimitive>, uint32_t indexOffset,
+void WebGPUDriver::draw(PipelineState pipelineState, Handle<HwRenderPrimitive> rph, uint32_t indexOffset,
```
kpiascik, line 867: nit: `rph` should be `renderPrimitiveHandle`, which I think is wha you're using here

jc3265 (approved):

PR #8795, opened 2025-05-30
materials: introduce mutable spec constants

elizagamedev:
Rationale & design of this feature has been discussed internally.

The current implementation uses a `FixedCapacityVector` to store the new program handles, but I wouldn't object to replacing it with a hasmap as discussed offline.

I have compiled but not tested this yet on Android, so I'm not certain that the API bindings are correctly wired up.


elizagamedev:
Addressed comments and also tested that it works on Android.

filament/src/details/Material.h
```diff
@@ -208,6 +212,12 @@ class FMaterial : public Material {
     }
     size_t getParameters(ParameterInfo* parameters, size_t count) const noexcept;
 
+    size_t getMutableConstantCount() const noexcept { return mMaterialMutableConstants.size(); }
+    std::optional<uint32_t> getMutableConstantId(std::string_view name) const noexcept;
```
bejado, line 216: `uint8_t` here since there's a max of 8 mutable constants?

samples/materials/bakedColorWithMutableConstants.mat
```diff
@@ -0,0 +1,41 @@
...
+            default: true,
+        },
+        {
+            name: darken,
+            type: bool,
+            mutable: true,
+        },
+        {
+            // This offsets the ID of the mutable constants by one,
+            // and is included here purely for testing purposes.
```
bejado: Shouldn't this be the first constant? Otherwise how does it offset the others' IDs?

NEW_RELEASE_NOTES.md
```diff
@@ -7,3 +7,6 @@ for next branch cut* header.
 appropriate header in [RELEASE_NOTES.md](./RELEASE_NOTES.md).
 
 ## Release notes for next branch cut
+
+- materials: sampler now export their type in the material binary [⚠️ **New Material Version**]
```
poweifeng, line 12: rebase?

docs_src/markdeep/Materials.md.html
```diff
@@ -1091,30 +1091,37 @@
...
-float                  | A single-precision GLSL float            |         0.0
-bool                   | A GLSL bool                              |         false
+    Constants may also be specified as mutable by setting the `mutable` property to `true`. Only
+    `bool` spec constants may be specified as mutable.
+
+| Type  | Description                   | Default | May be mutable? |
+|:------|:------------------------------|:--------|-----------------|
+| int   | A signed, 32 bit GLSL int     |         | no              |
+| float | A single-precision GLSL float | 0.0     | no              |
+| bool  | A GLSL bool                   | false   | yes             |
```
poweifeng, line 1101: mention default value for `mutable`?

poweifeng: actually, it should seem pretty obvious that it's false.

I know this was discussed in the doc, but maybe just use `mutable_bool` as a type would be more explicit? Because there's really not a case where you'd want `mutable: false`.  And also `mutable` only applies to bool.

samples/mutable_constants.cpp
```diff
@@ -0,0 +1,214 @@
+/*
+ * Copyright (C) 2018 The Android Open Source Project
```
poweifeng:
```suggestion
 * Copyright (C) 2025 The Android Open Source Project
```

samples/materials/bakedColorWithMutableConstants.mat
elizagamedev: Immutable constants are always generated before mutable constants. I updated the comment to make this clearer.

filament/src/details/Material.h
elizagamedev: I changed this to a `uint32_t` to match the actual type of `mMaterialMutableConstants.size()`. I figure it's probably best keep this value as CPU word size.

docs_src/markdeep/Materials.md.html
elizagamedev: I mentioned this a little bit in the design doc, but in order to keep forwards compatibility in case we want to extend this feature to other types, we should keep the spec very similar to the existing immutable constants. Treating mutable constants as different types also complicates the implementation slightly.

PR #8808, opened 2025-06-02
renderdiff: fix breakage in parsing commit message

poweifeng:


poweifeng:
bypassing review due to breakage.

PR #8809, opened 2025-06-03
webgpu: fix vertex buffer info for hellopbr

AndyHovingh:
BUGS = [397432947]

### DIFF
--- a/filament/backend/src/webgpu/WebGPUDriver.cpp
+++ b/filament/backend/src/webgpu/WebGPUDriver.cpp
@@ -58,9 +58,7 @@ WebGPUDriver::WebGPUDriver(WebGPUPlatform& platform,
               driverConfig.disableHandleUseAfterFreeCheck, driverConfig.disableHeapHandleTags) {
     mAdapter = mPlatform.requestAdapter(nullptr);
     mDevice = mPlatform.requestDevice(mAdapter);
-    wgpu::Limits supportedLimits{};
-    mDevice.GetLimits(&supportedLimits);
-    mMinUniformBufferOffsetAlignment = supportedLimits.minUniformBufferOffsetAlignment;
+    mDevice.GetLimits(&mDeviceLimits);
     mQueue = mDevice.GetQueue();
 }
 
@@ -337,9 +335,10 @@ void WebGPUDriver::createSwapChainR(Handle<HwSwapChain> sch, void* nativeWindow,
 void WebGPUDriver::createSwapChainHeadlessR(Handle<HwSwapChain> sch, uint32_t width,
         uint32_t height, uint64_t flags) {}
 
-void WebGPUDriver::createVertexBufferInfoR(Handle<HwVertexBufferInfo> vbih, uint8_t bufferCount,
-        uint8_t attributeCount, AttributeArray attributes) {
-    constructHandle<WGPUVertexBufferInfo>(vbih, bufferCount, attributeCount, attributes);
+void WebGPUDriver::createVertexBufferInfoR(Handle<HwVertexBufferInfo> vertexBufferInfoHandle,
+        uint8_t bufferCount, uint8_t attributeCount, AttributeArray attributes) {
+    constructHandle<WGPUVertexBufferInfo>(vertexBufferInfoHandle, bufferCount, attributeCount,
+            attributes, mDeviceLimits);
 }
 
 void WebGPUDriver::createVertexBufferR(Handle<HwVertexBuffer> vbh, uint32_t vertexCount,
@@ -965,15 +964,18 @@ void WebGPUDriver::bindPipeline(PipelineState const& pipelineState) {
     mRenderPassEncoder.SetPipeline(pipeline);
 }
 
-void WebGPUDriver::bindRenderPrimitive(Handle<HwRenderPrimitive> rph) {
-    auto* renderPrimitive = handleCast<WGPURenderPrimitive>(rph);
-    auto vbi = handleCast<WGPUVertexBufferInfo>(renderPrimitive->vertexBuffer->vbih);
-    for (const auto& webGPUSlotBindings: vbi->getWebGPUSlotBindingInfos()) {
-        mRenderPassEncoder.SetVertexBuffer(webGPUSlotBindings.slot,
-                renderPrimitive->vertexBuffer->buffers[webGPUSlotBindings.sourceBuffer],
-                webGPUSlotBindings.bufferOffset);
+void WebGPUDriver::bindRenderPrimitive(Handle<HwRenderPrimitive> renderPrimitiveHandle) {
+    const auto renderPrimitive = handleCast<WGPURenderPrimitive>(renderPrimitiveHandle);
+    const auto vertexBufferInfo =
+            handleCast<WGPUVertexBufferInfo>(renderPrimitive->vertexBuffer->vbih);
+    for (size_t slotIndex = 0; slotIndex < vertexBufferInfo->getVertexBufferLayoutCount();
+            slotIndex++) {
+        WGPUVertexBufferInfo::WebGPUSlotBindingInfo const& bindingInfo =
+                vertexBufferInfo->getWebGPUSlotBindingInfos()[slotIndex];
+        mRenderPassEncoder.SetVertexBuffer(slotIndex,
+                renderPrimitive->vertexBuffer->buffers[bindingInfo.sourceBufferIndex],
+                bindingInfo.bufferOffset);
     }
-
     mRenderPassEncoder.SetIndexBuffer(renderPrimitive->indexBuffer->getBuffer(),
             renderPrimitive->indexBuffer->indexFormat);
 }
@@ -1026,9 +1028,10 @@ void WebGPUDriver::updateDescriptorSetBuffer(Handle<HwDescriptorSet> dsh,
     auto buffer = handleCast<WGPUBufferObject>(boh);
     if (!bindGroup->getIsLocked()) {
         // TODO making assumptions that size and offset mean the same thing here.
-        FILAMENT_CHECK_PRECONDITION(offset % mMinUniformBufferOffsetAlignment == 0)
-                << "Binding offset must be multiple of " << mMinUniformBufferOffsetAlignment
-                << "But requested offset is " << offset;
+        FILAMENT_CHECK_PRECONDITION(offset % mDeviceLimits.minUniformBufferOffsetAlignment == 0)
+                << "Binding offset must be multiple of "
+                << mDeviceLimits.minUniformBufferOffsetAlignment << "But requested offset is "
+                << offset;
         wgpu::BindGroupEntry entry{ .binding = static_cast<uint32_t>(binding * 2),
             .buffer = buffer->getBuffer(),
             .offset = offset,
--- a/filament/backend/src/webgpu/WebGPUHandles.cpp
+++ b/filament/backend/src/webgpu/WebGPUHandles.cpp
@@ -197,89 +197,127 @@ void WGPUBufferBase::updateGPUBuffer(BufferDescriptor& bufferDescriptor, uint32_
     }
 }
 
-static constexpr uint32_t DUMMY_WEBGPU_SLOT = 0;
-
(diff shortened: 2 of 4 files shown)


filament/backend/src/webgpu/WebGPUDriver.cpp
```diff
@@ -965,15 +964,17 @@ void WebGPUDriver::bindPipeline(PipelineState const& pipelineState) {
...
 
-void WebGPUDriver::bindRenderPrimitive(Handle<HwRenderPrimitive> rph) {
-    auto* renderPrimitive = handleCast<WGPURenderPrimitive>(rph);
-    auto vbi = handleCast<WGPUVertexBufferInfo>(renderPrimitive->vertexBuffer->vbih);
-    for (const auto& webGPUSlotBindings: vbi->getWebGPUSlotBindingInfos()) {
-        mRenderPassEncoder.SetVertexBuffer(webGPUSlotBindings.slot,
-                renderPrimitive->vertexBuffer->buffers[webGPUSlotBindings.sourceBuffer],
-                webGPUSlotBindings.bufferOffset);
+void WebGPUDriver::bindRenderPrimitive(Handle<HwRenderPrimitive> renderPrimitiveHandle) {
+    auto* renderPrimitive = handleCast<WGPURenderPrimitive>(renderPrimitiveHandle);
```
bridgewaterrobbie: Not a huge issue and it existed before, but a bit inconsistent to do auto* for one handleCast, and auto for the other.

filament/backend/src/webgpu/WebGPUHandles.h
```diff
@@ -45,44 +45,43 @@ class WGPUProgram final : public HwProgram {
...
+    // (this is a vector and not an array due to size limitations of the handle in the handle
+    // allocator)
     std::vector<wgpu::VertexBufferLayout> mVertexBufferLayouts;
 
-    // This stores all wgpu::VertexAttribute structs, indexed by their original
-    // Filament attributeIndex.
-    std::vector<std::vector<wgpu::VertexAttribute>> mVertexAttributes;
+    // (this is a vector and not an array due to size limitations of the handle in the handle
+    // allocator)
+    std::vector<wgpu::VertexAttribute> mVertexAttributes;
```
bridgewaterrobbie, line 84: Unlike the other private members, this one now lacks a comment above it describing what it is (I know we didn't commit to comment guidelines in general, but calling out consistency here)

filament/backend/src/webgpu/WebGPUDriver.cpp
```diff
@@ -965,15 +964,17 @@ void WebGPUDriver::bindPipeline(PipelineState const& pipelineState) {
...
+    auto vertexBufferInfo = handleCast<WGPUVertexBufferInfo>(renderPrimitive->vertexBuffer->vbih);
+    for (size_t slotIndex = 0; slotIndex < vertexBufferInfo->getVertexBufferLayoutCount();
+            slotIndex++) {
+        WGPUVertexBufferInfo::WebGPUSlotBindingInfo const& bindingInfo =
+                vertexBufferInfo->getWebGPUSlotBindingInfos()[slotIndex];
+        mRenderPassEncoder.SetVertexBuffer(slotIndex,
+                renderPrimitive->vertexBuffer->buffers[bindingInfo.sourceBufferIndex],
```
kpiascik, line 976: was it really the difference between `sourceBuffer` vs `sourceBufferIndex` ?

filament/backend/src/webgpu/WebGPUHandles.cpp
```diff
@@ -197,89 +197,113 @@ void WGPUBufferBase::updateGPUBuffer(BufferDescriptor& bufferDescriptor, uint32_
...
-            .format = vertexFormat,
-            .offset = attribute.offset % attribute.stride,
-            .shaderLocation = attributeIndex,
-        });
-        mVertexBufferLayouts[assignedSlot].attributeCount++;
+        attributeInfos[currentAttributeIndex++] = {
+            .slot = existingSlot,
+            .attribute = {
+                .format = vertexFormat,
+                .offset = attribute.offset - mWebGPUSlotBindingInfos[existingSlot].bufferOffset,
```
jc3265, line 286: is there a case for when `attribute.offset` is 0 and `mWebGPUSlotBindingInfos[existingSlot].bufferOffset` is not? 
or cases in which `mWebGPUSlotBindingInfos[existingSlot].bufferOffset > attribute.offset` ? that would result in a negative offset

filament/backend/src/webgpu/WebGPUHandles.cpp
```diff
@@ -197,89 +197,113 @@ void WGPUBufferBase::updateGPUBuffer(BufferDescriptor& bufferDescriptor, uint32_
...
-                .bufferOffset = attribute.offset - (attribute.offset % attribute.stride),
-                .stride = attribute.stride,
+                .sourceBufferIndex = attribute.buffer,
+                .bufferOffset = attribute.offset,
+                .stride = attribute.stride
+            });
+            // we need to have a vertex buffer layout for each slot
+            mVertexBufferLayouts.push_back({
+                .stepMode = wgpu::VertexStepMode::Vertex,
+                .arrayStride = attribute.stride
```
jc3265, line 277: do we also need to set `.attributeCount` ?

filament/backend/src/webgpu/WebGPUHandles.h
```diff
@@ -45,44 +45,43 @@ class WGPUProgram final : public HwProgram {
 
 
 // WGPUVertexBufferInfo maps Filament vertex attributes to WebGPU buffer binding model.
-class WGPUVertexBufferInfo : public HwVertexBufferInfo {
+class WGPUVertexBufferInfo final : public HwVertexBufferInfo {
 public:
-    WGPUVertexBufferInfo(uint8_t bufferCount, uint8_t attributeCount,
-            AttributeArray const& attributes);
+    WGPUVertexBufferInfo(uint8_t bufferCount, uint8_t attributeCount, AttributeArray const&,
+            wgpu::Limits const&);
```
jc3265: why not `AttributeArray const& attributes, wgpu::Limits const& deviceLimits`?

filament/backend/src/webgpu/WebGPUDriver.cpp
bridgewaterrobbie: (Speculating mostly for fun until Andy can answer because I'm also curious) I don't think so, take a look at line 260/261 of handles.cpp, this looks to just be a more specific renaming. The old `.sourceBuffer = attribute.buffer, ` is replaced with `.sourceBufferIndex = attribute.buffer,`

filament/backend/src/webgpu/WebGPUHandles.cpp
AndyHovingh: There are cases where the offset could be less than the bufferOffset of a given slot, because we can (and do) have scenarios where the same buffer is used for multiple attributes, but the attributes are not interleaved. So, a given attribute can be placed in one partition of the buffer and another attribute in another partition. In that case, we need to specify them as separate slots/vertex buffer layouts to WebGPU (since WebGPU needs each set of attributes for a layout interleaved). The bufferOffset is where the partition begins for a given slot/layout. Thus, if you encounter an attribute at a partition earlier than another partition/slot this can happen. I was actually seeing this partitioning for different attributes, and , furthermore, attributes with identical strides in the same buffer (but different partitions), thus the check below that the stride does not exceed the attributes offset from the partition's offset. In any case, whether this happens or not (the difference being logically negative) in our samples, we should be robust and defensive about the possibility anyway. These are unsigned integers, so we have a bug if we subtract them and the logical result is negative (the actually result overflowing).

filament/backend/src/webgpu/WebGPUHandles.cpp
AndyHovingh: Yes, but we don't know what that is at this point. We set this in another pass at the end of the function.

filament/backend/src/webgpu/WebGPUDriver.cpp
AndyHovingh: Yes, just more specific naming 👍  they are the same

filament/backend/src/webgpu/WebGPUHandles.h
AndyHovingh: are the names needed in the header interface? does it help clarify something?

AndyHovingh:
> A few nits, but this all makes straightforward sense reading it, which is ideal for something complex.

Thank you @bridgewaterrobbie ! That's always wonderful to hear and important! Hopefully, we can be improve that some more as part of the refactor we already discussed before.

filament/backend/src/webgpu/WebGPUDriver.cpp
AndyHovingh: Good call. I'll make them more consistent.

filament/backend/src/webgpu/WebGPUHandles.h
AndyHovingh: Good point. A comment about what this is could help and would be more consistent.

AndyHovingh: done

filament/backend/src/webgpu/WebGPUDriver.cpp
AndyHovingh: done

filament/backend/src/webgpu/WebGPUHandles.cpp
AndyHovingh: @jc3265 I added a comment in the code to help clarify this. What do you think?

filament/backend/src/webgpu/WebGPUHandles.cpp
AndyHovingh: @jc3265 I added a comment to clarify this as well. What do you think?

filament/backend/src/webgpu/WebGPUHandles.cpp
AndyHovingh: Sorry, @jc3265 yes, there is a case, but not at this point in the code. We check that above when we select the slot. The comment is there.

jc3265: Sounds good Thanks for the explanation!

filament/backend/src/webgpu/WebGPUHandles.cpp
jc3265: sounds good!

filament/backend/src/webgpu/WebGPUHandles.cpp
AndyHovingh: Thank you!

filament/backend/src/webgpu/WebGPUHandles.h
jc3265: well, on this function we already have `bufferCount` and `attributeCount`
and all of the other functions do have the names, so maybe for consistency? or also remove `bufferCount` and `attributeCount`
?

AndyHovingh: I don't want to remove `bufferCount` and `attributeCount` because then we just have `uint8_t` and `uint8_t` with no idea here what those arguments are supposed to represent; the type doesn't provide enough of a clue. However, the types `AttributeArray` and `wgpu::Limits` do provide just as much of a clue as the names `attributes` and `limits`. I would favor concision to consistency in a line of code or function signature, as I think that is ultimately more readable. I think consistency is more of argument about how common the pattern of omitting obvious/redundant names is across the code base, which I do not think is a strong argument here, especially as it is not defined in the code style.  That said, it really is not a big deal; do you feel that keeping the `attributes` and `deviceLimits` names help? If so, I can add them back.

kpiascik: Even if C++ headers don't explicitly need them and your types are pretty descriptive, there's no har in keeping the variable names

AndyHovingh: True, there isn't any harm in them except unnecessary verbosity/redundancy, which is not a big a deal. I will add them back.

AndyHovingh: added back 👍

PR #8810, opened 2025-06-03
utils: add additional guards for Tracing

poweifeng:
The addition JobSystem.cpp allows for defining
FILAMENT_TRACING_ENABLED across targets.

Addingin FILAMENT_TRACING_ENABLED to the #if in Tracing.h prevents perfetto from being included.

### DIFF
--- a/libs/utils/src/JobSystem.cpp
+++ b/libs/utils/src/JobSystem.cpp
@@ -14,8 +14,13 @@
  * limitations under the License.
  */
 
+// TODO: Clean-up. We shouldn't need this #ifndef here, but a client has requested that perfetto be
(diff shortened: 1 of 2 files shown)


libs/utils/src/JobSystem.cpp
```diff
@@ -14,8 +14,13 @@
  * limitations under the License.
  */
 
+// TODO: Clean-up. We shouldn't need this #ifndef here, but a client has requested that perfetto be
+// disabled due to size increase.  In their case, this flag would be defined across targets. Hence
```
bejado, line 18: I'm confused why we need this `ifndef`. What is the client defining `FILAMENT_TRACING_ENABLED` to?

bejado (approved):

PR #8813, opened 2025-06-03
webgpu: Generate mipmaps when requested, deferring as needed

bridgewaterrobbie:


filament/backend/src/webgpu/WebGPUDriver.cpp
```diff
@@ -695,7 +695,60 @@ void WebGPUDriver::setupExternalImage2(Platform::ExternalImageHandleRef image) {
...
+            .aspect = texture->getAspect(),
+        };
+
+        uint32_t dstWidth = std::max(1u, width >> 1);
+        uint32_t dstHeight = std::max(1u, height >> 1);
+        uint32_t dstDepth = std::max(1u, depth >> 1);
+
+        wgpu::Extent3D copySize = { dstWidth, dstHeight,
+            (texture->target == SamplerType::SAMPLER_3D) ? dstDepth
+                                                         : texture->getArrayLayerCount() };
```
jc3265: for consistency I would change this to 
```
 wgpu::Extent3D{ .width = dstWidth, .height = dstHeight, .depthOrArrayLayers = (texture->target == SamplerType::SAMPLER_3D) ? dstDepth
                                                         : texture->getArrayLayerCount() };
```

bridgewaterrobbie: Good point. Will do

filament/backend/src/webgpu/WebGPUHandles.h
```diff
@@ -181,6 +181,7 @@ class WGPUTexture : public HwTexture {
     [[nodiscard]] const wgpu::Texture& getTexture() const { return mTexture; }
     [[nodiscard]] const wgpu::TextureView& getTextureView() const { return mTexureView; }
     [[nodiscard]] wgpu::TextureFormat getFormat() const { return mFormat; }
+    [[nodiscard]] uint32_t getArrayLayerCount() const { return mArrayLayerCount; }
```
AndyHovingh, line 187: As we have things coded at this time, wouldn't this _always_ return 1? It looks hardcoded below and not set or updated anywhere. Should we return something more meaningful here depending on the `samplerTargetType` or remove it entirely? Do we expect some other way of updating/interpreting this than that?

bridgewaterrobbie: The constructor uses filament-passed values for certain targets.

```    switch (target) {
        case SamplerType::SAMPLER_2D:
            mArrayLayerCount = 1;
            break;
        case SamplerType::SAMPLER_2D_ARRAY:
            mArrayLayerCount = textureDescriptor.size.depthOrArrayLayers;
            break;
        case SamplerType::SAMPLER_CUBEMAP:
            textureDescriptor.size.depthOrArrayLayers = 6;
            mArrayLayerCount = textureDescriptor.size.depthOrArrayLayers;
            break;
        case SamplerType::SAMPLER_EXTERNAL:
        case SamplerType::SAMPLER_3D:
            mArrayLayerCount = 1;
            break;
        case SamplerType::SAMPLER_CUBEMAP_ARRAY:
            textureDescriptor.size.depthOrArrayLayers = depth * 6;
            mArrayLayerCount = textureDescriptor.size.depthOrArrayLayers;
            break;
    }
```

edit: hang on I see what you mean I think. I think this PR is correct, but you may be pointing out a different valid problem to solve
edit2: Nope, I double checking the code, it comes form .size = { .width = width, .height = height, .depthOrArrayLayers = depth }, which ultimately is from the params passed.

AndyHovingh: Thank you @bridgewaterrobbie ! I stand corrected sir! Strangely, CLion was not finding these usages earlier. When I pulled down the changes again I see them now in usages as assigned in the constructor as you pointed out ¯\_(ツ)_/¯... I will blame it on the IDE 😂 Never mind, this makes sense 👍

AndyHovingh: Thank you @bridgewaterrobbie ! I stand corrected sir! Strangely, CLion was not finding these usages earlier. When I pulled down the changes again I see them now in usages as assigned in the constructor as you pointed out ¯\_(ツ)_/¯... I will blame it on the IDE 😂 Never mind, this makes sense 👍

filament/backend/src/webgpu/WebGPUDriver.cpp
```diff
@@ -805,6 +859,12 @@ void WebGPUDriver::makeCurrent(Handle<HwSwapChain> drawSch, Handle<HwSwapChain>
         .label = "command_encoder"
     };
     mCommandEncoder = mDevice.CreateCommandEncoder(&commandEncoderDescriptor);
+    if (!mMipQueue.empty()) {
+        for (auto& h: mMipQueue) {
```
AndyHovingh: nit: might be easier to grok if `h` were something like `textureHandle`

AndyHovingh:
Nice and quick work @bridgewaterrobbie ! 💯

PR #8815, opened 2025-06-03
Fix backend tests on macos vulkan.

mhoff12358:


filament/backend/test/test_LoadImage.cpp
```diff
@@ -449,6 +449,7 @@ TEST_F(LoadImageTest, UpdateImageSRGB) {
     RenderPassParams params = getClearColorRenderPass();
     params.viewport.width = kTexSize;
     params.viewport.height = kTexSize;
+    api.makeCurrent(swapChain, swapChain);
```
mhoff12358: These are seemingly redundant with api.makeCurrent calls earlier in the test, but specifically for vulkan without this they crash when resolving the readPixels inside EXPECT_IMAGE.

poweifeng (approved):

PR #8819, opened 2025-06-04
Add common CLI Args parser for the samples

jc3265:
add a common cli parser for the samples


kpiascik (approved):
lgtm

samples/common/arguments.cpp
```diff
@@ -0,0 +1,63 @@
...
+
+    while ((opt = getopt_long(argc, argv, OPTSTR, OPTIONS, &optionIndex)) >= 0) {
+        std::string arg(optarg ? optarg : "");
+        switch (opt) {
+            case 'a':
+                return parseArgumentsForBackend(arg);
+        }
+    }
+    // if no args were provided return OPENGL as the default
+    return filament::Engine::Backend::OPENGL;
```
AndyHovingh: We should not assume OpenGL as the default. We should return `filament::Engine::Backend::DEFAULT` and let Filament choose the backend as normal.

samples/animation.cpp
```diff
@@ -14,6 +14,8 @@
  * limitations under the License.
  */
 
+#include "common/arguments.h"
```
AndyHovingh, line 17: is this header really private? Seems like we would consider it public, as it is part of the library linked in these various sample binaries. Should it be `#include <common/arguments.h>`?

samples/common/arguments.cpp
jc3265: so most (if not all) of the samples already had a comment saying 
```
...
Specify the backend API: opengl (default), vulkan, metal, or webgpu
...
```
and thats why I left it as the default

samples/common/arguments.h
```diff
@@ -0,0 +1,28 @@
...
+#ifndef TNT_SAMPLES_ARGUMENTS_H
+#define TNT_SAMPLES_ARGUMENTS_H
+
+#include <filament/Engine.h>
+
+#include <string>
+
+namespace samples {
+    filament::Engine::Backend parseArgumentsForBackend(int argc, char* argv[]);
+    filament::Engine::Backend parseArgumentsForBackend(std::string backend);
```
AndyHovingh: can this be a `std::string_view` instead? Why a string? And, if it must be a string, why not `std::string const&`?

samples/gltf_instances.cpp
```diff
@@ -86,7 +88,7 @@ static void printUsage(char* name) {
         "   --help, -h\n"
         "       Prints this message\n\n"
         "   --api, -a\n"
-        "       Specify the backend API: opengl (default), vulkan, or metal\n\n"
+        "       Specify the backend API: opengl (default), vulkan, metal or webgpu\n\n"
```
poweifeng: can this substring also be part of the commons lib (somehow)?

samples/common/arguments.cpp
```diff
@@ -0,0 +1,63 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <filament/Engine.h>
+#include <getopt/getopt.h>
```
AndyHovingh, line 21: nit: space before the 3rd party library include, e.g.
```cpp
#include <filament/Engine.h>

#include <getopt/getopt.h>
```

samples/hellopbr.cpp
```diff
@@ -91,19 +93,7 @@ static int handleCommandLineArguments(int argc, char* argv[], App* app) {
...
-                } else if (arg == "metal") {
-                    app->config.backend = Engine::Backend::METAL;
-                } else if (arg == "webgpu") {
-                    app->config.backend = Engine::Backend::WEBGPU;
-                } else {
-                    std::cerr << "Unrecognized backend. Must be "
-                                 "'opengl'|'vulkan'|'metal'|'webgpu'.\n";
-                    exit(1);
-                }
+                app->config.backend = samples::parseArgumentsForBackend(arg);
```
poweifeng, line 98: Maybe just remove the 'a' case from the switch.  And assign it outside of the switch statment. 
Reasoning is that if we ever need to move to another letter (not 'a'), then it'll be easy. (not that we'll ever want to).

samples/common/arguments.cpp
```diff
@@ -0,0 +1,63 @@
...
+    } else if (backend == "webgpu") {
+        return filament::Engine::Backend::WEBGPU;
+    } else {
+        std::cerr << "Unrecognized target API. Must be 'opengl'|'vulkan'|'metal'|'webgpu'."
+                  << std::endl;
+        exit(1);
+    }
+}
+
+filament::Engine::Backend parseArgumentsForBackend(int argc, char* argv[]) {
```
AndyHovingh, line 70: Since this is for samples that do not already have a CLI, shouldn't we minimally include a help (-h|--help) option with a usage printout?

samples/common/arguments.cpp
AndyHovingh: But the implementation uses `filament::Engine::Backend::DEFAULT`, since the `Config` struct defaults the `backend` member as such (`mutable filament::Engine::Backend backend = filament::Engine::Backend::DEFAULT;` See libs/filamentapp/include/filamentapp/Config.h) and when the backend is not specified neither is that config field.

samples/gltf_instances.cpp
jc3265: yeah so I thought about adding a `printBackednUsage` but I would either have to call that after `printUsage` (which would mean that the options would not be in alphabetical order) or add the call in the middle of the string. 
if having
```
        "   --api, -a\n"
        "       Specify the backend API: opengl (default), vulkan, metal or webgpu\n\n"
```
at the bottom is not a dealbreaker then we can def do that

samples/common/arguments.cpp
jc3265: yeah I see what you mean and setting it to default makes the most sense. we could just set it to `DEFAULT`, update the `usage` comments and remove `opengl` as the default (since it sorta kinda isnt true)

@poweifeng thoughts on if we should leave `OPENGL` or switch to `DEFAULT`?

samples/gltf_instances.cpp
poweifeng: so the way I think I'd do is just to generalize this
```
    const std::string from("SHOWCASE");
    for (size_t pos = usage.find(from); pos != std::string::npos; pos = usage.find(from, pos)) {
        usage.replace(pos, from.length(), exec_name);
    }
```
into another utils in the commons lib, and just have a place holder (e.g. "BACKEND_OPTIONS") in the string here, and then replace it with the backend option string.

samples/common/arguments.cpp
poweifeng: I would vote for removing the "opengl (default)" part and use Backend::DEFAULT when not specified.

samples/gltf_instances.cpp
jc3265: Gotcha
Thanks

samples/hellopbr.cpp
jc3265: We would need to leave the `case a` (even if just a `break` statement) or we would have to modify 
```
            default:
            case 'h':
                printUsage(argv[0]);
                exit(0);
```
removing the `case 'a' makes it fall under the `default` and just continuously print the usage

samples/common/arguments.cpp
jc3265: this is a good idea. Added those options for the cases without a CLI

samples/hellopbr.cpp
poweifeng: ack

PR #8820, opened 2025-06-04
Fix Windows build error C2512

kpiascik:


filament/backend/src/webgpu/WebGPUHandles.cpp
```diff
@@ -218,6 +218,14 @@ WGPUVertexBufferInfo::WGPUVertexBufferInfo(uint8_t bufferCount, uint8_t attribut
     struct AttributeInfo final {
         uint8_t slot = IMPOSSIBLE_SLOT_INDEX;
         wgpu::VertexAttribute attribute = {};
+        AttributeInfo() {
```
AndyHovingh: We should use constructor initializer lists when we can, as they are more efficient, among other things, -> [ML Filament WebGPU C++ Coding Guidelines > Consider Doing (weaker opinions) > Prefer Constructor Initialization Lists When Possible](https://docs.google.com/document/d/1bBvYFjfAZ6sWXwYvbdQzh9egNItLMHzyGgLxpr1HY8M/edit?resourcekey=0-95-dhqXNMLgD_LG6haAmFw&tab=t.0#heading=h.1lpf8wdztltp).

Does this work?:
```cpp
AttributeInfo() : slot(IMPOSSIBLE_SLOT_INDEX), attribute({}) {}
AttributeInfo(uint8_t slot, wgpu::VertexAttribute attribute) : slot(slot), attribute(attribute) {}
```

kpiascik: done

filament/backend/src/webgpu/WebGPUHandles.cpp
```diff
@@ -279,14 +285,13 @@ WGPUVertexBufferInfo::WGPUVertexBufferInfo(uint8_t bufferCount, uint8_t attribut
                 // in a subsequent pass over the attributeInfos we collect in this loop.
             });
         }
-        attributeInfos[currentAttributeIndex++] = {
-            .slot = existingSlot,
-            .attribute = {
+        attributeInfos[currentAttributeIndex++] = AttributeInfo(existingSlot,
```
AndyHovingh, line 288: Can we still use a designated initializer at the top level? I suspect Microsoft is not smart enough to deduce the type on the left? Does this work?:
```cpp
attributeInfos[currentAttributeIndex++] = AttributeInfo{
  slot = existingSlot,
  attribute = { ...}
};
```

If not, no biggie.

AndyHovingh:
Thank you @kpiascik for checking the Windows build! Good to know we are not going off the track on that!

filament/backend/src/webgpu/WebGPUHandles.cpp
kpiascik: no, this doesn't work

AndyHovingh: 😭  ok. we gotta do what we gotta do ¯\_(ツ)_/¯ Thank you for checking!

PR #8823, opened 2025-06-04
Backend tests using webgpu

mhoff12358:
Add "Web GPU" as one of the supported backends for the unit tests. At this time this is just the initial implementation as the tests are currently not running

### DIFF
--- a/filament/backend/test/test_LoadImage.cpp
+++ b/filament/backend/test/test_LoadImage.cpp
@@ -453,7 +453,6 @@ TEST_F(LoadImageTest, UpdateImageSRGB) {
     RenderPassParams params = getClearColorRenderPass();
     params.viewport.width = kTexSize;
     params.viewport.height = kTexSize;
-    api.makeCurrent(swapChain, swapChain);
     PipelineState state = getColorWritePipelineState();
     shader.addProgramToPipelineState(state);
     state.primitiveType = PrimitiveType::TRIANGLES;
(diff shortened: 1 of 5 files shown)


filament/backend/test/test_LoadImage.cpp
```diff
@@ -453,6 +453,7 @@ TEST_F(LoadImageTest, UpdateImageSRGB) {
     RenderPassParams params = getClearColorRenderPass();
     params.viewport.width = kTexSize;
     params.viewport.height = kTexSize;
+    api.makeCurrent(swapChain, swapChain);
```
poweifeng: there's a makeCurrent above.  This doesn't seem like it should be necessary.  Maybe this is fixed by #8984

jc3265: yes, seems like it was redundant. Removed 👍

filament/backend/test/test_LoadImage.cpp
```diff
@@ -615,6 +616,7 @@ TEST_F(LoadImageTest, UpdateImage3D) {
         RenderPassParams params = getClearColorRenderPass();
         params.viewport.width = kTexSize;
         params.viewport.height = kTexSize;
+        api.makeCurrent(swapChain, swapChain);
```
poweifeng: this call as well.  There is also a makecurrent above

filament/backend/test/test_LoadImage.cpp
```diff
@@ -528,6 +528,7 @@ TEST_F(LoadImageTest, UpdateImageMipLevel) {
...
```
poweifeng: this call as well.  There is also a makecurrent above

jc3265: Removed redundant `makeCurrent`

filament/backend/test/test_LoadImage.cpp
jc3265: Removed redundant `makeCurrent`

poweifeng (approved):

PR #8827, opened 2025-06-05
webgpu: Add required feature RG11B10UfloatRenderable for gltf_viewer, with comment on why

bridgewaterrobbie:


filament/backend/src/webgpu/platform/WebGPUPlatform.cpp
```diff
@@ -56,7 +56,12 @@ namespace {
 constexpr std::array REQUIRED_FEATURES = { wgpu::FeatureName::TransientAttachments,
```
AndyHovingh, line 56: @poweifeng note that if the system does not have one of these REQUIRED_FEATURES WebGPU just fails right away with a message about not finding a compatible adapter.

Thus, this change requires the `wgpu::FeatureName::RG11B10UfloatRenderable`

bridgewaterrobbie: I will also note that I have attempted to have `isTextureFormatSupported` return false for the relevant Filament type, and the result is that Filament just panics. If we want to support devices that don't have this extension, then we also need a way to be certain Filament won't try to use it (Which might just indicate that making the extension required is the right thing to do, even if the other samples did not use it)

poweifeng: seems reasonable.

poweifeng (approved):

PR #8828, opened 2025-06-05
webgpu: Correct stencil state handling for depth-only formats

bridgewaterrobbie:


filament/backend/src/webgpu/WebGPUPipelineCreation.cpp
```diff
@@ -65,6 +65,17 @@ constexpr wgpu::CullMode toWebGPU(CullingMode cullMode) {
     }
 }
 
+bool hasStencilAspect(wgpu::TextureFormat format) {
```
AndyHovingh, line 68: nit: can this be `constexpr`? `[[nodiscard]]` (toWebGPU functions below could be too, but fine if we do that here or later )

bridgewaterrobbie: Time permitting I'll ad that as a followup- focused on just bringing it to life at the moment

PR #8830, opened 2025-06-05
vk: implement proper offset when uploading to texture

poweifeng:
We didn't take into account the top/left/stride parameters of the PixelBufferDescriptor.

### DIFF
--- a/filament/backend/src/vulkan/VulkanTexture.cpp
+++ b/filament/backend/src/vulkan/VulkanTexture.cpp
@@ -25,6 +25,7 @@
 #include <backend/DriverEnums.h>
 #include <private/backend/BackendUtils.h>
 
+#include <utils/compiler.h>
 #include <utils/Panic.h>
 
 using namespace bluevk;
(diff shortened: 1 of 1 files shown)


filament/backend/src/vulkan/VulkanTexture.cpp
```diff
@@ -218,6 +218,40 @@ VkImageUsageFlags getUsage(VulkanContext const& context, uint8_t samples,
     return usage;
 }
 
+void adjustedMemcpy(void* mapped, PixelBufferDescriptor const& p, size_t width, size_t height,
+        size_t depth) {
+    uint8_t* buf = (uint8_t*) p.buffer;
+    size_t const pixelSize = PixelBufferDescriptor::computeDataSize(p.format, p.type, 1, 1, 1);
+    size_t const pbdStride = p.stride ? p.stride : width;
+
+    // Slow path of copying row by row
```
bejado, line 228: Is the first branch the slow path? Maybe use `UTILS_UNLIKELY`

bejado (approved):

filament/backend/src/vulkan/VulkanTexture.cpp
poweifeng: good point.  done

PR #8832, opened 2025-06-06
webgpu: Fix Android material compilation

kpiascik:
Instead of requiring GL_EXT_shader_non_constant_global_initializers we can just make the new variables const

BUGS=[421457710]


bridgewaterrobbie:
lgtm, but please wait for Powei/Ben to approve too

PR #8834, opened 2025-06-06
Convert staging images to resources

anishmgoyal:
This will reduce the amount of computation required in gc() calls to determine if a staging image is still in use or not. This follows the change made to staging buffers.

### DIFF
--- a/libs/gltfio/src/TangentsJob.cpp
+++ b/libs/gltfio/src/TangentsJob.cpp
@@ -16,6 +16,7 @@
 
 #include "TangentsJob.h"
 
+#include <cstdlib>
 #include <memory>
 
 #include <geometry/SurfaceOrientation.h>


filament/backend/src/vulkan/VulkanStagePool.h
```diff
@@ -123,13 +123,70 @@ class VulkanStage {
...
+  private:
+    const VkFormat mFormat;
+    const uint32_t mWidth;
+    const uint32_t mHeight;
+    const VmaAllocation mMemory;
+    const VkImage mImage;
+
+    uint64_t mLastAccessed;
+    // Denote as a friend so that it can update mLastAccessed.
+    friend class VulkanStagePool;
```
phoenixxxx, line 192: curious why the StagePool can't ask the object to update the mLastAccessed variable? Also would this need to be mutable? It strikes me as the type of field that potentially could be updated on an otherwise constant object?

anishmgoyal: the 'last frame' that an image wasn't used on doesn't change after the stage pool gives it out - so i prefer having the stage pool manage this value. also, this object won't be const within the stagepool, so it can be modified already.

i don't want to use mutable if we don't *have* to use mutable.

last note - the field is private, it doesn't make sense for any other objects to even see the field.

filament/backend/src/vulkan/VulkanStagePool.cpp
```diff
@@ -156,27 +156,27 @@ void VulkanStagePool::destroyStage(VulkanStage const*&& stage) {
     delete stage;
 }
 
-VulkanStageImage const* VulkanStagePool::acquireImage(PixelDataFormat format, PixelDataType type,
+fvkmemory::resource_ptr<VulkanStageImage::Resource> VulkanStagePool::acquireImage(PixelDataFormat format, PixelDataType type,
```
poweifeng: nit: line > 100 chars

filament/backend/src/vulkan/VulkanStagePool.h
```diff
@@ -126,13 +126,70 @@ class VulkanStage {
...
+        Resource(const Resource& other) = delete;
+        Resource(Resource&& other) = delete;
+        Resource& operator=(const Resource& other) = delete;
+        Resource& operator=(Resource&& other) = delete;
+
+        VulkanStageImage* const mImage;
+        RecycleFn mOnRecycleFn;
+    };
+
+    VulkanStageImage(VkFormat format, uint32_t width, uint32_t height, VmaAllocation memory, VkImage image, uint64_t lastAccessed)
```
poweifeng: nit: line > 100 chars

anishmgoyal: updated!

filament/backend/src/vulkan/VulkanStagePool.cpp
anishmgoyal: updated!

phoenixxxx (approved):

PR #8836, opened 2025-06-06
VK: Disable VMA internal synchronization.

rafadevai:
The vulkan backend is currently single threads, so there's no need for VMA to use synchronization
internally. This will improve the CPU performance
when going through VMA.


poweifeng (approved):
nice find!

filament/backend/src/vulkan/VulkanDriver.cpp
```diff
@@ -86,6 +86,7 @@ VmaAllocator createAllocator(VkInstance instance, VkPhysicalDevice physicalDevic
 #endif
     };
     VmaAllocatorCreateInfo const allocatorInfo {
+        .flags = VMA_ALLOCATOR_CREATE_EXTERNALLY_SYNCHRONIZED_BIT,
```
poweifeng, line 92: actually, could you add a comment here for future gen?

rafadevai: Done

phoenixxxx (approved):

PR #8838, opened 2025-06-08
webgpu: refactor: break down handles into their own class files

AndyHovingh:
BUGS = [417598108]

### DIFF
--- a/filament/backend/CMakeLists.txt
+++ b/filament/backend/CMakeLists.txt
@@ -270,18 +270,18 @@ if (FILAMENT_SUPPORTS_WEBGPU)
             src/webgpu/WebGPUDriver.h
             src/webgpu/WebGPUIndexBuffer.cpp
             src/webgpu/WebGPUIndexBuffer.h
-            src/webgpu/WebGPURenderTarget.cpp
-            src/webgpu/WebGPURenderTarget.h
             src/webgpu/WebGPUPipelineCreation.cpp
             src/webgpu/WebGPUPipelineCreation.h
+            src/webgpu/WebGPUProgram.cpp
+            src/webgpu/WebGPUProgram.h
+            src/webgpu/WebGPURenderPrimitive.h
+            src/webgpu/WebGPURenderTarget.cpp
+            src/webgpu/WebGPURenderTarget.h
             src/webgpu/WebGPUStrings.h
             src/webgpu/WebGPUSwapChain.cpp
             src/webgpu/WebGPUSwapChain.h
             src/webgpu/WebGPUTexture.cpp
             src/webgpu/WebGPUTexture.h
-            src/webgpu/WebGPUProgram.cpp
-            src/webgpu/WebGPUProgram.h
-            src/webgpu/WebGPURenderPrimitive.h
             src/webgpu/WebGPUVertexBuffer.cpp
             src/webgpu/WebGPUVertexBuffer.h
             src/webgpu/WebGPUVertexBufferInfo.cpp


filament/backend/CMakeLists.txt
```diff
@@ -257,17 +257,35 @@ if (FILAMENT_SUPPORTS_WEBGPU)
...
-            src/webgpu/WGPUProgram.cpp
+            src/webgpu/WebGPUTexture.cpp
+            src/webgpu/WebGPUTexture.h
+            src/webgpu/WebGPUProgram.cpp
+            src/webgpu/WebGPUProgram.h
+            src/webgpu/WebGPURenderPrimitive.h
+            src/webgpu/WebGPUVertexBuffer.cpp
+            src/webgpu/WebGPUVertexBuffer.h
+            src/webgpu/WebGPUVertexBufferInfo.cpp
+            src/webgpu/WebGPUVertexBufferInfo.h
```
jc3265, line 288: nit
can we sort this list? 
```
src/webgpu/WebGPUBufferBase.cpp
src/webgpu/WebGPUBufferBase.h
src/webgpu/WebGPUBufferObject.cpp
src/webgpu/WebGPUBufferObject.h
src/webgpu/WebGPUConstants.h
src/webgpu/WebGPUDescriptorSet.cpp
src/webgpu/WebGPUDescriptorSet.h
src/webgpu/WebGPUDescriptorSetLayout.cpp
src/webgpu/WebGPUDescriptorSetLayout.h
src/webgpu/WebGPUDriver.cpp
src/webgpu/WebGPUDriver.h
src/webgpu/WebGPUIndexBuffer.cpp
src/webgpu/WebGPUIndexBuffer.h
src/webgpu/WebGPUPipelineCreation.cpp
src/webgpu/WebGPUPipelineCreation.h
src/webgpu/WebGPUProgram.cpp
src/webgpu/WebGPUProgram.h
src/webgpu/WebGPURenderPrimitive.h
src/webgpu/WebGPURenderTarget.cpp
src/webgpu/WebGPURenderTarget.h
src/webgpu/WebGPUStrings.h
src/webgpu/WebGPUSwapChain.cpp
src/webgpu/WebGPUSwapChain.h
src/webgpu/WebGPUTexture.cpp
src/webgpu/WebGPUTexture.h
src/webgpu/WebGPUVertexBuffer.cpp
src/webgpu/WebGPUVertexBuffer.h
src/webgpu/WebGPUVertexBufferInfo.cpp
src/webgpu/WebGPUVertexBufferInfo.h
```

AndyHovingh: done 👍

filament/backend/src/webgpu/WebGPUIndexBuffer.cpp
```diff
@@ -0,0 +1,36 @@
...
+#include <cstdint>
+
+namespace filament::backend {
+
+WebGPUIndexBuffer::WebGPUIndexBuffer(wgpu::Device const& device, const uint8_t elementSize,
+        const uint32_t indexCount)
+    : HwIndexBuffer{ elementSize, indexCount },
+      WebGPUBufferBase{ device, wgpu::BufferUsage::CopyDst | wgpu::BufferUsage::Index,
+          elementSize * indexCount, "index_buffer" },
+      mIndexFormat{ elementSize == 2 ? wgpu::IndexFormat::Uint16 : wgpu::IndexFormat::Uint32 } {}
```
jc3265, line 34: What was the `2` again? is that something that we could put in the `webgpu constants`?

AndyHovingh: number of bytes. I think this is fine, unless we could possibly see Uint64... which makes NO sense for number of vertex indices in a model... 32 bits gets us over 4 billion indices in a model... I think we are good :)

filament/backend/src/webgpu/WebGPUDescriptorSetLayout.cpp
```diff
@@ -0,0 +1,186 @@
...
+
+    std::vector<wgpu::BindGroupLayoutEntry> wEntries;
+    wEntries.reserve(layout.bindings.size() + samplerCount);
+    mBindGroupEntries.reserve(wEntries.capacity());
+
+    for (auto fEntry: layout.bindings) {
+        auto& wEntry = wEntries.emplace_back();
+        auto& entryInfo = mBindGroupEntries.emplace_back();
+        wEntry.visibility = filamentStageToWGPUStage(fEntry.stageFlags);
+        wEntry.binding = fEntry.binding * 2;
```
jc3265, line 85: is this `2` the same in `WebGPUIndexBuffer.cpp`? 
if this is also number of bytes i do think that it would be worth having this as a `const` somewhere so that is easier to read

filament/backend/src/webgpu/WebGPUDescriptorSetLayout.cpp
```diff
@@ -0,0 +1,186 @@
...
+            case DescriptorType::SAMPLER_3D_UINT:
+            case DescriptorType::SAMPLER_2D_MS_FLOAT:
+            case DescriptorType::SAMPLER_2D_MS_INT:
+            case DescriptorType::SAMPLER_2D_MS_UINT:
+            case DescriptorType::SAMPLER_2D_MS_ARRAY_FLOAT:
+            case DescriptorType::SAMPLER_2D_MS_ARRAY_INT:
+            case DescriptorType::SAMPLER_2D_MS_ARRAY_UINT: {
+                auto& samplerEntry = wEntries.emplace_back();
+                auto& samplerEntryInfo = mBindGroupEntries.emplace_back();
+                samplerEntry.binding = fEntry.binding * 2 + 1;
```
jc3265, line 116: same comment for the `2`
I know this was just moved from the handles but why `+1`?

filament/backend/src/webgpu/WebGPUDriver.cpp
```diff
@@ -1126,53 +1158,58 @@ void WebGPUDriver::resetState(int) {
...
+    const auto buffer = handleCast<WebGPUBufferObject>(bufferObjectHandle);
     if (!bindGroup->getIsLocked()) {
         // TODO making assumptions that size and offset mean the same thing here.
         FILAMENT_CHECK_PRECONDITION(offset % mDeviceLimits.minUniformBufferOffsetAlignment == 0)
                 << "Binding offset must be multiple of "
                 << mDeviceLimits.minUniformBufferOffsetAlignment << "But requested offset is "
                 << offset;
-        wgpu::BindGroupEntry entry{ .binding = static_cast<uint32_t>(binding * 2),
+        wgpu::BindGroupEntry entry{
+            .binding = static_cast<uint32_t>(binding * 2),
```
jc3265, line 1173: same comment about the `2`

filament/backend/src/webgpu/WebGPURenderPrimitive.h
```diff
@@ -0,0 +1,34 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef TNT_FILAMENT_BACKEND_WEBGPURENDERPRIMITIVE_H
+#define TNT_FILAMENT_BACKEND_WEBGPURENDERPRIMITIVE_H
+
+#include "DriverBase.h"
```
jc3265, line 20: im not sure as of to when we forward declare vs not
but couldnt we just forward declare 
`struct HwRenderPrimitive` and then not include this file?

filament/backend/src/webgpu/WebGPUDescriptorSetLayout.cpp
AndyHovingh: This is not number of bytes here. This is a trick we did to add samplers (splitting samplers from textures in the shaders, which is required by WebGPU/WGSL) and keep the binding consistent between this backend and the shaders themselves. We need to ensure to the bindings are unique.

filament/backend/src/webgpu/WebGPUDescriptorSetLayout.cpp
AndyHovingh: In short uniqueness. The binding for the sampler needs to differ from the corresponding texture

filament/backend/src/webgpu/WebGPUDriver.cpp
AndyHovingh: same answer as above

filament/backend/src/webgpu/WebGPURenderPrimitive.h
AndyHovingh: We can't get away with that because we extent that struct. If we forward declared it we would get a incomplete class/struct error by the compiler.

filament/backend/src/webgpu/WebGPUTexture.cpp
```diff
@@ -0,0 +1,474 @@
...
+        case TextureFormat::RED_GREEN_RGTC2:         return wgpu::TextureFormat::BC5RGUnorm;
+        case TextureFormat::SIGNED_RED_GREEN_RGTC2:  return wgpu::TextureFormat::BC5RGSnorm;
+        case TextureFormat::RGB_BPTC_UNSIGNED_FLOAT: return wgpu::TextureFormat::BC6HRGBUfloat;
+        case TextureFormat::RGB_BPTC_SIGNED_FLOAT:   return wgpu::TextureFormat::BC6HRGBFloat;
+        case TextureFormat::RGBA_BPTC_UNORM:         return wgpu::TextureFormat::BC7RGBAUnorm;
+        case TextureFormat::SRGB_ALPHA_BPTC_UNORM:   return wgpu::TextureFormat::BC7RGBAUnormSrgb;
+        case TextureFormat::RGB565:
+            // No direct mapping in wgpu. Could potentially map to RGBA8Unorm
+            // and discard the alpha and lower precision.
+            FWGPU_LOGW << "Requested Filament texture format RGB565 but getting "
```
jc3265, line 358: should we only include these logs for `debug`? or add a `TODO` and clean them up after?

jc3265 (approved):
Looks good!
just the comments about the `2`s but i see why we can just leave them as is :shipit:

filament/backend/src/webgpu/WebGPUTexture.cpp
AndyHovingh: I think we should put them behind a flag later, or remove them if we are convinced they are completely harmless. I added visibility into these things to help with our development.

PR #8841, opened 2025-06-09
webgpu: Fix copy/paste error with custom render target

kpiascik:


filament/backend/src/webgpu/WebGPURenderTarget.cpp
```diff
@@ -130,12 +130,12 @@ void WebGPURenderTarget::setUpRenderPassAttachments(wgpu::RenderPassDescriptor&
 
         const bool hasStencil =
                 customStencilTextureView ||
-                (customDepthFormat == wgpu::TextureFormat::Depth24PlusStencil8 ||
-                        customDepthFormat == wgpu::TextureFormat::Depth32FloatStencil8);
+                (customStencilFormat == wgpu::TextureFormat::Depth24PlusStencil8 ||
```
AndyHovingh, line 133: these are actually intentional. The idea was to infer whether we are doing stencil by either the presence of a stencil texture OR the presence of stencil bits in the depth format. So, we should actually not make this change.

However, after discussing with the team last week, we learned that we should not be determined `hasStencil` or `hasDepth` in this way _at all_. Instead, we should rely on Filament's params/flags input to so this. This has been in my queue to fix. Futhermore, we should also validate that if we get BOTH depth and stencil textures that they are actually pointing to the same underlying texture/buffer (i.e. FIlament is already setup to only do depth+stencil with a depth+stencil format in a single texture, never separate textures, which is great because WebGPU does not support another way).

filament/backend/src/webgpu/WebGPURenderTarget.cpp
```diff
@@ -130,12 +130,12 @@ void WebGPURenderTarget::setUpRenderPassAttachments(wgpu::RenderPassDescriptor&
...
+                        customStencilFormat == wgpu::TextureFormat::Depth32FloatStencil8);
 
         const bool hasDepth =
                 customDepthTextureView ||
-                (customStencilFormat == wgpu::TextureFormat::Depth24PlusStencil8 ||
+                (customDepthFormat == wgpu::TextureFormat::Depth24PlusStencil8 ||
                         customDepthFormat == wgpu::TextureFormat::Depth32FloatStencil8);
```
jc3265, line 139: @AndyHovingh I agree with your comment in that this was intentional. However, I believe this line should be changeed to `customStencilFormat == wgpu::TextureFormat::Depth32FloatStencil8` no?
either that or we just tae care of it when we modify this code to use the given params/flags

AndyHovingh: I see what you mean. This is still malformed. With the previous thinking this should have been:
```cpp
const bool hasDepth =
                customDepthTextureView ||
                (customStencilFormat == wgpu::TextureFormat::Depth24PlusStencil8 ||
                        customStencilFormat == wgpu::TextureFormat::Depth32FloatStencil8);
```

Nonetheless, we are barking up the wrong tree here.

kpiascik:
Closing based on feedback

PR #8845, opened 2025-06-09
webgpu: Pass through defaultDepthStencilFormat

jneljneljnel:
BUGS=[422705735]
using defaultDepthStencilFormat to check stencil loadOP and storeOP

### DIFF
--- a/filament/backend/src/webgpu/WebGPURenderTarget.cpp
+++ b/filament/backend/src/webgpu/WebGPURenderTarget.cpp
@@ -109,7 +109,8 @@ void WebGPURenderTarget::setUpRenderPassAttachments(wgpu::RenderPassDescriptor& 
             };
 
             if (defaultDepthStencilFormat == wgpu::TextureFormat::Depth24PlusStencil8 ||
-                    defaultDepthStencilFormat == wgpu::TextureFormat::Depth32FloatStencil8) {
+                    defaultDepthStencilFormat == wgpu::TextureFormat::Depth32FloatStencil8 ||
+                    defaultDepthStencilFormat == wgpu::TextureFormat::Stencil8) {
                 mDepthStencilAttachmentDescriptor.stencilLoadOp =
                         WebGPURenderTarget::getLoadOperation(params, TargetBufferFlags::STENCIL);
                 mDepthStencilAttachmentDescriptor.stencilStoreOp =
--- a/filament/backend/src/webgpu/WebGPURenderTarget.cpp
+++ b/filament/backend/src/webgpu/WebGPURenderTarget.cpp
@@ -101,20 +101,20 @@ void WebGPURenderTarget::setUpRenderPassAttachments(wgpu::RenderPassDescriptor& 
                 .depthClearValue = static_cast<float>(params.clearDepth),
                 .depthReadOnly =
                         (params.readOnlyDepthStencil & RenderPassParams::READONLY_DEPTH) > 0,
+                .stencilLoadOp = wgpu::LoadOp::Undefined,
+                .stencilStoreOp = wgpu::StoreOp::Undefined,
+                .stencilClearValue = params.clearStencil,
+                .stencilReadOnly =
+                        (params.readOnlyDepthStencil & RenderPassParams::READONLY_STENCIL) > 0,
             };
+
             if (defaultDepthStencilFormat == wgpu::TextureFormat::Depth24PlusStencil8 ||
                     defaultDepthStencilFormat == wgpu::TextureFormat::Depth32FloatStencil8) {
                 mDepthStencilAttachmentDescriptor.stencilLoadOp =
                         WebGPURenderTarget::getLoadOperation(params, TargetBufferFlags::STENCIL);
                 mDepthStencilAttachmentDescriptor.stencilStoreOp =
                         WebGPURenderTarget::getStoreOperation(params, TargetBufferFlags::STENCIL);
-            } else {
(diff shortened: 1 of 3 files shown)


filament/backend/src/webgpu/WebGPURenderTarget.cpp
```diff
@@ -73,6 +73,7 @@ wgpu::StoreOp WebGPURenderTarget::getStoreOperation(RenderPassParams const& para
 void WebGPURenderTarget::setUpRenderPassAttachments(wgpu::RenderPassDescriptor& outDescriptor,
         RenderPassParams const& params, wgpu::TextureView const& defaultColorTextureView,
         wgpu::TextureView const& defaultDepthStencilTextureView,
```
kpiascik, line 75: we're already passing in the view, can't we just call `getFormat()` instead of passing in a new param?

jneljneljnel: I don't think so, as the api for the TextureView doesn't expose that. For the custom depth and stencil views we have to pass the respective custom depth and custom stencil formats, so it made sense to do the same for defaults.

filament/backend/src/webgpu/WebGPURenderTarget.cpp
```diff
@@ -93,30 +94,38 @@ void WebGPURenderTarget::setUpRenderPassAttachments(wgpu::RenderPassDescriptor&
...
+                        WebGPURenderTarget::getLoadOperation(params, TargetBufferFlags::DEPTH),
                 .depthStoreOp =
                         WebGPURenderTarget::getStoreOperation(params, TargetBufferFlags::DEPTH),
                 .depthClearValue = static_cast<float>(params.clearDepth),
                 .depthReadOnly =
                         (params.readOnlyDepthStencil & RenderPassParams::READONLY_DEPTH) > 0,
-                .stencilLoadOp =
-                        WebGPURenderTarget::getLoadOperation(params, TargetBufferFlags::STENCIL),
-                .stencilStoreOp =
-                        WebGPURenderTarget::getStoreOperation(params, TargetBufferFlags::STENCIL),
```
kpiascik, line 105: we could keep the default here to be undefined for load/store. Then below not have any `else` clause.

filament/backend/src/webgpu/WebGPURenderTarget.cpp
```diff
@@ -93,30 +94,38 @@ void WebGPURenderTarget::setUpRenderPassAttachments(wgpu::RenderPassDescriptor&
...
-                .stencilClearValue = params.clearStencil,
-                .stencilReadOnly =
-                        (params.readOnlyDepthStencil & RenderPassParams::READONLY_STENCIL) > 0,
```
kpiascik: We can keep these defaults since they don't change

filament/backend/src/webgpu/WebGPURenderTarget.cpp
jneljneljnel: I agree that's better, thanks.

kpiascik (approved):

jc3265 (approved):

filament/backend/src/webgpu/WebGPURenderTarget.cpp
AndyHovingh: @jneljneljnel is correct. These C++ wgpu classes are essentially RAII wrappers around the C API handles, which get us resource management + ref counting, etc., but they don't provide any additional functionality the actual WebGPU API itself offers; Once cannot get the format from a TextureView object in the spec (https://www.w3.org/TR/webgpu/#gputextureview).

filament/backend/src/webgpu/WebGPURenderTarget.cpp
```diff
@@ -93,30 +94,38 @@ void WebGPURenderTarget::setUpRenderPassAttachments(wgpu::RenderPassDescriptor&
...
                 .depthStoreOp =
                         WebGPURenderTarget::getStoreOperation(params, TargetBufferFlags::DEPTH),
                 .depthClearValue = static_cast<float>(params.clearDepth),
                 .depthReadOnly =
                         (params.readOnlyDepthStencil & RenderPassParams::READONLY_DEPTH) > 0,
-                .stencilLoadOp =
-                        WebGPURenderTarget::getLoadOperation(params, TargetBufferFlags::STENCIL),
-                .stencilStoreOp =
-                        WebGPURenderTarget::getStoreOperation(params, TargetBufferFlags::STENCIL),
+                .stencilLoadOp = wgpu::LoadOp::Undefined,
```
AndyHovingh, line 104: @jneljneljnel this was the regression wasn't it?

filament/backend/src/webgpu/WebGPURenderTarget.cpp
```diff
@@ -93,30 +94,38 @@ void WebGPURenderTarget::setUpRenderPassAttachments(wgpu::RenderPassDescriptor&
...
+                .stencilStoreOp = wgpu::StoreOp::Undefined,
                 .stencilClearValue = params.clearStencil,
                 .stencilReadOnly =
                         (params.readOnlyDepthStencil & RenderPassParams::READONLY_STENCIL) > 0,
             };
+
+            if (defaultDepthStencilFormat == wgpu::TextureFormat::Depth24PlusStencil8 ||
```
AndyHovingh, line 111: For defensiveness (actually it could happen if we don't have depth in a pass, only stencil), should we not also check for `wgpu::TextureFormat::Stencil8`? We might want to have/use a helper/convenience function (if it doesn't already exist) to check if a format has stencil bits instead of doing this in various places.

PR #8849, opened 2025-06-11
Fix: Prevent deadlock when loading programs from cache

z3moon:
When a program was created directly from a cached blob in `ShaderCompilerService`, its associated token was not signaled as ready in the THREAD_POOL mode. This oversight caused a deadlock at the `token->wait()` call during program destruction.
    
This commit resolves the issue by skipping the token's readiness check upon destruction if the program was created from the cache blob.

BUGS=[423221474]

### DIFF
--- a/filament/backend/src/opengl/ShaderCompilerService.cpp
+++ b/filament/backend/src/opengl/ShaderCompilerService.cpp
@@ -122,12 +122,18 @@ struct ShaderCompilerService::OpenGLProgramToken : ProgramToken {
     }
 
     std::optional<CallbackManager::Handle> handle{};
+
+    // Only valid when the blob functions are provided by users. The validity of this variable
+    // doesn't guarantee that the program was created from the cache blob.
     BlobCacheKey key;
 
     // Used for the `THREAD_POOL` mode.
     mutable Mutex lock;
     mutable Condition cond;
     bool signaled = false;
+
+    // Indicate this program was created from the cache blob.
+    bool retrieved = false;
 };
 
 ShaderCompilerService::OpenGLProgramToken::~OpenGLProgramToken() {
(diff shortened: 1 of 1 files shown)


filament/backend/src/opengl/ShaderCompilerService.cpp
```diff
@@ -259,6 +259,9 @@ ShaderCompilerService::program_token_t ShaderCompilerService::createProgram(
     // Try retrieving the cached program blob if available.
     token->gl.program = mBlobCache.retrieve(&token->key, mDriver.mPlatform, program);
     if (token->gl.program) {
+        // `gl.program` is created directly from  the cached blob, so signal its readiness
+        // immediately.
+        token->signal();
```
poweifeng: this is ok with the caveat we talked about.  That we're still checking (and locking) the queue even though we never put an entry in there.

poweifeng (approved):

filament/backend/src/opengl/ShaderCompilerService.cpp
z3moon: Yeah actually I'm thinking of modifying this commit to totally remove the `dequeue` call for the retrieved program.

filament/backend/src/opengl/ShaderCompilerService.cpp
```diff
@@ -122,12 +122,18 @@ struct ShaderCompilerService::OpenGLProgramToken : ProgramToken {
...
+    // doesn't guarantee that the program was created from the cache blob.
     BlobCacheKey key;
 
     // Used for the `THREAD_POOL` mode.
     mutable Mutex lock;
     mutable Condition cond;
     bool signaled = false;
+
+    // Indicate this program was created from the cache blob.
+    bool retrieved = false;
```
poweifeng: nit: I'd name it something more specific like `retrievedFromBlobCache`.

z3moon: Sounds good. Let me update.

PR #8850, opened 2025-06-11
webgpu: texture changes to better support mipmap generation later

AndyHovingh:
BUGS = [423893882]

### DIFF
--- a/filament/backend/src/webgpu/WebGPUTexture.cpp
+++ b/filament/backend/src/webgpu/WebGPUTexture.cpp
@@ -28,6 +28,10 @@
 #include <utils/ostream.h>
 
 #include <webgpu/webgpu_cpp.h>
+#if FWGPU_ENABLED(FWGPU_PRINT_SYSTEM)
+#include <webgpu/webgpu_cpp_print.h>
+#include <sstream>
+#endif
 
 #include <cstdint>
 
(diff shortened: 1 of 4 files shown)


filament/backend/src/webgpu/WebGPUTexture.h
```diff
@@ -47,29 +47,47 @@ class WebGPUTexture : public HwTexture {
 
     [[nodiscard]] wgpu::TextureView getOrMakeTextureView(uint8_t mipLevel, uint32_t arrayLayer);
 
-    [[nodiscard]] wgpu::TextureFormat getWebGPUFormat() const { return mWebGPUFormat; }
+    [[nodiscard]] wgpu::TextureFormat getViewFormat() const { return mViewFormat; }
```
bridgewaterrobbie, line 50: Nit: We might want to have a getTextureFormat too (Assuming I'm not missing it somewhere else) even if we don't yet need it. On the other hand that might be possible to be confused about

filament/backend/src/webgpu/WebGPUTexture.h
```diff
@@ -47,29 +47,47 @@ class WebGPUTexture : public HwTexture {
...
     wgpu::TextureAspect mAspect = wgpu::TextureAspect::Undefined;
     // usage is inherited from HwTexture. This naming is to distinguish it from Filament's usage
     wgpu::TextureUsage mWebGPUUsage = wgpu::TextureUsage::None;
-    uint32_t mArrayLayerCount = 1;
+    wgpu::TextureUsage mViewUsage = wgpu::TextureUsage::None;
     size_t mBlockWidth = 0;
     size_t mBlockHeight = 0;
-    wgpu::TextureView mDefaultTextureView = nullptr;
+    uint32_t mArrayLayerCount = 1;
+    // CreateTextureR has info for a texture and sampler. Texture Views are needed for binding,
```
bridgewaterrobbie: Nit: I know you just moved this from above, but it actually isn't correct anymore with the changes that have been made, and improved understanding. We don't get the needed sampler info until updateDescriptorSetTexture

filament/backend/src/webgpu/WebGPUTexture.cpp
```diff
@@ -59,21 +63,113 @@ namespace {
...
+/**
+ * @param fUsage Filament's requested texture usage
+ * @param samples How many samples to use for MSAA
+ * @param supportsMultipleMipLevels if we support and (potentially) need multiple
+ *                                  mip levels for the texture
+ * @return The appropriate texture usage flags for the underlying texture
+ */
 [[nodiscard]] wgpu::TextureUsage fToWGPUTextureUsage(TextureUsage const& fUsage,
-        const uint8_t samples) {
+        const uint8_t samples, const bool supportsMultipleMipLevels) {
```
bridgewaterrobbie: nit: I'd adjust this variable name to reflect "needsComputeSupport" or similar. The reason we need compute is for mipmap, but conceptually they can be separated at this layer

filament/backend/src/webgpu/WebGPUTexture.cpp
```diff
@@ -59,21 +63,113 @@ namespace {
...
+    if (supportsMultipleMipLevels ||
+            any((TextureUsage::BLIT_DST | TextureUsage::UPLOADABLE) & fUsage)) {
         retUsage |= wgpu::TextureUsage::CopyDst;
     }
-    if (any(TextureUsage::SAMPLEABLE & fUsage)) {
+    if (supportsMultipleMipLevels || any(TextureUsage::SAMPLEABLE & fUsage)) {
         retUsage |= wgpu::TextureUsage::TextureBinding;
     }
+    if (supportsMultipleMipLevels) {
+        retUsage |= wgpu::TextureUsage::StorageBinding;
```
bridgewaterrobbie, line 166: Personally, I'd take a tactic of "If the format supports compute, just put the bit to true for the texture". If I understand right, having the bit set aggressively shouldn't be expensive.

Then, for any default texture views we should take the bit off every time. We don't have any intention of using any views we make here in a compute shader (We pass the whole texture to the mipmap generator and it makes it's own views as appropriate)

filament/backend/src/webgpu/WebGPUTexture.cpp
```diff
@@ -109,6 +204,21 @@ namespace {
...
+ * @param textureUsage The usage flags for the underlying texture for which we want a view
+ * @param viewFormat The format of the texture view
+ * @param textureFormat The format of the underlying texture
+ * @return The texture usage appropriate for the texture view to the underlying texture
+ */
+[[nodiscard]] constexpr wgpu::TextureUsage textureViewUsage(const wgpu::TextureUsage textureUsage,
+        const wgpu::TextureFormat viewFormat, const wgpu::TextureFormat textureFormat) {
+    if (viewFormat != textureFormat && !(isFormatStorageCompatible(viewFormat))) {
+        // remove the storage bit if the view format doesn't support it
+        return (textureUsage & (~wgpu::TextureUsage::StorageBinding));
```
bridgewaterrobbie: As I said above, I'd just unambiguously do `textureUsage & (~wgpu::TextureUsage::StorageBinding)`. We have no intention of ever using our **views** in a compute shader right now.

filament/backend/src/webgpu/WebGPUTexture.cpp
```diff
@@ -455,16 +589,20 @@ wgpu::TextureFormat WebGPUTexture::fToWGPUTextureFormat(TextureFormat const& fFo
 wgpu::TextureView WebGPUTexture::makeTextureView(const uint8_t& baseLevel,
         const uint8_t& levelCount, const uint32_t& baseArrayLayer, const uint32_t& arrayLayerCount,
         const SamplerType samplerType) const noexcept {
+    if (baseLevel > 0 && !mSupportsMultipleMipLevels) {
+        FWGPU_LOGW << "Trying to make a texture view into a level (" << baseLevel
```
bridgewaterrobbie: We discussed in chat, but noting here: IMO we can make this a verbose log of something. In the actual generate mippmap function we can check the usage at the time of use and panic if the bit didn't get set here.

filament/backend/src/webgpu/WebGPUTexture.h
AndyHovingh: We don't currently have a need for it, only the view format.  Right now we use this to get customDepthFormat and customStencilFormat in `WebGPUDriver::beginRenderPass`, where getViewFormat is fine. In fact, even that looks like it will be unnecessary in the scope of changes for bug 423880880. I vote we do not add functions until we need them.

filament/backend/src/webgpu/WebGPUTexture.h
AndyHovingh: Good catch! Thank you! Happily removing.

filament/backend/src/webgpu/WebGPUTexture.cpp
AndyHovingh: Very good point. We don't need this level of coupling. I changed the name to `needsComputeStorageSupport` and adjusted the comments 👍

filament/backend/src/webgpu/WebGPUTexture.cpp
AndyHovingh: Always adding the bit for compute could potentially cause problems for textures we just want to read from, sample, like in any other shader. That limits the possible texture formats a bit. Although, not a big deal now, because we only use the compute shader for one thing. That said, what I changed effectively does what I think you are saying anyway; I changed `supportsMultipleMipLevels` to `needsComputeStorageSupport` and this bit gets set in that case. Seems reasonable for now.

As for always taking the bit off for texture views, I agree. I don't see why we have any reason to add that bit at this stage or in the foreseeable future.

In short, I think we are on the same page 👍

filament/backend/src/webgpu/WebGPUTexture.cpp
AndyHovingh: Yep. Agreed. Removing this function altogether is much cleaner. I just reuse the `fToWGPUTextureUsage` function and set `needsComputeStorageSupport` to `false`.

filament/backend/src/webgpu/WebGPUTexture.cpp
AndyHovingh: absolutely, I have this behind the `FWGPU_DEBUG_VALIDATION` flag now, and I added a `FILAMENT_CHECK_PRECONDITION` in `WebGPUDriver::generateMipmaps`. If that is too aggressive for what we are doing, we can always downgrade that to a warning there.

AndyHovingh:
Thank you @bridgewaterrobbie for your feedback! Merging away! :shipit:

PR #8851, opened 2025-06-11
gltfio: enable passing in variant filters

poweifeng:


libs/gltfio/src/JitShaderProvider.cpp
```diff
@@ -49,15 +52,37 @@ class JitShaderProvider : public MaterialProvider {
...
 
-JitShaderProvider::JitShaderProvider(Engine* engine, bool optimizeShaders) : mEngine(engine),
-        mOptimizeShaders(optimizeShaders) {
+JitShaderProvider::JitShaderProvider(Engine* engine, bool optimizeShaders,
+        utils::FixedCapacityVector<char const*> const& variantFilters)
+    : mEngine(engine),
+      mOptimizeShaders(optimizeShaders) {
+
+    // Note that this is the same as the list in tools/matc/src/ParametersProcessor.cpp
+    static const std::unordered_map<std::string, filament::UserVariantFilterBit> strToEnum  = [] {
```
bejado, line 70: I'd be nice if we could find a way to not have this duplicated. I worry we'll forget to update this when we change variant bits.

poweifeng: yeah, I thought about this too. I'll try to move this into filamat or something similar next week.

PR #8852, opened 2025-06-11
VK: Simple staging bypass for free uniform buffers

rafadevai:
When the uniform VulkanBuffer is currently not in
use and under UMA, just perform a direct memcpy to GPU memory.

Add a flag to control disable this flow and always use staging to update buffers.
The default state is to always use staging.


filament/backend/include/backend/platforms/VulkanPlatform.h
```diff
@@ -152,6 +152,13 @@ class VulkanPlatform : public Platform, utils::PrivateImplementation<VulkanPlatf
          * presentation. Default is true.
          */
         bool transitionSwapChainImageLayoutForPresent = true;
+
+        /**
+         *
+         * Whether the buffer updates always use staging or not. Otherwise
+         * its allowed to also use memcpy for buffer updates. Default is true
+         */
+        bool forceStagingForBufferUpdates = true;
```
poweifeng: why have this flag at all?
Isn't it dictated by whether the backing buffer has mapped memory or not (that is, UMA).

filament/backend/src/vulkan/VulkanBufferProxy.cpp
```diff
@@ -35,7 +35,38 @@ VulkanBufferProxy::VulkanBufferProxy(VmaAllocator allocator, VulkanStagePool& st
       mUpdatedBytes(0) {}
 
 void VulkanBufferProxy::loadFromCpu(VulkanCommandBuffer& commands, const void* cpuData,
-        uint32_t byteOffset, uint32_t numBytes) {
+        uint32_t byteOffset, uint32_t numBytes, bool forceStaging) {
+    // The VulkanBuffer is available if the only object reference is hold by the
+    // `VulkanBufferProxy`. This means that the buffer is not currently in progress to be use or
+    // already in flight.
```
poweifeng: `in flight` here means on the GPU queue ready to be used? Is it safe to do a memcpy then?

filament/backend/include/backend/platforms/VulkanPlatform.h
rafadevai: It's more for debugging purposes in case we want to disable it completely on UMA

filament/backend/src/vulkan/VulkanBufferProxy.cpp
rafadevai: In flight => submitted to the GPU
In progress => we did an update already this frame and now its being tracked by the command buffer (hard to make a decision if we should allow the memcpy or not)

filament/backend/include/backend/platforms/VulkanPlatform.h
poweifeng: then I'd suggest not adding this *here* because this is an API interface. And removing it in the future is not as simple. But perhaps you could add a `FVK_DEBUG` define or the likes.

filament/backend/src/vulkan/VulkanBufferProxy.cpp
poweifeng: I'm still a bit confused by in-flight (or maybe my understanding of GPUS is incomplete).
So when the command buffer is submitted, I still have no guarantee on when an ubo will be read within that submission, is that correct? And "in flight" here only means the commands have been submitted?

filament/backend/include/backend/platforms/VulkanPlatform.h
rafadevai: I changed it for a constexpr value is that okay?

filament/backend/src/vulkan/VulkanBufferProxy.cpp
rafadevai: I updated the comment to say, if its referenced by a VulkanCommanBuffer it wont do a memcpy, that covers all cases

filament/backend/src/vulkan/VulkanConstants.h
```diff
@@ -214,4 +214,8 @@ constexpr static const int FVK_MAX_PIPELINE_AGE = FVK_MAX_COMMAND_BUFFERS;
 // destroying any unused pipeline object.
 static_assert(FVK_MAX_PIPELINE_AGE >= FVK_MAX_COMMAND_BUFFERS);
 
+// Whether the buffer updates always use staging or not. Otherwise its allowed to also use memcpy
+// for buffer updates. Default is false
+constexpr static const bool FVK_FORCE_STAGING_FOR_BUFFER_UPDATES = false;
```
poweifeng, line 219: this lgtm. we can also move it to a Engine feature flag latter, if there's a need.

PR #8853, opened 2025-06-11
webgpu: Use C++ converted mipmap generator. First pass, likely to be tweaked substantially in the future

bridgewaterrobbie:


poweifeng (commented):
Can you fix the formatting in general and then add a README.md about how this came about?

filament/backend/src/webgpu/SpdMipmapGenerator/SpdMipmapGenerator.cpp
```diff
@@ -0,0 +1,795 @@
+#include "SpdMipmapGenerator.h"
+#include <sstream>
+#include <stdexcept>
+// C++ port of https://github.com/JolifantoBambla/webgpu-spd for early experiments- do not merge like this
```
jc3265: not sure if you were going to update this comment _"do not merge like this"_

bridgewaterrobbie: Good point since we decided 🚢  though I want to clean it up further in a followup. To be honest I was surprised it worked this well on the first pass.

PR #8857, opened 2025-06-12
webgpu: fix logging

AndyHovingh:


filament/backend/src/webgpu/WebGPUStrings.h
```diff
@@ -115,54 +92,24 @@ utils::io::ostream& streamInsertWebGPUPrintable(utils::io::ostream& out,
...
- * Example usage (because including the template operator override itself does not compile):
- * template<typename STREAM_TYPE>
- * STREAM_TYPE& operator<<(STREAM_TYPE& out, wgpu::RequestAdapterOptions const& options) noexcept {
- *   return streamInsertRequestAdapterOptions(out, options);
- * }
- */
-template<typename STREAM_TYPE>
-STREAM_TYPE& streamInsertRequestAdapterOptions(STREAM_TYPE& out,
-        wgpu::RequestAdapterOptions const& options) noexcept {
+[[nodiscard]] std::string adapterOptionsToString(wgpu::RequestAdapterOptions const& options) {
```
jc3265: nit IMO this function name and `adapterInfoToString` are a little misleading because you are not converting the "adapterOptions" to a "string" its more of a "getAdapterOptionsDetails" or something like that

jc3265 (approved):

filament/backend/src/webgpu/WebGPUStrings.h
AndyHovingh: I would call this `toString` if the compiler didn't take issue with it. This is a convention elsewhere in Filament, which comes from a Java background, essentially the `toString()` method on Objects. I think "getAdapterOptionsDetails" is more confusing in terms of what it is doing at a glance. Other languages have similar convensions, such as Python's `str()`/`__str__()` and `repr()`/`__repr()__` (which is a different thing, but somewhat related).
I think if I were to change the name here, the same reason would apply to so many other functions, both in the WebGPU-specific code and throughout Filament. Thus, I think I will keep it.
Thank you for your take on it though; I will keep thinking about that.

filament/backend/src/webgpu/WebGPUStrings.h
```diff
@@ -21,50 +21,27 @@
...
- * #endif
- */
 template<typename WebGPUPrintable>
-utils::io::ostream& streamInsertWebGPUPrintable(utils::io::ostream& out,
-        const WebGPUPrintable printable) {
-    std::stringstream printableStream;
-    printableStream << printable;
-    out << printableStream.str();
-    return out;
+[[nodiscard]] std::string webGPUPrintableToString(const WebGPUPrintable printable) {
```
kpiascik: nit: inline otherwise we'll get a linker warning

filament/backend/src/webgpu/WebGPUStrings.h
kpiascik: nit: inline otherwise we'll get a linker warning

filament/backend/src/webgpu/WebGPUStrings.h
AndyHovingh: ok. no problem 👍

filament/backend/src/webgpu/WebGPUStrings.h
AndyHovingh: done 👍

PR #8861, opened 2025-06-12
webgpu: Mipmap driver changes

bridgewaterrobbie:
Note: The SPD changes are still not fully solid, which is known. The changes here should be incrementally better, as we likely do a more full rewrite in the future as time permits. 

BUGS = [423893882]

### DIFF
--- a/filament/backend/CMakeLists.txt
+++ b/filament/backend/CMakeLists.txt
@@ -257,6 +257,7 @@ if (FILAMENT_SUPPORTS_WEBGPU)
     list(APPEND SRCS
             include/backend/platforms/WebGPUPlatform.h
             src/webgpu/platform/WebGPUPlatform.cpp
+            src/webgpu/SpdMipmapGenerator/SpdMipmapGenerator.cpp
             src/webgpu/WebGPUBufferBase.cpp
             src/webgpu/WebGPUBufferBase.h
             src/webgpu/WebGPUBufferObject.cpp
--- a/filament/backend/src/webgpu/SpdMipmapGenerator/SpdMipmapGenerator.cpp
+++ b/filament/backend/src/webgpu/SpdMipmapGenerator/SpdMipmapGenerator.cpp
@@ -86,7 +86,9 @@ MipmapGenerator::MipmapGenerator(const wgpu::Device& device)
     // std::min(4u - 1, 11u) = 3.
     m_maxMipsPerPass =
             std::min(static_cast<uint32_t>(deviceLimits.maxStorageTexturesPerShaderStage) - 1, 11u);
-    
+    m_maxArrayLayers = deviceLimits.maxStorageTexturesPerShaderStage;
+
+
     // If we can generate more than 6 mips in a pass, we need an advanced BGL
     // that includes atomic counter and mid-mip buffer for levels > 5.
     if (m_maxMipsPerPass > 6) {
@@ -222,50 +224,58 @@ SPDPipeline& MipmapGenerator::GetOrCreatePipeline(const PipelineCacheKey& key) {
 
 void MipmapGenerator::Generate(wgpu::CommandEncoder& commandEncoder, wgpu::Texture srcTexture,
         const SPDPassConfig& config) {
-    // config.numMips now represents the total number of desired mip levels for the texture.
-    // config.sourceMipLevel is the starting level (e.g., 0 for a full chain).
     if (config.numMips <= 1 || config.numMips <= config.sourceMipLevel + 1) {
-        return; // Not enough mips to generate, or target is already met/exceeded.
-    }
-
-    if (m_maxMipsPerPass == 0) {
-        // This indicates an issue, e.g., deviceLimits.maxStorageTexturesPerShaderStage <= 1.
-        // Log an error or handle appropriately.
         return;
     }
 
-    // Prepare pipelines once based on the overall configuration for filter and precision.
-    // The actual number of mips per pipeline will be handled by GetOrCreatePipeline.
+    if (m_maxMipsPerPass == 0) {
+        return;
+    }
+
     PreparePipelines(srcTexture.GetFormat(), config.filter, config.halfPrecision);
 
-    uint32_t mipsToGenerateCount = config.numMips - (config.sourceMipLevel + 1);
-    uint32_t currentSourceMipLevel = config.sourceMipLevel;
+    uint32_t totalSrcArrayLayers = srcTexture.GetDepthOrArrayLayers();
 
-    while (mipsToGenerateCount > 0) {
-        uint32_t mipsThisPass = std::min(mipsToGenerateCount, m_maxMipsPerPass);
+    // Loop over array layer chunks
+    for (uint32_t currentBaseArrayLayer = 0; currentBaseArrayLayer < totalSrcArrayLayers;
+            currentBaseArrayLayer += m_maxArrayLayers) {
 
-        SPDPassConfig passSpecificConfig = {}; // Create a new config for this specific pass
-        passSpecificConfig.filter = config.filter;
-        passSpecificConfig.halfPrecision = config.halfPrecision;
-        passSpecificConfig.targetTexture = config.targetTexture ? config.targetTexture : srcTexture;
-        passSpecificConfig.numMips = mipsThisPass; // numMips for this specific pass
-        passSpecificConfig.sourceMipLevel = currentSourceMipLevel;
+        uint32_t arrayLayersThisPass =
+                std::min(m_maxArrayLayers, totalSrcArrayLayers - currentBaseArrayLayer);
 
-        generatePass(commandEncoder, srcTexture, passSpecificConfig);
+        uint32_t mipsToGenerateCount = config.numMips - (config.sourceMipLevel + 1);
+        uint32_t currentSourceMipLevel = config.sourceMipLevel;
 
-        mipsToGenerateCount -= mipsThisPass;
-        currentSourceMipLevel += mipsThisPass;
+        while (mipsToGenerateCount > 0) {
+            uint32_t mipsThisPass = std::min(mipsToGenerateCount, m_maxMipsPerPass);
+
+            SPDPassConfig passSpecificConfig = {};
+            passSpecificConfig.filter = config.filter;
+            passSpecificConfig.halfPrecision = config.halfPrecision;
+            passSpecificConfig.targetTexture =
+                    config.targetTexture ? config.targetTexture : srcTexture;
+            passSpecificConfig.numMips = mipsThisPass;
+            passSpecificConfig.sourceMipLevel = currentSourceMipLevel;
+
+            // Pass baseArrayLayer and numArrayLayers directly to generatePass
+            generatePass(commandEncoder, srcTexture, passSpecificConfig, currentBaseArrayLayer,
+                    arrayLayersThisPass);
+
+            mipsToGenerateCount -= mipsThisPass;
+            currentSourceMipLevel += mipsThisPass;
+        }
     }
 }
 
 void MipmapGenerator::generatePass(wgpu::CommandEncoder& commandEncoder, wgpu::Texture srcTexture,
-        const SPDPassConfig& passConfig) { // Renamed param to passConfig for clarity
+        const SPDPassConfig& passConfig, uint32_t baseArrayLayer, uint32_t numArrayLayers) {
     uint32_t width = srcTexture.GetWidth() >> passConfig.sourceMipLevel;
     uint32_t height = srcTexture.GetHeight() >> passConfig.sourceMipLevel;
-    uint32_t arrayLayerCount = srcTexture.GetDepthOrArrayLayers();
+
+    uint32_t arrayLayerCountForThisPass = numArrayLayers;
 
     wgpu::Texture target = passConfig.targetTexture ? passConfig.targetTexture : srcTexture;
-    uint32_t numMipsThisPass = passConfig.numMips; // This is now specific to this pass
+    uint32_t numMipsThisPass = passConfig.numMips;
 
     if (numMipsThisPass == 0) return;
 
@@ -280,8 +290,8 @@ void MipmapGenerator::generatePass(wgpu::CommandEncoder& commandEncoder, wgpu::T
     srcViewDesc.dimension = wgpu::TextureViewDimension::e2DArray;
     srcViewDesc.baseMipLevel = passConfig.sourceMipLevel;
     srcViewDesc.mipLevelCount = 1;
-    srcViewDesc.baseArrayLayer = 0;
-    srcViewDesc.arrayLayerCount = arrayLayerCount;
+    srcViewDesc.baseArrayLayer = baseArrayLayer;
+    srcViewDesc.arrayLayerCount = arrayLayerCountForThisPass;
 
     wgpu::BindGroupEntry srcEntry{};
     srcEntry.binding = 0;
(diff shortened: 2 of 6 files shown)


filament/backend/src/webgpu/WebGPUDriver.cpp
```diff
@@ -738,67 +738,60 @@ void WebGPUDriver::setExternalStream(Handle<HwTexture> textureHandle,
...
-        };
 
-        wgpu::TexelCopyTextureInfo destinationCopyInfo{
-            .texture = wgpuTexture,
-            .mipLevel = mipLevel + 1,
-            .aspect = texture->getAspect(),
-        };
+    // Determine the maximum number of mips we can generate in a single pass.
+    // The limit is on *storage* textures, and we need 1 binding for the source texture.
+    const uint32_t maxMipsPerPass = 2;//mDeviceLimits.maxStorageTexturesPerShaderStage - 1;
```
jc3265: why not set it to `mDeviceLimits.maxStorageTexturesPerShaderStage - 1;` instead of 2?

bridgewaterrobbie: That's because it wasn't working as desired, so for this WIP stage I set it to 2 as a safe value.

I'm working on removing that issue right now, just wanted this available for others to work past mipmap issues.

filament/backend/src/webgpu/SpdMipmapGenerator/SpdMipmapGenerator.cpp
```diff
@@ -73,69 +73,112 @@ MipmapGenerator::MipmapGenerator(const wgpu::Device& device)
...
     }
 }
 
 SPDScalarType MipmapGenerator::SanitizeScalarType(wgpu::TextureFormat format, bool halfPrecision) {
     std::string formatStr = to_string(format);
     std::transform(formatStr.begin(), formatStr.end(), formatStr.begin(), ::tolower);
 
-    SPDScalarType texelType = SPDScalarType::F32;
+    SPDScalarType texelType = SPDScalarType::F32; // Default to F32
     if (formatStr.find("sint") != std::string::npos) {
```
jc3265, line 143: outside of this PR but should we create functions for this
``` 
isSignedint(wgpu::TextureFormat format)
case (format)
   sint cases
       return true
return false
```

filament/backend/src/webgpu/SpdMipmapGenerator/SpdMipmapGenerator.cpp
```diff
@@ -73,69 +73,112 @@ MipmapGenerator::MipmapGenerator(const wgpu::Device& device)
...
+    // further capped at 11 output mips (which means 12 total texture bindings: 1 source + 11
+    // outputs). WebGPU minspec for maxStorageTexturesPerShaderStage is 4, so
+    // deviceLimits.maxStorageTexturesPerShaderStage >= 4. Thus, m_maxMipsPerPass will be at least
+    // std::min(4u - 1, 11u) = 3.
+    m_maxMipsPerPass =
+            std::min(static_cast<uint32_t>(deviceLimits.maxStorageTexturesPerShaderStage) - 1, 11u);
+    
+    // If we can generate more than 6 mips in a pass, we need an advanced BGL
+    // that includes atomic counter and mid-mip buffer for levels > 5.
+    if (m_maxMipsPerPass > 6) {
```
jc3265, line 94: should we make `6` a const somewhere? i think we are using it on line 202

filament/backend/src/webgpu/SpdMipmapGenerator/SpdMipmapGenerator.cpp
```diff
@@ -166,32 +214,71 @@ SPDPipeline& MipmapGenerator::GetOrCreatePipeline(wgpu::TextureFormat format, SP
         pipelineDesc.compute.entryPoint = "downsample";
 
         spdPipeline.pipeline = m_device.CreateComputePipeline(&pipelineDesc);
-        m_pipelines[format][scalarType][filter][numMips] = std::move(spdPipeline);
+        m_pipelines[key] = std::move(spdPipeline);
```
jc3265, line 219: +1 this is wayyyy more readable than what it was

filament/backend/src/webgpu/SpdMipmapGenerator/SpdMipmapGenerator.cpp
bridgewaterrobbie: It could be considered, but I don't think this code is going to be long lived. Just trying to stomp out all issues with compute shader based mip map generation in general right now.

filament/backend/src/webgpu/SpdMipmapGenerator/SpdMipmapGenerator.cpp
bridgewaterrobbie: Yeah that initial mess is like a textbook example of "Just because it works doesn't mean we should trust it", and I know there are further improvements to be had. I just wanted to fix the most egregious ones.

filament/backend/src/webgpu/SpdMipmapGenerator/SpdMipmapGenerator.cpp
bridgewaterrobbie: Honestly this should be converted to a switch statement if this code survives long enough. It's converting the format to a string so that the original typescript logic can do the checking, which is not ideal.

filament/backend/CMakeLists.txt
```diff
@@ -286,6 +286,7 @@ if (FILAMENT_SUPPORTS_WEBGPU)
             src/webgpu/WebGPUVertexBuffer.h
             src/webgpu/WebGPUVertexBufferInfo.cpp
             src/webgpu/WebGPUVertexBufferInfo.h
+            src/webgpu/SpdMipmapGenerator/SpdMipmapGenerator.cpp
```
AndyHovingh: nit: to keep sorted order we should probably list this under `src/webgpu/platform/...`

filament/backend/src/webgpu/WebGPUDriver.h
```diff
@@ -34,7 +34,7 @@
 
 #include <cstdint>
 #include <memory>
-
+#include "SpdMipmapGenerator/SpdMipmapGenerator.h"
```
AndyHovingh: nit: IMO this should be in the 3rd party library group with robin_map.h and webgpu_cpp.h and sorted in alphabetical order in the group.

filament/backend/src/webgpu/platform/WebGPUPlatform.cpp
```diff
@@ -628,7 +628,16 @@ wgpu::Device WebGPUPlatform::requestDevice(wgpu::Adapter const& adapter) {
     deviceDescriptor.defaultQueue.label = "default_queue";
     deviceDescriptor.requiredFeatureCount = enabledFeatures.size();
     deviceDescriptor.requiredFeatures = enabledFeatures.data();
-    deviceDescriptor.requiredLimits = &REQUIRED_LIMITS;
+
+    // It is helpful to increase maxStorageTexturesPerShaderStage if the hardware supports it, but
+    // not required. If we have more cases like this we can generalize it in the future.
+    wgpu::Limits supportedLimits{};
+    adapter.GetLimits(&supportedLimits);
```
AndyHovingh: We need to handle the case where `GetLimits(..)` returns/completes in error.

filament/backend/src/webgpu/platform/WebGPUPlatform.cpp
```diff
@@ -628,7 +628,16 @@ wgpu::Device WebGPUPlatform::requestDevice(wgpu::Adapter const& adapter) {
...
+    auto actualLimits = REQUIRED_LIMITS;
```
AndyHovingh: nit: can we have a little more self-describing name? `actual` is a little vague. Maybe something like `limitsToRequest`?

filament/backend/src/webgpu/platform/WebGPUPlatform.cpp
```diff
@@ -628,7 +628,16 @@ wgpu::Device WebGPUPlatform::requestDevice(wgpu::Adapter const& adapter) {
...
+    actualLimits.maxStorageTexturesPerShaderStage =
+            std::min(12u, supportedLimits.maxStorageTexturesPerShaderStage);
```
AndyHovingh: magic number: why `12` specifically? I suggest defining this as a constant, even it if just local to this file, where constant has a self describing name.

AndyHovingh: Another idea is to have another variable like `REQUIRED_LIMITS`,  `OPTIONAL_LIMITS` or something like that. And, we can more generically apply each such limit like this when determining the limits to request. That could be in a different PR, time and priority permitting _(or if we run into another use case)_. I see your comment above about generalizing it, which this could be. Thus, feel free to ignore :)

filament/backend/src/webgpu/WebGPUDriver.cpp
```diff
@@ -28,7 +28,6 @@
 #include "WebGPUVertexBuffer.h"
 #include "WebGPUVertexBufferInfo.h"
 #include <backend/platforms/WebGPUPlatform.h>
-
```
AndyHovingh: this space belongs here. It separates the group of WebGPU-specific includes from Filament backend API includes _(different layers called out in Filament's code style)_.

filament/backend/src/webgpu/WebGPUDriver.cpp
```diff
@@ -65,10 +64,11 @@ Driver* WebGPUDriver::create(WebGPUPlatform& platform, const Platform::DriverCon
 WebGPUDriver::WebGPUDriver(WebGPUPlatform& platform,
         const Platform::DriverConfig& driverConfig) noexcept
     : mPlatform(platform),
+      mAdapter(mPlatform.requestAdapter(nullptr)),
+      mDevice(mPlatform.requestDevice(mAdapter)),
```
AndyHovingh: I think we can add `mQueue` to the constructor initializer list after `mDevice` here, removing it from the body below, which is a minor improvement.

BTW thank you for moving these into the constructor initializer list :)

filament/backend/src/webgpu/platform/WebGPUPlatform.cpp
bridgewaterrobbie: Just panic, right? I was assuming a failure here wouldn't be possible since it has to be invoked when closing an adapter to get this far (A perfectly performant startup might cache the first invocation, but I didn't want to refactor all the code for this)

filament/backend/src/webgpu/platform/WebGPUPlatform.cpp
bridgewaterrobbie: Yeah a refactor could make senes if this kind of change grows, but given it was just one limit, that seemed like a bit of a large refactor for something so small, such that it would be deprioritized. 

12 is the maximum number of attachments the SPD can make use of effectively, so more than that does not help us.  I'll make a change so that fact is reflected in code.

filament/backend/src/webgpu/platform/WebGPUPlatform.cpp
bridgewaterrobbie: I can change it, though to my mind "actual" is more descriptive- We attach it to "Required Limits" which makes sense from the webgpu perspective, but less from a reader perspective IMO for this optional case.

filament/backend/src/webgpu/WebGPUDriver.cpp
```diff
@@ -65,10 +64,11 @@ Driver* WebGPUDriver::create(WebGPUPlatform& platform, const Platform::DriverCon
 WebGPUDriver::WebGPUDriver(WebGPUPlatform& platform,
         const Platform::DriverConfig& driverConfig) noexcept
     : mPlatform(platform),
+      mAdapter(mPlatform.requestAdapter(nullptr)),
```
AndyHovingh: nit: I know this was like this before this change, but ideally we should prefer to use uniform/list initialization over parentheses notation. This is a little more modern way to do things since C++11 and it has certain benefits, such as preventing narrowing conversions, as well as offering a more consistent syntax across a given C++ code base. Frankly, I would prefer if we would use this kind of initialization everywhere we can, even local variables etc. The parentheses is the traditional/older way to do things.

Thus, I would recommend:
```cpp
WebGPUDriver::WebGPUDriver(WebGPUPlatform& platform,
        const Platform::DriverConfig& driverConfig) noexcept
    : mPlatform{ platform },
      mAdapter { mPlatform.requestAdapter(nullptr) },
     ...
```

filament/backend/src/webgpu/WebGPUDriver.cpp
bridgewaterrobbie: Will bake it into the rest. I did this change mostly to avoid a default constructed mMipMapGenerator, which needs the mDevice.

filament/backend/src/webgpu/WebGPUDriver.cpp
```diff
@@ -738,67 +738,41 @@ void WebGPUDriver::setExternalStream(Handle<HwTexture> textureHandle,
...
     assert_invariant(texture);
     wgpu::Texture wgpuTexture = texture->getTexture();
     assert_invariant(wgpuTexture);
 
-    FILAMENT_CHECK_PRECONDITION(wgpuTexture.GetUsage() & wgpu::TextureUsage::CopySrc)
-            << "Texture intended for mipmap generation (as source) must have CopySrc usage.";
-    FILAMENT_CHECK_PRECONDITION(wgpuTexture.GetUsage() & wgpu::TextureUsage::CopyDst)
-            << "Texture intended for mipmap generation (as destination) must have CopyDst usage.";
+    const auto usage = wgpuTexture.GetUsage();
+    FILAMENT_CHECK_PRECONDITION(usage & wgpu::TextureUsage::TextureBinding)
```
AndyHovingh, line 748: I am not sure explicitly checking the texture and storage binding support here and _not_ the texture's `supportsMultipleMipLevels()` function is the best approach.

On the one hand, this provides better and more useful error messaging feedback, which is good _(as opposed to just checking `supportsMultipleMipLevels()` and potentially failing without a more concrete reason for why it is not supported)_.

However, these functions can get out of sync; both this `WebGPUDriver::generateMipmaps(...)` and `WebGPUTexture::WebGPUTexture(...)` independently make a determination of whether mipmaps can be generated for the texture. By using the `supportsMultipleMipLevels()` function we keep the code DRY.

Perhaps there is a best of both worlds here, where we reuse the same logic for determining if mipmap generation is supported, and we provide meaningful error messaging at the right place when we try to do said generation when we know we can't?

bridgewaterrobbie: IMO having a check before we actually do the mipmap generation against the wgpu-level concepts (The actual usage) is the ideal, and `supportsMultipleMipLevels` is better for determining if we **should** set those at texture creation time based on the format. This approach also ensures we catch any mismatch.

One potential "Both worlds" approach I could see is to note the result of supportsMultipleMipLevels in the postcondition check log, just to help sanity check a failure here?

filament/backend/src/webgpu/platform/WebGPUPlatform.cpp
AndyHovingh: I think you right that the likelihood of failure here is very low. Nonetheless, we should never have silent failures. I think panicking here is probably fine for now 👍

filament/backend/src/webgpu/WebGPUDriver.cpp
```diff
@@ -738,67 +738,41 @@ void WebGPUDriver::setExternalStream(Handle<HwTexture> textureHandle,
...
-        };
+    // We will record all passes into a single command encoder.
+    wgpu::CommandEncoderDescriptor encoderDesc = {};
+    encoderDesc.label = "Mipmap Command Encoder";
+    wgpu::CommandEncoder encoder = mDevice.CreateCommandEncoder(&encoderDesc);
 
-        uint32_t dstWidth = std::max(1u, width >> 1);
-        uint32_t dstHeight = std::max(1u, height >> 1);
-        uint32_t dstDepth = std::max(1u, depth >> 1);
+    spd::SPDPassConfig spdConfig = {};
```
AndyHovingh: I would recommend using a designated initializer here

filament/backend/src/webgpu/WebGPUDriver.cpp
bridgewaterrobbie: Adjusted log as described

PR #8863, opened 2025-06-13
webgpu: Use Target buffer flags for depth and stencil

jneljneljnel:
BUGS = [423880880]

### DIFF
--- a/filament/backend/src/webgpu/WebGPUDriver.cpp
+++ b/filament/backend/src/webgpu/WebGPUDriver.cpp
@@ -456,7 +456,7 @@ void WebGPUDriver::createRenderTargetR(Handle<HwRenderTarget> renderTargetHandle
         const uint8_t samples, const uint8_t layerCount, const MRT color,
         const TargetBufferInfo depth, const TargetBufferInfo stencil) {
     constructHandle<WebGPURenderTarget>(renderTargetHandle, width, height, samples, layerCount,
-            color, depth, stencil);
+            color, depth, stencil, targetFlags);
 }
 
 void WebGPUDriver::createFenceR(Handle<HwFence> fenceHandle, const int /* dummy */) {
@@ -823,10 +823,9 @@ void WebGPUDriver::beginRenderPass(Handle<HwRenderTarget> renderTargetHandle,
 
     std::array<wgpu::TextureView, MRT::MAX_SUPPORTED_RENDER_TARGET_COUNT> customColorViews{};
     uint32_t customColorViewCount = 0;
-    wgpu::TextureView customDepthView = nullptr;
-    wgpu::TextureFormat customDepthFormat = wgpu::TextureFormat::Undefined;
-    wgpu::TextureFormat customStencilFormat = wgpu::TextureFormat::Undefined;
-    wgpu::TextureView customStencilView = nullptr;
+
+    wgpu::TextureView customDepthStencilView = nullptr;
+    wgpu::TextureFormat customDepthStencilFormat = wgpu::TextureFormat::Undefined;
 
     mCurrentRenderTarget = renderTarget;
     if (renderTarget->isDefaultRenderTarget()) {
@@ -834,6 +833,24 @@ void WebGPUDriver::beginRenderPass(Handle<HwRenderTarget> renderTargetHandle,
         defaultColorView = mTextureView;
         defaultDepthStencilView = mSwapChain->getDepthTextureView();
         defaultDepthStencilFormat = mSwapChain->getDepthFormat();
+
+        if (any(renderTarget->getTargets() & TargetBufferFlags::STENCIL) &&
+            !(defaultDepthStencilFormat == wgpu::TextureFormat::Depth24PlusStencil8 ||
+              defaultDepthStencilFormat == wgpu::TextureFormat::Depth32FloatStencil8 ||
+              defaultDepthStencilFormat == wgpu::TextureFormat::Stencil8)) {
+            FILAMENT_CHECK_POSTCONDITION(false) << "Default render target requested stencil, but swap chain's depth format "
+                                                << (uint32_t)defaultDepthStencilFormat << " does not have a stencil aspect.";
+        }
+
+        if (any(renderTarget->getTargets() & TargetBufferFlags::DEPTH) &&
+            !(defaultDepthStencilFormat == wgpu::TextureFormat::Depth16Unorm ||
+              defaultDepthStencilFormat == wgpu::TextureFormat::Depth32Float ||
+              defaultDepthStencilFormat == wgpu::TextureFormat::Depth24Plus ||
+              defaultDepthStencilFormat == wgpu::TextureFormat::Depth24PlusStencil8 ||
+              defaultDepthStencilFormat == wgpu::TextureFormat::Depth32FloatStencil8)) {
+            FILAMENT_CHECK_POSTCONDITION(false) << "Default render target requested depth, but swap chain's depth format "
+                                                << (uint32_t)defaultDepthStencilFormat << " does not have a depth aspect.";
+        }
     } else {
         // Resolve views for custom render target
         const auto& colorInfos = renderTarget->getColorAttachmentInfos();
@@ -853,36 +870,55 @@ void WebGPUDriver::beginRenderPass(Handle<HwRenderTarget> renderTargetHandle,
                 }
             }
         }
-
         const auto& depthInfo = renderTarget->getDepthAttachmentInfo();
+        const auto& stencilInfo = renderTarget->getStencilAttachmentInfo();
+
+        Handle<HwTexture> depthStencilSourceHandle = {};
+        uint8_t dsMipLevel = 0;
+        uint32_t dsArrayLayer = 0;
+
         if (depthInfo.handle) {
-            FILAMENT_CHECK_POSTCONDITION(depthInfo.layer < renderTarget->getLayerCount())
-                    << "Depth attachment requests layer " << depthInfo.layer
-                    << "but render target has only " << renderTarget->getLayerCount() << ".";
-            auto depthTexture = handleCast<WebGPUTexture>(depthInfo.handle);
-            if (depthTexture) {
-                const uint8_t depthMipLevel = depthInfo.level;
-                const uint32_t depthArrayLayer = depthInfo.layer;
-                customDepthView =
-                        depthTexture->getOrMakeTextureView(depthMipLevel, depthArrayLayer);
-                customDepthFormat = depthTexture->getViewFormat();
-            }
+            depthStencilSourceHandle = depthInfo.handle;
+            dsMipLevel = depthInfo.level;
+            dsArrayLayer = depthInfo.layer;
+        } else if (stencilInfo.handle) {
+            depthStencilSourceHandle = stencilInfo.handle;
+            dsMipLevel = stencilInfo.level;
+            dsArrayLayer = stencilInfo.layer;
         }
 
-        const auto& stencilInfo = renderTarget->getStencilAttachmentInfo();
-        if (stencilInfo.handle) {
-            // If depth and stencil use the same texture handle, this will re-cast but that's fine.
-            auto stencilTexture = handleCast<WebGPUTexture>(stencilInfo.handle);
-            if (stencilTexture) {
-                FILAMENT_CHECK_POSTCONDITION(stencilInfo.layer < renderTarget->getLayerCount())
-                        << "Stencil attachment requests layer " << stencilInfo.layer
-                        << " but render target has only " << renderTarget->getLayerCount()
-                        << " layers.";
-                const uint8_t stencilMipLevel = stencilInfo.level;
-                const uint32_t stencilArrayLayer = stencilInfo.layer;
-                customStencilView =
-                        stencilTexture->getOrMakeTextureView(stencilMipLevel, stencilArrayLayer);
-                customStencilFormat = stencilTexture->getViewFormat();
+        if (depthStencilSourceHandle) {
+            auto dsTexture = handleCast<WebGPUTexture>(depthStencilSourceHandle);
+            if (dsTexture) {
+                customDepthStencilView = dsTexture->getOrMakeTextureView(dsMipLevel, dsArrayLayer);
+                customDepthStencilFormat = dsTexture->getViewFormat();
+
+                if (any(renderTarget->getTargets() & TargetBufferFlags::STENCIL) &&
+                        !(customDepthStencilFormat == wgpu::TextureFormat::Depth24PlusStencil8 ||
+                                customDepthStencilFormat ==
+                                        wgpu::TextureFormat::Depth32FloatStencil8 ||
+                                customDepthStencilFormat ==
+                                        wgpu::TextureFormat::Stencil8)) {
+                    FILAMENT_CHECK_POSTCONDITION(false)
+                            << "Custom render target requested stencil, but the provided texture"
+                               "format "
+                            << (uint32_t) customDepthStencilFormat
+                            << " does not have a stencil aspect.";
+                }
+                if (any(renderTarget->getTargets() & TargetBufferFlags::DEPTH) &&
+                        !(customDepthStencilFormat == wgpu::TextureFormat::Depth16Unorm ||
+                                customDepthStencilFormat == wgpu::TextureFormat::Depth32Float ||
+                                customDepthStencilFormat ==
+                                        wgpu::TextureFormat::Depth24Plus || // Modern WebGPU
+                                customDepthStencilFormat ==
+                                        wgpu::TextureFormat::Depth24PlusStencil8 ||
+                                customDepthStencilFormat ==
+                                        wgpu::TextureFormat::Depth32FloatStencil8)) {
+                    FILAMENT_CHECK_POSTCONDITION(false) << "Custom render target requested depth, "
+                                                           "but the provided texture format "
+                                                        << (uint32_t) customDepthStencilFormat
+                                                        << " does not have a depth aspect.";
+                }
             }
         }
     }
@@ -894,10 +930,10 @@ void WebGPUDriver::beginRenderPass(Handle<HwRenderTarget> renderTargetHandle,
             defaultDepthStencilFormat,
             customColorViews.data(),
             customColorViewCount,
-            customDepthView,
-            customStencilView,
-            customDepthFormat,
-            customStencilFormat);
+            customDepthStencilView,
+            customDepthStencilView,
+            customDepthStencilFormat,
+            customDepthStencilFormat);
 
     mRenderPassEncoder = mCommandEncoder.BeginRenderPass(&renderPassDescriptor);
 
@@ -932,9 +968,23 @@ void WebGPUDriver::makeCurrent(Handle<HwSwapChain> drawSch, Handle<HwSwapChain> 
     wgpu::Extent2D surfaceSize = mPlatform.getSurfaceExtent(mNativeWindow);
     mTextureView = mSwapChain->getCurrentSurfaceTextureView(surfaceSize);
     assert_invariant(mTextureView);
-    wgpu::CommandEncoderDescriptor commandEncoderDescriptor = {
-        .label = "command_encoder"
-    };
+
+    assert_invariant(mDefaultRenderTarget);
+
+    wgpu::TextureFormat depthFormat = mSwapChain->getDepthFormat();
+    mDefaultRenderTarget->mTargetFlags = filament::backend::TargetBufferFlags::NONE;
+
+    mDefaultRenderTarget->mTargetFlags |= filament::backend::TargetBufferFlags::COLOR;
+
+    if (depthFormat != wgpu::TextureFormat::Undefined) {
+        mDefaultRenderTarget->mTargetFlags |= filament::backend::TargetBufferFlags::DEPTH;
+
+        if (depthFormat == wgpu::TextureFormat::Depth24PlusStencil8 ||
+                depthFormat == wgpu::TextureFormat::Depth32FloatStencil8) {
+            mDefaultRenderTarget->mTargetFlags |= filament::backend::TargetBufferFlags::STENCIL;
+        }
+    }
+    wgpu::CommandEncoderDescriptor commandEncoderDescriptor = { .label = "command_encoder" };
     mCommandEncoder = mDevice.CreateCommandEncoder(&commandEncoderDescriptor);
     assert_invariant(mCommandEncoder);
 }
@@ -1069,14 +1119,15 @@ void WebGPUDriver::bindPipeline(PipelineState const& pipelineState) {
     assert_invariant(vertexBufferInfo);
 
     std::vector<wgpu::TextureFormat> pipelineColorFormats;
-    wgpu::TextureFormat pipelineDepthFormat = wgpu::TextureFormat::Undefined;
+    wgpu::TextureFormat pipelineDepthStencilFormat = wgpu::TextureFormat::Undefined;
     uint8_t pipelineSamples = 1;
+    bool requestedStencil = false;
 
     if (mCurrentRenderTarget->isDefaultRenderTarget()) {
         pipelineColorFormats.push_back(mSwapChain->getColorFormat());
-        pipelineDepthFormat = mSwapChain->getDepthFormat();
-        pipelineSamples =
-                mCurrentRenderTarget->getSamples();// Default RT should have samples (usually 1)
+        pipelineDepthStencilFormat = mSwapChain->getDepthFormat();
+        pipelineSamples = mCurrentRenderTarget->getSamples();
+        requestedStencil = any(mCurrentRenderTarget->getTargets() & TargetBufferFlags::STENCIL);
     } else {
         const auto& mrtColorAttachments = mCurrentRenderTarget->getColorAttachmentInfos();
         for (size_t i = 0; i < MRT::MAX_SUPPORTED_RENDER_TARGET_COUNT; ++i) {
@@ -1090,21 +1141,21 @@ void WebGPUDriver::bindPipeline(PipelineState const& pipelineState) {
 
         const auto& depthInfo = mCurrentRenderTarget->getDepthAttachmentInfo();
         const auto& stencilInfo = mCurrentRenderTarget->getStencilAttachmentInfo();
+        Handle<HwTexture> dsHandle = {};
         if (depthInfo.handle) {
-            FILAMENT_CHECK_POSTCONDITION(!stencilInfo.handle)
-                    << "depth and stencil attachments cannot both be provided for WebGPU";
-            const auto depthTexture = handleCast<WebGPUTexture>(depthInfo.handle);
-            if (depthTexture) pipelineDepthFormat = depthTexture->getTexture().GetFormat();
-        } else {
-            if (stencilInfo.handle) {
-                const auto stencilTexture = handleCast<WebGPUTexture>(stencilInfo.handle);
-                // Assuming combined depth/stencil format if only stencil is present
-                if (stencilTexture) {
-                    pipelineDepthFormat = stencilTexture->getTexture().GetFormat();
-                }
+            dsHandle = depthInfo.handle;
+        } else if (stencilInfo.handle) {
+             dsHandle = stencilInfo.handle;
+        }
+
+        if (dsHandle) {
+            const auto dsTexture = handleCast<WebGPUTexture>(dsHandle);
+            if (dsTexture) {
+                pipelineDepthStencilFormat = dsTexture->getTexture().GetFormat();
             }
         }
         pipelineSamples = mCurrentRenderTarget->getSamples();
+        requestedStencil = any(mCurrentRenderTarget->getTargets() & TargetBufferFlags::STENCIL);
     }
 
     // TODO: We expected this to be a sane check, however it complains when running shadowtest.
@@ -1116,7 +1167,7 @@ void WebGPUDriver::bindPipeline(PipelineState const& pipelineState) {
     wgpu::RenderPipeline pipeline = createWebGPURenderPipeline(mDevice, *program, *vertexBufferInfo,
             layout, pipelineState.rasterState, pipelineState.stencilState,
             pipelineState.polygonOffset, pipelineState.primitiveType, pipelineColorFormats,
-            pipelineDepthFormat, pipelineSamples);
+            pipelineDepthStencilFormat, pipelineSamples, requestedStencil);
     assert_invariant(pipeline);
     mPipelineMap[hash] = pipeline;
     mRenderPassEncoder.SetPipeline(pipeline);
--- a/filament/backend/src/webgpu/WebGPUDriver.cpp
+++ b/filament/backend/src/webgpu/WebGPUDriver.cpp
@@ -452,7 +452,7 @@ void WebGPUDriver::createDefaultRenderTargetR(Handle<HwRenderTarget> renderTarge
 }
 
 void WebGPUDriver::createRenderTargetR(Handle<HwRenderTarget> renderTargetHandle,
-        const TargetBufferFlags targets, const uint32_t width, const uint32_t height,
+        const TargetBufferFlags targetFlags, const uint32_t width, const uint32_t height,
         const uint8_t samples, const uint8_t layerCount, const MRT color,
         const TargetBufferInfo depth, const TargetBufferInfo stencil) {
     constructHandle<WebGPURenderTarget>(renderTargetHandle, width, height, samples, layerCount,
@@ -834,7 +834,7 @@ void WebGPUDriver::beginRenderPass(Handle<HwRenderTarget> renderTargetHandle,
         defaultDepthStencilView = mSwapChain->getDepthTextureView();
         defaultDepthStencilFormat = mSwapChain->getDepthFormat();
 
-        if (any(renderTarget->getTargets() & TargetBufferFlags::STENCIL) &&
+        if (any(renderTarget->getTargetFlags() & TargetBufferFlags::STENCIL) &&
             !(defaultDepthStencilFormat == wgpu::TextureFormat::Depth24PlusStencil8 ||
               defaultDepthStencilFormat == wgpu::TextureFormat::Depth32FloatStencil8 ||
               defaultDepthStencilFormat == wgpu::TextureFormat::Stencil8)) {
@@ -842,7 +842,7 @@ void WebGPUDriver::beginRenderPass(Handle<HwRenderTarget> renderTargetHandle,
                                                 << (uint32_t)defaultDepthStencilFormat << " does not have a stencil aspect.";
         }
 
-        if (any(renderTarget->getTargets() & TargetBufferFlags::DEPTH) &&
+        if (any(renderTarget->getTargetFlags() & TargetBufferFlags::DEPTH) &&
             !(defaultDepthStencilFormat == wgpu::TextureFormat::Depth16Unorm ||
               defaultDepthStencilFormat == wgpu::TextureFormat::Depth32Float ||
               defaultDepthStencilFormat == wgpu::TextureFormat::Depth24Plus ||
@@ -893,7 +893,7 @@ void WebGPUDriver::beginRenderPass(Handle<HwRenderTarget> renderTargetHandle,
                 customDepthStencilView = dsTexture->getOrMakeTextureView(dsMipLevel, dsArrayLayer);
                 customDepthStencilFormat = dsTexture->getViewFormat();
 
-                if (any(renderTarget->getTargets() & TargetBufferFlags::STENCIL) &&
+                if (any(renderTarget->getTargetFlags() & TargetBufferFlags::STENCIL) &&
                         !(customDepthStencilFormat == wgpu::TextureFormat::Depth24PlusStencil8 ||
                                 customDepthStencilFormat ==
                                         wgpu::TextureFormat::Depth32FloatStencil8 ||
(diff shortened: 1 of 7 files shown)


filament/backend/src/webgpu/WebGPUDriver.cpp
```diff
@@ -452,7 +452,7 @@ void WebGPUDriver::createRenderTargetR(Handle<HwRenderTarget> renderTargetHandle
         const uint8_t samples, const uint8_t layerCount, const MRT color,
         const TargetBufferInfo depth, const TargetBufferInfo stencil) {
     constructHandle<WebGPURenderTarget>(renderTargetHandle, width, height, samples, layerCount,
-            color, depth, stencil);
+            color, depth, stencil, targetFlags);
```
AndyHovingh, line 459: I'm getting a compilation error trying to find `targetFlags`. I suspect you want to rename the `targets` parameter above to `targetFlags`?

filament/backend/src/webgpu/WebGPUSwapChain.h
```diff
@@ -36,6 +36,8 @@ class WebGPUSwapChain final : public Platform::SwapChain, HwSwapChain {
 
     [[nodiscard]] wgpu::TextureFormat getDepthFormat() const { return mDepthFormat; }
 
+    [[nodiscard]] bool getNeedStencil() const { return mNeedStencil; }
```
AndyHovingh: This getter appears not to be used/needed. I would suggest removing it. If we kept it, I would recommend a more concise name, e.g. `needsStencil()`.

filament/backend/src/webgpu/WebGPURenderTarget.h
```diff
@@ -34,25 +34,23 @@ class WebGPURenderTarget : public HwRenderTarget {
 
     WebGPURenderTarget(uint32_t width, uint32_t height, uint8_t samples, uint8_t layerCount,
             MRT const& colorAttachments, Attachment const& depthAttachment,
-            Attachment const& stencilAttachment);
+            Attachment const& stencilAttachment, TargetBufferFlags targetFlags);
 
     // Default constructor for the default render target
     WebGPURenderTarget();
 
     // Updated signature: takes resolved views for custom RTs, and default views for default RT
```
AndyHovingh: nit: I know this is not related to your change, but I think we can safely remove this comment; it doesn't help.

filament/backend/src/webgpu/WebGPURenderTarget.h
```diff
@@ -69,6 +67,10 @@ class WebGPURenderTarget : public HwRenderTarget {
     [[nodiscard]] static wgpu::StoreOp getStoreOperation(RenderPassParams const& params,
             TargetBufferFlags buffer);
 
+    [[nodiscard]] TargetBufferFlags getTargets() const { return mTargetFlags; }
```
AndyHovingh: for clarity could this be `getTargetFlags()`?

filament/backend/src/webgpu/WebGPURenderTarget.h
```diff
@@ -69,6 +67,10 @@ class WebGPURenderTarget : public HwRenderTarget {
...
+
+    TargetBufferFlags mTargetFlags = TargetBufferFlags::NONE;
```
AndyHovingh: This should be `private`.

filament/backend/src/webgpu/WebGPUPipelineCreation.h
```diff
@@ -41,7 +41,7 @@ class WebGPUProgram;
         WebGPUProgram const&, WebGPUVertexBufferInfo const&, wgpu::PipelineLayout const&,
         RasterState const&, StencilState const&, PolygonOffset const&, PrimitiveType primitiveType,
         std::vector<wgpu::TextureFormat> const& colorFormats, wgpu::TextureFormat depthFormat,
-        uint8_t samplesCount);
+        uint8_t samplesCount, bool filamentRequestedStencil);
```
AndyHovingh: I think the `filament` part of the name is superfluous.

AndyHovingh: If we pass a flag in here about whether we need stencil bits, we should also pass a flag for whether you need depth bits (you can technically have depth only, stencil only, depth+stencil, or neither).

filament/backend/src/webgpu/WebGPURenderTarget.cpp
```diff
@@ -32,8 +32,9 @@ namespace filament::backend {
 
 WebGPURenderTarget::WebGPURenderTarget(const uint32_t width, const uint32_t height,
         const uint8_t samples, const uint8_t layerCount, MRT const& colorAttachmentsMRT,
-        Attachment const& depthAttachmentInfo, Attachment const& stencilAttachmentInfo)
+        Attachment const& depthAttachmentInfo, Attachment const& stencilAttachmentInfo, TargetBufferFlags targetFlags)
```
AndyHovingh: I would make `targetFlags` `const` here.

filament/backend/src/webgpu/WebGPURenderTarget.cpp
```diff
@@ -32,8 +32,9 @@ namespace filament::backend {
...
     : HwRenderTarget{ width, height },
+      mTargetFlags(targetFlags),
```
AndyHovingh: we should use curly braces (unity/list initialization). This is more consistent with the rest of the code around it, but it also generally a better idea, as doing so (as opposed to paratheses) can avoid certain bugs due to narrowing conversions. It is also more consistent syntactically with modern C++ (post C++11).

e.g. `mTargetFlags{ targetFlags },`

filament/backend/src/webgpu/WebGPURenderTarget.cpp
```diff
@@ -47,17 +48,19 @@ WebGPURenderTarget::WebGPURenderTarget(const uint32_t width, const uint32_t heig
 // Default constructor for the default render target
 WebGPURenderTarget::WebGPURenderTarget()
     : HwRenderTarget{ 0, 0 },
+      mTargetFlags(TargetBufferFlags::NONE),
```
AndyHovingh: same as above, curly braces `mTargetFlags{ TargetBufferFlags::NONE },`

filament/backend/src/webgpu/WebGPURenderTarget.cpp
```diff
@@ -81,6 +85,10 @@ void WebGPURenderTarget::setUpRenderPassAttachments(wgpu::RenderPassDescriptor&
     mColorAttachmentDescriptors.clear();
     mHasDepthStencilAttachment = false;
 
+    bool hasDepth = any(mTargetFlags & TargetBufferFlags::DEPTH);
```
AndyHovingh: `hasDepth` and `hasStencil` can both be `const`

filament/backend/src/webgpu/WebGPURenderTarget.cpp
```diff
@@ -81,6 +85,10 @@ void WebGPURenderTarget::setUpRenderPassAttachments(wgpu::RenderPassDescriptor&
...
+    bool hasStencil = any(mTargetFlags & TargetBufferFlags::STENCIL);
+
+    // Color attachments
     if (mDefaultRenderTarget) {
         assert_invariant(defaultColorTextureView);
         mColorAttachmentDescriptors.push_back({ .view = defaultColorTextureView,
```
AndyHovingh: nit: totally up to your preference, and I know the formatting according to .clang-format has the first attribute on the same line like this, but I find it less readable than just moving it to the next line. Again, up to you.

filament/backend/src/webgpu/WebGPURenderTarget.cpp
```diff
@@ -90,66 +98,81 @@ void WebGPURenderTarget::setUpRenderPassAttachments(wgpu::RenderPassDescriptor&
...
-                << "WebGPU CANNOT support separate texture views for depth + stencil. depth + "
-                   "stencil needs to be in one texture view";
-
-        const bool hasStencil =
-                customStencilTextureView ||
-                (customDepthFormat == wgpu::TextureFormat::Depth24PlusStencil8 ||
-                        customDepthFormat == wgpu::TextureFormat::Depth32FloatStencil8);
+    // Depth/Stencil attachment
+    if (mDefaultRenderTarget) {
+        if (hasDepth || hasStencil) {
```
AndyHovingh: could this `if (hasDepth || hasStencil) {` be moved outside the whole depth/stencil attachment logic here, e.g, `if (hasDepth || hasStencil) { ... if (mDefaultRenderTarget) { ...`?

filament/backend/src/webgpu/WebGPURenderTarget.cpp
```diff
@@ -176,9 +199,6 @@ void WebGPURenderTarget::setUpRenderPassAttachments(wgpu::RenderPassDescriptor&
             mHasDepthStencilAttachment = true;
```
AndyHovingh: could we set this forward the top of the function, something like `mHasDepthStencilAttachment = hasDepth || hasStencil;`?

filament/backend/src/webgpu/WebGPURenderTarget.cpp
```diff
@@ -90,66 +98,81 @@ void WebGPURenderTarget::setUpRenderPassAttachments(wgpu::RenderPassDescriptor&
...
-        FILAMENT_CHECK_POSTCONDITION(!(customDepthTextureView && customStencilTextureView))
-                << "WebGPU CANNOT support separate texture views for depth + stencil. depth + "
-                   "stencil needs to be in one texture view";
-
-        const bool hasStencil =
-                customStencilTextureView ||
-                (customDepthFormat == wgpu::TextureFormat::Depth24PlusStencil8 ||
-                        customDepthFormat == wgpu::TextureFormat::Depth32FloatStencil8);
+    // Depth/Stencil attachment
+    if (mDefaultRenderTarget) {
```
AndyHovingh: Could this logic be made more concise and less redundant? It looks at a glance the same logic for the default target as well as the custom scenario; only difference seems to be what is set as the `view` (+ the custom check for textures in the cutom case), right?

filament/backend/src/webgpu/WebGPUPipelineCreation.cpp
```diff
@@ -169,39 +169,48 @@ wgpu::RenderPipeline createWebGPURenderPipeline(wgpu::Device const& device,
         wgpu::PipelineLayout const& layout, RasterState const& rasterState,
         StencilState const& stencilState, PolygonOffset const& polygonOffset,
         const PrimitiveType primitiveType, std::vector<wgpu::TextureFormat> const& colorFormats,
-        const wgpu::TextureFormat depthFormat, const uint8_t samplesCount) {
+        const wgpu::TextureFormat depthStencilFormat, const uint8_t samplesCount, bool filamentRequestedStencil) {
```
AndyHovingh: again, I think the `filament` prefix is not needed. I could also make the parameter `const` here. Also, I believe we need to also get a flag for depth too.

filament/backend/src/webgpu/WebGPUPipelineCreation.cpp
```diff
@@ -169,39 +169,48 @@ wgpu::RenderPipeline createWebGPURenderPipeline(wgpu::Device const& device,
...
     assert_invariant(program.vertexShaderModule);
     wgpu::DepthStencilState depthStencilState{};
-    if (depthFormat != wgpu::TextureFormat::Undefined) {
-        depthStencilState.format = depthFormat;
+
+      if (depthStencilFormat != wgpu::TextureFormat::Undefined) {
```
AndyHovingh: shouldn't this be an `if` based on bool flag passed to this function? The render pipeline could potentially not have a depth bit (defensive at least).

filament/backend/src/webgpu/WebGPUPipelineCreation.cpp
```diff
@@ -169,39 +169,48 @@ wgpu::RenderPipeline createWebGPURenderPipeline(wgpu::Device const& device,
         wgpu::PipelineLayout const& layout, RasterState const& rasterState,
         StencilState const& stencilState, PolygonOffset const& polygonOffset,
         const PrimitiveType primitiveType, std::vector<wgpu::TextureFormat> const& colorFormats,
-        const wgpu::TextureFormat depthFormat, const uint8_t samplesCount) {
+        const wgpu::TextureFormat depthStencilFormat, const uint8_t samplesCount, bool filamentRequestedStencil) {
     assert_invariant(program.vertexShaderModule);
     wgpu::DepthStencilState depthStencilState{};
-    if (depthFormat != wgpu::TextureFormat::Undefined) {
-        depthStencilState.format = depthFormat;
+
```
AndyHovingh, line 176: we could validate the `depthStencil` format against boolean flags (one for depth, one for stencil) toward the top of this function right?

filament/backend/src/webgpu/WebGPUPipelineCreation.cpp
```diff
@@ -169,39 +169,48 @@ wgpu::RenderPipeline createWebGPURenderPipeline(wgpu::Device const& device,
...
+      if (depthStencilFormat != wgpu::TextureFormat::Undefined) {
+        depthStencilState.format = depthStencilFormat;
```
AndyHovingh, line 181: shouldn't we _always_ set the `format` attribute to this when we initially instantiate the `depthStencilState` above?

filament/backend/src/webgpu/WebGPUPipelineCreation.cpp
```diff
@@ -169,39 +169,48 @@ wgpu::RenderPipeline createWebGPURenderPipeline(wgpu::Device const& device,
...
+            depthStencilState.depthBias = static_cast<int32_t>(polygonOffset.constant);
+            depthStencilState.depthBiasSlopeScale = polygonOffset.slope;
+            depthStencilState.depthBiasClamp = 0.0f;
+        } else {
+            depthStencilState.depthBias = 0;
+            depthStencilState.depthBiasSlopeScale = 0.0f;
+            depthStencilState.depthBiasClamp = 0.0f;
+        }
+
+        if (hasStencilAspect(depthStencilFormat) && filamentRequestedStencil) {
```
AndyHovingh: I think if the flag is true, but the `depthStencilFormat` doesn't have stencil bits we panic.

filament/backend/src/webgpu/WebGPUPipelineCreation.cpp
```diff
@@ -245,7 +254,7 @@ wgpu::RenderPipeline createWebGPURenderPipeline(wgpu::Device const& device,
             .unclippedDepth = !rasterState.depthClamp &&
                               device.HasFeature(wgpu::FeatureName::DepthClipControl)
         },
-        .depthStencil = depthFormat != wgpu::TextureFormat::Undefined ? &depthStencilState: nullptr,
+        .depthStencil = depthStencilFormat != wgpu::TextureFormat::Undefined ? &depthStencilState: nullptr,
```
AndyHovingh: I think the condition should be based on the 2 flags, something like `= (requestedDepth || requestedStencil) ? &depthStencilState : nullptr;`?

The primary objective of this change is to use Filament's flags, not inferencing based on formats or provided textures, etc.

filament/backend/src/webgpu/WebGPURenderTarget.cpp
```diff
@@ -73,6 +76,7 @@ wgpu::StoreOp WebGPURenderTarget::getStoreOperation(RenderPassParams const& para
 void WebGPURenderTarget::setUpRenderPassAttachments(wgpu::RenderPassDescriptor& outDescriptor,
         RenderPassParams const& params, wgpu::TextureView const& defaultColorTextureView,
         wgpu::TextureView const& defaultDepthStencilTextureView,
+        wgpu::TextureFormat const& defaultDepthStencilFormat,
```
AndyHovingh: are `defaultDepthStencilFormat`, `customDepthFormat`, and `customStencilFormat` still needed?

filament/backend/src/webgpu/WebGPUSwapChain.h
jneljneljnel: removed with a check that became redundant.

filament/backend/src/webgpu/WebGPUDriver.cpp
```diff
@@ -879,13 +915,10 @@ void WebGPUDriver::beginRenderPass(Handle<HwRenderTarget> renderTargetHandle,
...
             defaultDepthStencilView,
-            defaultDepthStencilFormat,
             customColorViews.data(),
             customColorViewCount,
-            customDepthView,
-            customStencilView,
-            customDepthFormat,
-            customStencilFormat);
+            customDepthStencilView,
+            customDepthStencilView);
```
jc3265, line 940: `customDepthStencilView` on 920-921 should prob be `customStencilTextureView`

jneljneljnel: These are combined now so we only use depthstencil for default and custom

filament/backend/src/webgpu/WebGPURenderTarget.h
```diff
@@ -82,8 +79,8 @@ class WebGPURenderTarget : public HwRenderTarget {
     Attachment mStencilAttachment{};
 
     // Cached descriptors for the render pass
-    std::vector<wgpu::RenderPassColorAttachment> mColorAttachmentDescriptors;
-    wgpu::RenderPassDepthStencilAttachment mDepthStencilAttachmentDescriptor{};
+    std::vector<wgpu::RenderPassColorAttachment> mColorAttachmentDesc;
```
AndyHovingh, line 82: nit: if we want shorter names, how about `mColorAttachments` and `mDepthStencilAttachment`?

filament/backend/src/webgpu/WebGPURenderTarget.cpp
```diff
@@ -133,68 +110,62 @@ void WebGPURenderTarget::setUpRenderPassAttachments(wgpu::RenderPassDescriptor&
                         .a = params.clearColor.a } });
             }
         }
+    }
+    outDescriptor.colorAttachmentCount = mColorAttachmentDesc.size();
+    outDescriptor.colorAttachments = mColorAttachmentDesc.data();
+
+    // Depth/Stencil Attachments
+    if (mHasDepthStencilAttachment) {
+        wgpu::TextureView depthStencilViewToUse;
```
AndyHovingh: we should always initialize things when we can, otherwise we can get undefined behavior across different compilers. We don't necessarily initialize it below before we invoke it with `if (depthStencilViewToUse)`. Thus, a safer thing here would be something like `wgpu::TextureView depthStencilViewToUse{nullptr};`

filament/backend/src/webgpu/WebGPUPipelineCreation.cpp
```diff
@@ -169,39 +169,59 @@ wgpu::RenderPipeline createWebGPURenderPipeline(wgpu::Device const& device,
         wgpu::PipelineLayout const& layout, RasterState const& rasterState,
         StencilState const& stencilState, PolygonOffset const& polygonOffset,
         const PrimitiveType primitiveType, std::vector<wgpu::TextureFormat> const& colorFormats,
-        const wgpu::TextureFormat depthFormat, const uint8_t samplesCount) {
+        wgpu::TextureFormat const& depthStencilFormat, const uint8_t samplesCount, bool requestedDepth, bool requestedStencil) {
```
AndyHovingh: nit: `const bool requestedDepth, const bool requestedStencil` these variables can be `const` in the implementation.

filament/backend/src/webgpu/WebGPUSwapChain.cpp
```diff
@@ -231,6 +229,16 @@ void initConfig(wgpu::SurfaceConfiguration& config, wgpu::Device const& device,
     };
     if (needStencil) {
         descriptor.aspect = wgpu::TextureAspect::All;
+        descriptor.format = depthFormat;
+    } else {
+        descriptor.aspect = wgpu::TextureAspect::DepthOnly;
+        if (depthFormat == wgpu::TextureFormat::Depth32FloatStencil8) {
```
AndyHovingh, line 235: is this really happening that `needStencil` is `false` and we are getting a depth format with stencil bits? I wonder if we should be panicking in this case; that feels like a Filament bug.

filament/backend/src/webgpu/WebGPUSwapChain.cpp
```diff
@@ -220,7 +219,6 @@ void initConfig(wgpu::SurfaceConfiguration& config, wgpu::Device const& device,
         wgpu::TextureFormat depthFormat, bool needStencil) {
```
AndyHovingh: nit: I know this is not technically part of your change but `depthFormat` and `needStencil` can be `const` here.

filament/backend/src/webgpu/WebGPUDriver.cpp
jc3265: wait so we are passing `customDepthStencilView` twice to `setUpRenderPassAttachments` 
why not just pass it once?

filament/backend/src/webgpu/WebGPUDriver.cpp
```diff
@@ -841,36 +858,55 @@ void WebGPUDriver::beginRenderPass(Handle<HwRenderTarget> renderTargetHandle,
...
-                FILAMENT_CHECK_POSTCONDITION(stencilInfo.layer < renderTarget->getLayerCount())
-                        << "Stencil attachment requests layer " << stencilInfo.layer
-                        << " but render target has only " << renderTarget->getLayerCount()
-                        << " layers.";
-                const uint8_t stencilMipLevel = stencilInfo.level;
-                const uint32_t stencilArrayLayer = stencilInfo.layer;
-                customStencilView =
-                        stencilTexture->getOrMakeTextureView(stencilMipLevel, stencilArrayLayer);
-                customStencilFormat = stencilTexture->getViewFormat();
+        if (depthStencilSourceHandle) {
```
AndyHovingh, line 898: from what I can tell, we are never actually checking that if we have both depth and stencil info that the underlying texture for each is the same.

Could we do something like?:
```cpp
if (depthInfo.handle && stencilInfo.handle) {
  auto depthTexture = handleCast<WebGPUTexture>(depthInfo.handle).getTexture();
  auto stencilTexture = handleCast<WebGPUTexture>(stencilInfo.handle)getTexture();
  FILAMENT_CHECK_POSTCONDITION(depthTexture.Get() == stencilTexture.Get()) ...
}
```

AndyHovingh: Also, shouldn't we use the flags to get the handle(s) anyway?

If the flags indicate depth only, maybe we just get the depth handle (asserting it is there) amd just ignore the stencil handle entirely (we don't care if it is there or not).
If the flags indicate stencil only, the same thing as above, just swapped.
If the flags indicate both depth and stencil then we use the depth handle and verify if the stencil handle is also provided that the the underlying textures are the same.

filament/backend/src/webgpu/WebGPUDriver.cpp
```diff
@@ -920,9 +952,27 @@ void WebGPUDriver::makeCurrent(Handle<HwSwapChain> drawSch, Handle<HwSwapChain>
...
+        newTargetFlags |= filament::backend::TargetBufferFlags::DEPTH;
+
+        if (depthFormat == wgpu::TextureFormat::Depth24PlusStencil8 ||
+                depthFormat == wgpu::TextureFormat::Depth32FloatStencil8) {
+            newTargetFlags |= filament::backend::TargetBufferFlags::STENCIL;
+        }
+    }
+    mDefaultRenderTarget->setTargetFlags(newTargetFlags);
+
+    wgpu::CommandEncoderDescriptor commandEncoderDescriptor = { .label = "command_encoder" };
```
AndyHovingh: nit: maybe label it `"frame_command_encoder""` to differentiate it from another encoder, e.g. for mipmap generation or something. Otherwise, the label "command_encoder" doesn't really help us in debugging in any way; we already know it is a command encoder of some kind.

filament/backend/src/webgpu/WebGPUDriver.cpp
```diff
@@ -1057,14 +1107,17 @@ void WebGPUDriver::bindPipeline(PipelineState const& pipelineState) {
     assert_invariant(vertexBufferInfo);
 
     std::vector<wgpu::TextureFormat> pipelineColorFormats;
-    wgpu::TextureFormat pipelineDepthFormat = wgpu::TextureFormat::Undefined;
+    wgpu::TextureFormat pipelineDepthStencilFormat = wgpu::TextureFormat::Undefined;
     uint8_t pipelineSamples = 1;
+    bool requestedDepth = any(mCurrentRenderTarget->getTargetFlags() & TargetBufferFlags::DEPTH);
```
AndyHovingh: nit: requestedDepth and requestedStencil can be `const` right?

filament/backend/src/webgpu/WebGPUDriver.cpp
```diff
@@ -1057,14 +1107,17 @@ void WebGPUDriver::bindPipeline(PipelineState const& pipelineState) {
...
+    pipelineSamples = mCurrentRenderTarget->getSamples();
 
     if (mCurrentRenderTarget->isDefaultRenderTarget()) {
         pipelineColorFormats.push_back(mSwapChain->getColorFormat());
-        pipelineDepthFormat = mSwapChain->getDepthFormat();
-        pipelineSamples =
-                mCurrentRenderTarget->getSamples();// Default RT should have samples (usually 1)
+        pipelineDepthStencilFormat = mSwapChain->getDepthFormat();
+        pipelineSamples = mCurrentRenderTarget->getSamples();
+        requestedStencil = any(mCurrentRenderTarget->getTargetFlags() & TargetBufferFlags::STENCIL);
```
AndyHovingh: duplicate line

filament/backend/src/webgpu/WebGPUDriver.cpp
```diff
@@ -1078,21 +1131,21 @@ void WebGPUDriver::bindPipeline(PipelineState const& pipelineState) {
...
+        }
+
+        if (dsHandle) {
+            const auto dsTexture = handleCast<WebGPUTexture>(dsHandle);
+            if (dsTexture) {
+                pipelineDepthStencilFormat = dsTexture->getTexture().GetFormat();
             }
         }
         pipelineSamples = mCurrentRenderTarget->getSamples();
+        requestedStencil = any(mCurrentRenderTarget->getTargetFlags() & TargetBufferFlags::STENCIL);
```
AndyHovingh: duplicate line

filament/backend/src/webgpu/WebGPUDriver.cpp
AndyHovingh: agreed. @jneljneljnel already made the change to send one view 👍

filament/backend/src/webgpu/WebGPURenderTarget.cpp
```diff
@@ -137,71 +111,62 @@ void WebGPURenderTarget::setUpRenderPassAttachments(wgpu::RenderPassDescriptor&
...
-                mDepthStencilAttachmentDescriptor.depthLoadOp =
-                        depthReadonly ? wgpu::LoadOp::Undefined :
-                        WebGPURenderTarget::getLoadOperation(params, TargetBufferFlags::DEPTH);
-                mDepthStencilAttachmentDescriptor.depthStoreOp =
-                        depthReadonly ? wgpu::StoreOp::Undefined :
-                        WebGPURenderTarget::getStoreOperation(params, TargetBufferFlags::DEPTH);
-                mDepthStencilAttachmentDescriptor.depthClearValue =
-                        static_cast<float>(params.clearDepth);
-                mDepthStencilAttachmentDescriptor.depthReadOnly = depthReadonly;
+                mDepthStencilAttachmentDesc.depthLoadOp =  depthReadOnly ? wgpu::LoadOp::Undefined :getLoadOperation(params, TargetBufferFlags::DEPTH);
```
jc3265, line 135: nit: 
seeing as the `bool depthReadOnly` is only used here, shouldnt we also just define it here?
theres no need to define it every time/for the other cases

filament/backend/src/webgpu/WebGPUDriver.cpp
```diff
@@ -823,17 +823,34 @@ void WebGPUDriver::beginRenderPass(Handle<HwRenderTarget> renderTargetHandle,
...
+                                                << (uint32_t)defaultDepthStencilFormat << " does not have a stencil aspect.";
+        }
+
+        if (any(renderTarget->getTargetFlags() & TargetBufferFlags::DEPTH) &&
+            !(defaultDepthStencilFormat == wgpu::TextureFormat::Depth16Unorm ||
+              defaultDepthStencilFormat == wgpu::TextureFormat::Depth32Float ||
+              defaultDepthStencilFormat == wgpu::TextureFormat::Depth24Plus ||
+              defaultDepthStencilFormat == wgpu::TextureFormat::Depth24PlusStencil8 ||
+              defaultDepthStencilFormat == wgpu::TextureFormat::Depth32FloatStencil8)) {
+            FILAMENT_CHECK_POSTCONDITION(false) << "Default render target requested depth, but swap chain's depth format "
```
jc3265, line 851: should this be `but swap chain's stencil format` ...

filament/backend/src/webgpu/WebGPUDriver.cpp
```diff
@@ -853,36 +870,63 @@ void WebGPUDriver::beginRenderPass(Handle<HwRenderTarget> renderTargetHandle,
                 }
             }
         }
-
         const auto& depthInfo = renderTarget->getDepthAttachmentInfo();
+        const auto& stencilInfo = renderTarget->getStencilAttachmentInfo();
+
+        Handle<HwTexture> depthStencilSourceHandle = {};
+        uint8_t dsMipLevel = 0;
+        uint32_t dsArrayLayer = 0;
```
jc3265: i thought we were walking away from "short"/not descriptive variable names, could be `depthStencilArrayLayer` or something other than `ds`?

filament/backend/src/webgpu/WebGPUDriver.cpp
```diff
@@ -932,9 +972,27 @@ void WebGPUDriver::makeCurrent(Handle<HwSwapChain> drawSch, Handle<HwSwapChain>
...
-    wgpu::CommandEncoderDescriptor commandEncoderDescriptor = {
-        .label = "command_encoder"
-    };
+
+    assert_invariant(mDefaultRenderTarget);
+
+    wgpu::TextureFormat depthFormat = mSwapChain->getDepthFormat();
+    TargetBufferFlags newTargetFlags = filament::backend::TargetBufferFlags::NONE;
+
+    newTargetFlags = filament::backend::TargetBufferFlags::NONE;
```
jc3265: this looks unnecessary, line before already has `TargetBufferFlags newTargetFlags = filament::backend::TargetBufferFlags::NONE;`

filament/backend/src/webgpu/WebGPUDriver.cpp
```diff
@@ -932,9 +972,27 @@ void WebGPUDriver::makeCurrent(Handle<HwSwapChain> drawSch, Handle<HwSwapChain>
...
+
+    //Assuming Color and Depth are always present.
+    newTargetFlags |= filament::backend::TargetBufferFlags::COLOR;
+    if (depthFormat != wgpu::TextureFormat::Undefined) {
+        newTargetFlags |= filament::backend::TargetBufferFlags::DEPTH;
+
+        if (depthFormat == wgpu::TextureFormat::Depth24PlusStencil8 ||
+                depthFormat == wgpu::TextureFormat::Depth32FloatStencil8) {
```
jc3265, line 987: sometimes when we do these depth checks we also check `x == wgpu::TextureFormat::Stencil8`
should we also do it here?

filament/backend/src/webgpu/WebGPUDriver.cpp
```diff
@@ -1090,21 +1149,19 @@ void WebGPUDriver::bindPipeline(PipelineState const& pipelineState) {
...
-                    pipelineDepthFormat = stencilTexture->getTexture().GetFormat();
-                }
+            dsHandle = depthInfo.handle;
+        } else if (stencilInfo.handle) {
+             dsHandle = stencilInfo.handle;
+        }
+
+        if (dsHandle) {
+            const auto dsTexture = handleCast<WebGPUTexture>(dsHandle);
+            if (dsTexture) {
```
jc3265, line 1159: just the same comment about using `ds` for `dsHandle` and dsTexture`
we already have variables like `depthStencilSourceHandle` so it would just be consistency

filament/backend/src/webgpu/WebGPUPipelineCreation.cpp
```diff
@@ -169,39 +169,59 @@ wgpu::RenderPipeline createWebGPURenderPipeline(wgpu::Device const& device,
...
+        FILAMENT_CHECK_PRECONDITION(depthStencilFormat != wgpu::TextureFormat::Undefined)
+                << "Depth or Stencil requested for pipeline, but depthStencilFormat is "
+                   "wgpu::TextureFormat::Undefined.";
+        depthStencilState.format = depthStencilFormat;
+
+        if (requestedDepth) {
+            depthStencilState.depthWriteEnabled = rasterState.depthWrite;
+            depthStencilState.depthCompare = toWebGPU(rasterState.depthFunc);
+
+            if (polygonOffset.constant != 0.0f || polygonOffset.slope != 0.0f) {
```
jc3265: i dont get why we need this check. 
if `polygonOffset.constant` is NOT 0 then we set `depthStencilState.depthBias` to that. However, if `polygonOffset.constant == 0` we still set `depthStencilState.depthBias` to 0. 
why not remove the if and just always set 
`depthStencilState.depthBias = polygonOffset.constant` (same for `.slope`)

filament/backend/src/webgpu/WebGPUPipelineCreation.cpp
```diff
@@ -169,39 +169,59 @@ wgpu::RenderPipeline createWebGPURenderPipeline(wgpu::Device const& device,
...
+                depthStencilState.depthBiasSlopeScale = polygonOffset.slope;
+                depthStencilState.depthBiasClamp = 0.0f;
+            } else {
+                depthStencilState.depthBias = 0;
+                depthStencilState.depthBiasSlopeScale = 0.0f;
+                depthStencilState.depthBiasClamp = 0.0f;
+            }
+        } else {
+            depthStencilState.depthWriteEnabled = false;
+            depthStencilState.depthCompare = wgpu::CompareFunction::Always;
```
jc3265: this is for the case where `requestedDepth == false` if thats the case, then why is this `Always` and not `Undefined`?

filament/backend/src/webgpu/WebGPUPipelineCreation.cpp
```diff
@@ -169,39 +169,59 @@ wgpu::RenderPipeline createWebGPURenderPipeline(wgpu::Device const& device,
...
+            depthStencilState.stencilBack = {
+                .compare = toWebGPU(stencilState.back.stencilFunc),
+                .failOp = toWebGPU(stencilState.back.stencilOpStencilFail),
+                .depthFailOp = toWebGPU(stencilState.back.stencilOpDepthFail),
+                .passOp = toWebGPU(stencilState.back.stencilOpDepthStencilPass),
+            };
+            depthStencilState.stencilReadMask = stencilState.front.readMask;
+            depthStencilState.stencilWriteMask = stencilState.stencilWrite ? stencilState.front.writeMask : 0u;
+        } else {
             depthStencilState.stencilFront.compare = wgpu::CompareFunction::Always;
```
jc3265: Ahhh, i guess this is why for depth we also set this to `Always` and not `undefined`. However, should they both be `undefined`?

filament/backend/src/webgpu/WebGPURenderTarget.cpp
```diff
@@ -137,71 +111,62 @@ void WebGPURenderTarget::setUpRenderPassAttachments(wgpu::RenderPassDescriptor&
...
-    outDescriptor.colorAttachmentCount = mColorAttachmentDescriptors.size();
-    outDescriptor.colorAttachments = mColorAttachmentDescriptors.data();
-    outDescriptor.depthStencilAttachment =
-            mHasDepthStencilAttachment ? &mDepthStencilAttachmentDescriptor : nullptr;
+    } else {
+        outDescriptor.depthStencilAttachment = nullptr;
+        mDepthStencilAttachmentDesc = {};
+    }
 
     // descriptor.sampleCount was removed from the core spec. If your webgpu.h still has it,
```
jc3265, line 167: this is prob outside of this PR but this seems to be a leftover comment that doesnt apply. Our Dawn version also doesnt have those fields

filament/backend/src/webgpu/WebGPUDriver.cpp
jneljneljnel: We shouldn't have only a stencil on a default, I think we would want to panic if that were happening,

filament/backend/src/webgpu/WebGPUPipelineCreation.cpp
jneljneljnel: Yeah this seems better, thanks

filament/backend/src/webgpu/WebGPUDriver.cpp
jneljneljnel: I think this is right, as we're checking for a depth format, stencil is the above check on 837

filament/backend/src/webgpu/WebGPURenderTarget.cpp
jneljneljnel: we opted to define it at the top with hasDepth and hasStencil

filament/backend/src/webgpu/WebGPURenderTarget.cpp
jneljneljnel: ah nice, we can take that out in a subsequent fix

PR #8865, opened 2025-06-13
webgpu: Wait for work of first frame to be done before any presenting

bridgewaterrobbie:


filament/backend/src/webgpu/WebGPUDriver.cpp
```diff
@@ -53,6 +53,8 @@
 #include <sstream>
 #include <utility>
 
+using namespace std::chrono_literals;
```
jc3265, line 56: so we have this but we are only using it in some of the cases.
We could either just not use it or use it for every case?

filament/backend/src/webgpu/WebGPUDriver.cpp
```diff
@@ -923,6 +925,17 @@ void WebGPUDriver::commit(Handle<HwSwapChain> sch) {
...
     mQueue.Submit(1, &mCommandBuffer);
+
+    static bool firstRender = true;
+    // For the first frame rendered, we need to make sure the work is done before presenting or we
+    // get a purple flash
+    if (firstRender) {
+        auto f = mQueue.OnSubmittedWorkDone(wgpu::CallbackMode::WaitAnyOnly,
+                [=](wgpu::QueueWorkDoneStatus) {});
+        const wgpu::Instance instance = mDevice.GetAdapter().GetInstance();
+        instance.WaitAny(f, std::chrono::duration_cast<std::chrono::nanoseconds>(1s).count());
```
jc3265: (from above comment) 
we already have `std::chrono::nanoseconds` why not `std::chrono::seconds` as well for the `1s`?
```suggestion
        instance.WaitAny(f, std::chrono::duration_cast<std::chrono::nanoseconds>(std::chrono::seconds(1)).count());
```
and then remove the `using namespace std::chrono_literals;` ?

filament/backend/src/webgpu/WebGPUDriver.cpp
bridgewaterrobbie: Somewhat a matter of opinion. IMO this is far more readable and used it for my initial code

filament/backend/src/webgpu/WebGPUDriver.cpp
bridgewaterrobbie: Readability is the only reason- std::chrono::duration_cast<std::chrono::nanoseconds>(1s).count to me parses naturally as "the number of nanoseconds in 1 second". I could also just put 1000000000 instead.

AndyHovingh: I suggest we log a warning if this `WaitAny` fails

filament/backend/src/webgpu/WebGPUDriver.cpp
```diff
@@ -923,6 +925,17 @@ void WebGPUDriver::commit(Handle<HwSwapChain> sch) {
...
     mCommandEncoder = nullptr;
     mQueue.Submit(1, &mCommandBuffer);
+
+    static bool firstRender = true;
+    // For the first frame rendered, we need to make sure the work is done before presenting or we
+    // get a purple flash
+    if (firstRender) {
+        auto f = mQueue.OnSubmittedWorkDone(wgpu::CallbackMode::WaitAnyOnly,
+                [=](wgpu::QueueWorkDoneStatus) {});
+        const wgpu::Instance instance = mDevice.GetAdapter().GetInstance();
```
AndyHovingh: nit: I would suggest `mPlatform.getInstance()` or `mAdapter.GetInstance()`

PR #8870, opened 2025-06-16
Fix validation error on AntiqueCamera.glb

kpiascik:
Inline header defined method to remove compiler warning


AndyHovingh:
When I run this branch (on mac) I get the magenta screen with the errors:
```
WebGPU device error: VALIDATION Texture binding (group:0, binding:20) is TextureSampleType::UnfilterableFloat but used statically with a sampler (group:0, binding:21) that's SamplerBindingType::Filtering
 - While validating fragment stage ([ShaderModule "base_lit_opaque fragment shader"], entryPoint: "main").
 - While validating fragment state.
 - While calling [Device "graphics_device"].CreateRenderPipeline([RenderPipelineDescriptor ""base_lit_opaque pipeline""]).

WebGPU device error: VALIDATION Texture binding (group:2, binding:2) is TextureSampleType::UnfilterableFloat but used statically with a sampler (group:2, binding:3) that's SamplerBindingType::Filtering
 - While validating fragment stage ([ShaderModule "Skybox fragment shader"], entryPoint: "main").
 - While validating fragment state.
 - While calling [Device "graphics_device"].CreateRenderPipeline([RenderPipelineDescriptor ""Skybox pipeline""]).
```

When I run the same glb from the main branch it works without errors (albeit with the slow load of mipmaps and initial input slowness we are seeing across the board).

filament/backend/src/webgpu/WebGPUDescriptorSetLayout.cpp
```diff
@@ -151,8 +151,8 @@ WebGPUDescriptorSetLayout::WebGPUDescriptorSetLayout(DescriptorSetLayout const&
         if (isDepthDescriptor(fEntry.type)) {
             wEntry.texture.sampleType = wgpu::TextureSampleType::Depth;
         } else if (isFloatDescriptor(fEntry.type)) {
-            // TODO: Set once we have the filtering values
-            wEntry.texture.sampleType = wgpu::TextureSampleType::Float;
+            // TODO: Set correctly once we have the filtering values
+            wEntry.texture.sampleType = wgpu::TextureSampleType::UnfilterableFloat;
```
AndyHovingh: I get runtime errors with this + magenta screen. If I revert it back to `Float` the errors go away.

bridgewaterrobbie: Generally speaking, us setting Float or Unfilterable Float is unavoidably a hack until the Filament Engine gives us that info properly, but the existing code is the hack with the broadest compatibility. If we pinpointed this as what it takes for AntiqueCamera that is good because we know the eventual solution, but we can't merge this as-is

filament/backend/src/webgpu/WebGPUStrings.h
```diff
@@ -101,7 +101,7 @@ template<typename WebGPUPrintable>
     return out.str();
 }
 
-[[nodiscard]] std::string adapterInfoToString(wgpu::AdapterInfo const& info) {
+[[nodiscard]] inline std::string adapterInfoToString(wgpu::AdapterInfo const& info) {
```
kpiascik, line 104: IT looks like this is the only change that has any impact.  With or without my patch there is no difference in output from the AntiqueCamera.glb with Powei's change

filament/backend/src/webgpu/WebGPUDescriptorSetLayout.cpp
kpiascik: I've removed this from my commit.

kpiascik:
@AndyHovingh the validation errors will be resolved by @poweifeng's change to implement filterable.  This is just a small update to other code.  It may still end up being tweaked later.

kpiascik:
abandoning

PR #8871, opened 2025-06-16
DRAFT for webgpu Add unfilterable [DO NOT REVIEW]

poweifeng:


filament/backend/src/webgpu/WebGPUDescriptorSetLayout.cpp
```diff
@@ -15,6 +15,7 @@
  */
 
 #include "WebGPUDescriptorSetLayout.h"
+#include "WebGPUConstants.h"
```
AndyHovingh, line 18: this appears to be unused

poweifeng: indeed, but please don't review this.  It's only meant for testing

poweifeng:
Real PR is #8873 

This has how to disable msaa and ssao for gltf_viewer by default.

PR #8872, opened 2025-06-16
Fix staging buffer alignment

anishmgoyal:
Note: we were not aligning texture copies properly, so this broke certain use cases on the Pixel 4.

### DIFF
--- a/filament/backend/src/vulkan/VulkanStagePool.cpp
+++ b/filament/backend/src/vulkan/VulkanStagePool.cpp
@@ -62,7 +62,7 @@ VulkanStagePool::VulkanStagePool(VmaAllocator allocator, fvkmemory::ResourceMana
       mDeviceLimits(deviceLimits) {}
 
 fvkmemory::resource_ptr<VulkanStage::Segment> VulkanStagePool::acquireStage(uint32_t numBytes,
-        std::optional<uint32_t> alignment) {
+        uint32_t alignment) {
     // Apply alignment to the byte count to ensure that, when we later flush
     // data written by the host, we only flush the atoms that we modified, and
     // no adjacent atoms.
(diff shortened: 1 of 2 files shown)


filament/backend/src/vulkan/VulkanStagePool.cpp
```diff
@@ -64,17 +72,39 @@ fvkmemory::resource_ptr<VulkanStage::Segment> VulkanStagePool::acquireStage(uint
...
-    if (iter != mStages.end()) {
-        pStage = iter->second;
-        mStages.erase(iter);
-    } else {
+    // If we have potential buffer candidates, loop through and try to find one
+    // that works with our alignment. Also, set the offset for the buffer.
+    // If we have no buffer, create a new one.
+    VulkanStage* pStage = nullptr;
+    uint32_t segmentOffset = 0;
+    while (iter != mStages.end()) {
```
anishmgoyal, line 80: note - there should only be a handful of blocks to search through, i don't anticipate this being a very long search. also, since alignment is only upto 16 bytes for textures / inputs, we're searching for buffers that are within 16 bytes of the required space. that means that most unsuitable buffers will already have been ruled out on line 73.

filament/backend/src/vulkan/VulkanStagePool.h
```diff
@@ -144,7 +148,10 @@ class VulkanStagePool {
     // buffers so that we have less objects around that we have to keep track
     // of.
     // This function is NOT thread-safe.
-    fvkmemory::resource_ptr<VulkanStage::Segment> acquireStage(uint32_t numBytes);
+    // numBytes - the number of bytes required by this segment.
+    // alignment - (optional) the multiple that
+    fvkmemory::resource_ptr<VulkanStage::Segment> acquireStage(uint32_t numBytes,
+            std::optional<uint32_t> alignment = std::nullopt);
```
poweifeng: Maybe we don't need the optional and just make `alignment = 0`. I `alignValue()` handles that case by not doing any alignment.

filament/backend/src/vulkan/VulkanStagePool.cpp
```diff
@@ -64,17 +72,39 @@ fvkmemory::resource_ptr<VulkanStage::Segment> VulkanStagePool::acquireStage(uint
     // equal to the requested size.
     auto iter = mStages.lower_bound(numBytes);
```
poweifeng, line 73: I haven't thought too deeply about this, but I wonder if you could just do `lower_bound(numBytes + alignmentOffsetVal)` and not have to do the loop below.

anishmgoyal: the problem is that the alignment offset varies depending on the current offset of each buffer. e.g. a buffer at offset 3 might need alignment, whereas one at 4 might not.

filament/backend/src/vulkan/VulkanStagePool.h
anishmgoyal: done!

phoenixxxx (approved):

PR #8873, opened 2025-06-16
Add unfilterable to Material sampler definition

poweifeng:
`unfiterable : true` indicates that the sampling of the texture will not apply filtering.

This is mainly to satisfy the webgpu requirement for bindGroupLayouts.

 - Documentation has been updated for both `unfilterable` and `multisample`.
 - Internal materials have been updated where necessary (depth samplers have to be marked as `unfilterable`.)
 - webgpu code has been changed appropriately.

FIXES=420745987

### DIFF
--- a/filament/backend/src/webgpu/WebGPUPipelineCreation.cpp
+++ b/filament/backend/src/webgpu/WebGPUPipelineCreation.cpp
@@ -249,7 +249,7 @@ wgpu::RenderPipeline createWebGPURenderPipeline(wgpu::Device const& device,
         .multisample = {
             .count = samplesCount,
             .mask = 0xFFFFFFFF,
-            .alphaToCoverageEnabled = rasterState.alphaToCoverage
+            .alphaToCoverageEnabled = (samplesCount > 1) && rasterState.alphaToCoverage
         },
         .fragment = nullptr // will add below if fragment module is included
     };


kpiascik:
The change lgtm, but I do want to go on record that I've experienced pain with default argument values defined in headers in the past and am not a fan.  This is especially a problem if you have several of the same type of variable adjacent to each other.

docs/Materials.md.html
```diff
@@ -1027,8 +1027,11 @@
 [Table [materialParamsTypes]: Material parameter types]
 
 Samplers
-:     Sampler types can also specify a `format` which can be either `int` or `float` (defaults to
-      `float`).
+:     Sampler types can have the following fields:
+       - `format` : which can be either `int` or `float` (defaults to `float`).
+       - `multisample` : a boolean to indicate whether the sampler is meant for multisampling (defaults to `false`)
+       - `unfilterable` : a boolean to indicate whether the sampling is not filtered (defaults to `false`)
```
bejado, line 1033: I wonder if it would be clearer to phrase this in the positive—name this `filterable` and make the default `true`. I guess the advantage of using the name `unfilterable` is that it directly matches `wgpu::TextureSampleType::UnfilterableFloat`.

I think we should also specify what constraints this puts on the client. I.e., if a texture is not filterable, then you're only allowed to use it in calls to `texelFetch`. Is this something we can enforce at material compile time?

poweifeng: > I wonder if it would be clearer to phrase this in the positive—name this filterable and make the default true. I guess the advantage of using the name unfilterable is that it directly matches wgpu::TextureSampleType::UnfilterableFloat.

That's the reason why I went with unfilterable.  But also, for things like `multisample` and `mutable` (the ongoing mutable constant bit).  The default is false and enabling the bit implies a case different from the norm.

> I think we should also specify what constraints this puts on the client. I.e., if a texture is not filterable, then you're only allowed to use it in calls to texelFetch. Is this something we can enforce at material compile time?

So this is something I'm also not super clear on.  In webGPU, you could bind a depth or unfilterable float sample type to a bind point meant for a depth texture. As of now, we have no way to specifying a sampler2D as a depth format. So adding this `unfilterable` bit is the only way to tell the backend that we will bind a depth texture here. 

But does that mean a depth texture should only be used with `texelFetch`? Is there a case where we do want filtering with a depth texture?  If so, then `unfilterable` is not enough to convey that, but we'd need something like a `depth` format (in addition to `float` and `int`).

poweifeng:
> The change lgtm, but I do want to go on record that I've experienced pain with default argument values defined in headers in the past and am not a fan. This is especially a problem if you have several of the same type of variable adjacent to each other.

Good point and I'll update accordingly.

docs/Materials.md.html
bejado: Yeah, these are good areas for investigation. VSM shadows come to mind when I think of filtering a depth texture. I'm fine with this PR as-is to unblock WebGPU.

bejado (approved):

bridgewaterrobbie (approved):

jc3265 (approved):

PR #8877, opened 2025-06-16
webgpu: Misc Fixes

kpiascik:


filament/backend/src/webgpu/WebGPUDriver.cpp
```diff
@@ -612,11 +609,24 @@ size_t WebGPUDriver::getMaxUniformBufferSize() {
...
+    switch (target) {
+        case SamplerType::SAMPLER_2D:
+        case SamplerType::SAMPLER_2D_ARRAY:
+        case SamplerType::SAMPLER_EXTERNAL:
+            result = mDevice.maxTextureDimension2D;
+            break;
+        case SamplerType::SAMPLER_3D:
+        case SamplerType::SAMPLER_CUBEMAP:
+        case SamplerType::SAMPLER_CUBEMAP_ARRAY:
+            result = mDevice.maxTextureDimension3D;
```
bridgewaterrobbie: I am pretty sure that in this sense, only the sampler3D should query the maxTextureDimension3D. Trying to f ind documented proof to be certain

bridgewaterrobbie: "cube"
this.dimension must be "2d".

descriptor.arrayLayerCount must be 6.

this.width must equal this.height.

"cube-array"
this.dimension must be "2d".

descriptor.arrayLayerCount must be a multiple of 6.

this.width must equal this.height.

Below https://www.w3.org/TR/webgpu/#dom-gputextureviewdimension-cube


As well as 
maxTextureDimension2D	GPUSize32	maximum	8192
The maximum allowed value for the size.width and size.height of a texture created with dimension "2d".

maxTextureDimension3D	GPUSize32	maximum	2048
The maximum allowed value for the size.width, size.height and size.depthOrArrayLayers of a texture created with dimension "3d".

PR #8884, opened 2025-06-18
webgpu: implement fence

AndyHovingh:
Addresses lagginess, initial input unresponsiveness, and command stream using too much space issues.

### DIFF
--- a/filament/backend/src/webgpu/WebGPUFence.cpp
+++ b/filament/backend/src/webgpu/WebGPUFence.cpp
@@ -1,0 +1,50 @@
+/*
+ * Copyright (C) 2025 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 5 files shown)


jc3265 (approved):

filament/backend/src/webgpu/WebGPUFence.cpp
```diff
@@ -0,0 +1,50 @@
...
+            [](const wgpu::QueueWorkDoneStatus status,
+                    std::atomic<FenceStatus>* const fenceStatus) {
+                if (!fenceStatus) {
+                    return;
+                }
+                switch (status) {
+                    case wgpu::QueueWorkDoneStatus::Success:
+                        fenceStatus->store(FenceStatus::CONDITION_SATISFIED);
+                        break;
+                    case wgpu::QueueWorkDoneStatus::CallbackCancelled:
```
kpiascik, line 41: is `CallbackCancelled` really an error state?

AndyHovingh: Yes, I believe so. This is an unexpected scenario. Let's fail loudly until we know better.

PR #8885, opened 2025-06-18
webgpu: Add alphaToCoverage check

jc3265:
check for alpha to coverage when samples > 1

### DIFF
--- a/filament/backend/src/webgpu/WebGPUPipelineCreation.cpp
+++ b/filament/backend/src/webgpu/WebGPUPipelineCreation.cpp
@@ -249,7 +249,7 @@ wgpu::RenderPipeline createWebGPURenderPipeline(wgpu::Device const& device,
         .multisample = {
             .count = samplesCount,
             .mask = 0xFFFFFFFF,
-            .alphaToCoverageEnabled = rasterState.alphaToCoverage
+            .alphaToCoverageEnabled = (samplesCount > 1) && rasterState.alphaToCoverage
         },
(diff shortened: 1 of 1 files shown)


filament/backend/src/webgpu/WebGPUPipelineCreation.cpp
```diff
@@ -206,7 +206,8 @@ wgpu::RenderPipeline createWebGPURenderPipeline(wgpu::Device const& device,
             depthStencilState.stencilWriteMask = 0;
         }
     }
-
+    // Use the rasterState.alphaToCoverage only if there are more than one samples
+    const bool alphaToCoverage = (samplesCount > 1) ? rasterState.alphaToCoverage : false;
```
AndyHovingh: could this be simplified to `samplesCount > 1 && rasterState.alphaToCoverage`?

filament/backend/src/webgpu/WebGPUPipelineCreation.cpp
```diff
@@ -249,7 +250,7 @@ wgpu::RenderPipeline createWebGPURenderPipeline(wgpu::Device const& device,
         .multisample = {
             .count = samplesCount,
             .mask = 0xFFFFFFFF,
-            .alphaToCoverageEnabled = rasterState.alphaToCoverage
+            .alphaToCoverageEnabled = alphaToCoverage
```
kpiascik: we could also just have the calculation inline here instead of creating a variable\

PR #8894, opened 2025-06-24
webgpu: implement render pass based mipmap generation (supports more texture formats than compute pass based mipmap generation)

AndyHovingh:
BUGS = [427179627]

### DIFF
--- a/filament/backend/src/webgpu/WebGPURenderPassMipmapGenerator.h
+++ b/filament/backend/src/webgpu/WebGPURenderPassMipmapGenerator.h
@@ -1,0 +1,69 @@
+/*
+ * Copyright (C) 2025 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef TNT_FILAMENT_BACKEND_WEBGPURENDERPASSMIPMAPGENERATOR_H
+#define TNT_FILAMENT_BACKEND_WEBGPURENDERPASSMIPMAPGENERATOR_H
+
+#include <utils/Panic.h>
+
+#include <tsl/robin_map.h>
+#include <webgpu/webgpu_cpp.h>
+
+#include <cstdint>
+#include <string_view>
+
+namespace filament::backend {
+
+class WebGPURenderPassMipmapGenerator final {
+public:
+    struct FormatCompatibility final {
+        bool compatible{ false };
+        std::string_view reason;
+    };
+
+    [[nodiscard]] static FormatCompatibility getCompatibilityFor(wgpu::TextureFormat,
+            wgpu::TextureDimension, uint32_t sampleCount);
+
+    explicit WebGPURenderPassMipmapGenerator(wgpu::Device const&);
+
+    void generateMipmaps(wgpu::Queue const&, wgpu::Texture const&);
+
+private:
+    enum class ScalarSampleType : uint8_t {
+        F32,
+        I32,
+        U32,
+    };
+
+    [[nodiscard]] constexpr static ScalarSampleType getScalarSampleTypeFrom(
+            wgpu::TextureFormat format);
+
+    [[nodiscard]] wgpu::RenderPipeline& getOrCreatePipelineFor(wgpu::TextureFormat);
+
+    void generateMipmap(wgpu::CommandEncoder const&, wgpu::Texture const&,
+            wgpu::RenderPipeline const&, uint32_t layer, uint32_t mipLevel);
+
+    wgpu::Device const& mDevice;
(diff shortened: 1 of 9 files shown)


filament/backend/src/webgpu/WebGPURenderPassMipmapGenerator.h
```diff
@@ -0,0 +1,282 @@
...
+namespace filament::backend {
+
+class WebGPURenderPassMipmapGenerator final {
+public:
+    struct FormatCompatibility final {
+        bool compatible{ false };
+        std::string_view reason;
+    };
+
+    [[nodiscard]] constexpr static FormatCompatibility getCompatibilityFor(
```
kpiascik: why are you putting this implementation in the header?

filament/backend/src/webgpu/WebGPUSwapChain.cpp
```diff
@@ -42,7 +42,7 @@ void printSurfaceCapabilitiesDetails(wgpu::SurfaceCapabilities const& capabiliti
     if (capabilities.formatCount > 0 && capabilities.formats != nullptr) {
         std::for_each(capabilities.formats, capabilities.formats + capabilities.formatCount,
                 [](wgpu::TextureFormat const format) {
-                    FWGPU_LOGI << "    " << webGPUPrintableToString(format);
+                    FWGPU_LOGI << "    " << webGPUFormatToString(format);
```
kpiascik: why create a new method here?

filament/backend/src/webgpu/WebGPUStrings.h
```diff
@@ -123,6 +123,121 @@ template<typename WebGPUPrintable>
     }
 }
 
+[[nodiscard]] constexpr std::string_view webGPUFormatToString(const wgpu::TextureFormat format) {
```
kpiascik: the function name should probably be `webGPUTextureFormatToString`

filament/backend/src/webgpu/WebGPUTexture.cpp
```diff
@@ -258,6 +266,26 @@ namespace {
     }
 }
 
+[[nodiscard]] WebGPUTexture::MipmapGenerationStrategy determineMipmapGenerationStrategy(
+        const wgpu::TextureFormat format, const SamplerType samplerType, const uint8_t sampleCount,
+        const uint8_t mipmapLevels) {
+    if (mipmapLevels < 2) {
```
kpiascik: please align this check with the one in WebGPUDriver.cpp line 744

filament/backend/src/webgpu/WebGPURenderPassMipmapGenerator.h
```diff
@@ -0,0 +1,282 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef TNT_FILAMENT_BACKEND_WEBGPURENDERPASSMIPMAPGENERATORE_H
+#define TNT_FILAMENT_BACKEND_WEBGPURENDERPASSMIPMAPGENERATORE_H
```
jc3265: seems like theres an extra `E` `WEBGPURENDERPASSMIPMAPGENERATORE` 
```suggestion
#define TNT_FILAMENT_BACKEND_WEBGPURENDERPASSMIPMAPGENERATOR_H
```

filament/backend/src/webgpu/WebGPUTexture.cpp
jc3265: wouldnt it make more sense to say `<= 1` ? they are the same thing but when reading it it would prob make more sense

filament/backend/src/webgpu/WebGPURenderPassMipmapGenerator.h
AndyHovingh: thank you. good catch :)

filament/backend/src/webgpu/WebGPURenderPassMipmapGenerator.h
AndyHovingh: I had this as constexpr, which I could forward declare and define in the implementation file, but that only works if it is only used in the implementation file. Sharing would require it to be in the shared header file.

However, that said, we really don't need the constexpr for this and I would rather clean up the header. So, I will remove constexpr and move the implementation.

filament/backend/src/webgpu/WebGPUStrings.h
AndyHovingh: Works for me 👍

filament/backend/src/webgpu/WebGPUSwapChain.cpp
AndyHovingh: `webGPUPrintableToString` is a template which ultimately uses a std::stringstream to stream characters from Dawn's print functions in order to get a std::string... a bit more expensive than we typically want to do, but not terrible for one-time setup activities like these. However, now that I have defined an efficient function to just provide a static string from a LUD here I am using that instead, at least for wgpu::TextureFormat.

filament/backend/src/webgpu/WebGPUTexture.cpp
AndyHovingh: sure. no problem.

AndyHovingh: sure. no problem.

jc3265 (approved):

PR #8895, opened 2025-06-25
webgpu: remove float filterable requirement

poweifeng:
This will enable webgpu for devices such as pixel 4, which does not support this feature.

### DIFF
--- a/filament/backend/src/webgpu/platform/WebGPUPlatform.cpp
+++ b/filament/backend/src/webgpu/platform/WebGPUPlatform.cpp
@@ -56,10 +56,8 @@ namespace {
 // The SPD Algorithm can make use of up to 12 storage texture attachments
 constexpr uint32_t MAX_MIPMAP_STORAGE_TEXTURES_PER_STAGE = 12u;
 
-constexpr std::array REQUIRED_FEATURES = { wgpu::FeatureName::TransientAttachments,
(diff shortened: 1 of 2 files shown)


filament/backend/src/webgpu/platform/WebGPUPlatform.cpp
```diff
@@ -56,10 +56,8 @@ namespace {
 // The SPD Algorithm can make use of up to 12 storage texture attachments
 constexpr uint32_t MAX_MIPMAP_STORAGE_TEXTURES_PER_STAGE = 12u;
 
-constexpr std::array REQUIRED_FEATURES = { wgpu::FeatureName::TransientAttachments,
-    /*To make filtering assumptions like we want while waiting for Filament to provide that info,
-       float 32 needs to be filterable*/
-    wgpu::FeatureName::Float32Filterable,
+constexpr std::array REQUIRED_FEATURES = {
+    wgpu::FeatureName::TransientAttachments,
```
AndyHovingh, line 60: I MUCH prefer this syntax of having the elements on the line after the curly brace. I discovered last week that if you include a trailing comma on the last item (after `wgpu::FeatureName::RG11B10UfloatRenderable` in this case), then the clang formatting does that for us, which I find much easier to read 👍

PR #8897, opened 2025-06-25
Fix ASAN use-after-stack-free in VulkanPlatform

bejado:
`context` is a stack variable, but it is captured by reference.


poweifeng (approved):
I don't really understand this one. It is passed by `const&` but I don't keep that reference since it's ultimately copied [here](https://github.com/google/filament/blob/main/filament/backend/src/vulkan/VulkanDriver.cpp#L209).

But this change is good nonetheless.

PR #8900, opened 2025-06-26
No need for blending when upscaling

poweifeng:
No need for blending when upscaling
    
    The dst buffer is allocated and should not be blended against.
    The blending code is a relic of a previous code path that's
    no longer present [link].
    
    [link]: https://github.com/google/filament/blame/c049a1bfff0851098bf640159fb6c5daab88d5d4/filament/src/PostProcessManager.cpp
    
    FIXES=422593171

### DIFF
--- a/NEW_RELEASE_NOTES.md
+++ b/NEW_RELEASE_NOTES.md
@@ -7,6 +7,3 @@ We are chaning the way Vulkan buffers are handled. We need to switch over to a m
 appropriate header in [RELEASE_NOTES.md](./RELEASE_NOTES.md).
 
 ## Release notes for next branch cut
-
-- Add new shader define `VARIANT_DEPTH`, defined when a material is compiled for depth variants
-  (e.g., shadows) [**Requires recompiling materials**]
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ repositories {
 }
 
 dependencies {
-    implementation 'com.google.android.filament:filament-android:1.61.2'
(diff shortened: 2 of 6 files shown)


bejado (approved):

filament/src/PostProcessManager.cpp
```diff
@@ -3012,9 +3012,15 @@ FrameGraphId<FrameGraphTexture> PostProcessManager::upscaleBilinear(FrameGraph&
                 data.input = builder.sample(input);
                 data.output = builder.createTexture("upscaled output", outDesc);
                 data.output = builder.write(data.output, FrameGraphTexture::Usage::COLOR_ATTACHMENT);
+
+                TargetBufferFlags clearFlags = TargetBufferFlags::DEPTH;
+                if (translucent) {
+                    clearFlags |= TargetBufferFlags::COLOR0;
+                }
+
```
pixelflinger: I'm really not sure about this.  In theory we shouldn't need blending at all if the source has just been allocated.

I think this code tries to handle the case where the source "upscaled output" has been "forwarded" to an existing buffer, in this case you don't want to clear it.

I think that's why it's written this way; but I'm not entirely sure. Again, if we're going to blend into a cleared buffer, we don't need to actually blend.  So there must have been a reason the code was written this way.

poweifeng: I removed the translucent path per discussion.

pixelflinger (approved):

PR #8901, opened 2025-06-26
webgpu: update pipeline cache key to include render target

AndyHovingh:
Currently, we simply hash on the Filament PipelineState _(utils::hash::MurmurHashFn<filament::backend::PipelineState>())_, but when we create our pipeline we also use the mCurrentRenderTarget, as it contains color attachment info we use to define our pipeline. If we return/use a pipeline built for a different render target _(but with the same pipeline state)_ than is currently being used we can easily get mismatches on attachment states, which we see here. Indeed, I temporarily added some debug logs and was able to see evidence of this happening:

```
...
BIND PIPELINE bilateralBlur with pipeline HASH 3467939382 and render target 0x1203d8448
...
BIND PIPELINE bilateralBlur with pipeline HASH 3467939382 and render target 0x1203d8678
...
```

Thus, we need to use the render target as part of the pipeline hash.

BUGS = [427709441]

### DIFF
--- a/filament/backend/src/webgpu/WebGPUDriver.cpp
+++ b/filament/backend/src/webgpu/WebGPUDriver.cpp
@@ -1100,13 +1100,24 @@ void WebGPUDriver::blit(Handle<HwTexture> destinationTextureHandle, const uint8_
     // todo
 }
 
-void WebGPUDriver::bindPipeline(PipelineState const& pipelineState) {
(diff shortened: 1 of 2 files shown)


jneljneljnel:
Maybe we can combine the colorFormats and include them as part of the hash.

jneljneljnel (approved):

PR #8902, opened 2025-06-26
Fix adjustedMemcpy for compressed formats

anishmgoyal:
Right now, the way we calculate 'bytes per pixel' returns 0 in the case of compressed texture formats. The easy solution in these cases is to simply fall back to the full buffer size.


poweifeng (approved):
much thanks!

filament/backend/src/vulkan/VulkanTexture.cpp
```diff
@@ -248,7 +248,9 @@ void adjustedMemcpy(void* mapped, PixelBufferDescriptor const& p, size_t width,
             }
         }
     } else {
-        size_t const writeSize = pixelSize * (width * height * depth);
+        // Note: if pixelSize = 0, we're dealing with a compressed format; we
```
rafadevai, line 251: isnt this still going to break the case that powei added the fix for the case of copy a subsection? is that operation supported for compressed formats? @poweifeng

poweifeng: no it's not supported for compressed formats.

anishmgoyal: you can't reference individual pixels for a compressed format, in many cases, so you can't copy a subsection

PR #8905, opened 2025-06-26
materials: Add VARIANT_DEPTH define for depth shader variants

bejado:
This lets user materials check if their code is executing inside the depth variant.

BUGS=[428025097]

### DIFF
--- a/libs/filamat/src/shaders/ShaderGenerator.cpp
+++ b/libs/filamat/src/shaders/ShaderGenerator.cpp
@@ -63,6 +63,8 @@ void ShaderGenerator::generateSurfaceMaterialVariantDefines(io::sstream& out,
             filament::Variant::isVSMVariant(variant));
     CodeGenerator::generateDefine(out, "VARIANT_HAS_STEREO",
             hasStereo(variant, featureLevel));
+    CodeGenerator::generateDefine(out, "VARIANT_DEPTH",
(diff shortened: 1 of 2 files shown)


libs/filamat/src/shaders/ShaderGenerator.cpp
```diff
@@ -63,6 +63,8 @@ void ShaderGenerator::generateSurfaceMaterialVariantDefines(io::sstream& out,
             filament::Variant::isVSMVariant(variant));
     CodeGenerator::generateDefine(out, "VARIANT_HAS_STEREO",
             hasStereo(variant, featureLevel));
+    CodeGenerator::generateDefine(out, "VARIANT_DEPTH",
```
poweifeng, line 66: do we need to bump the material version?

bejado: I don't think so, since this is backwards-compatible. Clients that want the new feature will need to re-build materials.

poweifeng: I see, but maybe it'll be good to still add a line in the release note.  In case someone want's to depend on it and like to know which version of filament/matc would have it.

bejado: Yup, good idea!

PR #8906, opened 2025-06-26
docs: post-submit update flow

poweifeng:
- Move directories into more descriptive organization
    - src_mdbook, src_markdeep, src_raw
    - mdbook will be generated from mdbook
    - markdeep will be generated to html
    - raw will be copied without modification
 - Add script that will run on a succesful commit to main. This script will update /docs and commit the changes.


.github/workflows/postsubmit.yml
```diff
@@ -33,3 +33,25 @@ jobs:
...
+        with:
+          fetch-depth: 0
+      - uses: ./.github/actions/linux-prereq
+      - id: get_commit_msg
+        uses: ./.github/actions/get-commit-msg
+      - name: Prerequisites
+        run: pip install selenium
+      - name: Run update script
+        env:
+          GH_TOKEN: ${{ secrets.FILAMENTBOT_TOKEN }}
```
bejado, line 51: I think we should avoid using the Filament bot token. Github now provides you with a token to use in workflows:

https://docs.github.com/en/actions/how-tos/security-for-github-actions/security-guides/automatic-token-authentication

poweifeng: So it doesn't look like I can modify GITHUB_TOKEN to be read/write for the repo: 

image

Probably an org-level restriction

bejado: Oh interesting. Okay filament bot it is!

bejado (approved):

PR #8911, opened 2025-06-30
Metal: re-enable fastmath

bejado:


pixelflinger (commented):
I'm really curious why this is not a problem on Android. Or is it?

bejado:
> I'm really curious why this is not a problem on Android. Or is it?

The original motivation behind disabling fast math was that `saturateMediump` becomes a no-op in Metal with fast math enabled. Thus, any `INF` values in the BRDF calculations ended up as black pixels. Not sure if Android behaves the same way.

pixelflinger:
> > I'm really curious why this is not a problem on Android. Or is it?
> 
> The original motivation behind disabling fast math was that `saturateMediump` becomes a no-op in Metal with fast math enabled. Thus, any `INF` values in the BRDF calculations ended up as black pixels. Not sure if Android behaves the same way.

Okay, it looks like with `fastmath` it assumes it never gets inf/nan. metal doesn't have any other options, like a fast math + inf?

bejado:
> > > I'm really curious why this is not a problem on Android. Or is it?
> > 
> > 
> > The original motivation behind disabling fast math was that `saturateMediump` becomes a no-op in Metal with fast math enabled. Thus, any `INF` values in the BRDF calculations ended up as black pixels. Not sure if Android behaves the same way.
> 
> Okay, it looks like with `fastmath` it assumes it never gets inf/nan. metal doesn't have any other options, like a fast math + inf?

So it does actually, but unfortunately it's only available on newer versions of iOS (iOS 18 it seems).

shaders/src/common_math.glsl
```diff
@@ -3,22 +3,26 @@
...
+#define FLT_EPS                     1e-5
 #endif
 
-#define saturate(x)        clamp(x, 0.0, 1.0)
+#define saturate(x)                 clamp(x, 0.0, 1.0)
+
+#ifdef TARGET_MOBILE
+#define PREVENT_DIV0(n, d, magic)   (n) / max(d, magic)
+#else
+#define PREVENT_DIV0(n, d, magic)   (n) / (d)
```
pixelflinger: I think we need `()` around the whole expression too. So that:

`PREVENT_DIV0(a, b) * 0` is 0 and not `a/0`

pixelflinger (changes requested):

shaders/src/common_math.glsl
pixelflinger: I guess... it wouldn't be a/0 technically; but still, I think it'll be less risky to use the macro with `()`.

bejado: Ah yeah, good point

PR #8917, opened 2025-07-01
use RG32UI for picking instead of RG32F

### LINKED ISSUE #8762: ModelViewer.pick almost always returns 0 on certain devices
**Describe the bug**
In my application, I noticed that the pick() method works incorrectly on certain devices — it almost always returns 0. I verified this using the sample-gltf-viewer from this repository and confirmed that the issue is reproducible.

**To Reproduce**
1. Open the sample-gltf-viewer app on Samsung S22.
2. Tap on the model.
3. Observe that the returned pick result is 0 almost every time.

**Expected behavior**
Tapping on a visible part of the model should return a valid entity, not 0.

**Logs**
There are no errors or crashes in the logs. The issue is purely about unexpected behavior of the pick() method.

**Smartphone:**
Device with issue:
 - Device: Samsung Galaxy S22, S22+, S22 Ultra
 - OS: Android 12

**Additional context**
The issue was tested via BrowserStack. It consistently reproduces on Galaxy S22 devices, works fine on Galaxy S23, and is partially reproducible on Galaxy S24.

Video demonstration:
 - [samsung s22 (issue)](https://drive.google.com/file/d/1Afsl2DbHR851bm0TbIHw7NJo_fxyZ6U5/view?usp=sharing) 
 - [samsung s23 (working)](https://drive.google.com/file/d/1KCpUk89EeIgJoxQs-4_hMOhlQYIUOVoe/view?usp=sharing)

Software information:
 - [samsung s22](https://drive.google.com/file/d/1jTKUSdEXyrrTS324YajHwopEiP80cIK8/view?usp=sharing)
 - [samsung s23](https://drive.google.com/file/d/1mciQnptu_PeNxSwNMsDCnYFTBM4DWS5_/view?usp=sharing)

Filament 1.60.0

pixelflinger:
The reason for this is that according to the GLES 3.0 spec, RG32UI is  guaranteed to be supported for writing, unlike RG32F.

We also use RGBA_INTEGER / UINT for readPixels() which is also a guaranteed format.

Attempt to Fix #8762

### DIFF
--- a/shaders/src/surface_depth_main.fs
+++ b/shaders/src/surface_depth_main.fs
@@ -7,7 +7,7 @@ highp vec4 outPicking;
 #       if MATERIAL_FEATURE_LEVEL == 0
 layout(location = 0) out highp vec4 outPicking;
 #       else
-layout(location = 0) out highp vec2 outPicking;
+layout(location = 0) out highp uvec2 outPicking;
 #       endif
 #   endif
 #else
@@ -63,8 +63,8 @@ void main() {
     outPicking.g = mod(float(object_uniforms_objectId)        , 256.0) / 255.0;
     outPicking.r = vertex_position.z / vertex_position.w;
 #else
-    outPicking.x = intBitsToFloat(object_uniforms_objectId);
-    outPicking.y = vertex_position.z / vertex_position.w;
+    outPicking.x = uint(object_uniforms_objectId);
+    outPicking.y = floatBitsToUint(vertex_position.z / vertex_position.w);
 #endif
 #if __VERSION__ == 100
     gl_FragData[0] = outPicking;
--- a/filament/src/PostProcessManager.cpp
+++ b/filament/src/PostProcessManager.cpp
@@ -569,7 +569,7 @@ PostProcessManager::StructurePassOutput PostProcessManager::structure(FrameGraph
                     // FIXME: the DescriptorSetLayout must specify SAMPLER_FLOAT
                     data.picking = builder.createTexture("Picking Buffer", {
                             .width = width, .height = height,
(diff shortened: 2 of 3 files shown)


shaders/src/surface_depth_main.fs
```diff
@@ -63,8 +63,8 @@ void main() {
     outPicking.g = mod(float(object_uniforms_objectId)        , 256.0) / 255.0;
     outPicking.r = vertex_position.z / vertex_position.w;
 #else
-    outPicking.x = intBitsToFloat(object_uniforms_objectId);
-    outPicking.y = vertex_position.z / vertex_position.w;
+    outPicking.x = object_uniforms_objectId;
+    outPicking.y = floatBitsToInt(vertex_position.z / vertex_position.w);
 #endif
```
poweifeng, line 68: A potential workaround is just to add another path for metal.

pixelflinger: yeah, we could, but I think this is basically a public API issue too.  We're not clear about which readPIxels format are guaranteed to work; but I think using the same ones that GLES3.0 makes sense?

shaders/src/surface_depth_main.fs
```diff
@@ -7,7 +7,7 @@ highp vec4 outPicking;
 #       if MATERIAL_FEATURE_LEVEL == 0
 layout(location = 0) out highp vec4 outPicking;
 #       else
-layout(location = 0) out highp vec2 outPicking;
+layout(location = 0) out highp ivec2 outPicking;
```
bejado: Since you're writing to a RG32UI color texture this should be `uvec2` no? Otherwise Metal complains. This is separate from the missing format issue which I'll look into.

pixelflinger: done. fixes both vk and metal!

PR #8920, opened 2025-07-02
renderdiff: update get-mesa to use also macport on MacOS

poweifeng:


build/common/get-mesa.sh
```diff
@@ -77,16 +77,19 @@ if [[ "$OS_NAME" == "Linux" ]]; then
...
-            exit 1
-        fi
+    if command -v brew > /dev/null 2>&1; then
+        HOMEBREW_NO_INSTALLED_DEPENDENTS_CHECK=true brew install autoconf automake libx11 libxext libxrandr \
+                                                    llvm@${LLVM_VERSION} ninja meson pkg-config libxshmfence
+        # For reasons unknown, this is necessary for pkg-config to find homebrew's packages
+        LOCAL_PKG_CONFIG_PATH="/opt/homebrew/lib/pkgconfig:$PKG_CONFIG_PATH"
+    elif command -v port > /dev/null 2>&1; then
+        sudo port install autoconf automake libx11 libXext libXrandr llvm-${LLVM_VERSION} \
+             ninja meson pkgconfig libxshmfence
```
bejado, line 87:
```suggestion
        sudo port install autoconf automake xorg-libX11 xorg-libXext xorg-libXrandr llvm-${LLVM_VERSION} \
             ninja meson pkgconfig xorg-libxshmfence
```

build/common/get-mesa.sh
```diff
@@ -77,16 +77,19 @@ if [[ "$OS_NAME" == "Linux" ]]; then
...
-        if [ ! command -v brew > /dev/null 2>&1 ]; then
-            echo "Error: need to install homebrew to continue"
-            exit 1
-        fi
+    if command -v brew > /dev/null 2>&1; then
+        HOMEBREW_NO_INSTALLED_DEPENDENTS_CHECK=true brew install autoconf automake libx11 libxext libxrandr \
+                                                    llvm@${LLVM_VERSION} ninja meson pkg-config libxshmfence
+        # For reasons unknown, this is necessary for pkg-config to find homebrew's packages
+        LOCAL_PKG_CONFIG_PATH="/opt/homebrew/lib/pkgconfig:$PKG_CONFIG_PATH"
+    elif command -v port > /dev/null 2>&1; then
```
bejado, line 85:
```suggestion
    elif sudo command -v port > /dev/null 2>&1; then
```

Since `port` is run with `sudo` below, we should check that the command works in the sudoer environment.

bejado (commented):
This almost worked for me, there were just 3 additional things I had to set/run:

```
export LLVM_CONFIG=/opt/local/bin/llvm-config-mp-16
install_name_tool -add_rpath /opt/local/libexec/llvm-16/lib /Users/bendoherty/code/filament/mesa/out/lib/libGL.1.dylib
pip install tifffile
```

poweifeng:
> This almost worked for me, there were just 3 additional things I had to set/run:
> 
> ```
> export LLVM_CONFIG=/opt/local/bin/llvm-config-mp-16
> install_name_tool -add_rpath /opt/local/libexec/llvm-16/lib /Users/bendoherty/code/filament/mesa/out/lib/libGL.1.dylib
> pip install tifffile
> ```

The first two I added to the script, but the last line I think is covered [here](https://github.com/google/filament/blob/main/test/renderdiff/generate.sh#L27). So I think this might be because I gave you the wrong instruction for the venv part.  Should be `. /venv/bin/activate`.

PR #8923, opened 2025-07-02
API BREAKAGE: remove Texture::generatePrefilterMipmap

pixelflinger:
This feature is now supported by `libfilament-generatePrefilterMipmap.a`


0001-attempt-to-fix-external-streams-with-protected-conte.patch
```diff
@@ -0,0 +1,299 @@
+From 0c10259d1d69c6e4ffcc6892f175982d8291bbab Mon Sep 17 00:00:00 2001
```
poweifeng: not relevant to this?

z3moon (approved):

libs/generatePrefilterMipmap/src/generatePrefilterMipmap.cpp
```diff
@@ -0,0 +1,228 @@
+/*
+ * Copyright (C) 2021 The Android Open Source Project
```
poweifeng:
```suggestion
 * Copyright (C) 2025 The Android Open Source Project
```

libs/generatePrefilterMipmap/README.md
```diff
@@ -0,0 +1,43 @@
+# IBL Prefilter
```
poweifeng: This seems to be copied from iblprefilter. Needs to be updated.

libs/generatePrefilterMipmap/include/filament-generatePrefilterMipmap/generatePrefilterMipmap.h
```diff
@@ -0,0 +1,94 @@
+/*
+ * Copyright (C) 2021 The Android Open Source Project
```
poweifeng:
```suggestion
 * Copyright (C) 2025 The Android Open Source Project
```

CMakeLists.txt
```diff
@@ -786,6 +786,7 @@ add_subdirectory(${LIBRARIES}/filabridge)
 add_subdirectory(${LIBRARIES}/filaflat)
 add_subdirectory(${LIBRARIES}/filagui)
 add_subdirectory(${LIBRARIES}/filameshio)
+add_subdirectory(${LIBRARIES}/generatePrefilterMipmap)
```
poweifeng, line 789: nit: would `envprefilter` be a reasonable alternative name?  So that we'd get a lib name that's parallel to `iblprefilter` (two seem to be serving same purpose for different maps).

pixelflinger: They're actually mostly doing the same thing, but one is using the cpu and one is using the gpu.

PR #8928, opened 2025-07-04
webgpu: render pipeline cache

AndyHovingh:
BUGS = [427709441]


filament/backend/src/webgpu/WebGPUBufferBase.h
```diff
@@ -40,7 +40,7 @@ class WebGPUBufferBase /* intended to be extended */ {
 private:
     const wgpu::Buffer mBuffer;
     // WEBGPU_BUFFER_SIZE_MODULUS (e.g. 4) bytes to hold any extra chunk we need.
-    std::array<uint8_t, WEBGPU_BUFFER_SIZE_MODULUS> mRemainderChunk{};
+    std::array<uint8_t, F_WEBGPU_BUFFER_SIZE_MODULUS> mRemainderChunk{};
```
jc3265: is this `F` because of "filament"? why not just spell it out

filament/backend/src/webgpu/WebGPUBufferBase.h
```diff
@@ -40,7 +40,7 @@ class WebGPUBufferBase /* intended to be extended */ {
 private:
     const wgpu::Buffer mBuffer;
     // WEBGPU_BUFFER_SIZE_MODULUS (e.g. 4) bytes to hold any extra chunk we need.
```
jc3265: nit: this comment should reflect the new name

filament/backend/src/webgpu/WebGPURenderPassMipmapGenerator.h
```diff
@@ -56,7 +56,7 @@ class WebGPURenderPassMipmapGenerator final {
     void generateMipmap(wgpu::CommandEncoder const&, wgpu::Texture const&,
             wgpu::RenderPipeline const&, uint32_t layer, uint32_t mipLevel);
 
-    wgpu::Device const& mDevice;
+    wgpu::Device mDevice;
```
jc3265, line 59: why this change?

filament/backend/src/webgpu/WebGPUPipelineCache.cpp
```diff
@@ -0,0 +1,383 @@
...
+    outKey.stencilFrontWriteMask = request.stencilState.front.writeMask;
+    outKey.stencilBackCompare = request.stencilState.back.stencilFunc;
+    outKey.stencilBackFailOperation = request.stencilState.back.stencilOpStencilFail;
+    outKey.stencilBackDepthFailOperation = request.stencilState.back.stencilOpDepthFail;
+    outKey.stencilBackPassOperation = request.stencilState.back.stencilOpDepthStencilPass;
+    outKey.cullingMode = request.rasterState.culling;
+    outKey.inverseFrontFaces = toUint8(request.rasterState.inverseFrontFaces);
+    outKey.depthWriteEnabled = toUint8(request.rasterState.depthWrite);
+    outKey.depthCompare = request.rasterState.depthFunc;
+    outKey.depthClampEnable = toUint8(request.rasterState.depthClamp);
```
jc3265: out of curiosity, why not just name them the same thing? `depthClamp` instead of `depthClampEnable`

filament/backend/src/webgpu/WebGPUPipelineCache.cpp
```diff
@@ -0,0 +1,383 @@
...
+    return pipeline;
+}
+
+bool WebGPUPipelineCache::RenderPipelineKeyEqual::operator()(RenderPipelineKey const& key1,
+        RenderPipelineKey const& key2) const {
+    return 0 == memcmp(reinterpret_cast<void const*>(&key1), reinterpret_cast<void const*>(&key2),
+                        sizeof(key1));
+}
+
+void WebGPUPipelineCache::collectGarbage() {
```
jc3265: in theory this could also be called `removeExpiredPipelines`

filament/backend/src/webgpu/WebGPUBufferBase.h
AndyHovingh: I could and will. I just didn't earlier because some of these variable/constant names are really long already. The F isn't really important as much as we avoid naming collisions since this is in the already large `filament::backend` namespace. But, if it already causes friction to the mind reading it over, I agree that just spelling it out would probably be better. 👍

filament/backend/src/webgpu/WebGPUBufferBase.h
AndyHovingh: thank you. good catch

filament/backend/src/webgpu/WebGPURenderPassMipmapGenerator.h
AndyHovingh: Elsewhere we just have a copy of the wrapper, not a ref like this. The advantage is that the C++ wrapper classes do ref counting, so we can avoid a dangling pointers in certain scenarios; it is a little safer. My motivation for the change was that I was getting inexplicable Dawn null pointer bugs for a bit. I believe I squashed those and they were not related to this, but keeping this is probably not a bad idea anyway.

filament/backend/src/webgpu/WebGPUPipelineCache.cpp
AndyHovingh: I could. no problem 👍  no reason really, just putting things together quickly is all.

filament/backend/src/webgpu/WebGPUPipelineCache.cpp
AndyHovingh: Yes. I took the name from the Vulkan implementation, but I like that better, as it is more self describing. If the thing needed to garbage collect other things another `collectGarbage` function could be created to call the removal of all the individual things. I like it.

PR #8932, opened 2025-07-07
Change unfilterable to filterable

poweifeng:


libs/filabridge/include/private/filament/SamplerInterfaceBlock.h
```diff
@@ -84,7 +84,7 @@ class SamplerInterfaceBlock {
             Type type;                      // type of this sampler
             Format format;                  // format of this sampler
             Precision precision;            // precision of this sampler
-            bool unfilterable = false;      // whether the sampling should be unfiltered.
+            bool filterable = false;        // whether the sampling should be filterable.
```
pixelflinger: is `false` the correct default here?

tools/matc/src/matc/ParametersProcessor.cpp
```diff
@@ -267,7 +267,19 @@ static bool processParameter(MaterialBuilder& builder, const JsonishObject& json
         auto precision = precisionValue ? Enums::toEnum<ParameterPrecision>(
                 precisionValue->toJsonString()->getString()) : ParameterPrecision::DEFAULT;
 
-        auto unfilterable = unfilterableValue ? unfilterableValue->toJsonBool()->getBool() : false;
+        // For samplers without `filterable` defined, we use the following logic
+        //   - float sampler can be filterable or not, default to filterable
+        //   - int sampler is not filterable
+        //   - shadow sampler uses comparison operator and does not need filterability defined.
```
pixelflinger: shouldn't shadow samplers always be filterable?

pixelflinger (commented):
thanks for the change, just a couple questions...

libs/filabridge/include/private/filament/SamplerInterfaceBlock.h
poweifeng: you're right.  But I went and remove all the defaults and just explicitly defined them in `SibGenerator.cpp`. To reduce confusion later.

tools/matc/src/matc/ParametersProcessor.cpp
poweifeng: I will defer to you on this since I really don't know. 

But here I changed it so that we'd assert if we see a shadow sampler. I think the only permissable params for user materials is `float` and `int` for sampler formats?

pixelflinger: so `int` can never be filterable.
`shadow` is always filterable (because the texture most be DEPTH and the mode must be COMPARE, and in that case we're filterable)
`float` can be either.

So I think, we should never allow the `filterable` parameter for  `int` and `shadow` because it's always respectively `false` and `true`.

And only allow it for `float`, because that will actually depend at runtime on the type of texture bound.

poweifeng: good point and amended

PR #8933, opened 2025-07-07
resgen: output offset/size as constexpr

pixelflinger:
`resgen` now generates `constexpr int` variables for a resource size and offset instead of an `extern` variable. The assembly file also doesn't generate the corresponding global variables anymore, saving some binary space.


tools/resgen/src/main.cpp
```diff
@@ -245,7 +245,8 @@ int main(int argc, char* argv[]) {
     ostringstream headerStream;
     headerStream << "#ifndef " << packagePrefix << "H_" << endl
             << "#define " << packagePrefix << "H_" << endl << endl
-            << "#include <stdint.h>" << endl << endl
+            << "#include <stdint.h>\n"
+            << "#include <stddef.h>\n" << endl
```
poweifeng: I think you need to add `stddef.h` for line 267 as well.

tools/resgen/src/main.cpp
```diff
@@ -245,12 +245,12 @@ int main(int argc, char* argv[]) {
...
```
poweifeng: do we still need stddefs if you're no longer using `size_t`?

poweifeng (commented):
nit: would change the title of the commit (not PR) to be more accurate

tools/resgen/src/main.cpp
pixelflinger: 🦅 👀

PR #8937, opened 2025-07-08
Integrate opencv for image comparison in testing.

granade-work:
This adds the opencv and zlib libraries in support of a to-be-completed image comparison feature for backend tests.


poweifeng:
I have some reservation about adding this dependency. Do you have a design in mind for this work? It might be helpful to chat about this offline.

PR #8938, opened 2025-07-09
gl: Enhance error reporting for asynchronous shader operations

z3moon:
When ShaderCompilerService asynchronously compiles and links shaders, unexpected errors (e.g., GL_CONTEXT_LOST) can occur after the initial GL call, complicating diagnostics.

This change improves error context by collecting the GL error status immediately following relevant GL calls, and again when querying shader compilation and program link statuses. This provides a more detailed understanding of error origins.

BUGS=[373396840]

### DIFF
--- a/filament/backend/src/opengl/ShaderCompilerService.cpp
+++ b/filament/backend/src/opengl/ShaderCompilerService.cpp
@@ -66,8 +66,8 @@ static std::string to_string(int const i) { return std::to_string(i); }
 static std::string to_string(float const f) { return "float(" + std::to_string(f) + ")"; }
 
 static void logCompilationError(ShaderStage shaderType, const char* name, GLuint shaderId,
-        Program::ShaderBlob const& sourceCode) noexcept;
-static void logProgramLinkError(char const* name, GLuint program) noexcept;
+        GLenum callResult, Program::ShaderBlob const& sourceCode) noexcept;
+static void logProgramLinkError(char const* name, GLuint program, GLenum callResult) noexcept;
 
 static void process_GOOGLE_cpp_style_line_directive(OpenGLContext const& context, char* source,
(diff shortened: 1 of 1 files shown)


filament/backend/src/opengl/ShaderCompilerService.cpp
```diff
@@ -609,6 +611,7 @@ void ShaderCompilerService::executeTickOps() noexcept {
             GLuint const shaderId = glCreateShader(glShaderType);
             glShaderSource(shaderId, GLsizei(count), shaderStrings.data(), lengths.data());
             glCompileShader(shaderId);
+            token->gl.compilationCallResult = glGetError();
```
pixelflinger: I don't want to call glGetError() immediately after `glCompileShader` because that too might make `glCompileShader` synchronous. I'm actually not sure it will, but I think it's better to call glGetError() only once we see that the link has failed.

filament/backend/src/opengl/ShaderCompilerService.cpp
```diff
@@ -679,6 +683,7 @@ void ShaderCompilerService::executeTickOps() noexcept {
         }
     }
     glLinkProgram(program);
+    token->gl.linkCallResult = glGetError();
```
pixelflinger: same problem here, this might make `glLinkProgram` synchronous.

filament/backend/src/opengl/ShaderCompilerService.cpp
z3moon: I worried that the error of the initial gl compilation/link call might be overwritten by other subsequent gl calls for asynchronous mode. But it may be okay for the context lost case. Let me update it.

pixelflinger (approved):

PR #8939, opened 2025-07-09
webgpu: Return Feature Levels

jc3265:
BUGS = [420985749]

### DIFF
--- a/filament/backend/src/webgpu/WebGPUDriver.cpp
+++ b/filament/backend/src/webgpu/WebGPUDriver.cpp
@@ -621,8 +621,23 @@ bool WebGPUDriver::isWorkaroundNeeded(Workaround) {
 }
 
 FeatureLevel WebGPUDriver::getFeatureLevel() {
-    //todo
-    return FeatureLevel::FEATURE_LEVEL_1;
+
+    // If the max sampler counts do not meet FeatureLevel2 standards, then this is an FeatureLevel1
+    // device.
+    const auto& featureLevel2 = FEATURE_LEVEL_CAPS[+FeatureLevel::FEATURE_LEVEL_2];
+    if (mDeviceLimits.maxSamplersPerShaderStage < featureLevel2.MAX_VERTEX_SAMPLER_COUNT ||
+            mDeviceLimits.maxSamplersPerShaderStage < featureLevel2.MAX_FRAGMENT_SAMPLER_COUNT) {
+        return FeatureLevel::FEATURE_LEVEL_1;
+    }
+
+    // If the max sampler counts do not meet FeatureLevel3 standards, then this is an FeatureLevel2
+    // device.
+    const auto& featureLevel3 = FEATURE_LEVEL_CAPS[+FeatureLevel::FEATURE_LEVEL_3];
+    if (mDeviceLimits.maxSamplersPerShaderStage < featureLevel3.MAX_VERTEX_SAMPLER_COUNT ||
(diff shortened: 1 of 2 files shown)


filament/backend/src/webgpu/WebGPUDriver.cpp
```diff
@@ -621,8 +621,23 @@ bool WebGPUDriver::isWorkaroundNeeded(Workaround) {
 }
 
 FeatureLevel WebGPUDriver::getFeatureLevel() {
-    //todo
-    return FeatureLevel::FEATURE_LEVEL_1;
+
+    // If the max sampler counts do not meet FeatureLevel2 standards, then this is an FeatureLevel1
```
AndyHovingh, line 625: @jc3265 @poweifeng should we be defensive and also verify we can even support FeatureLevel1? I assume this is a highly unlikely scenario, but maybe we panic in that case (unless returning FeatureLevel0 makes any sense / would be supported)?

jc3265: so the case for FL 1 is

    // If cubemap arrays are not supported, then this is an FL1 device.
    if (!mContext.isImageCubeArraySupported()) {
        return FeatureLevel::FEATURE_LEVEL_1;
    }

 which if im not mistaken are supported by default on WebGPU 
 thats why i skipped that check and jumped to FL2

AndyHovingh: What I am asking is if we cannot even support FL1 requirements? So far we fallback to FL1 assuming we can, but as far as I can tell we never actually verify that we can even meet those settings. Maybe I am missing something.

poweifeng: maybe the webgpu minspec already meets the fl1 bar. 

FL0 is strictly for GLES2, so webgpu backend should not fall back to that.

filament/backend/src/webgpu/WebGPUDriver.cpp
```diff
@@ -621,8 +621,23 @@ bool WebGPUDriver::isWorkaroundNeeded(Workaround) {
 }
 
 FeatureLevel WebGPUDriver::getFeatureLevel() {
```
AndyHovingh, line 623: @pixelflinger @poweifeng @bejado This change certainly informs Filament what feature level the WebGPU backend driver can support at runtime, but it does not address how we configure our use of the GPU when the backend gets created, namely the device settings/limits/features. @pixelflinger, maybe I am misremembering, but I thought you mentioned something earlier about honoring the requested feature level the client requests when creating the Engine instance, that we both ensure we can support the minimum requirements for the level IN ADDITION to not requesting/using more resources than required for that level. Today, we basically request a WebGPU adapter and configure a device as if feature level 3 is requested (e.g. "give me everything you have"). This would imply that we may have different required/optional features and different limits per feature level that we would use when requesting a WebGPU adapter and creating a properly configured device for it. Is that correct? If so, do we have an example elsewhere where we do this? Or, maybe we do not have enough information at this stage to properly handle that functionality?

Or is limiting the use resource utilization ultimately handled in the Filament layer above the drivers/backends? In which case, what we have makes sense.

Also, related to this, we always prefer the `HighPerformance` power preference to `LowPower` when selecting an adapter _(see [WebGPUPlatform.cpp ~lines 513-561, the selectPreferredAdapter(...) function](https://github.com/google/filament/blob/29ed42fa9814b18725c1d547adb63d1504d6ebfc/filament/backend/src/webgpu/platform/WebGPUPlatform.cpp#L513-L561) for more detail)_. Should this depend on desktop vs mobile?

PR #8944, opened 2025-07-10
better UBO invalidation

pixelflinger:
Only invalidate UBOs when the data actually changes. It turns out it's a fairly common case for application to update MaterialInstance parameters with the same value; we now check for this before marking the UBO "dirty".

The overhead of this check should be very small but it could save many very expensive buffer updates.

### DIFF
--- a/filament/src/UniformBuffer.cpp
+++ b/filament/src/UniformBuffer.cpp
@@ -16,6 +16,17 @@
 
 #include "UniformBuffer.h"
 
+#include <utils/Allocator.h>
+#include <utils/compiler.h>
+#include <utils/debug.h>
+#include <utils/ostream.h>
+
+#include <math/mat3.h>
+
+#include <utility>
+
+#include <stdint.h>
+#include <stddef.h>
 #include <stdlib.h>
 #include <string.h>
 
@@ -94,7 +105,11 @@ void UniformBuffer::free(void* addr, size_t) noexcept {
 
 template<size_t Size>
 void UniformBuffer::setUniformUntyped(size_t const offset, void const* UTILS_RESTRICT v) noexcept{
-    setUniformUntyped<Size>(invalidateUniforms(offset, Size), 0ul, v);
+    if (UTILS_LIKELY(invalidateNeeded<Size>(offset, v))) {
+        void* const addr = getUniformAddress(offset);
+        setUniformUntyped<Size>(addr, v);
+        invalidateUniforms(offset, Size);
+    }
 }
 
 template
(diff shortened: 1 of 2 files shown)


filament/src/UniformBuffer.h
```diff
@@ -66,15 +69,22 @@ class UniformBuffer { // NOLINT(cppcoreguidelines-pro-type-member-init)
...
+    // Checks if a range needs to be invalidated by comparing the current value.
+    template<size_t Size>
+    bool invalidateNeeded(size_t const offset, void const* UTILS_RESTRICT v) const {
+        assert_invariant(offset + Size <= mSize);
+        void* const UTILS_RESTRICT addr = getUniformAddress(offset);
+        return bool(memcmp(addr, v, Size)); // inlined
+    }
+
+    // Invalidate a range of uniforms and return a pointer to it. offset and size given in bytes
+    void invalidateUniforms(size_t const offset, size_t const size) const {
```
poweifeng, line 81: Doesn't seem like `offset` and `size` are actually used here.  Couldn't we move `mSomethingDirty` to `invalidate()` and just have that?

filament/src/UniformBuffer.cpp
```diff
@@ -111,10 +126,9 @@ void UniformBuffer::setUniformUntyped<64ul>(size_t offset, void const* UTILS_RES
 template<size_t Size>
 void UniformBuffer::setUniformArrayUntyped(size_t const offset, void const* UTILS_RESTRICT begin, size_t const count) noexcept {
     constexpr size_t stride = (Size + 0xFu) & ~0xFu;
-    size_t arraySize = stride * count - stride + Size;
-    void* UTILS_RESTRICT p = invalidateUniforms(offset, arraySize);
```
z3moon, line 115: It looks like this function doesn't call invalidateUniforms. I wonder if it is okay not to invalidate it for the array values.

filament/src/UniformBuffer.h
pixelflinger: ah yes, you're right, but there is a reason for it -- I wanted this internal API to be able to invalidate ranges; currently, it just invalidate the whole thing; but I wanted to keep that info.

filament/src/UniformBuffer.cpp
pixelflinger: oof. very good catch!

pixelflinger: should be fixed now

poweifeng (approved):

z3moon (approved):

PR #8946, opened 2025-07-10
 vk: VulkanBuffer cannot skip staging buffer after a read

poweifeng:
The previous introduction of VulkanBufferProxy uses the refcount of the backing VulkanBuffer as an indication whether there's a write in progress or not. But we also want to make sure there is no read in progress (i.e. not referenced in a descriptor set). This commit fixes the bookkeeping while trying to clean up the relevant structs.

### DIFF
--- a/filament/backend/src/vulkan/VulkanHandles.h
+++ b/filament/backend/src/vulkan/VulkanHandles.h
@@ -26,6 +26,7 @@
 #include "VulkanFboCache.h"
 #include "VulkanSwapChain.h"
 #include "VulkanTexture.h"
+#include "vulkan/VulkanCommands.h"
 #include "vulkan/memory/Resource.h"
 #include "vulkan/utils/Definitions.h"
 #include "vulkan/utils/StaticVector.h"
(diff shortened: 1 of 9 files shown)


rafadevai:
LGTM

rafadevai:
so now with this change if I understanding it correctly, will negate the bypass?

filament/backend/src/vulkan/VulkanHandles.h
```diff
@@ -419,6 +421,9 @@ struct VulkanVertexBuffer : public HwVertexBuffer, fvkmemory::Resource {
             fvkmemory::resource_ptr<VulkanVertexBufferInfo> vbi);
     void setBuffer(fvkmemory::resource_ptr<VulkanBufferObject> bufferObject, uint32_t index);
 
+    // TODO: because VulkanBufferObject is backed by VulkanBufferProxy, which could switch out the
```
rafadevai, line 431: So in the case of switching a buffer is expected also to update the descriptor sets to keep things consistent, but we are not doing that right now.

So better put a comment in the VulkanBufferProxy class where the buffer change could be added?

bejado (approved):

filament/backend/src/vulkan/VulkanHandles.h
poweifeng: done

PR #8949, opened 2025-07-11
Add per-byte image comparison to backend tests

granade-work:
Instead of importing opencv as I had intended with https://github.com/google/filament/pull/8937 this open codes a very simple, almost trivial image comparison that can nevertheless be expanded to provide image diffs and adjustable image comparison thresholds for golden images.

I had not previously appreciated that this could be quite this simple since the captures generated in these tests are so uniform.

This code is mostly inactive as there are no examples of off-by-one-pixel images in the current test corpus. As such image artifacts are encountered, the allowance of number of non-matching bytes (for errant bad pixels) or the threshold for counting a byte as matching (for color selection variation) can be specified to the constructor of a ScreenshotParams struct for a particular test.

BUGS=[398199436]

### DIFF
--- a/filament/backend/test/ImageExpectations.cpp
+++ b/filament/backend/test/ImageExpectations.cpp
@@ -36,12 +36,14 @@
 namespace test {
 
 ScreenshotParams::ScreenshotParams(int width, int height, std::string fileName,
-        uint32_t expectedHash, bool isSrgb)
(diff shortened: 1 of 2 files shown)


bejado (approved):

bejado:
Thanks!

PR #8953, opened 2025-07-14
glTF comparison tests

mhoff12358:
Adds `gltf_comparison` as a cmake build target, designed to run a single parameterized test to be run for each glTF model in the test case directory.

A modified copy of the EXPECT_IMAGE image diffing macro is pulled in from the filament backend tests to do side-by-side image comparisons of the rendered output.

Currently something is wrong with the glTF loading/rendering and only a solid color is drawn with what seems to be artifacts from other graphics memory showing up randomly. The solid color is from the glTF file itself, so I believe the issue is with the uber archive of materials.

There is some awkward multithreading code required, as the filament app runs on the main thread for macOS reasons, and the gtest code needs to wait on it from another thread.


poweifeng:
closing this for now.  We can revisit if there's a need later.

PR #8955, opened 2025-07-14
separate all postfx materials into their own material resource and cpp file

pixelflinger:


z3moon (approved):

show50726:
So the plan is we will be able to exclude a specific post process through the cmake files, and at runtime if we find the resource bins of a specific post process effect doesn't exist we just skip the registration of it?

show50726 (approved):

poweifeng (approved):

pixelflinger:
> So the plan is we will be able to exclude a specific post process through the cmake files, and at runtime if we find the resource bins of a specific post process effect doesn't exist we just skip the registration of it?

The plan is not fully formalized in my head yet. Currently what i'm imagining is that postfx materials will be in their own .a and for the normal build we'll just include them all; but  other build systems could chose to not include them.

PR #8957, opened 2025-07-15
gl: Move program cache retrieval to the thread pool

z3moon:
This commit refactors the ShaderCompilerService to move the logic for retrieving compiled programs from the cache off the main thread and into the compiler thread pool.

Previously, the cache lookup was performed on the main thread before a compilation job was created. If the cache was slow or contained a large number of entries, this could cause a noticeable stall, impacting application responsiveness.

By moving the cache retrieval into the compilation task, the lookup now happens on the background thread. This alleviates the main thread stall.

BUGS=[430628799]


filament/backend/src/opengl/ShaderCompilerService.cpp
```diff
@@ -262,22 +262,20 @@ ShaderCompilerService::program_token_t ShaderCompilerService::createProgram(
...
     switch (mMode) {
         case Mode::THREAD_POOL: {
             mCompilerThreadPool.queue(priorityQueue, token,
                     [this, &gl, program = std::move(program), token]() mutable {
-                        compileShaders(gl, std::move(program.getShadersSource()),
-                                program.getSpecializationConstants(), program.isMultiview(), token);
-                        linkProgram(gl, token);
+                        // Try retrieving the program from the cache first. If no program found,
+                        // a normal compilation/linking process is performed.
+                        if (!tryRetrievingProgram(mBlobCache, mDriver.mPlatform, program, token)) {
```
poweifeng, line 273: if we successfully retrieved it from the cache, do we still want to try caching it below? You can then avoid having the `retrievedFromBlobCache` boolean.

z3moon: yeah, the `retrievedFromBlobCache` check is now done inside the `tryCachingProgram` function.

poweifeng: yes, I mean you don't even need `retrievedFromBlobCache` if you just keep it local to 271-277

z3moon: I see what you mean. But the intent is to call `tryCachingProgram` after we send a signal as described in the comments below. For additional context, `tryCachingProgram` is also called later in the `initialize` method for other modes (SYNC & ASYNC).

poweifeng: yeah, I guess there's also the non THREAD_POOL case to consider.

pixelflinger (approved):

PR #8964, opened 2025-07-16
gl: amortize shader compilation over N frames

elizagamedev:
On platforms which don't support parallel shader compilation (e.g. Firefox), we can sort of emulate it by queueing up any requests to `Material::compile()` and compiling each shader over multiple frames.

Compiling over consecutive frames slows things quite noticeably. Instead, space each compilation over a number of ticks (16 chosen arbitrarily).

If we need a program NOW, we can bump a program queued for much later to the front. We also avoid compiling any queued programs on this same frame to limit stuttering.

### DIFF
--- a/filament/backend/src/CompilerThreadPool.cpp
+++ b/filament/backend/src/CompilerThreadPool.cpp
@@ -32,6 +32,7 @@ CompilerThreadPool::~CompilerThreadPool() noexcept {
     assert_invariant(mCompilerThreads.empty());
     assert_invariant(mQueues[0].empty());
     assert_invariant(mQueues[1].empty());
+    assert_invariant(mQueues[2].empty());
 }
 
 void CompilerThreadPool::init(uint32_t threadCount,
@@ -55,7 +56,7 @@ void CompilerThreadPool::init(uint32_t threadCount,
                 });
 
                 FILAMENT_TRACING_VALUE(FILAMENT_TRACING_CATEGORY_FILAMENT, "CompilerThreadPool Jobs",
-                        mQueues[0].size() + mQueues[1].size());
+                        mQueues[0].size() + mQueues[1].size() + mQueues[2].size());
 
                 if (UTILS_LIKELY(!mExitRequested)) {
                     Job job;


elizagamedev:
Branch name says WIP but it's no longer WIP.

filament/backend/src/opengl/ShaderCompilerService.h
```diff
@@ -134,6 +134,11 @@ class ShaderCompilerService {
     using ContainerType = std::tuple<CompilerPriorityQueue, program_token_t, Job>;
     std::vector<ContainerType> mRunAtNextTickOps;
 
+    uint32_t mNumProgramsCreatedSynchronouslyThisTick = 0u;
+    uint32_t mNumTicksUntilNextSynchronousProgram = 0u;
+    using PendingSynchronousProgram = std::tuple<program_token_t, Program>;
+    std::vector<PendingSynchronousProgram> mPendingSynchronousPrograms;
+
```
pixelflinger: what are the threading conditions for all these new variables? do they need any concurrent access protection?

filament/src/details/Material.h
```diff
@@ -132,7 +132,8 @@ class FMaterial : public Material {
     // Must be called outside of backend render pass.
     // Must be called before getProgram() below.
     void prepareProgram(Variant const variant,
-            backend::CompilerPriorityQueue const priorityQueue = CompilerPriorityQueue::HIGH) const noexcept {
+            backend::CompilerPriorityQueue const priorityQueue = CompilerPriorityQueue::IMMEDIATE)
```
pixelflinger: Maybe let's remove that default parameter entirely.

For systems that have the parallel shader compilation, what does `IMMEDIATE` mean? Is it equivalent to `HIGH`?

filament/backend/include/backend/DriverEnums.h
```diff
@@ -674,6 +674,7 @@ enum class Precision : uint8_t {
  * Shader compiler priority queue
  */
 enum class CompilerPriorityQueue : uint8_t {
+    IMMEDIATE,
```
pixelflinger: This exposes `IMMEDIATE` to the public API in `Material.h` in `Material::compile()` and I don't think that's good. The public API should be unchanged here, and `IMMEDIATE` doesn't mean anything for them.

I feel like `IMMEDIATE` should be something completely internal to the `ShaderCompilerService`.

filament/src/details/Material.h
elizagamedev: Done.

`IMMEDIATE` is not given any special treatment on non-synchronous platforms, but it is technically one priority level over high. It's what's used whenever there's a cache miss. Technically this still benefits parallel platforms, since they'll now prioritize those cache misses over any other shaders which may be prioritized as `HIGH`.

filament/backend/include/backend/DriverEnums.h
elizagamedev: Is it necessarily bad to expose a way for clients to greedily compile shaders like this? For example, if you had a game with a loading screen or something during which you wanted to precompile shaders, you'd probably want them `IMMEDIATE`ly instead of relying on the 16-frame interval "background" compilation.

elizagamedev:
> In addition to the comments inline, I think that:
> 
> * we need a feature-flag to turn this off entirely; in case something unexpected goes wrong in prod (see Engine's feature flag system)
> * we need to be very confident that it doesn't change the existing behavior if the amortized mode is not used.
> 
> It would also be nice to have an easy way (Even if it's at compile time) to force the amortized mode, in order to test it.

I'll add a flag `backend.disable_amortized_shader_compile`. Setting `backend.disable_parallel_shader_compile` should be enough to force it on.

filament/backend/src/opengl/ShaderCompilerService.h
elizagamedev: These will only ever be accessed by the main thread. I left a comment explaining the details.

filament/backend/include/backend/DriverEnums.h
pixelflinger: ok, but then let's make sure to update the public docs to explain what this does and when to use it or not use it. It sounds like it's like this:

IMMEDIATE: I know I will need this right away
HIGH: I know I will need this
LOW: I'm not sure if I'll need this

sounds correct?

elizagamedev: Done.

pixelflinger (approved):

pixelflinger:
I would like @z3moon 's eyes on this too.

filament/backend/include/backend/DriverEnums.h
```diff
@@ -672,9 +672,21 @@ enum class Precision : uint8_t {
...
+ * order of priority, then insertion order. See Material::compile().
  */
 enum class CompilerPriorityQueue : uint8_t {
+    /** We need this program NOW.
+     *
+     * When passed as an argument to Material::compile(), if the platform doesn't support parallel
+     * compilation, but does support amortized shader compilation, the given shader program will be
+     * synchronously compiled.
+     */
+    IMMEDIATE,
```
z3moon: This will cause a crash at `mCompilerThreadPool.queue(priorityQueue, token,` in the (new) `compileProgram` method for the `priorityQueue` being LOW, which becomes 2 now. Please see the `queue` method to see how the value is used. You may want to update the CompilerThreadPool accordingly.

filament/backend/src/opengl/ShaderCompilerService.h
```diff
@@ -142,6 +146,14 @@ class ShaderCompilerService {
     void executeTickOps() noexcept;
     bool cancelTickOp(program_token_t const& token) noexcept;
 
+    bool shouldCompileSynchronousProgramThisTick() const noexcept;
+    void compilePendingSynchronousPrograms() noexcept;
+    void compilePendingSynchronousProgramNow(program_token_t const& token) noexcept;
+    void cancelPendingSynchronousProgram(program_token_t const& token) noexcept;
+
+    // Compile the program. Does NOT attempt to use the shader blob cache.
```
z3moon: The comment looks misinformed. It still uses the shader blob cache if available. It calls tryRetrievingProgram internally.

filament/backend/src/opengl/ShaderCompilerService.cpp
```diff
@@ -59,6 +59,11 @@ namespace filament::backend {
 
 using namespace utils;
 
+// How many ticks do we wait in between compiling non-IMMEDIATE programs synchronously?
+constexpr uint32_t NUM_TICKS_BETWEEN_PROGRAMS = 16;
+// How many non-IMMEDIATE synchronous programs are we allowed to compile on the same tick?
+constexpr uint32_t MAX_NUM_SYNCHRONOUS_PROGRAMS_PER_FRAME = 1;
```
z3moon, line 66: Can we make these values configurable?

filament/backend/src/opengl/ShaderCompilerService.cpp
```diff
@@ -255,14 +261,41 @@ void ShaderCompilerService::terminate() noexcept {
...
 
     // Create a token. A callback condition (handle) is internally created upon token creation.
     auto token = std::make_shared<OpenGLProgramToken>(*this, name);
     if (UTILS_UNLIKELY(gl.isES2())) {
         token->attributes = std::move(program.getAttributes());
     }
 
-    // Initiate program compilation.
+    if (mMode == Mode::SYNCHRONOUS) {
+        if (priorityQueue == CompilerPriorityQueue::IMMEDIATE ||
```
z3moon: The term `IMMEDIATE` appears to function as intended for SYNCHRONOUS mode, but not for other modes. Can we rename it more universally like CRITICAL, URGENT, HIGHEST, etc? (and make it work immediately for SYNCHRONOUS mode)

filament/backend/src/opengl/ShaderCompilerService.cpp
pixelflinger: I think that could come in a later PR;  but I agree we probably want control over this. Probably via DriverConfig.

filament/backend/include/backend/DriverEnums.h
elizagamedev: Oops, good catch

z3moon (approved):
LGTM, thanks for the change!

filament/backend/src/CompilerThreadPool.h
```diff
@@ -59,12 +59,11 @@ class CompilerThreadPool {
     bool mExitRequested{ false };
     utils::Mutex mQueueLock;
     utils::Condition mQueueCondition;
-    std::array<Queue, 2> mQueues;
+    std::array<Queue, 3> mQueues;
```
z3moon, line 62: nit: could you check CompilerThreadPool.cpp as well. There are a couple places to be updated.

elizagamedev: Done.

PR #8969, opened 2025-07-18
webgpu: Create command encoder and pass arrayLayerCount

jc3265:
BUGS = [432262069]


filament/backend/src/webgpu/WebGPUDriver.cpp
```diff
@@ -852,6 +852,10 @@ void WebGPUDriver::compilePrograms(CompilerPriorityQueue priority,
 
 void WebGPUDriver::beginRenderPass(Handle<HwRenderTarget> renderTargetHandle,
         RenderPassParams const& params) {
+    if (!mCommandEncoder) {
```
AndyHovingh, line 855: This is fine, but I think we should either _also_ do this nullptr check in makeCurrent before we create the encoder there or just remove creation there instead (only creating the command encoder here if it is not already created).

filament/backend/src/webgpu/WebGPUTexture.cpp
```diff
@@ -422,13 +422,15 @@ bool WebGPUTexture::supportsMultipleMipLevelsViaStorageBinding(const wgpu::Textu
     return storageBindingCompatibleFormatForViewFormat(format) != wgpu::TextureFormat::Undefined;
 }
 
-wgpu::TextureView WebGPUTexture::getOrMakeTextureView(const uint8_t mipLevel,
-        const uint32_t arrayLayer) {
+wgpu::TextureView WebGPUTexture::makeAttachmentTextureView(const uint8_t mipLevel,
+        const uint32_t arrayLayer, uint32_t layerCount) {
```
AndyHovingh: nit: `const uint32_t layerCount`

filament/backend/src/webgpu/WebGPUDriver.cpp
jc3265: good idea, i'll add the check

PR #8973, opened 2025-07-19
use a separate UBO for the postfx user passes

pixelflinger:
The per-view UBO is used in several different contexts, for instance:
- during the color pass
- during the shadow pass
- during the structure pass
- during app postfx passes

For shadows and other passes, its content is actually different because
the camera is very different. In that case, we use a different buffer.

However, the code that was used to set the values into the UBO was
duplicated and did slightly different things. In addition some
UBO values were (and are still) probably missing (e.g. the per-material
data).

To improve this situation we new use a utility class that does the
UBO set up of those "shared" values.

It turns out that the structure, ssr and ssao passes also need a 
slightly different content (because they run at half res) and currently
are sharing the color pass UBO, which causes mid-frame updates.

We introduce a new StructureDescriptorSet which is managing a new UBO,
it too is now implemented in terms of the new utility class. 
StructureDescriptorSet Is not used yet.

### DIFF
--- a/filament/src/ds/StructureDescriptorSet.cpp
+++ b/filament/src/ds/StructureDescriptorSet.cpp
@@ -24,6 +24,9 @@
 #include <private/filament/EngineEnums.h>
 #include <private/filament/UibStructs.h>
 
+#include <utils/compiler.h>
+#include <utils/debug.h>
+
 #include <math/vec4.h>
 
 #include <array>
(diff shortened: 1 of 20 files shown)


filament/src/ds/StructureDescriptorSet.cpp
```diff
@@ -0,0 +1,103 @@
...
+    mDescriptorSet.terminate(driver);
+    mUniforms.terminate(driver);
+}
+
+void StructureDescriptorSet::commit(DriverApi& driver) noexcept {
+    driver.updateBufferObject(mUniforms.getUboHandle(),
+            mUniforms.toBufferDescriptor(driver), 0);
+
+    assert_invariant(mDescriptorSetLayout);
+    if (mDescriptorSetLayout) {
```
poweifeng: nit: why do the conditional here?  Shouldn't we do something harsher (PRECONDITION assert) if the layout doesn't exisit?

pixelflinger: ah yes, I got annoyed by the IDE warning on this line. this situation cannot happen in practice; I should just remove the check.

PR #8975, opened 2025-07-21
the ssr-history texture is not an array

pixelflinger:
the descriptor used for SSR is different depending on the pass. In the color pass, it's used as the source of the SSRs buffers and is an array (refraction & reflection). In the SSR pass it's the history buffer and is just a 2D texture.


libs/filabridge/include/private/filament/EngineEnums.h
```diff
@@ -54,7 +54,8 @@ enum class PerViewBindingPoints : uint8_t  {
     IBL_DFG_LUT     =  7,   // user defined (128x128), RGB16F
     IBL_SPECULAR    =  8,   // user defined, user defined, CUBEMAP
     SSAO            =  9,   // variable, RGB8 {AO, [depth]}
-    SSR             = 10,   // variable, RGB_11_11_10, mipmapped
+    SSR             = 10,   // variable, 2d array, RGB_11_11_10, mipmapped
+    SSR_HISTORY     = 10,   // variable, 2d texture, RGB_11_11_10
```
kpiascik, line 58: should both `SSR` and `SSR_HISTORY` have the same value for the enum?

pixelflinger: yes definitely. they're both the same binding numbers but in different descriptor sets.

PR #8979, opened 2025-07-22
cleanup resgen code

pixelflinger:
A test for how Gemini can clean-up code. Changes:

- Configuration Struct: Global variables have been grouped into an  AppConfig struct. This makes the program's configuration explicit and  avoids polluting the global namespace.

- Clearer Argument Parsing: The handleArguments function now returns an  AppConfig object, encapsulating all parsing logic and making the main  function cleaner.

- Helper Functions: Repetitive tasks like string replacement, file I/O,  and writing file entries have been extracted into small, well-defined  helper functions (replaceAll, openOutputFile, readFile, etc.).

- Simplified main Function: The main function is now a high-level  coordinator, delegating work to helper functions. This makes the overall
 program flow much easier to follow.

- Modern C++ Idioms: The code now consistently uses C++ streams  (std::cout, std::ofstream) and std::string objects, which is more  idiomatic than a mix of C and C++ styles.

- added short comments

### DIFF
--- a/tools/resgen/src/main.cpp
+++ b/tools/resgen/src/main.cpp
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2018 The Android Open Source Project
+ * Copyright (C) 2025 The Android Open Source Project
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


tools/resgen/src/main.cpp
```diff
@@ -32,15 +35,22 @@
 using namespace std;
```
poweifeng: I think the previous way of prefixing with `std::` was actually better.  Even though there was an unnecessary `using namespace std`.  Maybe you can ask AI to undo just that part?

poweifeng (approved):
one nit, otherwise lgtm

tools/resgen/src/main.cpp
pixelflinger: yeah, lol, I was thinking the same thing. Let me try!

pixelflinger: done

PR #8981, opened 2025-07-22
add a unit test for Bitmap

pixelflinger:
- rename find(Value&) to findValue() to avoid issues if the key and values have the same type.

- fix a leak if Bitmap is destroyed and not empty (thankfully this didn't happen in practice in filament, because the Bimap was always empty when destroyed)

- fix a use after free in erase. we were destroying the Key before removing it from the maps, which theoretically could need to access them.

- slightly improved the API and added some comments about preconditions.


poweifeng (approved):
nit: replace Bitmap with Bimap in commit message

PR #8984, opened 2025-07-22
Tests: Create default render target before makeCurrent

rafadevai:


rafadevai:
@pixelflinger can you provide any guidance on this? 

is there a dependency between createDefaultRenderTarget, createSwapchain and makeCurrent API calls?

pixelflinger:
> @pixelflinger can you provide any guidance on this?
> 
> is there a dependency between createDefaultRenderTarget, createSwapchain and makeCurrent API calls?

`createDeaultRenderTarget()` at its name suggest is used to create a `Handle<HwRenderTarget>` that corresponds to the *currently bound* swap chain (view makeCurrent). This is a bit of "magic" handle that changes meaning when `makeCurrent` is called.

`createSwapchain` creates *a* swapchain (i.e. an object that can be made current and presented) from a native window. There can be several swapchains, if there are several windows.

`makeCurrent` associates the swapchain to the handle returned earlier by `createDeaultRenderTarget()`.

It doesn't make sense to need to have a swapchain created for `createDeaultRenderTarget()` to work, because there could be several swapchains.  It also doesn't make sense to have to have a `makeCurrent` call, because another `makeCurrent` could change it.

```
// returns a valid handle, but because nothing is current right now, it's not valid to use it as an attachment
Handle<HwRenderTarget> rt = createDefaultRenderTarget();

// create a couple swapchains
Handle<HwSwapChain> sc0 createSwapChain();
Handle<HwSwapChain> sc1 createSwapChain();

// make sc0 current, at this point rt can be used as an attachment and rendering will go into sc0
makeCurrent(sc0);

// make sc1 current, at this point rt can be used as an attachment and rendering will go into sc1
makeCurrent(sc1);
```

The sequence about should work.

filament/backend/test/test_Blit.cpp
```diff
@@ -498,11 +497,10 @@ TEST_F(BlitTest, BlitRegionToSwapChain) {
     constexpr int kNumLevels = 3;
 
     // Create a SwapChain and make it current.
+    Handle<HwRenderTarget> dstRenderTarget = mCleanup.add(api.createDefaultRenderTarget());
```
pixelflinger, line 500: This is fine. You're calling `createDefaultRenderTarget()` before `createSwapChain()` which is the important thing to test (i.e. that it is independent from the swapchain).

It seems to me the vk backend should be resilient to the order being different. It could *internally* create the default render target first and then just return that when `createDefaultRenderTarget()` is called.

I'm okay LGTM'ing this. But it seems trivial to fix it.

pixelflinger (approved):

filament/backend/test/test_Blit.cpp
rafadevai: Some of the tests, were creating the defaultRenderTarget after makeCurrent, which caused a crash on VK.

Thanks for the clarification

rafadevai:
Updated the PR with other changes from the 25 tests, 12 still fail 
Now we can do a full run of the test without any crashing.

PR #8986, opened 2025-07-22
materials: reintroduce mutable spec constants

elizagamedev:
This change was reverted because it broke stereoscopic rendering on Linux, but I can no longer reproduce that breakage. See also #8795.


filament/src/details/Material.h
```diff
@@ -122,36 +122,40 @@ class FMaterial : public Material {
...
     // prepareProgram creates the program for the material's given variant at the backend level.
     // Must be called outside of backend render pass.
     // Must be called before getProgram() below.
     void prepareProgram(Variant const variant,
+            backend::Program::MutableSpecConstantsInfo const mutableSpecConstants,
             backend::CompilerPriorityQueue const priorityQueue = CompilerPriorityQueue::HIGH) const noexcept {
         // prepareProgram() is called for each RenderPrimitive in the scene, so it must be efficient.
-        if (UTILS_UNLIKELY(!isCached(variant))) {
-            prepareProgramSlow(variant, priorityQueue);
+        if (UTILS_UNLIKELY(!isCached(variant, mutableSpecConstants))) {
```
pixelflinger, line 149: in the original code `isCached` was inlined so that we don't do a function call in the common case (This is called in the most inner loop).  Has `isCached` become too big to be inline?

elizagamedev: I think it's fine. I re-inlined it.

z3moon:
At that time, I couldn't reproduce it either in my cloudtop. You see the stereoscopic rendering is broken when you try in a local machine with a physical GPU. Let me try with this branch again.

z3moon:
This still has the same issue, printing the same error.

```
OpenGL error 0x502 (GL_INVALID_OPERATION) in "draw2" at line 4202
```

elizagamedev:
That's strange, I can't reproduce on my Linux workstation at all. What's the exact build command line you are invoking?

elizagamedev:
Finally fixed this. I was copying to a `FixedCapacityVector` incorrectly, causing spec constant which controls the number of eyes to change from its specialized value to its default of 2, breaking everything.

elizagamedev:
Closing this since we need to go back to the drawing board a bit given the feature requests we've received on this subject.

PR #8988, opened 2025-07-22
PPM: add material instance caching

poweifeng:
We add a class to manage a cache of material instances mapped by materials. This allows us to re-use instances across frames. Note hat also it's possible for the same frame to want to use the same instance multiple times (for example for bilateral blurring or upsampling), but we allocate separate instances for that case so that UBOs are not shared across render passes. This allows for bubble-reduction for vulkan (and possibly other backends).


filament/src/PostProcessManager.h
```diff
@@ -393,7 +384,60 @@ class PostProcessManager {
...
 
+    void resetForRender();
+
 private:
+    // This class manages a cache of material instances mapped by materials.  This allows us to
+    // re-use instances across frames. Note that also it's possible for the same frame to want to
+    // use the same instance multiple times (for example for bilateral blurring or upsampling), but
+    // we allocate separate instances for that case so that UBOs are not shared across render
+    // passes. This allows for bubble-reduction for vulkan (and possibly other backends).
+    class MaterialInstanceManager {
```
pixelflinger: Could we move this into its own file? And just keep the helper for PostProcessMaterial into PostProcessManager (so that MaterialInstanceManager doesn't need to depend on PostProcessMateral).
I think this class is 99% generic, we might as well make it 100%.

filament/src/PostProcessManager.cpp
```diff
@@ -590,7 +636,7 @@ PostProcessManager::StructurePassOutput PostProcessManager::structure(FrameGraph
                 FMaterial const* const ma = material.getMaterial(mEngine);
                 auto pipeline = getPipelineState(ma);
 
-                FMaterialInstance* const mi = PostProcessMaterial::getMaterialInstance(ma);
+                FMaterialInstance* const mi = mMaterialInstanceManager.getMaterialInstance(ma);
```
pixelflinger: For all these you could maybe have a helper on PostProcessManager so you don't have to repeat `mMaterialInstanceManager.` each time.

show50726:
Just out of curious, why is sharing a UBO across different render passes bad?

filament/src/PostProcessManager.h
poweifeng: done

filament/src/PostProcessManager.cpp

poweifeng:
> Just out of curious, why is sharing a UBO across different render passes bad?

Imagine one pass is reading from the UBO, and then the next pass wants to update a parameter in it before rendering.  This means that the accesses to that UBO needs to be synchronized across the passes, which can create a "bubble" (lack of overlap between vertex/fragment/upload) on the GPU. You'd see this explicitly in vulkan because you need to add a barrier in order to have the accesses synchronized. This work will help us not have no many barriers.

show50726:
> > Just out of curious, why is sharing a UBO across different render passes bad?
> 
> Imagine one pass is reading from the UBO, and then the next pass wants to update a parameter in it before rendering. This means that the accesses to that UBO needs to be synchronized across the passes, which can create a "bubble" (lack of overlap between vertex/fragment/upload) on the GPU. You'd see this explicitly in vulkan because you need to add a barrier in order to have the accesses synchronized. This work will help us not have no many barriers.

I see. Thanks for the clear explanation!

filament/src/MaterialInstanceManager.h
```diff
@@ -0,0 +1,84 @@
...
+
+    private:
+        FMaterial const* mMaterial = nullptr;
+        std::vector<FMaterialInstance*> mInstances;
+        utils::bitset32 mAvailable = {};
+
+        friend class MaterialInstanceManager;
+    };
+
+    ~MaterialInstanceManager() = default;
```
pixelflinger: Add a ctor and copy/move ctor and = operators (just to be clean). Implement them in the .ccp file to they're not inlined.

filament/src/MaterialInstanceManager.h
```diff
@@ -0,0 +1,84 @@
...
+    class Record {
+    public:
+        Record(FMaterial const* material)
+            : mMaterial(material) {}
+        ~Record() = default;
+
+        Record(Record const& rhs) noexcept = default;
+        Record& operator=(Record const& rhs) noexcept = default;
+        Record(Record&& rhs) noexcept = default;
+        Record& operator=(Record&& rhs) noexcept = default;
```
pixelflinger: all these ctor/dtors are inline. is it what you wanted? This will inline the std::vector construction/destruction. I think these should probably not be inline.

pixelflinger (changes requested):

filament/src/MaterialInstanceManager.h
poweifeng: moved to cpp

filament/src/MaterialInstanceManager.h
poweifeng: done

filament/src/MaterialInstanceManager.cpp
```diff
@@ -0,0 +1,89 @@
...
+
+FMaterialInstance* Record::getInstance() {
+    if (mAvailable.any()) {
+        int picked = -1;
+        mAvailable.forEachSetBit([&picked](size_t index) {
+            if (picked >= 0) {
+                return;
+            }
+            picked = index;
+        });
```
pixelflinger, line 36: it looks like here you're looking for the first bit set?  You could add a method in bitset to do this, it should be just one call to ctz, instead of a loop. Maybe as a follow-up.

poweifeng: yes, I can do that.  

I feel like I did the same elsewhere.  It should be a common pattern worthy of a method.

PR #8992, opened 2025-07-23
Better set default shader stages for samplers

bejado:
Now by default, we set the shader stages for samplers to **fragment only** unless:
- It's a post-process material
- There's a custom vertex shader

This helps avoid unnecessary dependencies between render passes.

### DIFF
--- a/filament/src/materials/skybox.mat
+++ b/filament/src/materials/skybox.mat
@@ -11,7 +11,8 @@ material {
         },
         {
            type : samplerCubemap,
-           name : skybox
+           name : skybox,
+           stages : [ "fragment" ]
         },
         {
            type : float4,
--- a/filament/src/materials/blitDepth.mat
+++ b/filament/src/materials/blitDepth.mat
@@ -19,7 +19,7 @@ material {
             target : depth,
             type : float
         }
-    ],    
+    ],
     variables : [
         vertex
     ],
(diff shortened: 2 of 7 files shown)


filament/src/materials/skybox.mat
```diff
@@ -11,7 +11,8 @@ material {
         },
         {
            type : samplerCubemap,
-           name : skybox
+           name : skybox,
+           stages : [ "fragment" ]
```
pixelflinger, line 15: Can you make a note somewhere that in theory we could inspect the vertex shader and figure this out automatically better if the stage is not specified? (maybe even file a bug).

pixelflinger (approved):

poweifeng (approved):
lgtm, but would echo Mathias' comment about maybe this can be further inferred (an item for a future fix-it or r&d day).

pixelflinger:
> Now by default, we set the shader stages for samplers to **fragment only** unless:
> 
> * It's a post-process material
> * There's a custom vertex shader
> 
> This helps avoid unnecessary dependencies between render passes.

This comment is misleading, postfx material always default to "fragment" regardless of which shaders are present.

filament/src/materials/skybox.mat
bejado: Updated the PR's description, sorry for the confusion.

Yes, in theory we could inspect the AST and see if materials are actually using the sampler. Will make a note and file a bug.

PR #8993, opened 2025-07-23
vk: better choice for semaphore wait stage

poweifeng:
Previously we always waited on the semaphore at
VK_PIPELINE_STAGE_ALL_COMMANDS_BIT. If we be a little bit more specific for the "in-frame" wait (between two command buffers within a frame), we can allow for better parallelism.

### DIFF
--- a/filament/backend/CMakeLists.txt
+++ b/filament/backend/CMakeLists.txt
@@ -567,6 +567,8 @@ if (APPLE OR LINUX)
         test/test_Scissor.cpp
         test/test_MipLevels.cpp
         test/test_Handles.cpp
+        test/test_CircularBuffer.cpp
+        test/test_CommandBufferQueue.cpp
     )
     set(BACKEND_TEST_LIBS
         absl::str_format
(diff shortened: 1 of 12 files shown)


filament/backend/src/vulkan/VulkanCommands.h
```diff
@@ -77,8 +77,9 @@ struct VulkanCommandBuffer {
 
     void reset() noexcept;
 
-    inline void insertWait(VkSemaphore sem) {
+    inline void insertWait(VkSemaphore sem, VkPipelineStageFlags waitStage) {
         mWaitSemaphores.push_back(sem);
+        mWaitSemaphoreStages.push_back(waitStage);
```
pixelflinger, line 82: wouldn't it make more sense  to have a single container and do something like `push_back({sem, stage})`?

poweifeng: but then I wouldn't be able to do this:
```
   VkSubmitInfo submitInfo{
        .sType = VK_STRUCTURE_TYPE_SUBMIT_INFO,
        .waitSemaphoreCount = mWaitSemaphores.size(),
        .pWaitSemaphores = mWaitSemaphores.data(),
        .pWaitDstStageMask = mWaitSemaphoreStages.data(),
        .commandBufferCount = 1u,
        .pCommandBuffers = &mBuffer,
        .signalSemaphoreCount = 1u,
        .pSignalSemaphores = &mSubmission,
    };
```
or I'd need to use SoA

pixelflinger: ah... makes sense.

pixelflinger (approved):

PR #8996, opened 2025-07-24
webgpu: Implement readPixels

bridgewaterrobbie:


filament/backend/src/webgpu/WebGPUSwapChain.cpp
```diff
@@ -373,6 +373,17 @@ wgpu::TextureView WebGPUSwapChain::getCurrentTextureView( wgpu::Extent2D const&
     return surfaceTexture.texture.CreateView(&textureViewDescriptor);
 }
 
+wgpu::Texture WebGPUSwapChain::getCurrentTexture(wgpu::Extent2D const& extent) {
+    if (isHeadless()) {
+        return mRenderTargetTextures[mHeadlessBufferIndex];
+    } else {
+        setExtent(extent);
+        wgpu::SurfaceTexture surfaceTexture;
+        mSurface.GetCurrentTexture(&surfaceTexture);
```
AndyHovingh: I am concerned about the implications of this, as this function really gets the next image in the swapchain, which is only something we want to do once per frame. We do this in WebGPU::makeCurrent right now. I would expect that we would essentially want to get at the texture for the current image in the swapchain anyway right (which is what is being done in the headless case above)?

filament/backend/src/webgpu/WebGPUTexture.cpp
```diff
@@ -324,6 +324,59 @@ namespace {
 
 } // namespace
 
+size_t WebGPUTexture::getWGPUTextureFormatPixelSize(wgpu::TextureFormat format) {
```
AndyHovingh: nit: `format` could be `const`

filament/backend/src/webgpu/WebGPUSwapChain.cpp
bridgewaterrobbie: I'm not sure I follow how this gets the next texture rather than the current one, is the function name misleading? Is there a way to confidently get the texture for what you say is the current one?

Admittedly this code path is as yet unused- getCurrentTexture is only called in the readPixels, and the only practical use of that we know of is headless. An "Unimplemented" abort might be possible if we don't know a good way to address this?

filament/backend/src/webgpu/WebGPUDriver.cpp
```diff
@@ -1110,8 +1110,125 @@ void WebGPUDriver::stopCapture(int) {
...
+    FILAMENT_CHECK_PRECONDITION(dstBytesPerPixel == srcBytesPerPixel && dstBytesPerPixel > 0)
+            << "Source texture pixel size (" << srcBytesPerPixel
+            << ") does not match destination pixel buffer pixel size (" << dstBytesPerPixel
+            << "), or the format is not supported for readPixels.";
+
+    // Create a staging buffer to copy the texture to. WebGPU requires 256 byte alignment
+    const size_t bytesPerPixel = dstBytesPerPixel;
+    const size_t unpaddedBytesPerRow = actualWidth * bytesPerPixel;
+    const size_t alignment = 256;
+    const size_t paddedBytesPerRow = (unpaddedBytesPerRow + alignment - 1) & ~(alignment - 1);
```
jc3265, line 1165: Prob the same result but isnt this the same as using 
`PixelBufferDescriptor::computeDataSize()` https://github.com/google/filament/blob/main/filament/backend/src/metal/MetalHandles.mm#L761

Edit: 
Actually nvm

filament/backend/src/webgpu/WebGPUDriver.cpp
```diff
@@ -1110,8 +1110,125 @@ void WebGPUDriver::stopCapture(int) {
...
+    assert_invariant(stagingBuffer);
+
+    wgpu::CommandEncoder encoder = mDevice.CreateCommandEncoder();
+    assert_invariant(encoder);
+
+    // WebGPU flips the y axis like Metal does
+    const uint32_t flippedY = srcHeight - y - height;
+    wgpu::TexelCopyTextureInfo source{ .texture = srcTexture,
+        .mipLevel = 0,
+        .origin = { x, flippedY, 0 } };
```
jc3265: im assuming for this we could also do `.origin = { .x = x, .y = flippedY, .z = 0 }`

filament/backend/src/webgpu/WebGPUDriver.cpp
```diff
@@ -1110,8 +1110,125 @@ void WebGPUDriver::stopCapture(int) {
...
+    mQueue.Submit(1, &commandBuffer);
+
+    // Map the buffer to read the data
+    struct UserData {
+        PixelBufferDescriptor pbd;
+        wgpu::Buffer buffer;
+        size_t unpaddedBytesPerRow;
+        size_t paddedBytesPerRow;
+        uint32_t height;
+        WebGPUDriver* that;
```
jc3265: whats this ? `that` is a bit confusing 😅

filament/backend/src/webgpu/WebGPUTexture.cpp
```diff
@@ -324,6 +324,59 @@ namespace {
...
+        case wgpu::TextureFormat::RGBA16Sint:
+        case wgpu::TextureFormat::RGBA16Float: return 8;
+        case wgpu::TextureFormat::RGBA32Float:
+        case wgpu::TextureFormat::RGBA32Uint:
+        case wgpu::TextureFormat::RGBA32Sint: return 16;
+        case wgpu::TextureFormat::Stencil8: return 1;
+        case wgpu::TextureFormat::Depth16Unorm: return 2;
+        case wgpu::TextureFormat::Depth24Plus: return 4;
+        case wgpu::TextureFormat::Depth24PlusStencil8: return 4;
+        case wgpu::TextureFormat::Depth32Float: return 4;
```
jc3265: why not group these guys and just do `return 4` once?
or why not move them to the other cases that return 4?

bridgewaterrobbie: That is a small improvement that can likely be made.

filament/backend/src/webgpu/WebGPUSwapChain.cpp
AndyHovingh: Yes, I would argue/agree that the WebGPU API naming here is misleading. See https://www.w3.org/TR/webgpu/#dom-gpucanvascontext-getcurrenttexture _"Get the GPUTexture that will be composited to the document by the GPUCanvasCOntext next"_ (the "next" word here being key).  "GetNextTexture" would make WAY more sense to me. Furthermore, looking at the interface to Surface, we just have Configure, GetCapabilities, GetCurrentTexture, Present, SetLabel, and Unconfigure. We need minimally need the ability to get the next swapchain image, which we use GetCurrentTexture. Another function for getting the in-flight swapchain image doesn't exist as far as I can see. For that, I suspect we need to just hold onto a handle in our own code.

I understand that technically this code path is unused, as we only need/want this for the headless case. Thus, an unimplemented abort/panic for this sounds good to me; no need to invest much time and energy on a path that may never be needed + it signals for us the use case when it is (and what that use case is, which is good for our learning).

filament/backend/src/webgpu/WebGPUSwapChain.h
```diff
@@ -42,6 +42,7 @@ class WebGPUSwapChain final : public Platform::SwapChain, HwSwapChain {
 
     [[nodiscard]] wgpu::TextureView getCurrentTextureView(wgpu::Extent2D const& extent);
```
AndyHovingh, line 43: Perhaps to help with the confusion on naming, we should rename this to `getNextTextureView`, remove `getCurrentHeadlessTextureView` below (I don't understand why we need that in addition to the function above it, as the caller doesn't need to care whether it is headless or not and the WebGPUSwapChain instance knows this). 

Then the meaning of `getCurrentTexture` below is more clear. Otherwise, at a glance, I would think this is the same as the currently existing `getCurrentTextureView` above, just for the underlying texture instead of the view.

filament/backend/src/webgpu/WebGPUTexture.h
```diff
@@ -118,6 +118,8 @@ class WebGPUTexture : public HwTexture {
     [[nodiscard]] static bool supportsMultipleMipLevelsViaStorageBinding(
             wgpu::TextureFormat format);
 
+    [[nodiscard]] static size_t getWGPUTextureFormatPixelSize(wgpu::TextureFormat format);
```
AndyHovingh: Not that you should change this; this is fine. But,  I just wanted to point out that likely this function could be improved by being `constexpr`. That said, the benefits of that would not be realized (yet) by how it is being used in this PR (it would only be nice in a future refactor that may or many never happen), and, of course, the downside of having it constexpr is that it would have to be implemented here in the header vs. the corresponding implementation .cpp file (since the function would need to be shared outside the WebGPUTexture translation unit). I just bring it up as food for thought if you have not already thought through that.

filament/backend/src/webgpu/WebGPUDriver.cpp
AndyHovingh: nit: formatting. If we have a trailing comma on all attributes, especially the last one, clang-format as the project has been configured, allows for something like the following, which I would argue is more readable:
```cpp
... source{ 
        .texture = srcTexture,
        .mipLevel = 0,
        .origin = { .x = x, .y = flippedY, .z = 0 },
};
```

This applies to other designated initializers below, such as `destination`, `copySize`, and `userData`.

filament/backend/src/webgpu/WebGPUDriver.cpp
```diff
@@ -1110,8 +1110,125 @@ void WebGPUDriver::stopCapture(int) {
...
+    wgpu::Extent3D copySize{ .width = actualWidth,
+        .height = actualHeight,
+        .depthOrArrayLayers = 1 };
+    encoder.CopyTextureToBuffer(&source, &destination, &copySize);
+    wgpu::CommandBuffer commandBuffer = encoder.Finish();
+    mQueue.Submit(1, &commandBuffer);
+
+    // Map the buffer to read the data
+    struct UserData {
+        PixelBufferDescriptor pbd;
```
AndyHovingh: For readability and self-describing code we have been trying to avoid using names like this, `pbd`, even if that pattern is used elsewhere. Maybe `bufferDescriptor` or something?

filament/backend/src/webgpu/WebGPUDriver.cpp
AndyHovingh: perhaps "driver"?

filament/backend/src/webgpu/WebGPUDriver.cpp
```diff
@@ -1110,8 +1110,125 @@ void WebGPUDriver::stopCapture(int) {
...
+        WebGPUDriver* that;
+    };
+    auto userData = std::make_unique<UserData>(UserData{ .pbd = std::move(pixelBufferDescriptor),
+        .buffer = std::move(stagingBuffer),
+        .unpaddedBytesPerRow = unpaddedBytesPerRow,
+        .paddedBytesPerRow = paddedBytesPerRow,
+        .height = height,
+        .that = this });
+
+    userData->buffer.MapAsync(
```
AndyHovingh, line 1217: I wonder if this could/would lead to a race condition with the caller. Wouldn't the caller want to know somehow that the texture has finished being written before using it? I wonder if ultimately some kind of synchronization mechanism would make sense, such as a field on HwTexture or a callback in this function or something. I realize this is more of a question for the other Filament folks. What do you think @poweifeng ?

filament/backend/src/webgpu/WebGPUDriver.cpp
bridgewaterrobbie: I can name it driver. I gave it this name because to my mind, I'm just passing around the `this` pointer in nearby code, necessary for the lambda. Intuitively it feels odd to name the this pointer something like that, but that's just probably me being weird so I'll switch it.

filament/backend/src/webgpu/WebGPUSwapChain.h
bridgewaterrobbie: I did strongly consider combining the textureview funcs, the caveat is the headless does not need an extent and I wasn't implicitly sure if the caller's invocation of 
```
 wgpu::Extent2D extent = mPlatform.getSurfaceExtent(mNativeWindow);
         mTextureView = mSwapChain->getCurrentTextureView(extent);
```
had an issue on headless or not.

filament/backend/src/webgpu/WebGPUDriver.cpp
bridgewaterrobbie: I had wondered that too, but with no way to signal the caller I tried this and it did in fact work. My hope is that the caller detects the scheduled destruction or something to know it can pull from the buffer it passed in

filament/backend/src/webgpu/WebGPUSwapChain.h
AndyHovingh: The reason we pass `extent` to the swapchain like this is to handle resizing _(in the event that the extent has changed in the swapchain instance)_. I can see that we don't have a _(new/updated)_ extent in the headless case in the driver, because we don't have a window to query. Thus, I would suggest we overload the function, `getNextTextureView(wgpu::Extent2D)` with `getNextTextureView()`. We should be able to reuse logic in those.

AndyHovingh: Maybe this is why Jeremy had the separate functions and their names, to be clear when NOT passing the extent would be "OK". Although, I don't have a strong opinion about this; not passing the extend vs. doing so would imply the first would not handle resizing _(but this may not be obvious to someone reading the code later)_. I think it being headless or not is not relevant to the caller though; it is more about whether the swapchain images should potentially be resized. In that case, having a separate `updateExtent` function may make more sense _(only being invoked in the non-headless path)_, and _only_ having `getNextTextureView()`.

bridgewaterrobbie: I agree with you that a change here would be cleaner separation of concerns, it was just not in the scope for this specific change and I didn't want to risk breaking something I hadn't considered without time to test it and be certain.

filament/backend/src/webgpu/WebGPUDriver.cpp
```diff
@@ -1110,8 +1110,129 @@ void WebGPUDriver::stopCapture(int) {
...
+    }
+    const size_t dstBytesPerPixel = PixelBufferDescriptor::computePixelSize(
+            pixelBufferDescriptor.format, pixelBufferDescriptor.type);
+    const size_t srcBytesPerPixel =
+            WebGPUTexture::getWGPUTextureFormatPixelSize(srcTexture.GetFormat());
+
+    FILAMENT_CHECK_PRECONDITION(dstBytesPerPixel == srcBytesPerPixel && dstBytesPerPixel > 0)
+            << "Source texture pixel size (" << srcBytesPerPixel
+            << ") does not match destination pixel buffer pixel size (" << dstBytesPerPixel
+            << "), or the format is not supported for readPixels.";
```
poweifeng, line 1159: I think you can use the reshaper if the dst and src formats don' match

filament/backend/src/webgpu/WebGPUDriver.cpp
poweifeng: when the PBD goes out of scope, it triggers a callback provided during the PBD construction, which passes the ownership of the filled-in buffer to the caller. The callback is how the caller will know that readPixels is finished.  This is why we always `move` PBD, and it implies you need to keep it in scope until you've filled in buffer.

AndyHovingh:
@poweifeng thank you! I appreciate learning about how the caller knows when the pixels are available via the PBD callback!

filament/backend/src/webgpu/WebGPUDriver.cpp
bridgewaterrobbie: I expect that is true, with the caveat that to do it fully "Right" would take a large bit of enum mapping. Though a "Needs one more channel" heuristic could probably be good enough reason to call the reshaper, and would usually work.

That hadn't come up in any of the cases I could test though, so I'd rather leave it until a case to test against presents itself.

poweifeng: well, you could just look at the other backends.  It's not that much work really.

But this check will do nothing dst=RGBA8 and src=BGRA8 I think, which is not necessarily what you want.

PR #9002, opened 2025-07-25
gl: Expedite shader cancellation in THREAD_POOL mode

z3moon:
Previously, cancelling shader compilation tokens in THREAD_POOL mode would block the engine thread until GL resource loading completed. This led to unexpected application stalls when numerous shaders were simultaneously cancelled.

This change moves the token cancellation logic to the compiler thread. By marking tokens for asynchronous cancellation, we prevent unwanted main thread stalls, ensuring a smoother user experience during bulk shader cancellation.

BUGS=[430068776]


filament/backend/src/opengl/ShaderCompilerService.cpp
```diff
@@ -134,9 +135,31 @@ struct ShaderCompilerService::OpenGLProgramToken : ProgramToken {
...
+        if (gl.program) {
+            glDetachShader(gl.program, shader);
+        }
+        glDeleteShader(shader);
+        shader = 0;
+    }
+    if (gl.program) {
+        glDeleteProgram(gl.program);
+        gl.program = 0;
+    }
```
poweifeng: is it safe to delete the program when the token goes out of scope?  I understand if the compilation was cancelled, but if it succeeded, wouldn't you want to keep the program around?

filament/backend/src/opengl/ShaderCompilerService.cpp
```diff
@@ -134,9 +135,31 @@ struct ShaderCompilerService::OpenGLProgramToken : ProgramToken {
...
+
+    // Indicate whether this token was canceled. Used for THREAD_POOL mode.
+    std::atomic_bool cancelled = false;
 };
 
 ShaderCompilerService::OpenGLProgramToken::~OpenGLProgramToken() {
+    // Clean up OpenGL resources. Typically, shader objects are destroyed, and ownership of the
+    // created program is transferred to `OpenGLProgram` if everything goes well. However, these GL
+    // resources may persist if the token was cancelled. This ensures thorough cleanup for the case.
+    for (GLuint& shader: gl.shaders) {
```
pixelflinger: This probably works, but I'm not a fan of doing GL calls from the dtor of an object managed by a shared_pointer. It means we have no real control when this is called. It's currently not the case, but this could happen from a non-GL thread and this would be bad.  I would prefer to have the destroy be more controlled.

filament/backend/src/opengl/ShaderCompilerService.cpp
z3moon: Yes, the created program is taken in the `initialize` call when it's used. So this wouldn't hit in that case.

poweifeng: taken, but is the program deleted?

z3moon: no, the ownership of the program is transferred to OpenGLProgram class and used for rendering. And the transferred program instance is deleted when the OpenGLProgram instance is destroyed.

poweifeng: right, so it's different than here, where you're deleting the program. But the program (maybe still in a OpenGLProgram) might still be needed.

filament/backend/src/opengl/ShaderCompilerService.cpp
z3moon: What do you mean by "no control"? Also why is it bad to delete these resources in non-GL thread (I guess you meant the engine thread) ? Do you want these resources to be deleted in the compiler thread?

filament/backend/src/opengl/ShaderCompilerService.cpp
poweifeng: ah I see. you mean `gl.program` is set to 0 in initialize().  But is it?

z3moon: Correct, it's set to 0 in initialize().

poweifeng: I see. I just read it. thanks for the clarification.

filament/backend/src/opengl/ShaderCompilerService.cpp
```diff
@@ -134,9 +135,19 @@ struct ShaderCompilerService::OpenGLProgramToken : ProgramToken {
 
     // Indicate this program was created from the cache blob.
     bool retrievedFromBlobCache = false;
+
+    // The current state of token. Used for THREAD_POOL mode.
+    enum class State : uint8_t {
+        WAITING,
+        LOADING,
+        COMPLETED,
+        CANCELED,
```
pixelflinger: can you add a comment for each explaining what the state means and what transtitions are expected to happen?

filament/backend/src/opengl/ShaderCompilerService.cpp
```diff
@@ -268,6 +279,16 @@ ShaderCompilerService::program_token_t ShaderCompilerService::createProgram(
...
             mCompilerThreadPool.queue(priorityQueue, token,
                     [this, &gl, program = std::move(program), token]() mutable {
+                        // If the token was canceled, we exit early. If the cancellation occurs
+                        // after this check, the created GL resources will be properly cleaned up
+                        // later in the `tick`.
+                        OpenGLProgramToken::State tokenState = OpenGLProgramToken::State::WAITING;
+                        if (!token->state.compare_exchange_strong(tokenState,
+                                    OpenGLProgramToken::State::LOADING,
+                                    std::memory_order_relaxed)) {
+                            assert_invariant(tokenState == OpenGLProgramToken::State::CANCELED);
```
pixelflinger, line 304: Can you explain why COMPLETE is impossible/not expected here?

filament/backend/src/opengl/ShaderCompilerService.cpp
```diff
@@ -268,6 +279,16 @@ ShaderCompilerService::program_token_t ShaderCompilerService::createProgram(
         case Mode::THREAD_POOL: {
             mCompilerThreadPool.queue(priorityQueue, token,
                     [this, &gl, program = std::move(program), token]() mutable {
+                        // If the token was canceled, we exit early. If the cancellation occurs
+                        // after this check, the created GL resources will be properly cleaned up
+                        // later in the `tick`.
+                        OpenGLProgramToken::State tokenState = OpenGLProgramToken::State::WAITING;
```
pixelflinger, line 297: would it be reasonable to add an assert that `token->state != COMPLETED or LOADING` (I *think* that's right, right?).
can you add a comment explaining which state are possible for `token->state` at this point.

filament/backend/src/opengl/ShaderCompilerService.cpp
```diff
@@ -284,6 +305,11 @@ ShaderCompilerService::program_token_t ShaderCompilerService::createProgram(
                         // to unblock the engine thread as soon as the token is ready while
                         // performing an expensive caching operation still in the pool.
                         tryCachingProgram(mBlobCache, mDriver.mPlatform, token);
+                        // Updates the token's state. If the token is canceled while this function
+                        // executes, this update notifies `tick` that GL resource loading is
+                        // complete, allowing `tick` to proceed with resource destruction.
```
pixelflinger, line 325: so that means it was canceled, but it did complete, so it's marked as complete? correct?

filament/backend/src/opengl/ShaderCompilerService.cpp
```diff
@@ -367,6 +394,38 @@ void ShaderCompilerService::tick() {
     // we don't need to run executeTickOps() if we're using the thread-pool
     if (UTILS_UNLIKELY(mMode != Mode::THREAD_POOL)) {
         executeTickOps();
+    } else {
+        // TODO: refactor (or rename) `executeTickOps` so that it performs properly
+        // (including below) based on mMode.
+        for (auto it = mCanceledTokens.begin(); it != mCanceledTokens.end();) {
```
pixelflinger, line 415: is there any risk that `mCanceledTokens` can be modified during this loop? i.e. does access to `mCanceledTokens` need to be protected?

filament/backend/src/opengl/ShaderCompilerService.h
```diff
@@ -134,6 +135,8 @@ class ShaderCompilerService {
     using ContainerType = std::tuple<CompilerPriorityQueue, program_token_t, Job>;
     std::vector<ContainerType> mRunAtNextTickOps;
 
+    std::list<program_token_t> mCanceledTokens;
```
pixelflinger, line 138: we don't expect this to be used very much right? This is only mostly for terminate()?  asking ebcause std::list is not great as it's doing memory allocations.

filament/backend/src/opengl/ShaderCompilerService.cpp
z3moon: done

filament/backend/src/opengl/ShaderCompilerService.cpp
z3moon: Yeah at this point, only `token->state == WAITING` is possible. Let me add an assert_invariant here.

z3moon: I misread it, CANCELED is also possible. Let me check this as well along with explanation.

filament/backend/src/opengl/ShaderCompilerService.cpp
z3moon: done

filament/backend/src/opengl/ShaderCompilerService.cpp
z3moon: Correct, if we failed to early-cancel the token (which means the cancellation occurred after the state transition at the beginning above) the job should proceed until completion so that GL reousrces can be gracefully destroyed later in the tick.

filament/backend/src/opengl/ShaderCompilerService.cpp
z3moon: There is no risk. The `mCanceledTokens` will shrink during the loop. It checks the token's state and, if the state refers to COMPLETED, put the token into the `mCompilerThreadPool` again for GL resource destruction job this time at a lower priority.

filament/backend/src/opengl/ShaderCompilerService.h
z3moon: I selected std::list for this container to simplify random element removal. The order in which tokens within this container complete is non-deterministic.

I expect this container to remain relatively small, as the most tokens are successfully early-canceled and therefore aren't added to this list in the first place.

PR #9003, opened 2025-07-25
webgpu: flush before reading pixels

poweifeng:
- Implement finish() to block until GPU work is done.
 - Implement flush() has the old finish() logic
 - Add flush() before reading pixels


filament/backend/src/webgpu/WebGPUDriver.cpp
```diff
@@ -1148,6 +1168,11 @@ void WebGPUDriver::readPixels(Handle<HwRenderTarget> sourceRenderTargetHandle, c
         scheduleDestroy(std::move(pixelBufferDescriptor));
         return;
     }
+
+    // Once we're ready to read the pixels, we need to flush all the previous work of this frame.
+    // This ensures that the readPixels will be ordered after all the draws.
+    flush();
```
pixelflinger, line 1174: it's weird to me that this is needed.

pixelflinger (approved):

filament/backend/src/webgpu/WebGPUDriver.cpp
poweifeng: why is that?

jc3265 (approved):
Lgtm

PR #9007, opened 2025-07-28
remove the explicit command queue flushes

pixelflinger:
Filament used to issue command queue flushes (E.g. glFlush()) after large passes like the color pass in order to kick the gpu into action sooner and reduce latency.  This however comes at a high cost on the gpu timeline as command queue flushes are expensive on vulkan and  metal. Moreover GL drivers are allowed to ignore glFlush().

So we just remove all these calls.

### DIFF
--- a/filament/src/RendererUtils.cpp
+++ b/filament/src/RendererUtils.cpp
@@ -254,11 +254,6 @@ RendererUtils::ColorPassOutput RendererUtils::colorPass(
                 driver.beginRenderPass(out.target, out.params);
                 passExecutor.execute(engine, driver);
                 driver.endRenderPass();
-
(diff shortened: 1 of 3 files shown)


rafadevai:
One question I was wondering with this change, if the flush is removed, does that mean some barriers needs to be added to synchronize the work between flushes?

pixelflinger:
> One question I was wondering with this change, if the flush is removed, does that mean some barriers needs to be added to synchronize the work between flushes?

it really should not. the flushes were not here for synchronization or ordering reasons.

poweifeng:
> One question I was wondering with this change, if the flush is removed, does that mean some barriers needs to be added to synchronize the work between flushes?

The barriers in place should be sufficient I think.

PR #9009, opened 2025-07-29
Fix incorrect buffer size and memory free issue with prefilter mipmap

bejado:


pixelflinger (approved):
What problem does this fix? is there a bug associated? If so can we update the PR?

show50726 (approved):
The fix looks good to me, but I'm also uncertain about why we should allocate `image` on the heap

bejado:
> What problem does this fix? is there a bug associated? If so can we update the PR?

This was for a failing G3 test

bejado:
`image` must be allocated on the heap so that it outlives ` texture->setImage` (which internally will queue up a data transfer operation on the backend thread).

PR #9010, opened 2025-07-29
webgpu: Return the swapchain texture

jc3265:
BUGS = [434731091]


filament/backend/src/webgpu/WebGPUDriver.cpp
```diff
@@ -1039,12 +1039,9 @@ void WebGPUDriver::makeCurrent(Handle<HwSwapChain> drawSch, Handle<HwSwapChain>
...
     mSwapChain = swapChain;
     assert_invariant(mSwapChain);
-    if(!mSwapChain->isHeadless()){
-         wgpu::Extent2D extent = mPlatform.getSurfaceExtent(mNativeWindow);
-         mTextureView = mSwapChain->getCurrentTextureView(extent);
-    } else {
-        mTextureView = mSwapChain->getCurrentHeadlessTextureView();
-    }
+
+     wgpu::Extent2D extent = mPlatform.getSurfaceExtent(mNativeWindow);
```
AndyHovingh: isn't a bit of a waste to even query the platform for the current surface extent in headless mode? I see that this simplifies things and maybe I am discussing premature optimization, but maybe this is even a problem in the headless context?

jc3265: should we have like an mExtent?

AndyHovingh: In the headless case there is no need for the extent argument here because we never resize the screen, as far as I can tell. For the non-headless case, we know what the previous extent was from mConfig (has width and height). That's how we know we need to resize/reconfigure. It is just not relevant in the headless case.

jc3265: yeah sorry i see what you mean i added the check to `is headless`  but if needed i can restore the `headless` function to separate the two of them

filament/backend/src/webgpu/WebGPUDriver.cpp
```diff
@@ -1039,12 +1039,12 @@ void WebGPUDriver::makeCurrent(Handle<HwSwapChain> drawSch, Handle<HwSwapChain>
...
-         wgpu::Extent2D extent = mPlatform.getSurfaceExtent(mNativeWindow);
-         mTextureView = mSwapChain->getCurrentTextureView(extent);
-    } else {
-        mTextureView = mSwapChain->getCurrentHeadlessTextureView();
+
+    wgpu::Extent2D extent{};
+    if (!mSwapChain->isHeadless()) {
+        extent = mPlatform.getSurfaceExtent(mNativeWindow);
     }
+    mTextureView = mSwapChain->getNextTextureView(extent);
```
AndyHovingh: This is confusing as well. Why would we NEED to send an extent at this point. Now, we are sending an empty (0,0) extent. Seems like at this point, we should either have an overload of getNextTextureView that does not take an extent. I think the reason there was a separate function before was to make clear when NOT sending an extent was OK. The key point has to do with resizing.

A suggestion:
```cpp
if (mSwapChain->isHeadless()) {
 mTextureView = mSwapChain->getNextTextureView();
} else {
  mTextureView = mSwapChain->getNextTextureView(mPlatform.getSurfaceExtent(mNativeWindow));
}
```
and/or maybe we do something with the naming of the functions for clarity (which was done before).

jc3265: okay, i just added a function that doesnt take an arg and added an assert

AndyHovingh: Good call on adding the assert! That addresses the issue of when it is OK to call that overload of the function 👍  I didn't think of that simple and elegant solution 🤦‍♂️ ! Nice @jc3265 !

PR #9011, opened 2025-07-29
WIP: remove the setDebugTag API in favor of tagged create API

bejado:


pixelflinger (approved):
Are we ever freeing the tags? I don't remember? I think we were not because we need them for use-after-free debugging?

PR #9012, opened 2025-07-30
wgpu: support blit

AndyHovingh:
This fixes the -i IBL issue and the image viewer. However, more work is needed (subsequent PR) to add robustness for us in the  WebGPUDriver::blit function, such as handling depth formats.

But this is probably a good place to integrate with main before proceeding with next steps.

BUGS=[434650196, 432815114]

### DIFF
--- a/filament/backend/src/webgpu/WebGPUDriver.cpp
+++ b/filament/backend/src/webgpu/WebGPUDriver.cpp
@@ -28,6 +28,7 @@
 #include "WebGPUStrings.h"
 #include "WebGPUSwapChain.h"
 #include "WebGPUTexture.h"
+#include "WebGPUTextureHelpers.h"
 #include "WebGPUVertexBuffer.h"
 #include "WebGPUVertexBufferInfo.h"
 #include <backend/platforms/WebGPUPlatform.h>
@@ -79,8 +80,9 @@ WebGPUDriver::WebGPUDriver(WebGPUPlatform& platform,
       mPipelineCache{ mDevice },
       mRenderPassMipmapGenerator{ mDevice },
       mSpdComputePassMipmapGenerator{ mDevice },
+      mBlitter{ mDevice },
       mHandleAllocator{ "Handles", driverConfig.handleArenaSize,
-          driverConfig.disableHandleUseAfterFreeCheck, driverConfig.disableHeapHandleTags } {
+          driverConfig.disableHandleUseAfterFreeCheck, driverConfig.disableHeapHandleTags }{
     mDevice.GetLimits(&mDeviceLimits);
 }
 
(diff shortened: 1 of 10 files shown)


filament/backend/src/webgpu/WebGPUDriver.cpp
```diff
@@ -741,21 +743,65 @@ void WebGPUDriver::update3DImage(Handle<HwTexture> textureHandle, const uint32_t
...
-    auto texture = handleCast<WebGPUTexture>(textureHandle);
+    const auto texture{ handleCast<WebGPUTexture>(textureHandle) };
+    FILAMENT_CHECK_PRECONDITION((texture->getTexture().GetWidth() + xoffset) >= width)
+            << "Blitting requires the destination region to have enough width ("
+            << texture->getTexture().GetWidth() << " to accommodate the requested " << width
+            << " with to write/blit (accounting for xoffset, which is " << xoffset << ").";
+    FILAMENT_CHECK_PRECONDITION((texture->getTexture().GetHeight() + yoffset) >= height)
+            << "Blitting requires the destination region to have enough height ("
+            << texture->getTexture().GetHeight() << " to accommodate the requested " << height
+            << " with to write/blit (accounting for yoffset, which is " << yoffset << ").";
```
jc3265: should this be `Blitting requires the destination region to have enough height (N) to accommodate the requested N height to write/blit...`

```suggestion
            << " height to write/blit (accounting for yoffset, which is " << yoffset << ").";
```

filament/backend/src/webgpu/WebGPUDriver.cpp
```diff
@@ -741,21 +743,65 @@ void WebGPUDriver::update3DImage(Handle<HwTexture> textureHandle, const uint32_t
...
     if (reshape(pixelBufferDescriptor, reshapedData)) {
-        data = &reshapedData;
+        inputData = &reshapedData;
     }
-    auto texture = handleCast<WebGPUTexture>(textureHandle);
+    const auto texture{ handleCast<WebGPUTexture>(textureHandle) };
+    FILAMENT_CHECK_PRECONDITION((texture->getTexture().GetWidth() + xoffset) >= width)
+            << "Blitting requires the destination region to have enough width ("
+            << texture->getTexture().GetWidth() << " to accommodate the requested " << width
+            << " with to write/blit (accounting for xoffset, which is " << xoffset << ").";
```
jc3265: same comment as below, should this be width? `width to write/blit `

jc3265 (approved):
Just the two questions/comments on the logs

filament/backend/src/webgpu/WebGPUDriver.cpp
AndyHovingh: Good call. I changed the wording as such 👍

filament/backend/src/webgpu/WebGPUDriver.cpp
AndyHovingh: yes

PR #9014, opened 2025-07-30
gl: Cleanup code flow in ShaderCompilerService

z3moon:


filament/backend/src/opengl/ShaderCompilerService.cpp
```diff
@@ -478,18 +446,17 @@ GLuint ShaderCompilerService::initialize(program_token_t& token) {
     FILAMENT_CHECK_POSTCONDITION(linked)
             << "OpenGL program " << token->name.c_str_safe() << " failed to link or compile";
 
-    // The program is successfully created. Try caching the program blob. In the THREAD_POOL mode,
-    // caching is performed in the pool.
+    GLuint const program = token->gl.program;
```
poweifeng, line 449: nit: why move this before the conditional.  It doesn't seem like there's a dependency

poweifeng (approved):

pixelflinger (approved):

filament/backend/src/opengl/ShaderCompilerService.cpp
z3moon: Yeah this doesn't have any dependency. I just thought it'd be easier to read the below in this way.

PR #9020, opened 2025-07-31
Materials: add useDefaultDepthVariant parameter

bejado:
Materials with expensive vertex shaders (for example, ones that sample from a texture) may only perform these computations for consumption by the fragment shader. If users enable the `useDefaultDepthVariant`, then Filament knows it can skip the custom vertex shader for depth-only passes.


bejado:
Accidentally added copilot..

pixelflinger (approved):

PR #9021, opened 2025-07-31
webgpu: support depth in blit + copy when we can.

AndyHovingh:
(in draft because of the handle allocator issue blocking manually testing. but that is happening on main as well, so I don't that is related to this change)

### DIFF
--- a/filament/backend/src/webgpu/WebGPUBlitter.cpp
+++ b/filament/backend/src/webgpu/WebGPUBlitter.cpp
@@ -18,6 +18,7 @@
 
 #include "WebGPUConstants.h"
 #include "WebGPUStrings.h"
+#include "WebGPUTextureHelpers.h"
 #include "webgpu/utils/StringPlaceholderTemplateProcessor.h"
 
 #include <backend/DriverEnums.h>
@@ -36,6 +37,32 @@
 namespace filament::backend {
 
 namespace {
+
+/**
+ * @return true if copyToTextureToTexture validation would succeed and we could use that for the
+ * "blit"
+ * ( see https://www.w3.org/TR/webgpu/#dom-gpucommandencoder-copytexturetotexture )
+ * NOTE: this does not validate usage of the source and/or destination.
+ *       If canDoDirectCopy(...) == true and the textures don't have the appropriate usages
+ *       we should panic (we should not be doing a renderpass blit when we could have been
+ *       doing a direct copy)
+ */
+[[nodiscard]] constexpr bool canDoDirectCopy(const uint32_t sourceSampleCount,
+        const wgpu::TextureFormat sourceFormat, const wgpu::Origin2D sourceOrigin,
+        const wgpu::Extent2D sourceExtent, const uint32_t destinationSampleCount,
+        const wgpu::TextureFormat destinationFormat, const wgpu::Origin2D destinationOrigin,
+        const wgpu::Extent2D destinationExtent) {
+    const wgpu::Extent2D sourceBlockSize{ getBlockSize(sourceFormat) };
+    const wgpu::Extent2D destinationBlockSize{ getBlockSize(destinationFormat) };
+    return sourceSampleCount == destinationSampleCount &&
+           areCopyCompatible(sourceFormat, destinationFormat) &&
+           (sourceExtent.width == destinationExtent.width) &&
+           (sourceExtent.height == destinationExtent.height) &&
+           ((sourceOrigin.x & sourceBlockSize.width) == 0) &&
+           ((sourceOrigin.y & sourceBlockSize.height) == 0) &&
+           ((destinationOrigin.x & destinationBlockSize.width) == 0) &&
+           ((destinationOrigin.y & destinationBlockSize.height) == 0);
+}
 
 struct BlitFragmentShaderArgs final {
     uint32_t depthPlane{ 0 };
@@ -61,11 +88,18 @@ constexpr std::string_view FRAGMENT_SHADER_ENTRY_POINT{ "fragmentShaderMain" };
 // otherwise be present in the template:
 constexpr std::string_view PLACEHOLDER_PREFIX{ "{{" };
 constexpr std::string_view PLACEHOLDER_SUFFIX{ "}}" };
-// texture_2d<f32> or texture_multisampled_2d<f32> or texture_3d<f32>
+// texture_2d<f32> or texture_multisampled_2d<f32> or texture_3d<f32> or
+// texture_depth_2d or texture_depth_multisampled_2d
 constexpr std::string_view TEXTURE_TYPE_PLACEHOLDER{ "TEXTURE_TYPE" };
+// texture_2d<f32> or texture_depth_2d
+[[maybe_unused]] constexpr std::string_view TEXTURE_2D_TYPE_PLACEHOLDER { "TEXTURE_2D_TYPE" };
 // "" for no sampler, otherwise "@group(0) @binding(2) var sourceSampler: sampler;"
 constexpr std::string_view SAMPLER_DECLARATION_PLACEHOLDER{ "SAMPLER_DECLARATION" };
 constexpr std::string_view FRAGMENT_SHADER_SNIPPET_PLACEHOLDER{ "FRAGMENT_SHADER_SNIPPET" };
+// @location(0) vec4<f32> or @builtin(frag_depth) f32
+constexpr std::string_view FRAGMENT_RETURN_ATTRIBUTE_AND_TYPE_PLACEHOLDER{
+    "FRAGMENT_RETURN_ATTRIBUTE_AND_TYPE"
+};
 constexpr std::string_view SHADER_SOURCE_TEMPLATE{ R"(
     struct BlitFragmentShaderArgs {
         depthPlane:        u32,
@@ -95,7 +129,7 @@ constexpr std::string_view SHADER_SOURCE_TEMPLATE{ R"(
 
     fn getNormalized2dSourceTextureCoordinates(
             position: vec2<f32>,
-            sourceTexture: texture_2d<f32>) -> vec2<f32> {
+            sourceTexture: {{TEXTURE_2D_TYPE}}) -> vec2<f32> {
         let sourceDimensions: vec2<u32> = textureDimensions(sourceTexture);
         return normalize2dSourceTextureCoordinates(
             getUnnormalizedSourceTextureCoordinates(position),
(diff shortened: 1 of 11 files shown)


filament/backend/src/webgpu/WebGPUBlitter.h
```diff
@@ -47,45 +48,52 @@ class WebGPUBlitter final {
...
+     * ONLY should be called if canDoDirectCopy(...) is true (see it in WebGPUBlitter.cpp)
+     */
+    void copy(wgpu::CommandEncoder const&, BlitArgs const&);
+
     void createSampler(SamplerMagFilter);
 
     [[nodiscard]] wgpu::RenderPipeline const& getOrCreateRenderPipeline(SamplerMagFilter,
             wgpu::TextureViewDimension sourceDimension, uint32_t sourceSampleCount,
-            wgpu::TextureFormat destinationTextureFormat);
+            bool sourceDepth, wgpu::TextureFormat destinationTextureFormat);
```
jc3265: sometimes we use `sourceDepth` and sometimes `depthSource` i think `depthSource` makes more sense

filament/backend/src/webgpu/WebGPUDriver.cpp
```diff
@@ -1424,43 +1424,47 @@ void WebGPUDriver::blit(Handle<HwTexture> destinationTextureHandle, const uint8_
         const uint8_t destinationLayer, const math::uint2 sourceOrigin, const math::uint2 size) {
     // TODO uncomment when texture format is taken into account with sampler settings in the
```
jc3265: cleanup

filament/backend/src/webgpu/WebGPUDriver.cpp
```diff
@@ -1424,43 +1424,47 @@ void WebGPUDriver::blit(Handle<HwTexture> destinationTextureHandle, const uint8_
...
+            .aspect = destinationTexture->getAspect(),
+            .origin = {.x = destinationOrigin.x, .y=destinationOrigin.y},
+            .extent = {.width=size.x, .height =size.y},
+            .mipLevel = destinationLevel,
+            .layerOrDepth = destinationLayer,
+        },
+        .filter = SamplerMagFilter::NEAREST,
+    };
+    mBlitter.blit(mQueue, mCommandEncoder, blitArgs);
+    if (!reusedCommandEncoder) {
```
jc3265, line 1490: why dont we `submit` in the case of a _reused_ command encoder?

filament/backend/src/webgpu/WebGPUDriver.cpp
```diff
@@ -768,34 +768,33 @@ void WebGPUDriver::update3DImage(Handle<HwTexture> textureHandle, const uint32_t
...
-                   << webGPUTextureFormatToString(texture->getViewFormat());
-        FWGPU_LOGD << "     Texture format (output):              "
-                   << webGPUTextureFormatToString(texture->getTexture().GetFormat());
-        FWGPU_LOGD << "     Linear Texture format (output):       "
-                   << webGPUTextureFormatToString(outputLinearFormat);
-        FWGPU_LOGD << "     Conversion Necessary:                 " << conversionNecessary;
-        FWGPU_LOGD << "     Do Blit:                              " << doBlit;
-    }
-#endif
+//#if FWGPU_ENABLED(FWGPU_DEBUG_VALIDATION)
```
jc3265: do we want to leave this since its under the `FWGPU_ENABLED(FWGPU_DEBUG_VALIDATION)` flag?

filament/backend/src/webgpu/WebGPUBlitter.cpp
```diff
@@ -37,6 +38,31 @@ namespace filament::backend {
 
 namespace {
 
+/**
+ * @return true of copyToTextureToTexture validation would succeed and we could use that for the
```
jc3265:
```suggestion
 * @return true if copyToTextureToTexture validation would succeed and we could use that for the
```

filament/backend/src/webgpu/WebGPUBlitter.cpp
```diff
@@ -139,9 +170,9 @@ constexpr std::string_view SHADER_SOURCE_TEMPLATE{ R"(
     {{FRAGMENT_SHADER_SNIPPET}}
 )" };
 
-constexpr std::string_view FRAGMENT_SHADER_SNIPPET_MSAA_INPUT{ R"(
+constexpr std::string_view FRAGMENT_SHADER_SNIPPET_MSAA_INPUT_TEMPLATE{ R"(
     @fragment
-    fn fragmentShaderMain(@builtin(position) position: vec4<f32>) -> @location(0) vec4<f32> {
+    fn fragmentShaderMain(@builtin(position) position: vec4<f32>) -> {{FRAGMENT_RETURN}} {
```
jc3265: i know what you mean, i wonder if something like `FRAGMENT_RETURN_TYPE` would be even clearer

filament/backend/src/webgpu/WebGPUBlitter.cpp
```diff
@@ -199,48 +230,63 @@ void WebGPUBlitter::blit(wgpu::Queue const& queue, wgpu::CommandEncoder const& c
...
-    FWGPU_LOGD << "      width: " << args.destination.texture.GetWidth();
-    FWGPU_LOGD << "    origin: x: " << args.destination.origin.x
-               << " y: " << args.destination.origin.y;
-    FWGPU_LOGD << "    extent: width: " << args.destination.extent.width
-               << " height: " << args.destination.extent.height;
-    FWGPU_LOGD << "    mipLevel: " << args.destination.mipLevel;
-    FWGPU_LOGD << "    layerOrDepth: " << args.destination.layerOrDepth;
-    FWGPU_LOGD << "  filter: " << (args.filter == SamplerMagFilter::NEAREST ? "NEAREST" : "LINEAR");
-#endif
+//#if FWGPU_ENABLED(FWGPU_DEBUG_VALIDATION)
```
jc3265: i really do believe that we should have a `printTexture(Texture)` function. Maybe in the `textureHelpers` or `webGPUStrings`?

filament/backend/src/webgpu/WebGPUBlitter.cpp
```diff
@@ -353,50 +399,94 @@ void WebGPUBlitter::blit(wgpu::Queue const& queue, wgpu::CommandEncoder const& c
...
+            .depthSlice = args.destination.texture.GetDimension() == wgpu::TextureDimension::e3D
+                                  ? args.destination.layerOrDepth
+                                  : wgpu::kDepthSliceUndefined,
+            // we don't need this resolveTarget because we are doing a resolve in software (the
+            // shader), but we probably should (leveraging a msaa sidecar texture) to take advantage
+            // of hardware acceleration. Thus, this might be an opportunity for optimization
+            .resolveTarget = nullptr,
+            .loadOp = wgpu::LoadOp::Clear,
+            .storeOp = wgpu::StoreOp::Store,
+            .clearValue = { .r = 0.0, .g = 0.0, .b = 0.0, .a = 1.0 },
```
jc3265, line 433: out of curiosity, why not pass `0` to alpha too?

filament/backend/src/webgpu/WebGPUBlitter.cpp
```diff
@@ -608,47 +716,84 @@ wgpu::BindGroupLayout WebGPUBlitter::createTextureBindGroupLayout(const SamplerM
...
+        } else {
+            if (sourceDimension == wgpu::TextureViewDimension::e3D) {
+                textureType = "texture_3d<f32>";
+            } else {
+                textureType = "texture_2d<f32>";
+            }
+        }
+    }
+    const std::string_view texture2dType{ depthSource ? "texture_depth_2d" : "texture_2d<f32>" };
+    // we don't declare a sample for the multisampled case
```
jc3265:
```suggestion
    // we don't declare a sampler for the multisampled case
```

filament/backend/src/webgpu/WebGPUBlitter.h
AndyHovingh: Looks like I generally agreed with you, as I used depthSource far more often. Good catch. I renamed them 👍

filament/backend/src/webgpu/WebGPUDriver.cpp
AndyHovingh: My thinking was that if a command encoder is already in flight it should get submitted elsewhere. Really, we should avoid superfluous submits for efficiency and lower bandwidth between the CPU and GPU. Honestly, we probably should not bother submitting even if we created the encoder here, as this will happen in commit, finish, and/or flush (either in this frame or the next). Even if it is the next frame, the blit was encoded/recorded prior to other things, so I think it is ok. But, I'm not sure. With the latest instability, I want to play it safe for now. We probably should remove these extra submits and verify everything else still works well, maybe once we have the test automation in place and tests passing. Then, we could get quick feedback that doing so doesn't break functionality, only improve performance 👍  What do you think?

filament/backend/src/webgpu/WebGPUDriver.cpp
AndyHovingh: If we do it REALLY spams the logs, even for our debugging case, which is why I took it out. So, I opted for the next thing, which is to allow for uncommenting when needed. Another option is to have another flag for this, such as FWGPU_DEBUG_UPDATE_IMAGE or something. I could do the same with blit. What do you think?

filament/backend/src/webgpu/WebGPUBlitter.cpp
AndyHovingh: I am all for more self-descriptive names, but `FRAGMENT_RETURN_TYPE` seems a _little_ misleading, because it is more than the type, it includes the annotation. But, maybe I am splitting hairs with that. Thinking of a little better name...

filament/backend/src/webgpu/WebGPUDriver.cpp
AndyHovingh: thank you! good catch :)

filament/backend/src/webgpu/WebGPUDriver.cpp
jc3265: Sounds like a plan!

filament/backend/src/webgpu/WebGPUBlitter.cpp
AndyHovingh: That's fine, especially since you feel strongly about it :) At the least, I could compile it conditionally behind a flag if we are worried at all with artifact size etc. So, I will add it.

filament/backend/src/webgpu/WebGPUDriver.cpp
jc3265: that would work!

filament/backend/src/webgpu/WebGPUBlitter.cpp
AndyHovingh: No particularly strong reason at all. I think it is more out of habit to visually indicate at a glance these are floating point numbers, not integers necessarily. I want to say in previous versions of C/C++, or maybe in another language, I ran into syntax errors, and so I have been more strict about it since. But, either way 🤷 .

filament/backend/src/webgpu/WebGPUDriver.cpp
AndyHovingh: 👍 ok. I am going to put some TODOs where we do this to consider removing them once we have test coverage and why we care.

filament/backend/src/webgpu/WebGPUBlitter.cpp
AndyHovingh: what about `FRAGMENT_RETURN_ATTRIBUTE_AND_TYPE`? A little long, but accurate.

jc3265: that works!

PR #9022, opened 2025-07-31
ppm: ensure a renderable descriptor set is bound

poweifeng:
- We just set a dummy set at the renderable bind point for post-processing passes, since its not used.
 - We also fix WebgpuDriver::bindDescriptorSet to properly "unbind" when receiving a null handle.

### DIFF
--- a/filament/backend/src/webgpu/WebGPUDriver.cpp
+++ b/filament/backend/src/webgpu/WebGPUDriver.cpp
@@ -136,6 +136,10 @@ void WebGPUDriver::setPresentationTime(int64_t monotonic_clock_ns) {
 void WebGPUDriver::endFrame(const uint32_t /* frameId */) {
     mPipelineLayoutCache.onFrameEnd();
     mPipelineCache.onFrameEnd();
+
+    for (size_t i = 0; i < MAX_DESCRIPTOR_SET_COUNT; i++) {
+        mCurrentDescriptorSets[i] = {};
+    }
 }
 
 void WebGPUDriver::flush(int) {
@@ -1480,8 +1484,8 @@ void WebGPUDriver::bindPipeline(PipelineState const& pipelineState) {
     assert_invariant(bindGroupLayouts.size() >= pipelineState.pipelineLayout.setLayout.size());
     size_t bindGroupLayoutCount{ 0 };
     for (size_t i{ 0 }; i < bindGroupLayouts.size(); i++) {
-        const auto handle{ pipelineState.pipelineLayout.setLayout[bindGroupLayoutCount] };
-        if (handle.getId() == HandleBase::nullid) {
+        const auto handle{ pipelineState.pipelineLayout.setLayout[i] };
+        if (!handle) {
             continue;
         }
         bindGroupLayouts[bindGroupLayoutCount++] =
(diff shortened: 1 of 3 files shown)


filament/src/PostProcessManager.h
```diff
@@ -413,6 +413,11 @@ class PostProcessManager {
     backend::VertexBufferInfoHandle mFullScreenQuadVbih;
     backend::DescriptorSetLayoutHandle mPerRenderableDslh;
 
+    // We need to have a dummy descriptor set because each post processing pass is expected to have
+    // a descriptor set bound at the renderable bind point. But the set itself contains dummy
+    // values.
```
pixelflinger, line 419: where does that constraint come from? Can we lift it for ppm?

jc3265 (approved):
lgtm

filament/src/PostProcessManager.cpp
```diff
@@ -484,6 +527,9 @@ void PostProcessManager::commitAndRenderFullScreenQuad(DriverApi& driver,
     driver.beginRenderPass(out.target, out.params);
     driver.draw(pipeline, mFullScreenQuadRph, 0, 3, 1);
     driver.endRenderPass();
+
+    driver.bindDescriptorSet({},
+            +DescriptorSetBindingPoints::PER_RENDERABLE , {});
```
pixelflinger: These can be replaced by:
`DescriptorSet::unbind(driver, DescriptorSetBindingPoints::PER_RENDERABLE);`

filament/src/PostProcessManager.cpp
```diff
@@ -476,6 +516,9 @@ void PostProcessManager::commitAndRenderFullScreenQuad(DriverApi& driver,
     FMaterial const* const ma = mi->getMaterial();
     PipelineState const pipeline = getPipelineState(ma, variant);
 
+    driver.bindDescriptorSet(mDummyPerRenderableDsh, +DescriptorSetBindingPoints::PER_RENDERABLE,
+            { { 0, 0 }, driver });
+
```
pixelflinger: I'm not sure this is best to do this here because some of these "render full screen quad" functions are called in a loop. It would be better to have the bind/unbind outside the loop.

filament/src/PostProcessManager.cpp
```diff
@@ -457,6 +491,9 @@ void PostProcessManager::renderFullScreenQuadWithScissor(
         PipelineState const& pipeline,
         backend::Viewport const scissor,
         DriverApi& driver) const noexcept {
+    driver.bindDescriptorSet(mDummyPerRenderableDsh, +DescriptorSetBindingPoints::PER_RENDERABLE,
+            { { 0, 0 }, driver });
```
pixelflinger: maybe make a helper on PostProcessManager for this.

filament/backend/src/webgpu/WebGPUDriver.cpp
```diff
@@ -136,6 +136,10 @@ void WebGPUDriver::setPresentationTime(int64_t monotonic_clock_ns) {
 void WebGPUDriver::endFrame(const uint32_t /* frameId */) {
     mPipelineLayoutCache.onFrameEnd();
     mPipelineCache.onFrameEnd();
+
```
pixelflinger, line 139: let's separate the backend changes from the frontend changes here (could be same PR, but at least different CLs).

filament/src/PostProcessManager.h
poweifeng: so the reason for this is that the previous PR unbound the color pass descriptor sets. So when we get to postprocessing passes, we have a pipeline associated with three descriptor set layouts where the second one is the renderable set layout, but we have nothing bound to the renderable bind point.

filament/src/PostProcessManager.cpp
poweifeng: I left out all the unbind code since I think your PR takes care of that.

filament/src/PostProcessManager.cpp
poweifeng: good point.  Moved to right below the per_view bind calls.

filament/src/PostProcessManager.cpp

filament/backend/src/webgpu/WebGPUDriver.cpp
poweifeng: done

pixelflinger (approved):
lgtm. but let's try to remove the PER_RENDERABLE descset entirely for postfx materials in a future PR.

PR #9026, opened 2025-08-01
utils: add precond/postcond macro guarded by feature flags

poweifeng:
We link the behavior of certain precondition/postcondition checks to feature flag states.

- If provided *flag* is true, then this macro will assert when the *condition* is false.
- If provided *flag* is fase, then this macro will output a warning when the condition is false.
- If the condition is true, then neither of the above will happen

These two macros will enable us to provide less restrictive behavior for certain correctness assertions, allowing clients to modify their before enabling these assertions by default.


libs/utils/include/utils/Panic.h
```diff
@@ -544,6 +545,27 @@ class TPanicStream : public PanicStream {
...
+            char const* condition)
+        : PanicStream(function, file, line, condition),
+          mEnablePanic(enable) {}
+    ~FlagGuardedStream() noexcept(false) {
+        if (mEnablePanic) {
+            PanicType::panic(mFunction, mFile, mLine, mLiteral, mStream.c_str());
+        } else {
+            LOG(WARNING) << "Failed [" << mLiteral << "]" << "\n  file=" << mFile
+                         << "\n  function=" << mFunction << "\n  line=" << mLine << "\n  info=\""
+                         << mStream.c_str() << "\"";
```
pixelflinger: This is not dependent on `PanicType` so there is no need to inline it. Basically I think the `else` should be a function call implemented in the CPP.

pixelflinger (changes requested):

libs/utils/include/utils/Panic.h
poweifeng: done

pixelflinger (approved):

PR #9029, opened 2025-08-04
Follow up of material parser PR

trie94:
- Fix CMakeLists to correctly include src files in matp test
- Replace std::string to std::string_view


libs/filament-matp/tests/test_matp.cpp
```diff
@@ -18,10 +18,12 @@
 
 #include "TestMaterialParser.h"
 
+#include <filamat/MaterialBuilder.h>
+
 #include <filament-matp/MaterialParser.h>
-#include "../src/JsonishLexer.h"
-#include "../src/JsonishParser.h"
-#include "../src/MaterialLexer.h"
+#include "JsonishLexer.h"
```
poweifeng, line 24: I think there's one more in `test_includer.cpp`

trie94: ah you are right. updated.

z3moon (approved):

PR #9032, opened 2025-08-05
webgpu: add visibility for debugging bind groups via logging (compile-time flag option)

AndyHovingh:
This becomes a fire hose of log spamming when switched on, but useful in correlating bind group items together for troubleshooting. If one switches it on, I recommend running it from the command line and piping stdout and stderr to a file, do the actions in the sample quickly, and analyze the log file afterward. This is because, the IDE's log output gets truncated and you lose information useful for correlation. And, due to frequency of logs, this truncation happens quickly.

BUGS = [435237421]

### DIFF
--- a/filament/backend/src/webgpu/WebGPUDriver.cpp
+++ b/filament/backend/src/webgpu/WebGPUDriver.cpp
@@ -16,6 +16,7 @@
 #include "webgpu/WebGPUDriver.h"
 
 #include "WebGPUBufferObject.h"
+#include "WebGPUConstants.h"
 #include "WebGPUDescriptorSet.h"
 #include "WebGPUDescriptorSetLayout.h"
 #include "WebGPUFence.h"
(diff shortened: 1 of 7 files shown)


filament/backend/src/webgpu/WebGPUDriver.cpp
```diff
@@ -1699,6 +1699,23 @@ void WebGPUDriver::updateDescriptorSetTexture(Handle<HwDescriptorSet> descriptor
             .binding = static_cast<uint32_t>(binding * 2 + 1),
             .sampler = sampler };
         bindGroup->addEntry(sEntry.binding, std::move(sEntry));
+#if FWGPU_ENABLED(FWGPU_DEBUG_BIND_GROUPS)
```
jc3265, line 1703: does the driver also need to include WebGPUConstants?

AndyHovingh: it should. looks like it wasn't added before. Adding it. 👍

PR #9034, opened 2025-08-05
bilateral blur & shadow map materials require unfilterable samplers

AndyHovingh:
This fixes the Display Shader Texture with the WebGPU implementation.

BUGS = [435237421]

### DIFF
--- a/filament/src/materials/ssao/bilateralBlur.mat
+++ b/filament/src/materials/ssao/bilateralBlur.mat
@@ -4,7 +4,8 @@ material {
         {
             type : sampler2dArray,
             name : ssao,
-            precision: medium
+            precision: medium,
+            filterable: false
         },
         {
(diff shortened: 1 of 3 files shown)


filament/src/materials/ssao/bilateralBlur.mat
```diff
@@ -4,7 +4,8 @@ material {
         {
             type : sampler2dArray,
             name : ssao,
-            precision: medium
+            precision: medium,
+            filterable: false
```
poweifeng, line 8: I think there's also this: filament/src/materials/ssao/bilateralBlurBentNormals.mat

AndyHovingh: I think you are right. Added that as well.

poweifeng (approved):

PR #9037, opened 2025-08-06
Add missing WebGPU checks to shader generation for tests

jc3265:
Test Shader generation was failing to return a valid shader for WebGPU
part of BUGS = [424157731]

### DIFF
--- a/filament/backend/test/SharedShaders.cpp
+++ b/filament/backend/test/SharedShaders.cpp
@@ -47,7 +47,7 @@ layout(location = 0) in vec4 mesh_position;
 void main() {
     gl_Position = vec4(mesh_position.xy, 0.0, 1.0);
 #if defined(TARGET_VULKAN_ENVIRONMENT)
-    // In Vulkan, clip space is Y-down. In OpenGL and Metal, clip space is Y-up.
+    // In Vulkan, clip space is Y-down. In OpenGL, Metal and WebGPU, clip space is Y-up.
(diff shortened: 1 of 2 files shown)


filament/backend/test/SharedShaders.cpp
```diff
@@ -46,7 +46,7 @@ layout(location = 0) in vec4 mesh_position;
 )", R"(
 void main() {
     gl_Position = vec4(mesh_position.xy, 0.0, 1.0);
-#if defined(TARGET_VULKAN_ENVIRONMENT)
+#if defined(TARGET_VULKAN_ENVIRONMENT) || defined(TARGET_WEBGPU_ENVIRONMENT)
```
AndyHovingh: Doesn't WebGPU align with Metal on clip space directions? I would think we shouldn't invert these, as Metal doesn't either.

filament/backend/test/SharedShaders.cpp
```diff
@@ -197,7 +197,7 @@ ShaderLanguage getShaderLanguage(const Backend& backend) {
         case Backend::METAL:
             return ShaderLanguage::MSL;
         case Backend::WEBGPU:
-            return ShaderLanguage::WGSL;
+//            return ShaderLanguage::WGSL;
```
AndyHovingh: But this should be returning WGSL. Why this change?

jc3265: yup it was just me testing

PR #9038, opened 2025-08-06
webgpu: Set Layercount to 1 instead of 0

jc3265:
part of BUGS = [424157731]

### DIFF
--- a/filament/backend/src/webgpu/WebGPURenderTarget.cpp
+++ b/filament/backend/src/webgpu/WebGPURenderTarget.cpp
@@ -125,7 +125,7 @@ WebGPURenderTarget::WebGPURenderTarget(const uint32_t width, const uint32_t heig
       mDefaultRenderTarget{ false },
       mTargetFlags{ targetFlags },
       mSamples{ samples },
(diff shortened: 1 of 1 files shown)


filament/backend/src/webgpu/WebGPURenderTarget.cpp
```diff
@@ -125,7 +125,7 @@ WebGPURenderTarget::WebGPURenderTarget(const uint32_t width, const uint32_t heig
       mDefaultRenderTarget{ false },
       mTargetFlags{ targetFlags },
       mSamples{ samples },
-      mLayerCount{ layerCount },
+      mLayerCount(layerCount == 0 ? 1 : layerCount),
```
AndyHovingh: this is side stepping one of the key advantages of using uniform initializer blocks, which prevents implicit narrowing conversions. Although, it really doesn't matter in this case with the literal value of 1, I would recommend for consistency: `mLayerCount{ layerCount <= 0 ? static_cast<uint8_t>(1) : layerCount }`. It is more verbose, but 🤷

pixelflinger (approved):

PR #9042, opened 2025-08-06
Add GEN_MIPMAPPABLE TextureUsage

poweifeng:
This will allow for more backend-specific usage mapping then before - where BLIT_SRC and BLIT_DST are assumed for textures whose mipmaps will be generated.

These mappings have been introduced in metal, vulkan, webgpu.


jc3265:
Should this change be added to `test_MipLevels` as well?

poweifeng:
> Should this change be added to `test_MipLevels` as well?

it doesn't seem like test is calling generateMipmaps?

filament/src/details/Texture.cpp
```diff
@@ -242,18 +242,30 @@ Texture* Texture::Builder::build(Engine& engine) {
...
+    auto const& featureFlags = downcast(engine).features.engine.debug;
+    if (mImpl->mLevels > 1 && (mImpl->mWidth > 1 || mImpl->mHeight > 1) && !mImpl->mExternal) {
+        bool const formatMipmappable =
+                downcast(engine).getDriverApi().isTextureFormatMipmappable(mImpl->mFormat);
+
+        FILAMENT_FLAG_GUARDED_CHECK_PRECONDITION(
+                formatMipmappable,
+                featureFlags.assert_texture_format_mipmappable)
+                << "Texture levels is > 1, but the format (" << int(mImpl->mFormat) << ") "
+                << " is not mipmppable";
```
pixelflinger: this check should be separate from "guessing" the usage bits.

filament/src/details/Texture.cpp
```diff
@@ -242,18 +242,30 @@ Texture* Texture::Builder::build(Engine& engine) {
...
+
+        if (formatMipmappable && !featureFlags.assert_texture_can_generate_mipmap) {
```
pixelflinger: if we want to do this, we should guard this whole block (From line 248) by the feature flag, because this whole code here is about "guessing" the usage flag to put. but in the future this should be explicit.

filament/src/details/Texture.cpp
```diff
@@ -242,18 +242,30 @@ Texture* Texture::Builder::build(Engine& engine) {
...
+            // Setting this for backwards compatibility,but should remove when safe.
+            mImpl->mUsage |= TextureUsage::GEN_MIPMAPPABLE;
         }
     }
 
+    if (any(mImpl->mUsage & TextureUsage::GEN_MIPMAPPABLE)) {
+        FILAMENT_CHECK_PRECONDITION(mImpl->mLevels > 1)
+                << "Texture has GEN_MIPMAPPABLE usage set, but levels is 1.";
+    }
```
pixelflinger: I think we shouldn't assert here, it's dumb to do this, but legal.

pixelflinger (changes requested):

filament/src/details/Texture.cpp
poweifeng: done

filament/src/details/Texture.cpp
poweifeng: done

filament/src/details/Texture.cpp
poweifeng: removed

jc3265:
> > Should this change be added to `test_MipLevels` as well?
> 
> it doesn't seem like test is calling generateMipmaps?

gotcha, I thought that https://github.com/google/filament/blob/main/filament/backend/test/test_MipLevels.cpp#L98
had to be updated to also include the new flag
but yes, doesnt seem like the test is calling the driver function

pixelflinger (approved):

PR #9044, opened 2025-08-06
webgpu: Refactor texture usage

jc3265:


filament/backend/src/webgpu/WebGPUTexture.cpp
```diff
@@ -121,47 +121,34 @@ namespace {
...
-    if (needsComputeStorageSupport || any(TextureUsage::SAMPLEABLE & fUsage)) {
-        retUsage |= wgpu::TextureUsage::TextureBinding;
-    }
-    if (needsComputeStorageSupport) {
-        retUsage |= wgpu::TextureUsage::StorageBinding;
-    }
-    if (needsRenderAttachmentSupport) {
-        retUsage |= wgpu::TextureUsage::RenderAttachment;
-    }
+    // Basing this mapping off of VulkanTexture.cpp's getUsage function
```
AndyHovingh: I don't think we need to mention that this is based on Vulkan's texture usage

filament/backend/src/webgpu/WebGPUTexture.cpp
```diff
@@ -121,47 +121,34 @@ namespace {
...
     if (any(TextureUsage::BLIT_SRC & fUsage)) {
+        retUsage |= wgpu::TextureUsage::CopySrc;
```
AndyHovingh: nit: for consistency and concision, perhaps:
`retUsage |= (wgpu::TextureUsage::CopySrc | wgpu::TextureUsage::TextureBinding);`

_(same comment for `if BLIT_DST`, `if needsComputeStorageSupport`, and `if (COLOR_ATTACHMENT || STENCIL_ATTACHMENT || DEPTH_ATTACHMENT`)_

filament/backend/src/webgpu/WebGPUTexture.cpp
```diff
@@ -121,47 +121,34 @@ namespace {
         const bool needsRenderAttachmentSupport) {
     wgpu::TextureUsage retUsage = wgpu::TextureUsage::None;
```
AndyHovingh: nit: uniform initialization, e.g. `wgpu::TextureUsage retUsage{ wgpu::TextureUsage::None };`

_(same comment for `transientAttachmentNeeded` and `useTransientAttachment`)_

filament/backend/src/webgpu/WebGPUTexture.cpp
```diff
@@ -121,47 +121,34 @@ namespace {
...
+        retUsage |= wgpu::TextureUsage::StorageBinding;
+        retUsage |= wgpu::TextureUsage::CopySrc;
+        retUsage |= wgpu::TextureUsage::CopyDst;
+        retUsage |= wgpu::TextureUsage::TextureBinding;
     }
 
-    // This is from Vulkan logic- if there are any issues try disabling this first, allows perf
-    // benefit though
+    wgpu::TextureUsage transientAttachmentNeeded = wgpu::TextureUsage::None;
     const bool useTransientAttachment =
```
AndyHovingh: We should add a parameter to this function, something like `const bool transientAttachmentsSupported` and use it here, e.g.:
```cpp
const bool useTransientAttachment {
        transientAttachmentsSupported &&
        ...
```

when calling `fToWGPUTextureUsage(...)` you can pass to that parameter `device.HasFeature(wgpu::FeatureName::TransientAttachments)`

jc3265: not sure if a parameter would make it more confusing. 
I updated the function call to 
```
mMipmapGenerationStrategy == MipmapGenerationStrategy::RENDER_PASS) &&
                      device.HasFeature(wgpu::FeatureName::TransientAttachments
```
thoughts?

AndyHovingh: Whether the TransientAttachments feature is supported or not has nothing to do with the mipmap generation strategy. I am suggesting that it needs to be part of our logic for deciding whether that usage flag ever gets enabled. We should never be using the wgpu::TextureUsage::TransientAttachment flag if the wgpu::FeatureName::TransientAttachments feature is not enabled on the graphics device.

jc3265: yes, sorry, I was confusing Transient with Render

filament/backend/src/webgpu/WebGPUTexture.cpp
```diff
@@ -114,55 +114,38 @@ namespace {
...
-                    TextureUsage::DEPTH_ATTACHMENT) &
-                fUsage)) {
-        retUsage |= wgpu::TextureUsage::RenderAttachment;
+    if (any(TextureUsage::SAMPLEABLE & fUsage)) {
+        retUsage |= wgpu::TextureUsage::TextureBinding;
+    }
+    // if needsComputeStorageSupport we need to read and write to the texture in a shader and thus
+    // require CopySrc, CopyDst, TextureBinding, & StorageBinding
+    if (needsComputeStorageSupport) {
+        retUsage |= (wgpu::TextureUsage::StorageBinding | wgpu::TextureUsage::CopySrc |
```
AndyHovingh: nit: for readability I would break from clang format a little here and keep each usage on a separate line, e.g.:
```cpp
retUsage |= (wgpu::TextureUsage::StorageBinding |
                      wgpu::TextureUsage::CopySrc |
                      wgpu::TextureUsage::CopyDst |
                      wgpu::TextureUsage::TextureBinding);
```

filament/backend/src/webgpu/WebGPUTexture.cpp
```diff
@@ -172,15 +155,38 @@ namespace {
             // restriction.
             // Note that the custom shader does not resolve stencil. We do need to move to vk 1.2
             // and above to be able to support stencil resolve (along with depth).
-            !(any(fUsage & TextureUsage::DEPTH_ATTACHMENT) && samples > 1);
-    if (useTransientAttachment) {
-        retUsage |= wgpu::TextureUsage::TransientAttachment;
+            !(any(fUsage & TextureUsage::DEPTH_ATTACHMENT) && samples > 1)};
+    if (useTransientAttachment && deviceSupportsTransientAttachments) {
```
AndyHovingh: I think `deviceSupportsTransientAttachments` needs to be part of how we set/compute the bool `useTransientAttachment`itself. Otherwise, we are making decisions about other things w.r.t. having a transient attachment when we don't... logical inconsistencies that can get very confusing and error prone.

jc3265: i dont think we should pass the device to this function. the other thing that i did is to move it to the texture helpers (Should prob also move the others in `namespace {}`

PR #9045, opened 2025-08-07
MetalBlitter: add support for blitting integer formats

bejado:
Fixes the Metal format issues seen in #8917.

### DIFF
--- a/filament/backend/CMakeLists.txt
+++ b/filament/backend/CMakeLists.txt
@@ -575,6 +575,12 @@ if (APPLE OR LINUX)
         test/test_CircularBuffer.cpp
         test/test_CommandBufferQueue.cpp
     )
+    if (APPLE)
+        # Metal-specific tests
+        list(APPEND BACKEND_TEST_SRC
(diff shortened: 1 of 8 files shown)


filament/backend/CMakeLists.txt
```diff
@@ -575,6 +575,12 @@ if (APPLE OR LINUX)
         test/test_CircularBuffer.cpp
         test/test_CommandBufferQueue.cpp
     )
+    if (APPLE)
+        # Metal-specific tests
+        list(APPEND BACKEND_TEST_SRC
+            test/test_MetalBlitter.mm
```
poweifeng, line 581: nit: maybe this can be `test/metal/test_MetalBlitter.mm` 
(parallel to `backend/src/metal`)

poweifeng (approved):

filament/backend/CMakeLists.txt
bejado: This is a good idea. Let me do this in a follow-up, as this requires a refactor of the backend tests CMake

PR #9047, opened 2025-08-07
add zstd to the build and use it for the DFG LUT

pixelflinger:


filament/src/zstd_helper.cpp
```diff
@@ -0,0 +1,42 @@
...
+#include "zstd_helper.h"
+
+#include <zstd.h>
+
+#include <cstddef>
+#include <cstdint>
+
+namespace filament {
+
+bool ZstdHelper::is_zstd_compressed(const void* src, size_t src_size) {
```
bejado: Nit: camelCase for these functions (and the file name)?

bejado (approved):

poweifeng (commented):
doesn't seem like zstd has a tnt folder with `README.md` how how it was imported?

It'd be nice to also know (from commit message) or something in the `tnt` folder what version of zstd or this is, or maybe just a commit hash.

poweifeng (approved):

filament/src/zstd_helper.cpp
pixelflinger: done

PR #9050, opened 2025-08-07
Fixes to frame_generator sample

patrikhuber:
The frame_generator sample was broken for me, and I found & fixes several issues:

1) `in.read(g_materialBuffer.data(), fileSize)` was always returning `false` for me, even though `in.is_open()` was successful. It turns out `std::ios::binary` is needed. It's possible that this is because I'm on Windows, whereas perhaps on other platforms `ifstream` reads in binary mode by default (or falls back to it).

2) `g_materialBuffer.reserve()` merely reserves memory, but does not actually change the size of the vector. We should actually allocate storage here and set the vector size appropriately, and not just reserve the memory.

3) `cleanup()` crashed with:
```
Precondition
in bool __cdecl filament::FEngine::destroy(const class filament::FMaterialInstance *):1243
in file filament-github-source\filament\src\details\Engine.cpp
reason: destroying MaterialInstance "LitOpaque" which is still in use by Renderable (entity=17, instance=11, index=0)
```
It looks like the material / renderables associated with the material are not destroyed in cleanup(), but they need to be destroyed before `g_materialInstance` can be destroyed.


patrikhuber:
CLA should now show as signed.

poweifeng (approved):

PR #9053, opened 2025-08-07
mat: Add CRC32 validation for material packages

z3moon:
This commit introduces a CRC32 checksum to material packages to ensure data integrity.

When a material is loaded, this checksum is verified. If the check fails, an error is logged, and the material fails to load. For older material packages without a CRC32 checksum, a warning is logged and proceed.

BUGS=[373396840]


pixelflinger (approved):

filament/src/details/Material.cpp
```diff
@@ -188,6 +188,34 @@ Material* Material::Builder::build(Engine& engine) const {
...
+            if (parsedCrc32 != expectedCrc32) {
+                CString name;
+                materialParser->getName(&name);
+                LOG(ERROR) << "The material '" << name.c_str_safe()
+                           << "' is corrupted: crc32_expected=" << expectedCrc32
+                           << ", crc32_parsed=" << parsedCrc32;
+                return nullptr;
+            }
+        } else {
+            // crc32 is not stored, probably due to an old version of material package.
```
poweifeng: Is this necessary?  I think we always assume that the current material version is the one being read here?

libs/filamat/src/MaterialBuilder.cpp
```diff
@@ -1374,10 +1374,25 @@ Package MaterialBuilder::build(JobSystem& jobSystem) {
         goto error;
     }
 
-    // Flatten all chunks in the container into a Package.
-    Package package(container.getSize());
+    // Flatten all container chunks into a single package and compute its CRC32 value, storing it as
+    // a separate chunk.
+    size_t originalContainerSize = container.getSize();
+    size_t crc32ChunkSize = sizeof(uint64_t) + sizeof(uint32_t) + sizeof(uint32_t);
```
poweifeng: can do `constexpr` here.

libs/utils/include/utils/Hash.h
```diff
@@ -20,12 +20,51 @@
...
 #include <stddef.h>
 
 namespace utils::hash {
 
+// The standard CRC-32 polynomial (used in Ethernet, ZIP, PNG, etc.)
+static constexpr uint32_t CRC32_POLYNOMIAL = 0xEDB88320;
+
+// Generates the lookup table for the CRC-32 algorithm.
+inline void crc32GenerateTable(std::vector<uint32_t>& table) {
+    if (table.size() != 256) {
```
poweifeng, line 35: this table seems like it could be statically generated?

libs/utils/include/utils/Hash.h
```diff
@@ -20,12 +20,51 @@
...
+    }
+}
+
+// Updates the CRC-32 value with a new chunk of data. This function can be called multiple times to
+// process a large buffer in smaller, separate parts.
+//  @param previous_crc The CRC value returned from the previous call to `crc32Update`.
+//  @param data A pointer to the data buffer for this chunk.
+//  @param length The length of the data buffer in bytes.
+//  @return The new CRC value.
+inline uint32_t crc32Update(uint32_t previous_crc, const void* data, size_t length, const std::vector<uint32_t>& table) {
```
poweifeng, line 58: does the table ever change?

filament/src/details/Material.cpp
bejado: This PR doesn't bump the material version (which maybe we want to do anyways), so it's possible user materials don't have the crc field.

z3moon: Yeah I think it's good to have, because we don't know how users maintain their material packages. They can use an old version of package files for some reasons.

poweifeng: does that mean you *don't* want to bump the material version for this?

libs/filamat/src/MaterialBuilder.cpp
z3moon: good point

libs/utils/include/utils/Hash.h
z3moon: It could, but it'll take up 1kb, which I want to avoid. We could revisit this later if this should be used more often.

libs/utils/include/utils/Hash.h
z3moon: No it won't change.

filament/src/details/Material.cpp
z3moon: This change itself doesn't require material version bump. But even after they upgrade the latest filament, this check is still valid because the package could be truncated.

z3moon: Actually I changed my mind. I'll bump up the material version and consolidate the error handling, which looks robust and simpler.

z3moon:
@bejado ah, good catch

bejado (approved):

libs/utils/include/utils/Hash.h
poweifeng: nit: if the parameter doesn't change, then maybe it should be within the function. But it's a minor thing.

z3moon: I see what you mean. I originally considered adding a fixed array but was uncertain about adding 1kb binary size only for this. We can revisit this later if needed.

poweifeng: you can store it as a static variable, of course, but then you'd have to deal with threading.

PR #9056, opened 2025-08-07
Fix: webgpu error regarding mismatched pipeline 

bejado:
Fixes an issue introduced by #9025

### DIFF
--- a/filament/src/details/Material.h
+++ b/filament/src/details/Material.h
@@ -96,6 +96,9 @@ public:
     DescriptorSetLayout const& getPerViewDescriptorSetLayout(
             Variant const variant, bool const useVsmDescriptorSetLayout) const noexcept;
 
+    // Returns the layout that should be used when this material is bound to the pipeline for the
+    // given variant. Shared variants use the Engine's default material's variants, so we should
+    // also use the default material's layout.
     DescriptorSetLayout const& getDescriptorSetLayout(Variant variant = {}) const noexcept {
         if (!isSharedVariant(variant)) {
             return mDescriptorSetLayout;
--- a/filament/src/details/Material.h
+++ b/filament/src/details/Material.h
@@ -96,8 +96,15 @@ public:
     DescriptorSetLayout const& getPerViewDescriptorSetLayout(
             Variant const variant, bool const useVsmDescriptorSetLayout) const noexcept;
 
-    DescriptorSetLayout const& getDescriptorSetLayout() const noexcept {
-        return mDescriptorSetLayout;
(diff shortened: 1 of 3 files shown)


poweifeng (approved):

filament/src/details/Material.h
```diff
@@ -96,8 +96,18 @@ class FMaterial : public Material {
     DescriptorSetLayout const& getPerViewDescriptorSetLayout(
             Variant const variant, bool const useVsmDescriptorSetLayout) const noexcept;
 
-    DescriptorSetLayout const& getDescriptorSetLayout() const noexcept {
-        return mDescriptorSetLayout;
+    // Returns the layout that should be used when this material is bound to the pipeline for the
+    // given variant. Shared variants use the Engine's default material's variants, so we should
+    // also use the default material's layout.
+    DescriptorSetLayout const& getDescriptorSetLayout(Variant variant = {}) const noexcept {
```
pixelflinger, line 102: should we even have a default parameter here? It seems safer to always pass the variant?

filament/src/details/Material.h
```diff
@@ -96,8 +96,18 @@ class FMaterial : public Material {
...
+        if (!isSharedVariant(variant)) {
+            return mDescriptorSetLayout;
+        }
+        FMaterial const* const pDefaultMaterial = mEngine.getDefaultMaterial();
+        if (UTILS_UNLIKELY(!pDefaultMaterial)) {
+            return mDescriptorSetLayout;
+        }
+        return pDefaultMaterial->getDescriptorSetLayout();
```
pixelflinger, line 110: this is ok for now, but I would like to revisit this to see if we can't do something much faster.

Ideally the code would look like : `if isShared() { return a; } else { return b }` and maybe `b` can be updated from some other place. or something.

filament/src/details/Material.h
bejado: Hm, I'm not sure what the best thing to do is here. For methods like `FMaterialInstance::setParameter`, we need the descriptor set layout, but only for validation. So in that case, we really want the "material's actual descriptor set" without taking into account shared variants. The default `Variant {}` here is a way to express that, but I agree it's not the safest.

In fact, there's only once place where we call `getDescriptorSetLayout` with a non-default argument (inside RenderPass). So maybe it's better to keep `getDescriptorSetLayout` as it is now, and add a new `getDescriptorSetLayoutForRendering` or something that takes the `Variant` and handles the shared variant logic.

WDYT?

bejado:
Merging to unblock the release.

PR #9064, opened 2025-08-08
webgpu: Update GLSL for mip test

jc3265:
part of BUGS [424157731]


poweifeng:
I think there's another one in `test_LoadImage.cpp`, would you mind doing the same replacement in there as well?

jc3265:
> I think there's another one in `test_LoadImage.cpp`, would you mind doing the same replacement in there as well?

Yes, will do this on a different PR along with other changes for `LoadImage`

PR #9065, opened 2025-08-08
vk: make default render target independent of call order

poweifeng:


filament/backend/src/vulkan/VulkanHandles.h
```diff
@@ -308,6 +308,17 @@ struct VulkanRenderTarget : private HwRenderTarget, fvkmemory::Resource {
...
+    VulkanRenderTarget(VulkanRenderTarget&& target) {
+        (*this) = std::move(target);
+    }
+
+    VulkanRenderTarget& operator=(VulkanRenderTarget&& target) {
+        std::swap(mOffscreen, target.mOffscreen);
+        std::swap(mProtected, target.mProtected);
+        std::swap(mInfo, target.mInfo);
+        return *this;
+    }
```
pixelflinger: I would prefer that we create a swap() method that both ctor&& and operator=&& call.

poweifeng: done

PR #9072, opened 2025-08-08
update glslang to version 15.4.0

pixelflinger:
add a nicer script to update


jc3265:
Would the script also work for a particular commit? 
Dawn points to individual commits and not particular releases. 
e.g. 
The latest Dawn Version [7350](https://dawn.googlesource.com/dawn/+/refs/heads/chromium/7350/third_party/glslang/src)
points to `eb77189a282b90e9856fc0ed5b08361a70025bec` And looking at `glslang`'s history. https://chromium.googlesource.com/external/github.com/KhronosGroup/glslang/+log
that commit is newer than `15.4.0`

pixelflinger:
> Would the script also work for a particular commit? Dawn points to individual commits and not particular releases. e.g. The latest Dawn Version [7350](https://dawn.googlesource.com/dawn/+/refs/heads/chromium/7350/third_party/glslang/src) points to `eb77189a282b90e9856fc0ed5b08361a70025bec` And looking at `glslang`'s history. https://chromium.googlesource.com/external/github.com/KhronosGroup/glslang/+log that commit is newer than `15.4.0`

I can make it work with commits, but the problem is that we can't really rely on a particular version, because internally google is using their own version (usually very up to date). So, bottom line is... we shouldn't sync to particular version.

PR #9073, opened 2025-08-08
update spirv-cross to d8e3e2b

pixelflinger:


jc3265:
Should we update the `Readme` and reflect the new commit https://github.com/google/filament/blob/main/third_party/spirv-cross/tnt/README.md?plain=1#L5 
would make it easier to track the version that we are using

jc3265:
> Should we update the `Readme` and reflect the new commit https://github.com/google/filament/blob/main/third_party/spirv-cross/tnt/README.md?plain=1#L5 would make it easier to track the version that we are using

nvm, just saw #9074 I guess it wont be needed

poweifeng (approved):

PR #9078, opened 2025-08-11
matc: improvements to LineDictionary compression

pixelflinger:
We can significantly improve the effectiveness of the line dictionary
by splitting lines strategically. spirv optimization tends to create
new variable names (e.g. _1234) that are not always stable between
variants, hereby preventing lines from being duplicated. By simply
splitting lines around these patterns, we improve the effectiveness
of the compression.

The decompression is almost unchanged. To make the code easier, we 
now keep the '\n' at the end of lines in the dictionary. That's the
only change to the decompression side.

On certain materials like aiDefault.mat the compression ratio is
improved from 2.8x to 15x.

This also fixes a small but where the line dictionary relied on the
last line of the input to be a newline.


bejado (approved):
Seems like a good candidate for some Gemini unit tests :)

PR #9084, opened 2025-08-11
github: add action for installing vulkansdk

poweifeng:


build/common/get-vulkan-sdk.sh
```diff
@@ -0,0 +1,119 @@
+# This script is from https://github.com/humbletim/install-vulkan-sdk
```
bejado, line 3: Nit: add shebang?

bejado (approved):

build/common/get-vulkan-sdk.sh
poweifeng: done

PR #9085, opened 2025-08-12
github: add script and action for getting gltf assets

poweifeng:


.github/actions/get-gltf-assets/action.yml
```diff
@@ -0,0 +1,25 @@
...
+runs:
+  using: "composite"
+  steps:
+    - uses: ./.github/actions/dep-versions  
+    - name: Cache glTF assets
+      id: cache-gltf
+      uses: actions/cache@v4
+      with:
+        path: gltf
+        key: ${{ runner.os }}-gltf-assets-${{ env.GITHUB_GLTF_SAMPLE_ASSETS_COMMIT }}
```
bejado: Does `runner.os` need to be part of the cache key? Seems like the assets are the same regardless of OS.

poweifeng: good point.  removed.

PR #9088, opened 2025-08-12
Rename shader std var

jc3265:
Running one of the samples resulted in an error due to `std` being a reserved word


pixelflinger (approved):
This code is not from us, it's from qcom's SGSR so it's unfortunate we have to modify it.

PR #9092, opened 2025-08-12
Begin to refactor test_LoadImage to use SharedShaders

bejado:
This updates _LoadImageTest.UpdateImage2D_ to use `SharedShaders`.

- Update `SharedShaders` to support different sampler types
- Add a `Builder` to `ScreenshotParams`
- Allow forcing an alpha value when comparing screenshots

### DIFF
--- a/filament/backend/test/ImageExpectations.cpp
+++ b/filament/backend/test/ImageExpectations.cpp
@@ -36,14 +36,16 @@
 namespace test {
 
 ScreenshotParams::ScreenshotParams(int width, int height, std::string fileName,
-        uint32_t expectedHash, bool isSrgb, int numAllowedDeviations, int pixelMatchThreshold)
+        uint32_t expectedHash, bool isSrgb, int numAllowedDeviations, int pixelMatchThreshold,
+        std::optional<float> forceAlphaValue)
     : mWidth(width),
       mHeight(height),
       mIsSrgb(isSrgb),
       mExpectedPixelHash(expectedHash),
       mFileName(std::move(fileName)),
       mAllowedPixelDeviations(numAllowedDeviations),
-      mPixelMatchThreshold(pixelMatchThreshold) {}
+      mPixelMatchThreshold(pixelMatchThreshold),
+      mForceAlphaValue(forceAlphaValue) {}
 
 int ScreenshotParams::width() const {
     return mWidth;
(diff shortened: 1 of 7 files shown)


filament/backend/test/SharedShaders.cpp
```diff
@@ -158,9 +158,19 @@ layout(binding = 0, set = 0) uniform Params {
 } params;
 )";
         }
-        case ShaderUniformType::Sampler: {
+        case ShaderUniformType::Sampler2D: {
```
AndyHovingh, line 161: Note that WebGPU/WGSL does not support combined image samplers. We need to define and bind both the texture and the sampler for it and use both in the call to sample the color from the texture.

In pure WGSL it is:
```
...
@group(0) @binding(0) var mySampler : sampler;
@group(0) @binding(1) var myTexture : texture_2d<f32>;
...
let color = textureSample(myTexture, mySampler, uv);
```

`textureLoad` can be used to lookup data directly from a texture without a sampler.

We deal with this in our shader generation code using a SPIRV-OPT pass that splits combined image samplers (see (libs/filamat/src/GLSLPostProcessor.cpp ~line 574)[https://github.com/google/filament/blob/5daa0cfe4b9d74a489117d8df393d01540af1a0a/libs/filamat/src/GLSLPostProcessor.cpp#L574]: `optimizer->RegisterPass(CreateSplitCombinedImageSamplerPass());` _(thanks to a lot of help from David Neto)_. We also, do some tricks to remap the bindings for them, as can be seen from the `rebindImageSamplerForWGSL(*spirv);` call below the pass. We account for this when creating bind group layouts and bind groups on the driver/backend side  _(see [filament/backend/src/webgpu/WebGPUDescriptorSetLayout.cpp ~lines 89-94](https://github.com/google/filament/blob/5daa0cfe4b9d74a489117d8df393d01540af1a0a/filament/backend/src/webgpu/WebGPUDescriptorSetLayout.cpp#L89-L94))_:
```cpp
// In WebGPU, textures and samplers are separate bindings.
/ We map the Filament binding index to two WebGPU binding indices:
// - texture: binding * 2
 // - sampler: binding * 2 + 1
wEntry.binding = fEntry.binding * 2;
entryInfo.binding = wEntry.binding;
```

I could be wrong, as I am not familiar with this side of the code, but I don't see where this is accounted for in the test shader generation. I think we need to modify the shader GLSL for the WebGPU case to handle this.

filament/backend/test/ImageExpectations.cpp
```diff
@@ -36,14 +36,16 @@
...
+        std::optional<float> forceAlphaValue)
     : mWidth(width),
       mHeight(height),
       mIsSrgb(isSrgb),
       mExpectedPixelHash(expectedHash),
       mFileName(std::move(fileName)),
       mAllowedPixelDeviations(numAllowedDeviations),
-      mPixelMatchThreshold(pixelMatchThreshold) {}
+      mPixelMatchThreshold(pixelMatchThreshold),
+      mForceAlphaValue(forceAlphaValue) {}
```
poweifeng, line 48: why is forcing an alpha value necessary?

PR #9094, opened 2025-08-12
GL: Fix memory leak when importing AHardwareBuffer

rafadevai:
In EGL Android, when importing an AHardwareBuffer from an ExternalImageHandle, the corresponding EGLImageKHR is not destroyed when calling destroyExternalImageTexture, causing a memory leak, that depending on the scenario will cause an OOM situation.


filament/backend/src/opengl/platforms/PlatformEGLAndroid.cpp
```diff
@@ -356,6 +361,9 @@ bool PlatformEGLAndroid::setImage(ExternalImageEGLAndroid const* eglExternalImag
         glBindTexture(GL_TEXTURE_2D, prevTexture);
         return false;
     }
+    ExternalTextureAndroid* outTexture = static_cast<ExternalTextureAndroid*>(texture);
+    outTexture->eglImage = eglImage;
```
poweifeng, line 371: What happens when you call `setImage` on the same texture with two different `ExternalImageEGLAndroid`?  

And can you clarify why it's better to associate the egl image with `ExternalTexture` as oppose to `ExternalImage`?

rafadevai: In this case will create a different eglImageKHR, so we ended up leaking memory in that case too. Let me adjust the PR to account for this case.

From what I see ExternalImage is tied to the ExternalImageHandle and can be destroyed after calling setExternalImage, so we need to tied to the lifecycle of the ExternalTexture instead.

rafadevai: Done

pixelflinger (approved):
good catch!

PR #9098, opened 2025-08-13
Update CMake min version to 3.22.1

poweifeng:
The reason for this bump is:
 - third_party/spirv-tools
 - third_party/dawn


jc3265:
should https://github.com/google/filament/blob/main/build.sh#L159 also be updated?

poweifeng:
> should https://github.com/google/filament/blob/main/build.sh#L159 also be updated?

Good catch.  It looks like those vars aren't really needed.

PR #9105, opened 2025-08-14
webgpu: Perform Blit when textures dont match

jc3265:


filament/backend/src/webgpu/WebGPUSwapChain.cpp
```diff
@@ -187,7 +187,7 @@ void initConfig(wgpu::SurfaceConfiguration& config, wgpu::Device const& device,
         wgpu::SurfaceCapabilities const& capabilities, wgpu::Extent2D const& extent,
         bool useSRGBColorSpace) {
     config.device = device;
-    config.usage = wgpu::TextureUsage::RenderAttachment | wgpu::TextureUsage::CopySrc;
+    config.usage = wgpu::TextureUsage::RenderAttachment | wgpu::TextureUsage::CopySrc | wgpu::TextureUsage::TextureBinding;
```
AndyHovingh: nit: line length

filament/backend/src/webgpu/WebGPUDriver.cpp
```diff
@@ -1367,22 +1367,77 @@ void WebGPUDriver::readPixels(Handle<HwRenderTarget> sourceRenderTargetHandle, c
     // This ensures that the readPixels will be ordered after all the draws.
     flush();
 
-    const size_t dstBytesPerPixel = PixelBufferDescriptor::computePixelSize(
-            pixelBufferDescriptor.format, pixelBufferDescriptor.type);
-    const size_t srcBytesPerPixel = getWGPUTextureFormatPixelSize(srcTexture.GetFormat());
+    const wgpu::TextureFormat srcFormat = srcTexture.GetFormat();
```
AndyHovingh: nit: uniform initialization

```cpp

const wgpu::TextureFormat srcFormat{ srcTexture.GetFormat() };
```

_(same for `dstFormat`, `needsFormatConversion`, `intermediateTexture`, `encoder`, `bytesPerPixel`, etc. ...)_

filament/backend/src/webgpu/WebGPUDriver.cpp
```diff
@@ -1367,22 +1367,77 @@ void WebGPUDriver::readPixels(Handle<HwRenderTarget> sourceRenderTargetHandle, c
...
+    const wgpu::TextureFormat dstFormat = toWebGPUFormat(pixelBufferDescriptor.format,
+                                                         pixelBufferDescriptor.type);
+
+    // If the source format is different from the destination (e.g. BGRA vs RGBA),
+    // we need to perform a conversion using an intermediate blit.
+    const bool needsFormatConversion = (srcFormat != dstFormat);
```
AndyHovingh: I think we should probably do the same as we do in update3DImage:

```cpp
const bool conversionNecessary{
    inputPixelFormat != outputLinearFormat && inputData->type != PixelDataType::COMPRESSED
}; // compressed formats should never need conversion
```
That implementation accounts for compressed formats and I think we can do a straight copy if the formats only differ by sRGB (but we should double check that).

Also, since we have the same logic/concern in both places, we should probably define a `constexpr` function that gets reused in both places.

filament/backend/src/webgpu/WebGPUDriver.cpp
```diff
@@ -1367,22 +1367,77 @@ void WebGPUDriver::readPixels(Handle<HwRenderTarget> sourceRenderTargetHandle, c
...
+
+    wgpu::Texture textureToReadFrom{srcTexture};
+    wgpu::Texture intermediateTexture;
```
AndyHovingh: maybe `stagingTexture` is more consistent naming?

filament/backend/src/webgpu/WebGPUDriver.cpp
```diff
@@ -1367,22 +1367,77 @@ void WebGPUDriver::readPixels(Handle<HwRenderTarget> sourceRenderTargetHandle, c
...
+                .format = dstFormat,
+                .mipLevelCount = 1,
+                .sampleCount = 1,
+        };
+        intermediateTexture = mDevice.CreateTexture(&intermediateDesc);
+        assert_invariant(intermediateTexture);
+        textureToReadFrom = intermediateTexture;
+
+        wgpu::CommandEncoderDescriptor desc{.label = "readPixels_blit_encoder"};
+        wgpu::CommandEncoder encoder = mDevice.CreateCommandEncoder(&desc);
```
AndyHovingh: shouldn't we follow the same pattern as we do elsewhere, where we reuse mCommandEncoder if it is there, otherwise create it there (and submit it if the encoder was not reused)?

filament/backend/src/webgpu/WebGPUDriver.cpp
```diff
@@ -1367,22 +1367,77 @@ void WebGPUDriver::readPixels(Handle<HwRenderTarget> sourceRenderTargetHandle, c
...
+                .usage = wgpu::TextureUsage::CopySrc | wgpu::TextureUsage::RenderAttachment,
+                .dimension = wgpu::TextureDimension::e2D,
+                .size = {
+                        .width = actualWidth,
+                        .height = actualHeight,
+                        .depthOrArrayLayers = 1,
+                },
+                .format = dstFormat,
+                .mipLevelCount = 1,
+                .sampleCount = 1,
```
AndyHovingh: how do we know it is single sampled? can we assume this?

filament/backend/src/webgpu/WebGPUDriver.cpp
```diff
@@ -1367,22 +1367,77 @@ void WebGPUDriver::readPixels(Handle<HwRenderTarget> sourceRenderTargetHandle, c
...
+                },
+                .filter = SamplerMagFilter::NEAREST, // Use NEAREST for a 1:1 copy
+        };
+        mBlitter.blit(mQueue, encoder, blitArgs);
 
-    FILAMENT_CHECK_PRECONDITION(dstBytesPerPixel == srcBytesPerPixel && dstBytesPerPixel > 0)
-            << "Source texture pixel size (" << srcBytesPerPixel
-            << ") does not match destination pixel buffer pixel size (" << dstBytesPerPixel
-            << "), or the format is not supported for readPixels.";
+        wgpu::CommandBuffer commandBuffer = encoder.Finish();
```
AndyHovingh: as mentioned above, maybe we reuse the command encoder and conditionally submit if we created a fresh encoder?

jc3265: good point

filament/backend/src/webgpu/WebGPUDriver.cpp
jc3265: good point

PR #9106, opened 2025-08-14
webgpu: enable choosing vulkan backend on MacOS

poweifeng:
- Add a Configuration struct to the WebGPUPlatform class.  This allows for client-side setting of WebGPU backend configurations.
 - Add a configuration for forcing the wgpu backend to pick a certain backend.
 - Add Vulkan as a potential backend for WebGPU + MacOS.
 - Locally modify Dawn so that it does not assume only switfshader is available for Vulkan on MacOS.
 - Enable vulkan support for Dawn (third_party/dawn/tnt/CMakeLists.txt)
 - Plumb option to pick different WebGPU backend through FilamentApp and gltf_viewer.


samples/gltf_viewer.cpp
```diff
@@ -215,7 +217,7 @@ static std::ifstream::pos_type getFileSize(const char* filename) {
 }
 
 static int handleCommandLineArguments(int argc, char* argv[], App* app) {
-    static constexpr const char* OPTSTR = "ha:f:i:usc:rt:b:evg:d";
+    static constexpr const char* OPTSTR = "ha:f:i:usc:rt:b:evg:dw:";
```
jc3265, line 220: Do we need to add the `w` option after like line 237 as well?

third_party/dawn/tnt/README.md
```diff
@@ -18,6 +18,7 @@ cd ..
 rm -rf dawn
 mv dawn_copy dawn
 patch -p2 < dawn/tnt/dawn-generator-CMakeList.patch
+patch -p2 < dawn/tnt/remove-vk-macos-restriction.patch
```
jc3265, line 21: I dont think the patch file was included in the PR

samples/gltf_viewer.cpp
poweifeng: yes, corrected.

third_party/dawn/tnt/README.md
poweifeng: added

libs/filamentapp/src/FilamentApp.cpp
```diff
@@ -26,6 +25,7 @@
 
 #include <imgui.h>
 
+#include <utils/Logger.h>
```
jc3265: nit: this should prob go after `EntityManager`
Should this include also be guarded by the `if webgpu`?

filament/backend/include/backend/platforms/WebGPUPlatform.h
```diff
@@ -52,6 +52,14 @@ class WebGPUPlatform final : public Platform {
     // either returns a valid device or panics
     [[nodiscard]] wgpu::Device requestDevice(wgpu::Adapter const& adapter);
 
+    struct Configuration {
+        wgpu::BackendType forceBackendType =  wgpu::BackendType::Undefined;
+    };
+
+    virtual Configuration getConfiguration() const noexcept {
```
jc3265: should this be `[[nodiscard]]`?

poweifeng: done

libs/filamentapp/src/FilamentApp.cpp
poweifeng: moved.

IMO we can keep it outside of `FILAMENT_SUPPORTS_WEBGPU` and not make the code even more complicated.

jc3265 (approved):

PR #9110, opened 2025-08-15
[WIP] Create exportable Syncs (fences)

anishmgoyal:
Note: creating this review as a WIP so we can discuss the API in further depth.

Creates exportable Fences on the Android platform, with the possibility of exporting to other platforms in the future. Exposes them to the frontend as "Sync" objects, which can only be exported to external sync objects (e.g. Fences).

### DIFF
--- a/filament/backend/include/backend/DriverEnums.h
+++ b/filament/backend/include/backend/DriverEnums.h
@@ -599,6 +599,20 @@ enum class FenceStatus : int8_t {
     TIMEOUT_EXPIRED = 1,        //!< wait()'s timeout expired. The Fence condition is not satisfied.
 };
 
+/**
+ * Error codes for DriverApi::getFenceFD()
+ * @see Fence, DriverApi::getFenceFD()
+ */
+enum class FenceConversionResult : int8_t {
+    HANDLE_NOT_AVAILABLE =
+            -3, //!< The underlying handle hasn't been allocated yet. Likely was requested too soon.
+    NOT_SUPPORTED =
+            -2,  //!< The current platform doesn't support converting fences to external handles.
+    ERROR = -1,  //!< An error occurred while converting the fence to an external handle.
+    SUCCESS = 0, //!< Successfully converted to external handle.
+    PREVIOUSLY_SIGNALED = 1, //!< The fence has already been signaled, can't convert.
+};
+
 static constexpr uint64_t FENCE_WAIT_FOR_EVER = uint64_t(-1);
 
 /**
--- a/filament/backend/include/backend/platforms/OpenGLPlatform.h
+++ b/filament/backend/include/backend/platforms/OpenGLPlatform.h
@@ -260,6 +260,18 @@ public:
     virtual Fence* UTILS_NULLABLE createFence() noexcept;
 
     /**
+     * Called by the driver to convert a fence to an external handle, if
+     * supported. Requires the fence to have been created with the appropriate
+     * flags, indicating that it can be exported. Will fail otherwise.
+     *
+     * @param fence The fence to convert to an external handle.
+     * @param fd A pointer that will be supplied with the external handle.
+     * @return A status code indicating if the conversion was successful.
+     */
+    virtual FenceConversionResult getFenceFD(Fence* UTILS_NONNULL fence,
+            int32_t* UTILS_NONNULL fd) noexcept;
+
+    /**
      * Destroys a Fence object. The default implementation does nothing.
      *
      * @param fence Fence to destroy.
--- a/filament/backend/include/backend/platforms/PlatformEGLAndroid.h
+++ b/filament/backend/include/backend/platforms/PlatformEGLAndroid.h
@@ -92,6 +92,8 @@ protected:
 
     Stream* createStream(void* nativeStream) noexcept override;
     void destroyStream(Stream* stream) noexcept override;
+    Fence* createFence() noexcept override;
+    FenceConversionResult getFenceFD(Fence* fence, int32_t* fd) noexcept override;
     void attach(Stream* stream, intptr_t tname) noexcept override;
     void detach(Stream* stream) noexcept override;
     void updateTexImage(Stream* stream, int64_t* timestamp) noexcept override;
--- a/filament/backend/include/backend/platforms/VulkanPlatform.h
+++ b/filament/backend/include/backend/platforms/VulkanPlatform.h
@@ -407,12 +407,16 @@ public:
         return {};
     }
 
+    virtual FenceConversionResult getFenceFD(VkFence fence, int32_t* fd) const noexcept;
+
 protected:
     virtual ExtensionSet getSwapchainInstanceExtensions() const;
 
     using SurfaceBundle = std::tuple<VkSurfaceKHR, VkExtent2D>;
     virtual SurfaceBundle createVkSurfaceKHR(void* nativeWindow, VkInstance instance,
             uint64_t flags) const noexcept;
+
+    virtual VkExternalFenceHandleTypeFlagBits getFenceExportFlags() const noexcept;
 
 private:
     // Platform dependent helper methods
--- a/filament/backend/include/backend/platforms/VulkanPlatformAndroid.h
+++ b/filament/backend/include/backend/platforms/VulkanPlatformAndroid.h
@@ -44,12 +44,16 @@ public:
 
     virtual ImageData createVkImageFromExternal(ExternalImageHandleRef image) const override;
 
+    virtual FenceConversionResult getFenceFD(VkFence fence, int32_t* fd) const noexcept override;
+
 protected:
     virtual ExtensionSet getSwapchainInstanceExtensions() const override;
 
     using SurfaceBundle = VulkanPlatform::SurfaceBundle;
     virtual SurfaceBundle createVkSurfaceKHR(void* nativeWindow, VkInstance instance,
             uint64_t flags) const noexcept override;
+
+    virtual VkExternalFenceHandleTypeFlagBits getFenceExportFlags() const noexcept override;
 
 private:
     struct ExternalImageVulkanAndroid : public Platform::ExternalImage {
--- a/filament/backend/include/private/backend/DriverAPI.inc
+++ b/filament/backend/include/private/backend/DriverAPI.inc
@@ -344,6 +344,8 @@ DECL_DRIVER_API_SYNCHRONOUS_N(void, setStreamDimensions, backend::StreamHandle, 
 DECL_DRIVER_API_SYNCHRONOUS_N(int64_t, getStreamTimestamp, backend::StreamHandle, stream)
 DECL_DRIVER_API_SYNCHRONOUS_N(void, updateStreams, backend::DriverApi*, driver)
 DECL_DRIVER_API_SYNCHRONOUS_N(backend::FenceStatus, getFenceStatus, backend::FenceHandle, fh)
+DECL_DRIVER_API_SYNCHRONOUS_N(backend::FenceConversionResult, getFenceFD, backend::FenceHandle, fh,
+        int32_t*, fd)
 DECL_DRIVER_API_SYNCHRONOUS_N(bool, isTextureFormatSupported, backend::TextureFormat, format)
 DECL_DRIVER_API_SYNCHRONOUS_0(bool, isTextureSwizzleSupported)
 DECL_DRIVER_API_SYNCHRONOUS_N(bool, isTextureFormatMipmappable, backend::TextureFormat, format)
--- a/filament/backend/src/noop/NoopDriver.cpp
+++ b/filament/backend/src/noop/NoopDriver.cpp
@@ -147,6 +147,10 @@ int64_t NoopDriver::getStreamTimestamp(Handle<HwStream> sh) {
 void NoopDriver::updateStreams(CommandStream* driver) {
 }
 
+FenceConversionResult NoopDriver::getFenceFD(Handle<HwFence> fh, int32_t* fd) {
+    return FenceConversionResult::NOT_SUPPORTED;
+}
+
 void NoopDriver::destroyFence(Handle<HwFence> fh) {
 }
 
--- a/filament/backend/src/opengl/OpenGLDriver.cpp
+++ b/filament/backend/src/opengl/OpenGLDriver.cpp
@@ -2235,6 +2235,22 @@ FenceStatus OpenGLDriver::getFenceStatus(Handle<HwFence> fh) {
     return FenceStatus::ERROR;
 }
 
+FenceConversionResult OpenGLDriver::getFenceFD(Handle<HwFence> fh, int32_t* fd) {
+    if (!fh) {
+        return FenceConversionResult::HANDLE_NOT_AVAILABLE;
+    }
+
+    GLFence* f = handle_cast<GLFence*>(fh);
+    if (f->fence == nullptr) {
+        if (!mPlatform.canCreateFence()) {
+            return FenceConversionResult::NOT_SUPPORTED;
+        }
+        return FenceConversionResult::HANDLE_NOT_AVAILABLE;
+    }
+
+    return mPlatform.getFenceFD(f->fence, fd);
+}
+
 bool OpenGLDriver::isTextureFormatSupported(TextureFormat format) {
     const auto& ext = mContext.ext;
     if (isETC2Compression(format)) {
--- a/filament/backend/src/opengl/platforms/PlatformEGLAndroid.cpp
+++ b/filament/backend/src/opengl/platforms/PlatformEGLAndroid.cpp
@@ -83,6 +83,7 @@ UTILS_PRIVATE PFNEGLGETCOMPOSITORTIMINGANDROIDPROC eglGetCompositorTimingANDROID
 UTILS_PRIVATE PFNEGLGETNEXTFRAMEIDANDROIDPROC eglGetNextFrameIdANDROID = {};
 UTILS_PRIVATE PFNEGLGETFRAMETIMESTAMPSUPPORTEDANDROIDPROC eglGetFrameTimestampSupportedANDROID = {};
 UTILS_PRIVATE PFNEGLGETFRAMETIMESTAMPSANDROIDPROC eglGetFrameTimestampsANDROID = {};
+UTILS_PRIVATE PFNEGLDUPNATIVEFENCEFDANDROIDPROC eglDupNativeFenceFDANDROID = {};
 }
 using namespace glext;
 
@@ -226,6 +227,8 @@ Driver* PlatformEGLAndroid::createDriver(void* sharedContext,
                 "eglGetFrameTimestampSupportedANDROID");
         eglGetFrameTimestampsANDROID = (PFNEGLGETFRAMETIMESTAMPSANDROIDPROC)eglGetProcAddress(
                 "eglGetFrameTimestampsANDROID");
+        eglDupNativeFenceFDANDROID =
+                (PFNEGLDUPNATIVEFENCEFDANDROIDPROC) eglGetProcAddress("eglDupNativeFenceFDANDROID");
     }
 
     mAssertNativeWindowIsValid = driverConfig.assertNativeWindowIsValid;
@@ -379,6 +382,20 @@ Platform::Stream* PlatformEGLAndroid::createStream(void* nativeStream) noexcept 
 
 void PlatformEGLAndroid::destroyStream(Platform::Stream* stream) noexcept {
     mExternalStreamManager.release(stream);
+}
+
+// For PlatformEGL on Android, create a fence that is exportable to an fd.
+Platform::Fence* PlatformEGLAndroid::createFence() noexcept {
+    return (Fence*) eglCreateSyncKHR(mEGLDisplay, EGL_SYNC_NATIVE_FENCE_ANDROID, nullptr);
+}
+
+FenceConversionResult PlatformEGLAndroid::getFenceFD(Platform::Fence* fence, int32_t* fd) noexcept {
+    EGLint tmpFd = eglDupNativeFenceFDANDROID(mEGLDisplay, (EGLSyncKHR) fence);
+    if (tmpFd == EGL_NO_NATIVE_FENCE_FD_ANDROID) {
+        return FenceConversionResult::ERROR;
+    }
+    *fd = tmpFd;
+    return FenceConversionResult::SUCCESS;
 }
 
 void PlatformEGLAndroid::attach(Stream* stream, intptr_t tname) noexcept {
--- a/filament/backend/src/vulkan/VulkanDriver.cpp
+++ b/filament/backend/src/vulkan/VulkanDriver.cpp
@@ -1033,6 +1033,19 @@ FenceStatus VulkanDriver::getFenceStatus(Handle<HwFence> fh) {
     return FenceStatus::TIMEOUT_EXPIRED;
 }
 
+FenceConversionResult VulkanDriver::getFenceFD(Handle<HwFence> fh, int32_t* fd) {
+    if (!fh) {
+        return FenceConversionResult::HANDLE_NOT_AVAILABLE;
+    }
+
+    VkFence fence = resource_ptr<VulkanFence>::cast(&mResourceManager, fh)->getVkFence();
+    if (fence == VK_NULL_HANDLE) {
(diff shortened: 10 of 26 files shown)


filament/backend/include/backend/DriverEnums.h
```diff
@@ -599,6 +599,20 @@ enum class FenceStatus : int8_t {
     TIMEOUT_EXPIRED = 1,        //!< wait()'s timeout expired. The Fence condition is not satisfied.
 };
 
+/**
+ * Error codes for DriverApi::getFenceFD()
+ * @see Fence, DriverApi::getFenceFD()
+ */
+enum class FenceConversionResult : int8_t {
```
poweifeng: Would suggest we consistently refer to it as sync

filament/backend/include/private/backend/DriverAPI.inc
```diff
@@ -344,6 +344,8 @@ DECL_DRIVER_API_SYNCHRONOUS_N(void, setStreamDimensions, backend::StreamHandle,
 DECL_DRIVER_API_SYNCHRONOUS_N(int64_t, getStreamTimestamp, backend::StreamHandle, stream)
 DECL_DRIVER_API_SYNCHRONOUS_N(void, updateStreams, backend::DriverApi*, driver)
 DECL_DRIVER_API_SYNCHRONOUS_N(backend::FenceStatus, getFenceStatus, backend::FenceHandle, fh)
+DECL_DRIVER_API_SYNCHRONOUS_N(backend::FenceConversionResult, getFenceFD, backend::FenceHandle, fh,
+        int32_t*, fd)
```
poweifeng: can we call it something like exportSync?

filament/backend/include/backend/DriverEnums.h
```diff
@@ -599,6 +599,20 @@ enum class FenceStatus : int8_t {
...
+    HANDLE_NOT_AVAILABLE =
+            -3, //!< The underlying handle hasn't been allocated yet. Likely was requested too soon.
+    NOT_SUPPORTED =
+            -2,  //!< The current platform doesn't support converting fences to external handles.
```
poweifeng: I think you can just assert in this case.

filament/backend/include/backend/platforms/OpenGLPlatform.h
```diff
@@ -259,6 +259,18 @@ class OpenGLPlatform : public Platform {
...
+     * Called by the driver to convert a fence to an external handle, if
+     * supported. Requires the fence to have been created with the appropriate
+     * flags, indicating that it can be exported. Will fail otherwise.
+     *
+     * @param fence The fence to convert to an external handle.
+     * @param fd A pointer that will be supplied with the external handle.
+     * @return A status code indicating if the conversion was successful.
+     */
+    virtual FenceConversionResult getFenceFD(Fence* UTILS_NONNULL fence,
+            int32_t* UTILS_NONNULL fd) noexcept;
```
poweifeng: instead of int32_t, we need to declare an opaque type.  We need it to be general enough to support other platforms.

filament/backend/src/opengl/OpenGLDriver.cpp
```diff
@@ -2235,6 +2235,22 @@ FenceStatus OpenGLDriver::getFenceStatus(Handle<HwFence> fh) {
     return FenceStatus::ERROR;
 }
 
+FenceConversionResult OpenGLDriver::getFenceFD(Handle<HwFence> fh, int32_t* fd) {
+    if (!fh) {
```
poweifeng: you should just `assert_invariant(fh)`

filament/src/details/Sync.cpp
```diff
@@ -0,0 +1,67 @@
...
+
+namespace filament {
+
+using DriverApi = backend::DriverApi;
+
+FSync::FSync(FEngine& engine)
+    : mEngine(engine) {
+    DriverApi& driverApi = engine.getDriverApi();
+    mHwFence = driverApi.createFence();
+    driverApi.queueCommand([this]() {
```
pixelflinger: you cannot capture `this` here because by the time this gets called, `FSync` could have been destroyed. e.g. it's legal for the user to do :
```
destroySync(createSync())
```
Instead, you have to use a `shared_ptr<>` that contains the state you care about and you capture it with a `weak_ptr<>`

filament/src/details/Sync.cpp
```diff
@@ -0,0 +1,67 @@
...
+    for (const auto& callback: mConversionCallbacks) {
+        processCallback(callback);
+    }
+}
+
+void FSync::processCallback(SyncConversionCallback callback) {
+    DriverApi& driverApi = mEngine.getDriverApi();
+    int32_t externalHandle;
+    SyncConversionResult result = driverApi.getFenceFD(mHwFence, &externalHandle);
+    callback(result, externalHandle);
```
pixelflinger: You cannot call user callbacks from the backend thread", this is very dangerous and we never allow it.

Look how other callbacks are handled in the backends (We use something called scheduleDestroy that's used for pixelbufferdescriptor. You need to do something similar).

filament/src/details/Sync.cpp
```diff
@@ -0,0 +1,67 @@
...
+    driverApi.queueCommand([this]() {
+        {
+            std::lock_guard lock(mMutex);
+            mHasHandle = true;
+        }
+        processAllCallbacks();
+    });
+}
+
+void FSync::terminate(FEngine& engine) noexcept { engine.getDriverApi().destroyFence(mHwFence); }
```
pixelflinger: not on one  line please

filament/src/details/Sync.cpp
```diff
@@ -0,0 +1,67 @@
...
+FSync::FSync(FEngine& engine)
+    : mEngine(engine) {
+    DriverApi& driverApi = engine.getDriverApi();
+    mHwFence = driverApi.createFence();
+    driverApi.queueCommand([this]() {
+        {
+            std::lock_guard lock(mMutex);
+            mHasHandle = true;
+        }
+        processAllCallbacks();
```
pixelflinger: this will call the callbacks from the backend thread, which we never allow (see below).

filament/src/details/Sync.h
```diff
@@ -0,0 +1,69 @@
...
+
+    /**
+     * Converts a sync object to one that can be used externally to
+     * wait for some GPU work to be completed within Filament before proceeding
+     *
+     * @param callback A function that will receive the external sync handle, if
+     *                 conversion was successful, along with a conversion result
+     *                 code (e.g. SUCCESS, ERROR, NOT_SUPPORTED).
+     */
+    void convertToExternalSync(SyncConversionCallback callback) noexcept;
```
pixelflinger: I would call this `getExternalSync`

filament/include/filament/Sync.h
```diff
@@ -0,0 +1,50 @@
...
+#include <backend/DriverEnums.h>
+
+#include <functional>
+
+namespace filament {
+
+class UTILS_PUBLIC Sync : public FilamentAPI {
+public:
+    using SyncConversionResult = backend::FenceConversionResult;
+    using SyncConversionCallback = std::function<void(SyncConversionResult, int32_t)>;
```
pixelflinger: what's the `int32_t` here?

You can't assume the "external sync" will be a `fd` or will fit on a `int32_t`.  (also isn't a fd an `int`, and not `int32_t`??).

We should do all this via opaque objects that exists only at the PlatformLevel. And then PlatformEGLAndroid will have a *custom* method on it that would convert the opaque object into a `fd`.

pixelflinger (changes requested):

filament/src/details/Sync.cpp
anishmgoyal: ahh - clang-format did this one, will undo

filament/include/filament/Sync.h
anishmgoyal: makes sense; will refactor these.

filament/backend/include/backend/platforms/OpenGLPlatform.h
poweifeng: actually mentioend elsewhere.

filament/backend/include/private/backend/DriverAPI.inc
anishmgoyal: how's getPlatformSync? could do exportSync, but since it isn't technically exporting the sync thought i'd check. the new function just returns a Platform::Sync once a callback resolves.

filament/backend/src/vulkan/platform/VulkanPlatform.cpp
```diff
@@ -1002,6 +1008,19 @@ SwapChainPtr VulkanPlatform::createSwapChain(void* nativeWindow, uint64_t flags,
     return swapchain;
 }
 
+Platform::Sync* VulkanPlatform::createSync(VkFence fence,
+        std::shared_ptr<VulkanCmdFence> fenceStatus) noexcept {
+    return new VulkanPlatform::Sync(fence, fenceStatus);
+}
+
+bool VulkanPlatform::convertSyncToFd(Platform::Sync* sync, int32_t* fd) const noexcept {
```
poweifeng: I don't hink you handled the part about `int32_t` not being an opaque type.

filament/backend/src/vulkan/platform/VulkanPlatformAndroid.cpp
```diff
@@ -389,6 +389,31 @@ VulkanPlatform::ImageData VulkanPlatformAndroid::createVkImageFromExternal(
     };
 }
 
+bool VulkanPlatformAndroid::convertSyncToFd(Platform::Sync* sync, int32_t* fd) const noexcept {
+    assert_invariant(sync != nullptr && fd != nullptr);
```
poweifeng: nit: filament idiom is just `assert_invariant(sync && fd)`

poweifeng: same with `getFenceStatus()` below

filament/backend/src/vulkan/VulkanDriver.cpp
```diff
@@ -795,6 +802,33 @@ void VulkanDriver::createFenceR(Handle<HwFence> fh, int) {
...
+        fence = mCommands.getMostRecentFence();
+        fenceStatus = mCommands.getMostRecentFenceStatus();
+    }
+
+    {
+        std::lock_guard<std::mutex> guard(sync->lock);
+        sync->sync = mPlatform->createSync(fence, fenceStatus);
+    }
+
+    for (auto& cbData : sync->conversionCallbacks) {
```
poweifeng, line 853: I'd clear `sync->conversionCallbacks` after this loop.

poweifeng: actually, don't you have to lock `conversionCallbacks` as well, since you're `push_back`-ing in `getPlatformSync`?

poweifeng: I think you can just `std::swap` with a locally defined `std::vector` in the mutex-locked block above.

filament/backend/src/vulkan/platform/VulkanPlatform.cpp
anishmgoyal: wasn't the point to not have `int32_t` in the public api? this is the platform api.

here, fd is an output variable, not an input. what would you prefer this function header be?

filament/backend/src/vulkan/VulkanDriver.cpp
anishmgoyal: ahh fair - i'm realizing now there is a narrow possibility where conversion callbacks could be written to concurrently, if false is read, and then true is written. i will update this

anishmgoyal: actually no - this is not possible.

```
lock;
check if sync->sync is null;
if yes:
  push_back;
unlock
```

here:

```
lock
sync->sync = value
unlock
read loop
```

so if a thread is pushing something onto the vector, that means it holds the lock and sync was found to be null with the lock held.

if a thread is reading the loop, it acquired the lock, and then set a value in sync. so while reading, it's not possible for anyone else to push_back onto the vector, because anyone else who acquires the lock will see that sync->sync is not null, and not push anything.

anishmgoyal: i will clear the vector, but i don't think it makes sense to add the swap to the mutex unless i'm wrong about there not being a race condition.

this has been updated. please resolve if this is sufficiently addressed, else we can discuss further.

filament/backend/src/vulkan/platform/VulkanPlatform.cpp
poweifeng: so the issue, I think, is that if I wanted to export a fence on windows (or some other platform), it might not be represented as int32_t. If you want to keep this as a VulkanPlatformAndroid API, then I think that's fine.

anishmgoyal: gotcha. i'll remove this from the base platform objects then

filament/backend/src/vulkan/VulkanDriver.cpp
poweifeng: yes, I think you're right.  thanks for the clarification. clearing is sufficient.

filament/backend/include/backend/platforms/PlatformEGLAndroid.h
```diff
@@ -26,6 +26,7 @@
 #include <utils/compiler.h>
 
 #include <chrono>
+#include <vector>
```
poweifeng: unused?

anishmgoyal: removed

poweifeng (approved):

filament/backend/src/vulkan/VulkanDriver.cpp
```diff
@@ -820,6 +827,36 @@ void VulkanDriver::createFenceR(Handle<HwFence> fh, utils::CString tag) {
...
+    auto sync = resource_ptr<VulkanSync>::cast(&mResourceManager, sh);
+    VkFence fence = VK_NULL_HANDLE;
+    std::shared_ptr<VulkanCmdFence> fenceStatus;
+    if (mCurrentRenderPass.commandBuffer) {
+        VulkanCommandBuffer* cmdBuff = mCurrentRenderPass.commandBuffer;
+        fence = cmdBuff->getVkFence();
+        fenceStatus = cmdBuff->getFenceStatus();
+        // If we're currently recording, flush so that the fence only applies
+        // to commands already issued.
+        mCommands.flush();
```
rafadevai, line 842: is it possible that this command is being processed while a render pass is being recorded? if so, I dont think we want to flush the current command buffer.

@poweifeng what do you think?

filament/backend/src/opengl/platforms/PlatformEGLAndroid.cpp
```diff
@@ -395,6 +398,33 @@ void PlatformEGLAndroid::destroyStream(Platform::Stream* stream) noexcept {
...
+    return new PlatformEGLAndroid::Sync(mEGLDisplay);
+}
+
+bool PlatformEGLAndroid::convertSyncToFd(Platform::Sync* sync, int32_t* fd) noexcept {
+    assert_invariant(sync != nullptr && fd != nullptr);
+    PlatformEGLAndroid::Sync& eglSync = static_cast<PlatformEGLAndroid::Sync&>(*sync);
+    *fd = eglDupNativeFenceFDANDROID(mEGLDisplay, eglSync.getSync());
+    // In the case where there was no native FD, -1 is returned. Return false
+    // to indicate there was an error in this case.
+    return *fd != -1;
```
rafadevai: instead of -1, use EGL_NO_NATIVE_FENCE_FD_ANDROID to know if there was an error

and also check the result of eglGetError

filament/backend/src/noop/NoopDriver.cpp
```diff
@@ -16,11 +16,13 @@
 
 #include <backend/DriverEnums.h>
 #include <backend/Handle.h>
+#include <backend/Platform.h>
 
 #include "noop/NoopDriver.h"
 #include "CommandStreamDispatcher.h"
 
 #include <backend/Handle.h>
+#include <backend/Platform.h>
```
poweifeng: duplicate includes

filament/backend/src/vulkan/VulkanDriver.cpp
poweifeng: I think it's not possible since the client-side can only call createSync between frames or calls to `render(view)`, which means that this will be inserted between render passes, and not in them.

filament/backend/src/opengl/platforms/PlatformEGLAndroid.cpp
anishmgoyal: done

filament/backend/src/noop/NoopDriver.cpp
anishmgoyal: removed

filament/backend/include/backend/Platform.h
```diff
@@ -43,6 +44,24 @@ class UTILS_PUBLIC Platform {
...
     struct Stream {};
 
+    class Sync {
+    public:
+        virtual ~Sync();
+
+    protected:
+        // Avoid direct instantiation.
+        Sync() = default;
+    };
```
pixelflinger: Why not use the same pattern as other types like Fence or Stream? We shouldn't need Sync do have a vtable, because all operations are done via a concrete type who knows which type it is.

filament/backend/include/backend/platforms/OpenGLPlatform.h
```diff
@@ -52,6 +52,12 @@ class OpenGLPlatform : public Platform {
 
     ~OpenGLPlatform() noexcept override;
 
+    class Sync : public Platform::Sync {
+    public:
+        Sync() noexcept = default;
+        virtual ~Sync() noexcept;
+    };
+
```
pixelflinger: I don't think that's needed at all.

filament/backend/include/backend/platforms/PlatformEGLAndroid.h
```diff
@@ -74,6 +85,20 @@ class PlatformEGLAndroid : public PlatformEGL {
...
+        virtual ~Sync() noexcept;
+
+        inline EGLSyncKHR getSync() {
+            return mSync;
+        }
+
+    private:
+        EGLDisplay mEGLDisplay;
+        EGLSyncKHR mSync;
+    };
```
pixelflinger: Doesn't need to have a virtual. Also this should be in PlatformEGL, this part is not restricted to  Android. I would also rename the type to `SyncEGL` to keep the same convention of other types.

filament/backend/include/backend/platforms/VulkanPlatform.h
```diff
@@ -29,9 +30,9 @@
 
 #include <cstddef>
 #include <functional>
+#include <string>
```
pixelflinger, line 33: <string> is banned from the backend. Use CString.

filament/backend/include/backend/platforms/VulkanPlatform.h
```diff
@@ -408,12 +432,35 @@ class VulkanPlatform : public Platform, utils::PrivateImplementation<VulkanPlatf
...
+
+        inline const std::shared_ptr<VulkanCmdFence> getFenceStatus() const noexcept {
+            return mFenceStatus;
+        }
+
+    private:
+        VkFence mFence;
+        std::shared_ptr<VulkanCmdFence> mFenceStatus;
+    };
+
```
pixelflinger: same as for EGL. Should be called SyncVK or SyncVulkan. doesn't need to have a vtable.

filament/backend/include/backend/platforms/VulkanPlatformAndroid.h
```diff
@@ -44,13 +44,26 @@ class VulkanPlatformAndroid : public VulkanPlatform {
...
+    /**
+     * Converts a sync to an external file descriptor, if possible. Accepts an
+     * opaque handle to a sync, as well as a pointer to where the fd should be
+     * stored.
+     * @param sync The sync to be converted to a file descriptor.
+     * @param fd   A pointer to where the file descriptor should be stored.
+     * @return `true` on success, `false` on failure. The default implementation
+     *         returns `false`.
+     */
+    bool convertSyncToFd(Platform::Sync* sync, int* fd) const noexcept;
```
pixelflinger, line 56: This should take a `SyncEGL*`

anishmgoyal: i disagree on this point - there's no way to get a SyncEGL* from the frontend filament::Sync, unless you cast from outside filament. seeing as createFence in the platform objects is doing its own sort of cast, I feel this should do the same.

*preferably* i'd rather even go one step further and have this be a virtual function on the platform sync class, but if we're avoiding virtual calls entirely, i think this is the cleanest api we can provide.

filament/backend/include/backend/Platform.h
anishmgoyal: edit: will change to a struct.

~Fence is using completely disjoint types, and casting to get around this. If we're going to go this route, what I'd propose is completely disjoint types here as well, and having the entire thing hinge on:~

~class Platform {
  typedef Sync* = void*;
}~

~If that's preferred, happy to make that change. I don't know if I agree that we should use disjoint types and use casts to get the compiler to ignore this.~

filament/backend/include/backend/platforms/OpenGLPlatform.h
anishmgoyal: edit: will use the base platform::sync struct.

~i'd rather not leave sync as nullptr, because downstream api's could come to code against this assumption in weird ways. do we feel strongly that it should be null in cases that it can't be converted to an fd?~

filament/backend/include/backend/platforms/PlatformEGLAndroid.h
anishmgoyal: i'll move it to platformegl, that's a fair point.

edit: will change to a struct.

~re virtual, if we don't have a vtable in a derived class, it raises a warning, which either we or impress is elevating to an error. so it either has to have a vtable, or be a disjoint type.~

filament/backend/include/backend/platforms/VulkanPlatform.h
anishmgoyal: if you check the diff, this is not introduced in this change. clang-format re-ordered the includes. i can revert the change in ordering, but i feel that removing the use of this header is out-of-scope of this change.

filament/backend/include/backend/platforms/VulkanPlatform.h
anishmgoyal: i'll rename this.

edit: will change to a struct.

~re virtual, if we don't have a vtable in a derived class, it raises a warning, which either we or impress is elevating to an error. so it either has to have a vtable, or be a disjoint type.~

filament/backend/include/backend/platforms/PlatformEGLAndroid.h
anishmgoyal: this can't be in platformEGL actually - we explicitly define the sync as an android sync when creating it, and the conversion method is android-only. we'd be overriding it anyways, i'd rather just introduce a concrete sync class in PlatformEGL later, if we need one.

filament/backend/include/backend/Platform.h
```diff
@@ -42,6 +43,16 @@ class UTILS_PUBLIC Platform {
...
+    struct Sync {};
+
+    using SyncCallback = void(*)(Sync* UTILS_NONNULL sync, void* UTILS_NULLABLE userData);
+
+    struct SyncCallbackData {
+        CallbackHandler* UTILS_NULLABLE handler;
+        SyncCallback UTILS_NULLABLE cb;
+        Sync* UTILS_NULLABLE sync;
+        void* UTILS_NULLABLE userData;
+    };
```
pixelflinger: This structure shouldn't be public right? It's completely an implementation detail of the concrete platform it seems?

pixelflinger:
@anishmgoyal ios and mac builds are failing, fyi.

filament/backend/include/backend/Platform.h
anishmgoyal: hmm - fair point. i'll move this over to be a sub-struct of the sync classes, if that makes sense? e.g.

```
struct VulkanSync : public HwSync {
  struct CallbackData {

  };
};
```

PR #9112, opened 2025-08-15
Guard against uninitialized thickness in refractionSolidSphere

PatRibas:
Duplicated from `refractionThinSphere`


pixelflinger (changes requested):
the correct way to fix this is to initialize thickness'd default value. This fix is fragile because if thickness is used somewhere else, we'll have the same issue. 
From a performance standpoint, there is no difference due to inlining and DCE.

PatRibas:
Added `|| defined(MATERIAL_HAS_REFRACTION)` to the `pixel.thickness` initialization line.

pixelflinger:
https://github.com/google/filament/pull/9135 should fix this

Issue #9111, opened 2025-08-15
matc error in Windows build on latest main


patrikhuber:
**Describe the bug**
When I compile the `filament` CMake target on Windows using Visual Studio, I get a bunch of `matc.exe` errors during the build:

<img width="435" height="339" alt="Image" src="https://github.com/user-attachments/assets/901aa021-8948-44d6-aed7-3742dcf183f9" />

This happens when matc is invoked during the build, and trying to compile some materials:

```
[53/172] Compiling material C:/Users/PatrikHuber/Projects/Dependencies/filament-github-source/filament/src/materials/bloom/bloomDownsample2x.mat
FAILED: filament/generated/material/bloomDownsample2x.filamat C:/Users/PatrikHuber/Projects/Dependencies/filament-github-source/out/build/x64-debug/filament/generated/material/bloomDownsample2x.filamat 
C:\WINDOWS\system32\cmd.exe /C "cd /D C:\Users\PatrikHuber\Projects\Dependencies\filament-github-source\out\build\x64-debug\filament && C:\Users\PatrikHuber\Projects\Dependencies\filament-github-source\out\build\x64-debug\tools\matc\matc.exe -a opengl -a vulkan -p desktop -d -g -o C:/Users/PatrikHuber/Projects/Dependencies/filament-github-source/out/build/x64-debug/filament/generated/material/bloomDownsample2x.filamat C:/Users/PatrikHuber/Projects/Dependencies/filament-github-source/filament/src/materials/bloom/bloomDownsample2x.mat"
[54/172] Compiling material C:/Users/PatrikHuber/Projects/Dependencies/filament-github-source/filament/src/materials/colorGrading/colorGrading.mat
FAILED: filament/generated/material/colorGrading.filamat C:/Users/PatrikHuber/Projects/Dependencies/filament-github-source/out/build/x64-debug/filament/generated/material/colorGrading.filamat 
[...]

Issue #9062, opened 2025-08-08
Random race condition during `engine->flushAndWait()` in offscreen rendring mode


elias-Mimetrik:
⚠️ **Issues not using this template will be systematically closed.**

**Describe the bug**
Rendering in headless mode crashes randomly, especially when more than a single renderer is used.

**To Reproduce**
I construct the renderer components as in this thread (https://github.com/google/filament/issues/3318), I know this will work given the comment at the end.

This is my code to render a frame:

```
 static void callback(void* buffer, size_t size, void* user) {
	cv::Mat image = cv::Mat(500, 500, CV_8UC4, static_cast<uint8_t*>(buffer)).clone();
	image.convertTo(image, CV_8U);
	cv::cvtColor(image, image, cv::COLOR_RGBA2RGB);
	cv::imwrite("aaa.png", image);
	delete[] static_cast<uint8_t*>(buffer);
}

...

// Offscreen rendering:
size_t size = 500 * 500 * 4;
uint8_t* buffer = new uint8_t[size];

filament::backend::PixelBufferDescriptor pd(buffer, size,
      filament::backend::PixelBufferDescriptor::PixelDataFormat::RGBA,
      filament::backend::PixelBufferDescriptor::PixelDataType::UBYTE,
      callback, nullptr);

this->filament_renderer->beginFrame(this->swap_chain);
this->filament_renderer->render(this->view);
this->filament_renderer->readPixels(0, 0, 500, 500, std::move(pd));
this->filament_renderer->endFrame();
this->engine->flushAndWait(); // <<<<<<<<<<<<<<<<<<<<<<<<< Exception here!
```

Basically when I initialise two renderer objects `std::unique_ptr<mimetrik::occlusioncamera::Renderer>`, the first time I render an image using one of them, it crashes at the `this->engine->flushAndWait();` command. I cannot understand why this is happening.. Are we limited to using a single Renderer?

**Expected behavior**
I expect a frame to be rendered successfully without any exceptions.

**Screenshots**

**Logs**
If applicable, copy **full** logs from your console here. Please *do not*
use screenshots of logs, copy them as text, use gist or attach an *uncompressed* file.

**Desktop (please complete the following information):**
 - OS: ubuntu 24.04
 - GPU: RTX 4080
 - Backend: Vulkan

**Smartphone (please complete the following information):**
 - Device: [e.g. Pixel 2]
 - OS: [e.g. Android Pie 9.0]

**Additional context**

pixelflinger:
[...]

PR #9113, opened 2025-08-16
Issue 9062 repro

patrikhuber:
This is a repro for https://github.com/google/filament/issues/9062, in the form of a filament sample.

I couldn't base it off the latest main branch because of #9111 - so this is based off of v1.63.0. The only relevant addition is the `samples/issue_9062_repro.cpp` file.

A few notes and new observations regarding the sample & #9062:
- I managed to force the crash to happen almost every time by increasing the size of the mesh - I'm just creating a 1000 x 800 grid of triangles in the XY plane. Note that more triangles doesn't necessarily mean the crash happens more often - some constellations, like e.g. if I use a grid of 1000 x 1000 triangles, and the crash only rarely happens.
- I've been mostly using a debug build, but the bug repros in a release build too.
- With that bigger mesh, the bug now reproduces on Windows **without calling readPixels()** too! So it looks like a more generic issue?
- On Windows (MSVC), the crash almost exclusively happens only when the renderer is destroyed after the final rendered frame (e.g. after frame 1 if `num_frames=1`, after frame 100 if `num_frames=100`. **However, on Ubuntu Linux, when setting `num_frames > 1` (e.g. `10000`), the crash happens on the same line (`loadFromCpu` with Vulkan backend) and with the same stack trace while in the middle of rendering frames (e.g. on frame 3153 or whatever - randomly), and not on the final frame**. (And on Linux it also crashes with and without `readPixels()` - and it crashes more often / reliably with a bigger mesh - same behaviour as on Windows).

Lastly, while pushing this example, I noticed that this code in `add_mesh()` allocates memory for `indices_arr`, `vertices_f`, and `tbns` using `new`, but never deletes it. I looked at some other Filament examples and some of them use the third/fourth argument of `VertexBuffer::BufferDescriptor` to specify a callback that deletes that memory. I'm not sure if this could be related to the issue (in which case there wouldn't be a bug in Filament, I guess), but I would have thought the worst that could do is produce a memory leak, because memory is allocated but never deleted - and it should not produce this crash.


patrikhuber:
I'm not sure why this PR shows as 963 file diff - it looks like there might have been a force-push to main since I branched off of v1.63.0 for this. In any case, the "Files changed" diff is small and the only file added here is `samples/issue_9062_repro.cpp`.

patrikhuber:
Force-pushed: Rebased on latest main, and printing a message now if it destructs successfully (if the message is not printed, it means it crashed).

poweifeng:
See comment here: https://github.com/google/filament/issues/9062#issuecomment-3202947540

PR #9115, opened 2025-08-18
Doc: Fix broken link

icheered:
The doc links to Filament and Materials pages on the index page are broken due to incorrect relative file location.


poweifeng:
so /docs is actually generated.  

I have another PR for fixing this and other broken links: https://github.com/google/filament/pull/9118

Thanks for bringing this to our attention.

poweifeng:
Fixed in #9118

PR #9120, opened 2025-08-19
wgpu: fix finish(), where it was erroneously returning if a command encoder was not in flight. Also, it was not waiting for async buffer maps to finish

AndyHovingh:


filament/backend/src/webgpu/WebGPUDriver.cpp
```diff
@@ -1543,6 +1544,7 @@ void WebGPUDriver::readPixels(Handle<HwRenderTarget> sourceRenderTargetHandle, c
                     FWGPU_LOGE << "Failed to map staging buffer for readPixels: " << message;
                 }
                 data->driver->scheduleDestroy(std::move(data->pixelBufferDescriptor));
+                data->driver->mReadPixelMapsCounter.finishTask();
```
jc3265, line 1547: so shouldnt we also add this do the destructor?

AndyHovingh: no. This should only happen when we actually finish mapping a buffer. That doesn't happen in the destructor; decrementing the counter there would be confusing. If we have not mapped all the buffers at destruction time, this counter should reflect that reality.

PR #9121, opened 2025-08-19
webgpu: Implement blitDeprecated

jc3265:
BUGS = [436887647]

### DIFF
--- a/filament/backend/src/webgpu/WebGPUBlitter.cpp
+++ b/filament/backend/src/webgpu/WebGPUBlitter.cpp
@@ -677,11 +677,11 @@ wgpu::BindGroupLayout WebGPUBlitter::createTextureBindGroupLayout(const SamplerM
             .texture = {
                 .sampleType = depthSource
                                       ? wgpu::TextureSampleType::Depth
-                                      : wgpu::TextureSampleType::Float, // only F32 scalar sample
-                                                                        // type supported for now
-                                                                        // (aside from depth)
-                                                                        // see assumptions listed
-                                                                        // in blit function
+                                      : (multisampledSource
+                                                ? wgpu::TextureSampleType::UnfilterableFloat
+                                                : wgpu::TextureSampleType::Float), // only F32 scalar sample
+                                                                                   // type supported for now
+                                                                                   // (aside from depth)
                 .viewDimension = sourceDimension,
                 .multisampled = multisampledSource,
             },
@@ -758,7 +758,7 @@ wgpu::ShaderModule WebGPUBlitter::createShaderModule(
         }
     } else {
         if (multisampledSource) {
-            textureType = "texture_multisampled_2d";
+            textureType = "texture_multisampled_2d<f32>";
         } else {
             if (sourceDimension == wgpu::TextureViewDimension::e3D) {
                 textureType = "texture_3d<f32>";
--- a/filament/backend/src/webgpu/WebGPUDriver.cpp
+++ b/filament/backend/src/webgpu/WebGPUDriver.cpp
@@ -1559,7 +1559,87 @@ void WebGPUDriver::blitDEPRECATED(TargetBufferFlags buffers,
         Handle<HwRenderTarget> destinationRenderTargetHandle, const Viewport destinationViewport,
         Handle<HwRenderTarget> sourceRenderTargetHandle, const Viewport sourceViewport,
         const SamplerMagFilter filter) {
-    PANIC_PRECONDITION("WebGPUDriver::blitDEPRECATED not supported");
+
+    auto const sourceTarget{ handleCast<WebGPURenderTarget>(sourceRenderTargetHandle) };
+    auto const destinationTarget{ handleCast<WebGPURenderTarget>(destinationRenderTargetHandle) };
+    assert_invariant(sourceTarget && destinationTarget);
+
+    FILAMENT_CHECK_PRECONDITION(buffers == TargetBufferFlags::COLOR0)
+            << "blitDEPRECATED only supports COLOR0";
+
+    FILAMENT_CHECK_PRECONDITION(sourceViewport.left >= 0 && sourceViewport.bottom >= 0 &&
+                                destinationViewport.left >= 0 && destinationViewport.bottom >= 0)
+            << "Source and destination viewports must be positive.";
+
+    // We always blit from/to the COLOR0 attachment.
+    auto sourceAttachment {sourceTarget->getColorAttachmentInfos()[0]};
+    Handle<HwTexture> sourceTextureHandle{ sourceAttachment.handle };
+    auto destinationAttachment {destinationTarget->getColorAttachmentInfos()[0]};
+    Handle<HwTexture> destinationTextureHandle{
+        destinationAttachment.handle
+    };
+
+    if (UTILS_UNLIKELY(!sourceTextureHandle || !destinationTextureHandle)) {
+        FWGPU_LOGE << "blitDEPRECATED could not find a valid color attachment to blit.";
+        return;
+    }
+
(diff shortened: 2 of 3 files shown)


filament/backend/src/webgpu/WebGPUDriver.cpp
```diff
@@ -1559,7 +1559,74 @@ void WebGPUDriver::blitDEPRECATED(TargetBufferFlags buffers,
         Handle<HwRenderTarget> destinationRenderTargetHandle, const Viewport destinationViewport,
         Handle<HwRenderTarget> sourceRenderTargetHandle, const Viewport sourceViewport,
         const SamplerMagFilter filter) {
-    PANIC_PRECONDITION("WebGPUDriver::blitDEPRECATED not supported");
+
+    auto const srcTarget = handleCast<WebGPURenderTarget>(sourceRenderTargetHandle);
```
AndyHovingh: nit: uniform initialization, e.g. `auto const srcTarget{ handleCast<WebGPURenderTarget>(sourceRenderTargetHandle) };`

_(same comment for `dstTarget`, `sourceTextureHandle`, `destinationTextureHandle`, `srcTexture`, `srcTextureHeight`, `flippedSourceY`, `dstTexture`, `dstTextureHeight`, `flippedDestinationY`, `sourceOrigin`, `destinationOrigin`, `sourceSize`, `destinationSize`,  ...)_

note the `=` is not necessary in `const wgpu::Origin2D sourceOrigin = { (uint32_t) sourceViewport.left, flippedSourceY };`; it can be `const wgpu::Origin2D sourceOrigin{ ...`

AndyHovingh: nit: looks like sometimes we use `src`, others `source`. Sometimes we use `dst` and other others `destination`. IMO I would prefer explicit naming, e.g. `source` and `destination`.

filament/backend/src/webgpu/WebGPUDriver.cpp
```diff
@@ -1559,7 +1559,74 @@ void WebGPUDriver::blitDEPRECATED(TargetBufferFlags buffers,
...
+    const uint32_t srcTextureHeight = srcTexture->height;
+    const uint32_t flippedSourceY =
+            srcTextureHeight - sourceViewport.bottom - sourceViewport.height;
+
+    auto const dstTexture = handleCast<WebGPUTexture>(destinationTextureHandle);
+    const uint32_t dstTextureHeight = dstTexture->height;
+    const uint32_t flippedDestinationY =
+            dstTextureHeight - destinationViewport.bottom - destinationViewport.height;
+
+    const wgpu::Origin2D sourceOrigin = { (uint32_t) sourceViewport.left, flippedSourceY };
```
AndyHovingh: please avoid C-style casts, using something like `static_cast<uint32_t>(...` instead.

filament/backend/src/webgpu/WebGPUDriver.cpp
```diff
@@ -1559,7 +1559,74 @@ void WebGPUDriver::blitDEPRECATED(TargetBufferFlags buffers,
...
+    if (mCommandEncoder) {
+        flush();
+    } else {
+        reusedCommandEncoder = false;
+        const wgpu::CommandEncoderDescriptor desc{ .label = "blit_deprecated_command" };
+        mCommandEncoder = mDevice.CreateCommandEncoder(&desc);
+    }
+
+    const WebGPUBlitter::BlitArgs blitArgs{
+        .source = { .texture = srcTexture->getTexture(), .aspect = srcTexture->getAspect(),
```
AndyHovingh: nit: formatting: separating the attributes on separate lines is easier to read, e.g.:
```cpp
.source = {
  .texture = sourceTexture->getTexture(),
  .aspect = sourceTexture->getAspect(),
  ...
},
...
```
If all the attributes are followed by commas, including the last one, then clang-format will allow it.

filament/backend/src/webgpu/WebGPUDriver.cpp
```diff
@@ -1559,7 +1559,74 @@ void WebGPUDriver::blitDEPRECATED(TargetBufferFlags buffers,
...
+                    .origin = sourceOrigin, .extent = sourceSize, .mipLevel = 0, .layerOrDepth = 0, },
```
AndyHovingh: can we safely assume mip level 0 and layer/depth 0? If both textures have the same number of mip levels and layers/depths could we not loop over each _(doing a blit for each)_? If they are not the same I assume we should panic right?

filament/backend/test/test_Blit.cpp
```diff
@@ -429,17 +430,18 @@ TEST_F(BlitTest, BlitRegion) {
...
     createBitmap(api, srcTexture, kSrcTexWidth, kSrcTexHeight, 0, float3(0.5, 0, 0), flipY);
     createBitmap(api, srcTexture, kSrcTexWidth, kSrcTexHeight, 1, float3(0, 0, 0.5), flipY);
 
     // Create a destination texture.
-    Handle<HwTexture> dstTexture = mCleanup.add(api.createTexture(
-            SamplerType::SAMPLER_2D, kNumLevels, kDstTexFormat, 1, kDstTexWidth, kDstTexHeight, 1,
-            TextureUsage::SAMPLEABLE | TextureUsage::COLOR_ATTACHMENT));
+    Handle<HwTexture> dstTexture = mCleanup.add(api.createTexture(SamplerType::SAMPLER_2D,
+            kNumLevels, kDstTexFormat, 1, kDstTexWidth, kDstTexHeight, 1,
+            TextureUsage::SAMPLEABLE | TextureUsage::COLOR_ATTACHMENT | TextureUsage::BLIT_DST));
```
AndyHovingh, line 444: does BlitRegionToSwapChain also need these TextureUsage::BLIT_X additions (even though we need to skip it for now)?

filament/backend/src/webgpu/WebGPUDriver.cpp
jc3265: I believe that they already are but can manually format if needed

AndyHovingh: Yes, the commas at the end of all of them doesn't necessarily direct clang format to format it one way or another, just the absence of the trailing comma prevents the line-by-line formatting. The comment is really more about that preferred format.

poweifeng (approved):
test_Blit changes lgtm

filament/backend/test/test_Blit.cpp
jc3265: possibly but its hard to tell, we get regardless
```
blitDEPRECATED could not find a valid color attachment to blit.
```

AndyHovingh: no biggie. If we re-enable the test is should be pretty obvious one way or another.

filament/backend/src/webgpu/WebGPUBlitter.cpp
```diff
@@ -677,11 +677,9 @@ wgpu::BindGroupLayout WebGPUBlitter::createTextureBindGroupLayout(const SamplerM
...
                 .sampleType = depthSource
                                       ? wgpu::TextureSampleType::Depth
-                                      : wgpu::TextureSampleType::Float, // only F32 scalar sample
-                                                                        // type supported for now
-                                                                        // (aside from depth)
-                                                                        // see assumptions listed
-                                                                        // in blit function
+                                      : (multisampledSource
+                                                ? wgpu::TextureSampleType::UnfilterableFloat
+                                                : wgpu::TextureSampleType::Float),
```
AndyHovingh: I would keep the comment about F32 scalar sample type support only, as this doesn't change that.

PR #9122, opened 2025-08-19
temporary test workaround for missing read_pixels texture usage

AndyHovingh:


filament/backend/test/Workarounds.h
```diff
@@ -0,0 +1,43 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef TNT_BACKEND_TEST_WORKAROUNDS_H
+#define TNT_BACKEND_TEST_WORKAROUNDS_H
+
+#if defined(FILAMENT_SUPPORTS_WEBGPU) || defined(FILAMENT_SUPPORTS_VULKAN)
```
jc3265: why vulkan?

AndyHovingh: I think Vulkan has the same problem with transient attachments. Please correct me if I am wrong @poweifeng

jc3265: tests seem to be passing of Vulkan though

bejado (approved):

filament/backend/test/Workarounds.h
AndyHovingh: You are right. I will remove it. It can be added later if needed.

PR #9126, opened 2025-08-20
webgpu: Remove texture logs

jc3265:


filament/backend/src/webgpu/WebGPUTextureHelpers.h
```diff
@@ -281,6 +219,12 @@ namespace filament::backend {
         case TextureFormat::DXT1_SRGBA:              return wgpu::TextureFormat::BC1RGBAUnormSrgb;
         case TextureFormat::DXT3_SRGBA:              return wgpu::TextureFormat::BC2RGBAUnormSrgb;
         case TextureFormat::DXT5_SRGBA:              return wgpu::TextureFormat::BC3RGBAUnormSrgb;
+        case TextureFormat::RGB565:
+        case TextureFormat::RGB5_A1:
+        case TextureFormat::RGBA4:
+            // No direct mapping in wgpu. Could potentially map to RGBA8Unorm
```
AndyHovingh: I think this case warrants a warning for visibility. I think we just want to remove the warnings about alpha.

PR #9128, opened 2025-08-20
vk: use firstBitSet API

poweifeng:
This is more efficient than the prevoius way

### DIFF
--- a/filament/backend/src/vulkan/VulkanQueryManager.cpp
+++ b/filament/backend/src/vulkan/VulkanQueryManager.cpp
@@ -46,17 +46,13 @@ fvkmemory::resource_ptr<VulkanTimerQuery> VulkanQueryManager::getNextQuery(
         return {};
     }
 
-    bool found = false;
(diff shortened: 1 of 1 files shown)


filament/backend/src/vulkan/VulkanQueryManager.cpp
```diff
@@ -46,17 +46,13 @@ fvkmemory::resource_ptr<VulkanTimerQuery> VulkanQueryManager::getNextQuery(
         return {};
     }
 
-    bool found = false;
     std::pair<uint32_t, uint32_t> queryIndices;
-    unused.forEachSetBit([&](size_t index) {
-        if (found) {
-            return;
-        }
+    size_t const firstUnused = unused.firstSetBit();
```
pixelflinger, line 50: you probably need to check fro the case where no bits are set?

poweifeng: so line 44 checks that `unused` is not `empty`

pixelflinger (approved):

PR #9129, opened 2025-08-20
wgpu: add texture swizzle support

AndyHovingh:
BUGS = [433944896]

### DIFF
--- a/filament/backend/src/webgpu/WebGPUDriver.cpp
+++ b/filament/backend/src/webgpu/WebGPUDriver.cpp
@@ -496,7 +496,40 @@ void WebGPUDriver::createTextureViewSwizzleR(Handle<HwTexture> textureHandle,
         Handle<HwTexture> sourceTextureHandle, const backend::TextureSwizzle r,
         const backend::TextureSwizzle g, const backend::TextureSwizzle b,
         const backend::TextureSwizzle a) {
-    PANIC_POSTCONDITION("Swizzle WebGPU Texture is not supported");
+
+    if (!isTextureSwizzleSupported()) {
+        FWGPU_LOGW << "WebGPUDriver::createTextureViewSwizzleR called while texture swizzling is "
+                      "not supported by the device/driver";
+    }
+
+    auto sourceTexture{ handleCast<WebGPUTexture>(sourceTextureHandle) };
+    assert_invariant(sourceTexture);
+
+    wgpu::TextureComponentSwizzle swizzle
+    {
+        .r = toWGPUComponentSwizzle(r),
+        .g = toWGPUComponentSwizzle(g),
+        .b = toWGPUComponentSwizzle(b),
+        .a = toWGPUComponentSwizzle(a),
+    };
+
+    wgpu::TextureComponentSwizzleDescriptor swizzleDesc {};
+    swizzleDesc.swizzle = swizzle;
+
+    const wgpu::TextureViewDescriptor viewDesc {
+        .nextInChain = &swizzleDesc,
+        .label = "swizzled_texture_view",
+        .format = sourceTexture->getTexture().GetFormat(),
+        .dimension = sourceTexture->getViewDimension(),
+        .baseMipLevel = 0,
+        .mipLevelCount = sourceTexture->getTexture().GetMipLevelCount(),
+        .baseArrayLayer = 0,
+        .arrayLayerCount = sourceTexture->getTexture().GetDepthOrArrayLayers(),
+    };
+
+    wgpu::TextureView swizzledView{ sourceTexture->getTexture().CreateView(&viewDesc) };
+    FILAMENT_CHECK_POSTCONDITION(swizzledView) << "Failed to create swizzled Texture view";
+    constructHandle<WebGPUTexture>(textureHandle, sourceTexture, swizzledView);
 }
 
 void WebGPUDriver::createTextureExternalImage2R(Handle<HwTexture> textureHandle,
@@ -641,7 +674,7 @@ bool WebGPUDriver::isTextureFormatSupported(const TextureFormat format) {
 }
 
 bool WebGPUDriver::isTextureSwizzleSupported() {
-    return false;
+    return mDevice.HasFeature(wgpu::FeatureName::TextureComponentSwizzle);
 }
 
 bool WebGPUDriver::isTextureFormatMipmappable(const TextureFormat format) {
--- a/filament/backend/src/webgpu/WebGPUTexture.h
+++ b/filament/backend/src/webgpu/WebGPUTexture.h
@@ -34,10 +34,25 @@ public:
         NONE,
     };
 
+    /**
+     * Creates a Filament texture and a texture view
+     */
     WebGPUTexture(SamplerType, uint8_t levels, TextureFormat, uint8_t samples, uint32_t width,
             uint32_t height, uint32_t depth, TextureUsage, wgpu::Device const&) noexcept;
 
+    /**
+     * Creates a "Filament Texture View", where the underlying texture is the same as the source,
+     * but the view elements differ
+     */
     WebGPUTexture(WebGPUTexture const* src, uint8_t baseLevel, uint8_t levelCount) noexcept;
+
+    /**
+     * @param textureView texture view to use for the respective texture. e.g. it can be a
+     *                    swizzled Texture view for that filament texture
+     *
+     * Associates the underlying texture with the given texture view
+     */
+    WebGPUTexture(WebGPUTexture const* source, wgpu::TextureView textureView) noexcept;
 
     [[nodiscard]] wgpu::TextureAspect getAspect() const { return mAspect; }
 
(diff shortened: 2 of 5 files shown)


AndyHovingh:
@jc3265 I would change the commit message to match the PR title

filament/backend/src/webgpu/WebGPUDriver.cpp
```diff
@@ -496,7 +496,32 @@ void WebGPUDriver::createTextureViewSwizzleR(Handle<HwTexture> textureHandle,
         Handle<HwTexture> sourceTextureHandle, const backend::TextureSwizzle r,
         const backend::TextureSwizzle g, const backend::TextureSwizzle b,
         const backend::TextureSwizzle a) {
-    PANIC_POSTCONDITION("Swizzle WebGPU Texture is not supported");
+
+    auto sourceTexture = handleCast<WebGPUTexture>(sourceTextureHandle);
```
AndyHovingh: nit: uniform initialization, e.g. `auto sourceTexture{ handleCast<WebGPUTexture>(sourceTextureHandle) };`

filament/backend/src/webgpu/WebGPUDriver.cpp
```diff
@@ -496,7 +496,32 @@ void WebGPUDriver::createTextureViewSwizzleR(Handle<HwTexture> textureHandle,
         Handle<HwTexture> sourceTextureHandle, const backend::TextureSwizzle r,
         const backend::TextureSwizzle g, const backend::TextureSwizzle b,
         const backend::TextureSwizzle a) {
-    PANIC_POSTCONDITION("Swizzle WebGPU Texture is not supported");
+
```
AndyHovingh, line 499: I would add a guard here, where this should not be getting called if texture swizzling is not supported, e.g.:
```cpp
FILAMENT_CHECK_PRECONDITION(isTextureSwizzleSupported()) << "Unexpected call to swizzle a texture when the backend/driver does not support it";
```

filament/backend/src/webgpu/WebGPUDriver.cpp
```diff
@@ -496,7 +496,32 @@ void WebGPUDriver::createTextureViewSwizzleR(Handle<HwTexture> textureHandle,
...
+    auto sourceTexture = handleCast<WebGPUTexture>(sourceTextureHandle);
+    assert_invariant(sourceTexture);
+
+    wgpu::TextureComponentSwizzle swizzle;
```
AndyHovingh: designated initializer?:
```cpp
const wgpu::TextureComponentSwizzle swizzle{
    .r = toWGPUComponentSwizzle(r),
    .g = toWGPUComponentSwizzle(g),
    .b = toWGPUComponentSwizzle(b),
    .a = toWGPUComponentSwizzle(a),
};
```

filament/backend/src/webgpu/WebGPUDriver.cpp
```diff
@@ -496,7 +496,32 @@ void WebGPUDriver::createTextureViewSwizzleR(Handle<HwTexture> textureHandle,
...
+    swizzle.r = toWGPUComponentSwizzle(r);
+    swizzle.g = toWGPUComponentSwizzle(g);
+    swizzle.b = toWGPUComponentSwizzle(b);
+    swizzle.a = toWGPUComponentSwizzle(a);
+
+    wgpu::TextureComponentSwizzleDescriptor swizzleDesc {};
+    swizzleDesc.swizzle = swizzle;
+
+    wgpu::TextureViewDescriptor viewDesc = {
```
AndyHovingh: can this be `const`? The `=` is not needed.

filament/backend/src/webgpu/WebGPUDriver.cpp
```diff
@@ -496,7 +496,32 @@ void WebGPUDriver::createTextureViewSwizzleR(Handle<HwTexture> textureHandle,
...
+        .label = "swizzled_texture_view",
+        .format = sourceTexture->getTexture().GetFormat(),
+        .dimension = sourceTexture->getViewDimension(),
+        .baseMipLevel = 0,
+        .mipLevelCount = sourceTexture->getTexture().GetMipLevelCount(),
+        .baseArrayLayer = 0,
+        .arrayLayerCount = sourceTexture->getTexture().GetDepthOrArrayLayers(),
+    };
+
+    wgpu::TextureView swizzledView = sourceTexture->getTexture().CreateView(&viewDesc);
```
AndyHovingh: nit: `const` and uniform initialization, e.g. `const wgpu::TextureView swizzledView{ sourceTexture->getTexture().CreateView(&viewDesc) };`.

We should also check that the view was created properly right? FILAMENT_CHECK_POSTCONDITION?

filament/backend/src/webgpu/WebGPUDriver.cpp
```diff
@@ -496,7 +496,32 @@ void WebGPUDriver::createTextureViewSwizzleR(Handle<HwTexture> textureHandle,
...
+    constructHandle<WebGPUTexture>(textureHandle, sourceTexture, (swizzledView));
```
AndyHovingh: why the parentheses around `swizzledView`?

filament/backend/src/webgpu/WebGPUDriver.cpp
```diff
@@ -641,6 +666,9 @@ bool WebGPUDriver::isTextureFormatSupported(const TextureFormat format) {
 }
 
 bool WebGPUDriver::isTextureSwizzleSupported() {
+    if (mDevice.HasFeature(wgpu::FeatureName::TextureComponentSwizzle)) {
```
AndyHovingh: perhaps more concisely?: `return mDevice.HasFeature(wgpu::FeatureName::TextureComponentSwizzle);`

filament/backend/src/webgpu/WebGPUTexture.h
```diff
@@ -39,6 +39,8 @@ class WebGPUTexture : public HwTexture {
 
     WebGPUTexture(WebGPUTexture const* src, uint8_t baseLevel, uint8_t levelCount) noexcept;
```
AndyHovingh, line 47: I think we could add a comment on this constructor calling it out as a "Filament Texture View", where the underlying texture is the same as the source, but the view elements differ. Similarly, we could comment on the new constructor as the "Filament Swizzle Texture View".

filament/backend/src/webgpu/WebGPUTextureHelpers.h
```diff
@@ -295,6 +295,18 @@ namespace filament::backend {
     }
 }
 
+[[nodiscard]] constexpr wgpu::ComponentSwizzle toWGPUComponentSwizzle(backend::TextureSwizzle swizzle) {
```
AndyHovingh: nit: `const backend::TextureSwizzle swizzle`

filament/backend/src/webgpu/WebGPUTextureHelpers.h
```diff
@@ -295,6 +295,18 @@ namespace filament::backend {
...
+    switch (swizzle) {
+        case backend::TextureSwizzle::CHANNEL_0:       return wgpu::ComponentSwizzle::R;
+        case backend::TextureSwizzle::CHANNEL_1:       return wgpu::ComponentSwizzle::G;
+        case backend::TextureSwizzle::CHANNEL_2:       return wgpu::ComponentSwizzle::B;
+        case backend::TextureSwizzle::CHANNEL_3:       return wgpu::ComponentSwizzle::A;
+        case backend::TextureSwizzle::SUBSTITUTE_ZERO: return wgpu::ComponentSwizzle::Zero;
+        case backend::TextureSwizzle::SUBSTITUTE_ONE:  return wgpu::ComponentSwizzle::One;
+        default:                                       return wgpu ::ComponentSwizzle::Undefined;
```
AndyHovingh: we should probably panic or warn in this case, yeah?

filament/backend/src/webgpu/WebGPUDriver.cpp
AndyHovingh: If this causes things to blow up, we could instead log a warning and return early. But, I would start with a panic and test out the samples and unit tests first.

filament/backend/src/webgpu/WebGPUDriver.cpp
jc3265: removed

filament/backend/src/webgpu/WebGPUDriver.cpp
jc3265: yeah the `gltf_viewer`'s batch test blows up

AndyHovingh: ok. Then, let's warn and return early 👍

filament/backend/src/webgpu/WebGPUTexture.h
jc3265: added comments

filament/backend/src/webgpu/WebGPUDriver.cpp
jc3265: not returning early since filament is still calling the function even when the feature is not supported

AndyHovingh:
@jc3265 lgtm, but I can't approve it since I created the PR for you. You can approve it for me :) if you would like 👍

PR #9130, opened 2025-08-21
backend: use VirtualAlloc for creating soft CircularBuffer on Windows

yrom:
Use `VirtualAlloc` for virtual buffer allocation and `VirtualProtect` to set a guard page on Windows, aligning the behavior of `mmap` + `mprotect`.


filament/backend/src/CircularBuffer.cpp
```diff
@@ -150,6 +170,10 @@ void CircularBuffer::dealloc() noexcept {
             mAshmemFd = -1;
         }
     }
+#elif defined(WIN32)
+    if (mData) {
+        VirtualFree(mData, 0, MEM_RELEASE);
```
pixelflinger, line 175: I'm unfamiliar with windows APIs, so wanted to make sure that we don't need to "unprotect" the guard page in addition to virtualfree?

z3moon:
The change itself looks good to me. Though I think we can improve it even further by implementing the mirrored memory approach as well.

z3moon (approved):

PR #9137, opened 2025-08-22
webgpu: Support skinning on WebGPU

jc3265:
BUGS = [436886048]

### DIFF
--- a/samples/helloskinningbuffer.cpp
+++ b/samples/helloskinningbuffer.cpp
@@ -174,6 +174,7 @@ int main(int argc, char** argv) {
                 .attribute(VertexAttribute::POSITION, 0, VertexBuffer::AttributeType::FLOAT2, 0, 12)
                 .attribute(VertexAttribute::COLOR, 0, VertexBuffer::AttributeType::UBYTE4, 8, 12)
                 .normalized(VertexAttribute::COLOR)
+                .advancedSkinning(true)
                 .build(*engine);
         app.vb2->setBufferAt(*engine, 0,
                 VertexBuffer::BufferDescriptor(TRIANGLE_VERTICES, 36, nullptr));


libs/filamat/src/shaders/CodeGenerator.cpp
```diff
@@ -966,6 +962,20 @@ utils::io::sstream& CodeGenerator::generatePushConstants(utils::io::sstream& out
         MaterialBuilder::PushConstantList const& pushConstants, size_t const layoutLocation) const {
     static constexpr char const* STRUCT_NAME = "Constants";
 
+    // This is a workaround for WebGPU not supporting push constants for skinning.
+    // We replace the push constant with a regular constant struct initialized to 0.
+    if (mTargetApi == TargetApi::WEBGPU) {
+        out << "struct " << STRUCT_NAME << " {\n";
+        for (auto const& constant: pushConstants) {
+            out << "    int " << constant.name.c_str() << ";\n";
```
AndyHovingh: We should not assume an int. See `constant.type`, which is an enum of either INT, FLOAT, BOOL. This is handled below with the `getType` lambda, which could be reused here. I know we expect one constant and it to be an int with value 0, but we can make the code a little easier to adjust in the event that spec constants become supported.

libs/filamat/src/shaders/CodeGenerator.cpp
```diff
@@ -966,6 +962,20 @@ utils::io::sstream& CodeGenerator::generatePushConstants(utils::io::sstream& out
...
+        }
+        out << "};\n";
+        out << "const " << STRUCT_NAME << " " << PUSH_CONSTANT_STRUCT_VAR_NAME << " = "
+            << STRUCT_NAME << "(0);\n";
```
AndyHovingh, line 996: I recommend an assert that the `pushConstants` has one value with the one name we expect. Otherwise, things/assumptions have changed and this part of the code should fail loudly. Also, it makes the hack here more obvious (in addition to your comment).

jc3265: i dont get this, what should we assert?

AndyHovingh: We are assuming that the pushConstants are just the one that we should hardcode to 0. We should probably assert that it is indeed just the one we are expecting to get. That assumption could change (in the future maybe multiple push constants are requested...).

libs/filamat/src/shaders/CodeGenerator.cpp
```diff
@@ -966,8 +962,6 @@ utils::io::sstream& CodeGenerator::generatePushConstants(utils::io::sstream& out
         MaterialBuilder::PushConstantList const& pushConstants, size_t const layoutLocation) const {
     static constexpr char const* STRUCT_NAME = "Constants";
```
AndyHovingh, line 967: I realize this wasn't here before, but should we consider the case where there are no push constants being requested for some reason (ever), maybe for code defensiveness? e.g.:
```cpp
if (UTILS_UNLIKELY(pushConstants.empty())) {
    return out;
}
```

libs/filamat/src/shaders/CodeGenerator.cpp
AndyHovingh: perhaps something like:
```cpp
assert_invariant(
        pushConstants.size() == 1 &&
        "The current workaround for WebGPU push constants assumes for now that only 1");
assert_invariant(pushConstants[0].name == CString("morphingBufferOffset") &&
                 "The current workaround for WebGPU push constants assumes only the "
                 "morphingBufferOffset constant is present.");
assert_invariant(pushConstants[0].type == ConstantType::INT &&
                 "The current workaround for WebGPU push constants assumes "
                 "morphingBufferOffset is an integer type.");
```

filament/backend/src/webgpu/WebGPUPipelineCache.cpp
```diff
@@ -177,6 +177,12 @@ void WebGPUPipelineCache::populateKey(RenderPipelineRequest const& request,
     outKey.targetRenderFlags = request.targetRenderFlags;
     outKey.multisampleCount = request.multisampleCount;
     outKey.depthStencilFormat = request.depthStencilFormat;
+    if (request.primitiveType == PrimitiveType::TRIANGLE_STRIP ||
```
AndyHovingh: I suggest a constexpr helper function, probably defined in DriverEnums.h, e.g.:
```cpp
[[nodiscard]] constexpr bool isStripType(const PrimitiveType type) {
    switch (type) {
        case PrimitiveType::POINTS:
        case PrimitiveType::LINES:
        case PrimitiveType::TRIANGLES:
            return false;
        case PrimitiveType::LINE_STRIP:
        case PrimitiveType::TRIANGLE_STRIP:
            return true;
    }
}
```

then:
```cpp
outKey.stripIndexFormat = isStripType(request.primitiveType) ? request.stripIndexFormat
                                                             : wgpu::IndexFormat::Undefined;
```

filament/backend/src/webgpu/WebGPUPipelineCache.cpp
```diff
@@ -312,11 +318,11 @@ wgpu::RenderPipeline WebGPUPipelineCache::createRenderPipeline(
         },
         .fragment = nullptr // will add below if fragment module is included
     };
-    // TODO:
     if (pipelineDescriptor.primitive.topology == wgpu::PrimitiveTopology::LineStrip ||
```
AndyHovingh: again, I would recommend a helper function, e.g.:
```cpp
[[nodiscard]] constexpr bool isStripTopology(const wgpu::PrimitiveTopology topology) {
    switch (topology) {
        case wgpu::PrimitiveTopology::Undefined:
        case wgpu::PrimitiveTopology::PointList:
        case wgpu::PrimitiveTopology::LineList:
        case wgpu::PrimitiveTopology::TriangleList:
            return false;
        case wgpu::PrimitiveTopology::LineStrip:
        case wgpu::PrimitiveTopology::TriangleStrip:
            return true;
    }
}
```
then this could be simplified to:
```cpp
FILAMENT_CHECK_POSTCONDITION(!isStripTopology(pipelineDescriptor.primitive.topology) ||
                             request.stripIndexFormat != wgpu::IndexFormat::Undefined)
        << "If the topology is a strip format, e.g. TriangleStrip, the stripIndexFormat cannot "
           "be Undefined.";
```
right?

filament/backend/src/webgpu/WebGPUPipelineCache.cpp
```diff
@@ -312,11 +318,11 @@ wgpu::RenderPipeline WebGPUPipelineCache::createRenderPipeline(
...
             pipelineDescriptor.primitive.topology == wgpu::PrimitiveTopology::TriangleStrip) {
-        PANIC_POSTCONDITION("stripIndexFormat must be set for strip topologies. "
-                            "This needs to be plumbed through from the RenderPrimitive.");
+        pipelineDescriptor.primitive.stripIndexFormat = request.stripIndexFormat;
+        FILAMENT_CHECK_POSTCONDITION(request.stripIndexFormat != wgpu::IndexFormat::Undefined)
+                << "Invalid format";
```
AndyHovingh: Wouldn't it be cleaner to set the stripIndexFormat directly above in the designated initializer list and do the FILAMENT_CHECK before or after that?

```cpp
...
.primitive = {
    .topology = toWebGPU(request.primitiveType),
    .stripIndexFormat = request.stripIndexFormat,
...
```

filament/backend/src/webgpu/WebGPUDriver.cpp
```diff
@@ -569,6 +569,7 @@ void WebGPUDriver::createRenderPrimitiveR(Handle<HwRenderPrimitive> renderPrimit
     renderPrimitive->vertexBuffer = vertexBuffer;
     renderPrimitive->indexBuffer = indexBuffer;
     renderPrimitive->type = primitiveType;
+    mCurrentRenderPrimitive = renderPrimitive;
```
AndyHovingh: Are we sure we are working with only one render primitive at a time? And, are we sure that this will always be called prior to calling bindPipeline? And, are we sure no other primitives could be created in between that may not have anything to do with the render primitive(s) being drawn with the pipeline being bound by Filament later? There seems to be a lot of assumptions here that do not feel right; are we sure this is a robust way to get this?

AndyHovingh: @poweifeng @pixelflinger @bejado any idea how we could plumb the relevant index buffer info, namely the index format, to the bindPipeline call?

libs/filamat/src/shaders/CodeGenerator.cpp
jc3265: not too sure if we need this one but maybe 
@poweifeng or @pixelflinger can chime in?

poweifeng: I think it's ok to keep this.  The concern is that this variable is referenced in the shader, but the path will never be taken unless we do morphing.

PR #9139, opened 2025-08-22
vk: check rebind flow if bindDescriptorSet called between draw2

poweifeng:
In the case with external samplers, if another descriptor set gets bound between draw2(), then we need to check to see if a rebinding of the pipeline is necessary.

### DIFF
--- a/filament/backend/src/vulkan/VulkanDriver.cpp
+++ b/filament/backend/src/vulkan/VulkanDriver.cpp
@@ -452,7 +452,7 @@ void VulkanDriver::updateDescriptorSetTexture(
     auto set = resource_ptr<VulkanDescriptorSet>::cast(&mResourceManager, dsh);
     auto texture = resource_ptr<VulkanTexture>::cast(&mResourceManager, th);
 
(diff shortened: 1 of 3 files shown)


filament/backend/src/vulkan/VulkanDriver.cpp
```diff
@@ -1939,6 +1939,18 @@ void VulkanDriver::bindDescriptorSet(
     if (dsh) {
         auto set = resource_ptr<VulkanDescriptorSet>::cast(&mResourceManager, dsh);
         mDescriptorSetCache.bind(setIndex, set, std::move(offsets));
+
+        if (mAppState.hasExternalSamplers()) {
```
pixelflinger, line 1943: unlikely?

pixelflinger (approved):

filament/backend/src/vulkan/VulkanDriver.cpp
poweifeng: it's a tough call on this one.  On Android I think it's actually likely, but on unlikely on others.

PR #9147, opened 2025-08-25
Update all per-object/instance buffers all at once 

pixelflinger:
FIXES=[433807402]

### DIFF
--- a/filament/src/details/InstanceBuffer.cpp
+++ b/filament/src/details/InstanceBuffer.cpp
@@ -29,6 +29,8 @@
 #include <backend/DriverEnums.h>
 #include <backend/Handle.h>
 
+#include <utils/compiler.h>
+#include <utils/debug.h>
 #include <utils/Panic.h>
 #include <utils/StaticString.h>
 
@@ -87,7 +89,7 @@ InstanceBuffer* InstanceBuffer::Builder::build(Engine& engine) const {
 
 // ------------------------------------------------------------------------------------------------
 
-FInstanceBuffer::FInstanceBuffer(FEngine& engine, const Builder& builder)
+FInstanceBuffer::FInstanceBuffer(FEngine&, const Builder& builder)
     : mName(builder.getName()) {
     mInstanceCount = builder->mInstanceCount;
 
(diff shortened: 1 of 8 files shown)


filament/src/details/InstanceBuffer.cpp
```diff
@@ -127,28 +116,22 @@ void FInstanceBuffer::setLocalTransforms(
     memcpy(mLocalTransforms.data() + offset, localTransforms, sizeof(math::mat4f) * count);
 }
 
-void FInstanceBuffer::prepare(FEngine& engine, math::mat4f const& rootTransform,
-        const PerRenderableData& ubo) {
-    DriverApi& driver = engine.getDriverApi();
+void FInstanceBuffer::prepare(
+            BufferObjectHandle ubh,
+            PerRenderableData* const UTILS_RESTRICT buffer, uint32_t const offset, uint32_t const count,
```
poweifeng, line 123: why do you need `count` if it's the same as `mInstanceCount` ?

filament/src/details/InstanceBuffer.h
```diff
@@ -48,11 +50,15 @@ class FInstanceBuffer : public InstanceBuffer {
 
     void setLocalTransforms(math::mat4f const* localTransforms, size_t count, size_t offset);
 
-    void prepare(FEngine& engine, math::mat4f const& rootTransform, const PerRenderableData& ubo);
+    void prepare(
+            backend::BufferObjectHandle ubh,
+            PerRenderableData* buffer, uint32_t offset, uint32_t count,
+            math::mat4f const& rootTransform, const PerRenderableData& ubo);
```
poweifeng: nit: `PerRenderableData const& ubo` for consistency

filament/src/details/View.cpp
```diff
@@ -916,6 +911,100 @@ void FView::computeVisibilityMasks(
...
+        FScene::RenderableSoa& renderableData,
+        utils::Range<uint32_t> visibleRenderables) noexcept {
+    FILAMENT_TRACING_CALL(FILAMENT_TRACING_CATEGORY_FILAMENT);
+
+    FRenderableManager::InstancesInfo const* instancesData = renderableData.data<FScene::INSTANCES>();
+    PerRenderableData const* const uboData = renderableData.data<FScene::UBO>();
+    mat4f const* const worldTransformData = renderableData.data<FScene::WORLD_TRANSFORM>();
+
+    // regular renderables count
+    size_t rcount = visibleRenderables.size();
```
poweifeng: nit const

filament/src/details/InstanceBuffer.cpp
pixelflinger: these basically come from two different places. On one side the user tells use how many instance they wants, on the other side, they give us a buffer -- ~that buffer better match. Currently there is an assert, but it probably should be a PRECONDITION check~
The doc says that the instance buffer mush be at least big enough. So the assert was wrong.

filament/src/details/InstanceBuffer.h
pixelflinger: done

filament/src/details/View.cpp
pixelflinger: done

filament/src/details/InstanceBuffer.cpp
```diff
@@ -127,28 +118,23 @@ void FInstanceBuffer::setLocalTransforms(
...
+            math::mat4f const& rootTransform, PerRenderableData const& ubo) {
+
+    // there is a precondition check for this, so this assert really should never trigger
+    assert_invariant(count <= mInstanceCount);
 
-    // TODO: allocate this staging buffer from a pool.
-    constexpr uint32_t stagingBufferSize = sizeof(PerRenderableUib);
-    PerRenderableData* stagingBuffer = static_cast<PerRenderableData*>(malloc(stagingBufferSize));
-    // TODO: consider using JobSystem to parallelize this.
     for (size_t i = 0, c = mInstanceCount; i < c; i++) {
```
poweifeng: should this then be `c = count` ?

pixelflinger: good catch, thanks.

PR #9149, opened 2025-08-26
webgpu: Handle stacked swizzles

jc3265:
BUGS = [433944896]


filament/backend/src/webgpu/WebGPUTexture.h
```diff
@@ -47,12 +47,10 @@ class WebGPUTexture : public HwTexture {
     WebGPUTexture(WebGPUTexture const* src, uint8_t baseLevel, uint8_t levelCount) noexcept;
 
     /**
-     * @param textureView texture view to use for the respective texture. e.g. it can be a
-     *                    swizzled Texture view for that filament texture
-     *
-     * Associates the underlying texture with the given texture view
+     * @param nextSwizzle the swizzle to apply to the source texture.
+     * Creates a new texture view by composing the nextSwizzle with the source's swizzle.
```
AndyHovingh: nit: (assuming the source is itself a swizzle view. the source could be a regular view or an actual texture)

filament/backend/src/webgpu/WebGPUTexture.h
```diff
@@ -144,6 +138,7 @@ class WebGPUTexture : public HwTexture {
     // by sampleCount or something like that.
     // For now that complexity and cost is not warranted due to WebGPU's restrictions.
     wgpu::Texture mMsaaSidecarTexture = nullptr;
+    wgpu::TextureComponentSwizzle mSwizzle{};
```
AndyHovingh, line 141: This instance of WebGPUTexture being a swizzle view or not doesn't seem very clear. A comment here suggesting that the swizzle parameters being undefined (default) implies this is NOT a swizzle view might be helpful. Alternatively, you could consider expressing that with `std::optional`. This applies to the corresponding getter. Also, could this be `const`, assigned at constructor time? Would it ever reasonably change?

filament/backend/src/webgpu/WebGPUTexture.cpp
```diff
@@ -41,6 +41,38 @@ namespace filament::backend {
 
 namespace {
 
+wgpu::TextureComponentSwizzle composeSwizzle(wgpu::TextureComponentSwizzle const& prev,
```
AndyHovingh: nit: `[[nodiscard]]`

filament/backend/src/webgpu/WebGPUTexture.cpp
```diff
@@ -41,6 +41,38 @@ namespace filament::backend {
...
+        wgpu::TextureComponentSwizzle const& next) {
+    auto const compose =
```
AndyHovingh: why is this extra lambda necessary?

filament/backend/src/webgpu/WebGPUTexture.cpp
```diff
@@ -298,8 +333,20 @@ WebGPUTexture::WebGPUTexture(const WebGPUTexture* src, const wgpu::TextureView v
       mTexture{ src->mTexture },
       mDefaultMipLevel{ src->mDefaultMipLevel },
       mDefaultBaseArrayLayer{ 0 },
-      mDefaultTextureView{ view },
-      mMsaaSidecarTexture{src->mMsaaSidecarTexture}{}
+      mMsaaSidecarTexture{src->mMsaaSidecarTexture},
+      mSwizzle{ composeSwizzle(src->getSwizzle(), nextSwizzle) } {
```
AndyHovingh, line 329: I can see here that compose should work fine if the source sizzle is undefined (not a sizzle view) as well as if it is. A comment to that effect might help someone working with the code (or if `std::optional` or something is used that becomes more explicit).

filament/backend/src/webgpu/WebGPUTexture.cpp
```diff
@@ -281,9 +314,11 @@ WebGPUTexture::WebGPUTexture(WebGPUTexture const* src, const uint8_t baseLevel,
       mDefaultBaseArrayLayer{ src->mArrayLayerCount },
       mDefaultTextureView{ makeTextureView(mDefaultMipLevel, levelCount, 0, mDefaultBaseArrayLayer,
               toWebGPUTextureViewDimension(src->target)) },
-      mMsaaSidecarTexture{ src->mMsaaSidecarTexture } {}
+      mMsaaSidecarTexture{ src->mMsaaSidecarTexture },
+      mSwizzle{}{}
```
AndyHovingh: i.e. not a swizzle texture... but what if the source texture was? If the source texture was a swizzle texture shouldn't we retain the source texture's swizzle parameters?

filament/backend/src/webgpu/WebGPUTexture.cpp
```diff
@@ -216,7 +248,8 @@ WebGPUTexture::WebGPUTexture(const SamplerType samplerType, const uint8_t levels
       mBlockWidth{ filament::backend::getBlockWidth(format) },
       mBlockHeight{ filament::backend::getBlockHeight(format) },
       mDefaultMipLevel{ 0 },
-      mDefaultBaseArrayLayer{ 0 } {
+      mDefaultBaseArrayLayer{ 0 },
+      mSwizzle{} {
```
AndyHovingh, line 221: i.e. not a swizzle view (maybe a comment can clarify this point)

filament/backend/src/webgpu/WebGPUTexture.cpp
```diff
@@ -41,6 +41,38 @@ namespace filament::backend {
...
+wgpu::TextureComponentSwizzle composeSwizzle(wgpu::TextureComponentSwizzle const& prev,
+        wgpu::TextureComponentSwizzle const& next) {
+    auto const compose =
+            [](wgpu::TextureComponentSwizzle const& p,
+                    wgpu::TextureComponentSwizzle const& n) -> wgpu::TextureComponentSwizzle {
+        wgpu::ComponentSwizzle vals[4] = { n.r, n.g, n.b, n.a };
+        for (auto& out: vals) {
+            switch (out) {
+                case wgpu::ComponentSwizzle::R:
+                    out = p.r;
```
AndyHovingh: what if previous was undefined? should we then use the next's R mapping?

PR #9152, opened 2025-08-27
fix: Make ZstdHelper compatible with UBSan

z3moon:
In ZstdHelper::isCompressed call, `src` may not be aligned to 4 bytes, which is violating the alignment requirement of
`UndefinedBehaviorSanitizer: misaligned-pointer-use`, thereby resuling in a runtime failure with UBSan enabled. So reconstruct the 32-bit integer as a workaround.


poweifeng (approved):
lgtm for now to unblock, but @pixelflinger ptal when you have a chane.

PR #9164, opened 2025-08-29
Stream API Vulkan Backend

phoenixxxx:
This is the implemenation of the Stream API on the Vulkan Backend. We leverage most of the work that was done by the external image manager, but we created a different manager to contain the code.

Tested on Moohan (manual merge, so some things could have gotten lost in the process). Also working on validating this on Android.

### DIFF
--- a/filament/backend/src/opengl/OpenGLContext.h
+++ b/filament/backend/src/opengl/OpenGLContext.h
@@ -735,7 +735,7 @@ void OpenGLContext::bindVertexArray(RenderPrimitive const* p) noexcept {
         // - the nameVersion is out of date *and* we're on the protected context, in this case:
         //      - the name must be stale from a previous use of this context because we always
         //        destroy the protected context when we're done with it.
-        bool const recreateVaoName = p != &mDefaultVAO &&
+        bool const recreateVaoName = vao != &mDefaultVAO &&
                 ((vao->vao[contextIndex] == 0) ||
                         (vao->nameVersion != state.age && contextIndex == 1));
         if (UTILS_UNLIKELY(recreateVaoName)) {
--- a/filament/backend/src/vulkan/VulkanPipelineCache.cpp
+++ b/filament/backend/src/vulkan/VulkanPipelineCache.cpp
@@ -191,9 +191,12 @@ VulkanPipelineCache::PipelineCacheEntry* VulkanPipelineCache::createPipeline() n
         .alphaToCoverageEnable = raster.alphaToCoverageEnable,
         .alphaToOneEnable = VK_FALSE,
     };
+    bool const enableDepthTest =
+        raster.depthCompareOp != SamplerCompareFunc::A ||
+        raster.depthWriteEnable;
     VkPipelineDepthStencilStateCreateInfo vkDs = {
         .sType = VK_STRUCTURE_TYPE_PIPELINE_DEPTH_STENCIL_STATE_CREATE_INFO,
-        .depthTestEnable = VK_TRUE,
+        .depthTestEnable = enableDepthTest ? VK_TRUE : VK_FALSE,
         .depthWriteEnable = raster.depthWriteEnable,
         .depthCompareOp = fvkutils::getCompareOp(raster.depthCompareOp),
         .depthBoundsTestEnable = VK_FALSE,
--- a/libs/filabridge/include/filament/MaterialEnums.h
+++ b/libs/filabridge/include/filament/MaterialEnums.h
@@ -28,7 +28,7 @@
 namespace filament {
 
 // update this when a new version of filament wouldn't work with older materials
-static constexpr size_t MATERIAL_VERSION = 67;
+static constexpr size_t MATERIAL_VERSION = 68;
 
 /**
  * Supported shading models


filament/backend/src/vulkan/VulkanStreamedImageManager.h
```diff
@@ -0,0 +1,68 @@
...
+#include <array>
+#include <vector>
+
+namespace filament::backend {
+
+class VulkanExternalImageManager;
+class VulkanDescriptorSetCache;
+class VulkanSamplerCache;
+
+// Manages the logic of streamed and streamed images.
```
poweifeng:
```suggestion
// Manages the logic of streamed images.
```

filament/backend/src/vulkan/VulkanStreamedImageManager.h
```diff
@@ -0,0 +1,68 @@
...
+public:
+    void bindStreamedTexture(fvkmemory::resource_ptr<VulkanDescriptorSet> set, uint8_t bindingPoint,
+            fvkmemory::resource_ptr<VulkanTexture> image, SamplerParams samplerParams);
+    void unbindStreamedTexture(fvkmemory::resource_ptr<VulkanDescriptorSet> set,
+            uint8_t bindingPoint);
+    void onStreamAcquireImage(fvkmemory::resource_ptr<VulkanTexture> image,
+            fvkmemory::resource_ptr<VulkanStream> stream, bool newImage);
+
+private:
+    struct streamedTextureBinding {
```
poweifeng: Should capitalize `Streamed`

filament/backend/include/backend/platforms/VulkanPlatformAndroid.h
```diff
@@ -28,6 +28,8 @@ class VulkanPlatformAndroid : public VulkanPlatform {
 public:
     ExternalImageHandle UTILS_PUBLIC createExternalImage(AHardwareBuffer const* buffer,
             bool sRGB) noexcept;
+    virtual ExternalImageHandle createExternalImageFromRaw(void* image,
+            bool sRGB) noexcept override;
```
poweifeng: I think we should just combine these methods into one called `createExternalImage(void* image, bool sRGB)`

filament/backend/src/vulkan/VulkanDriver.cpp
```diff
@@ -455,13 +456,18 @@ void VulkanDriver::updateDescriptorSetTexture(
     if (UTILS_UNLIKELY(mExternalImageManager.isExternallySampledTexture(texture))) {
         mExternalImageManager.bindExternallySampledTexture(set, binding, texture, params);
         mAppState.hasBoundExternalImages = true;
-    } else {
+    } else if (bool(texture->getStream())) {
+        mStreamedImageManager.bindStreamedTexture(set, binding, texture, params);
+        mAppState.hasBoundExternalImages = true;
+    }
+    else {
```
poweifeng:
```suggestion
    } else {
```

filament/backend/src/vulkan/VulkanDriver.cpp
```diff
@@ -991,25 +1002,99 @@ void VulkanDriver::destroyDescriptorSet(Handle<HwDescriptorSet> dsh) {
 }
 
 Handle<HwStream> VulkanDriver::createStreamNative(void* nativeStream) {
-    return {};
+    auto handle = mResourceManager.allocHandle<VulkanStream>();
+    auto stream = resource_ptr<VulkanStream>::make(&mResourceManager, handle);
+    stream.inc();
+    return handle;
```
poweifeng: Can you add an assert ( a filament panic) to say that this is not supported on Vulkan? 

The reason is that this is meant to be used with `SurfaceTexture` methods specifically designed for EGL.

filament/backend/src/vulkan/VulkanDriver.cpp
```diff
@@ -991,25 +1002,99 @@ void VulkanDriver::destroyDescriptorSet(Handle<HwDescriptorSet> dsh) {
...
 }
 
 void VulkanDriver::setAcquiredImage(Handle<HwStream> sh, void* image, const math::mat3f& transform,
         CallbackHandler* handler, StreamCallback cb, void* userData) {
+    FVK_SYSTRACE_SCOPE();
+    auto stream = resource_ptr<VulkanStream>::cast(&mResourceManager, sh);
+    stream->acquire({ image, cb, userData, handler });
+    mStreamsWithPendingAcquiredImage.push_back(stream);
```
poweifeng, line 1210: This is a synchronous call. I think you need a lock around this vector.

filament/backend/src/vulkan/VulkanTexture.h
```diff
@@ -36,6 +36,34 @@ namespace filament::backend {
 
 struct VulkanTexture;
 
+struct VulkanStream : public HwStream, fvkmemory::Resource {
```
poweifeng, line 39: This should be `fvkmemory::ThreadSafeResource`

filament/backend/src/vulkan/VulkanTexture.h
```diff
@@ -36,6 +36,34 @@ namespace filament::backend {
...
+        AcquiredImage previous = mPrevious;
+        mPrevious = { nullptr, nullptr, nullptr, nullptr };
+        return previous;
+    }
+    const AcquiredImage& getAcquired() const { return mAcquired; }
+
+private:
+    AcquiredImage mAcquired;
+    AcquiredImage mPrevious;
+    std::map<void*, fvkmemory::resource_ptr<VulkanTexture>> mTextures;
```
poweifeng: ThreadSafe resources can't have any `resource_ptr` fields unfortunately.  Maybe you can keep this map in `VulkanStreamedImageManager`

filament/backend/src/vulkan/VulkanTexture.h
```diff
@@ -182,6 +213,11 @@ struct VulkanTexture : public HwTexture, fvkmemory::Resource {
         return mState->mSidecarMSAA;
     }
 
+    void setStream(fvkmemory::resource_ptr<VulkanStream> stream) { mState->mStream = stream;
+    }
```
poweifeng: formatting here.

filament/backend/src/vulkan/VulkanDriver.cpp
```diff
@@ -991,25 +1002,99 @@ void VulkanDriver::destroyDescriptorSet(Handle<HwDescriptorSet> dsh) {
...
+    stream->height = height;
 }
 
 int64_t VulkanDriver::getStreamTimestamp(Handle<HwStream> sh) {
     return 0;
 }
 
 void VulkanDriver::updateStreams(CommandStream* driver) {
+    if (UTILS_UNLIKELY(!mStreamsWithPendingAcquiredImage.empty())) {
+        for (auto& stream: mStreamsWithPendingAcquiredImage) {
```
poweifeng, line 1227: This is a synchonous call.  You'll need a lock for `mStreamsWithPendingAcquiredImage`. and the loop below should be posted onto the driver thread as in [here](https://github.com/google/filament/blob/main/filament/backend/src/opengl/OpenGLDriver.cpp#L2143)

filament/backend/src/vulkan/VulkanDriver.cpp
```diff
@@ -991,25 +1002,99 @@ void VulkanDriver::destroyDescriptorSet(Handle<HwDescriptorSet> dsh) {
...
+                        metadata.height, metadata.layers, metadata.filamentUsage, mStagePool);
+
+                // We shouldn't need to do this when we allocate the texture
+                if (true) {
+                    auto& commands = mCommands.get();
+                    // Unlike uploaded textures or swapchains, we need to explicit transition this
+                    // texture into the read layout.
+                    newTexture->transitionLayout(&commands, newTexture->getPrimaryViewRange(),
+                            VulkanLayout::FRAG_READ);
+                }
```
poweifeng: why `(true)`?

filament/backend/src/vulkan/VulkanStreamedImageManager.cpp
```diff
@@ -0,0 +1,83 @@
...
+        fvkmemory::resource_ptr<VulkanStream> stream, bool newImage) {
+    for (streamedTextureBinding const& data: mStreamedTexturesBindings) {
+        if (data.image->getStream() == stream) {
+            if (newImage) {
+                mExternalImageManager->bindExternallySampledTexture(data.set, data.binding, image,
+                        data.samplerParams);
+            } else {
+                // For some reason, some of the frames coming to us, are on streams where the
+                // descriptor set isn't external...
+                if (data.set->getExternalSamplerVkSet()) {
```
rafadevai: Can you add a comment here saying that this is where the descriptor set gets updated with the corresponding stream texture?

When reading this code is not clear this is happening here

filament/backend/src/vulkan/VulkanDriver.cpp
```diff
@@ -455,13 +456,18 @@ void VulkanDriver::updateDescriptorSetTexture(
     if (UTILS_UNLIKELY(mExternalImageManager.isExternallySampledTexture(texture))) {
         mExternalImageManager.bindExternallySampledTexture(set, binding, texture, params);
         mAppState.hasBoundExternalImages = true;
-    } else {
+    } else if (bool(texture->getStream())) {
+        mStreamedImageManager.bindStreamedTexture(set, binding, texture, params);
```
rafadevai, line 510: do we have to handle the case where the texture is destroyed but not the stream? 
and in that case we have to call mStreamedImageManager.unbindStreamedTexture(....)

filament/backend/src/vulkan/VulkanDriver.cpp
rafadevai: this case is not supported, right? if so, maybe its better to log an error message

filament/backend/src/vulkan/VulkanDriver.cpp
```diff
@@ -991,25 +1002,99 @@ void VulkanDriver::destroyDescriptorSet(Handle<HwDescriptorSet> dsh) {
 }
 
 Handle<HwStream> VulkanDriver::createStreamNative(void* nativeStream) {
-    return {};
+    auto handle = mResourceManager.allocHandle<VulkanStream>();
```
rafadevai: add FVK_SYSTRACE_SCOPE();

filament/backend/src/vulkan/VulkanDriver.cpp
```diff
@@ -965,6 +971,11 @@ void VulkanDriver::destroySwapChain(Handle<HwSwapChain> sch) {
 }
 
 void VulkanDriver::destroyStream(Handle<HwStream> sh) {
+    if (!sh) {
```
rafadevai, line 1143: FVK_SYSTRACE_SCOPE();

filament/backend/src/vulkan/VulkanDriver.cpp
```diff
@@ -991,25 +1002,99 @@ void VulkanDriver::destroyDescriptorSet(Handle<HwDescriptorSet> dsh) {
...
-    return {};
+    auto handle = mResourceManager.allocHandle<VulkanStream>();
+    auto stream = resource_ptr<VulkanStream>::make(&mResourceManager, handle);
+    stream.inc();
+    return handle;
 }
 
 Handle<HwStream> VulkanDriver::createStreamAcquired() {
-    return {};
+    auto handle = mResourceManager.allocHandle<VulkanStream>();
```
rafadevai, line 1192: FVK_SYSTRACE_SCOPE();

filament/backend/src/vulkan/VulkanDriver.cpp
```diff
@@ -991,25 +1002,99 @@ void VulkanDriver::destroyDescriptorSet(Handle<HwDescriptorSet> dsh) {
...
 void VulkanDriver::setAcquiredImage(Handle<HwStream> sh, void* image, const math::mat3f& transform,
         CallbackHandler* handler, StreamCallback cb, void* userData) {
+    FVK_SYSTRACE_SCOPE();
+    auto stream = resource_ptr<VulkanStream>::cast(&mResourceManager, sh);
+    stream->acquire({ image, cb, userData, handler });
+    mStreamsWithPendingAcquiredImage.push_back(stream);
 }
 
 void VulkanDriver::setStreamDimensions(Handle<HwStream> sh, uint32_t width, uint32_t height) {
+    auto stream = resource_ptr<VulkanStream>::cast(&mResourceManager, sh);
```
rafadevai, line 1215: FVK_SYSTRACE_SCOPE();

filament/backend/src/vulkan/VulkanDriver.cpp
```diff
@@ -991,25 +1002,99 @@ void VulkanDriver::destroyDescriptorSet(Handle<HwDescriptorSet> dsh) {
...
+    stream->width = width;
+    stream->height = height;
 }
 
 int64_t VulkanDriver::getStreamTimestamp(Handle<HwStream> sh) {
     return 0;
 }
 
 void VulkanDriver::updateStreams(CommandStream* driver) {
+    if (UTILS_UNLIKELY(!mStreamsWithPendingAcquiredImage.empty())) {
```
rafadevai, line 1226: FVK_SYSTRACE_SCOPE();

filament/backend/src/vulkan/VulkanDriver.cpp
```diff
@@ -1331,6 +1416,11 @@ TimerQueryResult VulkanDriver::getTimerQueryValue(Handle<HwTimerQuery> tqh, uint
 }
 
 void VulkanDriver::setExternalStream(Handle<HwTexture> th, Handle<HwStream> sh) {
+    auto t = resource_ptr<VulkanTexture>::cast(&mResourceManager, th);
```
rafadevai, line 1695: FVK_SYSTRACE_SCOPE();

filament/backend/src/vulkan/VulkanTexture.h
```diff
@@ -36,6 +36,34 @@ namespace filament::backend {
 
 struct VulkanTexture;
 
+struct VulkanStream : public HwStream, fvkmemory::Resource {
+    fvkmemory::resource_ptr<VulkanTexture> getTexture(void* ahb) {
+        fvkmemory::resource_ptr<VulkanTexture> frame;
+        const auto& it = mTextures.find(ahb);
```
rafadevai: Probably a good idea to create a bug to limit the amount of AHB cached at a given time for a stream.
Its possible we always receive new buffers and they will never get released until the stream is destroyed.

filament/backend/src/vulkan/VulkanStreamedImageManager.cpp
```diff
@@ -0,0 +1,83 @@
...
+        VulkanSamplerCache* samplerCache)
+    : mExternalImageManager(manager),
+      mDescriptorSetCache(descriptorSet),
+      mSamplerCache(samplerCache) {
+
+}
+
+VulkanStreamedImageManager::~VulkanStreamedImageManager() = default;
+
+void VulkanStreamedImageManager::terminate() {}
```
rafadevai: Better to just remove it, since its empty

filament/backend/src/vulkan/VulkanDriver.cpp
```diff
@@ -455,13 +456,18 @@ void VulkanDriver::updateDescriptorSetTexture(
     if (UTILS_UNLIKELY(mExternalImageManager.isExternallySampledTexture(texture))) {
         mExternalImageManager.bindExternallySampledTexture(set, binding, texture, params);
         mAppState.hasBoundExternalImages = true;
-    } else {
+    } else if (bool(texture->getStream())) {
```
rafadevai, line 509: can a texture be external sampled and streamed at the same time? or are they mutually exclusive?

phoenixxxx: Logically yes. But that would mean that a texture that is external was attached to a stream. And that seems like something went wrong, because from the frontend's perspective you can create a texture one of 3 ways:
- 1) From a format
- 2) From an external image
- 3) By setting a stream
Although you could logically do 2) and then set a stream via 3, I question the meaning of that.

filament/backend/src/vulkan/VulkanStreamedImageManager.cpp
phoenixxxx: Good call, but I actually forgot to clear the vector.

filament/backend/src/vulkan/VulkanDriver.cpp
phoenixxxx: If the texture is destroyed and not the stream then eventually the stream should run out of handles. Is there a particular use case you have in mind?
Binding just adds one more unique tuple to check at draw time, but it doesn't itself hold the resource.

filament/backend/include/backend/platforms/VulkanPlatformAndroid.h
phoenixxxx: Yeah this issue is that Mathias didn't want to do that. 
The whole concept of ExternalImageHandleRef exists to avoid using raw pointer. Something that the Stream API totally breaks, it relies (on the GL and VK sides) on void*.

Second issue is we did a ton of changes on the Impress side to create, and send ExternalImageHandleRef (works for all external image types) so undoing this would need many Impress side changes.

filament/backend/src/vulkan/VulkanDriver.cpp
phoenixxxx: My bad this was an artifact of when that call was on the synchronous path. Back then allocation and usage were different.

filament/backend/include/backend/platforms/VulkanPlatform.h
```diff
@@ -406,6 +406,9 @@ class VulkanPlatform : public Platform, utils::PrivateImplementation<VulkanPlatf
     virtual ImageData createVkImageFromExternal(ExternalImageHandleRef image) const {
         return {};
     }
+    virtual ExternalImageHandle createExternalImageFromRaw(void* image, bool sRGB) noexcept {
+        return {};
+    }
```
pixelflinger: - all virtuals in this file should not be implemented in the header file
- all need documentation, as this *is* the public API.

filament/backend/include/backend/platforms/VulkanPlatformAndroid.h
pixelflinger: I don't even understand what the void* is here? This needs to be documented.

phoenixxxx: So that's because in the stream case, we call createExternalImageFromRaw insinde the driver when we process the pending stream. And inside VulkanDriver.cpp we don't know what the platform is, so we can't assume createExternalImage(AHardwareBuffer const*)

filament/backend/src/vulkan/VulkanTexture.h
phoenixxxx: I think I've found a way around this.

filament/backend/src/vulkan/VulkanTexture.h
phoenixxxx: I think i've found a way around this.

filament/backend/src/vulkan/VulkanDriver.cpp
```diff
@@ -1278,7 +1383,34 @@ void VulkanDriver::updateIndexBuffer(Handle<HwIndexBuffer> ibh, BufferDescriptor
 
 void VulkanDriver::registerBufferObjectStreams(Handle<HwBufferObject> boh,
         BufferObjectStreamDescriptor&& streams) {
-    // Noop
+
+    auto bo = resource_ptr<VulkanBufferObject>::cast(&mResourceManager, boh);
+    mStreamUniformDescriptors[bo->getVkBuffer()] = std::move(streams);
```
rafadevai: I think its better to use the bo handle instead of the vk one, in theory, the vk handle could change even tho we are not doing that right now, so better be safe here.

filament/backend/src/vulkan/VulkanDriver.cpp
```diff
@@ -1289,6 +1421,20 @@ void VulkanDriver::updateBufferObject(Handle<HwBufferObject> boh, BufferDescript
...
+        auto streamDescriptors = mStreamUniformDescriptors.find(bo->getVkBuffer());
+        if (streamDescriptors != mStreamUniformDescriptors.end()) {
+            for (auto const& [offset, stream, associationType]:
+                    streamDescriptors->second.mStreams) {
+                if (associationType == BufferObjectStreamAssociationType::TRANSFORM_MATRIX) {
+                    auto transform = getStreamTransformMatrix(stream);
+                    copyMat3f(bd.buffer, offset, transform);
+                }
+            }
+            mStreamUniformDescriptors.erase(streamDescriptors);
```
rafadevai, line 1623: why do we have to erase this?

filament/backend/include/backend/platforms/VulkanPlatformAndroid.h
pixelflinger: Ok I see why you need to do this.  This is because AcquiredImage has a void*. I think this is okay for now (for the sake of getting this done). But really, we should change the Acquired Stream API to only speak in terms of ExternalImageRef.

HOWEVER, I think this means that createExternalImageFromRaw() must not be public. It must be protected. And the concrete implementations must be private. This is needed only for internal machinery.

pixelflinger (changes requested):

filament/backend/include/backend/platforms/VulkanPlatformAndroid.h
phoenixxxx: Yes I am ok with that. Frankly, I think I agree, the stream taking void* isn't the right call.

filament/backend/src/vulkan/VulkanDriver.cpp
poweifeng: not needed if `mStreamsWithPendingAcquiredImage` is only used with main thread.

filament/backend/src/vulkan/VulkanDriver.cpp
```diff
@@ -1014,25 +1024,120 @@ void VulkanDriver::destroyDescriptorSet(Handle<HwDescriptorSet> dsh) {
...
+        for (auto& stream: mStreamsWithPendingAcquiredImage) {
+            if (stream->previousNeedsRelease()) {
+                scheduleRelease(stream->takePrevious());
+            }
+
+            // This executes on the backend thread (updateStreams is synchonous which means it
+            // executes on the user thread) Note: stream is captured by copy which is fine, this is
+            // a copy of a resource_ptr<VulkanStream>. We only need it find the associated stream
+            // inside the mStreamedImageManager texture bindings
+            driver->queueCommand([this, stream, s = stream.get(),
```
poweifeng, line 1236: why do the `s=stream.get`?

filament/backend/src/vulkan/VulkanDriver.cpp
```diff
@@ -1014,25 +1024,120 @@ void VulkanDriver::destroyDescriptorSet(Handle<HwDescriptorSet> dsh) {
...
+
+                    if (imgData.external.valid()) {
+                        mExternalImageManager.addExternallySampledTexture(newTexture,
+                                externalImage);
+                        // Cache the AHB backed image.
+                        s->pushImage(image, newTexture);
+                        newImage = true;
+                    }
+
+                    newTexture.inc();
```
poweifeng: why inc here?

filament/backend/src/vulkan/VulkanDriver.h
```diff
@@ -153,6 +154,12 @@ class VulkanDriver final : public DriverBase {
     VulkanDescriptorSetCache mDescriptorSetCache;
     VulkanQueryManager mQueryManager;
     VulkanExternalImageManager mExternalImageManager;
+    VulkanStreamedImageManager mStreamedImageManager;
+
+    // Stream transforms
+    std::unordered_map<VkBuffer, BufferObjectStreamDescriptor> mStreamUniformDescriptors;
```
poweifeng: can we keep this in `VulkanStreamedImageManager` ?

filament/backend/src/vulkan/VulkanDriver.cpp
poweifeng: since these resources are ref-counted, in fact, the binding map does hold the resource.

filament/backend/src/vulkan/VulkanStreamedImageManager.cpp
```diff
@@ -0,0 +1,87 @@
...
+}
+
+void VulkanStreamedImageManager::onStreamAcquireImage(fvkmemory::resource_ptr<VulkanTexture> image,
+        fvkmemory::resource_ptr<VulkanStream> stream, bool newImage) {
+    for (StreamedTextureBinding const& data: mStreamedTexturesBindings) {
+        if (data.image->getStream() == stream) {
+            if (newImage) {
+                mExternalImageManager->bindExternallySampledTexture(data.set, data.binding, image,
+                        data.samplerParams);
+            } else {
```
poweifeng: I don't think you'd only want to bindExternallySampledTexture if it's a newImage.  Instead don't we want something like

```
if (externallySampled) {
  mExternalImageManager->bindExternallySampledTexture()
} else {
  mDescriptorSetCache->updateSampler();
}
```

filament/backend/src/vulkan/VulkanTexture.h
```diff
@@ -36,6 +36,46 @@ namespace filament::backend {
...
+    void acquire(const AcquiredImage& image) {
+        user_thread.mPrevious = user_thread.mAcquired;
+        user_thread.mAcquired = image;
+    }
+    bool previousNeedsRelease() const { return (user_thread.mPrevious.image != nullptr); }
+    // this function will null the previous once the caller takes it.
+    // It ensures we don't schedule for release twice.
+    AcquiredImage takePrevious() {
+        AcquiredImage previous = user_thread.mPrevious;
+        user_thread.mPrevious = { nullptr, nullptr, nullptr, nullptr };
```
poweifeng: You can just write

mPrevious = {};

filament/backend/src/vulkan/VulkanTexture.h
```diff
@@ -74,6 +114,8 @@ struct VulkanTextureState : public fvkmemory::Resource {
 
     // The texture with the sidecar owns the sidecar.
     fvkmemory::resource_ptr<VulkanTexture> mSidecarMSAA;
+    // The stream this texture is associated with (I think cleaner than HwTexture::hwStream).
```
poweifeng: nit: no need for `(I think cleaner than HwTexture::hwStream).`

filament/backend/src/vulkan/VulkanDriver.cpp
phoenixxxx: So the logic is that the user has to submit an update to the stream descriptor every time the xform changes. Once we process it, we can remove it from the list.

filament/backend/src/vulkan/VulkanDriver.cpp
phoenixxxx: good call.

filament/backend/src/vulkan/VulkanStreamedImageManager.cpp
phoenixxxx: Good call

filament/backend/src/vulkan/VulkanDriver.h
phoenixxxx: I would rather not. It feels more nature for the stream to hold this data. It also means that once we have the stream we can access all the stream images. Since this object is not a Thread safe resource, it seems sensible?
But if you feel strongly about it, I can move it to the VulkanStreamedImageManager, that's one more level of indirection where we need a <stream> to map<VkBuffer, BufferObjectStreamDescriptor> inside the VulkanStreamedImageManager

filament/backend/src/vulkan/VulkanDriver.cpp
phoenixxxx: Because it's in the loop, and the stream variable will actually change, so I think we need a unique copy of the stream pointer (access underlying data) per queueCommand

filament/backend/src/vulkan/VulkanDriver.cpp
phoenixxxx: Haha yeah not sure... I am still fuzzy on the ref counting but your next comment clarified it.

phoenixxxx: So I can remove newTexture.inc(); ?

filament/backend/src/vulkan/VulkanDriver.cpp
poweifeng: So I think `stream` is a copy of the ref-counted pointer to the actual content already.  You shouldn't need to do `s = stream.get` (unless you pass it with `&stream`, then it's a reference.  But even this is ok I think).

filament/backend/src/vulkan/VulkanDriver.cpp
poweifeng: yes you can.  The inc is only for the `create/destroy` calls in driver.  So `inc` on `create` and `dec` on `destroy`.  This is just an optimization to avoid keep a list of references.

filament/backend/src/vulkan/VulkanStreamedImageManager.cpp
```diff
@@ -0,0 +1,87 @@
...
+                if (newImage) {
+                    mExternalImageManager->bindExternallySampledTexture(data.set, data.binding,
+                            image, data.samplerParams);
+                }
+                // Eventually the updateSampler and updateSamplerForExternalSamplerSet
+                // will call to vkUpdateDescriptorSets with a VkWriteDescriptorSet
+                // type VK_DESCRIPTOR_TYPE_COMBINED_IMAGE_SAMPLER the
+                // VkDescriptorImageInfo will contain the view of the new image frame.
+                mDescriptorSetCache->updateSamplerForExternalSamplerSet(data.set, data.binding,
+                        image);
```
poweifeng: I don't think you need this part even. This will be done in `ExternalImageManager::prepareBindSets`.

filament/backend/src/vulkan/VulkanStreamedImageManager.cpp
```diff
@@ -0,0 +1,87 @@
...
+
+void VulkanStreamedImageManager::onStreamAcquireImage(fvkmemory::resource_ptr<VulkanTexture> image,
+        fvkmemory::resource_ptr<VulkanStream> stream, bool newImage) {
+    for (StreamedTextureBinding const& data: mStreamedTexturesBindings) {
+        // Find the right stream
+        if (data.image->getStream() == stream) {
+            // For some reason, some of the frames coming to us, are on streams where the
+            // descriptor set isn't external...
+            if (data.set->getExternalSamplerVkSet()) {
+                if (newImage) {
```
poweifeng: I think we can always just call `bindExternallySampledTexture` even if it's not a `newImage` ? 

But we should probably add some logic in `bindExternallSampledTexture` to remove an entry if it shares the same `data.set` and `data.binding`.  (Basically it's overwriting the existing binding for a descriptor set).

filament/backend/src/vulkan/VulkanTexture.h
poweifeng: What way did you find?

filament/backend/src/vulkan/VulkanDriver.cpp
rafadevai: in which case will change? Now that i look at this again, do you actually need stream and s at the same time?

filament/backend/src/vulkan/VulkanDriver.cpp
rafadevai: yes, it should be removed if you are storing the ref to the object some where else. For example right now its stored in the stream.

filament/backend/include/backend/platforms/VulkanPlatformAndroid.h
phoenixxxx: How do I not make this public?

filament/backend/src/vulkan/VulkanTexture.h
phoenixxxx: So first the lifetime of the resource is only managed on the frontend (ie: the backend doesnt hold resources to the stream) so we don;t need to atomic incredment of the ThreadSafeResource. Stream is pretty much like a texture or any other fvkmemory::Resource derived class.

This follows the desing pattern of GL and like GL in updateStreams we just make sure to capture and access parts of the stream that the front end doesnt touch.

filament/backend/src/vulkan/VulkanDriver.cpp
phoenixxxx: Yeah so this follows the GL design pattern where it gets the underlying raw stream object and sends it to the backend (the backend only manipulates the raw data thereby ensure we don't need to share the actual VulkanStream object betwen FE and BE... I can explain this on the chat, it's a bit tricky (not great) but that's what the GL path does.

> in which case will change? Now that i look at this again, do you actually need stream and s at the same time?
This is because we use the [VulkanStream ](fvkmemory::resource_ptr<VulkanStream>) (not VulkanStream*)

filament/backend/src/vulkan/VulkanTexture.h
poweifeng: Backend does in fact hold all the resources.  This is due to the ref-counting mechanism of the vk backend.  This is not true for the GL backend.  This does have to be a `ThreadSafeResource` because the resource is accessed in both the backend thread and the frontend thread.  And `ThreadSafeResource` cannot hold references to other resources due to [this](https://github.com/google/filament/blob/main/filament/backend/src/vulkan/memory/ResourceManager.cpp#L45)

filament/backend/src/vulkan/VulkanDriver.cpp
poweifeng: I still don't understand why you can't just pass `stream` by copy.  `stream` is a pointer.  It can be dereferenced to get the content (`s`).

filament/backend/src/vulkan/VulkanStreamedImageManager.cpp
phoenixxxx: Ah yes, I think this is right. bindExternallySampledTexture pushes into mSetBindings and we allways update all the bindings.

filament/backend/src/vulkan/VulkanStreamedImageManager.cpp
phoenixxxx: I don't think so, we get these calls every frame, which means we will keep pushing int the binding vector and it will grow unbounded.

filament/backend/src/vulkan/VulkanDriver.cpp
```diff
@@ -988,6 +994,11 @@ void VulkanDriver::destroySwapChain(Handle<HwSwapChain> sch) {
 }
 
 void VulkanDriver::destroyStream(Handle<HwStream> sh) {
+    if (!sh) {
```
rafadevai, line 1143: FVK_SYSTRACE_SCOPE

filament/backend/src/vulkan/utils/Image.cpp
```diff
@@ -234,6 +234,15 @@ uint8_t reduceSampleCount(uint8_t sampleCount, VkSampleCountFlags mask) {
     return mostSignificantBit((sampleCount - 1) & mask);
 }
 
+Platform::ExternalImageHandle createExternalImageFromRaw(void* image,
+        bool sRGB) {
+#if defined(__ANDROID__)
+    return createExternalImage(reinterpret_cast<AHardwareBuffer const*>(image), sRGB);
```
poweifeng: don't you have to call this on the platform?

filament/backend/src/vulkan/VulkanTexture.h
```diff
@@ -36,6 +36,49 @@ namespace filament::backend {
...
+    }
+    void pushImage(void* ahb, fvkmemory::resource_ptr<VulkanTexture> tex) { mTextures[ahb] = tex; }
+
+private:
+    // These are only called from the frontend
+    struct {
+    AcquiredImage mAcquired;
+    AcquiredImage mPrevious;
+    math::mat3f mTransform;
+    } user_thread;
```
poweifeng: formatting here

filament/backend/src/vulkan/VulkanStreamedImageManager.cpp
poweifeng: You'll need to add logic in `bindExternallySampledTexture` to remove bindings to the same set and set binding.

filament/backend/src/vulkan/VulkanTexture.h
```diff
@@ -36,6 +36,49 @@ namespace filament::backend {
...
+
+    // #TODO b/442937292
+    math::mat3f mTransform;
+    std::map<void*, fvkmemory::resource_ptr<VulkanTexture>> mTextures;
```
poweifeng: unordered_map is better here.

filament/backend/src/vulkan/VulkanDriver.cpp
```diff
@@ -1014,25 +1025,122 @@ void VulkanDriver::destroyDescriptorSet(Handle<HwDescriptorSet> dsh) {
...
+        if (stream->streamType == StreamType::NATIVE) {
+            FILAMENT_CHECK_PRECONDITION(false) << "Native Stream not supported in Vulkan.";
+        } else {
+            // Backend call since getStreamTransformMatrix is called from async
+            // updateBufferObject.
+            return stream->getBackEndTransform();
+        }
+    }
+    return math::mat3f();
+}
```
poweifeng: This method isn't doing much, can't we just move it to the callsite?

poweifeng: maybe as a lambda if you prefer.

filament/backend/src/vulkan/VulkanDriver.cpp
phoenixxxx: It's not in the other methods around this one. Is it needed?

phoenixxxx: discussed internally.

filament/backend/src/vulkan/utils/Image.cpp
phoenixxxx: Yes, sorry I build on my windows, and I had to push and pull on my gLinux to bruild there as well.

filament/backend/src/vulkan/VulkanDriver.cpp
```diff
@@ -1087,25 +1099,106 @@ void VulkanDriver::destroyDescriptorSet(Handle<HwDescriptorSet> dsh) {
...
+                        mExternalImageManager.addExternallySampledTexture(newTexture,
+                                externalImage);
+                        // Cache the AHB backed image. Acquires the image here.
+                        s->pushImage(image, newTexture);
+                    }
+
+                    texture = newTexture;
+                }
+
+                // Note that we capture the
```
poweifeng: nit: remove unnecessary comment

filament/backend/src/vulkan/VulkanExternalImageManager.cpp
```diff
@@ -261,7 +261,18 @@ void VulkanExternalImageManager::bindExternallySampledTexture(
         fvkmemory::resource_ptr<VulkanTexture> image, SamplerParams samplerParams) {
     // Should we do duplicate validation here?
     auto& imageData = findImage(mImages, image);
-    mSetBindings.push_back({ bindingPoint, imageData.image, set, samplerParams });
+    auto itr = std::find_if(mSetBindings.begin(), mSetBindings.end(),
+            [&](SetBindingInfo const& binding) {
+                SamplerParams::EqualTo comparator;
+                return (binding.set == set && binding.binding == bindingPoint &&
+                        comparator(samplerParams, binding.samplerParams));
```
poweifeng: this is not necessary.

filament/backend/src/vulkan/VulkanExternalImageManager.cpp
```diff
@@ -261,7 +261,18 @@ void VulkanExternalImageManager::bindExternallySampledTexture(
...
+            });
+    if (itr == mSetBindings.end()) {
+        mSetBindings.push_back({ bindingPoint, imageData.image, set, samplerParams });
+    } else {
+        // override the image data in the binding point
+        itr->image = image;
```
poweifeng, line 223: and the samplerParams

filament/backend/src/vulkan/VulkanTexture.h
```diff
@@ -36,6 +36,49 @@ namespace filament::backend {
...
+    const math::mat3f& getFrontEndTransform() const { return user_thread.mTransform; }
+
+    //-- These methods are only called from the backend thread
+    void setBackendTransform(const math::mat3f& transform) { mTransform = transform; }
+    const math::mat3f& getBackEndTransform() const { return mTransform; }
+    fvkmemory::resource_ptr<VulkanTexture> getTexture(void* ahb) {
+        fvkmemory::resource_ptr<VulkanTexture> frame;
+        const auto& it = mTextures.find(ahb);
+        if (it != mTextures.end()) frame = it->second;
+        return frame;
```
poweifeng: can be simplified
```
if (auto itr = mTextures.find(ahb); itr != mTextures.end()) {
    return itr->second;
}
return {};
```

filament/backend/src/vulkan/VulkanDriver.cpp
```diff
@@ -1087,25 +1099,106 @@ void VulkanDriver::destroyDescriptorSet(Handle<HwDescriptorSet> dsh) {
...
+    FILAMENT_CHECK_PRECONDITION(false) << "createStreamNative not supported in Vulkan.";
     return {};
 }
 
 Handle<HwStream> VulkanDriver::createStreamAcquired(utils::CString tag) {
-    return {};
+    FVK_SYSTRACE_SCOPE();
+    auto handle = mResourceManager.allocHandle<VulkanStream>();
+    auto stream = resource_ptr<VulkanStream>::make(&mResourceManager, handle);
+    stream.inc();
```
poweifeng, line 1194: Can you put a TODO here?  This is still not thread-safe. We might have to revisit this question.

filament/backend/src/vulkan/VulkanExternalImageManager.cpp
phoenixxxx: Good catch.

filament/backend/src/vulkan/VulkanExternalImageManager.cpp
phoenixxxx: Good catch.

filament/backend/src/vulkan/VulkanDriver.cpp
```diff
@@ -1378,7 +1471,34 @@ void VulkanDriver::updateIndexBuffer(Handle<HwIndexBuffer> ibh, BufferDescriptor
...
         BufferObjectStreamDescriptor&& streams) {
-    // Noop
+
+    auto bo = resource_ptr<VulkanBufferObject>::cast(&mResourceManager, boh);
+    mStreamUniformDescriptors[bo.get()] = std::move(streams);
+}
+
+// This needs to be discussed because Vulkan has different default Matrix formats
+// right now this is following GL.
+static void copyMat3f(void* addr, size_t const offset, const math::mat3f& v) noexcept {
```
poweifeng: You can move this to the anonymous namespace above

filament/backend/src/vulkan/VulkanDriver.cpp
```diff
@@ -1087,25 +1124,106 @@ void VulkanDriver::destroyDescriptorSet(Handle<HwDescriptorSet> dsh) {
...
+    return handle;
 }
 
 void VulkanDriver::setAcquiredImage(Handle<HwStream> sh, void* image, const math::mat3f& transform,
         CallbackHandler* handler, StreamCallback cb, void* userData) {
+    FVK_SYSTRACE_SCOPE();
+    auto stream = resource_ptr<VulkanStream>::cast(&mResourceManager, sh);
+    stream->acquire({ image, cb, userData, handler});
+    stream->setFrontEndTransform(transform);
+    mStreamsWithPendingAcquiredImage.push_back(stream);
```
pixelflinger, line 1210: I think there is a few things missing here (but I'm not sure):
- maybe check that the stream is indeed for type ACQUIRED (maybe an assert at least)
- does this handle correctly if setAcquiredImage is called twice in a raw with different images (only the last one should be use and the previous one released)
- are there cases where push_pack() shouldn't be called if the method above failed or didn't have an image.

I'm just referring the GL backend implementation here.

pixelflinger (approved):
LGTM, just had a small comment

filament/backend/src/vulkan/VulkanDriver.cpp
phoenixxxx: Good calls. Addressed.

filament/backend/src/vulkan/VulkanStreamedImageManager.cpp
```diff
@@ -0,0 +1,79 @@
...
+}
+
+void VulkanStreamedImageManager::onStreamAcquireImage(fvkmemory::resource_ptr<VulkanTexture> image,
+        fvkmemory::resource_ptr<VulkanStream> stream) {
+    for (StreamedTextureBinding const& data: mStreamedTexturesBindings) {
+        // Find the right stream
+        if (data.image->getStream() == stream) {
+            // For some reason, some of the frames coming to us, are on streams where the
+            // descriptor set isn't external...
+            if (data.set->getExternalSamplerVkSet()) {
```
poweifeng: I think this needs to be `image->getVkFormat() == VK_FORMAT_UNDEFINED` that indicates the texture has a externally sampled format. The `ExternalSamplerVkSet` will only be created if the `bindExternallySampledTexture` goes through.

rafadevai: Nice catch!!

You are right, if the new acquiredImage needs an external sampler then bind it, otherwise just update the sampler

phoenixxxx: Good catch yes.

filament/backend/src/vulkan/VulkanDriver.cpp
```diff
@@ -1389,6 +1515,33 @@ void VulkanDriver::updateBufferObject(Handle<HwBufferObject> boh, BufferDescript
     commands.acquire(bo);
     bo->loadFromCpu(commands, bd.buffer, byteOffset, bd.size);
 
+    if (UTILS_UNLIKELY(!mStreamUniformDescriptors.empty())) {
```
rafadevai, line 1600: shouldnt the copyMat3f happen before bo->loadFromCpu? 
right now its copying the current bd.buffer to the gpu memory and then its updating the transform

filament/backend/src/vulkan/VulkanStreamedImageManager.cpp
```diff
@@ -0,0 +1,79 @@
...
+}
+
+void VulkanStreamedImageManager::onStreamAcquireImage(fvkmemory::resource_ptr<VulkanTexture> image,
+        fvkmemory::resource_ptr<VulkanStream> stream) {
+    for (StreamedTextureBinding const& data: mStreamedTexturesBindings) {
+        // Find the right stream
+        if (data.image->getStream() == stream) {
+            // For some reason, some of the frames coming to us, are on streams where the
+            // descriptor set isn't external...
+            if (data.image->getVkFormat() == VK_FORMAT_UNDEFINED) {
```
rafadevai: so im pretty sure here, you have to create a new descriptor set and copy the contents otherwise you might update one currently in flight

like every time, you acquire a new image, we need a new descriptor set regardless if the the sampler changed or not

poweifeng: and that's actually what [this PR](https://github.com/google/filament/pull/9251) is doing.  So you might want to wait for it to land and then rebase.

rafadevai: but even if the stream is not using an external sampler, you would still have to create a new descriptor set, which from what I understand, that PR wont help

Also in the case its also using the same colorspace but the texture is different, you would also require a new descriptor set because you want to update the entry but not change the one currently in flight.

poweifeng:
```but even if the stream is not using an external sampler, you would still have to create a new descriptor set, which from what I understand, that PR wont help
```
actually, you're right.  You'll need to make a new descriptor for the non-externally-sampled case.

phoenixxxx: OK so I think I might have lost track of all the temporary DS.... And I think a lot of cracks were found in the current ext sampler implementations. I am fine focusing on https://github.com/google/filament/pull/9251 and then re-evaluating how to manage this for the stream path (in this PR)

filament/backend/src/vulkan/VulkanDriver.cpp
phoenixxxx: Yes that's right, good catch.

PR #9167, opened 2025-08-29
additional performance analysis documentation, namely on locking GPU frequency

AndyHovingh:


docs_src/src_mdbook/src/notes/performance_analysis.md
```diff
@@ -6,12 +6,63 @@
...
+      echo 940000 > hint_max_freq
+      # you can typically verify the GPU is running at that frequency consistently by running something like
+      # the following a few times over time, which should show the frequency you want to lock the device to:
+      cat cur_freq
+      ```
+   1. You may need to re-apply the `hint_min_freq` just before starting the profiling trace and check before and after
+      that the frequency remained at the value expected. Some systems may adjust the frequency on you, but you
+      may want to ensure the frequency remains the same through the analysis.
+   1. After you have done your app profiling, just revert the state of the device back to the way you had it
+      initially, as needed.
```
poweifeng: afaict restarting the device will undo these changes.

AndyHovingh: The GPU frequency yes, but I was referring to the whole OS image etc.

PR #9169, opened 2025-08-29
remove all uses of std::string in filament

pixelflinger:
- libbackend (except webgpu)
- libfilament
- libutils

std::string generates a lot of code bloat, we use CString instead.

We also update CString to be more compatible with std::string's api.


poweifeng (approved):

romainguy:
🥳

PR #9174, opened 2025-09-02
add option to disable perfetto on android

pixelflinger:
The new cmake `FILAMENT_ENABLE_PERFETTO` must be set to enable  perfetto traces on Android. It is disabled by default on release builds, enabled otherwise.

The reason for this is that the perfetto SDK adds about 800K of code to the library. The text section goes from 2.2 to 1.4 MB


poweifeng:
I think JobSystem [here](https://github.com/google/filament/blob/main/libs/utils/src/JobSystem.cpp#L20) also needs to be updated?

z3moon (approved):

PR #9175, opened 2025-09-03
Add getEyeFromViewMatrix() for vertex shader

z3moon:
This new method, intended for the vertex shader, returns a matrix that transforms vertices from view space (also known as head space) to the current eye space.


filament/src/ds/PerViewDescriptorSetUtils.cpp
```diff
@@ -64,6 +64,7 @@ void PerViewDescriptorSetUtils::prepareCamera(PerViewUib& s,
         mat4f const& eyeFromHead = camera.eyeFromView[i];   // identity for monoscopic rendering
         mat4f const& clipFromEye = camera.eyeProjection[i];
         // clipFromEye * eyeFromHead * headFromWorld
```
poweifeng: I think you'd want to move this comment

poweifeng (changes requested):

filament/src/ds/PerViewDescriptorSetUtils.cpp
z3moon: Ah good catch.

poweifeng (approved):

libs/filabridge/include/private/filament/UibStructs.h
```diff
@@ -212,8 +213,8 @@ struct PerViewUib { // NOLINT(cppcoreguidelines-pro-type-member-init)
 };
 
 // 2 KiB == 128 float4s
-static_assert(sizeof(PerViewUib) == sizeof(math::float4) * 128,
-        "PerViewUib should be exactly 2KiB");
+static_assert(sizeof(PerViewUib) == sizeof(math::float4) * 144,
+        "PerViewUib should be 16 x 144 bytes");
```
pixelflinger, line 217: no this test should stay at 2KiB (see above). instead you need to change the '38' on line 212.

shaders/src/common_getters.glsl
```diff
@@ -46,6 +46,11 @@ highp mat4 getViewFromClipMatrix() {
     return frameUniforms.viewFromClipMatrix;
 }
 
+/** @public-api */
+highp mat4 getEyeFromViewMatrix() {
+    return frameUniforms.eyeFromViewMatrix[getEyeIndex()];
+}
```
pixelflinger, line 52: I think for this one we want a method that take the eyeIndex as a parameter. Maybe add an override. I think as part of this PR we can also add a version of getClipFromWorldMatrix(int) that takes the eyeIndex.

pixelflinger (commented):
You also need to update the material documentation for the new methods.

libs/filabridge/include/private/filament/UibStructs.h
z3moon: Oh I see. Let me correct it. Thanks for catching it.

shaders/src/common_getters.glsl
z3moon: Let me a overloading function.

PR #9176, opened 2025-09-03
Re-introduced Systrace

mdagois:
Added a way to switch the tracing mode between disabled, perfetto, and systrace.

I did those changes because I could not get Perfetto to work in my environment.
I learnt earlier that @pixelflinger is working on something similar, for a different reason (binary size impact of Perfetto).
Please let me know if you can use this (with eventual modifications), otherwise we can just drop the PR.

Note that I'll need to fix some of the BUILD files in g3 as well.


pixelflinger:
This PR https://github.com/google/filament/pull/9174 adds a flag to turn perfetto on/off.  I don't want to bring back systrace, this should work, if it doesn't we need to figure out why.

mdagois:
Got it. I will drop this PR then.
 As for perfetto, I was trying in g3. I assume a fix there will follow #9174?

PR #9182, opened 2025-09-04
Add Java bindings for setFrameScheduledCallback

bejado:


android/filament-android/src/main/java/com/google/android/filament/SwapChain.java
```diff
@@ -126,6 +126,78 @@ public void setFrameCompletedCallback(@NonNull Object handler, @NonNull Runnable
         nSetFrameCompletedCallback(getNativeObject(), handler, callback);
```
z3moon, line 126: nit: I wonder if isFrameCompletedCallbackSet is also needed.

z3moon (approved):

pixelflinger (approved):

android/filament-android/src/main/java/com/google/android/filament/SwapChain.java
bejado: Good point. We don't currently have this on the C++ side, so I left it out.

PR #9187, opened 2025-09-05
Metal: add debug labels to Metal resources

bejado:


filament/backend/src/metal/MetalHandles.h
```diff
@@ -479,6 +490,12 @@ struct MetalDescriptorSet : public HwDescriptorSet {
 
     void finalize(MetalDriver* driver);
 
+    void setLabel(const utils::CString& l) {
+#if FILAMENT_METAL_DEBUG_LABELS
+        label = l;
```
z3moon, line 502: nit: Do we not need to check the emptiness here like others? Or the other way around.

pixelflinger (approved):

PR #9188, opened 2025-09-05
View::setCamera() should accept nullptr

pixelflinger:
FIXES=[441757817]

### DIFF
--- a/android/filament-android/src/main/java/com/google/android/filament/View.java
+++ b/android/filament-android/src/main/java/com/google/android/filament/View.java
@@ -231,7 +231,7 @@ public class View {
      *
      * <p>
      * The View does not take ownership of the Scene pointer. Before destroying a Camera, be sure
-     * to remove it from all assoicated Views.
+     * to remove it from all associated Views. If the camera isn't set, Renderer::render() will result in a no-op.
      * </p>
      *
      * @see #getCamera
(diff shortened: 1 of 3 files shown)


z3moon:
I think we also need to modify `FCamera& getCameraUser() noexcept` and its call sites appropriately.

Also in `details/View.cpp`, `mCullingCamera` is referenced without null-pointer checks.

z3moon (changes requested):
I think we need to modify `FCamera& getCameraUser() noexcept` and its call sites appropriately.

Also in `details/View.cpp`, mCullingCamera is referenced without null-pointer checks.

pixelflinger:
> I think we need to modify `FCamera& getCameraUser() noexcept` and its call sites appropriately.
> Also in `details/View.cpp`, mCullingCamera is referenced without null-pointer checks.

`getCameraUser()` as well as `mCullingCamera` are only accessed by filament inside `Renderer::render()` and that has a check at the top. So we're good here.

However, it is true that `getCameraUser()` is called by `View& View::getCamera()` which is a public API (so we can't change it unfortunately). Basically the user is not allowed to call this if `hasCamera()` is `false`.  It's unfortunate that it doesn't return a pointer, but it's too late to change it.

z3moon (approved):
I see. Then maybe we need to update the comments for getCamera about the warning.

PR #9196, opened 2025-09-09
Add MSAA swapchain support

z3moon:
Users can now create a swapchain with MSAA support, which enables direct rendering to a multi-sampled swapchain image and automatic resolution.

This is only supported by EGL(Android). Other GL platforms (GLX, WGL, etc) don't support it because the swapchain MSAA settings must be configured before window creation.

Support for other backends will follow.

BUGS=[433327615]

### DIFF
--- a/filament/backend/src/vulkan/VulkanDriver.cpp
+++ b/filament/backend/src/vulkan/VulkanDriver.cpp
@@ -35,6 +35,7 @@
 #include <backend/DriverEnums.h>
 #include <backend/platforms/VulkanPlatform.h>
 
+#include <utils/compiler.h>
 #include <utils/CString.h>
 #include <utils/Panic.h>
 
@@ -484,6 +485,16 @@ void VulkanDriver::finish(int dummy) {
     FVK_SYSTRACE_SCOPE();
 
     endCommandRecording();
+
+    // It's not enough to wait on the fences of the buffers submitted.  Present calls are
+    // submitted to this same queue, so the more correct option is to call vkQueueWaitIdle.
+    vkQueueWaitIdle(mPlatform->getGraphicsQueue());
+    if (auto protectedQueue = mPlatform->getProtectedGraphicsQueue();
+            UTILS_UNLIKELY(protectedQueue != VK_NULL_HANDLE)) {
+        vkQueueWaitIdle(protectedQueue);
+    }
+
+    // This wait() is still necessary to update the fence status.
     mCommands.wait();
 
     mReadPixels.runUntilComplete();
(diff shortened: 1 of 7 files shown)


poweifeng (approved):

filament/backend/src/opengl/platforms/PlatformEGL.cpp
```diff
@@ -448,6 +448,11 @@ EGLConfig PlatformEGL::findSwapChainConfig(uint64_t flags, bool window, bool pbu
         configAttribs[EGL_RECORDABLE_ANDROID] = EGL_TRUE;
     }
 
+    if (flags & SWAP_CHAIN_CONFIG_MSAA_4_SAMPLES) {
+        configAttribs[EGL_SAMPLE_BUFFERS] = 1;
+        configAttribs[EGL_SAMPLES] = 4;
+    }
```
pixelflinger, line 454: I think if we're on a device that doesn't support 4 samples, this will fail the swapchain creation.

I think there needs to be a `isSupported" call. And we need to be very clear about what happens if MSAA_4_SAMPLES is used but not supported.

Here, it's a bit inconsistent because on on non-EGL it's ignored, but on EGL if it's not supported, it'll fail.

android/gradle.properties
```diff
@@ -22,8 +22,8 @@ android.defaults.buildfeatures.buildconfig=true
 android.nonTransitiveRClass=false
 android.nonFinalResIds=false
 
-com.google.android.filament.tools-dir=../../../out/release/filament
-com.google.android.filament.dist-dir=../out/android-release/filament
+com.google.android.filament.tools-dir=../../../out/debug/filament
+com.google.android.filament.dist-dir=../out/android-debug/filament
```
poweifeng: shouldn't change this

filament/backend/include/backend/platforms/PlatformEGL.h
```diff
@@ -98,8 +98,9 @@ class PlatformEGL : public OpenGLPlatform {
     void terminate() noexcept override;
 
     bool isProtectedContextSupported() const noexcept override;
-
     bool isSRGBSwapChainSupported() const noexcept override;
+    bool isMSAASwapChainSupported(uint32_t samples) const noexcept override;
```
poweifeng, line 102: can be a `uint8_t` I think

filament/backend/src/opengl/platforms/PlatformEGL.cpp
z3moon: Yeah makes sense. It's updated.

android/gradle.properties
z3moon: Oh, it was a WIP commit. I force-pushed it and it's ready now.

filament/backend/include/backend/platforms/PlatformEGL.h
z3moon: Yeah, it makes sense in terms of the range of the value. But let me leave in this way because uint8_t will be treated as 32bit or 64bit during the operation anyway.

PR #9199, opened 2025-09-10
vk: fix VulkanDriver::finish()

poweifeng:
VulkanDriver::finish() seems like it is waiting for all the
commands to be processed, but in reality, the present calls
are not waited on (because commands->wait() only waited on
the command buffers).

Here we just call vkQueueWaitIdle to address that case.

This fixes a validation error that occurs when running the
backend test BackendTest.MissingRequiredAttributes.


z3moon (approved):

pixelflinger (approved):

filament/backend/src/vulkan/VulkanCommands.cpp
```diff
@@ -467,8 +468,12 @@ void VulkanCommands::wait() {
     FVK_SYSTRACE_CONTEXT();
     FVK_SYSTRACE_START("commands::wait");
 
+    // It's not enough to wait on the fences of the buffers submitted.  Present calls are typically
+    // submitted to this same queue. So the more correct option is to call vkQueueWaitIdle.
+    vkQueueWaitIdle(mQueue);
```
rafadevai: I dont think this is a good idea. For present there should be another way to wait that is not waiting on the queue to be idle, this is a performance killer.

poweifeng: Well, wait() is a performance killer.  You're not suppose to call this while rendering. Only meant for tests and shutdown.

rafadevai: is wait never called in a normal app?

rafadevai: but also why we do care about present finishing? if the command buffer is completed, that should imply that all the rendering work is done

rafadevai: also this, doesnt wait to present to finish either

rafadevai: so looking at the backend code, when a swapchain gets resized, it might call wait.

poweifeng: yes, and also [this](https://github.com/google/filament/blob/main/filament/backend/src/vulkan/platform/VulkanPlatformSwapChainImpl.cpp#L359) will happen when swapchain is resized (destroyed).  So it's still necessary.

poweifeng: > but also why we do care about present finishing? if the command buffer is completed, that should imply that all the rendering work is done

We care on shutdown.  Otherwise, we can't destroy the semaphore associated with the present.

poweifeng: > also this, doesnt wait to present to finish either

so in fact, if you read the comment in the swapchain destroy bit, you'll see that it's guaranteed to wait for present to finish.

PR #9201, opened 2025-09-10
vk: fix blitDEPRECATED implementation

poweifeng:
blitDEPRECATED can also be used in the context of a multisampled texture resolve. Fix that case to use the resolve() path in the VulkanBlitter.

Also fixed backend test BlitTest.ColorResolve to add the correct usage bit (otherwise it'd have validation error).


pixelflinger (approved):
we need to do an audit and see if we can remove all uses of blit DEPRECATED

PR #9204, opened 2025-09-11
Refactor backend getShaderLanguage functions.

rafadevai:
Allow returning a list of supported shader languages instead of only one.

Notify the backend which is the preferred language. In the case of metal, the preferred language will be the first element in the list.

This change will allow a backend to support multiple languages as needed, in the case of the noop driver, it will support a material with any shader languages.

### DIFF
--- a/filament/backend/include/backend/DriverEnums.h
+++ b/filament/backend/include/backend/DriverEnums.h
@@ -197,6 +197,7 @@ constexpr std::string_view to_string(Backend const backend) noexcept {
  * - The Metal backend can prefer precompiled Metal libraries, while falling back to MSL.
  */
 enum class ShaderLanguage {
+    UNSPECIFIED = -1,
     ESSL1 = 0,
     ESSL3 = 1,
     SPIRV = 2,
@@ -219,6 +220,8 @@ constexpr const char* shaderLanguageToString(ShaderLanguage shaderLanguage) noex
             return "Metal precompiled library";
         case ShaderLanguage::WGSL:
             return "WGSL";
+        case ShaderLanguage::UNSPECIFIED:
+            return "Unspecified";
     }
     return "UNKNOWN";
 }
--- a/filament/backend/include/backend/DriverEnums.h
+++ b/filament/backend/include/backend/DriverEnums.h
@@ -223,6 +223,27 @@ constexpr const char* shaderLanguageToString(ShaderLanguage shaderLanguage) noex
     return "UNKNOWN";
 }
 
+/*
+ * Sets a preferred shader language for Filament to use.
+ *
+ * The Metal backend supports two shader languages: MSL (Metal Shading Language) and
+ * METAL_LIBRARY (precompiled .metallib). This option controls which shader language is
+ * used when materials contain both.
+ *
+ * By default, when preferredShaderLanguage is unset, Filament will prefer METAL_LIBRARY
+ * shaders if present within a material, falling back to MSL. Setting
+ * preferredShaderLanguage to ShaderLanguage::MSL will instead instruct Filament to check
+ * for the presence of MSL in a material first, falling back to METAL_LIBRARY if MSL is not
+ * present.
+ *
+ * When using a non-Metal backend, setting this has no effect.
+ */
+enum class PreferredShaderLanguage {
+    DEFAULT = 0,
+    MSL = 1,
+    METAL_LIBRARY = 2,
+};
+
 enum class ShaderStage : uint8_t {
     VERTEX = 0,
     FRAGMENT = 1,
--- a/filament/backend/include/backend/DriverEnums.h
+++ b/filament/backend/include/backend/DriverEnums.h
@@ -223,27 +223,6 @@ constexpr const char* shaderLanguageToString(ShaderLanguage shaderLanguage) noex
     return "UNKNOWN";
 }
 
-/*
- * Sets a preferred shader language for Filament to use.
- *
- * The Metal backend supports two shader languages: MSL (Metal Shading Language) and
- * METAL_LIBRARY (precompiled .metallib). This option controls which shader language is
- * used when materials contain both.
- *
- * By default, when preferredShaderLanguage is unset, Filament will prefer METAL_LIBRARY
- * shaders if present within a material, falling back to MSL. Setting
- * preferredShaderLanguage to ShaderLanguage::MSL will instead instruct Filament to check
- * for the presence of MSL in a material first, falling back to METAL_LIBRARY if MSL is not
- * present.
- *
- * When using a non-Metal backend, setting this has no effect.
- */
-enum class PreferredShaderLanguage {
-    DEFAULT = 0,
-    MSL = 1,
-    METAL_LIBRARY = 2,
-};
-
 enum class ShaderStage : uint8_t {
     VERTEX = 0,
     FRAGMENT = 1,
--- a/filament/backend/include/private/backend/Driver.h
+++ b/filament/backend/include/private/backend/Driver.h
@@ -75,6 +75,9 @@ public:
     // The shader languages used for shaders for this driver in order of preference, used to inform
     // matdbg.
     //
+    // The `preferredLanguage` is only a hint. If the backend supports it, it will appear first in
+    // the list.
+    //
     // For OpenGL, this distinguishes whether the driver's shaders are powered by ESSL1 or ESSL3.
     // This information is used by matdbg to display the correct shader code to the web UI and patch
     // the correct chunk when rebuilding shaders live.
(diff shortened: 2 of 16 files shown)


filament/backend/include/backend/DriverEnums.h
```diff
@@ -213,6 +213,27 @@ constexpr const char* shaderLanguageToString(ShaderLanguage shaderLanguage) noex
...
+ * used when materials contain both.
+ *
+ * By default, when preferredShaderLanguage is unset, Filament will prefer METAL_LIBRARY
+ * shaders if present within a material, falling back to MSL. Setting
+ * preferredShaderLanguage to ShaderLanguage::MSL will instead instruct Filament to check
+ * for the presence of MSL in a material first, falling back to METAL_LIBRARY if MSL is not
+ * present.
+ *
+ * When using a non-Metal backend, setting this has no effect.
+ */
```
poweifeng: I think this comment should still go into Engine::Config

rafadevai: can we still include this from the Driver.h files? I think that's the part I wasnt sure about or will it introduce some weird dependency?

poweifeng: There are other enums in Engine::Config like this one. So it should be ok.

filament/backend/include/backend/DriverEnums.h
```diff
@@ -213,6 +213,27 @@ constexpr const char* shaderLanguageToString(ShaderLanguage shaderLanguage) noex
...
+enum class PreferredShaderLanguage {
+    DEFAULT = 0,
+    MSL = 1,
+    METAL_LIBRARY = 2,
+};
+
```
pixelflinger: I don't think we need this, just use `ShaderLanguage`.

filament/backend/include/private/backend/Driver.h
```diff
@@ -79,7 +80,8 @@ class Driver {
     //
     // Metal shaders can either be MSL or Metal libraries, but at time of writing, matdbg can only
     // interface with MSL.
-    virtual ShaderLanguage getShaderLanguage() const noexcept = 0;
+    virtual utils::FixedCapacityVector<ShaderLanguage> getShaderLanguage(
+            PreferredShaderLanguage preferredLanguage) const noexcept = 0;
```
pixelflinger: use `ShaderLanguage` here, and document that's only hint. If the backend supports the asked prefered language, it must appear first in the list.

filament/backend/src/noop/NoopDriver.cpp
```diff
@@ -46,8 +46,16 @@ ShaderModel NoopDriver::getShaderModel() const noexcept {
...
-    return ShaderLanguage::ESSL3;
+utils::FixedCapacityVector<ShaderLanguage> NoopDriver::getShaderLanguage(
+        PreferredShaderLanguage preferredLanguage) const noexcept {
+    return {
+        ShaderLanguage::ESSL1,
+        ShaderLanguage::ESSL3,
+        ShaderLanguage::SPIRV,
+        ShaderLanguage::MSL,
+        ShaderLanguage::METAL_LIBRARY,
+        ShaderLanguage::WGSL,
```
pixelflinger: I think ESSL3 should appear first to keep the old behavior.

filament/backend/include/backend/DriverEnums.h
rafadevai: then should I add a convert function from the Config::ShaderLanguage to the backend one?

filament/src/details/Engine.h
```diff
@@ -268,16 +268,7 @@ class FEngine : public Engine {
 
     // Return a vector of shader languages, in order of preference.
     utils::FixedCapacityVector<backend::ShaderLanguage> getShaderLanguage() const noexcept {
```
bejado, line 273: Since `getShaderLanguage` returns a vector, I think we should rename it and the new `Driver` method to `getShaderLanguages` (plural).

filament/backend/include/backend/DriverEnums.h
rafadevai: Done

filament/backend/include/backend/DriverEnums.h
rafadevai: I followed mathias suggestion and just did a conversion between both types

filament/backend/src/noop/NoopDriver.cpp
rafadevai: done

filament/src/details/Engine.h
rafadevai: done

filament/backend/include/private/backend/Driver.h
rafadevai: done

filament/include/filament/Engine.h
```diff
@@ -376,6 +375,26 @@ class UTILS_PUBLIC Engine {
...
+        * for the presence of MSL in a material first, falling back to METAL_LIBRARY if MSL is not
+        * present.
+        *
+        * When using a non-Metal backend, setting this has no effect.
+        */
+        enum class ShaderLanguage {
+            DEFAULT = 0,
+            MSL = 1,
+            METAL_LIBRARY = 2,
+        };
```
pixelflinger, line 398: why do we need this at all? why not just use `backend::ShaderLanguage` ?

rafadevai: This is what we currently have right now, I reverted back what I changed. This is comparings against my other commits of this PR

rafadevai: are we allowed to change it since its a public api?

filament/src/details/Engine.h
```diff
@@ -268,16 +268,37 @@ class FEngine : public Engine {
...
+                    case Backend::VULKAN:
+                        preferredLanguage = backend::ShaderLanguage::SPIRV;
+                        break;
+                    case Backend::METAL:
+                        preferredLanguage = backend::ShaderLanguage::MSL;
+                        break;
+                    case Backend::WEBGPU:
+                        preferredLanguage = backend::ShaderLanguage::WGSL;
+                        break;
                 }
```
poweifeng: Why is this necessary?

rafadevai: because the the 2 types are different. so we need to convert from DEFAULT to a Backend::ShaderLanguage and I also wanted to make sure we passed down something valid for each backend.

poweifeng: I interpret `DEFAULT` as let the backend decide what the preference is.  In fact, I don't think we need to do any backend-dependent action here.  Since the driver has all the relevant info.

rafadevai: so I have it as mathias suggested where the getShaderLanguages accept a backend::ShaderLanguage parameter, so that means I need to convert from one to the other.

I guess i have 3 options here.
- I can do what I have right now, where we convert default to the appropriate backend one
- I can use a default value like ESSL3, and only set a value for the metal backend
- I can change the getShaderLanguages to accept a Config::ShaderLanguage parameter instead.

Let me know which one is preferred

poweifeng: conversion is only necessary when it's not default I think. I prefer not doing any conversion except for metal.

@pixelflinger , do you have a preference?

poweifeng: Maybe there should be a ShaderLanguage::UNDEFINED? or DEFAULT?

pixelflinger: ah I see, I didn't realize we had two types with the same name.

So that makes me think that the new API we just added is flawed. Powei is right. We need to add a `backend::ShaderLanguage::UNSPECIFIED = -1` that can only be used with `getShaderLanguages()` but is never returned by it. And that returns the default list. 

This way we can easily map DEFAULT to that. Makes sense?

rafadevai: done, had to make a small change in MaterialParser otherwise it wouldn't compile

filament/backend/src/metal/MetalDriver.mm
```diff
@@ -1002,8 +1002,12 @@
 #endif
 }
 
-ShaderLanguage MetalDriver::getShaderLanguage() const noexcept {
-    return ShaderLanguage::MSL;
+utils::FixedCapacityVector<ShaderLanguage> MetalDriver::getShaderLanguages(
+        ShaderLanguage preferredLanguage) const noexcept {
+    if (preferredLanguage == backend::ShaderLanguage::MSL) {
+        return { backend::ShaderLanguage::MSL, backend::ShaderLanguage::METAL_LIBRARY};
```
z3moon: nit: space at the end?

z3moon (approved):

filament/backend/src/metal/MetalDriver.mm
rafadevai: done

PR #9208, opened 2025-09-11
Fix: Correctly handle unsupported SwapChain flags

z3moon:
FSwapChain initialization logic would filter out unsupported flags but would incorrectly use the original, unfiltered flags when creating the swap chain.

This commit ensures the filtered flags are passed to the driver. It also adds the warning to log which ones are removed.

BUGS=[433327615]

### DIFF
--- a/filament/src/details/SwapChain.cpp
+++ b/filament/src/details/SwapChain.cpp
@@ -36,13 +36,12 @@ namespace filament {
 namespace {
 
 utils::CString getRemovedFlags(uint64_t originalFlags, uint64_t modifiedFlags) {
+    const uint64_t diffFlags = originalFlags ^ modifiedFlags;
     utils::CString removed;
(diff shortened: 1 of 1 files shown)


filament/src/details/SwapChain.cpp
```diff
@@ -31,14 +33,31 @@
 
 namespace filament {
 
+namespace {
+
+utils::CString getRemovedFlags(uint64_t originalFlags, uint64_t modifiedFlags) {
+    utils::CString removed;
+    if ((originalFlags & backend::SWAP_CHAIN_CONFIG_SRGB_COLORSPACE) &&
+        !(modifiedFlags & backend::SWAP_CHAIN_CONFIG_SRGB_COLORSPACE)) {
```
poweifeng: I think you can just do an xor here?

z3moon: Yeah makes sense. Updated.

PR #9214, opened 2025-09-12
Free List UB: Undefined Behavior due to object lifetime issues

### LINKED ISSUE #9210: Free List UB: Undefined Behavior due to object lifetime issues
**The Problem**

The function `FreeList::push` performs a `static_cast `from `void*` to `Node*`

`Node* const head = static_cast<Node*>(p);` -> doesn't start Node lifetime
`head->next = mHead;` -> UB`

Accessing non-static members like `head->next` on a pointer to non-existent objects is a violation of the C++ object lifetime rules [basic.life].

**Solution**
The correct way to handle this is to use **placement new**/**std::launder** to explicitly construct the Node object in the provided memory. This ensures the object has a valid lifetime before its members are accessed.

**NB**
While this might "work" on some compilers, it's a violation of the C++ standard. The correct approach is to use something to explicitly construct the Node object, ensuring its lifetime and guaranteeing safe member access. This prevents potential, hard-to-debug crashes and ensures future code stability.

Best regards!

nacht-wache:
Resolves #9210 UB in AtomicFreeList::push and FreeList::push: access to non-statiic data member without proper lifetime management

### DIFF
--- a/libs/utils/include/utils/Allocator.h
+++ b/libs/utils/include/utils/Allocator.h
@@ -239,7 +239,9 @@ public:
     void push(void* p) noexcept {
         assert(p);
         assert(p >= mBegin && p < mEnd);
-        Node* const head = static_cast<Node*>(p);
+        // we use placement-new to properly manage the lifetime
+        // this is noop already under O1
+        Node* const head = new (p) Node;
         head->next = mHead;
         mHead = head;
     }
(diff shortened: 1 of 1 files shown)


nacht-wache:
@pixelflinger i guess i dont' have permission to run workflow checks after rebase onto main :)

PR #9217, opened 2025-09-16
add support for memory mapped buffers in the backend

pixelflinger:


pixelflinger:
note: only GL backend is implemented. I will add other backends before checking this in (at least a fallback).

filament/backend/test/test_MemoryMappedBuffer.cpp
```diff
@@ -0,0 +1,376 @@
...
+    }
+};
+
+TEST_F(MemoryMappedTest, MapCopyUnmap) {
+    auto& api = getDriverApi();
+    Cleanup cleanup(api);
+
+    // Create a buffer object.
+    BufferObjectHandle const bufferObject = cleanup.add(api.createBufferObject(1024,
+            BufferObjectBinding::VERTEX, BufferUsage::DYNAMIC | BufferUsage::SHARED_WRITE_BIT));
```
show50726: Should it be `DYNAMIC_BIT`?

filament/backend/test/test_MemoryMappedBuffer.cpp
```diff
@@ -0,0 +1,376 @@
...
+        render(3, 0, swapChain, renderTarget, renderPrimitive, state);
+
+        EXPECT_IMAGE(renderTarget, ScreenshotParams(screenWidth(), screenHeight(), screenshotName, 0));
+    }
+};
+
+TEST_F(MemoryMappedTest, MapCopyUnmap) {
+    auto& api = getDriverApi();
+    Cleanup cleanup(api);
+
```
poweifeng, line 211: Can you skip the other backends for now?

filament/backend/include/backend/DriverEnums.h
```diff
@@ -556,12 +556,21 @@ constexpr TargetBufferFlags getTargetBufferFlagsAt(size_t index) noexcept {
...
- * for the driver to make better decisions about managing memory internally.
+ * How the buffer will be used.
  */
 enum class BufferUsage : uint8_t {
-    STATIC,      //!< content modified once, used many times
-    DYNAMIC,     //!< content modified frequently, used many times
+    STATIC              = 0,    //!< (legacy) content modified once, used many times
+    DYNAMIC             = 1,    //!< (legacy) content modified frequently, used many times
+    DYNAMIC_BIT         = 0x1,  //!< buffer can be modified frequently, used many times
+    SHARED_WRITE_BIT    = 0x04, //!< buffer can be memory mapped for write operations
```
poweifeng, line 565: nit: why jump to 0x4?

poweifeng (approved):
minor comments

filament/backend/test/test_MemoryMappedBuffer.cpp
pixelflinger: It should be but both are compatible.

filament/backend/include/backend/DriverEnums.h
pixelflinger: I just left space for `SHARED_READ_BIT`, so they stay together if we add another `BIT` later.

filament/backend/test/test_MemoryMappedBuffer.cpp
pixelflinger: done

filament/backend/test/test_MemoryMappedBuffer.cpp
pixelflinger: I did better than that, I implemented basic support for the API so that the unit tests will pass (and the api should be fully functional, albeit, not fast).

filament/backend/src/vulkan/VulkanDriver.cpp
```diff
@@ -476,6 +476,22 @@ void VulkanDriver::updateDescriptorSetTexture(
     }
 }
 
+void VulkanDriver::copyToMemoryMappedBuffer(MemoryMappedBufferHandle mmbh, size_t offset,
+        size_t size, BufferDescriptor&& data) {
+    FVK_SYSTRACE_SCOPE();
+    auto mmb = resource_ptr<VulkanMemoryMappedBuffer>::cast(&mResourceManager, mmbh);
+
+    assert_invariant(any(mmb->access & MapBufferAccessFlags::WRITE_BIT));
+    assert_invariant(size <= data.size);
```
poweifeng: This is probably ok as a temporary implementation.  

I do have a concern. The assumption here is that we're using a subset of `data` to update a portion of `size` in the mapped memory.   But `updateBufferObject` assumes that we're updating a buffer using the size of `data`. (that is, `size` is not pass to `updateBufferObject`).

poweifeng (approved):
vk lgtm in general.

one clarification needed.

filament/backend/src/vulkan/VulkanDriver.cpp
pixelflinger: This is a very good point! I just don't need the size parameter in this backend method.

filament/backend/include/private/backend/DriverAPI.inc
```diff
@@ -291,27 +291,11 @@ DECL_DRIVER_API_TAGGED_R_N(backend::DescriptorSetLayoutHandle, createDescriptorS
 DECL_DRIVER_API_TAGGED_R_N(backend::DescriptorSetHandle, createDescriptorSet,
         backend::DescriptorSetLayoutHandle, dslh)
 
-DECL_DRIVER_API_N(updateDescriptorSetBuffer,
-        backend::DescriptorSetHandle, dsh,
-        backend::descriptor_binding_t, binding,
+DECL_DRIVER_API_TAGGED_R_N(backend::MemoryMappedBufferHandle, mapBuffer,
```
rafadevai, line 294: So why not just map the whole memory when you create the buffer and unmap when you destroy it? 

This is really expensive call to make and there's no benefit on mapping/unmapping sections of the memory.

filament/backend/include/private/backend/DriverAPI.inc
```diff
@@ -447,6 +432,26 @@ DECL_DRIVER_API_N(compilePrograms,
...
+)
+
+DECL_DRIVER_API_N(updateDescriptorSetTexture,
+        backend::DescriptorSetHandle, dsh,
+        backend::descriptor_binding_t, binding,
+        backend::TextureHandle, th,
+        SamplerParams, params
+)
+
+DECL_DRIVER_API_N(copyToMemoryMappedBuffer,
```
rafadevai, line 450: why not just use the unsynchronized copy? if the BO is mapped, just copy directly to memory

filament/backend/include/private/backend/DriverAPI.inc
pixelflinger: In GLES you can't use the buffer in the GPU if it's mapped; so we need this.

But for your vulkan implementation you don't have to map/unmap each time this is called. This is just an API/interface. You can absolutely map the buffer once when it's created (And you now know that because I give you the SHARED_WRITE flag).

Also for the case where you don't have UMA, you might need this to flush the content efficiently.

filament/backend/include/private/backend/DriverAPI.inc
pixelflinger: but it's not mapped at the time you call mapBuffer, these are commands queued into the backend command queue.   But it doesn't matter because that's exactly what `copyToMemoryMappedBuffer` does, it doesn't do any extra copy, it just "Queues" a `memcpy` that will happen later on the backend thread.

Also, you need this for drivers that don't support memory mapped buffers.

The filament side can *never* have access to the pointer, this is by design so we are free to implement it however we want.

filament/backend/include/backend/DriverEnums.h
```diff
@@ -556,12 +556,21 @@ constexpr TargetBufferFlags getTargetBufferFlagsAt(size_t index) noexcept {
...
+};
+
+/**
+ * How the buffer will be mapped.
+ */
+enum class MapBufferAccessFlags : uint8_t {
+    WRITE_BIT               = 0x2,  //!< buffer is mapped from writing
+    INVALIDATE_RANGE_BIT    = 0x4,  //!< the mapped range content is lost
```
bejado, line 573: Should the first comment read "buffer is mapped _for_ writing"?

Can you elaborate on `INVALIDATE_RANGE_BIT`? Does this mean that when the CPU maps a range, the CPU cannot assume anything about the contents of that range?

filament/backend/src/metal/MetalHandles.h
```diff
@@ -507,6 +507,18 @@ struct MetalDescriptorSet : public HwDescriptorSet {
...
 };
 
+
+struct MetalMemoryMappedBuffer : public HwMemoryMappedBuffer {
+    MetalMemoryMappedBuffer(BufferObjectHandle boh, size_t const offset,
+            size_t const size, MapBufferAccessFlags const access)
+        : boh(boh), access(access), size(size), offset(offset) {
+    }
+    BufferObjectHandle boh{};
+    MapBufferAccessFlags access{};
```
bejado, line 517: Should these go in `HwMemoryMappedBuffer`?

filament/backend/include/backend/DriverEnums.h
pixelflinger: > Should the first comment read "buffer is mapped _for_ writing"?
> 
> Can you elaborate on `INVALIDATE_RANGE_BIT`? Does this mean that when the CPU maps a range, the CPU cannot assume anything about the contents of that range?

yes. It essentially implies that the CPU will fill the whole range with new data. For non UMA systems, it means you don't have to copy-back the content of the buffer before mapping.

filament/backend/src/metal/MetalHandles.h
pixelflinger: they could, but I decided that it was a mistake to stash stuff in  the `Hw*` because it forces things to be saved that a given backend might not even need, wasting space. I'd like to go through and change all of them, one day.

PR #9219, opened 2025-09-16
renderdiff: enable webgpu and improvements

poweifeng:
- Allow for presets to override selected models when presented in order.
 - Add 'gltf' for model search path
 - Add a few simple gltf models to the 'base' preset
 - Improve UI so that missing tests do not generate any html bits.
 - Add documentation on using the viewer
 - Add renderdiff documentation to the project webpage.

RDIFF_BRANCH=pf/renderdiff-enable-webgpu


test/renderdiff/src/viewer_html/app.js
```diff
@@ -420,12 +399,27 @@ class App extends LitElement {
     this.selectedTests = [];
     this.comparisonContent = null;
     this.compareMode = false;
+    this.missingFile = {
+      ['/d/undefined']: true,
```
bejado: Maybe use `getDiffUrl` or something here for readability? I wasn't sure what the `/d/` meant at first.

poweifeng: good point. changed!

PR #9221, opened 2025-09-17
fix OpenGL backend tests on macOS

pixelflinger:
multi-sampled textures are only supported at feature level 2, but on macOS OpenGL is only feature level 1.

### DIFF
--- a/filament/backend/test/test_Blit.cpp
+++ b/filament/backend/test/test_Blit.cpp
@@ -269,6 +269,10 @@ TEST_F(BlitTest, ColorResolve) {
     SKIP_IF(Backend::WEBGPU, "test cases fail in WebGPU, see b/424157731");
     auto& api = getDriverApi();
 
+    if (api.getFeatureLevel() < FeatureLevel::FEATURE_LEVEL_2) {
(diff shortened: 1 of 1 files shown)


filament/backend/test/test_Blit.cpp
```diff
@@ -269,6 +269,10 @@ TEST_F(BlitTest, ColorResolve) {
     SKIP_IF(Backend::WEBGPU, "test cases fail in WebGPU, see b/424157731");
     auto& api = getDriverApi();
 
+    if (api.getFeatureLevel() < FeatureLevel::FEATURE_LEVEL_2) {
+        GTEST_SKIP() << "Skipping test because multi-sampled textures are not supported at Feature level < 2";
+    }
```
poweifeng, line 274: I wasn't aware of this.  But I guess vk and metal both have ms textures by default?

What does this mean in practice?  The client couldn't request msaa unless they also request FL2?

pixelflinger: no, MSAA works regardless because it uses MSRTT, or, uses an intermediate renderbuffer. I mean, Filament deals with it. Renderbuffers (i.e. Textures that are not sampleable) don't have that limitation in GLES.

PR #9229, opened 2025-09-18
Fix Metal LoadImageTest.UpdateImageMipLevel

bejado:
`EXPECT_IMAGE` must be called within a `RenderFrame` block, otherwise `~RenderFrame` calls `api.endFrame()` too soon.


pixelflinger (approved):

poweifeng (changes requested):
there's a call to `beginFrame` on line 522 that should be removed i think.

bejado:
> there's a call to `beginFrame` on line 522 that should be removed i think.

Good call, that's redundant now.

PR #9235, opened 2025-09-19
Cache isMSAASwapChainSupported() result for EGL

z3moon:
This change introduces a caching mechanism to store the results of MSAA support queries. The results are stored in a vector, indexed by the power-of-two of the sample count.

### DIFF
--- a/filament/backend/include/backend/platforms/PlatformEGL.h
+++ b/filament/backend/include/backend/platforms/PlatformEGL.h
@@ -28,6 +28,7 @@
 #include <utils/Invocable.h>
 
 #include <initializer_list>
+#include <optional>
 #include <utility>
 #include <vector>
 
@@ -168,6 +169,9 @@ protected:
     EGLConfig mEGLConfig = EGL_NO_CONFIG_KHR;
     Config mContextAttribs;
     std::vector<EGLContext> mAdditionalContexts;
+    // A cache for MSAA support queries. The index of the vector corresponds to the power of 2 of
+    // the sample count. For example, index 0 is for 2 samples, 1 is for 4 samples, etc.
+    mutable std::vector<std::optional<bool>> mMSAASupport{4}; // caches for 2x, 4x, 8x, 16x
 
     // supported extensions detected at runtime
     struct {
--- a/filament/backend/include/backend/platforms/PlatformEGL.h
+++ b/filament/backend/include/backend/platforms/PlatformEGL.h
@@ -28,7 +28,6 @@
 #include <utils/Invocable.h>
 
 #include <initializer_list>
-#include <optional>
 #include <utility>
 #include <vector>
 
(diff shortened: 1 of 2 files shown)


filament/backend/include/backend/platforms/PlatformEGL.h
```diff
@@ -168,6 +169,9 @@ class PlatformEGL : public OpenGLPlatform {
     EGLConfig mEGLConfig = EGL_NO_CONFIG_KHR;
     Config mContextAttribs;
     std::vector<EGLContext> mAdditionalContexts;
+    // A cache for MSAA support queries. The index of the vector corresponds to the power of 2 of
+    // the sample count. For example, index 0 is for 2 samples, 1 is for 4 samples, etc.
+    mutable std::vector<std::optional<bool>> mMSAASupport{4}; // caches for 2x, 4x, 8x, 16x
```
pixelflinger: that should be a `std::array<bool>`, no need for `optional` and not need for a `std::vector`.

poweifeng: how about just a bitset?

pixelflinger (commented):
This is too complicated. We don't need to handle anything else than x4 for now, until someone asks for it. x8 and x16 are very uncommon on mobile, and x2 is virtually useless.

I don't want any code in `isMSAA...Supported()` this should just be a `return` statement. 

The check for x4 can be done once during init.

pixelflinger (changes requested):
see comments

filament/backend/include/backend/platforms/PlatformEGL.h
z3moon: done

z3moon: Yeah that was my first thought, but the original design had three states (not cached yet / not supported / supported).

filament/backend/src/opengl/platforms/PlatformEGL.cpp
```diff
@@ -491,34 +493,13 @@ bool PlatformEGL::isMSAASwapChainSupported(uint32_t samples) const noexcept {
...
-    if (!eglChooseConfig(mEGLDisplay, configAttribs.data(), &config, 1, &configsCount)) {
-        return false;
+    if (samples == 4) {
+        return mMSAA4XSupport;
     }
 
-    return configsCount > 0;
+    // Other sample counts are not cached, retrieve it.
+    LOG(INFO) << "MSAA sample count " << samples << " is queried, consider caching it.";
+    return checkIfMSAASwapChainSupported(samples);
```
pixelflinger: just return false. I don't want anything else in this function so it can be `noexcept`.

filament/backend/src/opengl/platforms/PlatformEGL.cpp
```diff
@@ -107,7 +107,9 @@ void PlatformEGL::clearGlError() noexcept {
 
 // ---------------------------------------------------------------------------------------------
 
-PlatformEGL::PlatformEGL() noexcept = default;
+PlatformEGL::PlatformEGL() noexcept : OpenGLPlatform() {
+    mMSAA4XSupport = checkIfMSAASwapChainSupported(4);
```
pixelflinger: this needs to be called from createDriver, because at this point `eglInitialize()` has not even been called.

pixelflinger (changes requested):

filament/backend/src/opengl/platforms/PlatformEGL.cpp
z3moon: done

filament/backend/src/opengl/platforms/PlatformEGL.cpp
z3moon: done

PR #9237, opened 2025-09-19
Metal: use atomic for buffer tracking

bejado:
BUGS = [438574508]


z3moon (approved):

pixelflinger (changes requested):
I don't think those should be static. There could be several instances of  filament in the process.

In fact we should review *all* static variables usage.

It would be preferable that those live in MetalDriver or MetalContext.

PR #9240, opened 2025-09-22
gl: disable fb fetch for llvmpipe

poweifeng:


filament/backend/src/opengl/OpenGLDriver.cpp
```diff
@@ -2478,6 +2478,9 @@ bool OpenGLDriver::isRenderTargetFormatSupported(TextureFormat format) {
 
 bool OpenGLDriver::isFrameBufferFetchSupported() {
     auto& gl = mContext;
+    if (gl.bugs.disable_framebuffer_fetch_extension) {
+        return false;
+    }
```
pixelflinger: I think it would be better to do this where (or just after) `gl.ext.EXT_shader_framebuffer_fetch` is set. This way the whole backend is coherent (in case gl.ext.EXT_shader_framebuffer_fetch is checked directly somewhere else), and we don't have to do to checks here.

pixelflinger (changes requested):

filament/backend/src/opengl/OpenGLDriver.cpp
poweifeng: done

PR #9251, opened 2025-09-24
vk: fix update after bind (wrt descriptor set) validation error

poweifeng:
In filament, once a descriptor set is bound, we no longer make updates
to it. However, this guarrantee was broken in the external sampler/image
because a colorspace change might necessitate a layout change, and
thereby a new descriptor needs to be generated with the appropriate
externally sampled image updated in the new set.

Another use case is the Stream API, where each frame we might be
getting a different AHardwareBuffer, meaning we need to update one
or multiple sampler bindings in a set that might have been bound in
a previous frame.

In this change, we always create a new set when there's a change
in the image currently bound to an existing set (while accounting for
whether we need to use a new externally sampled layout or not).


rafadevai:
but if the image changed its colorspace, wouldnt that mean we got a new texture called with setParameter (not in the stream) which implies the frontend will create a new descriptor set?

filament/backend/src/vulkan/VulkanDriver.cpp
```diff
@@ -2115,7 +2116,7 @@ void VulkanDriver::draw2(uint32_t indexOffset, uint32_t indexCount, uint32_t ins
         }
         mPipelineState.bindInDraw.first = false;
     }
-    mDescriptorSetCache.commit(mCurrentRenderPass.commandBuffer, mPipelineState.pipelineLayout,
+    mDescriptorSetCache.commit(bufferHolder, mPipelineState.pipelineLayout,
```
pixelflinger: is there a way this could behind a very fast "if unlikely"? I feel like `commit` is pretty heavy.

filament/backend/src/vulkan/VulkanHandles.h
```diff
@@ -159,38 +162,14 @@ struct VulkanDescriptorSet : public HwDescriptorSet, fvkmemory::Resource {
...
-            uint8_t uniqueDynamicUboCount,
-            OnRecycle&& onRecycleFn, VkDescriptorSet vkSet)
-        : dynamicUboMask(dynamicUboMask),
-          uniqueDynamicUboCount(uniqueDynamicUboCount),
-          mVkSet(vkSet),
-          mOnRecycleFn(std::move(onRecycleFn)) {}
+            fvkmemory::resource_ptr<VulkanDescriptorSetLayout> layout,
+            OnRecycle&& onRecycleFn, VkDescriptorSet vkSet);
 
     // NOLINTNEXTLINE(bugprone-exception-escape)
```
rafadevai, line 168: no need for this anymore?

phoenixxxx (approved):

rafadevai (changes requested):
Last time there was more validation errors, then before. We need to make sure those are fixed before merging this

poweifeng:
merged (and fixed) as part of #9164

PR #9250, opened 2025-09-24
Fix matdbg build

pixelflinger:

PR #9252, opened 2025-09-24
Fix a compile error

z3moon:
FMaterial accesses a private static method of the MaterialDefinition class, which incurs a compile error for certain compilers.


poweifeng:
There's the same fix in #9250

z3moon:
Ah let me close this.

PR #9253, opened 2025-09-24
material cache: address Powei's comments

elizagamedev:


filament/src/MaterialParser.cpp
```diff
@@ -111,6 +112,17 @@ bool MaterialParser::operator==(MaterialParser const& rhs) const noexcept {
...
+    std::optional<uint32_t> rhsCrc32;
+    if (!getMaterialCrc32(&lhsCrc32.value())) {
+        lhsCrc32 = mCrc32.load(std::memory_order_relaxed);
+    }
+    if (!rhs.getMaterialCrc32(&rhsCrc32.value())) {
+        rhsCrc32 = rhs.mCrc32.load(std::memory_order_relaxed);
+    }
+    if (lhsCrc32 && rhsCrc32 && *lhsCrc32 != *rhsCrc32) {
+        return false;
+    }
```
poweifeng: this logic is pretty dense. What if you just did `computeCrc32() != rhs.computeCrc32()` ? I guess that would miss the stored crc32, but maybe that's ok?

elizagamedev: I pushed another commit to simplify it a bit. What do you think?

filament/src/MaterialParser.cpp
```diff
@@ -172,6 +183,11 @@ MaterialParser::ParseResult MaterialParser::parse() noexcept {
 }
 
 uint32_t MaterialParser::computeCrc32() const noexcept {
+    std::optional<uint32_t> cachedCrc32 = mCrc32.load(std::memory_order_relaxed);
+    if (cachedCrc32) {
+        return *cachedCrc32;
+    }
```
pixelflinger: Why use a std::optional<> here?

filament/src/MaterialParser.cpp
```diff
@@ -181,7 +197,21 @@ uint32_t MaterialParser::computeCrc32() const noexcept {
...
+    uint32_t crc32 = utils::hash::crc32Update(0, payload, originalSize, crc32Table);
+    mCrc32.store(crc32, std::memory_order_relaxed);
+    return crc32;
+}
+
+std::optional<uint32_t> MaterialParser::getPrecomputedCrc32() const noexcept {
+    std::optional<uint32_t> cachedCrc32 = mCrc32.load(std::memory_order_relaxed);
+    if (cachedCrc32) {
+        return cachedCrc32;
+    }
```
pixelflinger, line 211: same here, the line below returns a uint32_t, this should use the same pattern, no need to create a std::optional explicitely

filament/src/MaterialParser.h
```diff
@@ -69,7 +69,11 @@ class MaterialParser {
 
     ParseResult parse() noexcept;
 
+    // Compute the CRC32 of the material or return the cached value.
     uint32_t computeCrc32() const noexcept;
+    // Return the cached computed CRC32 or the CRC32 built into the material file if one exists.
+    std::optional<uint32_t> getPrecomputedCrc32() const noexcept;
```
pixelflinger, line 75: why not use 0 to indicate it's not ready?  I find std::optional overkill here.

filament/src/MaterialParser.h
```diff
@@ -187,6 +191,7 @@ class MaterialParser {
     filaflat::ChunkContainer& getChunkContainer() noexcept;
     filaflat::ChunkContainer const& getChunkContainer() const noexcept;
     MaterialParserDetails mImpl;
+    mutable std::atomic<std::optional<uint32_t>> mCrc32;
```
pixelflinger: I don't like this because this doesn't gurantee that mCrc32 won't use a lock. it may or may not.
if you need an atomic, just store the crc and use 0 to mean it's not there.

pixelflinger (changes requested):

pixelflinger:
also change the title and commit message, because I don't think public commit should have our names in them, it's not very informative :-)

filament/src/MaterialParser.h
elizagamedev: What if the CRC32 is actually 0?

pixelflinger (approved):

PR #9267, opened 2025-09-26
new fenceWait() API in the backend

pixelflinger:


filament/backend/src/vulkan/VulkanDriver.cpp
```diff
@@ -1156,29 +1158,30 @@ void VulkanDriver::destroyFence(Handle<HwFence> fh) {
...
+        return FenceStatus::TIMEOUT_EXPIRED;
     }
 
-
-    // Two other states are possible:
-    //   - VK_INCOMPLETE: the corresponding buffer has not yet been submitted.
-    //   - VK_NOT_READY: the buffer has been submitted but not yet signaled.
-    //     In either case, we return TIMEOUT_EXPIRED to indicate the fence has not been signaled.
-    return FenceStatus::TIMEOUT_EXPIRED;
+    // now we are holding a reference to our VulkanCmdFence, so we now it can't
```
poweifeng: ### ### 
```suggestion
    // now we are holding a reference to our VulkanCmdFence, so we know it can't
```

filament/backend/src/vulkan/VulkanDriver.cpp
```diff
@@ -1156,29 +1158,30 @@ void VulkanDriver::destroyFence(Handle<HwFence> fh) {
...
+    // disappear while we wait
+    return cmdfence->wait(mPlatform->getDevice(), timeout, until);
```
poweifeng, line 1184: nit: technically, the timeout here needs to be subtracted by the amount waited above, same with until.

poweifeng (commented):
vk and webgpu lgtm

filament/backend/src/vulkan/VulkanDriver.cpp
pixelflinger: yes, lol, I was thinking about that and then I thought... whatever :-)

PR #9269, opened 2025-09-27
Fix a crash for MaterialCache

z3moon:
The parser object was accessed to generate a key after being passed to the lambda using std::move(). This led to a crash on certain platforms (at least on Windows).

This change fixes it by creating key first before moving the parser object into the lambda.


CYH0613:
save my ass ,bro

PR #9271, opened 2025-09-29
fix several issues CircularQueue

pixelflinger:
- pop_back() didn't call the object's dtor
- added a dtor so objects can't be leaked
- made it moveable but not copiable
- added unit tests

### DIFF
--- a/filament/src/FrameInfo.h
+++ b/filament/src/FrameInfo.h
@@ -51,14 +51,36 @@ struct FrameInfo {
 struct FrameInfoImpl : public details::FrameInfo {
     using clock = std::chrono::steady_clock;
     using time_point = clock::time_point;
-    uint32_t const frameId;
+    uint32_t frameId;
     time_point beginFrame;           // main thread beginFrame time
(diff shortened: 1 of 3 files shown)


filament/src/FrameInfo.h
```diff
@@ -51,14 +51,38 @@ struct FrameInfo {
...
+    }
+
+    FrameInfoImpl(FrameInfoImpl&& rhs) noexcept : 
+        details::FrameInfo(rhs),
+        frameId(rhs.frameId),
+        beginFrame(rhs.beginFrame),
+        endFrame(rhs.endFrame),
+        backendBeginFrame(rhs.backendBeginFrame),
+        backendEndFrame(rhs.backendEndFrame),
+        gpuFrameComplete(rhs.gpuFrameComplete),
```
poweifeng: this field is not defined it seems

pixelflinger: ah yes, rebase error. thanks.

poweifeng (approved):

z3moon (approved):

PR #9273, opened 2025-09-30
Fix compile issues

z3moon:


filament/src/MaterialParser.cpp
```diff
@@ -124,7 +125,7 @@ bool MaterialParser::operator==(MaterialParser const& rhs) const noexcept {
             return false;
         }
     }
-    return !std::memcmp(mImpl.mManagedBuffer.data(), rhs.mImpl.mManagedBuffer.data(),
+    return !memcmp(mImpl.mManagedBuffer.data(), rhs.mImpl.mManagedBuffer.data(),
```
elizagamedev, line 128: Why not `#include <cstring>` and `std::memcmp` here?

z3moon: Yeah, we prefer C headers to C++ in general if there're multiple options.

elizagamedev: Hmm, okay

elizagamedev (approved):

poweifeng (approved):

poweifeng (approved):

pixelflinger (approved):

PR #9274, opened 2025-09-30
Support C++17 for RefCountedMap

z3moon:
Some of our users are sticking to C++17 for a while.

### DIFF
--- a/filament/backend/src/opengl/OpenGLDriver.cpp
+++ b/filament/backend/src/opengl/OpenGLDriver.cpp
@@ -1827,7 +1827,7 @@ void OpenGLDriver::createFenceR(Handle<HwFence> fh, CString&& tag) {
 
     // This is the case where we need to use OpenGL fences
 #ifndef FILAMENT_SILENCE_NOT_SUPPORTED_BY_ES2
-    std::weak_ptr const weak = f->state;
+    std::weak_ptr<GLFence::State> const weak = f->state;
     whenGpuCommandsComplete([weak] {
         if (auto const state = weak.lock()) {
             std::lock_guard const lock(state->lock);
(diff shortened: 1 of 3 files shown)


libs/utils/include/utils/RefCountedMap.h
```diff
@@ -22,24 +22,62 @@
...
 namespace refcountedmap {
 
+#if FILAMENT_CPP20
+
+template<typename T>
+concept is_pointer_like = requires(T a) { *a; };
+
 template<typename T>
-concept IsPointer = requires(T a) { *a; };
+inline constexpr bool is_pointer_like_v = is_pointer_like<T>;
```
poweifeng: Why change this part?

libs/utils/include/utils/RefCountedMap.h
```diff
@@ -22,24 +22,62 @@
...
+template<typename T, typename = void>
+struct PointerTraits {
+    using element_type = T;
+};
+
+template<typename T>
+struct PointerTraits<T, std::enable_if_t<is_pointer_like_v<T>>> {
+    using element_type = typename std::pointer_traits<T>::element_type;
+};
+
```
poweifeng, line 51: Maybe just follow the existing name of IsPointer?

poweifeng:
> It would be a lot simpler to just keep the C++17 implementation and delete the C++20 one IMO. That way we don't have to maintain two versions.

This sounds good to me too.

z3moon:
Yeah I also thought about it if you're okay. Let me remove them.

libs/utils/include/utils/RefCountedMap.h
z3moon: I followed the implicit naming convention. But maybe that's better. Done.

Issue #8611, opened 2025-04-09
Abseil combile/install breaks projects that depend on Abseil themselves


emezeske:
**Describe the bug**
PR #8387 appears to break any projects that depend on both Filament and Abseil. For example my project depends on Abseil directly (using vcpkg), and after this commit, CMake configuration fails with a bunch of errors like:

```
[cmake] CMake Error at /.../vcpkg/scripts/buildsystems/vcpkg.cmake:639 (_add_library):
[cmake]   _add_library cannot create ALIAS target "absl::bad_variant_access" because
[cmake]   another target with the same name already exists.
[cmake] Call Stack (most recent call first):
[cmake]   lib/filament/third_party/abseil/CMake/AbseilHelpers.cmake:356 (add_library)
[cmake]   lib/filament/third_party/abseil/absl/types/CMakeLists.txt:202 (absl_cc_library)
```

It seems that prior to this commit, Abseil was a private dependency of Filament, and it was possible for clients to use their own version of Abseil without conflict. But now, clients have to use whatever version of Abseil is baked-into Filament?

Is there a way around this, i.e. a flag to revert to the prior private dependency behavior?

**To Reproduce**
1. Build a project that depends on Abseil directly with Filament checked out post #8387 

**Expected behavior**
It compiles

**Screenshots**
N/A

**Logs**
N/A

**Desktop (please complete the following information):**
 - OS: Windows, MacOS

**Smartphone (please complete the following information):**
N/A

**Additional context**
N/A

emezeske:
Okay, I think that the issue is the new unconditional call to add_subdirectory in third_party\abseil\tnt\CMakeLists.txt:

`add_subdirectory(../ "${CMAKE_CURRENT_BINARY_DIR}/abseil")
`
If I comment this out, it's possible to build my project without issues linking against Abseil, because my parent CMakeLists.txt file calls add_subdirectory() for the version of Abseil that I want to use.

I think that the right thing for a library like Filament to do is wrap all the add_subdirectory() calls for third_party dependencies in something roughly like this:

```
find_package(absl CONFIG QUIET)
if(NOT absl_FOUND)
  add_subdirectory(../ "${CMAKE_CURRENT_BINARY_DIR}/abseil")
endif()
```

[...]

PR #9275, opened 2025-09-30
Some trivial changes to compile under most recent MSVC as a CMake sub-project

emezeske:
I pulled the most recent Filament changes this week into my project that uses it as a CMake sub-project, compiling under the most recent MSVC, and these changes were required to get it to compile.

This is related also to #8611.

### DIFF
--- a/third_party/abseil/tnt/CMakeLists.txt
+++ b/third_party/abseil/tnt/CMakeLists.txt
@@ -1,5 +1,8 @@
-find_package(absl CONFIG QUIET)
-if(NOT absl_FOUND)
+option(FILAMENT_USE_EXTERNAL_ABSL "Use find_package to locate an external Abseil dependency" OFF)
+
+if (FILAMENT_USE_EXTERNAL_ABSL)
+  find_package(absl CONFIG QUIET)
+else()
   add_subdirectory(../ "${CMAKE_CURRENT_BINARY_DIR}/abseil")
 endif()
 


filament/CMakeLists.txt
```diff
@@ -729,6 +729,7 @@ if (MSVC)
 else()
     set(FILAMENT_WARNINGS
             -Wall -Wextra -Wno-unused-parameter -Wno-missing-field-initializers
+            -Wno-return-type
```
z3moon: do you know what functions don't return a value? we want to fix them unless they're third_party dependencies.

libs/utils/include/utils/RefCountedMap.h
```diff
@@ -22,6 +22,8 @@
 
 #include <tsl/robin_map.h>
 
+#include <optional>
```
z3moon, line 25: this is fixed. please rebase.

third_party/abseil/tnt/CMakeLists.txt
```diff
@@ -1,4 +1,7 @@
-add_subdirectory(../ "${CMAKE_CURRENT_BINARY_DIR}/abseil")
+find_package(absl CONFIG QUIET)
```
z3moon: do you specifically want to use your own absl already installed on your windows? what's the context here?

z3moon (changes requested):

libs/utils/include/utils/RefCountedMap.h
emezeske: It's MSVC being ridiculous here -- the functions are things where there is a switch statement, and the cases fully cover the possible values for an enum:

switch (x) {
  case A: return 1;
  case B: return 2;
}

This is obviously safe as long as the cases are exhaustive, but MSVC still complains.

third_party/abseil/tnt/CMakeLists.txt
emezeske: Yes, my project uses its own absl.

The change here is extremely standard for CMake projects, in general *all* cases where a library like Filament has a dependency, this is the correct thing to do, so that CMake projects that include Filament can make their own choices about what library versions to use. CMake sub-projects should ideally never force library choices in a way that cannot be customized by the caller.

(I did not go through and correct all cases, though.)

filament/CMakeLists.txt
emezeske: It's MSVC being ridiculous here -- the functions are things where there is a switch statement, and the cases fully cover the possible values for an enum:

switch (x) {
case A: return 1;
case B: return 2;
}

This is obviously safe as long as the cases are exhaustive, but MSVC still complains.

libs/utils/include/utils/RefCountedMap.h
emezeske: oops wrong thread --will rebase!

emezeske (commented):
Even after rebasing, this <optional> is needed for MSVC

emezeske:
PTAL

emezeske:
Here are all the warnings I'm seeing in latest MSVC, in case you'd prefer to add default return statements to the functions instead:

image

Note that the View.h warnings also really should be fixed up (the syntax is wrong, should be `#pragma warning(push)` and `#pragma warning(disable : 4996)`

third_party/abseil/tnt/CMakeLists.txt
pixelflinger: I'm not familiar with `find_package()`, but what we do not want is to have to rely on abseil being installed on the system.  For filament we want a hermetic build (you download it and it builds without having to install dependencies).  If you have the need to use your own abseil, I'm okay with this as long as it's an option that's off by default.

pixelflinger:
> Here are all the warnings I'm seeing in latest MSVC, in case you'd prefer to add default return statements to the functions instead:

If there aren't too many, I think it would be best to add a return statement everywhere it's missing, that's a bit safer/cleaner.

emezeske:
> If there aren't too many, I think it would be best to add a return statement everywhere it's missing, that's a bit safer/cleaner.

I looked into it a bit, and there are more than I saw originally. This pattern is used all over the codebase (see below). For now I've reverted my -Wno-return-type in this PR and will find a way to handle this in my own CMake code.

A general note: there are a TON of warnings of all kinds (thousands?) that show up compiling under a modern MSVC, with the warning flags that Filament sets by default. How many are the compiler being paranoid, and how many are useful, I cannot say. But it may make sense for someone to compile it under MSVC and do a triage to figure out which ones are useful.

image

third_party/abseil/tnt/CMakeLists.txt
emezeske: The pattern in this PR should do what you want: it will use the hermetic Filament-provided abseil by default, but allows the super-project to override that if desired by providing its own abseil. This is an extremely common CMake pattern for open source libraries that have their own dependencies.

Somehow you have to solve the diamond problem, or else Filament is simply not compatible with projects that share any of the dependencies that Filament has. AFAICT this is the standard way of doing that.

If you want to make it impossible for Filament to find a system abseil without the super-project explicitly requesting it, we could wrap the `find_package(absl CONFIG QUIET)` in a condition like FILAMENT_USE_EXTERNAL_ABSL. I don't think this is strictly necessary, but would guarantee that using an external abseil cannot be done by accident.

third_party/abseil/tnt/CMakeLists.txt
```diff
@@ -1,4 +1,7 @@
...
+if(NOT absl_FOUND)
+  add_subdirectory(../ "${CMAKE_CURRENT_BINARY_DIR}/abseil")
+endif()
```
pixelflinger: I don't know if I like this. This breaks the hermetic build. Now it's impossible to tell if two versiona of filaments are built the same way.
This check should be based on a compile time flag. Like `-DFILAMENT_USE_SYSTEM_ABSEIL` or something like that

poweifeng (approved):

z3moon (approved):

third_party/abseil/tnt/CMakeLists.txt
pixelflinger: could we do it this way for now? i.e. have a `FILAMENT_USE_EXTERNAL_ABSL`?

pixelflinger (changes requested):
for now I'd prefer that we introduce `FILAMENT_USE_EXTERNAL_ABSL`.

third_party/abseil/tnt/CMakeLists.txt
emezeske: Yessir -- PTAL

patrikhuber:
I'm also seeing `Dependencies\filament-v1.65.0-windows\include\filament\Color.h(213): warning C4715: 'filament::Color::toLinear': not all control paths return a value`, which is an issue if one treats warnings as errors in their projects and includes filament. What fix would you prefer for this?

emezeske:
Hey @pixelflinger does this look OK? I added the FILAMENT_USE_EXTERNAL_ABSL flag.

emezeske:
Thanks!

PR #9276, opened 2025-09-30
utils: split Slice into mutable and constant types

elizagamedev:
We were doing some Not Good™️ const casting stuff with `Slice`. This CL changes `Slice` to a const type and introduces `MutSlice`. If we expect that clients are depending on `Slice` and that this will be a breaking change for them, we can further rename `Slice` to `ConstSlice` and reintroduce a deprecated compatibility `Slice` which provides the old bad behavior.

### DIFF
--- a/filament/backend/src/opengl/OpenGLDriver.cpp
+++ b/filament/backend/src/opengl/OpenGLDriver.cpp
@@ -394,7 +394,7 @@ void OpenGLDriver::setPushConstant(ShaderStage const stage, uint8_t const index,
     }
 #endif
 
-    Slice<std::pair<GLint, ConstantType>> constants;
+    Slice<const std::pair<GLint, ConstantType>> constants;
     if (stage == ShaderStage::VERTEX) {
         constants = mCurrentPushConstants->vertexConstants;
     } else if (stage == ShaderStage::FRAGMENT) {
--- a/filament/backend/src/opengl/OpenGLProgram.h
+++ b/filament/backend/src/opengl/OpenGLProgram.h
@@ -43,8 +43,8 @@ namespace filament::backend {
 class OpenGLDriver;
 
 struct PushConstantBundle {
-    utils::Slice<std::pair<GLint, ConstantType>> vertexConstants;
-    utils::Slice<std::pair<GLint, ConstantType>> fragmentConstants;
+    utils::Slice<const std::pair<GLint, ConstantType>> vertexConstants;
+    utils::Slice<const std::pair<GLint, ConstantType>> fragmentConstants;
 };
 
 class OpenGLProgram : public HwProgram {
@@ -94,8 +94,8 @@ public:
     PushConstantBundle getPushConstants() {
         auto fragBegin = mPushConstants.begin() + mPushConstantFragmentStageOffset;
         return {
-            .vertexConstants = utils::Slice(mPushConstants.begin(), fragBegin),
-            .fragmentConstants = utils::Slice(fragBegin, mPushConstants.end()),
+                .vertexConstants = { mPushConstants.begin(), fragBegin },
+                .fragmentConstants = { fragBegin, mPushConstants.end() },
         };
     }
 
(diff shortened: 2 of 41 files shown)


pixelflinger:
I don't understand why we need Slice and MutSlice, conceptually Slice is just like a std::string_view and we don't need a mut_string_view or a const_string_view.
The const_cast<> was there only to simplify the implementation of some functions, it didn't have to use const_cast<>.

elizagamedev:
> I don't understand why we need Slice and MutSlice, conceptually Slice is just like a std::string_view and we don't need a mut_string_view or a const_string_view. The const_cast<> was there only to simplify the implementation of some functions, it didn't have to use const_cast<>.

You can't write to the memory pointed at by an `std::string_view`. Likewise, with this change, you can't write to the memory pointed at by `Slice`, but you can for a `MutSlice`.

For example, take a look at `fxaa.cpp`. Without this change, you could compile the following code:

```c++
utils::Slice<StaticMaterialInfo> foo = filament::getFxaaMaterialList();
foo[0] = bar; // segfault
```

But with this change this becomes a compiler error.

pixelflinger:
> > I don't understand why we need Slice and MutSlice, conceptually Slice is just like a std::string_view and we don't need a mut_string_view or a const_string_view. The const_cast<> was there only to simplify the implementation of some functions, it didn't have to use const_cast<>.
> 
> You can't write to the memory pointed at by an `std::string_view`. Likewise, with this change, you can't write to the memory pointed at by `Slice`, but you can for a `MutSlice`.
> 
> For example, take a look at `fxaa.cpp`. Without this change, you could compile the following code:
> 
> ```c++
> utils::Slice<StaticMaterialInfo> foo = filament::getFxaaMaterialList();
> foo[0] = bar; // segfault
> ```
> 
> But with this change this becomes a compiler error.

How about this...

Can we make Slice<> be the mutable one, and create a ConstSlice<> that's the non mutable one, and go change the APIs where it makes sense to?

I think it's better because I believe Slice<> is a public API, so I would prefer if it didn't change.

Also we need unit tests for this.


But I don't understand why we couldn't use `const Slice<>` to mean we want an immutable one? Why is it better to create a whole separate object?

elizagamedev:
> Can we make Slice<> be the mutable one, and create a ConstSlice<> that's the non mutable one, and go change the APIs where it makes sense to?

We could, but technically this would still be a breaking change, since we would forbid creating a Slice from mutable pointers. If we wanted perfect backwards compatibility we'd have to do it like i described in the OP

> But I don't understand why we couldn't use const Slice<> to mean we want an immutable one? Why is it better to create a whole separate object?

It's like the difference between `const char*` and `char* const`.

```c++
const Slice foo = bar;
Slice baz = foo;

//might as well be
char* const foo = bar;
char* baz = foo;
```

elizagamedev:
>We could, but technically this would still be a breaking change, since we would forbid creating a Slice from mutable pointers. If we wanted perfect backwards compatibility we'd have to do it like i described in the OP

Actually I mean the opposite: we would forbid creating a Slice pointing to constant data, as in this universe Slice refers to mutable data whereas ConstSlice refers to immutable data

> Also we need unit tests for this.

👍

elizagamedev:
I pushed another change which renames `Slice` to `ConstSlice` and reintroduces the original `Slice` for compatibility. However, all internal Filament code now uses either `ConstSlice` or `MutSlice`.

I realize that I also messed up a few of the usages of `const` on `BaseSlice`: they should actually almost all be `const` unless they literally affect the `mBegin` and `mEnd` pointers for reasons I explained both in comments in this PR and also in some new comments I introduced in `Slice.h`.

pixelflinger (changes requested):
Thanks for this change and the comments.  I'd like to rename `MutSlice` to `MutableSlice` because filament's style guide normally bans abbreviations, and "mut" is not a style we've used before.

Also let's mark Slice<> deprecated
`@deprecated Use ConstSlice<> or MutableSlice<> instead`
We've done stuff like that before to the public API

And let's byte the bullet with `getFeatureFlags` and return a ConstSlice. This is technically a breaking change, but hopefully it should be easy to fix.

elizagamedev:
As discussed offline, I realized that there was no need for two types after all. Just a single Slice but with fixed memory semantics will do the trick.

There's a lot of really redundant and messy commits in this PR so I'm going to squash/force push them, but I pushed the commit independently for the sake of history.

pixelflinger (approved):

PR #9278, opened 2025-09-30
new utility AsyncJobQueue

pixelflinger:
this is a very simple job queue, it spawns a thread and runs the jobs
pushed to the queue in sequence.

### DIFF
--- a/libs/utils/src/AsyncJobQueue.cpp
+++ b/libs/utils/src/AsyncJobQueue.cpp
@@ -1,0 +1,75 @@
+/*
+ * Copyright (C) 2025 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <utils/AsyncJobQueue.h>
+#include <utils/JobSystem.h>
+#include <utils/debug.h>
+
+#include <mutex>
+#include <utility>
+
+namespace utils {
+
+AsyncJobQueue::AsyncJobQueue(const char* name, Priority priority) {
+    mQueue.reserve(2);
+    mThread = std::thread([this, name, priority]() {
+        JobSystem::setThreadName(name);
+        JobSystem::setThreadPriority(priority);
+        auto& queue = mQueue;
+        bool exitRequested;
+        do {
+            std::unique_lock lock(mLock);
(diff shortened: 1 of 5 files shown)


libs/utils/src/AsyncJobQueue.cpp
```diff
@@ -0,0 +1,75 @@
...
+        bool exitRequested;
+        do {
+            std::unique_lock lock(mLock);
+            mCondition.wait(lock, [this, &queue]() -> bool {
+                return mExitRequested || !queue.empty();
+            });
+            exitRequested = mExitRequested;
+            if (!queue.empty()) {
+                Job const job(std::move(queue.front()));
+                queue.erase(queue.begin());
```
z3moon, line 39: this will shift all elements in the container using copy or move. maybe std::queue is more suitable?

libs/utils/src/AsyncJobQueue.cpp
```diff
@@ -0,0 +1,75 @@
...
+
+AsyncJobQueue::~AsyncJobQueue() noexcept {
+    assert_invariant(mQueue.empty());
+}
+
+void AsyncJobQueue::push(Job&& job) {
+    std::unique_lock const lock(mLock);
+    if (!mExitRequested) {
+        mQueue.push_back(std::move(job));
+        mCondition.notify_one();
```
z3moon, line 57: nit: putting this outside the lock scope is a bit more optimal

libs/utils/src/AsyncJobQueue.cpp
```diff
@@ -0,0 +1,75 @@
...
+AsyncJobQueue::AsyncJobQueue(const char* name, Priority priority) {
+    mQueue.reserve(2);
+    mThread = std::thread([this, name, priority]() {
+        JobSystem::setThreadName(name);
+        JobSystem::setThreadPriority(priority);
+        auto& queue = mQueue;
+        bool exitRequested;
+        do {
+            std::unique_lock lock(mLock);
+            mCondition.wait(lock, [this, &queue]() -> bool {
```
z3moon: maybe you don't need to capture queue as it's mQueue?

libs/utils/src/AsyncJobQueue.cpp
```diff
@@ -0,0 +1,75 @@
...
+    std::unique_lock const lock(mLock);
+    if (!mExitRequested) {
+        mQueue.push_back(std::move(job));
+        mCondition.notify_one();
+    }
+}
+
+void AsyncJobQueue::drainAndExit() {
+    std::unique_lock lock(mLock);
+    mCondition.wait(lock, [this] { return mQueue.empty(); });
```
z3moon, line 63: I think the wait here is risky because it depends on spurious wakeups.

libs/utils/src/AsyncJobQueue.cpp
pixelflinger: yes, but I expect very few elements, std::queue is a lot more costly than vector.

libs/utils/src/AsyncJobQueue.cpp
pixelflinger: I don't think so because because `wait` actually handles that (it retries if there is a spurious wakeup).

z3moon (approved):

PR #9279, opened 2025-09-30
Introduce utils::Status

trie94:
Introduce utils::Status

FIXES=436235264


libs/utils/include/utils/Status.h
```diff
@@ -0,0 +1,109 @@
...
+                break;
+        }
+        os << ", error message: " << status.mErrorMessage;
+        return os;
+    }
+
+private:
+    StatusCode mStatusCode;
+    // Reason for the error if exists.
+    std::string mErrorMessage;
```
pixelflinger: `std::string` is banned from filament. Use `utils::CString`.

libs/utils/include/utils/Status.h
```diff
@@ -0,0 +1,109 @@
...
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+
+#ifndef TNT_UTILS_STATUS_H
+#define TNT_UTILS_STATUS_H
+
+#include <ostream>
+#include <string>
```
pixelflinger: <string> is banned use utils/CString.h

libs/utils/include/utils/Status.h
```diff
@@ -0,0 +1,109 @@
...
+
+    static Status internal(std::string_view message) {
+        return {StatusCode::INTERNAL, message};
+    }
+
+    static Status invalidArgument(std::string_view message) {
+        return {StatusCode::INVALID_ARGUMENT, message};
+    }
+
+    friend std::ostream& operator<<(std::ostream& os, const Status& status) {
```
pixelflinger: I would like this to be in a .cpp file so it's not inlined.

libs/utils/include/utils/Status.h
```diff
@@ -0,0 +1,109 @@
...
+    INVALID_ARGUMENT,
+    INTERNAL,
+};
+
+// Returns the StatusCode to indicate whether the request was successful.
+// If successful, it returns OK with no error message, if not it returns
+// other codes with an optional error message.
+class UTILS_PUBLIC Status {
+public:
+    // Default to StatusCode::OK
```
pixelflinger: add doxygen style comments to all public methods (except copy ctors / operator= / trivial ones)

pixelflinger (changes requested):
just a few small-ish comments. looks really good. thanks.

libs/utils/include/utils/Status.h
```diff
@@ -0,0 +1,109 @@
...
+    StatusCode getCode() const {
+        return mStatusCode;
+    }
+
+    std::string_view getErrorMessage() const {
+        return mErrorMessage;
+    }
+
+    // Convenient static functions that help construct Status.
+    // E.g., Usage `return utils::Status::internal("internal error");`
```
poweifeng: comment doesn't seem to match

libs/utils/include/utils/Status.h
```diff
@@ -0,0 +1,109 @@
...
+#include <string>
+#include <string_view>
+#include <utils/compiler.h>
+
+namespace utils {
+
+enum class StatusCode {
+    OK,
+    INVALID_ARGUMENT,
+    INTERNAL,
```
poweifeng, line 36: Can you add a bit of comment on this?

libs/utils/test/test_Status.cpp
```diff
@@ -0,0 +1,127 @@
...
+
+TEST(StatusTest, DefaultConstructorOkStatus) {
+    Status actual = Status();
+    Status expected = Status(StatusCode::OK, "");
+    EXPECT_EQ(actual, expected);
+}
+
+TEST(StatusTest, Constructor) {
+    std::string_view errorMessage = "invalid";
+    Status status = Status(StatusCode::INVALID_ARGUMENT, errorMessage);
```
poweifeng: just curious, why do you not write it like
```
Status stats(StatusCode::INVALID_ARGUMENT, errorMessage); ? 
```

libs/utils/include/utils/Status.h
trie94: done

libs/utils/include/utils/Status.h
trie94: oh i thought std::string was only banned in the core part. but on a second thought, this can be used in core libs so makes sense to use CString.

i made some changes but kept std::string_view. does this change look good?

libs/utils/include/utils/Status.h
trie94: ditto

libs/utils/include/utils/Status.h
trie94: you mean only this one `friend std::ostream& operator<<(std::ostream& os, const Status& status)`, right? done.

i also move `getErrorMessage` cuz it has a bit more logic then simply returning the error message..

libs/utils/include/utils/Status.h
trie94: oh i think you reviewed the outdated patch

libs/utils/test/test_Status.cpp
trie94: no particular reason.. updated them as you suggested

libs/utils/include/utils/Status.h
trie94: this is updated in the last commit

PR #9281, opened 2025-10-01
fix many exceptions qualifiers

pixelflinger:


z3moon (approved):
maybe you missed updating `PlatformCocoaTouchGL.h`

PR #9283, opened 2025-10-01
engine: change spec constants to simple list

elizagamedev:
For a shader program cache to be keyed on the set of spec constants that a program has set, we need to know full exact list of constants, including their default values. If we're going to always hold a list of all constants all the time, then we may as well store them as a simple list where each index is the ID number of the spec constant.

As part of this change, we now write the default values of spec constants into the material file metadata.

### DIFF
--- a/filament/backend/src/opengl/OpenGLDriver.cpp
+++ b/filament/backend/src/opengl/OpenGLDriver.cpp
@@ -2617,6 +2617,8 @@ bool OpenGLDriver::isWorkaroundNeeded(Workaround const workaround) {
             return mContext.bugs.powervr_shader_workarounds;
         case Workaround::DISABLE_DEPTH_PRECACHE_FOR_DEFAULT_MATERIAL:
             return mContext.bugs.disable_depth_precache_for_default_material;
+        case Workaround::EMULATE_SRGB_SWAPCHAIN:
+            return mContext.isES2() && !mPlatform.isSRGBSwapChainSupported();
         default:
             return false;
     }
--- a/filament/backend/src/opengl/OpenGLDriver.cpp
+++ b/filament/backend/src/opengl/OpenGLDriver.cpp
@@ -764,29 +764,6 @@ void OpenGLDriver::createRenderPrimitiveR(Handle<HwRenderPrimitive> rph,
 void OpenGLDriver::createProgramR(Handle<HwProgram> ph, Program&& program, CString&& tag) {
     DEBUG_MARKER()
 
-
-    if (UTILS_UNLIKELY(mContext.isES2())) {
-        // Here we patch the specialization constants to enable or not the rec709 output
-        // color space emulation in this program. Obviously, the backend shouldn't know about
-        // specific spec-constants, so we need to handle failures gracefully. This cannot be
-        // done at Material creation time because only the backend has access to
-        // Platform.isSRGBSwapChainSupported().
-        if (!mPlatform.isSRGBSwapChainSupported()) {
-            auto& specializationConstants = program.getSpecializationConstants();
-            auto const pos = std::find_if(specializationConstants.begin(), specializationConstants.end(),
-                    [](auto&& sc) {
-                        // This constant must match
-                        // ReservedSpecializationConstants::CONFIG_SRGB_SWAPCHAIN_EMULATION
-                        // which we can't use here because it's defined in EngineEnums.h.
-                        // (we're breaking layering here, but it's for the good cause).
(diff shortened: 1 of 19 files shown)


filament/backend/src/opengl/OpenGLDriver.cpp
```diff
@@ -764,29 +764,6 @@ void OpenGLDriver::createRenderPrimitiveR(Handle<HwRenderPrimitive> rph,
...
-                        // which we can't use here because it's defined in EngineEnums.h.
-                        // (we're breaking layering here, but it's for the good cause).
-                        return sc.id == 3;
-                    });
-            if (pos != specializationConstants.end()) {
-                pos->value = true;
-            }
-        }
-    }
-
```
pixelflinger, line 789: where did that go?

filament/backend/src/vulkan/utils/Spirv.cpp
```diff
@@ -42,25 +42,15 @@ void getTransformedConstantInst(SpecConstantValue* value, uint32_t* inst) {
...
-    if (!value) {
-        // If a specialization wasn't provided, just switch the opcode.
-        uint32_t const op = inst[OP] & 0x0000FFFF;
-        if (op == spv::Op::OpSpecConstantFalse) {
-            inst[OP] = (3 << 16) | spv::Op::OpConstantFalse;
-        } else if (op == spv::Op::OpSpecConstantTrue) {
-            inst[OP] = (3 << 16) | spv::Op::OpConstantTrue;
-        } else if (op == spv::Op::OpSpecConstant) {
-            inst[OP] = (4 << 16) | spv::Op::OpConstant;
-        }
```
pixelflinger, line 54: where did that go?

filament/src/details/Material.h
```diff
@@ -291,8 +291,8 @@ class FMaterial : public Material {
     backend::Program getProgramWithVariants(Variant variant,
             Variant vertexVariant, Variant fragmentVariant) const;
 
-    void processSpecializationConstants(FEngine& engine, Builder const& builder);
-    void processPushConstants(FEngine& engine);
+    utils::FixedCapacityVector<backend::Program::SpecializationConstant>
+    processSpecializationConstants(Builder const& builder);
```
pixelflinger: indent?

poweifeng (approved):
this lgtm, but please wait for Mathias' feedback.

filament/backend/src/opengl/OpenGLDriver.cpp
elizagamedev: This constant now gets set in MaterialDefinition.cpp, line 366

filament/backend/src/vulkan/utils/Spirv.cpp
elizagamedev: It's dead code now, since all spec constants are set all the time, so I removed it. I can add it back if you'd like.

poweifeng: no need to add it back. I was worried that the defaults in the shader would get overwritten, but I see that you packed it into the compiled material. So this is safe to remove

elizagamedev:
I'm running this through our downstream tests just in case before I submit.

filament/backend/src/opengl/OpenGLDriver.cpp
pixelflinger: ah but this will actually break because:

```
bool OpenGLDriver::isSRGBSwapChainSupported() {
    if (UTILS_UNLIKELY(mContext.isES2())) {
        // On ES2 backend (i.e. feature level 0), we always pretend to the client that sRGB
        // SwapChain are available. If we actually have that feature, it'll be used, otherwise
        // we emulate it in the shaders.
        return true;
    }
    return mPlatform.isSRGBSwapChainSupported();
}
```

We need to query the platform directly not the driver.

This is because we want clients to always use the srgb supported code path (otherwise some might actually fail because they don't have the option to go to the postprocessing). but if it's actually not supported, we need to enable the emulation, and only in that case.

pixelflinger (changes requested):

filament/backend/src/opengl/OpenGLDriver.cpp
elizagamedev: Oh, I see. I pushed another commit which I think solves this.

elizagamedev:
> I'm running this through our downstream tests just in case before I submit.

All tests passed :)

pixelflinger (approved):

PR #9284, opened 2025-10-01
utils: RefCountedInternPool/RefCountedMap

elizagamedev:
First, introduce RefCountedInternPool, a reference counted intern pool of Slice<const T>. Just acquire() a slice that you want and you're guaranteed to get exactly one canonical value-equal Slice<const T> back.

Additionally, introduce the concept of NullValue to RefCountedMap. A NullValue defines what should be considered an uninitialized value; by default, it's the default value of that type (0 for ints, nullptr for pointers, etc). This allows us to lazily-initialize values in the map. A client can acquire() a bunch of different resources which will be initialized only when get(factory) is called. If a client attempts to get() a value without specifying a factory, and the value is not initialized (i.e. equal to NullValue{}()), RefCountedMap will panic.

### DIFF
--- a/libs/utils/include/utils/RefCountedInternPool.h
+++ b/libs/utils/include/utils/RefCountedInternPool.h
@@ -1,0 +1,151 @@
+/*
+ * Copyright (C) 2025 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#ifndef TNT_UTILS_REFCOUNTEDINTERNPOOL_H
+#define TNT_UTILS_REFCOUNTEDINTERNPOOL_H
+
+#include <utils/Slice.h>
+#include <utils/FixedCapacityVector.h>
+#include <utils/Hash.h>
+#include <utils/Panic.h>
+#include <utils/debug.h>
+
+#include <tsl/robin_map.h>
+
+namespace utils {
+
+/** A reference-counted intern pool of slices of T. */
+template<typename T, typename Hash = std::hash<T>>
+class RefCountedInternPool {
+    struct HashSlice {
+        inline size_t operator()(Slice<const T> const& slice) const noexcept {
+            return slice.template hash<Hash>();
+        }
+    };
+
+    struct Entry {
+        uint32_t referenceCount;
+        FixedCapacityVector<T> value;
+    };
+
+    using Map = tsl::robin_map<Slice<const T>, Entry, HashSlice>;
+
+    static constexpr const char* UTILS_NONNULL MISSING_ENTRY_ERROR_STRING =
+            "RefCountedInternPool is missing entry";
+
+public:
+    RefCountedInternPool() = default;
+    RefCountedInternPool(RefCountedInternPool const& rhs) = delete;
+    RefCountedInternPool& operator=(RefCountedInternPool const& rhs) = delete;
+    RefCountedInternPool(RefCountedInternPool&& rhs) = default;
+    RefCountedInternPool& operator=(RefCountedInternPool&& rhs) = default;
+
+    /** Acquire an interned copy of value. */
+    Slice<const T> acquire(Slice<const T> slice, size_t hash) noexcept {
+        if (slice.empty()) {
+            return { nullptr, nullptr };
+        }
+        auto it = mMap.find(slice, hash);
+        if (it != mMap.end()) {
+            it.value().referenceCount++;
+            return it.key();
+        }
+        FixedCapacityVector<T> value(slice);
+        // TODO: how to use above computed hash here?
+        return mMap.insert({ value.as_slice(), Entry{
+                .referenceCount = 1,
+                .value = std::move(value),
+            }}).first.key();
+    }
+
+    inline Slice<const T> acquire(Slice<const T> slice) noexcept {
+        return acquire(slice, HashSlice{}(slice));
+    }
+
+    Slice<const T> acquire(FixedCapacityVector<T>&& value, size_t hash) noexcept {
+        if (value.empty()) {
+            return { nullptr, nullptr };
+        }
+        Slice<const T> slice = value.as_slice();
+        auto it = mMap.find(slice, hash);
+        if (it != mMap.end()) {
+            it.value().referenceCount++;
+            return it.key();
+        }
+        // TODO: how to use above computed hash here?
+        return mMap.insert({ slice, Entry{
+                .referenceCount = 1,
+                .value = std::move(value),
+            }}).first.key();
+    }
+
+    inline Slice<const T> acquire(FixedCapacityVector<T>&& value) noexcept {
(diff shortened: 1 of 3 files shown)


elizagamedev:
I'm planning on adding more unit tests for this bug I wanted to get the PR made early.

libs/utils/include/utils/RefCountedMap.h
```diff
@@ -103,40 +111,46 @@ class RefCountedMap {
...
-                return &mMap.insert({key, Entry{ 1, std::move(r.value()) }}).first.value().value;
-            }
+        T r = factory();
+        if (r == NullValue{}()) {
             return nullptr;
         }
+        // TODO: how to use above computed hash here?
+        return &deref(mMap.insert({key, Entry{
+                        .referenceCount = 1,
+                        .value = std::move(r),
```
z3moon: same here (including below)

elizagamedev:
> I'm planning on adding more unit tests for this bug I wanted to get the PR made early.

Okay, tests added.

libs/utils/include/utils/RefCountedInternPool.h
```diff
@@ -0,0 +1,151 @@
...
+        auto it = mMap.find(slice, hash);
+        if (it != mMap.end()) {
+            it.value().referenceCount++;
+            return it.key();
+        }
+        FixedCapacityVector<T> value(slice);
+        // TODO: how to use above computed hash here?
+        return mMap.insert({ value.as_slice(), Entry{
+                .referenceCount = 1,
+                .value = std::move(value),
```
z3moon: this is a c++20 feature, which we shouldn't use yet.

libs/utils/include/utils/RefCountedInternPool.h
```diff
@@ -0,0 +1,151 @@
...
+        Slice<const T> slice = value.as_slice();
+        auto it = mMap.find(slice, hash);
+        if (it != mMap.end()) {
+            it.value().referenceCount++;
+            return it.key();
+        }
+        // TODO: how to use above computed hash here?
+        return mMap.insert({ slice, Entry{
+                .referenceCount = 1,
+                .value = std::move(value),
```
z3moon: ditto

libs/utils/include/utils/RefCountedInternPool.h
```diff
@@ -0,0 +1,151 @@
...
+            }}).first.key();
+    }
+
+    inline Slice<const T> acquire(FixedCapacityVector<T>&& value) noexcept {
+        return acquire(std::move(value), HashSlice{}(value.as_slice()));
```
z3moon: although this code works now since the parameters are evaluated from right to left, it has a potential risk of generating dangling pointers in `value.as_slice` if the order is ever changed.

libs/utils/include/utils/InternPool.h
```diff
@@ -0,0 +1,151 @@
...
+        return acquire(slice, HashSlice{}(slice));
+    }
+
+    /** Release interned value. */
+    void release(Slice<const T> slice, size_t hash) noexcept {
+        if (slice.empty()) {
+            return;
+        }
+        auto it = mMap.find(slice, hash);
+        FILAMENT_CHECK_PRECONDITION(it != mMap.end()) << MISSING_ENTRY_ERROR_STRING;
```
z3moon, line 108: so the caller shouldn't misuse this? or is it allowed to print a warning and proceed?

libs/utils/include/utils/RefCountedMap.h
```diff
@@ -186,6 +225,8 @@ class RefCountedMap {
     TValue& get(KeyRef key, size_t hash) noexcept {
         auto it = mMap.find(key);
         FILAMENT_CHECK_PRECONDITION(it != mMap.end()) << MISSING_ENTRY_ERROR_STRING;
+        FILAMENT_CHECK_PRECONDITION(it.value().value != NullValue{}())
+                << MISSING_VALUE_ERROR_STRING;
```
z3moon, line 223: this as well, I'm not sure if we should panic upon failure. is it critical to misuse this?

z3moon (changes requested):

show50726:
Just out of curious, why make all functions inlined?

z3moon:
> Just out of curious, why make all functions inlined?

They're template methods, which are subject to inlining implicitly. Actually explicit `inline` for template methods is unnecessary. Even if explicitly specified, it's up to compiler to decide whether inlining is performed.

show50726:
@z3moon I see. Thanks for your explanation!

libs/utils/include/utils/RefCountedInternPool.h
```diff
@@ -0,0 +1,151 @@
...
+#include <utils/Panic.h>
+#include <utils/debug.h>
+
+#include <tsl/robin_map.h>
+
+namespace utils {
+
+/** A reference-counted intern pool of slices of T. */
+template<typename T, typename Hash = std::hash<T>>
+class RefCountedInternPool {
```
pixelflinger: There is a lot of code in this class and everything is inlined. I would like to see if there is a way to use the inlines for type safety but have the "heavy" methods be in a cpp, acquire() and release() are going to generate tons of code.

Also unless we have multiple uses of this, maybe we keep it inside filament instead of libutils initially? 

I can't immediately think of a better name (naming is hard!) but I don't like "RefCounted" because it's an abbreviation.  Do we need to qualify it with "RefCounted" would "InternPool" be enough?

pixelflinger (changes requested):

libs/utils/include/utils/InternPool.h
elizagamedev: The contract of `get()` is that the value must already exist, so it's critical that we crash here. If we didn't, we'd just be moving the responsibility to the call site.

libs/utils/include/utils/RefCountedInternPool.h
elizagamedev: From my understanding, template functions are always "inlined" according to the language spec, but compilers generally only keep one the function at link time. I doubt this code is literally going to be inlined everywhere in practice. Like @z3moon said there's probably not even a point in including the inline keyword in the functions I did explicitly mark inline.

Moving the implementation to a C++ file is troublesome because we depend on copy constructors, `std::move()`, etc, which would depend on the types.

I could rename this to just `InternPool`, but should I then rename `RefCountedMap` to `ReferenceCountedMap`?

libs/utils/include/utils/RefCountedInternPool.h
elizagamedev: Good catch, fixed.

libs/utils/include/utils/RefCountedInternPool.h
z3moon: Non-template methods, which don't require T, can be in cpp. But all methods here require T.

And template functions with a couple lines will mostly be inlined. But functions with longer body (including for-loop or many branches) are not usually inlined but converted to a function call at link time.

pixelflinger: some ideas for the RefCountedMap name:  SharedMap, SharedObjectMap?

for the inlining, yeah maybe there is no choice -- but in that case, I would make sure to wrap it into a typed class, basically making sure that these big methods don't get inlined more than once. A bit like Bimap.h is used.

libs/utils/include/utils/InternPool.h
pixelflinger: if you're using FILAMENT_CHECK_PRECONDITION the method can't be `noexcept`

libs/utils/include/utils/RefCountedMap.h
pixelflinger: I tihnk it's okay if that's a precondition, but then the method cannot be noexcept.

pixelflinger (changes requested):
just that apis that call FILAMENT_CHECK_*** can't be noexcept.

PR #9285, opened 2025-10-02
Use utils::Status in MaterialParser

trie94:
Replace bool with utils::Status in MaterialParser.
FIXES=448126610

This is actually a breaking change, so it needs one line update in the Impress code.


libs/filament-matp/src/MaterialParser.cpp
```diff
@@ -155,100 +148,95 @@ static bool reflectParameters(const MaterialBuilder& builder) {
...
     if (!value) {
-        std::cerr << "'material' block does not have a value, one is required." << std::endl;
-        return false;
+        return utils::Status::invalidArgument(
+                "'material' block does not have a value, one is required.");
     }
 
     if (value->getType() != JsonishValue::OBJECT) {
-        std::cerr << "'material' block has an invalid type: "
+        std::stringstream errorMessage;
```
pixelflinger: std::stringstream will add tons of bloat.  use utils::ostream instead.

pixelflinger (changes requested):

trie94:
as mentioned in the pr description, this needs impress code changes in the next filament update. is that ok?

poweifeng:
> as mentioned in the pr description, this needs impress code changes in the next filament update. is that ok?

yeah, that's fine

libs/filament-matp/src/MaterialParser.cpp
```diff
@@ -500,12 +504,13 @@ bool MaterialParser::parse(filamat::MaterialBuilder& builder,
         builder.shaderDefine(define.first.c_str(), define.second.c_str());
     }
 
-    if (!processMaterialParameters(builder, config)) {
+    if (utils::Status processedStatus = processMaterialParameters(builder, config);
+            !processedStatus.isOk()) {
         std::cerr << "Error while processing material parameters." << std::endl;
```
poweifeng: is this still necessary?

libs/filament-matp/src/MaterialParser.cpp
poweifeng: if there are no more std::cerr, maybe we can remove iostream?

poweifeng: Also if possible, following the include order described [here](https://github.com/google/filament/blob/main/CODE_STYLE.md#headers)

trie94: i kept it because i left `reflectParameters` to use std::cout.

i changed it so that it's not returning OK status with the error message that contains the param string, and updated the MaterialCompiler:
```
    if (config.getReflectionTarget() != matp::Config::Metadata::NONE) {
        // When reflecting parameters, the string is in the error message.
        std::cout << status.getErrorMessage() << std::endl;
        return true;
    }
```
but this is kind of weird as it's... now using error message for general purposes.

i can think of three options:
1) move `reflectParameters` in `MaterialCompiler`, and use cout there.
2) rename `mErrorMessage` to `mMessage` in `utils::Status`, and use the message as general purpose
3) keep the current version.

any preference?

-
also update the header order!

libs/filament-matp/src/MaterialParser.cpp
trie94: i was debating whether to remove or leave those cerr logs, but i guess it makes sense to remove them as they will be propagated.

libs/filament-matp/src/MaterialParser.cpp
pixelflinger: I like both 1 and 2 together :-)

Like we should do 1, maybe it's cleaner?  And because it's called "Status", it's not really an "Error", so let's make it just a message. wdyt?

trie94: for now i just did 2, which is just making the error message more generic.

i think we could do 1 if necessary... i agree that it might be cleaner to have that in MaterialCompiler but maybe it's useful to have it in MaterialParser since we might use it for debugging? we can revisit this later?

trie94:
also i think i was getting some release note check failure. is this necessary for this change? otherwise can either of you help me mark this as internal?

PR #9286, opened 2025-10-02
Fix public headers check on github

z3moon:


z3moon:
> Actually we can't do this.
> 
> I think we (and myself included) have been handling DIST_HDRS wrong. These headers are supposed to be ONLY the headers we want to export to the public, e.g. for use by our clients for their own code for instance.
> 
> We certainly don't want to export JobSystem for instance.
> 
> The correct fix is actually to remove AsyncJobQueue.h from the list, I added it by mistake.

oh I see. let me update this.

PR #9287, opened 2025-10-02
optional per-channel depth clear

pixelflinger:
- increase the number of channels from 4 to 8
- new api on View to enable per-channel depth clear
FIXES=[447648764]


z3moon (approved):

filament/src/details/Renderer.cpp
```diff
@@ -1104,23 +1105,36 @@ void FRenderer::renderJob(RootArenaScope& rootArenaScope, FView& view) {
...
+    });
+
     // color-grading as subpass is done either by the color pass or the TAA pass if any
     auto colorGradingConfigForColor = colorGradingConfig;
     colorGradingConfigForColor.asSubpass = colorGradingConfigForColor.asSubpass && !taaOptions.enabled;
 
     if (colorGradingConfigForColor.asSubpass) {
         // append color grading subpass after all other passes
-        passBuilder.customCommand(3,
+        passBuilder.customCommand(7,
```
poweifeng: could probably have a constant, like `LAST_CHANNEL` for all of the 7s.

filament/src/materials/clearDepth.mat
```diff
@@ -0,0 +1,8 @@
+material {
+    name : clearDepth,
+    culling: none,
+    colorWrite : false,
+    depthWrite : true,
+    depthCulling : true,
```
poweifeng: should this be true? I guess it doesn't matter?

filament/src/PostProcessManager.cpp
```diff
@@ -2602,6 +2602,42 @@ FrameGraphId<FrameGraphTexture> PostProcessManager::customResolveUncompressPass(
...
 }
 
+
+void PostProcessManager::clearAncillaryBuffersPrepare(DriverApi& driver) noexcept {
+    auto const& material = getPostProcessMaterial("clearDepth");
+    auto ma = material.getMaterial(mEngine, PostProcessVariant::OPAQUE);
+    auto [mi, fixedIndex] = mMaterialInstanceManager.getFixedMaterialInstance(ma);
+    mFixedMaterialInstanceIndex.clearDepth = fixedIndex;
+    mi->commit(driver);
+    material.getMaterial(mEngine);
```
poweifeng, line 2612: what's this for?

filament/src/PostProcessManager.cpp
```diff
@@ -2602,6 +2602,42 @@ FrameGraphId<FrameGraphTexture> PostProcessManager::customResolveUncompressPass(
...
+}
+
+void PostProcessManager::clearAncillaryBuffers(DriverApi& driver,
+        TargetBufferFlags attachments) const noexcept {
+    attachments &= TargetBufferFlags::DEPTH;
+    if (none(attachments & TargetBufferFlags::DEPTH)) {
```
poweifeng, line 2619: I dont think you need the line above if you're doing the `&` here?

filament/src/RenderPass.h
poweifeng: nit: `CCC`

filament/src/details/Renderer.cpp
pixelflinger: yeah probably a good idea :-)

filament/src/materials/clearDepth.mat
pixelflinger: yeah, maybe I should try without.

filament/src/PostProcessManager.cpp
pixelflinger: I think it's for guaranteeing the material is loaded at that point

filament/src/PostProcessManager.cpp
pixelflinger: ah yes. I forgot to document that. the line above should really be DEPTH_AND_STENCIL one day, but for now, I wanted to just handle depth.

filament/src/details/Renderer.cpp
pixelflinger: done

poweifeng (approved):

filament/src/materials/clearDepth.mat
pixelflinger: it worked without.

PR #9288, opened 2025-10-02
Minor changes in utils::Status

trie94:
- << operator doesn't have to be friend
- simplify getErrorMessage to not use strlen internally

### DIFF
--- a/libs/utils/include/utils/Status.h
+++ b/libs/utils/include/utils/Status.h
@@ -136,7 +136,7 @@ private:
     utils::CString mErrorMessage;
 };
 
-std::ostream& operator<<(std::ostream& os, const Status& status);
+utils::io::ostream& operator<<(utils::io::ostream& os, const Status& status);
 
 } // namespace utils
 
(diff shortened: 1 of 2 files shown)


libs/utils/include/utils/Status.h
```diff
@@ -128,14 +130,14 @@ class UTILS_PUBLIC Status {
...
 
-    friend std::ostream& operator<<(std::ostream& os, const Status& status);
-
 private:
     StatusCode mStatusCode;
     // Reason for the error if exists.
     utils::CString mErrorMessage;
 };
 
+std::ostream& operator<<(std::ostream& os, const Status& status);
```
poweifeng: nit: this is used in Impress' codebase?  It's much more common for filament components to override
```
utils::io::ostream& operator<<(utils::io::ostream&, const Status& status);
```

poweifeng (approved):

z3moon (approved):

libs/utils/include/utils/Status.h
pixelflinger: good catch!

trie94: oh i see. updated

PR #9289, opened 2025-10-02
internal: fix some editor indentation rules

elizagamedev:


.clang-format
```diff
@@ -15,7 +15,7 @@ AllowShortLoopsOnASingleLine: true
 AlwaysBreakAfterReturnType: None
 AlwaysBreakTemplateDeclarations: Yes
 BreakBeforeBraces: Custom
-BracedInitializerIndentWidth: 4
+BracedInitializerIndentWidth: 8
```
poweifeng: I actually prefer 4 here.  It's been working ok so far I think.

elizagamedev: I prefer 4 too, but most of the existing code that I've seen uses 8 for some reason. I think maybe because Mathias uses CLion?

poweifeng: yeah, I think it's one of those minor differences that no one complains about. So IMO it's ok to keep it at 4.

pixelflinger: the rule is that our "continuation indent" is 8.

elizagamedev: Does a braced initializer list count as a continuation indent?

pixelflinger: can I see an example?

poweifeng:
```
            VulkanAttachment src {
                .level = uint8_t(level),
                .layer = layer,
            };
```

vs. 
```
            VulkanAttachment src {
                    .level = uint8_t(level),
                    .layer = layer,
            };
```

pixelflinger: oh yeah, okay, I've considered these to be continuation indent. I'm okay either way. If we change the rule, let's document it and we'll update the files over time (or if someone wants to do a pass at updating all of them).

elizagamedev: In that case I switched the rules to make these 4 instead of 8 and pushed changes to CODE_STYLE.md.

PR #9291, opened 2025-10-03
New ImmutableCString string class

pixelflinger:
ImmutableCString is a string class similar to CString except it's immutable. ImmutableCString occupies 16 bytes instead of 8 for CString.

However, ImmutableCString is able to avoid memory allocation when constructed from a string literal, and in that way it us similar to StaticString.

ImmutableCString can be auto converted from StaticString.

The backend tag tracking is updated to use ImmutableCString and the FrameGraph resource manager us updated to use StaticString.

Together these changes significantly cut down heap allocations due to internal tagging.

We also add optional tracking to {Immutable}CString.

### DIFF
--- a/libs/utils/include/utils/ImmutableCString.h
+++ b/libs/utils/include/utils/ImmutableCString.h
@@ -1,0 +1,199 @@
+/*
+ * Copyright (C) 2025 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef TNT_UTILS_IMMUTABLECSTRING_H
+#define TNT_UTILS_IMMUTABLECSTRING_H
+
+#include <utils/compiler.h>
+
+#include <utils/StaticString.h>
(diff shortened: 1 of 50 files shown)


pixelflinger:
PSA: The best way to review this large PR is to look at libutils first, then the framegraph changes, then the backend changes.

libs/utils/include/utils/ImmutableCString.h
```diff
@@ -0,0 +1,191 @@
...
+    using const_reference = const value_type&;
+    using const_pointer   = const value_type*;
+    using const_iterator  = const value_type*;
+
+    ImmutableCString() noexcept {
+        track(true, mIsStatic);
+    }
+
+    template<size_t N>
+    ImmutableCString(const char (&str)[N]) noexcept : mData(str), mSize(N - 1) { // NOLINT(*-explicit-constructor)
```
z3moon, line 56: nit: this won't work as intended with a fixed size array like `const char arr[16] = "hello"`

but also this usage is obsolete so maybe safe to ignore.

libs/utils/include/utils/StaticString.h
```diff
@@ -41,7 +45,7 @@ class UTILS_PUBLIC StaticString {
     template<unsigned int M>
     constexpr StaticString(const char (&str)[M]) noexcept : mString(str, M - 1) {} // NOLINT(*-explicit-constructor)
 
-    constexpr StaticString() noexcept = default;
+    constexpr StaticString() noexcept : mString("", 0) {}
```
z3moon, line 48: maybe just `mString()`?

pixelflinger: I wanted to be able to do c_str() without having to worry about nullptr.

libs/utils/include/utils/ImmutableCString.h
pixelflinger: yeah that's the same behavior as CString. But that's a good point, we want to be able to be in static mode even in that case, I'll see if I can make this better.

poweifeng (approved):

PR #9293, opened 2025-10-03
Improvements to FrameInfo

pixelflinger:
- return the GPU Complete timestamp
- return the app VSYNC timestamp
- works TimerQueries are not supported

The VSYNC time is just a convenience as it is the same value provided by the application during Renderer::beginFrame() or via Renderer::setVsyncTime().

### DIFF
--- a/filament/src/FrameInfo.h
+++ b/filament/src/FrameInfo.h
@@ -25,6 +25,7 @@
 
 #include <utils/compiler.h>
 #include <utils/debug.h>
+#include <utils/AsyncJobQueue.h>
 #include <utils/FixedCapacityVector.h>
 
 #include <array>
@@ -56,9 +57,16 @@ struct FrameInfoImpl : public details::FrameInfo {
     time_point endFrame;             // main thread endFrame time
     time_point backendBeginFrame;    // backend thread beginFrame time (makeCurrent time)
     time_point backendEndFrame;      // backend thread endFrame time (present time)
+    time_point gpuFrameComplete;     // the frame is done rendering on the gpu
+    time_point vsync;                // vsync time
+    backend::FenceHandle fence{};    // the fence used for gpuFrameComplete
     std::atomic_bool ready{};        // true once backend thread has populated its data
     explicit FrameInfoImpl(uint32_t const id) noexcept
         : frameId(id) {
+    }
+
+    ~FrameInfoImpl() noexcept {
+        assert_invariant(!fence);
     }
 
     FrameInfoImpl(FrameInfoImpl&& rhs) noexcept : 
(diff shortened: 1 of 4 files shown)


filament/src/FrameInfo.h
```diff
@@ -237,6 +260,8 @@ class FrameInfoManager {
     uint32_t mLast = 0;                 // index of oldest query still active
     FrameInfoImpl* pFront = nullptr;    // the most recent slot with a valid frame time
     FrameHistoryQueue mFrameTimeHistory{};
+    utils::AsyncJobQueue mJobQueue;
+    bool mHasTimerQueries = false;
```
poweifeng: nit: could be `bool const mHasTimerQueries;`

filament/src/FrameInfo.h
```diff
@@ -56,18 +57,27 @@ struct FrameInfoImpl : public details::FrameInfo {
...
+
     FrameInfoImpl(FrameInfoImpl&& rhs) noexcept : 
         details::FrameInfo(rhs),
         frameId(rhs.frameId),
         beginFrame(rhs.beginFrame),
         endFrame(rhs.endFrame),
         backendBeginFrame(rhs.backendBeginFrame),
         backendEndFrame(rhs.backendEndFrame),
+        gpuFrameComplete(rhs.gpuFrameComplete),
+        vsync(rhs.vsync),
```
z3moon, line 80: I think `fence` also needs to be moved?

pixelflinger: done. thanks.

filament/src/FrameInfo.h
pixelflinger: done

libs/utils/src/AsyncJobQueue.cpp
```diff
@@ -51,7 +58,11 @@ AsyncJobQueue::~AsyncJobQueue() noexcept {
 
 void AsyncJobQueue::push(Job&& job) {
     std::unique_lock lock(mLock);
-    if (!mExitRequested) {
+    if (UTILS_UNLIKELY(mExitRequested)) {
+        LOG(WARNING) << "AsyncJobQueue::push() called after drainAndExit()";
+    }
```
z3moon, line 63: nit: or use an early return here and remove the if statement below.

z3moon (approved):

libs/utils/src/AsyncJobQueue.cpp
pixelflinger: I wanted it to always assert in DEBUG mode but still have a log in release (this would be a low frequency log).

z3moon: Oh, I meant you could combine `if (UTILS_UNLIKELY(mExitRequested)) {` and `if (UTILS_LIKELY(!mExitRequested)) {` as one if-else or an early return pattern. You can move the assert to the top.

pixelflinger:
back to draft because this breaks CI due to a WebGPU backend bug. @poweifeng is on it.

poweifeng:
> back to draft because this breaks CI due to a WebGPU backend bug. @poweifeng is on it.

Issue should be resolved.  Re-running `test-renderdiff` to see.

PR #9297, opened 2025-10-06
engine: add program cache

elizagamedev:
This is another chunky change.

The core of this change is to cache programs in MaterialCache according to a "specialization" (ProgramSpecialization) which is defined as the program cache ID (the same key used for the OpenGL binary blob cache), the variant, and the set of spec constants.

As part of this change, a lot of the implementation details of shader compilation were refactored from Material to MaterialDefinition. The resulting flow is a lot cleaner and easier to reason about, since shader compilation is now a pure function of the MaterialDefinition + ProgramSpecialization.

Since the global cache program lookups might take a bit of time to compute hashes, etc, I left the set of cached programs in Material as well, which kind of acts like an L1 cache. The effect is that prepareProgram() and getProgram() should be no slower than HEAD, even with the more complex caching requirements.

I'm planning on writing a document about this (and all changes up until this point), but I'm being asked to work on higher priority things and I wanted to have this PR out for review in the meantime so it doesn't bitrot.


filament/backend/src/Program.cpp
```diff
@@ -32,14 +33,53 @@ namespace filament::backend {
...
+          mMultiview(std::move(rhs.mMultiview)) {
+    rhs.mSpecializationConstantsInternPool = nullptr;
+    rhs.mSpecializationConstants.clear();
 }
 
-Program::Program(Program&& rhs) noexcept = default;
-
-Program& Program::operator=(Program&& rhs) noexcept = default;
+Program& Program::operator=(Program&& rhs) noexcept {
+    mSpecializationConstantsInternPool = rhs.mSpecializationConstantsInternPool;
```
z3moon: I think you also need to do
```
    if (mSpecializationConstantsInternPool) {
        mSpecializationConstantsInternPool->release(mSpecializationConstants);
    }
```
at the top of this method to prevent memory leak for lhs?

z3moon (changes requested):

filament/src/MaterialCache.cpp
```diff
@@ -56,18 +63,61 @@ MaterialDefinition* UTILS_NULLABLE MaterialCache::acquire(FEngine& engine,
...
             });
 }
 
+backend::Handle<backend::HwProgram> MaterialCache::acquireProgram(
+        ProgramSpecialization const& specialization) noexcept {
+    backend::Handle<backend::HwProgram> const* program = mPrograms.acquire(specialization);
+    if (program) {
+        return *program;
+    }
+    return backend::Handle<backend::HwProgram>();
```
show50726: Will the newly created program be added into `mPrograms` at some point?

filament/backend/src/Program.cpp
elizagamedev: Ah, good catch

filament/src/MaterialCache.cpp
elizagamedev: Yep. The flow is:

1. `MaterialDefinition::acquirePrograms()` acquires all programs, but doesn't necessarily compile them.
2. Later, when a specific program is needed, `MaterialDefinition::prepareProgram()` is called, which compiles an already-acquired program
3. Finally, `MaterialDefinition::getProgram()` is called to actually get the program when it's needed

show50726: Oh, I see. Thanks for the explanation!

show50726: BTW the refactoring looks really nice to me!

libs/filabridge/src/Variant.cpp
```diff
@@ -218,8 +214,16 @@ utils::Slice<const Variant> getUnlitVariants() noexcept {
...
+utils::Slice<const Variant> getLitDepthVariants() noexcept {
+    return { details::gLitDepthVariants.data(), details::gLitDepthVariants.size() };
+}
+
+utils::Slice<const Variant> getUnlitDepthVariants() noexcept {
+    return { details::gUnlitDepthVariants.data(), details::gUnlitDepthVariants.size() };
+}
+
+utils::Slice<const Variant> getPostProcessVariants() noexcept {
+    return { details::gPostProcessVariants.data(), details::gPostProcessVariants.size() };
```
pixelflinger: Something is odd here. A depth variant can't be lit or unlit; it's just a depth variant. In fact the DIR, DYN and SRE bits don't have meaning if DEP is on.

filament/backend/include/backend/Program.h
```diff
@@ -61,7 +63,7 @@ class Program {
 
     using DescriptorBindingsInfo = utils::FixedCapacityVector<Descriptor>;
     using DescriptorSetInfo = std::array<DescriptorBindingsInfo, MAX_DESCRIPTOR_SET_COUNT>;
-    using SpecializationConstantsInfo = utils::FixedCapacityVector<SpecializationConstant>;
+    using SpecializationConstantsInfo = utils::Slice<const SpecializationConstant>;
```
pixelflinger: so Program is a public API of the backend. I think it's weird that Program must be given an InternPool.

imho, the api/interface to Program should stay an array of spec constant -- but *internally* in the concrete implementations of each backend (or maybe in DriverBase) it can use the internpool

I'm surprised this change doesn't break a bunch of unit tests.


Program is supposed to be a "bundle" that describes the program (mostly shader + spec constants + binding infos). I don't like that it references data outside of itself. It should be a simple value type imho.

filament/src/details/Material.cpp
```diff
@@ -274,6 +271,46 @@ void FMaterial::compile(CompilerPriorityQueue const priority,
...
 
+backend::Handle<backend::HwProgram> FMaterial::prepareProgramSlow(Variant const variant,
+        backend::CompilerPriorityQueue const priorityQueue) const noexcept {
+    if (isSharedVariant(variant)) {
+        FMaterial const* defaultMaterial = mEngine.getDefaultMaterial();
+        FILAMENT_CHECK_PRECONDITION(defaultMaterial);
+        return mCachedPrograms[variant.key] =
+                defaultMaterial->prepareProgram(variant, priorityQueue);
+    }
+    return mCachedPrograms[variant.key] = mEngine.getMaterialCache().prepareProgram(mEngine,
```
z3moon: program is cached to `mCachedPrograms` but the reference counter is not increased. Looks like it could lead to a dangling pointer for a certain case. Maybe call `acquireAndPrepareProgram` instead?

filament/src/details/Material.cpp
```diff
@@ -274,6 +271,46 @@ void FMaterial::compile(CompilerPriorityQueue const priority,
...
+[[nodiscard]]
+backend::Handle<backend::HwProgram> FMaterial::getProgramSlow(
+        Variant const variant) const noexcept {
+    if (isSharedVariant(variant)) {
+        FMaterial const* defaultMaterial = mEngine.getDefaultMaterial();
+        FILAMENT_CHECK_PRECONDITION(defaultMaterial);
+        return mCachedPrograms[variant.key] = defaultMaterial->getProgram(variant);
+    }
+    return mCachedPrograms[variant.key] =
+                   mEngine.getMaterialCache().getProgram(getProgramSpecialization(variant));
```
z3moon: this may also need to be replaced with `acquireProgram`

filament/backend/src/Program.cpp
```diff
@@ -32,14 +33,57 @@ namespace filament::backend {
...
+          mMultiview(std::move(rhs.mMultiview)) {
+    rhs.mSpecializationConstantsInternPool = nullptr;
+    rhs.mSpecializationConstants.clear();
 }
 
-Program::Program(Program&& rhs) noexcept = default;
-
-Program& Program::operator=(Program&& rhs) noexcept = default;
+Program& Program::operator=(Program&& rhs) noexcept {
+    if (mSpecializationConstantsInternPool) {
```
z3moon: probably

    if (this == &rhs) {
        return *this;
    }

is needed for the case of self-assignment.

libs/filabridge/src/Variant.cpp
elizagamedev: I could have sworn I saw a case in Variant.h where they differed, but I don't see it on further examination. I will revert this.

filament/backend/include/backend/Program.h
elizagamedev: I'll revert this change too.

filament/backend/src/Program.cpp
elizagamedev: Reverting this bit

filament/src/details/Material.cpp
elizagamedev: I implemented this with the same assumption as HEAD, in that any given `Material` will always outlive the default material. I could refactor `MaterialDefinition::acquirePrograms()` to additionally acquire the variants in the default material but it feels kind of messier than to just rely on the cache owned by the default material.

filament/backend/include/backend/Program.h
```diff
@@ -19,7 +19,9 @@
 
 #include <utils/CString.h>
 #include <utils/FixedCapacityVector.h>
+#include <utils/InternPool.h>
```
z3moon: probably this is no longer needed.

filament/src/details/Engine.cpp
```diff
@@ -527,14 +525,11 @@ void FEngine::init() {
...
-                                +ReservedSpecializationConstants::CONFIG_DEBUG_FROXEL_VISUALIZATION,
+                        FMaterial::SpecializationConstantsBuilder constants =
+                                material->getSpecializationConstantsBuilder();
+                        constants.set(+ReservedSpecializationConstants::
+                                              CONFIG_DEBUG_FROXEL_VISUALIZATION,
                                 debug.lighting.debug_froxel_visualization);
-
-                        material->invalidate(
-                                Variant::DYN | Variant::DEP,
-                                Variant::DYN);
```
z3moon, line 564: looks like `material->setSpecializationConstants(std::move(constants));` is missing here.

filament/src/details/Material.cpp
```diff
@@ -596,142 +461,30 @@ void FMaterial::onQueryCallback(void* userdata, VariantList* pActiveVariants) {
...
+    mDefinition.acquirePrograms(mEngine, mCachedPrograms.as_slice(), mSpecializationConstants,
+            mIsDefaultMaterial);
 
-    mSpecializationConstants[id] = value;
-
-    return true;
+    // Then release old ones.
+    mDefinition.releasePrograms(mEngine, mCachedPrograms.as_slice(), prevSpecializationConstants,
+            mIsDefaultMaterial);
+    internPool.release(prevSpecializationConstants);
```
z3moon: It looks like the order should be changed. My concern is that the newly acquired program handles in `mCachedPrograms` is being cleared by releasePrograms.

filament/src/MaterialCache.cpp
```diff
@@ -56,18 +63,61 @@ MaterialDefinition* UTILS_NULLABLE MaterialCache::acquire(FEngine& engine,
...
+    return *program;
+}
+
+backend::Handle<backend::HwProgram> MaterialCache::getProgram(
+        ProgramSpecialization const& specialization) {
+    return mPrograms.get(specialization);
+}
+
+void MaterialCache::releaseProgram(FEngine& engine, ProgramSpecialization const& specialization) {
+    return mPrograms.release(specialization,
```
z3moon: Please remove `return` as it's a void function.

filament/src/details/Material.cpp
elizagamedev: Ah, you're right. The reason they're ordered like this is because I didn't want to potentially free any programs and then recreate them immediately after, but given that this should only ever be called if spec constants change, and that the default material manages the lifetime of its own spec constants, that shouldn't ever be a problem.

elizagamedev:
I actually think there's still some work that needs to be done here w/ matdbg, so please wait before reviewing again

filament/backend/include/backend/Program.h
```diff
@@ -135,11 +136,7 @@ class Program {
...
     CompilerPriorityQueue getPriorityQueue() const noexcept { return mPriorityQueue; }
 
-    SpecializationConstantsInfo const& getSpecializationConstants() const noexcept {
-        return mSpecializationConstants;
-    }
-
-    SpecializationConstantsInfo& getSpecializationConstants() noexcept {
+    SpecializationConstantsInfo getSpecializationConstants() const noexcept {
         return mSpecializationConstants;
     }
```
z3moon, line 147: You can probably revert this as you explicitly copy the array below?

z3moon:
BTW, this change looks really good. And thanks for your patience.

elizagamedev:
I finally managed to finish this PR. matdbg works and I addressed the bitrot that happened over vacation.

elizagamedev:
Fixed scuba failure and added a feature flag to use the program cache; defaults to off.

pixelflinger:
> Fixed scuba failure and added a feature flag to use the program cache; defaults to off.

make sure to follow the process at go/filament-feature-flags (create the bug and reference it from this PR's commit message with BUGS=[])

elizagamedev:
> > Fixed scuba failure and added a feature flag to use the program cache; defaults to off.
> 
> make sure to follow the process at go/filament-feature-flags (create the bug and reference it from this PR's commit message with BUGS=[])

Got it, done

libs/utils/include/private/utils/FeatureFlagManager.h
```diff
@@ -90,6 +90,7 @@ class FeatureFlagManager {
             bool disable_handle_use_after_free_check = false;
             bool disable_heap_handle_tags = true; // FIXME: this should be false
             bool enable_asynchronous_operation = false;
+            bool enable_program_cache = false;
```
pixelflinger: move this feature flag under `engine`

libs/utils/src/FeatureFlagManager.cpp
```diff
@@ -91,6 +91,9 @@ FeatureFlagManager::FeatureFlagManager() : mFeatures{{
         { "backend.enable_asynchronous_operation",
           "Enable asynchronous operation for resource management.",
           &features.backend.enable_asynchronous_operation },
+        { "backend.enable_program_cache",
```
pixelflinger: move under `engine.`

libs/filabridge/src/Variant.cpp
```diff
@@ -79,7 +81,7 @@ constexpr inline size_t variant_count(bool lit) noexcept {
     size_t count = 0;
     for (size_t i = 0; i < VARIANT_COUNT; i++) {
         Variant variant(i);
-        if (!Variant::isValid(variant)) {
+        if (!Variant::isValidStandardVariant(variant)) {
```
pixelflinger, line 84: why this change here? Why exclude the depth variants?

filament/src/details/Material.h
```diff
@@ -73,6 +72,56 @@ class FMaterial : public Material {
         DefaultMaterialBuilder();
     };
 
+    // Used to change specialization constants at runtime internally by Filament.
+    //
+    // It would be nice to reuse Builder for this, but Builder allocates an entire map and asserts
+    // if specified constants don't exist. In constrast, SpecializationConstantsBuilder is
+    // copy-on-write and allows non-existent specialization constants to be named.
```
pixelflinger, line 84: better documentation needed:
- how do I get one of these?
- document the life-time constraints (it holds a raw pointer)

Also is this intended to become public eventually?

filament/src/details/Engine.cpp
```diff
@@ -538,14 +538,12 @@ void FEngine::init() {
...
+                        FMaterial::SpecializationConstantsBuilder constants =
+                                material->getSpecializationConstantsBuilder();
+                        constants.set(+ReservedSpecializationConstants::
+                                              CONFIG_DEBUG_DIRECTIONAL_SHADOWMAP,
                                 debug.shadowmap.debug_directional_shadowmap);
-
-                        material->invalidate(
-                                Variant::DIR | Variant::SRE | Variant::DEP,
-                                Variant::DIR | Variant::SRE);
+                        material->setSpecializationConstants(std::move(constants));
```
pixelflinger, line 546: just to be clear, where before this would just invalidate the HwProgram cache, now this will cache the new program as they become needed. And therefore switching between these settings shouldn't incur a recompilation each time anymore. correct?

the invalidate() call is no longer needed because now things happen on demand anyways, right?

What happens here when the feature flag is off? Is the old behvior preserved?

filament/src/MaterialCache.cpp
```diff
@@ -26,26 +26,33 @@
...
-        LOG(WARNING) << "MaterialCache was destroyed but wasn't empty";
+        LOG(WARNING) << "MaterialCache was destroyed but definitions cache wasn't empty";
+    }
+    if (!mPrograms.empty()) {
+        LOG(WARNING) << "MaterialCache was destroyed but program cache wasn't empty";
+    }
+    if (!mSpecializationConstantsInternPool.empty()) {
+        LOG(WARNING) << "MaterialCache was destroyed but specialization constants intern pool "
+                        "wasn't empty";
     }
```
pixelflinger: I don't like using warnings LOG because we don't get them from prod.

So either this should never happen by construction, we should use `assert_invariant()` (i.e. it's our bug). Or, this would happen if the user misused the API, in that case I think  we should:
- have precondition checks in the public APIs that would catch this (by definition if the user misused the aPI, we should be able to catch it at the point of misuse). and then use a assert_invariant() here (beause now this should not happen since we checked the preconditions. make sense?

filament/src/MaterialCache.h
```diff
@@ -35,29 +38,49 @@ class Material;
...
-    struct Key {
+    struct MaterialKey {
         struct Hash {
-            size_t operator()(Key const& x) const noexcept;
+            size_t operator()(MaterialKey const& x) const noexcept;
         };
-        bool operator==(Key const& rhs) const noexcept;
+        bool operator==(MaterialKey const& rhs) const noexcept;
 
         MaterialParser const* UTILS_NONNULL parser;
```
pixelflinger, line 49: how is it guaranteed that MaterialParser* outlives MaterialKey? 
I think MaterialParser* is owned by MaterialDefinition, so MaterialKey's lifetime depends on MaterialDefinition's lifetime.

Should MaterialKey no non-copiable? (so at least it can't end-up in places holding a pointer to a dead object)?

regardless, this should be called out and well documented.

filament/src/MaterialCache.cpp
```diff
@@ -56,15 +63,16 @@ MaterialDefinition* UTILS_NULLABLE MaterialCache::acquire(FEngine& engine,
...
     return mDefinitions.acquire(key, [&engine, parser = std::move(parser)]() mutable {
         return MaterialDefinition::create(engine, std::move(parser));
     });
 }
 
-void MaterialCache::release(FEngine& engine, MaterialDefinition const& definition) noexcept {
-    mDefinitions.release(Key{ &definition.getMaterialParser() },
+void MaterialCache::releaseMaterial(FEngine& engine,
+        MaterialDefinition const& definition) noexcept {
+    mDefinitions.release(MaterialKey{ &definition.getMaterialParser() },
```
pixelflinger, line 68: if I understand this correctly, the cached programs are aggressively removed from the cache, right?

e.g. if I have one Material configured with some spec constants, and then I change the spec constants on it, the old HwProgram<> will be destroyed right away (no more reference) -- it wouldn't be the case if I had create two Material from the same blob and used 2 different spec constants on it.

Should this be revisited at some point? e.g. maybe we want the cache to be smarter (whatever that means) about the evictions? 

Also I'm wondering if we don't have a problem with the life-time of HwProgram here?

e.g.:  I create a Material with a set of spec constant, and I attach this material into a renderable. We;re saying the renderable CANNOT cache the Handle<HwProgram>, right?
I think we're saying that once a spec constant changes, all the existing Handle<HwProgram> become invalid -- before they would become invalid only when you destroyed the Material.

Is there a risk here?

filament/src/details/Material.h
elizagamedev: It's not intended to become public since we changing the value of constants on a `Material` has always been a kind of internal hack. Clients will eventually be able to set constants on `MaterialInstance`s.

libs/filabridge/src/Variant.cpp
elizagamedev: The lit/unlit variants are treated in a mutual-exclusive capacity with the depth variants in `MaterialDefinition::acquirePrograms()` and `releasePrograms()`. Without this change we were double acquiring/freeing.

filament/src/details/Engine.cpp
elizagamedev: The cache at present is just reference counted per program. When that reference hits zero, that program is unceremoniously evicted. It's designed in a way that we could keep things around even after the reference hits zero in an LRU cache, but this will have to be done in a future PR. So regardless of whether the flag is on or off, we're recompiling each time even with this PR.

filament/src/MaterialCache.cpp
elizagamedev: This could be triggered by client code if they have a Material which outlives an Engine. What's the expected behavior if this happens without this PR? Do we already crash? If so, then I will replace these warnings with asserts. If not, we introduce creating new crashes.

filament/src/MaterialCache.h
elizagamedev: `MaterialKey` has to be copyable since that's a requirement of being usable as a key in `RefCountedMap`.

We can guarantee at least at the `MaterialCache` API boundary that we're not exceeding lifetimes because the only way to acquire a material is through a raw pointer/size pair to a material package, and the only way to free is through a reference to a `MaterialDefinition&`. Within `MaterialCache` we're also careful about lifetimes.

filament/src/MaterialCache.cpp
elizagamedev: Yep, exactly.

> Should this be revisited at some point? e.g. maybe we want the cache to be smarter (whatever that means) about the evictions?

I'd like to pair this with an LRU cache so we can keep around programs likely to be used again. This could be a configurable setting at the engine level too, e.g. how many programs to keep around?

> Also I'm wondering if we don't have a problem with the life-time of HwProgram here?

Yes, that's all true. That's precisely why we don't expose the ability to change the value of `Material`'s spec constants to clients. Every case where we do that internally in Filament is kind of a hack. Once we add the ability to override spec constants on `MaterialInstance`s themselves, we could remove this hack for good and simplify the code of `Material` a good deal.

...That said, the same could be said about matdbg. So we can't really reliably cache handles in places that aren't ultimately "owned" by Material anyways.

filament/src/MaterialCache.cpp
pixelflinger: ok, so in theory it's impossible for a Material to outlive the Engine because Engine cleans-up everything when it shuts down. So I think this means these should be `assert_invariant()` but we should definitely test this.

It's been sort of part of the public API that you don't have to destroy your objects because Engine::shutdown will do it for you.

filament/src/MaterialCache.h
pixelflinger: sgtm. can you add a comment to that effect near MaterialKey?

filament/src/MaterialCache.cpp
pixelflinger: can you update your doc with info about this? I'd like it to be captured somewhere. thanks!

elizagamedev: Done.

filament/src/MaterialCache.h
elizagamedev: Yep I added a comment.

Issue #1921, opened 2019-11-30
Using shared context on Windows fails in PlatformWGL::makeCurrent() and questiong about `createSwapChain()` 


roxlu:
As I'm trying to get shared contexts to work on Windows, especially with GLFW and Filament (_but any other windowing lib will probably have similar issues_) I've read up on OpenGL context creation on Windows to figure out what is causing the issue described in below.  As I've mostly used GLFW I'm probably still lacking some know-how about this. Please bare with me when I'm talking sh**t here :-) 

The problem I have, is that `PlatformWGL::makeCurrent()` triggers the error _reason: wglMakeCurrent() failed. hdc = 0000000014010D59 Windows error code: 2000. (null)._ In this code:

```c++
void PlatformWGL::makeCurrent(Platform::SwapChain* drawSwapChain,
                              Platform::SwapChain* readSwapChain) noexcept {
    ASSERT_PRECONDITION_NON_FATAL(drawSwapChain == readSwapChain,
                                  "PlatformWGL does not support distinct draw/read swap chains.");
    HDC hdc = (HDC)(drawSwapChain);
    if (hdc != NULL) {
        BOOL success = wglMakeCurrent(hdc, mContext);
        if (!ASSERT_POSTCONDITION_NON_FATAL(success, "wglMakeCurrent() failed. hdc = %p", hdc)) {
            reportLastWindowsError();
            wglMakeCurrent(0, NULL);
        }
    }
}
```

It's unclear what is causing this error, but while looking into this I ran into a couple of questions which someone can hopefully clarify. The call to `PlatformWGL::createDriver()` is executed form a different thread then the main thread which is important to understand the code there. Because when the code was executed on the same thread as where the `sharedGLContext` was created (when given), one could have created a new context by making that context current, retrieving `wglCreateContextAttribsARB()` and create a new one (we would have to find a way to get a `HDC`   for the curent window because we need this to pass this into the `wglCreateContextAttribsARB()` function, but Windows provides several solutions for this).

Because we *are* creating a context in a different thread, we have to create a dummy window, use that to create a context and retrieve a pointer to `wglCreateContextAttribsARB()`.  Once we have access to `wglCreateContextAttribsARB()` we can create the GL 4.1 context.  This is basically my understanding of what's going on in `PlatformWGL::createDriver()`.  

[...]

PR #9298, opened 2025-10-07
gl: Enable shared contexts for Windows

z3moon:
This change addresses a platform-specific issue on Windows where shared OpenGL contexts must be created on the same thread as the primary context.

To resolve this, we now pre-create a pool of shared contexts when the driver is initialized. These contexts are then distributed to other threads as needed, ensuring shared context creation requirement on Windows. This allows us to be able to use THREAD_POOL mode for shader compiler service on Windows. So Windows build uses THREAD_POOL mode instead of ASYNCHRONOUS mode as of this change.

This change also refactors the shader compiler service to catch and report errors in the same thread where the program compilation and linking actually performed as it's the correct way to use shared contexts. Otherwise it may crashes.


pixelflinger:
This might be fixing #1921

PR #9299, opened 2025-10-08
webgpu: fix queue submission and fence

poweifeng:
- Add WebGPUQeuueManager to manage queue submissions
 - The manager will keep a "last status" field that will track the status of the last submission.
 - Any created HwFence will be associated with a status object.
 - The object updates in a callback that indicates that the submitted workfload has been completed.


filament/backend/src/webgpu/WebGPUFence.cpp
```diff
@@ -15,76 +15,49 @@
...
+    return wait(0);
 }
 
-FenceStatus WebGPUFence::wait(uint64_t const timeout) {
+FenceStatus WebGPUFence::wait(uint64_t timeout) {
+    {
+        std::lock_guard<std::mutex> const lock(mLock);
+        if (!mState) {
+            return FenceStatus::TIMEOUT_EXPIRED;
+        }
```
pixelflinger: Why can't we wait that the state be set? or do something similar to what I did with Vulkan, with a secondary state?

I really would like it if the wait(timeout) call behaved properly instead of "failing" if the fence is not submitted yet.

filament/backend/src/webgpu/WebGPUFence.cpp
```diff
@@ -15,76 +15,49 @@
...
-                                        message;
-                                break;
-                        }
-                    }
-                }
-            });
+    {
+        std::unique_lock<std::mutex> lock(mLock);
+        bool const success = mCond.wait_until(lock, until, [this] {
+            return bool(mState);
```
pixelflinger, line 48: how can this be null, since we checked it wasn't null before?

filament/backend/src/webgpu/WebGPUFence.cpp
poweifeng: Done.

yeah sorry, this is bad logic. We'll just rely on the wait below.

filament/backend/src/webgpu/WebGPUFence.cpp
```diff
@@ -15,76 +15,43 @@
...
+        });
+        if (!success) {
+            return FenceStatus::TIMEOUT_EXPIRED;
+        }
+    }
+    return mState->waitForCompletion(timeout);
```
z3moon: It looks like the original `timeout` may have already been consumed while waiting for `mState` right above.

filament/backend/src/webgpu/WebGPUFence.cpp
poweifeng: removed redundancy

filament/backend/src/webgpu/WebGPUDriver.cpp
```diff
@@ -655,9 +605,7 @@ void WebGPUDriver::createRenderTargetR(Handle<HwRenderTarget> renderTargetHandle
 void WebGPUDriver::createFenceR(Handle<HwFence> fenceHandle, utils::ImmutableCString&& tag) {
     // The handle is constructed synchronously in createFenceS.
     const auto fence = handleCast<WebGPUFence>(fenceHandle);
-    assert_invariant(mQueue);
-    fence->addMarkerToQueueState(mQueue);
-    setDebugTag(fenceHandle.getId(), std::move(tag));
```
z3moon, line 660: nit: `tag` is not used. maybe revive `setDebugTag`? (the function is empty though)

filament/backend/src/webgpu/WebGPUQueueManager.h
```diff
@@ -0,0 +1,85 @@
...
+
+private:
+    std::mutex mLock;
+    std::condition_variable mCond;
+    FenceStatus mStatus;
+};
+
+struct WebGPUSubmission {
+    std::shared_ptr<WebGPUSubmissionState> state;
+};
```
z3moon: it looks like this is not used in the code. maybe remove it?

filament/backend/src/webgpu/WebGPUDriver.cpp
poweifeng: returned the setDebugTag

filament/backend/src/webgpu/WebGPUFence.cpp
pixelflinger: yeah I have this similar bug in the other implementations; I think it's an acceptable bug (or at least very low priority).

poweifeng: adjusted for time spent

filament/backend/src/webgpu/WebGPUQueueManager.h
poweifeng: done

PR #9301, opened 2025-10-08
renderdiff: add two params to local_test.sh

poweifeng:
- --no_rebuild will skip building the gltf_viewer again
 - --num_threads will limit the number of threads used in running the rendering test.

Both of these options are meant for debugging locally.

### DIFF
--- a/test/renderdiff/README.md
+++ b/test/renderdiff/README.md
@@ -51,14 +51,18 @@ renderings, do the following step
   bash test/renderdiff/generate.sh
   ```
 
-## Filtering Tests
+## Command-line Options
 
-You can run a subset of the tests by passing the `--test_filter` flag to `local_test.sh`. The
-filter supports wildcards (`*`) to match test names. For example, to run all tests in the
-`ClearCoat` suite, you can use the following command:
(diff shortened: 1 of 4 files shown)


test/renderdiff/README.md
```diff
@@ -51,14 +51,18 @@ renderings, do the following step
...
-## Filtering Tests
+## Command-line Options
 
-You can run a subset of the tests by passing the `--test_filter` flag to `local_test.sh`. The
-filter supports wildcards (`*`) to match test names. For example, to run all tests in the
-`ClearCoat` suite, you can use the following command:
+You can control the behavior of the test scripts with the following flags passed to `local_test.sh`:
+
+- `--test_filter=<filter>`: Run a subset of tests. The filter supports wildcards (`*`) to match test names.
+- `--no_rebuild`: Skip rebuilding the `gltf_viewer` executable.
```
z3moon, line 59: just curious. Is this because gltf_viewer takes long time?

poweifeng: it takes 10s of seconds on my laptop, so it's not a trivial amount of time waiting, when debugging.

PR #9302, opened 2025-10-08
Make JsonishParser use utils::Status

trie94:
Make JsonishParser use utils::Status, and return that status in MaterialParser if it's an error.
also:
- add unsupported error in utils::Status
- add invalid case in the test

### DIFF
--- a/libs/filament-matp/src/JsonishParser.cpp
+++ b/libs/filament-matp/src/JsonishParser.cpp
@@ -264,12 +264,12 @@ JsonishValue* JsonishParser::parseString() noexcept {
     } else {
         tmp = std::string(strLexeme->getStart(), strLexeme->getSize());
     }
-    std::pair<utils::Status, std::string> resolved = resolveEscapes(tmp);
-    if (!resolved.first.isOk()) {
-        mStatus = resolved.first;
+    auto [status, resolvedStr] = resolveEscapes(tmp);
+    if (!status.isOk()) {
+        mStatus = status;
         return nullptr;
     }
-    return new JsonishString(resolved.second);
+    return new JsonishString(resolvedStr);
 }
 
 JsonishValue* JsonishParser::parseValue() noexcept {


libs/filament-matp/src/JsonishParser.cpp
```diff
@@ -260,7 +267,7 @@ JsonishValue* JsonishParser::parseString() noexcept {
     } else {
         tmp = std::string(strLexeme->getStart(), strLexeme->getSize());
     }
-    return new JsonishString(tmp);
+    return new JsonishString(resolveEscapes(tmp, mStatus));
```
poweifeng: This feels like it should return null if `resolveEscapes` failed.  It's more in line with the rest of the error handling.

libs/filament-matp/src/JsonishParser.cpp
```diff
@@ -16,15 +16,15 @@
...
+#include <utils/Status.h>
 
 #include <string.h>
 
-#include <utils/string.h>
-
 namespace matp {
 
-static std::string resolveEscapes(const std::string& s) {
+static std::string resolveEscapes(const std::string& s, utils::Status& status) {
```
poweifeng: nit: return `std::pair<std::string, utils::Status>`

libs/filament-matp/src/JsonishParser.cpp
```diff
@@ -76,25 +79,29 @@ static std::string resolveEscapes(const std::string& s) {
...
-        std::cerr << "Incomplete escape sequence." << std::endl;
+        status = utils::Status::invalidArgument("Incomplete escape sequence.");
     }
 
     return out;
 }
 
 JsonishString::JsonishString(const std::string& string) : JsonishValue(STRING) {
-    mString = resolveEscapes(string);
+    mString = string;
```
poweifeng: nit: use constructor initializer list.

libs/filament-matp/src/JsonishParser.cpp
pixelflinger: yes! I've used that pattern of returning a pair<> and it've liked it.

libs/filament-matp/src/JsonishParser.cpp
trie94: yeah i added the status check before assigning it.

libs/filament-matp/src/JsonishParser.cpp
trie94: now i am tempted to add utils::StatusOr 😂  i changed to pair.

libs/filament-matp/src/JsonishParser.cpp
trie94: good point. actually i moved this to the header.

libs/filament-matp/src/JsonishParser.cpp
```diff
@@ -260,7 +264,12 @@ JsonishValue* JsonishParser::parseString() noexcept {
     } else {
         tmp = std::string(strLexeme->getStart(), strLexeme->getSize());
     }
-    return new JsonishString(tmp);
+    std::pair<utils::Status, std::string> resolved = resolveEscapes(tmp);
```
poweifeng: You could do
```
auto [status, resolvedStr] = resolveEscapes(tmp);
```

Or maybe
```
std::tie(mStatus, tmp) = resolveEscapes(tmp);
```

trie94: oh wow i didn't know about std::tie. i went with auto

PR #9310, opened 2025-10-10
filamat: remove call to Glslang's SPIR-V remapper

dneto0:
The Glslang SPIR-V remapper was removed, and replaced by the CanonicalizeIds pass in spirv-opt.  See https://github.com/KhronosGroup/SPIRV-Tools/pull/6174

Filament's GLSL post-optimization algorithm already runs CanonicalizeIds.

Bug: b:450891564


libs/filamat/src/GLSLPostProcessor.cpp
```diff
@@ -1002,17 +1000,15 @@ std::shared_ptr<Optimizer> GLSLPostProcessor::createOptimizer(
...
     optimizer->RegisterPass(CreateCanonicalizeIdsPass());
 
     // run optimizer
     if (!optimizer->Run(spirv.data(), spirv.size(), &spirv)) {
         slog.e << "SPIR-V optimizer pass failed" << io::endl;
         return;
     }
-
-    // Remove dead module-level objects: functions, types, vars
-    SpirvRemapWrapperRemap(spirv);
```
pixelflinger, line 1015: so this one is not needed anymore either?

dneto0: That's right.  The remapper functionality was moved from Glslang to spirv-opt's CanonicalizeIds pass.

Backstory is: When SPIR-V was originally developed, it was all inside Glslang. Along with initial code generation work, LunarG added an ID remapper in the same code directory.  This was long before SPIRV-Tools existed.
Over time SPIRV-Tools was created, and gained an optimizer framework.  Now the remapper is a maintenance burden, and was recently migrated to be a transform inside spirv-opt, and the original remapper was removed from Glslang's codebase.

If I remember correctly, the original remapper did some extra things like removing some dead variables at global scope.  But that should already be covered by other passes that Filament invokes.  See registerSizePasses https://github.com/google/filament/blob/5db8e0b9ab9a8f1e2eb18cd4fc560ab521b7e242/libs/filamat/src/GLSLPostProcessor.cpp#L1118

dneto0: I don't have a way to verify that the resulting sizes of binaries is comparable.  If not, then make sure the flow invokes the size-reduction passes as noted above.   I did not dig into what transforms are run under what conditions. 

Note that the canonical IDs transform is meant to make IDs for the same logical objects be consistent across a large number of SPIR-V modules.  E.g. the type for an unsigned 32 bit integer will have the same ID across all modules. That gives compression gains when using something like gzip to compress a large number of modules together.  (anything that uses a dictionary-based algorithm will benefit).    Gains are very significant, I hear.

NEW_RELEASE_NOTES.md
```diff
@@ -1,7 +1,12 @@
 # Filament Release Notes log
 
 **If you are merging a PR into main**: please add the release note below, under the *Release notes
-We are chaning the way Vulkan buffers are handled. We need to switch over to a managed (or view-based) model where the data stored inside the object is a proxy to a Vulkan object that can dynamically be swapped around.
+
+- We are chaning the way Vulkan buffers are handled. We need to switch over to a managed (or view-based) model where the data stored inside the object is a proxy to a Vulkan object that can dynamically be swapped around.
+
+- Vulkan: Removed a dependency on Glslang's deprecated SPIR-V remapper.
```
poweifeng: you'll need to rebase this on head, this looks to be very old.

And please prefix with `filamat:` instead of vulkan (we actually use spirv-opt then spirv-cross to target material for each backend).

libs/filamat/src/GLSLPostProcessor.cpp
pixelflinger: yes the canonical ids have helped our line dictionary a lot!

NEW_RELEASE_NOTES.md
dneto0: It's only one commit newer than main.

See https://raw.githubusercontent.com/google/filament/refs/heads/main/NEW_RELEASE_NOTES.md

dneto0: I've changed the prefix to `filamat:`

NEW_RELEASE_NOTES.md
```diff
@@ -1,7 +1,12 @@
 # Filament Release Notes log
 
 **If you are merging a PR into main**: please add the release note below, under the *Release notes
-We are chaning the way Vulkan buffers are handled. We need to switch over to a managed (or view-based) model where the data stored inside the object is a proxy to a Vulkan object that can dynamically be swapped around.
+
+- We are chaning the way Vulkan buffers are handled. We need to switch over to a managed (or view-based) model where the data stored inside the object is a proxy to a Vulkan object that can dynamically be swapped around.
+
+- filamat: Removed a dependency on Glslang's deprecated SPIR-V remapper.
+  The functionality is already implemented by calling the CanonicalizeIds pass
+  in the SPIRV-Tools, and should be a non-functional change.
```
poweifeng, line 10: Ok, I see what the confusion is now.  

Can you help us remove this line?  It was added errorneously a while back.
```
- We are chaning the way Vulkan buffers are handled. We need to switch over to a managed (or view-based) model where the data stored inside the object is a proxy to a Vulkan object that can dynamically be swapped around.
```

Next is, would you mind moving your description to below line 12?  That is, after
```
*If you are cherry-picking a commit into an rc/ branch**: add the release note under the
```
We have a script that needs the new commit descriptions to be at the very bottom.

Thank you!

dneto0: I think I put it in the right spot now.

PR #9312, opened 2025-10-11
fix wasm build.

pixelflinger:
The new AsyncJobQueue can't work with wasm, which doesn't support  threads.

### DIFF
--- a/libs/utils/include/utils/AsyncJobQueue.h
+++ b/libs/utils/include/utils/AsyncJobQueue.h
@@ -48,13 +48,17 @@ public:
     // adds a job to the queue. no-op if drainAndExit() was called.
     void push(Job&& job);
 
+    bool isValid() const noexcept;
+
 private:
+#if !defined(__EMSCRIPTEN__)
     using Container = std::vector<Job>;
     std::thread mThread;
     Mutex mLock; // NOLINT(*-include-cleaner)
     Condition mCondition; // NOLINT(*-include-cleaner)
     Container mQueue;
     bool mExitRequested = false;
+#endif
 };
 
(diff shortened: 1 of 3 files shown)


libs/utils/include/utils/AsyncJobQueue.h
```diff
@@ -48,13 +48,17 @@ class AsyncJobQueue {
     // adds a job to the queue. no-op if drainAndExit() was called.
     void push(Job&& job);
 
+    bool isValid() const noexcept;
+
 private:
+#if !defined(__EMSCRIPTEN__)
```
z3moon, line 54: Do we need this? We have other source code that already use std::thread and mutex, etc. Do you plan on removing them as well?

pixelflinger: We're already doing that for emscripten.  All the places where we spawn threads are ifdef'ed out on emscripten.

z3moon: I see. Although ShaderCompilerService is not instanced for web, code itself doesn't seem to be protected with ifdef. I guess the intention behind this is for binary size optimization or something? In any case, would it be enough for us not to use ShaderCompilerService class on web?

pixelflinger: the web doesn't support threads, so if you were trying to instantiate it it would fail at runtime.  So we need to make sure that we're not using threads in ShaderCompilerService when we're on web.. If it's not protected, it probably should be (or we need to make sure we're never using it). Maybe a `#error` directive in that case would be good.

PR #9313, opened 2025-10-11
Add a Renderer API to force skipping frames

pixelflinger:
- Add a Renderer API to force skipping frames
Renderer::skipNextFrames(size_t) can be used to force filament to
pretend the next N frames must be skipped. This is mostly useful for
debugging.

- Add DebugOptions to Settings
We still need to move the "Debug" features of gltf_viewer to this,
but this give us a framework to do it. 
Currently there is one debug option that allows to set a number of
frames to skip.

- ViewerGui propose a button to skip 10 frames using this framework

### DIFF
--- a/libs/viewer/src/Settings.cpp
+++ b/libs/viewer/src/Settings.cpp
@@ -556,7 +556,7 @@ static int parse(jsmntok_t const* tokens, int i, const char* jsonChunk, DebugOpt
             i = parse(tokens, i + 1);
         }
         if (i < 0) {
-            slog.e << "Invalid viewer options value: '" << STR(tok, jsonChunk) << "'" << io::endl;
+            slog.e << "Invalid debug options value: '" << STR(tok, jsonChunk) << "'" << io::endl;
             return i;
         }
     }


filament/src/FrameSkipper.cpp
```diff
@@ -75,4 +83,14 @@ void FrameSkipper::submitFrame(DriverApi& driver) noexcept {
     fences[last] = driver.createFence();
 }
 
+void FrameSkipper::skipNextFrames(size_t frameCount) const noexcept {
+    frameCount = std::min(frameCount, size_t(std::numeric_limits<decltype(mFrameToSkip)>::max()));
+    mFrameToSkip = uint16_t(frameCount);
```
z3moon, line 88: This always overwrites the existing value (if it's not 0) of `mFrameToSkip`. This is right if any subsequent call's `frameCount` is properly adjusted.

libs/viewer/src/Settings.cpp
```diff
@@ -681,6 +702,12 @@ void applySettings(Engine* engine, const ViewerOptions& settings, Camera* camera
     }
 }
 
+void applySettings(Engine* engine, const DebugOptions& settings, Renderer* renderer) {
+    if (!renderer->getFrameToSkipCount()) {
+        renderer->skipNextFrames(settings.skipFrames);
+    }
+}
```
z3moon, line 709: nit: this debug option (e.g., like 1000 frames skipping for automation test) won't be set if there's any frame skip (like 1~2) left in the renderer. Honestly I'm not sure if we should handle this, or maybe okay. But this slightly contradicts the other part I left a comment.

filament/src/FrameSkipper.cpp
pixelflinger: Right, this is intended to be called once. There is a getter if the caller wants to do an adjustment.

libs/viewer/src/Settings.cpp
pixelflinger: Yeah, this is a tough one.  The problem is that applySetting is called from the wasm UI, we can't put any logic in the "app" (because there is no app, just the Settings framework).  So the Settings "API" is that this value is only set when there is not a skipFrame in progress -- this is different from the Renderer version of the API (They don't have to be the same). 
Also, I put this is the debug settings, where I think it's acceptable to have special behaviors.

filament/src/FrameSkipper.cpp
poweifeng, line 88: nit: pedantically, `uint16_t` should be `decltype(mFrameToSkip)`

libs/viewer/src/Settings.cpp
```diff
@@ -544,6 +543,26 @@ static int parse(jsmntok_t const* tokens, int i, const char* jsonChunk, ViewerOp
...
+static int parse(jsmntok_t const* tokens, int i, const char* jsonChunk, DebugOptions* out) {
+    CHECK_TOKTYPE(tokens[i], JSMN_OBJECT);
+    int size = tokens[i++].size;
+    for (int j = 0; j < size; ++j) {
+        const jsmntok_t tok = tokens[i];
+        CHECK_KEY(tok);
+        if (compare(tok, jsonChunk, "skipFrames") == 0) {
+            i = parse(tokens, i + 1, jsonChunk, &out->skipFrames);
+        } else {
+            slog.w << "Invalid viewer options key: '" << STR(tok, jsonChunk) << "'" << io::endl;
```
poweifeng:
```suggestion
            slog.w << "Invalid debug options key: '" << STR(tok, jsonChunk) << "'" << io::endl;
```

libs/viewer/src/Settings.cpp
```diff
@@ -544,6 +543,26 @@ static int parse(jsmntok_t const* tokens, int i, const char* jsonChunk, ViewerOp
...
+            i = parse(tokens, i + 1);
+        }
+        if (i < 0) {
+            slog.e << "Invalid viewer options value: '" << STR(tok, jsonChunk) << "'" << io::endl;
```
poweifeng:
```suggestion
            slog.e << "Invalid debug options value: '" << STR(tok, jsonChunk) << "'" << io::endl;
```

PR #9314, opened 2025-10-13
Fix, Abseil compilation error with StaticString

bejado:


bejado:
> is there a way to make Abseil aware of StaticString and even pass the information that it is static and therefore can be kept?

Yes, I think through [`AbslStringify`](https://abseil.io/docs/cpp/guides/abslstringify). Though at the moment it would only work with "real" Abseil and not our custom minimal implementation.

PR #9316, opened 2025-10-13
Revert "Make check-headers test use c++20"

bejado:
This reverts commit decf3168025c7b0eb1af840f2ecb9f3773fb8203.

Reverting this as some clients still rely on C++17-compatible public headers.

### DIFF
--- a/test/check-headers/test.sh
+++ b/test/check-headers/test.sh
@@ -66,6 +66,8 @@ for include in "${includes[@]}"; do
         echo "#define SYSTRACE_TAG SYSTRACE_TAG_DISABLED" >> ${TMP_FILE}
     fi
     echo "#include <${include}>" >> ${TMP_FILE}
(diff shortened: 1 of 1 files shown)


test/check-headers/test.sh
```diff
@@ -66,6 +66,6 @@ for include in "${includes[@]}"; do
         echo "#define SYSTRACE_TAG SYSTRACE_TAG_DISABLED" >> ${TMP_FILE}
     fi
     echo "#include <${include}>" >> ${TMP_FILE}
-    clang -std=c++20 -I "${FILAMENT_HEADERS}" ${TMP_FILE} -c -o /dev/null
+    clang -std=c++17 -I "${FILAMENT_HEADERS}" ${TMP_FILE} -c -o /dev/null
```
poweifeng, line 71: I think we need to add a comment as to why the headers need to compile against c++17 vs the rest of filament.

PR #9318, opened 2025-10-14
webgpu: fix picking

poweifeng:
Context:
 - Picking has switched to use integer textures and this texture is meant to have only RG components (#8917).
 - WebGPU only supports 4 component textures, and so we need to to a blit-based transform to turn 2 components into 4.
 - WebGPUBlitter supports only float textures

In this commit, we add support for integer textures to WebGPUBlitter. In doing so, we refactor parts of the shader construction code in the blitter class and introduce a couple helper functions for texture formats.

This fixes picking (verifiable with gltf_viewer)

### DIFF
--- a/.github/actions/get-commit-msg/action.yml
+++ b/.github/actions/get-commit-msg/action.yml
@@ -1,43 +1,53 @@
+# This action retrieves the latest commit message from a push or pull_request event
+# and makes it available as an output variable named 'msg'.
 name: 'Get commit message'
 outputs:
   msg:
     value: ${{ steps.action_output.outputs.msg }}
+  hash:
+    value: ${{ steps.action_output.outputs.hash }}
 runs:
   using: "composite"
   steps:
     - name: Find commit message (on push)
       if: github.event_name == 'push'
       shell: bash
+      env:
+        COMMIT_MESSAGE: ${{ github.event.head_commit.message }}
       run: |
         AUTHOR_NAME="${{ github.event.head_commit.author.name }}"
         AUTHOR_EMAIL="${{ github.event.head_commit.author.email }}"
         TSTAMP="${{ github.event.head_commit.timestamp }}"
-        echo "commit ${{ github.event.head_commit.id }}" >> /tmp/commit_msg.txt
+        HASH="${{ github.event.head_commit.id }}"
+        echo "commit $HASH" >> /tmp/commit_msg.txt
         echo "Author: ${AUTHOR_NAME}<${AUTHOR_EMAIL}>" >> /tmp/commit_msg.txt
         echo "Date: ${TSTAMP}" >> /tmp/commit_msg.txt
         echo "" >> /tmp/commit_msg.txt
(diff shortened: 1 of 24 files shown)


pixelflinger (changes requested):
So specifically for picking, the format is not exposed to the public API (in fact we're doing something different for ES2).
So could we just use RGBA on the filament side where we're on WebGPU? This would void these changes in the backend?

poweifeng:
> So specifically for picking, the format is not exposed to the public API (in fact we're doing something different for ES2). So could we just use RGBA on the filament side where we're on WebGPU? This would void these changes in the backend?

Yeah, that's a good point.  But this is a general improvement to the existing blitter code. I think it'll help WebGPU pass additional backend tests (in particular `LoadImageTest.UpdateImage2D`).

poweifeng:
separated wgsl part into a `.inc`

filament/backend/src/webgpu/WebGPUBlitter.cpp
```diff
@@ -720,63 +613,109 @@ wgpu::ShaderModule const& WebGPUBlitter::getOrCreateShaderModule(ShaderModuleKey
...
+    // Fill in fragment in the shader module, second pass
+    source = replace(source, {
+        { TEXTURE_TYPE, srcTextureType },
+        { RET_ATTRIBUTE, retAttribute },
+        { RET_TYPE, retType },
+    });
+
+    // Fill in fragment in the shader module, third dpass
+    // TEXTURE_TYPE, RET_ATTRIBUTE, RET_TYPE have a depedency on SRC_PRIM_TYPE, DST_PRIM_TYPE,
+    // VECTOR_DIM so we need another pass to fill them in.
```
bejado: nit
```suggestion
    // Fill in fragment in the shader module, third pass
    // TEXTURE_TYPE, RET_ATTRIBUTE, RET_TYPE have a dependency on SRC_PRIM_TYPE, DST_PRIM_TYPE,
    // VECTOR_DIM so we need another pass to fill them in.
```

filament/backend/src/webgpu/WebGPUBlitter.cpp
```diff
@@ -720,63 +613,109 @@ wgpu::ShaderModule const& WebGPUBlitter::getOrCreateShaderModule(ShaderModuleKey
...
+    if (isDepthDst) {
+        retAttribute = "@builtin(frag_depth)";
+        retType = "f32";
+    } else {
+        retAttribute = "@location(0)";
+        retType = "vec4{{DST_PRIM_TYPE}}";
+    }
+
+    using ReplacementMap = std::unordered_map<std::string_view, std::string_view>;
+    auto const replace = [&](std::string_view src, ReplacementMap const& replacements) {
```
bejado, line 678: Does this lambda need to be capturing?

poweifeng: Yes, but only for `PLACEHOLDER_PREFIX` and `PLACEHOLDER_SUFFIX`. I could be more explicit if you prefer.

pixelflinger (approved):

PR #9320, opened 2025-10-14
Add new public API to query a sampler transform name field.

rafadevai:
This new API will let filament users query a Material object the value of the `transformName` field of a specified sampler parameter.

The transformName is an optional field, so if its not defined by the user, it will return a nullptr value.

- A new test was added to test_filamat to validate the serialization.

### DIFF
--- a/android/filament-android/src/main/cpp/Material.cpp
+++ b/android/filament-android/src/main/cpp/Material.cpp
@@ -284,7 +284,8 @@ Java_com_google_android_filament_Material_nGetParameterTransformName(JNIEnv* env
         jlong nativeMaterial, jstring samplerName_) {
     Material* material = (Material*) nativeMaterial;
     const char* samplerName = env->GetStringUTFChars(samplerName_, 0);
-    jstring transformName_ = env->NewStringUTF(material->getParameterTransformName(samplerName));
+    const char* transformName = material->getParameterTransformName(samplerName);
+    jstring transformName_ = env->NewStringUTF(transformName ? transformName : "");
     env->ReleaseStringUTFChars(samplerName_, samplerName);
     return transformName_;
 }
--- a/android/filament-android/src/main/java/com/google/android/filament/Material.java
+++ b/android/filament-android/src/main/java/com/google/android/filament/Material.java
@@ -789,8 +789,8 @@ public class Material {
 
     /**
      * 
-     * Returns the name of the transform name parameter associated with the given sampler parameter. 
-     * In the case the parameter don't have a transform name field, it will return an empty value.
+     * Returns the name of the transform parameter associated with the given sampler parameter.
+     * In the case the parameter doesn't have a transform name field, it will return an empty string.
      * 
      * @param samplerName the name of the sampler parameter to query.
      * 
--- a/filament/include/filament/Material.h
+++ b/filament/include/filament/Material.h
@@ -353,9 +353,8 @@ public:
 
     /**
      *
-     * Gets the name of the transform name field associated for the given sampler parameter.
-     * In the case where the parameter does not have a transform name field, it will return an empty
-     * value.
+     * Gets the name of the transform field associated for the given sampler parameter.
+     * In the case where the parameter does not have a transform name field, it will return nullptr.
      *
      * @param samplerName the name of the sampler parameter to query.
      *


rafadevai:
One question I have is, do I have to update the MATERIAL_VERSION?

I ran the filament_test_material_parser and PASSED but just want to double check

libs/filamat/tests/test_filamat.cpp
```diff
@@ -936,6 +936,25 @@ TEST_F(MaterialCompiler, FeatureLevel0Sampler2D) {
   EXPECT_TRUE(result.isValid());
 }
 
+TEST_F(MaterialCompiler, SamplerTransformName) {
```
z3moon, line 939: This test doesn't exercise the newly introduced method `getParameterTransformName`. You may want to add one.

show50726:
Is this related to the job of removing `commitStreamUniformAssociations` ?

rafadevai:
> Is this related to the job of removing `commitStreamUniformAssociations` ?

Yes

poweifeng:
> One question I have is, do I have to update the MATERIAL_VERSION?
> 
> I ran the filament_test_material_parser and PASSED but just want to double check

You don't need to modify `MATERIAL_VERSION`, but you do need to add an entry in `NEW_RELEASE_NOTES.md` like [here](https://github.com/google/filament/commit/075726db8bf0767bbb05689e88ec582a3fd5d581#diff-d38ed88bacefa6d931992ebe309c31042c16962c2c688d347334b16506e8a99bR7-R11)

libs/filamat/src/eiff/MaterialInterfaceBlockChunk.cpp
```diff
@@ -79,6 +79,12 @@ void MaterialSamplerInterfaceBlockChunk::flatten(Flattener& f) {
         f.writeUint8(static_cast<uint8_t>(sInfo.precision));
         f.writeBool(sInfo.filterable);
         f.writeBool(sInfo.multisample);
+
+        bool hasTransformNameField = !sInfo.transformName.empty();
+        f.writeBool(hasTransformNameField);
```
poweifeng: Why do you need two fields?  Can an empty transform name indicate that there wasn't one provided? (I don't know what happens in `writeString`, but I imagine you could write a 0 in the worst case.)

filament/include/filament/Material.h
```diff
@@ -350,6 +350,11 @@ class UTILS_PUBLIC Material : public FilamentAPI {
     //! Indicates whether an existing parameter is a sampler or not.
     bool isSampler(const char* UTILS_NONNULL name) const noexcept;
 
+    //! Returns the name of the transform name parameter associated with the given parameter, or
+    //! nullptr if the given parameter doesn't have a transform name field.
+    const char* UTILS_NULLABLE getParameterTransformName(
```
poweifeng, line 363: I think you'll need a corresponding java binding.  See [here](https://github.com/google/filament/blob/main/android/filament-android/src/main/java/com/google/android/filament/Material.java)

pixelflinger: and in jsbingings.cpp too.

filament/include/filament/Material.h
```diff
@@ -350,6 +350,11 @@ class UTILS_PUBLIC Material : public FilamentAPI {
     //! Indicates whether an existing parameter is a sampler or not.
     bool isSampler(const char* UTILS_NONNULL name) const noexcept;
 
+    //! Returns the name of the transform name parameter associated with the given parameter, or
+    //! nullptr if the given parameter doesn't have a transform name field.
```
pixelflinger: use /** */ syntax for multiline doxygen. Make sure to explain what the "parameter" is (it's the name of a sampler right?).

Also maybe document that the returned pointer is valid as long as Material is valid.

filament/include/filament/Material.h
```diff
@@ -350,6 +350,11 @@ class UTILS_PUBLIC Material : public FilamentAPI {
...
+    const char* UTILS_NULLABLE getParameterTransformName(
+            const char* UTILS_NONNULL name) const noexcept;
```
pixelflinger: maybe use a less generic name than "name" (e.g. "samplerName")?

filament/src/MaterialParser.cpp
```diff
@@ -571,13 +573,22 @@ bool ChunkSamplerInterfaceBlock::unflatten(Unflattener& unflattener,
             return false;
         }
 
+        if (!unflattener.read(&hasFieldTransformName)) {
+            return false;
+        }
+
+        if (hasFieldTransformName && !unflattener.read(&fieldTransformName)) {
```
pixelflinger: I don't love having the call to unflatterner on the same line.  It's not super clear if it's always called (it's not) or not. Just split into several lines.

libs/filabridge/include/private/filament/SamplerInterfaceBlock.h
```diff
@@ -64,6 +64,7 @@ class SamplerInterfaceBlock {
         bool filterable;            // whether the sampling should be filterable.
         bool multisample;           // multisample capable
         ShaderStageFlags stages;    // stages the sampler can be accessed from
+        utils::CString transformName; // name of the uniform holding the transform matrix for this sampler
```
pixelflinger: you should now use `utils::ImmutableCString` everywhere in this change.

libs/filamat/src/eiff/MaterialInterfaceBlockChunk.cpp
pixelflinger: yes I think it might make things simpler?

libs/filamat/tests/test_filamat.cpp
pixelflinger: agreed, if we're adding tests, we should check it returns the right name.

also I'd like this test to be slightly more complete by having several samplers, some with, some without the transform name, and make sure we get back the right name from the right sampler.

pixelflinger (changes requested):

libs/filamat/src/eiff/MaterialInterfaceBlockChunk.cpp
rafadevai: When I do the that, it fails to compile the materials

filament/include/filament/Material.h
```diff
@@ -350,6 +350,19 @@ class UTILS_PUBLIC Material : public FilamentAPI {
     //! Indicates whether an existing parameter is a sampler or not.
     bool isSampler(const char* UTILS_NONNULL name) const noexcept;
 
+
+    /**
+     * 
+     * Gets the name of the transform name field associated for the given sampler parameter. 
+     * In the case the parameter don't have a transform name field, it will return an empty value.
```
poweifeng:
```suggestion
     * In the case where the parameter does not have a transform name field, it will return an empty value.
```

libs/filamat/tests/test_filamat.cpp
rafadevai: Done

libs/filamat/src/eiff/MaterialInterfaceBlockChunk.cpp
rafadevai: Done, changed so it doesnt need to extra bool

filament/include/filament/Material.h
rafadevai: Done

filament/include/filament/Material.h
rafadevai: Done

filament/src/MaterialParser.cpp
rafadevai: Done

libs/filamat/tests/test_filamat.cpp
rafadevai: This test is for the serialization part but added more tests for querying the values.

filament/include/filament/Material.h
rafadevai: Done

libs/filabridge/include/private/filament/SamplerInterfaceBlock.h
rafadevai: Done

filament/test/filament_test_material.cpp
```diff
@@ -0,0 +1,86 @@
+/*
+ * Copyright (C) 2020 The Android Open Source Project
```
poweifeng: ### 
```suggestion
 * Copyright (C) 2025 The Android Open Source Project
```

filament/test/filament_test_material.cpp
```diff
@@ -0,0 +1,86 @@
...
+
+using namespace filament;
+
+TEST(MaterialTransformName, QuerySamplerWithTransform) {
+    Engine* engine = Engine::create(Engine::Backend::DEFAULT);
+
+    Material* material = Material::Builder()
+    .package(FILAMENT_TEST_RESOURCES_TEST_MATERIAL_TRANSFORMNAME_DATA, FILAMENT_TEST_RESOURCES_TEST_MATERIAL_TRANSFORMNAME_SIZE)
+    .build(*engine);
+    ASSERT_NE(material, nullptr);
```
poweifeng, line 33: nit: the indentation here looks off

poweifeng: also for the other tests

filament/test/update_test_material.sh
```diff
@@ -6,11 +6,15 @@ set -ex
...
-OUTPUT="filament/test/test_material.filamat"
+OUTPUT_DIR="filament/test/"
 
 ./build.sh -p desktop release matc
 MATC=out/cmake-release/tools/matc/matc
-${MATC} --platform all --api all -o ${OUTPUT} samples/materials/sandboxLit.mat
+${MATC} --platform all --api all -o ${OUTPUT_DIR}/test_material.filamat samples/materials/sandboxLit.mat
+
+./build.sh -p desktop release matc
+MATC=out/cmake-release/tools/matc/matc
```
poweifeng: Shouldn't need these two lines

libs/filamat/tests/test_filamat.cpp
```diff
@@ -936,6 +936,25 @@ TEST_F(MaterialCompiler, FeatureLevel0Sampler2D) {
...
+  std::string shaderCode(R"(
+        void material(inout MaterialInputs material) {
+            prepareMaterial(material);
+            vec3 uvw = materialParams.sampler_transform * vec3(0.0, 0.0, 0.0);
```
poweifeng, line 943: Can you add a negative case, where the transform name does not exists and it fails?

NEW_RELEASE_NOTES.md
```diff
@@ -6,3 +6,6 @@
 appropriate header in [RELEASE_NOTES.md](./RELEASE_NOTES.md).
 
 ## Release notes for next branch cut
+
+materials: Add a new API getParameterTransformName that will return the value of the transformName field of a sampler
+parameter.
```
poweifeng: needs to have `[⚠️ **Recompile Materials**]` at the end of this line

Add a `-` at the beginning of this line

filament/include/filament/Material.h
rafadevai: Done

filament/test/filament_test_material.cpp
rafadevai: Done

filament/test/update_test_material.sh
rafadevai: Done

libs/filamat/tests/test_filamat.cpp
rafadevai: Done

NEW_RELEASE_NOTES.md
rafadevai: Done

filament/test/filament_test_material.cpp
rafadevai: Done

android/filament-android/src/main/java/com/google/android/filament/Material.java
```diff
@@ -787,6 +787,21 @@ public boolean hasParameter(@NonNull String name) {
         return nHasParameter(getNativeObject(), name);
     }
 
+    /**
+     * 
+     * Returns the name of the transform name parameter associated with the given sampler parameter.
```
pixelflinger:
```suggestion
     * Returns the name of the transform parameter associated with the given sampler parameter. 
```

android/filament-android/src/main/java/com/google/android/filament/Material.java
```diff
@@ -787,6 +787,21 @@ public boolean hasParameter(@NonNull String name) {
         return nHasParameter(getNativeObject(), name);
     }
 
+    /**
+     * 
+     * Returns the name of the transform name parameter associated with the given sampler parameter. 
+     * In the case the parameter don't have a transform name field, it will return an empty value.
```
pixelflinger:
```suggestion
     * In the case the parameter doesn't have a transform name field, it will return an empty string.
```

filament/include/filament/Material.h
```diff
@@ -350,6 +350,20 @@ class UTILS_PUBLIC Material : public FilamentAPI {
     //! Indicates whether an existing parameter is a sampler or not.
     bool isSampler(const char* UTILS_NONNULL name) const noexcept;
 
+
+    /**
+     *
+     * Gets the name of the transform name field associated for the given sampler parameter.
```
pixelflinger:
```suggestion
     * Gets the name of the transform field associated for the given sampler parameter.
```

filament/include/filament/Material.h
```diff
@@ -350,6 +350,20 @@ class UTILS_PUBLIC Material : public FilamentAPI {
...
+     * In the case where the parameter does not have a transform name field, it will return an empty
+     * value.
```
pixelflinger:
```suggestion
     * In the case where the parameter does not have a transform name field, it will return nullptr.
```

android/filament-android/src/main/cpp/Material.cpp
```diff
@@ -279,6 +279,16 @@ Java_com_google_android_filament_Material_nHasParameter(JNIEnv* env, jclass,
     return (jboolean) hasParameter;
 }
 
+extern "C" JNIEXPORT jstring JNICALL
+Java_com_google_android_filament_Material_nGetParameterTransformName(JNIEnv* env, jclass,
+        jlong nativeMaterial, jstring samplerName_) {
+    Material* material = (Material*) nativeMaterial;
+    const char* samplerName = env->GetStringUTFChars(samplerName_, 0);
+    jstring transformName_ = env->NewStringUTF(material->getParameterTransformName(samplerName));
```
pixelflinger: `getParameterTransformName` could return `nullptr` if the samplerName didn't exist. In that case, you need to return an empty string according to the javadoc.

rafadevai: Done

filament/include/filament/Material.h
rafadevai: Done

filament/include/filament/Material.h
rafadevai: Done

android/filament-android/src/main/java/com/google/android/filament/Material.java
rafadevai: Done

android/filament-android/src/main/java/com/google/android/filament/Material.java
rafadevai: Done

PR #9323, opened 2025-10-15
Various fixes to support gcc compilation.

haroonq:
Various code changes to allow filament to be built using gcc.  Most of these changes should hopefully be non-controversial.

I'm working on a separate set of CMake changes that will let users set a FILAMENT_ENABLE_EXPERIMENTAL_GCC_SUPPORT  option to allow it to work. I'm happy to maintain (to a certain degree) this experimental support.

### DIFF
--- a/libs/gltfio/CMakeLists.txt
+++ b/libs/gltfio/CMakeLists.txt
@@ -196,10 +196,7 @@ if (WEBGL_PTHREADS)
     target_compile_definitions(gltfio_core PUBLIC -DFILAMENT_WASM_THREADS)
 endif()
 
-set(GLTFIO_WARNINGS -Wall -Werror)
-if (NOT MSVC)
-    target_compile_options(gltfio_core PRIVATE ${GLTFIO_WARNINGS})
-endif()
+target_compile_options(gltfio_core PRIVATE ${FILAMENT_WARNINGS})
 
 if (NOT WEBGL AND NOT ANDROID AND NOT IOS)
 


libs/gltfio/CMakeLists.txt
```diff
@@ -196,10 +196,7 @@ if (WEBGL_PTHREADS)
     target_compile_definitions(gltfio_core PUBLIC -DFILAMENT_WASM_THREADS)
 endif()
 
-set(GLTFIO_WARNINGS -Wall -Werror)
```
poweifeng, line 199: why remove this?

libs/gltfio/src/extended/TangentSpaceMeshWrapper.h
```diff
@@ -67,11 +67,6 @@ struct TangentSpaceMeshWrapper {
     short4* getQuats() noexcept;
     uint3* getTriangles();
 
-    template<typename T>
-    using is_supported_aux_t = typename std::enable_if<
-            std::is_same<float2*, T>::value || std::is_same<float3*, T>::value ||
-            std::is_same<float4*, T>::value || std::is_same<ushort3*, T>::value ||
-            std::is_same<ushort4*, T>::value>::type;
     template<typename T, typename = is_supported_aux_t<T>>
```
poweifeng, line 70: how does this compile without `is_supported_aux_t` defined?

filament/src/fg/FrameGraphRenderPass.h
```diff
@@ -33,14 +34,20 @@ namespace filament {
...
     static constexpr size_t ATTACHMENT_COUNT = backend::MRT::MAX_SUPPORTED_RENDER_TARGET_COUNT + 2;
     struct Attachments {
-        union {
-            FrameGraphId<FrameGraphTexture> array[ATTACHMENT_COUNT] = {};
-            struct {
-                FrameGraphId<FrameGraphTexture> color[backend::MRT::MAX_SUPPORTED_RENDER_TARGET_COUNT];
-                FrameGraphId<FrameGraphTexture> depth;
-                FrameGraphId<FrameGraphTexture> stencil;
-            };
-        };
```
pixelflinger, line 43: why is this not working with gcc?

filament/src/fg/FrameGraphRenderPass.h
```diff
@@ -33,14 +34,20 @@ namespace filament {
...
+        FrameGraphId<FrameGraphTexture> color[backend::MRT::MAX_SUPPORTED_RENDER_TARGET_COUNT];
+        FrameGraphId<FrameGraphTexture> depth;
+        FrameGraphId<FrameGraphTexture> stencil;
+
+        FrameGraphId<FrameGraphTexture>& operator[](int index) {
+            FILAMENT_CHECK_PRECONDITION(index >= 0 && index < ATTACHMENT_COUNT);
```
pixelflinger: we definitely don't want a FILAMENT_CHECK_* here maybe use assert_invariant() instead.

filament/src/fg/FrameGraphRenderPass.h
```diff
@@ -33,14 +34,20 @@ namespace filament {
...
-                FrameGraphId<FrameGraphTexture> color[backend::MRT::MAX_SUPPORTED_RENDER_TARGET_COUNT];
-                FrameGraphId<FrameGraphTexture> depth;
-                FrameGraphId<FrameGraphTexture> stencil;
-            };
-        };
+        FrameGraphId<FrameGraphTexture> color[backend::MRT::MAX_SUPPORTED_RENDER_TARGET_COUNT];
+        FrameGraphId<FrameGraphTexture> depth;
+        FrameGraphId<FrameGraphTexture> stencil;
+
+        FrameGraphId<FrameGraphTexture>& operator[](int index) {
```
pixelflinger: add a `const` version, also should be `noexcept` (once you remove the FILAMENT_CHECK)

filament/src/ds/SsrPassDescriptorSet.cpp
```diff
@@ -70,7 +70,7 @@ void SsrPassDescriptorSet::setFrameUniforms(FEngine const& engine,
 
 void SsrPassDescriptorSet::prepareHistorySSR(FEngine const& engine, Handle<HwTexture> ssr) noexcept {
     mDescriptorSet.setSampler(engine.getPerViewDescriptorSetLayoutSsrVariant(),
-            +PerViewBindingPoints::SSR_HISTORY, ssr, {
+            +PerViewBindingPoints::SSR_HISTORY, ssr, SamplerParams{
```
pixelflinger, line 73: Why is this not working?  we're bound to constantly break the GCC build because that's a pattern that's used everywhere. Also the IDE / tidy will propose to remove these redundant types.

haroonq:
_>> we can easily break gcc compatibility because we only build with clang_

Yes, I completely understand and am willing to support it myself (to a certain extent).  As you can see, there really wasn't much work needed to get to compile with gcc.

This is also why I plan to wrap this support under an "experimental" flag (or some such thing) to make it clear that it's not officially supported by the filament team.

libs/gltfio/src/extended/TangentSpaceMeshWrapper.h
haroonq: This is a duplicate definition. The original is defned above on line 34.

libs/gltfio/CMakeLists.txt
haroonq: It's not removed; it's replaced with ${FILAMENT_WARNINGS} which also adds "-Wall -Werror".  

The intention is that I will provide a different set of compiler flags for gcc vs clang.  Moreover, for gcc, I don't want -Werror because otherwise I'd have to make several more changes.  This seemed like a reasonable compromise.

filament/src/fg/FrameGraphRenderPass.h
haroonq: This is the error we get with gcc:
error: member 'depth' with constructor not allowed in anonymous aggregate

filament/src/ds/SsrPassDescriptorSet.cpp
haroonq: This is the error I get with gcc:
cannot convert ‘\<brace-enclosed initializer list\>’ to ‘filament::backend::SamplerParams’

I understand it's a pattern that is used and code is likely to break.  As mentioned above, I'm willing to fix these issues as they arise (albeit probably a bit delayed since we won't be updating our integration with filament as fast as you're making these changes).

I don't expect an IDE to remove these types.

filament/src/fg/FrameGraphRenderPass.h
haroonq: Done.

filament/src/fg/FrameGraphRenderPass.h
haroonq: I've marked the function as `noexcept`.  I've not added a `const` version as it was currently unneeded, but happy to do so if you think it's better.

poweifeng (approved):

filament/backend/src/vulkan/VulkanPipelineCache.cpp
```diff
@@ -267,8 +267,8 @@ void VulkanPipelineCache::bindVertexArray(VkVertexInputAttributeDescription cons
             mPipelineRequirements.vertexAttributes[i] = attribDesc[i];
             mPipelineRequirements.vertexBuffers[i] = bufferDesc[i];
         } else {
-            mPipelineRequirements.vertexAttributes[i] = {};
-            mPipelineRequirements.vertexBuffers[i] = {};
+            mPipelineRequirements.vertexAttributes[i] = VertexInputAttributeDescription{};
+            mPipelineRequirements.vertexBuffers[i] = VertexInputBindingDescription{};
```
pixelflinger, line 271: what was the error for that one?

filament/src/fg/FrameGraphRenderPass.h
pixelflinger: ok, gcc only allows fields with trivial ctors in anonymous structs. what a drag.

pixelflinger (commented):
the `SamplerParams` really bothers me.  What was the error?

filament/backend/src/vulkan/VulkanPipelineCache.cpp
haroonq: ambiguous overload for ‘operator=’ (operand types are ‘`filament::backend::VulkanPipelineCache::VertexInputBindingDescription`’ and ‘`<brace-enclosed initializer list>`’)

See https://godbolt.org/z/63qP99bdq

haroonq:
Re `SamplerParams`.  This is the error: cannot convert ‘`<brace-enclosed initializer list>`’ to ‘`filament::backend::SamplerParams`’

See https://godbolt.org/z/M8q38d53o

This one is definitely odd.  It only happens as a combination of a member being both an `enum class` and packing it into a bit field.  According to [this](https://stackoverflow.com/questions/52194486/enum-bitfield-and-aggregate-initialization), it seems like an bug in gcc.

pixelflinger:
> Re `SamplerParams`. This is the error: cannot convert ‘`<brace-enclosed initializer list>`’ to ‘`filament::backend::SamplerParams`’
> 
> See https://godbolt.org/z/M8q38d53o
> 
> This one is definitely odd. It only happens as a combination of a member being both an `enum class` and packing it into a bit field. According to [this](https://stackoverflow.com/questions/52194486/enum-bitfield-and-aggregate-initialization), it seems like an bug in gcc.

ok, looks like a gcc bug.  hopefully they fix it, it makes the code ugly.  thanks. lgtm.

PR #9329, opened 2025-10-15
backend: add two command line arguments to backend test

poweifeng:
`--headless_only`
Headless-only mode indicates that we will only use headless
swapchains.  This is particularly useful if we want to test in
a CI (continuous integration) environment.

`--ci`
Having CONTINUOUS_INTEGRATION as an OS will allow us to have
CI-only temporary exceptions. Though it's expected that some
exceptions will be unfixable and will remain as workarounds.

### DIFF
--- a/.github/actions/get-commit-msg/action.yml
+++ b/.github/actions/get-commit-msg/action.yml
@@ -1,3 +1,5 @@
+# This action retrieves the latest commit message from a push or pull_request event
+# and makes it available as an output variable named 'msg'.
 name: 'Get commit message'
 outputs:
   msg:
@@ -8,6 +10,8 @@ runs:
     - name: Find commit message (on push)
       if: github.event_name == 'push'
       shell: bash
+      env:
+        COMMIT_MESSAGE: ${{ github.event.head_commit.message }}
       run: |
         AUTHOR_NAME="${{ github.event.head_commit.author.name }}"
         AUTHOR_EMAIL="${{ github.event.head_commit.author.email }}"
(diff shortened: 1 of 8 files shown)


filament/backend/test/Arguments.cpp
```diff
@@ -23,13 +23,17 @@
...
-Backend parseArgumentsForBackend(int argc, char* argv[]) {
-    Backend backend = Backend::OPENGL;
+TestArguments parseArguments(int argc, char* argv[]) {
+    TestArguments arguments = {};
+    arguments.backend = Backend::OPENGL;
+
     // The first colon in OPTSTR turns on silent error reporting. This is important, as the
     // arguments may also contain gtest parameters we don't know about.
-    static constexpr const char* OPTSTR = ":a:";
+    static constexpr const char* OPTSTR = ":a:c:";
```
z3moon: You may miss adding `h` here. Also `--ci` doesn't require a value. Maybe `ch` instead?

filament/backend/test/mac_runner.mm
```diff
@@ -32,29 +32,36 @@
...
+        }
+        if (self.backend == test::Backend::METAL) {
+            nativeView.ptr = (void*) view.layer;
+        }
+        if (self.backend == test::Backend::VULKAN) {
+            nativeView.ptr = (void*) view.layer;
+        }
+        if (self.backend == test::Backend::WEBGPU) {
+            nativeView.ptr = (void*) view.layer;
+        }
```
z3moon: nit: maybe 
```
if (self.backend == test::Backend::OPENGL) {
    ...
} else {
    ...
}
```

filament/backend/test/Arguments.cpp
poweifeng: done

filament/backend/test/mac_runner.mm
poweifeng: done. switched to switch

z3moon (approved):

pixelflinger (approved):

filament/backend/test/Arguments.cpp
```diff
@@ -23,13 +23,17 @@
...
+    TestArguments arguments = {};
+    arguments.backend = Backend::OPENGL;
+
     // The first colon in OPTSTR turns on silent error reporting. This is important, as the
     // arguments may also contain gtest parameters we don't know about.
-    static constexpr const char* OPTSTR = ":a:";
+    static constexpr const char* OPTSTR = ":a:kc";
     static const struct option OPTIONS[] = {
             { "api", required_argument, nullptr, 'a' },
+            { "headless_only", no_argument, nullptr, 'k' },
```
bejado, line 35: Just curious, why not use `h` for the flag?

poweifeng: Because h is usually for help.  And maybe it should be passed on to gtest.

PR #9330, opened 2025-10-16
github: fix get-commit-msg

poweifeng:
If we assign the message (which might contain quotes) to an
environment variable and then echo it, this should prevent the
problem of having a double quote (").


poweifeng:
> don't know what any of this does. I trust it's working :-)

good point, I've added a little documentation on the github action.

PR #9343, opened 2025-10-21
GL: Fix importing non srgb using the external image API

rafadevai:
When importing a non srgb AHB but setting the srgb as true, makes the eglCreateImageKHR function to fail because it tries to set the colorspace as srgb for non srgb formats.

Example:
createExternalImage(yuv_ahardwaredbuffer, /*sRGB=*/true);

The fix is to only make sure that flag is set to true if the corresponding filament format requires sRGB and false otherwise.


filament/backend/src/opengl/platforms/PlatformEGLAndroid.cpp
```diff
@@ -251,12 +251,13 @@ Platform::ExternalImageHandle PlatformEGLAndroid::createExternalImage(AHardwareB
...
         p->aHardwareBuffer = hardwareBuffer;
-        p->sRGB = sRGB;
         AHardwareBuffer_Desc hardwareBufferDescription = {};
         AHardwareBuffer_describe(hardwareBuffer, &hardwareBufferDescription);
         p->height = hardwareBufferDescription.height;
         p->width = hardwareBufferDescription.width;
-        auto textureFormat = mapToFilamentFormat(hardwareBufferDescription.format, sRGB);
+        auto const textureFormat = mapToFilamentFormat(hardwareBufferDescription.format, sRGB);
+        // Only set sRGB as true if the filament format requires it, otherwise the eglCreateImage might fail.
+        p->sRGB = textureFormat == TextureFormat::SRGB8 || textureFormat == TextureFormat::SRGB8_A8;
```
z3moon, line 428: it looks like there's no behavioral change in the new code. i.e., if `sRGB` is true, p->sRGB will be set to true, and false otherwise. How does this new code help?

pixelflinger: @z3moon ~I *think* that if the format was for example YUV and `sRGB` was `true`, then `textureFormat` wouldn't be a SRGB format and `p->sRGB` would stay `false`.~
I think @z3moon is right, because `mapToFilamentFormat` always returns a SRGB format if `sRGB` is `true`.

rafadevai: so right now the issue is that when sRGB = true and the AHB is a YUV format when PlatformEGLAndroid::setImage is called, this path is executed

    if (eglExternalImage->sRGB) {
        imageAttrs[attrIndex++] = EGL_GL_COLORSPACE;
        imageAttrs[attrIndex++] = EGL_GL_COLORSPACE_SRGB;
    }
    
    which then makes the eglCreateImageKHR fails,
    
    so what this change will do, if just make sure to not set eglExternalImage->sRGB=true when the format doesnt need SRGB

pixelflinger: understood, but both @z3moon and I don't understand how this code does anything different than before? if `sRGB` is `true`, then `p->sRGB` will be `true` because `mapToFilamentFormat` will return a SRGB format.

rafadevai: The current issue is when calling eglCreateImageKHR for YUV images and the colorspace is set to srgb, the call fails and you ended up seeing a black textures.

This happens when eglExternalImage->sRGB = true, even for YUV images, which is this PR is fixing.

So what it will change is that, for non sRGB images, it will sanitize the input and prevent from setting sRGB to true when its not valid.

z3moon: So you mean the code works differently with this new change?

`mapToFilamentForma` function always returns one of these two, `TextureFormat::SRGB8` or `TextureFormat::SRGB8_A8`, if the input argument `sRGB` is true, resulting in setting this `p->sRGB` to true regardless of what value the `hardwareBufferDescription.format` has. (in the same way, p->sRGB is always set to false if the input sRGB is false)

So this new code looks identical as the original one, `p->sRGB = sRGB;`

I think you may want to modify `mapToFilamentForma` function or somewhere else?

rafadevai: its different because before the value of sRGB could be set as true when its not valid, like

Before this PR
createExternalImage(YUV_AHB, true) then this will set p->sRGB = true, which is not correct, we should store it as false.

After this PR
createExternalImage(YUV_AHB, true)  then this will set p->sRGB = false because the format isnt TextureFormat::SRGB8 or TextureFormat::SRGB8_A8

The value of sRGB will affect what is happening later on when in PlatformEGLAndroid::setImage

if (eglExternalImage->sRGB) {  <--------- executing this code path will cause problem for YUV images
    imageAttrs[attrIndex++] = EGL_GL_COLORSPACE; 
    imageAttrs[attrIndex++] = EGL_GL_COLORSPACE_SRGB; 
}

z3moon: As I said, `mapToFilamentFormat(XXX, true)` will always returns either `SRGB8` or `SRGB8_A8` because the second parameter is true, which I think is weird. So you probably want to modify `mapToFilamentFormat` as well.

rafadevai: ahhh sorry, I didnt understood what you were saying, yeah that part also should be changed.

rafadevai: Done

pixelflinger (approved):

PR #9349, opened 2025-10-22
cleanup PlatformEGL swapchain

pixelflinger:
SwapChainEGL is now implemented more idiomatically. It has a constructor that does the initialization. This allows subclasses to be implemented properly and safely.

PlatformEGL::createSwapChain() now just constructs SwapChainEGL and does nothing else, which allows a subclass of PlatformEGL to "add" to the PlatformEGL swapchain, buy simply subclassing from it and constructing it normally.

The SwapChain doesn't need to have a vtable because polymorphism is handled by the concrete Platform implementation.

Here we also implement a no-op SwapChainEGLAndroid to show how easy it is to extend the swapchain. This functionality will be used soon.

### DIFF
--- a/filament/backend/include/backend/platforms/PlatformEGLAndroid.h
+++ b/filament/backend/include/backend/platforms/PlatformEGLAndroid.h
@@ -18,12 +18,15 @@
 #define TNT_FILAMENT_BACKEND_OPENGL_OPENGL_PLATFORM_EGL_ANDROID_H
 
 #include <backend/AcquiredImage.h>
+#include <backend/DriverEnums.h>
 #include <backend/Platform.h>
 #include <backend/platforms/OpenGLPlatform.h>
 #include <backend/platforms/PlatformEGL.h>
 
 #include <utils/android/PerformanceHintManager.h>
 #include <utils/compiler.h>
+
+#include <math/mat3.h>
 
 #include <chrono>
 
@@ -67,7 +70,7 @@ public:
      * @return `true` on success, `false` on failure. The default implementation
      *         returns `false`.
      */
-    bool convertSyncToFd(Platform::Sync* sync, int* fd) noexcept;
+    bool convertSyncToFd(Sync* sync, int* fd) noexcept;
 
 protected:
     // --------------------------------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


filament/backend/include/backend/platforms/PlatformEGLAndroid.h
```diff
@@ -140,12 +143,24 @@ class PlatformEGLAndroid : public PlatformEGL {
     bool setImage(ExternalImageEGLAndroid const* eglExternalImage,
             ExternalTexture* texture) noexcept;
 
-protected:
```
z3moon, line 143: Is this intended? It'll make the `makeCurrent` and `SwapChainEGLAndroid` below visible outside of this class.

filament/backend/src/opengl/platforms/PlatformEGL.cpp
```diff
@@ -849,10 +757,108 @@ bool PlatformEGL::checkIfMSAASwapChainSupported(uint32_t samples) const noexcept
...
+}
+
+PlatformEGL::SwapChainEGL::SwapChainEGL(PlatformEGL const& platform,
+        uint32_t const width, uint32_t const height, uint64_t flags) {
+    attribs = {
+        { EGL_WIDTH, EGLint(width) },
+        { EGL_HEIGHT, EGLint(height) },
+    };
+
+    if (platform.ext.egl.KHR_gl_colorspace) {
```
z3moon, line 815: nit: maybe `isSRGBSwapChainSupported()`? to match the other constructor's code.

filament/backend/src/opengl/platforms/PlatformEGL.cpp
```diff
@@ -849,10 +757,108 @@ bool PlatformEGL::checkIfMSAASwapChainSupported(uint32_t samples) const noexcept
...
+        if (flags & SWAP_CHAIN_CONFIG_SRGB_COLORSPACE) {
+            attribs[EGL_GL_COLORSPACE_KHR] = EGL_GL_COLORSPACE_SRGB_KHR;
+        }
+    } else {
+        flags &= ~SWAP_CHAIN_CONFIG_SRGB_COLORSPACE;
+    }
+
+    if (platform.ext.egl.EXT_protected_content) {
```
z3moon, line 823: nit: isProtectedContextSupported ?

filament/backend/include/backend/platforms/PlatformEGLAndroid.h
pixelflinger: not really, because there was another `protected` above.

filament/backend/src/opengl/platforms/PlatformEGL.cpp
pixelflinger: I didn't want to change the existing logic too much.

filament/backend/src/opengl/platforms/PlatformEGL.cpp
pixelflinger: I didn't want to change the existing logic too much.

PR #9354, opened 2025-10-23
new frame and compositor timing APIs for Platform

pixelflinger:


filament/backend/src/opengl/platforms/PlatformEGLAndroid.cpp
```diff
@@ -139,38 +131,19 @@ bool PlatformEGLAndroid::makeCurrent(ContextType const type,
...
-            int const err = pWindow->query(dsc->nativeWindow, NativeWindow::IS_VALID, &isValid);
-            if (UTILS_LIKELY(err >= 0)) { // in case the IS_VALID enum is not recognized
-                // query call succeeded
-                FILAMENT_CHECK_POSTCONDITION(isValid) << kNativeWindowInvalidMsg << dsc->sur;
-            }
-        }
+    // anw can be nullptr if we're using a pbuffer surface
+    if (dsc->nativeWindow) {
+        auto [err, valid] = NativeWindow::isValid(dsc->nativeWindow);
+        FILAMENT_CHECK_POSTCONDITION(err || valid) << kNativeWindowInvalidMsg << dsc->sur;
```
z3moon: it looks like `ANativeWindow_getBuffersDataSpace` may return a negative value. In this case, `err || valid` will be still true, resulting in evading this assertion.

filament/backend/include/backend/Platform.h
```diff
@@ -247,6 +287,60 @@ class UTILS_PUBLIC Platform {
...
+
+    /**
+     * Associate a generic frameId which must be monotonically increasing (albeit not strictly) with
+     * the next frame to be presented on the specified swapchain.
+     *
+     * This must be called from the backend thread.
+     *
+     * @param swapchain
+     * @param frameId
+     * @return
```
z3moon: nit: missing description about return?

filament/backend/src/opengl/platforms/PlatformEGLAndroid.cpp
```diff
@@ -254,6 +248,128 @@ PlatformEGLAndroid::ExternalImageEGLAndroid::~ExternalImageEGLAndroid() {
...
+bool PlatformEGLAndroid::queryFrameTimestamps(SwapChain const* swapchain, uint64_t frameId,
+        FrameTimestamps* outFrameTimestamps) const noexcept {
+
+    if (!swapchain) {
+        return false;
+    }
+
+    SwapChainEGLAndroid const* const sc = static_cast<SwapChainEGLAndroid const*>(swapchain);
+    uint64_t const hwFrameId = sc->getFrameId(frameId);
+    if (hwFrameId == std::numeric_limits<uint64_t>::max()) {
```
poweifeng, line 315: nit: would use a constant instead of `std::numeric_limits<uint64_t>::max()`

PR #9355, opened 2025-10-23
Do not process third_party CMakeLists if those targets have already been defined.

haroonq:


pixelflinger:
can you give more context/details?

haroonq:
Yes, of course, sorry.

Basically, we have an executable that depends on a few libraries including absl, sdl2, gtest (for testing) and filament (for rendering).  In our CMakeLists.txt file, we download those dependencies from github and add them to our project.

Because those targets (absl, gtest, etc.) are defined in their own CMakeLists.txt files, we get a conflict when filament also tries to defined those dependencies in the filament/third_party folder.

For SDL2, filament's CMakeLists.txt file provides this check

```
if (FILAMENT_SKIP_SDL2)
    return()
endif()
```

which we set, but nothing similar exists for those other libraries.  Rather than introduce more flags (once for each target), I thought it would be safer to just check to see if the target is already loaded.

z3moon (approved):
While this pattern is common for CMake ecosystem, I wonder if we should apply this to other third party libs as well.

In that case, we probably leave the responsibility for resolving build errors due to mispatched or incompatible library versions to users.

pixelflinger:
@haroonq there was a similar/related change recently here : https://github.com/google/filament/pull/9275

It added a build option `FILAMENT_USE_EXTERNAL_ABSL`. Should we do it this way for now? See my comments in that PR.

haroonq:
@pixelfinger I personally think it's a bit cleaner to use functions built-in to cmake [1] for checking if a target exists rather than proliferating the number of options exposed by filament.  But, either solution would work for us, so it's really your call.  I'm also very new to cmake so I'm not sure about the best practice.

[1] https://cmake.org/cmake/help/latest/command/if.html#target

pixelflinger:
> @pixelfinger I personally think it's a bit cleaner to use functions built-in to cmake [1] for checking if a target exists rather than proliferating the number of options exposed by filament. But, either solution would work for us, so it's really your call. I'm also very new to cmake so I'm not sure about the best practice.
> 
> [1] https://cmake.org/cmake/help/latest/command/if.html#target

What I'm worried about is to have Frankenstein versions of filament in the wild  and start getting bug reports that we have no way to reproduce.

So I wanted to make it very obvious and explicit for developers that they're building with untested versions of various libs. On our side, we should be doing a better job at updating our dependencies regularly.

wdyt?  I also don't have the perfect answer here.

haroonq:
(Sorry about the late reply.)

Making the opt-out explicit in order to better diagnose bug reports makes sense.  I've added options to benchmark and glibtest that follow the pattern in absl that you referenced.  PTAL.

pixelflinger (approved):

PR #9359, opened 2025-10-24
Shared context support for GLX

z3moon:


filament/backend/include/backend/platforms/OpenGLPlatform.h
```diff
@@ -430,7 +430,7 @@ class OpenGLPlatform : public Platform {
     virtual bool isExtraContextSupported() const noexcept;
 
     /**
-     * Creates an OpenGL context with the same configuration than the main context and makes it
+     * Creates an OpenGL context with the same configuration with the main context and makes it
```
bejado:
```suggestion
     * Creates an OpenGL context with the same configuration as the main context and makes it
```

nit: "as" is the clearest term to use here

bejado (approved):

filament/backend/src/opengl/platforms/PlatformGLX.cpp
```diff
@@ -230,11 +231,70 @@ Driver* PlatformGLX::createDriver(void* sharedGLContext,
...
+
+    g_glx.setCurrentContext(mGLXDisplay,
+            (GLXDrawable)None, (GLXDrawable)None, context);
+
+    std::lock_guard<std::shared_mutex> lock(mAdditionalContextsLock);
+    mAdditionalContexts[currentThreadId] = context;
+}
+
+void PlatformGLX::releaseContext() noexcept {
+    g_glx.setCurrentContext(mGLXDisplay,
```
poweifeng, line 279: I assume you can only have one mGLXDisplay but one context per thread?  If so, don't you only want to do this if the current context attached to the display is also the one you are releasing?

z3moon: Each shared context is attached to the current thread by calling setCurrentContext after it's created.
In the same way, each shared context needs to be detached first by this line (passing nullptr) before being destroyed.

poweifeng: So let's say we do the following: 

Thread1: createContext() -> new context A, current context of display is A
Thread2: createContext() -> new context B, current context of display is B

Thread1: destroyContext() -> destroy context A, current context of display is B, but is set to null.

But B is still really the context attached to the display.

Would this be a problem?

z3moon: The shared context is not a global state for the display, but something like a thread-local state.

Each shared context is bound only to the calling thread. So destroying context A in your scenario doesn't affect the context B for the display. Also an implicit rule for these interface methods is that a pair of `createContext` and `releaseContext` should be invoked from the same thread. So this shouldn't be a problem. I tested this on linux.

poweifeng: > The shared context is not a global state for the display, but something like a thread-local state.

If this is true, then yes sgtm.

PR #9360, opened 2025-10-24
WebGPU: use staging buffers for updateGPUBuffer

yrcloud:
Use staging buffers and MapAsync instead of WriteBuffer for CPU-to-GPU buffer update

### DIFF
--- a/filament/backend/src/webgpu/WebGPUBufferBase.cpp
+++ b/filament/backend/src/webgpu/WebGPUBufferBase.cpp
@@ -99,7 +99,7 @@ void WebGPUBufferBase::updateGPUBuffer(BufferDescriptor const& bufferDescriptor,
     // Make sure the padded memory is set to 0 to have deterministic behaviors
     if (remainder != 0) {
         uint8_t* paddingStart = static_cast<uint8_t*>(mappedRange) + bufferDescriptor.size;
-        std::memset(paddingStart, 0, FILAMENT_WEBGPU_BUFFER_SIZE_MODULUS - remainder);
+        memset(paddingStart, 0, FILAMENT_WEBGPU_BUFFER_SIZE_MODULUS - remainder);
     }
 
     stagingBuffer.Unmap();


filament/backend/src/webgpu/WebGPUBufferBase.cpp
```diff
@@ -78,24 +79,62 @@ void WebGPUBufferBase::updateGPUBuffer(BufferDescriptor const& bufferDescriptor,
...
-        queue.WriteBuffer(mBuffer, byteOffset + legalSize, &mRemainderChunk,
-                FILAMENT_WEBGPU_BUFFER_SIZE_MODULUS);
+    // create the staging buffer
+    wgpu::BufferDescriptor descriptor{
+        .label = "staging buffer",
+        .usage =  wgpu::BufferUsage::MapWrite | wgpu::BufferUsage::CopySrc,
+        .size = mBuffer.GetSize(),
+        .mappedAtCreation = true };
+    wgpu::Buffer stagingBuffer = mDevice.CreateBuffer(&descriptor);
+    std::cout << "Run Yu: creating a new staging buffer, with size of " << descriptor.size << std::endl;
```
poweifeng: remove debug line

filament/backend/src/webgpu/WebGPUBufferBase.cpp
```diff
@@ -28,6 +28,7 @@
 
 #include <cstdint>
 #include <cstring>
+#include <iostream>
```
poweifeng: remove if not needed

filament/backend/src/webgpu/WebGPUBufferBase.cpp
poweifeng: Can you refactor `wgpu::Queue` to use [WebGPUQueueManager](https://github.com/google/filament/blob/main/filament/backend/src/webgpu/WebGPUQueueManager.h#L53) instead?

filament/backend/src/webgpu/WebGPUBufferBase.cpp
```diff
@@ -78,24 +79,62 @@ void WebGPUBufferBase::updateGPUBuffer(BufferDescriptor const& bufferDescriptor,
...
-        std::memset(mRemainderChunk.data() + remainder, 0,
-                FILAMENT_WEBGPU_BUFFER_SIZE_MODULUS - remainder);
-
-        queue.WriteBuffer(mBuffer, byteOffset + legalSize, &mRemainderChunk,
-                FILAMENT_WEBGPU_BUFFER_SIZE_MODULUS);
+    // create the staging buffer
+    wgpu::BufferDescriptor descriptor{
+        .label = "staging buffer",
+        .usage =  wgpu::BufferUsage::MapWrite | wgpu::BufferUsage::CopySrc,
+        .size = mBuffer.GetSize(),
```
poweifeng: I think the staging buffer should be of the `bufferDescriptor.size`.  

So imagine buffer A is of size X.  Update is called with size Y and offset O, what we want is to 
``` 
update A starting at O and copy Y bytes
```
But here `mBuffer.GetSize() == X`.

filament/backend/src/webgpu/WebGPUBufferBase.cpp
```diff
@@ -78,24 +79,62 @@ void WebGPUBufferBase::updateGPUBuffer(BufferDescriptor const& bufferDescriptor,
...
+                    break;
+                case wgpu::QueueWorkDoneStatus::Error:
+                case wgpu::QueueWorkDoneStatus::CallbackCancelled:
+                    std::cout << "Run Yu: queue submission met error or cancellation" << std::endl;
+                    break;
+            }
+        });
+
+    while (!ready) {
+        mDevice.Tick();
```
poweifeng: I don't think this part is needed.  The copy will be part of the encoded command in the queue, and webgpu will take care of the synchronization in that case. Any commands that come after will be able to use the buffer.

filament/backend/src/webgpu/WebGPUBufferBase.cpp
```diff
@@ -78,24 +79,76 @@ void WebGPUBufferBase::updateGPUBuffer(BufferDescriptor const& bufferDescriptor,
...
-                FILAMENT_WEBGPU_BUFFER_SIZE_MODULUS - remainder);
-
-        queue.WriteBuffer(mBuffer, byteOffset + legalSize, &mRemainderChunk,
-                FILAMENT_WEBGPU_BUFFER_SIZE_MODULUS);
+    // Fetch a staging buffer from the pool or create a new one
+    std::unique_ptr<wgpu::Buffer> stagingBuffer;
+    if (!mStagingBuffers.empty()) {
+        stagingBuffer = std::move(mStagingBuffers.back());
+        mStagingBuffers.pop_back();
+        std::cout << "Run Yu: fetched a staging buffer." << std::endl;
```
z3moon: what does `Run Yu` mean? Also please use LOG(INFO) instead as it's the way Filament logs work.

filament/backend/src/webgpu/WebGPUBufferBase.cpp
```diff
@@ -78,24 +79,76 @@ void WebGPUBufferBase::updateGPUBuffer(BufferDescriptor const& bufferDescriptor,
...
+    } else {
+        wgpu::BufferDescriptor descriptor{ .label = "staging buffer",
+            .usage = wgpu::BufferUsage::MapWrite | wgpu::BufferUsage::CopySrc,
+            .size = mBuffer.GetSize(),
+            .mappedAtCreation = true };
+        std::cout << "Run Yu: creating a new staging buffer, with size of " << descriptor.size
```
z3moon: ditto

filament/backend/src/webgpu/WebGPUBufferBase.cpp
```diff
@@ -78,24 +79,76 @@ void WebGPUBufferBase::updateGPUBuffer(BufferDescriptor const& bufferDescriptor,
...
+                  << std::endl;
+        stagingBuffer = std::make_unique<wgpu::Buffer>(mDevice.CreateBuffer(&descriptor));
+    }
+
+    // Calculate some alignment related sizes
+    const size_t copySizeRemainder = bufferDescriptor.size % FILAMENT_WEBGPU_BUFFER_SIZE_MODULUS;
+    const size_t legalCopySize = bufferDescriptor.size - copySizeRemainder;
+    // WGPU requires the offset of GetMappedRange to be a multiple of 8
+    const int stagingOffsetShift = -(byteOffset % FILAMENT_WEBGPU_MAPPED_RANGE_OFFSET_MODULUS);
```
z3moon: Is this necessary? You could always just copy it to the beginning of the staging buffer, which I assume to be already 8-bytes aligned, and let the `CopyBufferToBuffer` call below to handle the offset like:

```
commandEncoder.CopyBufferToBuffer(*stagingBuffer, 0, mBuffer,
            byteOffset, ...
```

filament/backend/src/webgpu/WebGPUBufferBase.h
```diff
@@ -23,6 +23,7 @@
 
 #include <array>
 #include <cstdint>
+#include <queue>
```
z3moon: I don't see any use case of std::queue, but std::vector. did you mean `#include <vector>`?

filament/backend/src/webgpu/WebGPUBufferBase.cpp
yrcloud: Done

filament/backend/src/webgpu/WebGPUBufferBase.cpp
yrcloud: Done with the new commit.

filament/backend/src/webgpu/WebGPUBufferBase.cpp
yrcloud: Deleted.

filament/backend/src/webgpu/WebGPUBufferBase.cpp
yrcloud: Sorry it was a WIP commit, "Run Yu" is just my name to help me filter out the debug messages I need. 
They are deleted now.

filament/backend/src/webgpu/WebGPUBufferBase.cpp
yrcloud: Yep now the staging buffer is just as large as needed, instead of being the same size as the member buffer.

filament/backend/src/webgpu/WebGPUBufferBase.h
yrcloud: Deleted.

filament/backend/src/webgpu/WebGPUBufferBase.cpp
yrcloud: Now I'm using a `wgpu::Device` directly. I think both this and using WebGPUQueueManager would behave the same iiuc.

poweifeng: Please still use QueueManager.  The purpose is to have a central place where we manage queue submissions.  This way we're able to create a "fence" to signify when a submission is complete.

yrcloud: Done.

filament/backend/src/webgpu/WebGPUBufferBase.cpp
```diff
@@ -78,24 +78,35 @@ void WebGPUBufferBase::updateGPUBuffer(BufferDescriptor const& bufferDescriptor,
...
+        uint8_t* paddingStart = static_cast<uint8_t*>(mappedRange) + bufferDescriptor.size;
+        std::memset(paddingStart, 0, FILAMENT_WEBGPU_BUFFER_SIZE_MODULUS - remainder);
     }
+
+    stagingBuffer.Unmap();
+
+    // Copy the staging buffer contents to the destination buffer.
+    webGPUQueueManager.getCommandEncoder().CopyBufferToBuffer(stagingBuffer, 0, mBuffer, byteOffset,
+            stagingBufferSize);
+    webGPUQueueManager.flush();
```
poweifeng: Can you try it without flushing here.  I don't think it's necessary if the copy is already part of an encoder.

filament/backend/src/webgpu/WebGPUBufferBase.h
```diff
@@ -41,7 +40,8 @@ class WebGPUBufferBase /* intended to be extended */ {
      * happen after draw commands encoded in the encoder. Submitting any commands up to this point
      * ensures the calls happen in the expected sequence.
      */
-    void updateGPUBuffer(BufferDescriptor const&, uint32_t byteOffset, wgpu::Queue const&);
+    void updateGPUBuffer(BufferDescriptor const&, uint32_t byteOffset, wgpu::Device const& device,
+            WebGPUQueueManager& webGPUQueueManager);
```
poweifeng: You could pass by pointer `WebGPUQueueManager*` and forward declare the class.  That way your dependency on `WebGPUQueueManager.h` can be in cpp-only

poweifeng: Can this comment be addressed?

yrcloud: Yes I deleted the `flush()`. Thanks!

yrcloud: Sorry I mis-read the comments. I'll address it here.

filament/backend/src/webgpu/WebGPUBufferBase.cpp
yrcloud: Done.

filament/backend/src/webgpu/WebGPUBufferBase.h
yrcloud: Done.

filament/backend/src/webgpu/WebGPUBufferBase.cpp
```diff
@@ -78,24 +79,34 @@ void WebGPUBufferBase::updateGPUBuffer(BufferDescriptor const& bufferDescriptor,
...
-        memcpy(mRemainderChunk.data(), remainderStart, remainder);
-        // Pad the remainder with zeros to ensure deterministic behavior, though GPU shouldn't
-        // access this
-        std::memset(mRemainderChunk.data() + remainder, 0,
-                FILAMENT_WEBGPU_BUFFER_SIZE_MODULUS - remainder);
-
-        queue.WriteBuffer(mBuffer, byteOffset + legalSize, &mRemainderChunk,
-                FILAMENT_WEBGPU_BUFFER_SIZE_MODULUS);
+        uint8_t* paddingStart = static_cast<uint8_t*>(mappedRange) + bufferDescriptor.size;
+        std::memset(paddingStart, 0, FILAMENT_WEBGPU_BUFFER_SIZE_MODULUS - remainder);
```
z3moon: nit: please remove std as filament prefers C style.

yrcloud: Good catch. Done.

PR #9361, opened 2025-10-24
Metal: support MSAA SwapChain flag

bejado:
This adds Metal support for the `CONFIG_MSAA_4_SAMPLES` SwapChain flag. Currently, this flag should not be used when  rendering multiple Views into the SwapChain.

BUGS = [433327615]


filament/backend/src/metal/MetalDriver.mm
```diff
@@ -1821,25 +1820,35 @@
     auto [fragment, vertex] = functions.getRasterFunctions();
 
     // Pipeline state
+    std::vector<NSUInteger> sampleCounts;
```
pixelflinger: do we really need a std::vector here? bindPipeline is on the critical path, I'd rather avoid heap allocations

filament/backend/src/metal/MetalDriver.mm
```diff
@@ -1821,25 +1820,35 @@
...
+        sampleCounts.push_back(depthAttachment.getSampleCount());
     }
     MTLPixelFormat stencilPixelFormat = MTLPixelFormatInvalid;
     const auto& stencilAttachment = mContext->currentRenderTarget->getStencilAttachment();
     if (stencilAttachment) {
         stencilPixelFormat = stencilAttachment.getPixelFormat();
+        sampleCounts.push_back(stencilAttachment.getSampleCount());
         assert_invariant(isMetalFormatStencil(stencilPixelFormat));
     }
+    UTILS_UNUSED_IN_RELEASE bool allSampleCountsEqual =
```
poweifeng: I think you could do this without storing all the samples?  Would save some space and iterating through the vector.

filament/backend/test/test_MsaaSwapChain.cpp
```diff
@@ -0,0 +1,103 @@
+/*
+ * Copyright (C) 2022 The Android Open Source Project
```
poweifeng:
```suggestion
 * Copyright (C) 2025 The Android Open Source Project
```

filament/backend/src/metal/MetalDriver.mm
```diff
@@ -1149,8 +1149,7 @@
 }
 
 bool MetalDriver::isMSAASwapChainSupported(uint32_t) {
-    // TODO: support MSAA swapchain
-    return false;
+    return true;
```
z3moon, line 1153: Do we need to check for actual support of hardware or software capabilities?

filament/backend/src/metal/MetalHandles.mm
```diff
@@ -1076,105 +1150,102 @@ static void func(void* user) {
...
-            assert_invariant(descriptor.colorAttachments[i].loadAction != MTLLoadActionLoad);
-            assert_invariant(!defaultRenderTarget);
-
-            id<MTLTexture> sidecar = attachment.getMSAASidecarTexture();
-            assert_invariant(sidecar);
+            // This might happen if the user is rendering multiple views into a MSAA SwapChain,
+            // which is not supported. In that case, ignore the load action.
+            if (descriptor.colorAttachments[i].loadAction == MTLLoadActionLoad) {
+                descriptor.colorAttachments[i].loadAction = MTLLoadActionDontCare;
+            }
```
z3moon, line 1185: I'm not an expert on Metal, but it seems like the load action should be MTLLoadActionDontCare.

Maybe consider adding
```
assert_invariant(descriptor.colorAttachments[i].loadAction == MTLLoadActionDontCare);
```
in case more options are added to MTLLoadAction? (for depth and stencil attachments as well)

filament/backend/src/metal/MetalDriver.mm
bejado: Yup, good point.

filament/backend/src/metal/MetalDriver.mm
bejado: We don't need to, as all Metal hardware/software should support it.

filament/backend/src/metal/MetalHandles.mm
bejado: Hm, I'm not sure I follow. For MSAA rendering, there's two allowed load actions: "Clear", and "DontCare". "Load" is not allowed. `getLoadAction` will return one of those three, so here we're just checking if it's been set to "Load" and falling back to "DontCare".

z3moon: I see, I didn't know Clear is also allowed. Then leaving comments to clarify those two allowed options may be a good idea. Is there any chance more more enums are added to MTLLoadAction later?

bejado: I'll add some comments. I think it's unlikely that additional enums are added to `MTLLoadAction`. The load action is determined by `getLoadAction`, which only returns one of 3 load actions, so even if more were added those are still the only possibilities.

z3moon (approved):

PR #9362, opened 2025-10-24
Use the new setProducerThrottling API

pixelflinger:


filament/backend/src/opengl/platforms/PlatformEGLAndroid.cpp
```diff
@@ -114,9 +116,30 @@ PlatformEGLAndroid::PlatformEGLAndroid() noexcept
...
+            mHasProducerThrottlingControl = true;
+            LOG(INFO) << "Producer Throttling API available";
+        }
+    }
 }
 
 PlatformEGLAndroid::~PlatformEGLAndroid() noexcept {
+    // note: we don't need to dlclose() mNativeWindowLib here, because the library will be cleaned
+    // when the process ends and dlopen() are ref-counted. dclose() NDK documentation commands
+    // not to call dlclose().
```
poweifeng: ```dclose() NDK documentation commands not to call dlclose().``` needs clarification?

filament/backend/include/backend/platforms/PlatformEGLAndroid.h
```diff
@@ -203,7 +203,11 @@ class PlatformEGLAndroid : public PlatformEGL {
     using clock = std::chrono::high_resolution_clock;
     clock::time_point mStartTimeOfActualWork;
 
+    void* mNativeWindowLib = nullptr;
```
poweifeng: if dclose is not needed, can you make this scoped to the function?

PR #9370, opened 2025-10-27
webgpu: fix FrameScheduled callback and test

poweifeng:
Fix the callback by using a shared_ptr to store the callback. The same logic exists in all other backends.

Re-enable corresponding backend test for webgpu.

Remove viewport size check in beingRenderPass. There is no similar check in other backends.

### DIFF
--- a/filament/backend/src/webgpu/WebGPUSwapChain.cpp
+++ b/filament/backend/src/webgpu/WebGPUSwapChain.cpp
@@ -390,11 +390,11 @@ void WebGPUSwapChain::present(DriverBase& driver) {
     if (!isHeadless()) {
         mSurface.Present();
     }
-    if (frameScheduled.callback) {
-        driver.scheduleCallback(frameScheduled.handler,
-                [callback = std::move(frameScheduled.callback)]() {
+    if (mFrameScheduled.callback) {
+        driver.scheduleCallback(mFrameScheduled.handler,
+                [callback = mFrameScheduled.callback]() {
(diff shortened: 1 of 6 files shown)


filament/backend/src/webgpu/WebGPUSwapChain.cpp
```diff
@@ -390,11 +390,11 @@ void WebGPUSwapChain::present(DriverBase& driver) {
     if (!isHeadless()) {
         mSurface.Present();
     }
-    if (frameScheduled.callback) {
-        driver.scheduleCallback(frameScheduled.handler,
-                [callback = std::move(frameScheduled.callback)]() {
+    if (mFrameScheduled.callback) {
+        driver.scheduleCallback(mFrameScheduled.handler,
+                [callback = mFrameScheduled.callback]() {
```
z3moon, line 395: The change looks good. But, just curious, could you fill me in how the original code ends up a failure (or crash)?

poweifeng: note the `std::move(frameScheduled.callback)`, this would have cleared `mFrameScheduled.callback` after a single invocation.

z3moon: I see, and the same callback is supposed to be repeatedly used for `present`.

z3moon (approved):

pixelflinger (approved):

PR #9374, opened 2025-10-28
Move resolving #includes from MaterialBuilder to MaterialCompiler

trie94:
Move resolving #includes from MaterialBuilder to MaterialCompiler before parsing the material.

Some details:
- resolving #includes was happening after parsing, now moving before parsing. this is because we could offload this resolution at build time for RuntimeMaterialCompiler
- filamat::resolveIncludes and include related classes are moved to MaterialCompiler. Now MaterialBuilder always assumes the input is already inlined.
- resolveIncludes used to have an assumption where the given text was already a shader block. this assumption is now broken so it finds the line offset internally. thus the line offset field is removed from IncludeResult.

Follow up (in a separate PR):
We will add a resolveIncludes flag in the CommandlineConfig and allows matc to return unresolved mat to resolved mat.

### DIFF
--- a/tools/matc/src/matc/Includes.cpp
+++ b/tools/matc/src/matc/Includes.cpp
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-#include <filamat/Includes.h>
+#include "Includes.h"
 
 #include <utils/Log.h>
 #include <utils/compiler.h>
@@ -22,7 +22,7 @@
 
 #include <string>
 
-namespace filamat {
+namespace matc {
 
 static bool isWhitespace(char c) {
     return (c == ' ' || c == '\f' || c == '\n' || c == '\r' || c == '\t' || c == '\v');
@@ -308,4 +308,4 @@ std::vector<FoundInclude> parseForIncludes(const utils::CString& source) {
     return results;
 }
 
-} // namespace filamat
+} // namespace matc
--- a/tools/matc/src/matc/Includes.h
+++ b/tools/matc/src/matc/Includes.h
@@ -14,16 +14,16 @@
  * limitations under the License.
  */
 
-#ifndef TNT_FILAMAT_INCLUDES_H
-#define TNT_FILAMAT_INCLUDES_H
+#ifndef TNT_INCLUDES_H
+#define TNT_INCLUDES_H
 
-#include <filamat/IncludeCallback.h>
+#include "IncludeCallback.h"
 
-#include <utils/CString.h>
+#include "utils/CString.h"
 
 #include <vector>
 
-namespace filamat {
+namespace matc {
 
 struct ResolveOptions {
     // If true, insert #line directives before / after each include.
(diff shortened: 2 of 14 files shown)


tools/matc/src/matc/Includes.cpp
```diff
@@ -83,10 +83,16 @@ bool resolveIncludes(IncludeResult& root, IncludeCallback callback,
...
-    const size_t lineNumberOffset = root.lineNumberOffset;
     utils::CString& text = root.text;
 
     std::vector<FoundInclude> includes = parseForIncludes(text);
+    if (includes.empty()) {
+        // No more to resolve.
+        return true;
+    }
+
+    const size_t lineNumberOffset = includes[0].line;
```
z3moon, line 94: I see `parseForIncludes` could return an empty container. It needs a proper handling for this case.

tools/matc/src/matc/MaterialCompiler.cpp
```diff
@@ -69,16 +69,37 @@ bool MaterialCompiler::run(const matp::Config& config) {
...
+
+    if (!filamat::resolveIncludes(result, std::move(includer), options)) {
+        return false;
+    }
+
+    // Now that the buffer is mutated, we need to update the buffer pointer and the size
+    // before parsing.
+    size = result.text.size();
+    auto modifiedBuffer = std::make_unique<char[]>(size);
+    std::strcpy(modifiedBuffer.get(), result.text.c_str());
```
z3moon: strcpy is not safe, so maybe use memcpy or strncpy instead.

tools/matc/src/matc/Includes.cpp
```diff
@@ -169,12 +176,13 @@ bool resolveIncludes(IncludeResult& root, IncludeCallback callback,
...
 
-    // Add a line directive on the first line for the root include.
+    // Add a line directive on the previous line for the root include.
     if (options.insertLineDirectives && depth == 0) {
         utils::io::sstream lineDirective;
-        insertLineDirective(lineDirective, lineNumberOffset + 1, rootIncludeName);
+        insertLineDirective(lineDirective, lineNumberOffset, rootIncludeName);
         lineDirective << '\n';
-        text.insert(0, utils::CString(lineDirective.c_str()));
+        const size_t startPos = includes[0].startPosition;
```
z3moon, line 184: ditto

tools/matc/src/matc/MaterialCompiler.cpp
trie94: ah you are right, replace it with `strncpy`

tools/matc/src/matc/Includes.cpp
trie94: i am a bit confused by the recursion but i believe i am handling this correctly with this block. am i missing something?
```
    if (includes.empty()) {
        // No more to resolve.
        return true;
    }
```

z3moon: Oh, I think I missed it. Sorry.

poweifeng (approved):

tools/matc/src/matc/Includes.h
```diff
@@ -48,11 +48,11 @@ struct FoundInclude {
     utils::CString name;
     size_t startPosition;
     size_t length;
-    size_t line;        // the line number the include was found on (first line is 1)
+    size_t line;        // the line number the #include was found on.
```
bejado: Is the first line now 0? Maybe we should update the comment here

PR #9377, opened 2025-10-29
fix a possible deadlock on exit

pixelflinger:
The deadlock happened because FrameInfoManager needs to wait that all fences have signaled before it can be destroyed. 

The fix here is simply to destroy the fence without waiting (for safety we do wait after we destroy them).

This uncovered another problem where all backends didn't handle this  correctly.

We now explicitly handle this when destroying a fence: we make sure it unblocks all the waiters and returns an error.

### DIFF
--- a/filament/backend/src/opengl/OpenGLDriver.cpp
+++ b/filament/backend/src/opengl/OpenGLDriver.cpp
@@ -2287,9 +2287,14 @@ mat3f OpenGLDriver::getStreamTransformMatrix(Handle<HwStream> sh) {
 
 void OpenGLDriver::destroyFence(Handle<HwFence> fh) {
     if (fh) {
-        GLFence const* f = handle_cast<GLFence*>(fh);
+        GLFence const* const f = handle_cast<GLFence*>(fh);
         if (mPlatform.canCreateFence() || mContext.isES2()) {
             mPlatform.destroyFence(f->fence);
+        } else {
+            // signal waiters it's time to give-up
+            std::unique_lock const lock(f->state->lock);
+            f->state->status = FenceStatus::ERROR;
+            f->state->cond.notify_all();
         }
         destruct(fh, f);
     }
--- a/filament/backend/src/vulkan/VulkanAsyncHandles.h
+++ b/filament/backend/src/vulkan/VulkanAsyncHandles.h
@@ -56,6 +56,10 @@ struct VulkanCmdFence {
     FenceStatus wait(VkDevice device, uint64_t timeout,
         std::chrono::steady_clock::time_point until);
 
+    void cancel() {
+        setStatus(VK_ERROR_UNKNOWN);
+    }
+
 private:
     std::shared_mutex mLock; // NOLINT(*-include-cleaner)
     std::condition_variable_any mCond;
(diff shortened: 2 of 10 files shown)


filament/backend/src/opengl/OpenGLDriver.cpp
```diff
@@ -2290,6 +2290,13 @@ void OpenGLDriver::destroyFence(Handle<HwFence> fh) {
         GLFence const* f = handle_cast<GLFence*>(fh);
         if (mPlatform.canCreateFence() || mContext.isES2()) {
             mPlatform.destroyFence(f->fence);
+        } else {
+            if (f->state.use_count() > 1) {
```
z3moon: I think this line has a race condition with [this shared_ptr copy](https://github.com/google/filament/blob/379b63d11bc37a462bdc56a8a3bc3ff9df20d63b/filament/backend/src/opengl/OpenGLDriver.cpp#L2320), where `fenceWait` is called from the AsyncJob thread. E.g., this line could hit first prior to that shared_ptr copy, resulting in skipping this if block. So I think we can just remove `if (f->state.use_count() > 1)`.

There is another race condition with [this line](https://github.com/google/filament/blob/379b63d11bc37a462bdc56a8a3bc3ff9df20d63b/filament/backend/src/opengl/OpenGLDriver.cpp#L1812). Although both  are called from the engine thread, one could overwrite the other. (very unlikely, but theoretically possible). However, this rare overwriting doesn't affect the deadlock but just report a wrong result. So maybe it's okay to ignore this case.

My suggestion is to
1. either just remove `if (f->state.use_count() > 1)` to fix the first issue (skipping), embracing the second issue (overwriting)
2. or update the status only once by checking if it still has the default value (TIMEOUT_EXPIRED or a new one like NONE), which means it hasn't changed yet, to also handle the second issue. (including [this part](https://github.com/google/filament/blob/379b63d11bc37a462bdc56a8a3bc3ff9df20d63b/filament/backend/src/opengl/OpenGLDriver.cpp#L1812))

```
std::unique_lock const lock(f->state->lock);
if (f->state->status == FenceStatus::TIMEOUT_EXPIRED) {
    f->state->status = FenceStatus::ERROR;
    f->state->cond.notify_all();
}
```

pixelflinger: Good catch for (1).  For (2) I think I don't care if the status is one or the other; since we're canceling it doesn't matter if the waiter gets ERROR or SATISFIED, imho.

z3moon (approved):
I confirmed this fixes the issue.

filament/backend/src/vulkan/VulkanAsyncHandles.h
```diff
@@ -70,28 +74,43 @@ struct VulkanFence : public HwFence, fvkmemory::ThreadSafeResource {
...
+        }
+        state->status = FenceStatus::ERROR;
+        state->cond.notify_all();
+    }
+
 private:
     struct State {
         std::mutex lock;
         std::condition_variable cond;
+        FenceStatus status = FenceStatus::TIMEOUT_EXPIRED;
```
poweifeng: We don't really need the additional semantics of FenceStatus. It'll be easier to parse with a well-named boolean like `hasCancelled` IMO.

filament/backend/src/vulkan/VulkanDriver.cpp
```diff
@@ -1180,9 +1181,9 @@ FenceStatus VulkanDriver::fenceWait(FenceHandle const fh, uint64_t const timeout
         until = now + nanoseconds(timeout);
     }
 
-    std::shared_ptr const cmdfence = fence->wait(until);
-    if (!cmdfence) {
-        return FenceStatus::TIMEOUT_EXPIRED;
+    auto const [cmdfence, status] = fence->wait(until);
+    if (!cmdfence || status == FenceStatus::ERROR) {
+        return status;
```
poweifeng: could be
```
if (!cmdfence || wasCancelled) {
    return FenceStatus::ERROR;
}
```

PR #9378, opened 2025-10-29
Add display present time as well as compositor timings to FrameInfo

pixelflinger:
to reviewers: the best way to review this PR is one CL at a time.

### DIFF
--- a/filament/src/FrameInfo.cpp
+++ b/filament/src/FrameInfo.cpp
@@ -101,7 +101,7 @@ void FrameInfoManager::beginFrame(FSwapChain* swapChain, DriverApi& driver,
     }
 
     // create a new entry
-    auto& front = history.emplace_front(frameId);
+    FrameInfoImpl& front = history.emplace_front(frameId);
 
     // store the current time
     front.vsync = vsync;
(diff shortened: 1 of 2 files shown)


libs/utils/include/utils/MonotonicRingMap.h
```diff
@@ -0,0 +1,154 @@
...
+
+    /**
+     * Inserts a new key-value pair.
+     * The key must be greater than the key of the last inserted element.
+     * If the map is full, the oldest element is overwritten.
+     * @param key The key to insert.
+     * @param value The value to associate with the key.
+     */
+    UTILS_NOINLINE void insert(key_type key, mapped_type value) {
+        assert(empty() || key > back().first); // assert monotonic
```
z3moon, line 67: probably assert_invariant, including others in this file

filament/src/FrameInfo.cpp
```diff
@@ -249,22 +270,60 @@ FixedCapacityVector<Renderer::FrameInfo> FrameInfoManager::getFrameInfoHistory(
...
+    std::swap(mUserFrameHistory, result);
+}
+
+FixedCapacityVector<Renderer::FrameInfo> FrameInfoManager::getFrameInfoHistory(
+        size_t const historySize) const {
+    auto result = mUserFrameHistory;
+    if (result.capacity() >= historySize) {
+        result.resize(historySize);
+    }
     return result;
```
z3moon, line 340: nit: maybe explicit std::move here? some toolchians/platforms may not support RVO even in c++17 yet.

z3moon (approved):

filament/src/FrameInfo.cpp
pixelflinger: you should never do an explicit move on a return, because it can disable RVO actually.  I think RVO is mandatory in C++17, so we're going to assume good compilers!

z3moon: ah sorry, I meant NRVO, which is still optional in c++17. (most major compilers will do tho)
IIUC, NRVO still may not work in debug build. but probably fine.

PR #9385, opened 2025-10-30
vkWaitForFences() must be called with the read-lock held

pixelflinger:


poweifeng (approved):

filament/backend/src/vulkan/VulkanAsyncHandles.cpp
```diff
@@ -74,7 +72,8 @@ FenceStatus VulkanCmdFence::wait(VkDevice device, uint64_t const timeout,
     }
 
     if (status == VK_SUCCESS) {
-        std::lock_guard const l(mLock);
+        rl.unlock();
+        std::lock_guard const wl(mLock);
         mStatus = status;
         return FenceStatus::CONDITION_SATISFIED;
     }
```
z3moon, line 79: There might be a race condition here you may want to avoid.

e.g., if cancel() is called while the other thread is blocked in vkWaitForFences(), and it returns VK_SUCCESS, the cancel() call, which was waiting for this lock to be freed, may squeeze in between `rl.unlock();` and `std::lock_guard const wl(mLock);`. This wil ignore the cancellation request.

If the cancel() should be honored. I think this block should check if a cancellation was issued after the lock_guard, something like

```
        if (mCanceled) {
            return FenceStatus::ERROR;
        }
        mStatus = status;
        return FenceStatus::CONDITION_SATISFIED;
```

pixelflinger: yesh but I think it doesn't matter because we'll return CONDITION_SATISFIED, so it's one or the other code is correct.

PR #9388, opened 2025-10-31
Guard MonotonicRingMap.MonotonicityDeathTest with GTEST_HAS_DEATH_TEST

poweifeng:


libs/utils/test/test_MonotonicRingMap.cpp
```diff
@@ -127,7 +127,7 @@ TEST(MonotonicRingMap, FindAfterWrapAround) {
     ASSERT_NE(nullptr, map.find(80));
 }
 
-#if !defined(NDEBUG)
+#if !defined(NDEBUG) && defined(GTEST_HAS_DEATH_TEST)
```
z3moon, line 130: do we want to remove silent warnings? what is the context of this change?

pixelflinger (approved):

PR #9389, opened 2025-10-31
Check for null fences to avoid segfault in tests

anishmgoyal:
In some test environments, creating a sync or fence backed by an Android fence returns null. In order to avoid NPEs, cover this scenario with a null check.

### DIFF
--- a/filament/backend/src/opengl/platforms/PlatformEGLAndroid.cpp
+++ b/filament/backend/src/opengl/platforms/PlatformEGLAndroid.cpp
@@ -551,6 +551,8 @@ Platform::Sync* PlatformEGLAndroid::createSync() noexcept {
         if (sync == EGL_NO_SYNC_KHR) {
             LOG(ERROR) << "Failed to create sync: " << eglGetError();
         }
+    } else {
+        LOG(WARNING) << "Native fences not supported on this device.";
     }
     return new(std::nothrow) SyncEGLAndroid{ .sync = sync };
 }
@@ -559,13 +561,13 @@ bool PlatformEGLAndroid::convertSyncToFd(Sync* sync, int* fd) noexcept {
     assert_invariant(sync && fd);
 
     if (UTILS_UNLIKELY(!ext.egl.ANDROID_native_fence_sync)) {
+        LOG(WARNING) << "Native fences not supported, cannot convert to fd.";
         return false;
     }
 
     SyncEGLAndroid const& eglSync = static_cast<SyncEGLAndroid&>(*sync);
     if (eglSync.sync == EGL_NO_SYNC_KHR) {
-        LOG(ERROR) << "Cannot convert sync to fd, because the underlying fence "
-                      "does not exist - does this display support fences?";
+        LOG(ERROR) << "Invalid fence, cannot convert to fd.";
         return false;
     }
 


anishmgoyal:
@phoenixxxx @pixelflinger @poweifeng mind reviewing?

filament/backend/src/opengl/platforms/PlatformEGLAndroid.cpp
```diff
@@ -546,6 +546,9 @@ void PlatformEGLAndroid::destroyStream(Stream* stream) noexcept {
 
 Platform::Sync* PlatformEGLAndroid::createSync() noexcept {
     auto const sync = eglCreateSyncKHR(getEglDisplay(), EGL_SYNC_NATIVE_FENCE_ANDROID, nullptr);
+    if (sync == nullptr) {
+        LOG(ERROR) << "Failed to create sync: " << eglGetError();
+    }
     return new(std::nothrow) SyncEGLAndroid{ .sync = sync };
 }
```
z3moon, line 558: should this return SyncEGLAndroid anyway even for the error case? If Sync wouldn't work because of the null `sync`, I think we may as well return a nullptr and handle the exception case at the call site.

z3moon: I'm lacking contexts here, so my comment could be over engineering or doesn't fit the current design.

anishmgoyal: hmm - i think it breaks a few assumptions, as the null-ness of the handle is how we determine if it's been set at all, for example. that said, this will only be null in the case of cuttlefish

poweifeng (approved):

filament/backend/src/opengl/platforms/PlatformEGLAndroid.cpp
pixelflinger: Actually these are EGL extensions, and we should check the extension is here before using them. And we should have a `isSyncSupported()` which the filament side can query. So it avoids from using the feature if it's not supported.

anishmgoyal: ok i definitely should check that, and will add it. but, i'm not sure if this will actually fix the cuttlefish scenario, because the error being returned isn't that the extension isn't present - the error code is EGL_BAD_DISPLAY.

tl;dr, i think we should do both.

anishmgoyal:
@pixelflinger updated to add checks for the native sync fence extension where relevant. this now matches the pattern followed by createSync() in PlatformEGL.

note - since there currently isn't any platform-independent functionality provided by these syncs atm, we haven't been providing functions to check for platform compatibility. i'd be happy to follow-up with a proper cl to introduce something like that, but the main piece of functionality that one may or may not have is the ability to convert to an FD.

filament/backend/src/opengl/platforms/PlatformEGLAndroid.cpp
```diff
@@ -557,6 +563,12 @@ bool PlatformEGLAndroid::convertSyncToFd(Sync* sync, int* fd) noexcept {
     }
 
     SyncEGLAndroid const& eglSync = static_cast<SyncEGLAndroid&>(*sync);
+    if (eglSync.sync == EGL_NO_SYNC_KHR) {
+        LOG(ERROR) << "Cannot convert sync to fd, because the underlying fence "
+                      "does not exist - does this display support fences?";
```
pixelflinger: I would shorten this message. If `ANDROID_native_fence_sync` is supported, this should not fail here.
so I would do something like:
```
if (sync == NO_SYNC) {
 if (ANDROID_native_fence_sync)
   LOG(ERROR) << "invalid fence"
 else
   LOG(WARNING) << "fences not supported"
}
```

filament/backend/src/opengl/platforms/PlatformEGLAndroid.cpp
```diff
@@ -545,7 +545,13 @@ void PlatformEGLAndroid::destroyStream(Stream* stream) noexcept {
...
 
 Platform::Sync* PlatformEGLAndroid::createSync() noexcept {
-    auto const sync = eglCreateSyncKHR(getEglDisplay(), EGL_SYNC_NATIVE_FENCE_ANDROID, nullptr);
+    EGLSyncKHR sync = EGL_NO_SYNC_KHR;
+    if (UTILS_LIKELY(ext.egl.ANDROID_native_fence_sync)) {
+        sync = eglCreateSyncKHR(getEglDisplay(), EGL_SYNC_NATIVE_FENCE_ANDROID, nullptr);
+        if (sync == EGL_NO_SYNC_KHR) {
+            LOG(ERROR) << "Failed to create sync: " << eglGetError();
+        }
+    }
```
pixelflinger, line 556: maybe add `else LOG(WARNING)<<"not supported"`?

pixelflinger (approved):
LGTM but minor comments about the error logs.

filament/backend/src/opengl/platforms/PlatformEGLAndroid.cpp
anishmgoyal: done!

filament/backend/src/opengl/platforms/PlatformEGLAndroid.cpp
anishmgoyal: this can fail even if ANDROID_native_fence_sync is supported. while cuttlefish technically reports that the feature is available, the actual creation returns an error message. i can shorten the message, though.

anishmgoyal:
@pixelflinger updated log statements

PR #9393, opened 2025-10-31
use choreographer to access the system's expected present time

pixelflinger:


filament/backend/src/AndroidFrameCallback.cpp
```diff
@@ -0,0 +1,111 @@
...
+    if (__builtin_available(android 33, *)) {
+        // request exit
+        mExitRequested.store(true, std::memory_order_relaxed);
+        // wake the looper right away
+        ALooper_wake(mLooper);
+        // release our reference to the looper
+        ALooper_release(mLooper);
+        mLooper = nullptr;
+        // and make sure to wait for the thread to terminate
+        mLooperThread.join();
```
z3moon: You need to call `joinable` to see if `join` can be called safely. Otherwise it could crash.

filament/backend/src/AndroidFrameCallback.cpp
```diff
@@ -0,0 +1,111 @@
...
+        });
+    }
+}
+
+void AndroidFrameCallback::terminate() {
+    if (__builtin_available(android 33, *)) {
+        // request exit
+        mExitRequested.store(true, std::memory_order_relaxed);
+        // wake the looper right away
+        ALooper_wake(mLooper);
```
z3moon: this mLooper has a race issue with the thread. For example, if `init` and `terminate` are called consecutively in a short period of time,  ALooper_wake(nullptr) could be called resulting in a crash probably.

filament/backend/src/AndroidFrameCallback.cpp
```diff
@@ -0,0 +1,111 @@
...
+        // release our reference to the looper
+        ALooper_release(mLooper);
```
z3moon: you could move this to the end of the thread, which is safer.

filament/backend/include/backend/platforms/VulkanPlatform.h
```diff
@@ -464,6 +468,10 @@ class VulkanPlatform : public Platform, utils::PrivateImplementation<VulkanPlatf
             uint64_t flags) noexcept;
 
     friend struct VulkanPlatformPrivate;
+
+#ifdef __ANDROID__
+    AndroidFrameCallback mAndroidFrameCallback;
```
rafadevai: There's an VulkanPlatformAndroid class, seems like a better place to put this

filament/backend/src/AndroidFrameCallback.cpp
pixelflinger: Why would it not be joinable?  And even if I did call joinable() there would be a race no?

```
if (mThread.joinable()) {
  mThread.join(); // how do I know it's still joinable here?
} else {
   // and how do I know it didn't become joinable here?
}
```

filament/backend/include/backend/platforms/VulkanPlatform.h
pixelflinger: ah yes, I didn't know it existed.

filament/backend/src/AndroidFrameCallback.cpp
z3moon: You may ignore my comments if you think is very unlikely.  But technically it won't be joinable when either `terminate` is called without `init`, or `terminate` is called twice. So it's not idempotent.

filament/backend/include/backend/platforms/VulkanPlatform.h
pixelflinger: done

filament/backend/src/AndroidFrameCallback.cpp
pixelflinger: done

filament/backend/src/AndroidFrameCallback.cpp
pixelflinger: done

poweifeng (approved):

PR #9398, opened 2025-11-03
vk: refactor to use polymorphism for platform specialization

poweifeng:
We refactor VulkanPlatform so that getSwapchainInstanceExtensions() and createVkSurfaceKHR() are virtual functions that are implemented by each platform. This will enable the use of polymorphism for platform-dependent bits.

### DIFF
--- a/NEW_RELEASE_NOTES.md
+++ b/NEW_RELEASE_NOTES.md
@@ -6,9 +6,3 @@
 appropriate header in [RELEASE_NOTES.md](./RELEASE_NOTES.md).
 
 ## Release notes for next branch cut
-
-- Metal: Add support for the `SwapChain::CONFIG_MSAA_4_SAMPLES` flag.
-
-- third_party: Optionally add libwebp to build
-  - controlled by cmake flag FILAMENT_SUPPORTS_WEBP_TEXTURES, defaults to OFF
-  - actual webp texture support for libs/gltfio coming in subsequent change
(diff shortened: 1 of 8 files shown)


filament/backend/src/vulkan/platform/VulkanPlatformWindows.cpp
```diff
@@ -0,0 +1,73 @@
...
+    VulkanPlatform::ExtensionSet const ret = {
+        VK_KHR_WIN32_SURFACE_EXTENSION_NAME,
+    };
+    return ret;
+}
+
+VulkanPlatform::SurfaceBundle VulkanPlatformWindows::createVkSurfaceKHR(void* nativeWindow,
+        VkInstance instance, uint64_t flags) const noexcept {
+    VkSurfaceKHR surface;
+    VkExtent2D extent;
```
z3moon, line 49: `extent` is not initialized.

z3moon: Sorry, I should've added more details. I thought this `extent` needs to be populated with values. Is this method intentionally returning { 0, 0 } for the second element? Then maybe this `VkExtent2D extent;` can be removed.

pixelflinger (approved):

filament/backend/src/vulkan/platform/VulkanPlatformWindows.cpp
poweifeng: it's not needed for this implementation, but needed for others.

PR #9371, opened 2025-10-27
Add Optional Inclusion of libwebp in Third Party Libaries

benn-herrera:
# Add libwebp to third_party 
This PR broken out as 2 of 3 from #9363 per request by @pixelflinger 

* Addresses [Issue 7962](https://github.com/google/filament/issues/7962)
* Added third_party/libwebp at version 1.6.0
  * Vast majority of the file changes 
* Added integration in third_party/libwebp/tnt/
  * update_libwebp.sh (modified version of third_party/cgltf/tnt/update_cgltf.sh from #9369
  * CMakeLists.txt that sets build options and wraps main libwebp project cmake
* Added FILAMENT_SUPPORTS_WEBP_TEXTURES cmake build option
  * controls inclusion of third_party/libwebp/tnt
  * adds compile time macro definition FILAMENT_SUPPORTS_WEBP_TEXTURES when ON
  * defaults to OFF

PR #9363, opened 2025-10-25
bh/7962 feature webp

benn-herrera:
# Webp Texture Support
Addresses [Issue 7962](https://github.com/google/filament/issues/7962)

## Changes
* added third_party/libwebp
  * this is the vast majority of the changed file count
  * added tnt/ CMakeLists.txt, README, and update_libwebp.sh
* third_party/cgltf
  * converted README instructions to update_cgltf.sh utility script 
  * updated to 1.15 to get webp texture extension handling
* added third_party/models/DamagedHelmet/DamagedHelmetWebp.glb
  * images converted to webp, textures updated to use EXT_texture_webp extension
  * geometry draco compressed
  * modifications done via [gltf-transform](https://gltf-transform.dev/) CLI utility
* libs/gltfio
  * added WebpProvider and createWebpProvider
    * adapted from StbProvider with minimal changes (substituted webp equivalents for stbi info and decoding operations) 
  * added handling for cgltf webp_image in places where image and basisu_image were being referenced
  * added test case for loading webp version of damaged helmet
* propagated use of createWebpProvider and registration as image/webp handler to samples
* updated jvm and javascript bindings to include webp texture provider

Issue #7962, opened 2024-07-04
Update cgltf to add support for webp textures in gltf/glb scenes


hannojg:
**Is your feature request related to a problem? Please describe.**

Use webp for textures in gltf/glb scenes.

**Describe the solution you'd like**

In a recent PR cgltf added support for webp:

- https://github.com/jkuhlmann/cgltf/pull/253

I think all we'd need to do in filament is update cgltf and maybe add some small changes to filament to make it work?

**Describe alternatives you've considered**

/

**OS and backend**

its not, I think cgltf is used cross platform.

hannojg:
Note: I just realised that since the mentioned PR get merged no new release has been created. So we have to wait for a bit…

poweifeng:
I think this has been completed after @benn-herrera 's PRs

PR #9406, opened 2025-11-04
GLTFIO Webp Texture Support

benn-herrera:
# Webp Texture Support
* Addresses [Issue 7962](https://github.com/google/filament/issues/7962) (Webp Texture Support)
* This PR broken out as 3 of 3 from #9363 per request by @pixelflinger 
* Preceding PRs: #9369 #9371

## Changes
* cmake option FILAMENT_SUPPORTS_WEBP_TEXTURES now actually does all of the thing.  defaults to OFF
* libs/gltfio
  * added WebpProvider and createOptionalWebpProvider
    * adapted from StbProvider with minimal changes (substituted webp equivalents for stbi 
info and decoding operations) 
    * createOptionalWebpProvider() is always available regardless of FILAMENT_SUPPORTS_WEBP_TEXTURES
      * returns new texture provider when ON
      * returns nullptr when OFF 
  * added handling for cgltf webp_image in places where image and basisu_image were being referenced
  * added test case for loading webp version of damaged helmet (see below)
* propagated use of createOptionalWebpProvider and registration as image/webp handler to samples
* updated jvm and javascript bindings to include optional webp texture provider
* added third_party/models/DamagedHelmet/DamagedHelmetWebp.glb
  * images converted to webp, textures updated to use EXT_texture_webp extension
  * geometry draco compressed
  * modifications done via [gltf-transform](https://gltf-transform.dev/) CLI utility
  * used in updated gltfio_test

### DIFF
--- a/filament/CMakeLists.txt
+++ b/filament/CMakeLists.txt
@@ -199,6 +199,7 @@ set(PRIVATE_HDRS
         src/components/LightManager.h
         src/components/RenderableManager.h
         src/components/TransformManager.h
+        src/details/AsyncHelpers.h
         src/details/BufferAllocator.h
         src/details/BufferObject.h
         src/details/Camera.h
--- a/filament/backend/CMakeLists.txt
+++ b/filament/backend/CMakeLists.txt
@@ -28,6 +28,7 @@ set(SRCS
         src/BackendUtils.cpp
         src/BlobCacheKey.cpp
         src/Callable.cpp
+        src/CallbackHandler.cpp
         src/CallbackManager.cpp
         src/CircularBuffer.cpp
         src/CommandBufferQueue.cpp
--- a/filament/backend/include/backend/CallbackHandler.h
+++ b/filament/backend/include/backend/CallbackHandler.h
@@ -64,7 +64,7 @@ public:
     virtual void post(void* user, Callback callback) = 0;
 
 protected:
-    virtual ~CallbackHandler() = default;
+    virtual ~CallbackHandler();
 };
 
 } // namespace filament::backend
--- a/filament/backend/include/backend/DriverEnums.h
+++ b/filament/backend/include/backend/DriverEnums.h
@@ -1715,6 +1715,12 @@ using CompositorTiming = Platform::CompositorTiming;
 
 using AsynchronousMode = Platform::AsynchronousMode;
 
+using AsyncCallId = uint32_t;
+
+static constexpr AsyncCallId InvalidAsyncCallId = std::numeric_limits<AsyncCallId>::max();
+
+using AsynchronousMode = Platform::AsynchronousMode;
+
 } // namespace filament::backend
 
 template<> struct utils::EnableBitMaskOperators<filament::backend::ShaderStageFlags>
--- a/filament/backend/include/backend/platforms/PlatformWGL.h
+++ b/filament/backend/include/backend/platforms/PlatformWGL.h
@@ -64,7 +64,7 @@ protected:
     std::vector<int> mAttribs;
 
     // For shared contexts
-    static constexpr int SHARED_CONTEXT_NUM = 1;
+    static constexpr int SHARED_CONTEXT_NUM = 2;
     std::vector<HGLRC> mAdditionalContexts;
     std::atomic<int> mNextFreeSharedContextIndex{0};
 };
--- a/filament/backend/include/private/backend/Driver.h
+++ b/filament/backend/include/private/backend/Driver.h
@@ -70,6 +70,11 @@ public:
     // is where the driver can execute user callbacks.
     virtual void purge() noexcept = 0;
 
+    // Called from the engine thread (render-thread) to execute the `callback` via the `handler` if
+    // it is available. Otherwise, if `handler` is null, the `callback` is executed on the main
+    // thread via `purge()`.
+    virtual void scheduleCallback(CallbackHandler* handler, void* user, CallbackHandler::Callback callback) = 0;
+
     virtual ShaderModel getShaderModel() const noexcept = 0;
 
     // The shader languages used for shaders for this driver in order of preference, used to inform
--- a/filament/backend/include/private/backend/DriverAPI.inc
+++ b/filament/backend/include/private/backend/DriverAPI.inc
@@ -74,8 +74,9 @@
 #define APPLY10(M, A, ...) EXPAND(M(A)), EXPAND(APPLY9(M, __VA_ARGS__))
 #define APPLY11(M, A, ...) EXPAND(M(A)), EXPAND(APPLY10(M, __VA_ARGS__))
 #define APPLY12(M, A, ...) EXPAND(M(A)), EXPAND(APPLY11(M, __VA_ARGS__))
-#define APPLY_N__(_0, _1, _2, _3, _4, _5, _6, _7, _8, _9, _10, _11, _12, X, ...) APPLY##X
-#define APPLY(M, ...) EXPAND(EXPAND(APPLY_N__(M, __VA_ARGS__, 12, 11, 10, 9, 8, 7, 6, 5, 4, 3, 2, 1, 0))(M, __VA_ARGS__))
+#define APPLY13(M, A, ...) EXPAND(M(A)), EXPAND(APPLY12(M, __VA_ARGS__))
+#define APPLY_N__(_0, _1, _2, _3, _4, _5, _6, _7, _8, _9, _10, _11, _12, _13, X, ...) APPLY##X
+#define APPLY(M, ...) EXPAND(EXPAND(APPLY_N__(M, __VA_ARGS__, 13, 12, 11, 10, 9, 8, 7, 6, 5, 4, 3, 2, 1, 0))(M, __VA_ARGS__))
 
 
 #define PAIR_ARGS_0(M, ...)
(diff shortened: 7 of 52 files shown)


libs/gltfio/include/gltfio/TextureProvider.h
```diff
@@ -179,6 +179,13 @@ TextureProvider* createStbProvider(filament::Engine* engine);
  */
 TextureProvider* createKtx2Provider(filament::Engine* engine);
 
+/**
+ * If webp support is enabled at build time, creates a decoder that can handle "image/webp"
+ * lossless and lossy content.
+ * If webp support is not enabled at build time, returns nullptr.
+ */
+TextureProvider* createOptionalWebpProvider(filament::Engine* engine);
+
```
benn-herrera, line 188: @pixelflinger - is the behavior described here reasonable or is there another pattern that would be more consistent with the project?

docs/viewer/filament-viewer.js
```diff
@@ -307,7 +307,7 @@ class FilamentViewer extends LitElement {
                 asyncInterval: 30
             };
 
-            const doneAddingResources = (resourceLoader, stbProvider, ktx2Provider) => {
+            const doneAddingResources = (resourceLoader, stbProvider, ktx2Provider, webpProvider) => {
```
poweifeng: please don't change this file since this whole directory is auto updated.  
Changes should go to `docs_src/src_raw/viewer/filament-viewer.js`

CMakeLists.txt
```diff
@@ -58,7 +58,7 @@ option(FILAMENT_USE_ABSEIL_LOGGING "Use Abseil to log, may increase binary size"
 
 option(FILAMENT_ENABLE_EXPERIMENTAL_GCC_SUPPORT "Enable GCC support (unsupported)" OFF)
 
-option(FILAMENT_SUPPORTS_WEBP_TEXTURES "Enable webp texture support for Filament (builds libwebp when ON)" OFF)
+option(FILAMENT_SUPPORTS_WEBP_TEXTURES "Enable webp texture support for Filament (builds libwebp when ON)" ON)
```
poweifeng: I think we want this to default OFF?

NEW_RELEASE_NOTES.md
```diff
@@ -8,7 +8,7 @@ appropriate header in [RELEASE_NOTES.md](./RELEASE_NOTES.md).
...
 
 - Metal: Add support for the `SwapChain::CONFIG_MSAA_4_SAMPLES` flag.
-
-- third_party: Optionally add libwebp to build
-  - controlled by cmake flag FILAMENT_SUPPORTS_WEBP_TEXTURES, defaults to OFF
-  - actual webp texture support for libs/gltfio coming in subsequent change
+- gltfio: Add optional support for webp textures (EXT_texture_webp)
+  - Controlled via FILAMENT_SUPPORTS_WEBP_TEXTURES cmake option
+    - Defaults to ON
+  - When enabled also builds third_party/libwebp
```
poweifeng: Please keep this a single line description - doesn't have to have too much detail.

libs/gltfio/src/WebpProvider.cpp
```diff
@@ -0,0 +1,295 @@
...
+    for (auto& info : mTextures) {
+        if (info->state != TextureState::DECODING) {
+            continue;
+        }
+        // Deleting data here should be safe thread-wise as the only other place where
+        // decodedTexelsBaseMipmap is loaded is in the job threads, and we have waited them to
+        // completion above. We also expect the TextureProvider API calls to be made only from one
+        // thread.
+        if (intptr_t data = info->decodedTexelsBaseMipmap.load()) {
+            delete [] (uint8_t*) data;
```
z3moon: I think you should delete the data using `WebPFree`?

web/filament-js/jsbindings.cpp
```diff
@@ -2054,6 +2055,11 @@ class_<Ktx2Provider>("gltfio$Ktx2Provider")
         return Ktx2Provider { createKtx2Provider(engine) };
     }));
 
+class_<OptionalWebpProvider>("gltfio$WebpProvider")
```
z3moon: Maybe `gltfio$OptionalWebpProvider` instead?

libs/gltfio/include/gltfio/TextureProvider.h
```diff
@@ -179,6 +179,13 @@ TextureProvider* createStbProvider(filament::Engine* engine);
  */
 TextureProvider* createKtx2Provider(filament::Engine* engine);
 
+/**
+ * If webp support is enabled at build time, creates a decoder that can handle "image/webp"
+ * lossless and lossy content.
+ * If webp support is not enabled at build time, returns nullptr.
+ */
+TextureProvider* createOptionalWebpProvider(filament::Engine* engine);
```
bejado: Curious what others think, but I'm not sure we need "Optional" in the name here: maybe simply `createWebpProvider`. Alternatively, I'm also in favor of making this function alwasy return a valid `TextureProvider*` but guarding it with `FILAMENT_SUPPORTS_WEBP_TEXTURES`.

libs/gltfio/src/WebpProvider.cpp
benn-herrera: Good catch. Fixed.

NEW_RELEASE_NOTES.md
benn-herrera: Fixed.

CMakeLists.txt
benn-herrera: Fixed.

docs/viewer/filament-viewer.js
benn-herrera: Fixed.

benn-herrera:
> Can you also add to `third_party/models/DamagedHelmet/README.md` to indicate how the Webp version came about?

Fixed.

web/filament-js/jsbindings.cpp
benn-herrera: Good catch. This was a mismatch with the usage
``` const webpProvider = new Filament.gltfio$OptionalWebpProvider(this.engine);```
from filament-viewer.js

Fixed.

File: libs/gltfio/include/gltfio/TextureProvider.h (hunk 7 above)
Comment (benn-herrera) [reply libs/gltfio/include/gltfio/TextureProvider.h]: This touches on my biggest questions:
* Naming to communicate that this option might not have been enabled at build time
* Behavior and correct API consumption pattern when the option was not enabled at build time.

Let me know what the consensus is and I'll do that.

Comment (z3moon) [reply libs/gltfio/include/gltfio/TextureProvider.h]: I'm also leaning towards removing Optional. It might be more cumbersome to rename the function rather than remove the defines when this support become permanent later.

Comment (benn-herrera) [reply libs/gltfio/include/gltfio/TextureProvider.h]: @bejado, @z3moon - renamed functions and structs to eliminate 'Optional'

**Review by bejado — APPROVED** [review]

File: android/gltfio-android/CMakeLists.txt [hunk 8]
```diff
@@ -102,8 +113,12 @@ target_link_libraries(gltfio-jni
         PRIVATE basis_transcoder
         PRIVATE uberarchive
         PRIVATE dracodec
+        PRIVATE webpdecoder
```
poweifeng: I think you'd want to remove this.  Might be why the android build is failing.

benn-herrera: Looking at that now. I'll need to conditionalize the dependency on enablement.

benn-herrera: Ah - already have the conditional dependency. Needed to just remove the straggler (as. you indicated).
Fixed.

libs/gltfio/include/gltfio/TextureProvider.h
pixelflinger: if the function is not enabled at build time, maybe it shouldn't be there at all? Can we guard the whole function? What do people think?

benn-herrera: The problem with guarding the whole function is that the bound functions in dynamic languages can't be guarded in the same way at the points of usage. Also it would impose guarding use of the function on consumers of the C++ API.

The deeper issue is that API consumers have to use the texture provider mechanism for standard types in the first place. The supported image types should work out of the box and the texture provider mechanism should just be a way of overriding or adding support for custom image types.

That seems like it would have been much too big a change for this PR, however. The current implementation feels like a sane middle ground.

libs/gltfio/include/gltfio/TextureProvider.h
pixelflinger: if it's going to return null, then we need a `isWebpSupported()` call, so  users of the API can know.

benn-herrera: Makes sense given the current pattern. 

That also means isWebpSupported() needs to be plumbed through to the JS and Java bindings and then used in those examples.

This adds to the argument for a subsequent PR eliminating the need for API consumers to plumb texture providers in the first place.

I don't see any good reason that GLTFIO shouldn't set that up automatically and allow users to provide additional non-standard texture providers at need (or override a default one)

benn-herrera: Added isWebpSupported(), bindings, testing, and usage thereof.

pixelflinger:
@benn-herrera we're almost there! Looks like the iOS build is failing and the PR needs a rebase.

benn-herrera:
> @benn-herrera we're almost there! Looks like the iOS build is failing and the PR needs a rebase.

on it.

EDIT: should be fixed with last commit. iOS builds for me locally.

benn-herrera:
Fixed a couple issues that only showed up when WebP is enabled.

pixelflinger:
@benn-herrera I merged the change!  Thank you for your contribution!

benn-herrera:
> @benn-herrera I merged the change! Thank you for your contribution!

💪
Glad we got it over the line.

PR #9410, opened 2025-11-06
vk: fix leaking swapchains again

### LINKED ISSUE #9403: GPU memory leak when frequently calling resetSwapChain() on Linux/Wayland/Vulkan/GLFW
Description:

Filament is running on Linux + Wayland + Vulkan + GLFW.

Here is the code I'm using:

void FilamentApp::Window::resetSwapChain()
{
    mFilamentApp->mEngine->destroy(mSwapChain);
    mSwapChain = mFilamentApp->mEngine
                     ->createSwapChain(getNativeWindow(mWindow),
                                       filament::SwapChain::CONFIG_HAS_STENCIL_BUFFER);
}

When I frequently call FilamentApp::Window::resetSwapChain(), the GPU memory usage increases drastically. After investigating, I found that the reason is that VulkanPlatformSwapChainBase::destroy() is not called—only the subclass's VulkanPlatformSurfaceSwapChain::destroy() is invoked. As a result, the VkImage objects in mSwapChainBundle are not released.

Question:
Why is VulkanPlatformSwapChainBase::destroy() declared as a pure virtual function? This seems to prevent proper cleanup of the swap chain resources when destroying swap chain objects.

poweifeng:
- Adjust order of destroy call in both headless and platform swapchains.  We need to be careful of the order due to assumptions made by the base class's destroy().
 - remove `=0` since VulkanPlatformSwapChainBase::destroy() has an implementation.

Fixes #9403

### DIFF
--- a/filament/backend/src/vulkan/platform/VulkanPlatformSwapChainImpl.cpp
+++ b/filament/backend/src/vulkan/platform/VulkanPlatformSwapChainImpl.cpp
@@ -408,7 +408,6 @@ VkResult VulkanPlatformSurfaceSwapChain::recreate() {
 }
 
 void VulkanPlatformSurfaceSwapChain::destroy() {
-    VulkanPlatformSwapChainBase::destroy();
     // The next part is not ideal. We don't have a good signal on when it's ok to destroy
     // a swapchain. This is a spec oversight and mentioned as much:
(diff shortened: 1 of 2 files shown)


bejado (approved):
Please also cherry-pick to rc/1.67.1

z3moon (approved):

pixelflinger (approved):

PR #9413, opened 2025-11-06
Fix a race condition when tearing down FrameInfo 

pixelflinger:
It is actually invalid to destroy a Handle<HwFence> while inside
fenceWait().

Updated the HwFence implementations so that they don't pretend they can
handle being destroyed during fenceWait(), they can't.

FrameInfo now cancels all the pending callbacks and waits for the 
currently executing one to terminate, *before* destroying the
handles.

Reenable the gpuFrameComplete metric, as it should be working now.

### DIFF
--- a/filament/backend/src/vulkan/VulkanAsyncHandles.h
+++ b/filament/backend/src/vulkan/VulkanAsyncHandles.h
@@ -77,46 +77,41 @@ struct VulkanFence : public HwFence, fvkmemory::ThreadSafeResource {
     VulkanFence() {}
 
     void setFence(std::shared_ptr<VulkanCmdFence> fence) {
-        std::lock_guard const lock(mState->lock);
-        mState->sharedFence = std::move(fence);
-        mState->cond.notify_all();
+        std::lock_guard const l(lock);
+        sharedFence = std::move(fence);
+        cond.notify_all();
     }
 
     std::shared_ptr<VulkanCmdFence>& getSharedFence() {
-        std::lock_guard const lock(mState->lock);
-        return mState->sharedFence;
+        std::lock_guard const l(lock);
+        return sharedFence;
     }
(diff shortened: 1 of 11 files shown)


filament/src/FrameInfo.h
```diff
@@ -130,6 +131,50 @@ class CircularQueue {
     using reference = value_type&;
     using const_reference = value_type const&;
 
+    class iterator {
```
bejado: Could we make the iterator a templated class based on the pointer type (either `T*` or `const T*`)? This way we don't have duplication between `iterator` and `const_iterator`:

```
template<typename Ptr>
class iterator {
    ...
    using pointer = Ptr;
    ...
};
```

filament/backend/src/vulkan/VulkanAsyncHandles.h
```diff
@@ -77,46 +77,41 @@ struct VulkanFence : public HwFence, fvkmemory::ThreadSafeResource {
...
-        return { state->sharedFence, state->canceled };
+        return { sharedFence, canceled };
     }
 
     void cancel() const {
-        std::shared_ptr const state{ mState };
-        std::unique_lock const lock(state->lock);
-        if (state->sharedFence) {
-            state->sharedFence->cancel();
+        std::unique_lock const l(lock);
```
z3moon: nit: maybe lock_guard would be preferrable for lightness and consistency

filament/src/FrameInfo.h
pixelflinger: done

filament/backend/src/vulkan/VulkanAsyncHandles.h
pixelflinger: good catch! done.

PR #9414, opened 2025-11-06
Move the include resolution functionality to matp

trie94:
- matp::MaterialParser now has a function resolveIncludes that returns a pair of status and resolve string.
- all the include resolution classes are moved to matp private src.
- matp::resolveIncludes is renamed to matp::resolveIncludesRecursively and only used internally.
- added insertLineDirectives and insertLineDirectiveChecks in Config; add those in the CommandlineConfig.
- moved output format to public use.

Note: MaterialParser::resolveIncludes could take a includer instead of the materialFilePath, but i decided to go with the materialFilePath because the most common use case is resolving from the file's directory. This allows the parser to just create the default DirIncluder internally, and we don't need to expose it publicly.

### DIFF
--- a/libs/filament-matp/include/filament-matp/MaterialParser.h
+++ b/libs/filament-matp/include/filament-matp/MaterialParser.h
@@ -42,7 +42,7 @@ public:
     // Returns the string of resolved material with the state that indicates whether the include
     // resolution was successful.
     std::pair<utils::Status, utils::CString> resolveIncludes(
-            std::unique_ptr<const char[]>& buffer, ssize_t size,
+            const std::unique_ptr<const char[]>& buffer, ssize_t size,
             const utils::Path& materialFilePath,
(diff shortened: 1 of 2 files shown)


libs/filament-matp/include/filament-matp/MaterialParser.h
```diff
@@ -36,7 +38,13 @@ class MaterialLexeme;
 class MaterialParser {
 public:
     MaterialParser();
-
+    // Resolves #include directives in the material by looking at the given file structure.
+    // Returns the string of resolved material with the state that indicates whether the include
+    // resolution was successful.
+    std::pair<utils::Status, utils::CString> resolveIncludes(
+            std::unique_ptr<const char[]>& buffer, ssize_t size,
```
z3moon: nit: probably adding `const` to the `buffer` variable

z3moon (approved):

Issue #9411, opened 2025-11-06
add loadTexture from ByteArrays in filament-utils


rma6:
The current implementation of loadTexture in filament.utils.TextureLoader only works for loading the bitmap from Resources.

I propose overloading the function to include an implementation that also loads the bitmap from a ByteArray.

This would be a Android only change AFAIK. I only have used filament for Android so far.

poweifeng:
This is a reasonable additional IMO.  Would you like to open a pull request yourself for this?

rma6:
Sure, I'll open a PR when I' back from work.

romainguy:
@pixelflinger The PR was merged, I think you can close this issue.

PR #9416, opened 2025-11-06
add overload of loadTexture that loads bitmap from ByteArray

rma6:
Implements Issue #9411 

Added an overload filament-utils.TextureLoader.loadTexture that uses the Android BitmapFactory.decodeByteArray to contruct a bitmap from a ByteArray.
The section of the code that uses the bitmap to build the texture was separated in a private function to avoid code duplicity, and making it easy to implement the other BitmapFactory factories if needed.

### DIFF
--- a/android/filament-utils-android/src/main/java/com/google/android/filament/utils/TextureLoader.kt
+++ b/android/filament-utils-android/src/main/java/com/google/android/filament/utils/TextureLoader.kt
@@ -40,7 +40,21 @@ fun loadTexture(engine: Engine, resources: Resources, resourceId: Int, type: Tex
     options.inPremultiplied = type == TextureType.COLOR
 
     val bitmap = BitmapFactory.decodeResource(resources, resourceId, options)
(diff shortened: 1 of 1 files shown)


android/filament-utils-android/src/main/java/com/google/android/filament/utils/TextureLoader.kt
```diff
@@ -40,7 +40,21 @@ fun loadTexture(engine: Engine, resources: Resources, resourceId: Int, type: Tex
     options.inPremultiplied = type == TextureType.COLOR
 
     val bitmap = BitmapFactory.decodeResource(resources, resourceId, options)
+    return buildTexture(engine, bitmap, type)
+}
+
+fun loadTexture(engine: Engine, bytes: ByteArray, type: TextureType, offset: Int = 0, length: Int = bytes.size): Texture {
```
romainguy, line 46: It would make more sense to expose the buildTexture() method that's below as it works with any Bitmap.

PR #9417, opened 2025-11-07
Properly use SDK 26+ APIs

pixelflinger:
With minSDK less than 26, it's unsafe to use APIs in nativewindow.h  that appeared at API 26 or more, even when using the weak symbol technique. This is because pre 26, libnativewindow.h didn't exist and therefore the build system cannot link against it.

Currently Filament's minSdk is 21, which means we should never use nativewindow.h API after 26.

In this PR we chose the strategy at compile time; either the more modern weak symbol or the old way with dlsym based on __ANDROID_API__.


filament/backend/src/AndroidNativeWindow.cpp
```diff
@@ -30,11 +31,16 @@
 namespace filament::backend {
 
 std::pair<int, bool> NativeWindow::isValid(ANativeWindow* const anw) noexcept {
+#if __ANDROID_API__ >= 26
+    // libnativewindow.so is not available before API level 26, this means we can't call
+    // any method above 25 (even protected by __builtin_available()).
     if (__builtin_available(android 28, *)) {
```
bejado, line 37: Should this still be `android 28`? Where does that come from?

bejado: Oh I see, the `ANativeWindow_getBuffersDataSpace` only exists in API level 28.

filament/backend/src/AndroidNdk.cpp
```diff
@@ -0,0 +1,60 @@
...
+
+#include <utils/compiler.h>
+
+#include <mutex>
+
+#include <dlfcn.h>
+
+namespace filament::backend {
+
+UTILS_UNUSED
```
bejado, line 29: Why `UTILS_UNUSED` here (and for `loadSymbol`?)

filament/backend/src/AndroidNativeWindow.cpp
pixelflinger: yes. This specific API is only available starting at 28. But we can only attempt to call it (with a 28 check) if *we* have a minSdk of at least 26.

In this particular case, if we're at minSdk < 26 (but the device could be like 33), I just use the private API call (I could have used the dlsym version too).  If we're >= 26, then I do the check with the builtin.

filament/backend/src/AndroidNdk.cpp
pixelflinger: that's for the case where we our minSdk>=26, in that case the code with loadSymbol is ifdef'ed out.

bejado: Ah I see. I thought we had `UTILS_MAYBE_UNUSED`, but looks like we just define that as `UTILS_UNUSED`.

bejado (approved):

PR #9419, opened 2025-11-07
add View::getLastDynamicResolutionScale()

pixelflinger:
This method return the last dynamic resolution scaling factor used by the view.

FIXES=[457753622]

### DIFF
--- a/android/filament-android/src/main/java/com/google/android/filament/View.java
+++ b/android/filament-android/src/main/java/com/google/android/filament/View.java
@@ -682,8 +682,8 @@ public class View {
      * @see Renderer#render(View)
      */
     @NonNull @Size(min = 2)
-    public float[] getLastDynamicResolutionScale(@Nullable @Size(min = 4) float[] out) {
-        out = Asserts.assertFloat4(out);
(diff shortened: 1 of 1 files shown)


android/filament-android/src/main/java/com/google/android/filament/View.java
```diff
@@ -674,6 +673,21 @@ public DynamicResolutionOptions getDynamicResolutionOptions() {
...
+     * Returns the last dynamic resolution scale factor used by this view. This value is updated
+     * when Renderer::render(View*) is called
+     * @param out A 2-float array where the value will be stored, or null in which case the array is
+     *            allocated.
+     * @return A 2-float array containing the horizontal and the vertical scale factors
+     * @see Renderer#render(View)
+     */
+    @NonNull @Size(min = 2)
+    public float[] getLastDynamicResolutionScale(@Nullable @Size(min = 4) float[] out) {
+        out = Asserts.assertFloat4(out);
```
z3moon: I think you meant `@Size(min = 2) float[] out` and `out = Asserts.assertFloat2(out);` for these two lines.

pixelflinger: fixed

z3moon (approved):

PR #9420, opened 2025-11-08
improve camera precondition checks and default state

pixelflinger:
- Camera is now initialized with a default projection matrix with left/right, top/bottom set to (-1,1) and near/far set to (0.1,1).

- setLensProjection() and setProjection(FOV) now check the near/far precondition (near>0 and far>near).

- sanitize user's light near/far to avoid infs and nans during froxelization

- finally froxelizer asserts that near>0 and far>near on debug builds.

FIXES=[458030186]

### DIFF
--- a/filament/src/Froxelizer.cpp
+++ b/filament/src/Froxelizer.cpp
@@ -396,10 +396,10 @@ bool Froxelizer::update() noexcept {
         if (zLightFar < zLightNear) {
             std::swap(zLightFar, zLightNear);
         }
-        if (zLightNear < mNear || zLightNear > mFar) {
+        if (zLightNear < mNear || zLightNear >= mFar) {
             zLightNear = mNear;
         }
-        if (zLightFar > mFar || zLightFar < mNear) {
+        if (zLightFar > mFar || zLightFar <= mNear) {
             zLightFar = mFar;
         }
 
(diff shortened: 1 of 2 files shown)


filament/src/Froxelizer.cpp
```diff
@@ -383,8 +385,28 @@ bool Froxelizer::update() noexcept {
...
+            std::swap(zLightFar, zLightNear);
+        }
+        if (zLightNear < mNear || zLightNear > mFar) {
+            zLightNear = mNear;
+        }
+        if (zLightFar > mFar || zLightFar < mNear) {
+            zLightFar = mFar;
+        }
+
+        assert_invariant(zLightNear < zLightFar);
```
z3moon, line 406: this line may result in a crash if `zLightFar == mNear`.

filament/src/details/Camera.cpp
```diff
@@ -84,19 +92,27 @@ mat4 FCamera::projection(double const focalLengthInMillimeters,
  * All methods for setting the projection funnel through here
  */
 
-void UTILS_NOINLINE FCamera::setCustomProjection(mat4 const& p,
-        mat4 const& c, double const near, double const far) noexcept {
+void UTILS_NOINLINE FCamera::setCustomProjection(mat4 const& projection,
+        mat4 const& projectionForCulling, double const near, double const far) noexcept {
+
+    FILAMENT_CHECK_PRECONDITION(near != far)
+            << "Camera preconditions not met in setCustomEyeProjection(): near = far = " << near;
```
z3moon: I think it meant `setCustomProjection`

filament/src/Froxelizer.cpp
pixelflinger: good catch! fixed.

filament/src/details/Camera.cpp
pixelflinger: fixed

z3moon (approved):

PR #9423, opened 2025-11-10
VK: Better support for renderdoc captures

rafadevai:
Doing a renderdoc capture when using transient memory or importing AHardwareBuffers into Vulkan, makes it fail, crash or not replay properly.

To improve the capture support, added a new constant to let the backend know when renderdoc is expected to be used and configure things appropietly.

- In the case of transient attachments that uses lazy allocated memory, it will be disabled.
- In the case of AHardwareBuffers, a more robust memory heap selection algorithm will be used.

Also a new VulkanPlatform function will be added, to let other subclasses know that the transient attachments are enabled or not.

### DIFF
--- a/filament/backend/include/backend/platforms/VulkanPlatform.h
+++ b/filament/backend/include/backend/platforms/VulkanPlatform.h
@@ -444,6 +444,11 @@ protected:
 
     virtual VkExternalFenceHandleTypeFlagBits getFenceExportFlags() const noexcept;
 
+    /**
+     * Query if transient attachments are supported by the backend.
+     */
+    bool isTransientAttachmentSupported() const noexcept;
+
 private:
     friend struct VulkanPlatformPrivate;
 };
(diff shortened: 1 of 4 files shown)


filament/backend/include/backend/platforms/VulkanPlatform.h
```diff
@@ -444,6 +444,11 @@ class VulkanPlatform : public Platform, utils::PrivateImplementation<VulkanPlatf
 
     virtual VkExternalFenceHandleTypeFlagBits getFenceExportFlags() const noexcept;
 
+    /**
+     * Query if transient attachments are supported by the backend.
+     */
+    bool isTransientAttachmentSupported() const noexcept;
```
poweifeng, line 450: can you elaborate on why is this method necessary?

poweifeng: resolved offline

poweifeng (approved):

PR #9425, opened 2025-11-11
Set the minimum required Metal version

emezeske:
Apple's documentation says that the language version will be set to the most recent version if it's not specified, but this is flat out false in some environments. On Apple Silicon I believe it is true, but I recently ported my app to run as a universal binary, and I have an Intel Mac where in some contexts, Filament's shader compilation fails with errors like:

```
Compilation failed:

program_source:113:41: error: 'id' attribute requires Metal language standard macos-metal2.0 or higher
constant FrameUniforms* frameUniforms [[id(0)]];
                                        ^~~~~

```

My app works when running as a standalone executable, but this issue comes up when running as a plugin (dynamically loaded object) inside another application. It appears that the host application can somehow cause the Metal API to assume a different Metal version than the latest.

Setting this explicitly to the minimum version that Filament requires fixes the problem. It also seems like a good idea, because it makes the version requirement explicit -- if a new Filament version is going to use newer Metal API features, it will be clear to everyone what version of Metal is being targeted.


filament/backend/src/metal/MetalShaderCompiler.mm
```diff
@@ -111,9 +111,14 @@ bool isReady() const noexcept {
...
             case ShaderLanguage::MSL: {
-                // By default, Metal uses the most recent language version.
                 MTLCompileOptions* options = [MTLCompileOptions new];
                 options.fastMathEnabled = YES;
+                // The docs for MTLCompileOptions say that Metal uses the most
+                // recent language version if it is not specified, but this is
+                // not always true: it is possible for the environment to be set
+                // up in a way where an older version is used by default. So we
+                // explicitly set the minimum version we require.
+                options.languageVersion = MTLLanguageVersion2_3;
```
bejado: I think it's good to be explicit with the language version here, but this should match what we pass to spirv-cross while compiling materials:

https://github.com/google/filament/blob/ad27d48fd309e1e8940e1a23a95ec992f07c481c/libs/filamat/src/GLSLPostProcessor.cpp#L359-L369

I think it's safe to go with MSL 2.0 on mobile, and MSL 2.3 on desktop. You can use a `#if defined(FILAMENT_IOS)` check here.

emezeske: Thanks. Sorry it took me so long to get back to this -- updated!

emezeske:
I think the latest version of this PR addresses feedback, ptal!  :D

bejado:
There were some build errors. Looks like we need to surround `MTLLanguageVersion2_3` with an `@available` check. Can you modify the code to the following?

```obj-c
                #if defined(FILAMENT_IOS)
                    options.languageVersion = MTLLanguageVersion2_0;
                #else
                    if (@available(macOS 11.0, *)) {
                        options.languageVersion = MTLLanguageVersion2_3;
                    } else {
                        options.languageVersion = MTLLanguageVersion2_2;
                    }
                #endif
```

emezeske:
Done, PTAL!

emezeske:
Thanks!

PR #9426, opened 2025-11-11
android: remove checked in asset

poweifeng:
For sample-gltf-viewer, an asset was checked into the source tree. But other assets are generated or copied from asset directories.

We remove the checked in gltf and copy the asset over during build (as with other existing files).

### DIFF
--- a/NEW_RELEASE_NOTES.md
+++ b/NEW_RELEASE_NOTES.md
@@ -8,3 +8,5 @@ appropriate header in [RELEASE_NOTES.md](./RELEASE_NOTES.md).
 ## Release notes for next branch cut
 
 - engine: add `View::getLastDynamicResolutionScale()` (b/457753622)
+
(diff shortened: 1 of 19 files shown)


z3moon (approved):

android/samples/sample-gltf-viewer/build.gradle
```diff
@@ -17,12 +17,20 @@ filamentTools {
 }
 
 // don't forget to update MainACtivity.kt when/if changing this.
-tasks.register('copyMesh', Copy) {
+tasks.register('copyDamagedHelmetGltf', Copy) {
+    from file("../../../third_party/models/DamagedHelmet/DamagedHelmet.glb")
```
pixelflinger, line 21: there is no better way than doing the ../../.. ?

poweifeng: it seems to be the convention in all samples' `build.gradle`.  I suppose we could do better with a variable to indicate the root directory, but that's a larger refactor.

PR #9429, opened 2025-11-12
vk: record timestmps on command buffers

poweifeng:
Timer queries are inserted into the beginning and the end of each command buffer. Refactor vk's implementation of Filament's timer queries so that they are now computed as a sum of the time taken by one or more command buffers (and not direct insertion of queries into the command stream).

This is to address a potential issue where the timestamps returned between command buffers are not consistent.


filament/backend/src/vulkan/VulkanAsyncHandles.cpp
```diff
@@ -90,4 +89,35 @@ void VulkanCmdFence::resetFence(VkDevice device) {
...
+        startState = mBeginState.get();
+        endState = mEndState.get();
+    }
+    // Here we sum up the time taken by each command buffer that were marked by this timer query.
+    uint64_t duration = 0;
+    do {
+        uint64_t const stDuration = startState->getBufferDuration();
+        assert_invariant(stDuration != VulkanCmdBufferState::UNKNOWN_BUFFER_DURATION);
+        duration += stDuration;
+        startState = startState->getNextState().get();
```
pixelflinger, line 113: isn't there a race here? what if setNextState() is called concurently?

filament/backend/src/vulkan/VulkanAsyncHandles.h
```diff
@@ -128,38 +156,50 @@ struct VulkanSync : fvkmemory::ThreadSafeResource, public HwSync {
...
 
-private:
-    uint32_t mStartingQueryIndex;
-    uint32_t mStoppingQueryIndex;
+    uint64_t getResult();
 
-    std::shared_ptr<VulkanCmdFence> mFence;
-    utils::Mutex mFenceMutex;
+private:
+    std::shared_mutex mMutex; // NOLINT(*-include-cleaner)
```
pixelflinger, line 198: I don't think we need a shared_mutex here because there is never a case where the "read lock" is held while blocking.

filament/backend/src/vulkan/VulkanAsyncHandles.h
```diff
@@ -128,38 +156,50 @@ struct VulkanSync : fvkmemory::ThreadSafeResource, public HwSync {
...
+    std::mutex mDurationLock;
```
pixelflinger, line 199: Also given when the contention looks like (small), I don't think we need two locks?

filament/backend/src/vulkan/VulkanAsyncHandles.cpp
```diff
@@ -90,4 +89,35 @@ void VulkanCmdFence::resetFence(VkDevice device) {
...
+    } while (startState != endState);
```
z3moon, line 114: Is it possible that `startState` and `endState` have the same value? (i.e., same buffer state)
In that case, this do-while block is still run regardless so `startState = startState->getNextState().get();` may result in a wrong behavior in the next frame.

filament/backend/src/vulkan/VulkanAsyncHandles.h
```diff
@@ -62,6 +74,18 @@ struct VulkanCmdFence {
         mCond.notify_all();
     }
 
+    void setNextState(std::shared_ptr<VulkanCmdBufferState> next) {
+        mNextState = std::move(next);
+    }
+
+    std::shared_ptr<VulkanCmdBufferState>& getNextState() {
```
bejado, line 81: Do you need to reference a reference here?

filament/backend/src/vulkan/VulkanAsyncHandles.h
bejado, line 198: If you do decide to go with two separate mutexes, maybe rename  `mMutex` to be clearer what it's guarding (like `mStateLock` to distinguish it from `mDurationLock`). That being said I think a single `std::atomic<uint64_t>` could be used in place of `mDurationLock` and `mDurationMutex`.

poweifeng:
turns out this is not required, but appreciate all the reviews.

PR #9430, opened 2025-11-12
Metal: recreate sidecar texture if sample count changes

bejado:
This fixes a crash when the "automagic" resolve sample count changes, for example, when VSM is used and the user adjusts the VSM sample count from 2 to 4.


pixelflinger (approved):

bejado:
The new tests fails on OpenGL and Vulkan. Will investigate

PR #9431, opened 2025-11-13
Fix several problems with RenderPass descriptor sets

pixelflinger:
The root of the problem is that in the main rendering loop we need to set the correct per-view descriptor-set based on the material and variants.

But we have two cases, either the descriptor set is always constant,  which is the case with ssr, structure, shadows and postfx passes, or it need to be dynamically changed based on the material & variant.

In the 2nd case, where was a problem where the postfx descriptor set could be used, which is wrong (e.g. while rendering shadows we need the shadow UBO, even with a postfx material), and would corrupt the correct descriptor set, which would never be set back.

Another issue was related to running a custom command, it could  change the state without updating the local copy, causing corruptions or validation errors.

In this CL we:
- use the same descriptorsetlayout for both postfx and depth
- invalidate the state after a custom command
- only switch to postfx descriptor set in dynamic mode (color pass)

This fixes setChannelDepthClearEnabled() which could cause  validation error, corruption or crashes.

FIXES=[459567258]

### DIFF
--- a/filament/src/details/Material.cpp
+++ b/filament/src/details/Material.cpp
@@ -217,16 +217,17 @@ void FMaterial::terminate(FEngine& engine) {
 
 filament::DescriptorSetLayout const& FMaterial::getPerViewDescriptorSetLayout(
         Variant const variant, bool const useVsmDescriptorSetLayout) const noexcept {
-    if (Variant::isValidDepthVariant(variant)) {
-        assert_invariant(mDefinition.materialDomain == MaterialDomain::SURFACE);
-        return mEngine.getPerViewDescriptorSetLayoutDepthVariant();
+    if (mDefinition.materialDomain == MaterialDomain::SURFACE) {
+        // `variant` is only sensical for MaterialDomain::SURFACE
+        if (Variant::isValidDepthVariant(variant)) {
+            return mEngine.getPerViewDescriptorSetLayoutDepthVariant();
+        }
+        if (Variant::isSSRVariant(variant)) {
+            return mEngine.getPerViewDescriptorSetLayoutSsrVariant();
+        }
     }
-    if (Variant::isSSRVariant(variant)) {
-        assert_invariant(mDefinition.materialDomain == MaterialDomain::SURFACE);
-        return mEngine.getPerViewDescriptorSetLayoutSsrVariant();
-    }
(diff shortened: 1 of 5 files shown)


poweifeng (approved):

filament/src/details/Material.cpp
```diff
@@ -217,16 +217,17 @@ void FMaterial::terminate(FEngine& engine) {
...
-        return mEngine.getPerViewDescriptorSetLayoutSsrVariant();
+    if (mDefinition.materialDomain == MaterialDomain::SURFACE) {
+        // `variant` is only sensical for MaterialDomain::SURFACE
+        if (Variant::isValidDepthVariant(variant)) {
+            return mEngine.getPerViewDescriptorSetLayoutDepthVariant();
+        }
+        if (Variant::isSSRVariant(variant)) {
+            return mEngine.getPerViewDescriptorSetLayoutSsrVariant();
+        }
     }
```
z3moon, line 228: This removes the assert_invariant checks, which means incorrect combination of variant and material will silently fall through and make an issue elsewhere. But please ignore my comments if you think it's okay.

filament/src/details/Material.cpp
```diff
@@ -217,16 +217,17 @@ void FMaterial::terminate(FEngine& engine) {
...
+    // mDefinition.perViewDescriptorSetLayout{Vsm} is already resolved for MaterialDomain
     if (useVsmDescriptorSetLayout) {
-        assert_invariant(mDefinition.materialDomain == MaterialDomain::SURFACE);
         return mDefinition.perViewDescriptorSetLayoutVsm;
     }
```
z3moon, line 232: Can this `if` block also be located inside the above `if (mDefinition.materialDomain == MaterialDomain::SURFACE)`?

filament/src/details/Material.cpp
pixelflinger: the assert was incorrect before because calling `isValidDepthVairant()` on a material domain that is not SURFACE is nonsensical, i.e. `isValidDepthVairant()` won't return the right value. And of course we can't make the assert before, because postfx material domains are allowed.

filament/src/details/Material.cpp
pixelflinger: Yes it could, but "logically' it shouldn't. good questions!

PR #9434, opened 2025-11-13
Log GPU errors through Platform debugStat API

bejado:
BUGS = [431665753]


filament/backend/include/backend/Platform.h
```diff
@@ -533,10 +539,26 @@ class UTILS_PUBLIC Platform {
...
+    void debugUpdateStat(const char* UTILS_NONNULL key, uint64_t intValue);
+
+    /**
+     * To track backend-specific statistics, the backend implementation can call the
+     * application-provided callback function debugUpdateStatFunc to associate or update a value
+     * with a given key. It is possible for this function to be called multiple times with the
+     * same key, in which case newer values should overwrite older values.
+     *
+     * This function is guaranteed to be called only on a single thread, the Filament driver
+     * thread.
```
pixelflinger, line 562: We should document in big capital letters that only short and non blocking work should be done in the callback and that GL calls are forbidden, because it's called from the backend thread.

filament/backend/src/metal/MetalDriver.mm
```diff
@@ -254,6 +254,16 @@
...
+
+    // Notify platform of GPU errors.
+    auto& platform = mPlatform;
+    mContext->commandBufferErrors.flush([&platform](NSError* error) {
+        if (UTILS_VERY_UNLIKELY(!error)) {
+            return;
+        }
+        const utils::CString errorString(error.localizedDescription.UTF8String);
+        platform.debugUpdateStat("filament.metal.command_buffer_error", errorString);
+    });
```
pixelflinger: could put all this inside a `if (UNLIKELY(!empty))`

filament/backend/src/metal/MetalErrorQueue.h
```diff
@@ -0,0 +1,63 @@
...
+#include <functional>
+#include <mutex>
+#include <vector>
+
+class MetalErrorQueue {
+public:
+    void push(NSError* error) {
+        std::lock_guard<std::mutex> lock(mMutex);
+        mErrors.push_back(error);
+        mHasErrors.store(true);
```
pixelflinger: `std::memory_order_relaxed` because you're actually not "publishing" the result -- you are, but the publishing part is done with the lock.

filament/backend/src/metal/MetalErrorQueue.h
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    void flush(const std::function<void(NSError*)>& callback) {
+        if (UTILS_LIKELY(!mHasErrors.load())) {
```
pixelflinger: `std::memory_order_relaxed`

filament/backend/src/metal/MetalErrorQueue.h
```diff
@@ -0,0 +1,63 @@
...
+            return;
+        }
+
+        std::vector<NSError*> errors;
+        {
+            std::lock_guard<std::mutex> lock(mMutex);
+            std::swap(mErrors, errors);
+            mHasErrors.store(false);
```
pixelflinger: `std::memory_order_relaxed`

filament/backend/src/Platform.cpp
```diff
@@ -172,9 +172,15 @@ bool Platform::hasDebugUpdateStatFunc() const noexcept {
     return bool(mDebugUpdateStat);
 }
 
-void Platform::debugUpdateStat(const char* key, uint64_t value) {
+void Platform::debugUpdateStat(const char* key, uint64_t intValue) {
     if (mDebugUpdateStat) {
```
pixelflinger: UNLIKELY?

filament/backend/src/Platform.cpp
```diff
@@ -172,9 +172,15 @@ bool Platform::hasDebugUpdateStatFunc() const noexcept {
...
-        mDebugUpdateStat(key, value);
+        mDebugUpdateStat(key, intValue, "");
+    }
+}
+
+void Platform::debugUpdateStat(const char* key, utils::CString stringValue) {
+    if (mDebugUpdateStat) {
```
pixelflinger: UNLIKELY?

filament/backend/src/Platform.cpp
```diff
@@ -172,9 +172,15 @@ bool Platform::hasDebugUpdateStatFunc() const noexcept {
...
+        mDebugUpdateStat(key, 0, stringValue);
```
pixelflinger: isn't there a race here with `setDebugUpdateStatFunc`?

pixelflinger (changes requested):

filament/backend/src/metal/MetalDriver.mm
bejado: I like that check to be inside `flush`, so that the caller doesn't have to remember to check first if it's empty.

filament/backend/src/Platform.cpp
bejado: I'm not sure it's likely or unlikely one way or another. Some clients will always have `mDebugUpdateStat` set.

filament/backend/src/Platform.cpp
bejado: Hm, yes it looks like there is. How does `setBlobFunc` avoid this?

filament/backend/src/metal/MetalDriver.mm
pixelflinger: Sure but you're systematically doing a function call for nothing -- it would be fine to do both, since the rare case is that the test will pass.

filament/backend/src/Platform.cpp
pixelflinger: looks like it has the race too :-)  The documentation does say "Cache functions may only be specified once during the lifetime of a Platform", which helps a bit. We could avoid the race without a lock using an atomic with acquires/release memory orders.

bejado: We can't use `std::atomic<RetrieveBlobFunc>`, for example, because `RetrieveBlobFunc` is not trivially-copiable. We could use a `std::mutex` to lock all of them, or maybe do `std::atomic<std::shared_ptr<FuncType>>`.

filament/backend/include/backend/Platform.h
```diff
@@ -564,6 +570,8 @@ class UTILS_PUBLIC Platform {
     InsertBlobFunc mInsertBlob;
     RetrieveBlobFunc mRetrieveBlob;
     DebugUpdateStatFunc mDebugUpdateStat;
+    mutable utils::Mutex mBlobFuncsMutex;
+    mutable utils::Mutex mDebugUpdateStatFuncMutex;
```
pixelflinger: I would actually use the same mutex for both, since it is expected to be very low contention.

filament/backend/src/Platform.cpp
```diff
@@ -139,46 +139,55 @@ bool Platform::queryFrameTimestamps(SwapChain const*, uint64_t, FrameTimestamps*
...
     return bool(mInsertBlob);
 }
 
 bool Platform::hasRetrieveBlobFunc() const noexcept {
+    std::lock_guard<decltype(mBlobFuncsMutex)> lock(mBlobFuncsMutex);
     return bool(mRetrieveBlob);
 }
 
 void Platform::insertBlob(void const* key, size_t keySize, void const* value, size_t valueSize) {
+    std::lock_guard<decltype(mBlobFuncsMutex)> lock(mBlobFuncsMutex);
```
pixelflinger: we can't call `mInsertBlob` with the lock held. We need to get the callback to a local with the lock held, release the lock and then call the callback.

bejado: > We need to get the callback to a local

Since `utils::Invocable` isn't copyable, the only way to do so is by wrapping the callbacks in a shared_ptr

pixelflinger: Unfortunate, but necessary. maybe we need to have a version of invocable that's copyable, or, bite the bullet and allow std::function<> in public APIs.

PR #9436, opened 2025-11-14
Simplify buffer object creation

z3moon:
Simplify the buffer object creation logic to streamline and help make the future integration of asynchronous features easier.


filament/src/details/VertexBuffer.cpp
```diff
@@ -283,55 +283,51 @@ FVertexBuffer::FVertexBuffer(FEngine& engine, const Builder& builder)
...
+    auto shouldCreateBuffer = [this](size_t attributeIndex) {
+        if (mBufferObjectsEnabled) {
+            if (mAdvancedSkinningEnabled) {
+                // For advanced skinning mode, only relevant buffers (BONE_INDICES & BONE_WEIGHTS) are
+                // created. We already manually handled the data for those specific buffers above.
+                if (attributeIndex != BONE_INDICES && attributeIndex != BONE_WEIGHTS) {
+                    return false;
+                }
+            } else {
+                return false;
```
z3moon: Although this is behaviorally identical to the original code, I'm not sure why we don't create a buffer in this case. @pixelflinger could you double check?

filament/src/details/VertexBuffer.cpp
```diff
@@ -283,55 +283,49 @@ FVertexBuffer::FVertexBuffer(FEngine& engine, const Builder& builder)
     mHandle = driver.createVertexBuffer(mVertexCount, mVertexBufferInfoHandle,
             utils::ImmutableCString{ builder.getName() });
 
+    auto shouldCreateBuffer = [this](size_t attributeIndex) {
+        if (mBufferObjectsEnabled) {
+            if (!mAdvancedSkinningEnabled) {
+                return false;
```
z3moon: Although this is behaviorally identical to the original code, I'm not sure why we don't create a buffer in this case. @pixelflinger  what do you think?

filament/src/details/VertexBuffer.cpp
```diff
@@ -283,55 +283,49 @@ FVertexBuffer::FVertexBuffer(FEngine& engine, const Builder& builder)
     mHandle = driver.createVertexBuffer(mVertexCount, mVertexBufferInfoHandle,
             utils::ImmutableCString{ builder.getName() });
 
+    auto shouldCreateBuffer = [this](size_t attributeIndex) {
+        if (mBufferObjectsEnabled) {
+            if (!mAdvancedSkinningEnabled) {
```
pixelflinger: Looks like this case can be entirely taken outside the loop. I think that will make it even easier to read.

filament/src/details/VertexBuffer.cpp
pixelflinger: I think it's because if `mBufferObjectsEnabled` is `true`, *we* don't manage the buffers, the user does. But there is an exception to that: If `mAdvancedSkinningEnabled` is `true`, then *we* need to manage the skinning/morphing buffers.

filament/src/details/VertexBuffer.cpp
```diff
@@ -283,55 +283,49 @@ FVertexBuffer::FVertexBuffer(FEngine& engine, const Builder& builder)
...
-                if (!mBufferObjects[i]) {
-                    BufferObjectHandle const bo = driver.createBufferObject(bufferSizes[i],
-                            BufferObjectBinding::VERTEX, BufferUsage::STATIC,
-                            utils::ImmutableCString{ builder.getName() });
-                    driver.setVertexBufferObject(mHandle, i, bo);
-                    mBufferObjects[i] = bo;
-                }
-            }
+    // create buffers
+    for (size_t i = 0; i < MAX_VERTEX_BUFFER_COUNT; ++i) {
```
pixelflinger, line 318: Could we not fold the 2nd loop into the first one now?

filament/src/details/VertexBuffer.cpp
z3moon: So you mean `if (!mAdvancedSkinningEnabled)` always takes precedence over mBufferObjectsEnabled? It's a different behavior through, let me update this.

filament/src/details/VertexBuffer.cpp
z3moon: Yeah, the final `bufferSizes` should be calculated in the 1st loop prior to buffer creation (2nd loop).

filament/src/details/VertexBuffer.cpp
```diff
@@ -283,55 +283,46 @@ FVertexBuffer::FVertexBuffer(FEngine& engine, const Builder& builder)
     mHandle = driver.createVertexBuffer(mVertexCount, mVertexBufferInfoHandle,
             utils::ImmutableCString{ builder.getName() });
 
+    auto shouldCreateBuffer = [this](size_t attributeIndex) {
+        if (mAdvancedSkinningEnabled) {
```
poweifeng: this doesn't seem like it's equivalent to the previous?  Shouldn't it be
```
if (mBufferObjectsEnabled && mAdvancedSkinningEnabled) 
```

z3moon: It was my original behavior. But @pixelflinger seems to think this can take precedence (in the other comments)

pixelflinger: no sorry, I meant that mAdvancedSkinningEnabled and mBufferObjectsEnabled are constants through the loop that calls shouldCreateBuffer() so we should be able to skip the whole outer loop.

pixelflinger (changes requested):

filament/src/details/VertexBuffer.cpp
z3moon: updated

filament/src/details/VertexBuffer.cpp
```diff
@@ -285,53 +285,45 @@ FVertexBuffer::FVertexBuffer(FEngine& engine, const Builder& builder)
...
-            const size_t end = offset + mVertexCount * stride;
-            if (slot != Attribute::BUFFER_UNUSED) {
-                assert_invariant(slot < MAX_VERTEX_BUFFER_COUNT);
-                bufferSizes[slot] = std::max(bufferSizes[slot], end);
-            }
-        }
-    }
+
+    auto shouldCreateBuffer = [this](size_t attributeIndex) {
+        uint8_t slot = mAttributes[attributeIndex].buffer;
```
poweifeng: nit: const

poweifeng (approved):

PR #9438, opened 2025-11-17
buffer update opt: Some optimizations

show50726:
### Changes
- Early return in `allocateOnDemand` when previous allocation has already failed
- Core frame functions now remove large instance map argument and the logic in each function becomes simpler. 
- Instead, `UboManager` now internally manages material instances which have batching enabled
- `allocateOnDemand` uses the new internal list
- Add tracing calls monitor performance of memory allocation

I'd like to know your thoughts about managing a separate list for batching-enabled material instances. 

The benefit of this is that the logic inside functions become clearer. For `allocateOnDemand`, the logic becomes simpler and potentially more efficient as well since it only traverses required material instances. The downside of this is that it requires additional memory to maintain a separate list.


pixelflinger:
> I'd like to know your thoughts about managing a separate list for batching-enabled material instances.
> 
> The benefit of this is that the logic inside functions become clearer. For `allocateOnDemand`, the logic becomes simpler and potentially more efficient as well since it only traverses required material instances. The downside of this is that it requires additional memory to maintain a separate list.

That seems like a good change. I would assume that adding/removing materialinstance is a low frequency operation. the extra memory seems okay, but we need to make sure those lists are updated when a MaterialInstance is destroyed.

filament/src/details/MaterialInstance.cpp
```diff
@@ -167,6 +168,7 @@ FMaterialInstance::FMaterialInstance(FEngine& engine,
 
     if (mUseUboBatching) {
         mUboData = BufferAllocator::UNALLOCATED;
+        engine.getUboManager()->manageMaterialInstance(this);
```
pixelflinger, line 171: don't we need a way to "unmanage" the MaterialInstance when it's destroyed?

filament/src/details/UboManager.cpp
```diff
@@ -135,25 +135,19 @@ void UboManager::beginFrame(DriverApi& driver,
...
     // Invalidate the migrated MIs, so that next commit() call must be triggered.
-    for (const auto& materialInstance : materialInstances) {
-        materialInstance.second.forEach([](const FMaterialInstance* mi) {
-            if (!mi->isUsingUboBatching()) {
-                return;
-            }
-
-            mi->getUniformBuffer().invalidate();
-        });
+    for (const auto& mi : mManagedInstances) {
```
pixelflinger: this should be a `*` here not a `&` (I think you're basically getting a reference to a pointer, which is less efficient).

filament/src/details/UboManager.cpp
```diff
@@ -164,24 +158,16 @@ void UboManager::finishBeginFrame(DriverApi& driver) {
...
-        materialInstance.second.forEach([&allocator, &allocationIds](const FMaterialInstance* mi) {
-            if (!mi->isUsingUboBatching()) {
-                return;
-            }
-
-            const AllocationId id = mi->getAllocationId();
-            if (!BufferAllocator::isValid(id)) {
-                return;
-            }
+    for (const auto& mi : mManagedInstances) {
```
pixelflinger: * instead of &

filament/src/details/UboManager.cpp
```diff
@@ -201,69 +187,76 @@ void UboManager::updateSlot(DriverApi& driver, AllocationId id,
...
 
-void UboManager::retireSlot(BufferAllocator::AllocationId id) {
+void UboManager::manageMaterialInstance(FMaterialInstance* instance) {
+    mPendingInstances.insert(instance);
+}
+
+void UboManager::unmanageMaterialInstance(const FMaterialInstance* materialInstance) {
+    AllocationId id = materialInstance->getAllocationId();
     if (!BufferAllocator::isValid(id))
         return;
```
pixelflinger: what does that mean here that the isValid() could be false? Shouldn't we remove the material instance from the list anyway? should this be an assert? or is this a valid case?

filament/src/details/UboManager.cpp
```diff
@@ -201,69 +187,76 @@ void UboManager::updateSlot(DriverApi& driver, AllocationId id,
...
-                existingInstances.push_back(mi);
-            } else {
-                newInstances.push_back(mi);
-            }
-        });
+    // const_cast should be safe here since this cast is just to match the container type.
+    auto mi = const_cast<FMaterialInstance*>(materialInstance);
+    if (UTILS_UNLIKELY(mPendingInstances.contains(mi))) {
+        mPendingInstances.erase(mi);
+    }
```
pixelflinger: do we need to do the `if`? is it not okay to just to erase(mi), if it's not contained, it won't erase it!

filament/src/details/UboManager.cpp
```diff
@@ -201,69 +187,76 @@ void UboManager::updateSlot(DriverApi& driver, AllocationId id,
...
+    FILAMENT_TRACING_CALL(FILAMENT_TRACING_CATEGORY_FILAMENT);
     bool reallocationNeeded = false;
 
     // Pass 1: Allocate slots for new material instances (that don't have a slot yet).
-    for (FMaterialInstance* mi : newInstances) {
+    for (FMaterialInstance* mi : mPendingInstances) {
+        mManagedInstances.insert(mi);
         auto [newId, newOffset] = mAllocator.allocate(mi->getUniformBuffer().getSize());
         mi->assignUboAllocation(mUbHandle, newId, newOffset);
         if (!BufferAllocator::isValid(newId)) {
```
pixelflinger, line 223: this deserves some comments. e.g. we assignUboAllocation, but just after we check isValid()? what if it is not valid, was it okay to do the assignment?

filament/src/details/UboManager.cpp
```diff
@@ -201,69 +187,76 @@ void UboManager::updateSlot(DriverApi& driver, AllocationId id,
...
     // Pass 2: Allocate slots for existing material instances that need to be orphaned.
-    for (FMaterialInstance* mi : existingInstances) {
-        if (mi->getUniformBuffer().isDirty() && mAllocator.isLockedByGpu(mi->getAllocationId())) {
-            mAllocator.retire(mi->getAllocationId());
-            auto [newId, newOffset] = mAllocator.allocate(mi->getUniformBuffer().getSize());
-            mi->assignUboAllocation(mUbHandle, newId, newOffset);
-            if (!BufferAllocator::isValid(newId)) {
-                reallocationNeeded = true;
-            }
+    for (auto& mi: mManagedInstances) {
```
pixelflinger: * instead of &

filament/src/details/UboManager.cpp
```diff
@@ -201,69 +187,76 @@ void UboManager::updateSlot(DriverApi& driver, AllocationId id,
...
-        miList.forEach([this](FMaterialInstance* mi) {
-            if (!mi->isUsingUboBatching()) {
-                return;
-            }
-            auto [newId, newOffset] = mAllocator.allocate(mi->getUniformBuffer().getSize());
-            assert_invariant(BufferAllocator::isValid(newId));
-            mi->assignUboAllocation(mUbHandle, newId, newOffset);
-        });
+void UboManager::allocateAllInstances() {
+    for (const auto& mi: mManagedInstances) {
```
pixelflinger: * instead of &

filament/src/details/Engine.cpp
```diff
@@ -1287,7 +1286,7 @@ bool FEngine::destroy(const FMaterialInstance* p) {
 
     if (p->isUsingUboBatching()) {
         assert_invariant(isUboBatchingEnabled());
-        mUboManager->retireSlot(p->getAllocationId());
+        mUboManager->unmanageMaterialInstance(p);
```
pixelflinger: why not do this in FMaterialInstance::terminate ?

show50726: That looks like a better place to put this logic. Updated!

filament/src/details/MaterialInstance.cpp
show50726: Yes, it was in `Engine::destroy` and it is moved to `FMaterialInstance::terminate()` now

filament/src/details/UboManager.cpp
show50726: Done.

filament/src/details/UboManager.cpp
show50726: Done.

filament/src/details/UboManager.cpp
show50726: Oh right it would not cause any exception. Removed the if

filament/src/details/UboManager.cpp
show50726: Done.

filament/src/details/UboManager.cpp
show50726: Added the comment to the line above.

filament/src/details/UboManager.cpp
show50726: Done

filament/src/details/UboManager.cpp
show50726: If you create a material instance and destroy it immediately, the id here would be invalid (though it should be rare).

And you're right, we still need to remove it from the list, but should not retire this slot since the id is invalid. I moved this check after the erase. Thanks for the catch!

filament/src/details/UboManager.cpp
```diff
@@ -164,24 +158,16 @@ void UboManager::finishBeginFrame(DriverApi& driver) {
...
-            }
-
-            const AllocationId id = mi->getAllocationId();
-            if (!BufferAllocator::isValid(id)) {
-                return;
-            }
+    for (const auto* mi : mManagedInstances) {
+        const AllocationId id = mi->getAllocationId();
+        if (!BufferAllocator::isValid(id)) {
+            return;
```
poweifeng: should this be continue?

filament/src/details/UboManager.cpp
```diff
@@ -201,69 +187,82 @@ void UboManager::updateSlot(DriverApi& driver, AllocationId id,
...
+}
+
+void UboManager::unmanageMaterialInstance(const FMaterialInstance* materialInstance) {
+    AllocationId id = materialInstance->getAllocationId();
+    // const_cast should be safe here since this cast is just to match the container type.
+    auto mi = const_cast<FMaterialInstance*>(materialInstance);
+    mPendingInstances.erase(mi);
+    mManagedInstances.erase(mi);
+
     if (!BufferAllocator::isValid(id))
```
poweifeng: nit: bracket

show50726: Done.

filament/src/details/UboManager.cpp
show50726: Good catch! We really need an unit test for this class...will try to add one

PR #9440, opened 2025-11-17
android: add readPixels callback for ModelViewer

poweifeng:


android/filament-utils-android/src/main/java/com/google/android/filament/utils/ModelViewer.kt
```diff
@@ -284,7 +288,7 @@ class ModelViewer(
      * @param frameTimeNanos time in nanoseconds when the frame started being rendered,
      *                       typically comes from {@link android.view.Choreographer.FrameCallback}
      */
-    fun render(frameTimeNanos: Long) {
+    fun render(frameTimeNanos: Long, callback: ((Bitmap) -> Unit)? = null) {
```
pixelflinger: This is essentially an public API so two questions:
- do we want to add this callback in a public API?
- this needs to be documented in the javadoc above

pixelflinger (changes requested):

android/filament-utils-android/src/main/java/com/google/android/filament/utils/ModelViewer.kt
poweifeng: > * do we want to add this callback in a public API?

You mean like `ModelViewer.setRenderedFrameCallback()` ? The case I need it for is to get a single frame and not every frame. And that might be the common case IMO.  wdyt

> * this needs to be documented in the javadoc above

done

pixelflinger: what I don't like about this API is that the user doesn't control the format of the bitmap, nor the region captured.

Format: what if the rendering is HDR?

It seems like a debug API, but it's right there on the main render() API.

Maybe a ModelViewer.debugGetNextFrameCallback()? Or, if we can make the API more generic?

poweifeng: yeah, your comment makes sense.  Actually, this might be a breaking api if users are using `render()` from java.  

I'll go with the `debugGetNextFrameCallback()`.

poweifeng: done

PR #9445, opened 2025-11-19
Fix push constants with ext sampler (VK)

anishmgoyal:
Currently, in Vulkan, we set the pipeline layout when binding the pipeline, which doesn't happen until draw time when external samplers are present. In cases like that, we push constants with the layout VK_NULL_HANDLE, which is not valid. Instead, we can wait until the pipeline layout is known, and then flush all push constants.


anishmgoyal:
@poweifeng @phoenixxxx @pixelflinger @rafadevai

poweifeng (approved):

filament/backend/src/vulkan/VulkanDriver.cpp
```diff
@@ -2233,6 +2233,9 @@ void VulkanDriver::draw2(uint32_t indexOffset, uint32_t indexCount, uint32_t ins
         if (pipelineLayout != mPipelineState.pipelineLayout) {
             bindPipelineImpl(bundle.pipelineState, pipelineLayout, bundle.descriptorSetMask);
         }
+        // If we're doing bind in draw, it's possible there are some push constants that weren't
+        // written out because we didn't have the pipelineLayout yet. Push them now.
+        program->flushPushConstants(pipelineLayout);
```
rafadevai: this should be inside the above if, only need to do this when we bind a new pipeline

filament/backend/src/vulkan/VulkanDriver.cpp
pixelflinger, line 2534: is this expected to be the common case? If not, it should be UTILS_UNLIKELY and in fact I think this should all be moved into a separate function.

filament/backend/src/vulkan/VulkanHandles.h
```diff
@@ -277,7 +286,13 @@ struct VulkanProgram : public HwProgram, fvkmemory::Resource {
 
     inline void writePushConstant(VkCommandBuffer cmdbuf, VkPipelineLayout layout,
             backend::ShaderStage stage, uint8_t index, backend::PushConstantVariant const& value) {
-        mInfo->pushConstantDescription.write(cmdbuf, layout, stage, index, value);
+        // It's possible that we don't have the layout yet. But, we don't want to "forget"
+        // about the push constant! We can flush the constants later, once the layout is set.
```
pixelflinger: can you explain in which scenario we have or don't have a layout?

filament/backend/src/vulkan/VulkanHandles.h
```diff
@@ -294,8 +309,16 @@ struct VulkanProgram : public HwProgram, fvkmemory::Resource {
...
+    struct PushConstantInfo {
+        VkCommandBuffer cmdbuf;
+        backend::ShaderStage stage;
+        uint8_t index;
+        backend::PushConstantVariant value;
+    };
+
     PipelineInfo* mInfo;
     VkDevice mDevice = VK_NULL_HANDLE;
+    std::vector<PushConstantInfo> mQueuedPushConstants;
```
pixelflinger, line 323: should we do a "reasonable" .reserve() for this, so we avoid too many heap allocations?

filament/backend/src/vulkan/VulkanDriver.cpp
poweifeng: it's common if you have externally sampled textures, otherwise will never be hit. So not sure how one would quantify that with `UTILS_LIKELY/UNLIKELY`

filament/backend/src/vulkan/VulkanHandles.h
anishmgoyal: i did consider that, but it's hard to know what a reasonable number is. the driver api takes one push constant at a time, and only the frontend knows how many push constants will be pushed. it's possible that actually, only the app knows about the push constants.

filament/backend/src/vulkan/VulkanHandles.h
anishmgoyal: for external samplers, in bindPipeline(), there's an early return - prior to setting the layout. push constants are then set after that, and *finally*, the layout is selected and bound in draw2() (called after binding push constants in the render pass).

filament/backend/src/vulkan/VulkanDriver.cpp
anishmgoyal: i think this is out of scope for this change, and a much broader question

filament/backend/src/vulkan/VulkanDriver.cpp
anishmgoyal: technically we probably shouldn't even have that if-statement, because it's guaranteed that condition will be true with the code written as it is today. that said, i'm happy to move this into the if block.

poweifeng: I do think that if block is still necessary, because you could have multiple draw calls for one pipeline.  It's just an optomization.

anishmgoyal: ahhhhh ok - then in that case, yes, this absolutely should be in the if block.

poweifeng: sorry, thinking about it some more.  I think it needs to be outside the conditional.  This is also valid

```
bindPipeline
setPushConstant
draw
setPushConstant
draw
```
We need to be able do that second update of push constants.

anishmgoyal: in the second case, we wouldn't have the push constants in a list that needs to be flushed. we're only storing values if the layout is null. in the second case, the layout should not be null. unless i'm missing something?

poweifeng: Ah yes you're right. The `mPipelinState.pipelineLayout` thing.   Can you add a comment to that effect above the `program->flushPushConstants(pipelineLayout);`? Something like, "Further push constant updates after the pipeline is bound will immediately set the constants because the pipeline is known (see `VulkanProgram::writePushConstant`)". 

Because this logic is getting harder to follow with intertwining paths.

anishmgoyal: added exactly that comment, sums it up fairly well.

anishmgoyal:
@pixelflinger please re-review when you get a chance!

filament/backend/src/vulkan/VulkanHandles.h
pixelflinger: I meant, in the comment :-)

filament/backend/src/vulkan/VulkanHandles.h
pixelflinger: Actually, we have `MAX_PUSH_CONSTANT_COUNT` (32). So I think we can be pragmatic here and maybe do a reserve with 8 for instance.

rafadevai: can we just use an array with the fixed MAX_PUSH_CONSTANT_COUNT size?

filament/backend/src/vulkan/VulkanDriver.cpp
```diff
@@ -2232,6 +2232,11 @@ void VulkanDriver::draw2(uint32_t indexOffset, uint32_t indexCount, uint32_t ins
         VkPipelineLayout const pipelineLayout = mPipelineLayoutCache.getLayout(vklayouts, program);
         if (pipelineLayout != mPipelineState.pipelineLayout) {
             bindPipelineImpl(bundle.pipelineState, pipelineLayout, bundle.descriptorSetMask);
+            // Any push constants received while the layout was null were not written out. They were
+            // added to a list of pending constants, so we'll flush them now.
+            // Further push constant updates after the pipeline is bound will immediately set
+            // the constants because the layout is non-null (see VulkanProgram::writePushConstant).
+            program->flushPushConstants(pipelineLayout);
```
rafadevai, line 2560: On a different approach and this for me is better, is to just use the RGB layout of the pipeline to set the push constants?

I expect the pipeline layouts to be compatible as described here https://registry.khronos.org/VulkanSC/specs/1.0-extensions/html/vkspec.html#descriptorsets-compatibility

anishmgoyal: vkCmdPushConstants takes `layout` as a parameter. i don't see it implied in the docs that it's sufficient to use a compatible layout instead of the actual layout that the constants are being pushed for, so i don't believe we should make that change.

while they technically might currently have the same push constant layouts, it'd be fairly flimsy. any change where, say, we intro a push constant specific to external samplers (for some reason we can't anticipate today), and we'd likely be existing in the realm of undefined behavior.

rafadevai: Youre right, it does say it have to identically layout, with all the creation parameters the same, which is not true for external samplers.
 
Too bad :(

filament/backend/src/vulkan/VulkanHandles.h
pixelflinger: the common case is that we have 0 push constants, and the less common case is that we have 1. So always allocating 32 seems a lot.  Maybe you were right from the start that not even doing a reserve() is the better choice (maybe add a comment that the common case is 0 push constants for now)

filament/backend/src/vulkan/VulkanHandles.h
anishmgoyal: ahh, makes sense! updated.

filament/backend/src/vulkan/VulkanHandles.h
anishmgoyal: i'll leave this be then

PR #9449, opened 2025-11-20
don't assume compositor and frame timing are always available

pixelflinger:
- this is not true fro headless swapchains
- and potentially some timings might not be available on a given nativewindow

Previously the code would handle these errors gracefully, but the  errors were still generated. Now, we query the availability and only make the calls  if supported.


poweifeng (approved):
lgtm again

PR #9453, opened 2025-11-22
Fix bindVAO for OpenGL

z3moon:
Currently bindVertexArray call generates an object for the default VAO, which is not originally intended.


pixelflinger:
does this fix a leak? or a real bug? is there a buganizer for this?

z3moon:
> does this fix a leak? or a real bug? is there a buganizer for this?

No buganizer for this. While normally harmless, generating a vertex array object (VAO) for mDefaultVAO created a problem in a shared context environment because VAOs cannot be shared between OpenGL shared contexts. It was a latent bug.

PR #9456, opened 2025-11-25
Validation layer issue nullsampler

phoenixxxx:
Addressing the validation layer issue where we encounter RGB samplers next to YUV.


phoenixxxx:
We can close this as it was merged in the Stream PR.

PR #9462, opened 2025-12-01
Add config option for asynchronous mode

z3moon:


filament/include/filament/Engine.h
```diff
@@ -742,6 +748,14 @@ class UTILS_PUBLIC Engine {
      */
     bool isStereoSupported(StereoscopicType stereoscopicType) const noexcept;
 
+    /**
+     * Queries the device and platform for support of asynchronous operation.
+     *
+     * @param mode The asynchronous mode that the caller wants to check whether to support.
+     * @return The mode the engine currently supports. Returns NONE if none of them is supported.
+     */
+    AsynchronousMode isAsynchronousOperationSupported(AsynchronousMode mode) const noexcept;
```
bejado: Maybe this should be renamed to `resolveAsynchronousMode` or something, and the argument can be called `hint`? I expect a method that starts with "is" to return a bool.

z3moon: Yeah it makes sense. Done.

filament/backend/include/backend/Platform.h
```diff
@@ -255,6 +255,34 @@ class UTILS_PUBLIC Platform {
...
+        /**
+         * Asynchronous operations are disabled. This is the default.
+         */
+        NONE,
+
+        /**
+         * Attempts to use a dedicated worker thread for asynchronous tasks. If threading is not
+         * supported by the platform, it automatically falls back to using an amortization strategy.
+         */
+        THREAD_PREFERRED,
```
poweifeng, line 271: I feel like the client is able to do the "preferred" logic themselves given that you've given them an API to check the mode's availability.

filament/src/details/Engine.cpp
```diff
@@ -1503,6 +1511,19 @@ Engine::FeatureLevel FEngine::setActiveFeatureLevel(FeatureLevel featureLevel) {
     return (mActiveFeatureLevel = std::max(mActiveFeatureLevel, featureLevel));
 }
 
+AsynchronousMode FEngine::resolveAsynchronousMode(AsynchronousMode hint) const noexcept {
```
poweifeng: don't you have to plumb this to the backends?  Right now, no backend supports async mode right?

z3moon: Yeah it'll be done in followup PRs. This is a preliminary change for that.

filament/backend/include/backend/Platform.h
z3moon: That resolve function can be called/used only after the engine is created. So users can't rely on it to determine which option to use for engine creation. And actually I realized the resolve function is not what I originally intended. Let me update this PR.

filament/backend/include/backend/Platform.h
```diff
@@ -255,6 +255,34 @@ class UTILS_PUBLIC Platform {
...
+
+        /**
+         * Mandates the use of a dedicated worker thread for asynchronous tasks. If the platform
+         * does not support threads, engine initialization will fail.
+         */
+        THREAD,
```
pixelflinger: do we need this? I don't think it adds value. when a user uses THREAD the engine creation can fail, and they might not know why. I would get rid of it until someone asks for it.

z3moon: Users will see the error message in that case, so they'll know why it failed. But I agree that this may not be necessary at this point. Let me remove it.

filament/include/filament/Engine.h
```diff
@@ -742,6 +748,14 @@ class UTILS_PUBLIC Engine {
      */
     bool isStereoSupported(StereoscopicType stereoscopicType) const noexcept;
 
+    /**
+     * Checks if the engine is set up for asynchronous operation. If it returns true, the
+     * asynchronous versions of the APIs are available for use.
+     *
+     * @return true if the engine supports asynchronous operation.
+     */
+    bool isAsynchronousOperationSupported() const noexcept;
```
poweifeng, line 757: If neither thread nor amoritization are supported, don't we crash on engine creation?  So the client would already know if asynchronous op is supported or not. So do we still need this?

z3moon: Amortization is always available, so it doesn't crash. (The only option that may crash, which was THREAD, has been removed)
If the feature flag is turned off, then the user's option is ignored and the engine is always configured with NONE. This method returns the current engine setting, so it allows users to decide which functions to use depending on the returned value.

poweifeng: I think they can get that from `getConfig` below?

z3moon: Ah, I missed adding additional check. Updated.

poweifeng: looking at the code, if there was no feature flag, then this API is not needed correct?  If so, IMO we shouldn't introduce an API that's dependent on a feature flag.

z3moon: Well, I think it depends on what the API offers to users. This is a helper function to help users easily decide which API sets they should use without considering two different factors. (config setting they provided originally and feature flag). I expect users will prepare two different code paths (sync vs. async) and branch them using this function. We could remove this and have them handle the branching properly with the feature flag and the original setting, which I think is error-prone and not convenient. Even without the feature flag, I still think it's better the engine has this helper function.

poweifeng: If there was no feature flag, isn't it always supported? because amortization doesn't require threading, and you'll at least have that?

z3moon: Yeah, armotization should be always supported if they ask. Still this is a basic query function that I believe the engine should offer. Otherwise users should understand all enums and use them properly with their original setting for branching. Also they may need to change code if we add more enums later. (which I think is less likely tough, but possible)

poweifeng (commented):
I think you'll also need to add java bindings.

z3moon:
> I think you'll also need to add java bindings.

Yup, I'm wring the bindings now, and they'll be following as separate PRs.

PR #9465, opened 2025-12-01
Fix race condition with garbage collection

z3moon:
This change fixes an intermittent crash caused by a thread race condition.

The crash happened when a CallbackHandler::Callback (triggered by engine.flush()) attempted to create an instance of a Structure of Arrays (SoA) like TransformManager, which adds items to SoA) at the same time that the garbage collection job (FEngine::gc, which removes items from SoA) was running.

This conflict is resolved by modifying the logic to ensure that the garbage collection task completes synchronously before the callback operations begin.


filament/src/details/Renderer.cpp
```diff
@@ -296,12 +296,9 @@ void FRenderer::skipFrame(uint64_t vsyncSteadyClockTimeNano) {
     // WARNING: while doing this we can't access any component manager
     auto& js = engine.getJobSystem();
 
-    auto *job = js.runAndRetain(jobs::createJob(js, nullptr, &FEngine::gc, &engine)); // gc all managers
+    js.runAndWait(jobs::createJob(js, nullptr, &FEngine::gc, &engine)); // gc all managers
 
     engine.flush();     // flush command stream
-
-    // make sure we're done with the gcs
-    js.waitAndRelease(job);
```
poweifeng, line 304: does the ordering matter here?  Would you want to gc after the flush()?

z3moon: Yeah looks like it's better.

poweifeng: I actually don't know if my suggestion is valid or not.  LGTM to me, but please wait for pixelflinger to approve.

poweifeng (approved):

filament/src/details/Renderer.cpp
z3moon: it's marginally better because SoA components can be deleted when `engine.flush()` is called, and then those deleted components are immediately garbage collected.

pixelflinger: probably doesn't matter since before it was done in parallel.

PR #9470, opened 2025-12-02
Implement setPresentationTime for Metal

bejado:
BUGS = [447688607]

### DIFF
--- a/filament/backend/src/metal/MetalContext.h
+++ b/filament/backend/src/metal/MetalContext.h
@@ -132,11 +132,6 @@ struct MetalContext {
     uint32_t currentFrame = 0;
     MetalErrorQueue commandBufferErrors;
 
-    // If zero, the next presentation should happen as soon as possible.
-    // Otherwise, this is the timestamp when the present should happen.
-    // Resets to 0 after the present.
-    int64_t presentationTimeNs = 0;
-
     std::atomic<bool> memorylessLimitsReached = false;
 
     // Supported features.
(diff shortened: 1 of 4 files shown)


filament/backend/src/metal/MetalContext.h
```diff
@@ -132,6 +132,11 @@ struct MetalContext {
     uint32_t currentFrame = 0;
     MetalErrorQueue commandBufferErrors;
 
+    // If zero, the next presentation should happen as soon as possible.
+    // Otherwise, this is the timestamp when the present should happen.
+    // Resetst to 0 after the present.
```
poweifeng:
```suggestion
    // Reset to 0 after the present.
```

pixelflinger:
```suggestion
    // Resets to 0 after the present.
```

filament/backend/src/metal/MetalDriver.mm
```diff
@@ -312,6 +312,7 @@
 }
 
 void MetalDriver::setPresentationTime(int64_t monotonic_clock_ns) {
```
pixelflinger, line 314: `Renderer::setPresentationTime()` must be called between beginFrame/enFrame, and this is because it acts on the current swapchain, i.e. the one set in `makeCurrent`.

So the "presentationTime" cannot be associate to the context, it needs to be associated to the current swapchain.

If the user is using several Renderer with different swapchains like, rendering in different windows, this will break.


The backend API is bad, my fault. It should have been `setPresentationTime(Handle<SwapChain>, int64_t)`.

bejado: Got it, I was wondering about that. I'll associate the presentation time with the SwapChain instead

PR #9471, opened 2025-12-02
Implement KHR_materials_dispersion

### LINKED ISSUE #8307: Support for more glTF material extensions.

Hello, I don't know if Filament is still under active development w.r.t. adding new features or primarily maintenance, but a couple of new glTF extensions dropped recently: 

https://github.com/KhronosGroup/glTF/tree/main/extensions/2.0/Khronos/KHR_materials_dispersion
https://github.com/KhronosGroup/glTF/tree/main/extensions/2.0/Khronos/KHR_materials_iridescence
https://github.com/KhronosGroup/glTF/tree/main/extensions/2.0/Khronos/KHR_materials_diffuse_transmission

These three seem reasonably easy to add (given that they include sample code) and would make Filament a more compliant glTF renderer.

support for three new glTF material extensions.

None

**OS and backend**
not backend specific

Nemi26:
Hi, this commit implements the gltf dispersion extension. Partially fixes https://github.com/google/filament/issues/8307

Dragon Dispersion demonstration:
image

### DIFF
--- a/shaders/src/surface_light_indirect.fs
+++ b/shaders/src/surface_light_indirect.fs
@@ -447,22 +447,22 @@ struct Refraction {
     float d;
 };
 
-void refractionSolidSphere(const PixelParams pixel,
+void refractionSolidSphere(float etaIR, float etaRI, float thickness,
     const vec3 n, vec3 r, out Refraction ray) {
-    r = refract(r, n, pixel.etaIR);
+    r = refract(r, n, etaIR);
     float NoR = dot(n, r);
-    float d = pixel.thickness * -NoR;
+    float d = thickness * -NoR;
     ray.position = vec3(shading_position + r * d);
     ray.d = d;
     vec3 n1 = normalize(NoR * r - n * 0.5);
-    ray.direction = refract(r, n1,  pixel.etaRI);
+    ray.direction = refract(r, n1,  etaRI);
 }
 
-void refractionSolidBox(const PixelParams pixel,
+void refractionSolidBox(float etaIR, float thickness,
     const vec3 n, vec3 r, out Refraction ray) {
-    vec3 rr = refract(r, n, pixel.etaIR);
+    vec3 rr = refract(r, n, etaIR);
     float NoR = dot(n, rr);
-    float d = pixel.thickness / max(-NoR, 0.001);
+    float d = thickness / max(-NoR, 0.001);
     ray.position = vec3(shading_position + rr * d);
     ray.direction = r;
     ray.d = d;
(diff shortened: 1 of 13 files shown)


pixelflinger (changes requested):
Thank you! A few comments:

Do you think you could refactor the code so that everything that goes into the "for loop" can be a function? I think this would make this clearer?

You need to update the material documentation at  ./docs_src/src_markdeep/Materials.md.html

pixelflinger:
@Nemi26 also you need to add to the release notes

pixelflinger:
oh, and also I think you need to add a unit test for that property.

shaders/src/surface_light_indirect.fs
```diff
@@ -494,63 +494,79 @@ vec3 evaluateRefraction(
...
-    // since it's a fairly subtle effect.
-    E *= 1.0 + pixel.transmission * (1.0 - E.g) / (1.0 + E.g);
+        // For thin surfaces, the light will bounce off at the second interface in the direction of
+        // the reflection, effectively adding to the specular, but this process will repeat itself.
+        // Each time the ray exits the surface on the front side after the first bounce,
+        // it's multiplied by E^2, and we get: E + E(1-E)^2 + E^3(1-E)^2 + ...
+        // This infinite series converges and is easy to simplify.
+        // Note: we calculate these bounces only on a single component,
+        // since it's a fairly subtle effect.
+        E *= 1.0 + pixel.transmission * (1.0 - E.g) / (1.0 + E.g);
```
z3moon: This `E` is inside the for-loop, meaning the modification for the first channel (red) will affect the second and third channels (green/blue). Not sure if this is intended, but otherwise you may want to make a local copy of E for each iteration as with etaRI and etaRI.

Nemi26: oops, good catch

Nemi26: Actually this doesn't affect dispersion, I forgot to add guard checks for this but dispersion only works on solid refraction

z3moon: I see this affects the output value `Ft` of evaluateRefraction, which will then added to `color.rgb` at call site. Could you double check if this is okay?

pixelflinger:
@Nemi26 see my comments above.

Nemi26:
@pixelflinger I added the things you mentioned, as well as some tiny fixes (checking that refraction type is solid in order for dispersion to be enabled). I wasn't able to pull out that refraction inner loop code into a separate function in a way that would make it more readable unfortunately.

shaders/src/surface_light_indirect.fs
```diff
@@ -494,63 +494,80 @@ vec3 evaluateRefraction(
...
 
-    // apply absorption
+        // apply absorption
 #if defined(MATERIAL_HAS_ABSORPTION)
-    Ft *= T;
+        t *= T;
+#endif
+
+#if defined(MATERIAL_HAS_DISPERSION) && (REFRACTION_TYPE == REFRACTION_TYPE_SOLID)
+        Ft[i] = t[i];
```
pixelflinger, line 567: Maybe I'm not reading this correctly, but it looks like in the loop you're computing 3 components of `t` and are discarding two of them. I don't know for a fact that the compiler optimizes that.

It seems to my that in the dispersion case, `t` should be a scalar.

It might make the code even harder to read though. Can you double-check that the spirv-opt compiler will eliminate the computation of the components that are discarded during each loop? If that's the case, then we can just add a comment. If not, I think we need to bite the bullet and force `t` to be a scalar.

wdyt?

pixelflinger (changes requested):
it looks pretty good, but I'm worried about doing too much math in the dispersion case. see my comment inline.

pixelflinger:
oops, looks like all builds are broken.

shaders/src/surface_light_indirect.fs
Nemi26: @romainguy said in the linked issue that he had a working version of dispersion (and diffuse transmission) somewhere. Maybe he solved this in a better way?

pixelflinger (approved):

z3moon (approved):

pixelflinger:
@Nemi26 the dispersion doesn't work with gltf_viewer's `-u` (ubershader) option; it crashes. FYI. 
Do you want to give it a shot at fixing it, or should I revert the change for now?

Also, I do see the effect working with khronos' DragonDispersion test, but not with the `DispersionTest`.

PR #9474, opened 2025-12-03
Correctly add flag to guard camera near/far assertion (#9472)

pixelflinger:
The previous attempt changed the precondition checks and made them too aggressive.  For instance, near=0 is allowed with an ORTHO projection.

### DIFF
--- a/filament/src/details/Camera.cpp
+++ b/filament/src/details/Camera.cpp
@@ -95,13 +95,9 @@ mat4 FCamera::projection(double const focalLengthInMillimeters,
 void UTILS_NOINLINE FCamera::setCustomProjection(mat4 const& projection,
         mat4 const& projectionForCulling, double const near, double const far) noexcept {
 
-    FILAMENT_CHECK_PRECONDITION(near > 0)
-            << "Camera preconditions not met in setCustomProjection(): near <= 0, near=" << near;
-    auto const& featureFlags = mEngine.features.engine.debug;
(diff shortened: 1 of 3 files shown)


filament/src/details/Camera.cpp
```diff
@@ -136,6 +128,32 @@ void UTILS_NOINLINE FCamera::setCustomEyeProjection(mat4 const* projection, size
     mFar = far;
 }
 
+void FCamera::setProjection(double const fovInDegrees, double const aspect,
+        double const near, double const far, Fov const direction) {
+
+    FILAMENT_CHECK_PRECONDITION(near > 0 && far > near)
+            << "Camera preconditions not met in setProjection(): near <= 0 or far <= near, near="
+            << near << ", far=" << far;
```
poweifeng: this is actually the one we want flag guarded.

filament/src/details/Camera.cpp
```diff
@@ -136,6 +128,32 @@ void UTILS_NOINLINE FCamera::setCustomEyeProjection(mat4 const* projection, size
...
+    setCustomProjection(
+            Camera::projection(direction, fovInDegrees, aspect, near),
+            Camera::projection(direction, fovInDegrees, aspect, near, far),
+            near, far);
+}
+
+void FCamera::setLensProjection(double const focalLengthInMillimeters,
+        double const aspect, double const near, double const far) {
+
+    FILAMENT_CHECK_PRECONDITION(near > 0 && far > near)
```
poweifeng: same here

PR #9481, opened 2025-12-03
Metal: ensure framebuffer fetch is only used on macOS 11.0+

bejado:


filament/backend/src/metal/MetalDriver.mm
```diff
@@ -1144,8 +1144,12 @@
 bool MetalDriver::isFrameBufferFetchSupported() {
     // FrameBuffer fetch is achievable via "programmable blending" in Metal, and only supported on
     // Apple GPUs with readWriteTextureSupport.
-    return mContext->highestSupportedGpuFamily.apple >= 1 &&
-            mContext->device.readWriteTextureSupport;
+    // On macOS, framebuffer fetch requires MSL 2.3, which is only available with macOS 11.0.
+    if (@available(macOS 11.0, *)) {
+        return mContext->highestSupportedGpuFamily.apple >= 1 &&
```
poweifeng, line 1149: what about iOS?  or is that somehow a subset?

bejado: All versions of iOS (that Filament supports) support framebuffer fetch (as long as that boolean expression returns true). The `@available(macOS 11.0, *)` statement will return true for all versions of iOS.

poweifeng (approved):

PR #9484, opened 2025-12-04
Allow embedding material source in cmat

trie94:
matc now takes in an optional flag, which embeds the source mat string into cmat.
a new chunk type is added for this, which will optionally compress the source material string chunk.

FIXES = 465855395

### DIFF
--- a/filament/backend/CMakeLists.txt
+++ b/filament/backend/CMakeLists.txt
@@ -36,6 +36,7 @@ set(SRCS
         src/Driver.cpp
         src/Handle.cpp
         src/HandleAllocator.cpp
+        src/JobQueue.cpp
         src/ostream.cpp
         src/noop/NoopDriver.cpp
         src/noop/PlatformNoop.cpp
@@ -59,6 +60,7 @@ set(PRIVATE_HDRS
         src/CompilerThreadPool.h
         src/DataReshaper.h
         src/DriverBase.h
+        src/JobQueue.h
 )
 
 # ==================================================================================================
@@ -568,6 +570,7 @@ if (APPLE OR LINUX)
         test/test_ReadPixels.cpp
         test/test_BufferUpdates.cpp
         test/test_Callbacks.cpp
+        test/test_JobQueue.cpp
         test/test_MemoryMappedBuffer.cpp
         test/test_MsaaSwapChain.cpp
         test/test_MRT.cpp
--- a/filament/backend/src/JobQueue.cpp
+++ b/filament/backend/src/JobQueue.cpp
@@ -1,0 +1,247 @@
+/*
+ * Copyright (C) 2025 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "JobQueue.h"
+
+#include <utils/compiler.h>
+#include <utils/debug.h>
+#include <utils/Panic.h>
+
+namespace filament::backend {
+
+JobQueue::JobQueue(PassKey) {}
+
+JobQueue::JobId JobQueue::push(Job job, JobId const preIssuedJobId/* = InvalidJobId*/) {
+    JobId jobId = preIssuedJobId;
+    {
+        std::lock_guard<std::mutex> lock(mQueueMutex);
+        if (mIsStopping) {
+            return InvalidJobId;
+        }
+
+        if (jobId == InvalidJobId) {
+            jobId = genNextJobId();
+            mJobsMap[jobId] = std::move(job);
+        } else {
+            // Use the job ID previously issued by `issueJobId()`
+            auto it = mJobsMap.find(jobId);
+            if (it == mJobsMap.end()) {
+                // Pre-issued job does not exist, either users passed a wrong id (unlikely)
+                // or the job must have been canceled (likely)
+                return InvalidJobId;
+            }
+            FILAMENT_CHECK_PRECONDITION(!static_cast<bool>(it->second))
+                    << "pre-issued job has already been populated";
+            it->second = std::move(job);
+        }
+        mJobOrder.push(jobId);
+    }
+
+    // Always notify. A ThreadWorker might be waiting.
+    mQueueCondition.notify_one();
+
+    return jobId;
+}
+
+JobQueue::Job JobQueue::pop(bool shouldBlock) {
+    std::unique_lock<std::mutex> lock(mQueueMutex);
+
+    decltype(mJobsMap)::iterator it;
+
+    while (true) {
+        if (shouldBlock) {
+            // Wait only if we're in blocking mode and the queue is empty
+            mQueueCondition.wait(lock, [this] { return !mJobOrder.empty() || mIsStopping; });
+        }
+
+        if (mJobOrder.empty()) {
+            // When `shouldBlock` is true, this means the queue is stopping now.
+            // When `shouldBlock` is false, this means there's no job.
+            return nullptr;
+        }
+
+        JobId jobId = mJobOrder.front();
+        mJobOrder.pop();
+
+        it = mJobsMap.find(jobId);
+        if (it != mJobsMap.end()) {
+            break;
+        }
+
+        // If execution reaches this line, the job must have been canceled right after being added.
+        // Therefore, we should continue the loop and attempt to retrieve the next available job.
+    }
+
+    Job job = std::move(it->second);
+    mJobsMap.erase(it);
+    return job;
+}
+
+utils::FixedCapacityVector<JobQueue::Job> JobQueue::popBatch(int const maxJobsToPop) {
+    utils::FixedCapacityVector<Job> jobs;
+
+    if (UTILS_UNLIKELY(maxJobsToPop == 0)) {
+        return jobs;
+    }
+
+    std::lock_guard<std::mutex> lock(mQueueMutex);
+    if (mJobOrder.empty()) {
+        return jobs;
+    }
+
+    // Calculate jobs to take. If maxJobsToPop is negative, we take all jobs.
+    size_t jobsToTake = mJobOrder.size();
+    if (0 < maxJobsToPop && maxJobsToPop < static_cast<int>(jobsToTake)) {
+        jobsToTake = maxJobsToPop;
+    }
+    jobs.reserve(jobsToTake);
+
+    while (0 < jobsToTake && !mJobOrder.empty()) {
+        JobId jobId = mJobOrder.front();
+        mJobOrder.pop();
+
+        auto it = mJobsMap.find(jobId);
+        if (UTILS_UNLIKELY(it == mJobsMap.end())) {
+            // The job was probably canceled.
+            continue;
+        }
+
+        jobs.push_back(std::move(it->second));
+        --jobsToTake;
+        mJobsMap.erase(it);
+    }
+
+    return jobs;
+}
+
+JobQueue::JobId JobQueue::issueJobId() noexcept {
+    std::lock_guard<std::mutex> lock(mQueueMutex);
+    JobId const jobId = genNextJobId();
+    // Preallocate a job, which serves two main purposes. It provides a valid jobId that can be
+    // checked for integrity when passed to the `push` method, and it enables job cancellation for
+    // tasks that are yet to be pushed.
+    mJobsMap[jobId];
+    return jobId;
+}
+
+bool JobQueue::cancel(JobId const jobId) noexcept {
+    std::lock_guard<std::mutex> lock(mQueueMutex);
+
+    auto it = mJobsMap.find(jobId);
+    if (it == mJobsMap.end()) {
+        return false; // Job not found, must have been completed or canceled.
+    }
+
+    mJobsMap.erase(it);
+
+    return true;
+}
+
+void JobQueue::stop() noexcept {
+    {
+        std::lock_guard<std::mutex> lock(mQueueMutex);
+        mIsStopping = true;
+    }
+    mQueueCondition.notify_all(); // Wake up all waiting threads
+}
+
+JobQueue::JobId JobQueue::genNextJobId() noexcept {
+    // We assume this method is called within the critical section.
+    JobId newJobId = mNextJobId++;
+    // We assume the job ID won't overflow or wraps around to zero within the application's lifetime.
+    assert_invariant(newJobId != InvalidJobId);
+    return newJobId;
+}
+
+JobWorker::~JobWorker() = default;
+
+void JobWorker::terminate() {
+    // This is called from workers `terminate()`, which may hinder the concurrent use of multiple
+    // workers. Consider removing this line and require the owner/caller to explicitly invoke it to
+    // enable multiple worker instances.
+    if (mQueue) {
+        mQueue->stop();
+    }
+}
+
+AmortizationWorker::AmortizationWorker(JobQueue::Ptr queue, PassKey)
+    : JobWorker(std::move(queue)) {
+}
+
+AmortizationWorker::~AmortizationWorker() = default;
+
+void AmortizationWorker::process(int const jobCount) {
+    if (!mQueue || jobCount == 0) {
+        return;
+    }
+
+    if (jobCount == 1) {
+        // Handle single job without vector allocation.
+        if (auto job = mQueue->pop(false)) {
+            job();
+        }
+        return;
+    }
+
+    // Handle batch (jobCount > 1 or jobCount < 0 for "all pending jobs")
+    utils::FixedCapacityVector<JobQueue::Job> jobs = mQueue->popBatch(jobCount);
+    if (jobs.empty()) {
+        return;
+    }
+
+    for (auto& job: jobs) {
+        job();
+    }
+}
+
+void AmortizationWorker::terminate() {
+    JobWorker::terminate();
+
+    // Drain all pending jobs.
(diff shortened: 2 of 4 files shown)


filament/test/filament_test_material.cpp
```diff
@@ -14,7 +14,9 @@
  * limitations under the License.
  */
 
-#include <gtest/gtest.h>
+#include <gmock/gmock-matchers.h>
+
+#include <filamat/MaterialBuilder.h>
```
show50726, line 18: You might want to update cmake file for this dependency?

trie94: yeah.. what's weird is it works locally without the dep???
i just added filamat in CMakeLists. let's see if that makes the presubmit happy...

pixelflinger:
> follow up: we also need target api and variant filter to be embedded in cmat, will do it as a follow up.

This needs a better comment + a FIXES=[] line.

filament/include/filament/Material.h
```diff
@@ -371,7 +371,8 @@ class UTILS_PUBLIC Material : public FilamentAPI {
     //! Indicates whether an existing parameter is a sampler or not.
     bool isSampler(const char* UTILS_NONNULL name) const noexcept;
 
-
+    //! Returns the material source (ASCII) if exists.
+    const char* UTILS_NULLABLE getSource() const noexcept;
```
pixelflinger: You need to document clearly the lifetime of the `const char*`. Might be better to return a `std::string_view`.

libs/filamat/include/filamat/MaterialBuilder.h
```diff
@@ -660,6 +660,11 @@ class UTILS_PUBLIC MaterialBuilder : public MaterialBuilderBase {
      */
     MaterialBuilder& useDefaultDepthVariant() noexcept;
 
+    /**
+     * Sets the source ASCII material.
+     */
+    MaterialBuilder& materialSource(const char* code, size_t size = 0) noexcept;
```
pixelflinger: same here maybe a string_view is better. Also document the life-time of 'code'.

filament/include/filament/Material.h
pixelflinger: Also document better what we mean by "material source". I think we mean the ".mat", that is the json-ish file.

pixelflinger (changes requested):

libs/filabridge/include/filament/MaterialChunkType.h
```diff
@@ -101,6 +101,8 @@ enum UTILS_PUBLIC ChunkType : uint64_t {
     DictionaryMetalLibrary = charTo64bitNum("DIC_MLIB"),
 
     MaterialCrc32 = charTo64bitNum("MAT_CRC "),
+
+    MaterialSource = charTo64bitNum("MAT_SOURCE"),
```
pixelflinger: this needs to be 8 bytes.

libs/filamat/include/filamat/MaterialBuilder.h
z3moon: `size = 0` seems to be error-prone here. E.g., if users forgot to pass size, they would expect the size to be calculated internally, but it's not. This leads to a null CString. Maybe consider using string_view as mathias suggested.

libs/filabridge/include/filament/MaterialChunkType.h
trie94: yeah changed it to MAT_SRC

libs/filamat/include/filamat/MaterialBuilder.h
trie94: removed the size field, changing to std::string_view

libs/filamat/src/eiff/CompressedStringChunk.cpp
```diff
@@ -0,0 +1,29 @@
...
+* limitations under the License.
+*/
+
+#include "CompressedStringChunk.h"
+
+namespace filamat {
+
+void CompressedStringChunk::flatten(filamat::Flattener& f) {
+    if (mCompression == Compression::NONE) {
+        // use zstd to compress??
```
trie94: i understand at high level i need to compress here, and uncompress when reading the data. but i am having trouble what function/library to use to actually achieve that. i am not very familiar with zstd, but i guess i need to do some treatment here to actually compress the chunk?

libs/filamat/src/eiff/CompressedStringChunk.cpp
```diff
@@ -0,0 +1,29 @@
...
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+
+#include "CompressedStringChunk.h"
+
+namespace filamat {
+
+void CompressedStringChunk::flatten(filamat::Flattener& f) {
+    if (mCompression == Compression::NONE) {
```
pixelflinger:
```suggestion
    if (mCompression != Compression::NONE) {
```

libs/filamat/src/eiff/CompressedStringChunk.cpp
pixelflinger: yes, zstd is very easy to use. there are some example in the source tree already. look at resgen for instance. also there are uncompression usages in filament itself and even a helper class there.

libs/filamat/src/eiff/CompressedStringChunk.h
```diff
@@ -0,0 +1,43 @@
...
+
+    CompressedStringChunk(ChunkType type, Compression compression, std::string_view string)
+            : Chunk(type), mString(string) {}
+    ~CompressedStringChunk() = default;
+
+private:
+    void flatten(Flattener& f) override;
+
+    Compression mCompression;
+    std::string_view mString;
```
pixelflinger: I think you have to copy the string here because the lifetime of the string_view content is not guaranteed after the constructor is called.

libs/filamat/src/MaterialBuilder.cpp
```diff
@@ -1701,6 +1707,9 @@ void MaterialBuilder::writeCommonChunks(ChunkContainer& container, MaterialInfo&
                     hasher({ frag.data(), frag.size() })));
 
     container.emplace<uint64_t>(MaterialCacheId, materialId);
+    if (!mMaterialSource.empty()) {
+        container.push<CompressedStringChunk>(MaterialSource, mMaterialSource.data());
```
pixelflinger:
```suggestion
        container.push<CompressedStringChunk>(MaterialSource, mMaterialSource);
```

filament/include/filament/Material.h
```diff
@@ -371,7 +371,12 @@ class UTILS_PUBLIC Material : public FilamentAPI {
     //! Indicates whether an existing parameter is a sampler or not.
     bool isSampler(const char* UTILS_NONNULL name) const noexcept;
 
-
+    /**
+     * Returns a view of the material source (.mat which is a JSON-ish file) string,
+     * if it has been set. Otherwise, it returns a view of an empty string.
+     * The lifetime of the string_view is tied to the lifetime of the Material.
+     */
+    std::string_view getSource() const noexcept;
```
bejado, line 379:
```suggestion
    std::string_view getSource() const noexcept;
 
```

tools/matc/src/matc/CommandlineConfig.cpp
```diff
@@ -163,6 +164,8 @@ static void usage(char* name) {
             "       #if defined(GL_GOOGLE_cpp_style_line_directive)\n"
             "       Some drivers may complain about the use of cpp style #line directives\n"
             "       if they don't support it.\n\n"
+            "   --include-source-mat\n"
+            "       Embeds the source ASCII material if set.\n"
```
bejado: We should use the noun "material definition" to refer to an uncompiled material.

```suggestion
            "       Embeds the source ASCII material definition.\n"
```

libs/filabridge/include/filament/MaterialChunkType.h
```diff
@@ -101,6 +101,8 @@ enum UTILS_PUBLIC ChunkType : uint64_t {
     DictionaryMetalLibrary = charTo64bitNum("DIC_MLIB"),
 
     MaterialCrc32 = charTo64bitNum("MAT_CRC "),
+
+    MaterialSource = charTo64bitNum("MAT_SRC"),
```
bejado: I think these chunk types need to be 8 chars long (see `MAT_CRC `) above. So we need a space at the end.

```suggestion
    MaterialSource = charTo64bitNum("MAT_SRC "),
```

libs/filamat/src/eiff/CompressedStringChunk.h
trie94: sounds good. will use CString

trie94:
i see
"""
ld: warning: duplicate symbol '_ZSTD_compressBlock_lazy2_dedicatedDictSearch' in:
    /Users/yeinj/filament/out/cmake-release/third_party/basisu/tnt/libzstd-basisu.a[2](zstd.c.o)
    /Users/yeinj/filament/out/cmake-release/third_party/zstd/tnt/libzstd.a[19](zstd_lazy.c.o)
"""
when running the build script, how can i fix this? but i need to include zstd...

show50726:
> i see """ ld: warning: duplicate symbol '_ZSTD_compressBlock_lazy2_dedicatedDictSearch' in: /Users/yeinj/filament/out/cmake-release/third_party/basisu/tnt/libzstd-basisu.a[2](zstd.c.o) /Users/yeinj/filament/out/cmake-release/third_party/zstd/tnt/libzstd.a[19](zstd_lazy.c.o) """ when running the build script, how can i fix this? but i need to include zstd...

The issue is that we have a copy of `zstd` under `third_party/`, but `basisu` also contains a partial bundled copy. Introducing `zstd` as a dependency to `filamat` causes targets to link against both copies simultaneously, resulting in duplicate symbol errors.

I believe the correct fix is to standardize on a single copy of `zstd` across the project. We should update `basisu` to link against the version in `third_party/` instead of its own. @pixelflinger If you agree with this approach, I can file a PR to fix it and unblock Yein. WDYT?

pixelflinger:
> > i see """ ld: warning: duplicate symbol '_ZSTD_compressBlock_lazy2_dedicatedDictSearch' in: /Users/yeinj/filament/out/cmake-release/third_party/basisu/tnt/libzstd-basisu.a[2](zstd.c.o) /Users/yeinj/filament/out/cmake-release/third_party/zstd/tnt/libzstd.a[19](zstd_lazy.c.o) """ when running the build script, how can i fix this? but i need to include zstd...
> 
> The issue is that we have a copy of `zstd` under `third_party/`, but `basisu` also contains a partial bundled copy. Introducing `zstd` as a dependency to `filamat` causes targets to link against both copies simultaneously, resulting in duplicate symbol errors.
> 
> I believe the correct fix is to standardize on a single copy of `zstd` across the project. We should update `basisu` to link against the version in `third_party/` instead of its own. @pixelflinger If you agree with this approach, I can file a PR to fix it and unblock Yein. WDYT?

I'm wondering why filamat has a dependency on basisu? maybe that would be the first thing to look at.

Otherwise I agree with unifying the zstd libs, but when I ran into this myself, it ended-up being harder than I thought. I don't remember why.

Another option should be to link against zstd PRIVATEly.

libs/filamat/CMakeLists.txt
```diff
@@ -81,7 +83,7 @@ include_directories(${CMAKE_BINARY_DIR})
 add_library(${TARGET} STATIC ${HDRS} ${PRIVATE_HDRS} ${SRCS})
 target_include_directories(${TARGET} PUBLIC ${PUBLIC_HDR_DIR})
 set_target_properties(${TARGET} PROPERTIES FOLDER Libs)
-target_link_libraries(${TARGET} backend_headers shaders filabridge utils smol-v)
+target_link_libraries(${TARGET} backend_headers shaders filabridge utils smol-v zstd)
```
pixelflinger, line 86: try to add PRIVATE just after {TARGET}

trie94: i couldn't mix the plain and keyword `target_link_libraries`, so i had to update all other `target_link_libraries` to link `zstd` privately. though, this doesn't get rid of the duplicate symbol warning :/

i am not entirely sure whether PRIVATE keyword helps here. my understanding with PRIVATE is that it doesn't do transitive linking, so zstd is linked to filamat, but if filament depends on basisu that links zstd, we would still get the duplicate symbol issue?

show50726:
> > > i see """ ld: warning: duplicate symbol '_ZSTD_compressBlock_lazy2_dedicatedDictSearch' in: /Users/yeinj/filament/out/cmake-release/third_party/basisu/tnt/libzstd-basisu.a[2](zstd.c.o) /Users/yeinj/filament/out/cmake-release/third_party/zstd/tnt/libzstd.a[19](zstd_lazy.c.o) """ when running the build script, how can i fix this? but i need to include zstd...
> > 
> > 
> > The issue is that we have a copy of `zstd` under `third_party/`, but `basisu` also contains a partial bundled copy. Introducing `zstd` as a dependency to `filamat` causes targets to link against both copies simultaneously, resulting in duplicate symbol errors.
> > I believe the correct fix is to standardize on a single copy of `zstd` across the project. We should update `basisu` to link against the version in `third_party/` instead of its own. @pixelflinger If you agree with this approach, I can file a PR to fix it and unblock Yein. WDYT?
> 
> I'm wondering why filamat has a dependency on basisu? maybe that would be the first thing to look at.
> 
> Otherwise I agree with unifying the zstd libs, but when I ran into this myself, it ended-up being harder than I thought. I don't remember why.
> 
> Another option should be to link against zstd PRIVATEly.

Just took a through look, it seems at least `imageio` is depending on `basis_encoder`, which depends on `zstd`. And the duplicate symbol conflict at least happens in backend test target, which depends on `imageio` and `filamat` at the same time.

libs/filamat/CMakeLists.txt
show50726: Since `zstd` is a static library, setting it to `PRIVATE` may not help. 

Another way to fix it is making `zstd` a shared library, but I believe unifying them to a single source might be a better fix.

trie94:
> > > > i see """ ld: warning: duplicate symbol '_ZSTD_compressBlock_lazy2_dedicatedDictSearch' in: /Users/yeinj/filament/out/cmake-release/third_party/basisu/tnt/libzstd-basisu.a[2](zstd.c.o) /Users/yeinj/filament/out/cmake-release/third_party/zstd/tnt/libzstd.a[19](zstd_lazy.c.o) """ when running the build script, how can i fix this? but i need to include zstd...
> > > 
> > > 
> > > The issue is that we have a copy of `zstd` under `third_party/`, but `basisu` also contains a partial bundled copy. Introducing `zstd` as a dependency to `filamat` causes targets to link against both copies simultaneously, resulting in duplicate symbol errors.
> > > I believe the correct fix is to standardize on a single copy of `zstd` across the project. We should update `basisu` to link against the version in `third_party/` instead of its own. @pixelflinger If you agree with this approach, I can file a PR to fix it and unblock Yein. WDYT?
> > 
> > 
> > I'm wondering why filamat has a dependency on basisu? maybe that would be the first thing to look at.
> > Otherwise I agree with unifying the zstd libs, but when I ran into this myself, it ended-up being harder than I thought. I don't remember why.
> > Another option should be to link against zstd PRIVATEly.
> 
> Just took a through look, it seems at least `imageio` is depending on `basis_encoder`, which depends on `zstd`. And the duplicate symbol conflict at least happens in backend test target, which depends on `imageio` and `filamat` at the same time.

i will wait until https://github.com/google/filament/pull/9493 lands. in the meantime, can y'all review the implementation?

libs/filamat/CMakeLists.txt
trie94: i reverted this change, since the correct fix is merged

libs/filamat/src/eiff/CompressedStringChunk.h
```diff
@@ -0,0 +1,46 @@
...
+
+namespace filamat {
+
+class CompressedStringChunk final : public Chunk {
+public:
+    enum class CompressionLevel { MIN, MAX, DEFAULT };
+    CompressedStringChunk(
+            ChunkType type, std::string_view string, CompressionLevel compressionLevel)
+            : Chunk(type),
+              mString(utils::CString(string.data())),
```
pixelflinger:
```suggestion
              mString(utils::CString(string.data(), string.size())),
```

libs/filamat/src/eiff/CompressedStringChunk.cpp
```diff
@@ -0,0 +1,54 @@
...
+        case CompressedStringChunk::CompressionLevel::DEFAULT:
+            return ZSTD_defaultCLevel();
+    }
+}
+} // namespace
+
+void CompressedStringChunk::flatten(filamat::Flattener& f) {
+    const size_t bufferBound = ZSTD_compressBound(mString.size());
+    std::vector<std::uint8_t> compressed;
+    compressed.reserve(bufferBound);
```
pixelflinger:
```suggestion
    std::vector<std::uint8_t> compressed(bufferBound);
```

libs/filamat/src/eiff/CompressedStringChunk.cpp
```diff
@@ -0,0 +1,54 @@
...
+
+    const size_t compressedSize = ZSTD_compress(
+            compressed.data(), bufferBound,
```
pixelflinger:
```suggestion
            compressed.data(), compressed.size(),
```

trie94: actually for this one `bufferBound` is correct. basically,
1) get the capacity (which is bufferBound)
2) pass that capacity to the ZSTD_compress
3) then get the actual compressed size (or an error)
4) use that actual size

pixelflinger: Well, `bufferBound` is the size now, so `compressed.size()` and `bufferBound` are equal, because:
```
std::vector<std::uint8_t> compressed(bufferBound);
```
sets the vector's size -- it's like doing a `push_back()` `bufferBound` times.

Before, with the `reserve` call, the code was incorrect because you were using `data()` with an empty vector (it had a capacity, but it was empty).
The C++ standard doesn't guarantee that `data()` returns a valid pointer if `size()` is 0, even if `capacity()` is not zero.

PR #9487, opened 2025-12-04
material: remove checked-in .filamat for transform name test

poweifeng:
filament_test_material should not need to depend on a "fixed" (or checked in) material.

Now the dependent .filamat is generated as needed as is packaged with the checked-in test_material.filamat as before.


z3moon (approved):

poweifeng:
> Delete `filament/test/test_material_transformname.filamat`?

thanks.  deleted.

PR #9488, opened 2025-12-04
Allow PlatformEGLHeadless to load GLES

PatRibas:
PlatformEGLHeadless only loads regular OpenGL, not GLES. This change allows PlatformEGLHeadless to also use GLES to help write tests against FL0.

BUGS = [462485141, 449755952]


filament/backend/src/opengl/platforms/PlatformEGL.cpp
```diff
@@ -179,7 +179,7 @@ Driver* PlatformEGL::createDriver(void* sharedContext, const DriverConfig& drive
         requestES2Context = bool(atoi(property));
     }
 #else
-    constexpr bool requestES2Context = false;
+    bool requestES2Context = isOpenGL();
```
pixelflinger: I don't understand this logic here. 

1. the only thing `requestES2Context` does is to set `        contextAttribs[EGL_CONTEXT_OPENGL_BACKWARDS_COMPATIBLE_ANGLE] = EGL_FALSE;` which would only work with ANGLE and only if the `EGL_ANGLE_create_context_backwards_compatible` extension was present. I don't understand why you would need that.

2. because it is set only when `isOpenGL()` is true, that means that we're actually requesting an OpenGL 4.1 context. Is this the intended behavior?

So the code means this:

"if we're not on ANDROID and using OpenGL (as opposed to GLES) then request a GL 4.1 context from ANGLE if it supports it".

I don't understand what this buys you.

filament/backend/src/opengl/platforms/PlatformEGLHeadless.cpp
```diff
@@ -35,16 +35,31 @@ PlatformEGLHeadless::PlatformEGLHeadless() noexcept
 }
 
 bool PlatformEGLHeadless::isOpenGL() const noexcept {
-    return  true;
+#if defined(BACKEND_OPENGL_VERSION_GL)
```
pixelflinger, line 38: Okay this is fine. This means we're compiling with OpenGL headers (as opposed to OpenGLES headers).

filament/backend/src/opengl/platforms/PlatformEGL.cpp
pixelflinger: I think you can leave this the way it was. The variable name is confusing, it's not actually requesting an ES2 context, that is done on line 193 based on `isOpenGL()`.

pixelflinger:
Also please provide a better description for the change.

filament/backend/src/opengl/platforms/PlatformEGL.cpp
PatRibas: Done

PatRibas:
> Also please provide a better description for the change.

Added a description to the merge comment in Github.

bejado (approved):
LGTM. Maybe also add a `BUGS = [b/xyz]` to the PR description to associate with a Buganizer bug.

z3moon (approved):

filament/backend/src/opengl/gl_headers.h
```diff
@@ -37,7 +37,7 @@
  */
 
 
-#if defined(__ANDROID__) || defined(FILAMENT_USE_EXTERNAL_GLES3) || defined(__EMSCRIPTEN__)
```
aWilson41, line 40: Doesn't this mean EGL only ever works with GLES?

Tried to build headless but can't because it keeps giving me GLES.

Or is there some other pathway that supposed to be used to use EGL with regular GL? Or is EGL just not supported in filament for regular OpenGL?

New to filament just trying to figure out why it won't work.

PatRibas: Hmm, you should be able to still use regular GL with these changes, but reading this again it's possible something got mucked up. If you're unable to get it to work with `BACKEND_OPENGL_VERSION_GL` defined and need to make changes, feel free to open a PR and add anyone you see approving changes as a reviewer. Wonder if this is supposed to be something like `|| (defined(FILAMENT_SUPPORTS_EGL_ON_LINUX) && !defined(BACKEND_OPENGL_VERSION_GL)`.

aWilson41: I took a slightly deeper look and it looks like it goes a bit deeper than that since it branches off into the bluegl code and it doesn't seem to have any edge cases with EGL support or anything. Though the initialization is probably very similar I don't think that code is shared unfortunately and it seemed like much more work to remove GLX and X11 includes to get a EGL+GL without any GLX or X11. Prob just going to stick with Vulkan in this case.

PR #9490, opened 2025-12-05
improve the memcpy benchmark

pixelflinger:
- remove the multi-threaded tests, they added too much noise
- added a memset test
- make sure we allocate the buffer aligned to 16KB

### DIFF
--- a/libs/utils/benchmark/benchmark_memcpy.cpp
+++ b/libs/utils/benchmark/benchmark_memcpy.cpp
@@ -18,25 +18,125 @@
 
 #include <benchmark/benchmark.h>
 
+#include <utils/memalign.h>
 
-static void BM_memcpy(benchmark::State& state) {
-    char* src = new char[state.range(0)];
-    char* dst = new char[state.range(0)];
-    memset(src, 'x', (size_t)state.range(0));
+#include <cstddef>
+#include <cstdint>
+#include <cstdlib>
+#include <cstring>
+#include <vector>
 
-    {
-        PerformanceCounters pc(state);
-        for (auto _ : state) {
-            memcpy(dst, src, (size_t)state.range(0));
-            benchmark::DoNotOptimize(dst);
-            benchmark::DoNotOptimize(src);
-            benchmark::ClobberMemory();
-        }
-    }
-    state.SetBytesProcessed(int64_t(state.iterations()) * int64_t(state.range(0)));
+// ------------------------------------------------------------------------------------------------
 
-    delete[] src;
-    delete[] dst;
+#ifdef __linux__
+#include <sched.h>
+#include <unistd.h>
+#include <sys/syscall.h>
+#endif
+
+// Returns true if successful, false if failed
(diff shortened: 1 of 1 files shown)


libs/utils/benchmark/benchmark_memcpy.cpp
```diff
@@ -18,25 +18,127 @@
...
+static constexpr size_t ALIGNMENT = 16 << 10;
+
+template<class T, size_t Align>
+struct AlignedAllocator {
+    using value_type = T;
+
+    template<class U>
+    struct rebind {
+        using other = AlignedAllocator<U, Align>;
+    };
```
poweifeng, line 64: unused?

poweifeng (approved):

libs/utils/benchmark/benchmark_memcpy.cpp
pixelflinger: yeah, but needed, doesn't compile without it.

libs/utils/benchmark/benchmark_memcpy.cpp
```diff
@@ -18,25 +18,127 @@
...
+    cpu_set_t cpuset;
+    CPU_ZERO(&cpuset);
+    CPU_SET(core_id, &cpuset);
+
+    // 0 = current thread
+    int const result = sched_setaffinity(0, sizeof(cpu_set_t), &cpuset);
+    return (result == 0);
+#else
+    (void)core_id;
+    return -1;
```
z3moon: this will be evaluated as true. just returns true or false?
but the callers don't use this function (commented out), so maybe just remove this function?

libs/utils/benchmark/benchmark_memcpy.cpp
```diff
@@ -18,25 +18,127 @@
...
+        return nullptr;
+    }
+
+    void deallocate(T* p, size_t) noexcept {
+        std::free(p);
+    }
+};
+
+template<typename T>
+using PageAlignedVector = std::vector<T, AlignedAllocator<char, ALIGNMENT>>;
```
z3moon:
```suggestion
using PageAlignedVector = std::vector<T, AlignedAllocator<T, ALIGNMENT>>;
```

libs/utils/benchmark/benchmark_memcpy.cpp
```diff
@@ -18,25 +18,127 @@
...
+// ------------------------------------------------------------------------------------------------
+
+#ifdef __linux__
+#include <sched.h>
+#include <unistd.h>
+#include <sys/syscall.h>
+#endif
+
+// Returns true if successful, false if failed
+static bool pinThreadToCore(int core_id) {
```
bejado, line 38: Do we still need this function? Seems commented-out everywhere.

pixelflinger: yeah, it's so that we can manually use it for local benchmark. it's complicated to make it work for all platforms.

PR #9491, opened 2025-12-05
VK: Do memcpy if the uniform buffer is WRITE_SHARED

rafadevai:
With UBO batching, the frontend guarantees that doing a memcpy to the buffer if safe.

BUGS=[448181318]


filament/backend/src/vulkan/VulkanBufferProxy.cpp
```diff
@@ -62,8 +62,7 @@ void VulkanBufferProxy::loadFromCpu(VulkanCommandBuffer& commands, const void* c
     // a memcpy if its available.
     bool const isStaticOrShared =
             any(mUsage & (BufferUsage::STATIC | BufferUsage::SHARED_WRITE_BIT));
-    bool const useMemcpy = (isUniformAndAvailable || isStaticOrShared) && isMemcopyable &&
-                           mStagingBufferBypassEnabled;
+    bool const useMemcpy = ((isUniformAndAvailable && mStagingBufferBypassEnabled) || isStaticOrShared) && isMemcopyable;
```
poweifeng: nit: line is a little long

rafadevai: Done

PR #9496, opened 2025-12-08
Fix(FrameInfo): Improve robustness of frame info history management

pixelflinger:
Address potential issues in FrameInfoImpl and FrameInfoManager:
 - Modify FrameInfoManager::beginFrame to handle cases where the  circular queue is full and the oldest frame is not yet ready,  logging a warning and skipping the frame.  This prevents potential use-after-free or data corruption by  ensuring only ready frames are removed from the history.

FIXES=[466081317]

### DIFF
--- a/filament/src/FrameInfo.cpp
+++ b/filament/src/FrameInfo.cpp
@@ -101,11 +101,22 @@ void FrameInfoManager::beginFrame(FSwapChain* swapChain, DriverApi& driver,
     auto& history = mFrameTimeHistory;
     // don't exceed the capacity, drop the oldest entry
     if (UTILS_LIKELY(history.size() == history.capacity())) {
-        if (!mDisableGpuFrameComplete) {
-            assert_invariant(history.back().fence);
-            driver.destroyFence(std::move(history.back().fence));
+        FrameInfoImpl& frameInfo = history.back();
+        if (frameInfo.ready.load(std::memory_order_relaxed)) {
+            if (!mDisableGpuFrameComplete) {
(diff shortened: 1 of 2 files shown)


filament/src/FrameInfo.cpp
```diff
@@ -101,11 +101,22 @@ void FrameInfoManager::beginFrame(FSwapChain* swapChain, DriverApi& driver,
...
+                driver.destroyFence(std::move(frameInfo.fence));
+            }
+            history.pop_back();
+        } else {
+            // This is a big problem, we ran out of space in the circular queue and that entry
+            // hasn't been processed yet. Because the code below keeps a reference to the
+            // front element of the queue, we can't pop/push. Our only option is to not record
+            // a new entry for this frame, which will create a false skipped frame in the
+            // data.
+            LOG(WARNING) << "FrameInfo's circular queue is full, but the latest item hasn't "
```
z3moon, line 117:
```suggestion
            LOG(WARNING) << "FrameInfo's circular queue is full, but the oldest item hasn't "
```

filament/src/FrameInfo.cpp
```diff
@@ -101,11 +101,22 @@ void FrameInfoManager::beginFrame(FSwapChain* swapChain, DriverApi& driver,
...
+                            " been processed yet. Skipping this frame, id = " << frameId;
         }
```
z3moon, line 119: shouldn't we actually skip this frame by `return;` or something? otherwise it looks like it'll crash at `history.emplace_front(frameId);` right below.

PR #9497, opened 2025-12-08
vk: remove precondition for createStreamNative

poweifeng:
Keep it's original behavior of no-op


filament/backend/src/vulkan/VulkanDriver.cpp
```diff
@@ -1156,7 +1154,6 @@ void VulkanDriver::destroyDescriptorSet(Handle<HwDescriptorSet> dsh) {
 }
```
pixelflinger, line 1157: that seems wrong that we have to remove this. that means someone depends on it not working. At least, we should have LOG(ERROR) or LOG(WARNING).

poweifeng: yeah, it's an internal test.  I can switch it to LOG(WARNING) for now.

poweifeng: done

PR #9503, opened 2025-12-09
Add missing VulkanDriver.h header

bejado:


poweifeng (approved):

poweifeng:
This might not be needed after this clean-up PR: https://github.com/google/filament/pull/9464

bejado:
> This might not be needed after this clean-up PR: #9464

Got it, looks like that include was deleted. Thanks!

PR #9504, opened 2025-12-10
a series of CL to add RenderTarget support for external textures

pixelflinger:


android/samples/sample-texture-target/src/main/res/values/strings.xml
```diff
@@ -0,0 +1,3 @@
+<resources>
+  <string name="app_name">Texture View</string>
```
z3moon: probably "Texture Target"

android/filament-android/src/main/cpp/Texture.cpp
```diff
@@ -388,6 +393,58 @@ Java_com_google_android_filament_Texture_nSetExternalImage(JNIEnv*, jclass, jlon
...
+        auto* vulkanPlatform = (VulkanPlatformAndroid*) platform;
+        auto ref = vulkanPlatform->createExternalImage(nativeBuffer, false);
+        texture->setExternalImage(*engine, ref);
+    }
+#endif // FILAMENT_SUPPORTS_VULKAN
+    // success!
+    return JNI_TRUE;
+#else
+    // other platforms could come here
+#endif // __ANDROID__
```
z3moon: I think the `return JNI_FALSE` below should be inside this `#else - #endif` block here. otherwise the code will be like `return JNI_TRUE;` and `return JNI_FALSE;` together for `__ANDROID__`

android/samples/sample-texture-target/src/main/res/values/strings.xml
pixelflinger: done

android/filament-android/src/main/cpp/Texture.cpp
pixelflinger: done

android/build.gradle
```diff
@@ -124,6 +124,7 @@ buildscript {
 
     ext.cmakeArgs = [
         "--no-warn-unused-cli",
+        "-DANDROID_WEAK_API_DEFS=ON",
```
poweifeng, line 127: q: what is this for?  just wondering if we need to make the same change in our internal build file.

pixelflinger: this is so that we can use `if (__builtin_available(android 26, *)) {`

PR #9507, opened 2025-12-11
EMSCRIPTEN always supports fences, even at feature level 0

pixelflinger:
FIXES=[467783405]

### DIFF
--- a/filament/backend/src/opengl/OpenGLDriver.cpp
+++ b/filament/backend/src/opengl/OpenGLDriver.cpp
@@ -1790,16 +1790,16 @@ void OpenGLDriver::createFenceR(Handle<HwFence> fh, ImmutableCString&& tag) {
     GLFence* const f = handle_cast<GLFence*>(fh);
     assert_invariant(f->state);
 
-    bool const platformCanCreateFence = mPlatform.canCreateFence();
-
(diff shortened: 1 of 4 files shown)


pixelflinger:
@PatRibas

filament/backend/src/opengl/OpenGLDriver.cpp
```diff
@@ -2331,30 +2333,30 @@ FenceStatus OpenGLDriver::fenceWait(FenceHandle fh, uint64_t const timeout) {
...
+        std::unique_lock lock(f->state->lock);
+        if (f->fence == nullptr) {
+            // we've been called before the fence was created asynchronously,
+            // so we need to wait for that, before using the real fence.
+            // By construction, "f" can't be destroyed while we wait, because its
+            // construction call is in the queue and a destroy call will have to come later.
+            f->state->cond.wait_until(lock, until, [f] {
+                return f->fence != nullptr;
+            });
             if (f->fence == nullptr) {
```
poweifeng, line 2346: this should already be true here?

pixelflinger: what do you mean? either the condition returned because f->fence wasn't null, or it returned because it timed-out?

PR #9508, opened 2025-12-11
Implement asynchronous resource management

z3moon:
This commit introduces async methods and functions across both the public and backend APIs, enabling non-blocking creation and updates for Texture, VertexBuffer, and IndexBuffer resources.

Asynchronous resource management is disabled by default. Users should enable this to use the feature.

Java bindings and a C++ code sample will follow in subsequent PRs.

BUGS=[442921995]


filament/include/filament/Engine.h
```diff
@@ -1023,6 +1024,38 @@ class UTILS_PUBLIC Engine {
     size_t getRenderTargetCount() const noexcept;
     /**  @} */
 
+    /**
+     * An asynchronous version of `CommandStream::queueCommand()`.
```
bejado: `CommandStream::queueCommand` is private, yes? I think if `queueCommandAsync` is meant to be used publicly, we shouldn't refer to a private API in its documentation.

filament/include/filament/IndexBuffer.h
```diff
@@ -131,7 +150,7 @@ class UTILS_PUBLIC IndexBuffer : public FilamentAPI {
     };
 
     /**
-     * Asynchronously copy-initializes a region of this IndexBuffer from the data provided.
+     * copy-initializes a region of this IndexBuffer from the data provided.
```
bejado:
```suggestion
     * Copy-initializes a region of this IndexBuffer from the data provided.
```

filament/include/filament/IndexBuffer.h
```diff
@@ -141,12 +160,45 @@ class UTILS_PUBLIC IndexBuffer : public FilamentAPI {
      */
     void setBuffer(Engine& engine, BufferDescriptor&& buffer, uint32_t byteOffset = 0);
 
+    /**
+     * An asynchronous version of `setBuffer()`.
+     * copy-initializes a region of this IndexBuffer from the data provided.
```
bejado:
```suggestion
     * Copy-initializes a region of this IndexBuffer from the data provided.
```

filament/include/filament/Texture.h
```diff
@@ -261,6 +264,26 @@ class UTILS_PUBLIC Texture : public FilamentAPI {
          */
         Builder& external() noexcept;
 
+        /**
+         * Specifies a callback that will execute once the resource's data has been fully allocated
+         * within the GPU memory. This enables the resource creation process to be handled
+         * asynchronously.
+         *
+         * To use this method, the engine must be configured for asynchronous operation. Otherwise,
+         * the program will crash or fail.
```
bejado: What's the difference between crashing and failing? I think we should be specific here.

filament/include/filament/Texture.h
```diff
@@ -547,6 +646,17 @@ class UTILS_PUBLIC Texture : public FilamentAPI {
      */
     void generateMipmaps(Engine& engine) const noexcept;
 
+    /**
+     * This non-blocking method checks if the resource has finished creation. If the resource
+     * creation was initiated asynchronously, it will return true only after all related
+     * asynchronous tasks are complete.
```
bejado: I think we should document what this method returns if the resource was _not_ created asynchronously. In that scenario, it should probably always return true.

filament/include/filament/Texture.h
```diff
@@ -261,6 +264,26 @@ class UTILS_PUBLIC Texture : public FilamentAPI {
          */
         Builder& external() noexcept;
 
+        /**
+         * Specifies a callback that will execute once the resource's data has been fully allocated
+         * within the GPU memory. This enables the resource creation process to be handled
+         * asynchronously.
```
bejado, line 271: We should document what happens if this `Texture` (and likewise for `VertexBuffer`, `IndexBuffer`, etc) is used before the async creation process has finished. Is it an error?

filament/backend/src/DriverBase.h
```diff
@@ -214,6 +221,13 @@ class DriverBase : public Driver {
 protected:
     class CallbackDataDetails;
 
+    [[nodiscard]] Platform::DriverConfig const& getDriverConfig() const noexcept {
+        return mDriverConfig;
+    }
+
+    [[nodiscard]] JobQueue* getJobQueue() const noexcept { return mJobQueue.get(); }
+    [[nodiscard]] JobWorker* getJobWorker() const noexcept { return mJobWorker.get(); }
```
poweifeng: does this need to be part of the driverBase? I had imagine that it would only be for gl backend.

filament/backend/include/private/backend/DriverAPI.inc
```diff
@@ -254,6 +295,20 @@ DECL_DRIVER_API_TAGGED_R_N(backend::TextureHandle, importTexture,
         uint32_t, depth,
         backend::TextureUsage, usage)
 
+DECL_DRIVER_API_TAGGED_R_N(backend::TextureHandle, importTextureAsync,
```
pixelflinger, line 298: Just to be sure I understand, this actually wouldn't need to be async (not doing anything expensive), but we have async version so that it can be queued with the async texture creation, correct?

filament/backend/include/private/backend/Driver.h
```diff
@@ -70,6 +70,8 @@ class Driver {
     // is where the driver can execute user callbacks.
     virtual void purge() noexcept = 0;
 
+    virtual void scheduleCallback(CallbackHandler* handler, void* user, CallbackHandler::Callback callback) = 0;
```
pixelflinger, line 76: maybe add a comment about what this is for.

filament/backend/src/opengl/OpenGLContext.cpp
```diff
@@ -926,7 +926,7 @@ void OpenGLContext::bindBuffer(GLenum target, GLuint buffer) noexcept {
         size_t const targetIndex = getIndexForBufferTarget(target);
         update_state(state.buffers.genericBinding[targetIndex], buffer, [&]() {
             glBindBuffer(target, buffer);
-        });
+        }, true); // TODO: split the GL state for shared contexts
```
pixelflinger: This will affect the performance of the GL backend even when async mode is not used, and also for non-async objects. I don't think we can check it in this way.

filament/backend/src/opengl/OpenGLDriver.cpp
```diff
@@ -616,6 +637,30 @@ Handle<HwTexture> OpenGLDriver::importTextureS() noexcept {
     return initHandle<GLTexture>();
 }
 
+Handle<HwTexture> OpenGLDriver::importTextureAsyncS() noexcept {
+    return initHandle<GLTexture>();
+}
+
+AsyncCallId OpenGLDriver::update3DImageAsyncS() noexcept {
+    FILAMENT_CHECK_PRECONDITION(getJobQueue()) << "Engine not configured for async operations";
```
pixelflinger: I think here we should only have `assert_invariant`, the `FILAMENT_CHECK` should only exist on the filament side. i.e. we should check that the client doesn't make async calls when async is not enabled, as early as possible. After that, if we get here that would mean a bug on *our* side, and these should not use `FILAMENT_CHECK`.

filament/backend/src/opengl/OpenGLDriver.cpp
```diff
@@ -697,17 +741,60 @@ void OpenGLDriver::createIndexBufferR(
...
+
+    getJobQueue()->push([this, ibh, elementType, indexCount, usage, handler, callback, user,
+            tag=std::move(tag)]() mutable {
+        DEBUG_MARKER_NAME("createIndexBufferAsyncR")
+        createIndexBufferCommon(ibh, elementType, indexCount, usage, std::move(tag));
+        // glFlush() should be called when using a shared context for this operation. Without it,
+        // the driver may delay submitting commands to the GPU, preventing other contexts from
+        // seeing the changes immediately. This ensures submitting the current commands right away.
+        glFlush();
```
pixelflinger, line 803: That's an interesting problem; can you add something about it in your design doc. I can't think about a better way right now but a few comments:
- doing a glFlush for every command might be very costly
- instead of a flush we could maybe use a gl fence and use a wait(0) on the other side?
- I'm actually not even sure that the gl spec has to do anything on a flush; maybe implementations don't do anything. And using a fence might be the only safe way. I think we should look into this more.

filament/backend/src/opengl/OpenGLDriver.cpp
```diff
@@ -1994,49 +2225,57 @@ void OpenGLDriver::destroyTexture(Handle<HwTexture> th) {
...
             }
+            destruct(th, t);
+        };
+
+        if (t->asynchronous) {
+            getJobQueue()->push(std::move(destroyTextureCommon));
         } else {
-            gl.unbindTexture(t->gl.target, t->gl.id);
+            destroyTextureCommon();
         }
```
pixelflinger, line 2300: I want to make sure that this pattern doesn't inline `destroyCommon` (code bloat). It might be better to extract it into its own method (instead of lambda) and mark it NO_INLINE.

filament/backend/src/DriverBase.h
```diff
@@ -72,22 +74,24 @@ struct HwVertexBuffer : public HwBase {
 };
 
 struct HwBufferObject : public HwBase {
-    uint32_t byteCount{};
+    uint32_t byteCount : 31;
+    uint32_t asynchronous : 1;
```
pixelflinger, line 76: 👍

filament/backend/src/DriverBase.h
```diff
@@ -72,22 +74,24 @@ struct HwVertexBuffer : public HwBase {
...
-    explicit HwBufferObject(uint32_t byteCount) noexcept : byteCount(byteCount) {}
+    explicit HwBufferObject(uint32_t byteCount, bool async) noexcept : byteCount(byteCount), asynchronous(async) {}
 };
 
 struct HwMemoryMappedBuffer : public HwBase {
 };
 
 struct HwIndexBuffer : public HwBase {
-    uint32_t count : 27;
+    uint32_t count : 26;
```
pixelflinger:
```suggestion
    uint32_t count : 26; // 67M indices
```

filament/backend/src/DriverBase.h
pixelflinger: I don't think these need to be [[nodiscard]] nothing bad happens if you discard them (it's not like malloc).

filament/backend/src/vulkan/VulkanDriver.cpp
```diff
@@ -601,6 +603,12 @@ void VulkanDriver::createIndexBufferR(Handle<HwIndexBuffer> ibh, ElementType ele
     mResourceManager.associateHandle(ibh.getId(), std::move(tag));
 }
 
+void VulkanDriver::createIndexBufferAsyncR(Handle<HwIndexBuffer> ibh, ElementType elementType,
+        uint32_t indexCount, BufferUsage usage, CallbackHandler* handler,
+        CallbackHandler::Callback callback, void* user, utils::ImmutableCString&& tag) {
+    // TODO: implement this.
```
pixelflinger, line 609: so don't we have a `isAsyncModeSupported` somewhere in the backends? (so that the front end can know if it can use it or not). I think we will need that if we want to be able to check it in without all backends supporting it.

And also a public `Engine::isAsyncSupported`

filament/include/filament/Engine.h
pixelflinger: This should be in FEngine. It's not a public API.

filament/include/filament/Texture.h
pixelflinger: To be even more specific, Builder methods don't do any checks. The method that will "crash" (`FILAMENT_CHECK`) is `Builder::build()`.

filament/include/filament/Texture.h
```diff
@@ -261,6 +264,26 @@ class UTILS_PUBLIC Texture : public FilamentAPI {
...
+         *
+         * To use this method, the engine must be configured for asynchronous operation. Otherwise,
+         * the program will crash or fail.
+         *
+         * This method and the `external()` method are mutually exclusive. You cannot use both
+         * because external texture's contents are filled later by calling `setExternalImage()`.
```
pixelflinger, line 283: the precondition check for this will be done in `build()`

filament/include/filament/VertexBuffer.h
```diff
@@ -61,6 +61,9 @@ class UTILS_PUBLIC VertexBuffer : public FilamentAPI {
 public:
     using AttributeType = backend::ElementType;
     using BufferDescriptor = backend::BufferDescriptor;
+    using AsyncCallbackType = std::function<void(VertexBuffer* UTILS_NONNULL, void* UTILS_NULLABLE)>;
```
pixelflinger, line 65: you probably need `<functional>`

filament/src/details/AsyncHelpers.h
```diff
@@ -0,0 +1,143 @@
...
+        // 6. Executes the *REAL* user's callback.
+        if (thisObject->mUserCallback) {
+            thisObject->mUserCallback(thisObject->mUserParam1, thisObject->mUserParam2);
+        }
+        delete thisObject;
+    }
+
+    // Increase the countdown. Users are required to call this for each async call they make.
+    uint32_t increaseCountdown() {
+        auto prev = mCount.fetch_add(1, std::memory_order_relaxed);
```
pixelflinger, line 123: you need to explain why `memory_order_relaxed`. I think this usage is correct because the `inc` doesn't publish or acquires anything.

filament/src/details/AsyncHelpers.h
```diff
@@ -0,0 +1,143 @@
...
+        return prev + 1;
+    }
+
+private:
+    // Reduces the countdown and returns true upon reaching zero.
+    bool decreaseCountdown() {
+        auto prev = mCount.fetch_sub(1, std::memory_order_relaxed);
```
pixelflinger: you need to explain why `memory_order_relaxed`. I think this one is *not* correct because when `decreaseCountDown` reaches 0, it actually publishes `mUserCallback` and all its dependents. 
I believe the correct memory order here is `acq_rel`.  If you want to be fancy, you can do `relaxed`, but when it reaches 0, you need to use `atomic_thread_fence(acq_rel)`

filament/src/details/IndexBuffer.cpp
```diff
@@ -80,11 +94,42 @@ FIndexBuffer::FIndexBuffer(FEngine& engine, const Builder& builder)
...
+        auto copiedCompletionCallback = builder->mAsyncCreationCallback;
+
+        using IndexBufferCountdownCallbackHandler = CountdownCallbackHandler<IndexBuffer>;
+        auto* cdHandler = IndexBufferCountdownCallbackHandler::make(
+                /* handler */ builder->mAsyncCreationHandler,
+                /* userCallback */ std::move(copiedCompletionCallback),
+                /* userParam1 */ this,
+                /* userParam2 */ builder->mAsyncCreationUserData,
+                /* onCountdownComplete */ [this]() {
+                    mCreationComplete.store(true, std::memory_order_relaxed);
```
pixelflinger, line 116: you need to explain why `memory_order_relaxed`

filament/src/details/IndexBuffer.h
```diff
@@ -42,10 +42,20 @@ class FIndexBuffer : public IndexBuffer {
 
     void setBuffer(FEngine& engine, BufferDescriptor&& buffer, uint32_t byteOffset = 0);
 
+    AsyncCallId setBufferAsync(FEngine& engine, BufferDescriptor&& buffer, uint32_t byteOffset,
+            backend::CallbackHandler* handler, AsyncCallbackType callback, void* user);
+
+    bool isCreationComplete() const noexcept { return mCreationComplete.load(std::memory_order_relaxed); }
```
pixelflinger, line 48: Normally in filament all public API must be called from the main thread. If this one is different we should be *extremely* clear about it.
I don't think this needs to be atomic.

filament/src/details/IndexBuffer.cpp
```diff
@@ -80,11 +94,42 @@ FIndexBuffer::FIndexBuffer(FEngine& engine, const Builder& builder)
...
+                },
+                /* driver */ &engine.getDriver());
+
+        mHandle = driver.createIndexBufferAsync(
+                backend::ElementType(builder->mIndexType),
+                uint32_t(builder->mIndexCount),
+                backend::BufferUsage::STATIC,
+                cdHandler, &IndexBufferCountdownCallbackHandler::countdownCallback, cdHandler,
+                std::move(name));
+        cdHandler->increaseCountdown();
```
pixelflinger, line 126: `CountdownCallbackHandler` could always be created with a count of `1` so that we don't have to call `increaseCountdown`  just after creation.

filament/src/details/IndexBuffer.cpp
```diff
@@ -80,11 +94,42 @@ FIndexBuffer::FIndexBuffer(FEngine& engine, const Builder& builder)
...
+        // again to reuse the builder for an additional new object.
+        auto copiedCompletionCallback = builder->mAsyncCreationCallback;
+
+        using IndexBufferCountdownCallbackHandler = CountdownCallbackHandler<IndexBuffer>;
+        auto* cdHandler = IndexBufferCountdownCallbackHandler::make(
+                /* handler */ builder->mAsyncCreationHandler,
+                /* userCallback */ std::move(copiedCompletionCallback),
+                /* userParam1 */ this,
+                /* userParam2 */ builder->mAsyncCreationUserData,
+                /* onCountdownComplete */ [this]() {
```
pixelflinger, line 113: I'm worried about capturing `this` here. How do we know that the object hasn't been destroyed by the time the callback is invoked?

filament/src/details/IndexBuffer.cpp
```diff
@@ -64,6 +69,15 @@ IndexBuffer::Builder& IndexBuffer::Builder::name(utils::StaticString const& name
...
+        AsyncCallbackType callback, void* user) noexcept {
+    mImpl->mAsynchronous = true;
+    mImpl->mAsyncCreationHandler = handler;
+    mImpl->mAsyncCreationCallback = std::move(callback);
+    mImpl->mAsyncCreationUserData = user;
+    return *this;
+}
+
 IndexBuffer* IndexBuffer::Builder::build(Engine& engine) {
     return downcast(engine).createIndexBuffer(*this);
```
pixelflinger, line 85: This is where you should check `FILAMENT_CHECK` that async mode is enabled if `mAsynchronous` is `true`. All validation of the `Builder` should happen here.

filament/src/details/IndexBuffer.cpp
```diff
@@ -80,11 +94,42 @@ FIndexBuffer::FIndexBuffer(FEngine& engine, const Builder& builder)
             << "Invalid index type " << static_cast<int>(builder->mIndexType) << ", tag=" << tag;
 
     FEngine::DriverApi& driver = engine.getDriverApi();
-    mHandle = driver.createIndexBuffer(
-            backend::ElementType(builder->mIndexType),
-            uint32_t(builder->mIndexCount),
-            backend::BufferUsage::STATIC,
-            std::move(name));
+
```
pixelflinger, line 100: if you want, you could have an assert_invariant here to check that `mAsynchronous` matches the engine config.

filament/src/details/Material.cpp
```diff
@@ -15,6 +15,7 @@
  */
 
 #include "details/Material.h"
+
```
pixelflinger: change not needed

pixelflinger (changes requested):

filament/include/filament/Engine.h
z3moon: I update it as a new API. I found this is useful while writing a sample app to load images asynchronously in sequence. Later, it will be also useful for loading gltf assets and addressing various users' needs.

filament/include/filament/IndexBuffer.h
z3moon: done

filament/include/filament/IndexBuffer.h
z3moon: done

filament/include/filament/Texture.h
z3moon: done

filament/include/filament/Texture.h
z3moon: good point! done

filament/include/filament/Texture.h
z3moon: yes it's undefined behavior, I updated this.

filament/backend/include/private/backend/DriverAPI.inc
z3moon: Yes, you're correct. It doesn't necessarily have to be asynchronous, but I had to do it because it's part of the texture creation process.

filament/backend/include/private/backend/Driver.h
z3moon: done

filament/backend/src/opengl/OpenGLContext.cpp
z3moon: Yeah I removed them.

filament/backend/src/opengl/OpenGLDriver.cpp
z3moon: I see, I moved the checks to Builder::build, replacing these with assert_invariant

filament/backend/src/opengl/OpenGLDriver.cpp
z3moon: I updated the design doc with the details. wait may not work because the command queue will remain unflushed until new commands are queued.

filament/src/details/IndexBuffer.cpp
z3moon: I see what you mean. But it may look confusing because there will be no increase call only for the first asyn call. The consistency will make it clearer to users. WDYT?

filament/src/details/IndexBuffer.cpp
z3moon: Good point. A simple plan could be deferring the destruction until the creation is finished. Then they can be garbage-collected. Let me add another commit for it.

filament/src/details/IndexBuffer.cpp
z3moon: done

filament/src/details/IndexBuffer.cpp
z3moon: it's done in the build method

filament/src/details/Material.cpp
z3moon: done

filament/backend/src/DriverBase.h
z3moon: The plan is to use the thread worker for other backends as well. Resources for those backends can be created/updated in a separate thread for better performance.

poweifeng: and if they wanted to use more than one thread?

z3moon: yes, but if others can't use threads for some reasons, the amortization worker should still work.

poweifeng: right, and you have a class for that right? Wouldn't the backend just have an instance of it?
Or they could just advertise async not supported if threads are not there.

z3moon: I'd like to maintain a consistent interface across all backends if possible to keep the init code in one place and simplify future maintenance/expansion. If a specific backend has threading limitations, we can make the initialization fail for threading mode and advertise that it only supports the amortization for asynchronous methods. If I'm not wrong, both Vulkan and Metal support threads. Is your concern specifically for WebGPU?

poweifeng: my concern is more like other backends might want to manage their own threads. There's some variance here. I could imagine that this same thread, or pool could be used for readPixels on Vulkan. 

wrt webgpu should support thread, but I don't know what limitations there are for that.

z3moon: I see, yeah let's discuss this further in our chat. I assumed other backends will do the same (or similar) approach to create or update resources. But I might be wrong.

filament/backend/src/opengl/OpenGLDriver.cpp
z3moon: done

filament/backend/src/DriverBase.h
z3moon: done

filament/backend/src/DriverBase.h
z3moon: yup, removed them

filament/backend/src/vulkan/VulkanDriver.cpp
z3moon: Yeah we have `Engine::isAsynchronousOperationSupported()` (public api) to inform users whether they can use async methods.

filament/include/filament/Texture.h
z3moon: done

filament/include/filament/VertexBuffer.h
z3moon: done

filament/src/details/AsyncHelpers.h
z3moon: done

filament/src/details/AsyncHelpers.h
z3moon: To be honest, I'm not sure it needs in the current implementation because there are no reads/writes for member fields after its creation. But let me update it to be more conservative in case things may change later.

filament/src/details/IndexBuffer.cpp
z3moon: done

filament/src/details/IndexBuffer.h
z3moon: This is an atomic variable because mCreationComplete.store(...) is called from ServiceThread. It could've been posted to the main thread, but it would require more posting steps, leading to slower responsiveness in terms of checking `isCreationComplete()`.

filament/backend/src/DriverBase.h
z3moon: I moved those job-relevant member variables to OpenGLDriver.

filament/src/details/IndexBuffer.cpp
pixelflinger: This is how RefBase works on Android, it always create the object with a refcount of 1. Personally, I feel like that if you don't create the object with a refcount of one, it's like it's in "limbo", it exists, but there is no reference to it.  ymmv, but to me, it's weird. I think it's easy enough to document it. And it's slightly more efficient too.

pixelflinger (approved):
lgtm, except for the [this] capture that will be addressed in another CL.

filament/backend/src/opengl/OpenGLDriver.cpp
```diff
@@ -312,6 +315,28 @@ OpenGLDriver::OpenGLDriver(OpenGLPlatform* platform, const Platform::DriverConfi
 #endif
 
     mShaderCompilerService.init();
+
+    if (driverConfig.asynchronousMode != AsynchronousMode::NONE) {
+        mJobQueue = JobQueue::create();
+
+        bool useThreadWorker = false;
```
poweifeng: how about just
```
bool const useThreadWorker = driverConfig.asynchronousMode == AsynchronousMode::THREAD_PREFERRED &&
                UTILS_HAS_THREADING;
```

filament/backend/src/vulkan/VulkanDriver.cpp
poweifeng: but there is not a backend api?  Can you do maybe a follow-up to add one?

filament/backend/src/DriverBase.h
```diff
@@ -72,22 +72,24 @@ struct HwVertexBuffer : public HwBase {
 };
 
 struct HwBufferObject : public HwBase {
-    uint32_t byteCount{};
+    uint32_t byteCount : 31;
+    uint32_t asynchronous : 1;
```
poweifeng, line 76: can this be
```
bool asynchronous: 1; 
```

filament/include/filament/Engine.h
```diff
@@ -1023,6 +1024,39 @@ class UTILS_PUBLIC Engine {
...
+     * alongside other asynchronous operations (see Texture, VertexBuffer, and IndexBuffer) and
+     * guaranteed to be executed in the exact order they were invoked.
+     *
+     * Users can call the `Engine::cancelAsyncCall()` method with the returned ID to cancel the
+     * asynchronous call.
+     *
+     * To use this method, the engine must be configured for asynchronous operation. Otherwise,
+     * calling async method will cause the program to terminate.
+     *
+     * @param command The custom command to be executed.
```
poweifeng, line 1045: would it suffice to do away with "command"?  I imagine you just want to queue something empty in the async queue and have the `onComplete` be where the user do the work?  

Or do you expect user to actually do work on the async thread?

filament/src/details/IndexBuffer.cpp
```diff
@@ -80,11 +97,44 @@ FIndexBuffer::FIndexBuffer(FEngine& engine, const Builder& builder)
...
+    if (builder->mAsynchronous) {
+        // Copy the callback instead of moving it because moving would leave the builder with an
+        // empty completion callback. If the builder were emptied, users would need to call `async`
+        // again to reuse the builder for an additional new object.
+        auto copiedCompletionCallback = builder->mAsyncCreationCallback;
+
+        using IndexBufferCountdownCallbackHandler = CountdownCallbackHandler<IndexBuffer>;
+        auto* cdHandler = IndexBufferCountdownCallbackHandler::make(
```
poweifeng, line 108: what destroys this?

filament/src/details/VertexBuffer.cpp
```diff
@@ -344,30 +402,40 @@ size_t FVertexBuffer::getVertexCount() const noexcept {
...
+        AsyncCallbackType callback, void* user) {
+    FILAMENT_CHECK_PRECONDITION(!mBufferObjectsEnabled)
+            << "buffer objects enabled, use setBufferObjectAt() instead";
+    FILAMENT_CHECK_PRECONDITION(bufferIndex < mBufferCount)
+            << "bufferIndex must be < bufferCount";
+    FILAMENT_CHECK_PRECONDITION((byteOffset & 0x3) == 0)
+        << "byteOffset must be a multiple of 4";
+
+    using VertexBufferCallbackAdapter = CallbackAdapter<VertexBuffer>;
+        auto* const cbWrapper = VertexBufferCallbackAdapter::make(std::move(callback), this, user);
```
poweifeng: indentation here

filament/src/details/VertexBuffer.cpp
```diff
@@ -378,6 +446,30 @@ void FVertexBuffer::setBufferObjectAt(FEngine& engine, uint8_t const bufferIndex
...
+    FILAMENT_CHECK_PRECONDITION(bufferObject->getBindingType() == BufferObject::BindingType::VERTEX)
+            << "bufferObject binding type must be VERTEX but is "
+            << to_string(bufferObject->getBindingType());
+    FILAMENT_CHECK_PRECONDITION(bufferIndex < mBufferCount)
+            << "bufferIndex must be < bufferCount";
+
+    auto const hwBufferObject = bufferObject->getHwHandle();
+
+    using VertexBufferCallbackAdapter = CallbackAdapter<VertexBuffer>;
+        auto* const cbWrapper = VertexBufferCallbackAdapter::make(std::move(callback), this, user);
```
poweifeng: indentation here

filament/backend/src/DriverBase.h
z3moon: changing the type may break the bitfield packing and start a new packing in a certain compiler (especially MSVC), bloating the size unintentionally. I would not risk it.

filament/backend/src/opengl/OpenGLDriver.cpp
z3moon: done

filament/include/filament/Engine.h
z3moon: Yeah, I expect users to do something in the thread, like loading image, meshes, etc, before they perform the next operations with them.

filament/src/details/IndexBuffer.cpp
z3moon: this object is deleted inside the CountdownCallbackHandler class, I followed the existing pattern.

filament/src/details/VertexBuffer.cpp
z3moon: done

filament/src/details/VertexBuffer.cpp
z3moon: done

filament/src/details/Texture.cpp
```diff
@@ -317,30 +341,76 @@ FTexture::FTexture(FEngine& engine, const Builder& builder)
...
+        if (UTILS_LIKELY(!isImported)) {
+            mHandle = driver.createTextureAsync(
+                    mTarget, mLevelCount, mFormat, mSampleCount, mWidth, mHeight, mDepth, mUsage,
+                    cdHandler, &TextureCountdownCallbackHandler::countdownCallback, cdHandler, std::move(tag));
+            cdHandler->increaseCountdown();
+        } else {
+            mHandle = driver.importTextureAsync(builder->mImportedId,
+                    mTarget, mLevelCount, mFormat, mSampleCount, mWidth, mHeight, mDepth, mUsage,
+                    cdHandler, &TextureCountdownCallbackHandler::countdownCallback, cdHandler, std::move(tag));
+            cdHandler->increaseCountdown();
```
poweifeng, line 378: nit: I think you can introduce a method like:
```
CountdownCallbackHandler* CountdownCallbackHandler::ref() {
    increaseCountdown();
    return this;
}
```

and then in the async call it you would be
```
cdHandler->ref()
```

filament/src/details/IndexBuffer.cpp
poweifeng: just curious, can you point me to that pattern?

filament/include/filament/Engine.h
poweifeng: that's unusual though. The async thread is directly working on graphics api.  If the user don't want to block the main thread, wouldn't they spin up their own threads?

filament/backend/src/DriverBase.h
poweifeng: noted

filament/include/filament/Engine.h
z3moon: The idea is that it's not limited to loading only graphics resource. Users can take advantage of it if they need. This will be also used for loading gltf assets.

filament/src/details/IndexBuffer.cpp
z3moon: It's in the [Material::compile](https://github.com/google/filament/blob/main/filament/src/details/Material.cpp#L290)

filament/include/filament/Engine.h
poweifeng: Can you add a comment that mentions that the workload done on this thread will affect performance of the async calls?  I imagine the user could do all kinds of things on the async thread (reading from disk etc), and then the async uploads would suffer.

filament/src/details/IndexBuffer.cpp
poweifeng: noted

filament/backend/src/vulkan/VulkanDriver.cpp
z3moon: Ah yeah, let me add it as a follow-up.

filament/include/filament/Engine.h
z3moon: That is a valid point. Let add the comments to clarify.

filament/src/details/Texture.cpp
z3moon: Yeah I considered this approach. But the instance is used for both `handler` and `user`, which made me rather choose this explicit method.

PR #9509, opened 2025-12-11
Fix for ES2 mat3f uploads

PatRibas:
FIXES=[468092808]


z3moon:
there's a duplicate https://github.com/google/filament/pull/9510

PR #9510, opened 2025-12-12
fix mat33 in UBOs at feature level 0

pixelflinger:
When emulating UBOs for feature level 0, we need to use packed mat33 as source of glUniformMatrix3fv which doesn't expect the std140 alignments of UBOs.

FIXES=[468092808]

### DIFF
--- a/filament/backend/src/opengl/OpenGLProgram.cpp
+++ b/filament/backend/src/opengl/OpenGLProgram.cpp
@@ -39,6 +39,7 @@
 #include <new>
 #include <string_view>
 #include <utility>
+#include <vector>
 
 #include <stddef.h>
 #include <stdint.h>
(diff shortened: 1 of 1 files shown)


pixelflinger:
@PatRibas

filament/backend/src/opengl/OpenGLProgram.cpp
```diff
@@ -328,9 +329,40 @@ void OpenGLProgram::updateUniforms(
...
+                    explicit mat3(std140::mat3 const& other) noexcept
+                        : std::array<std::array<GLfloat, 3>, 3>({
+                            { other[0][0], other[0][1], other[0][2] },
+                            { other[1][0], other[1][1], other[1][2] },
+                            { other[2][0], other[2][1], other[2][2] }
+                        }) {
+                    }
+                };
+
+                std::vector<mat3> const temp{ b, b + u.size };
```
z3moon, line 360: I think we could use a local variable of array, not to allocate memory, if `u.size` is always small. Can it be large?

pixelflinger: we don't know how large it could be. but also I think it doesn't matter, that's the emulated ubo case, it's already incredibly inefficient and it's for GLES2, also only for mat3.

PR #9511, opened 2025-12-12
don't trigger an assert failure when getFenceStatus fails

pixelflinger:
this is not an error that should warrant an assertion, since it's possible for fences to not be supported at all.

FIXES=[467782111]

### DIFF
--- a/.github/actions/mac-prereq/action.yml
+++ b/.github/actions/mac-prereq/action.yml
@@ -15,7 +15,7 @@ runs:
       uses: actions/cache@v4 # Use a specific version
       with:
         path: ~/Library/Caches/Homebrew
-        key: ${{ runner.os }}-brew-20250424
+        key: ${{ runner.os }}-brew-20251211
     - name: Install Mac Prerequisites
       shell: bash
       run: |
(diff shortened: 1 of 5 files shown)


pixelflinger:
@PatRibas

filament/src/FrameSkipper.cpp
```diff
@@ -62,7 +62,10 @@ bool FrameSkipper::shouldRenderFrame(DriverApi& driver) const noexcept {
             // The fence hasn't signaled yet, skip this frame
             return false;
         }
-        assert_invariant(status == FenceStatus::CONDITION_SATISFIED);
+        // If we get a FenceStatus::ERROR, it doesn't necessarily indicate a "bug", it could
+        // just be that fences are not supported. Regardless, we should return `true` in that
+        // case.
+        assert_invariant(status != FenceStatus::TIMEOUT_EXPIRED);
```
z3moon, line 68: Do we need this? This assertion will be always true since the execution returns `false` above if status == TIMEOUT_EXPIRED.

pixelflinger: not strictly needed, it was more a way to indicate that this is en expectation at this point in the code -- eg in case we change the code above some day.

PR #9515, opened 2025-12-12
Add thread pool and callback manager / knobs for enabling + disabling prewarm

anishmgoyal:
Note: this doesn't add the cache prewarming logic. I'm aiming to break this into a few different CLs so they can be reviewed more easily. This just adds configurations, refactors the pipeline create method, and adds the thread pool to the pipeline cache class.


anishmgoyal:
@pixelflinger @poweifeng @phoenixxxx @rafadevai

filament/backend/src/vulkan/VulkanPipelineCache.cpp
```diff
@@ -87,9 +103,11 @@ VulkanPipelineCache::PipelineCacheEntry* VulkanPipelineCache::getOrCreatePipelin
         pipeline.lastUsed = mCurrentTime;
         return &pipeline;
     }
-    auto ret = createPipeline();
-    ret->lastUsed = mCurrentTime;
-    return ret;
+    PipelineCacheEntry cacheEntry {
+        .handle = createPipeline(mPipelineRequirements),
```
z3moon, line 107: `createPipeline` can return nullptr, which in turn could cause a crash.

filament/backend/src/vulkan/VulkanPipelineCache.h
```diff
@@ -86,14 +94,21 @@ class VulkanPipelineCache {
...
-    VulkanPipelineCache(VkDevice device, VulkanContext const& context);
+    VulkanPipelineCache(DriverBase& driver, VkDevice device, VulkanContext const& context);
 
-    void bindLayout(VkPipelineLayout layout) noexcept;
+    // Loads a fake pipeline into memory on a separate thread, with the intent of
+    // preloading the Vulkan cache with enough information to have a cache hit when
+    // compiling the pipeline on the main thread at draw time. This is very dependent
+    // on the implementation of the driver on the current device; it's expected to work
+    // on devices with VK_EXT_vertex_input_dynamic_state and VK_KHR_dynamic_rendering.
+    void asyncPreloadCache(fvkmemory::resource_ptr<VulkanProgram> program,
```
z3moon, line 107: This method has no implementation. Probably remove it?

z3moon (changes requested):

filament/backend/include/backend/Platform.h
```diff
@@ -347,6 +347,17 @@ class UTILS_PUBLIC Platform {
...
         bool vulkanEnableStagingBufferBypass = false;
 
+        /**
+         * Enables asynchronous pipeline cache preloading.
+         * This is only supported for:
+         *      - VulkanPlatform
+         * When the following device extensions are available:
+         *      - VK_KHR_dynamic_rendering
+         *      - VK_EXT_vertex_input_dynamic_state
+         * And is dependent on Vulkan driver implementation on the current device.
```
rafadevai: we should also say its only for Qualcomm GPU

filament/backend/src/vulkan/VulkanContext.h
```diff
@@ -179,13 +191,20 @@ struct VulkanContext {
 
     VkExternalFenceHandleTypeFlags mFenceExportFlags = {};
 
+    // These are options that can be enabled or disabled at an application level.
+    bool mAsyncPipelineCachePrewarmingEnabled = false;
```
rafadevai: Is better to say if prewarming is supported, I take the driver config more as a hint since depends on the hardware at the end.

filament/backend/src/vulkan/VulkanPipelineCache.cpp
```diff
@@ -107,27 +125,27 @@ void VulkanPipelineCache::bindPipeline(VulkanCommandBuffer* commands) {
     }
 }
 
-VulkanPipelineCache::PipelineCacheEntry* VulkanPipelineCache::createPipeline() noexcept {
-    assert_invariant(mPipelineRequirements.shaders[0] && "Vertex shader is not bound.");
-    assert_invariant(mPipelineRequirements.layout && "No pipeline layout specified");
+VkPipeline VulkanPipelineCache::createPipeline(const PipelineKey& key) noexcept {
```
rafadevai, line 129: whats the idea behind this change?

filament/backend/src/vulkan/platform/VulkanPlatform.cpp
```diff
@@ -722,6 +722,7 @@ Driver* VulkanPlatform::createDriver(void* sharedContext,
     VulkanContext& context = mImpl->mContext;
 
     // Pass along relevant driver config (feature flags)
+    context.mAsyncPipelineCachePrewarmingEnabled = driverConfig.vulkanEnableAsyncPipelineCachePrewarming;
```
rafadevai: The driverConfig in this case should be taken as a hint and then better say, if the prewarm is supported or not.

This should be true only if using a QC gpu, maybe we can also specify the deviceID and if we have the other 2 extensions.

filament/backend/src/vulkan/platform/VulkanPlatform.cpp
```diff
@@ -869,12 +870,17 @@ Driver* VulkanPlatform::createDriver(void* sharedContext,
     if (!mImpl->mSharedContext) {
```
rafadevai: can you also add these 2 extensions to the device extension list in this file?

filament/backend/src/vulkan/VulkanPipelineCache.cpp
```diff
@@ -66,13 +69,26 @@ void printPipelineFeedbackInfo(VkPipelineCreationFeedbackCreateInfo const& feedb
...
+VulkanPipelineCache::VulkanPipelineCache(DriverBase& driver, VkDevice device, VulkanContext const& context)
         : mDevice(device),
+          mCallbackManager(driver),
           mContext(context) {
     VkPipelineCacheCreateInfo createInfo = {
         .sType = VK_STRUCTURE_TYPE_PIPELINE_CACHE_CREATE_INFO,
     };
     bluevk::vkCreatePipelineCache(mDevice, &createInfo, VKALLOC, &mPipelineCache);
+
+    if (isAsyncPrewarmingSupported(context)) {
```
poweifeng: nit: this could be better named as Enabled as oppose to Supported.  

Enabled = Supported + Hinted (as mentioned by rafadevai)

filament/backend/src/vulkan/VulkanPipelineCache.cpp
anishmgoyal: that's not new behavior here - this was the case before this change as well. also, i'd argue that you'd want the application to crash if it cannot create a pipeline for drawing. would you prefer if this was wrapped in an assert?

filament/backend/src/vulkan/VulkanPipelineCache.cpp
anishmgoyal: instead of having two separate create pipeline functions that do the same thing, we just need the one. this will be used for the prewarm step as well.

filament/backend/src/vulkan/VulkanPipelineCache.h
anishmgoyal: ahh yeah - this was intended to be a part of the *next* cl. will remove.

filament/backend/src/vulkan/VulkanContext.h
anishmgoyal: supported and enabled are different - this flag specifically says if it's enabled. this being true doesn't mean it's supported.

support is determined elsewhere, by checking if vertex input dynamic state and dynamic rendering are supported.

anishmgoyal: to be more specific, what you're saying - that the driver config is a hint that doesn't determine if it's *supported* - that's exactly why this variable isn't called "supported", and instead called "enabled". i'm open to changing "enabled" to "allowed" if that's preferred

filament/backend/src/vulkan/VulkanPipelineCache.cpp
anishmgoyal: ok. i can change the driver config to say "allowed", and change this to say "enabled"

poweifeng: thinking more about it.  Do you even need a driverConfig? Don't you just want to enable it if the device matches?

filament/backend/include/backend/Platform.h
anishmgoyal: i don't think we should be so specific in the comment, personally. i can add a note saying that it will not work on all gpu's, though.

filament/backend/src/vulkan/platform/VulkanPlatform.cpp
anishmgoyal: what you're describing is how it's actually implemented - please see VulkanPipelineCache.cpp and VulkanDriver.cpp. we check for specific extensions as well as this flag.

i do not think we should be hardcoding gpu id's.

filament/backend/src/vulkan/platform/VulkanPlatform.cpp
anishmgoyal: sure

filament/backend/src/vulkan/VulkanPipelineCache.cpp
z3moon: ah true, looks like it's intended

filament/backend/src/vulkan/VulkanPipelineCache.cpp
anishmgoyal: i don't want to limit us to hardcoded lists of devices; it's an experimental feature that works on moohan, and depending on how quickly uber shader fallbacks are implemented, might have to be enabled on future androidXR devices. so, i'd prefer to have this be something we can enable and disable in the code manually.

that said - i'd be curious to know if you have a better place in mind for experimental feature flags

filament/backend/src/vulkan/VulkanPipelineCache.cpp
poweifeng: actually, that's probably a bug in the old code.  I'd prefer you'd add an assertion.  The alternative is to skip rendering drawing altogether when you don't have a pipeline, but that's too much for this PR I think.

filament/backend/src/vulkan/VulkanPipelineCache.cpp
poweifeng: I think we'd have to gate this on the particular hardware anyways.  This wouldn't work for for a pixel 8 for example. 

There is already a precedence for hardcoded driver/vendor in the GL backend.

By experimental, I think you mean that there is a baseline that you'd like to run an experiment on. AFAICT that is not the case here.  This feature is necessary to enable that baseline.

filament/backend/include/backend/Platform.h
rafadevai: You're right the comment about dependent on the vulkan driver is enough.

filament/backend/src/vulkan/VulkanContext.h
rafadevai: I gave it more though, it makes more sense to leave it as enabled like you did

filament/backend/include/backend/platforms/VulkanPlatform.h
```diff
@@ -83,6 +83,9 @@ class VulkanPlatform : public Platform, utils::PrivateImplementation<VulkanPlatf
         bool debugUtilsSupported = false;
         bool debugMarkersSupported = false;
         bool multiviewSupported = false;
+        bool dynamicRenderingSupported = false;
```
rafadevai: With the other PR you commented on, we can remove this. I definitely want to avoid starting adding more things here and better to have everything in one place

filament/backend/src/vulkan/VulkanPipelineCache.cpp
```diff
@@ -66,13 +69,26 @@ void printPipelineFeedbackInfo(VkPipelineCreationFeedbackCreateInfo const& feedb
 
 } // namespace
 
-VulkanPipelineCache::VulkanPipelineCache(VkDevice device, VulkanContext const& context)
+VulkanPipelineCache::VulkanPipelineCache(DriverBase& driver, VkDevice device, VulkanContext const& context)
```
rafadevai, line 72: Lets not pass the driver object here to avoid adding a cyclic dependency between this two classes.

So lets just pass the CallbackManager from the driver side

filament/backend/src/vulkan/VulkanPipelineCache.cpp
rafadevai: Lets just do 

if (context.asyncPipelineCachePrewarmingEnabled()) { ..... }

filament/backend/src/vulkan/VulkanPipelineCache.cpp
rafadevai: I dont see any value on having an unified function, one has a lot of hardcoded state and some things are not properly handled by the pipeline key and doing the prewarm is not even trying to create a partial pipeline.

filament/backend/src/vulkan/VulkanPipelineCache.h
```diff
@@ -46,6 +48,12 @@ class VulkanPipelineCache {
     VulkanPipelineCache(VulkanPipelineCache const&) = delete;
     VulkanPipelineCache& operator=(VulkanPipelineCache const&) = delete;
 
+    static inline bool isAsyncPrewarmingSupported(VulkanContext const& context) {
```
rafadevai: This should be done in the platform like we do for other features.
if the prewarm is enabled, then all the conditions are met

filament/backend/src/vulkan/VulkanPipelineCache.h
```diff
@@ -211,6 +226,14 @@ class VulkanPipelineCache {
     // Current bindings for the pipeline and descriptor sets.
     PipelineKey mBoundPipeline = {};
 
+    // Thread pool that allows us to "prewarm" the pipeline cache, reducing draw-time
+    // pipeline compilation time.
+    CompilerThreadPool mCompilerThreadPool;
```
rafadevai, line 234: Inside VulkanPipelineCache::terminate() we need to call mCompilerThreadPool.terminate too

filament/backend/src/vulkan/VulkanPipelineCache.h
```diff
@@ -211,6 +226,14 @@ class VulkanPipelineCache {
...
+
+    // Callback manager that allows us to notify the frontend when a set of pipelines have
+    // been prewarmed, signifying that it is safe to compile pipelines at draw time.
+    CallbackManager mCallbackManager;
```
rafadevai, line 238: In  VulkanPipelineCache::terminate() , we are missing calling the mCallbackManager.terminate()

filament/backend/src/vulkan/platform/VulkanPlatform.cpp
rafadevai: The reason we need to fix the gpu id is because so far that's the only one we know it works. 
I tried it on my desktop GPU that has all the required extensions and it doesnt work.

also we need to check the value of driverConfig.disableParallelShaderCompile

filament/backend/src/vulkan/VulkanContext.h
anishmgoyal: i'm going to switch it to 'allowed' actually, and the combination of supported and allowed will equal enabled.

filament/backend/src/vulkan/VulkanPipelineCache.cpp
anishmgoyal: i disagree - please see the follow-up cl. there is a LOT of duplicated code, and much less varies between the two than would be obvious.

having a full duplicate would be fairly unmaintainable.

anishmgoyal: additionally - i think adding support for dynamic vertex input + dynamic rendering WITHIN this function might be useful for future builds. nothing about this function will be tied specifically to the prewarming, and so none of it will have to be removed when we remove prewarming.

that said, we have to accept the possibility that prewarming won't go away, if we cannot achieve parity with this approach using other methods.

filament/backend/src/vulkan/platform/VulkanPlatform.cpp
anishmgoyal: we will have this as a function that platform must implement. xrplatform will hardcode the gpu id.
we now have 5 conditions for enabling:
- device id supported
- cache prewarming allowed by application
- parallel shader compilation must not be disabled
- dynamic rendering supported
- vertex input dynamic state supported

filament/backend/include/backend/platforms/VulkanPlatform.h
anishmgoyal: this is how extensions are marked enabled / disabled on shared context, so i disagree. this has to be here.

this is especially true because this is what moohan uses. without this, we will not know if dynamic rendering is supported on moohan.

if you'd like to refactor the shared context stuff, that's fine, but that's out of scope for this change.

filament/backend/src/vulkan/VulkanPipelineCache.cpp
anishmgoyal: i'm fairly confident callbackmanager isn't meant to be copy constructed, but will double check

anishmgoyal: callbackmanager cannot be copy constructed. also - driverbase is not a cyclic dependency. vulkandriver would be a cyclic dependency.

filament/backend/src/vulkan/VulkanPipelineCache.cpp
anishmgoyal: moved all of this logic to VulkanPlatform, which i think sums up the different cases nicely. the hardcoded device id will live in XrVulkanPlatform, and that's just a virtual function here.

filament/backend/src/vulkan/VulkanPipelineCache.h
anishmgoyal: moved to platform.

filament/backend/include/backend/platforms/VulkanPlatform.h
rafadevai: Like I said before, I already refectored some things and this is not needed anymore. 

So please remove it

filament/backend/src/vulkan/VulkanPipelineCache.h
```diff
@@ -86,14 +88,32 @@ class VulkanPipelineCache {
...
+     *
+     * @param driver The driver this is being instantiated for. This is used only for construction of
+     *               the callback manager, which references the driver for scheduling callbacks.
+     * @param device The device that the pipelines will be created and run on.
+     * @param context Information about the current instance of Vulkan, such as supported extensions,
+     *                and enabled features.
+     * @param isAsyncPrewarmingEnabled true if async cache prewarming is enabled (in which case a
+     *                                 threadpool for such jobs will be spawned), false if not.
+     */
+    VulkanPipelineCache(DriverBase& driver, VkDevice device, VulkanContext const& context, bool isAsyncPrewarmingEnabled = false);
```
rafadevai: Lets not pass around the driver object, this just created a circular dependency between both objects. Its better to pass the callbackmanager object.

filament/backend/src/vulkan/platform/VulkanPlatform.cpp
```diff
@@ -227,8 +227,9 @@ ExtensionSet getDeviceExtensions(VkPhysicalDevice device) {
 #if defined(__APPLE__)
         VK_KHR_PORTABILITY_SUBSET_EXTENSION_NAME,
 #endif
+        VK_EXT_VERTEX_INPUT_DYNAMIC_STATE_EXTENSION_NAME,
         VK_KHR_MULTIVIEW_EXTENSION_NAME,
-
+        VK_KHR_DYNAMIC_RENDERING_EXTENSION_NAME,
```
rafadevai, line 232: We also need VK_KHR_depth_stencil_resolve extension its a dependency for dynamic rendering

filament/backend/src/vulkan/platform/VulkanPlatform.cpp
```diff
@@ -916,18 +930,28 @@ void VulkanPlatform::queryAndSetDeviceFeatures(Platform::DriverConfig const& dri
 
     // Store the extension support in the context
     if (!mImpl->mSharedContext) {
-        context.mDebugUtilsSupported = setContains(instExts, VK_EXT_DEBUG_UTILS_EXTENSION_NAME);
+        context.mDebugUtilsSupported =
+                setContains(instExts, VK_EXT_DEBUG_UTILS_EXTENSION_NAME);
         context.mDebugMarkersSupported =
                 setContains(deviceExts, VK_EXT_DEBUG_MARKER_EXTENSION_NAME);
+        context.mDynamicRenderingSupported =
```
rafadevai: We need to check if the feature is actually supported and enable it, having the extension is not enough

https://docs.vulkan.org/refpages/latest/refpages/source/VkPhysicalDeviceDynamicRenderingFeatures.html

filament/backend/src/vulkan/VulkanContext.h
```diff
@@ -179,13 +195,21 @@ struct VulkanContext {
...
+    bool mIsUnifiedMemoryArchitecture = false;
     bool mLazilyAllocatedMemorySupported = false;
+    bool mPipelineCreationFeedbackSupported = false;
     bool mProtectedMemorySupported = false;
-    bool mIsUnifiedMemoryArchitecture = false;
+    bool mVertexInputDynamicStateSupported = false;
+
+    // These are options that can be enabled or disabled at an application level.
+    bool mAsyncPipelineCachePrewarmingAllowed = false;
+    bool mParallelShaderCompileDisabled = false;
```
rafadevai, line 220: mParallelShaderCompileDisabled is not needed, we only need the prewarm one

filament/backend/include/backend/Platform.h
rafadevai, line 1: Update the comment for `disableParallelShaderCompile` and mention Vulkan too, on line 291

filament/backend/src/vulkan/VulkanPipelineCache.cpp
```diff
@@ -66,13 +69,27 @@ void printPipelineFeedbackInfo(VkPipelineCreationFeedbackCreateInfo const& feedb
...
-          mContext(context) {
+          mCallbackManager(driver),
+          mContext(context),
+          mIsAsyncPrewarmingEnabled(isAsyncPrewarmingEnabled) {
     VkPipelineCacheCreateInfo createInfo = {
         .sType = VK_STRUCTURE_TYPE_PIPELINE_CACHE_CREATE_INFO,
     };
     bluevk::vkCreatePipelineCache(mDevice, &createInfo, VKALLOC, &mPipelineCache);
+
+    if (mIsAsyncPrewarmingEnabled) {
```
rafadevai: why just not do mContext.asyncPipelineCachePrewarmingAllowed() ?

filament/src/details/Engine.h
```diff
@@ -783,6 +783,10 @@ class FEngine : public Engine {
                 bool assert_native_window_is_valid = false;
             } opengl;
             struct {
+                // In certain GPU drivers, graphics pipelines are cached based on a subset of their
+                // parameters. In those cases, we can create fake pipelines ahead of time to ensure
+                // a cache hit when creating graphics pipelines at draw time, eliminating hitching.
+                bool allow_pipeline_cache_prewarming = false;
```
pixelflinger: Rename this to `enable_` since it should be seen as a feature flag enabling a feature. it's intended to be removed once the feature is functional and safe.
This can't be used as a "setting".

filament/backend/include/backend/Platform.h
anishmgoyal: maybe we should defer until we implement gpl?

filament/backend/src/vulkan/VulkanContext.h
anishmgoyal: we should have both. eventually, we will remove the prewarming one, when gpl is supported.

filament/backend/src/vulkan/VulkanPipelineCache.h
anishmgoyal: as mentioned before, this is necessary. the callback manager is not copy constructible, and this is not a circular dependency in the strictest sense since we are using DriverBase, and not VulkanDriver. the callback manager will hold a driver reference no matter how you implement this.

this pattern is followed by the gl backend.

filament/backend/include/backend/platforms/VulkanPlatform.h
anishmgoyal: will review the other pr again and adjust

filament/backend/src/vulkan/platform/VulkanPlatform.cpp
anishmgoyal: adding it to target_exts, but not vulkancontext

filament/backend/src/vulkan/platform/VulkanPlatform.cpp
anishmgoyal: updated!

filament/backend/include/backend/Platform.h
anishmgoyal: updated for now

anishmgoyal:
@pixelflinger @poweifeng @rafadevai lmk if there are any other concerns here

pixelflinger (approved):

PR #9516, opened 2025-12-12
webgpu: refactor WebGPUPlatform

yrcloud:
Refactor WebGPUPlatform so each OS environment has its own WebGPU Platform class derived from the same base, and use virtual functions for their own implementations.


pixelflinger (approved):

pixelflinger:
please wait for powei's review as well.

PR #9517, opened 2025-12-12
Fix Build.builder.feature() for Java bindings

z3moon:
The "const char*" used for the string literal parameter of the `feature` method becomes invalid once it passes outside the Java binding scope, leading to invalid data access. This fix ensures the string is safely stored.


z3moon:
E.g., this didn't work

```
engine = Engine.Builder()
        .config(config)
        .feature("backend.enable_asynchronous_operation", true)
        .build()
```

filament/src/details/Engine.cpp
```diff
@@ -132,7 +132,7 @@ struct Engine::BuilderDetails {
     FeatureLevel mFeatureLevel = FeatureLevel::FEATURE_LEVEL_1;
     void* mSharedContext = nullptr;
     bool mPaused = false;
-    std::unordered_map<std::string_view, bool> mFeatureFlags;
+    std::unordered_map<std::string, bool> mFeatureFlags;
```
pixelflinger: use `utils::CString`, std::string is banned.

z3moon: done

PR #9519, opened 2025-12-12
Workaround for harmless warning about descriptor sets

pixelflinger:
The proper fix is difficult and will be handled later. Here we simply suppress the warning at FL0.

BUGS=[468072646]

### DIFF
--- a/filament/src/ds/DescriptorSet.cpp
+++ b/filament/src/ds/DescriptorSet.cpp
@@ -113,7 +113,7 @@ void DescriptorSet::commitSlow(DescriptorSetLayout const& layout,
     //  is difficult and risky, and will be done at a later time.
     //  Note: that the correct fix is actually needed to properly support FL3 once we want
     //  to take advantage of having more samplers.
-    if (UTILS_UNLIKELY(driver.getFeatureLevel() > backend::FeatureLevel::FEATURE_LEVEL_0)) {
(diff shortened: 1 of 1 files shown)


filament/src/ds/DescriptorSet.cpp
```diff
@@ -106,13 +106,22 @@ void DescriptorSet::commitSlow(DescriptorSetLayout const& layout,
...
-        });
-        mSetUndefinedParameterWarning = true;
+    // FIXME: see [b/468072646]
+    //  We only validate empty descriptors at FEATURE_LEVEL_1 and above.
+    //  This is because at FL0 it's expected that some descriptor won't be set. In theory, the
+    //  corresponding layouts should not even contain those descriptors. However, making that change
+    //  is difficult and risky, and will be done at a later time.
+    //  Note: that the correct fix is actually needed to properly support FL3 once we want
+    //  to take advantage of having more samplers.
+    if (UTILS_UNLIKELY(driver.getFeatureLevel() > backend::FeatureLevel::FEATURE_LEVEL_0)) {
```
poweifeng: isn't this LIKELY?

pixelflinger:
```suggestion
    if (UTILS_LIKELY(driver.getFeatureLevel() > backend::FeatureLevel::FEATURE_LEVEL_0)) {
```

PR #9521, opened 2025-12-13
docs: fix freezing when generating docs

poweifeng:
For reasons unknown, selenium will freeze when trying to make
request to the local server serving the markdown files. We add
a delay between the server and the selenium driver to hopefully
address this.

DOCS_FORCE


z3moon:
did you find this working? let me blindly LGTM anyway to unblock your progress.

poweifeng:
> did you find this working? let me blindly LGTM anyway to unblock your progress.

it does according to my test.

PR #9522, opened 2025-12-15
Add descriptor set layouts to Program

anishmgoyal:
This will help later for parallel compilation

@pixelflinger @poweifeng @phoenixxxx @rafadevai

### DIFF
--- a/filament/backend/src/Platform.cpp
+++ b/filament/backend/src/Platform.cpp
@@ -194,7 +194,7 @@ void Platform::debugUpdateStat(const char* key, uint64_t intValue) {
         callback = mDebugUpdateStat;
     }
     if (callback) {
-        (*callback)(key, intValue, "");
+        (*callback)(key, intValue, utils::CString(""));
     }
 }
 
--- a/filament/backend/src/opengl/OpenGLDriver.cpp
+++ b/filament/backend/src/opengl/OpenGLDriver.cpp
@@ -2182,30 +2182,53 @@ void OpenGLDriver::destroyVertexBuffer(Handle<HwVertexBuffer> vbh) {
     }
 }
 
+void OpenGLDriver::destroyIndexBufferCommon(Handle<HwIndexBuffer> ibh) {
+    GLIndexBuffer const* ib = handle_cast<const GLIndexBuffer*>(ibh);
+    auto& gl = mContext;
+    gl.deleteBuffer(ib->gl.buffer, GL_ELEMENT_ARRAY_BUFFER);
+    destruct(ibh, ib);
+}
+
 void OpenGLDriver::destroyIndexBuffer(Handle<HwIndexBuffer> ibh) {
     DEBUG_MARKER()
 
     if (ibh) {
-        auto& gl = mContext;
         GLIndexBuffer const* ib = handle_cast<const GLIndexBuffer*>(ibh);
-        gl.deleteBuffer(ib->gl.buffer, GL_ELEMENT_ARRAY_BUFFER);
-        destruct(ibh, ib);
+        if (ib->asynchronous) {
+            getJobQueue()->push([this, ibh]() {
+                destroyIndexBufferCommon(ibh);
+            });
+        } else {
+            destroyIndexBufferCommon(ibh);
+        }
     }
+}
+
+void OpenGLDriver::destroyBufferObjectCommon(Handle<HwBufferObject> boh) {
+    GLBufferObject const* bo = handle_cast<const GLBufferObject*>(boh);
+    auto& gl = mContext;
+    // check we're not destroying a buffer that has active mappings
+    assert_invariant(bo->mappingCount == 0);
+    if (UTILS_UNLIKELY(bo->bindingType == BufferObjectBinding::UNIFORM && gl.isES2())) {
+        free(bo->gl.buffer);
+    } else {
+        gl.deleteBuffer(bo->gl.id, bo->gl.binding);
+    }
+    destruct(boh, bo);
 }
 
 void OpenGLDriver::destroyBufferObject(Handle<HwBufferObject> boh) {
     DEBUG_MARKER()
+
     if (boh) {
-        auto& gl = mContext;
         GLBufferObject const* bo = handle_cast<const GLBufferObject*>(boh);
-        // check we're not destroying a buffer that has active mappings
-        assert_invariant(bo->mappingCount == 0);
-        if (UTILS_UNLIKELY(bo->bindingType == BufferObjectBinding::UNIFORM && gl.isES2())) {
-            free(bo->gl.buffer);
+        if (bo->asynchronous) {
+            getJobQueue()->push([this, boh]() {
+                destroyBufferObjectCommon(boh);
+            });
         } else {
-            gl.deleteBuffer(bo->gl.id, bo->gl.binding);
+            destroyBufferObjectCommon(boh);
         }
-        destruct(boh, bo);
     }
 }
 
--- a/filament/backend/src/opengl/OpenGLDriver.h
+++ b/filament/backend/src/opengl/OpenGLDriver.h
@@ -387,6 +387,8 @@ private:
     void updateIndexBufferCommon(Handle<HwIndexBuffer> ibh, BufferDescriptor&& p,
             uint32_t const byteOffset);
     void destroyTextureCommon(Handle<HwTexture> th);
+    void destroyBufferObjectCommon(Handle<HwBufferObject> boh);
+    void destroyIndexBufferCommon(Handle<HwIndexBuffer> ibh);
 
     // state required to represent the current render pass
     Handle<HwRenderTarget> mRenderPassTarget;
--- a/filament/src/PostProcessManager.cpp
+++ b/filament/src/PostProcessManager.cpp
@@ -2533,9 +2533,10 @@ static float4 getVignetteParameters(VignetteOptions const& options,
 }
 
 void PostProcessManager::colorGradingPrepareSubpass(DriverApi& driver,
-        const FColorGrading* colorGrading, ColorGradingConfig const& colorGradingConfig,
-        VignetteOptions const& vignetteOptions, uint32_t const width, uint32_t const height) noexcept {
-
+        FColorGrading const* colorGrading,
+        ColorGradingConfig const& colorGradingConfig,
+        VignetteOptions const& vignetteOptions,
+        uint32_t const width, uint32_t const height) noexcept {
     auto& material = getPostProcessMaterial("colorGradingAsSubpass");
     FMaterialInstance const* const mi =
             configureColorGradingMaterial(material, colorGrading, colorGradingConfig,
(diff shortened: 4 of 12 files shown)


filament/src/MaterialDefinition.h
```diff
@@ -56,10 +57,16 @@ struct MaterialDefinition {
...
+    // may be used by some backends in parallel compilation of programs.
+    backend::DescriptorSetLayout perViewDescriptorSetLayoutDescription;
+    backend::DescriptorSetLayout perViewDescriptorSetLayoutVsmDescription;
+    backend::DescriptorSetLayout descriptorSetLayoutDescription;
+
     // try to order by frequency of use
-    DescriptorSetLayout perViewDescriptorSetLayout;
-    DescriptorSetLayout perViewDescriptorSetLayoutVsm;
-    DescriptorSetLayout descriptorSetLayout;
+    filament::DescriptorSetLayout perViewDescriptorSetLayout;
```
anishmgoyal, line 67: note - this is to avoid ambiguity in the namespacing for `DescriptorSetLayout`

filament/src/MaterialDefinition.cpp
```diff
@@ -477,42 +476,42 @@ void MaterialDefinition::processDescriptorSets(FEngine& engine) {
...
-    descriptorSetLayout.label = CString{ name }.append("_perMat");
-    perViewDescriptorSetLayout.label = CString{ name }.append("_perView");
-    perViewDescriptorSetLayoutVsm.label = CString{ name }.append("_perViewVsm");
+    this->descriptorSetLayoutDescription.label = CString{ name }.append("_perMat");
+    this->perViewDescriptorSetLayoutDescription.label = CString{ name }.append("_perView");
+    this->perViewDescriptorSetLayoutVsmDescription.label = CString{ name }.append("_perViewVsm");
 
     // get the PER_RENDERABLE and PER_VIEW descriptor binding info
-    for (auto&& [bindingPoint, descriptorSetLayout] : {
+    for (auto&& [bindingPoint, dsl] : {
```
anishmgoyal, line 491: note - this is to avoid ambiguity vs the field, `descriptorSetLayout`

filament/backend/include/backend/Program.h
```diff
@@ -147,6 +147,16 @@ class Program {
         return mDescriptorBindings;
     }
 
+    inline Program& descriptorLayout(backend::descriptor_set_t set,
+            DescriptorSetLayout descriptorLayout) noexcept {
+        mDescriptorLayouts[set] = descriptorLayout;
```
pixelflinger:
```suggestion
        mDescriptorLayouts[set] = std::move(descriptorLayout);
```

filament/src/details/Material.cpp
```diff
@@ -499,6 +517,16 @@ void FMaterial::createAndCacheProgram(Program&& p, Variant const variant) const
         }
     }
 
+    // Set descriptor sets for the program.
+    // Note: right now, we're going to assume VSM is disabled. In the future, we may
+    // want to provide both to the backend, so that both can be built.
+    p.descriptorLayout(+DescriptorSetBindingPoints::PER_VIEW,
+        getPerViewDescriptorSetLayoutDescription(variant, /*useVsm=*/false));
```
pixelflinger: I think you must set `useVsm` to `Variant::isVSMVariant(variant)` here

filament/src/details/Material.h
```diff
@@ -81,18 +81,39 @@ class FMaterial : public Material {
         return mDefinition.uniformInterfaceBlock;
     }
 
-    DescriptorSetLayout const& getPerViewDescriptorSetLayout() const noexcept {
+    backend::DescriptorSetLayout const& getPerViewDescriptorSetLayoutDescription() const noexcept {
+        return mDefinition.perViewDescriptorSetLayoutDescription;
+    }
+
+    backend::DescriptorSetLayout const& getPerViewDescriptorSetLayoutDescription(
+            Variant const variant, bool const useVsmDescriptorSetLayout) const noexcept;
```
pixelflinger: we need comments here, why do we have two function with the same name that seem to be doing vastly different things?

filament/src/details/Material.h
```diff
@@ -81,18 +81,39 @@ class FMaterial : public Material {
...
+        if (!isSharedVariant(variant)) {
+            return mDefinition.descriptorSetLayoutDescription;
+        }
+        FMaterial const* const pDefaultMaterial = mEngine.getDefaultMaterial();
+        if (UTILS_UNLIKELY(!pDefaultMaterial)) {
+            return mDefinition.descriptorSetLayoutDescription;
+        }
+        return pDefaultMaterial->getDescriptorSetLayoutDescription();
+    }
+
```
pixelflinger: don't put the implementation in the header.

filament/src/MaterialDefinition.cpp
```diff
@@ -477,42 +476,42 @@ void MaterialDefinition::processDescriptorSets(FEngine& engine) {
             refractionMode == RefractionMode::SCREEN_SPACE;
     bool const hasFog = !(variantFilterMask & UserVariantFilterMask(UserVariantFilterBit::FOG));
 
-    auto perViewDescriptorSetLayout = descriptor_sets::getPerViewDescriptorSetLayout(
+    this->perViewDescriptorSetLayoutDescription = descriptor_sets::getPerViewDescriptorSetLayout(
```
pixelflinger, line 479: it looks like you don't actually need the `this->` in most cases here.

pixelflinger (changes requested):

filament/backend/include/backend/Program.h
anishmgoyal: we might copy this layout multiple times, for multiple programs, which is why it wasn't moved

filament/src/details/Material.cpp
anishmgoyal: i thought that at first, but upon checking with powei, it seems like whether or not vsm is enabled actually depends on the render pass, and that even if a shader has vsm baked in, it might not be used at runtime? and the shadow type is set by the view?

at least on moohan, we know vsm is never enabled. lmk your thoughts on this!

filament/src/details/Material.h
anishmgoyal: the problem is, i'm following the same pattern as the `getPerViewDescriptorSetLayout` method, but i'm not sure why that was written how it was either 🤔 so i'm not sure how to describe this

filament/src/details/Material.h
anishmgoyal: should i move the implementation for `getDescriptorSetLayout` to the cpp file as well? sorry - was trying to follow the same approach as existed in the file already

filament/src/MaterialDefinition.cpp
anishmgoyal: yeah - i mostly added it because this is a struct, and doesn't use the `mVarName` syntax. i was worried it wouldn't be clear that these were fields, not locals. lmk though - happy to remove `this->`, or even convert this to a class and change the fields to `mVarName`

filament/backend/include/backend/Program.h
anishmgoyal: oh wait - nvm, i see what you mean

filament/backend/include/backend/Program.h
```diff
@@ -147,6 +147,16 @@ class Program {
         return mDescriptorBindings;
     }
 
+    inline Program& descriptorLayout(backend::descriptor_set_t set,
+            DescriptorSetLayout descriptorLayout) noexcept {
+        mDescriptorLayouts[set] = std::move(descriptorLayout);
+        return *this;
+    }
+
+    const std::array<DescriptorSetLayout, MAX_DESCRIPTOR_SET_COUNT>& getDescriptorSetLayouts() const noexcept {
```
poweifeng: nit: prefer a `using DescriptorSetLayoutArray = std::array<DescriptorSetLayout, MAX_DESCRIPTOR_SET_COUNT>`

filament/src/details/Material.h
```diff
@@ -81,18 +81,30 @@ class FMaterial : public Material {
         return mDefinition.uniformInterfaceBlock;
     }
 
-    DescriptorSetLayout const& getPerViewDescriptorSetLayout() const noexcept {
+    inline backend::DescriptorSetLayout const& getPerViewDescriptorSetLayoutDescription() const noexcept {
```
poweifeng: is this used anywhere?

filament/src/details/Material.h
```diff
@@ -81,18 +81,30 @@ class FMaterial : public Material {
...
+        return mDefinition.perViewDescriptorSetLayoutDescription;
+    }
+
+    backend::DescriptorSetLayout const& getPerViewDescriptorSetLayoutDescription(
```
poweifeng: can these description methods be private?

filament/src/details/Material.cpp
pixelflinger: yeah, but this part of the code can't be moohan specific.  It seems to me that you could end-up with a Program (on the backend side) that thinks it doesn't need the VSM descriptor set layout (which is different), when in fact the actual shader does.

I think you don't have a bug now because you're only using this info for the cache, but in reality, if you were validating that the layout in the pipeline matches the layout given here... they wouldn't.

filament/src/details/Material.h
pixelflinger: only trivial getters can be in headers.

pixelflinger (changes requested):

filament/src/details/Material.cpp
anishmgoyal: looking into this a bit more. i don't have an answer just yet, but i *do* know that simply using Variant::isVSMVariant(variant) causes the prewarm step to crash. there seem to be cases where a variant is a vsm variant, but the descriptor set layout doesn't match the shader. not sure why this is, because descriptor_sets::getPerViewDescriptorSetLayoutWithVariant seems to believe that that flag is sufficient to find a matching descriptor set layout for the given material.

anishmgoyal: huh.. after testing again, it *isn't* crashing now. possible that the crash from earlier was a red herring / unrelated. going to take some more traces.

filament/src/details/Material.h
anishmgoyal: it's used in the following cl. it makes the most sense to be introduced in this cl, as it is contextually related to the api being introduced for the backend.

filament/src/details/Material.h
anishmgoyal: moved these to private.

filament/src/details/Material.h
anishmgoyal: removing the no-arg version of this, instead of maintaining symmetry with the other api

filament/src/MaterialDefinition.cpp
anishmgoyal: i'm gonna leave this as is, because it feels out of scope to refactor this entire struct for this cl. i strongly feel that either this should become a class, OR `this->` should be retained for readability. otherwise, it is not clear these are fields.

pixelflinger (approved):

PR #9523, opened 2025-12-15
VK: Refactor platform Vulkan initialization

rafadevai:
Refactor the platform code to allow overriding how the vkInstance, vkPhysicalDevice and vkDevice are created.

This refactor will also consolidate where the extensions and features are enabled/disabled into VulkanPlatform, instead of each platform doing its own individual setup if needed. It also address the issue of knowing if a feature or extension has been enabled or not, unless we keep adding entries to the SharedContext which is not scalable.

The createDriver function now have a more defined structure, mostly how and where features are queried and enabled.

- Get shared context data.
- Create instance if missing.
- Select physical device if missing.
- Query physical device support and set supported engine features into the VulkanContext.
- Select the queues
- Create the device and queues

### DIFF
--- a/filament/backend/src/vulkan/platform/VulkanPlatform.cpp
+++ b/filament/backend/src/vulkan/platform/VulkanPlatform.cpp
@@ -562,7 +562,7 @@ Driver* VulkanPlatform::createDriver(void* sharedContext,
         instExts.merge(getRequiredInstanceExtensions());
     }
     if (mImpl->mInstance == VK_NULL_HANDLE) {
-        mImpl->mInstance = createInstance(instExts);
+        createInstance(instExts);
     }
     assert_invariant(mImpl->mInstance != VK_NULL_HANDLE);
 
@@ -581,7 +581,8 @@ Driver* VulkanPlatform::createDriver(void* sharedContext,
     printDeviceInfo(mImpl->mInstance, mImpl->mPhysicalDevice);
 
     ExtensionSet deviceExts;
-    // If using a shared context, we do not assume any extensions.
+    // If a shared context is not used, we will use our own provided list; otherwise, we do not
+    // assume any extensions.
     if (!mImpl->mSharedContext) {
         deviceExts = getDeviceExtensions(mImpl->mPhysicalDevice);
         auto [prunedInstExts, prunedDeviceExts] =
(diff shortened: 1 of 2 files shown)


filament/backend/src/vulkan/platform/VulkanPlatform.cpp
```diff
@@ -761,27 +574,98 @@ Driver* VulkanPlatform::createDriver(void* sharedContext,
...
-    // physicalDeviceFeatures, graphicsQueueFamilyIndex.
-    vkGetPhysicalDeviceProperties2(mImpl->mPhysicalDevice, &context.mPhysicalDeviceProperties);
-    vkGetPhysicalDeviceFeatures2(mImpl->mPhysicalDevice, &context.mPhysicalDeviceFeatures);
-    vkGetPhysicalDeviceMemoryProperties(mImpl->mPhysicalDevice, &context.mMemoryProperties);
+    ExtensionSet deviceExts;
+    // If using a shared context, we do not assume any extensions.
+    if (!mImpl->mSharedContext) {
+        deviceExts = getDeviceExtensions(mImpl->mPhysicalDevice);
+        auto [prunedInstExts, prunedDeviceExts] =
+                pruneExtensions(mImpl->mPhysicalDevice, driverConfig, instExts, deviceExts);
```
poweifeng, line 589: I think you'd also need to be able to override this.  For example, if I overriden getInstance, but did not enable `VK_EXT_DEBUG_UTILS` in a debug build, then it's already deviated from the default.

poweifeng: And many if not all of the context.xSupported should probably be renamed to context.xEnabled IMO

filament/backend/src/vulkan/platform/VulkanPlatform.cpp
```diff
@@ -1035,4 +857,208 @@ bool VulkanPlatform::isTransientAttachmentSupported() const noexcept {
...
+        .pEnabledValidationFeatures = enables,
+    };
+    if (validationFeaturesSupported) {
+        chainStruct(&instanceCreateInfo, &features);
+    }
+
+    return createVkInstance(instanceCreateInfo);
+}
+
+VkDevice VulkanPlatform::createLogicalDeviceAndQueues(const ExtensionSet& deviceExtensions,
```
poweifeng: nit: it's weird that this function return a device, but writes to mImpl->mGraphicsQueueIndex. I'd prefer it be purely functional (or otherwise not at all).

filament/backend/src/vulkan/platform/VulkanPlatform.cpp
```diff
@@ -761,27 +574,98 @@ Driver* VulkanPlatform::createDriver(void* sharedContext,
...
-    chainStruct(&context.mPhysicalDeviceFeatures, &context.mPhysicalDeviceVk11Features);
-
-    // Initialize the following fields: physicalDeviceProperties, memoryProperties,
-    // physicalDeviceFeatures, graphicsQueueFamilyIndex.
-    vkGetPhysicalDeviceProperties2(mImpl->mPhysicalDevice, &context.mPhysicalDeviceProperties);
-    vkGetPhysicalDeviceFeatures2(mImpl->mPhysicalDevice, &context.mPhysicalDeviceFeatures);
-    vkGetPhysicalDeviceMemoryProperties(mImpl->mPhysicalDevice, &context.mMemoryProperties);
+    ExtensionSet deviceExts;
+    // If using a shared context, we do not assume any extensions.
+    if (!mImpl->mSharedContext) {
```
anishmgoyal, line 586: nit: the comment and condition seem to mismatch? ("if *using* a shared context" -> `if !sharedContext`)

filament/backend/src/vulkan/platform/VulkanPlatform.cpp
```diff
@@ -761,27 +574,98 @@ Driver* VulkanPlatform::createDriver(void* sharedContext,
...
+        deviceExts = getDeviceExtensions(mImpl->mPhysicalDevice);
+        auto [prunedInstExts, prunedDeviceExts] =
+                pruneExtensions(mImpl->mPhysicalDevice, driverConfig, instExts, deviceExts);
+        instExts = prunedInstExts;
+        deviceExts = prunedDeviceExts;
+    }
 
+    // Query all the supported physical device features and Enable/Disable any feature as needed
```
anishmgoyal: nit: this feels like it could probably be a separate function, this one has gotten fairly long

filament/backend/src/vulkan/platform/VulkanPlatform.cpp
rafadevai: so maybe I should change what we override, I only want them to provide the function to create the objects and not let them change the the creation parameters because then we end up in those weird situations

rafadevai: Okay, so just added a comment about the expectation when using these functions. 

So in the case you mentioned they should respect the extensions used by filament and add their own on top, otherwise we cannot guarantee things work as expected.

filament/backend/src/vulkan/platform/VulkanPlatform.cpp
rafadevai: Changed it

filament/backend/src/vulkan/platform/VulkanPlatform.cpp
rafadevai: Its an old comment but I updated it, let me know if its better or if you have any suggestion.

filament/backend/src/vulkan/platform/VulkanPlatform.cpp
rafadevai: Done

Issue #9526, opened 2025-12-16
Crash compiling mipmapDepth program on Metal


kpeeters:
**Describe the bug**
Running on iPhone or iPad, my program crashes with an error compiling filament's `mipmapDepth` program. See log below. Works fine on Mac Catalyst (with no warning or error in the log). Also works fine on Android/OpenGL.

**To Reproduce**
Consistent crash on shader compilation. This is in a large project, will attempt to find minimal example next, just thought to flag it in case it rings any bells.

**Expected behavior**
No crash.

**Logs**
```
program_source:8:1: error: type 'const constant MaterialParams *' is not valid for attribute 'id'
constant MaterialParams* materialParams [[id(0)]];
^                                         ~~~~~
program_source:18:26: error: type 'const constant spvDescriptorSetBuffer2 &' is not valid for attribute 'buffer'
fragment main0_out main0(constant spvDescriptorSetBuffer2& spvDescriptorSet2 [[buffer(23)]], constant uint* spvDynamicOffsets [[buffer(25)]], float4 gl_FragCoord [[position]])
                         ^                                                     ~~~~~~~~~~
program_source:18:35: note: type 'spvDescriptorSetBuffer2' cannot be used in buffer pointee type
fragment main0_out main0(constant spvDescriptorSetBuffer2& spvDescriptorSet2 [[buffer(23)]], constant uint* spvDynamicOffsets [[buffer(25)]], float4 gl_FragCoord [[position]])
                                  ^

PanicLog
in compileProgram:168
reason: Failed to compile Metal program.
#0	alq4                                   0x10658be8c _ZN5utils7details8panicLogEPKcS2_iS2_z + 152
#1	alq4                                   0x10657d8c8 _ZN8filament7backend19MetalShaderCompiler14compileProgramERKNS0_7ProgramEPU19objcproto9MTLDevice11objc_object + 1144
#2	alq4                                   0x10657e488 _ZZN5utils9InvocableIFvvEEC1IZN8filament7backend19MetalShaderCompiler13createProgramERKNS_7CStringEONS5_7ProgramEE3$_0Li0EEEOT_ENUlPvE0_8__invokeESF_ + 56
#3	alq4                                   0x10657de00 _ZN8filament7backend19MetalShaderCompiler10getProgramERNSt3__110shared_ptrINS1_17MetalProgramTokenEEE + 240
#4	alq4                                   0x106579348 _ZN8filament7backend12MetalProgram12getFunctionsEv + 48
[...]

PR #9527, opened 2025-12-16
Revert "Set the minimum required Metal version (#9425)"

bejado:
This reverts commit f5071b2aae818dd28b5138df8cd80f7a7c08b3f5.

Apologies @emezeske but the "Set the minimum required Metal version" change is causing crashes on certain iOS devices, so I'm reverting it for now. See #9526.

I suspect that some of our materials require a higher version of MSL. I need to investigate further.

### DIFF
--- a/filament/backend/src/metal/MetalShaderCompiler.mm
+++ b/filament/backend/src/metal/MetalShaderCompiler.mm
@@ -111,28 +111,9 @@ bool MetalShaderCompiler::isParallelShaderCompileSupported() const noexcept {
         id<MTLLibrary> library = nil;
         switch (program.getShaderLanguage()) {
             case ShaderLanguage::MSL: {
+                // By default, Metal uses the most recent language version.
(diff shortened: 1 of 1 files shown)


poweifeng (approved):

emezeske:
Thanks for the heads up! Sorry for the breakage. I guess it's a good opportunity to be certain about what metal shader language Filament requires. Let me know if I can help with testing or anything.

emezeske:
So how do we follow up here? It would appear that the minimum metal version for iOS is not well-understood, and needs to be sorted out, is that correct?

The rollback here causes crashes for my use of Filament on desktop macOS with Intel macs in certain configurations, so I am keen to sort out the issue, let me know how I can help.

bejado:
Internally we need to audit usages of Filament and understand which Metal Shading Language versions we need to support. If we do need to support MSL 2.0, then we also need to fix the codegen issue that caused shader compilation to fail. Unfortunately I don't have a timeline for this at the moment.

PR #9528, opened 2025-12-17
android: fix texture rendertarget sample for vk

poweifeng:
- add a flag to run the sample using regular, non-external texture as render target.
 - Add a non-external version of the material
 - vk: fix y-flipping issue with uvToRenderTargetUV()


android/samples/sample-texture-target/src/main/materials/textured.mat
```diff
@@ -3,7 +3,7 @@ material {
     shadingModel : unlit,
     parameters : [
         {
-            type : samplerExternal,
+            type : sampler2d,
```
pixelflinger, line 6: interestingly, I believe the sample should work with sampler2d -- we could possibly extend this sample to also test that.

pixelflinger (approved):

android/samples/sample-texture-target/src/main/java/com/google/android/filament/texturetarget/MainActivity.kt
```diff
@@ -385,33 +397,41 @@ class MainActivity : Activity() {
...
-                .width(width)
-                .height(height)
-                .usage(Texture.Usage.COLOR_ATTACHMENT or Texture.Usage.SAMPLEABLE)
-                .sampler(Texture.Sampler.SAMPLER_EXTERNAL)
-                .format(Texture.InternalFormat.RGBA8)
-                .external()
-                .build(engine)
-
-            texture!!.setExternalImage(engine, hardwareBuffer!!)
+            if (useExternalTexture) {
```
rafadevai, line 400: because the format is RGB8, why would this require an external sampler? or its just for testing purpose?

poweifeng: yes, this is to test the case of using external texture as rendertarget.

rafadevai: so then why the need of an external sampler? because if the AHB requires an external sampler, then it can't be used as a rendertarget.

Because to me this seem to be mixing external image with external samplers or I'm missing something.

poweifeng: fair point, afaict the following are ok cases:

 1. sampler2d and internal texture
 2. sampler2d and external RGB texture
 3. samplerExternal and external RGB texture
 4. samplerExternal and external YUV texture (externally sampled)

I think we have 1, 3 here.  But could expand to include 2.  For 4, I feel like that's covered with the camera test.

rafadevai: Gotcha

poweifeng: I'll check this in for now, but will do an update to add the additional test.

rafadevai: sounds good!

pixelflinger: @poweifeng is right about the 4 cases.  Note that using an external texture as a render target (which this sample is about in the first place) is explicitly supported by the EGL_image extension. But there are restrictions, e.g. you can't render into a YUV buffer (unless some other extension allows it). The spec is unclear about which formats are supported. But on Android, all common RGB formats are -- Vulkan/Metal/WebGPU are their own thing too.

PR #9533, opened 2025-12-18
Remove the dependency of RenderPassBuilder on DriverApi

pixelflinger:
The main goal of this CL is to break the dependency of RenderPassBuilder
on DriverApi. The dependency only existed for the auto-instancing 
feature, which needed to allocate and update a UBO on the fly.

We want to break that dependency so that RenderPassBuilder::build().
which is an expensive call, could run on a separate thread. In the
future this could allow, for example, shadow and ssao passes to 
run in parallel.

A new method RenderPass::finalize(DriverApi&) is added which *must* be 
called before calling getExecutor(), this is where the temporary
UBO is allocated and updated. We need a new method because this must be 
called on the main thread before a render pass is started.

This implementation stores the UBO data needed for instancing until
finalize() is called, if any.

We introduce a compact() utility that efficiently scans a range finding
"equivalent" consecutive objects and removes duplicates. The previous
code did that with an extra remove pass which invalidated all 
iterators.

### DIFF
--- a/filament/src/RenderPass.cpp
+++ b/filament/src/RenderPass.cpp
@@ -359,6 +359,7 @@ RenderPass::Command* RenderPass::instanceify(
 
 
 void RenderPass::finalize(FEngine const& engine, DriverApi& driver) {
+    mFinalized = true;
     if (UTILS_UNLIKELY(!mInstancingStagingBuffer.empty())) {
         auto const* p = mInstancingStagingBuffer.data();
         size_t const size = sizeof(PerRenderableData) * mInstancingStagingBuffer.size();
@@ -1133,6 +1134,7 @@ RenderPass::Executor::Executor(RenderPass const& pass, Command const* b, Command
     mHasScissorViewport = mScissor != backend::Viewport{ 0, 0, INT32_MAX, INT32_MAX };
     assert_invariant(b >= pass.begin());
     assert_invariant(e <= pass.end());
+    assert_invariant(pass.isFinalized());
 }
 
 RenderPass::Executor::Executor() noexcept
--- a/filament/src/RenderPass.h
+++ b/filament/src/RenderPass.h
@@ -345,6 +345,8 @@ public:
     using DescriptorSetSharedHandle = SharedHandle<
             backend::HwDescriptorSet, DescriptorSetHandleDeleter>;
 
+    bool isFinalized() const noexcept { return mFinalized; }
+
     /*
      * Executor holds the range of commands to execute for a given pass
      */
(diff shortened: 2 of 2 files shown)


filament/src/RenderPass.cpp
```diff
@@ -298,138 +298,111 @@ RenderPass::Command* RenderPass::instanceify(DriverApi& driver,
...
-        }, 0);
+        instancedPrimitiveOffset += instanceCount;
+    };
 
-        stagingBuffer = nullptr;
+    auto const end = compact<CONFIG_MAX_INSTANCES>(curr, last, equivalent, processor);
+    // if (!mInstancingStagingBuffer.empty()) {
+    //     slog.d << "auto-instancing, saving " << drawCallsSavedCount << " draw calls, out of "
+    //             << count << io::endl;
+    // }
```
z3moon, line 356: Maybe you forgot to remove this.

z3moon (approved):

filament/src/RenderPass.cpp
pixelflinger: oh no I keep it for when I'm debugging.

filament/src/compact.h
```diff
@@ -0,0 +1,133 @@
...
+        if (UTILS_UNLIKELY(runSize > 1)) {
+            // We have a run to process. Chunk it and process each chunk.
+            for (auto chunkBegin = src; chunkBegin != runEnd; ) {
+                auto chunkSize = std::min(size_t(std::distance(chunkBegin, runEnd)), MaxRunSize);
+                auto chunkEnd = chunkBegin + chunkSize;
+
+                if constexpr (std::is_trivially_copyable_v<value_type>) {
+                    *dst = *chunkBegin;
+                } else {
+                    if (dst != chunkBegin) {
```
show50726, line 106: nit: Can we write it as an `else if()`?

show50726 (approved):

filament/src/RenderPass.h
```diff
@@ -310,6 +311,10 @@ class RenderPass {
     // allocated commands ARE NOT freed, they're owned by the Arena
     ~RenderPass() noexcept;
 
+    // this must be called before calling getExecutor(), but can't be called from within
```
poweifeng, line 314: nit: maybe this can be assert_invariant'd if you store a boolean? would help to avoid mistakes.

filament/test/test_compact.cpp
```diff
@@ -0,0 +1,261 @@
...
+        FAIL() << "Processor should not be called on an empty range.";
+    };
+    auto const end = compact<4>(data.begin(), data.end(), equivalent, processor);
+    EXPECT_EQ(end, data.end());
+    EXPECT_TRUE(data.empty());
+}
+
+TEST(CompactTest, NoRuns) {
+    std::vector data = { 1, 2, 3, 4, 5 };
+    std::vector<int> const original_data = data;
```
poweifeng, line 74: nit: many of the var names use underscore instead of camel case.  But optional.

pixelflinger: that's how you can tell AI wrote the code :-)

filament/src/compact.h
pixelflinger: because it's a `if constexpr` I prefer to keep it this way.

filament/src/RenderPass.h
pixelflinger: done

PR #9534, opened 2025-12-18
Fix missing dep in CompressedStringChunk

trie94:
adding `<cstdint>` explicitly as we are seeing build failure in moohan:

```
external/third_party+/filament/libs/filamat/src/eiff/CompressedStringChunk.cpp:38:22: error: no member named 'uint8_t' in namespace 'std'
   38 |     std::vector<std::uint8_t> compressed(bufferBound);
      |                 ~~~~~^
1 error generated.
```

### DIFF
--- a/libs/filamat/src/eiff/CompressedStringChunk.cpp
+++ b/libs/filamat/src/eiff/CompressedStringChunk.cpp
@@ -16,6 +16,7 @@
 
 #include "CompressedStringChunk.h"
 
+#include <stdint.h>
 #include <zstd.h>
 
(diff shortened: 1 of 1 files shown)


libs/filamat/src/eiff/CompressedStringChunk.cpp
poweifeng: nit: I think, informally, `uint8_t`from `#include <stdint.h>` is preferred.

trie94: done!

PR #9535, opened 2025-12-18
Add apiLevel and add parsing logic

trie94:
Add apiLevel parsing logic in MaterialParser and MaterialBuilder. This now allows the user to optionally include apiLevel, while rejecting the .mat if it requests invalid apiLevel (either below 0 or above unstable api level)

FIXES=468381963

### DIFF
--- a/libs/filamat/include/filamat/MaterialBuilder.h
+++ b/libs/filamat/include/filamat/MaterialBuilder.h
@@ -666,6 +666,9 @@ public:
      */
     MaterialBuilder& materialSource(std::string_view source) noexcept;
 
+    //! Set the (client requested) api level that the material is supposed to be compiled against.
+    MaterialBuilder& setApiLevel(uint32_t apiLevel) noexcept;
+
     /**
      * Build the material. If you are using the Filament engine with this library, you should use
      * the job system provided by Engine.
@@ -990,6 +993,9 @@ private:
     bool mNoSamplerValidation = false;
 
     bool mUseDefaultDepthVariant = false;
+
+    // Default api level is always 1.
+    uint32_t mApiLevel = 1;
 };
 
 } // namespace filamat
--- a/libs/filamat/src/MaterialBuilder.cpp
+++ b/libs/filamat/src/MaterialBuilder.cpp
@@ -1257,6 +1257,11 @@ MaterialBuilder& MaterialBuilder::materialSource(std::string_view source) noexce
     return *this;
 }
 
+MaterialBuilder& MaterialBuilder::setApiLevel(uint32_t apiLevel) noexcept {
+    mApiLevel = apiLevel;
+    return *this;
+}
+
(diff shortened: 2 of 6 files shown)


libs/filamat/include/filamat/MaterialBuilder.h
```diff
@@ -122,6 +122,15 @@ class UTILS_PUBLIC MaterialBuilderBase {
         ALL = 0xFFFFFFFFFFFFFFFF
     };
 
+    // Those are the api levels that are used in the source material file (.mat)
+    //
+    // The level used for the apis that are already public and stable. Any released api supports
+    // backward compatibility (i.e. No breaking changes will be introduced for those apis.)
+    static constexpr uint32_t RELEASED_MATERIAL_API_LEVEL = 1;
```
trie94: open for discussion: i added those in MaterialBuilder, but i could add it in MaterialEnums.h, next to cmat version (https://github.com/google/filament/blob/main/libs/filabridge/include/filament/MaterialEnums.h#L31)

any preference?

pixelflinger: I feel like MaterialEnums.h would be better. You could have it in both, like tehe source of truth in MaterialEnums and have a `MaterialBuilder::RELEASED_MATERIAL_API_LEVEL` that equals that.

libs/filamat/src/MaterialBuilder.cpp
```diff
@@ -1293,6 +1298,12 @@ Package MaterialBuilder::build(JobSystem& jobSystem) {
         goto error;
     }
 
+    if (mApiLevel < 1 || mApiLevel > UNSTABLE_MATERIAL_API_LEVEL) {
+        slog.e << "Error: api level can't be set below 1 or above unstable material level(" <<
```
pixelflinger, line 1302: I think we're using the `LOG(ERROR) << ` syntax these days.

trie94: i changed to `LOG`, but do we want to update all the other logs to use `LOG` as well? (if so maybe in a separate pr)

libs/filamat/include/filamat/MaterialBuilder.h
trie94: sounds good. i simply moved it to `MaterialEnums.h`.

libs/filamat/src/MaterialBuilder.cpp
pixelflinger: oh! sorry, I didn't realize this file had not been converted. my bad. In that case I agree you could keep it as slog if you wanted to. and yes, we'd need to update in a separate PR.

trie94: sg! i reverted back to slog, will do a follow up to change slog to LOG

PR #9536, opened 2025-12-19
Make FrameGraph resource management extensible

pixelflinger:
This refactors the FrameGraph's resource management to allow for 
user-defined custom resources.
 
A new `ResourceAllocator` template abstracts the creation and 
destruction of resources, with specializations for different resource 
types (e.g., generic, backend-dependent, textures). 
This removes hardcoded resource management logic from the `FrameGraph` 
execution loop.
 
A `ResourceCreationContext` is introduced to provide the necessary 
context for resource allocation.
 
Tests for generic and backend custom resources have been added to 
validate the new system.

### DIFF
--- a/filament/src/fg/FrameGraphDummyLink.h
+++ b/filament/src/fg/FrameGraphDummyLink.h
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2021 The Android Open Source Project
+ * Copyright (C) 2025 The Android Open Source Project
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.


show50726:
Is there any plan for adding other type of resources in the near future?

I will need to update the fgviewer function (`getFrameGraphInfo`) to get the correct properties for different resources...

pixelflinger:
> Is there any plan for adding other type of resources in the near future?
> 
> I will need to update the fgviewer function (`getFrameGraphInfo`) to get the correct properties for different resources...

Yes there is. I'm thinking about probably 2 types:
- a "dummy" resource (already added) only used to create dependencies between passes
- a "RenderPass" resource (will be purely a CPU resources)

filament/src/fg/FrameGraphDummyLink.h
```diff
@@ -0,0 +1,31 @@
+/*
+ * Copyright (C) 2021 The Android Open Source Project
```
poweifeng:
```suggestion
 * Copyright (C) 2025 The Android Open Source Project
```

PR #9541, opened 2025-12-19
vk: move swapchain acquire() to as late as possible

poweifeng:
makeCurrent is not meant to acquire the swapchain; we should do so at the momemnt we need it, to shorten the amount of time a swapchain image/buffer needs to be held.

### DIFF
--- a/filament/include/filament/Engine.h
+++ b/filament/include/filament/Engine.h
@@ -194,7 +194,7 @@ public:
     using Driver = backend::Driver;
     using GpuContextPriority = backend::Platform::GpuContextPriority;
     using AsynchronousMode = backend::AsynchronousMode;
-    using AsyncCallbackType = std::function<void(void* UTILS_NULLABLE)>;
+    using AsyncCompletionCallback = std::function<void(void* UTILS_NULLABLE)>;
     using AsyncCallId = backend::AsyncCallId;
 
     /**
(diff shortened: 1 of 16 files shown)


pixelflinger:
@poweifeng change looks good, but it fails the presubmit, triggering an assert.

filament/backend/src/vulkan/VulkanDriver.cpp
rafadevai, line 1: what is this trying to fix?

I think its better to do the acquire as early as possible, why wait until the middle of the frame to acquire the new image?

if all the images are currently in use, why start working on the next frame?

poweifeng: because we've set `vkAcquireNextImageKHR` to be a blocking call, and that's currently associated with makeCurrent, which happens at the beginning of the frame.  

Assume all of the swapchain images are currently being queued or presented, then we'll be blocking at the beginning of the frame.  But if we are rendering into a shadow map, let's say, then we shouldn't be blocking on trying to acquire the swapchain image at that point. So this change just moves it to the place where we're actually rendering into the swapchain.

rafadevai: I think if you already into the point of needing to wait for an image to be released, then the backend should just pace and not submit anymore work.

poweifeng: @pixelflinger can explain the finer points of this.

But in my mind, the swapchain images being queued does not equal to GPU being busy. The compositor has its own logic/reasoning for holding or releasing buffers. This is a bit like holding a lock; we want our critical section to be as little as possible.

PR #9542, opened 2025-12-20
Add async support check to backend

z3moon:
BUGS=[442921995]


filament/src/details/Engine.cpp
```diff
@@ -1528,8 +1528,10 @@ Engine::FeatureLevel FEngine::setActiveFeatureLevel(FeatureLevel featureLevel) {
 }
 
 bool FEngine::isAsynchronousOperationSupported() const noexcept {
+    DriverApi& driver = const_cast<FEngine*>(this)->getDriverApi();
     return features.backend.enable_asynchronous_operation &&
-        mConfig.asynchronousMode != AsynchronousMode::NONE;
+            mConfig.asynchronousMode != AsynchronousMode::NONE &&
+            driver.isAsynchronousOperationSupported();
```
pixelflinger: I think this is okay, but in reality `mConfig.asynchronousMode` should never be anything other than `NONE` if either the feature is off or the driver doesn't support it. And technically this value can be cached entirely, since it can never change after the engine is created.

I think it okay to document that setting the async mode can silently fail if not supported, as long as there is a query (e.g. getAsynchronousMode()) and it returns `false` when not supported or not enabled.

make sense?

z3moon: Done. Also I renamed the name of the query to Engine::isAsynchronousModeEnabled to be more specific.

pixelflinger: I don't understand why the whole expression cannot be cached into a single boolean, that is initialized at engine start?

z3moon: Yeah some comparisons were redundant so I removed them. Now it's cached at each backend. We cannot cache it at the frontend level because we don't know whether a backend supports asynchronous operations until it's initialized.

PR #9543, opened 2025-12-21
Fix dispersion with uber-shaders

### LINKED ISSUE #9540: the dispersion crashes with gltf_viewer's `-u` (ubershader) option
@Nemi26 the dispersion doesn't work with gltf_viewer's `-u` (ubershader) option; it crashes. FYI. 
Do you want to give it a shot at fixing it, or should I revert the change for now?

Also, I do see the effect working with khronos' DragonDispersion test, but not with the `DispersionTest`.

_Originally posted by @pixelflinger in https://github.com/google/filament/issues/9471#issuecomment-3676815915_

pixelflinger:
Fixes #9540


pixelflinger:
@Nemi26

PR #9544, opened 2025-12-21
improved the dispersion computation

pixelflinger:
- Split the function so that the 3-components loop is more apparent.
- Factored out the parts of the computation that don't depend on the index of refraction, so they're only computed once.
- Moved the lod-from-roughness computation out of the loop as well.  It does depend on the IOR, but so little that it's not worth it.

The most important result of this change is that the code is now in good shape if we wanted to change the implementation of dispersion for these cases (for e.g.) :
- do the computation in the XYZ space
- use more than 3 wavelengths samples
- or use a stochastic approach, trading banding for noise


pixelflinger:
@poweifeng the renderdiff difference is just one pixel and barely visible and only for gles. I think we can accept the new image as master (I just don't know how).

PR #9552, opened 2025-12-23
Cache prewarm base case

anishmgoyal:
Prevents hitching for most pipelines, on certain devices with Turnip-based drivers

### DIFF
--- a/filament/backend/include/backend/Program.h
+++ b/filament/backend/include/backend/Program.h
@@ -24,7 +24,6 @@
 #include <backend/DriverEnums.h>
 
 #include <array>
-#include <optional>
 #include <tuple>
 #include <utility>
 #include <variant>
--- a/filament/backend/src/vulkan/VulkanPipelineCache.cpp
+++ b/filament/backend/src/vulkan/VulkanPipelineCache.cpp
@@ -243,8 +243,6 @@ VkPipeline VulkanPipelineCache::createPipeline(
         .scissorCount = 1,
     };
 
-    // Note - if we ever add more possible states, we should call `reserve()`
-    // on this vector.
     constexpr size_t maxDynamicStates = 3;
     size_t numDynamicStates = 2;
     VkDynamicState enabledDynamicStates[maxDynamicStates] = {


filament/backend/src/vulkan/VulkanPipelineCache.cpp
```diff
@@ -107,70 +127,137 @@ void VulkanPipelineCache::bindPipeline(VulkanCommandBuffer* commands) {
...
+            .depthCompareOp = SamplerCompareFunc::L,
+            .depthBiasConstantFactor = 0.f,
+            .depthBiasSlopeFactor = 0.f,
+        },
+        .layout = layout,
+    };
+    PipelineDynamicOptions dynamicOptions {
+        .useDynamicVertexInputState = true,
+        .useDynamicRenderPasses = true,
+        .stereoscopicType = stereoscopicType,
```
rafadevai, line 172: we need to check if the program uses multiview or not

anishmgoyal:
@poweifeng @phoenixxxx @rafadevai

filament/backend/include/backend/Program.h
```diff
@@ -51,7 +52,7 @@ class Program {
     };
 
     using SpecializationConstant = std::variant<int32_t, float, bool>;
-    using DescriptorSetLayoutArray = std::array<DescriptorSetLayout, MAX_DESCRIPTOR_SET_COUNT>;
+    using DescriptorSetLayoutArray = std::array<std::optional<DescriptorSetLayout>, MAX_DESCRIPTOR_SET_COUNT>;
```
anishmgoyal: @poweifeng @pixelflinger fyi - realized that because these are structs, they aren't nullable in the same way as a descriptor set. this was my solution here, but let me know if there's an alternative you'd prefer.

note that we only set 3 of the 4 indices here, as MAX_DESCRIPTOR_SET_COUNT is 4. so, there is one guaranteed to be unused every time.

filament/backend/src/vulkan/VulkanPipelineCache.cpp
anishmgoyal: i agree that it's good practice to use that variable and will update this. that said, fyi - this value comes from the engine:

https://github.com/google/filament/blob/main/filament/backend/include/backend/Program.h#L202

filament/backend/include/backend/Program.h
rafadevai: Why do we need optional here?

filament/backend/src/vulkan/VulkanDriver.cpp
```diff
@@ -796,6 +797,30 @@ void VulkanDriver::createProgramR(Handle<HwProgram> ph, Program&& program, utils
...
+    }
+
+    // If async prewarming is enabled, let's find the proper layout and build the pipeline.
+    VulkanDescriptorSetLayout::DescriptorSetLayoutArray vkLayouts {};
+    for (size_t i = 0; i < MAX_DESCRIPTOR_SET_COUNT; ++i) {
+        if (!program.getDescriptorSetLayouts()[i].has_value()) {
+            continue;
+        }
+        DescriptorSetLayout layoutDescription = *program.getDescriptorSetLayouts()[i];
+        auto layoutHandle = mResourceManager.allocHandle<VulkanDescriptorSetLayout>();
```
rafadevai, line 809: are we leaking the handles? seems like it

filament/backend/src/vulkan/VulkanPipelineCache.cpp
```diff
@@ -126,7 +126,63 @@ void VulkanPipelineCache::bindPipeline(VulkanCommandBuffer* commands) {
...
+        .stereoscopicViewCount = stereoscopicViewCount,
+    };
+
+    CallbackManager::Handle cmh = mCallbackManager.get();
+    auto token = std::make_shared<ProgramToken>();
+    mCompilerThreadPool.queue(priority, token, [this, key, dynamicOptions, cmh]() mutable {
+        createPipeline(key, dynamicOptions);
+        mCallbackManager.put(cmh);
```
rafadevai, line 180: we need to destroy the dummy pipeline after creating it, otherwise we leak memory

filament/backend/include/backend/Program.h
pixelflinger: Yeah I don't love this.

I would prefer that we create a structure like

```
struct Program::DescriptorSetLayoutBinding {
  descriptor_set_t set;
  DescriptorSetLayout layout;
};

using DescriptorSetLayoutBindings = utils::FixedCapacityVector<Program::DescriptorSetLayoutBinding>;
```

Note that there is already structure called `backend::DescriptorSetLayoutBinding` and that's a really really bad  name (my fault!). That structure is actually more like `DescriptorSetLayoutDescriptor`. But we can for now deal with it because they're not in the same namespace.

If you can, I would appreciate it if you could do a pre-CL that cleans up my mess. Just rename `backend::DescriptorSetLayoutBinding` to `backend::DescriptorSetLayoutDescriptor`.

filament/backend/include/backend/Program.h
```diff
@@ -154,7 +155,7 @@ class Program {
         return *this;
     }
 
-    const DescriptorSetLayoutArray& getDescriptorSetLayouts() const noexcept {
+    inline const DescriptorSetLayoutArray& getDescriptorSetLayouts() const noexcept {
```
pixelflinger: don't need the `inline` keyword here.

filament/backend/src/vulkan/VulkanPipelineCache.cpp
```diff
@@ -180,15 +240,22 @@ VkPipeline VulkanPipelineCache::createPipeline(const PipelineKey& key) noexcept
         .viewportCount = 1,
         .scissorCount = 1,
     };
-    VkDynamicState dynamicStateEnables[] = {
+
+    // Note - if we ever add more possible states, we should call `reserve()`
+    // on this vector.
+    std::vector<VkDynamicState> enabledDynamicStates = {
```
rafadevai: since we are trying to use less std::vector, use an array with a fixed size?

filament/backend/src/vulkan/VulkanPipelineCache.h
```diff
@@ -104,8 +104,16 @@ class VulkanPipelineCache {
     // compiling the pipeline on the main thread at draw time. This is very dependent
     // on the implementation of the driver on the current device; it's expected to work
     // on devices with VK_EXT_vertex_input_dynamic_state and VK_KHR_dynamic_rendering.
-    void asyncPreloadCache(fvkmemory::resource_ptr<VulkanProgram> program,
-                           VkPipelineLayout layout);
+    void asyncPrewarmCache(const VulkanProgram& program, VkPipelineLayout layout,
+        StereoscopicType stereoscopicType, uint8_t stereoscopicViewCount,
```
rafadevai: fix format

filament/backend/src/vulkan/VulkanPipelineCache.h
```diff
@@ -182,6 +190,19 @@ class VulkanPipelineCache {
...
+    // pipeline (if supported).
+    struct PipelineDynamicOptions {
+        // Requires `VK_EXT_vertex_input_dynamic_state` to enable.
+        bool useDynamicVertexInputState = false;
+        // Requires `VK_KHR_dynamic_rendering` to enable.
+        bool useDynamicRenderPasses = false;
+        // Only used if `useDynamicRenderPasses` is true.
+        StereoscopicType stereoscopicType = StereoscopicType::NONE;
+        // Only used if `stereoscopicType` is `StereoscopicType::MULTIVIEW`.
+        uint8_t stereoscopicViewCount = 2;
```
rafadevai, line 203: its fine to use 0 here as default

filament/backend/src/vulkan/VulkanPipelineCache.cpp
```diff
@@ -150,27 +206,31 @@ VkPipeline VulkanPipelineCache::createPipeline(const PipelineKey& key) noexcept
...
     VkVertexInputAttributeDescription vertexAttributes[VERTEX_ATTRIBUTE_COUNT];
     VkVertexInputBindingDescription vertexBuffers[VERTEX_ATTRIBUTE_COUNT];
-    for (uint32_t i = 0; i < VERTEX_ATTRIBUTE_COUNT; i++) {
-        if (key.vertexAttributes[i].format > 0) {
-            vertexAttributes[numVertexAttribs++] = key.vertexAttributes[i];
-        }
-        if (key.vertexBuffers[i].stride > 0) {
-            vertexBuffers[numVertexBuffers++] = key.vertexBuffers[i];
+    if (!dynamicOptions.useDynamicVertexInputState) {
+        // Expand our size-optimized structs into the proper Vk structs.
```
poweifeng: nit: this comment should probably be before all the expansion.

filament/backend/src/vulkan/VulkanPipelineCache.cpp
poweifeng: how about just an std::array, and use either 2 or 3 as dynamicStateCount?

filament/backend/include/backend/Program.h
anishmgoyal: i'll remove, but i'm curious - why? this is a single-line, solely-returning function that is non-virtual, defined in a header. am i missing something?

filament/backend/src/vulkan/VulkanPipelineCache.cpp
```diff
@@ -351,6 +439,14 @@ void VulkanPipelineCache::bindVertexArray(VkVertexInputAttributeDescription cons
     }
 }
 
+void VulkanPipelineCache::notifyCachePrewarmComplete(CallbackHandler* handler,
```
poweifeng: nit: would prefer "addCachePrewarmCompleteCallback" over "notifyCachePrewarmComplete"

filament/backend/src/vulkan/VulkanPipelineCache.cpp
anishmgoyal: sure. i can do that or fixedcapacityvector, which do you prefer?

filament/backend/src/vulkan/VulkanPipelineCache.cpp
anishmgoyal: sure, i can move it down to before line 217

filament/backend/src/vulkan/VulkanPipelineCache.cpp
poweifeng: std::array, FixedCapacityVector still allocates on the heap afaik

filament/backend/src/vulkan/VulkanPipelineCache.cpp
anishmgoyal: opted to remove this comment. i feel it's redundant, as that's what the purpose of the entire function is

filament/backend/src/vulkan/VulkanDriver.cpp
anishmgoyal: these are ref counted, are they not? a resource_ptr is allocated, and then by the end of the function it goes out of scope, and therefore the ref count should hit zero.

filament/backend/src/vulkan/VulkanPipelineCache.h
anishmgoyal: the format within this file itself is inconsistent, but i will match the original formatting for this function.

filament/backend/src/vulkan/VulkanPipelineCache.h
anishmgoyal: 2 is used as the default across filament for this value. since it'll be overwritten anyways, i'd rather follow the convention.

filament/backend/src/vulkan/VulkanDriver.cpp
rafadevai: okay, so the handle is not leaked, you're right

filament/backend/src/vulkan/VulkanPipelineCache.cpp
```diff
@@ -180,15 +242,24 @@ VkPipeline VulkanPipelineCache::createPipeline(const PipelineKey& key) noexcept
         .viewportCount = 1,
         .scissorCount = 1,
     };
-    VkDynamicState dynamicStateEnables[] = {
+
+    // Note - if we ever add more possible states, we should call `reserve()`
```
rafadevai: no need for this comment anymore

filament/backend/src/vulkan/VulkanPipelineCache.h
rafadevai: okay

filament/backend/src/vulkan/VulkanPipelineCache.cpp
rafadevai: I didnt notice you already had the check before calling this function.

anishmgoyal: i didn't, added that in as per your comment :)

poweifeng (approved):

filament/backend/include/backend/Program.h
pixelflinger: It's just redundant (and IDEs complain about it).  when a method is defined inside a class, it's automatically `inline`, which means it won't export a symbol and might be actually inlined (compiler decides). Adding the `inline` keyword as no effect AFAIK in that case.

PR #9553, opened 2025-12-23
Fix garbage data getting added to the source material

trie94:
Need to explicitly add a null terminator so that we don't get garbage values in the source material. For example some source materials were returning extra two letters like (`} }`) at the end. (It doesn't always happen, and that's why the unit tests were passing)

Alternatively this can be fixed by passing in the size to the function, but since we are using string_view, this seems to make more sense.

FIXES=471237867


tools/matc/src/matc/MaterialCompiler.cpp
```diff
@@ -78,9 +78,12 @@ bool MaterialCompiler::run(const matp::Config& config) {
 
     // Now that the buffer is mutated, we need to update the buffer pointer and the size
     // before parsing.
-    size = resolvedString.size();
+    size = resolvedString.size() + 1;
     auto modifiedBuffer = std::make_unique<char[]>(size);
     std::strncpy(modifiedBuffer.get(), resolvedString.c_str(), size);
+    // We need to append the null terminator here, otherwise some garbage data might follow when
+    // including source material.
+    modifiedBuffer[size - 1] = '\0';
```
pixelflinger: I think that's okay, but ideally we would never rely on the nul temrinator and always use a length. I think the code might not be safe about that everywhere.

trie94: yeah, like you said, it works for the other functions since they are using "size" specifically. and actually appending source material does use the size, so i am a bit baffled what's causing this issue.

i have also tried `builder.materialSource(std::string_view(buffer.get(), size));`, but this still gets some garbage values.

but regardless i think it's safer to append null terminator after `strncpy`.

trie94: actually `builder.materialSource(std::string_view(buffer.get(), size));` --> i could use this approach. would this be preferred?

trie94: decided to go with this approach.

PR #9558, opened 2025-12-24
Add external format to cache prewarming

anishmgoyal:
Please refer to the latest commit. This depends on other base PRs.

### DIFF
--- a/filament/backend/src/vulkan/VulkanDriver.cpp
+++ b/filament/backend/src/vulkan/VulkanDriver.cpp
@@ -852,6 +852,10 @@ void VulkanDriver::createProgramR(Handle<HwProgram> ph, Program&& program, utils
         mStereoscopicEyeCount,
         program.getPriorityQueue());
 
+    if (!hasExternalSamplers) {
+        return;
+    }
+
     // If we have external samplers, let's do this again with the external samplers
     // specified.
     for (const auto& format : mContext.getPipelineCachePrewarmExternalFormats()) {


filament/backend/src/vulkan/VulkanDriver.cpp
```diff
@@ -803,26 +803,82 @@ void VulkanDriver::createProgramR(Handle<HwProgram> ph, Program&& program, utils
...
+
+        // Update all layouts to use the external samplers.
+        for (size_t i = 0; i < MAX_DESCRIPTOR_SET_COUNT; ++i) {
+            if (!layouts[i]) {
+                continue;
+            }
+
+            utils::FixedCapacityVector<VkSampler> externalSamplers (layouts[i]->bitmask.externalSampler.count(), externalSampler);
+            vkLayouts[i] = mDescriptorSetLayoutCache.getVkLayout(
+                layouts[i]->bitmask, layouts[i]->bitmask.externalSampler, externalSamplers);
```
poweifeng, line 882: this is probably wrong logically, but ok practically.  You're using the same external sampler parameter for all the external samplers in the layout.  That might not be true.  Would like a comment to explain the difference here.

filament/backend/include/backend/platforms/VulkanPlatform.h
```diff
@@ -328,6 +338,30 @@ class VulkanPlatform : public Platform, utils::PrivateImplementation<VulkanPlatf
...
+    /**
+     * For cache prewarming, if external samplers are present, we need to build
+     * the fake pipeline using the proper formats specified. Since there's no way to
+     * get these at material build time, we allow the app to register them before
+     * creating materials.
+     *
+     * @param format The format, containing the external format value which should be
+     *               extracted from an AHardwareBuffer.
+     */
+    inline void registerCachePrewarmExternalFormat(const ExternalYcbcrFormat& format) noexcept {
```
poweifeng: don't you expect this list to be static per platform?  why have a register method for it?  Why not just return a fixed list?

and does it have to be on the platform?  Can it live in VulkanContext?

filament/backend/src/vulkan/platform/VulkanPlatformAndroid.cpp
```diff
@@ -563,4 +618,66 @@ bool VulkanPlatformAndroid::queryFrameTimestamps(SwapChain const* swapchain, uin
     return vulkanSwapchain->queryFrameTimestamps(frameId, outFrameTimestamps);
 }
 
+void VulkanPlatformAndroid::registerAndroidExternalFormatsForCachePrewarm() {
```
poweifeng: can all this logic live somewhere in the subclass of VulkanPlatformAndroid?

filament/backend/src/vulkan/VulkanDriver.cpp
anishmgoyal: since this is a fake pipeline, we don't know what samplers will actually be used. it seems like what matters for prewarming is simply compiling each potential ycbcr conversion that may show up, on the device we're targeting.

filament/backend/include/backend/platforms/VulkanPlatform.h
anishmgoyal: i agree that this should be in vulkancontext. i plan on moving it there, just need to figure out one bug that's popped up first. i don't always want to call this since it creates a few fake ahardwarebuffers, but i will likely call this from createDriver() if driverconfig indicates that prewarming is enabled.

filament/backend/src/vulkan/platform/VulkanPlatformAndroid.cpp
```diff
@@ -202,6 +202,61 @@ uint32_t selectMemoryTypeForExternalImage(VkPhysicalDevice physicalDevice, VkDev
...
+#define AHARDWAREBUFFER_FORMAT_YCbCr_P210 static_cast<AHardwareBuffer_Format>(0x3c)
+#else
+#error "Please remove this #if / #define."
+#endif
+
+constexpr std::array kExternalSamplerFormats = {
+    // Standard video playback
+    ExternalSamplerFormatDescription {
+        AHARDWAREBUFFER_FORMAT_Y8Cb8Cr8_420,
+        AHARDWAREBUFFER_USAGE_GPU_SAMPLED_IMAGE | AHARDWAREBUFFER_USAGE_VIDEO_ENCODE,
```
rafadevai: Dont think encode flag is necessary because the buffer wont go through an encoding pippeline, in any case would be a decoder.

filament/backend/src/vulkan/platform/VulkanPlatformAndroid.cpp
anishmgoyal: i could move it into XrVulkanPlatform in impress, sure. i just need to make sure that class has a ref to the context

filament/backend/src/vulkan/platform/VulkanPlatformAndroid.cpp
```diff
@@ -202,6 +202,61 @@ uint32_t selectMemoryTypeForExternalImage(VkPhysicalDevice physicalDevice, VkDev
...
+        VK_SAMPLER_YCBCR_MODEL_CONVERSION_YCBCR_709,
+    },
+    // HDR video playback
+    ExternalSamplerFormatDescription {
+        AHARDWAREBUFFER_FORMAT_YCbCr_P010,
+        AHARDWAREBUFFER_USAGE_GPU_SAMPLED_IMAGE | AHARDWAREBUFFER_USAGE_VIDEO_ENCODE,
+        VK_SAMPLER_YCBCR_RANGE_ITU_NARROW,
+        VK_SAMPLER_YCBCR_MODEL_CONVERSION_YCBCR_2020,
+    },
+    // Camera preview. Note: ITU_FULL may cause linker errors with certain formats,
```
rafadevai: what do you mean about linker errors? it fails to compile the pipeline?

filament/backend/src/vulkan/platform/VulkanPlatformAndroid.cpp
rafadevai: why? seems fine to me to live in the android section, is all related to android

filament/backend/src/vulkan/VulkanDriver.cpp
poweifeng: yes, but then technically, you'll need to try all the possible combinations of external samplers in the descriptor set layout.

For example, if you have two external samplers in this descriptor set. one could be av1 video, and one could be hdr photo (or something).  But here you'll only consider the descriptor set layout for both av1 or both hdr.

It's not a big deal since this is an imperfect guess.  But just would like to have a comment on why we ignore the case I mentioned above.

filament/backend/src/vulkan/platform/VulkanPlatformAndroid.cpp
poweifeng: you could make the context protected if it's not already.

anishmgoyal: it currently lives in PrivateImplementation{}, and isn't a field in any of the platform classes thus far

filament/backend/src/vulkan/VulkanDriver.cpp
```diff
@@ -803,26 +803,82 @@ void VulkanDriver::createProgramR(Handle<HwProgram> ph, Program&& program, utils
...
-        *vprogram.get(),
-        layout,
-        stereoscopicType,
-        mStereoscopicEyeCount,
-        program.getPriorityQueue());
+    // If there are no external samplers, that's it! We have enough to compile the
+    // fake pipeline at this point. If there are, though, this is unnecessary - there
+    // MUST be some external samplers defined. In that case, we'll go on to start
+    // instantiating the YCbCr conversions for them.
+    if (!hasExternalSamplers) {
```
rafadevai: So we do need to build both version when an external sampler is present.

Is possible to have a material that has only 1 external sampler, which can be used with non-YUV images and in this case when building the pipeline it will hitch once.

filament/backend/src/vulkan/VulkanDriver.cpp
rafadevai: You are right and to avoid messing with all possible combinations of samplers, this is just the most common approach.

Definitely we should add a comment mentioning this

filament/backend/src/vulkan/VulkanDriver.cpp
```diff
@@ -803,26 +803,82 @@ void VulkanDriver::createProgramR(Handle<HwProgram> ph, Program&& program, utils
...
+                .chromaFilter = SamplerMagFilter::NEAREST,
+            },
+            .format = VK_FORMAT_UNDEFINED,
+            .externalFormat = format.externalFormat,
+        };
+        VkSamplerYcbcrConversion vkConversion = mYcbcrConversionCache.getConversion(ycbcrConversion);
+        VkSampler externalSampler = mSamplerCache.getSampler({.sampler = {}, .conversion = vkConversion});
+
+        // Update all layouts to use the external samplers.
+        for (size_t i = 0; i < MAX_DESCRIPTOR_SET_COUNT; ++i) {
```
rafadevai, line 871: actually we dont need to iterate over all the descriptor sets, we only need to check the DescriptorSetBindingPoints::PER_MATERIAL

filament/backend/src/vulkan/platform/VulkanPlatformAndroid.cpp
rafadevai: This is still really useful even when using GPL, so I dont see any problem being here

poweifeng: imo this is a pretty unusual for a typical android app using filament.  I can, at most, imagine AR/camera passthrough needing this.  But that's just one external sampler format.

rafadevai: well even the camera could have different versions but if the material doesnt use an external sampler, then this would do nothing

poweifeng: but if my material does have external sampler, then it'll iterate through additional formats (and build additional pipelines).

rafadevai: and thats expected because we dont know which format it will use, so thats the tradeoff when doing the parallel compilation for Vulkan otherwise it will spend a lot of time building the pipeline when recording the commands.

poweifeng: but I'll be doing compilation for formats that are not relevant for my app (outside of this use case).

poweifeng: > it currently lives in PrivateImplementation{}, and isn't a field in any of the platform classes thus far

yeah, this is pretty tricky.  I think we can keep that list in platform then.

filament/backend/src/vulkan/platform/VulkanPlatformAndroid.cpp
anishmgoyal: there's no USAGE_VIDEO_DECODE; and in any case, this was sufficient to get me a format id that was being used by the youtube video decoder

filament/backend/src/vulkan/platform/VulkanPlatformAndroid.cpp
anishmgoyal: yes, but this is something you'd notice really early on - i.e. the first time you try to watch a youtube video. this set of formats does not cause issues though.

filament/backend/src/vulkan/platform/VulkanPlatformAndroid.cpp
rafadevai: > but I'll be doing compilation for formats that are not relevant for my app (outside of this use case).

100% and even in the case we are targeting to use this, thats still the case but there's no way to know before hand which exactly are needed and which arent.
 
On the other hand, apps enable parallel compilation because they want to avoid missing frames because of pipeline compilations, so seems an okay trade off here.

filament/backend/src/vulkan/VulkanDriver.cpp
anishmgoyal: so i don't doubt that this could happen, but i've yet to run into a case where there are hitches with this approach in:

youtube 4k vids
youtube 360 vids
youtube sd vids
camera videos immersive
camera photos immersive
camera videos 2d
camera photos 2d
chrome

that tells me one of two things. either:

1. this isn't happening anywhere on our target device, or
2. it happens, but compiling with a ycbcr conversion in the descriptors generates enough of a stub in the cache for rgb external samplers to result in a cache hit

either way - unless we can show that this actually causes issues, i'd rather not add an extra pipeline compilation until we can prove we need it, as we're fighting to achieve parity between gl and vulkan performance.

if, with this build, you're able to demonstrate a case where there is a hitch, i'm happy to update this

filament/backend/src/vulkan/platform/VulkanPlatformAndroid.cpp
poweifeng: I guess I'm saying that we could probably have a list of external sampling format that's empty. And then the client (whoever subclasses) should be filling it up. And if I'm the app writer, I'll know, with some certainty through experimentation, what to put there.  

But as is right now, this code makes many assumption for one particular use case.

filament/backend/src/vulkan/VulkanDriver.cpp
anishmgoyal: hmm, makes sense. i'll update this. in any case, it's a no-op for any descriptor set that doesn't have an external sampler, and doesn't lead to any extra compiles, so i don't think it hurts us to have this.

filament/backend/src/vulkan/platform/VulkanPlatformAndroid.cpp
rafadevai: that's what im saying, just leave GPU_SAMPLED_IMAGE and remove the other one

filament/backend/src/vulkan/platform/VulkanPlatformAndroid.cpp
rafadevai: Not really, there are material that dont need the 6 some might need 2, so hard to say. So we just overdo it to cover a good amount of possibilities.

anishmgoyal: @poweifeng if you're saying you're fine with the final list living in platform, but prefer that the XrVulkanPlatform class fills that list vs VulkanPlatformAndroid, i'm happy to make that change. can you confirm that that's specifically the issue here?

i can also have VulkanPlatform use registerExternalFormat as a protected method that adds things to VulkanContext directly.

filament/backend/src/vulkan/platform/VulkanPlatformAndroid.cpp
anishmgoyal: technically that gives us one less possible format number, since then we'd get the same format number as the other two. i think that shouldn't be an issue for hitching but i will test

filament/backend/src/vulkan/VulkanDriver.cpp
anishmgoyal: will add a comment, sure

filament/backend/src/vulkan/platform/VulkanPlatformAndroid.cpp
poweifeng: yes, all of the above are correct.

filament/backend/include/backend/platforms/VulkanPlatform.h
anishmgoyal: this function is now protected in VulkanPlatform. it calls an adder method on VulkanContext, so the getter was removed from VulkanPlatform. this hides the API a little bit better. note that the externalycbcrformat struct still needs to exist on VulkanPlatform so that it's visible from impress.

filament/backend/src/vulkan/platform/VulkanPlatformAndroid.cpp
anishmgoyal: updated

filament/backend/src/vulkan/platform/VulkanPlatformAndroid.cpp
anishmgoyal: we can revisit this conversation when discussing the change to impress, since this has been moved to that repo

filament/backend/src/vulkan/VulkanDriver.cpp
anishmgoyal: just a minor detail i need to sort out, which is that this constant is not defined somewhere accessible from the backend. will see the best way to reference that constant, or just copy it as is done in https://github.com/google/filament/blob/5eb0f7d3ecb286df88a4b820db2076ae9145167a/filament/backend/src/opengl/ShaderCompilerService.cpp#L733

anishmgoyal: after discussion with @poweifeng , this would break abstraction we'd like to maintain. since this should be a no-op for all other descriptor sets, let's keep this as is. thoughts, @rafadevai?

poweifeng: if you want the right pipeline layout, don't you need all the descriptor set layouts?

anishmgoyal: you do, it's just that the only one we expect to add external samplers to is the PER_MATERIAL one. in theory, none of the others should use external samplers

filament/backend/src/vulkan/VulkanContext.h
```diff
@@ -90,6 +91,19 @@ struct VulkanContext {
...
+    /**
+     * For pipeline cache prewarming, if external samplers are present, we need to build
+     * the fake pipeline using the proper formats specified. Since there's no way to
+     * get these at material build time, we allow the app to register them before
+     * creating materials.
+     *
+     * @param format The format, containing the external format value which should be
+     *               extracted from an AHardwareBuffer.
+     */
+    inline void addPipelineCachePrewarmExternalFormat(const VulkanPlatform::ExternalYcbcrFormat& format) noexcept {
```
poweifeng: I dont' think you can have `noexcept` here.

filament/backend/src/vulkan/VulkanDriver.cpp
poweifeng: I see.  If you have the right bind point for user material, then the loop can be outside of the format loop.

filament/backend/src/vulkan/VulkanContext.h
anishmgoyal: right - good point, vector::push_back is not noexcept. updated.

filament/backend/src/vulkan/VulkanDriver.cpp
rafadevai: Lets ignore my comment since it break how filament do things.

PR #9568, opened 2026-01-05
Remove unnecessary log statement

anishmgoyal:
It is not an error for a device to not have this API.


anishmgoyal:
@pixelflinger

PR #9570, opened 2026-01-05
Improve frame_generator sample tool

pixelflinger:
- Add support for vector parameters (float2, float3, float4) in the  parameters file using `{x, y, z}` syntax.
- Allow comments (`#`, `//`) and empty lines in the parameters file.
- Make the end value optional in the parameters file  (defaults to start value).
- Add `--size` / `-S` option to specify viewport dimensions.
- Optimize rendering by pre-filtering active material parameters.
- Remove hardcoded default parameters and silently ignore missing  parameters to support shared configuration files.
- Add documentation for the parameters file format.

### DIFF
--- a/samples/frame_generator.cpp
+++ b/samples/frame_generator.cpp
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2018 The Android Open Source Project
+ * Copyright (C) 2025 The Android Open Source Project
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


samples/frame_generator.cpp
```diff
@@ -180,18 +220,30 @@ static int handleCommandLineArgments(int argc, char* argv[], Config* config) {
...
-                    // keep scale of 1.0
+                    config->scale = 1.0;
                 }
                 break;
             case 'b':
                 try {
-                    g_clearColor = (uint32_t) std::stoul(arg, nullptr, 16);
+                    g_clearColor = uint32_t(std::stoul(arg, nullptr, 16));
+                } catch (std::invalid_argument& e) {
+                    g_clearColor = {};
```
poweifeng, line 232: nit: some warning logging here would be helpful

PR #9572, opened 2026-01-05
dispersion in XYZ space

pixelflinger:
- spectral reconstructions with 4 samples
- unroll the whole dispersion computation to improve performance (batch texture fetches and reuse common values)
- tool to generate the matrices for spectral integration

XYZ-space-dispersion

### DIFF
--- a/tools/specgen/specgen.cpp
+++ b/tools/specgen/specgen.cpp
@@ -1,0 +1,650 @@
+/*
+ * Copyright (C) 2025 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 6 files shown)


tools/specgen/specgen.cpp
show50726, line 1: This is so cool! Besides Wyman's work, are there any technical references you’d recommend for learning more about this topic?

pixelflinger: Gemini was very useful and on point. for that kind of stuff, you can use it as an interactive tutor. e.g. you can ask about the XYZ color space, and/or dispersion in rendering engines.

PR #9573, opened 2026-01-05
feat(taa): Overhaul TAA implementation for improved quality/upscaling

pixelflinger:
This commit introduces a significant rework of the Temporal Anti-Aliasing (TAA) system, focusing on improving reconstruction quality, robustness, and introducing flexible upscaling.

Core TAA Algorithm improvements:
- Replaced the Catmull-Rom filter with a more efficient 5-tap Lanczos filter for history sampling, which includes deringing to reduce artifacts.
- The input color buffer is now properly "unjittered" using a Lanczos reconstruction filter.
- Improved the history rejection algorithm by skipping the expensive accurate clipping when the history sample is already within the neighborhood's color gamut.
- Added a new `hdr` option to properly handle HDR content by tonemapping colors before blending and untonemapping the result.
- Removed the ineffective `VARIANCE` only history rejection method.
- Added protection against negative numbers in `sqrt()` for increased stability.

TAA Upscaling:
- Replaced the boolean `upscaling` flag with a float factor, allowing for variable upscaling ratios (e.g., 1.5x, 2x).
- Upscaling now correctly adjusts viewport and projection settings.
- The TAA shader now receives viewport and resolution information to correctly handle upscaled rendering.

API and Configuration Changes:
- Deprecated the `filterWidth` TAA option as it no longer has an effect.
- Introduced the `upscaling` float property to `TemporalAntiAliasingOptions`.
- Added the `hdr` boolean property to `TemporalAntiAliasingOptions`.

Other Changes:
- Updated UI elements in the viewer and material sandbox to reflect the new TAA options.
- Updated Javascript bindings and TypeScript definitions for the new TAA settings.
- Refactored shader code for clarity and performance.


filament/include/filament/Options.h
```diff
@@ -489,16 +489,15 @@ struct MultiSampleAntiAliasingOptions {
...
  */
 struct TemporalAntiAliasingOptions {
-    float filterWidth = 1.0f;   //!< reconstruction filter width typically between 1 (sharper) and 2 (smoother)
+    float filterWidth = 1.0f;   //!< @deprecated has no effect.
     float feedback = 0.12f;     //!< history feedback, between 0 (maximum temporal AA) and 1 (no temporal AA).
     float lodBias = -1.0f;      //!< texturing lod bias (typically -1 or -2)
     float sharpness = 0.0f;     //!< post-TAA sharpen, especially useful when upscaling is true.
     bool enabled = false;       //!< enables or disables temporal anti-aliasing
-    bool upscaling = false;     //!< 4x TAA upscaling. Disables Dynamic Resolution. [BETA]
+    float upscaling = 1.0f;     //!< Upscaling factor. Disables Dynamic Resolution. [BETA]
```
show50726, line 497: Is there any constraint for the value? e.g. it should be power of 2.

pixelflinger: Because this API (the upscaling part) is in BETA, constraints are not well defined yet. In fact, upscaling still has some bugs depending on the in and out resolution.

filament/src/details/Renderer.cpp
```diff
@@ -1307,16 +1309,26 @@ void FRenderer::renderJob(RootArenaScope& rootArenaScope, FView& view) {
...
+                view.getFrameHistory(), &FrameHistoryEntry::taa,
                 taaOptions, colorGradingConfig);
-        if (taaOptions.upscaling) {
+        if (taaOptions.upscaling > 1.0f) {
             scale = 1.0f;
             scaled = false;
+
+            // xvp = vp;
+            // xvp.left = xvp.bottom = 0;
+            // svp = xvp;
```
z3moon, line 1321: nit: you may want to remove unless it's for debugging

filament/src/details/Renderer.cpp
```diff
@@ -1307,16 +1309,26 @@ void FRenderer::renderJob(RootArenaScope& rootArenaScope, FView& view) {
...
+
             UTILS_UNUSED_IN_RELEASE auto const& inputDesc = fg.getDescriptor(input);
             svp.width = inputDesc.width;
             svp.height = inputDesc.height;
-            xvp.width *= 2;
-            xvp.height *= 2;
+            // FIXME: this rounds xvp incorrectly
+
+            xvp.left   *= taaOptions.upscaling;
```
z3moon, line 1328: Maybe you could use `xvp.left = std::round(xvp.left * taaOptions.upscaling);` to prevent truncation.

filament/src/materials/antiAliasing/taa/taa.mat
```diff
@@ -91,7 +112,13 @@ material {
 
 vertex {
     void postProcessVertex(inout PostProcessVertexInputs postProcess) {
-        postProcess.vertex.xy = uvToRenderTargetUV(postProcess.normalizedUV);
+        highp vec2 uv = postProcess.normalizedUV;
+        // input UVs
+        //postProcess.vertex.xy = uvToRenderTargetUV(
+        //        materialParams.colorViewport.xy + uv * materialParams.colorViewport.zw);
```
z3moon, line 118: maybe remove?

show50726 (approved):
I'd like to see if there's any screenshot compariing the original method and improved method.

filament/src/details/Renderer.cpp
pixelflinger: yeah, it here because the upscaling is still buggy [BETA], and that's basically debugging stuff I want to keep around.

filament/src/details/Renderer.cpp
pixelflinger: no the problem is actually deeper that than,, so while this might help (I'm not even sure), this wouldn't be the correct fix anyways.

filament/src/materials/antiAliasing/taa/taa.mat
pixelflinger: same thing, it's part of what's needed to make it really work, so it's still debug code for upscaling.

PR #9577, opened 2026-01-06
Add `user` parameter to runCommandAsync

z3moon:


filament/include/filament/Engine.h
```diff
@@ -193,6 +193,7 @@ class UTILS_PUBLIC Engine {
     using Driver = backend::Driver;
     using GpuContextPriority = backend::Platform::GpuContextPriority;
     using AsynchronousMode = backend::AsynchronousMode;
+    using AsyncCallbackType = std::function<void(void* UTILS_NULLABLE)>;
```
poweifeng, line 197: nit: how about just `AsyncCallback`

filament/include/filament/Engine.h
```diff
@@ -1047,14 +1048,15 @@ class UTILS_PUBLIC Engine {
...
      * @param handler The handler from which `onComplete` is invoked. If null, it's called from the
      * main thread.
+     * @param onComplete The callback function that runs once the command has finished.
+     * @param user    The custom data that will be passed as an argument to the `onComplete`.
      * @return A unique identifier for the asynchronous call.
      */
     AsyncCallId runCommandAsync(utils::Invocable<void()>&& command,
-            backend::CallbackHandler* UTILS_NULLABLE handler,
-            utils::Invocable<void()>&& onComplete);
+            backend::CallbackHandler* UTILS_NULLABLE handler, AsyncCallbackType onComplete,
```
pixelflinger, line 1059: you're passing `onComplete` by value here, is it intended?

filament/include/filament/Engine.h
pixelflinger: also you probably need `<functional>`

pixelflinger (commented):
`onComplete` is passed by value in many places, is that intended? (as opposed to `const&`)

filament/include/filament/Engine.h
z3moon: Yes, it's intended. This allows users to decide whether to move the completion callback or not (copy) depending on their situation.

filament/include/filament/Engine.h
z3moon: Included the header. Let me rename the type name (along with others) in a follow-up PR.

PR #9584, opened 2026-01-08
make getWorldFromModelMatrix() accessible in fragment shader

pixelflinger:
FIXES=[474353927]

### DIFF
--- a/shaders/src/surface_getters.vs
+++ b/shaders/src/surface_getters.vs
@@ -2,6 +2,7 @@
 // Uniforms access
 //------------------------------------------------------------------------------
 
+#if CLIENT_MATERIAL_API_LEVEL < UNSTABLE_MATERIAL_API_LEVEL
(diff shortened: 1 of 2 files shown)


pixelflinger:
@trie94 since this is a new API in materials, is there anything I need to do w.r.t to the new versioning?

trie94:
> @trie94 since this is a new API in materials, is there anything I need to do w.r.t to the new versioning?

yeah add this check in the shader
```
#if CLIENT_MATERIAL_API_LEVEL >= UNSTABLE_MATERIAL_API_LEVEL
```
e.g., https://github.com/google/filament/blob/d465d59c3a4b8c8c90e636849453feceed5559a2/libs/filamat/tests/test_filamat.cpp#L1046

btw do we want to use the `@unstable-api` annotation? this doesn't do anything functionally though

PR #9592, opened 2026-01-09
Add glibc check to support build filament on Linux musl-based distrib…

HanYunChenLuo:
I want to use filament on the alpine, but now all of steps for linux is Ubuntu ,so i change CMakeLists.txt and a bit code for alpine build.


libs/utils/src/CallStack.cpp
```diff
@@ -27,7 +27,7 @@
 #include <stdint.h>
 #include <stdio.h>
 
-#if !defined(__ANDROID__) && !defined(WIN32) && !defined(__EMSCRIPTEN__)
+#if !defined(__ANDROID__) && !defined(WIN32) && !defined(__EMSCRIPTEN__) && defined(__GLIBC__)
```
poweifeng: I'm not super comfortable with adding an conditional for all platform.  How about something like

```
 && (!defined(__linux__) || defined(__GLIBC__))
```

CMakeLists.txt
```diff
@@ -371,10 +371,18 @@ if (LINUX)
...
+            include(CheckSymbolExists)
+            check_symbol_exists(__GLIBC__ "features.h" CLANG_WITH_GLIBC)
+            if (CLANG_WITH_GLIBC)
+                set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -stdlib=libc++")
+                link_libraries("-static-libgcc -static-libstdc++")
+                link_libraries(libc++.a)
+                link_libraries(libc++abi.a)
+            else()
+                set(CMAKE_LINK_FLAGS "${CMAKE_LINK_FLAGS} -static")
+            endif()
```
poweifeng: can we reformat this so it looks like

```
if (FILAMENT_USING_GCC)
   ...
elseif (CLANG_WITH_GLIBC)
   ...
else() # e.g. musl-based distros
   ...
endif()
```

HanYunChenLuo:
I’ve updated the code to address the review comments. Please review it again at your convenience.

poweifeng (approved):

HanYunChenLuo:
fix a bit error for static lib link.

PR #9595, opened 2026-01-09
feat(engine): Add automatic frame skipping to manage CPU/GPU latency

pixelflinger:
Introduces a new feature, disabled by default, that allows the engine to automatically skip frames when the CPU gets too far ahead of the display's refresh rate. This helps to reduce overall latency by preventing a backlog of frames from building up in the driver queue.

The feature can be enabled with the "engine.skip_frame_when_cpu_ahead_of_display" property.

To implement this, the compositor timing mechanism has been refactored. Instead of reporting an absolute `expectedPresentTime`, the backend now provides an `expectedPresentLatency` relative to vsync. This is more robust against synchronization issues with platform callbacks (like Android's Choreographer), as the latency is generally a constant value.

BUGS=[474599530]

### DIFF
--- a/filament/backend/src/vulkan/platform/VulkanPlatformAndroid.cpp
+++ b/filament/backend/src/vulkan/platform/VulkanPlatformAndroid.cpp
@@ -27,6 +27,7 @@
 
 #include <backend/DriverEnums.h>
 
+#include <utils/compiler.h>
 #include <utils/Panic.h>
 
 #include <bluevk/BlueVK.h>
@@ -542,8 +543,14 @@ bool VulkanPlatformAndroid::queryCompositorTiming(SwapChain const* swapchain,
     //        the choreographer's callback can happen before (good) or after (bad) us.
     //        This problem is mitigated by storing the latency instead of the deadline,
     //        because it generally is constant frame to frame.
-    outCompositorTiming->expectedPresentLatency =
-            preferredTimeline.expectedPresentTime - preferredTimeline.frameTime;
+    if (UTILS_LIKELY(preferredTimeline.expectedPresentTime > preferredTimeline.frameTime)) {
+        // latency can never be negative, let's be safe
+        outCompositorTiming->expectedPresentLatency =
+                preferredTimeline.expectedPresentTime - preferredTimeline.frameTime;
+    } else {
+        // fake a reasonable value (33ms)
+        outCompositorTiming->expectedPresentLatency = 33'000'000;
+    }
     outCompositorTiming->compositeDeadline = CompositorTiming::INVALID;
     outCompositorTiming->compositeInterval = CompositorTiming::INVALID;
     outCompositorTiming->compositeToPresentLatency = CompositorTiming::INVALID;
--- a/filament/src/details/Renderer.h
+++ b/filament/src/details/Renderer.h
@@ -226,7 +226,7 @@ private:
     std::function<void()> mBeginFrameInternal;
     uint64_t mVsyncSteadyClockTimeNano = 0;
     std::unique_ptr<TextureCache> mResourceAllocator{};
-    mutable int64_t mLastFrameId = -1;
+    int64_t mLastFrameId = -1;
 };
 
 FILAMENT_DOWNCAST(Renderer)
--- a/libs/utils/include/private/utils/FeatureFlagManager.h
+++ b/libs/utils/include/private/utils/FeatureFlagManager.h
@@ -65,6 +65,8 @@ public:
             struct {
                 bool disable_gpu_frame_complete_metric = true;
             } frame_info;
+            // Automatically requests a frame skip when the CPU gets too much ahead of the display. This can prevent
+            // stalls on Android, e.g. when running out buffers, but this also keeps the expected latency in check.
             bool skip_frame_when_cpu_ahead_of_display = false;
         } engine;
         struct {


filament/backend/src/opengl/platforms/PlatformEGLAndroid.cpp
```diff
@@ -296,9 +296,12 @@ bool PlatformEGLAndroid::queryCompositorTiming(SwapChain const* swapchain,
...
             mAndroidDetails.androidFrameCallback.getPreferredTimeline() };
-    outCompositorTiming->frameTime = preferredTimeline.frameTime;
-    outCompositorTiming->expectedPresentTime = preferredTimeline.expectedPresentTime;
-    outCompositorTiming->frameTimelineDeadline = preferredTimeline.frameTimelineDeadline;
+    // FIXME: expectedPresentLatency might reflect the previous frame's value because
+    //        the choreographer's callback can happen before (good) or after (bad) us.
+    //        This problem is mitigated by storing the latency instead of the deadline,
+    //        because it generally is constant frame to frame.
+    outCompositorTiming->expectedPresentLatency =
+            preferredTimeline.expectedPresentTime - preferredTimeline.frameTime;
```
poweifeng: I know very little about this, but does that mean this value can be negative if the frameTime is actually past the expectedPresentTime?

filament/src/details/Engine.h
```diff
@@ -765,18 +765,16 @@ class FEngine : public Engine {
...
-                bool assert_material_instance_texture_descriptor_set_compatible =
-                        CORRECTNESS_ASSERTION_DEFAULT;
+                bool assert_vertex_buffer_attribute_stride_mult_of_4 = CORRECTNESS_ASSERTION_DEFAULT;
+                bool assert_material_instance_texture_descriptor_set_compatible = CORRECTNESS_ASSERTION_DEFAULT;
                 bool assert_texture_can_generate_mipmap = CORRECTNESS_ASSERTION_DEFAULT;
             } debug;
             struct {
                 bool disable_gpu_frame_complete_metric = true;
             } frame_info;
+            bool skip_frame_when_cpu_ahead_of_display = false;
```
poweifeng: Maybe we can add comments about this feature above this line? It'd be helpful when reading the code.

filament/backend/src/opengl/platforms/PlatformEGLAndroid.cpp
pixelflinger: the presentation time should always be after the frame time, but that's a good point, maybe I should at least clamp to 0 or discard the value if it's nonsensical.

pixelflinger: I use 33ms if the value is negative (Which should never happen)

filament/src/details/Renderer.cpp
```diff
@@ -299,49 +302,78 @@ void FRenderer::skipFrame(uint64_t vsyncSteadyClockTimeNano) {
...
     FEngine& engine = mEngine;
     FEngine::DriverApi& driver = engine.getDriverApi();
-    return mFrameSkipper.shouldRenderFrame(driver);
+    bool const renderFrame = mFrameSkipper.shouldRenderFrame(driver);
+    if (renderFrame && engine.features.engine.skip_frame_when_cpu_ahead_of_display) {
+        // see if we have another reason to skip
+        auto history = mFrameInfoManager.getFrameInfoHistory();
+        for (size_t i = 0, c = history.size(); i < c; i++) {
+            FrameInfo const& info = history[i];
+            if (int32_t(info.frameId - mLastFrameId) <= 0) {
```
z3moon, line 318: nit: `(int64_t)info.frameId <= mLastFrameId`

filament/src/details/Renderer.h
```diff
@@ -226,6 +226,7 @@ class FRenderer : public Renderer {
     std::function<void()> mBeginFrameInternal;
     uint64_t mVsyncSteadyClockTimeNano = 0;
     std::unique_ptr<TextureCache> mResourceAllocator{};
+    mutable int64_t mLastFrameId = -1;
```
z3moon: is this needed? I see it's only modified in beginFrame, which is not const

filament/backend/src/vulkan/platform/VulkanPlatformAndroid.cpp
```diff
@@ -538,9 +538,12 @@ bool VulkanPlatformAndroid::queryCompositorTiming(SwapChain const* swapchain,
...
     AndroidFrameCallback::Timeline const preferredTimeline{
         mAndroidDetails.androidFrameCallback.getPreferredTimeline() };
-    outCompositorTiming->frameTime = preferredTimeline.frameTime;
-    outCompositorTiming->expectedPresentTime = preferredTimeline.expectedPresentTime;
-    outCompositorTiming->frameTimelineDeadline = preferredTimeline.frameTimelineDeadline;
+    // FIXME: expectedPresentLatency might reflect the previous frame's value because
+    //        the choreographer's callback can happen before (good) or after (bad) us.
+    //        This problem is mitigated by storing the latency instead of the deadline,
+    //        because it generally is constant frame to frame.
+    outCompositorTiming->expectedPresentLatency =
```
poweifeng: same fix is needed here.

phoenixxxx:
Merge conflicts to be addressed locally.

z3moon (approved):

PR #9596, opened 2026-01-09
split FeatureFlagManager out of FEngine

pixelflinger:
- Separate the Feature flags management from FEngine.
- Make FeatureFlagManager available to backends through DriverConfig
- add a way to override a feature flag value at runtime using environment variables or system properties (on Android).

e.g.:

```
env "feature.name=true" gltf_viewer
```
or 
```
adb shell setprop debug.feature.name true
```

### DIFF
--- a/libs/utils/src/FeatureFlagManager.cpp
+++ b/libs/utils/src/FeatureFlagManager.cpp
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2024 The Android Open Source Project
+ * Copyright (C) 2026 The Android Open Source Project
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.


libs/utils/include/private/utils/FeatureFlagManager.h
```diff
@@ -0,0 +1,111 @@
...
+        } material;
+    } features;
+
+    FeatureFlagManager();
+
+    Slice<const FeatureFlag> getFeatureFlags() const noexcept {
+        return { mFeatures.data(), mFeatures.size() };
+    }
+
+    bool setFeatureFlag(char const* UTILS_NONNULL name, bool value) noexcept;
```
poweifeng, line 108: suggest we make this protected so that only FEngine gets to call it. Same with `getFeatureFlagPtr`. (backend doesn't get to).

libs/utils/src/FeatureFlagManager.cpp
```diff
@@ -0,0 +1,165 @@
...
+
+bool FeatureFlagManager::setFeatureFlag(char const* name, bool const value) noexcept {
+    auto* const p = getFeatureFlagPtr(name);
+    if (p) {
+        *p = value;
+    }
+    return p != nullptr;
+}
+
+std::optional<bool> FeatureFlagManager::getFeatureFlag(char const* name) const noexcept {
```
poweifeng, line 150: for flags that can be updated after engine creation.  Is there a potential race condition here now that it's shared across threads?

libs/utils/include/private/utils/FeatureFlagManager.h
pixelflinger: good idea.

libs/utils/src/FeatureFlagManager.cpp
pixelflinger: I think so, but I decided I wouldn't care about it. I think the ones you can change at runtime, it's mostly for debugging, or for using in a tool. I don't think it can cause crashes.

libs/utils/src/FeatureFlagManager.cpp
```diff
@@ -0,0 +1,167 @@
...
+    auto const pos = std::find_if(mFeatures.begin(), mFeatures.end(),
+            [name](FeatureFlag const& entry) {
+                return name == entry.name;
+            });
+
+    return (pos != mFeatures.end() && (!pos->constant || allowConstant)) ?
+           const_cast<bool*>(pos->value) : nullptr;
+}
+
+} // namespace filament
```
z3moon, line 167: namespace utils

libs/utils/include/private/utils/FeatureFlagManager.h
```diff
@@ -0,0 +1,117 @@
+/*
+ * Copyright (C) 2024 The Android Open Source Project
```
poweifeng:
```suggestion
 * Copyright (C) 2026 The Android Open Source Project
```

libs/utils/src/FeatureFlagManager.cpp
```diff
@@ -0,0 +1,167 @@
+/*
+ * Copyright (C) 2024 The Android Open Source Project
```
poweifeng:
```suggestion
 * Copyright (C) 2026 The Android Open Source Project
```

PR #9598, opened 2026-01-10
Re-enable iOS filamat builds

bejado:


libs/filamat/CMakeLists.txt
```diff
@@ -1,4 +1,10 @@
 cmake_minimum_required(VERSION 3.19)
+
+if(IOS)
+    # filamat depends on glslang, which requires iOS 13.0+
+    set(CMAKE_OSX_DEPLOYMENT_TARGET "13.0")
```
z3moon: just curious, can we put this in only one place and have others use the same one?

bejado: Yes I found a way! Good suggestion

PR #9600, opened 2026-01-10
Add Light Cookies support for Spot and Point lights

bludosDevv:
-i hope this time the code is complete (I am beginner)

If it is not enough please add lightcookies like unity where you can add custom texture such as d.o.t.s for dots projector 
v4-460px-Dots-Projector-Phasmophobia-Step-1


bludosDevv:
Imma do it again wait

PR #9601, opened 2026-01-10
Add Light Cookies support for Spot and Point lights (I hope this is correct)

bludosDevv:
- I've remade it tried keeping everything in place have a review
- if it's still wrong please add light cookies because I wanna make a game with this dots projector shown below, let us have light cookies for spot light and point light
v4-460px-Dots-Projector-Phasmophobia-Step-1


pixelflinger:
thank you for trying but unfortunately you cannot use an extra sampler for filament, because we are limited to a total of 16 samplers, and they're all used (we reserve 9 for the application). this has been the main limiting factor for us to implement new features

I wish there was an easy solution but there is not.

one solution is to make these features available only at Feature Level 3 -- but there are some important plumbing missing at this time.

another solution would be to use texture atlases, but again, we need more plumbing.

Unfortunately we cannot accept this PR, as it will break some existing materials.

bludosDevv:
I will try fixing

PR #9604, opened 2026-01-12
Handle deferred destruction for asynchronous objects

z3moon:
When an asynchronous object is destroyed immediately after creation, the background creation process may still be active, holding a reference to the object. We now defer destruction in these cases until the creation process completes to ensure the reference remains valid.

Implementation details:
1. Add a mechanism to stop the ServiceThread to drain pending asynchronous tasks earlier. This ensures the mCreationComplete flag in the frontend object is set to true.
2. Later, the engine's terminate performs garbage collection for the deferred destruction. All of them must be successfully processed.

BUGS=[442921995]

### DIFF
--- a/filament/src/details/Engine.cpp
+++ b/filament/src/details/Engine.cpp
@@ -1199,6 +1199,8 @@ bool FEngine::terminateAndDestroy(const T* ptr, ResourceList<T>& list) {
             p->terminate(*this);
 
             if (p->isCreationComplete()) {
+                // If creation is complete, we free the frontend object immediately. Note that in
+                // regular (non-async) mode, the `isCreationComplete` method always return true.
                 mHeapAllocator.destroy(p);
             } else {
                 // We defer the destruction of the frontend object until the creation process


filament/src/details/Engine.cpp
```diff
@@ -1194,11 +1245,19 @@ bool FEngine::destroy(const FBufferObject* p) {
 
 UTILS_NOINLINE
 bool FEngine::destroy(const FVertexBuffer* p) {
+    if (p && !p->isCreationComplete()) {
+        terminateAndDeferAsyncObjectDestruction(p, mVertexBuffers);
+        return true;
```
pixelflinger: shouldn't we return `false` if `list.remove()` failed in `terminateAndDeferAsyncDestruction`?

filament/src/details/Engine.cpp
```diff
@@ -1185,6 +1206,36 @@ bool FEngine::terminateAndDestroyLocked(Lock& lock, const T* p, ResourceList<T>&
     return success;
 }
 
+template<typename T>
+UTILS_NOINLINE
+void FEngine::terminateAndDeferAsyncObjectDestruction(const T* p, ResourceList<T>& list) {
```
pixelflinger: instead of duplication most of the code here, wouldn't it make more sense to add a "bool async = false" parameter to `terminateAndDestroy`?  For most objects it would be `false` and no extra code would be added.

filament/src/details/Engine.cpp
```diff
@@ -1194,11 +1245,19 @@ bool FEngine::destroy(const FBufferObject* p) {
 
 UTILS_NOINLINE
 bool FEngine::destroy(const FVertexBuffer* p) {
+    if (p && !p->isCreationComplete()) {
```
pixelflinger: shouldn't this be `isAsyncMode()`?  Also I think the `p && !p->isCreationComplete` should be inside terminateAndDeferAsync... so that it's not duplicated 3 times.

filament/src/details/Engine.h
```diff
@@ -708,6 +717,8 @@ class FEngine : public Engine, public utils::FeatureFlagManager {
     // Creation parameters
     Config mConfig;
 
+    std::vector<std::function<bool()>> mDeferredAsyncObjectDestruction;
```
pixelflinger, line 712: just to be clear, in practice this shouldn't grow too much right? (because we should be completed most of the time).

filament/src/details/Renderer.cpp
```diff
@@ -296,9 +296,14 @@ void FRenderer::skipFrame(uint64_t vsyncSteadyClockTimeNano) {
...
-    // Run the component managers' GC
-    auto& js = engine.getJobSystem();
-    js.runAndWait(jobs::createJob(js, nullptr, &FEngine::gc, &engine)); // gc all managers
+    // Run GC
+    JobSystem& js = engine.getJobSystem();
+    auto *rootJob = js.setRootJob(js.createJob());
+    js.run(jobs::createJob(js, nullptr, &FEngine::gcManagers, &engine)); // gc all component managers
+    if (engine.isAsynchronousModeEnabled()) {
+        js.run(jobs::createJob(js, nullptr, &FEngine::gcDeferredAsyncObjectDestruction, &engine));
+    }
```
pixelflinger: if you don't make the jobs children of the `rootJob`, the rootJob has basically no effect.

filament/src/details/Renderer.cpp
```diff
@@ -467,9 +472,14 @@ void FRenderer::endFrame() {
...
```
pixelflinger: jobs must be children of the rootJob

filament/src/details/Renderer.cpp
z3moon: If the parent parameter of `createJob` is null, it internally takes the parent value set by `setRootJob`. (FRenderer::renderInternal does the same thing) So it works fine.

But let me update it in an explicit way.

filament/src/details/Engine.cpp
z3moon: Yeah I consolidated them, which looks better.

filament/src/details/Engine.cpp
z3moon: done

filament/src/details/Engine.cpp
z3moon: done

filament/src/details/Engine.h
z3moon: Yeah, normally it wouldn't grow at all.

filament/src/details/Renderer.cpp
pixelflinger: oh yes, for this use case you wouldn't use setRootJob

filament/src/details/Engine.cpp
```diff
@@ -1176,9 +1183,40 @@ bool FEngine::terminateAndDestroy(const T* p, ResourceList<T>& list) {
...
+            // immediately as they are no longer referenced. However, we defer the destruction of
+            // the frontend object if it's still being loaded.
+            // Reason: The creation process is still active and holds a reference to the frontend
+            // object to set `mCreationComplete` to true (see FTexture::FTexture). Deleting it now
+            // may cause a crash, so we wait until creation completes.
+
+            // Terminate the backend resource immediately as they're unnecessary from this point.
+            p->terminate(*this);
+
+            if (p->isCreationComplete()) {
```
pixelflinger, line 1201: it might be "obvious", but I think it would be helpful to have a comment here that says that if the async mode is not enabled, `isCreationComplete()` is guaranteed to be `true`.

filament/src/details/Engine.cpp
```diff
@@ -1176,9 +1183,40 @@ bool FEngine::terminateAndDestroy(const T* p, ResourceList<T>& list) {
...
 
     if (ASSERT_PRECONDITION_NON_FATAL(success,
-            "Object %s at %p doesn't exist (double free?)", typeNameCStr, p)) {
-        const_cast<T*>(p)->terminate(*this);
-        mHeapAllocator.destroy(const_cast<T*>(p));
+            "Object %s at %p doesn't exist (double free?)", typeNameCStr, ptr)) {
+
+        T* p = const_cast<T*>(ptr);
+
+        if constexpr (HasIsCreationComplete<T>::value) {
```
pixelflinger, line 1189: nicely done!

PR #9605, opened 2026-01-13
backend: CallbackHandler dstor inlined in header

poweifeng:
Defining the destructor in a cpp will break downstream clients.

The previous move to cpp was part of #9508.

### DIFF
--- a/filament/backend/src/vulkan/VulkanHandles.cpp
+++ b/filament/backend/src/vulkan/VulkanHandles.cpp
@@ -265,24 +265,31 @@ void VulkanDescriptorSet::addNewSet(VkDescriptorSet vkSet, OnRecycle&& onRecycle
 
 PushConstantDescription::PushConstantDescription(backend::Program const& program) {
     mRangeCount = 0;
-    for (auto stage : { ShaderStage::VERTEX, ShaderStage::FRAGMENT, ShaderStage::COMPUTE }) {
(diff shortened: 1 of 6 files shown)


filament/backend/include/backend/CallbackHandler.h
```diff
@@ -17,6 +17,11 @@
 #ifndef TNT_FILAMENT_BACKEND_CALLBACKHANDLER_H
 #define TNT_FILAMENT_BACKEND_CALLBACKHANDLER_H
 
+#ifdef __clang__
+#pragma clang diagnostic push
```
pixelflinger, line 21: do you need this if you put back the dtor in the header? this is the way it was before.

poweifeng: Yes, because we now have an internal use of this API (see AsyncHelper.h and #9508 ). And we have both `-Wweak-vtables` and `-Werror`

z3moon: Sorry for this inconvenience. The change looks good. Could you add some comments on why this is needed?

z3moon (approved):

filament/backend/include/backend/CallbackHandler.h
poweifeng: done

PR #9606, opened 2026-01-13
Refactor implicit DriverApi dependency

pixelflinger:
Removed implicit dependency on FEngine::getDriverApi() in critical  rendering paths. This change makes the DriverApi dependency explicit by  passing it as a parameter.

Key changes:

- FRenderer: Now accepts DriverApi& in renderInternal and renderJob.  Caches feature levels and capability flags in the constructor.

- PostProcessManager: Caches feature flags and workaround flags in  init(). Updated PostProcessMaterial and configure* methods to accept  DriverApi&.

- FMaterial: Caches support flags (stereo, parallel shader compile) and default material pointer in the constructor. 
Updated precacheDepthVariants and prepareProgram to rely on these  cached values or passed parameters.

- ShadowMapManager: Caches workaround flags in the constructor.

- FEngine: Updated prepare() to accept DriverApi&.

- RenderPass: Updated to accept DriverApi& in constructor and  appendCommands.

- Removed implicit DriverApi calls from hot paths to reduce overhead  and improve safety.

One common pattern in this change is to cache getters in constructors so that we don't need the DriverApi dependency later during rendering. The intention here is to remove DriverApi& from as many places as possible so that we can later have multithreading for command recording.

This PR should not change any behavior.

### DIFF
--- a/filament/src/details/Material.h
+++ b/filament/src/details/Material.h
@@ -134,11 +134,11 @@ public:
     // prepareProgram creates the program for the material's given variant at the backend level.
     // Must be called outside of backend render pass.
     // Must be called before getProgram() below.
-    void prepareProgram(Variant const variant,
+    void prepareProgram(backend::DriverApi& driver, Variant const variant,
             backend::CompilerPriorityQueue const priorityQueue) const noexcept {
         // prepareProgram() is called for each RenderPrimitive in the scene, so it must be efficient.
         if (UTILS_UNLIKELY(!isCached(variant))) {
-            prepareProgramSlow(variant, priorityQueue);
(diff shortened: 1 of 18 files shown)


filament/src/details/Material.h
```diff
@@ -325,6 +325,11 @@ class FMaterial : public Material {
     bool mIsDefaultMaterial = false;
 
     bool mUseUboBatching = false;
+    bool mIsStereoSupported = false;
+    bool mIsParallelShaderCompileSupported = false;
+    bool mDepthPrecacheDisabled = false;
```
show50726, line 330: Would it help if we use bitfield for these bools like in FRenderer?

PR #9607, opened 2026-01-13
VK: Add support for VK_KHR_global_priority

rafadevai:
Add support for application that want to change the priority of GPU queues at a system level.

### DIFF
--- a/filament/backend/src/vulkan/VulkanDriver.cpp
+++ b/filament/backend/src/vulkan/VulkanDriver.cpp
@@ -851,7 +851,7 @@ void VulkanDriver::createProgramR(Handle<HwProgram> ph, Program&& program, utils
 
     // Base case - build the pipeline without any external samplers.
     mPipelineCache.asyncPrewarmCache(
-        *vprogram.get(),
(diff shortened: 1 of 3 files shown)


poweifeng (approved):

filament/backend/src/vulkan/VulkanContext.h
```diff
@@ -240,6 +244,7 @@ struct VulkanContext {
     bool mPipelineCreationFeedbackSupported = false;
     bool mProtectedMemorySupported = false;
     bool mVertexInputDynamicStateSupported = false;
+    bool mGlobalPrioritySupported = false;
```
anishmgoyal, line 247: nitpicky nit: let's try to keep these alphabetized?

pixelflinger: That's not "a thing" in filament's coding style.  However, we generally try to order fields by temporal locality and/or to avoid padding. Generally this means ordering by large object first (e.g. pointers).
In the end we don't have strong rules for field ordering.

PR #9616, opened 2026-01-19
Optimize thread variables in ServiceThread

z3moon:
Wraps thread-related member variables in preprocessor checks to ensure they are excluded when multithreading is not supported. Previously, these variables were declared unconditionally, leading to unnecessary memory usage for platforms that don't support multithreading.

### DIFF
--- a/filament/backend/src/opengl/OpenGLDriver.cpp
+++ b/filament/backend/src/opengl/OpenGLDriver.cpp
@@ -390,9 +390,9 @@ void OpenGLDriver::terminate() {
     if (getJobWorker()) {
         getJobWorker()->terminate();
     }
-    if constexpr (UTILS_HAS_THREADING) {
-        stopServiceThread();
-    }
+#if UTILS_HAS_THREADING
+    stopServiceThread();
+#endif
 
     mContext.terminate();
     mPlatform.terminate();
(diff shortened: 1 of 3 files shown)


filament/backend/src/opengl/OpenGLDriver.cpp
```diff
@@ -390,9 +390,9 @@ void OpenGLDriver::terminate() {
     if (getJobWorker()) {
         getJobWorker()->terminate();
     }
-    if constexpr (UTILS_HAS_THREADING) {
-        stopServiceThread();
-    }
+#if UTILS_HAS_THREADING
```
pixelflinger, line 393: I don't like this because that means that the "non threaded" code can fail compilation without us noticing. I would like to keep using `if constexpr` as  much as possible.

z3moon: Not sure if I'm following you, but non-threaded code cannot fail compilation (as you see web-build passed) because we remove all relevant variables & functions, which are actually unnecessary. If we want to remove those unnecessary (thread-relevant) variables for non-threaded platforms, we need to use `#if` as constexpr is evaluated after all preprocessors are handled.

pixelflinger: I know but that means that when I compile locally, I cannot know for sure that I didn't break the build. I have to rely on the CI, which takes time and if a failure happens, it gets really annoying to reproduce locally.

How much of a problem are those variable anyways? I don't think they're using too much space.

z3moon: This is just a cleanup, not producing any problem. Let me close this.

PR #9617, opened 2026-01-20
Add helloasync sample

z3moon:
Introduces a new sample application demonstrating asynchronous resource management. It shows how to asynchronously create and update Filament resources (Textures, VertexBuffers, IndexBuffers) and provides examples of managing chains of asynchronous operations effectively.

BUGS=[442921995]

### DIFF
--- a/libs/fgviewer/web/app.js
+++ b/libs/fgviewer/web/app.js
@@ -15,8 +15,6 @@
 */
 
 import {LitElement, html, css, unsafeCSS, nothing} from "https://unpkg.com/lit@2.8.0?module";
-import { graphviz } from "https://cdn.skypack.dev/d3-graphviz@5.1.0";
-import * as d3 from "https://cdn.skypack.dev/d3@7";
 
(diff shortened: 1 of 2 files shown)


samples/helloasync.cpp
```diff
@@ -0,0 +1,582 @@
...
+        usage.replace(pos, apiUsage.length(), samples::getBackendAPIArgumentsUsage());
+    }
+    std::cout << usage;
+}
+
+static int handleCommandLineArguments(int argc, char* argv[], Config& config) {
+    static constexpr const char* OPTSTR = "ha:";
+    static const struct option OPTIONS[] = {
+        { "help", no_argument, nullptr, 'h' },
+        { "api", required_argument, nullptr, 'a' },
```
pixelflinger, line 99: should be in printUsage() no?

z3moon: I brought the base code from a sample, and I found all samples don't have this either. Let me handle this as a separate pr.

PR #9619, opened 2026-01-21
Fix buffer overflow in RenderPass command batching

pixelflinger:
The command batching logic in RenderPass::Executor::execute calculates a conservative `maxCommandSizeInBytes` to determine how many commands can safely fit in the driver's circular buffer.

However, this calculation failed to account for the dynamic memory allocation incurred by the Per-Material descriptor set binding (via `MaterialInstance::use`). This binding allocates a `DescriptorSetOffsetArray` in the command stream, which was unbudgeted.

In scenarios with frequent material and scissor changes, the accumulated deficit from these unbudgeted allocations could exceed the buffer  capacity, leading to a crash.

This change:
1. Adds the missing accounting for the Per-Material dynamic allocation  to `maxCommandSizeInBytes`.
2. Adds a 20% safety margin to the calculated size to robustly handle alignment padding and worst-case scenarios.

FIXES=[474264976]

### DIFF
--- a/filament/src/RenderPass.cpp
+++ b/filament/src/RenderPass.cpp
@@ -940,31 +940,32 @@ void RenderPass::Executor::execute(FEngine const& engine, DriverApi& driver,
 
         // Maximum space occupied in the CircularBuffer by a single `Command`. This must be
         // reevaluated when the inner loop below adds DriverApi commands or when we change the
-        // CommandStream protocol. Currently, the maximum is 248 bytes.
+        // CommandStream protocol. Currently, the maximum is 272 bytes.
         // The batch size is calculated by adding the size of all commands that can possibly be
         // emitted per draw call:
         constexpr size_t maxCommandSizeInBytes =
                 sizeof(COMMAND_TYPE(scissor)) +
                 sizeof(COMMAND_TYPE(bindDescriptorSet)) +
-                sizeof(COMMAND_TYPE(bindDescriptorSet)) +
+                sizeof(COMMAND_TYPE(bindDescriptorSet)) + backend::CustomCommand::align(sizeof(NoopCommand) + 4) +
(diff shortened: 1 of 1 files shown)


filament/src/RenderPass.cpp
```diff
@@ -940,31 +940,32 @@ void RenderPass::Executor::execute(FEngine const& engine, DriverApi& driver,
...
         constexpr size_t maxCommandSizeInBytes =
                 sizeof(COMMAND_TYPE(scissor)) +
                 sizeof(COMMAND_TYPE(bindDescriptorSet)) +
-                sizeof(COMMAND_TYPE(bindDescriptorSet)) +
+                sizeof(COMMAND_TYPE(bindDescriptorSet)) + backend::CustomCommand::align(sizeof(NoopCommand) + 4) +
                 sizeof(COMMAND_TYPE(bindPipeline)) +
                 sizeof(COMMAND_TYPE(bindRenderPrimitive)) +
                 sizeof(COMMAND_TYPE(bindDescriptorSet)) + backend::CustomCommand::align(sizeof(NoopCommand) + 8) +
                 sizeof(COMMAND_TYPE(setPushConstant)) +
                 sizeof(COMMAND_TYPE(draw2));
```
poweifeng, line 955: since this number is pretty crucial.  Can you maybe do a 
```
static_assert(maxCommandSizeInBytes == 272);
```

pixelflinger: > since this number is pretty crucial. Can you maybe do a
> 
> ```
> static_assert(maxCommandSizeInBytes == 272);
> ```

Actually it's not 272 that's crucial. i.e. if the size of a command changed, that number would change to something else and that would be perfectly fine.

The bigger problem is that if we add commands or add allocations without updating that number, we could get into trouble, and unfortunately, there is no way that I can think of to do this at compile time or check at compile time.

PR #9621, opened 2026-01-21
don't bucket (z-sort) opaque geometry

pixelflinger:
All modern mobile GPUs do this sorting
automatically, similarly to using a v-buffer.
Sorting at the application level is 
counter-productive because prioritize depth
over state, which is much more important.

Transparents are still sorted for correctness.


poweifeng:
> Opaques are still sorted for correctness.

Do you mean transparents?

poweifeng (approved):

pixelflinger:
after doing more research, depth sorting is still beneficial on many mobile GPUs. abandoning the PR.

PR #9633, opened 2026-01-26
VK: Only bypass the staging buffer when is not in use

rafadevai:
In the cases where a frame takes more than the vsync time, we start seeing some rendering artifacts because we write to a buffer that is currently in flight.

Not 100% sure why the current approach fails but for now, the staging buffer is bypassed only when the buffer is not in use or referenced by any other resource.

FIXED=[445455050]
FIXED=[477305399]


pixelflinger:
there should be a bug for tracking the problem

rafadevai:
> there should be a bug for tracking the problem

I linked the 2 bugs, is there an specific way I should do it?

rafadevai:
> Yeah, I think this is the right approach, but having getCount() == 1 to guard it seems brittle.

Its not brittle but kinda conservative, I'm pretty sure we can come up with a way that its being activated more often

poweifeng:
> > Yeah, I think this is the right approach, but having getCount() == 1 to guard it seems brittle.
> 
> Its not brittle but kinda conservative, I'm pretty sure we can come up with a way that its being activated more often

brittle in the sense that if you somehow keep the reference to the buffer elsewhere (but not in a command buffer), then the count would be greater than 1.

PR #9634, opened 2026-01-26
Swap logic of how the EGL display is initialized.

haroonq:
1. Iterate over system displays using eglQueryDevicesEXT and try to initialize them.

2. If no display initialized, then try to initialize the DEFAULT display.

BUGS=[478925865]
FIXES=[478925865]


pixelflinger:
looks good but why the change? is there a bug associated?

filament/backend/src/opengl/platforms/PlatformEGL.cpp
```diff
@@ -120,25 +120,30 @@ bool PlatformEGL::isOpenGL() const noexcept {
...
-    mEGLDisplay = eglGetDisplay(EGL_DEFAULT_DISPLAY);
-    assert_invariant(mEGLDisplay != EGL_NO_DISPLAY);
+    static constexpr int kMaxNumEGLDevices = 32;
 
     EGLint major, minor;
-    EGLBoolean initialized = eglInitialize(mEGLDisplay, &major, &minor);
+    EGLBoolean initialized = false;
+
+    EGLDeviceEXT eglDevices[kMaxNumEGLDevices];
+    EGLint numDevices;
```
z3moon: this is only used within the `if (eglQueryDevicesEXT != nullptr) {` block below. You can move this within the block.

filament/backend/src/opengl/platforms/PlatformEGL.cpp
```diff
@@ -120,25 +120,30 @@ bool PlatformEGL::isOpenGL() const noexcept {
...
+    PFNEGLQUERYDEVICESEXTPROC const eglQueryDevicesEXT =
+        PFNEGLQUERYDEVICESEXTPROC(eglGetProcAddress("eglQueryDevicesEXT"));
+    PFNEGLGETPLATFORMDISPLAYEXTPROC const getPlatformDisplay =
+        PFNEGLGETPLATFORMDISPLAYEXTPROC(eglGetProcAddress("eglGetPlatformDisplay"));
+
+    if (eglQueryDevicesEXT != nullptr) {
+        eglQueryDevicesEXT(kMaxNumEGLDevices, eglDevices, &numDevices);
```
z3moon: You need to check the return value of this call. Otherwise `numDevices` may have an invalid value, thereby producing undefined behavior for the following loop.

Also I would initialize the variable just in case.
```
EGLint numDevices = 0;
```

filament/backend/src/opengl/platforms/PlatformEGL.cpp
```diff
@@ -120,25 +120,30 @@ bool PlatformEGL::isOpenGL() const noexcept {
...
+    EGLBoolean initialized = false;
+
+    EGLDeviceEXT eglDevices[kMaxNumEGLDevices];
+    EGLint numDevices;
+    PFNEGLQUERYDEVICESEXTPROC const eglQueryDevicesEXT =
+        PFNEGLQUERYDEVICESEXTPROC(eglGetProcAddress("eglQueryDevicesEXT"));
+    PFNEGLGETPLATFORMDISPLAYEXTPROC const getPlatformDisplay =
+        PFNEGLGETPLATFORMDISPLAYEXTPROC(eglGetProcAddress("eglGetPlatformDisplay"));
+
+    if (eglQueryDevicesEXT != nullptr) {
```
z3moon: please check `getPlatformDisplay != nullptr` as well

haroonq:
PTAL

PR #9635, opened 2026-01-26
AtlasAllocator: Implement free() and upgrade to Buddy Allocator

pixelflinger:
This change upgrades AtlasAllocator from a simple linear allocator  (reset-only) to a full Buddy Allocator capable of both allocation and  deallocation. This enables persistent shadow map management where  individual shadow maps can be updated or resized without rebuilding the  entire atlas every frame.

Key changes:
- Implemented AtlasAllocator::free():
	- Uses the existing QuadTree structure to track allocations.
	- Automatically coalesces empty sibling nodes back into larger  parent blocks (standard Buddy Allocator behavior).
	- Optimized to be allocation-free and recursion-free, using direct  array indexing.

- Added comprehensive unit tests

Updated documentation.

The QuadTreeArray implementation remains unchanged but is now fully  utilized for bidirectional tree traversal (down for alloc, up for free).

This change will later allow us to cache shadow-maps. The AtlasAlocator is currently disabled by a feature flag.

### DIFF
--- a/filament/src/AtlasAllocator.cpp
+++ b/filament/src/AtlasAllocator.cpp
@@ -72,8 +72,56 @@ AtlasAllocator::Allocation AtlasAllocator::allocate(size_t const textureSize) no
         result.viewport.width  = dimension;
         result.viewport.height = dimension;
         result.layer = loc.code >> (2 * layer);
+        result.code = loc.code;
+        result.level = loc.l;
     }
     return result;
+}
+
+void AtlasAllocator::free(Allocation const& allocation) noexcept {
+    if (UTILS_UNLIKELY(!allocation.isValid())) {
+        return;
+    }
+
+    // 1. Mark the node as unallocated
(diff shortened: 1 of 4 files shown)


filament/src/AtlasAllocator.cpp
```diff
@@ -72,10 +72,58 @@ AtlasAllocator::Allocation AtlasAllocator::allocate(size_t const textureSize) no
...
+        return;
+    }
+
+    // 1. Mark the node as unallocated
+    const size_t i = QuadTreeUtils::index(allocation.level, allocation.code);
+    Node& node = mQuadTree[i];
+
+    // In debug builds, ensure we are freeing something that was actually allocated
+    assert_invariant(node.isAllocated());
+    assert_invariant(!node.hasChildren());
```
z3moon, line 92: To be clear, this means they're filament bugs, right? (For example, users cannot double free an allocation)

pixelflinger: correct.

PR #9645, opened 2026-01-27
wip: fog for opaques in applied as a post-process effect

pixelflinger:
Instead of computing the fog "inline", in the forward pass, we can instead compute it as post-process pass that is applied with a simple fullscreen quad blending.  On tilers, the operation entirely stays in the tile, on desktop GPU it is a blending operation.

This works only for opaque materials.

The benefit is that fog will become immune to overdraw, and the forward pass shader will be simplified, hopefully leading to less register  pressure. Overall performance should be improved.

Another benefit is that it will allow us to free the "fog" texture slot from all opaque materials.

Transparent materials are unchanged.

This feature is currently DISABLED, and still work in progress; but it should be mostly functional.

To test it:

```
env material.enable_fog_as_postprocess=true ./out/samples/gltf_viewer
```

This change refactor the fog code, but shouldn't have any impact on the current behavior.

### DIFF
--- a/filament/src/PostProcessManager.cpp
+++ b/filament/src/PostProcessManager.cpp
@@ -30,6 +30,7 @@
 #include "materials/colorGrading/colorGrading.h"
 #include "materials/dof/dof.h"
 #include "materials/flare/flare.h"
+#include "materials/fog/fog.h"
 #include "materials/fsr/fsr.h"
 #include "materials/sgsr/sgsr.h"
 #include "materials/ssao/ssao.h"
@@ -65,6 +66,7 @@
 
 #include <private/filament/EngineEnums.h>
 #include <private/filament/UibStructs.h>
+#include <private/filament/Variant.h>
 
 #include <backend/DriverEnums.h>
 #include <backend/DriverApiForward.h>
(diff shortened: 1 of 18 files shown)


filament/src/details/Renderer.cpp
```diff
@@ -1167,7 +1170,7 @@ void FRenderer::renderJob(DriverApi& driver, RootArenaScope& rootArenaScope, FVi
 
     // a non-drawing pass to prepare everything that need to be before the color passes execute
     fg.addPass<FrameGraph::Empty>("Prepare Color Passes",
-            [](FrameGraph::Builder& builder) {
+            [&](FrameGraph::Builder& builder) {
```
poweifeng: why this change?

filament/src/PostProcessManager.cpp
```diff
@@ -392,6 +393,9 @@ void PostProcessManager::init() noexcept {
         for (auto const& info: getSsaoMaterialList()) {
             registerPostProcessMaterial(info.name, info);
         }
+        for (auto const& info: getFogMaterialList()) {
+            registerPostProcessMaterial(info.name, info);
+        }
```
show50726: I thought fog should be supported under feature level 0? Should we move this outside of the feature level 1 check?

filament/src/details/Renderer.cpp
pixelflinger: oops, oversight.

shaders/src/surface_main.fs
```diff
@@ -66,12 +66,34 @@ void main() {
     highp vec3 view = getWorldPosition() - getWorldCameraPosition();
     view = frameUniforms.fogFromWorldMatrix * view;
 
-#if defined (FILAMENT_LINEAR_FOG)
-    fragColor = fogLinear(fragColor, view);
+#if MATERIAL_FEATURE_LEVEL > 0
+#   if defined (FILAMENT_LINEAR_FOG)
+    vec4 fogColor = fogLinear(view, sampler0_fog);
```
romainguy, line 71: Shouldn't you skip fog completely in opaque materials?

filament/src/PostProcessManager.cpp
pixelflinger: good catch!

shaders/src/surface_main.fs
pixelflinger: yes, it's disabled on the C++ side. We can't remove the code right now because we need to introduce this feature via a feature flag.

PR #9653, opened 2026-01-28
Remove instanced stereoscopic type

trie94:
an option to remove the instanced stereoscopic type was brought up in the conversation whether we can have multiview as the default value for the stereo rendering.

### DIFF
--- a/filament/backend/include/backend/Platform.h
+++ b/filament/backend/include/backend/Platform.h
@@ -206,10 +206,6 @@ public:
          */
         NONE,
         /**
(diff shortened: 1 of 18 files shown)


pixelflinger:
I would like the opinion of the whole team on that one.

z3moon:
This is what we need to do eventually. This work requires thorough tests. At least, hellostereo and gltf_viewer samples should work without any issue.

trie94:
closing this in favor of https://github.com/google/filament/pull/9682

PR #9662, opened 2026-01-29
fgviewer: add websocket server

poweifeng:
A preliminary commit to add a websocket server to the DebugServer. This will enable us to transfer large data (like images) across to the frontend.

This is part of the work to enable viewing intermediate render buffers in fgviewer.

### DIFF
--- a/ios/samples/gltf-viewer/gltf-viewer.xcodeproj/project.pbxproj
+++ b/ios/samples/gltf-viewer/gltf-viewer.xcodeproj/project.pbxproj
@@ -437,6 +437,7 @@
 					"-lviewer",
 					"-lcivetweb",
 					"-luberzlib",
+					"-limageio-lite",
 				);
 				PRODUCT_BUNDLE_IDENTIFIER = "${SAMPLE_CODE_DISAMBIGUATOR}.google.filament.gltf-viewer";
 				SAMPLE_CODE_DISAMBIGUATOR = "${DEVELOPMENT_TEAM}";
@@ -494,6 +495,7 @@
 					"-lviewer",
 					"-lcivetweb",
 					"-luberzlib",
+					"-limageio-lite",
 				);
 				PRODUCT_BUNDLE_IDENTIFIER = "${SAMPLE_CODE_DISAMBIGUATOR}.google.filament.gltf-viewer";
 				SAMPLE_CODE_DISAMBIGUATOR = "${DEVELOPMENT_TEAM}";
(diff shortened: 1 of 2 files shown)


libs/fgviewer/src/WebSocketHandler.cpp
```diff
@@ -0,0 +1,59 @@
...
+
+namespace filament::fgviewer {
+
+using namespace utils;
+
+WebSocketHandler::WebSocketHandler(DebugServer* server)
+        : mServer(server) {}
+
+bool WebSocketHandler::handleConnection(CivetServer* server, const struct mg_connection* conn) {
+    slog.i << "[fgviewer] WebSocket connected." << io::endl;
```
show50726: nit: Can we use `LOG(INFO)` here?

(And all other `slog.i` usage)

libs/fgviewer/src/WebSocketHandler.h
```diff
@@ -0,0 +1,49 @@
...
+    bool handleData(CivetServer* server, struct mg_connection* conn, int bits, char* data,
+            size_t data_len) override;
+    void handleClose(CivetServer* server, const struct mg_connection* conn) override;
+
+    void broadcast(const char* data, size_t len);
+
+private:
+    DebugServer* mServer;
+    utils::Mutex mMutex;
+    std::set<struct mg_connection*> mConnections;
```
show50726: Could it be an `unordered_set`?

libs/fgviewer/src/WebSocketHandler.h
```diff
@@ -0,0 +1,49 @@
...
+    bool handleConnection(CivetServer* server, const struct mg_connection* conn) override;
+    void handleReadyState(CivetServer* server, struct mg_connection* conn) override;
+    bool handleData(CivetServer* server, struct mg_connection* conn, int bits, char* data,
+            size_t data_len) override;
+    void handleClose(CivetServer* server, const struct mg_connection* conn) override;
+
+    void broadcast(const char* data, size_t len);
+
+private:
+    DebugServer* mServer;
```
show50726: nit: Could we make the pointer const? e.g. `DebugServer const* mServer;`

libs/fgviewer/src/WebSocketHandler.h
poweifeng: done

libs/fgviewer/src/WebSocketHandler.h
poweifeng: done

PR #9663, opened 2026-01-29
engine: encapsulate material cache handling

elizagamedev:
We will soon allow `MaterialInstance` to override the value of spec constants. To avoid code duplication, we introduce a new class `MaterialPrograms` which handles the chunky bits of managing the program cache and values of the spec constants.

### DIFF
--- a/filament/src/LocalProgramCache.cpp
+++ b/filament/src/LocalProgramCache.cpp
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-#include "MaterialPrograms.h"
+#include "LocalProgramCache.h"
 #include "MaterialParser.h"
 
 #include "details/Engine.h"
@@ -27,7 +27,7 @@ namespace filament {
 using namespace backend;
 using namespace utils;
 
-MaterialPrograms::MaterialPrograms(MaterialPrograms const& other)
+LocalProgramCache::LocalProgramCache(LocalProgramCache const& other)
         : mMaterial(other.mMaterial),
           mCachedPrograms(other.mCachedPrograms.size()),
           mSpecializationConstants((other.mMaterial != nullptr)
@@ -37,7 +37,7 @@ MaterialPrograms::MaterialPrograms(MaterialPrograms const& other)
                                                      .acquire(other.mSpecializationConstants)
                                            : SpecializationConstants()) {}
 
-MaterialPrograms& MaterialPrograms::operator=(MaterialPrograms const& other) {
+LocalProgramCache& LocalProgramCache::operator=(LocalProgramCache const& other) {
     assert_invariant(mMaterial == nullptr);
     assert_invariant(mCachedPrograms.empty());
     assert_invariant(mSpecializationConstants.empty());
@@ -54,7 +54,7 @@ MaterialPrograms& MaterialPrograms::operator=(MaterialPrograms const& other) {
     return *this;
 }
 
-void MaterialPrograms::initializeForMaterial(FEngine& engine, FMaterial const& material,
+void LocalProgramCache::initializeForMaterial(FEngine& engine, FMaterial const& material,
         utils::FixedCapacityVector<backend::Program::SpecializationConstant>
                 specializationConstants) {
     assert_invariant(mMaterial == nullptr);
@@ -85,13 +85,13 @@ void MaterialPrograms::initializeForMaterial(FEngine& engine, FMaterial const& m
             material.getMaterialParser(), mSpecializationConstants, material.isDefaultMaterial());
 }
 
-void MaterialPrograms::initializeForMaterialInstance(FEngine& engine, FMaterial const& material) {
+void LocalProgramCache::initializeForMaterialInstance(FEngine& engine, FMaterial const& material) {
     assert_invariant(mMaterial == nullptr);
     assert_invariant(mCachedPrograms.empty());
     assert_invariant(mSpecializationConstants.empty());
 
     mMaterial = &material;
-    MaterialPrograms const& programs = material.getPrograms();
+    LocalProgramCache const& programs = material.getPrograms();
 
     mSpecializationConstants =
             engine.getMaterialCache().getSpecializationConstantsInternPool().acquire(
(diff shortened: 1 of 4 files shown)


filament/src/details/Material.cpp
```diff
@@ -163,37 +163,15 @@ FMaterial::FMaterial(FEngine& engine, const Builder& builder, MaterialDefinition
           mUseUboBatching(shouldEnableBatching(engine, builder->mUboBatchingMode,
                   definition.materialDomain)),
           mEngine(engine),
-          mMaterialId(engine.getMaterialId()) {
+          mMaterialId(engine.getMaterialId()),
+          mPrograms(engine, *this, processSpecializationConstants(builder)) {
```
pixelflinger: this is slightly dangerous because when initializing `mPrograms`, the `FMaterial` passed as parameter is not fully constructed. we're doing that in other places, but that means we need a big comment in `MaterialProgram` ctor that it can only store the FMaterial and not do anything with it.

filament/src/LocalProgramCache.cpp
```diff
@@ -0,0 +1,149 @@
...
+        case filament::MaterialDomain::POST_PROCESS:
+            cachedProgramsSize = 1 << POST_PROCESS_VARIANT_BITS;
+            break;
+        case filament::MaterialDomain::COMPUTE:
+            cachedProgramsSize = 1;
+            break;
+    }
+    mCachedPrograms = FixedCapacityVector<backend::Handle<backend::HwProgram>>(cachedProgramsSize);
+
+    material.getDefinition().acquirePrograms(engine, mCachedPrograms.as_slice(),
```
pixelflinger, line 84: this is very dangerous because `material` is not fully constructed when this is called.

in other places we use the `init() / terminate()` pattern for this.

pixelflinger (changes requested):
I'm not in love calling a ctor from a ctor passing "this". (i.e. FMaterial calling MaterialProgram(..., this)). I've been bit by this before.

filament/src/details/Material.cpp
elizagamedev: I had the same thought while working on the child PR. I backported and pushed the changes I made there.

filament/src/LocalProgramCache.cpp
```diff
@@ -0,0 +1,182 @@
...
+
+namespace filament {
+
+using namespace backend;
+using namespace utils;
+
+void MaterialPrograms::initializeForMaterial(FEngine& engine, FMaterial const& material,
+        utils::FixedCapacityVector<backend::Program::SpecializationConstant>
+                specializationConstants) {
+    mMaterial = &material;
```
poweifeng, line 64: nit: add some assert_invariants like mSpecialzationConstants.size() == 0

to sanity check that this is not called twice somehow.  Same for `initializeForMaterialInstance`

filament/src/MaterialPrograms.cpp
```diff
@@ -0,0 +1,182 @@
...
+        }
+        return mCachedPrograms[variant.key] =
+                defaultPrograms.prepareProgram(driver, variant, priorityQueue);
+    }
+    return mCachedPrograms[variant.key] = mMaterial->getDefinition().prepareProgram(engine, driver,
+                   mMaterial->getMaterialParser(), getProgramSpecialization(variant), priorityQueue);
+}
+
+ProgramSpecialization MaterialPrograms::getProgramSpecialization(Variant variant) const noexcept {
+    FILAMENT_CHECK_PRECONDITION(mMaterial != nullptr);
```
poweifeng: can `mMaterial == nullptr` be due to user error?  If not then assert_invariant is better.

filament/src/MaterialPrograms.h
```diff
@@ -0,0 +1,128 @@
...
+#include <backend/DriverApiForward.h>
+#include <backend/Program.h>
+
+namespace filament {
+
+class FMaterial;
+
+// L0 cache for material programs. Manages recompiling them on-demand; owned by Material and
+// MaterialInstance.
+class MaterialPrograms {
```
poweifeng: nit: would call it `MaterialProgramCache` instead to distinguish it from the globalProgramCache.

filament/src/MaterialPrograms.cpp
```diff
@@ -0,0 +1,182 @@
...
+            cachedProgramsSize = 1;
+            break;
+    }
+    mCachedPrograms = FixedCapacityVector<Handle<HwProgram>>(cachedProgramsSize);
+
+    material.getDefinition().acquirePrograms(engine, mCachedPrograms.as_slice(),
+            material.getMaterialParser(), mSpecializationConstants, material.isDefaultMaterial());
+}
+
+void MaterialPrograms::initializeForMaterialInstance(FEngine& engine, FMaterial const& material) {
```
poweifeng: I feel that we really just need one init function.  And the call site can make that distinction between material vs material instance and pass in the appropriate params.

filament/src/MaterialPrograms.cpp
```diff
@@ -0,0 +1,182 @@
...
+
+    return ProgramSpecialization {
+        .materialCrc32 = mMaterial->getMaterialParser().getCrc32(),
+        .variant = variant,
+        .specializationConstants = mSpecializationConstants,
+    };
+}
+
+void MaterialPrograms::terminate(FEngine& engine) {
+    FILAMENT_CHECK_PRECONDITION(mMaterial != nullptr);
```
poweifeng: same here.

filament/src/MaterialPrograms.h
```diff
@@ -0,0 +1,128 @@
...
+    }
+
+    // Set constant by name. Not applied until flushConstants() is called.
+    template<typename T, typename = is_supported_constant_parameter_t<T>>
+    void setConstant(std::string_view name, T value) noexcept {
+        setConstantImpl(name, value);
+    }
+
+    // Apply any pending specialization constants. Invalidates programs as necessary.
+    void flushConstants();
```
poweifeng: is there a case where you don't want to flush right after setting a list of constants?  If not, then maybe we could just have setConstant() take an initializer_list and just have one call instead of two.

filament/src/MaterialPrograms.cpp
elizagamedev: I feel like the two different methods/names make it really clear why you would call one method vs the other. For example, using the same method obscures the meaning of passing in a `FixedCapacityVector` vs relying on what `Material` owns.

filament/src/MaterialPrograms.h
elizagamedev: That's an interesting idea. This is eventually going across the JNI boundary, so I'll have to investigate how to approach it there. But it would simplify the internal complexity of the class a bit.

elizagamedev: (It's going across the JNI boundary for `MaterialInstance`, not for `Material`.)

PR #9664, opened 2026-01-29
External sampler bind index bug

phoenixxxx:
On some video use cases when the same vkSampler is retreived from the cache, we still need to ensure it is bound to the same index, otherwise VK considers it as two separate layouts for immutable samplers (bind index and sampler object need to match). As it stands we fail to differenciate when the same sampler is bound to two different bind points. We needs two separate layouts.

### DIFF
--- a/filament/backend/src/vulkan/VulkanDescriptorSetLayoutCache.cpp
+++ b/filament/backend/src/vulkan/VulkanDescriptorSetLayoutCache.cpp
@@ -62,7 +62,7 @@ uint32_t appendBindings(VkDescriptorSetLayoutBinding* toBind, VkDescriptorType t
 
 uint32_t appendSamplerBindings(VkDescriptorSetLayoutBinding* toBind,
         fvkutils::SamplerBitmask const& mask, fvkutils::SamplerBitmask const& external,
-        utils::FixedCapacityVector<std::pair<uint32_t, VkSampler>> const& immutableSamplers) {
+        utils::FixedCapacityVector<std::pair<uint64_t, VkSampler>> const& immutableSamplers) {
     using Bitmask = fvkutils::SamplerBitmask;
     uint32_t count = 0;
     Bitmask alreadySeen;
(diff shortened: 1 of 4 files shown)


filament/backend/src/vulkan/VulkanDescriptorSetLayoutCache.cpp
```diff
@@ -92,20 +92,19 @@ uint32_t appendSamplerBindings(VkDescriptorSetLayoutBinding* toBind,
...
                                               : nullptr,
             };
         }
     });
     return count;
 }
 
-uint64_t computeImmutableSamplerHash(utils::FixedCapacityVector<VkSampler> const& samplers) {
+uint64_t computeImmutableSamplerHash(
+        utils::FixedCapacityVector<std::pair<uint8_t, VkSampler>> const& samplers) {
```
poweifeng: there's a potential packing issue here.  To be safe, you could probably just make this `std::pair<uint64_t, VkSampler>`

filament/backend/src/vulkan/VulkanDescriptorSetLayoutCache.cpp
```diff
@@ -92,20 +92,19 @@ uint32_t appendSamplerBindings(VkDescriptorSetLayoutBinding* toBind,
...
     size_t const size = samplers.size();
     if (size == 0) {
         return 0;
-    } else if (size == 1) {
-        return (uint64_t) samplers[0];
     }
     return utils::hash::murmur3((uint32_t*) samplers.data(), samplers.size() * 2, 0);
```
poweifeng: `samplers.size() * 2` is the wrong size.  This should be the number of `uint32_t` elements.

filament/backend/src/vulkan/VulkanDescriptorSetLayoutCache.cpp
```diff
@@ -62,7 +62,7 @@ uint32_t appendBindings(VkDescriptorSetLayoutBinding* toBind, VkDescriptorType t
 
 uint32_t appendSamplerBindings(VkDescriptorSetLayoutBinding* toBind,
         fvkutils::SamplerBitmask const& mask, fvkutils::SamplerBitmask const& external,
-        utils::FixedCapacityVector<std::pair<uint8_t, VkSampler>> const& immutableSamplers) {
+        utils::FixedCapacityVector<std::pair<uint32_t, VkSampler>> const& immutableSamplers) {
```
poweifeng: I really think you want `uint64_t` here to ensure 8-byte alignment on 64-bit systems.

phoenixxxx: Ah good call. I had already made the change on the gLinux side the version I had in the first PR was outdated. Ok I will switch to 64

poweifeng (approved):

PR #9667, opened 2026-01-30
ios: fix build

poweifeng:
Need to include the new imageio-lite library in the gltf_viewer
sample.


poweifeng:
merge to fix ios build.

PR #9668, opened 2026-01-30
diffimg: binary tool for comparing images

poweifeng:
This tool uses existing libraries: image, imageio, imageio-lite, imagediff to perform difference comparison for on-disk images.

We refactor renderdiff to use this tool instead of using python dependencies.


tools/diffimg/main.cpp
```diff
@@ -0,0 +1,212 @@
...
+
+    auto decodeWithImageIOLite = [](std::istream& stream, const std::string& name) {
+        return imageio_lite::ImageDecoder::decode(stream, CString(name.c_str()),
+                imageio_lite::ImageDecoder::ColorSpace::LINEAR);
+    };
+
+    std::string ext = path.getExtension();
+    std::transform(ext.begin(), ext.end(), ext.begin(), ::tolower);
+
+    bool preferLite = (ext == "tif" || ext == "tiff");
```
bejado, line 61: What's the benefit of using imageio lite?

tools/diffimg/main.cpp
```diff
@@ -0,0 +1,212 @@
...
+        maskImg = loadImage(maskPath);
+        if (!maskImg.isValid()) return 1;
+    }
+
+    imagediff::ImageDiffConfig config;
+    if (!configPath.isEmpty()) {
+        std::string jsonContent = readFile(configPath);
+        if (jsonContent.empty()) {
+             std::cerr << "Error: Could not read config file: " << configPath << std::endl;
+             return 1;
```
bejado: nit:
```suggestion
            std::cerr << "Error: Could not read config file: " << configPath << std::endl;
            return 1;
```

bejado (approved):

tools/diffimg/main.cpp
poweifeng: Yeah, this can read a bit strange.  So I tried to list reasons in the initial commit: 
https://github.com/google/filament/commit/b40530ad3cc1b2ddcb004002ff1bea7f1994855d

PR #9670, opened 2026-01-31
Add multiview support to clearDepth.mat

z3moon:
View::setChannelDepthClearEnabled utilizes `clearDepth.mat` internally, which is not compatible with multiview since clearDepth is post-process.

This update converts clearDepth.mat into a surface shader, allowing it to hold the STE variant. This change enables the shader to function correctly in multiview.

BUGS=[470198472]


filament/src/details/Renderer.cpp
```diff
@@ -1186,7 +1186,13 @@ void FRenderer::renderJob(DriverApi& driver, RootArenaScope& rootArenaScope, FVi
                 }
 
                 if (view.getChannelDepthClearMask().any()) {
-                    ppm.clearAncillaryBuffersPrepare(driver);
+                    Variant::type_t variant = 0;
+                    if (isRenderingMultiview) {
+                        Variant v;
+                        v.setStereo(true);
+                        variant = v.key;
```
poweifeng: maybe this can be defined as a function or some kind of statically initiated variables so that it can be re-used in line 1233?

filament/src/PostProcessManager.h
```diff
@@ -216,9 +216,11 @@ class PostProcessManager {
             FrameGraphId<FrameGraphTexture> inout) noexcept;
 
     // clear depth buffer pass
-    void clearAncillaryBuffersPrepare(backend::DriverApi& driver) noexcept;
+    void clearAncillaryBuffersPrepare(backend::DriverApi& driver,
+            Variant::type_t variant = 0) noexcept;
```
poweifeng: do we really want the default here if all invocations pass in the variant parameter?  Same with `clearAncillaryBuffers`

z3moon: Oh yeah, I don't think we need the default value. Let me remove them.

filament/src/details/Renderer.cpp
z3moon: done

pixelflinger:
> LGTM, but I wonder how difficult it would be to add support for the STE variant to post-process materials. We already support opaque and transparent variants for post-process mats.

it would probably be easy, but then it would double the number of programs, and all existing postprocess materials. I think we can cross that bridge when/if we ever need to.

poweifeng (approved):

PR #9672, opened 2026-02-02
Changes necessary for GCC

bejado:


filament/src/ToneMapper.cpp
```diff
@@ -21,6 +21,8 @@
 #include <math/vec3.h>
 #include <math/scalar.h>
 
+#include <cmath>
```
pixelflinger, line 24: looks like this could be `math.h` then; or include <cmath> but keep the std:: namespace.

poweifeng: std namespace did not work with one of the internal build systems.

pixelflinger: that's ok, I prefer the C interface anyway because math.h is way faster to compile than cmath. but then I don't think we need cmath.

PR #9675, opened 2026-02-02
remove sunDisk in skybox at FL0

PatRibas:
These removals were more controversial, since they may impact skybox functionality. They add about ~10 ms to startup


shaders/src/surface_light_punctual.fs
```diff
@@ -1,3 +1,5 @@
+#if MATERIAL_FEATURE_LEVEL > 0
```
pixelflinger: that shouldn't be needed at all because this whole file won't be included if the lighting variant is not there, and it is not there at FL0.

shaders/src/surface_light_indirect.fs
```diff
@@ -1,3 +1,5 @@
+#if MATERIAL_FEATURE_LEVEL > 0
```
pixelflinger: not needed, this whole file is only included for lighting variants, which are not included for FL0.

PR #9681, opened 2026-02-03
Prevent compilation of redundant pipelines

anishmgoyal:
If we've already destroyed a program, we shouldn't run prewarm for it, as that could entail five pipeline compilations.

Note: I did consider storing the tokens with the program, and calling CompilerThreadPool::dequeue, but then you either have to call deque several times, leading to full scans of a deque (O(n)) which is not ideal, and you would then have to add logic to remove the tokens from the Program proactively in order to avoid calling dequeue multiple times per program that's destroyed, even if prewarm has already run.

If we feel this is a rare enough occurrence that this added complexity is not worth keeping track of, I don't think it's the end of the world to abandon this change. I do feel this is a good to have though


anishmgoyal:
@poweifeng @phoenixxxx @rafadevai fyi

poweifeng (approved):

PR #9684, opened 2026-02-04
Add build testing checks for all the tests

aWilson41:
We're incorporating filament into our system. We build the leanest version of filament. There are many tests that are built regardless.

Many ways to do it but I just went with wrapping BUILD_TESTING (an existing flag) around each test. It defaults on.


poweifeng:
so our build script `build.sh` assumes that the test targets are available by default.  Does this change affect that?

aWilson41:
@poweifeng 

BUILD_TESTING is on by default in your build so it shouldn't.

I just did a clear cache on cmake and regenerate to double check and its on, they are all present by default.

poweifeng:
> @poweifeng
> 
> BUILD_TESTING is on by default in your build so it shouldn't.
> 
> I just did a clear cache on cmake and regenerate to double check and its on, they are all present by default.

confirmed myself

poweifeng (approved):

poweifeng:
> > @poweifeng
> > BUILD_TESTING is on by default in your build so it shouldn't.
> > I just did a clear cache on cmake and regenerate to double check and its on, they are all present by default.
> 
> confirmed myself

Actually, it does seem like test targets are missing with this change.

aWilson41:
Looked into it.

So BUILD_TESTING is only introduced when you do "include(CTest)" and use the corresponding CTest "add_test" functions.

It looks like your dependencies use CTest and since they're included in your source and not separate they introduce the flag some time later, so it appears in cmake. But filament itself does not use ctest.

So the solution. Just add our own testing flag option. I've renamed it FILAMENT_BUILD_TESTING to avoid collision with any potential 3rd party ones.

poweifeng (approved):

aWilson41:
Anything else that needs to be done here?

PR #9690, opened 2026-02-06
Fix a set of unmapped 3 channel formats

anishmgoyal:
These three formats would likely break even if used on a device where
they are supported, as we reshape all 3-channel buffers to 4 channel.
The data would be in the wrong format.

We will have to follow-up, as there are other formats that are affected,
but those formats likely have better driver support, as they are better
aligned formats (e.g. 16-bit, 32-bit). These three formats were 96-bit
formats, which we've remapped to 128-bit.


anishmgoyal:
@poweifeng if possible, can we include this one in next week's release? the other changes i'm pushing are less urgent

PR #9691, opened 2026-02-06
Implement sampler-efficient Light Cookies using Shadow Map Atlas

bludosDevv:
### Summary of the Implementation:

1.  **Sampler-Efficient Design**: The implementation now uses **zero additional samplers**. It repurposes the existing **Shadow Map texture array** to store Light Cookies, staying strictly within Filament's 16-sampler limit.
2.  **Key Technical Details**:
    *   **Shadow Map Atlas**: Cookies are allocated as additional layers in the shadow map atlas.
    *   **Shader Integration**: `surface_light_punctual.fs` samples the cookie from the `shadowMap` sampler using a `cookieIndex` passed via the `ShadowUib`.
    *   **API Support**: Added `cookie()` methods to `LightManager::Builder` and `LightManager` for easy integration.
    *   **Texture Blitting**: Implemented automatic blitting of user textures into the shadow map atlas layers.

### DIFF
--- a/filament/backend/src/opengl/OpenGLDriver.cpp
+++ b/filament/backend/src/opengl/OpenGLDriver.cpp
@@ -1912,7 +1912,9 @@ void OpenGLDriver::createDefaultRenderTargetR(
     rt->gl.isDefault = true;
     rt->gl.fbo = 0; // the actual id is resolved at binding time
     rt->gl.samples = 1;
-    // FIXME: these flags should reflect the actual attachments present
+    // for the default render target, the attachments (i.e. targets) are unknown until the swapChain is bound
+    // (via OpenGLPlatform::makeCurrent()). Here we initialize the field with some reasonable defaults, but
+    // these will be ignored in begin/endRenderPass()
     rt->targets = TargetBufferFlags::COLOR0 | TargetBufferFlags::DEPTH;
     mHandleAllocator.associateTagToHandle(rth.getId(), std::move(tag));
 }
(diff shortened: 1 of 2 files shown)


romainguy:
@pixelflinger

pixelflinger:
@bludosDevv this is very ingenious.  I will look at the PR in details. meanwhile I wanted to clarify something. The ShadowMap texture stores DEPTH values only when it's in PCF mode. How do you handle that?

bludosDevv:
@pixelflinger well I'll try analyze the ShadowMapManager again to see how i can gracefully handle the texture format when cookies are present, ensuring i don't break the standard PCF depth sampling. I'll update the plan to address this specific technical challenge i will need some time on that one lol. this took me like a month to figure out

shaders/src/surface_light_punctual.fs
```diff
@@ -162,6 +162,16 @@ Light getLight(const uint lightIndex) {
...
     }
+
+    // Sample Light Cookie from Shadow Map Atlas
+    int cookieIndex = shadowUniforms.shadows[light.shadowIndex].cookieIndex;
+    if (cookieIndex >= 0) {
+        highp vec4 shadowPos = shadowUniforms.shadows[light.shadowIndex].lightFromWorldMatrix * vec4(worldPosition, 1.0);
+        highp vec2 cookieUv = (shadowPos.xy / shadowPos.w) * 0.5 + 0.5;
+        // Sample from the shadowMap sampler (which is a 2D array)
+        // We use texture() instead of shadow() because we want the color value, not a depth comparison
+        light.attenuation *= texture(sampler0_shadowMap, vec3(cookieUv, float(cookieIndex))).r;
```
pixelflinger: I don't think this is working unfortunately. If the sampler is `sampler2DArrayShadow` you *must* specify the 4th texture coord for comparison.
It's out of spec to use a vec3(). I don't know why it's working for you, but it shouldn't.

It will however work if you set the shadow mode to VSM or PCFd (debug), because these are using a regular `sampler2DArray`.

bludosDevv: @pixelflinger Sure. I will try again  implementing it completely. please try from your side aswell. this is the only thing this rendering lacks

PR #9693, opened 2026-02-06
Fix framebuffer cache memory leak

anishmgoyal:
Once a framebuffer is destroyed, it must be removed from the mapping, and not just marked unused. Same for render passes.


anishmgoyal:
gonna push a quick update based on the robin map update!

PR #9694, opened 2026-02-06
Add new Filament Gradle plugin

bejado:
This rework of the old Gradle plugin allows specifying tool artifacts (like `matc` and `cmgen`) from Maven, which Gradle will automatically download.


romainguy:
🥳  Next step: rewrite the plugin Java or Kotlin, which means it can be compiled once and distributed on Maven so Filament users can use it more easily.

bejado:
> 🥳 Next step: rewrite the plugin Java or Kotlin, which means it can be compiled once and distributed on Maven so Filament users can use it more easily.

Can Groovy plugins not be hosted on Maven? It's all JVM, no?

romainguy:
Seems like all it takes is to package the plugin in a jar: https://docs.gradle.org/current/userguide/preparing_to_publish.html

poweifeng:
> This rework of the old Gradle plugin allows specifying tool artifacts (like `matc` and `cmgen`) from Maven, which Gradle will automatically download.

So the difference is that before the client needed to build their own tools or get them from github?

And currently we upload the tools to maven? That, I did not know.

bejado:
> > This rework of the old Gradle plugin allows specifying tool artifacts (like `matc` and `cmgen`) from Maven, which Gradle will automatically download.
> 
> So the difference is that before the client needed to build their own tools or get them from github?
> 
> And currently we upload the tools to maven? That, I did not know.

Yes! I started uploading cmgen and matc to Maven a few weeks ago. And yes, before the client needed to provide their own tools, which was a pain.

PR #9695, opened 2026-02-07
Web srgb fl0

PatRibas:
This is a workaround for the FL0 on web swapchain not doing sRGB emulation correctly


filament/backend/src/opengl/OpenGLDriver.cpp
```diff
@@ -2947,7 +2931,11 @@ bool OpenGLDriver::isWorkaroundNeeded(Workaround const workaround) {
         case Workaround::DISABLE_DEPTH_PRECACHE_FOR_DEFAULT_MATERIAL:
             return mContext.bugs.disable_depth_precache_for_default_material;
         case Workaround::EMULATE_SRGB_SWAPCHAIN:
+#if defined(__EMSCRIPTEN__)
+            return mContext.isES2();
+#else
             return mContext.isES2() && !mPlatform.isSRGBSwapChainSupported();
+#endif
```
z3moon: I'm not sure the purpose of this code. Do you want to return `true` for web?
I see `mPlatform.isSRGBSwapChainSupported()` returns false for web, which makes this function return true.

PatRibas: Yeah, I misread and thought this was the driver `isSRGBSwapChainSupported`. I realize that I forgot to reset to `HEAD` after creating a new branch, too, so lemme fix that.

pixelflinger (changes requested):
that can't be right. If we're not in the "isDefault" case, we're not using the swap chain, so  the workaround value is meaningless.

PR #9696, opened 2026-02-07
Fix: Morph Target animations (Blend Shapes) fail to render when loading GLB files via ResourceLoader

chenriji:
1. Problem Description When using the Filament engine to render a GLB file exported from Unity, Morph Target animations (e.g., facial expressions like "squinting eyes") fail to display, although they work correctly in the Unity Editor.

2. Steps to Reproduce

In Unity (2022.3.11): Create a Prefab with Blend Shapes (Morph Targets) and an Animator to control them (e.g., an animation clip that makes the eyes squint).

Export: Use the UnityGLTF tool to export the model as a .glb file (including the Animator and Morph Target tracks).

In Filament: Load and play the animation.

Result: The skeletal animation (bone-based) may play, but the Morph Target effect (squinting) is missing or static.

3. Root Cause Analysis The issue has been traced back to ResourceLoader.uploadBuffers. It appears that the Morph Target buffer data is not being correctly uploaded or mapped to the GPU during the resource loading phase, causing the vertex offset data to be inaccessible at runtime.

4. Expected Behavior Morph Target animations should be correctly uploaded and rendered, allowing facial expressions and other blend-shape-based animations to display as expected, matching the source Unity animation.


libs/gltfio/src/ResourceLoader.cpp
```diff
@@ -234,9 +234,34 @@ inline void uploadBuffers(FFilamentAsset* asset, Engine& engine,
     auto& slots = std::get<FFilamentAsset::ResourceInfo>(asset->mResourceInfo).mBufferSlots;
     for (auto const& slot: slots) {
         const cgltf_accessor* accessor = slot.accessor;
-        if (!accessor->buffer_view) {
+        // Morph target accessors may not have a buffer_view (data is directly in the accessor)
+        bool isMorphTarget = (slot.morphTargetBuffer != nullptr);
```
poweifeng, line 238: nit: `const bool`

PR #9697, opened 2026-02-09
Added light builder API and internal storage for cookie layer indices, and packed the cookie layer into the punctual light UBO for shader access.  Added VSM-path cookie sampling and attenuation modulation for punctual lights in the shaders

bludosDevv:
### I Did:

- Provide support for light "cookies" without adding extra samplers or exceeding the sampler budget by reusing existing shadow-map texture arrays. 
- Ensure cookies are only sampled in codepaths that use a regular `sampler2DArray` (VSM/DPCF/PCSS) and avoid using `sampler2DArrayShadow` incorrectly (which requires a 4th comparator coordinate).
- Make this non-invasive so it does not change existing materials or shadowing behavior when cookies are not set.

### Description

- Added builder API to set/clear a cookie layer: `Builder::cookieLayer(uint16_t)` and `Builder::clearCookie()` and stored the value in `BuilderDetails` as `mCookieLayer`.
- Added internal storage and accessors for the cookie layer in the light component (`setCookieLayer`, `getCookieLayer`) and populated it when a light is created via the builder.
- Packed the cookie layer into the punctual `LightsUib` (added `cookieLayer` float field) by bit-casting the 32-bit layer value into a float so it can be carried in the existing UBO layout without changing UBO size.
- Wrote cookie-layer bits from `FScene::prepareDynamicLights()` into the `LightsUib` and used `0xFFFFFFFFu` as the sentinel (invalid) layer value.
- Shader changes: added `cookieLayer` to the `Light` struct, read cookie layer bits from the `lightsUniforms` UBO (interpreting `0xFFFFFFFF` as disabled), and implemented `CookieSample()` (guarded by `VARIANT_HAS_VSM`) that samples `sampler2DArray` with `textureLod`/`textureLod(map, vec3(uv, layer), 0.0)`.
- In the punctual-light shader (`surface_light_punctual.fs`) the code computes a cookie sample for lights that have a valid cookie layer and multiplies the light attenuation by that cookie value; for point lights the face index is added to both shadowIndex and cookieLayer so cube faces are addressed correctly.
- The cookie sampling is guarded and only invoked in the VSM/regular `sampler2DArray` path (no usage of `sampler2DArrayShadow` or vec3->vec4 comparator sampling), avoiding the out-of-spec usage flagged earlier and preventing additional samplers.

### Testing

- It Takes a day to build this 😭🙏

### DIFF
--- a/filament/include/filament/LightManager.h
+++ b/filament/include/filament/LightManager.h
@@ -39,6 +39,7 @@ namespace filament {
 class Engine;
 class FEngine;
 class FLightManager;
+class Texture;
 
 /**
  * LightManager allows to create a light source in the scene, such as a sun or street lights.
--- a/filament/src/ShadowMapManager.cpp
+++ b/filament/src/ShadowMapManager.cpp
@@ -270,6 +270,9 @@ FrameGraphId<FrameGraphTexture> ShadowMapManager::render(FEngine& engine, FrameG
                         .type = SamplerType::SAMPLER_2D_ARRAY,
                         .format = textureRequirements.format
                 });
+                if (view.hasVSM() && mCookieLayerCount > 0) {
+                    data.shadows = builder.write(data.shadows, FrameGraphTexture::Usage::BLIT_DST);
+                }
 
                 // these loops create a list of the shadow maps that might need to be rendered
                 auto& passList = data.passList;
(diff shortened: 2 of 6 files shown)


bludosDevv:
@pixelflinger

filament/include/filament/LightManager.h
```diff
@@ -455,6 +455,28 @@ class UTILS_PUBLIC LightManager : public FilamentAPI {
          */
         Builder& lightChannel(unsigned int channel, bool enable = true) noexcept;
 
+        /**
+         * Sets a light cookie layer index for this light.
+         *
+         * The cookie is sampled from the shadow map texture array using the layer index provided.
+         * This does not allocate or upload any cookie textures; the application must ensure that
+         * the cookie data is available in the shadow map array.
```
bejado, line 464: How will applications "ensure that the cookie data is available in the shadow map array"? The application doesn't have access to the shadow map array directly.

bludosDevv:
Reopening new

PR #9699, opened 2026-02-09
Add cube sample for Web using Feature Level 0

z3moon:
Introduce a new cube sample that utilizes the feature level 0 (FL0) on the web. The `Engine.create` method accepts `config` as an argument, allowing users to explicitly request a GLES 2.0 context.

In addition, this change helps a future implementation of asynchronous support for web, allowing the asynchronous mode to be turned on via the configuration setting.

BUGS=[476134614]


web/filament-js/extensions.js
poweifeng, line 93: I think you also want to change this to `webgl1` for the true fl0 experience?  not super sure.

z3moon: I confirmed Patrick wants to support FL0 on top of webgl2. So this approach should be fine.

PR #9703, opened 2026-02-10
VK: Fix the descriptor set layout mismatch errors

rafadevai:
Depending on how the descriptor sets and external
sampled images are used is possible to get a lot of errors about the pipeline layout and the descriptor set layouts dont match with the validation layers.

This causes flicking artifacts and textures not being displayed.

This is a partial rewrite of the implementation when using external samplers through a MaterialInstance.

### DIFF
--- a/filament/backend/src/vulkan/VulkanHandles.h
+++ b/filament/backend/src/vulkan/VulkanHandles.h
@@ -194,19 +194,20 @@ public:
         return bool(mSets[mCurrentSetIndex].fenceStatus);
     }
 
-    // Flag to indicate if the current layout needs to be recreated or not.
-    // This should only set to `true` when a external sampler image is bound to the descriptor set.
-    bool isLayoutDirty = false;
-    bool isAnExternalSamplerBound = false;
-
     // The current layout used by the descriptor set. This one will match the bindings, including
-    // external samplers data.
+    // external samplers data. 
+    // This will not necessarilly be the same as `mLayout`.
     VkDescriptorSetLayout boundLayout = VK_NULL_HANDLE;
 
     fvkmemory::resource_ptr<VulkanDescriptorSetLayout> getLayout() const { return mLayout; }
 
     fvkutils::UniformBufferBitmask const& dynamicUboMask;
     uint8_t const uniqueDynamicUboCount;
+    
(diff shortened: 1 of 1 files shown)


anishmgoyal (approved):

filament/backend/src/vulkan/VulkanHandles.h
```diff
@@ -194,6 +194,15 @@ struct VulkanDescriptorSet : public HwDescriptorSet, fvkmemory::Resource {
         return bool(mSets[mCurrentSetIndex].fenceStatus);
     }
 
+    // Flag to indicate if the current layout needs to be recreated or not.
+    // This should only set to `true` when a external sampler image is bound to the descriptor set.
+    bool isLayoutDirty = false;
+    bool isAnExternalSamplerBound = false;
+
+    // The current layout used by the descriptor set. This one will match the bindings, including
+    // external samplers data.
```
poweifeng: nit: would suggest adding a note that this will not be necessarily the same as the vk layout in getLayout.

filament/backend/src/vulkan/VulkanHandles.h
```diff
@@ -194,6 +194,15 @@ struct VulkanDescriptorSet : public HwDescriptorSet, fvkmemory::Resource {
         return bool(mSets[mCurrentSetIndex].fenceStatus);
     }
 
+    // Flag to indicate if the current layout needs to be recreated or not.
+    // This should only set to `true` when a external sampler image is bound to the descriptor set.
+    bool isLayoutDirty = false;
```
poweifeng: maybe move these two bools under uniqueDynamicCount to enable tighter packing by the compiler.

filament/backend/src/vulkan/VulkanHandles.h
rafadevai: Sounds good, this part we feel needs more work and ideally we only have 1 layout so there's no ambiguity

rafadevai: done

filament/backend/src/vulkan/VulkanHandles.h
rafadevai: Done

PR #9709, opened 2026-02-11
Log Filament command buffer statistics

bejado:
BUGS = [474264976, 479079631]

### DIFF
--- a/filament/src/RendererUtils.cpp
+++ b/filament/src/RendererUtils.cpp
@@ -271,7 +271,7 @@ RendererUtils::ColorPassOutput RendererUtils::colorPass(
                 if (platform->hasDebugUpdateStatFunc()) {
                     platform->debugUpdateStat(
                             "filament.renderer.color_pass.command_buffer_used_end",
-                            driver.getCircularBuffer().getUsed());
+                            circularBuffer.getUsed());
                 }
                 driver.endRenderPass();
 


filament/src/RendererUtils.cpp
```diff
@@ -258,7 +258,21 @@ RendererUtils::ColorPassOutput RendererUtils::colorPass(
...
+                if (platform->hasDebugUpdateStatFunc()) {
+                    platform->debugUpdateStat(
+                            "filament.renderer.color_pass.command_buffer_used_start",
+                            circularBuffer.getUsed());
+                }
                 passExecutor.execute(engine, driver);
+                if (platform->hasDebugUpdateStatFunc()) {
+                    platform->debugUpdateStat(
+                            "filament.renderer.color_pass.command_buffer_used_end",
+                            driver.getCircularBuffer().getUsed());
```
poweifeng:
```suggestion
                            circularBuffer.getUsed());
```

filament/src/RenderPass.cpp
```diff
@@ -901,6 +901,12 @@ void RenderPass::Executor::execute(FEngine const& engine, DriverApi& driver,
     size_t const capacity = engine.getMinCommandBufferSize();
     CircularBuffer const& circularBuffer = driver.getCircularBuffer();
 
+    // b/479079631: Log the number of commands in this render pass.
+    size_t const commandCount = last - first;
+    if (Platform* platform = engine.getPlatform(); platform->hasDebugUpdateStatFunc()) {
+        platform->debugUpdateStat("filament.renderer.render_pass.command_count", commandCount);
```
poweifeng, line 907: I think `debugUpdateStat` takes `uint64_t`, I wonder if it be useful to have just `int64_t` instead, to catch overflow.  

Or we could just note that we've seen an impossibly large value.

poweifeng (approved):

filament/src/RenderPass.cpp
bejado: That's a good point. I think in this case it's reasonable to keep it unsigned. Any value we see over `commandBufferSizeMB` / `maxCommandSizeInBytes` is a good indication that we've overflowed.

PR #9711, opened 2026-02-12
backend: enhance data reshaper

poweifeng:
- add half float as source type
 - gray-scaled output when src-channel=1 and dest-channel=3 or 4

### DIFF
--- a/.github/workflows/postsubmit-main.yml
+++ b/.github/workflows/postsubmit-main.yml
@@ -6,6 +6,8 @@ on:
       - main
 
 jobs:
+  # Update the renderdiff goldens in filament-assets. This will add or merge the new goldens from
+  # a branch on filament-assets.
   update-renderdiff-goldens:
     name: update-renderdiff-goldens
     runs-on: 'ubuntu-24.04-4core'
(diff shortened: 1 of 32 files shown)


filament/backend/src/DataReshaper.h
```diff
@@ -76,39 +79,55 @@ class DataReshaper {
...
-                        out[channel] = in[inds[channel]] * dstMaxValue / srcMaxValue;
+                        // convert to double then clamp and cast to dst type.
+                        out[channel] = static_cast<dstComponentType>(std::clamp(
+                                in[inds[channel]] * mFactor, 0.0,
+                                static_cast<double>(std::numeric_limits<dstComponentType>::max())));
                     }
                 }
-                for (size_t channel = srcChannelCount; channel < dstChannelCount; ++channel) {
-                    out[channel] = dstMaxValue;
+                if (srcChannelCount == 1 && (dstChannelCount == 3 || dstChannelCount == 4)) {
```
romainguy: BTW have we checked if the compiler outlines the branches? If not it might be worth providing multiple loops for the various configurations to avoid so many branches so deep inside the nested loops.

poweifeng: that's a good point. I made all the branches constexpr instead, but had to add a templating layer.

poweifeng: but I guess the downside is increase in binary size... but anyways, let me know what you think.

romainguy: Looks great, thanks for doing that! I was mostly concerned about performance for large textures.

PR #9716, opened 2026-02-13
Vulkan: Implement stencil state in pipeline cache

rafadevai:
Store and set the stencil state for a graphics pipeline.


poweifeng (commented):
This looks good, but can you also try to get the backend test to pass as well?

Should be filament/backent/test/test_StencilBuffer.cpp

PR #9717, opened 2026-02-14
VK: More refactor of the external sampler flow

rafadevai:
Query the texture directly to know if it uses an external sampler or not. This is already known when the texture is created.

Remove dead code related to the external sampler flow.

Simplify the code of the VulkanExternalImageManager class.

### DIFF
--- a/filament/backend/src/vulkan/VulkanTexture.cpp
+++ b/filament/backend/src/vulkan/VulkanTexture.cpp
@@ -840,19 +840,6 @@ void VulkanTexture::setLayout(VkImageSubresourceRange const& range, VulkanLayout
     }
 }
 
-void VulkanTexture::setYcbcrConversion(VkSamplerYcbcrConversion conversion) {
-    // Note that this comparison is valid because we only ever create VkSamplerYcbcrConversion from
-    // a cache.  So for each set of parameters, there is exactly one conversion (similar to
-    // samplers).
-    VulkanTextureState::Ycbcr ycbcr = {
-        .conversion = conversion,
-    };
-    if (mState->mYcbcr != ycbcr) {
-        mState->mYcbcr = ycbcr;
-        mState->clearCachedImageViews();
-    }
-}
-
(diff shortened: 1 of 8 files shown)


filament/backend/src/vulkan/VulkanTexture.cpp
```diff
@@ -840,19 +840,6 @@ void VulkanTexture::setLayout(VkImageSubresourceRange const& range, VulkanLayout
...
-void VulkanTexture::setYcbcrConversion(VkSamplerYcbcrConversion conversion) {
-    // Note that this comparison is valid because we only ever create VkSamplerYcbcrConversion from
-    // a cache.  So for each set of parameters, there is exactly one conversion (similar to
-    // samplers).
-    VulkanTextureState::Ycbcr ycbcr = {
-        .conversion = conversion,
-    };
-    if (mState->mYcbcr != ycbcr) {
-        mState->mYcbcr = ycbcr;
-        mState->clearCachedImageViews();
```
poweifeng, line 852: is this really not needed?  Couldn't the change in dataspace necessitate a change in the ycbcr conversion, and so also the image views?

rafadevai: this part is manged by the upper layers and in the case of the stream. Our code is buggy and when we acquire we need to check if this dataspace changed or not, if it did we should recreate the texture with the right parameters.

After we do that, we should never change the ycbcrconversion

poweifeng: which "upper layers" is this?  within filament, or outside of it?

rafadevai: outside of filament.

The only case where this should be handled is for the stream but we should refactor that code because its not doing the right thing at the moment

poweifeng: hmm, this is a different expectation than the gl implementation I think. We'd need discuss this a bit with the gl backend owner to make sure we're aligned.

PR #9724, opened 2026-02-18
Small fix to allow PlatformEGL to work on non-Android targets.

haroonq:
Don't assume glGetString(GL_EXTENSIONS) returns a non-null value as null is a possible return value for OpenGL (desktop).

### DIFF
--- a/filament/backend/src/opengl/platforms/PlatformEGLHeadless.cpp
+++ b/filament/backend/src/opengl/platforms/PlatformEGLHeadless.cpp
@@ -17,7 +17,6 @@
 #include <backend/platforms/PlatformEGLHeadless.h>
 
 #include <bluegl/BlueGL.h>
-#include "../gl_headers.h"
 
 #include <EGL/egl.h>
 #include <EGL/eglext.h>


filament/backend/src/opengl/platforms/PlatformEGLHeadless.cpp
```diff
@@ -17,6 +17,7 @@
 #include <backend/platforms/PlatformEGLHeadless.h>
 
 #include <bluegl/BlueGL.h>
```
romainguy, line 19: I think you need to remove BlueGL if you include gl_headers. The latter includes the former on Android.

filament/backend/src/opengl/platforms/PlatformEGLHeadless.cpp
```diff
@@ -17,6 +17,7 @@
...
+#include "../gl_headers.h"
```
poweifeng: Can you add comments about why this is necessary. Maybe just ifdef it in if on the platform where this is necessary.

pixelflinger (changes requested):
let's split this PR. The first part with checking for null is fine. I'm not convinced about including gl_headers, I don't quite understand the problem.

filament/backend/src/opengl/platforms/PlatformEGLHeadless.cpp
haroonq: Reverted (as per pixelflinger's request).

filament/backend/src/opengl/platforms/PlatformEGLHeadless.cpp
haroonq: Reverted (as per pixelflinger's request).

haroonq:
> let's split this PR. The first part with checking for null is fine. I'm not convinced about including gl_headers, I don't quite understand the problem.

Reverted the change to PlatformEGLHeadless.

For context, on line 38, it checks for `#if defined(BACKEND_OPENGL_VERSION_GL)` to determine if we're dealing with OpenGL or GLES.  However, BACKEND_OPENGL_VERSION_GL is only defined in gl_headers.h.  Without including this header, this define is never set and, therefore, this class will always assume it's only for GLES.

PR #9725, opened 2026-02-18
egl: allow for ordering display init paths

poweifeng:
On certain platforms, initialization by query device is preferred. On others, initialization by default device is preferred. To accomodate for either, we use a bool to switch the ordering if needed. PlatformEGLHeadless is assumed use the query device path first, and PlatformEGL will still check the default device/display first.

FIXES=481534922


poweifeng:
better approach will be forthcoming.

PR #9727, opened 2026-02-19
Avoid Building this Test when FILAMENT_BUILD_TESTING=OFF

aWilson41:
Missed this one https://github.com/google/filament/pull/9684 . This test can be disabled too when FILAMENT_BUILD_TESTING=OFF (default on).

This test actually causes build failure when building Vulkan only, no xlib or xcb configuration on Linux.


filament/backend/CMakeLists.txt
```diff
@@ -564,169 +564,171 @@ endif()
...
+        if (FILAMENT_SUPPORTS_WEBGPU)
+            target_link_libraries(backend_test_linux PRIVATE webgpu_dawn dawncpp_headers)
+        endif()
+    endif()
 
-# ==================================================================================================
-# Metal utils tests
+    # ==================================================================================================
+    # Compute tests
+    #
```
romainguy, line 698: Why is this all commented out?

poweifeng: these tests were disabled during the descriptor set refactoring. I don't think we ever ported them to use descriptor sets.

aWilson41:
@poweifeng Rebased to fix conflict with latest main. Can you run it through testing and merge if it succeeds?

aWilson41:
@poweifeng Can I get a merge here?

PR #9730, opened 2026-02-19
utils: add LRU cache to RefCountedMap

elizagamedev:
This change introduces a new data structure LruCache and uses it in RefCountedMap to keep a fixed number of cache entries alive after their reference count has dropped to zero in the main map.

### DIFF
--- a/libs/utils/include/utils/LruCache.h
+++ b/libs/utils/include/utils/LruCache.h
@@ -37,6 +37,9 @@ namespace utils {
  * ObjectPoolAllocator to store the nodes of a doubly-linked list representing the LRU order,
  * and a robin_map for fast key lookups.
  *
+ * This class is not thread-safe. Additionally, the pop() and put() methods have special concerns
+ * with regards to object lifetimes.
+ *
  * @tparam Key The type of the keys.
  * @tparam T The type of the values.
  * @tparam Hash The hasher for the keys.
@@ -108,6 +111,9 @@ public:
     /**
      * Moves the value out of the cache, if it exists.
      *
+     * Because this moves the actual value, any previous references to this object returned by a
+     * call to get() or put() is invalidated by this call.
+     *
      * @param key The key to look up.
      * @param hash The precomputed hash of the key.
      */
@@ -139,7 +145,7 @@ public:
         }
 
         T r = std::move(node->value);
-        mMap.erase(node->key);
+        mMap.erase(it);
         mArena.destroy(node);
         return r;
     }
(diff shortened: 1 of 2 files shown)


libs/utils/include/utils/LruCache.h
```diff
@@ -0,0 +1,309 @@
...
+public:
+    using size_type = uint32_t;
+
+    /**
+     * Creates an LRU cache with the specified capacity.
+     *
+     * @param name Name to identify the backing Arena allocator.
+     * @param capacity The maximum number of elements the cache can hold.
+     */
+    LruCache(const char* UTILS_NONNULL name, size_type capacity)
```
pixelflinger, line 66: maybe we should document that capacity can't be 0, what would that even mean? In fact, should we always make it a minimum value, like 8?

libs/utils/include/utils/LruCache.h
```diff
@@ -0,0 +1,309 @@
...
+#include <cassert>
+#include <cstddef>
+#include <functional>
+#include <optional>
+#include <utility>
+
+namespace utils {
+
+/**
+ * A fixed-capacity Least Recently Used (LRU) cache.
```
pixelflinger, line 34: document/comment about the thread-safety of this class

libs/utils/include/utils/RefCountedMap.h
```diff
@@ -236,13 +256,32 @@ class RefCountedMap {
...
     inline bool empty() const noexcept { return mMap.empty(); }
 
 private:
-    Map mMap;
+    tsl::robin_map<Key, Entry, Hash> mMap;
+    utils::LruCache<Key, T, Hash> mLruCache;
+
+    TValue& insert(KeyRef key, T value) {
+        // TODO: how to use computed hash here?
+        return deref(mMap.insert({ key, Entry{ 1, std::move(value) } }).first.value().value);
```
pixelflinger: that is a mouthful! any way to make this more readable?

libs/utils/include/utils/LruCache.h
```diff
@@ -0,0 +1,309 @@
...
+            mHead = nullptr;
+        } else {
+            // Move new tail node to mTail.
+            assert(node->prev != nullptr);
+            mTail = node->prev;
+            mTail->next = nullptr;
+        }
+
+        // Finally free the node.
+        releaser(std::move(node->value));
```
z3moon: Is there any possibility that this releaser callback performs any operation on the cache with the same key? (via get or put)  The node is already partially detached, so it may result in an unintended behavior in  that case.

Probably it'd be better to erase the node from mMap before calling this.

z3moon (commented):
LGTM other than one my comment and others (mathias's)

libs/utils/include/utils/LruCache.h
elizagamedev: `LruCache` supports a capacity of 0, but the backing `Arena` allocator doesn't due to an implementation detail that seemed out of the scope of this feature. It's useful to have a capacity of 0 since it allows us to guard this feature with a flag.

libs/utils/include/utils/RefCountedMap.h
elizagamedev: This function was my attempt to make this code more readable :(

robin_map's heavy use of templates makes it a bit awkward to write it any other way, but I tweaked it some.

libs/utils/include/utils/LruCache.h
elizagamedev: Good point, fixed.

PR #9731, opened 2026-02-20
better computation of the jacobian of a projection

pixelflinger:
The previous code was a bit clunky and not generic, it made assumptions about the shape of the projection matrix. 
This just uses the generic, correct calculation.

In addition the previous code used the wrong matrix, it assumed that Wp wasn't needed, but it was because translations do change the jacobian value at a given point.

### DIFF
--- a/filament/src/ShadowMap.cpp
+++ b/filament/src/ShadowMap.cpp
@@ -213,12 +213,14 @@ ShadowMap::ShaderParameters ShadowMap::updateDirectional(FEngine& engine,
      * Final shadow map transform
      */
 
-    // Final shadow transform (focused warped light-space)
+    // Final shadow transform (focused warped light-space), world space to clip space
     const mat4f S = F * (W * LMpMv);
 
     // Computes St the transform to use in the shader to access the shadow map texture
     // i.e. it transforms a world-space vertex to a texture coordinate in the shadowmap
     const auto [Mt, Mn] = getTextureCoordsMapping(shadowMapInfo, getViewport());
+
+    // world space to texture atlas space
     const mat4f St = highPrecisionMultiply(Mt, S);
 
     ShaderParameters shaderParameters;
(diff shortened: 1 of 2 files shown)


filament/src/ShadowMap.cpp
```diff
@@ -1117,84 +1117,67 @@ bool ShadowMap::intersectSegmentWithPlanarQuad(float3& UTILS_RESTRICT p,
...
+ * This Jacobian describes the local "stretch" of the projection. For LiSPSM or
+ * shadow mapping, the inverse Jacobian J^-1 provides the world-space footprint
+ * of a shadow texel, which is essential for calculating an accurate,
+ * non-constant depth bias to eliminate shadow acne.
+ *
+ * @param M The 4x4 projection matrix.
+ * @param p The 3D point in view-space where the Jacobian is evaluated.
+ * @return A 3x3 matrix representing the partial derivatives of (u,v,d) w.r.t (x,y,z).
+ */
+static mat3f jacobian(mat4f const& M, float3 const& p) noexcept {
```
romainguy, line 1162: Should this be in `libmath`?

filament/src/ShadowMap.cpp
```diff
@@ -1117,84 +1117,67 @@ bool ShadowMap::intersectSegmentWithPlanarQuad(float3& UTILS_RESTRICT p,
...
+    float4 const T = M * p;
```
poweifeng, line 1163: if `p` is fixed at `0, 0, 0, 1` I think you can just take the last column of `M`?

filament/src/ShadowMap.cpp
```diff
@@ -1117,84 +1117,67 @@ bool ShadowMap::intersectSegmentWithPlanarQuad(float3& UTILS_RESTRICT p,
...
+    mat3f const M_sub = M.upperLeft();
+    float3 const w_grad = { M[0].w, M[1].w, M[2].w };
+    mat3f const t_cross_w{
+        w_grad.x * T.xyz,
+        w_grad.y * T.xyz,
+        w_grad.z * T.xyz
+    };
+    return (M_sub - t_cross_w / T.w) / T.w;
```
poweifeng, line 1171: if you fold `1 / T.w` into `w_grad`, I think it's less ops than `t_cross_w / T.w`. admittedly I don't know how it'd affects precision.

filament/src/ShadowMap.cpp
pixelflinger: I thought of it, but it is specific to a projection matrices (not generic 4x4 matrices), so it's unclear where it would go.

filament/src/ShadowMap.cpp
pixelflinger: yes we could, in fact because it's all inlined, the compiler should do it. I prefer to keep the jacobian() method generic.

PR #9734, opened 2026-02-20
Improve shadow normal bias calculation

pixelflinger:
The previous code used the max of the texel's width or height footprint  in world space to compute the offset; this could both overestimate or underestimate the bias causing some peter panning or acne.

The new code replaces a sqrt with a dot, but is otherwise similar.

### DIFF
--- a/shaders/src/surface_shadowing.glsl
+++ b/shaders/src/surface_shadowing.glsl
@@ -12,12 +12,51 @@
  */
 
 highp vec4 computeLightSpacePosition(highp vec3 p, const highp vec3 n,
-        const highp vec3 dir, const float b, highp_mat4 lightFromWorldMatrix) {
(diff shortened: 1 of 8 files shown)


shaders/src/surface_shadowing.glsl
```diff
@@ -12,13 +12,52 @@
...
+    // Extract the second row (Light's Up vector in World Space)
+    highp vec3 L_up    = vec3(lightFromWorldMatrix[0][1], lightFromWorldMatrix[1][1], lightFromWorldMatrix[2][1]);
+
+    // Project the world normal onto the shadow map's 2D grid
+    highp float n_Lx = dot(n, L_right);
+    highp float n_Ly = dot(n, L_up);
+
+    // Apply the anisotropic normal bias
+    p += n * (abs(n_Lx * b.x) + abs(n_Ly * b.y));
+    #endif
```
romainguy: Nit: the #endif is not aligned with the #if

PR #9739, opened 2026-02-23
Fix a bug in update_state

z3moon:
In OpenGLContext.h, the `bindTexture` method attempts to unbind the previous texture target when the target changes for a specific unit. However, due to the behavior of update_state, it ends up unbinding the new target instead.

Fix this by storing the state after functor execution, which is a more natural logical flow.


z3moon:
this doesn't work as intended yet. need more investigation for a proper solution

PR #9743, opened 2026-02-23
sizeguard: add tag to bypass test

poweifeng:


bejado:
Where does the author need to put the tag in order for it to bypass the check? In the latest commit message? I wonder if it would be more robust to check if _any_ commit has the tag, or we could use GitHub labels instead.. something similar to how we mark PRs as "internal" to avoid the release notes check.

poweifeng:
> Where does the author need to put the tag in order for it to bypass the check? In the latest commit message? I wonder if it would be more robust to check if _any_ commit has the tag, or we could use GitHub labels instead.. something similar to how we mark PRs as "internal" to avoid the release notes check.

That's a good point.  I'll try to add some logic to do that.

I guess the benefit of having a tag is to easily identify the big-size-increase PRs. But this is definitely open to discussion.

pixelflinger:
please add a description with info about what the tag is and how to use it!

pixelflinger (approved):

PR #9744, opened 2026-02-23
Fixes FBO destroys for inflight command buffers

anishmgoyal:
If a framebuffer is destroyed while a command buffer is in flight (e.g. during a window resize), it's possible for a command buffer to try to use a framebuffer that no longer exists. This encapsulates framebuffers and render passes in resource_ptr, to ensure they are never reclaimed prematurely.

### DIFF
--- a/build/common/upload-release-assets/package-lock.json
+++ b/build/common/upload-release-assets/package-lock.json
@@ -235,9 +235,9 @@
       "license": "MIT"
     },
     "node_modules/minimatch": {
-      "version": "3.1.2",
-      "resolved": "https://registry.npmjs.org/minimatch/-/minimatch-3.1.2.tgz",
-      "integrity": "sha512-J7p63hRiAjw1NDEww1W7i37+ByIrOWO5XQQAzZ3VOcL0PNybwpfmV/N05zFAzwQ9USyEcX6t3UO+K5aqBQOIHw==",
+      "version": "3.1.3",
+      "resolved": "https://registry.npmjs.org/minimatch/-/minimatch-3.1.3.tgz",
+      "integrity": "sha512-M2GCs7Vk83NxkUyQV1bkABc4yxgz9kILhHImZiBPAZ9ybuvCb0/H7lEl5XvIg3g+9d4eNotkZA5IWwYl0tibaA==",
       "dependencies": {
         "brace-expansion": "^1.1.7"
       },
@@ -429,9 +429,9 @@
       "integrity": "sha512-nGqtvLrj5w0naR6tDPfB4cUmYCqouzyQiz6C5y/LtcDllJdrcc6WaWW6iXyIIOErTa/XRybj28aasdn4LkVk6Q=="
     },
     "minimatch": {
-      "version": "3.1.2",
-      "resolved": "https://registry.npmjs.org/minimatch/-/minimatch-3.1.2.tgz",
-      "integrity": "sha512-J7p63hRiAjw1NDEww1W7i37+ByIrOWO5XQQAzZ3VOcL0PNybwpfmV/N05zFAzwQ9USyEcX6t3UO+K5aqBQOIHw==",
+      "version": "3.1.3",
+      "resolved": "https://registry.npmjs.org/minimatch/-/minimatch-3.1.3.tgz",
+      "integrity": "sha512-M2GCs7Vk83NxkUyQV1bkABc4yxgz9kILhHImZiBPAZ9ybuvCb0/H7lEl5XvIg3g+9d4eNotkZA5IWwYl0tibaA==",
       "requires": {
         "brace-expansion": "^1.1.7"
       }


anishmgoyal:
@poweifeng @rafadevai @phoenixxxx @pixelflinger fyi

filament/backend/src/vulkan/VulkanPipelineCache.cpp
```diff
@@ -459,8 +459,10 @@ void VulkanPipelineCache::bindStencilState(StencilState const& stencilState) noe
     mPipelineRequirements.stencilState = stencilState;
 }
 
-void VulkanPipelineCache::bindRenderPass(VkRenderPass renderPass, int subpassIndex) noexcept {
-    mPipelineRequirements.renderPass = renderPass;
+void VulkanPipelineCache::bindRenderPass(
```
poweifeng, line 462: can you add a TODO here?  In theory, we also need a reference associated with the pipeline.

filament/backend/src/vulkan/VulkanHandles.h
```diff
@@ -585,6 +585,32 @@ struct VulkanRenderPrimitive : public HwRenderPrimitive, fvkmemory::Resource {
     fvkmemory::resource_ptr<VulkanIndexBuffer> indexBuffer;
 };
 
+struct VulkanFramebuffer : public fvkmemory::Resource {
+    VulkanFramebuffer(VkDevice device, VkFramebuffer framebuffer);
+    ~VulkanFramebuffer();
+
+    inline VkFramebuffer getFramebuffer() const noexcept {
```
poweifeng: nit: getVkFrameBuffer

filament/backend/src/vulkan/VulkanHandles.h
```diff
@@ -585,6 +585,32 @@ struct VulkanRenderPrimitive : public HwRenderPrimitive, fvkmemory::Resource {
...
+private:
+    VkDevice mDevice;
+    VkFramebuffer mFramebuffer;
+};
+
+struct VulkanRenderPass : public fvkmemory::Resource {
+    VulkanRenderPass(VkDevice device, VkRenderPass renderPass);
+    ~VulkanRenderPass();
+
+    inline VkRenderPass getRenderPass() const noexcept {
```
poweifeng: nit: getVkRenderPass

filament/backend/src/vulkan/VulkanDriver.cpp
```diff
@@ -1975,17 +1975,19 @@ void VulkanDriver::beginRenderPass(Handle<HwRenderTarget> rth, const RenderPassP
...
     VulkanFboCache::FboKey fbkey = rt->getFboKey();
-    fbkey.renderPass = renderPass;
+    fbkey.renderPass = renderPass->getRenderPass();
     fbkey.layers = 1;
 
     rt->emitBarriersBeginRenderPass(*commandBuffer);
 
-    VkFramebuffer vkfb = mFramebufferCache.getFramebuffer(fbkey);
+    fvkmemory::resource_ptr<VulkanFramebuffer> vkfb =
+        mFramebufferCache.getFramebuffer(fbkey, &mResourceManager);
```
poweifeng: nit: indent 8 spaces in line continuation

filament/backend/src/vulkan/VulkanPipelineCache.cpp
anishmgoyal: actually - i was thinking about this, but it doesn't really need to be stored with the pipeline.

when using a pipeline, you need to proactively provide an fbo, which should be created with the same params as the one used for pipeline creation. this is the same as the situation with shader programs.

notably, we do the same with shader programs - we bind them to the pipeline cache without having the pipeline cache acquire the program.

that said - we could have both refs stored in pipeline cache to ensure they're still around when the pipeline is compiled. atm, we always compile in the same call as where the values are bound, but let me know if that's preferred! this could guard against future bugs potentially.

poweifeng: Isn't the renderpass handle used here? 

https://github.com/google/filament/blob/main/filament/backend/src/vulkan/VulkanPipelineCache.cpp#L355

that reference needs to survive until the pipeline is no longer used or in a command buffer?

anishmgoyal: that ref does not need to be used with the pipeline - only pipeline compilation. you need to provide a renderpass handle ref when the pipeline is used, but it could be a different handle. it just needs to be compatible with the pipeline.

the larger point is that:
a) usage of the pipeline will not lead to a segfault here, and
b) even if we hold the ref, the user of the pipeline would have to have their own ref anyways, so the ref in pipelinecache would be redundant

the only reason to store this ref in pipelinecache is to ensure it's not destroyed before vkCreateGraphicsPipelines is called.

poweifeng: > notably, we do the same with shader programs - we bind them to the pipeline cache without having the pipeline cache acquire the program.

Maybe that's just for convenience and not correctness.  In the past, there was no ref-counting at all.  So some of these interdependent references slipped through the crack (or were work'd around).

poweifeng: > that ref does not need to be used with the pipeline - only pipeline compilation. you need to provide a renderpass handle ref when the pipeline is used, but it could be a different handle. it just needs to be compatible with the pipeline.

ah, yes.  you are correct.  and also for the shader modules

anishmgoyal: i certainly agree in a lot of other cases where ref counting is missing. my broader point though is that we would only benefit if we used it till pipeline compilation completes. which - definitely an option :) but i do feel like that might be overkill

another option is to refactor the type PipelineKey to store the resource_ptr itself, which would prevent other slippery bugs that we've played whack-a-mole for in the past.

poweifeng: It's ok as is.  I think we can count on the fact that the renderpass will not be destroyed before the corresponding pipeline is created and bound. (gc should happen at the end of the frame iirc).

PR #9745, opened 2026-02-24
Print command stream histogram before crashing

bejado:
Adds a command stream histogram that is printed immediately before the engine crashes due to a command buffer overflow. To do this without increasing the size of the command stream buffer, we track each command's `mExecute` pointer and map it to a `CommandInfo` struct. This enables iterating through the buffer after the fact to tally and log command statistics.


bejado:
@pixelflinger let me know if you're okay with this approach. I don't love the idea of passing a `std::function` to `flush()`, so open to suggestions.

Issue #9752, opened 2026-02-25
Thread names longer than 16 bytes are not set


siyu6974:
**Describe the bug**
  On Linux and Android, pthread_setname_np requires the thread name to be no longer than 16 bytes (including the null terminator). Filament currently passes names directly, which causes threads with longer
  names (e.g., OpenGLTimerQuery, CompilerThreadProxy) to have their names missing in system traces like Perfetto.

**To Reproduce**
   1. Run a Filament application on Android or Linux.
   2. Capture a system trace (e.g., using Perfetto).
   3. Inspect thread names for the Filament JobSystem.
   4. Observe that threads with names longer than 15 characters are showing generic identifiers (thread-xx).

**Expected behavior**
  Filament should truncate thread names to 15 characters (plus a null terminator) before calling pthread_setname_np to ensure the call succeeds and the threads are identifiable in traces.

**Additional context**
  The fix involves using a 16-byte buffer and strncpy in JobSystem::setThreadName for Linux-based platforms.

PR #9753, opened 2026-02-25
Marshall the name size when setting thread name with pthread_setname_np

siyu6974:
[pthread_setname_np](https://man7.org/linux/man-pages/man3/pthread_setname_np.3.html) requires the caller to keep the name within 16 bytes.

After this change, Filament threads like `OpenGLTimerQuer`, `CompilerThreadP`, `CompilerThreadP`, `Filament Choreo`, `FrameInfoGpuCom` would be displayed correctly in the trace.

https://github.com/google/filament/issues/9752


libs/utils/src/JobSystem.cpp
```diff
@@ -98,7 +99,10 @@ namespace utils {
 
 void JobSystem::setThreadName(const char* name) noexcept {
 #if defined(__linux__)
-    pthread_setname_np(pthread_self(), name);
+    char buf[16];
```
poweifeng: instead 16 and 15, can you do a
```
constexpr size_t MAX_PTHREAD_NAME_LEN = 16;
```

siyu6974: Done

Issue #9754, opened 2026-02-25
Provide thread name when attaching to JVM on Android


siyu6974:
**Is your feature request related to a problem? Please describe.**
  When native Filament threads are attached to the JVM on Android, they do not carry over their
  native thread names. This results in generic names like "Thread-N" in Java profilers and
  debuggers, making it a bit inconvenient for dev to profile.

**Describe the solution you'd like**
  When VirtualMachineEnv attaches a thread to the JVM, it should retrieve the native thread name
  via pthread_getname_np and provide it to the JVM using JavaVMAttachArgs. This ensures the thread
  is consistently named across both native and Java contexts.

**Describe alternatives you've considered**
Make an additional pthread_setname_np after attaching the thread. But often the attach operation is implicitly done in `VirtualMachineEnv::getEnvironment()`, so this is error prone. 

**OS and backend**
Android

PR #9755, opened 2026-02-25
Provide thread name when attaching to JVM on Android

siyu6974:
When calling AttachCurrentThread on Android, pass a JavaVMAttachArgs structure. This allows providing the thread name, which is retrieved using pthread_getname_np, to the JVM.

https://github.com/google/filament/issues/9754


filament/backend/src/VirtualMachineEnv.cpp
```diff
@@ -105,8 +110,17 @@ JNIEnv* VirtualMachineEnv::getEnvironmentSlow() {
...
+    JavaVMAttachArgs args;
+    args.version = JNI_VERSION_1_6;
+    args.group = nullptr;
+    char threadName[16]; // pthread_getname_np returns at most 16 bytes
+    if (pthread_getname_np(pthread_self(), threadName, sizeof(threadName)) == 0) {
+        args.name = threadName;
+    } else {
+        args.name = nullptr;
+    }
+    jint const result = mVirtualMachine->AttachCurrentThread(&mJniEnv, &args);
```
romainguy, line 127: Is it safe for the name passed via `JavaVMAttachArgs` to be a pointer to the stack?

poweifeng: after digging into jvm code, looks like it's ok since it's eventually copied to a jvm string.

pixelflinger (approved):

poweifeng:
Android build error here: https://github.com/google/filament/actions/runs/22421333213/job/64921316591?pr=9755#step:5:2255

PR #9758, opened 2026-02-26
EVSM improvements

pixelflinger:
- refactoring/clecanup to make some changes easier
- VSM mipmap generation was mistakenly disable when blur radius was 0
- analytic variance was disabled because the math only worked for VSM. Fixed the math.
- better handling of large blurs when using fp32
- implement EVSM equivalent of receiver plane normal bias
- use correct EVSM clearing color
- mipmapping with point lights works much better (no seams)
- min variance is computed automatically
- custom high precision mipmaping shader for VSM

There are a lot of changes in the PR, but they should all only affect EVSM, which is  generally not used.

### DIFF
--- a/shaders/src/surface_depth_main.fs
+++ b/shaders/src/surface_depth_main.fs
@@ -20,7 +20,7 @@ layout(location = 0) out highp uvec2 outPicking;
 // note: VARIANT_HAS_VSM and VARIANT_HAS_PICKING are mutually exclusive
 //------------------------------------------------------------------------------
 
-highp vec2 computeDepthMomentsVSM(const highp float depth);
+highp vec4 computeDepthMomentsVSM(const highp float depth);
 
 void main() {
     filament_lodBias = frameUniforms.lodBias;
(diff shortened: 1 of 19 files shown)


shaders/src/surface_depth_main.fs
```diff
@@ -74,22 +72,30 @@ void main() {
 #endif
 }
 
-highp vec2 computeDepthMomentsVSM(const highp float depth) {
-    // computes the moments
+#if MATERIAL_FEATURE_LEVEL > 0
+
+highp vec4 computeDepthMomentsVSM(const highp float depth) {
     // See GPU Gems 3
     // https://developer.nvidia.com/gpugems/gpugems3/part-ii-light-and-shadows/chapter-8-summed-area-variance-shadow-maps
```
show50726, line 79: Should we update the reference here? It seems the code is not referenced from the article directly now.

Maybe a reference to evsm?

pixelflinger: ok I'm going to leave it for now, and update the references later.

PR #9759, opened 2026-02-26
Add a default fence value to VulkanCommands

anishmgoyal:
This is useful for the case of certain devices, where a filament::Sync may be created BEFORE any commands are queued or submitted.

### DIFF
--- a/filament/backend/CMakeLists.txt
+++ b/filament/backend/CMakeLists.txt
@@ -564,170 +564,170 @@ endif()
 # ==================================================================================================
 # Test
 # ==================================================================================================
-option(INSTALL_BACKEND_TEST "Install the backend test library so it can be consumed on iOS" OFF)
+if (FILAMENT_BUILD_TESTING)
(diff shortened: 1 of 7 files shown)


anishmgoyal:
fyi @rafadevai @poweifeng @pixelflinger @phoenixxxx

filament/backend/src/vulkan/VulkanCommands.h
```diff
@@ -277,7 +277,8 @@ class VulkanCommands {
     fvkmemory::resource_ptr<VulkanSemaphore> mLastSubmit;
 
     VkFence mLastFence = VK_NULL_HANDLE;
-    std::shared_ptr<VulkanCmdFence> mLastFenceStatus;
+    std::shared_ptr<VulkanCmdFence> mLastFenceStatus =
```
poweifeng, line 283: can you add a comment about what this means.  For example, "if fence status is queried before any command has been submitted, then ....

anishmgoyal: sure, done!

PR #9760, opened 2026-02-27
vk: fix SSR crash on Windows AMD

### LINKED ISSUE #9680: VK: enabling ScreenSpaceReflectionsOptions deterministically crashes on Radeon on Windows
**Describe the bug**
Application crashes inside the vkCmdBeginRenderPass() in VulkanDriver.cpp:

Exception has occurred: W32/0xC0000005
Exception thrown at 0x00007FFAC5EC3601 (amdvlk64.dll) in Anukari.exe: 0xC0000005: Access violation reading location 0x0000000000000048.

I first found this from a user report with Anukari, and then later reproduced with the latest gltf_viewer.exe, so it is not an issue in my code.

I have reproduced this on two separate machines, a laptop and a desktop. Both machines have integrated AMD Radeon graphics.

**To Reproduce**
1. Run gltf_viewer.exe on a machine with AMD Radeon graphics
2. Go to the View tab and enable screen space reflections
3. It crashes every time
 
**Expected behavior**
Doesn't crash

**Screenshots**
N/A

**Logs**
The process is terminating before any error handlers are called, so no logs, just the crash inside the amd drivers.

 - OS: Windows 11
 - GPU: AMD Radeon with latest AMD-supplied drivers
 - Backend: Vulkan

poweifeng:
Crash is due to the color output being used as the SSR history. Here we blit the color output to another texture and use the copy as the history.

Fixes #9680

### DIFF
--- a/filament/src/details/Renderer.cpp
+++ b/filament/src/details/Renderer.cpp
@@ -104,6 +104,8 @@ FRenderer::FRenderer(FEngine& engine) :
         mIsFrameBufferFetchSupported(false),
         mIsFrameBufferFetchMultiSampleSupported(false),
         mIsAutoDepthResolveSupported(false),
+        mWorkaroundBlitSsrHistory(
+                engine.getDriverApi().isWorkaroundNeeded(Workaround::BLIT_SSR_HISTORY)),
         mUserEpoch(engine.getEngineEpoch()),
         mResourceAllocator(std::make_unique<TextureCache>(
                 engine.getSharedResourceAllocatorDisposer(),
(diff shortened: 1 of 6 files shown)


filament/src/details/Renderer.cpp
```diff
@@ -1337,21 +1339,38 @@ void FRenderer::renderJob(DriverApi& driver, RootArenaScope& rootArenaScope, FVi
...
+
+        if (mWorkaroundBlitSsrHistory) {
+            history = ppm.blit(fg, false, history,
+                    { 0, 0, colorBufferDesc.width, colorBufferDesc.height },
+                    {
+                        .width = colorBufferDesc.width,
+                        .height = colorBufferDesc.height,
+                        .format = config.hdrFormat,
+                    },
+                    SamplerMagFilter::LINEAR, SamplerMinFilter::LINEAR);
```
pixelflinger, line 1355: you should be using NEAREST here to get an exact copy

filament/src/details/Renderer.cpp
```diff
@@ -1337,21 +1339,38 @@ void FRenderer::renderJob(DriverApi& driver, RootArenaScope& rootArenaScope, FVi
...
                 [&](FrameGraph::Builder& builder, auto& data) {
                     // We need to use sideEffect here to ensure this pass won't be culled.
                     // The "output" of this pass is going to be used during the next frame as
                     // an "import".
                     builder.sideEffect();
-
-                    // we can't use colorPassOutput here because it could be tonemapped
-                    data.history = builder.sample(colorPassOutput.linearColor); // FIXME: an access must be declared for detach(), why?
-                }, [&view, projection](FrameGraphResources const& resources, auto const& data) {
+                    data.history = builder.sample(history);
```
pixelflinger, line 1366: You could simplify (a bit) this code by letting the pass culling do its job. i.e. you would unconditionally do the blit pass. And here you'd do:

`data.history = builder.sample(workaround ? copy : original);`

You shouldn't need to change the name of the ExportSSRHistory pass because that is already captured into the name of the actual resource that will be used.

pixelflinger (changes requested):

pixelflinger:
I feel like the root cause of the problem is not well understood. And the driver detection looks flaky?

poweifeng:
An actual fix in #9771 9771

PR #9764, opened 2026-03-02
Add Sonatype publishing step to release workflow

bejado:


bejado:
> Sometime a release goes bad and we have to "delete" it. How will that flow affect the sonatype publishing? Will the new release just override the old?

Good question. So when a release gets "published" to Sonatype, it still has to be manually released by me in the Sonatype interface. So, if a release goes bad, I can delete the bad release and we can rebuild. Then when we're confident in the release, I'll do the final "publish" step in Sonatype. In the future maybe this could be automated too.

PR #9767, opened 2026-03-03
backend: Platform::getDeviceInfo for querying device and driver info

poweifeng:
This adds a pure virtual getDeviceInfo() method to the Platform interface to allow querying graphics API specific information like renderer, vendor, device name, and driver details.

- OpenGLPlatform: returns renderer and vendor strings.
- VulkanPlatform: returns device, driver name, and driver info (cached in VulkanContext).
- Metal/WebGPU/Noop: return empty strings.
- Added unit tests and death tests for mismatched info types.

### DIFF
--- a/filament/backend/include/backend/Platform.h
+++ b/filament/backend/include/backend/Platform.h
@@ -217,6 +217,17 @@ public:
     };
 
     /**
+     * Types of device/driver information that can be queried from the platform.
+     */
+    enum class DeviceInfoType {
+        OPENGL_RENDERER,    //!< glGetString(GL_RENDERER)
+        OPENGL_VENDOR,      //!< glGetString(GL_VENDOR)
+        VULKAN_DEVICE_NAME, //!< VkPhysicalDeviceProperties::deviceName
+        VULKAN_DRIVER_NAME, //!< VkPhysicalDeviceDriverProperties::driverName
+        VULKAN_DRIVER_INFO, //!< VkPhysicalDeviceDriverProperties::driverInfo
+    };
+
+    /**
      * This controls the priority level for GPU work scheduling, which helps prioritize the
      * submitted GPU work and enables preemption.
      */
@@ -316,7 +327,7 @@ public:
          */
         StereoscopicType stereoscopicType = StereoscopicType::NONE;
 
-        /*
+        /**
          * The number of eyes to render when stereoscopic rendering is enabled. Supported values are
          * between 1 and Engine::getMaxStereoscopicEyes() (inclusive).
          */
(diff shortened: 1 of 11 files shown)


filament/backend/src/vulkan/platform/VulkanPlatform.cpp
```diff
@@ -704,6 +704,30 @@ VulkanPlatform::VulkanPlatform() = default;
...
+        case DeviceInfoType::VULKAN_DRIVER_NAME: {
+            return context.isDriverPropertiesSupported() ? utils::CString(context.getDriverName())
+                                                         : utils::CString();
+        }
+        case DeviceInfoType::VULKAN_DRIVER_INFO: {
+            return context.isDriverPropertiesSupported() ? utils::CString(context.getDriverInfo())
+                                                         : utils::CString();
+        }
+        default:
+            FILAMENT_CHECK_POSTCONDITION(false) << "Unsupported DeviceInfoType for VulkanPlatform";
```
bejado: Should this be a `PRECONDITION` because it represents an error on the caller's behalf?

filament/backend/include/backend/Platform.h
```diff
@@ -377,6 +388,16 @@ class UTILS_PUBLIC Platform {
      */
     virtual int getOSVersion() const noexcept = 0;
 
+    /**
+     * Queries device/driver information of the graphics API.
```
bejado, line 392: Maybe add that the returned string may be empty in the case that the driver doesn't support retrieving properties (i.e., some Vulkan drivers).

filament/backend/src/vulkan/platform/VulkanPlatform.cpp
poweifeng: done

filament/backend/include/backend/Platform.h
poweifeng: done

filament/backend/include/backend/platforms/OpenGLPlatform.h
```diff
@@ -52,6 +52,9 @@ class OpenGLPlatform : public Platform {
 
     ~OpenGLPlatform() noexcept override;
 
+    utils::CString getDeviceInfo(DeviceInfoType infoType,
+            Driver* UTILS_NULLABLE driver) const noexcept override;
```
pixelflinger: why NULLABLE here?

filament/backend/include/backend/Platform.h
```diff
@@ -377,6 +388,18 @@ class UTILS_PUBLIC Platform {
...
+     * @param infoType the type of information to query.
+     * @param driver a pointer to the current driver.
+     * @return a CString containing the requested information. Can return empty string in cases
+     *         where such query is not supported (for example on Vulkan without
+     *         vkGetPhysicalDeviceProperties2).
+     */
+    virtual utils::CString getDeviceInfo(DeviceInfoType infoType,
+            Driver* UTILS_NULLABLE driver) const noexcept = 0;
```
pixelflinger, line 398: if driver is not required for all platform, it should be documented here. or, we should make it UTILS_NONNULL.

pixelflinger (changes requested):

filament/backend/include/backend/Platform.h
poweifeng: Makes sense.  I've updated the declaration and test to use UTILS_NONNULL instead

PR #9769, opened 2026-03-03
ppm: move stencil check right before driver calls

poweifeng:
The init block of addPass will be executed
immediately, which means that even for passes
that are culled, we will trip on incorrect format
given to the pass.

We move the stencil check to the execute block
of addPass so that it'll only assert when the
pass is actually present in the graph.

FIXES=489437881

### DIFF
--- a/filament/src/PostProcessManager.cpp
+++ b/filament/src/PostProcessManager.cpp
@@ -3672,8 +3672,12 @@ FrameGraphId<FrameGraphTexture> PostProcessManager::resolve(FrameGraph& fg,
     }
 
     // The Metal / Vulkan backends currently don't support depth/stencil resolve.
-    if (isDepthFormat(inDesc.format) && (!mDepthStencilResolveSupported)) {
-        return resolveDepth(fg, outputBufferName, input, outDesc);
+    // TODO: Stencil resolve is actually *not* supported. Trying to resolve a stencil texture will
+    //     trigger an assert on debug builds. We need to investigate how this can be accomplished
+    //     through shaders or some other manipulation.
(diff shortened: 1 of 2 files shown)


filament/src/PostProcessManager.cpp
```diff
@@ -3696,14 +3696,14 @@ FrameGraphId<FrameGraphTexture> PostProcessManager::resolve(FrameGraph& fg,
...
-                assert_invariant(!isStencilFormat(inDesc.format));
-
                 data.input = builder.read(input, FrameGraphTexture::Usage::BLIT_SRC);
                 data.output = builder.createTexture(outputBufferName, outDesc);
                 data.output = builder.write(data.output, FrameGraphTexture::Usage::BLIT_DST);
             },
-            [](FrameGraphResources const& resources, auto const& data, DriverApi& driver) {
+            [=](FrameGraphResources const& resources, auto const& data, DriverApi& driver) {
+                // we currently don't support stencil resolve.
+                assert_invariant(!isDepthFormat(inDesc.format) || !isStencilFormat(inDesc.format));
```
pixelflinger: why does this also check for depth format?

poweifeng: correct me if wrong, but I think `resolveDepth` is only meant for metal/vk, and `resolve` is actually used for resolving depth for gl?  Hence this is saying, either we're just resolving depth (and no stencil), or we're not resolving depth at all.

poweifeng: removed assert and improved conditional for the `resolveDepth` path

PR #9770, opened 2026-03-03
Add workaround for GLES2 mipmap requirements

PatRibas:
The requirements for mipmapped texture sampling in GLES2 are weirdly stringent, and will render awful black artifacts. We can work around them by ignoring user sampler settings when they are invalid.

### DIFF
--- a/filament/backend/src/opengl/platforms/PlatformEGL.cpp
+++ b/filament/backend/src/opengl/platforms/PlatformEGL.cpp
@@ -706,10 +706,11 @@ bool PlatformEGL::setExternalImage(ExternalImageHandleRef externalImage,
 // -----------------------------------------------------------------------------------------------
 
 void PlatformEGL::initializeGlExtensions() noexcept {
-    // We're guaranteed to be on an ES platform, since we're using EGL
     const char* const extensions = (const char*)glGetString(GL_EXTENSIONS);
-    GLUtils::unordered_string_set const glExtensions = GLUtils::split(extensions);
-    ext.gl.OES_EGL_image_external_essl3 = glExtensions.has("GL_OES_EGL_image_external_essl3");
+    if (extensions) {
+        GLUtils::unordered_string_set const glExtensions = GLUtils::split(extensions);
+        ext.gl.OES_EGL_image_external_essl3 = glExtensions.has("GL_OES_EGL_image_external_essl3");
+    }
 }
 
 EGLContext PlatformEGL::getContextForType(ContextType const type) const noexcept {


filament/backend/src/opengl/GLDescriptorSet.cpp
```diff
@@ -341,7 +341,43 @@ void GLDescriptorSet::bind(
...
+                    // "Suppose that a texture is accessed from a fragment shader or vertex shader and has set GL_TEXTURE_MIN_FILTER
+                    // to one of the functions that requires mipmaps. If either the dimensions of the texture images currently
+                    // defined (with previous calls to glTexImage2D, glCompressedTexImage2D, or glCopyTexImage2D) do not follow
+                    // the proper sequence for mipmaps (described above), or there are fewer texture images defined than are needed,
+                    // or the set of texture images were defined with different formats or types, then the texture image unit
+                    // will return (R, G, B, A) = (0, 0, 0, 1)."
+                    //
+                    // So we force the sampler to a valid state in those cases.
+                    auto isPowerOfTwo = [](uint32_t x) -> bool { return (x & (x - 1)) == 0 && x > 0; };
+                    bool textureIsNonPowerOfTwo = isPowerOfTwo(t->width) && isPowerOfTwo(t->height);
```
poweifeng: did you mean to say `textureIsPowerOfTwo`?

filament/backend/src/opengl/GLDescriptorSet.cpp
```diff
@@ -341,7 +341,43 @@ void GLDescriptorSet::bind(
...
+
+                    uint32_t requiredMipLevels = (std::floor(std::log2(std::max(t->width, std::max(t->height, t->depth)))) + 1);
+                    bool textureHasRequiredMipLevels = t->levels == requiredMipLevels;
+
+                    bool samplerCanBeInvalid = params.filterMin > SamplerMinFilter::LINEAR ||
+                        params.wrapS != SamplerWrapMode::CLAMP_TO_EDGE || params.wrapT != SamplerWrapMode::CLAMP_TO_EDGE;
+
+                    if (samplerCanBeInvalid && (textureIsNonPowerOfTwo || !textureHasRequiredMipLevels)) {
+                        if (params.filterMin > SamplerMinFilter::LINEAR) {
+                            params.filterMin = SamplerMinFilter::NEAREST;
```
pixelflinger, line 375: Why go to NEAREST here? LINEAR should be fine.

filament/backend/src/opengl/GLDescriptorSet.cpp
```diff
@@ -341,7 +341,43 @@ void GLDescriptorSet::bind(
...
+                        }
+                        params.wrapS = SamplerWrapMode::CLAMP_TO_EDGE;
+                        params.wrapT = SamplerWrapMode::CLAMP_TO_EDGE;
```
pixelflinger: This is unfortunately a bit dangerous. The material could be relying on REPEAT. Not sure what to do here. My point is that it could be as broken as without.

filament/backend/src/opengl/GLDescriptorSet.cpp
```diff
@@ -341,7 +341,43 @@ void GLDescriptorSet::bind(
                 if (arg.handle) {
                     GLTexture const* const t = handleAllocator.handle_cast<GLTexture*>(arg.handle);
                     gl.bindTexture(unit, t->gl.target, t->gl.id, t->gl.external);
-                    SamplerParams const params = arg.params;
+                    SamplerParams params = arg.params;
```
pixelflinger, line 344: This whole block should be protected by checking for the NPOT extension. On most android devices NPOT will be present and we don't have to do any of this.

filament/backend/src/opengl/GLDescriptorSet.cpp
PatRibas: Whoops

filament/backend/src/opengl/GLDescriptorSet.cpp
PatRibas: We can actually be a bit more specific here, and use `NEAREST` or `LINEAR` if the mipmap level that would get sampled would've used one or the other.

filament/backend/src/opengl/GLDescriptorSet.cpp
PatRibas: Rereading the docs, it is (emphasis mine)

> **either** the GL_TEXTURE_MIN_FILTER is set to one of the functions that requires mipmaps **or** the GL_TEXTURE_WRAP_S or GL_TEXTURE_WRAP_T is not set to GL_CLAMP_TO_EDGE

so for our purposes, I think changing the `filterMin` is enough, and that does seem to work. We can leave the clamp settings alone.

filament/backend/src/opengl/GLDescriptorSet.cpp
PatRibas: Some cursory searching seems to suggest that we can expect Android devices above GLES 1.0 to support this. I am willing to stick this in an `ifdef EMSCRIPTEN` block to be even more strict.

PR #9774, opened 2026-03-05
engine: override spec constants on instance-basis

elizagamedev:
This commit adds `getConstant()` and `setConstant()` methods to `MaterialInstance`.

When a material instance is created, it doesn't create or manage any spec constants on its own, but instead relies on `Material`. The client can change constants at any time, after which the necessary resources are copied and maintained by `MaterialInstance`.

As part of this change, most internal rendering methods which depended on `Material` have been refactored to instead depend on `MaterialInstance`.


filament/src/details/MaterialInstance.cpp
```diff
@@ -398,6 +446,53 @@ const char* FMaterialInstance::getName() const noexcept {
...
+        struct Callback {
+            Invocable<void(Material*)> f;
+            Material* m;
+            static void func(void* user) {
+                auto* const c = static_cast<Callback*>(user);
+                c->f(c->m);
+                delete c;
+            }
+        };
+        // TODO(exv): fix this const cast
```
z3moon: I think we need to resolve this?

filament/include/filament/MaterialInstance.h
```diff
@@ -94,6 +94,12 @@ class UTILS_PUBLIC MaterialInstance : public FilamentAPI {
             std::is_same_v<math::mat3f, T>
     >;
 
+    template<typename T>
+    using is_supported_constant_parameter_t = std::enable_if_t<
```
poweifeng, line 98: Can we re-use the one in Material.h?

filament/src/details/MaterialInstance.cpp
```diff
@@ -378,6 +417,15 @@ void FMaterialInstance::setTransparencyMode(TransparencyMode const mode) noexcep
     mTransparencyMode = mode;
 }
 
+RasterState FMaterialInstance::getRasterState() const noexcept {
```
poweifeng: why this fix?  should this be a separate commit?

filament/src/details/MaterialInstance.cpp
elizagamedev: Oops, I completely forgot about this. Let me see what I can do...

filament/src/details/MaterialInstance.h
```diff
@@ -86,6 +84,34 @@ class FMaterialInstance : public MaterialInstance {
...
+
+    // prepareProgram creates the program for the material's given variant at the backend level.
+    // Must be called outside of backend render pass.
+    // Must be called before getProgram() below.
+    backend::Handle<backend::HwProgram> prepareProgram(backend::DriverApi& driver,
+            Variant const variant,
+            backend::CompilerPriorityQueue const priorityQueue) const noexcept {
+        if (UTILS_UNLIKELY(!mPendingSpecializationConstants.empty())) {
+            flushSpecializationConstants();
+        }
```
poweifeng: nit: how about have an early return in `flushSpecializationConstants()` if `mPendingSpecializationConstants.empty()`.  Then you can remove the check here.

filament/src/RenderPass.cpp
```diff
@@ -446,10 +446,6 @@ void RenderPass::setupColorCommand(Command& cmdDraw, Variant variant,
             BlendFunction::ZERO : cmdDraw.info.rasterState.blendFunctionDstAlpha;
 
     cmdDraw.info.rasterState.inverseFrontFaces = inverseFrontFaces;
-    cmdDraw.info.rasterState.culling = mi->getCullingMode();
-    cmdDraw.info.rasterState.colorWrite = mi->isColorWriteEnabled();
-    cmdDraw.info.rasterState.depthWrite = mi->isDepthWriteEnabled();
-    cmdDraw.info.rasterState.depthFunc = mi->getDepthFunc();
```
pixelflinger, line 452: What happened to those?

filament/src/PostProcessManager.cpp
```diff
@@ -189,7 +189,8 @@ FMaterial* PostProcessManager::PostProcessMaterial::getMaterial(FEngine& engine,
     if (UTILS_UNLIKELY(mSize)) {
         loadMaterial(engine);
     }
-    mMaterial->prepareProgram(driver, Variant{ variant }, CompilerPriorityQueue::CRITICAL);
+    mMaterial->getDefaultInstance()->prepareProgram(driver, Variant{ variant },
```
pixelflinger: can you have a comment that explains why it's okay to use the default instance? What if it's not the one the user ends-up using?

filament/src/details/Material.cpp
```diff
@@ -237,47 +235,16 @@ filament::DescriptorSetLayout const& FMaterial::getPerViewDescriptorSetLayout(
...
-                if (mDefinition.hasVariant(variant, shaderModel, isStereoSupported)) {
-                    prepareProgram(driver, variant, priority);
-                }
-            }
-        }
-    }
-
+void FMaterial::compile(CompilerPriorityQueue const priority, UserVariantFilterMask variantSpec,
+        CallbackHandler* handler, Invocable<void(Material*)>&& callback) noexcept {
+    FMaterialInstance* mi = getDefaultInstance();
```
pixelflinger, line 242: Is Material::compile() essentially deprecated? 

I feel like at least its documentation needs to change. If I understand correctly, as long as you don't set a constant on a material instance; this works like before. But if you do, this basically doesn't help you.

pixelflinger (commented):
This looks good, but scares the hell out of me :-)

It would be really hard to guard this change with a feature flag; but do we have high confidence that NOT using the new feature should essentially not change anything?

filament/src/RenderPass.cpp
elizagamedev: I moved these to a new method `MaterialInstance::getRasterState()`, which is used instead of `Material::getRasterState()` above.

I just noticed we actually make two calls to `getRasterState()`, so I'll update it to remove one of them.

filament/src/PostProcessManager.cpp
elizagamedev: You're right, I think it'd probably be best to inline the `prepareProgram()` call here at all the call sites of `getMaterial()`; will do that.

poweifeng (approved):
some nits

filament/src/PostProcessManager.cpp
elizagamedev: I had to do quite a lot of refactoring to address this properly, please review.

filament/src/details/Material.cpp
elizagamedev: Yes, that's correct. I can update the documentation.

pixelflinger (approved):

show50726:
Note: I removed the release note since it has already been included in the previous release. I add the `internal` label to this PR

PR #9776, opened 2026-03-05
webgpu: fixes for deadlock, over-flushing

poweifeng:
- In finish(), instead of block to wait for read pixels to finish, we poll the state of the counter to ensure that all the readbacks are complete.  This allows us to complete the readpixel callbacks by advancing webgpu's internal callback counter.
 - Remove all the redundant flushes that are no longer needed now that buffer upload is put on the command queue.
 - Add a convenience method for getting swapchain dimensions. Will be useful.
 - Add webgpu as a valid backend for running backend tests on linux (now that it no longer deadlocks).

### DIFF
--- a/filament/src/Froxelizer.cpp
+++ b/filament/src/Froxelizer.cpp
@@ -227,7 +227,6 @@ bool Froxelizer::prepare(
         const mat4f& projection, float const projectionNear, float const projectionFar,
         float4 const& clipTransform) noexcept {
     assert_invariant(projectionFar > projectionNear);
-    assert_invariant(projectionNear > 0);
     setViewport(viewport);
     setProjection(projection, projectionNear, projectionFar);
 


filament/backend/src/webgpu/WebGPUDriver.cpp
```diff
@@ -192,8 +192,15 @@ void WebGPUDriver::flush(int) {
...
 
 void WebGPUDriver::finish(int /* dummy */) {
-    mReadPixelMapsCounter.waitForAllToFinish();
     mQueueManager.finish();
+
+    // We use polling to advance webgpu's callback counter until all the read backs have been
+    // processed. Note that blocking with mReadPixelMapsCounter.waitForAllToFinish will only
+    // deadlock since we could not advance the counter.
+    while (!mReadPixelMapsCounter.isIdle()) {
+        mAdapter.GetInstance().ProcessEvents();
```
yrcloud, line 201: I'm a bit confused about the lifecycle logic of webgpu here. I'm assuming something is already calling the `WebGPUDriver::tick()` function every frame, which already calls `ProcessEvents`? Why do we need to explicitly call `ProcessEvents` again in `finish()` here? 
And should we call `WebGPUDriver::tick()` here instead of `ProcessEvents`?

poweifeng: `tick()` is a driver api meant for certain backends (gl, for example) to advance some kind of internal callback.  In the webgpu backend's case, it makes sense to advance dawn's internal counter to advance callbacks. But it could also be used for other things that are not in dawn (maybe, in the future, a webgpu backend specfic queue of work). 

I guess what I'm saying that `tick` is meant to be a general interface between backend and filament.  

But here, we *just* want to advance dawn's callbacks (and not do anything else). That's why I chose not to call `tick()` here.

PR #9779, opened 2026-03-06
Allow child classes to configure the EGLDisplay.

haroonq:


filament/backend/include/backend/platforms/PlatformEGL.h
```diff
@@ -159,6 +159,8 @@ class PlatformEGL : public OpenGLPlatform {
     EGLConfig getEglConfig() const noexcept { return mEGLConfig; }
     EGLConfig getSuitableConfigForSwapChain(uint64_t flags, bool window, bool pbuffer) const;
 
+    void setEglDisplay(EGLDisplay display) noexcept { mEGLDisplay = display; }
```
pixelflinger: That's very dangerous method that must be well documented.

For instance:
- when can it be called?
- probably should be illegal if mEGLDisplay is not EGL_NO_DISPLAY (return an error)
- probably illegal to pass EGL_NO_DISPLAY
- specify if it takes ownership (who destroys it?)
- can it be called more than once with different displays? (probably not)
etc...

This is a public API it must be really well defined.

haroonq: It is a protected API so can only be callable by derived classes.  That said, I've added comments about when it should and should not be called.

poweifeng: I would also add a `FILAMENT_CHECK_PRECONDITION` to check that `mEGLDisplay == EGL_NO_DISPLAY` when this is called. And maybe putting in the cpp.

haroonq: Done.

filament/backend/src/opengl/platforms/PlatformEGL.cpp
```diff
@@ -119,31 +120,39 @@ bool PlatformEGL::isOpenGL() const noexcept {
 
 PlatformEGL::ExternalImageEGL::~ExternalImageEGL() = default;
 
+void PlatformEGL::setEglDisplay(EGLDisplay display) noexcept {
+    FILAMENT_CHECK_PRECONDITION(mEGLDisplay == EGL_NO_DISPLAY)
```
pixelflinger, line 124: you probably want to check that they don't pass EGL_NO_DISPLAY? Either we need to specifically allow it, or check for it.

haroonq: Done.

filament/backend/include/backend/platforms/PlatformEGL.h
```diff
@@ -159,6 +159,12 @@ class PlatformEGL : public OpenGLPlatform {
     EGLConfig getEglConfig() const noexcept { return mEGLConfig; }
     EGLConfig getSuitableConfigForSwapChain(uint64_t flags, bool window, bool pbuffer) const;
 
+    // Sets the EGLDisplay to be used by this platform. This should only be called by derived 
+    // classes before invoking createDriver. Calling it after that point will result in 
+    // undefined behaviour. This class will take owernship of the display and call eglTerminate
```
poweifeng:
```suggestion
    // undefined behaviour. This class will take ownership of the display and call eglTerminate
```

PR #9782, opened 2026-03-07
Debug Shadow Cascades: remove dependency on frameUniforms

pixelflinger:
The DebugShadowCasacade material was dependent on frameUniforms, instead we use material parameters. We also pass all the informations relative to cascades (it's not used yet, but will be in the future).


pixelflinger:
> wouldn't you also want to remove the fields from `UibStructs`?

oh no, this fields are needed for the regular shadowing code. This is just the debugging shader that runs as a postfx pass. It shouldn't depend on values that happen to be in the ubo for the color pass.
Basically I want to the debug shader to be self contained.

PR #9784, opened 2026-03-09
Metal: fix crash from exceeding max texture dimensions for Bloom buffer

emezeske:
I recently added a new fuzz test for my Filament-based app, which creates a window and then rapidly issues changes such as toggling View features, changing the window size, and other operations that the users of my app can do.

It quickly found a crash when Bloom is enabled and the window dimensions get resized to weird aspect ratios on a large monitor. Eventually I narrowed it down to the Bloom buffer's width exceeding the Metal max texture dimension (on my machine, 16K). Applying this patch fixes the crash.


emezeske:
Thanks, @pixelflinger !

PR #9790, opened 2026-03-10
fix: Resolve a TSan warning for OpenGLTimerQuery

z3moon:
The main thread invokes `getTimerQueryValue` synchronously, reading `tq->state`. Meanwhile, the driver thread handles `createTimerQuery` asynchronously and initializes `tq->state = std::make_shared<State>()`.

Eliminate this race by initializing the state in createTimerQueryS.

BUGS=[491522442]

### DIFF
--- a/.github/workflows/postsubmit.yml
+++ b/.github/workflows/postsubmit.yml
@@ -77,9 +77,6 @@ jobs:
         with:
           name: filament-mac
           path: out/filament-release-darwin.tgz
-      - name: Check public headers
-        run: |
-          test/check-headers/test.sh out/release/filament/include
 
   build-web:
     name: build-web
--- a/.github/workflows/presubmit.yml
+++ b/.github/workflows/presubmit.yml
@@ -64,9 +64,15 @@ jobs:
       - name: Run build script
         run: |
           cd build/mac && printf "y" | ./build.sh presubmit-with-test
-      - name: Test material parser
+      - name: Test - material parser
         run: |
           out/cmake-release/filament/test/test_material_parser
+      - name: Test - public headers
+        run: |
+          # out/cmake-release should have the artifacts ready for installation. Here we install it
+          #   to test the public headers
+          ninja -C out/cmake-release install
+          test/check-headers/test.sh out/release/filament/include
 
   build-desktop-linux:
     name: build-linux
--- a/test/check-headers/test.sh
+++ b/test/check-headers/test.sh
@@ -53,6 +53,12 @@ for f in $(find . -name '*.h'); do
 done
 popd >/dev/null
 
+# Check if the system has getopt
+HAS_SYSTEM_GETOPT=0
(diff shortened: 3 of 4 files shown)


filament/backend/src/opengl/OpenGLTimerQuery.cpp
```diff
@@ -88,8 +88,11 @@ TimerQueryFactoryInterface::~TimerQueryFactoryInterface() = default;
 // This is a backend synchronous call
 TimerQueryResult TimerQueryFactoryInterface::getTimerQueryValue(
         GLTimerQuery* tq, uint64_t* elapsedTime) noexcept {
-    if (UTILS_LIKELY(tq->state)) {
-        int64_t const elapsed = tq->state->elapsed.load(std::memory_order_relaxed);
+    // Retrieve the timer query state via `atomicState` as this method is synchronously invoked by
+    // the main thread.
+    GLTimerQuery::State* atomicState = tq->atomicState.load(std::memory_order_acquire);
```
poweifeng: I can't help but think this is also not safe.  That `State` (whose lifetime is tied to a shared_ptr) could have been destroyed at this point. 

How about we do something like another field in `GLTimerQuery` called 

```
std::atomic<int64_t> elapsedCopy {TimerQueryResult::ERROR};
```

And just return `TimerQueryResult::ERROR` if `elapsedCopy` is `TimerQueryResult::ERROR`

z3moon: Good catch! I updated it with a simpler approach.

filament/backend/src/opengl/OpenGLTimerQuery.cpp
poweifeng: I think we'll never hit this part then.  the semantic has changed - is it safe?

pixelflinger: It will return `NOT_READY` (0) which I think is actually better than before.

pixelflinger: I think that's safe by accident. Technically we need a release after we write `state` and an acquire before we read it here.
Maybe for safety we can add a `std::atomic_thread_fence(release/acquire)` ?

z3moon: Our command stream pipeline already guarantees the synchronization using std::mutex. E.g., CommandBufferQueue::flush() is called from the main thread and CommandBufferQueue::waitForCommands() wakes up in the backend thread. Both are synchronized by a mutex, which implicitly guarantees that all memory writes (including the shared_ptr allocation) are globally visible.

z3moon: yeah actually I can make it simpler. updated.

pixelflinger: Oh getTimerQueryValue is only called from the main thread, so is createTimerQueryS; so I think we're fine actually.

pixelflinger (approved):

PR #9792, opened 2026-03-11
fix: Resolve a TSan warning for ShaderCompilerService

z3moon:
When the backend thread checks the availability of the token, it directly accesses `gl.program`, which caused a TSan warning.

BUGS=[491819760]


filament/backend/src/opengl/ShaderCompilerService.cpp
```diff
@@ -524,8 +531,14 @@ GLuint ShaderCompilerService::initialize(program_token_t& token) {
 }
 
 void ShaderCompilerService::ensureTokenIsReady(program_token_t const& token) {
-    if (token->gl.program) {
-        return;// It's ready.
+    if (mMode == Mode::THREAD_POOL) {
+        if (token->isReady()) {
+            return;
```
pixelflinger, line 537: shouldn't we also check or assert that `token->gl.program` is not zero?

filament/backend/src/opengl/ShaderCompilerService.cpp
```diff
@@ -935,6 +948,7 @@ void ShaderCompilerService::cancelPendingSynchronousProgram(program_token_t cons
         return false;
     }
     token->retrievedFromBlobCache = true;
+    token->signal();
```
pixelflinger: mayeb add a comment that this acts as a "release" for `token->gl.program`

filament/backend/src/opengl/ShaderCompilerService.cpp
```diff
@@ -524,8 +531,14 @@ GLuint ShaderCompilerService::initialize(program_token_t& token) {
 }
 
 void ShaderCompilerService::ensureTokenIsReady(program_token_t const& token) {
-    if (token->gl.program) {
-        return;// It's ready.
+    if (mMode == Mode::THREAD_POOL) {
+        if (token->isReady()) {
```
pixelflinger, line 536: maybe add a comment that this acts as "acquire" for `token->gl.program`

poweifeng (approved):

filament/backend/src/opengl/ShaderCompilerService.cpp
z3moon: `signal()` is invoked after `token->gl.program` is populated. So it is guaranteed to be initialized in that case. Referencing `token->gl.program` directly was the cause for the TSan warning.

filament/backend/src/opengl/ShaderCompilerService.cpp
z3moon: done

filament/backend/src/opengl/ShaderCompilerService.cpp
z3moon: done

PR #9796, opened 2026-03-13
utils: refactor getopt into utils namespace

### LINKED ISSUE #7551: Bus Error 10 (crash) caused by conflicting symbols in third_party/getopt (With Patch)

**Describe the bug**
I'm on macOS. Xcode ships with getopt. the tools link against the getopt in third_party and (presumably) indirectly against the system provided getopt. everything builds fine, but when I run for example glslminifier it crashes with a Bus Error 10 in line 44 in getopt.c (optopt = c;). After I removed the global "optarg" and "optopt" variables from getopt.c, it still compiled and linked, but the crash was gone.

**To Reproduce**
1. clone and build filament.

Note: I did not have this problem on my M1 macbook, only on my Intel iMac. So there seems to be some other factor involved here. 

**Expected behavior**
The freshly compiled tools don't crash with a Bus Error 10 when invoked by successive build steps.

**Screenshots**
N/A

**Logs**
N/A

 - OS: macOS 13.2.1, Xcode 13.4.1 & Xcode 14.3.1
 - GPU: not Relevant
 - Backend: not Relevant

**Additional context**
Two possible solutions:
1. link against the system getopt when it is available. This involves changing the includes to ```#include <getopt.h>```, searching for the library, etc.
2. avoid the clash by moving all the symbols of the third_party getopt implementation into a separate namespace. It's a bit extreme as it involves dropping plain c support, but it was easy to do and considering that getopt here is used as a utility library that probably won't be touched again, I'm willing to accept this. Patch attached.

[filament-50f605e-fixes Bus Error 10 caused by symbols clashing with system provided getopt..patch](https://github.com/google/filament/files/14156766/filament-50f605e-fixes.Bus.Error.10.caused.by.symbols.clashing.with.system.provided.getopt.patch)

poweifeng:
On certain linux, macOS environment, there is already a system getopt. This often creates conflict when compiling filament. Here we alias utils::getopt to either the system getopt (if present) or third_party/getopt.

Fixes #7551

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -326,11 +326,13 @@ endif()
 if (FILAMENT_ENABLE_LTO)
     include(CheckIPOSupported)
 
-    check_ipo_supported(RESULT IPO_SUPPORT)
+    check_ipo_supported(RESULT IPO_SUPPORT OUTPUT IPO_ERROR)
 
     if (IPO_SUPPORT)
-        message(STATUS "LTO support is enabled")
+        message(STATUS "IPO / LTO is enabled")
         set(CMAKE_INTERPROCEDURAL_OPTIMIZATION TRUE)
+    else()
+        message(WARNING "IPO / LTO is not supported by this architecture: ${IPO_ERROR}")
     endif()
 endif()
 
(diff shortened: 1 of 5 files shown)


poweifeng:
> so there was no way to just always use our getopt and never rely on the system one?

I think the problem is that SDL has calls to certain system libraries, and they have dependencies on getopt.

PR #9801, opened 2026-03-16
Fix crash in OpenGLDriver::destroyStream

z3moon:
Fixes a SIGSEGV crash that occurs during
`ExternalStreamManagerAndroid::release()` when users provide a custom Platform instance during Engine creation.

When a custom Platform is used, ExternalStreamManagerAndroid is instantiated on the application's thread. Its constructor was caching a C++ reference (`VirtualMachineEnv& mVm`) to the `thread_local` VirtualMachineEnv instance belonging to the application thread.

Later, when the Filament backend thread executed `release()`, it accessed this cached reference. This caused the backend thread to retrieve the `JNIEnv*` from the application thread's `thread_local` instance, and subsequently call `env->DeleteGlobalRef()`.  A thread must use its own `JNIEnv` pointer, so using the application thread's pointer from the backend thread resulted in a crash.

BUGS=[489814416]


z3moon:
> there was a similar fix for PerformanceHIntManager here: [cff9585](https://github.com/google/filament/commit/cff958587de263823cd9785de2fa722e6fd2d924)
> 
> Can we do something similar here? (Like initialize ExternalStreamManagerAndroid on the backend thread).

In the current Filament design, a stream is created in the application thread via `createStreamNative` or `createStreamAcquired`, and they're destroyed in the backend thread via `destroyStream`. I think this is because the application thread (Java) should directly increase the global reference of the given surface texture instance. (it wouldn't be valid in the backend thread). So we should allow both threads to have its own JNIEnv.

filament/backend/src/opengl/platforms/ExternalStreamManagerAndroid.cpp
```diff
@@ -67,7 +66,7 @@ ExternalStreamManagerAndroid::~ExternalStreamManagerAndroid() noexcept = default
 
 UTILS_NOINLINE
 JNIEnv* ExternalStreamManagerAndroid::getEnvironmentSlow() noexcept {
-    JNIEnv* const env = mVm.getEnvironment();
+    JNIEnv* const env = VirtualMachineEnv::get().getEnvironment();
```
pixelflinger: `VirtualMachineEnv::get()` is documented as "must be called on the backend thread"
This is also the case for `VirtualMachineEnv::getEnvironment()`.

I think what should happen instead is that `ExternalStreamManager` should always be created from the backend thread (this broke,  guess, when we allowed custom Platforms).

And, we should use `VirtualMachineEnv::getThreadEnvironment()` to get a JNIEnv from any thread.

z3moon: Yeah I moved the creation to the backend side.

PR #9802, opened 2026-03-17
remove the remaining instances of std::string

pixelflinger:
libfilament (including libutils and libmath) are 100% std::string free.

std::string is pulled in the .so (on android) through libc++ for  exception handling, even if we're not using them. There is not much we can do here, but at least, it's not because of us!

utils::ostream still references it but only as an inline function, so if the inline is not called, std::string won't be pulled in.

It's also referenced from Path.cpp, but that's not included in libfilament.

### DIFF
--- a/libs/filament-matp/src/ParametersProcessor.cpp
+++ b/libs/filament-matp/src/ParametersProcessor.cpp
@@ -39,27 +39,24 @@ using namespace utils;
 namespace matp {
 
 template <class T>
-static utils::Status logEnumIssue(const std::string& key, const JsonishString& value,
-        const std::unordered_map<std::string, T>& map) noexcept {
-    utils::io::sstream errorMessage;
-    errorMessage << "Error while processing key '" << key << "' value." << utils::io::endl;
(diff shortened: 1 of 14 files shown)


libs/filament-matp/src/ParametersProcessor.cpp
```diff
@@ -39,27 +39,24 @@ using namespace utils;
...
-static bool isStringValidEnum(const std::unordered_map<std::string, T>& map,
-        const std::string& s) noexcept {
+static bool isStringValidEnum(const std::unordered_map<std::string_view, T>& map, const std::string_view& s) noexcept {
     return map.find(s) != map.end();
 }
 
 template <class T>
-static T stringToEnum(const std::unordered_map<std::string, T>& map,
-        const std::string& s) noexcept {
+static T stringToEnum(const std::unordered_map<std::string_view, T>& map, const std::string& s) noexcept {
```
poweifeng: still an `std::string` here?

pixelflinger: good catch! Eagle eyes!

PR #9807, opened 2026-03-17
Silently disable MSAA for FL0.

PatRibas:
This avoids a bug where the frame graph would insert a post-process pass and skip sRGB emulation.


PatRibas:
> I can't see where `mFeatureLevel` is initialized with the real value.

Whoops... Done.

z3moon:
LGTM. it'd be nice to add a comment clarifying why MSAA is disabled for FL0 to be helpful for future reference.

PatRibas:
> LGTM. it'd be nice to add a comment clarifying why MSAA is disabled for FL0 to be helpful for future reference.

Done.

PR #9808, opened 2026-03-18
vk: make fbo eviction time configurable

### LINKED ISSUE #9786: VK: massive vram usage from MSAA + SSR
**Describe the bug**
My Filament app can easily run out of VRAM on my NVIDIA 5070ti card with 16 GB of ram, even with an extremely simple scene, with certain view settings.

**To Reproduce**
1. Enable MSAA 4x
2. Enable SSR with default settings
3. Make the window fullscreen on a 4k monitor
4. Load a simple scene

Look at VRAM usage. It easily goes up to 10-14 GB for me with only these options enabled.

I am testing against the current Filament main HEAD.

**Expected behavior**
I would expect a few hundred MB of VRAM used for these options with such a large window, not 10+ GB. On the Metal backend, with the exact same application code, the RAM usage only goes up by ~400 MB when I enable MSAA 4x and SSR together, which seems more reasonable.

It seems like there's some kind of multiplicative thing going on?

**Screenshots**
In the following diagram, I annotated the VRAM usage on my machine:

1. Enable SSR alone
2. Enable MSAA 4x alone
3. Enable SSR + MSAA 4x together

Image

I don't think it's a leak, because when I later disable MSAA or SSR, the memory usage goes down again to normal levels. Also it is stable at the (extremely high) level, it doesn't keep growing or anything.

**Logs**
```
FEngine (64 bits) created at 0000021F892317E0 (threading is enabled)
Vulkan device driver: NVIDIA 591.74
Selected physical device 'NVIDIA GeForce RTX 5070 Ti' from 2 physical devices. (vendor 0x10de, device 0x2c05, driver 0x93d28000, api 1.4)
Backend feature level: 3
FEngine feature level: 1
vkCreateSwapchain: 3340x1885, 44, 0, swapchain-size=3, identity-transform=true, depth=126, protected=false
```

 - OS: Windows 11
 - GPU: NVIDIA RTX 5070 ti
 - Backend: Vulkan

poweifeng:
The previous eviction time is too large (40), then causes memory to bloat over time. We set the default to 3 (as in triple buffering), but make it a configurable option via VulkanPlatform.

Fixes #9786


rafadevai:
Do you know why are so many FB being created? and any reason why they arent being reused? My assumption from 1000mts is that there should just be a couple of them

poweifeng:
> Do you know why are so many FB being created? and any reason why they arent being reused? My assumption from 1000mts is that there should just be a couple of them

I can't say for the case out in the wild, but with respect to gltf_viewer on Android running with a Pixel 4.  The problem is with dynamic resolution and debug mode.  The combination of the debug and dynres caused ping-ponging of the frame sizes (full res -> half res -> full -> etc). This creates and destroys the intermediate render buffers, and so we get new image views for the rendertargets frequently. And because this cache is keyed on VkImageViews, we fill up the cache quickly.

Filling up the cache is not too bad because VkFrameBuffer doesn't consume that much resource, but because we also keep references of the rendertarget (and thereby the backing textures) with each VulkanFrameBuffer, those won't get released while the buffer is in the cache. So we end up keeping an excessive amount of gpu memory.

rafadevai:
Ideally would be great that if the rendertarget are destroyed this would remove them from the cache as is impossible to retrieve them again.

rafadevai (approved):

filament/backend/src/vulkan/VulkanFboCache.cpp
```diff
@@ -376,31 +378,36 @@ void VulkanFboCache::gc() noexcept {
     FVK_SYSTRACE_START("fbocache::gc");
 
     // If this is one of the first few frames, return early to avoid wrapping unsigned integers.
```
z3moon: this comment looks like a leftover.

PR #9809, opened 2026-03-18
engine: panic if prepareProgram() fails

elizagamedev:
This change crashes Filament with a helpful error message if we ever try to prepare a program and fail.


elizagamedev:
Context: gltf viewer crashes with both VSM and Froxel Visualization enabled.

```
reason: Requested variant 80 does not exist for material transparentColor
```

PR #9812, opened 2026-03-18
android: [render-validation] add more test result details

poweifeng:
Add the following information:
 - Android build fingerprint, version
 - GPU driver name, info, vendor name
 - Time elapsed for test
 - Rendered images (as oppose to diff image)

filament-utils:
 - Add DeviceUtils to hook into Platform methods for reading out strings about gpu vendor, driver.

Fix "tolerance" in test definition

### DIFF
--- a/.github/actions/renderdiff-generate/action.yml
+++ b/.github/actions/renderdiff-generate/action.yml
@@ -8,9 +8,15 @@ runs:
     - uses: ./.github/actions/get-mesa
     - uses: ./.github/actions/get-vulkan-sdk
     - name: Prerequisites
+      if: runner.os == 'macOS'
       run: |
         # Must have at least clang-16 for a webgpu/dawn build.
         sudo xcode-select -s /Applications/Xcode_16.2.app/Contents/Developer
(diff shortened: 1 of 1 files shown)


android/filament-utils-android/src/main/cpp/DeviceUtils.cpp
```diff
@@ -0,0 +1,77 @@
...
+
+                if (!deviceName.empty()) fullInfo += deviceName.c_str();
+                if (!driverName.empty()) {
+                    if (!fullInfo.empty()) fullInfo += " | ";
+                    fullInfo += driverName.c_str();
+                }
+                if (!driverInfo.empty()) {
+                    if (!fullInfo.empty()) fullInfo += " | ";
+                    fullInfo += driverInfo.c_str();
+                }
```
bejado: This bit is repeated down below. Maybe refactor to a helper function?

poweifeng: good point!  I simplified this.

PR #9814, opened 2026-03-19
Optimize Shader LineDictionary

pixelflinger:
Many improvements to the `LineDictionary` yields to significant material size reduction, for e.g.:

```
 filament-android.aar:   -9.8 KB
 gltfio-android.aar:   -293.3 KB
 libfilament.a:         -41.3 KB
 volume_masked.mat:     -30.1 KB
 ```

1. triple-stream, variable length indces
2. shader-stage aware dictionary
3. leading underscore removal for spirv generated ids
4. remove superfluous spaces in expressions

### DIFF
--- a/libs/filaflat/src/MaterialChunk.cpp
+++ b/libs/filaflat/src/MaterialChunk.cpp
@@ -25,6 +25,7 @@
 
 #include <filament/MaterialChunkType.h>
 
+#include <utils/compiler.h>
 #include <utils/Invocable.h>
 #include <utils/debug.h>
 #include <utils/Log.h>
@@ -255,7 +256,17 @@ bool MaterialChunk::getTextShader(Unflattener unflattener,
             globalIndex += mVertexStrings + mFragmentStrings;
         }
 
+        if (UTILS_UNLIKELY(globalIndex >= dictionary.size())) {
+            return false;
+        }
+
         const auto& content = dictionary[globalIndex];
+
+        // Ensure string is correctly formed and doesn't exceed reserved shader space.
+        if (UTILS_UNLIKELY(content.size() == 0 || cursor + content.size() - 1 > shaderSize)) {
+            return false;
+        }
+
         // remove the terminating null character.
         size_t const length = content.size() - 1;
         memcpy(&shaderContent[cursor], content.data(), length);

### CODE REVIEW

**Comment by romainguy** [comment]
Might be a silly question but did we ever sort lines in the dictionary by frequency? (so they would use indices in the 0..239 range for the most frequent lines).

**Comment by pixelflinger** [comment]
> Might be a silly question but did we ever sort lines in the dictionary by frequency? (so they would use indices in the 0..239 range for the most frequent lines).

That's what this change does, yes.

File: libs/filaflat/src/MaterialChunk.cpp [hunk 1]
```diff
@@ -232,4 +272,58 @@ void MaterialChunk::visitShaders(
...
+                    break;
+                }
+                lineIndex = 240 + (((b8 - 240) << 8) | ext);
+            } else {
+                uint8_t extb0, extb1;
+                if (!extUnflattener.read(&extb0) || !extUnflattener.read(&extb1)) {
+                    break;
+                }
+                lineIndex = 4080 + (extb0 | (extb1 << 8));
+            }
```
z3moon: One suggestion. Logic in this part is the same as the line 164-178. So maybe they can be factored out as a helper function like.

```
static bool someHelperFunction(Unflattener&base, Unflattener& ext, uint32_t& outIndex) {
    uint8_t b8;
    if (!base.read(&b8)) {
        return false;
    }

    if (b8 < 240) {
        outIndex = b8;
    } else if (b8 < 255) {
    ...
    return true;
}
```

then have both sides call this function.

libs/filaflat/include/filaflat/MaterialChunk.h
```diff
@@ -70,6 +74,11 @@ class MaterialChunk {
     const uint8_t* mBase = nullptr;
     tsl::robin_map<uint32_t, uint32_t> mOffsets;
 
+    uint16_t mSharedStrings = 0;
+    uint16_t mVertexStrings = 0;
+    uint16_t mFragmentStrings = 0;
+    uint16_t mComputeStrings = 0;
+
```
poweifeng, line 81: nit: can we group them as "chunk headers" as in
```
struct Header {
}; 
```

and then everywhere we need to skip them, will just offset the cursor by `sizeof(MaterialChunk::Header)`

PR #9815, opened 2026-03-19
github: [mesa] ensure right llvm is linked on macos

poweifeng:


poweifeng:
quick fix to renderdiff

PR #9825, opened 2026-03-22
Remove -fstrict-vtable-pointers  flag for GCC

NDYDHHC:
Remove -fstrict-vtable-pointers  flag for GCC


pixelflinger:
> Remove -fstrict-vtable-pointers flag for GCC

Why this change?

NDYDHHC:
> > Remove -fstrict-vtable-pointers flag for GCC
> 
> Why this change?

GCC does not support -fstrict-vtable-pointers — it is a Clang-specific compiler option.

libs/generatePrefilterMipmap/CMakeLists.txt
```diff
@@ -49,6 +49,13 @@ if (MSVC)
 elseif(WEBGL)
     # Avoid strict-vtable-pointers here, it is broken in WebAssembly.
     set(OPTIMIZATION_FLAGS -fvisibility-inlines-hidden)
+elseif(FILAMENT_USING_GCC)
+    set(OPTIMIZATION_FLAGS
+        -ffast-math
+        -fno-finite-math-only
+        -ffp-contract=fast
+        -fvisibility-inlines-hidden
+    )
```
poweifeng: Instead of copying from the `else`? Can you instead just have a `if (!FILAMENT_USING_GCC)` block in the `else` block below.

libs/generatePrefilterMipmap/CMakeLists.txt
```diff
@@ -59,9 +59,9 @@ else()
         -fvisibility-inlines-hidden
     )
 
-        if (NOT FILAMENT_USING_GCC)
-                list(APPEND OPTIMIZATION_FLAGS -fstrict-vtable-pointers)
-        endif()
+    if (NOT FILAMENT_USING_GCC)
```
poweifeng, line 62: please also add a simple comment as to why above this line.  Same with the other file.

poweifeng: gentle ping on this comment @NDYDHHC

PR #9830, opened 2026-03-24
OpenGL: Factor per-thread state out of OpenGLContext

z3moon:
The OpenGLContext class in the OpenGL backend currently conflates two distinct parts: immutable configuration (extensions, bugs, capabilities) and a mutable GL state cache. This single-object design is incompatible with a multi-threaded rendering architecture, where each thread requires a dedicated GL context and, critically, its own independent state machine/cache.

This refactoring implements the proposed architecture by:
1. Introducing a new OpenGLState class to exclusively manage the mutable GL state cache (texture bindings, buffer bindings, etc.) with explicit per-thread ownership (e.g., mBackendState and mWorkerState).
2. Ensuring OpenGLContext retains only the shared, immutable configuration.
3. Preserving the existing gl.XXX usage pattern by implementing inline accessors in OpenGLState that forward read-only queries (e.g., gl.ext.XXX, gl.isES2()) to the back-referenced OpenGLContext.
4. Updating common functions shared between backend and worker threads to explicitly accept an OpenGLState& parameter.

This change enables a multi-threaded approach by guaranteeing each thread operates on its own dedicated GL state cache, while reusing the shared, immutable context configuration.

BUGS=[487347473]


filament/backend/src/opengl/OpenGLState.cpp
```diff
@@ -0,0 +1,559 @@
...
+    // the default FBO could be invalid
+    mDefaultFbo[index].reset();
+    contextIndex = index;
+    resetState();
+}
+
+void OpenGLState::setDefaultState() noexcept {
+    // We need to make sure our internal state matches the GL state when we start.
+    // (some of these calls may be unneeded as they might be the gl defaults)
+    GLenum const caps[] = {
```
poweifeng: can this be constexpr?

z3moon: yup, updated.

filament/backend/src/opengl/OpenGLState.cpp
```diff
@@ -0,0 +1,559 @@
...
+    }
+#endif
+    CHECK_GL_ERROR()
+    mSamplerMap[params] = s;
+    return s;
+}
+#endif
+
+
+void OpenGLState::resetState() noexcept {
```
bejado: I think this was called `resetState` before, but I wonder if the name isn't a bit misleading? To me, "resetting" state would mean setting it to the OpenGL defaults.. but here we're syncing it to match Filament's. Maybe `syncState` would be a better name?

z3moon: Yeah the name is unchanged. But Let me rename it as it's clearer.

PR #9832, opened 2026-03-24
Fix memory safety vulnerabilities in filaflat parsing components

pixelflinger:
* Fix out-of-bounds string read by verifying null-terminator existence during extraction.
* Fix heap buffer overflow by validating dictionary string lengths  against target shader buffers before copying.
* Fix out-of-bounds array reads by validating chunk-provided lookup  indices against parsed dictionary sizes.
* Fix integer wrapping exploits by replacing pointer addition with  offset subtraction during chunk size verifications.

Add unit tests for these vulerabilities.

### DIFF
--- a/libs/filaflat/include/filaflat/Unflattener.h
+++ b/libs/filaflat/include/filaflat/Unflattener.h
@@ -49,14 +49,20 @@ public:
         return mCursor < mEnd;
     }
 
-    bool willOverflow(size_t size) const noexcept {
-        return (mCursor + size) > mEnd;
+    bool willOverflow(size_t const size) const noexcept {
+        // Evaluate the remaining valid buffer size instead of using pointer arithmetic,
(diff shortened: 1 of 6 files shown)


libs/filaflat/include/filaflat/Unflattener.h
```diff
@@ -49,14 +49,20 @@ class UTILS_PUBLIC Unflattener {
         return mCursor < mEnd;
     }
 
-    bool willOverflow(size_t size) const noexcept {
-        return (mCursor + size) > mEnd;
+    bool willOverflow(size_t const size) const noexcept {
+        // Evaluate the remaining valid buffer size instead of using pointer arithmetic,
+        // preventing arbitrary integer size overflows from pointer wrapping.
+        return size > size_t(mEnd - mCursor);
```
trie94, line 55: is `mEnd` guaranteed to be equal or greater than `mCursor`?

PR #9834, opened 2026-03-25
Validate KTX1 header fields against buffer size in Ktx1Bundle

geoo115:
## Summary                                                                                                          
                                                                                                                      
  The `Ktx1Bundle` deserialization constructor trusts header fields from the input buffer without validating them
  against the actual buffer size. This can cause out-of-bounds reads when parsing malformed `.ktx` files:             
                                                                                                         
  - `bytesOfKeyValueData` is used to compute the metadata region end pointer without checking it fits within the      
  buffer                                                                                                        
  - `keyAndValueByteSize` entries within metadata are not validated against the remaining metadata region
  - `imageSize` (and the derived `levelSize`) are used directly in `memcpy()` without checking against remaining
  buffer size                                                                                                         
  - `levelSize = faceSize * mNumCubeFaces * mArrayLength` can silently overflow `uint32_t`
                                                                                                                      
  ## Changes                                                                                                          
            
  Added bounds validation using `FILAMENT_CHECK_PRECONDITION` (the same pattern already used in this constructor for  
  magic and minimum size checks):                                                                                   
                                 
  - Validate `bytesOfKeyValueData` does not exceed `nbytes - sizeof(SerializationHeader)`
  - Validate each `keyAndValueByteSize` entry fits within the metadata region                                         
  - Validate `pdata` has not advanced past the buffer after metadata parsing (possible from padding)
  - Validate `imageSize` field can be read before dereferencing                                                       
  - Validate `levelSize` does not overflow using 64-bit intermediate arithmetic
  - Validate `levelSize` does not exceed remaining buffer before `memcpy()`
                                                                                                                      
  No behavior change for valid KTX files.


libs/image/src/Ktx1Bundle.cpp
```diff
@@ -151,14 +161,30 @@ Ktx1Bundle::Ktx1Bundle(uint8_t const* bytes, uint32_t nbytes) :
...
     // Extract blobs from the serialized byte stream.
-    const uint32_t totalSize = nbytes - (pdata - bytes);
+    const uint32_t totalSize = (uint32_t)(bufEnd - pdata);
     mBlobs->blobs.resize(totalSize);
     for (uint32_t mipmap = 0; mipmap < mNumMipLevels; ++mipmap) {
+        FILAMENT_CHECK_PRECONDITION(pdata + sizeof(uint32_t) <= bufEnd)
+                << "KTX mipmap: not enough data for imageSize field";
         const uint32_t imageSize = *((uint32_t const*) pdata);
+        FILAMENT_CHECK_PRECONDITION(facesPerMip > 0)
+                << "KTX mipmap: facesPerMip must not be zero";
```
pixelflinger: this can be outside of the loop

geoo115: Done, good catch. 
moved the facesPerMip check before the loop since it's loop-invariant.
Thanks for the review!

poweifeng (approved):

pixelflinger:
will will provide our own fixes. thank you for reporting.

PR #9839, opened 2026-03-26
Fix integer overflows in KTX1Bundle deserialization

mohammadmseet-hue:
## Summary

Add validation checks to the KTX1 texture file parser to prevent memory safety issues when parsing malformed KTX1 files:

1. **Integer overflow in blob sizes**: Validate that `mNumMipLevels * mArrayLength * mNumCubeFaces` does not overflow `uint32_t` before sizing the blob list. An overflow to a small value causes out-of-bounds vector access when `flatten()` computes indices.

2. **Out-of-bounds read in key-value parsing**: Validate that `bytesOfKeyValueData` does not exceed the remaining buffer size. Without this, the key-value parsing loop reads past the end of the input buffer.

3. **Integer overflow in levelSize / heap buffer overflow**: Validate that `faceSize * mNumCubeFaces * mArrayLength` does not overflow before passing it to `memcpy`. An overflow causes copying the wrong amount of data — either too much (heap buffer overflow) or too little.

4. **Bounds checking on data pointer**: Add checks to ensure `pdata` remains within the input buffer before dereferencing, preventing out-of-bounds reads.

## Test plan

- Existing KTX1 tests continue to pass
- Malformed KTX1 files with overflow-inducing header values now fail gracefully

### DIFF
--- a/libs/image/src/Ktx1Bundle.cpp
+++ b/libs/image/src/Ktx1Bundle.cpp
@@ -124,11 +124,23 @@ Ktx1Bundle::Ktx1Bundle(uint8_t const* bytes, uint32_t nbytes) :
     mNumMipLevels = header->numberOfMipmapLevels ? header->numberOfMipmapLevels : 1;
     mArrayLength = header->numberOfArrayElements ? header->numberOfArrayElements : 1;
     mNumCubeFaces = header->numberOfFaces ? header->numberOfFaces : 1;
-    mBlobs->sizes.resize(mNumMipLevels * mArrayLength * mNumCubeFaces);
+
+    // Validate that the sizes multiplication does not overflow.
+    FILAMENT_CHECK_PRECONDITION(
+            mArrayLength == 0 || mNumCubeFaces <= UINT32_MAX / mArrayLength)
+            << "KTX: arrayLength * numCubeFaces overflow";
+    uint32_t const facesTimesLayers = mArrayLength * mNumCubeFaces;
+    FILAMENT_CHECK_PRECONDITION(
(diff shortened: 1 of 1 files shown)


libs/image/src/Ktx1Bundle.cpp
```diff
@@ -150,15 +162,32 @@ Ktx1Bundle::Ktx1Bundle(uint8_t const* bytes, uint32_t nbytes) :
...
+                mNumCubeFaces == 0 || faceSize <= UINT32_MAX / mNumCubeFaces)
+                << "KTX: faceSize * numCubeFaces overflow";
+        uint32_t const faceSizeTimesFaces = faceSize * mNumCubeFaces;
+        FILAMENT_CHECK_PRECONDITION(
+                mArrayLength == 0 || faceSizeTimesFaces <= UINT32_MAX / mArrayLength)
+                << "KTX: levelSize overflow";
+        const uint32_t levelSize = faceSizeTimesFaces * mArrayLength;
+
         pdata += sizeof(uint32_t);
+        FILAMENT_CHECK_PRECONDITION(levelSize <= totalSize)
```
z3moon: I think you shouldn't use totalSize here.

maybe something like
```
        size_t const remainingData = (bytes + nbytes) - pdata;
        FILAMENT_CHECK_PRECONDITION(levelSize <= remainingData)
                << "KTX: levelSize exceeds remaining data";
```

mohammadmseet-hue: Good catch — updated to check against `(bytes + nbytes) - pdata` so it validates the actual remaining data at each miplevel iteration, not the total from the start.

PR #9841, opened 2026-03-26
Fix integer overflow in computeBindingSize for glTF accessors

mohammadmseet-hue:
## Summary

Add overflow validation to `computeBindingSize()`, **consistent with the existing overflow check already present in `decodeMeshoptCompression()` in the same file** (lines 100-104).

The multiplication `accessor->stride * (accessor->count - 1)` can overflow when both `stride` and `count` are large values from a glTF file. The result is truncated to `uint32_t`, producing an undersized value that is used for buffer allocation in `ResourceLoader`. This can lead to undersized buffer objects when processing crafted glTF files.

Additionally, if `accessor->count` is 0, the subtraction `count - 1` underflows to `SIZE_MAX`.

### Existing pattern (same file, line 100-104):
```cpp
size_t const theoreticalMaxCount = std::numeric_limits<size_t>::max() / compression->stride;
FILAMENT_CHECK_PRECONDITION(compression->count <= theoreticalMaxCount)
        << "gltfio: meshopt decompression exceeds maximum count...";
```

This PR applies the same overflow-prevention pattern to `computeBindingSize()`.

## Test plan

- Existing glTF loading tests continue to pass
- Crafted glTF files with overflow-inducing stride/count values now fail gracefully with a precondition error

### DIFF
--- a/libs/gltfio/src/Utility.cpp
+++ b/libs/gltfio/src/Utility.cpp
@@ -168,7 +168,20 @@ bool primitiveHasVertexColor(cgltf_primitive* inPrim) {
 // private but its implementation file is available in this cpp file.
 uint32_t computeBindingSize(cgltf_accessor const* accessor) {
     cgltf_size element_size = cgltf_calc_size(accessor->type, accessor->component_type);
-    return uint32_t(accessor->stride * (accessor->count - 1) + element_size);
+    // Validate against overflow, consistent with the check in decodeMeshoptCompression above.
+    FILAMENT_CHECK_POSTCONDITION(accessor->count > 0)
+            << "gltfio: accessor count must be > 0";
+    cgltf_size const countMinus1 = accessor->count - 1;
+    if (accessor->stride > 0) {
+        cgltf_size const maxCount = std::numeric_limits<cgltf_size>::max() / accessor->stride;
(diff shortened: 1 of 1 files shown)


libs/gltfio/src/Utility.cpp
```diff
@@ -168,7 +168,20 @@ bool primitiveHasVertexColor(cgltf_primitive* inPrim) {
...
+            << "gltfio: accessor count must be > 0";
+    cgltf_size const countMinus1 = accessor->count - 1;
+    if (accessor->stride > 0) {
+        cgltf_size const maxCount = std::numeric_limits<cgltf_size>::max() / accessor->stride;
+        FILAMENT_CHECK_PRECONDITION(countMinus1 <= maxCount)
+                << "gltfio: binding size overflow (stride=" << accessor->stride
+                << ", count=" << accessor->count << ")";
+    }
+    cgltf_size const result = accessor->stride * countMinus1 + element_size;
+    FILAMENT_CHECK_PRECONDITION(result <= std::numeric_limits<uint32_t>::max())
```
pixelflinger: these should be post condition checks

pixelflinger (changes requested):

pixelflinger:
The fix was incomplete. See #9896 for a more complete fix.

PR #9842, opened 2026-03-26
Fix integer overflow in KTX2 texture transcoding allocation

mohammadmseet-hue:
## Summary

Fix two integer overflow vulnerabilities in `transcodeImageLevel()` that can result in undersized `malloc` allocations:

1. **Compressed format** (line 137): `sizeof(uint64_t) * qwordsPerBlock * m_total_blocks` — the `uint32_t` operands are multiplied before widening to `size_t`, which can overflow on 32-bit platforms (e.g., Android armv7).

2. **Uncompressed format** (line 152): `bytesPerPix * m_orig_width * rowCount` — all three operands are `uint32_t`. The first multiplication (`bytesPerPix * m_orig_width`) is evaluated as `uint32_t` arithmetic and can overflow before the result is widened to `size_t`. For example, `bytesPerPix=4` and `m_orig_width=0x40000000` produces `0x100000000` which wraps to `0` in `uint32_t`.

Both cases lead to `malloc(0)` or `malloc(small_value)`, followed by the BasisU transcoder writing the full image data into the undersized buffer, causing a heap buffer overflow.

**Fix**: Explicitly cast operands to `size_t` before multiplication and add post-multiplication overflow verification.

## Test plan

- Existing KTX2 texture loading tests continue to pass
- Crafted KTX2 files with overflow-inducing dimensions now return failure instead of triggering heap corruption

### DIFF
--- a/libs/ktxreader/src/Ktx2Reader.cpp
+++ b/libs/ktxreader/src/Ktx2Reader.cpp
@@ -134,7 +134,13 @@ static Result transcodeImageLevel(ktx2_transcoder& transcoder,
 
     if (formatInfo.isCompressed) {
         const uint32_t qwordsPerBlock = basisu::get_qwords_per_block(destFormat);
-        const size_t byteCount = sizeof(uint64_t) * qwordsPerBlock * levelInfo.m_total_blocks;
+        // Cast to size_t before multiplication to prevent uint32_t overflow.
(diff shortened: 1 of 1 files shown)


pixelflinger:
The fix was incomplete. See #9896 for a more complete fix.

PR #9847, opened 2026-03-27
docs: document 2D coordinate spaces

poweifeng:


docs_src/src_mdbook/src/notes/upload_readpixels.md
```diff
@@ -0,0 +1,52 @@
...
+The process begins with a standard image buffer in CPU memory.
+
+  - **Memory Layout:** Data is ordered from **low addresses** (top of the image) to **high addresses** (bottom of the image).
+  - **Orientation:** The image is naturally "upright."
+
+## 2\. Backend-Specific Behaviors
+
+### OpenGL (The Bottom-Left API)
+
+OpenGL is unique because its texture coordinate system origin \\((0, 0)\\) is at the **bottom-left**.
```
pixelflinger: that's not "unique". That's actually typically how we learn math in school.  Usually when you have a piece of paper, you naturally put the origin at the bottom left.

docs_src/src_mdbook/src/notes/upload_readpixels.md
```diff
@@ -0,0 +1,52 @@
...
+
+  - **Texture Upload:** To ensure texture sampling works "normally" (where $v=0$ is the top), images are typically flipped during upload. This results in the texture being stored **upside-down** in GPU memory.
```
pixelflinger: No I think this is confusing.
Images are not technically "flipped", it's just that the opengl spec defines glTexImage2D's data pointer to be ordered with the bottom of the image at the low addresses.

This *is* contrary to how images are *typically* stored on a computer.  So, yes, you have to "flip" your "Typical" bitmap if you use glTexImage2D and you want the bottom of the image to be at v=0.

We don't know how the image is stored in the GPU's memory (because in GL you can't read that buffer), so it's no correc to say that it's stored "upside down".

docs_src/src_mdbook/src/notes/upload_readpixels.md
```diff
@@ -0,0 +1,52 @@
...
+  - **Standard `glReadPixels`:** Because OpenGL reads from the bottom-left, a standard `glReadPixels` call effectively flips the data back, resulting in an **upright** image in the CPU buffer.
```
pixelflinger: Another way to say it is the glReadPIxel is consistent with glTexImage2D.
Neither operation is flipping the image you give it.
glReadPixel just transfers line 0, 1, 2... (with 0 at the bottom of the screen) in order.

Note that typically you don't do a glTexImage2D, then bind the texture as an attachment and then do a readPixels().

This exact combination of operation is *inconsistant* in Filament (because filament actually flips the result of readPixels in GL). This was a compromise we had to do, to be compatible with other APIs.

docs_src/src_mdbook/src/notes/upload_readpixels.md
```diff
@@ -0,0 +1,52 @@
...
+  - **Filament's Handling:** To maintain consistency or specific internal alignment, the Filament OpenGL backend performs an additional **CPU flip** after the read. This results in a final buffer where the image is **upside-down** relative to the original source.
```
pixelflinger: The reason for this is so that given a framebuffer (e.g. you draw a triangle into it), and you do a readpixels, the result is the *same* on all APIs. (with the "Typical" bitmap layout, low address are the top of the image).

This however "breaks" the consistency of uploading data into a texture and then reading it back with readPixel with the texture as an attachment.  This is actually the only situation in which filament doesn't work the same across all backends!

docs_src/src_mdbook/src/notes/upload_readpixels.md
```diff
@@ -0,0 +1,52 @@
...
+
+### Metal, Vulkan, & WebGPU (The Top-Left APIs)
+
+Metal, Vulkan, and WebGPU use a \\((0, 0)\\) **top-left** coordinate system, which aligns more closely with standard image memory layouts.
+
+  - **Texture Upload:** No flip is required. The image is stored in GPU memory in its **upright** orientation.
+  - **Filament `ReadPixels`:** The data is read directly. Since the GPU storage and the memory layout both favor top-left, the resulting CPU buffer is **upright**.
```
pixelflinger: yes, however all these APIs are inconsistent imho.  Because their clip-space puts the origin at the bottom left! so in these APIs the clip space is "inverted".

pixelflinger (changes requested):

docs_src/src_mdbook/src/notes/upload_readpixels.md
pixelflinger: Note that from a user's point of view, GL, VK and Metal are consistent.

i.e. if you do a glTexImage2D without inverting the texture, in all cases you end-up with the origin at the *top* of the image. No changes needed in the shader.  it's just that "conceptually* in GL the image is "upside down" (whatever that means) and the y axis goes "up" (whatever that means). Whereas is Metal/VK the image is right side up, and the y-axis goes down.

So... unintuitively, you do not have to do anything special to have your textures uploads and texture coords work on all platforms.

The problem only arises with render targets and is **compounded** when you upload data into a texture (as opposed to drawing into it) **and** use it as an attachment **and** readpixels from it.

pixelflinger: I say "whatever that means" because the definition of "up" is the direction of the Y-axis.

docs_src/src_mdbook/src/notes/upload_readpixels.md
poweifeng: removed and rephrased.

docs_src/src_mdbook/src/notes/upload_readpixels.md
poweifeng: noted and rephrased

docs_src/src_mdbook/src/notes/upload_readpixels.md
poweifeng: rephrased

docs_src/src_mdbook/src/notes/upload_readpixels.md
poweifeng: this case has been noted.

docs_src/src_mdbook/src/notes/upload_readpixels.md
poweifeng: this has also been noted.

docs_src/src_mdbook/src/notes/2d_coordinate_spaces.md
```diff
@@ -0,0 +1,221 @@
...
+**tree** is near the point of \\(((0.65, 0.35)\\))", to communicate the position of a tree in
+the picture to another person.
+
+Given the description above, we conclude this section with the following statements
+ - Programmers communicate with computers using coordinate systems like the above to draw 2D images
+   onto a display.
+ - In 3D graphics, similar coordinate systems are used to texturize 3D models with 2D images. The
+   2D images are called *textures*.
+ - Filament defines its coordinate system using the *top-left* of the image as \\((0, 0)\\) and
+   the *bottom-right* is \\((1, 1)\\).
```
pixelflinger, line 30: Is that true though? (like it's not true for textures).  Maybe give an example of a filament public API that uses that convention?

docs_src/src_mdbook/src/notes/2d_coordinate_spaces.md
```diff
@@ -0,0 +1,221 @@
...
+
+### 1.1. OpenGL (The Bottom-Left API)
+
+OpenGL has its texture coordinate origin \\((0, 0)\\) at the **bottom-left**, and
+\\((1, 1)\\) is at the top-right. Note that this differs from Filament's assumed texture
+coordinate system.
+
+The GL function for uploading texture is `glTexImage2D`. Even though GL interprets \\((0, 0)\\) as
+the bottom-left of a 2D image, `glTexImage2D` uploads the lower-bits as the *bottom* of the image
```
pixelflinger, line 39: I would say "lower-address bytes/data/pixels".

docs_src/src_mdbook/src/notes/2d_coordinate_spaces.md
```diff
@@ -0,0 +1,221 @@
...
+as opposed to the *top*. (This behavior is opposite of the other backends' texture upload
+APIs). The two differences cancel each other, meaning there is no difference between the different
+backends in upload and sampling of textures.
+
+We can restate the above as
+
+| |
+|---|
+| *Filament's internal texture coordinate space defines the **top-left** corner of the image as the origin.* |
```
pixelflinger, line 48: I see what you mean, but strangely I think of it the other way around.

docs_src/src_mdbook/src/notes/2d_coordinate_spaces.md
```diff
@@ -0,0 +1,221 @@
...
+Reading the result of rendering into CPU memory can be very useful; for example, one might want to
+take snapshots of the rendering and output to image formats such as PNG or JPEG. Another use is to
+perform image analysis or processing over a rendered image. In Filament, the readback path
+corresponds to the `readPixels`. In the OpenGL backend, this is implemented using the `glReadPixels`
+function. GL's interpretation of texture coordinate space holds true for `glReadPixels`; this means
+that it treats the bottom-left of the image as the origin and the readback output *starts* from the
+origin. Hence, the first row of the image in CPU memory is the bottom-most row of the image (when
+displayed).
+
+This creates two related inconsistencies:
```
pixelflinger, line 65: This bothers me to say this because GL is actually very consistant (it's in fact the most consistant of the APIs, imho).

docs_src/src_mdbook/src/notes/2d_coordinate_spaces.md
```diff
@@ -0,0 +1,221 @@
...
+ - First, (not a real use case), if one were to upload a texture, attach it as an attachment, and
```
pixelflinger, line 66: True, but since glTexImage2D also did that, this all works out perfectly. The image wouldn't be "upside down". You'd get exactly the same bytes that you used for glTexImage2D.

docs_src/src_mdbook/src/notes/2d_coordinate_spaces.md
```diff
@@ -0,0 +1,221 @@
...
+   read the attachment, the GL backend would place the bottom of the image as the first row of
+   pixels in CPU memory.
+ - Second, the other backends consider the top-left corner as the origin, which means their
+   readbacks would place the top of the image as the first row of the image in CPU memory.
```
pixelflinger, line 70: This also works out because their upload is consistant.

docs_src/src_mdbook/src/notes/2d_coordinate_spaces.md
```diff
@@ -0,0 +1,221 @@
...
+
+To account for these inconsistencies, *the GL backend vertically flips the rows of the readback
+image*.
```
pixelflinger, line 73: That's actually not correct. The GL backend inverts read pixel for a different reason!

It's for the case where you *RENDER* into an attachment. We want to get the same data back on all backends when using readPixels.

pixelflinger (changes requested):
I think i would structure things differently.

The first paragraph is good.  I would then describe what each backend does:

GL:
- uploads: low-address is bottom
- readbacks: bottom is low address
- rendering: origin is bottom-left
- sampling: origin is bottom-left

Others:
- uploads: low-address is top
- readbacks: top is low address
- rendering: origin is top-left
- sampling: origin is top-left

Then I would explain that this means that:
- upload + read: works in all backend
- but: render + read: produces different results

Then I would say that filament addresses this by:
- inverting reads when reading
- this fixes the render + read path
- this sacrifices the upload + read path
  - this is an unsolved problem in filament, only case where backends behave differently

bejado (commented):
Are the diagrams inside art/diagrams no longer accurate? If so, maybe we could update them instead of getting rid of them entirely.

poweifeng:
> Are the diagrams inside art/diagrams no longer accurate? If so, maybe we could update them instead of getting rid of them entirely.

Yes initially I just wanted to add text to the pictures because I found the pictures themselves a little confusing. But as I started writing, I thought words enough would be sufficient.  But we could do something in the middle where the pictures are included in this doc as embedded illustrations.

PR #9851, opened 2026-03-28
fix Ktx1Bundle metadata key null termination out of bounds read

Sakura-501:
## Summary

This PR fixes an out-of-bounds read in `image::Ktx1Bundle` metadata deserialization.

## Vulnerability

`Ktx1Bundle` parsed KTX key/value metadata by constructing the key with `std::string(const char*)` on attacker-controlled bytes. If a malformed metadata entry omitted a NUL terminator inside the current entry, the parser would continue scanning past the end of the metadata payload until it found a zero byte. In practice this allows a crafted `.ktx` file to trigger an out-of-bounds read during deserialization and crash any application or service that consumes untrusted KTX input.

The vulnerable path was in `libs/image/src/Ktx1Bundle.cpp` during metadata traversal.

## Fix

The patch makes metadata parsing range-aware before any string construction:

- verify that `bytesOfKeyValueData` stays inside the source buffer;
- verify that each metadata record header and record body are fully present;
- require a NUL terminator inside the current metadata entry before constructing the key;
- construct both key and value from explicit byte ranges instead of relying on unbounded C string scanning;
- verify that metadata padding also remains inside the metadata block.

## Tests

Added a regression test in `libs/image/tests/test_image.cpp` that feeds a malformed metadata entry without a NUL terminator and verifies the constructor rejects it with a controlled precondition failure instead of walking out of bounds.

This change only addresses malformed metadata key parsing. It does not include the independent image blob length validation fix or the uberz archive validation fix.

### DIFF
--- a/libs/image/src/Ktx1Bundle.cpp
+++ b/libs/image/src/Ktx1Bundle.cpp
@@ -19,6 +19,7 @@
 #include <utils/Panic.h>
 #include <utils/string.h>
 
+#include <cstring>
 #include <string>
 #include <unordered_map>
 #include <vector>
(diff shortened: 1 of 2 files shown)


pixelflinger:
we will do our own fixes. thanks for reporting.

PR #9852, opened 2026-03-28
fix Ktx1Bundle image blob copy heap buffer overflow

Sakura-501:
## Summary

This PR fixes a heap buffer overflow in `image::Ktx1Bundle` image blob deserialization.

## Vulnerability

`Ktx1Bundle` sized its destination blob storage from the number of bytes remaining in the serialized input, but then trusted attacker-controlled `imageSize` metadata when calculating the number of bytes to copy with `memcpy`. A minimal KTX payload that declared a much larger image body than was actually present could therefore drive out-of-bounds reads and writes during blob extraction.

This affects any code path that deserializes untrusted `.ktx` data through `image::Ktx1Bundle`.

## Fix

The patch enforces consistency between declared blob sizes and the actual serialized buffers:

- verify that an image size header is fully present before reading it;
- compute the effective level size and verify that it fits inside the remaining source bytes;
- verify that the computed level size also fits inside the allocated destination blob buffer before calling `memcpy`.

If any of those checks fail, deserialization now aborts with a controlled precondition failure instead of corrupting memory.

## Tests

Added a regression test in `libs/image/tests/test_image.cpp` that provides a truncated KTX image body and verifies the constructor rejects it instead of accepting the malformed blob layout.

This change only addresses the image blob copy length mismatch. It does not include the independent metadata key parsing fix or the uberz offset validation fix.

### DIFF
--- a/libs/image/src/Ktx1Bundle.cpp
+++ b/libs/image/src/Ktx1Bundle.cpp
@@ -150,16 +150,23 @@ Ktx1Bundle::Ktx1Bundle(uint8_t const* bytes, uint32_t nbytes) :
     // One aspect of the KTX spec is that the semantics differ for non-array cubemaps.
     const bool isNonArrayCube = mNumCubeFaces > 1 && mArrayLength == 1;
     const uint32_t facesPerMip = mArrayLength * mNumCubeFaces;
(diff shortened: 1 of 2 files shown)


pixelflinger:
we will do our own fixes. thanks for reporting.

PR #9853, opened 2026-03-28
fix uberz offset to pointer out of bounds write

Sakura-501:
## Summary

This PR fixes out-of-bounds reads and writes in `uberz::convertOffsetsToPointers` by making archive pointer conversion size-aware.

## Vulnerability

`convertOffsetsToPointers` rewrote attacker-controlled offsets from a decompressed uberz archive into live pointers without validating that the offsets remained inside the archive buffer. In release builds the existing `assert_invariant` checks compile away, so malformed values such as `specsOffset`, `flagsOffset`, `packageOffset`, or `nameOffset` could force the converter to write invalid pointers into the archive structure and then continue traversing those invalid addresses.

Real call sites include:

- `gltfio::ArchiveCache::load`
- the append path in `tools/uberz`

Any application or service that loads untrusted `.uberz` content could therefore crash or hit broader memory corruption.

## Fix

The patch changes the conversion API to accept the decompressed archive size and validates all offset-derived regions before any pointer rewriting happens:

- validate archive header size, magic, version, and alignment requirements;
- validate the `specs` array range before converting `specsOffset`;
- validate each `flags` array, package range, and NUL-terminated flag name string;
- keep these checks active in release builds;
- update both `gltfio` and `tools/uberz` to pass the decompressed archive size into the validator.

## Tests

Added `libs/uberz/tests/test_ReadableArchive.cpp` with regression coverage for:

- a top-level `specsOffset` that points outside the archive buffer;
- a nested `flag.nameOffset` that points outside the archive buffer.

This change only addresses uberz archive validation and call-site propagation. It is intentionally separate from the two independent KTX deserialization fixes.

### DIFF
--- a/libs/gltfio/src/ArchiveCache.cpp
+++ b/libs/gltfio/src/ArchiveCache.cpp
@@ -67,7 +67,7 @@ void ArchiveCache::load(const void* archiveData, uint64_t archiveByteCount) {
     uint64_t* basePointer = (uint64_t*) utils::aligned_alloc(decompSize, 8);
     ZSTD_decompress(basePointer, decompSize, archiveData, archiveByteCount);
     mArchive = (ReadableArchive*) basePointer;
-    convertOffsetsToPointers(mArchive);
+    convertOffsetsToPointers(mArchive, decompSize);
     mMaterials = FixedCapacityVector<Material*>(mArchive->specsCount, nullptr);
 }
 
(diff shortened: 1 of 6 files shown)


pixelflinger:
The fix was incomplete. See https://github.com/google/filament/pull/9896 for a more complete fix.

PR #9855, opened 2026-03-28
fp32 texture formats are not necessarily filterable

pixelflinger:
Mobile devices don't always support filtering of fp32 textures. This PR adds a query at the driver level.

High precision EVSM is now conditional to this feature being supported.

### DIFF
--- a/filament/backend/src/metal/MetalDriver.mm
+++ b/filament/backend/src/metal/MetalDriver.mm
@@ -1261,6 +1261,9 @@ bool MetalDriver::isTextureFormatMipmappable(TextureFormat format) {
 
 bool MetalDriver::isTextureFormatFilterable(TextureFormat format) {
     if (isFp32ColorFormat(format)) {
+        if (@available(macOS 11.0, iOS 14.0, *)) {
+            return mContext->device.supports32BitFloatFiltering;
+        }
         return mContext->highestSupportedGpuFamily.apple >= 7 ||
                mContext->highestSupportedGpuFamily.mac >= 1;
     }


pixelflinger:
@bejado @poweifeng metal, vulkan and webgpu implementations by gemini. please review :-)

filament/backend/src/opengl/OpenGLDriver.cpp
```diff
@@ -2775,6 +2775,39 @@ bool OpenGLDriver::isTextureFormatMipmappable(TextureFormat const format) {
...
+        case TextureFormat::R16F:
+        case TextureFormat::RG16F:
+        case TextureFormat::RGB16F:
+        case TextureFormat::RGBA16F:
+            return gl.ext.OES_texture_half_float_linear;
+        case TextureFormat::R32F:
+        case TextureFormat::RG32F:
+        case TextureFormat::RGB32F:
+        case TextureFormat::RGBA32F:
+            return gl.ext.OES_texture_float_linear;
```
romainguy, line 2804: If GL desktop is still supported, 16F and 32F are filterable by default

filament/backend/src/metal/MetalDriver.mm
```diff
@@ -1259,6 +1259,18 @@
     }
 }
 
+bool MetalDriver::isTextureFormatFilterable(TextureFormat format) {
+    if (isFp32ColorFormat(format)) {
+        return mContext->highestSupportedGpuFamily.apple >= 7 ||
+               mContext->highestSupportedGpuFamily.mac >= 1;
+    }
```
bejado, line 1269: I think we may also need to do a [`MTLDevice supports32BitFloatFiltering`](https://developer.apple.com/documentation/metal/mtldevice/supports32bitfloatfiltering?language=objc) check here.

filament/backend/src/opengl/OpenGLDriver.cpp
pixelflinger: yes, `OES_texture_float_linear` is forced to `true` in that case.

filament/backend/src/metal/MetalDriver.mm
pixelflinger: I added the check, I think it's correct!

bejado: LGTM

PR #9857, opened 2026-03-30
Support Engine::compile overload to generate view-dependent variants 

pixelflinger:
Implemented an `Engine::compile()` method to infer and generate shader 
permutations directly based on feature states from a provided `View` 
alongside `tribool` settings (skinning, shadowReceiver). 

- Added a new `Material::compile()` overload that directly accepts 
  a precalculated `FixedCapacityVector<Variant>`, decoupling the 
  Material compilation process from specific feature flag checks.

FIXES=[468058969]

### DIFF
--- a/filament/src/details/Material.cpp
+++ b/filament/src/details/Material.cpp
@@ -240,8 +240,7 @@ filament::DescriptorSetLayout const& FMaterial::getPerViewDescriptorSetLayout(
 void FMaterial::compile(CompilerPriorityQueue const priority,
         UserVariantFilterMask variantSpec,
         CallbackHandler* handler,
-        Invocable<void(Material*)>&& callback) noexcept {
-
+        Invocable<void(Material*)>&& callback) const noexcept {
     DriverApi& driver = mEngine.getDriverApi();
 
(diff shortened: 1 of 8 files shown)


filament/src/details/Material.cpp
```diff
@@ -240,17 +240,15 @@ filament::DescriptorSetLayout const& FMaterial::getPerViewDescriptorSetLayout(
...
 
     // Turn off the STE variant if stereo is not supported.
     if (!mIsStereoSupported) {
         variantSpec &= ~UserVariantFilterMask(UserVariantFilterBit::STE);
     }
 
-    UserVariantFilterMask const variantFilter =
-            ~variantSpec & UserVariantFilterMask(UserVariantFilterBit::ALL);
+    UserVariantFilterMask const variantFilter = ~variantSpec & UserVariantFilterMask(UserVariantFilterBit::ALL);
     ShaderModel const shaderModel = mEngine.getShaderModel();
```
poweifeng, line 252: nit: can just call the other `compile` with a filtered list of variants.

PR #9862, opened 2026-03-31
fgviewer: visualize intermediate buffers

poweifeng:
* Intermediate Buffer Visualization: Enabled live monitoring of internal `FrameGraph` render targets directly in the `fgviewer` web UI.
* HTTP Polling Architecture: Switched from WebSocket binary pushes to native `` polling (`/api/image`), eliminating severe frontend memory leaks and GC thrashing.
* Robust Resource Tracking: Replaced string-based lookups with `(ViewId, ResourceId)` composite keys to prevent cross-view collisions and ensure accurate reads.
* Format Post-Processing: Extracted readback conversions (HDR tonemapping, depth normalization, MSAA downsampling, single-channel expansion) into `DebugServer`.
* UI Polish: Added a live-updating full-screen image modal and explicitly filtered internal debug passes from the Graphviz/JSON exports to prevent DOM thrashing.
* Currently Unsupported:
  * Mipmaps & Subresources: Reading specific mip levels or array layers is explicitly skipped.
  * Shadowmaps: Variance Shadow Maps (VSM) will physically evaluate to `0.0` and appear completely empty/black in scenes without active shadow casters (due to inverted-Z).
  * Stencil: Resolving and reading stencil buffer data is not supported by the backend.

image


show50726:
It looks great! Thank you!

Couple questions:
1. Since a texture could be written several times within a frame, is the image shown on the web page always the final state of the texture, or can we specify the timing?
2. Is it tested on android as well?

filament/src/details/Renderer.cpp
```diff
@@ -1568,10 +1574,223 @@ void FRenderer::renderJob(DriverApi& driver, RootArenaScope& rootArenaScope, FVi
...
+    }
+#endif
+
     // save the current history entry and destroy the oldest entry
     view.commitFrameHistory(engine);
 
     recordHighWatermark(commandArena.getListener().getHighWatermark());
 }
 
+#if FILAMENT_ENABLE_FGVIEWER
```
pixelflinger: as the fgviewer supports gets bigger, do you think we could refactor this out  into its own file or class (kind of like postprocessmanager)

filament/src/fg/FrameGraph.cpp
```diff
@@ -15,7 +15,11 @@
  */
 
 #include "fg/FrameGraph.h"
+
+#include <utils/Log.h>
+
```
pixelflinger: move with other utils/ includes

pixelflinger (changes requested):

filament/src/details/Renderer.cpp
poweifeng: done.  I moved all that code into its own source file

filament/src/fg/FrameGraph.cpp
poweifeng: removed.

poweifeng:
> It looks great! Thank you!
> 
> Couple questions:
> 
> 1. Since a texture could be written several times within a frame, is the image shown on the web page always the final state of the texture, or can we specify the timing?
> 2. Is it tested on android as well?

1. There is no way to specify timing.  It's always the result of the buffer after the last pass (dependents on the output of the previous passes).  Which buffer is written more than once?
2. No, it's not tested beyond MacOS + metal/vk.  The changes made here are pretty complicated (to get all the formats working).  If ok, I'd like to check this in and then iterate on the other backends and platforms.

show50726:
> > It looks great! Thank you!
> > Couple questions:
> > 
> > 1. Since a texture could be written several times within a frame, is the image shown on the web page always the final state of the texture, or can we specify the timing?
> > 2. Is it tested on android as well?
> 
> 1. There is no way to specify timing.  It's always the result of the buffer after the last pass (dependents on the output of the previous passes).  Which buffer is written more than once?
> 2. No, it's not tested beyond MacOS + metal/vk.  The changes made here are pretty complicated (to get all the formats working).  If ok, I'd like to check this in and then iterate on the other backends and platforms.

Got it! For your question, like Structure Buffer and Bloom Out Texture in your screenshot are both written by 2 passes?

filament/src/fg/FgviewerManager.h
```diff
@@ -0,0 +1,46 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef TNT_FILAMENT_FGVIEWERMANAGER_H
+#define TNT_FILAMENT_FGVIEWERMANAGER_H
+
+#include <stdint.h>
+#include <backend/Handle.h>
```
show50726: Should `#include <backend/Handle.h>` be in front of `#include <stdint.h>`?

filament/src/details/Renderer.cpp
```diff
@@ -1549,25 +1550,23 @@ void FRenderer::renderJob(DriverApi& driver, RootArenaScope& rootArenaScope, FVi
 //    auto debug = structure
 //    fg.forwardResource(fgViewRenderTarget, debug ? debug : input);
 
-    fg.forwardResource(fgViewRenderTarget, input);
+    FgviewerManager::handleFgReadbacks(engine, fg, viewRenderTarget, view.getViewHandle());
```
show50726: Should these `FgviewerManager` calls be guarded by `#if FILAMENT_ENABLE_FGVIEWER`?

poweifeng:
> > > It looks great! Thank you!
> > > Couple questions:
> > > 
> > > 1. Since a texture could be written several times within a frame, is the image shown on the web page always the final state of the texture, or can we specify the timing?
> > > 2. Is it tested on android as well?
> > 
> > 
> > 
> > 1. There is no way to specify timing.  It's always the result of the buffer after the last pass (dependents on the output of the previous passes).  Which buffer is written more than once?
> > 2. No, it's not tested beyond MacOS + metal/vk.  The changes made here are pretty complicated (to get all the formats working).  If ok, I'd like to check this in and then iterate on the other backends and platforms.
> 
> Got it! For your question, like Structure Buffer and Bloom Out Texture in your screenshot are both written by 2 passes?

I see, I think different mip levels and layers are written to by different passes, but I'm guessing, for one layer/level in one texture, it's only ever written to once per-frame.

filament/src/details/Renderer.cpp
poweifeng: FgviewerManager itself has that `#define` guard.  These methods a no-op in the usual case.

filament/src/fg/FgviewerManager.h
poweifeng: done

show50726 (approved):

filament/src/fg/FgviewerManager.h
```diff
@@ -0,0 +1,47 @@
...
+class FView;
+class FrameGraph;
+
+class FgviewerManager {
+public:
+#if FILAMENT_ENABLE_FGVIEWER
+    static void handleFgReadbacks(FEngine& engine, FrameGraph& fg,
+            backend::Handle<backend::HwRenderTarget> viewTarget, uint32_t viewId);
+    static void update(FEngine& engine, FrameGraph& fg, FView const& view);
+    static void tick(FEngine& engine);
```
pixelflinger: some documentation would be nice

filament/src/details/Engine.h
```diff
@@ -570,6 +570,26 @@ class FEngine : public Engine, public utils::FeatureFlagManager {
...
+
+    void requestTextureReadback(fgviewer::ViewHandle viewId, uint32_t id,
+            utils::CString const& name,
+            std::function<void(fgviewer::DebugServer::PixelBuffer, uint32_t, uint32_t,
+                    fgviewer::DebugServer::PixelDataFormat, fgviewer::DebugServer::FormatInfo)>&&
+                    callback);
+
+    utils::Mutex mReadbackRequestsMutex;
+    std::vector<ReadbackRequest> mReadbackRequests;
+#endif
```
pixelflinger: does this needs to be on Engine? Could it be on the new FgviewerManager?

filament/backend/src/vulkan/utils/Conversion.cpp
```diff
@@ -770,6 +770,7 @@ PixelDataType getComponentType(VkFormat format) {
         case VK_FORMAT_D16_UNORM: return PixelDataType::USHORT;
         case VK_FORMAT_D32_SFLOAT: return PixelDataType::FLOAT;
         case VK_FORMAT_X8_D24_UNORM_PACK32: return PixelDataType::UINT;
+        case VK_FORMAT_B10G11R11_UFLOAT_PACK32: return PixelDataType::UINT_10F_11F_11F_REV;
```
pixelflinger, line 773: I would move all the VK changes in a different CL (or PR), seems unrelated.

filament/src/details/Renderer.cpp
pixelflinger: does this have to be done before present()?

IF not, you could hide all this inside the FRameGraph itself.

Like you could call handleFgReadbacks at the beginging of compile(), update() at the end of compile() and tick() at the end of execute().

It would be all contained inside the FrameGraph,  and FgviewerManager could become an implementation detail of FrameGraph.

pixelflinger (changes requested):

filament/src/details/Engine.h
poweifeng: moved to FgviewerManager

filament/src/fg/FgviewerManager.h
poweifeng: done

filament/backend/src/vulkan/utils/Conversion.cpp
poweifeng: Done

filament/src/details/Renderer.cpp
poweifeng: I've moved all the fgviewer logic to FrameGraph

pixelflinger (approved):

PR #9866, opened 2026-04-01
postfx materials now support sRGB emulation in FL0

pixelflinger:


shaders/src/post_process_main.fs
```diff
@@ -4,6 +4,16 @@ void main() {
     // Invoke user code
     postProcess(inputs);
 
+#if MATERIAL_FEATURE_LEVEL == 0
+    if (CONFIG_SRGB_SWAPCHAIN_EMULATION) {
+        if (frameUniforms.rec709 != 0) {
+            inputs.FRAG_OUTPUT0.r = pow(inputs.FRAG_OUTPUT0.r, 0.45454);
```
romainguy: This can be written as `inputs.FRAG_OUTPUT0 = pow(inputs.FRAG_OUTPUT0, vec3(1.0 / 2.2))` instead of using 3 separate lines.

PR #9867, opened 2026-04-02
FL0 materials are always compiled via spirv-cross

pixelflinger:


libs/filamat/src/GLSLPostProcessor.cpp
```diff
@@ -950,6 +950,38 @@ bool GLSLPostProcessor::fullOptimization(const TShader& tShader,
...
         }
+
+        // Validate the transpiled ESSL1 shader dynamically before considering it successful.
+        // This proactively catches unsupported SPIR-V -> ESSL1 translation quirks (like textureLod)
+        // at compile-time since we can't easily test all variants on physical GLES 2.0 devices.
+        if (config.featureLevel == 0) {
+            // preampitively forbid spirv-cross from cheating and polyfilling disabled features
+            auto const& exts = glslCompiler.get_required_extensions();
+            for (auto const& ext : exts) {
+                if (ext != "GL_OES_standard_derivatives" &&
```
romainguy, line 961: You should have a map/set of allowed extensions and just do an `if (!allowedMap.contains(ext))`. It'll make it easier to track/maintain the list of extension instead of finding this somewhat random piece of code.

shaders/src/post_process_main.fs
romainguy, line 10: You should just write 1.0 / 2.2, I think it's a bit more obvious for the reader (also what we use in other places that do sRGB conversion).

shaders/src/surface_main.fs
romainguy, line 113: See above

Issue #9294, opened 2025-10-06
[split_config.arm64_v8a.apk!libfilament-jni.so] utils::TPanic<utils::PostconditionPanic>::panic(char const*, char const*, int, char const*, std::__ndk1::basic_string<char, std::__ndk1::char_traits<char>, std::__ndk1::allocator<char>>)


sanchit-cars24:
Android app faces crash and I got "[split_config.arm64_v8a.apk!libfilament-jni.so] utils::TPanic<utils::PostconditionPanic>::panic(char const*, char const*, int, char const*, std::__ndk1::basic_string<char, std::__ndk1::char_traits<char>, std::__ndk1::allocator<char>>)" from logs. It crashes in android 15 and also below versions

**To Reproduce**
1. This is random (no evidance so far to mention specific steps)

**Expected behavior**
I expected crash not to happen
If crash happens then it should give more details in log to know what causes this error

**Screenshots**
NA

**Logs**
[split_config.arm64_v8a.apk!libfilament-jni.so] utils::TPanic<utils::PostconditionPanic>::panic(char const*, char const*, int, char const*, std::__ndk1::basic_string<char, std::__ndk1::char_traits<char>, std::__ndk1::allocator<char>>)

**Desktop (please complete the following information):**
 - OS: Android

**Smartphone (please complete the following information):**
 - Device: [Redmi Redmi Note 13 Pro 5G]
 - OS: [e.g. Android 15.0 and lowers other devices

z3moon:
Thanks for the report. But we need more information. Can you grab a stack trace? if not, what was the app performing during the crash?

sanchit-cars24:
When we launch app, after the splash screen disappear, we see few APIs(15-16) including analytics , app config , user data and server images in the logs after these api calls it suddenly crashes with this reason, it is completely random , there is no specific api after which it occurs and no as such no of api calls. 

- For some users it happens after 2-3 min usage of app
- play store stack trace only gives 
-*** *** *** *** *** *** *** *** *** *** *** *** *** *** *** ***

backtrace:
  #00  pc 0x000000000005d7d8  /apex/com.android.runtime/lib64/bionic/libc.so (abort+172)
[...]

PR #9885, opened 2026-04-08
Improved TPanic message for Android

paulorrfilho:
Register the full panic message as the tombstone “Abort message:” so it appears in crash reports (e.g., Google Play Console) that rely on device tombstones instead of logcat output.
Related to #9294 #9075

### DIFF
--- a/libs/utils/src/Panic.cpp
+++ b/libs/utils/src/Panic.cpp
@@ -36,6 +36,9 @@
 #include <new>
 #include <string_view>
 #include <utility>
+#if defined(__ANDROID__)
+#    include <android/set_abort_message.h>
+#endif
 
 namespace utils {
 
@@ -221,6 +224,13 @@ void TPanic<T>::panic(char const* function, char const* file, int line, char con
 
     // Call the user provided handler
     UserPanicHandler::get().call(e);
+
+    // Register the full panic message as the tombstone "Abort message:" so that it
+    // appears in crash reports collected by Google Play Console and other tools that
+    // read tombstones from field devices (which never have access to logcat output).
+#if defined(__ANDROID__)
+    android_set_abort_message(e.what());
+#endif
 
     // if exceptions are enabled, throw now.
 #ifdef __EXCEPTIONS


romainguy:
@pixelflinger

libs/utils/src/Panic.cpp
```diff
@@ -222,6 +225,13 @@ void TPanic<T>::panic(char const* function, char const* file, int line, char con
     // Call the user provided handler
     UserPanicHandler::get().call(e);
 
+    // Register the full panic message as the tombstone "Abort message:" so that it
+    // appears in crash reports collected by Google Play Console and other tools that
+    // read tombstones from field devices (which never have access to logcat output).
+#if defined(__ANDROID__)
+    android_set_abort_message(e.what());
+#endif
+
```
pixelflinger, line 234: this should be moved just above the `std::abort()` call, and after the `throw e` (if exceptions are enabled, we might not abort).

pixelflinger (changes requested):

pixelflinger (changes requested):
Unfortunately  `android_set_abort_message()` is not an official public API. However:

```
__android_log_default_aborter(const char *abort_message)
```

Seems to be a public API as of API level 30.

So, we should be able to use `__android_log_default_aborter` instead of `abort()` on android if we surround the code with
 
```
if (__builtin_available(android 30, *)) {
}
```

(on Android only).

If we're not on API level 30, then as a workaround, we can maybe use `android_set_abort_message()` + `abort()`. But we have to be careful. The best way would be to do this:

```
extern "C" __attribute__((weak)) void android_set_abort_message(const char *);
if (&android_set_abort_message) {
    android_set_abort_message(str);
}
```
and avoid including the header entirely.

poweifeng:
@paulorrfilho are you still interested in finishing this PR?  If not, I can pick it up.
Just so we can reduce PR docket by a little.

paulorrfilho:
@poweifeng yes, please go ahead, I don't have time to review this item right now.
Thank you

poweifeng:
Closing this one.  Cloned and addressed comments in #10061 10061

PR #9891, opened 2026-04-10
Fix ssao precision issue around silhouettes

poweifeng:
In the ssao pass, for silhouette pixels,

```
highp vec3 v = p - origin;
```

can be a "long" vector where the length cannot be captured with mediump on mobile. This means that

```
float vv = dot(v, v);
float vn = dot(v, normal);
```
and other variables can overflow. This caused the silhouettes to be dark when ssao is turned on a PowerVR gpu.

image

The fix is use highp for the relevant variables.


show50726 (approved):

filament/src/materials/ssao/saoImpl.fs
```diff
@@ -73,21 +73,21 @@ void computeAmbientOcclusionSAO(inout float occlusion, inout vec3 bentNormal,
...
-    float vv = dot(v, v);       // squared distance
-    float vn = dot(v, normal);  // distance * cos(v, normal)
+    highp float vv = dot(v, v);       // squared distance
+    highp float vn = dot(v, normal);  // distance * cos(v, normal)
 
     // discard samples that are outside of the radius, preventing distant geometry to
     // cast shadows -- there are many functions that work and choosing one is an artistic
     // decision.
-    float w = sq(max(0.0, 1.0 - vv * materialParams.invRadiusSquared));
+    highp float w = sq(max(0.0, 1.0 - vv * materialParams.invRadiusSquared));
```
pixelflinger, line 83: `w` should be between [0, 1]  so I think it can stay `mediump`

filament/src/materials/ssao/saoImpl.fs
```diff
@@ -73,21 +73,21 @@ void computeAmbientOcclusionSAO(inout float occlusion, inout vec3 bentNormal,
...
 
     // discard samples that are too close to the horizon to reduce shadows cast by geometry
     // not sufficently tessellated. The goal is to discard samples that form an angle 'beta'
     // smaller than 'epsilon' with the horizon. We already have dot(v,n) which is equal to the
     // sin(beta) * |v|. So the test simplifies to vn^2 < vv * sin(epsilon)^2.
     w *= step(vv * materialParams.minHorizonAngleSineSquared, vn * vn);
 
-    float sampleOcclusion = max(0.0, vn + (origin.z * materialParams.bias)) / (vv + materialParams.peak2);
+    highp float sampleOcclusion = max(0.0, vn + (origin.z * materialParams.bias)) / (vv + materialParams.peak2);
```
pixelflinger: same here `sampleOcclusion` should be between [0, 1]

pixelflinger (changes requested):

filament/src/materials/ssao/saoImpl.fs
poweifeng: Removed `highp`.

filament/src/materials/ssao/saoImpl.fs
poweifeng: I tried it, but the artifact is present without the `highp`.

pixelflinger (approved):

Issue #9890, opened 2026-04-10
Clearcoat IBL attenuation inconsistant between doc and code


kai-z99:
Documentation for clearcoat IBL says:

`iblSpecular *= sq(1.0 - Fc)`

while in surface_light_indirect.fs we have

`Fr *= attenuation;` (not squared)

where
`float attenuation = 1.0 - Fc;`

For both importance sampling and split sum path

romainguy:
The document you are referring to describes Filament's lighting and material model at a certain point in time and does not necessarily accurately represents the current state of the engine. It remains however a perfectly valid reference on how to implement a PBR system. Removing the square attenuation was done after energy conservation was added. This extra term was already trying to account for it.

romainguy:
What I'm basically saying is that the squaring of the specular attenuation with a clear coat layer should just be removed from the documentation.

romainguy:
@pixelflinger This can be closed

PR #9892, opened 2026-04-10
update clearcoat ibl doc

kai-z99:
#9890 Fix

### DIFF
--- a/docs/Filament.md.html
+++ b/docs/Filament.md.html
@@ -892,7 +892,7 @@ void BRDF(...) {
     float Frc = (Dc * Vc) * Fc;
 
     // account for energy loss in the base layer
-    return color * ((Fd + Fr * (1.0 - Fc)) * (1.0 - Fc) + Frc);
+    return color * ((Fd + Fr) * (1.0 - Fc) + Frc);
 }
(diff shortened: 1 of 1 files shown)


docs/Filament.md.html
```diff
@@ -2362,8 +2362,8 @@
 float Fc = F_Schlick(0.04, 1.0, clearCoat_NoV) * clearCoat;
 // base layer attenuation for energy compensation
 iblDiffuse  *= 1.0 - Fc;
-iblSpecular *= sq(1.0 - Fc);
-iblSpecular += specularIBL(r, clearCoatPerceptualRoughness) * Fc;
+iblSpecular *= 1.0 - Fc;
+iblSpecular += evaluateSpecularIBL(r, clearCoatPerceptualRoughness) * Fc;
```
romainguy, line 2366: Thanks. In section 4.9.13 ("Clear coat parameterization"), can you also update the direct lighting sample? Instead of:

```
return color * ((Fd + Fr * (1.0 - Fc)) * (1.0 - Fc) + Frc);
```

It should be:

```
return color * ((Fd + Fr) * (1.0 - Fc) + Frc);
```

kai-z99: Resolved, thanks

PR #9896, opened 2026-04-14
security fixes

pixelflinger:


libs/gltfio/src/ResourceLoader.cpp
```diff
@@ -273,10 +314,45 @@ inline void uploadBuffers(FFilamentAsset* asset, Engine& engine,
...
+
+                const size_t components = cgltf_num_components(accessor->type);
+                if (components > 0 && safeCount > std::numeric_limits<size_t>::max() / components) {
+                    continue;
+                }
+                const size_t floatsCount = safeCount * components;
+                
+                if (floatsCount > std::numeric_limits<size_t>::max() / sizeof(float)) {
+                    continue;
+                }
```
poweifeng, line 359: I think we can also continue `if (floatsCount == 0)` ? 

This ca happen if `maxCount == 0` or when `elementSize > availableBytes`

libs/gltfio/src/ResourceLoader.cpp
```diff
@@ -313,8 +389,39 @@ inline void uploadBuffers(FFilamentAsset* asset, Engine& engine,
...
+
+            const size_t components = cgltf_num_components(accessor->type);
+            if (components > 0 && safeCount > std::numeric_limits<size_t>::max() / components) {
+                continue;
+            }
+            const size_t floatsCount = safeCount * components;
+            
+            if (floatsCount > std::numeric_limits<size_t>::max() / sizeof(float)) {
+                continue;
+            }
```
poweifeng, line 431: some comment as above

libs/gltfio/src/ResourceLoader.cpp
pixelflinger: done

libs/gltfio/src/ResourceLoader.cpp
pixelflinger: done

libs/image/src/Ktx1Bundle.cpp
```diff
@@ -103,10 +104,18 @@ Ktx1Bundle::~Ktx1Bundle() = default;
 
 Ktx1Bundle::Ktx1Bundle(uint32_t numMipLevels, uint32_t arrayLength, bool isCubemap) :
         mBlobs(new KtxBlobList), mMetadata(new KtxMetadata) {
+    uint64_t const totalBlobs = (uint64_t)numMipLevels * arrayLength * (isCubemap ? 6 : 1);
+    if (totalBlobs > (uint64_t)std::numeric_limits<uint32_t>::max()) {
+        utils::slog.w << "KTX dimensions overflow" << utils::io::endl;
```
poweifeng: prefer `LOG(WARNING)`

libs/uberz/src/ReadableArchive.cpp
```diff
@@ -27,21 +30,110 @@ static_assert(sizeof(ReadableArchive) == 4 + 4 + 8 + 8);
...
+        return nullptr;
+    }
+    return string;
+}
+
+} // namespace
+
+bool convertOffsetsToPointers(ReadableArchive* archive, size_t archiveSize) {
+    if (archiveSize < sizeof(ReadableArchive)) {
+        slog.e << "Uberz archive header is truncated" << io::endl;
```
poweifeng: prefer `LOG(ERROR)` and other places in this file.

libs/ktxreader/src/Ktx2Reader.cpp
```diff
@@ -259,6 +278,7 @@ Texture* Ktx2Reader::load(const void* data, size_t size, TransferFunction transf
             return nullptr;
         }
         texture->setImage(mEngine, levelIndex, std::move(*pbd));
+        delete pbd;
```
poweifeng, line 281: is it safe to delete here since you also moved it? What is it actually pointing to?

poweifeng (commented):
one actual question and some nits (nits can be ignored or addressed later).

libs/ktxreader/src/Ktx2Reader.cpp
pixelflinger: it's not moving the pointer, it's moving the pointee. You still need to delete the object. It's similar to:

{
std::vector<int> a;
auto b = std::move(a);
// here the dtor of a and b are called
}

poweifeng (approved):

libs/uberz/src/ReadableArchive.cpp
pixelflinger: done

libs/image/src/Ktx1Bundle.cpp
pixelflinger: done

PR #9897, opened 2026-04-14
remove the now unneeded FL0 materials

pixelflinger:
Filament needed to compile FL0 materials for both FL0 and FL1, but this is now done automatically when a material is a FL0 material.

### DIFF
--- a/filament/src/details/Engine.cpp
+++ b/filament/src/details/Engine.cpp
@@ -350,10 +350,6 @@ void FEngine::init() {
 
     mActiveFeatureLevel = std::min(mActiveFeatureLevel, driverApi.getFeatureLevel());
 
-#ifndef FILAMENT_ENABLE_FEATURE_LEVEL_0
-    assert_invariant(mActiveFeatureLevel > FeatureLevel::FEATURE_LEVEL_0);
-#endif
-
     LOG(INFO) << "Backend feature level: " << int(driverApi.getFeatureLevel());
     LOG(INFO) << "FEngine feature level: " << int(mActiveFeatureLevel);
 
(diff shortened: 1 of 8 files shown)


show50726 (approved):

z3moon (approved):

filament/src/details/Engine.cpp
```diff
@@ -465,7 +465,7 @@ void FEngine::init() {
     if (UTILS_UNLIKELY(mActiveFeatureLevel == FeatureLevel::FEATURE_LEVEL_0)) {
         FMaterial::DefaultMaterialBuilder defaultMaterialBuilder;
         defaultMaterialBuilder.package(
-                MATERIALS_DEFAULTMATERIAL_FL0_DATA, MATERIALS_DEFAULTMATERIAL_FL0_SIZE);
+                MATERIALS_DEFAULTMATERIAL_DATA, MATERIALS_DEFAULTMATERIAL_SIZE);
         mDefaultMaterial = downcast(defaultMaterialBuilder.build(*const_cast<FEngine*>(this)));
     } else
 #endif
```
poweifeng: can we just remove this whole thing?

filament/src/details/Skybox.cpp
```diff
@@ -134,7 +134,7 @@ FMaterial const* FSkybox::createMaterial(FEngine& engine) {
     Material::Builder builder;
 #ifdef FILAMENT_ENABLE_FEATURE_LEVEL_0
     if (UTILS_UNLIKELY(engine.getActiveFeatureLevel() == Engine::FeatureLevel::FEATURE_LEVEL_0)) {
-        builder.package(MATERIALS_SKYBOX_FL0_DATA, MATERIALS_SKYBOX_FL0_SIZE);
+        builder.package(MATERIALS_SKYBOX_DATA, MATERIALS_SKYBOX_SIZE);
     } else
 #endif
```
poweifeng: remove?

filament/src/details/Engine.cpp
pixelflinger: oh yes we can. good catch!

filament/src/details/Skybox.cpp
pixelflinger: done

poweifeng (approved):

show50726 (approved):

PR #9899, opened 2026-04-15
Backend: Centralize fence synchronization and fix potential data races.

pixelflinger:
Transitioned fence synchronization from per-fence condition variables to a centralized shared condition variable in DriverBase, enabling future global error propagation and interruption.

Encapsulated synchronization by hiding the mutex and condition variable within DriverBase and exposing `waitForFence` and `signalFence`.

Fixed a critical bug in Metal backend where spurious wakeups would reset the relative timeout duration, potentially leading to much longer waits than intended. Resolved by using absolute time points (wait_until).

Fixed potential data races in Metal, OpenGL, and WebGPU backends by ensuring that shared state (status or state pointers) is always accessed under the protection of the mutex, specifically by capturing it inside the predicate.

Simplified Vulkan and WebGPU fences to be purely passive data holders.

Trade-off: Spurious wakeups are increased due to the shared condition variable, but are handled correctly by mandatory predicates in all backends.

### DIFF
--- a/filament/backend/src/metal/MetalHandles.mm
+++ b/filament/backend/src/metal/MetalHandles.mm
@@ -20,6 +20,7 @@
 #include "MetalEnums.h"
 #include "MetalUtils.h"
 #include "MetalBufferPool.h"
+#include "MetalDriver.h"
 
 #include <filament/SwapChain.h>
 
@@ -1404,12 +1405,13 @@ void MetalFence::encode() {
 
         // Using a weak_ptr here because the Fence could be deleted before the block executes.
         std::weak_ptr<State> weakState = state;
+        MetalDriver* driver = context.driver;
         [event notifyListener:context.eventListener atValue:value block:^(id <MTLSharedEvent> o,
                 uint64_t value) {
             if (auto s = weakState.lock()) {
-                std::lock_guard<std::mutex> guard(s->mutex);
-                s->status = FenceStatus::CONDITION_SATISFIED;
-                s->cv.notify_all();
+                driver->signalFence([&] {
+                    s->status = FenceStatus::CONDITION_SATISFIED;
+                });
             }
         }];
     }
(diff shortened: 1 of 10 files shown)


filament/backend/src/vulkan/VulkanDriver.cpp
```diff
@@ -1486,7 +1490,18 @@ FenceStatus VulkanDriver::fenceWait(FenceHandle const fh, uint64_t const timeout
         until = now + nanoseconds(timeout);
     }
 
-    auto const [cmdfence, canceled] = fence->wait(until);
+    std::shared_ptr<VulkanCmdFence> cmdfence;
+    bool canceled = false;
+    waitForFence([&] {
+        auto status = fence->getStatus();
```
poweifeng, line 1496: can use `auto [curFence, curCanceled]`

poweifeng (approved):

filament/backend/src/metal/MetalHandles.mm
```diff
@@ -1422,24 +1424,31 @@ static void func(void* user) {
...
     if (@available(iOS 12, *)) {
         using ns = std::chrono::nanoseconds;
-        std::unique_lock<std::mutex> guard(state->mutex);
-        while (state->status == FenceStatus::TIMEOUT_EXPIRED) {
-            if (timeoutNs == FENCE_WAIT_FOR_EVER) {
-                state->cv.wait(guard);
-            } else if (timeoutNs == 0 ||
-                    state->cv.wait_for(guard, ns(timeoutNs)) == std::cv_status::timeout) {
-                return state->status;
+
```
bejado: nit: trailing whitepsace

```suggestion

```

filament/backend/src/metal/MetalHandles.mm
```diff
@@ -1422,24 +1424,31 @@ static void func(void* user) {
...
+        
+        FenceStatus result = FenceStatus::TIMEOUT_EXPIRED;
+        auto predicate = [&] {
+            if (state->status != FenceStatus::TIMEOUT_EXPIRED) {
+                result = state->status;
+                return true;
             }
+            return false;
+        };
+
```
bejado:
```suggestion

```

bejado (approved):

Issue #3135, opened 2020-09-29
Shader compilation error on some Android phones


NujnaH:
**Describe the bug**
We're trying to render a basic 3D model and while it works on most devices, some devices such as Moto G6 Play have an error like the following in its logs:

```
09-29 16:55:04.183  7486  7839 E Filament: COMPILE ERROR:
09-29 16:55:04.184  7486  7839 E Filament: Fragment shader compilation failed.
09-29 16:55:04.184  7486  7839 E Filament: ERROR: 0:90: 'Error' : Cannot offset into the vector
09-29 16:55:04.184  7486  7839 E Filament: ERROR: 0:90: 'Error' : Cannot offset into the vector
09-29 16:55:04.184  7486  7839 E Filament: ERROR: 0:90: '=' :  cannot convert from '3-component vector of float' to 'float'
09-29 16:55:04.184  7486  7839 E Filament: ERROR: 0:93: 'Error' : Cannot offset into the vector
09-29 16:55:04.184  7486  7839 E Filament: ERROR: 0:93: 'Error' : Cannot offset into the vector
09-29 16:55:04.184  7486  7839 E Filament: ERROR: 0:93: 'Error' : Cannot offset into the vector
09-29 16:55:04.184  7486  7839 E Filament: ERROR: 0:93: 'constructor' : too many arguments
09-29 16:55:04.184  7486  7839 E Filament: ERROR: 0:93: 'Error' : Cannot offset into the vector
09-29 16:55:04.184  7486  7839 E Filament: ERROR: 0:93: '*' :  w
09-29 16:55:04.184  7486  7839 E Filament: 1:   #version 300 es
09-29 16:55:04.185  7486  7839 E Filament: 2:   precision mediump float;
09-29 16:55:04.185  7486  7839 E Filament: 3:   precision mediump int;
09-29 16:55:04.185  7486  7839 E Filament: 4:   struct PostProcessInputs
09-29 16:55:04.185  7486  7839 E Filament: 5:   {
09-29 16:55:04.185  7486  7839 E Filament: 6:   vec4 color;
09-29 16:55:04.185  7486  7839 E Filament: 7:   };
[...]

PR #9903, opened 2026-04-15
backend: Propagate backend thread exceptions to the main thread.

### LINKED ISSUE #8197: Filament panics unrecoverably in background threads for potentially recoverable issues

pixelflinger:
Introduce a mechanism to catch exceptions thrown on the backend thread and
rethrow them on the main thread. This prevents deadlocks and allows the
application to handle fatal backend failures gracefully.

BUGS=[407545700]
Fixes #8197, #3135

### DIFF
--- a/android/common/JniUtils.h
+++ b/android/common/JniUtils.h
@@ -1,0 +1,123 @@
+/*
+ * Copyright (C) 2026 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef TNT_ANDROID_COMMON_JNIUTILS_H
+#define TNT_ANDROID_COMMON_JNIUTILS_H
+
+#include <jni.h>
+#include <exception>
+#include <type_traits>
+#include <utils/Panic.h>
+#include <utils/compiler.h>
+
+namespace filament {
+namespace android {
+
+#ifdef __EXCEPTIONS
+
+// Non-templated helpers implemented in JniUtils.cpp
+void wrapJniHelper(JNIEnv* env, void (*invoker)(void*), void* userData);
+void wrapJniBackendHelper(JNIEnv* env, void (*invoker)(void*), void* userData);
+
+// For JNI methods that return a value
+template<typename R, typename F>
+R wrapJni(JNIEnv* env, F const& f) {
+    if constexpr (std::is_void_v<R>) {
+        auto invoker = [](void* data) {
+            auto& f_ref = *reinterpret_cast<const F*>(data);
+            f_ref();
+        };
+        wrapJniHelper(env, invoker, (void*)&f);
+    } else {
+        struct Context {
+            const F& f;
+            R result;
+        };
+        Context ctx{ f, R{} };
+        auto invoker = [](void* data) {
+            auto& context = *reinterpret_cast<Context*>(data);
+            context.result = context.f();
+        };
+        wrapJniHelper(env, invoker, &ctx);
+        return ctx.result;
+    }
+}
+
+// Overload for JNI methods that return void
+template<typename F>
+inline void wrapJni(JNIEnv* env, F const& f) {
+    wrapJni<void, F>(env, f);
+}
+
+// For JNI methods that can return backend errors (mapped to java.lang.Error)
+template<typename R, typename F>
+R wrapJniBackend(JNIEnv* env, F const& f) {
+    if constexpr (std::is_void_v<R>) {
+        auto invoker = [](void* data) {
+            auto& f_ref = *reinterpret_cast<const F*>(data);
+            f_ref();
+        };
+        wrapJniBackendHelper(env, invoker, (void*)&f);
+    } else {
+        struct Context {
+            const F& f;
+            R result;
+        };
+        Context ctx{ f, R{} };
+        auto invoker = [](void* data) {
+            auto& context = *reinterpret_cast<Context*>(data);
+            context.result = context.f();
+        };
+        wrapJniBackendHelper(env, invoker, &ctx);
+        return ctx.result;
+    }
+}
+
+template<typename F>
+inline void wrapJniBackend(JNIEnv* env, F const& f) {
+    wrapJniBackend<void, F>(env, f);
+}
+
+#else
+
+// For JNI methods that return a value
+template<typename R, typename F>
+R wrapJni(JNIEnv* env, F const& f) {
+    return f();
+}
+
+// Overload for JNI methods that return void
+template<typename F>
+inline void wrapJni(JNIEnv* env, F const& f) {
+    f();
+}
+
+template<typename R, typename F>
+R wrapJniBackend(JNIEnv* env, F const& f) {
+    return f();
+}
+
+template<typename F>
+inline void wrapJniBackend(JNIEnv* env, F const& f) {
+    f();
+}
+
+#endif
+
+} // namespace android
+} // namespace filament
+
+#endif // TNT_ANDROID_COMMON_JNIUTILS_H
--- a/android/filament-android/src/main/cpp/Engine.cpp
+++ b/android/filament-android/src/main/cpp/Engine.cpp
@@ -511,6 +511,12 @@ Java_com_google_android_filament_Engine_nIsAutomaticInstancingEnabled(JNIEnv*, j
     return (jboolean)engine->isAutomaticInstancingEnabled();
 }
 
+extern "C" JNIEXPORT jboolean JNICALL
+Java_com_google_android_filament_Engine_nHasUnrecoverableFailure(JNIEnv*, jclass, jlong nativeEngine) {
+    Engine* engine = (Engine*) nativeEngine;
+    return (jboolean)engine->hasUnrecoverableFailure();
+}
+
 extern "C" JNIEXPORT jlong JNICALL
 Java_com_google_android_filament_Engine_nGetMaxStereoscopicEyes(JNIEnv*, jclass, jlong nativeEngine) {
     Engine* engine = (Engine*) nativeEngine;
--- a/android/filament-android/src/main/cpp/Engine.cpp
+++ b/android/filament-android/src/main/cpp/Engine.cpp
@@ -16,6 +16,8 @@
 
 #include <jni.h>
 
+#include <exception>
+
 #include <filament/Camera.h>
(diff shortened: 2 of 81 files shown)


filament/src/details/Fence.h
```diff
@@ -41,11 +41,15 @@ class FFence : public Fence {
...
 
+public:
+    static void setUnrecoverableError() noexcept;
+
 private:
     // We assume we don't have a lot of contention of fence and have all of them
     // share a single lock/condition
     static utils::Mutex sLock;
     static utils::Condition sCondition;
+    static bool sHasUnrecoverableError;
```
z3moon: Do we allow multiple Engine instances in the same process? If yes, this will affect other processes' behavior.

filament/src/details/Engine.cpp
```diff
@@ -1581,14 +1597,32 @@ bool FEngine::execute() {
...
+            if (UTILS_LIKELY(item.begin)) {
+                driver.execute(item.begin);
+                mCommandBufferQueue.releaseBuffer(item);
+            }
         }
+#ifdef __EXCEPTIONS
+    } catch (...) {
+        mCommandBufferQueue.setUnrecoverableException(std::current_exception());
+        mDriver->setUnrecoverableError();
+        FFence::setUnrecoverableError();
```
z3moon: when an exception is raised, do we need to `releaseBuffer` for the remaining buffers?

filament/backend/include/private/backend/CommandBufferQueue.h
```diff
@@ -75,6 +77,21 @@ class CommandBufferQueue {
...
 
+#ifdef __EXCEPTIONS
+    bool hasUnrecoverableError() const noexcept {
+        return mHasUnrecoverableError.load(std::memory_order_relaxed);
+    }
+    void setUnrecoverableException(std::exception_ptr e) noexcept {
+        mBackendException = e;
+        mHasUnrecoverableError.store(true, std::memory_order_relaxed);
+    }
+    void propagateBackendException() const;
```
z3moon, line 88: probably it's safer to use `memory_order_acquire / release` here for `mBackendException` to be visible to the other thread.

filament/backend/src/CommandBufferQueue.cpp
```diff
@@ -79,8 +80,22 @@ bool CommandBufferQueue::isExitRequested() const {
 }
 
 
+#ifdef __EXCEPTIONS
+void CommandBufferQueue::propagateBackendException() const {
+    if (UTILS_VERY_UNLIKELY(mHasUnrecoverableError.load(std::memory_order_relaxed))) {
```
z3moon: you could use `hasUnrecoverableError()` instead.

filament/src/details/Fence.h
pixelflinger: oh that's a very good catch! thanks.

filament/src/details/Engine.cpp
pixelflinger: good question, this deserves investigation.

filament/backend/include/private/backend/CommandBufferQueue.h
pixelflinger: yes, without it mBackendException wouldn't be necessarily visible.

filament/src/details/Fence.h
pixelflinger: done

filament/src/details/Engine.cpp
pixelflinger: done + similar problem in flush()

filament/backend/include/private/backend/CommandBufferQueue.h
pixelflinger: done

filament/backend/src/CommandBufferQueue.cpp
pixelflinger: done

filament/src/details/MaterialInstance.cpp
```diff
@@ -216,6 +216,24 @@ void FMaterialInstance::commit(FEngine& engine) const {
 }
 
 void FMaterialInstance::commit(FEngine::DriverApi& driver, UboManager* uboManager) const {
+    if (!mTextureParameters.empty()) {
```
poweifeng, line 219: why did you move this part?

libs/filamentapp/src/FilamentApp.cpp
```diff
@@ -262,7 +269,7 @@ void FilamentApp::run(const Config& config, SetupCallback setupCallback,
             cameraFocalLength = mCameraFocalLength;
             cameraNear = mCameraNear;
             cameraFar = mCameraFar;
-        }
+            }
```
poweifeng, line 272: nit: accidentally added space here

poweifeng (approved):
lgtm with nits and a question

filament/src/details/Engine.h
```diff
@@ -147,6 +148,14 @@ class FEngine : public Engine, public utils::FeatureFlagManager {
     using Epoch = clock::time_point;
     using duration = clock::duration;
 
+#ifdef __EXCEPTIONS
+    void propagateBackendException() const {
```
bejado, line 152: Would it be worthwhile to make this available to clients? That way they don't have to render a frame to check if there have been any errors.

bejado (changes requested):
The `noexcept` must be removed from `MetalDriver::execute` in order for Metal driver exceptions to be caught.

filament/src/details/MaterialInstance.cpp
pixelflinger: oh the idea was to do the checks early, before changing any state.

filament/src/details/Engine.h
pixelflinger: yes that's a good idea. Like a "hasException()".

android/filament-android/src/main/cpp/Engine.cpp
```diff
@@ -415,17 +459,21 @@ Java_com_google_android_filament_Engine_nIsValidSwapChain(JNIEnv*, jclass,
 }
 
 extern "C" JNIEXPORT jboolean JNICALL
-Java_com_google_android_filament_Engine_nFlushAndWait(JNIEnv*, jclass,
+Java_com_google_android_filament_Engine_nFlushAndWait(JNIEnv *env, jclass,
         jlong nativeEngine, jlong timeout) {
     Engine* engine = (Engine*) nativeEngine;
-    return engine->flushAndWait((uint64_t)timeout);
+    return wrapJni<jboolean>(env, std::function<jboolean()>([=]() {
```
z3moon: do we need an explicit std::function conversion here?

android/filament-android/src/main/cpp/Engine.cpp
```diff
@@ -415,17 +459,21 @@ Java_com_google_android_filament_Engine_nIsValidSwapChain(JNIEnv*, jclass,
...
+    }));
 }
 
 extern "C" JNIEXPORT void JNICALL
-Java_com_google_android_filament_Engine_nFlush(JNIEnv*, jclass,
+Java_com_google_android_filament_Engine_nFlush(JNIEnv *env, jclass,
         jlong nativeEngine) {
     Engine* engine = (Engine*) nativeEngine;
-    engine->flush();
+    wrapJni(env, std::function<void()>([=]() {
```
z3moon: here as well

android/filament-android/src/main/cpp/Engine.cpp
```diff
@@ -436,21 +484,32 @@ Java_com_google_android_filament_Engine_nIsPaused(JNIEnv*, jclass,
...
     JniCallback* jniCallback = JniCallback::make(env, handler, runnable);
-    engine->compile(
-            (backend::CompilerPriorityQueue) priority,
-            material, view,
-            (utils::tribool::value_t) shadowReceiver,
-            (utils::tribool::value_t) skinning,
-            jniCallback->getHandler(), [jniCallback](Material*){
-                JniCallback::postToJavaAndDestroy(jniCallback);
-            });
+    wrapJni(env, std::function<void()>([=]() {
```
z3moon: ditto.
not sure if this is to workaround a compiler error or something.

android/common/JniUtils.h
```diff
@@ -0,0 +1,117 @@
...
+        return R{};
+    } catch (const std::exception& e) {
+        jclass exClass = env->FindClass("java/lang/RuntimeException");
+        env->ThrowNew(exClass, e.what());
+        return R{};
+    }
+}
+
+// Overload for JNI methods that return void
+UTILS_NOINLINE inline void wrapJni(JNIEnv* env, std::function<void()> const& f) {
```
z3moon: `UTILS_NOINLINE inline` used together. Do we need to pick one?

z3moon (approved):
LGTM except for minor issues.

android/filament-android/src/main/cpp/Engine.cpp
pixelflinger: done

android/filament-android/src/main/cpp/Engine.cpp
pixelflinger: done

android/filament-android/src/main/cpp/Engine.cpp
pixelflinger: done

android/common/JniUtils.h
pixelflinger: I completely changed this wrapper to be a lot more efficient.

PR #9904, opened 2026-04-15
ppm: compile resolveDepth only for vk/metal

poweifeng:
Sampling from a multi-sampled texture is not allowed for webgpu. From the comments in the material, we only need this for vk and metal.

### DIFF
--- a/filament/CMakeLists.txt
+++ b/filament/CMakeLists.txt
@@ -442,7 +442,12 @@ foreach(mat_dir ${MATERIAL_DIRS})
         check_in_list(MATERIAL_VULKAN_METAL_SRCS ${mat_src} is_vk_metal)
         if (is_vk_metal)
             # Remove webgpu and related parameters
+            # The variant-filter=stereo is added in the root folder's CMakeLists.txt for webgpu
+            # Context is that webgpu does not support the stereo variant
             string(REPLACE "-a;webgpu;--variant-filter=stereo" "" MATC_CUSTOM_FLAGS "${MATC_CUSTOM_FLAGS}")
+
+            # Remove gl
+            string(REPLACE "-a;opengl" "" MATC_CUSTOM_FLAGS "${MATC_CUSTOM_FLAGS}")
 
             # Remove any empty args
             string(REPLACE ";;" ";" MATC_CUSTOM_FLAGS "${MATC_CUSTOM_FLAGS}")
--- a/filament/CMakeLists.txt
+++ b/filament/CMakeLists.txt
@@ -319,6 +319,12 @@ set(MATERIAL_MULTIVIEW_SRCS
         src/materials/skybox.mat
 )
 
+# This set of materials should only be compiled for vulkan or metal because they contain operations
+# not allowed on other platforms. (e.g. sampling from a multi-sampled texture).
+set(MATERIAL_VULKAN_METAL_SRCS
+        src/materials/resolveDepth.mat
+)
+
 # ==================================================================================================
 # Configuration
 # ==================================================================================================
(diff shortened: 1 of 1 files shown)


pixelflinger (changes requested):
If I understand this change correctly, the MATERIAL_VULKAN_METAL removes these materials from the list.

Could we do it differently? I would prefer that, in the contrary, they add the materials to the list for the vulkan/metal backend (and you would remove them from the main list).

I think this would be less confusing?

poweifeng:
> If I understand this change correctly, the MATERIAL_VULKAN_METAL removes these materials from the list.
> 
> Could we do it differently? I would prefer that, in the contrary, they add the materials to the list for the vulkan/metal backend (and you would remove them from the main list).
> 
> I think this would be less confusing?

It's not exactly removing anything.
But I did make some helper functions to make the logic a little clearer.  PTAL. Thanks!

filament/CMakeLists.txt
```diff
@@ -409,29 +436,30 @@ foreach(mat_dir ${MATERIAL_DIRS})
...
-        )
-        list(APPEND FILAMAT_FILES_FOR_GROUP ${output_path})
-        list(APPEND FILAMAT_TARGETS_FOR_GROUP ${output_path})
+        set(MATC_CUSTOM_FLAGS "${MATC_BASE_FLAGS}")
+        set(COMMENT_STR "Compiling material ${fullname}")
+
+        check_in_list(MATERIAL_VULKAN_METAL_SRCS ${mat_src} is_vk_metal)
+        if (is_vk_metal)
+            # Remove webgpu and related parameters
+            string(REPLACE "-a;webgpu;--variant-filter=stereo" "" MATC_CUSTOM_FLAGS "${MATC_CUSTOM_FLAGS}")
```
bejado, line 447: I'm a bit confused by this line. We're only removing `webgpu` here? What about `opengl`? This seems fragile.

poweifeng: Yeah, I thought removing gl as well. I can put that in. 

I think the `--variant-filter=stereo` is the fragile part.  But with the upcoming dawn update.  This can be removed as well.

poweifeng: the context is that `webgpu`/`tint` does not support the stereo variant because there is no support for clip_distance extension in the shader.  But that has changed in the newer releases.

bejado: > webgpu/tint does not support the stereo variant because there is no support for clip_distance extension in the shader

Gotcha. Maybe comment that next to this line.

filament/CMakeLists.txt
```diff
@@ -409,29 +436,30 @@ foreach(mat_dir ${MATERIAL_DIRS})
...
+
+            # Remove any empty args
+            string(REPLACE ";;" ";" MATC_CUSTOM_FLAGS "${MATC_CUSTOM_FLAGS}")
+            string(REGEX REPLACE "^;|;$" "" MATC_CUSTOM_FLAGS "${MATC_CUSTOM_FLAGS}")
```
bejado, line 454: Is this necessary? Seems "empty" args will just end up as extra spaces in the command line, which is probably fine. Or is there something else going wrong here?

poweifeng: according to my ai friend, yes they are.  So `;;` gets translated to the string `""` (empty string with quotes) that gets passed to the cli tool.  

So
```
matc;-a;opengl;;
``` 

will turn into

```
matc -a opengl ""
```

which is not what we want.

pixelflinger (approved):

z3moon (approved):

filament/CMakeLists.txt
bejado: Ah I see, thanks for the clarification!

bejado (approved):

PR #9905, opened 2026-04-15
Harden filamesh parser and LinearImage against malformed inputs

sharadboni:
## Summary

Hardens two attacker-reachable parsers against malformed input while fixing three latent correctness bugs that exist in `main` today.

### `libs/filameshio` — MeshReader

- **New `loadMeshFromBuffer` overload that accepts `dataSize`** and bounds-checks every read against the buffer end (magic, header, vertexSize, indexSize, parts array, materialCount, each material nameLength). The existing overloads are kept as thin wrappers forwarding with `dataSize = SIZE_MAX`, so external callers compile unchanged.
- **`loadMeshFromFile`** forwards the known file size through the new bounds-validated path (plus the pre-existing `fd < 0` guard).
- **Alignment / strict-aliasing fix**: `Header`, `CompressionHeader`, and each `Part` are now read via `memcpy` into a local instead of cast through the caller-supplied buffer, which may not be naturally aligned for `uint32_t`.
- **`materialCount` / `nameLength` read fix**: the original code wrote `uint32_t x = (uint32_t) *p;` — this dereferences a single byte and then advances by `sizeof(uint32_t)`, silently truncating any value ≥ 256 and breaking on big-endian hosts. Replaced with `memcpy` so all four bytes emitted by the writer (`tools/filamesh/src/MeshWriter.cpp:253-255`) are read.
- **`nameLength + 1` overflow fix**: promote to `size_t` before `+ 1`, otherwise `nameLength == UINT32_MAX` wraps to zero and the bounds check is bypassed.
- **Header `version` check**: reject unknown versions before interpreting fields whose layout could differ.
- **Integer-overflow checks on compressed buffer sizing** (`indexSize * indexCount`, `vertexSize * vertexCount`) and on `header.parts * sizeof(Part)` before pointer arithmetic.
- **`malloc` return checks** on the two compressed-decode paths; the scratch buffer is `free`'d if decoding fails instead of leaking.

### `libs/image` — LinearImage

- `LinearImage(width, height, channels)` previously computed `uint32_t nfloats = width * height * channels;` — silently truncating on values like 65536 × 65536 × 3 and producing an undersized allocation followed by out-of-bounds `memset` / pixel writes.
- Now saturates in two stages (`width * height` fits in `uint64_t`; the multiply by `channels` is checked against `UINT64_MAX / channels`), then validates against `SIZE_MAX / sizeof(float)` before allocation. Throws `std::runtime_error` on overflow, matching the convention already used in `libs/imageio` (`ImageDecoder.cpp`, `ImageEncoder.cpp`).

## Test plan

- [ ] `test_filamesh` (`NonInterleaved`, `Interleaved`) — updated to call the bounds-validated overload; should continue to pass.
- [ ] Craft a filamesh buffer with `parts = 0xFFFFFFFF` and verify the overflow check rejects it.
- [ ] Craft a filamesh buffer that claims `materialCount = 300` to confirm that, with the `memcpy` fix, all 300 material entries are parsed (old code would silently read 300 as 0x2C = 44 on little-endian).
- [ ] Craft a filamesh buffer with `nameLength = 0xFFFFFFFF` and verify the `size_t`-promoted bounds check rejects it.
- [ ] Truncated buffer (every payload field one byte short of the file) — each should return an empty `Mesh` instead of reading past the end.
- [ ] `LinearImage(65536, 65536, 3)` throws `std::runtime_error`.
- [ ] Existing `.filamesh` samples load unchanged via the backward-compat overload.

### DIFF
--- a/libs/filameshio/src/MeshReader.cpp
+++ b/libs/filameshio/src/MeshReader.cpp
@@ -32,11 +32,11 @@
 #include <utils/Log.h>
 #include <utils/Path.h>
 
-#include <string>
-#include <vector>
+#include <cstdint>
+#include <cstring>
 #include <map>
 #include <string>
-#include <climits>
+#include <vector>
 
 #include <fcntl.h>
 #if !defined(WIN32)
@@ -79,7 +79,7 @@ MeshReader::MaterialRegistry::~MaterialRegistry() {
 }
 
 // Default move construction
-MeshReader::MaterialRegistry::MaterialRegistry(MaterialRegistry&& rhs) 
+MeshReader::MaterialRegistry::MaterialRegistry(MaterialRegistry&& rhs)
     : mImpl(nullptr) {
     std::swap(mImpl, rhs.mImpl);
 }
@@ -181,7 +181,7 @@ MeshReader::Mesh MeshReader::loadMeshFromFile(filament::Engine* engine, const ut
         p += sizeof(MAGICID);
 
         if (!strncmp(MAGICID, magic, 8)) {
-            mesh = loadMeshFromBuffer(engine, data, nullptr, nullptr, materials);
+            mesh = loadMeshFromBuffer(engine, data, size, nullptr, nullptr, materials);
         }
 
         Fence::waitAndDestroy(engine->createFence());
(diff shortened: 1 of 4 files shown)


sharadboni:
@poweifeng Could you review this security fix? It adds bounds validation to the filamesh parser, fixes a uint32 overflow in LinearImage dimensions, and adds overflow checks in compressed mesh decompression.

romainguy:
See https://github.com/google/filament/pull/9910 for `MeshReader` fixes that conflict with these. The other PR has better validation of the data since it checks there are no reads outside of the bounds of the file.

sharadboni:
Thanks @romainguy — I took a look at #9910 and agree the `dataSize` approach is the right foundation. I have updated this PR to adopt a bounds-validated `loadMeshFromBuffer(data, dataSize, ...)` overload (old overloads kept as backward-compatible wrappers so external callers continue to compile).

I also took the opportunity to address three latent issues that neither PR was fixing, which I think make this version strictly stronger than #9910:

1. **`(uint32_t) *p` reads only one byte.** Both the existing code and #9910 keep lines like `uint32_t materialCount = (uint32_t) *p;` — that dereferences a single `uint8_t` and then advances by `sizeof(uint32_t)`. On little-endian it silently truncates any value ≥ 256; on big-endian it always reads 0. The writer at `tools/filamesh/src/MeshWriter.cpp:253-255` emits a full 4-byte `uint32_t`, so this is a latent correctness bug for `materialCount` > 255 and any filamesh file on a big-endian reader. Fixed via `memcpy`.

2. **`Header` / `Part` / `CompressionHeader` casts are strict-aliasing + alignment UB** on unaligned caller buffers (e.g. JS bindings, `mmap` + offset). Replaced with `memcpy` into a local.

3. **`nameLength + 1u` can wrap to zero.** In #9910 the bounds check is `if (size_t(end - p) < nameLength + 1u)`. Because `nameLength` is `uint32_t`, the `+ 1u` is `uint32_t` arithmetic, so `nameLength == UINT32_MAX` wraps to 0 and the check is vacuous. Fixed here by promoting to `size_t` first.

Plus: a `header.version` check, `malloc` null-return checks, and the `LinearImage` fix (which #9910 does not touch).

If you would prefer I land the `LinearImage` part as a separate PR or drop something in scope, happy to split — just let me know.

libs/filameshio/src/MeshReader.cpp
```diff
@@ -187,67 +192,146 @@ MeshReader::Mesh MeshReader::loadMeshFromFile(filament::Engine* engine, const ut
...
+            utils::slog.e << "Buffer too small for material name length." << utils::io::endl;
+            return {};
+        }
+        uint32_t nameLength;
+        memcpy(&nameLength, base + consumed, sizeof(uint32_t));
+        consumed += sizeof(uint32_t);
+
+        // Promote to size_t before adding 1 so that nameLength == UINT32_MAX
+        // cannot wrap to 0 and bypass the bounds check.
+        const size_t nameSpan = size_t(nameLength) + 1;
```
z3moon, line 297: On 32bit platforms, size_t is 32-bit. So this will overflow to 0 if `nameLength` is UINT32_MAX.

libs/filameshio/src/MeshReader.cpp
```diff
@@ -187,67 +192,146 @@ MeshReader::Mesh MeshReader::loadMeshFromFile(filament::Engine* engine, const ut
...
+        if (nameSpan > dataSize - consumed) {
+            utils::slog.e << "Invalid material name length." << utils::io::endl;
+            return {};
+        }
+        partsMaterial[i] = (const char*) (base + consumed);
```
z3moon, line 302: You just validated `nameSpan` to ensure it fits in the buffer and then used this assignment that ignores the value. This will call `strlen`, which will incur a security hole if the string is not nul-terminated.

libs/filameshio/src/MeshReader.cpp
```diff
@@ -187,67 +192,146 @@ MeshReader::Mesh MeshReader::loadMeshFromFile(filament::Engine* engine, const ut
...
-    if (header->flags & COMPRESSION) {
-        size_t indexSize = header->indexType == UI16 ? sizeof(uint16_t) : sizeof(uint32_t);
-        size_t indexCount = header->indexCount;
+    const size_t indicesSize = header.indexSize;
+    if (header.flags & COMPRESSION) {
+        size_t indexSize = header.indexType == UI16 ? sizeof(uint16_t) : sizeof(uint32_t);
+        size_t indexCount = header.indexCount;
+        if (indexCount > 0 && indexSize > SIZE_MAX / indexCount) {
+            utils::slog.e << "Index buffer size overflow." << utils::io::endl;
+            return {};
```
z3moon, line 323: What happens for the created `mesh.indexBuffer` if we return here?

z3moon (changes requested):

libs/filameshio/src/MeshReader.cpp
sharadboni: Good catch. The `size_t` cast was a no-op on 32-bit targets, so `UINT32_MAX + 1` still wraps there. Switched to `if (nameLength >= dataSize - consumed) return {};` before doing the `+ 1`, so the add only ever runs on a bounded value.

libs/filameshio/src/MeshReader.cpp
sharadboni: Right, the length check was pointless if the copy itself falls back to strlen. Changed to `partsMaterial[i].assign((const char*)(base + consumed), nameLength)` so it uses the validated length directly and never walks looking for a NUL.

libs/filameshio/src/MeshReader.cpp
sharadboni: It leaks. Same issue applies to the malloc/decode failure paths below it and to the vertex-buffer block further down. Fix was to move both compressed-size overflow checks above the `IndexBuffer::Builder` / `VertexBuffer::Builder` calls (so those rejections return before anything is built), and to call `engine->destroy(mesh.indexBuffer)` / `engine->destroy(mesh.vertexBuffer)` on the malloc and decode failure paths. Every early return from `loadMeshFromBuffer` should be leak-free now.

sharadboni:
@z3moon can you please look into the pr it got closed due to inactivity

PR #9909, opened 2026-04-16
vk: handle non-error cases for swapchain acquire

### LINKED ISSUE #9732: VulkanSwapChain::acquire returns early on VK_SUBOPTIMAL_KHR instead of treating it as successful acquire
In VulkanSwapChain::acquire, the current logic does:

call mPlatform->acquire(...)
if result != VK_SUCCESS, return early
This means VK_SUBOPTIMAL_KHR is treated as a failure, even though it is a valid successful acquire result in Vulkan (with a hint that swapchain is suboptimal).

poweifeng:
When acquiring an image in a swapchain, we could get VK_ERROR_OUT_OF_DATE_KHR or VK_SUBOPTIMAL_KHR for acceptable states like window resizing.  For those cases, instead of failing to acquire, we recreate the swapchain and try to acquire swapchain image again

Fixes #9732


bejado (approved):

filament/backend/src/vulkan/VulkanSwapChain.cpp
```diff
@@ -149,28 +155,48 @@ void VulkanSwapChain::present(DriverBase& driver) {
...
+    if (mPlatform->hasResized(swapChain)) {
+        recreateSwapchain();
     }
 
     VulkanPlatform::ImageSyncData imageSyncData;
-    VkResult const result = mPlatform->acquire(swapChain, &imageSyncData);
+    VkResult result = VK_NOT_READY;
+    size_t acquireTryCount = 0;
+
+    while (acquireTryCount++ < MAX_ACQUIRE_TRY_COUNT && result != VK_SUCCESS) {
```
rafadevai, line 184: why do we need to retry?  and also if you flushed/waited already, why do it again?

poweifeng: > why do we need to retry?

I'm just trying to account for the case where a user is drag-resizing a window, in which case, I imagine we might hit multiple `VK_SUBOPTIMAL_KHR` or `VK_ERROR_OUT_OF_DATE_KHR`.  But that's just my guess and not based on any particular platform.

> and also if you flushed/waited already, why do it again?

That's a good point, but `flush()` and `wait()` is essentially no-op when there is no buffer being recorded. So calling it multiple times is fine.

poweifeng: let me know if you prefer changes to either of the above.

filament/backend/src/vulkan/VulkanSwapChain.cpp
```diff
@@ -149,28 +155,48 @@ void VulkanSwapChain::present(DriverBase& driver) {
...
+void VulkanSwapChain::acquire(bool& recreated) {
     // It's ok to call acquire multiple times due to it being linked to Driver::makeCurrent().
     if (mAcquired) {
         return;
     }
 
-    // Check if the swapchain should be resized.
-    if ((resized = mPlatform->hasResized(swapChain))) {
+    auto recreateSwapchain = [&]() {
+        recreated = true;
```
z3moon, line 167: `recreated` is only updated to `true` throughout the code flow. do we need to set it to false as default at the beginning of this function?

poweifeng: sure, done

filament/backend/src/vulkan/VulkanSwapChain.cpp
rafadevai: So I think we should try it once and that's it.

In the case of out of date, it it fails, we just record, submit and wont present anything.
Will that break anything?

poweifeng: so I had to do more work, but I added some guards so that calls that concern the renderpass are skipped, because we cannot construct a valid renderpass when the acquire fails.

filament/backend/src/vulkan/VulkanSwapChain.cpp
```diff
@@ -149,41 +149,76 @@ void VulkanSwapChain::present(DriverBase& driver) {
...
-        return;
+        return { mAcquired, swapchainRecreated };
     }
 
-    // Check if the swapchain should be resized.
-    if ((resized = mPlatform->hasResized(swapChain))) {
+    // Used for cases where the backing swapchain needs to be recreated.
+    auto recreate = [&]() {
+        // Calling flush multiptle times is ok, since it's no-op if not recording.
         if (mFlushAndWaitOnResize) {
```
rafadevai: I think its better to just move this function inside the if, easier to read it.

filament/backend/src/vulkan/VulkanDriver.cpp
rafadevai, line 2214: what about here? in the case of skip, there's no need to do present but do we still have to be aware that things have to be submitted?

filament/backend/src/vulkan/VulkanSwapChain.cpp
poweifeng: done

filament/backend/src/vulkan/VulkanDriver.cpp
poweifeng: it'll be no-op since the swapchain was not acquired. Other commands will be submitted with the `flush` call, or on the next frame

rafadevai (approved):

PR #9912, opened 2026-04-17
Add support for custom outputs for unlit surface materials

bejado:
Unlit surface materials can now use `outputs` to define a single output that will bypass the typical color code path. This allows users to write directly `materialInputs.customOutput`.

More aggressive validation will be done in a follow-up PR.

BUGS=[490401325]


pixelflinger (approved):

pixelflinger (commented):
could we add a couple tests?

poweifeng (approved):

PR #9913, opened 2026-04-17
feat: Make C++ exceptions and RTTI configurable

pixelflinger:
Introduced the `FILAMENT_ENABLE_EXCEPTIONS` CMake option (default ON) to control both C++ exceptions and RTTI globally. This replaces previous hardcoded, platform-specific overrides and allows users to opt-out to save binary size.

Added a `-E` flag to `build.sh` to easily disable exceptions and RTTI during build configuration.

Removed hardcoded `-fno-exceptions` and `-fno-rtti` from `android/build.gradle` to allow CMake to control these flags uniformly.

Documented in `CMakeLists.txt` and `BUILDING.md` that the JNI library (Android/Java build) requires exceptions to be enabled. Users of Filament as a pure native library on Android can disable them to save space.

Maintained hardcoded `-fno-rtti` for WebGL and Cygwin due to specific platform constraints (e.g., `emscripten::val` usage).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -26,6 +26,20 @@ project(TNT)
 # ==================================================================================================
 option(FILAMENT_USE_EXTERNAL_GLES3 "Experimental: Compile Filament against OpenGL ES 3" OFF)
 
+# Note: If you intend to use the JNI library (Android/Java build), you need to have
+# exceptions enabled (the default). If you are using Filament on Android as a pure
+# native library and want to save space, you can disable them.
+# We disable exceptions on iOS by default. This fixes an availability error we see when using
+# std::visit and std::get, which are not supported on iOS 11.0 when exceptions are enabled.
+if (IOS)
+    set(FILAMENT_ENABLE_EXCEPTIONS_DEFAULT OFF)
+else()
+    set(FILAMENT_ENABLE_EXCEPTIONS_DEFAULT ON)
+endif()
+option(FILAMENT_ENABLE_EXCEPTIONS "Enable C++ exceptions" ${FILAMENT_ENABLE_EXCEPTIONS_DEFAULT})
+
+option(FILAMENT_ENABLE_RTTI "Enable C++ RTTI" OFF)
+
 option(FILAMENT_ENABLE_LTO "Enable link-time optimizations if supported by the compiler" OFF)
 option(FILAMENT_SKIP_SAMPLES "Don't build samples" OFF)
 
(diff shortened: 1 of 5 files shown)


pixelflinger:
The extra 290KB is due to the unwind tables (verified with zbloat).

bejado (approved):

CMakeLists.txt
```diff
@@ -435,30 +457,19 @@ if (NOT MSVC)
...
+if (ANDROID OR WEBGL)
+    # On Android and WebGL RELEASE builds, we omit unwind info to save space.
+    # Omitting unwind info prevents the generation of readable stack traces in crash reports on iOS.
+    set(CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_RELEASE} -fno-unwind-tables -fno-asynchronous-unwind-tables")
 endif()
 
-# Turn off exceptions on iOS debug as well. This fixes an availability error we see when using
-# std::visit, which is not supported on iOS 11.0 when exceptions are enabled.
-if (IOS)
-    set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -fno-exceptions")
```
poweifeng, line 453: where's this part that disables exception for ios?

CMakeLists.txt
```diff
@@ -435,30 +457,19 @@ if (NOT MSVC)
...
-    set(CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_RELEASE} -fno-exceptions -fno-rtti")
-    set(SPIRV_CROSS_EXCEPTIONS_TO_ASSERTIONS ON)
-
-    if (ANDROID OR WEBGL)
-        # Omitting unwind info prevents the generation of readable stack traces in crash reports on iOS
-        set(CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_RELEASE} -fno-unwind-tables -fno-asynchronous-unwind-tables")
-    endif()
+if (ANDROID OR WEBGL)
+    # On Android and WebGL RELEASE builds, we omit unwind info to save space.
+    # Omitting unwind info prevents the generation of readable stack traces in crash reports on iOS.
```
poweifeng: this comment seems unreleated

poweifeng:
btw the sizeguard failure can be bypassed by adding this

```
SIZEGUARD_BYPASS
```

in the commit message (not the PR message)

CMakeLists.txt
pixelflinger: oh, now you have to do it manually. That's why the default for iOS is off.

I did this because someone might want to compile with exceptions for iOS, and it works as long as you're not using iOS 11. In fact it worked locally for me, but failed only in CI.

So on iOS it's off by default. But if someone wants exceptions they can turn them on at compile time. If we updated  our lowest iOS version on the CI, I could have it on by default.

PR #9917, opened 2026-04-20
Implement grid-based world origin snapping in View

pixelflinger:
Implement a grid-based world origin snapping system in View to avoid per-frame transform updates in the future. This will allow to improve CPU performance and to enable future caching of acceleration structures like BVH.

The feature is protected by the 'view.enable_grid_based_world_origin' feature flag. The grid size can be set
manually via View::setGridSize or calculated automatically as 10% of the camera's far plane distance.

A 50% hysteresis ratio is applied to prevent rapid origin flipping near grid edges. 

Exposed the new API to Java and JavaScript bindings and added unit tests in filament_test.cpp.

BUGS=[504726278]


filament/src/details/View.cpp
```diff
@@ -529,27 +530,100 @@ void FView::prepareLighting(FEngine& engine, CameraInfo const& cameraInfo) noexc
...
+    mat3 rotation{ 1.0f };
 
     /*
      * Calculate all camera parameters needed to render this View for this frame.
      */
     FCamera const* const camera = mViewingCamera ? mViewingCamera : mCullingCamera;
+    double3 const cameraPosition = camera->getPosition();
+
+    // Internal policy controlled by feature flag and debug flags
     if (engine.debug.view.camera_at_origin) {
```
poweifeng, line 630: not sure if I have this right, but this public api is dependent on a debug flag?

filament/src/details/View.cpp
```diff
@@ -529,27 +530,100 @@ void FView::prepareLighting(FEngine& engine, CameraInfo const& cameraInfo) noexc
...
+        if (engine.features.view.enable_grid_based_world_origin) {
+            // This moves the camera to a snapped grid origin, improving floating point precision
+            // while avoiding per-frame transform updates for objects as long as the camera
+            // stays within the grid cell (plus hysteresis).
+            double gridSize = mGridSize;
+            if (gridSize <= 0.0) {
+                gridSize = calculateAutomaticGridSize(camera);
+            }
+            mEffectiveGridSize = gridSize;
+            const double hysteresisRatio = 0.5; // Automatic 50% hysteresis
```
poweifeng: nit constexpr

filament/src/details/View.cpp
```diff
@@ -529,27 +530,100 @@ void FView::prepareLighting(FEngine& engine, CameraInfo const& cameraInfo) noexc
...
+ * 
+ * camera: The camera to use for the calculation.
+ * Returns the calculated grid size.
+ */
+double FView::calculateAutomaticGridSize(const FCamera* camera) const noexcept {
+    double zf = camera->getCullingFar();
+    if (std::isinf(zf) || zf <= 0.0) {
+        // Fallback for infinite far plane or invalid values
+        // Use near plane as a baseline, assuming a reasonable ratio (e.g., 1000x)
+        return camera->getNear() * 1000.0;
```
poweifeng: does this assume that `getNear()` > 0 ?  I think Ortho projection can have negative near?

filament/src/details/View.cpp
```diff
@@ -529,27 +530,100 @@ void FView::prepareLighting(FEngine& engine, CameraInfo const& cameraInfo) noexc
...
+    }
+    // Default to 10% of the far plane distance
+    return zf * 0.1;
```
z3moon: why 10%?

filament/src/details/View.cpp
```diff
@@ -529,27 +530,100 @@ void FView::prepareLighting(FEngine& engine, CameraInfo const& cameraInfo) noexc
...
+    double3 newOrigin;
+    newOrigin.x = std::abs(cameraPosition.x - currentOrigin.x) > threshold ? targetOrigin.x : currentOrigin.x;
+    newOrigin.y = std::abs(cameraPosition.y - currentOrigin.y) > threshold ? targetOrigin.y : currentOrigin.y;
+    newOrigin.z = std::abs(cameraPosition.z - currentOrigin.z) > threshold ? targetOrigin.z : currentOrigin.z;
+
+    if (newOrigin != currentOrigin) {
+        mGridOrigin = newOrigin;
+        LOG(INFO) << "Grid snap: size=" << gridSize
+                      << ", origin=(" << newOrigin.x << ", " << newOrigin.y << ", " << newOrigin.z << ")";
+    }
```
z3moon: it might be okay to compare float points directly in this case. but usually it's an anti-pattern. is this code tested to see if this log doesn't spam when you hover near on the hysteresis boundary?

filament/src/details/View.cpp
pixelflinger: yeah, this is so that I can turn off the snapping entirely in gltf_viewer UI for debugging. 
this is one of these thing we did wrong in the past where the debug flag DEFAULT/NORMAL value is `true`

filament/src/details/View.cpp
pixelflinger: that's a good point. I  think camera->getNear() assumes a near plan in front of the camera. I'll double check.

filament/src/details/View.cpp
pixelflinger: it's just a heuristic that seemed reasonable. The actual value doesn't really matter in "automatic" mode. I was planning to tune this over time based on experiments.
It's possible a good value might actually be 10x the frustum. Unclear yet.
maybe this should also be a setting. But I think for now it doesn't matter too much.

filament/src/details/View.cpp
pixelflinger: oh yes, the LOG(INFO) is a leftover, I forgot to remove it.

The compare is intentional, but you're right that maybe using a small deadzone here could be advantageous.

pixelflinger:
@z3moon @poweifeng I think I addressed your comments. thanks.

filament/src/details/View.cpp
```diff
@@ -533,27 +533,132 @@ void FView::prepareLighting(FEngine& engine, CameraInfo const& cameraInfo) noexc
...
+            double newGridSize = mGridSize;
+            if (newGridSize <= 0.0) {
+                newGridSize = calculateAutomaticGridSize(camera);
+            }
+            
+            // Force snap if user manually changed grid size to a positive value
+            bool const forceSnap = (mGridSize > 0.0 && mGridSize != currentGridSize);
+            
+            constexpr double hysteresisRatio = 0.5; // Automatic 50% hysteresis
+            translation = -computeGridOrigin(cameraPosition, currentGridSize, newGridSize, hysteresisRatio, forceSnap);
```
poweifeng, line 654: Do you need to update `mEffectiveGridSize` with `newGridSize`?

filament/src/details/View.cpp
```diff
@@ -533,27 +533,132 @@ void FView::prepareLighting(FEngine& engine, CameraInfo const& cameraInfo) noexc
...
-        // very far from the origin, objects are still rendered and lit properly.
-        translation = -camera->getPosition();
+        if (engine.features.view.enable_grid_based_world_origin) {
+            // This moves the camera to a snapped grid origin, improving floating point precision
+            // while avoiding per-frame transform updates for objects as long as the camera
+            // stays within the grid cell (plus hysteresis).
+            double currentGridSize = mEffectiveGridSize;
+            if (currentGridSize <= 0.0) {
+                // First time initialization
+                currentGridSize = mGridSize;
```
poweifeng: nit: lines 638-641 is identifical to lines 645-648

Can we maybe have a anon func with a descriptive name for this small logic?  It'd make reading the code a bit easier.

filament/src/details/View.cpp
```diff
@@ -533,27 +533,132 @@ void FView::prepareLighting(FEngine& engine, CameraInfo const& cameraInfo) noexc
...
+                if (currentGridSize <= 0.0) {
+                    currentGridSize = calculateAutomaticGridSize(camera);
+                }
+                mEffectiveGridSize = currentGridSize;
+            }
+            
+            double newGridSize = mGridSize;
+            if (newGridSize <= 0.0) {
+                newGridSize = calculateAutomaticGridSize(camera);
```
z3moon: nit: this could  be called twice within this function.

z3moon (approved):

poweifeng (approved):

PR #9930, opened 2026-04-23
Metal: implement asynchronous resource loading

bejado:
This change implements asynchronous resource uploads for the Metal backend, allowing Filament to transfer data to GPU resources without blocking driver execution.

To achieve this, we use two threads: the **main driver thread** and a **background worker thread**. The Metal backend now creates a `JobQueue` to manage this background worker and offload heavy data transfers from the driver thread.

_Note: Depending on the engine's configuration, this might not actually use a dedicated background thread. If the engine is set to use an amortization strategy (or if the platform does not support threads), these tasks are instead processed in small batches during the driver's normal execution cycle._

When an upload command is received (such as `updateIndexBufferAsyncR` or `update3DImageAsyncR`):
1. The driver thread creates a Metal command buffer.
2. The worker thread writes copy commands into that command buffer.
3. The worker thread adds a completion handler to the command buffer to trigger the user's asynchronous callback.
4. The worker thread commits the command buffer to the GPU.

Destructors for these asynchronous resources are also pushed onto the worker thread. This prevents race conditions where a resource might be destroyed on the main thread while an offloaded upload operation is still actively pending on the worker thread.

**Risks and Limitations:**
The overall risk of this change is low for clients who don't use asynchronous resources, as the new code paths are isolated and only active when asynchronous modes are explicitly enabled. The primary risk when using the new asynchronous methods comes from their limited testing. While I successfully verified the implementation using the `helloasync` sample, more extensive testing is needed before before using asynchronous resource loading with the Metal backend in production.

BUGS=[506166350]


filament/backend/src/metal/MetalBufferPool.h
```diff
@@ -58,6 +58,7 @@ class MetalBumpAllocator {
     TrackedMetalBuffer mCurrentUploadBuffer = nil;
     size_t mHead = 0;
     size_t mCapacity;
+    std::mutex mMutex;
```
pixelflinger: Would be nice to document which attributes it protects.
e.g. looks like mCapacity doesn't need the lock (from code below)

filament/backend/src/metal/MetalBufferPool.mm
pixelflinger: This call is made with the lock held, and it doesn't seem like it needs to.

pixelflinger (changes requested):
This should probably have a bug number associated.
I'd like the checkin comment to:
- explain better what this does
- what the risks are (e.g. seems like risks are low if async feature not used)

Ideally there should be a backend test for this.

filament/backend/src/metal/MetalDriver.mm
```diff
@@ -1485,7 +1601,8 @@
 void MetalDriver::setVertexBufferObjectAsyncR(AsyncCallId jobId, Handle<HwVertexBuffer> vbh,
         uint32_t index, Handle<HwBufferObject> boh, CallbackHandler* handler,
         CallbackHandler::Callback const callback, void* user) {
-    // TODO: implement this.
+    setVertexBufferObject(vbh, index, boh);
+    scheduleCallback(handler, user, callback);
 }
```
z3moon, line 1606: jobId is not used here. Is this expected?

z3moon (commented):
have you tested this with `helloasync` desktop sample? the sample makes use of the async APIs to create/update textures and renderables (vertex & index buffers) asynchronously.

filament/backend/src/metal/MetalDriver.mm
```diff
@@ -538,10 +572,26 @@
...
+    auto& sc = mContext->sampleCountLookup;
+    samples = sc[std::min(MAX_SAMPLE_COUNT, samples)];
+
+    MetalTexture* texture = construct_handle<MetalTexture>(th, *mContext, target, levels, format,
+            samples, width, height, depth, usage, true);
+    mContext->textures.insert(texture);
+
+    texture->setLabel(tag);
+
+    DEBUG_LOG("createTextureAsyncR(th = %d, target = %s, levels = %d, format = ?, samples = %d, "
```
poweifeng, line 588: intended?

filament/backend/src/metal/MetalDriver.mm
```diff
@@ -1009,7 +1061,13 @@
 }
 
 void MetalDriver::destroyIndexBuffer(Handle<HwIndexBuffer> ibh) {
-    if (ibh) {
+    if (UTILS_UNLIKELY(!ibh)) {
+        return;
+    }
+    auto* ib = handle_cast<MetalIndexBuffer>(ibh);
+    if (ib->asynchronous) {
+        getJobQueue()->push([this, ibh]() mutable { destruct_handle<MetalIndexBuffer>(ibh); });
```
poweifeng, line 1069: just curious.  Is it necessary to destroy on the same thread it was created?

poweifeng (approved):
lgtm minus other's comments

filament/backend/src/metal/MetalDriver.mm
bejado: Yes, it's expected. For the Metal backend, I chose to leave `setVertexBufferObjectAsync` (as well as some other construction methods) on the driver thread since they are light and not doing any resource uploads.

filament/backend/src/metal/MetalDriver.mm
bejado: Yes, the `DEBUG_LOG` macro is off by default. This is just mirroring the log from `createTextureR` and is helpful when I need to do debugging related to textures.

filament/backend/src/metal/MetalDriver.mm
bejado: Yes! Otherwise the driver thread could destroy a resource while it is being uploaded to by the worker thread. So, we push the destruction to the worker thread to sequence it correctly.

bejado:
> have you tested this with `helloasync` desktop sample? the sample makes use of the async APIs to create/update textures and renderables (vertex & index buffers) asynchronously.

Yes I did and it appears to be working well! Thank you for that sample, it was very helpful when implementing this.

bejado:
> Ideally there should be a backend test for this.

I'll work on adding a backend test for async loading.

pixelflinger (approved):

z3moon (approved):

PR #9931, opened 2026-04-23
github: restore platform-specific status badges in README

poweifeng:
Consolidating the CI workflows into a single postsubmit.yml broke the individual platform status badges in the README, as GitHub Actions only supports status badges at the workflow level, not the job level.

Added shadow workflows (e.g., status-android.yml) that trigger on the completion of the main "Postsubmit CI" workflow. These scripts query the GitHub Actions API to determine the success or failure of their respective job (e.g., build-android) and reflect that status. Updated the README to point to these new workflow badges.


bejado (approved):
LGTM, but maybe we don't need status badges at all? I worry this is adding more complexity to our workflows. That said, we might as well try it.. if this gets too unwieldy then I say just remove the status badges.

pixelflinger:
> LGTM, but maybe we don't need status badges at all? I worry this is adding more complexity to our workflows. That said, we might as well try it.. if this gets too unwieldy then I say just remove the status badges.

I would love to keep them.

z3moon (approved):

PR #9932, opened 2026-04-23
webgpu: implement push constants with immediates

poweifeng:
- Add test for skinning for gl, vk, and webgpu

RDIFF_ACCEPT_NEW_GOLDENS

### DIFF
--- a/libs/viewer/src/ViewerGui.cpp
+++ b/libs/viewer/src/ViewerGui.cpp
@@ -578,7 +578,10 @@ void ViewerGui::applyAnimation(double currentTime, FilamentInstance* instance) {
         mCurrentStartTime = currentTime;
         mResetAnimation = false;
     }
-    const double elapsedSeconds = currentTime - mCurrentStartTime;
+    double elapsedSeconds = currentTime - mCurrentStartTime;
+    if (mSettings.animation.time >= 0.0f) {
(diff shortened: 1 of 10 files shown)


bejado (approved):

libs/viewer/src/ViewerGui.cpp
```diff
@@ -578,7 +578,10 @@ void ViewerGui::applyAnimation(double currentTime, FilamentInstance* instance) {
         mCurrentStartTime = currentTime;
         mResetAnimation = false;
     }
-    const double elapsedSeconds = currentTime - mCurrentStartTime;
+    double elapsedSeconds = currentTime - mCurrentStartTime;
+    if (mSettings.animation.time >= 0.0f) {
+        elapsedSeconds = mSettings.animation.time;
+    }
```
pixelflinger, line 584: seems unrelated, what is this trying to do?

pixelflinger (approved):

libs/viewer/src/ViewerGui.cpp
poweifeng: This sets the animation at a particular time, instead of relying on the frame time.

poweifeng: so this is fixing a bug that's needed for the test

PR #9933, opened 2026-04-23
webgpu: Clean up spec constants and fix validation for unused attributes

poweifeng:
- WebGPUProgram.cpp: Remove legacy replaceSpecConstants string injection since WebGPU handles specialization constants natively.
- WebGPUVertexBufferInfo.cpp: Pad unused vertex attributes safely into slot 0 using 4-byte formats to satisfy WebGPU stride validation.
- CodeGenerator.cpp: Remove legacy fallback generating 'const' structs for specialization constants.
- ShaderGenerator.cpp: Unset unused skinning/morphing attributes to prevent WebGPU validation crashes.
- dawn/tnt/CMakeLists.txt: Fix JNI linker error by dropping missing Tint AST reader libraries and adding tint_utils_system.

This commit fixes combination use of ubershader and skinning.


filament/backend/src/webgpu/WebGPUVertexBufferInfo.cpp
```diff
@@ -297,9 +297,55 @@ void createBufferLayoutsBindingSlotsAndAttributeInfos(AttributeArray const& attr
...
+     */
+    if (currentWebGPUSlotIndex > 0) {
+        for (uint32_t attributeIndex = 0; attributeIndex < attributes.size(); ++attributeIndex) {
+            if (attributes[attributeIndex].buffer == Attribute::BUFFER_UNUSED) {
+                // Determine base type to satisfy WebGPU validation without exceeding stride.
+                // Float32 and Uint32 are both 4 bytes, fitting almost any stride >= 4.
+                wgpu::VertexFormat padFormat = wgpu::VertexFormat::Float32;
+                if (attributeIndex == 5 || // BONE_INDICES
+                    attributeIndex == 8 || // CUSTOM0
+                    attributeIndex == 9) { // CUSTOM1
```
z3moon: Maybe we should use VertexAttribute::BONE_INDICES instead?

poweifeng: these are not accessible to the backend unfortunately.  

This is a definite hack.

poweifeng: I removed the attributeIndex matching part.  Should be more generic without leaking filament-side ATTRIBUTES now.

pixelflinger (approved):

PR #9935, opened 2026-04-23
Implement batch callback system for scene cache incremental updates.

pixelflinger:
This change adds a batch callback mechanism to track changes in component managers and the entity manager, avoiding full cache rebuilds in the future scene cache implementation.

1. SingleInstanceComponentManager:
- Added SingleInstanceComponentManagerBase to handle callbacks without template bloat.
- Implemented batch delivery using Slice<const Entity> and a fixed-size array of 16 for dirty entities to avoid heap allocations.
- Added compile-time option for sorted insertion (default to linear).

2. Manager Integration:
- RenderableManager: Added notifications to setters.
- LightManager: Added notifications to setters
- TransformManager: Handled transactions by deferring notifications until commit, and supported hierarchical updates in children.

3. EntityManager:
- Added a new batching callback API for entity destruction.
- Used std::function for callbacks to allow thread-safe copying of the callback list outside the lock.
- Simple accumulation in loop and flush when full, no bypass.

4. Tests:
- Added formal unit tests for all managers and entity manager callbacks.

### DIFF
--- a/filament/test/filament_test.cpp
+++ b/filament/test/filament_test.cpp
@@ -36,6 +36,9 @@
 #include <private/filament/UibStructs.h>
 #include <private/backend/BackendUtils.h>
 
+#include <utils/Invocable.h>
+#include <utils/Slice.h>
+
 #include "Allocators.h"
 #include "details/Material.h"
 #include "details/Camera.h"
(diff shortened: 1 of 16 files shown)


libs/utils/src/EntityManagerImpl.h
```diff
@@ -142,6 +159,32 @@ class UTILS_PRIVATE EntityManagerImpl : public EntityManager {
...
+        std::lock_guard<Mutex> const lock(mListenerLock);
+        mChangeCallbacks.erase(
+                std::remove_if(mChangeCallbacks.begin(), mChangeCallbacks.end(),
+                        [token](auto const& info) { return info.token == token; }),
+                mChangeCallbacks.end());
+    }
+
+    void flushNotifications() noexcept {
+        std::unique_lock<Mutex> lock(mFreeListLock);
+        if (mDirtyCount > 0) {
```
poweifeng, line 178: nit: can we add a sanity assert here that `mDirtyCount < MAX_DIRTY_COUNT`?

libs/utils/src/EntityManagerImpl.h
```diff
@@ -180,6 +241,16 @@ class UTILS_PRIVATE EntityManagerImpl : public EntityManager {
...
+    struct CallbackInfo {
+        void const* token;
+        ChangeCallback callback;
+    };
+    std::vector<CallbackInfo> mChangeCallbacks;
+
+    static constexpr size_t MAX_DIRTY_COUNT = 16;
+    Entity mDirtyEntities[MAX_DIRTY_COUNT];
+    size_t mDirtyCount = 0;
+
```
poweifeng, line 255: same exact definition in SingleInstanceComponentManager; is this intentional or can this be abstracted into a class?

poweifeng (approved):

libs/utils/src/EntityManagerImpl.h
pixelflinger: yeah it could, but they're fundamentally different things, so I thought it would better to not add a dependency between them, just for something trivial (and that might change in the future).

filament/test/filament_test.cpp
```diff
@@ -301,6 +304,127 @@ TEST(FilamentTest, TransformManager) {
     EXPECT_EQ(c, tcm.getChildCount(newParent));
 }
 
+TEST(FilamentTest, RenderableManagerCallback) {
+    FEngine* engine = downcast(Engine::create());
```
pixelflinger:
```suggestion
    FEngine* engine = downcast(Engine::create(Engine::Backend::NOOP));
```

filament/test/filament_test.cpp
```diff
@@ -301,6 +304,127 @@ TEST(FilamentTest, TransformManager) {
...
+    rm.flushNotifications();
+    EXPECT_EQ(callbackCount, 1);
+    
+    rm.unregisterChangeCallback(&rm);
+    em.destroy(e);
+    Engine::destroy((Engine**)&engine);
+}
+
+TEST(FilamentTest, LightManagerCallback) {
+    FEngine* engine = downcast(Engine::create());
```
pixelflinger:
```suggestion
    FEngine* engine = downcast(Engine::create(Engine::Backend::NOOP));
```

filament/test/filament_test.cpp
```diff
@@ -301,6 +304,127 @@ TEST(FilamentTest, TransformManager) {
...
+    lm.flushNotifications();
+    EXPECT_EQ(callbackCount, 1);
+    
+    lm.unregisterChangeCallback(&lm);
+    em.destroy(e);
+    Engine::destroy((Engine**)&engine);
+}
+
+TEST(FilamentTest, TransformManagerCallback) {
+    FEngine* engine = downcast(Engine::create());
```
pixelflinger:
```suggestion
    FEngine* engine = downcast(Engine::create(Engine::Backend::NOOP));
```

libs/utils/src/EntityManagerImpl.h
```diff
@@ -142,6 +159,32 @@ class UTILS_PRIVATE EntityManagerImpl : public EntityManager {
...
+                std::remove_if(mChangeCallbacks.begin(), mChangeCallbacks.end(),
+                        [token](auto const& info) { return info.token == token; }),
+                mChangeCallbacks.end());
+    }
+
+    void flushNotifications() noexcept {
+        std::unique_lock<Mutex> lock(mFreeListLock);
+        if (mDirtyCount > 0) {
+            Entity localBuffer[MAX_DIRTY_COUNT];
+            std::copy(mDirtyEntities, mDirtyEntities + mDirtyCount, localBuffer);
```
pixelflinger, line 181:
```suggestion
            assert_invariant(mDirtyCount <= MAX_DIRTY_COUNT);
            std::copy(mDirtyEntities, mDirtyEntities + mDirtyCount, localBuffer);
```

libs/utils/src/EntityManagerImpl.h
```diff
@@ -20,25 +20,31 @@
 #include <utils/EntityManager.h>
 
 #include <utils/compiler.h>
-#include <utils/Entity.h>
-#include <utils/Mutex.h>
 #include <utils/CallStack.h>
```
pixelflinger, line 24:
```suggestion
#include <utils/debug.h>
#include <utils/CallStack.h>
```

libs/utils/src/SingleInstanceComponentManager.cpp
```diff
@@ -0,0 +1,74 @@
...
+        flushNotifications();
+    }
+}
+
+void SingleInstanceComponentManagerBase::flushNotifications() noexcept {
+    if (mDirtyCount > 0) {
+        Slice<const Entity> const slice(mDirtyEntities, mDirtyCount);
+        for (auto const& [token, callback] : mChangeCallbacks) {
+            callback(slice);
+        }
```
z3moon, line 69: mChangeCallbacks could be corrupted if `unregisterChangeCallback` is called within the `callback(slice)`. So maybe iterating a temporary vector would be more ideal.

e.g.,
```
auto changeCallbacks = std::move(mChangeCallbacks);
for (auto ... : changeCallbacks) { ... }
```

libs/utils/include/utils/SingleInstanceComponentManager.h
```diff
@@ -33,6 +38,66 @@ namespace utils {
 
 class EntityManager;
 
+/**
+ * Base class for SingleInstanceComponentManager to handle callbacks without template bloat.
+ */
+class UTILS_PUBLIC SingleInstanceComponentManagerBase {
+public:
+    using ChangeCallback = Invocable<void(Slice<const Entity>)>;
```
z3moon, line 46: EntityManager's ChangeCallback uses std::function, which is different than what's used here. Do we need to unify the type? Or is this intended?

pixelflinger: this was intended, for the same reason you ran into something similar. It's because the EntityManager one needs to be thread-safe.

libs/utils/src/SingleInstanceComponentManager.cpp
pixelflinger: good point. I will address it in another PR, since I committed before I saw this comment. thanks!

PR #9938, opened 2026-04-27
mesa: update mesa version and ensure linux build

poweifeng:
- Update mesa version to 25.0.3 so that it builds with clang-17.
 - Make sure that abseil checks do not block compilation (this was due to unknown incompatiblity that appeared when trying to build abseil on linux and clang-17).


bejado (approved):
Nice, thank you!

PR #9943, opened 2026-04-30
osmesa: fix race in terminate()

poweifeng:


filament/backend/src/opengl/platforms/PlatformOSMesa.cpp
```diff
@@ -145,10 +145,13 @@ Driver* PlatformOSMesa::createDriver(void* sharedGLContext,
 void PlatformOSMesa::terminate() noexcept {
     OSMesaAPI* api = (OSMesaAPI*) mOsMesaApi;
 
-    for (auto& it: mAdditionalContexts) {
-        api->fOSMesaDestroyContext(it.second.context);
+    {
+        std::lock_guard<std::shared_mutex> lock(mAdditionalContextsLock);
+        for (auto& it: mAdditionalContexts) {
+            api->fOSMesaDestroyContext(it.second.context);
```
pixelflinger: it it safe to call this API with the lock held?
it seems like the lock should only be protecting mAdditionalContexts?

poweifeng: I do think it is safe, but to be extra safe, I moved it outside of the lock.

pixelflinger (changes requested):

pixelflinger (approved):

bejado (approved):

bejado:
Sorry about the race condition, thanks for investigating!

PR #9944, opened 2026-04-30
Fix push constant CString storing garbage past null terminator

bejado:
Fixed a subtle bug in `MaterialDefinition::processPushConstants` where push constant names were storing data past the null terminating character.

The issue occurred because the local buffer `char buf[MAX_NAME_LEN]` is a fixed-size array. When passed to the `CString(buf)` constructor, it matched the `StringLiteral` template overload `(const char(&)[N])` instead of decaying to a pointer. This caused `CString` to unconditionally copy N - 1 (59) bytes, including garbage data after the null terminator.

The fix explicitly passes the string length using `strlen(buf)` to ensure only the valid characters are copied.


poweifeng (approved):
This is really subtle.  

Was it causing problem?

bejado:
> This is really subtle.
> 
> Was it causing problem?

No, not in any production scenario. I am running a local experiment where I need to hash the names of the push constants. Passing the full `CString` with garbage data to the hash function obviously gave different hashes each run.

pixelflinger (approved):

PR #9945, opened 2026-05-01
utils: Secure EntityManager and hide implementation details

pixelflinger:
Move the `mGens` array and `isAlive()` implementation from `EntityManager` to `EntityManagerImpl` to hide implementation details and maintain a clean public header. Make `isAlive()` strictly thread-safe by acquiring the `mFreeListLock`.

isAlive() was relying on undefined behavior, and the comment about the memory barrier being provided by the lock was misleading (wrong) because the lock only provided the "release", but isAlive() didn't have an "acquire". 

Better safe than sorry, we now have to acquire the lock to test isAlive().

### DIFF
--- a/libs/utils/src/EntityManagerImpl.h
+++ b/libs/utils/src/EntityManagerImpl.h
@@ -40,6 +40,7 @@
 #include <deque>
 #include <memory>
 #include <mutex> // for std::lock_guard
+#include <new>
 #include <utility>
 #include <vector>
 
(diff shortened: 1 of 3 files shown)


libs/utils/src/EntityManagerImpl.h
```diff
@@ -98,31 +122,27 @@ class UTILS_PRIVATE EntityManagerImpl : public EntityManager {
...
+        std::unique_lock lock(mFreeListLock);
         for (size_t i = 0; i < n; i++) {
             if (!entities[i]) {
                 // behave like free(), ok to free null Entity.
                 continue;
             }
 
             // it's an error to delete an Entity twice...
-            assert(isAlive(entities[i]));
+            bool const isAlive = getGeneration(entities[i]) == mGens[getIndex(entities[i])];
```
romainguy, line 137: I assume you don't call `isAlive()` here to not acquire the lock in the loop? You should probably leave a comment about this.

pixelflinger: yes, otherwise it would deadlock, since isAlive() now acquires the lock.

poweifeng (approved):

bejado (approved):

PR #9947, opened 2026-05-04
utils: Implement high-performance sparse bitsets for ECS indices.

pixelflinger:
Introduce PagedArenaBitset to the utils library.
This implementation is optimized for handling large, sparse integer domains (up to 2^27) and is specifically tailored for Entity Component System (ECS) workloads.

1. PagedArenaBitset:
- 4-level hierarchical design (Master/Summary Masks, Directory, Arena).
- "Trust the Mask" architecture enables ultra-fast clears (<20ns).
- Arena-based storage ensures cache-coherent page access.
- Optimized multi-way intersections and defragmentation support.
- Strict move semantics to prevent accidental expensive copies.

2. Testing and Benchmarking:
- Comprehensive benchmarks comparing performance across different bit densities and operations.
- Full unit test coverage for the implementation.

### DIFF
--- a/libs/utils/include/utils/PagedArenaBitset.h
+++ b/libs/utils/include/utils/PagedArenaBitset.h
@@ -1,0 +1,540 @@
+/*
+ * Copyright (C) 2026 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef TNT_UTILS_PAGEARENABITSET_H
+#define TNT_UTILS_PAGEARENABITSET_H
+
+#include <utils/compiler.h>
+#include <utils/Slice.h>
+#include <utils/algorithm.h>
+
+#include <array>
+#include <cstddef>
+#include <cstdint>
+#include <cassert>
+#include <type_traits>
+#include <vector>
+
+namespace utils {
+
+/**
+ * @class PagedArenaBitset
+ * @brief A highly optimized, cache-coherent sparse bitset tailored for Generational ECS indices.
+ *
+ * @details
+ * This structure implements a 4-level hierarchical bitset designed to handle large,
+ * sparse integer domains (e.g., 2^27) where bits are densely clustered in "islands".
+ * It is specifically optimized for Entity Component Systems where Entity IDs combine a
+ * monotonically increasing generation value with a dense, recyclable base index.
+ *
+ * ### Hierarchical Memory Architecture
+ * @code
+ * ENTITY ID [27 bits]
+ * [ Master Word | Master Bit | Summary Bit | Word Idx | Bit Index ]
+ * [   3 bits    |   6 bits   |   6 bits    |  6 bits  |   6 bits  ]
+ * |             |            |             |          |
+ * |             |            |             |          +->  [Level 3: Bit position in u64]
+ * |             |            |             +------------>  [Level 3: Word position in Page]
+ * |             |            +-------------------------->  [Level 1: Page presence in Summary]
+ * |             +--------------------------------------->  [Level 0: Summary Word presence]
+ * +------------------------------------------------------> [Level 0: Master Word selection]
+ *
+ * VISUAL TOPOLOGY:
+ * +-------------------------------------------------------+
+ * | [Level 0] MASTER MASK (8 x u64)                       | <-- High-level Skip (512 bits)
+ * | [ 11110000... ]                                       |
+ * +-----------|-------------------------------------------+
+ *             v
+ * +-------------------------------------------------------+
+ * | [Level 1] SUMMARY MASK (512 x u64)                    | <-- Mid-level Skip (4 KB)
+ * | [ ...00101100... ]                                    |
+ * +-----------|-------------------------------------------+
+ *             v
+ * +-------------------------------------------------------+
+ * | [Level 2] DIRECTORY (32,768 x u16)                    | <-- Logical -> Physical (64 KB)
+ * | [ 0 | 5 | X | 2 | X | X | 1 | ... ] (X = Gated/Dirty) |
+ * +-----------|-------------------------------------------+
+ *             v
+ * +-------------------------------------------------------+
+ * | [Level 3] ARENA (Contiguous Pages)                    | <-- Dense Storage
+ * | +---------------------------------------------------+ |
+ * | | Page N:                                           | |
+ * | | [ activeWordsMask ] (1 x u64)                     | | <-- Word-level Skip
+ * | | [ words          ] (64 x u64)                     | |
+ * | +---------------------------------------------------+ |
+ * +-------------------------------------------------------+
+ * @endcode
+ *
+ * ### Architecture (The "Trust the Mask" Design)
+ * - **Level 0 (Master Mask):** A 512-bit mask where each bit maps 1:1 to a single word of the
+ * Summary Mask. Enables ultra-fast O(1) skipping of 256k-bit empty generational gaps.
+ * - **Level 1 (Summary Mask):** A 4KB mask where each bit flags the allocation of a single page.
+ * - **Level 2 (Directory):** A 64KB array mapping logical page indices to physical arena indices.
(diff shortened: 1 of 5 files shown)


libs/utils/include/utils/PagedArenaBitset.h
```diff
@@ -0,0 +1,541 @@
...
+    void extractTo(std::vector<uint32_t>& outBuffer) const;
+
+    /**
+     * @brief Iterates over all set bits in ascending numerical order.
+     * @tparam Func A callable type accepting a uint32_t.
+     * @param callback The function to execute for each set bit.
+     */
+    template <typename Func>
+    void forEachSetBit(Func&& callback) const {
+        for (uint32_t m = 0; m < MASTER_WORDS; ++m) {
```
poweifeng, line 391: can this use `mMinMaskWord` and `mMaxMaskWord`? 
Same in other methods.

libs/utils/include/utils/PagedArenaBitset.h
```diff
@@ -0,0 +1,541 @@
...
+    static PagedArenaBitset& intersect(PagedArenaBitset* out,
+            const PagedArenaBitset& first,
+            const PagedArenaBitset& second,
+            const PagedArenaBitset& third,
+            const Rest&... rest) {
+        static_assert((std::is_same_v<PagedArenaBitset, Rest> && ...),
+                "All arguments to intersect must be PagedArenaBitset");
+
+        std::array<const PagedArenaBitset*, sizeof...(rest) + 3> inputs = {&first, &second, &third, &rest...};
+        return intersectSpan(out, inputs);
```
poweifeng: passing it to `std::span` will lose the static size information when working std::array, disabling loop unrolling

libs/utils/include/utils/PagedArenaBitset.h
pixelflinger: Interesting point. It was written this way so the compiler would see a fixed size loop and could potentially unroll it. But maybe it's worth benchmarking.

libs/utils/include/utils/PagedArenaBitset.h
pixelflinger: Ah yes. That correct. That is actually the desired behavior. This should have been better documented. 

The code is written in terms of a generic COLLECTION so that it could be instantiated either as a dynamic loop (current choice) or a static loop (would have to be moved into the header file).

Invoking it with std::span causes it to collapse to the dynamic loop version.

libs/utils/include/utils/PagedArenaBitset.h
```diff
@@ -0,0 +1,541 @@
...
+ * |         |          |          |
+ * |         |          |          +--------> [Level 3: Arena Word (Gated by Page Mask)]
+ * |         |          +-------------------> [Level 2: Directory Lookup]
+ * |         +------------------------------> [Level 1: Summary Mask Bit]
+ * +----------------------------------------> [Level 0: Master Mask Bit]
+ *
+ * VISUAL TOPOLOGY:
+ * +-------------------------------------------------------+
+ * | [Level 0] MASTER MASK (8 x u64)                       | <-- High-level Skip
+ * | [ 11110000... ]                                       |
```
z3moon: nit: this comment may be misleading if the Master is 9 bits. This illustrates 8 bits. (same error applies below)

libs/utils/include/utils/PagedArenaBitset.h
```diff
@@ -0,0 +1,541 @@
...
+ * @details
+ * This structure implements a 4-level hierarchical bitset designed to handle large,
+ * sparse integer domains (e.g., 2^27) where bits are densely clustered in "islands".
+ * It is specifically optimized for Entity Component Systems where Entity IDs combine a
+ * monotonically increasing generation value with a dense, recyclable base index.
+ *
+ * ### Hierarchical Memory Architecture
+ * @code
+ * ENTITY ID [27 bits]
+ * [ Master | Summary | Page Idx | Word Idx | Bit ]
```
z3moon: looks like we have only 4 segments, not 5?

libs/utils/src/PagedArenaBitset.cpp
```diff
@@ -0,0 +1,1370 @@
...
+    for (uint32_t m = 0; m < MASTER_WORDS; ++m) {
+        uint64_t myMaster = mMasterMask[m];
+
+        // Level 1: Master Directory Traversal
+        while (myMaster) {
+            int const maskBit = std::countr_zero(myMaster);
+            myMaster &= myMaster - 1; // Clear lowest set bit to advance loop
+
+            // Compute index into Level 2 (Summary Masks)
+            uint32_t const maskIdx = (m << 6) | maskBit;
```
z3moon: maybe use WORD_SHIFT instead?

libs/utils/src/PagedArenaBitset.cpp
```diff
@@ -0,0 +1,1370 @@
...
+    }
+
+    for (uint32_t m = 0; m < MASTER_WORDS; ++m) {
+        uint64_t overlapMaster = mMasterMask[m] & other.mMasterMask[m];
+
+        while (overlapMaster) {
+            int const maskBit = std::countr_zero(overlapMaster);
+            overlapMaster &= overlapMaster - 1;
+
+            uint32_t const maskIdx = (m << 6) | maskBit;
```
z3moon: ditto

libs/utils/src/PagedArenaBitset.cpp
```diff
@@ -0,0 +1,1370 @@
...
+    for (uint32_t m = 0; m < MASTER_WORDS; ++m) {
+        uint64_t otherMaster = other.mMasterMask[m];
+        if (!otherMaster) continue;
+
+        // Note: Don't update mMasterMask[m] yet.
+
+        while (otherMaster) {
+            int const maskBit = std::countr_zero(otherMaster);
+            otherMaster &= otherMaster - 1;
+            uint32_t const maskIdx = (m << 6) | maskBit;
```
z3moon: ditto, and more in this file

z3moon (approved):

z3moon:
Great addition!

libs/utils/include/utils/PagedArenaBitset.h
poweifeng: I think the point is that `intersect` is generic, but `intersectSpan` is not, so that you would *never* get the static unrolling benefit because `intersect` always calls `intersectSpan`.

libs/utils/src/PagedArenaBitset.cpp
```diff
@@ -0,0 +1,1370 @@
...
+
+        while (subL0) {
+            int const maskBit = std::countr_zero(subL0);
+            subL0 &= subL0 - 1;
+            uint32_t const maskIdx = (m << 6) | maskBit;
+
+            uint64_t subL1 = subset.mSummaryMask[maskIdx];
+            uint64_t const superL1 = superset.mSummaryMask[maskIdx];
+
+            if ((subL1 & ~superL1) != 0) return false;
```
poweifeng: there's a subtle bug here.  Because ghost pages are allowed (pages where the mask is 1, but content is all zero).  Even if `subL1` has a bit set where `superL1` doesn't, that bit set could be pointing to a page of all zeros.  So `subset` could still be a subset of `superset`.

libs/utils/include/utils/PagedArenaBitset.h
pixelflinger: in fact those are not used anymore, that's a good catch. I'll remove them.

libs/utils/include/utils/PagedArenaBitset.h
pixelflinger: yes you're correct. I wrote the code this way so that if I change my mind someday, we wouldn't have to rewrite the code.

libs/utils/include/utils/PagedArenaBitset.h
pixelflinger: Those 8 bits are just intended to be the first 8 bits of a 64 word (hence the ...). Each 1 represents the presence of a 64 word in the summary table.

libs/utils/include/utils/PagedArenaBitset.h
pixelflinger: made it more clear

libs/utils/src/PagedArenaBitset.cpp
pixelflinger: fixed

libs/utils/src/PagedArenaBitset.cpp
pixelflinger: fixed

libs/utils/src/PagedArenaBitset.cpp
pixelflinger: fixed them all!

libs/utils/src/PagedArenaBitset.cpp
pixelflinger: good catch! added a unit test.

PR #9957, opened 2026-05-05
Support attribute-less procedural rendering

z3moon:
Introduce support for rendering geometries with zero vertex attributes and omitted index buffers, enabling fully procedural shader-driven rendering (e.g., full-screen triangles, billboards, or particles) via getVertexIndex() / gl_VertexID.

Some core changes include:

- C++ Core & Driver API:
  * Relax validation constraints in VertexBuffer to allow `bufferCount(0)` and 0 declared attributes.
  * Add new non-indexed `geometry()` and `setGeometryAt()` overloads to RenderableManager, omitting the IndexBuffer parameter.
  * Update RenderPrimitive to accept a null IndexBuffer to initiate non-indexed drawing modes.
  * Introduce a new `drawArrays` command to DriverAPI to delegate non-indexed instanced draw calls natively across all backend drivers (OpenGL, Vulkan, Metal, WebGPU, and Noop).
  * Guard pipeline vertex input states to gracefully accept 0 attributes.
- Corresponding Java bindings
- Desktop & Android Samples

BUGS=[342018953]

Samples:

image

image


filament/backend/src/metal/MetalDriver.mm
```diff
@@ -2357,10 +2358,15 @@
...
     }
 
-    const auto bufferCount = maxBufferIndex + 1;
-    MetalBuffer::bindBuffers(getPendingCommandBuffer(mContext), mContext->currentRenderPassEncoder,
-            USER_VERTEX_BUFFER_BINDING_START, MetalBuffer::Stage::VERTEX, vertexBuffers,
-            vertexBufferOffsets, bufferCount);
+    // For attribute-less primitives `vbi->bufferMapping` is empty, in which case the loop above
+    // never runs. Don't bind a phantom buffer in that case (would otherwise be a 1-slot bind of
+    // a nil pointer).
+    if (!vbi->bufferMapping.empty()) {
```
pixelflinger: unlikely?

filament/backend/src/metal/MetalDriver.mm
```diff
@@ -2470,6 +2476,49 @@
...
+    FILAMENT_CHECK_PRECONDITION(mContext->currentRenderPassEncoder != nullptr)
+            << "drawArrays() without a valid command encoder.";
+    DEBUG_LOG("drawArrays(...)\n");
+
+    if (FILAMENT_ENABLE_MATDBG && UTILS_UNLIKELY(!mContext->validPipelineBound)) {
+        return;
+    }
+
+    // Bind the offset data.
+    if (mContext->dynamicOffsets.isDirty()) {
```
pixelflinger: unlikely?

filament/backend/src/vulkan/VulkanDriver.cpp
```diff
@@ -2622,6 +2630,12 @@ void VulkanDriver::draw2(uint32_t indexOffset, uint32_t indexCount, uint32_t ins
     }
     mDescriptorSetCache.commit(mCurrentRenderPass.commandBuffer, mPipelineState.pipelineLayout,
             mPipelineState.descriptorSetMask);
+}
+
+void VulkanDriver::draw2(uint32_t indexOffset, uint32_t indexCount, uint32_t instanceCount) {
+    FVK_SYSTRACE_SCOPE();
+    VkCommandBuffer cmdbuffer = mCurrentRenderPass.commandBuffer->buffer();
+    prepareDraw();
```
poweifeng: Can you reorder the two commands so that assigning cmdbuffer comes after prepareDraw()?

filament/backend/src/vulkan/VulkanDriver.cpp
```diff
@@ -2631,6 +2645,16 @@ void VulkanDriver::draw2(uint32_t indexOffset, uint32_t indexCount, uint32_t ins
     vkCmdDrawIndexed(cmdbuffer, indexCount, instanceCount, firstIndex, vertexOffset, firstInstId);
 }
 
+void VulkanDriver::drawArrays(uint32_t vertexOffset, uint32_t vertexCount,
+        uint32_t instanceCount) {
+    FVK_SYSTRACE_SCOPE();
+    VkCommandBuffer cmdbuffer = mCurrentRenderPass.commandBuffer->buffer();
```
poweifeng: Can you reorder the two commands so that assigning cmdbuffer comes after prepareDraw()?

android/gradle.properties
```diff
@@ -22,8 +22,8 @@ android.defaults.buildfeatures.buildconfig=true
 android.nonTransitiveRClass=false
 android.nonFinalResIds=false
 
-com.google.android.filament.tools-dir=../../../out/release/filament
-com.google.android.filament.dist-dir=../out/android-release/filament
+com.google.android.filament.tools-dir=../../../out/debug/filament
+com.google.android.filament.dist-dir=../out/android-debug/filament
 com.google.android.filament.abis=all
```
poweifeng, line 27: shouldn't change this file I think

android/filament-android/src/main/cpp/RenderableManager.cpp
```diff
@@ -108,6 +108,25 @@ Java_com_google_android_filament_RenderableManager_nBuilderGeometry__JIIJJIIII(J
             (size_t) count);
 }
 
+extern "C" JNIEXPORT void JNICALL
+Java_com_google_android_filament_RenderableManager_nBuilderGeometry__JIIJII(JNIEnv*,
```
poweifeng, line 112: What is that postfix "JIIJII" for?

poweifeng: I think it's for the parameters, but none of the other methods have these.

filament/src/details/VertexBuffer.cpp
```diff
@@ -168,7 +168,22 @@ VertexBuffer::Builder& VertexBuffer::Builder::async(CallbackHandler* handler,
...
     FILAMENT_CHECK_PRECONDITION(mImpl->mVertexCount > 0) << "vertexCount cannot be 0";
-    FILAMENT_CHECK_PRECONDITION(mImpl->mBufferCount > 0) << "bufferCount cannot be 0";
+
+    // Attribute-less rendering: allow bufferCount(0) only when no attributes are declared.
+    // This is used for procedural geometry that reads gl_VertexID/gl_VertexIndex/[[vertex_id]].
+    bool const isAttributeless =
+            mImpl->mBufferCount == 0 && mImpl->mDeclaredAttributes.none();
+
+    if (!isAttributeless) {
+        FILAMENT_CHECK_PRECONDITION(mImpl->mBufferCount > 0) << "bufferCount cannot be 0";
```
poweifeng, line 178: I think by definition this cannot be false. Could consider removing

filament/src/details/VertexBuffer.cpp
```diff
@@ -221,8 +236,10 @@ VertexBuffer* VertexBuffer::Builder::build(Engine& engine) {
         attributedBuffers.set(attributes[j].buffer);
     });
 
-    FILAMENT_CHECK_PRECONDITION(attributedBuffers.count() == mImpl->mBufferCount)
-            << "At least one buffer slot was never assigned to an attribute.";
+    if (!isAttributeless) {
```
poweifeng: I'm still a bit confused. If the vertexbuffer is attribute-less, then shouldn't the number of provided buffers (mImpl->mBufferCount = 0) match the declared attributes ( attributedBuffers.count() = 0) ?

android/filament-android/src/main/cpp/RenderableManager.cpp
romainguy: It might be necessary to distinguish between overloads. Ideally you would not use it if it's not necessary. Overloads can also be handled manually with manual registration (parts of the JNI bindings do manual registration, which is much better because it catches errors earlier, and I believe it's fast to initialize).

filament/backend/src/vulkan/VulkanDriver.cpp
z3moon: done

filament/backend/src/vulkan/VulkanDriver.cpp
z3moon: done

android/gradle.properties
z3moon: It's a leftover to test android. Thanks for catching it!

android/filament-android/src/main/cpp/RenderableManager.cpp
z3moon: yes, it's for auto resolution for overloads. you can see others have the similar postfix.

filament/src/details/VertexBuffer.cpp
z3moon: I see it's possible to hit this line if users set vertex attributes but forget to set the buffer count.

filament/src/details/VertexBuffer.cpp
z3moon: I think you're right. This `if` is redundant. Let me revert it.

filament/backend/src/metal/MetalDriver.mm
z3moon: yup, added.

filament/backend/src/metal/MetalDriver.mm
z3moon: I copied this from `draw2`. and yes it's unlikely. I added UTILS_UNLIKELY for both.

@bejado what do you think?

filament/src/details/VertexBuffer.cpp
poweifeng: yeah got it.

filament/include/filament/RenderableManager.h
```diff
@@ -224,6 +224,37 @@ class UTILS_PUBLIC RenderableManager : public FilamentAPI {
...
+         * in vertex shader. The associated VertexBuffer may have \c bufferCount == 0 with
+         * no declared attributes (see \c VertexBuffer::Builder). Attribute-less rendering requires
+         * \c FEATURE_LEVEL_1 or higher as GLES2 has no `gl_VertexID` and is incompatible with
+         * skinning and morphing.
+         *
+         * @param index zero-based index of the primitive, must be less than the count passed to Builder constructor
+         * @param type specifies the topology of the primitive (e.g., \c RenderableManager::PrimitiveType::TRIANGLES)
+         * @param vertices specifies the vertex buffer, which in turn specifies a set of attributes
+         * @param offset specifies where in the vertex buffer to start reading (expressed as a number of vertices)
+         * @param count number of vertices to read (for triangles, this should be a multiple of 3)
```
poweifeng, line 250: so there is also a vertexBuffer `count` parameter, what's the relationship between the two?

z3moon: `vertices->getVertexCount()` is passed as this param from the second overloads (below)

other than this, users can specify the number of vertices to be rendered for non-indexed rendering.

poweifeng: Can I suggest that we simplify the API and just have `geometry(index, type, vertices)` and omit this one?  So that we'd only have one source of vertex count. Maybe the offset part can be done in the material.

z3moon: In that way, we don't fully support non-indexed rendering. Users may want to have one vertex buffer shared among multiple renderables each of which has a pair of offset and count. And the original `geometry` above has also this overload to support that use case.

filament/backend/src/metal/MetalDriver.mm
bejado: Sounds good to me.

filament/src/RenderPass.cpp
```diff
@@ -1114,7 +1115,12 @@ void RenderPass::Executor::execute(FEngine const& engine, DriverApi& driver,
                             +PushConstantIds::MORPHING_BUFFER_OFFSET, int32_t(info.morphingOffset));
                 }
 
-                driver.draw2(info.indexOffset, info.indexCount, info.instanceCount);
+                if (UTILS_LIKELY(info.isIndexed)) {
+                    driver.draw2(info.indexOffset, info.indexCount, info.instanceCount);
+                } else {
+                    // For non-indexed draws, indexOffset/indexCount carry vertexOffset/vertexCount.
+                    driver.drawArrays(info.indexOffset, info.indexCount, info.instanceCount);
```
bejado: I worry keeping these fields named `indexOffset` and `indexCount` might lead to confusion. Maybe we rename them? At a minimum I think we should add a comment next to these fields in RenderPass.h.

bejado (approved):
LGTM, one minor comment. The new samples are neat! Would be cool to be able to use the mouse to manipulate the camera view.

filament/src/RenderPass.cpp
pixelflinger: we can also use a union with both names

pixelflinger (approved):

pixelflinger:
> LGTM, one minor comment. The new samples are neat! Would be cool to be able to use the mouse to manipulate the camera view.

maybe attach a screenshot with the PR?

filament/include/filament/RenderableManager.h
poweifeng: ok got it

poweifeng (approved):

filament/src/RenderPass.cpp
z3moon: Yeah I renamed them with comments.

PR #9960, opened 2026-05-06
Use fence pool in Vulkan to ensure conversion to FD always provides an FD

anishmgoyal:
In certain applications, even if a fence has already signaled by the time it's converted to an fd, it's important to be able to convert the fence to an fd to calculate things like frame timing. This is useful for apps that are pacing frames, like games.

### DIFF
--- a/filament/backend/include/backend/platforms/VulkanPlatform.h
+++ b/filament/backend/include/backend/platforms/VulkanPlatform.h
@@ -262,12 +262,10 @@ public:
     /**
      * Creates a Platform::Sync object, which tracks a fence and its status,
      * and allows conversion to an external sync.
-     * @param fence         The underlying VkFence to use for synchronization.
-     * @param fenceStatus   An object tracking the fence's state
+     * @param fenceStatus   An object tracking the fence and its current state.
      * @return              A Platform::Sync object tracking the provided fence.
      */
-    virtual Platform::Sync* createSync(VkFence fence,
-            std::shared_ptr<VulkanCmdFence> fenceStatus) noexcept;
+    virtual Platform::Sync* createSync(std::shared_ptr<VulkanCmdFence> fenceStatus) noexcept;
 
     /**
      * Destroys a sync. If called with a sync not created by this platform
@@ -458,7 +456,6 @@ public:
 
 protected:
     struct VulkanSync : public Platform::Sync {
-        VkFence fence;
         std::shared_ptr<VulkanCmdFence> fenceStatus;
     };
 
--- a/filament/backend/src/vulkan/VulkanAsyncHandles.cpp
+++ b/filament/backend/src/vulkan/VulkanAsyncHandles.cpp
@@ -200,6 +200,21 @@ std::shared_ptr<VulkanCmdFence> VulkanCmdFence::completed() noexcept {
     return cmdFence;
 }
 
+VkResult VulkanCmdFence::updateStatus(VkDevice device) {
+    VkResult status = mStatus;
+    {
+        std::shared_lock rl(mLock);
+        if (mFence != VK_NULL_HANDLE) {
+            status = vkGetFenceStatus(device, mFence);
+        }
+    }
+
+    if (status == VK_SUCCESS) {
+        setStatus(status);
+    }
+    return mStatus;
+}
+
 FenceStatus VulkanCmdFence::wait(VkDevice device, uint64_t const timeout,
     std::chrono::steady_clock::time_point const until) {
 
--- a/filament/backend/src/vulkan/VulkanAsyncHandles.cpp
+++ b/filament/backend/src/vulkan/VulkanAsyncHandles.cpp
@@ -210,41 +210,41 @@ VkResult VulkanCmdFence::updateStatus(VkDevice device) {
 
 FenceStatus VulkanCmdFence::wait(VkDevice device, uint64_t const timeout,
     std::chrono::steady_clock::time_point const until) {
-    // Hold the fence until we've determined we've checked the status, and
-    // want to wait on the fence.
-    std::shared_lock rl(mLock);
+    {
+        // Hold the fence until we've determined we've checked the status, and
+        // want to wait on the fence.
+        std::shared_lock rl(mLock);
 
-    // If the vulkan fence has not been submitted yet, we need to wait for that before we
-    // can use vkWaitForFences()
-    if (mStatus == VK_INCOMPLETE) {
-        bool const success = mCond.wait_until(rl, until, [this] {
-            // Internally we use the VK_INCOMPLETE status to mean "not yet submitted".
-            // When this fence gets submitted, its status changes to VK_NOT_READY.
-            return mStatus != VK_INCOMPLETE || mCanceled;
-        });
-        if (!success) {
-            // !success indicates a timeout or cancel
-            return mCanceled ? FenceStatus::ERROR : FenceStatus::TIMEOUT_EXPIRED;
+        // If the vulkan fence has not been submitted yet, we need to wait for that before we
+        // can use vkWaitForFences()
+        if (mStatus == VK_INCOMPLETE) {
+            bool const success = mCond.wait_until(rl, until, [this] {
+                // Internally we use the VK_INCOMPLETE status to mean "not yet submitted".
+                // When this fence gets submitted, its status changes to VK_NOT_READY.
+                return mStatus != VK_INCOMPLETE || mCanceled;
+            });
+            if (!success) {
+                // !success indicates a timeout or cancel
+                return mCanceled ? FenceStatus::ERROR : FenceStatus::TIMEOUT_EXPIRED;
+            }
         }
-    }
 
-    // The fence could have already signaled, avoid calling into vkWaitForFences()
-    if (mStatus == VK_SUCCESS) {
-        return FenceStatus::CONDITION_SATISFIED;
-    }
+        // The fence could have already signaled, avoid calling into vkWaitForFences()
+        if (mStatus == VK_SUCCESS) {
+            return FenceStatus::CONDITION_SATISFIED;
+        }
 
-    // Or it could have been canceled, return immediately
-    if (mCanceled) {
-        return FenceStatus::ERROR;
+        // Or it could have been canceled, return immediately
+        if (mCanceled) {
+            return FenceStatus::ERROR;
+        }
     }
 
     // If we're here, we know that vkQueueSubmit has been called (because it sets the status
     // to VK_NOT_READY).
     // However, since we're going to wait on the fence, and because the fence is guaranteed
-    // to stay in scope as long as this reference is still alive, we can unlock the lock here.
-    rl.unlock();
-
-    // Wait on the fence, without any lock.
+    // to stay in scope as long as this VulkanCmdFence is still alive, we do not need a lock
+    // here.
     VkResult const status = vkWaitForFences(device, 1, &mFence, VK_TRUE, timeout);
     if (status == VK_TIMEOUT) {
         return FenceStatus::TIMEOUT_EXPIRED;
--- a/filament/backend/src/vulkan/VulkanAsyncHandles.cpp
+++ b/filament/backend/src/vulkan/VulkanAsyncHandles.cpp
@@ -17,6 +17,7 @@
 #include "VulkanAsyncHandles.h"
 
 #include "VulkanConstants.h"
+#include "VulkanFencePool.h"
 
 #include "vulkan/utils/Spirv.h"
 
@@ -181,8 +182,20 @@ void VulkanProgram::flushPushConstants(VkPipelineLayout layout) {
     mQueuedPushConstants.clear();
 }
 
+VulkanCmdFence::VulkanCmdFence(VkFence fence, std::function<void(VkFence)> recycleFn)
+    : mFence(fence), mRecycleFn(recycleFn) {
+}
+
+VulkanCmdFence::~VulkanCmdFence() {
+    // Ensure the fence can't be cleared during this operation.
+    std::shared_lock rl(mLock);
+    if (mFence != VK_NULL_HANDLE && mRecycleFn != nullptr) {
+        mRecycleFn(mFence);
+    }
+}
+
 std::shared_ptr<VulkanCmdFence> VulkanCmdFence::completed() noexcept {
-    auto cmdFence = std::make_shared<VulkanCmdFence>(VK_NULL_HANDLE);
+    auto cmdFence = std::make_unique<VulkanCmdFence>(VK_NULL_HANDLE);
     cmdFence->mStatus = VK_SUCCESS;
     return cmdFence;
 }
@@ -224,6 +237,12 @@ FenceStatus VulkanCmdFence::wait(VkDevice device, uint64_t const timeout,
     // place simultaneously. vkResetFence is only called once it knows the fence has signaled,
     // which guaranties that vkResetFence won't have to wait too long, just enough for
     // all the vkWaitForFences() to return.
+    if (mFence == VK_NULL_HANDLE) {
+        // This can happen if this is a 'completed' fence, or if
+        // the fence has been cleared because Filament is being
+        // terminated.
+        return FenceStatus::CONDITION_SATISFIED;
+    }
     VkResult const status = vkWaitForFences(device, 1, &mFence, VK_TRUE, timeout);
     if (status == VK_TIMEOUT) {
         return FenceStatus::TIMEOUT_EXPIRED;
@@ -237,15 +256,6 @@ FenceStatus VulkanCmdFence::wait(VkDevice device, uint64_t const timeout,
     }
 
     return FenceStatus::ERROR; // not supported
-}
-
-void VulkanCmdFence::resetFence(VkDevice device) {
-    // This lock prevents vkResetFences() from being called simultaneously with vkWaitForFences(),
-    // but by construction, when we're here, we know that the fence has signaled and
-    // vkWaitForFences() will return shortly.
-    std::lock_guard const l(mLock);
-    assert_invariant(mStatus == VK_SUCCESS);
-    vkResetFences(device, 1, &mFence);
 }
 
 } // namespace filament::backend
--- a/filament/backend/src/vulkan/VulkanAsyncHandles.cpp
+++ b/filament/backend/src/vulkan/VulkanAsyncHandles.cpp
@@ -195,7 +195,7 @@ VulkanCmdFence::~VulkanCmdFence() {
 }
 
 std::shared_ptr<VulkanCmdFence> VulkanCmdFence::completed() noexcept {
-    auto cmdFence = std::make_unique<VulkanCmdFence>(VK_NULL_HANDLE);
+    auto cmdFence = std::make_shared<VulkanCmdFence>(VK_NULL_HANDLE);
     cmdFence->mStatus = VK_SUCCESS;
     return cmdFence;
 }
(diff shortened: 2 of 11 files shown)


anishmgoyal:
cc @rafadevai

filament/backend/include/backend/platforms/VulkanPlatform.h
```diff
@@ -262,12 +262,10 @@ class VulkanPlatform : public Platform, utils::PrivateImplementation<VulkanPlatf
...
      * Creates a Platform::Sync object, which tracks a fence and its status,
      * and allows conversion to an external sync.
-     * @param fence         The underlying VkFence to use for synchronization.
-     * @param fenceStatus   An object tracking the fence's state
+     * @param fenceStatus   An object tracking the fence and its current state.
      * @return              A Platform::Sync object tracking the provided fence.
      */
-    virtual Platform::Sync* createSync(VkFence fence,
-            std::shared_ptr<VulkanCmdFence> fenceStatus) noexcept;
+    virtual Platform::Sync* createSync(std::shared_ptr<VulkanCmdFence> fenceStatus) noexcept;
```
anishmgoyal, line 268: is this change ok? it's in VulkanPlatform, so it may technically be 'public' - although it's intended to be used by the vulkan driver specifically.

if this is unsafe, i'm happy to revert this detail.

anishmgoyal: in theory, a public caller shouldn't be able to invoke this api because they can't instantiate a VulkanCmdFence object

poweifeng: should be ok since there is no other overriding of this method that I know of.

filament/backend/src/vulkan/VulkanFencePool.cpp
```diff
@@ -0,0 +1,170 @@
...
+    std::unique_lock<std::mutex> lock (mFenceListMutex);
+    if (!mFences.empty()) {
+        VkFence fence = mFences.back().second;
+        mFences.pop_back();
+        return fence;
+    } else {
+        VkFence fence = allocateFence();
+        if (fence != VK_NULL_HANDLE) {
+            ++mNumFences;
+        }
```
poweifeng: Can it be null?  Won't it break caller's assumption if it it null?

filament/backend/src/vulkan/VulkanAsyncHandles.h
```diff
@@ -173,12 +177,30 @@ struct VulkanCmdFence {
...
     }
 
-    void resetFence(VkDevice device);
+    // Clears the fence from this struct. It assumes that either
+    // a) the recycleFn was defined in the ctor, and knows how to
+    // dispose of the fence, or b) the creator of this object owns
+    // the fence.
+    void clearFence();
+
+    template <typename T>
```
poweifeng: if bool is the only return value needed right now, let's just assume T == bool and not be so general.

poweifeng: I also wonder if it's safe to hold a lock while calling a client callback. I guess we can be more specific in the documentation

filament/backend/src/vulkan/VulkanAsyncHandles.h
```diff
@@ -173,12 +177,30 @@ struct VulkanCmdFence {
...
         std::shared_lock const l(mLock);
         return mStatus;
     }
 
-    void resetFence(VkDevice device);
+    // Clears the fence from this struct. It assumes that either
+    // a) the recycleFn was defined in the ctor, and knows how to
+    // dispose of the fence, or b) the creator of this object owns
+    // the fence.
+    void clearFence();
```
poweifeng: can you make this private and VulkanFencePool a friend class

filament/backend/src/vulkan/VulkanAsyncHandles.h
```diff
@@ -173,12 +177,30 @@ struct VulkanCmdFence {
         mCond.notify_all();
     }
 
+    // This is safe to use in the backend thread, as the backend
+    // thread is the only thread that will be calling clearFence().
+    inline VkFence getVkFence() {
```
poweifeng, line 185: nit: const

filament/backend/src/vulkan/VulkanFencePool.cpp
```diff
@@ -0,0 +1,170 @@
...
+    const auto fenceStatus = std::make_shared<VulkanCmdFence>(fence, recycleFn);
+    // Store a weak pointer to the fence status, so we can clear the fence
+    // later if terminate() is called.
+    mFenceStatuses.push_back(std::weak_ptr<VulkanCmdFence>(fenceStatus));
+    return fenceStatus;
+}
+
+void VulkanFencePool::gc() noexcept {
+    // Clear any fence statuses that no longer exist.
+    for (auto it = mFenceStatuses.begin(); it != mFenceStatuses.end();) {
```
poweifeng: nit: can do a `erase(std::remove())` here.  For example: https://en.cppreference.com/cpp/algorithm/remove

filament/backend/src/vulkan/VulkanFencePool.h
```diff
@@ -0,0 +1,77 @@
...
+
+    // Reclaims fences from all fenceStatus objects, and then clears the
+    // fence list.
+    void terminate() noexcept;
+
+private:
+    std::mutex mFenceListMutex;
+    VulkanContext const& mContext;
+    VkDevice mDevice;
+    uint32_t mMinPoolSize;
```
poweifeng: can be const

filament/backend/src/vulkan/VulkanAsyncHandles.h
poweifeng: seems kind of strange for this class to own the fence and the status and have another party set the status of fence.  But can be revisited later.

filament/backend/src/vulkan/VulkanAsyncHandles.cpp
```diff
@@ -181,6 +182,18 @@ void VulkanProgram::flushPushConstants(VkPipelineLayout layout) {
     mQueuedPushConstants.clear();
 }
 
+VulkanCmdFence::VulkanCmdFence(VkFence fence, std::function<void(VkFence)> recycleFn)
+    : mFence(fence), mRecycleFn(recycleFn) {
+}
+
+VulkanCmdFence::~VulkanCmdFence() {
+    // Ensure the fence can't be cleared during this operation.
+    std::shared_lock rl(mLock);
```
rafadevai: if the destructor is being called, there's no need to lock anything. No other method should be called in this object otherwise something else is wrong

filament/backend/src/vulkan/VulkanAsyncHandles.cpp
```diff
@@ -239,13 +258,14 @@ FenceStatus VulkanCmdFence::wait(VkDevice device, uint64_t const timeout,
     return FenceStatus::ERROR; // not supported
 }
 
-void VulkanCmdFence::resetFence(VkDevice device) {
-    // This lock prevents vkResetFences() from being called simultaneously with vkWaitForFences(),
-    // but by construction, when we're here, we know that the fence has signaled and
-    // vkWaitForFences() will return shortly.
+void VulkanCmdFence::clearFence() {
     std::lock_guard const l(mLock);
```
rafadevai: Its better if the lock its only for updating the members to NULL.

You can do
- copy the members to temporary variables
- update to NULL
- release the lock
- call recycleFn if needed

filament/backend/src/vulkan/VulkanFencePool.cpp
```diff
@@ -0,0 +1,170 @@
...
+    // fences from all fenceStatus objects.
+    for (const auto& fence : mFences) {
+        destroyFence(fence.second);
+    }
+    mFences.clear();
+    mNumFences = 0;
+}
+
+VkFence VulkanFencePool::acquireFence() noexcept {
+    std::unique_lock<std::mutex> lock (mFenceListMutex);
```
rafadevai: can the main thread call this?

filament/backend/src/vulkan/VulkanAsyncHandles.h
anishmgoyal: fair - we could move that logic into this class, and have it be "updateStatus()" instead?

filament/backend/src/vulkan/VulkanFencePool.cpp
anishmgoyal: ahh ok - wasn't sure if we liked the functional methods. i'll switch to `remove_if` then w/ this predicate

filament/backend/src/vulkan/VulkanFencePool.cpp
anishmgoyal: if it's null, essentially it's because we're out of memory or something like that. if we're confident that we'll always crash in that case, i think it's fine not to handle it. but if there isn't a crash, not having this could put us in an invalid state

filament/backend/src/vulkan/VulkanAsyncHandles.h
anishmgoyal: another option is i could have the method return a read lock along with the fence?
i could get rid of getVkFence() completely and only have a method that returns the lock with the fence

filament/backend/src/vulkan/VulkanFencePool.cpp
poweifeng: Currently, the vk backend just crash out on failures like this, which is not ideal, but a better way to handle this has not been thought through.  So I'd suggest we keep it consistent and crash out (precondtion/postcondition) and file a bug to track a better resolution for later.

filament/backend/src/vulkan/VulkanAsyncHandles.h
poweifeng: yeah, maybe something like that is good

filament/backend/src/vulkan/VulkanFencePool.cpp
anishmgoyal: gotcha. i can add an assert, although those do get stripped for prod, so not sure if there's something else i could put in here

poweifeng: `FILAMENT_CHECK_POSTCONDITION `

filament/backend/src/vulkan/VulkanAsyncHandles.h
anishmgoyal: ok i reviewed it a bit more, and it seems like the status isn't one-to-one exactly drawn from the fence itself. the 'success' status is, but there's an intermediate step before that which is not. i'm gonna refactor this though, i think we can make it a bit cleaner.

- markSubmitted() - will be a public function that transitions the fence into the not ready state
- VkResult checkStatus(VkDevice device) - updates the status based on current fence status, and returns it

filament/backend/src/vulkan/VulkanAsyncHandles.cpp
```diff
@@ -181,12 +182,39 @@ void VulkanProgram::flushPushConstants(VkPipelineLayout layout) {
...
+VulkanCmdFence::VulkanCmdFence(VkFence fence, std::function<void(VkFence)> recycleFn)
+    : mFence(fence), mRecycleFn(recycleFn) {
+}
+
+VulkanCmdFence::~VulkanCmdFence() {
+    // Ensure the fence can't be cleared during this operation.
+    std::shared_lock rl(mLock);
+    if (mFence != VK_NULL_HANDLE && mRecycleFn != nullptr) {
+        mRecycleFn(mFence);
+    }
```
poweifeng: nit: call clearFence?

poweifeng (approved):

filament/backend/src/vulkan/VulkanAsyncHandles.cpp
anishmgoyal: d'oh - that makes sense

filament/backend/src/vulkan/VulkanAsyncHandles.cpp
anishmgoyal: fair point - although now, we're just calling through to clearFence here

filament/backend/src/vulkan/VulkanAsyncHandles.cpp
anishmgoyal: fair - will update

filament/backend/src/vulkan/VulkanFencePool.cpp
anishmgoyal: we'll remove the locks from fence pool

filament/backend/src/vulkan/VulkanAsyncHandles.cpp
```diff
@@ -219,33 +240,26 @@ FenceStatus VulkanCmdFence::wait(VkDevice device, uint64_t const timeout,
...
     VkResult const status = vkWaitForFences(device, 1, &mFence, VK_TRUE, timeout);
     if (status == VK_TIMEOUT) {
         return FenceStatus::TIMEOUT_EXPIRED;
     }
 
     if (status == VK_SUCCESS) {
-        rl.unlock();
-        std::lock_guard const wl(mLock);
-        mStatus = status;
+        setStatus(status);
```
anishmgoyal, line 254: this will notify any observers of the cond var, but they should've been notified by now & woken up, so i think this is ok

anishmgoyal:
@poweifeng @rafadevai updated the synchronization logic (in the most recent commit) to remove some of the excess locking semantics we were concerned about.

filament/backend/src/vulkan/VulkanAsyncHandles.cpp
```diff
@@ -219,33 +240,26 @@ FenceStatus VulkanCmdFence::wait(VkDevice device, uint64_t const timeout,
...
+    rl.unlock();
+
+    // Wait on the fence, without any lock.
     VkResult const status = vkWaitForFences(device, 1, &mFence, VK_TRUE, timeout);
     if (status == VK_TIMEOUT) {
         return FenceStatus::TIMEOUT_EXPIRED;
     }
 
     if (status == VK_SUCCESS) {
-        rl.unlock();
```
poweifeng, line 233: I don't think you want to change the semantics here.  The lock was held to make sure the fence is externally synchronized.  I think there are places where wait() is called from different threads.

filament/backend/src/vulkan/VulkanAsyncHandles.cpp
```diff
@@ -181,16 +182,36 @@ void VulkanProgram::flushPushConstants(VkPipelineLayout layout) {
...
+}
+
 std::shared_ptr<VulkanCmdFence> VulkanCmdFence::completed() noexcept {
     auto cmdFence = std::make_shared<VulkanCmdFence>(VK_NULL_HANDLE);
     cmdFence->mStatus = VK_SUCCESS;
     return cmdFence;
 }
 
+VkResult VulkanCmdFence::updateStatus(VkDevice device) {
+    if (mFence != VK_NULL_HANDLE) {
```
poweifeng, line 202: I think you need a lock here for the same reason wait() has a read lock

filament/backend/src/vulkan/VulkanAsyncHandles.cpp
rafadevai: You need to keep the unlock because the setStatus will try to acquire the lock and still not avaiable.

filament/backend/src/vulkan/VulkanFencePool.cpp
```diff
@@ -0,0 +1,161 @@
...
+        [](const auto& fenceStatus) {
+            return fenceStatus.expired();
+        });
+    mFenceStatuses.erase(expiredStatuses, mFenceStatuses.end());
+
+    if (++mCurrFrame <= TIME_BEFORE_EVICTION) {
+        return;
+    }
+
+    // Lock now to avoid race conditions with other threads that may be
```
rafadevai: Update the comments related to locking

filament/backend/src/vulkan/VulkanFencePool.cpp
```diff
@@ -0,0 +1,161 @@
...
+static constexpr uint32_t TIME_BEFORE_EVICTION = 3;
+
+}  // namespace
+
+VulkanFencePool::VulkanFencePool(VulkanContext const& context,
+                                 VkDevice device,
+                                 uint32_t minPoolSize)
+    : mContext(context), mDevice(device), mMinPoolSize(minPoolSize)
+{
+    for (size_t i = 0; i < minPoolSize; ++i) {
```
rafadevai, line 41: call mFences.reserve(minPoolSize) before the for

filament/backend/src/vulkan/VulkanFencePool.h
```diff
@@ -0,0 +1,82 @@
...
+
+    // Clears idle fences, as well as cmdFences that are no longer alive.
+    void gc() noexcept;
+
+    // Reclaims fences from all fenceStatus objects, and then clears the
+    // fence list.
+    void terminate() noexcept;
+
+private:
+    VulkanContext const& mContext;
```
rafadevai: put the members at the end of the class definition

filament/backend/src/vulkan/VulkanFencePool.cpp
anishmgoyal: this is a deque - deques don't have reserve. deques don't reallocate fully as they grow.

filament/backend/src/vulkan/VulkanAsyncHandles.cpp
anishmgoyal: the lock is unlocked earlier in this function

anishmgoyal: @poweifeng re the synchronization semantics - this was a read lock, which meant that every thread calling wait can call it concurrently. this was important because there was a resetFence function - as you can see described in the comment that i removed. now, locking would simply prevent the status from being updated while threads are waiting, but at the new point where the lock is unlocked, we never check status again, and only ever update status to set it to success if we update it at all.

filament/backend/src/vulkan/VulkanAsyncHandles.cpp
anishmgoyal: i'm not sure we do - as of the new change, mFence is immutable. lemme actually mark it as const to make that clear. mFence cannot change during the lifecycle of a VulkanCmdFence now.

filament/backend/src/vulkan/VulkanAsyncHandles.cpp
poweifeng: hmm, ok.  I think understand now. the lock used to lock on the mFence because it could be reset, but now that reset happens only when it has been recycled, we don't need to worry about that case.

filament/backend/src/vulkan/VulkanAsyncHandles.cpp
poweifeng: ack below.

poweifeng (approved):

filament/backend/src/vulkan/VulkanAsyncHandles.cpp
anishmgoyal: sorry - pushed onnne additional change just now. basically just restructured so the lock is using raii / a scoped block so that the code is a bit more readable.

filament/backend/src/vulkan/VulkanFencePool.cpp
rafadevai: Ack

filament/backend/src/vulkan/VulkanFencePool.cpp
```diff
@@ -0,0 +1,153 @@
...
+    };
+    const auto fenceStatus = std::make_shared<VulkanCmdFence>(fence, recycleFn);
+    // Store a weak pointer to the fence status, so we can clear the fence
+    // later if terminate() is called.
+    mFenceStatuses.push_back(std::weak_ptr<VulkanCmdFence>(fenceStatus));
+    return fenceStatus;
+}
+
+void VulkanFencePool::gc() noexcept {
+    // Clear any fence statuses that no longer exist.
```
rafadevai, line 67: can you add a marker for for this function, so we can see it in the traces.

rafadevai (approved):

filament/backend/src/vulkan/VulkanFencePool.cpp
anishmgoyal: added!

PR #9964, opened 2026-05-07
utils: reduce PagedArenaBitset bloat

pixelflinger:
Remove a few not so useful methods and prevented inlining of others. This reduces the size of PagedArenaBitset by about 4KB of code. Also add restrict and nonnull annotation.
Overall benchmarks actually improve slightly.

### DIFF
--- a/libs/utils/include/utils/PagedArenaBitset.h
+++ b/libs/utils/include/utils/PagedArenaBitset.h
@@ -181,19 +181,24 @@ public:
     /**
      * @brief Returns the total number of set bits.
      */
-    size_t size() const;
+    size_t size() const noexcept {
+        assert(mSummaryMask.size() == MASK_WORDS && "FATAL: Attempted to use a moved-from PagedArenaBitset!");
+        return mSize;
+    }
 
     /**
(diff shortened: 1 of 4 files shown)


libs/utils/include/utils/PagedArenaBitset.h
```diff
@@ -181,19 +181,24 @@ class UTILS_PUBLIC PagedArenaBitset {
     /**
      * @brief Returns the total number of set bits.
      */
-    size_t size() const;
+    size_t size() const noexcept {
+        assert(mSummaryMask.size() == MASK_WORDS && "FATAL: Attempted to use a moved-from PagedArenaBitset!");
```
poweifeng, line 185: assert_invariant or FILAMENT_POSTCONDITION?

poweifeng (approved):

libs/utils/include/utils/PagedArenaBitset.h
pixelflinger: I consider this class internal, but I had to put it into the public headers because EntityManager is public.  I have to revisit that to see if I can do better here. So I think assert() is good then.

PR #9969, opened 2026-05-08
backend: typed clear-color value for integer-format attachments

z3moon:
clearWithRasterPipe always called glClearBufferfv, which is undefined behavior on integer-format color attachments and crashed apps that attached an R*UI / R*I texture to a RenderTarget.

Replace `RenderPassParams::clearColor` (was math::float4) and the matching FrameGraphRenderPass descriptors with `ClearColorValue`, a tagged union of float[4] / int32[4] / uint32[4].

`ClearColorValue` is implicitly constructible from math::float4 / int4 / uint4, so existing float-based call sites compile unchanged.

The public `Renderer::ClearOptions` API is not touched yet as there's no request for it now. Updating the public API to support typed clear-color is tricky because Java has no native concept of a union.

BUGS=[510864041]


filament/backend/include/backend/DriverEnums.h
```diff
@@ -1583,6 +1583,48 @@ struct RenderPassFlags {
...
+    // Convenience to keep `{r, g, b, a}` brace-init syntax working at call sites.
+    constexpr ClearColorValue(float r, float g, float b, float a) noexcept
+            : type(Type::FLOAT), asFloat{ r, g, b, a } {}
+    constexpr ClearColorValue(int32_t r, int32_t g, int32_t b, int32_t a) noexcept
+            : type(Type::INT), asInt{ r, g, b, a } {}
+    constexpr ClearColorValue(uint32_t r, uint32_t g, uint32_t b, uint32_t a) noexcept
+            : type(Type::UINT), asUint{ r, g, b, a } {}
+
+    Type type;
+    union {
```
poweifeng: have you considered std::variant?  Then you don't need a type.

filament/backend/include/backend/DriverEnums.h
```diff
@@ -1583,6 +1583,48 @@ struct RenderPassFlags {
...
+        float asFloat[4];
```
poweifeng: also why not `math::float4` and similar types?

filament/backend/src/metal/MetalHandles.mm
```diff
@@ -1201,8 +1201,30 @@ static void func(void* user) {
...
         descriptor.colorAttachments[i].storeAction = getStoreAction(params,
                 getTargetBufferFlagsAt(i));
-        descriptor.colorAttachments[i].clearColor = MTLClearColorMake(
-                params.clearColor.r, params.clearColor.g, params.clearColor.b, params.clearColor.a);
+        // Metal's MTLClearColor is always 4 doubles. The texture's pixel format determines whether those doubles are
+        // interpreted as floats, signed ints, or unsigned ints. A double has a 53-bit mantissa,
+        // so any int32_t / uint32_t value round-trips exactly.
+        switch (params.clearColor.type) {
+            case ClearColorValue::Type::FLOAT:
+                descriptor.colorAttachments[i].clearColor = MTLClearColorMake(
```
poweifeng: maybe you could have a `double4 toDouble() const` method for `ClearColorValue`

filament/backend/include/backend/DriverEnums.h
z3moon: Yeah I chose that at first, then changed later in favor of debugability and readability. Also POD is a bit more performant and better for compilation time. But could return to variant if we want.

pixelflinger:
I don't understand how this fixes anything since there is no way to set the ClearColor  to anything but float?

z3moon:
> I don't understand how this fixes anything since there is no way to set the ClearColor to anything but float?

Ah I forgot to bring the actual fix after the refactor.

z3moon:
Actually I think just using `double[4]` (like Metal and WebGPU do) would be much simpler. It'll also cover the Java binding concern except that we need to change the public field, which is trivial I think. Let me update this change with that approach.

filament/backend/src/vulkan/VulkanDriver.cpp
```diff
@@ -2068,10 +2062,26 @@ void VulkanDriver::beginRenderPass(Handle<HwRenderTarget> rth, const RenderPassP
         for (int i = 0; i < MRT::MAX_SUPPORTED_RENDER_TARGET_COUNT; i++) {
             if (fbkey.color[i]) {
                 VkClearValue &clearValue = clearValues[renderPassInfo.clearValueCount++];
-                clearValue.color.float32[0] = params.clearColor.r;
-                clearValue.color.float32[1] = params.clearColor.g;
-                clearValue.color.float32[2] = params.clearColor.b;
-                clearValue.color.float32[3] = params.clearColor.a;
+                switch (params.clearColor.type) {
```
rafadevai: Are you sure this is right? The spec says that the color type needs to match the attachment or format
https://docs.vulkan.org/refpages/latest/refpages/source/VkClearColorValue.html

filament/backend/include/backend/DriverEnums.h
```diff
@@ -1583,6 +1583,48 @@ struct RenderPassFlags {
     TargetBufferFlags discardEnd;
 };
 
+/**
+ * A typed clear-color value for a color attachment.
```
rafadevai: high level comment, please add some tests

filament/backend/include/backend/DriverEnums.h
z3moon: I reworked this with `math::double4`.

filament/backend/src/vulkan/VulkanDriver.cpp
z3moon: Yeah it was incomplete. I reworked it.

filament/backend/include/backend/DriverEnums.h
z3moon: done

filament/backend/include/backend/DriverEnums.h
```diff
@@ -1583,6 +1583,64 @@ struct RenderPassFlags {
...
+    // caller explicitly tags an integer family.
+    constexpr ClearColorValue(math::double4 const& d, Type t = Type::AUTO) noexcept
+            : type(t),
+              color(d) {}
+
+    // Note: `{r, g, b, a}` brace-init is intentionally not supported. It picks an overload by
+    // literal type (1 -> INT, 1.f -> FLOAT, 1.0 -> ambiguous), which silently changes the type
+    // tag in ways the caller cannot see. Use math::float4 / int4 / uint4 / double4 explicitly.
+
+    Type type;
```
pixelflinger: do we even need a Type? `double` can encode any value.  We can just document that the value you put there must be consistant with your RT and that we will "simply" convert to the right type as-is.
You can still keep the overrides for conveniance.

filament/include/filament/Renderer.h
```diff
@@ -160,6 +160,20 @@ class UTILS_PUBLIC Renderer : public FilamentAPI {
...
+         * matching family at clear time. Use `FLOAT` / `INT` / `UINT` only when you specifically
+         * need to override the default type with a non-zero clear color value.
+         */
+        enum class Type : uint8_t {
+            AUTO,   //!< Auto-detect based on attachment format
+            FLOAT,  //!< Float / normalized
+            INT,    //!< Signed integer
+            UINT    //!< Unsigned integer
+        };
+
```
pixelflinger: I don't think that's needed.

pixelflinger (changes requested):

filament/backend/src/vulkan/VulkanHandles.h
```diff
@@ -284,6 +284,21 @@ struct VulkanRenderTarget : private HwRenderTarget, fvkmemory::Resource {
         return mInfo->attachments[0];
     }
 
+    // Returns the attachment for color slot `slot`, or nullptr if that slot is unused.
+    // Note: `attachments` is stored compactly (only populated slots are pushed), so we map the
+    // slot index to its compact position via the `colors` bitset.
+    inline VulkanAttachment const* getColorAttachment(uint32_t slot) const {
```
rafadevai: Lets not do this, its very confusing and the index is already been decided when its created, so after that it shouldn't be modified anymore. 

Also from what I see all the colors are already put together at the beginning of the list and then you have the MSAA and depth/stencil attachments.

What was the reason you needed to do this?

filament/backend/include/backend/DriverEnums.h
z3moon: Yeah I kind of over-engineered it in case it was mis-used. The `type` was purely for safe-guard or debugging purpose. Let me remove it.

filament/include/filament/Renderer.h
z3moon: Yeah removed.

filament/backend/src/vulkan/VulkanHandles.h
z3moon: Yeah you're right. Let me simplify it. It's used to find out the format of the current attachment so that it picks the matching union arm of the attachment.

filament/backend/src/opengl/OpenGLDriver.cpp
```diff
@@ -4525,30 +4525,47 @@ void OpenGLDriver::clearWithRasterPipe(TargetBufferFlags const clearFlags,
...
-            glClearBufferfv(GL_COLOR, 7, linearColor.v);
-        }
+        GLRenderTarget const* rt = handle_cast<GLRenderTarget*>(mRenderPassTarget);
+        auto clearColorBuffer = [&](int i, TargetBufferFlags flag) {
+            if (!any(clearFlags & flag)) {
+                return;
+            }
+            // Pick the glClearBuffer*v variant from the attachment's format. Calling the wrong
+            // variant on an integer attachment is undefined behavior in GL.
+            TextureFormat const format = rt->gl.color[i] ? rt->gl.color[i]->format
```
pixelflinger: I think we could cache the type of the format inside `rt`, so we don't have to do all those `if` each time.

filament/backend/src/opengl/OpenGLDriver.cpp
```diff
@@ -4525,30 +4525,47 @@ void OpenGLDriver::clearWithRasterPipe(TargetBufferFlags const clearFlags,
...
+                                                         : TextureFormat::RGBA8;
+            if (isUnsignedIntFormat(format)) {
```
pixelflinger: The common case is actually float formats. So I think the code should be written like this:
```
if LIKELY(isFloatFormat()) {   // or !isIntegerFormat() if it's easier to test
} else {
 if (isSigned()) {
 } else {
 }
}
```

So we avoid two `if` that will almost always fail.

filament/backend/src/vulkan/VulkanDriver.cpp
```diff
@@ -2065,13 +2065,30 @@ void VulkanDriver::beginRenderPass(Handle<HwRenderTarget> rth, const RenderPassP
...
                 VkClearValue &clearValue = clearValues[renderPassInfo.clearValueCount++];
-                clearValue.color.float32[0] = params.clearColor.r;
-                clearValue.color.float32[1] = params.clearColor.g;
-                clearValue.color.float32[2] = params.clearColor.b;
-                clearValue.color.float32[3] = params.clearColor.a;
+                // VkClearColorValue is a union the driver reads according to the attachment's
+                // format. Pick the matching arm from the attachment's TextureFormat -- writing the
+                // wrong arm produces silent corruption.
+                TextureFormat const format = rt->getColor(colorIdx++).texture->format;
+                if (isUnsignedIntFormat(format)) {
```
pixelflinger: same comment as for gl

pixelflinger (changes requested):

filament/backend/src/opengl/OpenGLDriver.cpp
z3moon: Yup, now they're cached.

filament/src/details/Renderer.cpp
```diff
@@ -574,7 +574,7 @@ void FRenderer::copyFrame(FSwapChain* dstSwapChain, filament::Viewport const& ds
     // Clear color to black if the CLEAR flag is set.
     if (flags & CLEAR) {
         RenderPassParams params = {};
-        params.clearColor = {0.f, 0.f, 0.f, 1.f};
+        params.clearColor = math::float4{ 0.f, 0.f, 0.f, 1.f };
```
poweifeng: should this be backend::ClearColorValue?

z3moon: Yeah good catch. Updated.

pixelflinger (approved):

PR #9970, opened 2026-05-08
engine: fix VSM mipmaping and blur

pixelflinger:
It's incorrect to use the MaterialInstanceManager directly because  it doesn't prepare the programs. Instead we need to go through PostProcessManager.


pixelflinger:
@romainguy

romainguy (approved):

poweifeng (approved):

PR #9971, opened 2026-05-08
webgpu: fix web examples and add viewer

poweifeng:
- Fix broken web samples
 - Move viewer.js/html to web/examples (able to serve with web/examples/server.py)
 - Remove the docs_src/src_raw version of the viewer and use the same duplication logic as for the other examples
 - Move canvas in viewer.js from shadowDom to light dom.  This is necessary for WebGPU's emscripten binding


poweifeng:
bypassed review to fix broken web samples

PR #9972, opened 2026-05-08
Metal: ensure driver lifetime in MetalFence::encode

bejado:
BUGS=[510874687]

### DIFF
--- a/filament/backend/src/metal/MetalHandles.mm
+++ b/filament/backend/src/metal/MetalHandles.mm
@@ -1409,17 +1409,23 @@ void MetalFence::encode() {
         }
         [getPendingCommandBuffer(&context) encodeSignalEvent:event value:value];
 
-        // Using a weak_ptr here because the Fence could be deleted before the block executes.
+        // Using a weak_ptr for the Fence state and driver lifetime because the Fence could be
+        // deleted, or the driver could be shut down, before this block executes asynchronously.
         std::weak_ptr<State> weakState = state;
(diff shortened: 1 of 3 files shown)


pixelflinger (approved):
I can't convince myself that it doesn't work :-)
I think this really deserves some comments in both the dtor and notifyListener. 
In particular what makes it work is that the fence is signaled with the lock held.

filament/backend/src/metal/MetalHandles.mm
```diff
@@ -1409,17 +1409,23 @@ static void func(void* user) {
...
+        [event notifyListener:context.eventListener
+                      atValue:value
+                        block:^(id<MTLSharedEvent> o, uint64_t value) {
+                          auto s = weakState.lock();
+                          auto lifetime = weakDriverLifetime.lock();
+                          if (s && lifetime) {
+                              std::lock_guard<std::mutex> lock(lifetime->mutex);
+                              if (lifetime->driver) {
+                                  lifetime->driver->signalFence(
+                                          [&] { s->status = FenceStatus::CONDITION_SATISFIED; });
```
poweifeng, line 1427: why doesn't setting the status also require a lock?

poweifeng (approved):

filament/backend/src/metal/MetalHandles.mm
bejado: `signalFence` itself takes a mutex

PR #9973, opened 2026-05-08
android: [render-val] add page generation for results

poweifeng:
- Add script to re-run the compare step in python against adjusted tolerance.
 - Add web page to view the results of the test result each one from testing on device.
 - Add script to generate said web page.


z3moon (approved):

z3moon:
It's almost impossible to review these code. LGTM to unblock the progress though. Is this meant to be run by git hooks or manually?

poweifeng:
> It's almost impossible to review these code. LGTM to unblock the progress though. Is this meant to be run by git hooks or manually?

they are just helper scripts for when you have results from the render-validation app.  Reviewing is not really needed as they are not used 99% of the time, but nice to have for keeping a record in case I need to revisit.

PR #9975, opened 2026-05-11
Add enum to_string for ElementType

mhansen:
I'd like to use this for a Dear ImGUI debug UI, and I figured it's better to upstream this function than to copy this string code into a downstream project. Less risk of them getting out of sync.

I've done this in a way that's consistent with other `to_string` overloads in DriverEnums.h

Towards internal google bug b/509716623 and internal cl/911084590.

Towards #9979

### DIFF
--- a/filament/backend/include/backend/DriverEnums.h
+++ b/filament/backend/include/backend/DriverEnums.h
@@ -826,6 +826,38 @@ enum class ElementType : uint8_t {
     HALF4,
 };
 
+constexpr std::string_view to_string(ElementType type) noexcept {
+    switch (type) {
+        case ElementType::BYTE:    return "BYTE";
+        case ElementType::BYTE2:   return "BYTE2";
+        case ElementType::BYTE3:   return "BYTE3";
+        case ElementType::BYTE4:   return "BYTE4";
+        case ElementType::UBYTE:   return "UBYTE";
+        case ElementType::UBYTE2:  return "UBYTE2";
+        case ElementType::UBYTE3:  return "UBYTE3";
+        case ElementType::UBYTE4:  return "UBYTE4";
+        case ElementType::SHORT:   return "SHORT";
+        case ElementType::SHORT2:  return "SHORT2";
+        case ElementType::SHORT3:  return "SHORT3";
+        case ElementType::SHORT4:  return "SHORT4";
+        case ElementType::USHORT:  return "USHORT";
+        case ElementType::USHORT2: return "USHORT2";
+        case ElementType::USHORT3: return "USHORT3";
+        case ElementType::USHORT4: return "USHORT4";
+        case ElementType::INT:     return "INT";
+        case ElementType::UINT:    return "UINT";
+        case ElementType::FLOAT:   return "FLOAT";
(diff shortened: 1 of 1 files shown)


filament/backend/include/backend/DriverEnums.h
```diff
@@ -826,6 +826,64 @@ enum class ElementType : uint8_t {
     HALF4,
 };
 
+constexpr std::string_view to_string(ElementType type) noexcept {
```
romainguy, line 829: Same feedback as for your to_string proposal.

romainguy:
Why only these two enums btw? It would make more sense to do this consistently for all public enums. Have you also considered using something like `magic_enum` in your client code?

filament/backend/include/backend/DriverEnums.h
romainguy: Also: why is this function returning a `string_view` when the other one returns a `char*`?

pixelflinger: @romainguy we are now using string_view when possible for new to_string APIs. And we've started to use the `to_string` convention as well.  DriverEnums.h has, unfortunately, different conventions. At some point we should deprecate the old ones (and add new ones).

filament/backend/include/backend/DriverEnums.h
```diff
@@ -826,6 +826,64 @@ enum class ElementType : uint8_t {
...
+  switch (type) {
+    case ElementType::BYTE:
+      return "BYTE";
+    case ElementType::BYTE2:
```
pixelflinger: indentation is 4 not 2. also for these we put the `case` and `return` on the same line for readability.

pixelflinger (changes requested):

mhansen:
> Why only these two enums btw? It would make more sense to do this consistently for all public enums.

These are the only two I have a use case for so far :-). For some more background, I'm building a Dear ImGUI debug editor for our meshes that shows a table for the mesh showing the size of each custom column of data we're passing to the shader. Here's a screenshot of what I have:

image

I agree it'd be nice to do this for all enums.

> Have you also considered using something like `magic_enum` in your client code?

No, I hadn't heard of magic_enum - thanks for the pointer! I looked it up: magic_enum is discouraged in google (among other reasons) because it uses nonstandard compiler extensions that might get standardised differently from how magic_enum does it, and be hard to migrate if it's standardised differently, and it's hard to tell what happens with an int that's not in the enum.

filament/backend/include/backend/DriverEnums.h
mhansen: OK, sounds like we should stick with `constexpr std::string_view to_string` as this is a new API

filament/backend/include/backend/DriverEnums.h
mhansen: Sorry, that was google3 formatter sneaking in. Fixed now.

pixelflinger (approved):

mhansen:
Thanks. What's next step to get this merged? Do we wait for 2 reviews? I see "Verify Release Notes" is failing; I don't think this merits a change in release notes, but I don't seem to have access to add 'internal' label.

PR #9976, opened 2026-05-11
Add to_string for VertexAttribute enum

mhansen:
For using in a debug user interface. I could write this myself outside of filament, but then it seems likely to get out of sync if new enums are added.

I've tried to do this in a way consistent with the to_string functions in DriverEnums.h.

Towards internal google bug b/511875500 and b/509716623

Towards #9975

### DIFF
--- a/libs/filabridge/include/filament/MaterialEnums.h
+++ b/libs/filabridge/include/filament/MaterialEnums.h
@@ -24,6 +24,7 @@
 
 #include <stddef.h>
 #include <stdint.h>
+#include <string_view>
 
 namespace filament {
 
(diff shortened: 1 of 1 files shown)


libs/filabridge/include/filament/MaterialEnums.h
```diff
@@ -162,6 +162,42 @@ enum VertexAttribute : uint8_t {
     // this is limited by driver::MAX_VERTEX_ATTRIBUTE_COUNT
 };
 
+constexpr const char* to_string(VertexAttribute attr) noexcept {
```
romainguy: To follow Filament's naming conventions, it should be called toString().

Indentation should be 4 spaces, not 2.

romainguy (changes requested):

pixelflinger (changes requested):
what is this for? filabridge is linked into filament, this will add bloat.

A similar method exists in libfilamat:
const char* toString(VertexAttribute attribute) noexcept

mhansen:
> what is this for?

I'm making a Dear ImGUI debug interface to show mesh sizes and size of custom columns on the mesh. Here's what I have so far: 

image

More information on the google-internal bugs and CLs linked at the top.

> filabridge is linked into filament, this will add bloat.

I'd hoped the compiler would remove this as unused code if it's unused; is that not right? Is there some setup where it's all separately compiled and linked without LTO? I expect for non-optimised builds this will be a small increase in size, but given we already do this for many enums, I'd hoped it would be a tolerable increase.

> A similar method exists in libfilamat: const char* toString(VertexAttribute attribute) noexcept

Nice, yeah I expect there are many callers that would benefit from having one central definition of this. Perhaps we can remove that libfilamat function after we centralise the tostring next to the enum definitinon.

libs/filabridge/include/filament/MaterialEnums.h
mhansen: Sounds like from https://github.com/google/filament/pull/9975#discussion_r3220224460 we want this to be std::string_view to_string now. So I changed this to string_view to_string.

Changed indentation too (sorry, that was a google3-ism that snuck in). Moved the 'return' to one line like some other examples I saw in filament code too, to try and match style.

pixelflinger (approved):

PR #9981, opened 2026-05-12
Optimize Color Grading and Tone Mapping Pipeline with NEON SIMD

pixelflinger:
Optimize Color Grading and Tone Mapping Pipeline with NEON SIMD

This performance refactor vectorizes Filament's 3D LUT generation
pipeline and core tone mapping operators using 128-bit NEON intrinsics.
By restructuring the color grading tile processing loop and reducing
scalar overhead, LUT computation times on mobile (Pixel 8 arm64) have
been decreased by up to 79.7%, delivering significant throughput gains
across all quality tiers while maintaining strict numerical accuracy.

Key Performance Achievements (Pixel 8, 32x32x32 LUT):
- lutGenerationWithAdjustmentsInteger: 4.93x speedup (CPU time decreased
  from 12.76 ms to 2.59 ms; throughput improved to 12.65 M/s).
- lutGenerationDefault: 3.45x speedup (CPU time decreased from 5.57 ms
  to 1.61 ms; throughput improved to 20.32 M/s).
- lutGenerationAdvanced32: 2.97x speedup (CPU time decreased from 15.54 ms
  to 5.23 ms; throughput improved to 6.26 M/s).

Per-ToneMapper SIMD Kernel Acceleration:
- ACES / ACES Legacy: 3.92x speedup (5.57 ms -> 1.42 ms).
- PBR Neutral ToneMapper: 3.25x speedup (1.65 ms -> 0.51 ms).
- AgX ToneMapper: 3.10x speedup (3.10 ms -> 1.00 ms).

Architectural & Accuracy Highlights:
- Instruction Count Reduction of 83%: Total instructions executed per tile
  on the integer adjustment path decreased from 2257.7 to 380.6.
- Zero Virtual Dispatch Penalty: The virtual SIMD dispatch architecture
  achieves up to 4.93x net speedups with no measurable overhead on scalar
  fallback paths.
- Strict Precision Hygiene: An expanded 127,000-sample validation suite
  confirms < 1.0 LSB accuracy for all vectorized tone mappers. Empirical
  case studies confirm GT7ToneMapper and gamutMapping_sRGB are correctly
  preserved on the scalar fallback path to prevent transcendental drift
  (83.67 LSBs) and polynomial evaluation overhead (5x slowdown).

BUGS=[512331543]


android/filament-android/src/main/java/com/google/android/filament/ColorGrading.java
```diff
@@ -572,6 +572,18 @@ public Builder customLut(@NonNull java.nio.Buffer buffer, int dimension) {
...
         }
 
+        /**
+         * Enables or disables SIMD fast math optimizations during LUT generation.
+         *
+         * @param fastMath true to enable fast math (default), false to use standard scalar math.
+         * @return This Builder, for chaining calls
+         */
+        @NonNull
+        public Builder fastMath(boolean fastMath) {
```
romainguy, line 586: Is this truly necessary?

filament/src/details/ColorGrading.cpp
```diff
@@ -971,4 +1024,75 @@ float4 FColorGrading::hdrColorAt(Builder const& builder, Config const& config,
     return {v, 0.0f};
 }
 
+#if defined(__ARM_NEON)
+
+UTILS_NOINLINE
+void FColorGrading::generateDefaultLUTNeon(void* data, Config const& config) noexcept {
```
romainguy: Now this makes me feel we should do this for all cases, even if partially. A good chunk of the LUT generation loop could easily be vectorized I bet (and we could ignore some of the features like the night adaptation and keep those scalar).

filament/src/details/ColorGradingNeon.h
```diff
@@ -0,0 +1,305 @@
...
+ * Vectorized sRGB Opto-Electronic Transfer Function (OETF).
+ * Evaluates the piece-wise linear/exponential sRGB transfer function on four RGB color
+ * values simultaneously. Evaluated branchlessly using vector selects.
+ * 
+ * @param cg_r Reference to vector of Red channels, modified in-place.
+ * @param cg_g Reference to vector of Green channels, modified in-place.
+ * @param cg_b Reference to vector of Blue channels, modified in-place.
+ */
+UTILS_ALWAYS_INLINE
+inline void v_oetf_sRGB(float32x4_t& cg_r, float32x4_t& cg_g, float32x4_t& cg_b) {
```
romainguy, line 153: This should be done in `OETF_sRGB` in ColorSpaceUtils.h so we benefit everywhere? Same for `EOTF_sRGB`. These are used inb `Color::sRGBToLinear` for instance.

android/filament-android/src/main/java/com/google/android/filament/ColorGrading.java
pixelflinger: Well, because the optimized path does affect precision a bit, it could be useful to have an opt-out -- that's why it's enabled by default.  It's also needed for unit-test.
And finally, because the optimized path is not available on all platforms (only armv8), it's useful to be able to turn it off, if exact output is needed (e.g. for tests).

filament/src/details/ColorGrading.cpp
pixelflinger: The vectorization essentially falls apart as soon as you have a function call in the loop. So it makes it really hard to do.  We would have to write vectorized versions of all tone mappers too, which would be hard to maintain and would bloat the code.
We could however, remove the 32 size constraint (must be POT and multiple 4).
The intention here was to speed-up the default case, which we assume most people will unintentionally use.

filament/src/details/ColorGradingNeon.h
pixelflinger: The problem is that those methods process 4 values at a time (this does 4x3 "pow"), and OETF_sRGB /  EOTF_sRGB, obviously only process one.
What we could do is unroll the main loop by a factor 4, and call this in the case of sRGB.

android/filament-android/src/main/java/com/google/android/filament/ColorGrading.java
romainguy: Sure but I don't think it's super useful for users of Filament? It's not like ColorGrading is making any particular promise about precision to begin with. Makes sense for testing

filament/src/details/ColorGrading.cpp
romainguy: I understand the intention but it's also a bummer if it becomes 4x slower if you just want to add a bit of saturation. Would we have to vectorize all tone mappers? We could do the easy one (and honestly you already did the most complicated one, by far).

filament/src/details/ColorGradingNeon.h
romainguy: Ah, yeah for Color::sRGBToLinear we would implement the conversion on a single `float32x4` (since r, g, and b are processed the same way).

filament/src/details/ColorGrading.cpp
pixelflinger: So we could probably do the whole "hasAdjustments" passes in one NEON block, because it doesn't depend on any function calls. To avoid code bloat, we could take a slightly slower path with function calls, something like this:
```
if (super_fast_path) {
  generate_default_lut_neon();
} if (fast_path) {
  for (i ...) {
  adjustments_neon(); 
  (*neon_tonemapper)(); // we'd need to decide which tonemappers we neonize
  oetf_neon();
  }
}  else {
 cpp_path();
}
```

filament/src/details/ColorGradingNeon.h
pixelflinger: yes, we could filp the loop.

filament/src/details/ColorGrading.cpp
romainguy: Yes that's what I had in mind. Ideally even the tone mappers could be NEON since they never change really, and you already took care of the most complex one.

romainguy (approved):

filament/src/details/ColorGrading.cpp
```diff
@@ -805,45 +917,53 @@ FColorGrading::FColorGrading(FEngine& engine, const Builder& builder) {
...
                 FILAMENT_TRACING_NAME(FILAMENT_TRACING_CATEGORY_FILAMENT, "ColorGrading::job");
                 uint32_t const dim = config.lutDimension;
-                half4* UTILS_RESTRICT const p = static_cast<half4*>(data) + (b * dim * dim);
-                for (uint32_t g = 0; g < dim; g++) {
-                    for (uint32_t r = 0; r < dim; r++) {
-                        p[g * dim + r] = half4{hdrColorAt(builder, config, r, g, b)};
+                uint32_t const sliceCount = dim * dim;
+
+                constexpr uint32_t TILE_SIZE = 64;
+                float4 tile_buffer[TILE_SIZE];
```
poweifeng, line 941: nit: camelCase

poweifeng (approved):
albeit I've only looked closer at the API and not the actual neon code.

pixelflinger:
reviewers... I have squashed all the "NEON" optimizations into a single PR. This now adds a more comprehensive fast path, for almost all cases (exceptions are: FP16, non POT dimensions). Confidence that this works relies on the new unit tests and some manual verification.

romainguy (approved):

poweifeng (approved):

z3moon (approved):

PR #9982, opened 2026-05-12
Refactor: Replace Component Instances with Entities in SoAs

pixelflinger:
The gist of this change is to remove the RenderableManager::Instance and LightManager::Instance dependency in SceneData.  Basically, the SceneData (lights and renderable) SoA used to store an Instance to each component, instead we store the Entity itself and we resolve the Instance at the point we need it.  This change makes SceneData immune to adding/removing components.

For some data, we instead copy the component data we need directly into the SoA, for other, we resolve it using the Entity (depending on cost).

This is a pre-requisite to a future change where we want to keep the SceneData around.

There shouldn't be *any* behavior changes.

### DIFF
--- a/filament/src/details/Scene.cpp
+++ b/filament/src/details/Scene.cpp
@@ -224,7 +224,7 @@ void FScene::prepare(JobSystem& js,
             size_t const index = std::distance(first, p) + i;
             assert_invariant(index < sceneData.size());
 
-            sceneData.elementAt<RENDERABLE_INSTANCE>(index) = ri;
+            sceneData.elementAt<RENDERABLE_ENTITY>(index) = rcm.getEntity(ri);
             sceneData.elementAt<WORLD_TRANSFORM>(index)     = shaderWorldTransform;
             sceneData.elementAt<VISIBILITY_STATE>(index)    = visibility;
             sceneData.elementAt<SKINNING_BUFFER>(index)     = rcm.getSkinningBufferInfo(ri);
@@ -245,8 +245,7 @@ void FScene::prepare(JobSystem& js,
         }
     };
 
-    auto lightWork = [first = lightInstances.data(), &lcm, &tcm, &worldTransform,
-            &lightData](auto* p, auto c) {
+    auto lightWork = [first = lightInstances.data(), &lcm, &tcm, &worldTransform, &lightData](auto* p, auto c) {
         FILAMENT_TRACING_NAME(FILAMENT_TRACING_CATEGORY_FILAMENT, "lightWork");
         for (size_t i = 0; i < c; i++) {
             auto [li, ti] = p[i];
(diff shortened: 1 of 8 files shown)


filament/src/details/Scene.cpp
```diff
@@ -438,11 +439,11 @@ void FScene::prepareDynamicLights(const CameraInfo& camera,
     LightsUib* const lp = driver.allocatePod<LightsUib>(positionalLightCount);
 
     auto const* UTILS_RESTRICT directions       = lightData.data<DIRECTION>();
-    auto const* UTILS_RESTRICT instances        = lightData.data<LIGHT_INSTANCE>();
+    auto const* UTILS_RESTRICT instances        = lightData.data<LIGHT_ENTITY>();
```
poweifeng: nit: rename to `entities`

pixelflinger: done

poweifeng (approved):

filament/src/details/Scene.cpp
```diff
@@ -264,7 +264,8 @@ void FScene::prepare(JobSystem& js,
             assert_invariant(index < lightData.size());
             lightData.elementAt<POSITION_RADIUS>(index) = float4{ position.xyz, lcm.getRadius(li) };
             lightData.elementAt<DIRECTION>(index) = d;
-            lightData.elementAt<LIGHT_INSTANCE>(index) = li;
+            lightData.elementAt<SPOT_PARAMS>(index) = float2{lcm.getCosOuterSquared(li), lcm.getSinInverse(li)};
+            lightData.elementAt<LIGHT_ENTITY>(index) = li ? engine.getLightManager().getEntity(li) : utils::Entity{};
```
z3moon: maybe use captured `lcm` like above?

filament/src/details/Scene.cpp
```diff
@@ -326,9 +327,9 @@ void FScene::prepare(JobSystem& js,
         lightData.elementAt<DIRECTION>(0) = normalize(d);
         lightData.elementAt<SHADOW_DIRECTION>(0) = normalize(s);
         lightData.elementAt<SHADOW_REF>(0) = lsReferencePoint;
-        lightData.elementAt<LIGHT_INSTANCE>(0) = li;
+        lightData.elementAt<LIGHT_ENTITY>(0) = li ? engine.getLightManager().getEntity(li) : utils::Entity{};
```
z3moon: `lcm` here as well

z3moon (approved):

filament/src/details/Scene.cpp
pixelflinger: done

filament/src/details/Scene.cpp
pixelflinger: done

PR #9984, opened 2026-05-12
Address VK mem leak on skipFrame

anishmgoyal:
If a lot of frames are being skipped, it's possible for an accumulation of a) unsubmitted command buffer commands, and b) a long period between collectGarbage() calls. This ensures that when frames are skipped, any pending commands are flushed, and if there are a lot of flushes with no gc calls, that gc eventually is called every few ticks.

As per discussion yesterday. Wanted to open up discussion on this change and see if there's a better approach. A couple options considered:
* Adding a skipFrame method to DriverApi, so that backends can do any periodic cleanup they need to do
* Adding a way for engine.gc() to somehow also invoke driver's gc if a backend has a gc method
* Always call collectGarbage() in tick, and remove it from endFrame()
* add flush + gc to tick() if command buffer is still recording

This seemed like the least intrusive approach, that does still ensure that cleanup happens. Note: we can't use the current frame count for the tick count, as it does not advance when skipping frames.

### DIFF
--- a/filament/backend/src/vulkan/VulkanDriver.cpp
+++ b/filament/backend/src/vulkan/VulkanDriver.cpp
@@ -432,8 +432,9 @@ void VulkanDriver::tick(int) {
     mCommands.updateFences();
 
     // If the Renderer is skipping a lot of frames, it's possible for
-    // memory to accumulate. This ensures that gc does still run in
-    // these cases.
+    // commands to accumulate, since we only flush command buffers if
+    // driver.flush() is called, or in endFrame(). This ensures that
+    // flush and gc run regularly, even during skipped frames.
     if (++mTicksSinceLastGc == MAX_TICKS_BETWEEN_GC) {
         flush();
         collectGarbage(); // Resets mTicksSinceLastGc to 0.
--- a/filament/backend/src/vulkan/VulkanDriver.cpp
+++ b/filament/backend/src/vulkan/VulkanDriver.cpp
@@ -64,6 +64,8 @@ namespace filament::backend {
 
 namespace {
 
+static constexpr uint8_t MAX_TICKS_BETWEEN_GC = 3;
+
 VmaAllocator createAllocator(VkInstance instance, VkPhysicalDevice physicalDevice,
         VkDevice device) {
     VmaAllocator allocator;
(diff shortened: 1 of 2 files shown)


anishmgoyal:
CC @rafadevai

poweifeng (approved):
Please wait for Mathias' review

filament/src/details/Renderer.cpp
```diff
@@ -302,6 +302,8 @@ void FRenderer::skipFrame(uint64_t vsyncSteadyClockTimeNano) {
 
     // Gives the backend a chance to execute periodic tasks. This must be called before
     // the frame skipper.
+    // Also, flush in case any commands have accumulated during skipped frames.
+    driver.flush();
```
pixelflinger: I don't think we should put the flush here, if you need to flush, just do it in your tick().

pixelflinger (changes requested):

filament/src/details/Renderer.cpp
anishmgoyal: gotcha. i'll keep the tick count logic in though, does that make sense? so it'll be every 3 ticks it'll force a flush + gc if none has happened

anishmgoyal: @pixelflinger updated!

filament/backend/src/vulkan/VulkanDriver.cpp
```diff
@@ -428,6 +430,14 @@ void VulkanDriver::terminate() {
 
 void VulkanDriver::tick(int) {
     mCommands.updateFences();
+
+    // If the Renderer is skipping a lot of frames, it's possible for
+    // memory to accumulate. This ensures that gc does still run in
```
rafadevai: instead of memory, say some operation are in pending state and not are being processed because we rely on endFrame to be called.

rafadevai (approved):

poweifeng (approved):

filament/src/details/Renderer.cpp
pixelflinger: yes I'll let powei comment on the vk side of things -- you guys are the experts :-) sgtm.

pixelflinger (approved):

PR #9988, opened 2026-05-13
Add parameter to set colored penumbrae in shadows

romainguy:
A colored penumbra can be added to shadows by using the new material
parameter `coloredPenumbra` and setting it to true.  When this effect
is enabled, the transition zone from full shadow to fully lit (the
penumbra) is tinted using a color assumed to be the result of subsurface
scattering. How this color is defined depends on the shading model
(see below for details). To maximize the visual appearance of this
effect, see the material and lighting setup recommendations below.

The color of the penumbra varies based on the `shadingModel`:
- `lit`: the color is derived by saturating the computed diffuse color,
  itself based on `baseColor`
- `subsurface`: the color is based on `subsurfaceColor`
- `cloth`: the color is based on `subsurfaceColor`
- `unlit`: does not apply
- `specularGlossiness`: same as `lit`

This effect is always enabled when the shading model is set to
`subsurface`.

Before/after animation:
colored_penumbra


romainguy:
@pixelflinger

libs/filamat/include/filamat/MaterialBuilder.h
```diff
@@ -307,9 +303,6 @@ class UTILS_PUBLIC MaterialBuilder : public MaterialBuilderBase {
     //! Set the name of this material.
     MaterialBuilder& name(const char* name) noexcept;
 
-    //! Set the file name of this material file. Used in error reporting.
-    MaterialBuilder& fileName(const char* name) noexcept;
```
show50726, line 311: Is there any reason of removing this?

pixelflinger: yeah, it was not implemented! this has been like that since 2020, so I think we're okay. @bejado ?

pixelflinger:
@trie94 do we need to do something for this change w.r.t. the new versioning?

libs/viewer/src/ViewerGui.cpp
```diff
@@ -974,7 +974,7 @@ void ViewerGui::updateUserInterface() {
         if (ImGui::CollapsingHeader("Shadows")) {
             ImGui::Checkbox("Enable shadows", &light.enableShadows);
             int mapSize = light.sunlight.shadowOptions.mapSize;
-            ImGui::SliderInt("Shadow map size", &mapSize, 32, 1024);
+            ImGui::SliderInt("Shadow map size", &mapSize, 512, 4096);
```
pixelflinger: I want to keep the very very small shadowmap. Also technially 4096 can fail on some implementations. So I would keep the max to 2048 (gles minspec).

libs/filamat/src/shaders/ShaderGenerator.cpp
```diff
@@ -107,9 +104,9 @@ void ShaderGenerator::generateSurfaceMaterialVariantDefines(io::sstream& out,
                 material.refractionMode != RefractionMode::NONE);
         if (material.refractionMode != RefractionMode::NONE) {
             CodeGenerator::generateDefine(out, "REFRACTION_MODE_CUBEMAP",
-                    uint32_t(RefractionMode::CUBEMAP));
```
pixelflinger, line 110: so elsewhere in the code we've been using the ctor-style cast for non-pointer types. generally we only use static_cast for pointer types.

libs/filamat/include/filamat/MaterialBuilder.h
```diff
@@ -876,35 +876,33 @@ class UTILS_PUBLIC MaterialBuilder : public MaterialBuilderBase {
...
-    bool isLit() const noexcept { return mShading != filament::Shading::UNLIT; }
+    bool isLit() const noexcept { return mShading != Shading::UNLIT; }
 
     utils::CString mMaterialName;
     utils::CString mCompilationParameters;
 
     class ShaderCode {
     public:
-        void setLineOffset(size_t offset) noexcept { mLineOffset = offset; }
+        void setLineOffset(const size_t offset) noexcept { mLineOffset = offset; }
```
pixelflinger: I try tot use the `type const` syntax everywhere for new code (I know it's not always consistant because of existing code).

libs/filamat/include/filamat/MaterialBuilder.h
```diff
@@ -876,35 +876,33 @@ class UTILS_PUBLIC MaterialBuilder : public MaterialBuilderBase {
...
-    bool checkLiteRequirements() noexcept;
-
     bool checkMaterialLevelFeatures(MaterialInfo const& info) const noexcept;
 
     void writeCommonChunks(ChunkContainer& container, MaterialInfo& info) const noexcept;
     void writeSurfaceChunks(ChunkContainer& container) const noexcept;
 
     bool generateShaders(
             utils::JobSystem& jobSystem,
-            const std::vector<filamat::Variant>& variants, ChunkContainer& container,
```
pixelflinger, line 894: this one is a bit dangerous inside libfilamat because there are indeed a filament::Variant and a filament::Variant. I think if it compiles it's fine though.

pixelflinger (changes requested):

libs/viewer/src/ViewerGui.cpp
romainguy: I doubt you can run Android with a max texture size of 2048.

libs/filamat/include/filamat/MaterialBuilder.h
romainguy: There was no implementation indeed :/

pixelflinger (approved):
thanks!

docs/Materials.md.html
poweifeng, line 1: Can you move these changes to `docs_src/src_markdeep` and the image would need to be in `src_mdbook/src/images/`

poweifeng (changes requested):

NEW_RELEASE_NOTES.md
```diff
@@ -6,3 +6,5 @@
 appropriate header in [RELEASE_NOTES.md](./RELEASE_NOTES.md).
 
 ## Release notes for next branch cut
+- New `coloredPenumbra` material property can be used to simulate light scattering in shadow 
+  transitions. See Filament's material guide for more information
```
poweifeng, line 10: I think we want `[⚠️ New Material Version]` here?

romainguy: I don’t think so? It’s backwards compatible. It doesn’t introduce new uniforms etc. Older versions of filament would ignore the new chunk in the filamat file when using mat info for instance, but the shaders will work.

docs/Materials.md.html
romainguy: How does this work nowadays btw? Are the files in docs/ modified/committed by a GitHub workflow?

show50726 (approved):

samples/material_sandbox.cpp
```diff
@@ -125,24 +127,24 @@ static void printUsage(char* name) {
     std::cout << usage;
 }
 
-static int handleCommandLineArgments(int argc, char* argv[], Config* config) {
+static int handleCommandLineArgments(const int argc, char* argv[], Config* config) {
     static constexpr const char* OPTSTR = "ha:vps:i:d:c:";
-    static const utils::getopt::option OPTIONS[] = {
```
poweifeng, line 130: if possible, would you mind keeping the namespace `utils::getopt`?  It helps to disambiguate it from either the system getopt of the third_party/getopt.  See #9796

NEW_RELEASE_NOTES.md
poweifeng: ack'd

samples/material_sandbox.cpp
romainguy: Will do.

poweifeng (approved):

PR #9989, opened 2026-05-13
Add missing async method for VertexBuffer

z3moon:
Previously, an addition of `createVertexBufferAsync` was intentionally omitted because what it performs at backend is trivial. However, this missing piece can cause a crash if a VertexBuffer object is destroyed while an associated async operation is in the JobQueue because the synchronous destruction makes the handle invalid immediately.

Add the `createVertexBufferAsync` implementation so that all VertexBuffer-relevant API calls are sequentially processed.


pixelflinger (approved):

bejado (approved):
Metal LGTM

PR #9991, opened 2026-05-13
Cpu decoded yuv workaround

phoenixxxx:
On Qualcomm, CPU decoded frames from the Android decoder are strided. Although the VK driver allows that VkMemory to back a VkImage, sampling it yields black or green. We are adding a manual conversion and CPU to GPU upload path to circumvent this.

SIZEGUARD_BYPASS


filament/backend/include/backend/platforms/VulkanPlatform.h
```diff
@@ -425,6 +425,17 @@ class VulkanPlatform : public Platform, utils::PrivateImplementation<VulkanPlatf
          * Ycbcr y chroma offset
          */
         VkChromaLocation yChromaOffset;
+
+        /*
+         * YUV is software decoded (YV12 or 8Cb8Cr8_420)
+         */
```
pixelflinger, line 430: would like more details about what it means to be "software decoded". I mean the implication of it? Is it that the buffer needs CPU access? i.e. is it more about CPU access or CPU/GPU sync, than about "being software decoded"?

filament/backend/include/backend/platforms/VulkanPlatform.h
```diff
@@ -425,6 +425,17 @@ class VulkanPlatform : public Platform, utils::PrivateImplementation<VulkanPlatf
...
+       bool isSoftwareDecodedYUV;
+
+       /*
+        * Adding an explicit field for chroma conversion
+        * requirement.
+        */
+       bool isChromaConversionRequired;
```
pixelflinger, line 439: what does that mean? conversion from what to what?

filament/backend/src/vulkan/platform/VulkanPlatformAndroid.cpp
```diff
@@ -294,55 +368,87 @@ VulkanPlatform::ExternalImageMetadata VulkanPlatformAndroid::extractExternalImag
...
-    metadata.allocationSize = properties.allocationSize;
-    metadata.memoryTypeBits = properties.memoryTypeBits;
+        metadata.allocationSize = properties.allocationSize;
+        metadata.memoryTypeBits = properties.memoryTypeBits;
+
+        metadata.isChromaConversionRequired =
+            metadata.format == VK_FORMAT_UNDEFINED ||
+            fvkutils::isVKYcbcrConversionFormat(metadata.format);
+
+        // After more discussion we decided to take the driver's suggestion for these fields.
```
poweifeng: Let's leave out the "after more discussion we decided"

filament/backend/src/vulkan/platform/VulkanPlatformAndroid.cpp
```diff
@@ -294,55 +368,87 @@ VulkanPlatform::ExternalImageMetadata VulkanPlatformAndroid::extractExternalImag
...
-    };
-    VkAndroidHardwareBufferPropertiesANDROID properties = {
-        .sType = VK_STRUCTURE_TYPE_ANDROID_HARDWARE_BUFFER_PROPERTIES_ANDROID,
-        .pNext = &formatInfo,
-    };
-    VkResult result = vkGetAndroidHardwareBufferPropertiesANDROID(getDevice(), buffer, &properties);
-    FILAMENT_CHECK_POSTCONDITION(result == VK_SUCCESS)
+        VkAndroidHardwareBufferFormatPropertiesANDROID formatInfo = {
+            .sType = VK_STRUCTURE_TYPE_ANDROID_HARDWARE_BUFFER_FORMAT_PROPERTIES_ANDROID,
+        };
```
poweifeng, line 377: why did you move this into the `    if (__builtin_available(android 26, *)) {` block?

filament/backend/src/vulkan/VulkanDriver.cpp
poweifeng, line 832: Are you saying that this assumption is false?  That the passed in width and height is different than the one deduced from the metadata?

filament/backend/src/vulkan/VulkanDriver.cpp
```diff
@@ -1399,7 +1471,7 @@ void VulkanDriver::updateStreams(CommandStream* driver) {
                 if (!texture) {
                     auto externalImage = fvkutils::createExternalImageFromRaw(mPlatform, image, false);
                     auto metadata = mPlatform->extractExternalImageMetadata(externalImage);
-                    auto imgData = mPlatform->createVkImageFromExternal(externalImage);
+                    auto imgData = mPlatform->createVkImageFromExternal(externalImage, texture->width, texture->height);
```
z3moon, line 1497: this will crash as `texture == nullptr` in this block

filament/backend/src/vulkan/VulkanDriver.cpp
```diff
@@ -64,6 +64,69 @@ namespace filament::backend {
...
+    uint32_t const chromaHeight = h / 2;
+    uint32_t const totalBufferSize = yPlaneSize + (chromaWidth * chromaHeight * 2);
+
+    VkBuffer stagingBuffer = texture->getYUVStagingBuffer();
+    VkDeviceMemory stagingMemory = texture->getYUVStagingMemory();
+    VkDevice device = platform->getDevice();
+
+    // CPU mapped data
+    void* mappedData;
+    vkMapMemory(device, stagingMemory, 0, totalBufferSize, 0, &mappedData);
```
z3moon: I think it'd be safer to check the return value of this call.

filament/backend/src/vulkan/platform/VulkanPlatformAndroid.cpp
```diff
@@ -442,22 +550,104 @@ VulkanPlatform::ImageData VulkanPlatformAndroid::createVkImageFromExternal(
...
+
+        // Staging buffer used as a source for the buffer to image blit AND AHB* cpu copy
+        VkBufferCreateInfo const bufferInfo = {
+            .sType = VK_STRUCTURE_TYPE_BUFFER_CREATE_INFO,
+            .size = totalBufferSize,
+            .usage = VK_BUFFER_USAGE_TRANSFER_SRC_BIT,
+            .sharingMode = VK_SHARING_MODE_EXCLUSIVE,
+        };
+        VkBuffer stagingBuffer;
+        vkCreateBuffer(device, &bufferInfo, VKALLOC, &stagingBuffer);
```
z3moon: return value check

filament/backend/src/vulkan/platform/VulkanPlatformAndroid.cpp
```diff
@@ -442,22 +550,104 @@ VulkanPlatform::ImageData VulkanPlatformAndroid::createVkImageFromExternal(
...
+
+        VkMemoryAllocateInfo const bufAllocInfo = {
+            .sType = VK_STRUCTURE_TYPE_MEMORY_ALLOCATE_INFO,
+            .allocationSize = bufMemReqs.size,
+            .memoryTypeIndex = bufMemTypeIndex,
+        };
+        VkDeviceMemory stagingMemory;
+        // @TODO: When this CL gets stubmitted to github, we need to revisit this
+        vkAllocateMemory(device, &bufAllocInfo, VKALLOC, &stagingMemory);
+        vkBindBufferMemory(device, stagingBuffer, stagingMemory, 0);
```
z3moon: these two as well

filament/backend/src/vulkan/platform/VulkanPlatformAndroid.cpp
```diff
@@ -442,22 +550,104 @@ VulkanPlatform::ImageData VulkanPlatformAndroid::createVkImageFromExternal(
...
+            .mipLevels = 1,
+            .arrayLayers = metadata.layers,
+            .samples = metadata.samples,
+            .tiling = VK_IMAGE_TILING_OPTIMAL,
+            .usage = metadata.usage, // Correctly contains DST | SAMPLED based on our metadata tweak
+            .sharingMode = VK_SHARING_MODE_EXCLUSIVE,
+            .initialLayout = VK_IMAGE_LAYOUT_UNDEFINED,
+        };
+        VkImage finalTextureImage;
+        vkCreateImage(device, &imageInfo, VKALLOC, &finalTextureImage);
```
z3moon: return value check

filament/backend/src/vulkan/platform/VulkanPlatformAndroid.cpp
```diff
@@ -442,22 +550,104 @@ VulkanPlatform::ImageData VulkanPlatformAndroid::createVkImageFromExternal(
...
+                VK_MEMORY_PROPERTY_DEVICE_LOCAL_BIT);
+
+        VkMemoryAllocateInfo const imgAllocInfo = {
+            .sType = VK_STRUCTURE_TYPE_MEMORY_ALLOCATE_INFO,
+            .allocationSize = imgMemReqs.size,
+            .memoryTypeIndex = imgMemTypeIndex,
+        };
+        VkDeviceMemory finalTextureMem;
+        vkAllocateMemory(device, &imgAllocInfo, VKALLOC, &finalTextureMem);
+        vkBindImageMemory(device, finalTextureImage, finalTextureMem, 0);
```
z3moon: ditto

filament/backend/src/vulkan/platform/VulkanPlatformAndroid.cpp
```diff
@@ -268,22 +278,86 @@ VulkanPlatformAndroid::ExternalImageDescAndroid VulkanPlatformAndroid::getExtern
...
+                    uvRowDst[col * 2 + 0] = uRowSrc[col * uPixelStride];
+                    uvRowDst[col * 2 + 1] = vRowSrc[col * vPixelStride];
+                }
+            }
+            AHardwareBuffer_unlock(buffer, nullptr);
+            return true;
+        }
+        LOG(ERROR) << "[VulkanPlatformAndroid::copyExternalImageToMemory] AHardwareBuffer_lockPlanes failed";
+    } else {
+        LOG(ERROR) << "no fallback path in extractExternalImageMetadata [Android 26]";
```
z3moon: copyExternalImageToMemory instead

z3moon (dismissed):

filament/backend/include/backend/platforms/VulkanPlatform.h
phoenixxxx: I can add more comments directly in the code, but we can also discuss what more information is needed here. On Android, some .mp4 older formats can be encoded in ways that the GPU doesn't support. For these paths, the fallback will be on the CPU decoder.
So the AHB* that stores this frame will be CPU accessible (although it will also have GPU access flags). This is a source of confusion and why we though the GPU should have been able to render it. The fact that the memory is GPU accessible, doesn't imply that the GPU will sample it.
In particular for strided memory the Qualcomm hardware does not support GPU texture sampling. Which means reading that image on the GPU returns green/black.

filament/backend/include/backend/platforms/VulkanPlatform.h
phoenixxxx: This is just the standard Vulkan YUV to RGB chroma conversion definition. All images in the pixel shader have to be in the RGB space, so for all YUV images, a YUV to RGB conversion structure is required.

Since this is part of the ExternalImageMetadata, it technically also supports RGB external images. This was added to be clear about the type of processing this external image requires.

anishmgoyal: my only thought here is, maybe it makes more sense to specify the specific conversion? e.g. https://docs.vulkan.org/refpages/latest/refpages/source/VkSamplerYcbcrModelConversion.html <- maybe we have a filament enum that maps these out?

filament/backend/include/backend/platforms/VulkanPlatform.h
```diff
@@ -434,10 +445,19 @@ class VulkanPlatform : public Platform, utils::PrivateImplementation<VulkanPlatf
...
+    virtual bool copyExternalImageToMemory(ExternalImageHandleRef image, void* dstData,
+            uint32_t width, uint32_t height) const {
+        return false;
+    }
+
     struct ImageData {
         struct Bundle {
             VkImage image = VK_NULL_HANDLE;
             VkDeviceMemory memory = VK_NULL_HANDLE;
+            // For CPU decoded YUV images we need a CPU staging buffer
```
anishmgoyal, line 459: is it possible for us to use VulkanStagePool::Segment for this?

filament/backend/src/vulkan/platform/VulkanPlatformAndroid.cpp
```diff
@@ -76,6 +76,16 @@ bool isFormatSrgb(VkFormat format) {
   return format == VK_FORMAT_R8G8B8A8_SRGB || format == VK_FORMAT_R8G8B8_SRGB;
 }
 
+#ifndef AHARDWAREBUFFER_FORMAT_YV12
+#define AHARDWAREBUFFER_FORMAT_YV12 0x32315659 //842094169
+#endif
+bool isSoftwareDecodedYUV(uint32_t format, uint64_t usage){
+    bool const isYV12 = (format == AHARDWAREBUFFER_FORMAT_YV12);
+    bool const isFlexYUV = (format == AHARDWAREBUFFER_FORMAT_Y8Cb8Cr8_420);
+    bool const hasCpuWrite = (usage & AHARDWAREBUFFER_USAGE_CPU_WRITE_MASK) != 0;
```
anishmgoyal, line 85: i'm curious - have we attempted to allocate an ahardwarebuffer that isn't cpu writeable, and copying the buffer using those apis when ingesting it?

not that it's something we need to change, just trying to better understand

filament/backend/include/backend/platforms/VulkanPlatform.h
rafadevai: this should be renamed to isCpuVisible

filament/backend/include/backend/platforms/VulkanPlatform.h
rafadevai: why do we need this? this can already be inferred from externalFormat

filament/backend/src/vulkan/platform/VulkanPlatformAndroid.cpp
```diff
@@ -351,15 +457,16 @@ VulkanPlatform::ImageData VulkanPlatformAndroid::createVkImageFromExternal(
     VkDevice const device = getDevice();
     VkPhysicalDevice const physicalDevice = getPhysicalDevice();
     auto buildImage = [&](ExternalImageMetadata const& metadata) {
-        bool const isExternal = metadata.externalFormat != 0;
+        bool const isExternalFormat = metadata.isChromaConversionRequired &&
```
anishmgoyal, line 459: do we no longer care if an external format num is provided?

filament/backend/src/vulkan/platform/VulkanPlatformAndroid.cpp
```diff
@@ -268,22 +278,86 @@ VulkanPlatformAndroid::ExternalImageDescAndroid VulkanPlatformAndroid::getExtern
     };
 }
 
-VulkanPlatform::ExternalImageMetadata VulkanPlatformAndroid::extractExternalImageMetadata(
-        ExternalImageHandleRef image) const {
-    auto const* fvkExternalImage = static_cast<ExternalImageVulkanAndroid const*>(image.get());
+bool VulkanPlatformAndroid::copyExternalImageToMemory(
```
rafadevai: this should be renamed to something related to YUV since it only work for that type of format

filament/backend/src/vulkan/platform/VulkanPlatformAndroid.cpp
```diff
@@ -268,22 +278,86 @@ VulkanPlatformAndroid::ExternalImageDescAndroid VulkanPlatformAndroid::getExtern
...
+        ExternalImageMetadata metadata = {};
+
         AHardwareBuffer_Desc bufferDesc;
         AHardwareBuffer_describe(buffer, &bufferDesc);
+
         metadata.width = bufferDesc.width;
         metadata.height = bufferDesc.height;
         metadata.layers = bufferDesc.layers;
+        metadata.samples = VK_SAMPLE_COUNT_1_BIT;
+        metadata.isSoftwareDecodedYUV = isSoftwareDecodedYUV(bufferDesc.format, bufferDesc.usage);
```
rafadevai: this function should stay as it is, the only thing that is okay to add is the `isSoftwareDecodedYUV` entry.

the metadata of an externalImage is a fixed value and depending on the format it shouldnt return something different.

If you want the metadata for another type of buffer, make a call with a different AHB or externalImage

filament/backend/src/vulkan/platform/VulkanPlatformAndroid.cpp
```diff
@@ -442,22 +550,104 @@ VulkanPlatform::ImageData VulkanPlatformAndroid::createVkImageFromExternal(
...
-                << "vkBindImageMemory error=" << static_cast<int32_t>(result);
+            << "vkBindImageMemory error=" << static_cast<int32_t>(result);
+
         return memory;
     };
 
     VulkanPlatform::ImageData::Bundle internal = {}, external = {};
-    auto img = buildImage(metadata);
-    auto mem = allocMem(img, metadata);
+    if (metadata.isSoftwareDecodedYUV) {
```
rafadevai: all of the copy and creation of the new AHB should happen outside of this function

filament/backend/src/vulkan/platform/VulkanPlatformAndroid.cpp
```diff
@@ -442,22 +550,104 @@ VulkanPlatform::ImageData VulkanPlatformAndroid::createVkImageFromExternal(
...
+
+        uint32_t const w = metadata.width;
+        uint32_t const h = metadata.height;
+        uint32_t const yPlaneSize = w * h;
+        uint32_t const chromaPlaneSize = (w / 2) * (h / 2);
+        uint32_t const totalBufferSize = yPlaneSize + (chromaPlaneSize * 2);
+
+        // Staging buffer used as a source for the buffer to image blit AND AHB* cpu copy
+        VkBufferCreateInfo const bufferInfo = {
```
anishmgoyal, line 563: i see - so we can't get this buffer from stagepool because it's happening outside of driver. i'm curious though, do we want to at least try to use vma, since we use it for all of our other allocations?

filament/backend/src/vulkan/platform/VulkanPlatformAndroid.cpp
anishmgoyal: this is a *very* long function. can we break out things like allocating buffers into different functions?

filament/backend/src/vulkan/platform/VulkanPlatformAndroid.cpp
anishmgoyal: we should also at LEAST log errors that pop up during the conversion process.

filament/backend/include/backend/platforms/VulkanPlatformAndroid.h
```diff
@@ -49,8 +49,11 @@ class VulkanPlatformAndroid : public VulkanPlatform, public AndroidNdk {
 
     ExternalImageMetadata extractExternalImageMetadata(
             ExternalImageHandleRef image) const override;
+    bool copyExternalImageToMemory(ExternalImageHandleRef image, void* dstData,
```
anishmgoyal: let's make this a protected function

filament/backend/include/backend/platforms/VulkanPlatformAndroid.h
```diff
@@ -49,8 +49,11 @@ class VulkanPlatformAndroid : public VulkanPlatform, public AndroidNdk {
...
+                                   uint32_t width, uint32_t height) const override;
 
-    ImageData createVkImageFromExternal(ExternalImageHandleRef image) const override;
+    ImageData createVkImageFromExternal(ExternalImageHandleRef image,
```
anishmgoyal, line 55: i kinda want this to be hidden from outside of filament somehow, but i'm not sure what we could do to ensure that, since i think this is called from driver?

filament/backend/include/backend/platforms/VulkanPlatform.h
pixelflinger: actually I don't think "isCpuVisible" is correct either. It sounds like it's more a "isGpuSamplable" or "isGpuAccessible".

filament/backend/include/backend/platforms/VulkanPlatform.h
pixelflinger: if this is needed (which apparently we're not sure), I would like the name to be descriptive about the convertion and the comment have a link to what the YUV format is.

phoenixxxx: So for ExternalImageHandleRef we have 3 use cases:
YUV Case: 
- Directly sampleable YUV
- Workaround for YUV that are in a strided format (this is what we are adding here)
- RGB Case
Both YUV cases require a VkSamplerYcbcrModelConversion object. The RGB case is not currently exercised only for external apps that SysUI composites.

filament/backend/include/backend/platforms/VulkanPlatform.h
phoenixxxx: The name was changed the name to something more explicit.

filament/backend/src/vulkan/VulkanDriver.cpp
phoenixxxx: You are correct, this path is not working, 513573228 is tracking the fix. I have added a comment.

filament/backend/src/vulkan/platform/VulkanPlatformAndroid.cpp
phoenixxxx: Yes we did think about that, the issue is that at the point when you create a GPU only ahardwarebuffer, you are effectively doing something similar to copying to a GPU only image.

filament/backend/src/vulkan/platform/VulkanPlatformAndroid.cpp
phoenixxxx: we do, it's just that prior to this we ONLY checked it to know if a YUV conversion was necessary.

It's hard to follow and that's because the nature of this workaound is convoluted. So an VkExternalFormatANDROID is a construct that allows the vendor to hide the details about their YUV format. But there is one type of YUV that is not internal and that is defined by Vulkan. BOTH will require a conversion to YUV ONLY the internal format (the path we take 99% of the time) will ALSO require a VkExternalFormatANDROID added to the VkExternalMemoryImageCreateInfo.

filament/backend/src/vulkan/platform/VulkanPlatformAndroid.cpp
poweifeng: also, does it have to happen on the platform?  Can it be done in the driver?  And maybe use staging pool?

filament/backend/src/vulkan/platform/VulkanPlatformAndroid.cpp
poweifeng: would prefer something like `readExternalImage` as it's doing a read into cpu memory

filament/backend/src/vulkan/platform/VulkanPlatformAndroid.cpp
```diff
@@ -268,22 +278,86 @@ VulkanPlatformAndroid::ExternalImageDescAndroid VulkanPlatformAndroid::getExtern
     };
 }
 
-VulkanPlatform::ExternalImageMetadata VulkanPlatformAndroid::extractExternalImageMetadata(
-        ExternalImageHandleRef image) const {
-    auto const* fvkExternalImage = static_cast<ExternalImageVulkanAndroid const*>(image.get());
+bool VulkanPlatformAndroid::copyExternalImageToMemory(
+        ExternalImageHandleRef image, void* dstData, uint32_t w, uint32_t h) const {
```
poweifeng, line 282: Is this saying you are copying a subregion of the image?  Hence the w, and h?  If so, don't you also need to specify an offset?

filament/backend/src/vulkan/platform/VulkanPlatformAndroid.cpp
```diff
@@ -268,22 +278,86 @@ VulkanPlatformAndroid::ExternalImageDescAndroid VulkanPlatformAndroid::getExtern
...
```
poweifeng, line 283: should probably assert that is intepreable bits before you read.  That is - `isSoftwareDecodedYUV` or something similar.

filament/backend/src/vulkan/VulkanDriver.cpp
poweifeng: Can we maybe just leverage `update3DImage`?

filament/backend/src/vulkan/platform/VulkanPlatformAndroid.cpp
phoenixxxx: yeah it was done part of the other comment from Rafa.

filament/backend/src/vulkan/platform/VulkanPlatformAndroid.cpp
phoenixxxx: What errors we you thinking about? I think several logs were added from a comment from Sungun.

filament/backend/include/backend/platforms/VulkanPlatformAndroid.h
phoenixxxx: The driver is calling this method. I don't think it can be protected?

filament/backend/include/backend/platforms/VulkanPlatformAndroid.h
phoenixxxx: Yeah this is the way the Fronted End wanted the API to be. Where the user has to explicitely call a createTextureExternalImage2R with a opaque structure.

filament/backend/src/vulkan/platform/VulkanPlatformAndroid.cpp
phoenixxxx: If AHardwareBuffer_describe isn't available, you are severley limited in the logic of this code. Returning partially formed metadata seemed incorrect.

filament/backend/src/vulkan/VulkanDriver.cpp
phoenixxxx: I am saying that we should ensure that they match. We are getting geometry information (width, height) from two sources (AHB* and caller), we should ensure both are saying the same thing, otherwise there is an issue somewhere.

filament/backend/include/backend/platforms/VulkanPlatformAndroid.h
anishmgoyal: ahh nvm then

filament/backend/src/vulkan/platform/VulkanPlatformAndroid.cpp
phoenixxxx: We are copying each plane of the YV12 buffer into a NV12 Vulkan buffer. There is one conversion from muti planar to tighgtly packed.

filament/backend/src/vulkan/platform/VulkanPlatformAndroid.cpp
phoenixxxx: Correct, we check it in texture->isYUVStaging()

filament/backend/src/vulkan/VulkanDriver.cpp
phoenixxxx: Perhaps, how so?

filament/backend/src/vulkan/platform/VulkanPlatformAndroid.cpp
anishmgoyal: things like vkCreateBuffer, vkGetBufferMemoryRequirements, etc. those functions return statuses, we should log or do something if they fail

filament/backend/src/vulkan/platform/VulkanPlatformAndroid.cpp
```diff
@@ -268,22 +278,86 @@ VulkanPlatformAndroid::ExternalImageDescAndroid VulkanPlatformAndroid::getExtern
...
+            uint32_t uPixelStride = planes.planes[1].pixelStride;
+
+            uint8_t* vSrc = static_cast<uint8_t*>(planes.planes[2].data);
+            uint32_t vRowStride = planes.planes[2].rowStride;
+            uint32_t vPixelStride = planes.planes[2].pixelStride;
+
+            for (uint32_t row = 0; row < chromaHeight; ++row) {
+                uint8_t* uvRowDst = uvDst + (row * w);
+                uint8_t* uRowSrc = uSrc + (row * uRowStride);
+                uint8_t* vRowSrc = vSrc + (row * vRowStride);
```
pixelflinger, line 324: you may be able to help the compiler vectorizer by specifying __restrict__ (or UTILS_RESTRICT) on there 3 pointers.

pixelflinger (commented):
LGTM to me, I'm deferring to @poweifeng for final approval.

filament/backend/src/vulkan/platform/VulkanPlatformAndroid.cpp
```diff
@@ -442,22 +549,131 @@ VulkanPlatform::ImageData VulkanPlatformAndroid::createVkImageFromExternal(
...
+        result = vkBindImageMemory(device, image, memory, 0);
         FILAMENT_CHECK_POSTCONDITION(result == VK_SUCCESS)
-                << "vkBindImageMemory error=" << static_cast<int32_t>(result);
+            << "vkBindImageMemory error=" << static_cast<int32_t>(result);
+
         return memory;
     };
 
+    // Software Decoded (Staging) Path Lambdas (following desing pattern from buildImage)
+    auto buildStagingBuffer = [&](uint32_t size) {
```
anishmgoyal, line 561: nit: generally, i think it's cleaner to have these as functions defined in an anonymous namespace at the top of the cpp file, but i think this is fine for this change

anishmgoyal (approved):

filament/backend/src/vulkan/platform/VulkanPlatformAndroid.cpp
poweifeng: I'm saying why can't w, h be inferred?

filament/backend/src/vulkan/platform/VulkanPlatformAndroid.cpp
poweifeng: I mean it should be asserted here.

filament/backend/src/vulkan/VulkanDriver.cpp
poweifeng: I mean the same as above, why do you need to pass width and height, if it can be inferred from metadata.

filament/backend/src/vulkan/platform/VulkanPlatformAndroid.cpp
poweifeng: was this resolved?

filament/backend/src/vulkan/VulkanDriver.cpp
poweifeng: Teh same map, unmap upload is done in update3DImage.  Could they be consolidated.

filament/backend/include/backend/platforms/VulkanPlatform.h
poweifeng: I don't think i has been; at least not in this latest commit

poweifeng (approved):
discussed offline.  

lgtm after the lower-hanging items (like renaming) can be addressed.

PR #9994, opened 2026-05-14
Fix warnings when generating the web pages

romainguy:
- Install mdbook 0.4.*. The script would otherwise install mdbook 0.5 and the current configuration and theme used by Filament are not compatible and rely on deprecated APIs
- Fix  tags that were triggering dozens of warning when buidling with mdbook


romainguy:
@poweifeng

poweifeng (approved):

PR #9995, opened 2026-05-14
Feature: Introduce View-specific caching via SceneCacheData

pixelflinger:
Motivation:
Previously, caching intermediate rendering state (like culling results and visible lights) was difficult due to the engine's architecture. Scene contains the flat list of objects but cannot cache state because it depends on the camera. View has the camera, but caching scene-data there means the cache is lost if the scene changes. Renderer is transient and only receives a view at render time. We needed a cache location dependent on the intersection of both a Scene and a View.

Implementation:
Introduced SceneCacheData to hold view-dependent scene state (e.g., RenderableSoa, LightSoa, and contact shadow flags).

Instead of storing this globally, Scene now owns a collection of these caches, keyed by View. When a View is assigned to a Scene, it registers itself and is granted a dedicated cache slot. This allows Renderer and ShadowMapManager to fetch state specific to the current View without interference.

Performance Characteristics:
This architecture is explicitly optimized for the common case: a single Scene rendered by multiple Views simultaneously. Each view maintains persistent state within the scene. However, calling View::setScene() explicitly invalidates the cache for that view. Therefore, using a single View to sequentially render multiple different scenes is intentionally not optimized and will result in cache recreation.

### DIFF
--- a/filament/src/ShadowMapManager.cpp
+++ b/filament/src/ShadowMapManager.cpp
@@ -175,7 +175,7 @@ ShadowMapManager::ShadowTechnique ShadowMapManager::update(
     calculateTextureRequirements(engine, view, lightData);
 
     // Compute scene-dependent values shared across all shadow maps
-    ShadowMap::SceneInfo const info{ view.getScene()->getRenderableData(), view.getVisibleLayers() };
+    ShadowMap::SceneInfo const info{ view.getRenderableData(), view.getVisibleLayers() };
 
     shadowTechnique |= updateCascadeShadowMaps(
             engine, view, cameraInfo, renderableData, lightData, info);
@@ -238,8 +238,7 @@ FrameGraphId<FrameGraphTexture> ShadowMapManager::render(FEngine& engine, FrameG
         FView& view, CameraInfo const& mainCameraInfo,
         float4 const& userTime) noexcept {
 
-    FScene* scene = view.getScene();
-    assert_invariant(scene);
+    assert_invariant(view.getScene());
 
     // make a copy here, because it's a very small structure
     VsmShadowOptions const& vsmShadowOptions = view.getVsmShadowOptions();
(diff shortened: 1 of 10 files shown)


filament/src/details/View.cpp
```diff
@@ -202,7 +202,20 @@ FView::FView(FEngine& engine)
 
 FView::~FView() noexcept = default;
 
+void FView::setScene(FScene* scene) {
+    if (mScene != scene) {
+        invalidateCache(mScene);
+        mScene = scene;
+        if (mScene) {
+            mScene->registerView(this);
+            mCurrentViewCache = mScene->mViewCaches[this].get();
```
z3moon: maybe `registerView` just returns `SceneCacheData*` and setScene assigns it directly. no map lookup and no need for friend access in this way.

filament/src/details/View.h
```diff
@@ -103,6 +103,13 @@ class FView : public View {
     explicit FView(FEngine& engine);
     ~FView() noexcept;
 
+    FScene::RenderableSoa& getRenderableData() const noexcept {
+        return mCurrentViewCache->renderableData;
```
z3moon, line 108: maybe `assert_invariant(mCurrentViewCache);` here and below.

also these return non-const references from const methods. do we need overloads or is this intentional?

filament/src/ShadowMapManager.cpp
```diff
@@ -345,7 +344,7 @@ FrameGraphId<FrameGraphTexture> ShadowMapManager::render(FEngine& engine, FrameG
             [=, this,
                     passBuilder = passBuilder,
                     &engine = const_cast<FEngine /*const*/ &>(engine), // FIXME: we want this const
-                    &view = view]
+                    &view = const_cast<FView const&>(view)]
```
z3moon: nit: const_cast is usually for const removal. so maybe static_cast

filament/src/details/View.cpp
pixelflinger: done

filament/src/details/View.h
pixelflinger: I think the non-const from const method is appropriate here, because we're not changing the state of View, it's more like View is acting as a manager for the cache; it just holds the cache entries.

filament/src/ShadowMapManager.cpp
pixelflinger: done

filament/src/details/View.cpp
```diff
@@ -202,7 +202,19 @@ FView::FView(FEngine& engine)
 
 FView::~FView() noexcept = default;
 
+void FView::setScene(FScene* scene) {
+    if (mScene != scene) {
+        invalidateCache(mScene);
```
poweifeng: If mScene == null I don't think you want to call invalidateCache?

poweifeng: I guess if mScene is null, then mCurrentViewCache would be null too.

poweifeng (approved):

filament/src/details/View.cpp
pixelflinger: good points. fixed.

PR #9998, opened 2026-05-15
skybox: fix sun disk parameters

pixelflinger:
- the public API is always using degrees, but there was some inconsistencies.

- sanitize the sun disk calculations so that the halo size can be set independently from the sun angular radius without blowing up.

- update gltf_viewer default values to better match Earth's Sun (i.e. keep the default values, which are now working).


poweifeng (approved):
You'll need to add `RDIFF_ACCEPT_NEW_GOLDENS` to the commit message I think

PR #9910, opened 2026-04-17
Validate filamesh buffer bounds in MeshReader

TristanInSec:
## Summary

- Add a `dataSize` parameter to `loadMeshFromBuffer` to enable bounds validation
- Validate all header fields (`vertexSize`, `indexSize`, `parts`, `materialCount`, `nameLength`) against the buffer boundary before advancing the parse pointer
- Add overflow checks for compressed index and vertex buffer size calculations
- Fix missing `fd < 0` check in `loadMeshFromFile`
- Update all callers (samples, iOS samples, JS bindings, tests) to pass the buffer size

## Test plan

- [ ] Existing `test_filamesh` tests pass (both NonInterleaved and Interleaved)
- [ ] Samples compile and load built-in meshes correctly
- [ ] Truncated or malformed `.filamesh` files return empty Mesh instead of reading past the buffer

PR #10005, opened 2026-05-17
gltfio: fix integer overflow in uint8 index buffer conversion

TristanInSec:
When converting uint8 index buffers to uint16, the size is doubled via
`size * 2`. For very large buffers this multiplication can overflow,
leading to an undersized allocation followed by an out-of-bounds write.

Add an overflow check before the multiplication and skip the buffer
if the size would overflow.


TristanInSec:

PR #10006, opened 2026-05-17
gltfio: validate inverse bind matrices buffer bounds in createSkins

TristanInSec:
The createSkins function copies inverse bind matrices from the glTF
source buffer without verifying the buffer contains enough data for
the requested joint count. A malformed asset with a large joints_count
or small buffer can cause an out-of-bounds read.

Add bounds checking that validates the source buffer size against
the required copy size before the memcpy. Also adds null checks
for buffer_view and buffer pointers to guard against malformed input.

Follows the same validation pattern used in normalizeSkinningWeights
(added in 8d16ad88).


TristanInSec:

PR #10007, opened 2026-05-17
smol-v: add output bounds checking to SPIR-V decoder

TristanInSec:
The SMOL-V decoder allocates the output buffer based on a size field
in the encoded header, then writes decoded SPIR-V words without
checking whether the output pointer stays within bounds. If the
encoded data expands beyond the declared size, writes go past the
allocated buffer.

Add a bounds-checked write function (smolv_Write4Safe) and use it
for all output writes in Decode(). The function returns false on
overflow, causing the decoder to abort cleanly.

Also fixes unchecked smolv_Read4 return values during header parsing
and adds a missing bounds check before reading the swizzle byte.

### DIFF
--- a/third_party/smol-v/source/smolv.cpp
+++ b/third_party/smol-v/source/smolv.cpp
@@ -1116,6 +1116,15 @@ static void smolv_Write4(uint8_t*& buf, uint32_t v)
 	buf += 4;
 }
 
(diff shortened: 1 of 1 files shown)


romainguy:
This is a 3rd party library, so the fix should be done upstream, and then merged into Filament.

TristanInSec:
Fix submitted upstream: https://github.com/aras-p/smol-v/pull/16

PR #10008, opened 2026-05-17
ktx1: validate numberOfFaces in KTX1 deserialization

TristanInSec:
The KTX1 spec requires numberOfFaces to be either 1 (non-cubemap) or
6 (cubemap). The deserializer accepts any value, but the flatten()
indexing function assumes the face count is exactly 1 or 6. When
numberOfFaces is between 2 and 5, the blob array is sized for N faces
but flatten() computes indices assuming 6, causing out-of-bounds access.

Add validation that rejects numberOfFaces values other than 0, 1, or 6
during deserialization, consistent with the KTX1 specification.


TristanInSec:

PR #10013, opened 2026-05-18
Update material documentation

z3moon:
to include
 - non-index rendering
 - attribute-less / procedural rendering

### DIFF
--- a/docs_src/src_markdeep/Materials.md.html
+++ b/docs_src/src_markdeep/Materials.md.html
@@ -2259,15 +2259,6 @@ material {
 }
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-### Non-indexed rendering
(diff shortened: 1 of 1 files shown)


docs_src/src_markdeep/Materials.md.html
```diff
@@ -2259,6 +2259,104 @@
 }
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
+### Non-indexed rendering
```
romainguy: Is this section relevant to materials? It mentions C++ APIs but not shader APIs.

z3moon: I think you're right. This doesn't need to be in the doc as it's not relevant. Let me remove it.

pixelflinger (approved):

PR #10014, opened 2026-05-18
vk: add offset to image blit api

poweifeng:
- Offset wasn't accounted for previously.
 - Failed backend test LoadImageTest.UpdateImage2D
 - Also fixed buffer overflow bug in adustedMemcpy where the pbd is larger than the mapped buffer.

### DIFF
--- a/filament/backend/src/vulkan/VulkanTexture.cpp
+++ b/filament/backend/src/vulkan/VulkanTexture.cpp
@@ -233,6 +233,23 @@ VkImageUsageFlags getUsage(VulkanContext const& context, uint8_t samples,
 
 void adjustedMemcpy(void* mapped, PixelBufferDescriptor const& p, size_t width, size_t height,
         size_t depth) {
+    // This method copies a sub-region from the source PixelBufferDescriptor (p)
+    // into the destination mapped memory.
+    //
+    // Source Buffer (p):                                Destination Memory (mapped):
+    // <------------------ pbdStride ------------------->
+    // +------------------------------------------------+
+    // |                                     ^          |
+    // |            padding / top            | p.top    |
+    // |                                     v          |
+    // |  <-- p.left --> <------- width -------->       |    <------- width -------->
+    // |  +------------+------------------------+-------+ ^  +----------------------+ ^
+    // |  | (skipped)  |         (copy)         | (skip)| |  |        (copy)        | |
+    // |  |            |                        |       | |  |                      | | height
+    // |  +------------+------------------------+-------+ v  +----------------------+ v
+    // |                                                |
+    // +------------------------------------------------+
+    //
     uint8_t* buf = (uint8_t*) p.buffer;
     size_t const pixelSize = PixelBufferDescriptor::computeDataSize(p.format, p.type, 1, 1, 1);
     size_t const pbdStride = p.stride ? p.stride : width;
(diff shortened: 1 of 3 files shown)


filament/backend/src/vulkan/VulkanTexture.cpp
```diff
@@ -245,11 +245,12 @@ void adjustedMemcpy(void* mapped, PixelBufferDescriptor const& p, size_t width,
         // Size of a row to write
         size_t const writeSize = std::min(pbdStride - p.left, width) * pixelSize;
 
+        size_t const rowsToCopy = std::min(height, pbdHeight - p.top);
```
pixelflinger: is it guaranteed in some way that `pbdHeight >= p.top`? Or is the intention of the `min`, that if it overflows, it will fall back to `height`?

filament/backend/src/vulkan/VulkanTexture.cpp
```diff
@@ -245,11 +245,12 @@ void adjustedMemcpy(void* mapped, PixelBufferDescriptor const& p, size_t width,
...
         for (size_t z = 0; z < depth; z++) {
-            for (size_t y = p.top; y < pbdHeight; y++) {
+            for (size_t y = 0; y < rowsToCopy; y++) {
                 uint8_t* buf = (uint8_t*) p.buffer +
-                               ((p.left * pixelSize) + (y * pbdRowSize) + (z * pbdLayerSize));
-                uint8_t* curMapped = (uint8_t*) mapped + ((y - p.top) * rowSize + z * layerSize);
+                               ((p.left * pixelSize) + ((y + p.top) * pbdRowSize) + (z * pbdLayerSize));
+                uint8_t* curMapped = (uint8_t*) mapped + (y * rowSize + z * layerSize);
```
pixelflinger, line 277: I would really like it if there was intermediate variables with readable names here.

Or maybe even an ASCII art about what all these values are.

filament/backend/src/vulkan/VulkanTexture.cpp
poweifeng: Yes, I think `pdbHeight` should be greater than `top`.  The `std::min` part seems wrong, the pbd should be asserted to be larger than the copy size.  I've added a assert_invariant to sanity check.

filament/backend/src/vulkan/VulkanTexture.cpp
poweifeng: ascii diagram added.

pixelflinger (approved):

filament/backend/src/vulkan/VulkanTexture.cpp
```diff
@@ -647,20 +665,36 @@ void VulkanTexture::updateImageWithBlit(const PixelBufferDescriptor& data, uint3
...
+    VkOffset3D const srcRect[2] { {0, 0, 0}, {int32_t(width), int32_t(height), 1} };
+    VkOffset3D const dstRect[2] {
+        { int32_t(xoffset), int32_t(yoffset), int32_t(dstZOffset) },
+        { int32_t(xoffset + width), int32_t(yoffset + height), int32_t(dstZOffset + dstDepth) }
+    };
 
     VkImageAspectFlags const aspect = getImageAspect();
 
     VkImageBlit const blitRegions[1] = {{
         .srcSubresource = { aspect, 0, 0, 1 },
```
phoenixxxx, line 697: Should the "1" be "layerCount"?

filament/backend/src/vulkan/VulkanTexture.cpp
```diff
@@ -647,20 +665,36 @@ void VulkanTexture::updateImageWithBlit(const PixelBufferDescriptor& data, uint3
...
+    if (    target == SamplerType::SAMPLER_2D_ARRAY ||
+            target == SamplerType::SAMPLER_CUBEMAP ||
+            target == SamplerType::SAMPLER_CUBEMAP_ARRAY) {
+        layer = zoffset;
+        layerCount = depth;
+        dstZOffset = 0;
+        dstDepth = 1;
+    }
+
+    VkOffset3D const srcRect[2] { {0, 0, 0}, {int32_t(width), int32_t(height), 1} };
```
phoenixxxx, line 688: Should this be "int32_t(dstZOffset + dstDepth)"? To match the destination.

poweifeng: no because the src is a 2D staging image - single layer.

filament/backend/src/vulkan/VulkanTexture.cpp
poweifeng: see above comment

filament/backend/src/vulkan/VulkanTexture.cpp
poweifeng: actually, you're pointing out that this won't work for 3D images, which is correct.  I'll address that in a follow-up.

rafadevai (approved):

PR #10017, opened 2026-05-19
Allocate exact-sized interleaved (strided) buffers

### LINKED ISSUE #10009: OpenGL backend: slightly over-allocates interleaved buffer size by adding full stride to last element offset
**Describe the bug**

For interleaved vertex buffers, e.g. a vertex buffer with a few custom fields. e.g. we have this vertex array with vertexCount= 845, stride = 68B. Therefore, actual data size needed: 845x68B=57460B.

e.g. in RenderDoc:

Image

We see we have a byte length of 57516 in this interleaved buffer. If in RenderDoc I click through Buffer 271 and see its lifetime, we see something strange. It was initially created with a size of 57516 bytes (more than needed). But then it was filled with the right amount of data (57460 bytes).

Image

**Expected behavior**

I'd expect to allocate the required memory, and no more. Ideally we could get by with a single glBufferData call, rather than a glBufferData+glBufferSubData call.

I think Filament also has some logic to detect if we're "updating the entire array" and call glBufferData rather than glBufferSubData:

https://github.com/google/filament/blob/25263572d24a322ae9f771dd9d48e209ff896b2a/filament/backend/src/opengl/OpenGLDriver.cpp#L3316-L3323

That's not working in this case. I'm new to this space, but my understanding is this enables "buffer orphaning"; i.e. the driver allocating new memory rather than reusing the old memory. I think I'd expect that to work too. I'd be worried that some drivers might have to copy-on-write the backing array for a glBufferSubData call if the data's in use, where they wouldn't for a glBufferData call.

This was seen on an Android Automotive seahawk device, but that probably doesn't matter.

**Suspected Cause**

https://github.com/google/filament/blob/25263572d24a322ae9f771dd9d48e209ff896b2a/filament/src/details/VertexBuffer.cpp#L327

**Suspected fix**

```
    -    const size_t end = offset + mVertexCount * stride;
    +    const size_t elementSize = Driver::getElementTypeSize(mAttributes[attributeIndex].type);
    +    const size_t end = offset + (mVertexCount - 1) * stride + elementSize;
```

I see there's a check that mVertexCount > 0 here: https://github.com/google/filament/blob/25263572d24a322ae9f771dd9d48e209ff896b2a/filament/src/details/VertexBuffer.cpp#L170. Can we rely on that? Or is it stripped out in some builds?

mhansen:
Fixes #10009

### DIFF
--- a/filament/src/details/VertexBuffer.cpp
+++ b/filament/src/details/VertexBuffer.cpp
@@ -323,9 +323,13 @@ FVertexBuffer::FVertexBuffer(FEngine& engine, const Builder& builder)
         const uint32_t offset = mAttributes[attributeIndex].offset;
         const uint8_t stride = mAttributes[attributeIndex].stride;
         const uint8_t slot = mAttributes[attributeIndex].buffer;
-        const size_t end = offset + mVertexCount * stride;
+        const size_t elementSize = Driver::getElementTypeSize(mAttributes[attributeIndex].type);
+        // mVertexCount is always > 0, checked by precondition, so this should
+        // not underflow.
+        const size_t end = offset + (mVertexCount - 1) * stride + elementSize;
+        const size_t rounded = ((end + stride - 1) / stride) * stride;
         assert_invariant(slot < mBufferCount);
-        bufferSizes[slot] = std::max(bufferSizes[slot], end);
+        bufferSizes[slot] = std::max(bufferSizes[slot], rounded);
     };
 
     if (!mBufferObjectsEnabled) {
--- a/filament/test/test_VertexBuffer.cpp
+++ b/filament/test/test_VertexBuffer.cpp
@@ -1,0 +1,190 @@
+/*
+ * Copyright (C) 2026 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "MockDriver.h"
+#include "details/Engine.h"
+#include "CommandStreamDispatcher.h"
+
+#include <gtest/gtest.h>
+#include <gmock/gmock.h>
+
+#include "private/backend/CommandStream.h"
+#include "private/backend/Driver.h"
+
+#include <filament/Engine.h>
+#include <filament/VertexBuffer.h>
+#include <backend/DriverEnums.h>
+#include <backend/Platform.h>
+
+#include <atomic>
+#include <chrono>
+#include <condition_variable>
+#include <mutex>
+#include <thread>
+
+using namespace filament;
+using namespace filament::backend;
+using ::testing::_;
+using ::testing::Return;
+
+namespace {
+
+class CustomTestDriver : public MockDriver {
+public:
+    explicit CustomTestDriver(std::atomic<uint32_t>* sizeDest, std::mutex* mutex, std::condition_variable* cv)
+            : mSizeDest(sizeDest), mMutex(mutex), mCv(cv) {
+        // Set up default nice mock behaviors to pass Engine validation checks
+        ON_CALL(*this, getFeatureLevel())
+                .WillByDefault(Return(FeatureLevel::FEATURE_LEVEL_1));
+        ON_CALL(*this, isTextureFormatSupported(_))
+                .WillByDefault(Return(true));
+        ON_CALL(*this, isRenderTargetFormatSupported(_))
+                .WillByDefault(Return(true));
+        ON_CALL(*this, isTextureFormatMipmappable(_))
+                .WillByDefault(Return(true));
+        ON_CALL(*this, isTextureFormatFilterable(_))
+                .WillByDefault(Return(true));
+        ON_CALL(*this, getMaxTextureSize(_))
+                .WillByDefault(Return(2048));
+    }
+    ~CustomTestDriver() override = default;
+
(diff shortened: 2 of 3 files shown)


pixelflinger:
@mhansen the unit tests seem to fail under ASAN.

mhansen:
Whoops, sorry for the noise. I'll look at the tests tomorrow.

filament/src/details/VertexBuffer.cpp
```diff
@@ -324,7 +324,8 @@ FVertexBuffer::FVertexBuffer(FEngine& engine, const Builder& builder)
         const uint32_t offset = mAttributes[attributeIndex].offset;
         const uint8_t stride = mAttributes[attributeIndex].stride;
         const uint8_t slot = mAttributes[attributeIndex].buffer;
-        const size_t end = offset + mVertexCount * stride;
+        const size_t elementSize = Driver::getElementTypeSize(mAttributes[attributeIndex].type);
+        const size_t end = offset + (mVertexCount - 1) * stride + elementSize;
```
z3moon, line 329: This has an underflow issue. e.g., if mVertexCount == 0, `mVertexCount - 1` will underflow for the type uint32_t.

pixelflinger, line 329: The correct calculation is `mVertexCount * stride`.

Indeed the existing code was wrong because it added the "offset" to the size calculation, the offset should be irrelevant. If we compare this to a `std::vector<Foo>`, `stride` plays the role of `sizeof(Foo)` and `mVertexCount` plays the role of `.size()`.  `offset`  plays the role of `offsetof(Foo, field)`, it shouldn't be part of the size calculation.

filament/test/test_VertexBuffer.cpp
```diff
@@ -0,0 +1,164 @@
...
+using ::testing::Return;
+
+namespace {
+
+class CustomTestDriver : public MockDriver {
+public:
+    explicit CustomTestDriver(std::atomic<uint32_t>* sizeDest, std::mutex* mutex, std::condition_variable* cv)
+            : mSizeDest(sizeDest), mMutex(mutex), mCv(cv) {
+        // Set up default nice mock behaviors to pass Engine validation checks
+        ON_CALL(*this, getFeatureLevel())
```
pixelflinger, line 50: neat!

filament/test/test_VertexBuffer.cpp
```diff
@@ -0,0 +1,164 @@
...
+    std::condition_variable mCv;
+    CustomTestPlatform mCustomPlatform;
+};
+
+TEST_F(VertexBufferTest, InterleavedBufferSize) {
+    // Replicate the user's interleaved vertex buffer scenario:
+    // VertexCount = 845, Stride = 68
+    // Attr 0: offset 0, size 16 (FLOAT4)
+    // Attr 1: offset 56, size 12 (FLOAT3)
+    // Correct Size: 56 + (845 - 1) * 68 + 12 = 57460 bytes.
```
pixelflinger: no, correct size is actually `845*68 = 57460`.

The original bug was that filament calculated it as `56 + 845*68`.

pixelflinger (changes requested):

filament/test/test_VertexBuffer.cpp
mhansen: Yeah, I was happy to be able to write tests for this. I was wondering if we should put a fake/mock engine somewhere more common, this feels maybe more generally useful. But for now this is the only test using it so I thought let's leave it here.

filament/src/details/VertexBuffer.cpp
mhansen: D'oh. You're right, I completely bamboozled myself here.

mhansen: The simplification is `offset + elementSize = stride`, so `offset + (mVertexCount - 1) * stride + elementSize` collapses to `stride + (mVertexCount - 1) * stride` collapses to `mVertexCount * stride`. Sorry, I did this in an accretive not simplififying way.

mhansen: On https://github.com/google/filament/issues/10009 I asked about this:

I see there's a check that mVertexCount > 0 here: https://github.com/google/filament/blob/25263572d24a322ae9f771dd9d48e209ff896b2a/filament/src/details/VertexBuffer.cpp#L170. Can we rely on that? Or is it stripped out in some builds?

filament/test/test_VertexBuffer.cpp
mhansen: I'll update the comment as you say (your suggestion is a simpler way to calculate it for this interleaved case).

filament/src/details/VertexBuffer.cpp
mhansen: Wait, no, I remember why I did this - I was worried about if attributes are struct-of-array. While in this case (array of struct), `mVertexCount * stride` is correct, filament also seems to support vertices that are offset, e.g. here we have position at bytes 0-23 and color at bytes 23-35:

https://github.com/google/filament/blob/180e91f5511675aa94afd5c206e0b0145876a37f/samples/vbotest.cpp#L72-L75

If we used mVertexCount * stride, this would ignore colour's starting offset of 24. And the max-sized attribute would be computed as 24, so we'd allocate a 24-byte buffer.

I've added a test demonstrating this. And also another test for what happens with 32-byte-padded rows (ensuring that we allocate for the padding too)

z3moon: Ah yeah we can rely on it then. Could you add comments about why it's safe?

pixelflinger: good catch! I completely agree.

filament/src/details/VertexBuffer.cpp
```diff
@@ -324,9 +324,11 @@ FVertexBuffer::FVertexBuffer(FEngine& engine, const Builder& builder)
...
+        const size_t rounded = ((end + stride - 1) / stride) * stride;
```
pixelflinger, line 330: yes perfect, this works.

pixelflinger (approved):

filament/src/details/VertexBuffer.cpp
mhansen: Done

z3moon (approved):

PR #10018, opened 2026-05-19
Make release split build as default

z3moon:
`-y release` now becomes the default option for ./build.sh.

Fix a conflict with clean options (-c or -C) as split build will always be executed for clean options otherwise.


build.sh
```diff
@@ -223,8 +224,8 @@ OSMESA_OPTION=""
 IOS_BUILD_SIMULATOR=false
 BUILD_UNIVERSAL_LIBRARIES=false
 
-ISSUE_SPLIT_BUILD=false
-SPLIT_BUILD_TYPE=""
+ISSUE_SPLIT_BUILD=true
```
pixelflinger, line 228: so there no way to set this to false anymore? maybe -y should take like a `none` option to force the build to how it was before?  do we even want a non split build option?

z3moon: I added `none` option to disable the split build and verified it worked as intended.

bejado (approved):

pixelflinger (approved):

PR #10024, opened 2026-05-19
debug: allow environment vars to override DebugRegistry defaults

poweifeng:


filament/src/details/DebugRegistry.cpp
```diff
@@ -16,23 +16,61 @@
...
+
+constexpr size_t MAX_PROPERTY_NAME_LENGTH = 128;
+constexpr size_t MAX_PROPERTY_VALUE_LENGTH = 128;
+
+#ifdef __ANDROID__
+std::string_view getPropertyFromEnvironment(char const* const name,
+        std::array<char, MAX_PROPERTY_VALUE_LENGTH>& storage) {
+    // since API 26 there is no limit to the property name
+    char propertyName[MAX_PROPERTY_NAME_LENGTH];
+    snprintf(propertyName, MAX_PROPERTY_NAME_LENGTH, "debug.%s", name);
```
z3moon, line 93: It looks like propertyName could be truncated if `storage` is already almost full. Or is it guaranteed that MAX_PROPERTY_NAME_LENGTH is sufficient for propertyName?

poweifeng: so `MAX_PROPERTY_NAME_LENGTH` is the length of a (name, value) pair.  And `storage` is for storing the value.  I don't think the size of one should affect the size of the other. 

For `MAX_PROPERTY_NAME_LENGTH`, it's more like an implicit contract that our debug property name should not exceed 128.

bejado (approved):

z3moon (approved):

filament/src/details/DebugRegistry.cpp
```diff
@@ -59,11 +97,97 @@ void const* FDebugRegistry::getPropertyAddress(const char* name) const noexcept
...
+                }
+                case INT: {
+                    int v;
+                    if (sscanf(val, "%d", &v) == 1) {
+                        if (*static_cast<int*>(p) != v) {
+                            *static_cast<int*>(p) = v;
+                            changed = true;
+                        }
+                    }
+                    break;
```
pixelflinger: it looks like a lot of those could be a template? (where you pass the format string).

pixelflinger (approved):

pixelflinger (changes requested):
I think we should print/log each variable that is being overriden (including in release).

filament/src/details/DebugRegistry.cpp
poweifeng: done

poweifeng:
> I think we should print/log each variable that is being overriden (including in release).

yes, this is done with LOG(INFO)

pixelflinger (approved):

PR #10028, opened 2026-05-19
Trust VkFormat from Vulkan for external images

anishmgoyal:
Right now, we have a hardcoded mapping for VkFormats that we expect from AHardwareBuffers, which can fail in the case of legacy formats that Android CTS expects us to support. The driver knows the right VkFormat to use, and we're currently using that format for validation. Instead, let's have that format now be the source-of-truth, and trust it - that way, any legacy formats that we don't anticipate do not crash the headset.

This would be using the VkFormat we already get from the driver. The difference now is, if the format is one we don't recognize, we'll still be able to provide some level of support for it. This will not affect android apps - it will only affect system processes.

Alternatives considered:
-  Add support for the format directly (HAL pixel format constants that are not exposed through AHardwareBuffer)
  - This felt like a bad approach as it would involve hardcoding internal formats, not anchored by a source of truth
- Remove the postcondition check
  - This might not work, as the underlying format wouldn't be set
 - Reverse lookups from VkFormat to get Filament format
   - I think we should defer on this, because these formats should not be accessible by Android apps, and even system apps likely don't use them. They only appear in some CTS tests. Any format that isn't in our old mapping is one that has long been deprecated. But we could use the VkFormat from the query to find the Filament format as well, if we wanted to go further from this change.

### DIFF
--- a/filament/backend/src/vulkan/platform/VulkanPlatformAndroid.cpp
+++ b/filament/backend/src/vulkan/platform/VulkanPlatformAndroid.cpp
@@ -90,64 +90,10 @@ bool isProtectedFromUsage(uint64_t usage) {
     return usage & AHARDWAREBUFFER_USAGE_PROTECTED_CONTENT;
 }
 
-std::pair<VkFormat, VkImageUsageFlags> getVKFormatAndUsage(const AHardwareBuffer_Desc& desc,
-        bool sRGB) {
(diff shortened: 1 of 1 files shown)


anishmgoyal:
cc @rafadevai

rafadevai (approved):

poweifeng (approved):

PR #10037, opened 2026-05-20
gltfio: Support textures bound to both sRGB and Linear parameters

### LINKED ISSUE #8449: Assertion failure when reusing texture in different material properties

**Describe the bug**
When loading a GLTF model that reuses the same texture across different material properties (e.g., the same texture used for both baseColorMap and occlusionMap), an assertion fails in Debug builds. The assertion `assert_invariant(info.bindings.size() == 0 || info.flags == flags)` in the texture loading code triggers when attempting to bind the same texture with different usage flags.

**To Reproduce**
1. Use Filament's GLTF viewer in Debug build
2. Load the attached [VirtualCity.glb](https://github.com/user-attachments/files/18817347/VirtualCity.zip) which reuses the same texture image for different material properties
3. Assertion triggers during texture binding phase

**Expected behavior**
The same texture should be able to be used across different material properties without triggering an assertion, as this is valid according to the GLTF specification.

**Logs**
```
FEngine (64 bits) created at 0000020AD3882660 (threading is enabled)
Vulkan device driver: NVIDIA 566.36
Selected physical device 'NVIDIA GeForce RTX 3050 Laptop GPU' from 2 physical devices. (vendor 0x10de, device 0x25a2, driver 0x8d890000, api 1.3)
Backend feature level: 3
FEngine feature level: 3
Requested linear format but KTX contains a sRGB format.
Requested linear format but KTX contains a sRGB format.
Texture Decoder has 18 background threads.
vkCreateSwapchain: 1024x640, VK_FORMAT_B8G8R8A8_UNORM, VK_COLOR_SPACE_SRGB_NONLINEAR_KHR, swapchain-size=3, identity-transform=true, depth=VK_FORMAT_D32_SFLOAT_S8_UINT, protected=false
PanicLog
in void __cdecl utils::panic(const char *,const char *,int,const char *) noexcept:34
in file [...]/filament/libs/utils/src/debug.cpp
reason: [...]/filament/libs/gltfio/src/FilamentAsset.cpp:119: failed assertion 'info.bindings.size() == 0 || info.flags == flags'
```

 - OS: Windows 11
 - GPU: NVIDIA GeForce RTX 3050 Laptop GPU
 - Backend: Vulkan

bejado:
Previously, gltfio assumed that each `cgltf_texture` slot could only ever be bound in a single color space, mapping it to a single Filament Texture format. If a single texture index was bound to parameters with conflicting color-space requirements (for example, an sRGB `baseColorTexture` and a Linear `occlusionTexture`), the loader would trigger an assertion panic.

This PR resolves this limitation by allowing a single `cgltf_texture` to resolve to separate sRGB and Linear Filament texture resources under the hood when conflicting color space requirements are requested.

Fixes #8449


libs/gltfio/src/ResourceLoader.cpp
```diff
@@ -834,17 +701,20 @@ std::pair<Texture*, CacheResult> ResourceLoader::Impl::getOrCreateTexture(FFilam
     TextureProvider* provider = foundProvider->second;
     assert_invariant(provider);
 
+    const size_t cacheIdx = (uint64_t) flags & 1;
```
poweifeng: can we make this a little more clear like 

```
    const size_t cacheIdx = any(flags & TextureFlags::sRGB) ? 1 : 0;
```

poweifeng (approved):

pixelflinger (approved):

z3moon (approved):

libs/gltfio/src/ResourceLoader.cpp
bejado: Yup great suggestion.

PR #10040, opened 2026-05-21
concurrency: Modernization and compile-time thread capability annotations

pixelflinger:
Audit, modernize, and compile-verify the concurrency design of the
filament core utility layer (`libs/utils/`), core filament, 
and the rendering backend (`filament/backend/`). 
This change activates Clang's static thread
analyzer globally across core and backend targets, translating runtime 
lock patterns into compiler-checked invariants to prevent future data 
race regressions.

Key Modifications:
1. Activated standard std::mutex annotations in Bionic libc++ by defining
   -D_LIBCPP_ENABLE_THREAD_SAFETY_ANNOTATIONS globally inside Android
   toolchains.
2. Created dynamic preprocessor protection in compiler.h to safely disable
   thread annotations on standard types on host/CI compilers where
   standard capability decorations are absent by default.
3. Restored blocking and scheduler-critical synchronization blocks
   (JobQueue, CommandBufferQueue, DriverBase) to standard std::mutex &
   std::condition_variable to preserve native Priority Inheritance (PI)
   and kernel sleep queue wakeups.
4. Manually inlined CV wait loops as C++ while loops (CommandBufferQueue,
   JobQueue, ShaderCompilerService, DriverBase) to bypass Clang lambda
   wait closure tracing limits, enabling 100% static UTILS_GUARDED_BY
   coverage on queue states.
5. Resolved two actual concurrent data races in utils/EntityManagerImpl.h
   by safely locking mFreeListLock during debug active entity map copy
   passes.
6. Restructured CompilerThreadPool::terminate() into lexical scoped
   LockGuard const blocks to eliminate manual locking, ensuring
   deadlock-free RAII thread shutdowns.
7. Cleaned, sorted, and layered includes across all 13 modified files to
   comply fully with Filament's alphabetical and layering guidelines.


filament/backend/src/CompilerThreadPool.cpp
```diff
@@ -54,36 +54,29 @@ void CompilerThreadPool::init(uint32_t threadCount,
...
-                            (!std::all_of( std::begin(mQueues), std::end(mQueues),
-                                    [](auto&& q) { return q.empty(); }));
-                });
+            while (true) {
+                UniqueLock lock(mQueueLock);
+                while (!mExitRequested && mQueues[0].empty() && mQueues[1].empty() && mQueues[2].empty()) {
+                    mQueueCondition.wait(lock);
+                }
+
+                if (mExitRequested) {
```
poweifeng: nit: UTILS_UNLIKELY?

filament/backend/src/CompilerThreadPool.cpp
```diff
@@ -54,36 +54,29 @@ void CompilerThreadPool::init(uint32_t threadCount,
...
-                    // execute the job without holding any locks
-                    lock.unlock();
-                    job();
-                }
+                Job job;
+                // use the first queue that's not empty
+                auto& queue = mQueues[0].empty() ? (mQueues[1].empty() ? mQueues[2] : mQueues[1]) : mQueues[0];
+                assert_invariant(!queue.empty());
+                std::swap(job, queue.front().second);
+                queue.pop_front();
```
poweifeng: so I guess this means that we can't put a object that will can be locked in lambdas anymore?

filament/backend/src/vulkan/VulkanDriver.cpp
```diff
@@ -1121,7 +1121,7 @@ void VulkanDriver::createSyncR(Handle<HwSync> sh, utils::ImmutableCString&& tag)
     }
 
     {
-        std::lock_guard<std::mutex> guard(sync->lock);
+        std::lock_guard guard(sync->lock);
```
poweifeng, line 1125: should this be LockGuard?

poweifeng (approved):

filament/backend/src/CompilerThreadPool.cpp
```diff
@@ -54,36 +54,29 @@ void CompilerThreadPool::init(uint32_t threadCount,
...
-                    std::swap(job, queue.front().second);
-                    queue.pop_front();
-
-                    // execute the job without holding any locks
-                    lock.unlock();
-                    job();
-                }
+                Job job;
+                // use the first queue that's not empty
+                auto& queue = mQueues[0].empty() ? (mQueues[1].empty() ? mQueues[2] : mQueues[1]) : mQueues[0];
```
bejado: Is there no way to keep the lambda and still appease Clang? That's really unfortunate

filament/backend/src/CompilerThreadPool.cpp
pixelflinger: yes, unless you use the annotation to disable the checks in that function. but in practice I found that it's not a common case.

filament/backend/src/CompilerThreadPool.cpp
pixelflinger: I fixed the unprotected access to mExitRequested and kept the lambda with the "do not check" annotation (I think it's ok in this case).

filament/backend/src/vulkan/VulkanDriver.cpp
pixelflinger: the backend implementations have not been annotated yet.

filament/backend/src/DriverBase.h
```diff
@@ -315,18 +322,18 @@ class DriverBase : public Driver {
...
-    std::mutex mFenceMutex;
-    std::atomic<bool> mHasUnrecoverableError{false};
+    mutable std::mutex mServiceThreadLock;
+    mutable std::condition_variable mServiceThreadCondition;
+    std::vector<std::tuple<CallbackHandler*, CallbackHandler::Callback, void*>> mServiceThreadCallbackQueue UTILS_GUARDED_BY(mServiceThreadLock);
+    bool mExitRequested UTILS_GUARDED_BY(mServiceThreadLock) = false;
+
+    mutable std::condition_variable mFenceCondition;
+    mutable std::mutex mFenceMutex;
+    std::atomic<bool> mHasUnrecoverableError UTILS_GUARDED_BY(mFenceMutex){false};
```
z3moon, line 337: maybe we could just remove UTILS_GUARDED_BY since it's not always true. e.g., `hasUnrecoverableError` doesn't require the mutex.

filament/backend/src/CompilerThreadPool.cpp
pixelflinger: done

filament/backend/src/DriverBase.h
pixelflinger: This one is tricky, we could go both ways. I think I prefer to see it as protected by the lock (because it participate to the condition variable).

PR #10046, opened 2026-05-22
reorg: Reorganize C++ header includes by logical layering

pixelflinger:
Implement a systematic include reordering policy to ensure clean, acyclic dependency hierarchies across the entire library suite. Formulate exact inclusion, syntax, sorting, and visual spacing guidelines to optimize compilation time and catch missing headers.

Key changes:
- Defined a 7-level include layering hierarchy: corresponding headers, Windows cleanup (unwindows.h), private local headers, internal proprietary libraries, third-party libraries, C++ standard library, C POSIX system headers, and legacy standard C headers.
- Standardized include syntax: mandatory angle brackets < > for all public and private library includes (e.g., <private/backend/...>), and quotes " " strictly for module-local headers under src/.
- Implemented flat-first alphabetical sorting and subdirectory-prefix spacing separation (adding newlines between groups) for private local includes.
- Created a permanent formatting tool under tools/reorganize_headers.py that dynamically parses CMakeLists.txt target linkage graphs, builds an acyclic dependency tree, and performs a topological sort to get the perfect layer priority dynamically without hardcoding target order.
- Added robust preprocessor safety checks in the formatter tool to isolate simple guarded includes (such as #if __EXCEPTIONS) and keep them intact while safely skipping complex macro files (e.g. ostream.cpp).
- Recursively reorganized all includes in libs/utils, libs/math, libs/filaflat, libs/filabridge, and filament/ targets.
- Corrected a pre-existing self-containment include bug in ResourceNode.h.
- Formally documented the updated standards and helper tool in CODE_STYLE.md.

All targets compile cleanly with zero warnings; all 230 unit tests pass.

### DIFF
--- a/filament/backend/include/backend/platforms/PlatformWGL.h
+++ b/filament/backend/include/backend/platforms/PlatformWGL.h
@@ -17,15 +17,16 @@
 #ifndef TNT_FILAMENT_BACKEND_OPENGL_OPENGL_PLATFORM_WGL_H
 #define TNT_FILAMENT_BACKEND_OPENGL_OPENGL_PLATFORM_WGL_H
 
-#include <stdint.h>
+#include <utils/unwindows.h>
+
+#include <backend/DriverEnums.h>
+#include <backend/platforms/OpenGLPlatform.h>
 
 #include <windows.h>
-#include "utils/unwindows.h"
-
-#include <backend/platforms/OpenGLPlatform.h>
-#include <backend/DriverEnums.h>
 
 #include <vector>
+
+#include <stdint.h>
 
 namespace filament::backend {
 
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,63 @@
+# Filament AI Coding & Agent Guidelines
+
+This document defines machine-readable instructions and technical guardrails for AI coding assistants, agents, and LLMs working on the Filament repository. All AI-generated code, refactorings, and cleanups must strictly adhere to these standards.
+
+---
+
+## Skill 1: Dynamic C++ Header Inclusion & Reordering
+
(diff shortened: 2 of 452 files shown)


romainguy:
> Formally documented the updated standards and helper tool in CODE_STYLE.md.

Would it be helpful to also formalize this in AGENTS.md or an agent skill for anyone using agents?

pixelflinger:
> > Formally documented the updated standards and helper tool in CODE_STYLE.md.
> 
> Would it be helpful to also formalize this in AGENTS.md or an agent skill for anyone using agents?

yes that's a good idea.

filament/backend/include/backend/platforms/PlatformWGL.h
```diff
@@ -17,16 +17,17 @@
...
 
-#include <stdint.h>
-
-#include <windows.h>
-#include "utils/unwindows.h"
+#include <utils/unwindows.h>
 
-#include <backend/platforms/OpenGLPlatform.h>
 #include <backend/DriverEnums.h>
+#include <backend/platforms/OpenGLPlatform.h>
```
poweifeng, line 23: so utils is above backends?  that's different than what I've been doig.

poweifeng (commented):
@pixelflinger can you check in your script to somewhere like `build` ? I think we can set up a CI workflow to run it against incoming PRs and notify the PR author.

poweifeng (approved):

filament/backend/include/backend/platforms/PlatformWGL.h
pixelflinger: ~yeah that's a mistake. good catch.~

`utils/unwindows.h` is special, it always needs to be first.

PR #10053, opened 2026-05-25
backend: workaround broken WebGL spec-constant UBO array sizes

Erkko68:
## Symptom

On Chromium-on-Windows and several Android Chromium configurations, every `glDrawElementsInstanced` call spams the WebGL2 validator:

```
GL_INVALID_OPERATION: glDrawElementsInstanced:
  It is undefined behaviour to use a uniform buffer that is too small.
```

On the affected drivers the draw is silently dropped, so any lit material is invisible. Chromium-on-macOS (ANGLE-Metal) tolerates the mismatch and renders correctly, which is why this had gone unnoticed.

This problem was noticed in one of my projects while testing a scene on different browsers, you can check the broken version on the following Github Pages -> Selection the button "Solar".

https://erkko68.github.io/filament-kmp/

Before
image002

After
image001

## Cause

ANGLE-on-D3D11 and some Android GLES drivers don't propagate `const int CONFIG_FOO = SPIRV_CROSS_CONSTANT_ID_<id>` into the uniform-block array size — they layout the block using the matc-baked default and disagree with the runtime-bound range.

Same shape as the bug originally fixed by [`fe3790cb9`](https://github.com/google/filament/commit/fe3790cb9) (#5859, "WASM: Set CONFIG_MAX_INSTANCES to a small value + hack the GLSL"). That hack was removed when spec constants were assumed portable.

## Fix

Spec-constant-aware revival of the 2022 hack: in `ShaderCompilerService::compileShaders`, gate on `#ifdef __EMSCRIPTEN__`, walk the shader source, and rewrite each symbolic array length to a literal (padded with spaces to keep byte count). Affects three arrays in shipped FL1 materials: `CONFIG_MAX_INSTANCES`, `CONFIG_FROXEL_BUFFER_HEIGHT`, `CONFIG_FROXEL_RECORD_BUFFER_HEIGHT`.

### DIFF
--- a/filament/backend/src/opengl/ShaderCompilerService.cpp
+++ b/filament/backend/src/opengl/ShaderCompilerService.cpp
@@ -42,6 +42,8 @@
 #include <array>
 #include <atomic>
 #include <cctype>
+#include <cstdio>
+#include <cstring>
 #include <iterator>
 #include <memory>
 #include <mutex>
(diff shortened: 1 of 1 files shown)


show50726:
Are you testing on HEAD? I wonder if #10000 fixes your issue.

Erkko68:
> Are you testing on HEAD? I wonder if #10000 fixes your issue.

You are right! Thanks

PR #10062, opened 2026-05-27
utils: Implement Epoch-Based Reclamation (EBR) for Entity Management 

pixelflinger:
This major architectural update introduces an Epoch-Based Reclamation (EBR)
system to the EntityManager. It delays the recycling of destroyed Entity IDs
until all active readers and component managers have advanced past the epoch
in which the destruction occurred.

The primary motivation for EBR is to decouple the lifecycle of the 32-bit
Entity ID from the dense tracking structures. By guaranteeing that an entity's
physical index is never safely recycled while any subsystem is still processing
its destruction, component managers can safely transition to tracking entities
purely by their raw index, dropping the generation count from their internal
bitsets. This paves the way for drastically shrinking the memory footprint
of structures like PagedArenaBitset (e.g., from 16KB down to 256B) in future
updates.

Key changes:
- Implement `advanceEpoch()`, `registerWatermark()`, and `reclaimSafeEpochs()`
  in EntityManager to orchestrate the global timeline.
- Introduce `catchupGarbage()` in SingleInstanceComponentManager to
  periodically query the central timeline for missed garbage and mark
  entities logically dead.
- Amortize the physical destruction of component data into a `gc()` phase,
  safely holding zombie components in dense arrays until they can be cleaned up.
- Prevent use-after-free ID collisions by keeping destroyed Entity indices
  in a purgatory epoch until the global safe threshold clears them.
- Introduce exhaustive concurrent stress tests to ensure thread-safety of
  the timeline and bitset pools.

### DIFF
--- a/libs/utils/include/utils/PagedArenaBitset.h
+++ b/libs/utils/include/utils/PagedArenaBitset.h
@@ -365,6 +365,9 @@ public:
 
     static PagedArenaBitset merge(const PagedArenaBitset& a, const PagedArenaBitset& b);
 
+    static PagedArenaBitset& mergeSpan(PagedArenaBitset* UTILS_RESTRICT UTILS_NONNULL out,
+            Slice<const PagedArenaBitset* const> inputs);
+
     // ----------------------------------------------------------------------------------------------------------------
     // difference
     // ----------------------------------------------------------------------------------------------------------------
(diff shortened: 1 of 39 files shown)


libs/utils/include/utils/PagedArenaBitset.h
```diff
@@ -394,13 +394,36 @@ class UTILS_PUBLIC PagedArenaBitset {
      */
     void extractTo(std::vector<uint32_t>& outBuffer) const;
 
+    /**
+     * @brief Iterates over all set bits in ascending numerical order.
+     * @tparam Func A callable type accepting a uint32_t.
+     * @param callback The function to execute for each set bit.
+     */
```
poweifeng: extraneous?

libs/utils/src/EntityManagerImpl.h
```diff
@@ -77,97 +85,155 @@ class UTILS_PRIVATE EntityManagerImpl : public EntityManager {
...
 
-        LockGuard const lock(mFreeListLock);
-        bool const alive = (getGeneration(e) == mGens[getIndex(e)]);
+        LockGuard const lock(mDestructionLock);
+        uint32_t const index = getIndex(e);
+
+        // 1. Is this entity the current tenant of this memory slot?
+        bool const alive = (getGeneration(e) == mGens[index]);
+
+        // RESTORED: Since both track the 32-bit ID, this is perfectly valid again.
```
poweifeng: `RESTORED`?

poweifeng (approved):

libs/utils/include/utils/PagedArenaBitset.h
pixelflinger: fixed

libs/utils/src/EntityManagerImpl.h
pixelflinger: fixed

PR #10063, opened 2026-05-27
utils: Implement PagedArenaBitsetPool transient resource pool

pixelflinger:
Introduce utils::PagedArenaBitsetPool, a zero-allocation, RAII-driven transient resource pool for PagedArenaBitset. This pool optimizes steady-state execution in hot rendering and ECS loops by caching bitsets, avoiding randomized OS heap allocator placements.

- Implement ScopedBitset RAII wrapper with C++ reference constructors.
- Add clear() and size() for pool size tracking and garbage sweeps.
- Add comprehensive Doxygen documentation and unit test coverage.

### DIFF
--- a/libs/utils/include/utils/PagedArenaBitsetPool.h
+++ b/libs/utils/include/utils/PagedArenaBitsetPool.h
@@ -1,0 +1,195 @@
+/*
+ * Copyright (C) 2026 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


libs/utils/include/utils/PagedArenaBitsetPool.h
```diff
@@ -0,0 +1,193 @@
...
+         * @brief Constructs an empty, null ScopedBitset wrapper in a moved-from state.
+         */
+        ScopedBitset() noexcept = default;
+
+        /**
+         * @brief Constructs a ScopedBitset wrapping a checked-out PagedArenaBitset.
+         * @param bitset Reference to the checked-out bitset.
+         * @param pool Reference to the parent pool.
+         */
+        ScopedBitset(PagedArenaBitset& bitset, PagedArenaBitsetPool& pool) noexcept
```
poweifeng, line 96: nit: I think this is only meant for the pool?  Can we move this to private and add the Pool class as a friend?

poweifeng (approved):

libs/utils/include/utils/PagedArenaBitsetPool.h
pixelflinger: good catch, it's actually not good it's public, it could cause real harm.

z3moon (approved):

PR #10072, opened 2026-05-29
filamentapp: refactor to enable different windowing implementation

poweifeng:
- Introduce new abstraction DisplayManager to separate windowing
  specifics from FilamentApp
- This allows us to refactor out all of the SDL specific bits
  into its own implementation of the DisplayManager
- Add an initial HTML/webserver based implementation.
  HtmlDisplayManager will render headlessly and stream the output
  to a page.
- Refactor Platform related code its own directory
- Refactor Window related code to better separate concerns between
  FilamentApp and FilamentApp::Window.


pixelflinger:
Do you envision this will allow use to run our samples on both Desktop and Android?

poweifeng:
> Do you envision this will allow use to run our samples on both Desktop and Android?

Yes, I think that's a definite possibility with a bit more refactoring.

pixelflinger (approved):

PR #10077, opened 2026-06-01
backend: workaround broken WebGL spec-constant UBO array sizes

Erkko68:
## Symptom

On ANGLE-on-D3D11 (Chromium/Firefox on Windows, and several Android GLES configs), every `glDrawElementsInstanced` trips the WebGL2 validator:

```
GL_INVALID_OPERATION: glDrawElementsInstanced:
  It is undefined behaviour to use a uniform buffer that is too small.
```

The draw is dropped, so lit/instanced materials render black or invisible. ANGLE-Metal (Chromium on macOS) tolerates the mismatch, which is why it goes unnoticed there.

## Cause

ANGLE-D3D11 doesn't fold a spec-constant-initialized `const int` into a uniform-block array length:

```glsl
const int CONFIG_MAX_INSTANCES = SPIRV_CROSS_CONSTANT_ID_1;
...
PerRenderableData data[CONFIG_MAX_INSTANCES];
```

The block is laid out with the matc-baked default (`CONFIG_MAX_INSTANCES` is 64 in matc, which isn't built with `__EMSCRIPTEN__`), while the engine binds the runtime size (8 on web) — so the bound range is smaller than the declared block.

Same shape as the bug originally fixed by fe3790cb9 (#5859), removed when spec constants were assumed portable. #10000 later fixed the froxel variant but not `CONFIG_MAX_INSTANCES`.

## Fix

Add a `bugs.spec_constant_array_size_not_folded` workaround flag, set for ANGLE-D3D11 renderers. When set, `ShaderCompilerService` rewrites the symbolic array length to the literal the engine actually binds (read from the spec-constant array, so it's correct on every platform/feature level), padding with spaces to preserve byte count. Gating on the driver flag rather than `__EMSCRIPTEN__` keeps the spec constant on compliant drivers and also covers affected native GLES drivers; more renderer strings can be added to the match as offenders are confirmed.

Tested on Chromium/Windows and Android (both previously black, now correct).

## Note

This supersedes my earlier #10053, which I closed assuming #10000 covered it. After further testing, the instanced-rendering case was unrelated to #10000 (which only addressed the froxel buffers) and remained broken — this change actually resolves it.

### DIFF
--- a/filament/backend/src/opengl/OpenGLContext.cpp
+++ b/filament/backend/src/opengl/OpenGLContext.cpp
@@ -359,6 +359,16 @@ void OpenGLContext::initBugs(Bugs* bugs, Extensions const& exts,
     (void)shader;
 
     const bool isAngle = strstr(renderer, "ANGLE");
+
+    // ANGLE's D3D11 path doesn't fold a spec-constant-initialized `const int` into a
+    // uniform-block array length, so instanced draws bind a range smaller than the block the
+    // shader declares ("uniform buffer too small" -> dropped draw -> black materials). Renderer
+    // string looks like "ANGLE (NVIDIA, NVIDIA GeForce ... Direct3D11 vs_5_0 ps_5_0)". This is
+    // observed on Chromium/Firefox-on-Windows, including through WebGL. (b/...)
+    if (isAngle && strstr(renderer, "Direct3D11")) {
+        bugs->spec_constant_array_size_not_folded = true;
+    }
+
     if (!isAngle) {
         if (strstr(renderer, "Adreno")) {
             // Qualcomm GPU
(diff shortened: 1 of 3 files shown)


Erkko68:
Idk why the Windows CI failed, this doesn't seem related to my changes:

D:\a\filament\filament\out\cmake-mt\samples\generated\resources\dummy.c(1,1): error C1083: Cannot open source file: 'D:\a\filament\filament\out\cmake-mt\samples\generated\resources\dummy.c': Permission denied [D:\a\filament\filament\out\cmake-mt\samples\sample-resources.vcxproj]
  (compiling source file 'generated/resources/dummy.c')

show50726 (approved):

poweifeng (approved):

z3moon (approved):

PR #10079, opened 2026-06-02
ViewCache invalidation needs to unset mScene

poweifeng:
This addresses the following bug:
- Assume there's a scene at address 0x0a.
- We call Scene::terminate() and delete the scene.
- Scene::terminate() calls invalidateCache(), which unsets mCurrentViewCache.
- Allocate a new scene, which happens to be at 0x0a.
- Call View::setScene with 0x0a.
  - Because mScene was not unset from before, mScene == 0x0a and no new view cache has been created and set to mCurrentViewCache.


pixelflinger (approved):

pixelflinger:
> This addresses the following bug:
> 
> * Assume there's a scene at address 0x0a.
> * We call Scene::terminate() and delete the scene.
> * Scene::terminate() calls invalidateCache(), which unsets mCurrentViewCache.
> * Allocate a new scene, which happens to be at 0x0a.
> * Call View::setScene with 0x0a.
>   
>   * Because mScene was not unset from before, mScene == 0x0a and no new view cache has been created and set to mCurrentViewCache.

That looks correct because when `FScene::terminate()` calls `invalidateCache()`, not clearing `mScene` basically leaves it as a dangling pointer.

PR #10082, opened 2026-06-03
filamentapp: display manager driven rendering

poweifeng:
Different platforms might have different ways to drive the rendering. Some guided by display vsync, others by perceived latency or other throughput metrics.

Here we refactor rendering to be driven by the display manager and not a blocking loop inside FilamentApp. For SDL2, the actual implementation is still a blocking loop. But other implementation will be different.

### DIFF
--- a/libs/filamentapp/include/filamentapp/FilamentApp.h
+++ b/libs/filamentapp/include/filamentapp/FilamentApp.h
@@ -18,6 +18,8 @@
 #define TNT_FILAMENT_SAMPLE_FILAMENTAPP_H
 
 #include "Config.h"
+#include "Cube.h"
+#include "Grid.h"
 #include "IBL.h"
 
 #include <filament/Engine.h>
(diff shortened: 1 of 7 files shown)


libs/filamentapp/include/filamentapp/FilamentApp.h
```diff
@@ -302,6 +317,21 @@ class FilamentApp {
 
     filament::backend::Platform* mVulkanPlatform = nullptr;
     filament::backend::Platform* mWebGPUPlatform = nullptr;
+
+    std::unique_ptr<Window> mWindow;
+    CleanupCallback mCleanupCallback;
+    ImGuiCallback mImguiCallback = ImGuiCallback();
+    PreRenderCallback mPreRender = PreRenderCallback();
+    PostRenderCallback mPostRender = PostRenderCallback();
```
pixelflinger: Why can't it be `Foo callback{};`syntax?

libs/filamentapp/src/FilamentApp.cpp
```diff
@@ -101,14 +98,328 @@ FilamentApp::~FilamentApp() {
     }
 }
 
-View* FilamentApp::getGuiView() const noexcept {
-    return mImGuiHelper->getView();
+View* FilamentApp::getGuiView() const noexcept { return mImGuiHelper->getView(); }
+
+
+bool FilamentApp::doFrame() {
```
pixelflinger, line 241: is there a way to order method in this file so that this block doesn't appear as a diff?

libs/filamentapp/include/filamentapp/FilamentApp.h
poweifeng: done.

libs/filamentapp/src/FilamentApp.cpp
poweifeng: I tried, maybe now is better?

pixelflinger (approved):

PR #10087, opened 2026-06-04
Refactor: Simplify Scene cache ownership and lifecycle

pixelflinger:
Move FScene::SceneCacheData ownership to FView, simplifying the circular dependency and cleanup callbacks between FScene and FView.

Previously, FScene owned the SceneCacheData map keyed by FView, requiring complex and messy invalidateCache() and unregisterView() callbacks to tear down state. FView also had to use mutable qualifiers on mScene and mCurrentViewCache.

With this change:
- FView owns its SceneCacheData directly via a std::unique_ptr.
- FScene tracks referencing views using a simple std::vector<FView*> since the set of views rendering a scene is expected to be small.
- FView::detachScene() is introduced to allow FScene to safely nullify referencing views and clear their caches during scene destruction.
- Public View::setScene() documentation is updated to specify that scenes are automatically dissociated from the view when destroyed.
- mScene and mSceneCache are now const-correct (non-mutable).

### DIFF
--- a/filament/src/details/View.cpp
+++ b/filament/src/details/View.cpp
@@ -205,19 +205,26 @@ FView::~FView() noexcept = default;
 void FView::setScene(FScene* scene) {
     if (mScene != scene) {
         if (mScene) {
-            invalidateCache(mScene);
+            mScene->unregisterView(this);
         }
         mScene = scene;
         if (scene) {
-            mCurrentViewCache = scene->registerView(this);
+            scene->registerView(this);
+            if (!mSceneCache) {
+                mSceneCache = std::make_unique<FScene::SceneCacheData>();
+            }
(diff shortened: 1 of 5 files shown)


filament/src/details/View.cpp
```diff
@@ -205,19 +205,26 @@ FView::~FView() noexcept = default;
...
+                mSceneCache = std::make_unique<FScene::SceneCacheData>();
+            }
+        } else {
+            mSceneCache.reset();
         }
     }
 }
 
 void FView::terminate(FEngine& engine) {
     if (mScene) {
```
poweifeng, line 223: I think this could be just `setScene(nullptr)`

poweifeng (approved):

PR #10088, opened 2026-06-04
TextureCache: Recycle texture resources when discarded from history

pixelflinger:
When TAA or SSR history textures were popped from the frame history FIFO, they were destroyed directly through the driver via the disposer, deleting them permanently from the GPU. By registering the active TextureCache with the TextureCacheDisposer, we redirect these discard events back to the cache, allowing them to be successfully recycled and reused next frame instead of being recreated from scratch.

### DIFF
--- a/filament/src/TextureCache.cpp
+++ b/filament/src/TextureCache.cpp
@@ -140,6 +140,7 @@ TextureCache::TextureCache(std::shared_ptr<TextureCacheDisposer> disposer,
 }
 
 TextureCache::~TextureCache() noexcept {
+    mDisposer->removeTextureCache(this);
     assert_invariant(mTextureCache.empty());
 }
 
@@ -231,7 +232,7 @@ TextureHandle TextureCache::createTexture(StaticString const name,
             handle = swizzledHandle;
         }
     }
-    mDisposer->checkout(handle, key);
+    mDisposer->checkout(this, handle, key);
     return handle;
 }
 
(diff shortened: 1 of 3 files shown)


filament/src/TextureCache.cpp
```diff
@@ -330,9 +333,13 @@ void TextureCacheDisposer::terminate() noexcept {
 
 void TextureCacheDisposer::destroy(TextureHandle const handle) noexcept {
     if (handle) {
-        auto r = checkin(handle);
-        if (r.has_value()) {
-            mBackend.destroyTexture(handle);
+        if (mTextureCache) {
```
poweifeng: I'm little confused by the relationship between textureCache and TextureCacheDisposer.  It seems like TextureCache is tied to a renderer and TextureCacheDisposer to the engine.  Could this cause an issue if you have multiple renderers?

filament/src/TextureCache.cpp
```diff
@@ -330,9 +333,13 @@ void TextureCacheDisposer::terminate() noexcept {
...
+            mTextureCache->destroyTexture(handle);
+        } else {
```
poweifeng: I think maybe the point of having TextureCacheDisposer is to keep a list of alive textures across renderers? But then we'll almost never be in this block (and do `checkin`), am I reading this correct?

pixelflinger:
all very good points, I'm rewriting this.

filament/src/TextureCache.cpp
pixelflinger: addressed

filament/src/TextureCache.cpp
pixelflinger: addressed

bejado (approved):

PR #10089, opened 2026-06-04
Fix compilation errors due to incompatible c++20 use

poweifeng:
One of our clients is using a toolchain with incomplete implementation of c++20.


pixelflinger (approved):
I think we'll keep running into std::ranges though.

PR #10092, opened 2026-06-06
Implement Phase 1 (amortization) async loading for Vulkan

z3moon:
The Vulkan backend had all async method stubs defined but unimplemented. This change wires them up using the shared JobQueue + AmortizationWorker infrastructure, mirroring the OpenGL backend's behavior.

In this Phase 1, deliver the full async API surface--job ordering, cancellation, completion callbacks, and deferred destruction--without taking on any threading complexity. Vulkan has no shared-context model, so true CPU/GPU parallelism is deferred to Phase 2. Phase 1 decouples "we have a working async API" from "we have parallel uploads": all async jobs are time-sliced on the backend thread during tick(), so the device-capability question never blocks landing async loading.

In Phase 2, will add a ThreadWorker that offloads the amortizable work--staging fill, vkCmdCopy* recording, and layout barriers--into a thread own command pool, while resource allocation and vkQueueSubmit stay on the backend thread (kept synchronous here precisely so this is additive). The backend thread gathers the worker's command buffers into its own submit, or the worker submits to a dedicated transfer queue and signals a semaphore. This will require thread-safe VMA, StagePool, and BufferCache, plus queue-ownership transfer between different queues.

### DIFF
--- a/filament/backend/src/opengl/OpenGLDriver.cpp
+++ b/filament/backend/src/opengl/OpenGLDriver.cpp
@@ -404,12 +404,13 @@ void OpenGLDriver::terminate() {
     if (getJobWorker()) {
         getJobWorker()->terminate();
     }
-    // wait for the GPU again because JobWorker might have queued more work.
-    glFinish();
-
     if constexpr (UTILS_HAS_THREADING) {
+        // Flush any callbacks the drained jobs posted via scheduleCallback().
         stopServiceThread();
     }
+
+    // wait for the GPU again because JobWorker might have queued more work.
+    glFinish();
 
     mContext.destroyState(mBackendState);
     mBackendState = nullptr;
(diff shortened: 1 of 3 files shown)


jorgeag-google (commented):
LGTM.
However, wait for the approval of someone more familiar with the VK driver

bejado (approved):

filament/backend/src/opengl/OpenGLDriver.cpp
```diff
@@ -404,13 +404,14 @@ void OpenGLDriver::terminate() {
...
-    glFinish();
-
     if constexpr (UTILS_HAS_THREADING) {
+        // Flush any callbacks the drained jobs posted via scheduleCallback().
         stopServiceThread();
     }
 
+    // wait for the GPU again because JobWorker might have queued more work.
+    glFinish();
+
```
poweifeng, line 414: why this change?

poweifeng (approved):

filament/backend/src/opengl/OpenGLDriver.cpp
z3moon: I thought this potential blocking call can wait, which I think is slightly better. Same pattern applied to the VulkanDriver::terminate().

PR #10105, opened 2026-06-09
Make getMaterialCompileVariants respect the material shading model

trie94:
`getMaterialCompileVariants` was generating `DIR` variants even for unlit materials, because it only checked the `View`, ignoring the material's unlit shading model.

Also manually tested on Moohan with an unlit builtin material.

FIXES = 521419315


filament/src/details/Engine.cpp
```diff
@@ -1768,19 +1769,25 @@ FixedCapacityVector<Variant> FEngine::getMaterialCompileVariants(
...
+    baseVariant.setDynamicLighting(isMaterialLit && view->hasDynamicLighting());
     baseVariant.setFog(view->hasFog());
-    baseVariant.setShadowSampler2D(view->hasShadowing() && (view->getShadowType() != ShadowType::PCF));
+    baseVariant.setShadowSampler2D(isMaterialLit && view->hasShadowing() && (view->getShadowType() != ShadowType::PCF));
     baseVariant.setStereo(view->hasStereo());
 
     variants.push_back(baseVariant);
     apply(0, skinning, &Variant::setSkinning);
-    apply(0, shadowReceiver, &Variant::setShadowReceiver);
+    // Only respect shadowReceiver variant when the material is lit.
```
romainguy: Make sure none of this breaks the `shadowMultiplier` case (https://google.github.io/filament/Materials.md.html#materialdefinitions/materialblock/lighting:shadowmultiplier), used to cast shadows on unlit models, specifically for transparent shadow casters

trie94: const bool isMaterialLit = material->getDefinition().isVariantLit; checks both the material shading model and whether it has shadow multiplier set.

i added another unit test to verify the behavior.

trie94: `isVariantLit = shading != Shading::UNLIT || hasShadowMultiplier;`
https://github.com/google/filament/blob/main/filament/src/MaterialDefinition.cpp#L353

pixelflinger: yes, so if you look in GLSLPostProcessor.cpp, you'll discover that a material is considered "lit" if it uses the lit model *or* if it has `shadowMultiplier`.  In both cases, the descriptor set layout for *lit* is used -- but the shader code included is the unlit one (with some special lit stuff in it).

So for your work on variants, an unlit material can have the DIR  and/or SRE (but not DYN) variants active (and they will have the lit desc layout).

@romainguy thanks, good call out. 
@trie94 an easy way to test is to activate the ground plane in gltf_viewer

pixelflinger (dismissed):

filament/src/details/Engine.cpp
romainguy: Ah right, yeah that's who we did it initially. Thanks for double-checking, and even more so for adding a unit test. `shadowMultiplier` is a weird enough use case that it'd be easy to break.

trie94: i tested gltf_viewer, verifying that the ground plane shows correctly
shadowmultipliertest

i think the current version makes sense, unless i am missing something

trie94:
are there anything i should address on this pr?

show50726 (approved):

PR #10108, opened 2026-06-10
backend: add SwapChain::setFrameRate

pixelflinger:
This is currently only implemented on Android.

### DIFF
--- a/android/samples/sample-gltf-viewer/src/main/java/com/google/android/filament/gltf/MainActivity.kt
+++ b/android/samples/sample-gltf-viewer/src/main/java/com/google/android/filament/gltf/MainActivity.kt
@@ -29,6 +29,7 @@ import com.google.android.filament.Fence
 import com.google.android.filament.IndirectLight
 import com.google.android.filament.Material
 import com.google.android.filament.Skybox
+import com.google.android.filament.SwapChain
 import com.google.android.filament.View
 import com.google.android.filament.View.OnPickCallback
 import com.google.android.filament.utils.*
@@ -72,6 +73,7 @@ class MainActivity : Activity() {
     private var loadStartFence: Fence? = null
     private val viewerContent = AutomationEngine.ViewerContent()
     private var useStaticModel = false
+    private var currentFrameRate = -1.0f
 
     @SuppressLint("ClickableViewAccessibility")
     override fun onCreate(savedInstanceState: Bundle?) {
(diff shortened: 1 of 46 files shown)


filament/src/details/SwapChain.cpp
```diff
@@ -128,4 +134,18 @@ bool FSwapChain::isProtectedContentSupported(FEngine& engine) noexcept {
     return engine.getDriverApi().isProtectedContentSupported();
 }
 
+utils::tribool FSwapChain::isFrameRateChangeSupported() const noexcept {
+    if (mFrameRateSupportState.is_indeterminate() && mNativeWindow) {
```
bejado, line 138: `UTILS_UNLIKELY`?

libs/utils/include/utils/tribool.h
```diff
@@ -21,43 +21,58 @@
...
 
-    constexpr tribool() noexcept : mValue(Indeterminate) {}
-    constexpr tribool(bool v) noexcept : mValue(v ? True : False) {}
-    constexpr tribool(value_t v) noexcept : mValue(v) {}
+    using value_t = Value;
+
+    static constexpr Value kFalse = Value::kFalse;
+    static constexpr Value kTrue = Value::kTrue;
+    static constexpr Value kIndeterminate = Value::kIndeterminate;
+    static constexpr Value indeterminate = Value::kIndeterminate;
```
bejado: I don't think `indeterminate` (lowercase) is used anywhere.

android/samples/sample-gltf-viewer/src/main/java/com/google/android/filament/gltf/MainActivity.kt
```diff
@@ -393,6 +395,20 @@ class MainActivity : Activity() {
         modelViewer.cameraNear = automation.cameraSettings.near
         modelViewer.cameraFar = automation.cameraSettings.far
         updateRootTransform()
+
+        val targetFps = automation.viewerOptions.cameraFrameRate
+        if (targetFps != currentFrameRate) {
+            modelViewer.swapChain?.let { sc ->
+                if (sc.isFrameRateChangeSupported) {
```
bejado, line 402: If the platform does not support frame rate changes, then this method will always return false, which means we'll keep calling this every frame. Maybe we should make the Java API also return a tribool?

pixelflinger: I think it's okay for now because on android it's been supported for a while. java doesn't have tribool. also the overhead is not very big.

PR #10113, opened 2026-06-12
utils: Improve Slice compatibility with std::span

pixelflinger:
Enhance utils::Slice by adding missing std::span features and APIs, improving interoperability across the engine.

Key improvements:
- Add non-explicit constructors from C-arrays, std::array, and any contiguous container (like std::vector or FixedCapacityVector).
- Implement subview operations: first(), last(), and subspan().
- Add standard type aliases (difference_type, reverse_iterator) and reverse iterator accessors (rbegin(), rend(), etc.).
- Provide as_bytes() and as_writable_bytes() non-member helpers alongside a size_bytes() alias.
- Add C++17 deduction guides for container ranges.

Fully verified with new unit tests in test_Slice.cpp.

### DIFF
--- a/libs/utils/include/utils/Slice.h
+++ b/libs/utils/include/utils/Slice.h
@@ -21,13 +21,31 @@
 #include <utils/Hash.h>
 
 #include <algorithm>
+#include <array>
+#include <iterator>
 #include <type_traits>
 #include <utility>
 
 #include <assert.h>
 #include <stddef.h>
+#include <stdint.h>
 
 namespace utils {
+
+template<typename T>
+class Slice;
+
+constexpr size_t dynamic_extent = static_cast<size_t>(-1);
+
+namespace details {
+
+template<typename T>
+struct is_slice : std::false_type {};
+
+template<typename T>
+struct is_slice<Slice<T>> : std::true_type {};
+
+} // namespace details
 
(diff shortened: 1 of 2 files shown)


libs/utils/include/utils/Slice.h
```diff
@@ -39,25 +57,41 @@ class Slice {
...
     Slice() = default;
-    Slice(iterator begin, iterator end) noexcept : mBegin(begin), mEnd(end) {}
+    Slice(iterator const begin, iterator const end) noexcept : mBegin(begin), mEnd(end) {}
     Slice(pointer begin, size_type count) noexcept : mBegin(begin), mEnd(begin + count) {}
 
-    Slice(Slice<T> const& rhs) : mBegin(rhs.begin()), mEnd(rhs.end()) {}
+    template<size_t N>
+    Slice(element_type (&arr)[N]) noexcept : mBegin(arr), mEnd(arr + N) {}
+
+    template<typename Cont,
```
bejado, line 86: Maybe a comment or `using` declaration here to make this `enable_if_t` bit more readable?

bejado (approved):

libs/utils/include/utils/Slice.h
pixelflinger: done

PR #10114, opened 2026-06-12
filameshio: fix heap OOB write in compressed vertex decode (CWE-787)

foodlook:
## Summary

Fix a heap out-of-bounds write in MeshReader::loadMeshFromBuffer when parsing a crafted .filamesh file with the COMPRESSION flag set (non-interleaved path).

## Root Cause

The uncompressed vertex buffer is allocated based on hasUV1, which is derived from Header.offsetUV1/Header.strideUV1:

~~~cpp
const bool hasUV1 = header.offsetUV1 != uintmax && header.strideUV1 != uintmax;
size_t vertexSize = sizeof(half4) + sizeof(short4) + sizeof(ubyte4) + sizeof(ushort2)
                  + (hasUV1 ? sizeof(ushort2) : 0);   // 24 bytes when hasUV1==false
void* uncompressed = malloc(vertexSize * vertexCount);
~~~

However, the decode dispatch keys the uv1 write off CompressionHeader.uv1, an **independent** field embedded in the vertex payload:

~~~cpp
if (sizes.uv1) {                              // from CompressionHeader, not hasUV1
    dstdata += sizeof(ushort2) * vertexCount;  // advances to byte 24*N (end of alloc)
    err |= decode(dstdata, vertexCount, sizeof(ushort2), srcdata, sizes.uv1);
                                               // writes 4*N bytes past end -> OOB
}
~~~

The only existing guard (\compressedSum > verticesSize\) bounds the **source** reads, not the destination size.

## Fix

Add a consistency check right after reading the \CompressionHeader\: if \sizes.uv1 != 0\ but \hasUV1\ is false, reject the input.

## Regression Test

\CompressedUV1MismatchRejected\ in \	est_filamesh.cpp\: crafts a .filamesh buffer that triggers the inconsistency (COMPRESSION flag, hasUV1=false, CompressionHeader.uv1=1) and verifies \loadMeshFromBuffer\ returns an empty mesh.

## Reproduction

Under AddressSanitizer (Docker \silkeh/clang:latest\, linking the vendored meshoptimizer):

~~~
==ERROR: AddressSanitizer: heap-buffer-overflow
WRITE of size 1024 at ... 0 bytes after 6144-byte region
  #2 meshopt_decodeVertexBuffer vertexcodec.cpp:1229
  #3 main poc_mem001_filament_meshreader_oob.cpp:203
~~~

After the fix, the same input is rejected cleanly with no ASan error.


poweifeng (approved):

poweifeng:
Please fix failed check: https://github.com/google/filament/actions/runs/29948342390/job/89019455000?pr=10114

PR #10117, opened 2026-06-12
android: Proactively detect unsupported timestamps

### LINKED ISSUE #10104: TextureView swapchain: eglGetFrameTimestampsANDROID logs EGL_BAD_ACCESS every frame
**Describe the bug**
When using a TextureView-based swap chain, libEGL prints an error on every frame:
    eglGetFrameTimestampsANDROIDImpl:2408 error 3002 (EGL_BAD_ACCESS)

**To Reproduce**
1. Use the TextureView as sample below:
  filament/android/samples/sample-texture-view/src/main/java/com/google/android/filament/textureview/MainActivity.kt
2. Create a TextureView + UiHelper-based swap chain, call renderer.beginFrame every vsync.

**Expected behavior**
Filament should gracefully detect that this particular EGLSurface doesn't provide frametimestamps and stop calling eglGetFrameTimestampsANDROID. No EGL error should be printed every frame.

 - OS: [Android 15]
 - GPU: [Adreno 663]
 - Backend: [OpenGL]

**Additional context**
This only happens with the TextureView swapchain. When we switch to a SurfaceView swap chain, the issue does not occur and timestamps can be queried successfully.

we think that's because TextureView does not create a separate layer, so it don't have independent SurfaceControl, which would link to SurfaceFlinger. FrameTimeStamps are recorded for SurfaceFlinger layers, not for the intermediate SurfaceTexture surface. Therefore, although eglGetFrameTimestampSupportedANDROID reports that the timestamp type is supported for this EGLSurface, for any concrete frameId, it doesn't have frameEvent history, so eglGetFrameTimestampsANDROID returns EGL_BAD_ACCESS.

Note that this doesn't affect rendering or Filament’s fence-based frame skipping logic; it only results in noisy EGL error logs and unnecessary EGL calls.

It would be helpful if Filament could degrade gracefully to avoid unnecessary EGL calls in this scenario.

For example, in the TextureView case, if eglGetFrameTimestampsANDROID returns EGL_BAD_ACCESS continuously for a given swapchain and no timestamps have ever been successfully retrieved, Filament could set frameTimestampsSupported as false and stop querying timestamps for it. 

This would keep the existing behaviour unchanged for SurfaceView and other cases where frameEvent history is actually available.

If I’ve misunderstood anything about the expected behavior of the timestamp APIs or TextureView surfaces, please let me know. Thank you for your time.

pixelflinger:
When creating a swap chain from a native window on Android, perform an early timestamp test query for a dummy frame ID (0) to definitively detect whether the window provides frame event history (e.g. SurfaceView) or not (e.g. SurfaceTexture / TextureView). If the test returns EGL_BAD_ACCESS, the window lacks history and frame timestamps are immediately disabled, eliminating noisy log spam every frame.

Also adds a runtime LOG(WARNING) if querying timestamps unexpectedly fails during active rendering.

Fixes #10104.

### DIFF
--- a/filament/backend/src/opengl/platforms/PlatformEGLAndroid.cpp
+++ b/filament/backend/src/opengl/platforms/PlatformEGLAndroid.cpp
@@ -47,6 +47,7 @@
 #include <jni.h>
 
 #include <array>
+#include <atomic>
 #include <chrono>
 #include <limits>
 #include <mutex>
@@ -103,7 +104,9 @@ struct PlatformEGLAndroid::SwapChainEGLAndroid : public SwapChainEGL {
     bool setPresentFrameId(uint64_t frameId) const noexcept;
     uint64_t getFrameId(uint64_t frameId) const noexcept;
     bool compositorTimingSupported = false;
-    bool frameTimestampsSupported = false;
+    mutable std::atomic<bool> frameTimestampsSupported{false};
+    mutable std::atomic<bool> frameTimestampsEverRetrieved{false};
+    std::atomic<uint32_t> presentCount{0};
 private:
     AndroidSwapChainHelper mImpl{};
 };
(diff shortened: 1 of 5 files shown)


filament/src/FrameInfo.cpp
```diff
@@ -318,7 +318,8 @@ void FrameInfoManager::updateUserHistory(FSwapChain* swapChain, DriverApi& drive
         auto& entry = history[i];
 
         // retrieve the displayPresentTime only we don't already have it
```
show50726:
```suggestion
        // retrieve the displayPresentTime only when we don't already have it
```

filament/backend/src/opengl/platforms/PlatformEGLAndroid.cpp
```diff
@@ -154,13 +157,23 @@ bool PlatformEGLAndroid::makeCurrent(ContextType const type,
...
-    if (dsc->nativeWindow) {
+    if (dsc && dsc->nativeWindow) {
         auto [err, valid] = NativeWindow::isValid(dsc->nativeWindow);
         FILAMENT_CHECK_POSTCONDITION(!err && valid) << kNativeWindowInvalidMsg << dsc->sur;
     }
     return PlatformEGL::makeCurrent(type, drawSwapChain, readSwapChain);
 }
 
+void PlatformEGLAndroid::commit(SwapChain* swapChain) noexcept {
+    if (swapChain) {
```
show50726: Would it help to add `UTILS_LIKELY` here?

PR #10118, opened 2026-06-12
fix: TAA failed when the view was translucent

pixelflinger:


show50726 (approved):
It would be good to add some descriptions about postprocess variants in `Variant.h`, as I found the code a bit confusing when I first saw it.

PR #10121, opened 2026-06-13
fix(gltfio): reject oversized 8-bit index accessors

aleister1102:
## Summary

This fixes an out-of-bounds read when loading glTF assets with 8-bit index accessors whose declared size exceeds the backing buffer.

- Return `false` from `loadResources()` when `loadCgltfBuffers()` fails (including debug `cgltf_validate` failures)
- Reject index accessors whose binding size is larger than the available buffer bytes before 8-bit index expansion
- Add a regression test with a crafted GLB that previously triggered the crash

## Test plan

- [x] `./out/cmake-debug/libs/gltfio/test_gltfio` (all 6 tests pass)
- [x] New test `RejectsOversizedEightBitIndexAccessor` confirms `loadResources()` returns `false` without crashing

Made with [Cursor](https://cursor.com)

Issue: [issu.ee/523642979](https://issuetracker.google.com/issues/523642979)

### DIFF
--- a/libs/gltfio/src/ResourceLoader.cpp
+++ b/libs/gltfio/src/ResourceLoader.cpp
@@ -284,7 +284,29 @@ inline void createSkins(cgltf_data const* gltf, bool normalize,
     }
 }
 
-inline void uploadBuffers(FFilamentAsset* asset, Engine& engine,
+static bool indexAccessorFitsBuffer(cgltf_accessor const* accessor, uint32_t bindingSize) {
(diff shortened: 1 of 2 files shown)


aleister1102:
Hi @pixelflinger! When you have a moment, would you mind taking a look at this gltfio fix? It touches resource loading and index buffer upload, so your eyes on it would be really appreciated. Thanks so much!

poweifeng:
Fixed and credited in #10145.

PR #10122, opened 2026-06-13
Fix heap buffer overflow in MaterialParser constant-count unflatten

evilgensec:
## Summary

`ChunkMaterialConstants::unflatten` and `ChunkMaterialPushConstants::unflatten` in `src/MaterialParser.cpp` read a 64-bit constant count from a compiled material package and pass it directly to `FixedCapacityVector::reserve()`/`resize()`. `FixedCapacityVector::size_type` is `uint32_t`, so a count of 2^32 or more is silently truncated to a small (or zero) allocation, while the write loop iterates the full 64-bit count. `(*vec)[i].name = ...` for any `i` beyond the truncated capacity writes past the heap buffer. The only per-index guard is `operator[]`'s `assert`, which is compiled out under `NDEBUG` (release builds).

As an example, a count of `0x1_0000_0001` allocates one element; supplying two valid constant entries after it forces an out-of-bounds `CString` write at index 1.

This is reachable from `Material::Builder::package(payload, size)` followed by `build(engine)`: `MaterialDefinition::create` calls `processSpecializationConstants` then `getConstants` (and `processPushConstants` then `getPushConstants`) unconditionally whenever the package contains a `MaterialConstants` or `MaterialPushConstants` chunk. The chunk container validates only the chunk byte length, which is independent of the encoded count.

The other chunk readers in this file size their containers from `uint8_t`/`uint16_t` counts and use `push_back`, so they are not affected.

## Fix

Reject counts that do not fit the container `size_type`, or that exceed the bytes remaining in the chunk, before `reserve`/`resize`. `willOverflow(numConstants)` is a conservative lower bound since each entry occupies more than one byte.

## Testing

Built with `-DNDEBUG -fsanitize=address,undefined`. Before the change, a package with a `MaterialConstants` count of `0x1_0000_0001` followed by two entries triggers `heap-buffer-overflow` in `CString::operator=` through `ChunkMaterialConstants::unflatten`. After the change the oversized count is rejected and the parser returns false. A well-formed package with a valid count parses unchanged.


pixelflinger (approved):

pixelflinger:
@evilgensec all build fail.

evilgensec:
I have made changes to pass the build. Can you please check @pixelflinger?

evilgensec:
Hi Team @pixelflinger @poweifeng ,

Anything needed from my side for the merge? Please let me know.

PR #10123, opened 2026-06-15
fix(gltfio): handle meshopt decode allocation failures

aleister1102:
## Summary

Hi folks! This fixes a release-build crash when `decodeMeshoptCompression()` cannot allocate its decode buffer for attacker-controlled `EXT_meshopt_compression` assets.

- Replace the `assert_invariant(destination)` malloc check with a runtime null check that returns `false`
- Reject decode buffers larger than 256 MiB before allocation (same cap used by `ArchiveCache`)
- Propagate the error in `AssetLoaderExtended`
- Add a regression test for oversized meshopt metadata

## Why this matters

In release builds, `assert_invariant` is compiled out, so a failed `malloc()` could pass a null pointer into meshoptimizer decode routines and crash the host process. `ResourceLoader::loadResources()` already handles a `false` return from `decodeMeshoptCompression()`; this change makes that path actually reachable.

## Test plan

- [x] `./out/cmake-debug/libs/gltfio/test_gltfio --gtest_filter='*Meshopt*'`
- [x] `./out/cmake-debug/libs/utils/test_utils`
- [x] Release PoC with malloc fault injection exits cleanly (`loadResources_returned=0`, no crash)

## Review

@pixelflinger would you mind taking a look when you have a moment? Happy to adjust anything.

Issue: https://issuetracker.google.com/issues/524013826


libs/gltfio/src/Utility.cpp
```diff
@@ -108,9 +108,21 @@ bool decodeMeshoptCompression(cgltf_data* data) {
                 << " (actual=" << compression->count << ") given stride of " << compression->stride
                 << ".";
 
+        static constexpr size_t MAX_MESHOPT_DECODE_SIZE = 256ull * 1024ull * 1024ull;
```
pixelflinger: I don't like these arbitrary limits. I there something in the spec that says there is a limit?

aleister1102: I have dropped that in https://github.com/google/filament/pull/10123/changes/41a6a8523aa33dec19518bed0729807a3147bdbc.

poweifeng (approved):

poweifeng:
Please fix header errors: https://github.com/google/filament/actions/runs/27835064967/job/82381016756

aleister1102:
Hi @poweifeng, the header errors is fixed in https://github.com/google/filament/pull/10141. Thanks!

PR #10128, opened 2026-06-16
Mitigate compiler thread priority inversion

pixelflinger:
Introduce a cross-platform helper, ThreadPriorityOverride, to prevent priority inversion in ShaderCompilerService when the main thread blocks waiting for background shader compilation.

If the main thread enters a condition variable wait, it temporarily elevates the compiling thread's priority to DISPLAY / USER_INTERACTIVE. Once compilation concludes, the priority is restored to its baseline.

This behavior is gated behind the new FeatureFlag
'backend.enable_priority_override_mitigation', which defaults to false. When enabled, compiler thread pool threads default to BACKGROUND priority rather than DISPLAY, conserving system resources.

BUGS=[524748747]


filament/backend/src/opengl/ShaderCompilerService.cpp
```diff
@@ -118,9 +121,11 @@ struct ShaderCompilerService::OpenGLProgramToken : ProgramToken {
     // This may block until the token is ready to be used.
     void wait() const noexcept {
         UniqueLock l(lock);
+        priorityOverride.startOverride();
```
show50726: Is it possible that a job completes before `wait()` is called, causing us to raise the thread's priority but never restore it?

pixelflinger: the priority is always restored at the top of the main loop as a safety measure.

show50726 (approved):

filament/backend/src/opengl/ShaderCompilerService.cpp
```diff
@@ -241,7 +255,8 @@ void ShaderCompilerService::init() noexcept {
         // By default, we use one thread at the same priority as the gl thread. This is the
```
trie94, line 255: should we update this comment that we allow priority override?


also a qq: if it was using the same priority as the gl thread, why was it still causing the priority inversion problem? is it because it is still lower than the main render thread?

trie94 (approved):

filament/backend/src/opengl/ShaderCompilerService.cpp
pixelflinger: It was only causing a priority inversion problem on ANGLE / PowerVR, on other platforms, we mistakenly set the priority to DISPLAY.
I know there was some _internal_ reports of priority inversion on other platforms, but these reports were wrong, they were observing something else.

filament/backend/src/ThreadPriorityOverride.h
```diff
@@ -0,0 +1,204 @@
...
+            SetThreadPriority(thread, THREAD_PRIORITY_HIGHEST);
+        }
+#endif
+    }
+
+    /**
+     * Called by the waiting thread to release the temporary priority elevation.
+     * On Apple, this tears down the pthread QoS override.
+     */
+    void endOverride() noexcept {
```
poweifeng, line 129: nit: is it possible to just combine `endOverride` and `restoreDefaultPriority`?  Semantically, they seem to want to do the same thing (return priority to the before state, or tear down).

Maybe let `startOverride` take the default priority as a parameter?

PR #10131, opened 2026-06-17
Add Perfetto traces in shader compilation code path

trie94:
Add Perfetto traces in shader compilation code path

Also added support for Perfetto event tracing

FIXES=523327225

### DIFF
--- a/filament/backend/src/opengl/ShaderCompilerService.cpp
+++ b/filament/backend/src/opengl/ShaderCompilerService.cpp
@@ -41,6 +41,7 @@
 #include <utils/Mutex.h>
 #include <utils/ostream.h>
 #include <utils/Panic.h>
+#include <utils/sstream.h>
 
 #include <algorithm>
 #include <array>
@@ -100,6 +101,7 @@ struct ShaderCompilerService::OpenGLProgramToken : ProgramToken {
 
     ShaderCompilerService& compiler;
     CString const& name;
+    CString programString;
     FixedCapacityVector<std::pair<CString, uint8_t>> attributes;
     Program::ShaderSource shaderSourceCode;
     void* user = nullptr;
@@ -255,7 +257,7 @@ void ShaderCompilerService::init() noexcept {
         // By default, we use one thread at the same priority as the gl thread. This is the
         // safest choice that avoids priority inversions.
         uint32_t poolSize = 1;
-        JobSystem::Priority priority = mPriorityOverrideEnabled ? 
+        JobSystem::Priority priority = mPriorityOverrideEnabled ?
                 JobSystem::Priority::BACKGROUND : JobSystem::Priority::DISPLAY;
 
         auto const& renderer = mDriver.getContext().renderer;
@@ -313,6 +315,10 @@ ShaderCompilerService::program_token_t ShaderCompilerService::createProgram(
 
     // Create a token. A callback condition (handle) is internally created upon token creation.
     auto token = std::make_shared<OpenGLProgramToken>(*this, name, mPriorityOverrideEnabled);
+    utils::io::sstream ss;
+    ss << program;
+    token->programString = utils::CString(ss.c_str());
+
     if (UTILS_UNLIKELY(gl.isES2())) {
         token->attributes = std::move(program.getAttributes());
     }
@@ -348,7 +354,12 @@ void ShaderCompilerService::compileProgram(
     switch (mMode) {
         case Mode::THREAD_POOL: {
             mCompilerThreadPool.queue(priorityQueue, token,
-                    [this, &gl, program = std::move(program), token]() mutable {
+                    [this, &gl, program = std::move(program), token, priorityQueue]() mutable {
+                        UTILS_UNUSED const CompilerPriorityQueue priority = priorityQueue;
+                        FILAMENT_TRACING_EVENT(FILAMENT_TRACING_CATEGORY_FILAMENT,
+                                "compileProgram", "name", token->name.c_str_safe(),
+                                "program", token->programString.c_str_safe(),
+                                "priority", static_cast<uint32_t>(priority));
                         // The compilation job has just begun. At this point, the token's state must
                         // be either WAITING or CANCELED.
                         assert_invariant(token->state == OpenGLProgramToken::State::WAITING ||
@@ -571,7 +582,15 @@ void ShaderCompilerService::ensureTokenIsReady(program_token_t const& token) {
         case Mode::THREAD_POOL: {
             // We need this program right now, make sure the job is finished.
             if (auto job = mCompilerThreadPool.dequeue(token)) {
+                FILAMENT_TRACING_EVENT(FILAMENT_TRACING_CATEGORY_FILAMENT,
+                        "compile(Job not ready)", "name", token->name.c_str_safe(),
+                        "program", token->programString.c_str_safe());
                 job();// The job hasn't started yet, so execute it now.
+            } else if (!token->isReady()) {
+                // The job is running but hasn't finished.
+                FILAMENT_TRACING_EVENT(FILAMENT_TRACING_CATEGORY_FILAMENT,
+                        "compile(Wait)", "name", token->name.c_str_safe(),
+                        "program", token->programString.c_str_safe());
             }
 
             // This may block if the job was already taken by a thread ahead of the `dequeue`
--- a/filament/backend/src/opengl/ShaderCompilerService.cpp
+++ b/filament/backend/src/opengl/ShaderCompilerService.cpp
@@ -350,16 +350,15 @@ ShaderCompilerService::program_token_t ShaderCompilerService::createProgram(
 void ShaderCompilerService::compileProgram(
         program_token_t const& token, Program&& program) {
     auto& gl = mDriver.getContext();
-    CompilerPriorityQueue const priorityQueue = program.getPriorityQueue();
+    CompilerPriorityQueue const priority = program.getPriorityQueue();
     switch (mMode) {
         case Mode::THREAD_POOL: {
-            mCompilerThreadPool.queue(priorityQueue, token,
-                    [this, &gl, program = std::move(program), token, priorityQueue]() mutable {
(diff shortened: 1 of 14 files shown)


filament/backend/src/opengl/ShaderCompilerService.cpp
```diff
@@ -297,6 +299,10 @@ ShaderCompilerService::program_token_t ShaderCompilerService::createProgram(
 
     // Create a token. A callback condition (handle) is internally created upon token creation.
     auto token = std::make_shared<OpenGLProgramToken>(*this, name);
+    utils::io::sstream ss;
+    ss << program;
+    token->programString = utils::CString(ss.c_str());
```
show50726, line 320: I wonder why you do this instead of 
```
token->programString = utils::CString(program.getName());
```

filament/src/MaterialDefinition.cpp
```diff
@@ -670,6 +670,7 @@ Handle<HwProgram> MaterialDefinition::compileProgram(
         ProgramSpecialization const& specialization,
         backend::CompilerPriorityQueue const priorityQueue) const noexcept {
     assert_invariant(engine.hasFeatureLevel(featureLevel));
+
```
show50726: remove?

filament/src/details/MaterialInstance.cpp
```diff
@@ -473,10 +475,31 @@ void FMaterialInstance::compile(CompilerPriorityQueue const priority,
...
+                            "name", getMaterial()->getName().c_str_safe(),
+                            "variantKey", static_cast<uint32_t>(variant.key),
+                            "priority", static_cast<uint32_t>(priority));
                     prepareProgram(driver, variant, priority);
+                } else {
+                    FILAMENT_TRACING_EVENT(FILAMENT_TRACING_CATEGORY_FILAMENT,
+                            "requested variant missing",
+                            "name", getMaterial()->getName().c_str_safe(),
+                            "variantKey", static_cast<uint32_t>(variant.key),
+                            "priority", static_cast<uint32_t>(priority));
```
show50726, line 492: I'm not sure if tracing here adds much value, as it just means the material doesn't have the variant.

What I would imagine for how this tracing work is:

1. `materialInstance->prepareProgram` so we only have to add it once.
2. The point where a material instance compiles a `Critical` program. This would help us catch instances where we missed precompilation or the precompilation didn't finish in time.

@pixelflinger I'd like to know about your thoughts.

filament/backend/src/opengl/ShaderCompilerService.cpp
```diff
@@ -553,7 +564,15 @@ void ShaderCompilerService::ensureTokenIsReady(program_token_t const& token) {
         case Mode::THREAD_POOL: {
             // We need this program right now, make sure the job is finished.
             if (auto job = mCompilerThreadPool.dequeue(token)) {
+                FILAMENT_TRACING_EVENT(FILAMENT_TRACING_CATEGORY_FILAMENT,
+                        "compile(Job not ready)", "name", token->name.c_str_safe(),
```
pixelflinger: "job not ready" is a bit confusing because the if below is "!token->isReady()", so what's the difference?
Could be "compile(processing)" vs "compile(waiting)" maybe?

filament/backend/src/opengl/ShaderCompilerService.cpp
pixelflinger: ah I see, I got fooled by this too. It's because doing it like this, will log the variants too.

filament/src/details/Material.cpp
```diff
@@ -254,6 +260,10 @@ void FMaterial::compile(CompilerPriorityQueue const priority,
         FixedCapacityVector<Variant> const& variants,
         CallbackHandler* handler,
         Invocable<void(Material*)>&& callback) noexcept {
+    FILAMENT_TRACING_CALL(FILAMENT_TRACING_CATEGORY_FILAMENT,
+            "name", getName().c_str(), "variant count", variants.size(),
+            "parallel shader compilation supported",
```
pixelflinger: maybe use a shorted string, like just "supported"?

filament/src/details/Material.cpp
```diff
@@ -264,9 +274,28 @@ void FMaterial::compile(CompilerPriorityQueue const priority,
...
         }
+    } else {
+        for (UTILS_UNUSED_IN_RELEASE auto const variant : variants) {
+            FILAMENT_TRACING_EVENT(FILAMENT_TRACING_CATEGORY_FILAMENT,
+                    "parallel compilation disabled",
+                    "name", getName().c_str(),
+                    "variantKey", static_cast<uint32_t>(variant.key),
+                    "priority", static_cast<uint32_t>(priority));
+        }
     }
```
pixelflinger, line 305: these 3 logs seem high verbosity -- every variant for every materials. I would prefer to have these behind a compile time flag, or in debug build only.

filament/src/details/MaterialInstance.cpp
```diff
@@ -473,10 +475,31 @@ void FMaterialInstance::compile(CompilerPriorityQueue const priority,
...
+    } else {
+        for (auto const variant: definition.getVariants()) {
+            if (!variantFilter || variant == Variant::filterUserVariant(variant, variantFilter)) {
+                FILAMENT_TRACING_EVENT(FILAMENT_TRACING_CATEGORY_FILAMENT,
+                        "parallel compilation disabled",
+                        "name", getMaterial()->getName().c_str_safe(),
+                        "variantKey", static_cast<uint32_t>(variant.key),
+                        "priority", static_cast<uint32_t>(priority));
+            }
+        }
```
pixelflinger, line 507: same comment as above, this seems very high verbosity.

filament/backend/src/opengl/ShaderCompilerService.cpp
trie94: yeah i want the full program description. this gives me something like `Program{Filament Default Material, variant=(0x10), vertexVariant=(0x10), fragmentVariant=(0x10)}` (though sometimes the name is empty)

filament/src/details/MaterialInstance.cpp
trie94: i added this because it actually caught a bug where app-side precompilation request was asking for wrong variants. (https://github.com/google/filament/pull/10105)

filament/backend/src/opengl/ShaderCompilerService.cpp
trie94: the difference is 1) job hasn't started so kicking off now vs 2) job is running but hans't finished so waiting. maybe we don't need to distinguish the difference anyway both are waiting on the job. i merged them back.

filament/src/details/Material.cpp
trie94: i changed it to "parallel comp supported". "supported" isn't super clear maybe?

filament/src/details/Material.cpp
trie94: yeah sounds good, wrapped them with `#ifndef NDEBUG`

filament/src/details/MaterialInstance.cpp
trie94: wrapped them with `#ifndef NDEBUG`

filament/backend/src/opengl/ShaderCompilerService.cpp
```diff
@@ -348,7 +354,12 @@ void ShaderCompilerService::compileProgram(
     switch (mMode) {
         case Mode::THREAD_POOL: {
             mCompilerThreadPool.queue(priorityQueue, token,
-                    [this, &gl, program = std::move(program), token]() mutable {
+                    [this, &gl, program = std::move(program), token, priorityQueue]() mutable {
+                        UTILS_UNUSED const CompilerPriorityQueue priority = priorityQueue;
```
poweifeng: why do you need another variable here?

trie94: otherwise it fails to compile for release builds as it's unused other than tracing

poweifeng: elsewhere you put `UTILS_UNUSED_IN_RELEASE`. Should it be that here? 

But also, isn't it possible to have trace enabled for release builds?

trie94: ah actually you are right that we should be able to trace it on release builds too.

actually it was failing on non-android builds. i looked into this a bit more, and it seems that it is considered as unused for those platforms because tracing is disabled by default, thus it falls back to empty macros.

i worked around this one by getting the priority from the program instead. and i also removed `UTILS_UNUSED_IN_RELEASE` in the other file (vulkan), but i had to do a hack that i am not a super fan of: `(void) priority`. any other options that i could use for this one?

filament/backend/src/vulkan/VulkanPipelineCache.cpp
```diff
@@ -182,23 +188,29 @@ void VulkanPipelineCache::asyncPrewarmCache(
...
     // destroyed before we call createPipeline(). We can catch this in some cases, to avoid
     // compiling too many unnecessary already-destroyed-materials.
-    mCompilerThreadPool.queue(priority, token, [this, vprogram, key, dynamicOptions, cmh]() mutable {
-        if (vprogram->isParallelCompilationCanceled()) {
-            FVK_LOGD << "Skipping prewarm for a program that has been destroyed already.";
-            return;
-        }
+    mCompilerThreadPool.queue(
+        priority, token, [this, vprogram, key, dynamicOptions, cmh, priority]() mutable {
+            (void)priority; // Suppresses "unused variable" warnings when tracing is disabled
```
poweifeng: Sorry to go back and forth.  I guess here you'd need `UTILS_UNUSED` to indicate that it sometimes might be unused.  But to fix it properly, we'd probably need another macro like `UTILS_USED_WITH_TRACE`

trie94: gotcha, i added `UTILS_UNUSED_WITHOU_TRACING`. PTAL

pixelflinger (approved):

PR #10132, opened 2026-06-18
Fix race condition in FilamentTest.FrameHistoryStreamTest

poweifeng:
Do so by introducing a few APIs that might be useful for testing and debugging.


filament/test/filament_test.cpp
```diff
@@ -1706,6 +1707,7 @@ TEST(FilamentTest, FrameHistoryStreamTest) {
             renderer->endFrame();
         }
         engine->flushAndWait();
+        downcast(renderer)->waitForFrameHistory();
```
pixelflinger, line 1710: maybe we need a comment to explain why we would NOT want to do this in flushAndWait (because when shutting down we don't care about the result of the framehistory).

pixelflinger (approved):

filament/test/filament_test.cpp
poweifeng: done.  Added an extended comment for `FRenderer::waitForFrameHistory()`.

PR #10144, opened 2026-06-22
filamentapp: implement web/html based display manager

poweifeng:
HtmlDisplayManager starts a local server that streams out rendered images to a web client.

- Fleshed out client-side implementation (html, js, css)
- Add WebSocket handshake with magic number verification and event ID mapping
- Refactor HtmlDisplayManager to use utils::CString instead of std::string
- Implement 3-second client-side reconnection polling on the frontend
- Pump Cocoa events on macOS main thread to force graphics backing store allocation

image


libs/filamentapp/src/display_managers/HtmlDisplayManager.cpp
```diff
@@ -287,59 +264,132 @@ bool HtmlDisplayManager::WebSocketHandler::handleConnection(CivetServer* server,
...
             event.type = AppEvent::Type::QUIT;
             break;
-        case 1:
-        case 2:
+        case EVENT_KEYDOWN:
+        case EVENT_KEYUP:
             if (data_len < 5) return true;
-            event.type = (type == 1) ? AppEvent::Type::KEYDOWN : AppEvent::Type::KEYUP;
+            event.type = (type == EVENT_KEYDOWN) ? AppEvent::Type::KEYDOWN : AppEvent::Type::KEYUP;
             event.key.code = (AppKey) * (uint32_t*) &data[1];
```
bejado: I think it might be UB to do this, because `data[1]` won't be 4 byte aligned. You might need a `std::memcpy` here instead.

bejado (approved):
LGTM. I'm curious what use case you have in mind for this?

poweifeng:
> LGTM. I'm curious what use case you have in mind for this?

So the problem I'm trying to solve is to be able to interact with the software renderer (mesa) using different backends.  And to do that I needed to separate the windowing part from the backend+graphics driver part.  For SDL2, its windowing is tied to the graphics API + graphics driver - I'm guessing because different graphics API means different ways for building the swapchain.  So I couldn't just swap out the system GL driver with the mesa one, since that affects SDL2 windowing code.  

Incidentally, as proposed by Mathias in another PR, this work of separating windowing from rendering might allow us to easily port `/samples` to Android (not sure about iOS).

PR #10141, opened 2026-06-21
gltfio: fix header organization in test and extended loader

aleister1102:
## Summary
- Run reorganize-headers on files flagged by test-code-correctness CI in #10123

## Test plan
- [x] `bash test/code-correctness/check-headers.sh libs/gltfio/test/gltfio_test.cpp libs/gltfio/src/extended/AssetLoaderExtended.cpp`

Made with [Cursor](https://cursor.com)

PR #10135, opened 2026-06-19
gltfio: Always validate glTF buffers and check load status

deezsecc:
### Summary
This PR fixes a heap out-of-bounds read vulnerability in the glTF loader by ensuring structural validation always runs and is correctly handled.

### Details
1. **Enable Validation in Release Builds:** In `libs/gltfio/src/Utility.cpp`, the validator wrapper `cgltf_validate` was compiled out under release configurations (`#ifndef NDEBUG`). This allowed malformed glTF models specifying invalid accessor elements/bounds to bypass verification, triggering out-of-bounds heap memory access during unpacking (e.g., in `TangentsJob::run` via `cgltf_accessor_unpack_floats`). I removed the debug guard to ensure validation is always performed.
2. **Handle Validation Failure:** In `libs/gltfio/src/ResourceLoader.cpp`, I updated the call to `loadCgltfBuffers` to check its return status and return `false` (aborted load) if validation or buffer loading fails.

Fixes #10134
Security Report Reference: https://issuetracker.google.com/issues/525730913

PR #10133, opened 2026-06-18
Prevent OOB reads in gltfio animation import

Serotav:
Summary:
  - validate gltfio animation sampler accessors before importing them
  - reject non-scalar / non-float input timelines
  - reject sampler accessors whose declared counts do not fit their backing buffer view
  - guard applyAnimation against out-of-range animation indexes

  Testing:
  - ninja -C out/cmake-debug libs/gltfio/CMakeFiles/gltfio_core.dir/src/Animator.cpp.o libgltfio_core.a libgltfio.a test_gltfio
  - ./out/cmake-debug/libs/gltfio/test_gltfio

PR #10119, opened 2026-06-13
gltfio: bound createSkins inverse bind matrix copies to bufferView

aleister1102:
## Summary

- validate inverse bind matrix copies against the declared accessor count and `bufferView` span in `ResourceLoader::createSkins()`
- keep the whole-buffer bounds check for uncompressed buffers as a second line of defense
- add a regression test for malformed external-buffer glTF skins that previously disclosed bytes outside the declared view

## Testing

- `cmake --build out/cmake-debug --target test_gltfio`
- `./out/cmake-debug/libs/gltfio/test_gltfio`
- `./out/cmake-debug/libs/gltfio/test_gltfio --gtest_filter=glTFIOTest.SkipsInverseBindMatricesOutsideBufferView`

Issue: [issu.ee/523503237](https://issuetracker.google.com/issues/523503237)

PR #10145, opened 2026-06-22
gltfio: import security and stability fixes

poweifeng:
Combine several security and stability fixes to gltfio:

- Reorganize header order in test and extended loader (#10141)
- Validate animation channel path against sampler output type (#10139)
- Always validate glTF buffers and check loading status (#10135)
- Harden animation validation and accessor fits checks (#10133)
- Reject index accessors that exceed buffer capacity (#10121)
- Bound inverse bind matrix copies to bufferView bounds (#10119)

### DIFF
--- a/libs/gltfio/include/gltfio/AssetLoader.h
+++ b/libs/gltfio/include/gltfio/AssetLoader.h
@@ -17,12 +17,12 @@
 #ifndef GLTFIO_ASSETLOADER_H
 #define GLTFIO_ASSETLOADER_H
 
-#include <filament/Engine.h>
-#include <filament/Material.h>
-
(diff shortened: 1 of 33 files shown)


z3moon:
Just curious why we cannot accept those PRs individually. Are there additional changes other than them?

poweifeng:
> Just curious why we cannot accept those PRs individually. Are there additional changes other than them?

The reason is not interesting.  It's just too much burden to review so many patches. So instead I took the security issue each is trying to address, and asked a trusted agent to do it. This way I'm more confident that the code is not malicious.  (not saying that any of the original PRs is). This way I can also ensure that the code looks idiomatically filament.

bejado (approved):

PR #10147, opened 2026-06-24
gltfio: out-of-bounds write / crash in createRenderable

### LINKED ISSUE #10136: gltfio: out-of-bounds write / crash in createRenderable when a mesh has morph-target names but no primitives
## Description

Loading certain malformed (but structurally parseable) glTF assets crashes `gltfio` during
`AssetLoader::createAsset()`. The problem is in `FAssetLoader::createRenderable()`
(`libs/gltfio/src/AssetLoader.cpp`): the morph-target **names** are copied into a vector sized by
the mesh's morph-target **count**, but the copy loop iterates over `mesh->target_names_count`, which
is a separate value parsed from `mesh.extras.targetNames`.

When a mesh has **no primitives**, its morph-target count (`numMorphTargets`) is `0`, so
`morphTargetNames` is allocated with capacity `0` — but the loop still runs `target_names_count`
times, writing past the end of the vector:

```cpp
// libs/gltfio/src/AssetLoader.cpp  (createRenderable)
const cgltf_size numMorphTargets = inputPrim ? inputPrim->targets_count : 0;  // 0 when no primitives
...
FixedCapacityVector<CString> morphTargetNames(numMorphTargets);              // capacity 0
for (cgltf_size i = 0, c = mesh->target_names_count; i < c; ++i) {           // c is independent
    morphTargetNames[i] = CString(mesh->target_names[i]);                   // out of bounds
}
```

`FixedCapacityVector::operator[]` is unchecked in release builds, so this is an out-of-bounds write
(and the move-assignment also touches an uninitialized destination). I noticed it while running the
loader under AddressSanitizer.

## Steps to reproduce

Load this ~0.5 KB glTF (no buffers/images needed) via `AssetLoader::createAsset()`:

```json
{
  "asset": { "version": "2.0" },
  "scene": 0,
  "scenes": [ { "nodes": [0] } ],
  "nodes": [ { "mesh": 0 } ],
  "meshes": [ { "extras": { "targetNames": ["t0","t1","t2","t3","t4","t5","t6","t7"] } } ]
}
```

The mesh has no `primitives`, but its `extras.targetNames` lists 8 names. A headless NOOP-backend
harness reaches the fault directly in `createAsset`:

```
AddressSanitizer: heap-buffer-overflow
  #0 filament::gltfio::FAssetLoader::createRenderable(...)
  #1 filament::gltfio::FAssetLoader::recurseEntities(...)
  #2 filament::gltfio::FAssetLoader::createInstance(...)
  #3 filament::gltfio::FAssetLoader::createInstancedAsset(...)
  #4 filament::gltfio::AssetLoader::createAsset(...)
0x... is located 0 bytes after a 1-byte region allocated by createRenderable(...)
```

## Affected versions

Reproduced on `main` at `8b7f2c74`; the same code is present at the current tip `38cf0d2a`
(2026-06-18). The mesh passes the optional `cgltf_validate()` check, because cgltf only compares
`target_names_count` to the morph-target count when the mesh has primitives.

## Proposed fix

Bound the copy by the destination capacity (the morph-target count), e.g.:

```cpp
for (cgltf_size i = 0, c = std::min(numMorphTargets, mesh->target_names_count); i < c; ++i) {
    morphTargetNames[i] = CString(mesh->target_names[i]);
}
```

I'm happy to send a small PR with this fix plus a gltfio unit test that loads the asset above and
verifies it parses without an out-of-bounds access.

## Related (separate, lower priority)

While here I also noticed that `createRenderable()` iterates the morph-target **slot** loop using the
uncapped `numMorphTargets`, whereas `createPrimitives()` caps the per-primitive slot count at
`MAX_MORPH_TARGETS` (256). A mesh with more than 256 morph targets can therefore index
`outputPrim->slotIndices` out of range. This is a distinct issue from the one above; I can file it
separately or fold a bound into the same PR if you prefer.

f0r:
Fixes #10136.

`FAssetLoader::createRenderable()` has two morph-target count mismatches that index/write past their
storage when loading a structurally-parseable but malformed glTF mesh (reachable directly from
`AssetLoader::createAsset()`). Both are in the same function and the same "size by one count, iterate
by another" shape, so they're fixed together here.

## 1. Morph-target **names** vs. morph-target count

`morphTargetNames` is allocated with capacity equal to the mesh's morph-target count
(`numMorphTargets`) but filled by iterating over `mesh->target_names_count`, parsed independently from
`mesh.extras.targetNames`. When a mesh has no primitives, `numMorphTargets` is `0`, so a mesh that
still declares `targetNames` overruns the (zero-capacity) vector:

```cpp
  FixedCapacityVector<CString> morphTargetNames(numMorphTargets);            // capacity 0
- for (cgltf_size i = 0, c = mesh->target_names_count; i < c; ++i) {         // independent count
+ for (cgltf_size i = 0, c = std::min(numMorphTargets, mesh->target_names_count); i < c; ++i) {
      morphTargetNames[i] = CString(mesh->target_names[i]);
  }
```

## 2. Morph-target **slot** loop vs. the `MAX_MORPH_TARGETS` cap

`createPrimitives()` caps a primitive's morph-target count at `MAX_MORPH_TARGETS` and sizes its
`slotIndices` vector to that cap, but `createRenderable()` iterated the slot loop over the uncapped
`numMorphTargets`, so a mesh declaring more than `MAX_MORPH_TARGETS` morph targets indexed
`outputPrim->slotIndices` out of range (and then used the result to index the buffer-slot vector):

```cpp
+ const size_t numSlots = outputPrim->slotIndices.size();
- for (int tindex = 0; tindex < numMorphTargets; ++tindex) {
+ for (int tindex = 0; tindex < numMorphTargets && (size_t) tindex < numSlots; ++tindex) {
```

For well-formed assets both loops are unchanged (`numMorphTargets == target_names_count` and
`numMorphTargets == numSlots`); the only effect is that a malformed mesh's excess names/targets are
ignored instead of overrunning their storage, matching the truncation `createPrimitives()` already
performs.

## Tests

Adds two tests to `libs/gltfio/test/gltfio_test.cpp`:

- `MalformedMeshTargetNamesWithoutPrimitives` — loads a mesh declaring `extras.targetNames` with no
  primitives; pre-fix this trips a heap-buffer-overflow write in `createRenderable()` under ASan.
- `MorphTargetsExceedingMaxDoNotOverflow` — loads a mesh with 300 morph targets (> the 256 cap);
  pre-fix this trips an out-of-bounds `slotIndices` access under ASan.

Both parse cleanly post-fix.

## Notes

- Confined to `createRenderable()` plus the two tests; no public API change. Adheres to
  `CODE_STYLE.md` (4-space indent, 100 columns).
- The mesh in (1) passes the optional `cgltf_validate()` check, which only compares
  `target_names_count` to the morph-target count when the mesh has primitives.

### DIFF
--- a/libs/gltfio/src/AssetLoader.cpp
+++ b/libs/gltfio/src/AssetLoader.cpp
@@ -14,17 +14,19 @@
  * limitations under the License.
  */
 
-#include <gltfio/Animator.h>
-#include <gltfio/AssetLoader.h>
-#include <gltfio/MaterialProvider.h>
-#include <gltfio/math.h>
-
+#include "downcast.h"
 #include "FFilamentAsset.h"
 #include "FNodeManager.h"
 #include "FTrsTransformManager.h"
 #include "GltfEnums.h"
 #include "Utility.h"
+
 #include "extended/AssetLoaderExtended.h"
+
+#include <gltfio/Animator.h>
+#include <gltfio/AssetLoader.h>
+#include <gltfio/MaterialProvider.h>
+#include <gltfio/math.h>
 
 #include <filament/Box.h>
 #include <filament/BufferObject.h>
@@ -40,24 +42,21 @@
 #include <filament/TransformManager.h>
 #include <filament/VertexBuffer.h>
 
-#include <math/mat4.h>
-#include <math/vec3.h>
-#include <math/vec4.h>
-
 #include <private/utils/Tracing.h>
 
 #include <utils/compiler.h>
 #include <utils/EntityManager.h>
 #include <utils/FixedCapacityVector.h>
 #include <utils/Log.h>
-#include <utils/Panic.h>
 #include <utils/NameComponentManager.h>
+#include <utils/Panic.h>
 
-#include <tsl/robin_map.h>
+#include <math/mat4.h>
+#include <math/vec3.h>
+#include <math/vec4.h>
 
 #include <cgltf.h>
-
-#include "downcast.h"
+#include <tsl/robin_map.h>
 
 #include <codecvt>
 #include <locale>
@@ -900,7 +899,11 @@ void FAssetLoader::createRenderable(const cgltf_node* node, Entity entity, const
     }
 
     FixedCapacityVector<CString> morphTargetNames(numMorphTargets);
-    for (cgltf_size i = 0, c = mesh->target_names_count; i < c; ++i) {
+    // A mesh's morph-target name count (mesh.extras.targetNames) is parsed independently of its
+    // morph-target count, and is not constrained when the mesh has no primitives (numMorphTargets
+    // is then 0). Bound the copy by the destination capacity so that a mesh declaring more target
+    // names than morph targets does not write past morphTargetNames.
+    for (cgltf_size i = 0, c = std::min(numMorphTargets, mesh->target_names_count); i < c; ++i) {
         morphTargetNames[i] = CString(mesh->target_names[i]);
     }
     auto& nm = mNodeManager;
--- a/libs/gltfio/src/AssetLoader.cpp
+++ b/libs/gltfio/src/AssetLoader.cpp
@@ -822,7 +822,11 @@ void FAssetLoader::createRenderable(const cgltf_node* node, Entity entity, const
             }
 
             UTILS_UNUSED_IN_RELEASE cgltf_accessor const* previous = nullptr;
-            for (int tindex = 0; tindex < numMorphTargets; ++tindex) {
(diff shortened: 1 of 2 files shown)


poweifeng (approved):

poweifeng:
please fix failure: https://github.com/google/filament/actions/runs/28220046660/job/83599077225

f0r:
Sorry for the force-push — I rebased the two fixes onto the latest main (rather than keeping the merge) and ran tools/reorganize-headers/run.py on AssetLoader.cpp to clear the header-organization CI check that was failing. Should be green now; PTAL.

PR #10151, opened 2026-06-25
Remove Texture::setImage(...FaceOffsets) and FaceOffsets

trie94:
All the usages of this setImage overload & FaceOffsets in G3 (including Moohan) are removed.

Note that there's another FaceOffsets in Filament (https://github.com/google/filament/blob/main/libs/generatePrefilterMipmap/include/filament-generatePrefilterMipmap/generatePrefilterMipmap.h#L33), I didn't remove that one.

FIXES=521419320


poweifeng (approved):
this lgtm, but please wait for Mathias' review because this is a public API change.

pixelflinger (approved):

PR #10153, opened 2026-06-27
vulkan: Support external textures with mipmaps

z3moon:
Support importing external textures (Android Hardware Buffers) with a full mipmap chain under the Vulkan backend, leveraging Android's AHARDWAREBUFFER_USAGE_GPU_MIPMAP_COMPLETE flag.

- Add a `mipLevels` field to VulkanPlatform::ExternalImageMetadata.
- Extract and calculate the mipmap chain length from the hardware buffer descriptor in VulkanPlatformAndroid if the GPU_MIPMAP_COMPLETE flag is set.
- Pass the dynamic mipmap level count to VkImageCreateInfo when allocating the VkImage, keeping YUV staging images at 1 mip level.
- Update the specialized VulkanTexture constructor to accept the levels parameter and propagate it to HwTexture and VulkanTextureState.
- Update driver stream updates and swapchain color/depth allocations to pass the correct level count to VulkanTexture.

BUGS=[498943553]

### DIFF
--- a/filament/backend/src/vulkan/platform/VulkanPlatformAndroid.cpp
+++ b/filament/backend/src/vulkan/platform/VulkanPlatformAndroid.cpp
@@ -293,9 +293,8 @@ VulkanPlatform::ExternalImageMetadata VulkanPlatformAndroid::extractExternalImag
         metadata.samples = VK_SAMPLE_COUNT_1_BIT;
         metadata.isStagingRequired = isSoftwareDecodedYUV(bufferDesc.format, bufferDesc.usage);
 
-        // Calculate mip levels. Software-decoded (staging) images are always single-level.
-        // Use ilogbf (exact, exponent-based) instead of floor(log2(...)) to avoid an
-        // off-by-one for power-of-two dimensions. Mirrors FTexture::maxLevelCount().
+        // Calculate mip levels. Use ilogbf (exact, exponent-based) instead of floor(log2(...)) to
+        // avoid an off-by-one for power-of-two dimensions. Mirrors FTexture::maxLevelCount().
         if (bufferDesc.usage & AHARDWAREBUFFER_USAGE_GPU_MIPMAP_COMPLETE) {
             uint32_t const maxDimension = std::max(bufferDesc.width, bufferDesc.height);
             metadata.mipLevels = std::max(1, std::ilogbf(float(maxDimension)) + 1);
--- a/filament/backend/src/vulkan/platform/VulkanPlatformAndroid.cpp
+++ b/filament/backend/src/vulkan/platform/VulkanPlatformAndroid.cpp
@@ -277,11 +277,6 @@ bool VulkanPlatformAndroid::copyExternalImageToMemoryYUV(
     return false;
 }
 
-// Define fallback preprocessor guards to avoid compilation errors on older Android SDKs.
-#ifndef AHARDWAREBUFFER_USAGE_GPU_MIPMAP_COMPLETE
-#define AHARDWAREBUFFER_USAGE_GPU_MIPMAP_COMPLETE 0x4000000ULL
-#endif
-
 VulkanPlatform::ExternalImageMetadata VulkanPlatformAndroid::extractExternalImageMetadata(
     ExternalImageHandleRef image) const {
     if (__builtin_available(android 26, *)) {
--- a/filament/backend/src/vulkan/platform/VulkanPlatformAndroid.cpp
+++ b/filament/backend/src/vulkan/platform/VulkanPlatformAndroid.cpp
@@ -35,6 +35,8 @@
 #include <android/hardware_buffer.h>
 #include <android/native_window.h>
 
+#include <algorithm>
+#include <cmath>
 #include <new>
 #include <utility>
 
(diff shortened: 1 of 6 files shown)


filament/backend/src/vulkan/platform/VulkanPlatformAndroid.cpp
```diff
@@ -275,6 +277,11 @@ bool VulkanPlatformAndroid::copyExternalImageToMemoryYUV(
     return false;
 }
 
+// Define fallback preprocessor guards to avoid compilation errors on older Android SDKs.
+#ifndef AHARDWAREBUFFER_USAGE_GPU_MIPMAP_COMPLETE
+#define AHARDWAREBUFFER_USAGE_GPU_MIPMAP_COMPLETE 0x4000000ULL
+#endif
```
pixelflinger: I don't think this is supported at every API levels? How do we handle running on device that doesn't support this?

z3moon: This was introduced as of the API level 28 (Android 9, Pie). And devices that don't support it will fallback to `metadata.mipLevels = 1;` below since `bufferDesc.usage` won't have 0x4000000ULL bit set.

filament/backend/src/vulkan/platform/VulkanPlatformAndroid.cpp
```diff
@@ -291,6 +298,17 @@ VulkanPlatform::ExternalImageMetadata VulkanPlatformAndroid::extractExternalImag
         metadata.samples = VK_SAMPLE_COUNT_1_BIT;
         metadata.isStagingRequired = isSoftwareDecodedYUV(bufferDesc.format, bufferDesc.usage);
 
+        // Calculate mip levels. Software-decoded (staging) images are always single-level.
+        // Use ilogbf (exact, exponent-based) instead of floor(log2(...)) to avoid an
+        // off-by-one for power-of-two dimensions. Mirrors FTexture::maxLevelCount().
+        if (!metadata.isStagingRequired &&
```
rafadevai: isnt enough to just check the flag? 

Also there are more formats that dont support mipmapping that are not related to software decoded YUVs.

z3moon: Good point. Updated.

filament/backend/src/vulkan/platform/VulkanPlatformAndroid.cpp
poweifeng: do you really need to define this?  I though the `__builtin_available(android 26, *)` would have protected the usage here?  Maybe you'd need a `__builtin_available(android 28, *)` somewhere.

z3moon: Yeah I think it's needed. `__builtin_available` is a runtime check, not a compile-time check. The `__builtin_available(android 26, *)` check here exists to support AHardwareBuffer. And  AHARDWAREBUFFER_USAGE_GPU_MIPMAP_COMPLETE was later introduced to support mipmap for AHardwareBuffer.

Also this ifndef fallback is safe because `bufferDesc.usage` wouldn't contain 0x4000000ULL for older versions anyway. Similarly AHARDWAREBUFFER_FORMAT_YV12 uses the same trick here.

poweifeng: `__builtin_available` is a static check at compile time (maybe also a runtime thing?).  I don't know what magic makes it work, but it guards features at compile time [ref](https://developer.android.com/ndk/guides/using-newer-apis#enabling_weak_api_references_in_your_build).  

For `AHARDWAREBUFFER_FORMAT_YV12`, it needed that define because it's a private API, whereas `AHARDWAREBUFFER_USAGE_GPU_MIPMAP_COMPLETE` is a public api.

z3moon: it looks like __builtin_available works for both compile time and runtime. And you're right about AHARDWAREBUFFER_FORMAT_YV12. It's private and this is public, so the example is not relevant here.

However, this won't build against android 27 (even if I wrap it with `__builtin_available(android 28, *)`) because AHARDWAREBUFFER_USAGE_GPU_MIPMAP_COMPLETE is not defined there. To my understanding, preprocessors are evaluated first prior to __builtin_available.

poweifeng: how are you building it?  I just tried and have no issue. 

I did do a bit of research, and `AHARDWAREBUFFER_USAGE_GPU_MIPMAP_COMPLETE` is defined even in ndk r26d.

z3moon: Hm yeah, I see android API 27 was introduced in NDK r18, which is really old. So you're right that I don't need that ifndef.

I didn't test the edge case myself, but just wrote it conservatively to make sure it should work in any case. But let me remove it.

poweifeng (approved):

filament/backend/src/vulkan/platform/VulkanPlatformAndroid.cpp
```diff
@@ -291,6 +293,16 @@ VulkanPlatform::ExternalImageMetadata VulkanPlatformAndroid::extractExternalImag
         metadata.samples = VK_SAMPLE_COUNT_1_BIT;
         metadata.isStagingRequired = isSoftwareDecodedYUV(bufferDesc.format, bufferDesc.usage);
 
+        // Calculate mip levels. Software-decoded (staging) images are always single-level.
```
rafadevai: "Software-decoded (staging) images are always single-level." this comment is not relevant anymore

rafadevai (approved):

rafadevai:
How did you test this? and did you check if you had any validation errors because of this?

filament/backend/src/vulkan/platform/VulkanPlatformAndroid.cpp
z3moon: Sure, removed.

z3moon:
> How did you test this? and did you check if you had any validation errors because of this?

I only conducted a local build test. I'll work with the user to verify its functionality tho.

PR #10154, opened 2026-06-28
gltfio: fix OOB read in WEIGHTS animation playback caused by duplicate timestamps

destro4evr-rgb:
## Problem

 `AnimatorImpl::applyAnimation()` computes `valuesPerKeyframe` by dividing `sampler->values.size()` by `times.size()`, where `times` is a `std::map<float, size_t>`. If an animation input accessor contains duplicate timestamp values, duplicate keys collapse in the map and `times.size()` ends up smaller than `timelineAccessor->count`. The values buffer is sized from `output->count` (based on the full keyframe count), so the computed stride is inflated. Subsequent accesses using `iter->second` — which stores the raw accessor index and can be as large as `count - 1` — then read past the end of the `sampler->values` buffer. This fires on every animation frame while the animation plays.

The assert at line 556 (`assert(sampler->values.size() % times.size() == 0)`) is compiled out under NDEBUG and provides no protection in release builds. Even in debug builds, for inputs where `values.size()` happens to be divisible by `times.size()`, the assert passes silently.

  ## Fix

  1. Detect duplicate timestamps in `createSampler()` and disable the sampler early, consistent with how other malformed inputs are handled.

  2. Store `inputCount` in the `Sampler` struct and use it instead of `times.size()` when computing `valuesPerKeyframe`, so the stride is always derived from the accessor's declared count rather than the deduplicated map size.

### DIFF
--- a/libs/gltfio/src/Animator.cpp
+++ b/libs/gltfio/src/Animator.cpp
@@ -53,6 +53,7 @@ struct Sampler {
     TimeValues times;
     SourceValues values;
     enum { LINEAR, STEP, CUBIC } interpolation;
+    size_t inputCount = 0;
 };
 
 struct Channel {
@@ -102,6 +103,12 @@ static void createSampler(const cgltf_animation_sampler& src, Sampler& dst) {
     }
     for (size_t i = 0, len = timelineAccessor->count; i < len; ++i) {
         dst.times[timelineFloats[i]] = i;
+    }   
+    dst.inputCount = timelineAccessor->count;
+    if (UTILS_UNLIKELY(dst.times.size() != dst.inputCount)) {
+        GLTFIO_WARN("Animation sampler has duplicate timestamps; animation disabled.");
+        dst.times.clear();
+        return;
     }
 
     // Convert source data to float.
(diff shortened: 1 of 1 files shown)


pixelflinger:
@destro4evr-rgb is an animation with duplicate timestamp actually malformed? is there a risk that this will cause some existing animations to stop playing (I'm thinking about broken tools for example).

destro4evr-rgb:
Hi @pixelflinger  

  The glTF 2.0 spec requires animation sampler input times to be strictly increasing - `time[n + 1] > time[n]` - so duplicate timestamps are formally malformed.

In this case, duplicate keys collapse in `times`, making `times.size()` smaller than `timelineAccessor->count`, which inflates `valuesPerKeyframe` and causes the WEIGHTS path to read past the end of `sampler->values` on every frame. The assert at the division site is compiled out under `NDEBUG`, so it provides no protection in release builds. Disabling the sampler for duplicate timestamps is safer than continuing with undefined behavior, and it is consistent with how other malformed inputs are already handled in `createSampler`.

pixelflinger (approved):

PR #10155, opened 2026-06-28
web: sync filament.d.ts with embind bindings and fix typing bugs

Erkko68:
## Background

While building a Kotlin Multiplatform wrapper around Filament (which generates
its Web/JS externals from `web/filament-js/filament.d.ts`), I found that the
hand-maintained TypeScript annotations have drifted behind the actual embind
bindings in `jsbindings.cpp`/`jsenums.cpp`: a number of functions are registered
and callable at runtime but are absent from (or mistyped in) the `.d.ts`, so
TypeScript consumers can't see them. This PR closes that gap. It's a
typings-only change — no binding or behavior changes, no wasm rebuild required.

## What this changes

**`filament.d.ts` — declare already-bound-but-undocumented API:**

- Added missing method declarations to existing classes, including `Engine`
  (feature-level queries, `getBackend`, `getEntityManager`, the `isValid*`
  family, `getSteadyClockTimeNano`), `View` (`getViewport`, `getSampleCount`,
  `setDithering`, dynamic-resolution getters, `getFogEntity`, …), `Scene`
  (`getSkybox`, `getIndirectLight`, `getEntityCount`), `Renderer`
  (`getClearOptions`, frame-pacing getters), `RenderableManager`
  (`getChannel`/`setChannel`, `getFogEnabled`, per-instance light channels),
  `TransformManager` (`getParent`, `getChildren`), `Camera` (`getShift`), and
  `Skybox` (`setLayerMask`), among others.
- Added the `TransparencyMode` and `FeatureLevel` enums (the latter with the
  explicit `= 1` ordinals matching the registered values).

**`filament.d.ts` — fix incorrect declarations:**

- `View` blend-mode getter/setter typed against the global `BlendMode` instead
  of `View$BlendMode`.
- `gltfio$FilamentAsset.geAssetInstances` typo → `getAssetInstances`.
- `RenderTarget.getMipLevel`/`getFace`/`getLayer` were missing their
  `attachment` parameter; added `getTexture(attachment)` too.
- `Animator.applyAnimation` was missing its `time` argument.
- `PickingQueryResult.renderable` typed as `number` instead of `Entity`.

**`test.ts` — restructure into a per-subsystem smoke harness:**

The buildability test was reorganized into focused `smoke_*` functions (one per
embind class/subsystem) driven by a single registry array, and extended to
exercise **every class currently declared in `filament.d.ts`** — so the typings
buildability check now actually covers the surface it's validating, and the
typings can't silently drift from the bindings again.

## Testing

`npx tsc --noEmit ../../third_party/gl-matrix/gl-matrix.d.ts ./filament.d.ts ./test.ts`
passes cleanly.


pixelflinger:
@poweifeng shouldn't filament.d.ts be generated automatically?

poweifeng (approved):

poweifeng:
> @poweifeng shouldn't filament.d.ts be generated automatically?

probably, but I'm going to propose deprecating the typescript support entirely. It's maintenance overhead with small benefit when we already support js.

Issue #10146, opened 2026-06-23
Clear coat BRDF documentation listing attenuates `Fr` twice


zyz23333:
In the clear coat section of the Filament documentation, the formula and GLSL listing appear to disagree.

Documentation location:

- `docs_src/src_markdeep/Filament.md.html`
- Section: `Clear coat model` -> `Integration in the surface response`
- Around lines 849-895 in the current tree

The formula says the base diffuse and specular lobes are each attenuated once:

```text
f(v,l) = fDiffuse(v,l) * (1 - Fc) + fSpecular(v,l) * (1 - Fc) + fc(v,l)
```

But the GLSL listing below it says:

```
return color * ((Fd + Fr * (1.0 - Fc)) * (1.0 - Fc) + Frc);
```

This expands to:

```
Fd * (1 - Fc) + Fr * (1 - Fc)^2 + Frc
```

so the base specular lobe `Fr` is attenuated twice.

The current shader implementation seems to match the formula rather than the listing:

- `shaders/src/surface_shading_model_standard.fs`
- `surfaceShading(...)`
- Around lines 132-153

Relevant code:

```
vec3 color = Fd + Fr * pixel.energyCompensation;

float Fcc;
float clearCoat = clearCoatLobe(pixel, h, NoH, LoH, Fcc);
float attenuation = 1.0 - Fcc;

color *= attenuation;
color += clearCoat;
```

The indirect lighting path also appears to attenuate both `Fd` and `Fr` once:

- `shaders/src/surface_light_indirect.fs`
- `isEvaluateClearCoatIBL(...)`
- Around lines 349-364

Relevant code:

```
float Fc = F_Schlick(0.04, 1.0, clearCoatNoV) * pixel.clearCoat;
float attenuation = 1.0 - Fc;
Fd *= attenuation;
Fr *= attenuation;
```

Should the documentation listing be updated to match the formula and current implementation, for example:

```
return color * ((Fd + Fr) * (1.0 - Fc) + Frc);
```

romainguy:
The documentation is an outdated attempt at accounting for energy loss in the clear coat layer. The code itself is correct.

zyz23333:
@romainguy Thanks for the clarification! I understand now. The implementation is clear, and I also really appreciate how well written the PBR documentation is overall. I’ll submit a PR to fix the outdated section.

romainguy:
@pixelflinger You can close this

PR #10157, opened 2026-06-29
Fix clear coat BRDF doc listing

zyz23333:
The clear coat documentation formula attenuates both base diffuse and base specular lobes once by `(1 - Fc)`, but the GLSL listing attenuated `Fr` twice.

This updates the listing to match the formula and the current shader implementation.

Issue:
#10146


romainguy (approved):

romainguy:
@pixelflinger This can go in

PR #10159, opened 2026-06-29
Add support for single-threaded execution in Engine

gsimko:
Handle single-threaded execution by directly executing commands instead of using a fence.


poweifeng (approved):

filament/src/details/Engine.cpp
```diff
@@ -878,6 +878,14 @@ bool FEngine::flushAndWait(uint64_t const timeout) {
     // enqueue finish command -- this will stall in the driver until the GPU is done
     getDriverApi().finish();
 
+    if constexpr (!UTILS_HAS_THREADING) {
+      // When executing in a single-threaded (e.g. wasm) environment, fence would
+      // fail. Instead, directly execute the outstanding commands.
+      execute();
+      getDriver().purge();
+      return true;
+    }
```
bejado: nit: spacing
```suggestion
    if constexpr (!UTILS_HAS_THREADING) {
        // When executing in a single-threaded (e.g. wasm) environment, fence would
        // fail. Instead, directly execute the outstanding commands.
        execute();
        getDriver().purge();
        return true;
    }
```

bejado (approved):

z3moon (approved):

filament/src/details/Engine.cpp
```diff
@@ -878,6 +878,14 @@ bool FEngine::flushAndWait(uint64_t const timeout) {
     // enqueue finish command -- this will stall in the driver until the GPU is done
     getDriverApi().finish();
 
+    if constexpr (!UTILS_HAS_THREADING) {
+        // When executing in a single-threaded (e.g. wasm) environment, fence would
+        // fail. Instead, directly execute the outstanding commands.
+        execute();
+        getDriver().purge();
+        return true;
```
pixelflinger: I don't like how the code is written with the early return. It would be easy in the future to add more stuff at the end and not realize there is an early exit.

Could we write this as just:

```
status = CONDITION_SATISFIED;
if (UTILS_HAS_THREADING) {
  do_fence_stuff()
} else {
  execute();
}
```

pixelflinger (changes requested):

filament/src/details/Engine.cpp
gsimko: Done.

gsimko:
I had to rebase my solution because Ben's suggestion was from his private account and that was not CLA approved. Ready for review.

pixelflinger (approved):

PR #10161, opened 2026-06-30
Part 1: don't crash if swapchain has no depth

anishmgoyal:
If the swapchain bundle doesn't have a depth buffer, right now, we'll still attempt to wrap it in a Texture, and eventually crash. Downstream (e.g. in VulkanHandles), the lack of a depth buffer is already handled.

It's likely that the scenario where the depth buffer isn't present is where the swapchain is being derived by an app and provided with custom behavior. To that end, right now, the priority is simply supporting those usecases by avoiding a crash. There will be a followup to allow creation of a swapchain without a depth buffer intentionally, using the built-in vulkan swapchain.

Tested by removing the depth buffer from the swapchain, and running hellotriangle (with the skybox set to render first).

### DIFF
--- a/filament/backend/src/vulkan/VulkanSwapChain.cpp
+++ b/filament/backend/src/vulkan/VulkanSwapChain.cpp
@@ -99,12 +99,16 @@ void VulkanSwapChain::update() {
         mColors.push_back(colorTexture);
     }
 
-    mDepth = fvkmemory::resource_ptr<VulkanTexture>::construct(mResourceManager, mContext, device,
-            mAllocator, mResourceManager, mCommands, bundle.depth, VK_NULL_HANDLE,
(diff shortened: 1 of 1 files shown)


poweifeng (approved):

poweifeng:
Actually, can you add the following to the *commit* message (not the PR description)

```
FIXES=528385261
```

anishmgoyal:
> Actually, can you add the following to the _commit_ message (not the PR description)
> 
> ```
> FIXES=528385261
> ```

sure - i'm gonna add it as a bug tag though, because i still don't know if they'd like a flag for no depth buffer on swapchain creation. it looks like they don't need one, but i want to verify that first

```
BUG=528385261
```

PR #10162, opened 2026-06-30
EXPECT_DEATH is not available on certain platforms

poweifeng:
In particular, iOS and web does not fork() to spin up a child process. Hence, we should only check for death when GTEST_HAS_DEATH_TEST.


z3moon (approved):

poweifeng:
no longer needed (test reverted).

PR #10163, opened 2026-06-30
webgpu: fix depthBias precision bug

poweifeng:
WebGPU api allows for only integer depth bias. This means that we cannot represent a bias of 0.5f (used in PCF shadow map). The workaround here is to lose half a unit of precision and just round to the nearest integer.


filament/backend/src/webgpu/WebGPUPipelineCache.cpp
```diff
@@ -147,7 +147,11 @@ void WebGPUPipelineCache::populateKey(RenderPipelineRequest const& request,
     outKey.fragmentShaderModuleHandle =
             request.fragmentShaderModule ? request.fragmentShaderModule.Get() : nullptr;
     outKey.pipelineLayoutHandle = request.pipelineLayout ? request.pipelineLayout.Get() : nullptr;
-    outKey.depthBias = static_cast<int32_t>(request.polygonOffset.constant);
+
+    // A note on the depth bias.  This is meant to increase the depth by constant units of depth
+    // and so rounding here is technically ok because we are missing only half of a unit in terms
+    // of precision.
+    outKey.depthBias = static_cast<int32_t>(std::round(request.polygonOffset.constant));
```
pixelflinger: I think we should take the ceil() otherwise there is a risk of not offsetting enough and creating z-fighting or self shadowing.

pixelflinger (changes requested):

filament/backend/src/webgpu/WebGPUPipelineCache.cpp
poweifeng: done

pixelflinger (approved):

PR #10165, opened 2026-07-01
Fix srgb conversion for FL0 with transparent blending mode

cardoso-victore:
In `shaders/src/surface_main.fs`, the custom sRGB emulation for FL0 applies gamma correction to `fragColor.rgb`.
However, for materials using `blending : transparent`, `fragColor.rgb` is pre-multiplied by alpha (Color_pre = Color_linear × Alpha) before this stage.

Applying `pow` directly to the pre-multiplied color is mathematically incorrect:

Output_filament = pow(color_linear * alpha, 0.4545) = pow(color_linear, 0.4545) * pow(alpha, 0.4545)

Since alpha lies in range [0,1], then pow(alpha, 0.4545) is significantly larger than alpha for semi-transparent pixels.
This makes the source color much brighter and more opaque than intended when it is blended by the hardware.

For example, with alpha = 0.047, we have pow(alpha, 0.4545) ~ 0.25, which is more than 5x increase in effective color intensity before blending.

### DIFF
--- a/shaders/src/surface_main.fs
+++ b/shaders/src/surface_main.fs
@@ -116,7 +116,13 @@ void main() {
 #if MATERIAL_FEATURE_LEVEL == 0
     if (CONFIG_SRGB_SWAPCHAIN_EMULATION) {
         if (frameUniforms.rec709 != 0) {
+#if defined(BLEND_MODE_TRANSPARENT) || defined(BLEND_MODE_FADE)
+            unpremultiply(fragColor);
             fragColor.rgb = pow(fragColor.rgb, vec3(0.45454));
+            fragColor.rgb *= fragColor.a;
+#else
+            fragColor.rgb = pow(fragColor.rgb, vec3(0.45454));
+#endif
         }
     }
 #endif


bejado (approved):

shaders/src/surface_main.fs
```diff
@@ -116,7 +116,15 @@ void main() {
 #if MATERIAL_FEATURE_LEVEL == 0
     if (CONFIG_SRGB_SWAPCHAIN_EMULATION) {
         if (frameUniforms.rec709 != 0) {
+#if defined(BLEND_MODE_TRANSPARENT)
+            if (fragColor.a > 0.0) {
```
romainguy: You can use the `unpremultiply()` function instead (from `common_graphics.fs`). It does this to avoid the division by 0:

```
color.rgb /= max(color.a, FLT_EPS);
```

That function is used in `surface_shading_lit.fs`.

romainguy (changes requested):

shaders/src/surface_main.fs
cardoso-victore: Done

shaders/src/surface_main.fs
```diff
@@ -116,7 +116,13 @@ void main() {
 #if MATERIAL_FEATURE_LEVEL == 0
     if (CONFIG_SRGB_SWAPCHAIN_EMULATION) {
         if (frameUniforms.rec709 != 0) {
+#if defined(BLEND_MODE_TRANSPARENT)
```
z3moon: blending == fade may also have the same issue as they're performed similarly. Probably need to add `|| defined(BLEND_MODE_FADE)`?

cardoso-victore: done

poweifeng (approved):

pixelflinger (approved):

PR #10167, opened 2026-07-01
filamentapp: add asset loading class

poweifeng:
This abstraction allows us to load assets from non-desktop
platforms such as Android or web.

We mark the auto-disk-load behavior in gltfio deprecated so that
disk-related action can eventually be removed from gltfio.

### DIFF
--- a/samples/gltf_instances.cpp
+++ b/samples/gltf_instances.cpp
@@ -16,7 +16,20 @@
 
 #include "common/arguments.h"
 
+#include "generated/resources/gltf_demo.h"
+
+#include "materials/uberarchive.h"
+
+#include <viewer/ViewerGui.h>
+
+#include <gltfio/AssetLoader.h>
+#include <gltfio/FilamentAsset.h>
+#include <gltfio/ResourceLoader.h>
+#include <gltfio/TextureProvider.h>
+
+#include <filamentapp/AssetLoader.h>
 #include <filamentapp/Config.h>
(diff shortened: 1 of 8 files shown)


samples/gltf_instances.cpp
```diff
@@ -222,6 +216,26 @@ int main(int argc, char** argv) {
             }
         }
 
+        // We explicitly fetch the raw bytes and push them into ResourceLoader via addResourceData.
+        // This pre-populates the cache and completely bypasses ResourceLoader's internal disk I/O,
+        // which is required for Android compatibility (where assets are packed in the APK).
+        for (size_t i = 0, c = app.asset->getResourceUriCount(); i < c; i++) {
+            const char* uri = app.asset->getResourceUris()[i];
+            utils::Path uriPath = filename.getParent() + uri;
```
z3moon, line 224: What if multiple textures reference the same file? would it load the same file multiple times? if it's not the intention, would the check as follows help?
```
if (app.resourceLoader->hasResourceData(uri)) {
    continue;
}
```

poweifeng: it's a good point, but after code inspection, seems like the uri is guarranteed to be unique -> [reference](https://github.com/google/filament/blob/main/libs/gltfio/src/AssetLoader.cpp#L534)

bejado (approved):

z3moon (approved):

PR #10169, opened 2026-07-02
utils: Implement Epoch-Based Reclamation (EBR) for Entity Management

pixelflinger:
This major architectural update introduces an Epoch-Based Reclamation (EBR)
system to the EntityManager. It delays the recycling of destroyed Entity IDs
until all active readers and component managers have advanced past the epoch
in which the destruction occurred.

The primary motivation for EBR is to decouple the lifecycle of the 32-bit
Entity ID from the dense tracking structures. By guaranteeing that an entity's
physical index is never safely recycled while any subsystem is still processing
its destruction, component managers can safely transition to tracking entities
purely by their raw index, dropping the generation count from their internal
bitsets. This paves the way for drastically shrinking the memory footprint
of structures like PagedArenaBitset (e.g., from 16KB down to 256B) in future
updates.

Key changes:
- Implement `advanceEpoch()`, `registerWatermark()`, and `reclaimSafeEpochs()`
  in EntityManager to orchestrate the global timeline.
- Introduce `catchupGarbage()` in SingleInstanceComponentManager to
  periodically query the central timeline for missed garbage and mark
  entities logically dead.
- Amortize the physical destruction of component data into a `gc()` phase,
  safely holding zombie components in dense arrays until they can be cleaned up.
- Prevent use-after-free ID collisions by keeping destroyed Entity indices
  in a purgatory epoch until the global safe threshold clears them.
- Introduce exhaustive concurrent stress tests to ensure thread-safety of
  the timeline and bitset pools.

### DIFF
--- a/android/filament-android/src/main/cpp/EntityManager.cpp
+++ b/android/filament-android/src/main/cpp/EntityManager.cpp
@@ -15,9 +15,9 @@
  */
 
 
-#include <jni.h>
-
 #include <utils/EntityManager.h>
+
+#include <jni.h>
 
 using namespace utils;
 
@@ -75,4 +75,23 @@ Java_com_google_android_filament_EntityManager_nIsAlive(JNIEnv*, jclass,
     EntityManager *em = (EntityManager *) nativeEntityManager;
     Entity& entity = *reinterpret_cast<Entity*>(&entity_);
     return (jboolean) em->isAlive(entity);
+}
+
+extern "C" JNIEXPORT void JNICALL
+Java_com_google_android_filament_EntityManager_nAdvanceEpoch(JNIEnv*, jclass,
+        jlong nativeEntityManager) {
+    EntityManager *em = (EntityManager *) nativeEntityManager;
+    em->advanceEpoch();
+}
+
+extern "C" JNIEXPORT jint JNICALL
+Java_com_google_android_filament_EntityManager_nGetMaxEntityCount(JNIEnv*, jclass) {
+    return (jint) EntityManager::getMaxEntityCount();
+}
+
+extern "C" JNIEXPORT jint JNICALL
+Java_com_google_android_filament_EntityManager_nGetEntityCount(JNIEnv*, jclass,
+        jlong nativeEntityManager) {
+    EntityManager *em = (EntityManager *) nativeEntityManager;
+    return (jint) em->getEntityCount();
 }
--- a/android/filament-android/src/main/java/com/google/android/filament/EntityManager.java
+++ b/android/filament-android/src/main/java/com/google/android/filament/EntityManager.java
@@ -77,10 +77,25 @@ public class EntityManager {
         return mNativeObject;
     }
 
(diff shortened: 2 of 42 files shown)


z3moon (approved):

bejado:
Gemini suggested a potential **AB-BA Deadlock** between the main engine thread and background timeline operations. I tried investigating myself and it seems possible? Though the timeline is a little hairy.

**Lock A:** `mEbrEntitiesLock` (Component Manager)
**Lock B:** `mTimelineLock` (EntityManager)

*   **Trace 1 (Main Thread):** `removeComponentsImpl()` acquires **Lock A** $\rightarrow$ calls `suspend()` $\rightarrow$ invokes `unregisterWatermark()`, which attempts to lock **Lock B**.
*   **Trace 2 (Background/Engine Thread):** `advanceEpoch()` (or `reclaimSafeEpochs()`) acquires **Lock B** $\rightarrow$ calls `getExpiredEpochsLocked()`, which attempts to acquire **Lock A** via the `info.entityBitsetLock` pointer.

I don't think this is a vulnerability at the moment, because `advanceEpoch()` is always called from the main thread. But it could be in the future, according to the comments in SingleInstanceComponentManager.h:

> However, the global EntityManager timeline (advanceEpoch() -> getExpiredEpochsLocked()) may be executed concurrently from background helper threads (e.g. background asset loaders).

pixelflinger:
> Gemini suggested a potential **AB-BA Deadlock** between the main engine thread and background timeline operations. I tried investigating myself and it seems possible? Though the timeline is a little hairy.
> 
> **Lock A:** `mEbrEntitiesLock` (Component Manager) **Lock B:** `mTimelineLock` (EntityManager)
> 
> * **Trace 1 (Main Thread):** `removeComponentsImpl()` acquires **Lock A** 
>     →
>    calls `suspend()` 
>     →
>    invokes `unregisterWatermark()`, which attempts to lock **Lock B**.
> * **Trace 2 (Background/Engine Thread):** `advanceEpoch()` (or `reclaimSafeEpochs()`) acquires **Lock B** 
>     →
>    calls `getExpiredEpochsLocked()`, which attempts to acquire **Lock A** via the `info.entityBitsetLock` pointer.
> 
> I don't think this is a vulnerability at the moment, because `advanceEpoch()` is always called from the main thread. But it could be in the future, according to the comments in SingleInstanceComponentManager.h:
> 
> > However, the global EntityManager timeline (advanceEpoch() -> getExpiredEpochsLocked()) may be executed concurrently from background helper threads (e.g. background asset loaders).

yes good catch. I fixed it, but also changed how we use the lock. Because we're guaranteed there can't be concurrent writes to `mEntities`, we can actually omit to acquire the lock at all for read operations on the SingleInstanceManager side. The lock is only needed when EntityManager does the EBR (reading) and when mEntities is mutated. This reduces the window we need the lock for significantly.

bejado (approved):

PR #10174, opened 2026-07-03
gltfio: validate KHR_materials_variants mapping index before use

adilburaksen:
**gltfio: validate KHR_materials_variants mapping index before use**

`FAssetLoader::createMaterialVariants` reads the per-primitive `KHR_materials_variants` mapping index
(`srcPrim.mappings[i].variant`) straight from the glTF and uses it to index `instance->mVariants`
(`AssetLoader.cpp:969`) with no bounds check. `mVariants` is a `FixedCapacityVector<Variant>` sized to the
top-level `variants_count` (AssetLoader.cpp:443-446), and `FixedCapacityVector::operator[]` is assert-only
(no check in release builds). A glTF whose mapping `variant` is `>= variants_count` therefore indexes past
the allocation, and the following `.mappings.push_back(...)` dereferences/writes through an out-of-bounds
`std::vector` — heap out-of-bounds access.

cgltf does reject this, but only in `cgltf_validate()` (cgltf.h:1721), which runs later in
`ResourceLoader::loadResources`; `createMaterialVariants` executes earlier during `createAsset`, so the
out-of-bounds access happens before validation.

Fix: skip a mapping whose `variant` index is out of range of `mVariants`, matching the existing
"index out of range -> continue" handling used for TEXCOORD/JOINTS set indices in this file.


libs/gltfio/src/AssetLoader.cpp
```diff
@@ -956,6 +956,13 @@ void FAssetLoader::createMaterialVariants(const cgltf_mesh* mesh, Entity entity,
         const cgltf_primitive& srcPrim = mesh->primitives[prim];
         for (size_t i = 0, m = srcPrim.mappings_count; i < m; i++) {
             const size_t variantIndex = srcPrim.mappings[i].variant;
+            // A malformed glTF (KHR_materials_variants) can reference a variant index that is out of
+            // range of the top-level `variants` array. cgltf only rejects this in cgltf_validate(),
+            // which runs later (in ResourceLoader::loadResources), so guard here to avoid an
+            // out-of-bounds access of mVariants before validation happens.
+            if (UTILS_UNLIKELY(variantIndex >= instance->mVariants.size())) {
+                continue;
```
z3moon, line 966: Could you add a warning here?

libs/gltfio/test/gltfio_test.cpp
```diff
@@ -462,6 +462,35 @@ TEST_F(glTFIOTest, MorphTargetsExceedingMaxDoNotOverflow) {
...
+            R"("nodes":[{"mesh":0}],"materials":[{}],)"
+            R"("meshes":[{"primitives":[{"attributes":{"POSITION":0},"mode":4,"material":0,)"
+            R"("extensions":{"KHR_materials_variants":{"mappings":[{"material":0,"variants":[1000]}]}}}]}],)"
+            R"("accessors":[{"bufferView":0,"componentType":5126,"count":3,"type":"VEC3","min":[0,0,0],"max":[1,1,1]}],)"
+            R"("bufferViews":[{"buffer":0,"byteOffset":0,"byteLength":36}],)"
+            R"("buffers":[{"byteLength":36}]})";
+
+    AssetLoader* assetLoader = AssetLoader::create({ mEngine, mMaterialProvider, mNameManager });
+    FilamentAsset* const asset = assetLoader->createAsset(
+            reinterpret_cast<uint8_t const*>(kGltf), uint32_t(std::strlen(kGltf)));
```
z3moon: nit: static_cast<uint32_t> instead

libs/gltfio/src/AssetLoader.cpp
adilburaksen: Done — added a `slog.w` warning before the `continue` in e6b89dc4.

libs/gltfio/test/gltfio_test.cpp
adilburaksen: Done in e6b89dc4.

PR #10176, opened 2026-07-06
Print readable material variant in error message

poweifeng:


bejado (approved):

libs/filabridge/include/private/filament/Variant.h
```diff
@@ -266,6 +266,34 @@ struct Variant {
...
+        };
+        if (variant.key & STE) print("STE");
+        if (variant.key & DEP) {
+            if (variant.key & MNT) print("MNT");
+            if (variant.key & PCK) print("PCK");
+        } else {
+            if (variant.key & S2D) print("S2D");
+            if (variant.key & FOG) print("FOG");
+        }
+        if (variant.key & DEP) print("DEP");
```
show50726: Can we put this into the if above?

poweifeng: done

show50726 (approved):

bejado (approved):

filament/src/MaterialDefinition.cpp
```diff
@@ -830,10 +830,9 @@ Program MaterialDefinition::getProgramWithVariants(FEngine const& engine,
             .diagnostics(name,
                     [variant, vertexVariant, fragmentVariant](CString const& name,
                             io::ostream& out) -> io::ostream& {
-                        return out << name.c_str_safe() << ", variant=(" << io::hex << +variant.key
-                                   << io::dec << "), vertexVariant=(" << io::hex
-                                   << +vertexVariant.key << io::dec << "), fragmentVariant=("
-                                   << io::hex << +fragmentVariant.key << io::dec << ")";
+                        return out << name.c_str_safe() << ", variant=(" << variant
+                                   << "), vertexVariant=(" << vertexVariant
+                                   << "), fragmentVariant=(" << fragmentVariant << ")";
```
z3moon: Looks like you could remove these parentheses as they're already wrapped in the function.

z3moon (approved):

PR #10177, opened 2026-07-08
material: fix colored penumbra enablement

poweifeng:
In the shader the use of color penumbra is guarded with "#if defined(HAS_COLORED_PENUMBRA)"

However,
"CodeGenerator::generateDefine(fs, "HAS_COLORED_PENUMBRA", hasColoredPenumbra ? 1u : 0u);"

means that HAS_COLORED_PENUMBRA is always defined.  This means we're accidentally enabling this feature.

The fix is use to boolean overload of
`CodeGenerator::generateDefine`, which ensures that the literal is not defined when the feature is not enabled in the material definition.


z3moon (approved):
Good catch! Could you add a comment why we need a boolean overload as opposed to others?

pixelflinger (approved):

bejado (approved):

PR #10178, opened 2026-07-08
gltfio: reject unsupported morph normal/tangent accessor types (avoid heap OOB write)

ataberk-xyz:
## Summary

`gltfio` never validates the element type of a morph-target NORMAL/TANGENT accessor. Such an accessor is later unpacked, during tangent-space computation in `ResourceLoader`, into a scratch buffer sized for 3 floats per vertex. If the accessor declares a wider element type (e.g. `MAT4`, 16 components), the unpacker writes more floats per element than the buffer holds — a heap out-of-bounds write triggered by loading an untrusted `.glb`/`.gltf`.

This validates the accessor type at load time (the same `getElementType` check already applied to other attributes), so an unsupported type is rejected before it can reach the tangent-space job.

## Details

In `FAssetLoader::createPrimitive`, the morph-attribute loop explicitly skips normal/tangent accessors before the `getElementType` type check:

```cpp
// The glTF normal and tangent data are ignored here, but honored in ResourceLoader.
if (atype == cgltf_attribute_type_normal || atype == cgltf_attribute_type_tangent) {
    continue;
}
```

`getElementType` accepts only scalar/vec2/vec3/vec4 and rejects everything else, but morph normal/tangent accessors never reach it. `cgltf_validate()` also imposes no semantic type constraint on morph attributes (it checks counts and sparse-index bounds only), so a `MAT4`-typed morph tangent accessor passes all validation.

Downstream, `TangentsJob` allocates `morphDeltas` as `new float3[vertexCount]` (3 floats/vertex) and calls `cgltf_accessor_unpack_floats(mtTangentsInfo, &morphDeltas[0].x, vertexCount * 3)`. For a sparse accessor, `cgltf_accessor_unpack_floats`'s sparse pass writes `cgltf_num_components(type)` floats per element (16 for `MAT4`) at `writer_index * 16`, past the 3-float-per-vertex allocation.

This path is the default, all-platform loader; loading untrusted glTF assets is the library's purpose.

## Fix

Run `getElementType` on morph normal/tangent accessors (rejecting unsupported types such as matrix types) before the existing `continue`. Valid `VEC3`/`VEC4` morph deltas are unaffected; only unsupported element types are rejected.

## Test

`MorphTargetTangentInvalidTypeIsRejected` loads a minimal crafted `.glb` whose morph target declares a `MAT4` tangent accessor. Before the fix this produced a heap-buffer-overflow (visible under AddressSanitizer) during tangent computation; after the fix `createAsset` returns null and no out-of-bounds access occurs. The crafted asset is embedded in the test (`morph_tangent_typeconf_glb.inc`) so it is self-contained. All existing `gltfio` tests continue to pass.

Verified with an ASan build (`-DFILAMENT_ENABLE_ASAN_UBSAN=ON`): pre-fix, `heap-buffer-overflow WRITE` in the tangent-space unpack; post-fix, the asset is rejected and the suite is clean.


ataberk-xyz:
Closing this in favor of a clean re-submission of the same fix (single commit, CLA pre-signed). See the replacement PR.

PR #10181, opened 2026-07-08
gltfio: represent every morph target in the extended loader (fix heap OOB write)

### LINKED ISSUE #10180: gltfio: heap out-of-bounds write in `AssetLoaderExtended::createPrimitive` (slotIndices sized by filtered morph count, indexed by raw index)
### Summary

Loading an untrusted `.glb`/`.gltf` through the extended asset loader can cause a heap out-of-bounds write in `AssetLoaderExtended::createPrimitive`. `slotIndices` is sized by the *filtered* morph-target count but written at the *raw* morph-target index, so when the morph-target filter drops an earlier target, a later target's raw index exceeds the array bounds.

### Affected code (branch `main`)

`libs/gltfio/src/extended/AssetLoaderExtended.cpp`:

```cpp
// createPrimitive
out->slotIndices.resize(morphTargets.size());   // sized by the FILTERED count
...
    if (slot.offset == 0xdeadbeef) {
        out->slotIndices[slot.slot] = outSlots.size() + i;   // indexed by the RAW morph index
    }
```

`morphTargets` (built earlier in `createPrimitive`) keeps a morph target only if it has a TANGENT attribute, or (no tangent) if the primitive has a non-unlit material. A POSITION-only morph target on a material-less primitive is dropped. But `slot.slot` is the raw morph-target index (`computeGeometries` sets `.slot = target.morphTarget`), so when target 0 is dropped and target 1 (raw index 1) is kept, `slotIndices` is `resize(1)` and the code writes `slotIndices[1]` — one element past the allocation.

### Reproduction (AddressSanitizer)

`.glb`: one primitive, **no material**, two morph targets — target[0] POSITION-only (dropped by the filter), target[1] with a TANGENT (kept, raw index 1). Loaded through the extended loader (`AssetConfiguration::ext` set; `AssetConfigurationExtended::isSupported()` true on desktop).

ASan trace (full-engine `test_gltfio` build, `-DFILAMENT_ENABLE_ASAN_UBSAN=ON`):

```
==ERROR: AddressSanitizer: heap-buffer-overflow ... WRITE of size 4 ...
    #0 AssetLoaderExtended::createPrimitive(...)  AssetLoaderExtended.cpp:558
    #1 FAssetLoader::createPrimitives(...)         AssetLoader.cpp:695
    ...
    #6 AssetLoader::createAsset(...)               AssetLoader.cpp:1821
0x... is located 0 bytes after 4-byte region  [ resize(1) == one int ]
allocated by:
    std::vector<int>::resize(...) <- AssetLoaderExtended.cpp:543   // the resize(morphTargets.size())
SUMMARY: AddressSanitizer: heap-buffer-overflow AssetLoaderExtended.cpp:558 in ...::createPrimitive(...)
```

I can attach the crafted `.glb` and a builder script, or a `test_gltfio` test case, whichever you prefer.

### Root cause / suggested direction

This is more than a size mismatch — the extended loader's morph-target *filter* (which decides which targets need a tangent-space job) is being used to size and populate the morph-slot bookkeeping, but the consumer (`FAssetLoader::createRenderable`) indexes `slotIndices` by the raw target index and asserts `slot.slot == tindex` for every POSITION-bearing target:

```cpp
// createRenderable
for (int tindex = 0; tindex < numMorphTargets && (size_t) tindex < numSlots; ++tindex) {
    ...
    BufferSlot& slot = slots[outputPrim->slotIndices[tindex]];
    assert_invariant(slot.slot == tindex);
}
```

So a target that the extended filter drops has no slot, yet the consumer still expects one (and, separately, that target's morph POSITION data is not uploaded). The non-extended loader (`createPrimitive` in `AssetLoader.cpp`) sizes `slotIndices` to the full `targetsCount` and writes an entry for every target, keeping this invariant.

Simply changing the extended `resize(morphTargets.size())` to `resize(targetsCount)` stops the out-of-bounds write but leaves the dropped targets pointing at the wrong slot (it trips the `slot.slot == tindex` assertion in debug builds). The correct fix likely needs the extended path to create a morph slot for every target (decoupling the tangent-computation filter from the morph-slot/POSITION bookkeeping), matching the non-extended loader — which is why I'm filing this as an issue rather than a one-line PR, since it touches the extended loader's slot model.

### Impact

Heap out-of-bounds write from an untrusted, `cgltf_validate`-clean `.glb`, during asset loading, on the desktop/tools extended loader path (`isSupported()` is false on Android/iOS/Emscripten). Controlled write offset; the written value is a small non-negative integer.

Happy to help with a repro or a patch once you confirm the intended direction for the slot model.

ataberk-xyz:
## Summary

`AssetLoaderExtended::createPrimitive` builds its `slotIndices` array from a **filtered** subset of a primitive's morph targets (only those carrying a TANGENT attribute, or — when there is no tangent — those on a non-unlit material), but it **indexes that array by the raw morph-target index** (`slot.slot`, set to the target's original index). When an earlier target is filtered out, a later kept target's raw index exceeds the array bounds, producing a heap out-of-bounds write while loading an untrusted glTF asset through the extended loader. Filtered-out targets also have their POSITION deltas silently dropped.

Fixes #10180.

## Details

`FAssetLoader::createRenderable` consumes `slotIndices` indexed by the **raw** morph-target index and asserts each slot's `.slot` equals that index:

```cpp
for (int tindex = 0; tindex < numMorphTargets && (size_t) tindex < numSlots; ++tindex) {
    ...
    BufferSlot& slot = slots[outputPrim->slotIndices[tindex]];
    assert_invariant(slot.slot == tindex);
}
```

The non-extended loader (`FAssetLoader::createPrimitive`) sizes `slotIndices` to the full `targetsCount` and writes an entry for every target, so this invariant holds there. The extended loader's filter breaks it: a target that is dropped has no slot, yet the consumer still indexes `slotIndices` at its raw index — reading past the (shorter) array on the write side (`slotIndices[slot.slot]`) and mismatching `slot.slot` vs `tindex` on the read side.

Reproduced with AddressSanitizer (a material-less primitive whose morph target[0] is POSITION-only and target[1] carries a tangent): `heap-buffer-overflow WRITE` at the `slotIndices[slot.slot]` store, with the allocation frame at the `slotIndices.resize(...)`.

## Fix

Represent **every** morph target in `createPrimitive` (rather than the tangent/material-filtered subset), matching the non-extended loader. This keeps `slotIndices` sized to and indexed consistently with the consumer, and ensures every target's POSITION deltas are uploaded.

Note for reviewers: the previous filter skipped tangent-space computation for morph targets without a tangent attribute on unlit/material-less primitives. Since a material-less primitive is assigned a (lit) default material, those targets generally do need tangents anyway; representing every target also uploads their POSITION data, which the filter previously dropped. If you'd prefer to preserve the tangent-computation skip as an optimization while still giving every target a POSITION slot, I'm happy to rework it that way — this PR takes the simplest correct approach.

## Test

`ExtendedLoaderMorphTargetSlotIndicesInBounds` loads a small crafted asset (embedded) exercising the filtered-target case; it previously overflowed `slotIndices` and now loads cleanly. All existing `gltfio` tests pass. Verified under an ASan build (`-DFILAMENT_ENABLE_ASAN_UBSAN=ON`).


libs/gltfio/test/ext_slotindices_glb.inc
```diff
@@ -0,0 +1,98 @@
+// Crafted glb: material-less primitive, morph target[0] POSITION-only, target[1] with TANGENT.
+static const unsigned char EXT_SLOTINDICES_GLB[] = {
```
z3moon, line 2: this raw hex dump makes it difficult to extend this file in the future. please consider providing an readable/editable file with an auto-generation pipeline.

libs/gltfio/src/extended/AssetLoaderExtended.cpp
```diff
@@ -484,16 +483,14 @@ bool AssetLoaderExtended::createPrimitive(Input* input, Output* out,
...
-        if (!morphTargetHasNormals && prim->material && !prim->material->unlit) {
-            morphTargets.push_back(targetIndex);
         }
+        // Every morph target contributes a buffer slot (its POSITION deltas are uploaded and its
+        // tangent space is computed). The slotIndices array populated below is indexed downstream
+        // in FAssetLoader::createRenderable by the *raw* morph-target index, which also requires
+        // each slot's .slot to equal that index. Representing only a filtered subset of targets
+        // here desynchronizes slotIndices from that consumer and underflows it -- an out-of-bounds
+        // write when an earlier target is skipped -- and drops the skipped targets' POSITION data.
+        morphTargets.push_back(targetIndex);
```
z3moon, line 493: yes, we'd prefer the tangent-computation skip as before. could you consider doing a rework for this?

PR #10183, opened 2026-07-08
android: add NativeViewer to wrap FilamentApp

poweifeng:
This introduces NativeViewer.kt and its C++ JNI bridge to map
Android lifecycle methods (surface creation, resizing, and
touch events) to native FilamentApp callbacks.

Changes:
 - Added NativeViewer.kt for Kotlin event handling.
 - Implemented NativeViewer.cpp JNI bridge mapping.
 - Added lifecycle stubs to FilamentApp.cpp.
 - Left ModelViewer.kt unmodified as a reference for java/kotlin
   api.
 - Changed filamentapp inclusion in root CMakeLists.txt
 - Separate imgui deps into its own files to avoid including
   imgui in android.
 - Avoid depending on sdl2 and civetweb for android build.


android/gradle.properties
```diff
@@ -22,8 +22,8 @@ android.defaults.buildfeatures.buildconfig=true
 android.nonTransitiveRClass=false
 android.nonFinalResIds=false
 
-com.google.android.filament.tools-dir=../../../out/release/filament
-com.google.android.filament.dist-dir=../out/android-release/filament
+com.google.android.filament.tools-dir=../../../out/debug/filament
+com.google.android.filament.dist-dir=../out/android-debug/filament
```
pixelflinger: probably unintended?

pixelflinger (changes requested):

android/gradle.properties
poweifeng: removed

bejado (approved):

PR #10187, opened 2026-07-09
gl: MacOS deadlock due to blocking in destroySwapChain

poweifeng:
In a recent rework/addition of headless swapchain to opengl+macos (CocoaGL), a bug was introduced by having destroySwapchain block on mainthread work (dispatch_sync).  This causes a deadlock on shutdown because the main thread is waiting on the backend thread to complete, but the backend thread is destroying the swapchain and waiting on the main thread to destroy the window and view associated with the swapchain.

The solution is just to do it with dispatch_async() on destroy as opppose to dispatch_sync().


filament/backend/src/opengl/platforms/PlatformCocoaGL.mm
```diff
@@ -292,19 +303,34 @@
...
+    }
 
+    if (headlessView) {
+       // Delete the no longer needed headless view and window on the main thread. Note that this
+       // call is non-blocking.
+       dispatch_async(dispatch_get_main_queue(), ^{
+            utils::LockGuard const lock(pImpl->mHeadlessResourcesLock);
+            auto& views = pImpl->mHeadlessSwapChains;
+            auto& windows = pImpl->mHeadlessWindows;
+            if (auto it = std::find(views.begin(), views.end(), cocoaSwapChain->view);
```
z3moon: I think `cocoaSwapChain->view` could crash because the dispatch call has became non-blocking and cocoaSwapChain is delete right after this dispatch_async. Maybe you could use headlessView instead?

poweifeng: yeah good catch, that was actually the original intention.

poweifeng: done

bejado (approved):

PR #10188, opened 2026-07-09
implements EVSSM (a.k.a. PCSS with EVSM)

pixelflinger:
PCSS is typically extremely expensive due to the large number of samples required for blocker search and shadow blurring.

By leveraging EVSM, we can achieve PCSS with a handful of samples. This improves performance significantly on the shadow sampling side. Of course, we pay a price on the shadow map generation side, since EVSM is typically 2 to 4 times slower.

With this change, we deprecate and remove support for DPCF, which doesn't make sense anymore. It was expensive and low quality.

evssm

RDIFF_ACCEPT_NEW_GOLDENS
SIZEGUARD_BYPASS


filament/src/ShadowMapManager.cpp
```diff
@@ -748,7 +858,77 @@ FrameGraphId<FrameGraphTexture> ShadowMapManager::vsmMipmapPass(
...
+                auto const pipeline = ppm.getPipelineState(mi);
+                backend::Viewport const scissor = { 0, 0, dim, dim };
+
+                mi->setParameter("color", in, SamplerParams{
+                        .filterMag = SamplerMagFilter::NEAREST,
+                        .filterMin = SamplerMinFilter::NEAREST_MIPMAP_NEAREST
+                });
+                // FIXME: use real scissor
+                mi->setParameter("srcRect", int4{ 0, 0, width, width });
+                mi->setParameter("dstRect", int4{ scissor.left, scissor.bottom, scissor.width, scissor.height });
```
bejado, line 926: Does `dstRect` take into account the shadow atlas?

filament/src/ShadowMapManager.cpp
```diff
@@ -131,6 +142,103 @@ void ShadowMapManager::terminate(FEngine& engine) {
...
+
+    // no limits by default
+    float maxSearchRadius = std::numeric_limits<half>::max();
+    if (shadowMap.getShadowLightType() == ShadowLightType::DIRECTIONAL) {
+        // The maximum search radius the shader will ever need for a directional light is given by
+        // bulbRadius * (f - n)
+        // But we're limited by the number of mip levels we have relative to the shadowmap resolution.
+        // We must always use the first cascade parameters for this calculation.
+        float2 const wsTexelSizeAtOneMeter0 = getCascadeShaderParameters(0).texelSizeAtOneMeterWs;
+        maxSearchRadius = computeDynamicMaxSearchRadius(PCSS_MIP_LEVELS,
```
bejado: The first argument to `computeDynamicMaxSearchRadius` is `maxMipLevel`. Should this be `PCSS_MIP_LEVELS - 1`?

shaders/src/surface_shadowing.fs
```diff
@@ -408,25 +189,202 @@ float ShadowSample_VSM(const bool DIRECTIONAL, const highp sampler2DArray shadow
...
+    // 2. Setup the Spatial Grid (Decoupled Base)
+    highp vec2 r = (penumbraWidthInTexels * texelSize) * 0.5 * lodDeficit;
+    highp vec2 rotX = vec2(r.x, 0.0);
+    highp vec2 rotY = vec2(0.0, r.y);
+    highp vec2 anchor = position.xy;
+
+    // 3. Inject Noise (Rotates the grid and jitters the anchor)
+    #if defined(SHADOW_SAMPLING_EVSSM_NOISE) && SHADOW_SAMPLING_EVSSM_NOISE == 1
+        highp float noise = interleavedGradientNoise(gl_FragCoord.xy);
+        highp float theta = noise * 6.28318530718;
```
romainguy: Is that 2*PI? Would be good to have a comment or a named constant

filament/src/ShadowMapManager.cpp
pixelflinger: so the shadow atlas is not enabled right now and still WIP, so I'm not worrying too much about it yet.

shaders/src/surface_shadowing.fs
pixelflinger: yes good catch.

filament/src/ShadowMapManager.cpp
pixelflinger: yes. fixed!

shaders/src/surface_shadowing.fs
pixelflinger: fixed

bejado (approved):

PR #10195, opened 2026-07-10
Add utils::debug::Mutex lock-order debugger 

pixelflinger:
Introduce an opt-in, graph-based lock debugging tool (`debug::Mutex`)
to detect multi-threaded lock-order inversions and self-deadlocks at
runtime.

Key changes:
- utils::debug::Mutex: Implements a global dependency graph checked via
  BFS during lock() and try_lock(). When a cyclic dependency or
  recursive lock on a non-recursive mutex is detected, logs a detailed
  diagnostic report to LOG(ERROR) and calls std::abort().
- Callstack Tracing: Captures exact CallStack traces for lock creation
  points (`getCreationStack()`) and lock acquisition points (`lock()`
  and `try_lock()`), stripping internal wrapper frames.
- Condition Variable Integration: Updated generic and Linux condition
  variables (`Condition::wait`) to support templated mutex types so
  they work transparently with debug::Mutex.
- Build & Sizing Integration: Enabled via the `-u` build.sh flag or
  `-DFILAMENT_DEBUG_MUTEX=ON`. Enlarges backend HandleAllocator pool
  bucket sizes when enabled (`P2 = 312` bytes) so instrumented mutexes
  inside objects like GLSyncFence fit without pool overflow.

### DIFF
--- a/filament/backend/test/test_Handles.cpp
+++ b/filament/backend/test/test_Handles.cpp
@@ -28,10 +28,13 @@ using namespace filament::backend;
 static constexpr uint32_t HANDLE_HEAP_FLAG = 0x80000000u;
 static constexpr size_t POOL_SIZE_BYTES = 8 * 1024U * 1024U;
 // NOTE: actual count may be lower due to alignment requirements
+#if defined(FILAMENT_DEBUG_MUTEX) || defined(UTILS_DEBUG_MUTEX)
+constexpr size_t const POOL_HANDLE_COUNT = POOL_SIZE_BYTES / (32 + 96 + 312);
+#define HandleAllocatorTest  HandleAllocator<32,  96, 312>
+#else
 constexpr size_t const POOL_HANDLE_COUNT = POOL_SIZE_BYTES / (32 + 96 + 184); // 31775
-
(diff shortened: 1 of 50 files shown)


bejado (approved):

filament/backend/test/test_Handles.cpp
```diff
@@ -28,10 +28,13 @@ using namespace filament::backend;
 static constexpr uint32_t HANDLE_HEAP_FLAG = 0x80000000u;
 static constexpr size_t POOL_SIZE_BYTES = 8 * 1024U * 1024U;
 // NOTE: actual count may be lower due to alignment requirements
+#if defined(FILAMENT_DEBUG_MUTEX) || defined(UTILS_DEBUG_MUTEX)
```
poweifeng, line 31: How does `UTILS_DEBUG_MUTEX` differ from `FILAMENT_DEBUG_MUTEX`?

pixelflinger: this was just a way to be able to use libutils  kind of independently of filament. One is project level, the other is library level. maybe it doesn't add much value.

PR #10203, opened 2026-07-13
Adds an extra command pool field in vkCommands

jorgeag-google:
This add an extra command pool to the vKCommands struct that will be used for the async work.

This is the first step to make the async commands in the VK backend really asynchronous.


filament/backend/src/vulkan/VulkanCommands.cpp
```diff
@@ -386,19 +386,23 @@ void CommandBufferPool::insertEvent(char const* marker) {
 #endif // FVK_DEBUG_GROUP_MARKERS
 
 VulkanCommands::VulkanCommands(VkDevice device, VkQueue queue, uint32_t queueFamilyIndex,
-        VkQueue protectedQueue, uint32_t protectedQueueFamilyIndex, VulkanContext const& context,
+        VkQueue protectedQueue, uint32_t protectedQueueFamilyIndex,
+        std::unique_ptr<CommandBufferPool> asyncPool, VulkanContext const& context,
```
rafadevai: this seems the wrong approach to me. Its strange to pass down one pool but then internally it creates one. Also what about the mSemaphoreManager. 

In this case for the async, accessing the semaphore manager is that thread safe?

jorgeag-google: I changed the constructor. I am thinkinng if the semaphore manager needs to be thread safe. I know it is referenced in the command pool

PR #10204, opened 2026-07-14
gltf_Viewer: don't set framerate if not supported

pixelflinger:


libs/viewer/src/ViewerGui.cpp
```diff
@@ -1072,7 +1072,11 @@ void ViewerGui::updateUserInterface() {
         ImGui::Checkbox("Show skybox", &mSettings.viewer.skyboxEnabled);
         ImGui::ColorEdit3("Background color", &mSettings.viewer.backgroundColor.r);
 
-        ImGui::SliderFloat("Frame rate", &mSettings.viewer.cameraFrameRate, 0.0f, 120.0f, "%.0f Hz");
+        // we don't have a way to report the real value of SwapChain::isFrameRateChangeSupported(),
+        // for now we assume it's supported in remote more only
```
poweifeng:
```suggestion
        // for now we assume it's supported in remote mode only
```

poweifeng (approved):

PR #10211, opened 2026-07-15
driver: remove DriverConfig fields that are feature flags

poweifeng:
We no longer need them because of FeatureFlagManager, which is passed along with DriverConfig.


filament/backend/src/opengl/ShaderCompilerService.cpp
```diff
@@ -201,8 +201,17 @@ ShaderCompilerService::OpenGLProgramToken::~OpenGLProgramToken() {
 ShaderCompilerService::ShaderCompilerService(OpenGLDriver& driver)
         : mDriver(driver),
           mBlobCache(driver.getContext()),
-          mCallbackManager(driver) {
-}
+          mCallbackManager(driver),
+          mParallelShaderCompileDisabled(driver.getDriverConfig().featureFlagManager
```
bejado, line 205: I'm wondering if `getDriverConfig().featureFlagManager` can ever actually be `nullptr`. I know we have it marked as `UTILS_NULLABLE`, but `getDriverConfig` always returns a config with a non-null feature manager. Maybe `featureFlagManager` should be `UTILS_NONNULL`.

bejado (approved):

filament/backend/src/opengl/ShaderCompilerService.cpp
pixelflinger: I think the backend unit tests it can be absent

pixelflinger (approved):

PR #10212, opened 2026-07-15
vk: fix missing validation error for VK_KHR_create_renderpass2

poweifeng:
It's needed because of VK_KHR_depth_stencil_resolve, which is needed for dynamic rendering.


filament/backend/src/vulkan/platform/VulkanPlatform.cpp
```diff
@@ -215,23 +215,31 @@ ExtensionSet getDeviceExtensions(VkPhysicalDevice device) {
         // We only support external image for Android for now, but nothing bars us from
         // supporting other platforms.
 #if defined(__ANDROID__)
-        VK_KHR_EXTERNAL_MEMORY_EXTENSION_NAME,
+        // TODO: is this needed?
```
rafadevai: yes this is needed for external images same for VK_EXT_QUEUE_FAMILY_FOREIGN_EXTENSION_NAME

rafadevai (approved):

filament/backend/src/vulkan/platform/VulkanPlatform.cpp
poweifeng: noted

anishmgoyal (approved):

PR #10214, opened 2026-07-16
vk: set the renderpass name when DebugUtils is on

poweifeng:
- This PR also enables DebugUtils by default on debug builds. (when to enable it will be re-visited).
 - Separated validation from DebugUtils so that they are not tied together in enablement.
 - Together with Android Performance Analyzer, you should be able to see named renderpasses in the GPU non-fragment, fragment workload on the timeline.

### DIFF
--- a/filament/backend/src/vulkan/VulkanAsyncHandles.cpp
+++ b/filament/backend/src/vulkan/VulkanAsyncHandles.cpp
@@ -17,6 +17,7 @@
 #include "VulkanAsyncHandles.h"
 
 #include "VulkanConstants.h"
+#include "VulkanDriver.h"
 #include "VulkanFencePool.h"
 
 #include "vulkan/utils/Spirv.h"
--- a/filament/backend/src/vulkan/VulkanDriver.cpp
+++ b/filament/backend/src/vulkan/VulkanDriver.cpp
@@ -193,7 +193,7 @@ VKAPI_ATTR VkBool32 VKAPI_CALL debugReportCallback(VkDebugReportFlagsEXT flags,
 }
 #endif // FVK_ENABLED(FVK_DEBUG_VALIDATION)
 
-#if FVK_ENABLED(FVK_DEBUG_DEBUG_UTILS)
+#if FVK_ENABLED(FVK_DEBUG_DEBUG_UTILS) && FVK_ENABLED(FVK_DEBUG_VALIDATION)
 VKAPI_ATTR VkBool32 VKAPI_CALL debugUtilsCallback(VkDebugUtilsMessageSeverityFlagBitsEXT severity,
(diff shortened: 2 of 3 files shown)


pixelflinger (approved):

filament/backend/src/vulkan/VulkanAsyncHandles.cpp
```diff
@@ -17,6 +17,7 @@
 #include "VulkanAsyncHandles.h"
 
 #include "VulkanConstants.h"
+#include "VulkanDriver.h"
```
rafadevai, line 20: why do we need to include this here?

filament/backend/src/vulkan/VulkanDriver.cpp
```diff
@@ -2387,10 +2379,13 @@ void VulkanDriver::beginRenderPass(Handle<HwRenderTarget> rth, const RenderPassP
 // Assign a label to the framebuffer for debugging purposes.
 #if FVK_ENABLED(FVK_DEBUG_GROUP_MARKERS | FVK_DEBUG_DEBUG_UTILS)
     auto const topMarker = mCommands.getTopGroupMarker();
+    uint64_t fbVk = (uint64_t) vkfb->getVkFramebuffer();
```
rafadevai: put them inside the if block

rafadevai (approved):

filament/backend/src/vulkan/VulkanDriver.cpp
```diff
@@ -193,7 +193,7 @@ VKAPI_ATTR VkBool32 VKAPI_CALL debugReportCallback(VkDebugReportFlagsEXT flags,
 }
 #endif // FVK_ENABLED(FVK_DEBUG_VALIDATION)
 
-#if FVK_ENABLED(FVK_DEBUG_DEBUG_UTILS)
+#if FVK_ENABLED(FVK_DEBUG_DEBUG_UTILS) && FVK_ENABLED(FVK_DEBUG_VALIDATION)
```
anishmgoyal, line 196: could we just check the validation flag?

anishmgoyal (approved):

filament/backend/src/vulkan/VulkanAsyncHandles.cpp
poweifeng: there's a reference in line 165.  Only compiled when `FVK_DEBUG_DEBUG_UTILS` is enabled.

filament/backend/src/vulkan/VulkanDriver.cpp
poweifeng: good call. done

filament/backend/src/vulkan/VulkanDriver.cpp
poweifeng: they are independent.  You could have one without the other.

pixelflinger (approved):

PR #10217, opened 2026-07-17
Fix bug in set Store/Load op for DepthStencil in VK

jorgeag-google:
In the case the Depth/Stencil attachment exists, but the use is a pure one (pure Stencil or pure Depth) we set the flags of the other operations (the one we do not use) wrongly.

This fix checks the use of Depth and Stencil separately and in case any of those do not exists, sets his corresponding operation to VK_ATTACHMENT_STORE_OP_DONT_CARE and
VK_ATTACHMENT_LOAD_OP_DONT_CARE


poweifeng (approved):

poweifeng:
lgtm but I have to say what does this mean in the context of the spec?  If the buffer is not present (say depth buffer), does it matter what I put for those ops?

rafadevai (approved):

PR #10218, opened 2026-07-20
Fix build issues in certain environments

z3moon:


poweifeng (approved):

bejado (approved):

pixelflinger:
`std::ostream` is banned from filament (mega-bloat), and we never had to need it before.

PR #10219, opened 2026-07-21
filamentapp: FilamentApp2 to replace FilamentApp

poweifeng:
- FilamentApp2 differs from FilamentApp in two aspects:
 - Not a singleton (o allow for better life-time management
 - Use Builder pattern to set parameters
 - Refactor all cpp samples to use FilamentApp2
 - Mark FilamentApp as deprecated
 - Move Config to be a sample specific class - SampleConfig

### DIFF
--- a/libs/filamentapp/CMakeLists.txt
+++ b/libs/filamentapp/CMakeLists.txt
@@ -24,6 +24,7 @@ set(PUBLIC_HDRS
         include/filamentapp/Cube.h
         include/filamentapp/Grid.h
         include/filamentapp/FilamentApp.h
+        include/filamentapp/FilamentApp2.h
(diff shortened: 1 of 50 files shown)


pixelflinger:
quick question... would it be possible to reimplement FilamentApp in terms of FilamentApp2 -- so we could keep backward source compatibility. Maybe it could be a header-only file.

poweifeng:
> quick question... would it be possible to reimplement FilamentApp in terms of FilamentApp2 -- so we could keep backward source compatibility. Maybe it could be a header-only file.

yes, most likely.  I'll do that in a follow-up.

pixelflinger (approved):

PR #10229, opened 2026-07-21
github: [header reorg check] add tag for skipping

poweifeng:
Writing SKIP_HEADER_CHECK on its own line will now allow
skipping the header check. SKIP_HEADER_CHECK=a,b can be used
to skip specific files.

This is useful in the case when third party headers do not match
filament expectations.


z3moon:
so `SKIP_HEADER_CHECK` allows us to bypass the header check? could we have the detail in the commit/PR message?

poweifeng:
> so `SKIP_HEADER_CHECK` allows us to bypass the header check? could we have the detail in the commit/PR message?

done

z3moon (approved):

PR #10231, opened 2026-07-21
Fix washed-out textures in Filament FL0 GLES2 emulation.

cardoso-victore:
This change corrects sRGB texture handling in Filament's Feature Level 0 (GLES2 emulation) mode by initializing the GL_EXT_sRGB extension flag when running on GLES 3.0+ or when the extension is present. Instead of downgrading SRGB8_A8 textures to linear RGBA8, the OpenGL backend now correctly retains sRGB formats (GL_SRGB_ALPHA_EXT / GL_SRGB_EXT) and ensures matching format parameters during storage allocation and data updates as required by the specification. Finally, FL0 Scuba goldens have been updated to reflect the corrected, non-washed-out rendering, allowing both FL0 and FL1 raster tests to pass.

FIXES=533462283


filament/backend/src/opengl/OpenGLContext.cpp
```diff
@@ -679,6 +679,7 @@ void OpenGLContext::initExtensionsGLES(Extensions* ext, GLint major, GLint minor
     ext->EXT_texture_compression_bptc = exts.has("GL_EXT_texture_compression_bptc"sv);
     ext->EXT_texture_cube_map_array = exts.has("GL_EXT_texture_cube_map_array"sv) || exts.has("GL_OES_texture_cube_map_array"sv);
     ext->EXT_texture_filter_anisotropic = exts.has("GL_EXT_texture_filter_anisotropic"sv);
+    ext->EXT_texture_sRGB = exts.has("GL_EXT_sRGB"sv) || (major >= 3);
```
poweifeng: the `(major >= 3)` part is unusual.  Shouldn't that be captured in `exts` already?

filament/backend/src/opengl/OpenGLDriver.cpp
```diff
@@ -1005,6 +1005,11 @@ void OpenGLDriver::textureStorage(OpenGLState& gl, GLTexture* t,
             else {
                 // FIXME: handle compressed texture format
                 auto [format, type] = textureFormatToFormatAndType(t->format);
+                if (t->gl.internalFormat == GL_SRGB_ALPHA_EXT) {
+                    format = GL_SRGB_ALPHA_EXT;
+                } else if (t->gl.internalFormat == GL_SRGB_EXT) {
+                    format = GL_SRGB_EXT;
+                }
```
pixelflinger: you shouldn't need to do any of this because the _EXT enums are compatible with the ES3 enums, i.e.:

GL_SRGB_ALPHA_EXT == GL_SRGB8_ALPHA8

pixelflinger (changes requested):

filament/backend/src/opengl/OpenGLContext.cpp
cardoso-victore: Removed and verified that render tests still work.

filament/backend/src/opengl/OpenGLDriver.cpp
cardoso-victore: Removed and verified that render tests still work.

poweifeng (approved):

pixelflinger (approved):

z3moon (approved):

pixelflinger:
@cardoso-victore fails on iOS.

pixelflinger (approved):

PR #10111, opened 2026-06-12
fix(third_party): update tinyexr and reject negative tile coords

aleister1102:
## Summary
- Upgrade vendored tinyexr to upstream `release` @ `743d872` (adds required `exr_reader.hh` and `streamreader.hh`).
- Reject negative tiled chunk coordinates during parse and before RGBA assembly to prevent heap OOB writes on crafted EXR input reaching `LoadEXRFromMemory()`.

## Test plan
- [x] `./build.sh -ip desktop debug` succeeds (`tinyexr`, `imageio`, `cmgen`)
- [x] Audit PoC: valid 64x64 tiled EXR decodes successfully
- [x] Audit PoC: malicious tiled EXR with `tile_coordinates[0] = -1` fails cleanly under UBSan (no crash)

Made with [Cursor](https://cursor.com)

PR #10234, opened 2026-07-22
tinyexr: update to 3.2.0

poweifeng:
To address the issue raised in #10111


z3moon:
would it be possible to trim unnecessary files or directories like tests?

bejado (approved):

z3moon (approved):

PR #10244, opened 2026-07-23
Support SINGLE_THREADED mode in JobSystem

pixelflinger:
Introduce a JobSystem::SINGLE_THREADED constant to allow explicitly configuring the job system in single-threaded mode. When this special value is passed as the userThreadCount, the job system guarantees zero worker threads are spawned and the adoptableThreadsCount is forced to 1, ensuring all jobs safely execute on the calling thread.

Additional changes:
- Change constructor parameters from size_t to uint32_t.
- Replace slog.w with standard LOG(WARNING).
- Fix variable shadowing in setThreadName().
- Improve const correctness for internal method parameters.
- Simplify getStateToStealFrom logic.

### DIFF
--- a/libs/utils/include/utils/JobSystem.h
+++ b/libs/utils/include/utils/JobSystem.h
@@ -97,7 +97,19 @@ public:
     static_assert(sizeof(Job) == 64);
 #endif
 
-    explicit JobSystem(size_t threadCount = 0, size_t adoptableThreadsCount = 1) noexcept;
+    /**
+     * Special value for userThreadCount to configure the JobSystem in single-threaded mode.
+     */
+    static constexpr uint32_t SINGLE_THREADED = std::numeric_limits<uint32_t>::max();
+
+    /**
+     * Create a JobSystem and initializes its thread pool. The pool can have zero threads if SINGLE_THREADED is used
+     * as the userThreadCount, in that case adoptableThreadsCount is forced to 1.
+     *
+     * @param userThreadCount number of threads in the JobSystem thread pool.
+     * @param adoptableThreadsCount number of calling threads outside the thread pool.
+     */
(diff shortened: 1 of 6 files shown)


poweifeng (approved):

poweifeng (approved):

libs/utils/include/utils/JobSystem.h
```diff
@@ -97,7 +97,19 @@ class JobSystem {
...
 #endif
 
-    explicit JobSystem(size_t threadCount = 0, size_t adoptableThreadsCount = 1) noexcept;
+    /**
+     * Special value for userThreadCount to configure the JobSystem in single-threaded mode.
+     */
+    static constexpr uint32_t SINGLE_THREADED = std::numeric_limits<uint32_t>::max();
+
+    /**
+     * Create a JobSystem and initializes its thread pool. The pool can have zero threads if SINGLE_THREADED is used
```
z3moon, line 106: Can we ad comment on what the default value, which is zero, of userThreadCount means? Because I got an impression that both parameters can have JobSystem have zero threads.

libs/utils/src/JobSystem.cpp
```diff
@@ -178,33 +179,40 @@ void JobSystem::setThreadAffinityById(size_t id) noexcept {
...
-        // default value, system dependant
-        unsigned int hwThreads = std::thread::hardware_concurrency();
-        if (UTILS_HAS_HYPER_THREADING) {
-            // For now we avoid using HT, this simplifies profiling.
-            // TODO: figure-out what to do with Hyper-threading
-            // since we assumed HT, always round-up to an even number of cores (to play it safe)
-            hwThreads = (hwThreads + 1) / 2;
+
+    if (userThreadCount != SINGLE_THREADED) {
+        if (threadPoolCount == 0) {
```
z3moon, line 192: threadPoolCount is same as userThreadCount. How about consolidating them into one as it makes confusions?

PR #10247, opened 2026-07-27
Fix link error on Android

z3moon:
The current source code raises an "undefined symbol" link error for ANativeWindow_getBuffersDataSpace when minSdk is set to API level 26.

ANativeWindow_getBuffersDataSpace was introduced in Android 9 (API level 28). However, `__builtin_available` does not necessarily hide the ANativeWindow_getBuffersDataSpace call when compiling with API levels < 28.

To fix this, we dynamically retrieve the symbol.


pixelflinger (changes requested):
This is wrong, the call is protected by `__builtin_available(android 28, *)` and the symbols should be weak so there shouldn't be any link error. I think the real problem is that their build system is missing `__ANDROID_UNAVAILABLE_SYMBOLS_ARE_WEAK__`.

z3moon:
Yeah, you were right. It was their build env issue. Let me close this.

PR #10248, opened 2026-07-28
engine: support up to 4 additional shadowless directional lights

yuvaltassa:
Allow a scene to contain up to `CONFIG_MAX_EXTRA_DIRECTIONAL_LIGHTS` (4) directional lights in addition to the dominant one; the additional lights are evaluated without shadows. The dominant directional light (highest intensity) behaves exactly as before — it owns slot 0 of the light SoA, cascaded shadow maps, and the sun disc — and directional lights beyond the cap are ignored, preserving the previous dominant-light selection semantics as the degenerate case.

The feature is automatic and free for the common single-directional-light case: a new dynamic specialization constant (`RUNTIME_CONFIG_HAS_EXTRA_DIRECTIONAL_LIGHTS`, following the `RUNTIME_CONFIG_HAS_DYNAMIC_LIGHTING` pattern from #10095) is set per view when the scene contains more than one directional light, so the extra-lights loop compiles away entirely in programs specialized for single-light scenes.

Core changes:
- **Scene**: `FScene::prepare` collects all directional lights and partial-sorts them by intensity; the dominant one is stored at slot 0 as before, the next 4 are kept in `SceneCacheData` side arrays (direction + `LightManager` instance). They never enter the froxelizer or `ShadowMapManager`.
- **Uniforms**: new `extraLightDirection[4]`, `extraLightColorIntensity[4]` and `extraLightCount` fields in `PerViewUib`, carved out of the existing `reserved` slack (the struct remains exactly 2 KiB). Light channels are packed into `extraLightDirection[i].w`.
- **Shaders**: new `evaluateExtraDirectionalLights()` loop in `surface_light_directional.fs` under the existing `VARIANT_HAS_DIRECTIONAL_LIGHTING` — no new variants. The loop is gated by the dynamic specialization constant; extras skip shadowing and the sun disc.
- `DynamicSpecConstKey` gets an `EXTRA_DIRECTIONAL_LIGHTS` bit, set from `FView::hasExtraDirectionalLights()` when building render passes; `canSupportExtraDirectionalLights` limits it to lit surface variants with directional lighting, bounding the program-key growth.
- `MATERIAL_VERSION` bumped to 75 (new `PerViewUib` layout and dynamic specialization constant).
- `LightManager.h` documentation and release notes updated.

Tests:
- `FilamentTest.MultipleDirectionalLights`: dominant selection, extras ordering by intensity, direction transforms, truncation past the cap, and the single-light case.
- `RenderingTest.MultipleDirectionalLights`: headless render of a quad first lit by a single green directional light (verifying no red contribution), then with a weaker red directional light added (verifying red appears) — which also exercises the automatic program re-specialization.

### DIFF
--- a/filament/include/filament/LightManager.h
+++ b/filament/include/filament/LightManager.h
@@ -86,8 +86,11 @@ class FLightManager;
  * To create a directional light use Type.DIRECTIONAL or Type.SUN, both are similar, but the later
  * also draws a sun's disk in the sky and its reflection on glossy objects.
  *
- * @warning Currently, only a single directional light is supported. If several directional lights
- * are added to the scene, the dominant one will be used.
+ * Several directional lights can be added to the scene. The dominant directional light (the one
+ * with the highest intensity) is the only one that can cast shadows and draw a sun's disk; up to
+ * CONFIG_MAX_EXTRA_DIRECTIONAL_LIGHTS (4) additional directional lights are evaluated without
+ * shadows, and any further directional lights are ignored. Scenes with a single directional
+ * light don't pay any cost for this feature.
  *
  * @see Builder.direction(), Builder.sunAngularRadius()
  *
--- a/filament/include/filament/LightManager.h
+++ b/filament/include/filament/LightManager.h
@@ -86,11 +86,12 @@ class FLightManager;
  * To create a directional light use Type.DIRECTIONAL or Type.SUN, both are similar, but the later
  * also draws a sun's disk in the sky and its reflection on glossy objects.
  *
- * Several directional lights can be added to the scene. The dominant directional light (the one
- * with the highest intensity) is the only one that can cast shadows and draw a sun's disk; up to
- * CONFIG_MAX_EXTRA_DIRECTIONAL_LIGHTS (4) additional directional lights are evaluated without
- * shadows, and any further directional lights are ignored. Scenes with a single directional
- * light don't pay any cost for this feature.
+ * By default, only the dominant directional light (the one with the highest intensity) of a
+ * scene is evaluated. Several directional lights can be used by enabling
+ * Engine::Config::enableMultipleDirectionalLights: the dominant one is still the only one that
+ * can cast shadows and draw a sun's disk, and up to CONFIG_MAX_EXTRA_DIRECTIONAL_LIGHTS (4)
+ * additional directional lights are evaluated without shadows; any further directional lights
+ * are ignored. Scenes with a single directional light don't pay any cost for this feature.
  *
  * @see Builder.direction(), Builder.sunAngularRadius()
  *
--- a/filament/src/details/Scene.cpp
+++ b/filament/src/details/Scene.cpp
@@ -118,13 +118,13 @@ void FScene::prepare(JobSystem& js,
 
     FILAMENT_TRACING_NAME_BEGIN(FILAMENT_TRACING_CATEGORY_FILAMENT, "InstanceLoop");
 
-    // find the max intensity directional light index in our local array
-    float maxIntensity = 0.0f;
-    std::pair<LightManager::Instance, TransformManager::Instance> directionalLightInstances{};
+    // directional lights are handled separately from the punctual lights
+    LightInstanceContainer directionalInstances{
+            LightInstanceContainer::with_capacity(entities.size(), localArenaScope.getArena()) };
 
     /*
      * First compute the exact number of renderables and lights in the scene.
-     * Also find the main directional light.
+     * Also collect the directional lights.
      */
     entities.forEachSetBit([&](uint32_t id) {
         Entity const e = Entity::import(int32_t(id));
@@ -133,13 +133,9 @@ void FScene::prepare(JobSystem& js,
             auto li = lcm.getInstance(e);
             auto ri = rcm.getInstance(e);
             if (li) {
-                // we handle the directional light here because it'd prevent multithreading below
+                // we handle the directional lights here because it'd prevent multithreading below
                 if (UTILS_UNLIKELY(lcm.isDirectionalLight(li))) {
-                    // we don't store the directional lights, because we only have a single one
-                    if (lcm.getIntensity(li) >= maxIntensity) {
-                        maxIntensity = lcm.getIntensity(li);
-                        directionalLightInstances = { li, ti };
-                    }
+                    directionalInstances.emplace_back(li, ti);
                 } else {
                     lightInstances.emplace_back(li, ti);
                 }
@@ -149,6 +145,24 @@ void FScene::prepare(JobSystem& js,
             }
         }
     });
+
+    // Sort the directional lights by decreasing intensity. The most intense one is the
+    // dominant directional light (stored at index 0 of the LightSoa); it is the only one
+    // that casts shadows and can be a sun. Up to CONFIG_MAX_EXTRA_DIRECTIONAL_LIGHTS
+    // additional directional lights are evaluated without shadows; the rest are ignored.
+    if (UTILS_UNLIKELY(directionalInstances.size() > 1)) {
+        size_t const keep = std::min<size_t>(directionalInstances.size(),
+                DIRECTIONAL_LIGHTS_COUNT + CONFIG_MAX_EXTRA_DIRECTIONAL_LIGHTS);
+        std::partial_sort(directionalInstances.begin(),
+                directionalInstances.begin() + keep, directionalInstances.end(),
+                [&lcm](auto const& lhs, auto const& rhs) {
+                    return lcm.getIntensity(lhs.first) > lcm.getIntensity(rhs.first);
+                });
+    }
+    std::pair<LightManager::Instance, TransformManager::Instance> const directionalLightInstances =
+            directionalInstances.empty() ?
+                    std::pair<LightManager::Instance, TransformManager::Instance>{} :
+                    directionalInstances[0];
 
     FILAMENT_TRACING_NAME_END(FILAMENT_TRACING_CATEGORY_FILAMENT);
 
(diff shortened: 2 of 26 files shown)


filament/src/details/Scene.cpp
```diff
@@ -150,6 +146,24 @@ void FScene::prepare(JobSystem& js,
         }
     });
 
+    // Sort the directional lights by decreasing intensity. The most intense one is the
+    // dominant directional light (stored at index 0 of the LightSoa); it is the only one
+    // that casts shadows and can be a sun. Up to CONFIG_MAX_EXTRA_DIRECTIONAL_LIGHTS
+    // additional directional lights are evaluated without shadows; the rest are ignored.
```
pixelflinger, line 152: I think this will do for now, but I'm starting to think we should have a `setDominantLight()` on scene or view, or something.

pixelflinger (commented):
LGTM. I'd like @show50726 to review the spec constant stuff.

shaders/src/surface_light_directional.fs
```diff
@@ -95,3 +95,51 @@ void evaluateDirectionalLight(const MaterialInputs material,
...
+        const PixelParams pixel, inout vec3 color) {
+    int channels = object_uniforms_flagsChannels & 0xFF;
+    // RUNTIME_CONFIG_HAS_EXTRA_DIRECTIONAL_LIGHTS is a dynamic specialization constant,
+    // set automatically when the scene contains more than one directional light. It is
+    // false by default, in which case this function compiles away entirely.
+    if (!RUNTIME_CONFIG_HAS_EXTRA_DIRECTIONAL_LIGHTS) {
+        return;
+    }
+    for (int i = 0; i < CONFIG_MAX_EXTRA_DIRECTIONAL_LIGHTS; i++) {
+        if (i < frameUniforms.extraLightCount) {
```
show50726, line 121: What's the reason of not breaking the loop earlier?

show50726:
We might want to update the material precompile api as well:
https://github.com/google/filament/blob/50aa36845c89201af3e4ef45b72ca2d09d61be60/filament/src/details/Engine.cpp#L1837-L1848

You will have to update `getMaterialCompileDynamicSpecConstKey` to set the extra light according to the `view` parameter

filament/src/details/Scene.cpp
yuvaltassa: Acknowledged.

shaders/src/surface_light_directional.fs
yuvaltassa: Feature level 0, mostly. This loop also compiles and, with several directional lights in the scene, runs at ES2, where the constant is emulated via the injected SPIRV_CROSS_CONSTANT_ID define rather than a real spec constant. Strict GLSL ES 1.00 front-ends only guarantee constant-bounded loops, so I didn't want to rely on a data-dependent break being accepted there. Happy to change it if you'd rather special-case FL0.

yuvaltassa:
@show50726 `getMaterialCompileDynamicSpecConstKey` is updated in this PR. Your link points at the file on main, so the hunk may have been easy to miss. Let me know if you had something beyond that in mind.

```c
baseKey.setExtraDirectionalLights(isMaterialLit && view->hasExtraDirectionalLights());
```

yuvaltassa:
Rebased onto main to resolve a release-notes conflict from the 1.74.1 branch. While at it, `canSupportExtraDirectionalLights` now excludes depth/SSR variants before reading the directional-lighting bit, mirroring the dynamic-lighting guard after #10241.

filament/src/ds/ColorPassDescriptorSet.cpp
```diff
@@ -347,6 +347,24 @@ void ColorPassDescriptorSet::prepareDirectionalLight(FEngine& engine,
...
+    auto& s = mUniforms.edit();
+    for (size_t i = 0; i < count; i++) {
+        LightManagerInstance const li = instances[i];
+        // the channel mask fits in 8 bits, so the float conversion is exact
+        s.extraLightDirection[i] = float4{
+                -sceneSpaceDirections[i], float(lcm.getLightChannels(li)) };
+        s.extraLightColorIntensity[i] = float4{
+                lcm.getColor(li), lcm.getIntensity(li) * exposure };
+    }
+    s.extraLightCount = int32_t(count);
```
show50726, line 365: nit: use `static_cast<int32_t>`

show50726 (approved):
The spec const part looks good to me!

yuvaltassa:
Rebased.

filament/include/filament/LightManager.h
```diff
@@ -86,8 +86,11 @@ class FLightManager;
  * To create a directional light use Type.DIRECTIONAL or Type.SUN, both are similar, but the later
  * also draws a sun's disk in the sky and its reflection on glossy objects.
  *
- * @warning Currently, only a single directional light is supported. If several directional lights
- * are added to the scene, the dominant one will be used.
+ * Several directional lights can be added to the scene. The dominant directional light (the one
```
bejado: Can you please update the associated doc in LightManager.java as well?

bejado (approved):

shaders/src/surface_light_directional.fs
pixelflinger: generally I prefer to special-case FL0 (which is a fallback) if there is a chance the code will be more efficient at other FL (I think it's the case here?).

filament/src/ds/ColorPassDescriptorSet.cpp
pixelflinger: @show50726 technically our style guide is to use constructor syntax for basic type casting, so `int(foo)` is correct.

filament/src/details/Scene.h
```diff
@@ -185,6 +188,14 @@ class FScene : public Scene {
...
         RenderableSoa renderableData;
         LightSoa lightData;
+        // Directional lights in addition to the dominant one (which is stored at index 0 of
+        // lightData). These are evaluated without shadows, so we only need their direction
+        // and LightManager instance.
+        std::array<math::float3, CONFIG_MAX_EXTRA_DIRECTIONAL_LIGHTS>
+                extraDirectionalLightDirections{};
+        std::array<FLightManager::Instance, CONFIG_MAX_EXTRA_DIRECTIONAL_LIGHTS>
+                extraDirectionalLightInstances{};
+        size_t extraDirectionalLightCount = 0;
```
pixelflinger:
```suggestion
        uint8_t extraDirectionalLightCount = 0;
```

filament/src/details/View.cpp
```diff
@@ -802,6 +811,9 @@ void FView::prepare(FEngine& engine, DriverApi& driver, RootArenaScope& rootAren
         // now we know if we have dynamic lighting (i.e.: dynamic lights are visible)
         mHasDynamicLighting = lightData.size() > FScene::DIRECTIONAL_LIGHTS_COUNT;
 
+        // we also know if we have extra directional lights
+        mHasExtraDirectionalLights = mSceneCache->extraDirectionalLightCount > 0;
```
pixelflinger, line 815: That worries me a little because technically this breaks backward compatibility. i.e. a scene could have several directional lights and the user might expect only one to be active, as it was documented.

Unfortunately, I think we need an "enableMultipleDirectionalLights", probably in `Engine::Config`, I don't think it's necessary to do this per-view (and if so, the java/javascript bindings need to be updated).

I'm happy to hear about other's opinion on this?

pixelflinger (changes requested):

shaders/src/surface_light_directional.fs
yuvaltassa: Done. The constant-bounded loop with the guard is now only used at FL0 where GLSL ES 1.00 requires it.

filament/src/details/Scene.h
yuvaltassa: Done.

filament/include/filament/LightManager.h
yuvaltassa: Done.

filament/src/details/View.cpp
yuvaltassa: Done. Multiple directional lights are now opt-in via `Engine::Config::enableMultipleDirectionalLights`, defaulting to `false`, so an existing scene with several directional lights keeps using only the dominant one. Plumbed through the Java and JavaScript bindings, and a rendering test asserting that a second directional light has no effect with the default configuration.

If you want my opinion: Hyrum's Law notwithstanding, if someone put N>1 directional lights maybe they wanted N>1 directional lights? That said, I also respect the never break it edict. I could argue both sides 🤷‍♂️

yuvaltassa:
Addressed the review comments in the latest commit. Mainly: multiple directional lights are now opt-in via `Engine::Config::enableMultipleDirectionalLights` (default `false`) so existing scenes are unaffected.

show50726:
It seems many unrelated files are uploaded?

yuvaltassa:
Apologies @show50726, fixed now.

pixelflinger (approved):

show50726 (approved):

poweifeng:
Please fix this: https://github.com/google/filament/actions/runs/31153234968/job/92955472247#step:9:793

I think you should probably use the no-op driver/backend for the test.  There should be examples of doing so in other tests.

yuvaltassa:
@poweifeng 

Fixed. CI already excludes `RenderingTest.*`, but the filter matches exact suite names so my derived fixture slipped through, widened to `-*RenderingTest.*`. NOOP won't work for this one... it reads back pixels to check the second light contributes. The CPU-side logic is covered under NOOP by `FilamentTest.MultipleDirectionalLights`.

filament/test/filament_rendering_test.cpp
```diff
@@ -119,6 +137,178 @@ TEST_F(RenderingTest, ClearRed) {
     EXPECT_TRUE(callbackCalled);
 }
 
+// Holds a scene with a full-viewport lit quad, a dominant green directional light and a
+// weaker red one. Shared by the extra-directional-lights tests below.
+class LitQuadScene {
```
poweifeng: actually, given the complexity of this setup, would you mind not adding it as a test here but as part of `test/renderdiff/tests/presubmit.json`?  

Or probably better if you filed an issue to add a test for this on the github issue tracker

yuvaltassa: Removed. I'll file an issue for the renderdiff coverage.

poweifeng (approved):

PR #10250, opened 2026-07-28
Re-enable the backend_test build on iOS

tim4724:
## What

The iOS backend_test build has been disabled for a while with a "breakage wrt glslang update" TODO. This re-enables it (opt-in via `-DINSTALL_BACKEND_TEST=ON`, so default builds are unchanged) and fixes the actual breakages, which turned out to be five small things:

1. `imageio` in the test link list is a host-only library — its link entry is now added only under `IS_HOST_PLATFORM`, the same gate imageio itself is built behind. (Not `if (TARGET imageio)`: imageio is added after `filament/backend` in the root CMakeLists, so the target never exists at this point and the guard would silently drop imageio on desktop builds too.)
2. The test target lacked the backend's `-fno-objc-exceptions`: ObjC++ TUs defined `__EXCEPTIONS` while C++ exceptions are off, breaking robin-map.
3. `EXPECT_DEATH` without a `GTEST_HAS_DEATH_TEST` guard (death tests are unavailable on iOS-family platforms) — `test_Platform.cpp` and `test_MetalBlitter.mm`.
4. `test_Handles.cpp` hardcodes the `HandleAllocatorGL` template sizes, which don't exist in any build without the OpenGL backend (latent for every GL-less configuration, not just iOS).
5. `test_MetalBlitter.mm` simulator compatibility: depth textures use the default (shared) storage mode, which Metal-in-Simulator rejects for depth formats (hard validation abort) — fixed with `MTLStorageModePrivate` where contents are never CPU-accessed, and a `TARGET_OS_SIMULATOR` skip for the CPU fill/verify depth-blit test, which needs shared depth textures by construction. Both depth tests run and pass on real devices.

A third commit records the missing iOS-family golden hashes (they were placeholder `0`/`1` values, so those comparisons could never pass). Recorded on the Apple silicon simulator at the harness's canonical 512×512 size and cross-checked on an Apple TV 4K (A10X): every value except BasicAsyncFlow is bit-identical on both. **Every recorded hash is reproducible from the desktop goldens already in this repo**: it equals `utils::hash::murmur3` (seed 0) over the decoded RGBA pixels of the corresponding `expected_images/<name>.png`, i.e. the exact computation `RenderTargetDump::hash()` performs on the framebuffer readback (`LoadedPng::hash()` in the harness computes the same thing). To verify without building anything:

```python
# python3 hash_goldens.py path/to/filament/backend/test/expected_images
import sys, struct
from PIL import Image

def murmur3(words, seed=0):
    M = 0xFFFFFFFF
    h = seed
    for k in words:
        k = (k * 0xcc9e2d51) & M
        k = ((k << 15) | (k >> 17)) & M
        k = (k * 0x1b873593) & M
        h ^= k
        h = ((h << 13) | (h >> 19)) & M
        h = (h * 5 + 0xe6546b64) & M
    h ^= len(words)
    h ^= h >> 16
    h = (h * 0x85ebca6b) & M
    h ^= h >> 13
    h = (h * 0xc2b2ae35) & M
    return (h ^ (h >> 16))

for name in ["WriteAndRender", "MapWithOffset", "CopyWithOffset",
        "MapAndCopyWithOffsets", "MultipleCopies", "UpdatePartial_before",
        "UpdatePartial_after", "MsaaSwapChain", "BasicAsyncFlow", "scissor"]:
    rgba = Image.open(f"{sys.argv[1]}/{name}.png").convert("RGBA").tobytes()
    words = list(struct.unpack(f"<{len(rgba)//4}I", rgba))
    print(name, murmur3(words))
```

This prints exactly the ten values this PR writes into the tests. It also confirms the simulator renders match the blessed goldens byte for byte (verified separately by diffing dumped simulator pixels against the PNGs — zero channel deltas). The only value not derivable this way is TestTemplate, which has no desktop golden; it is the simulator-measured value. The pre-existing scissor hash matched no current rasterization (its render, however, matches `scissor.png` exactly) and is updated to the value all configurations agree on. Hashes are only consumed on iOS-family builds; desktop keeps comparing PNG goldens.

## Validation

Suite run via a minimal harness app (square 512×512 drawable, `initTests`/`runTests` per `PlatformRunner.h`), current main:

- iPhone simulator: 84 tests / 18 suites — **82 pass, 2 skipped (the simulator-incompatible depth blits), 0 failures**, 0 crashes, 0 Metal validation errors.
- Real A-series hardware (via tvOS, same Metal backend): 84 ran, 80 pass, 0 crashes — the 4 remaining diffs (BasicAsyncFlow, FeedbackLoops, ColorMinify, UpdateImageSRGB) are per-GPU rasterization drift between Apple silicon and A-series GPUs in blit-filtering/sRGB tests; exact-hash comparison cannot capture both GPU families with a single value. Switching iOS to the desktop path's tolerance-based comparison would be the durable fix; out of scope here.
- macOS (`backend_test_mac -a metal`, current main + this change): **87/87 pass** — no regressions on the desktop path (expected hashes are not consumed on desktop; it compares PNG goldens).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/filament/backend/CMakeLists.txt
+++ b/filament/backend/CMakeLists.txt
@@ -637,10 +637,16 @@ if (FILAMENT_BUILD_TESTING)
             absl::str_format
             backend
             gtest
-            imageio
             filamat
             SPIRV
             spirv-cross-glsl)
+        # imageio drives the golden-image comparison and is only built for host
+        # platforms (see the IS_HOST_PLATFORM block in the root CMakeLists).
+        # NOTE: this must not be `if (TARGET imageio)` — imageio is added after
+        # filament/backend, so the target does not exist yet at this point.
+        if (IS_HOST_PLATFORM)
+            list(APPEND BACKEND_TEST_LIBS imageio)
+        endif()
         # Create input/output directories for test result images.
         file(MAKE_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/images/actual_images)
         file(MAKE_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/images/diff_images)
@@ -648,13 +654,24 @@ if (FILAMENT_BUILD_TESTING)
     endif()
 
     # TODO: Disabling IOS test due to breakage wrt glslang update
-    if (APPLE AND NOT IOS)
-        # TODO: we should expand this test to Linux and other platforms.
-        list(APPEND BACKEND_TEST_SRC
-            test/test_RenderExternalImage.cpp)
+    # (tvOS builds go through the same IOS toolchain but are re-enabled: filamat,
+    # glslang and spirv-cross all compile for appletv* targets.)
+    if (APPLE)
+        if (NOT IOS)
+            # Uses CVPixelBuffer-backed external images; host-only.
+            # TODO: we should expand this test to Linux and other platforms.
+            list(APPEND BACKEND_TEST_SRC
+                test/test_RenderExternalImage.cpp)
+        endif()
         add_library(backend_test STATIC ${BACKEND_TEST_SRC})
         target_link_libraries(backend_test PUBLIC ${BACKEND_TEST_LIBS})
         target_compile_options(backend_test PRIVATE "-fobjc-arc")
+        if (IOS)
+            # Match the backend target: without this, ObjC++ files define __EXCEPTIONS
+            # (via ObjC exceptions) while C++ exceptions are disabled, which breaks
+            # headers like robin-map that key their throw usage off __EXCEPTIONS.
+            target_compile_options(backend_test PRIVATE "-fno-objc-exceptions")
+        endif()
 
         set(BACKEND_TEST_DEPS
                 OSDependent
--- a/filament/backend/CMakeLists.txt
+++ b/filament/backend/CMakeLists.txt
@@ -5,6 +5,19 @@ set(TARGET backend)
 set(PUBLIC_HDR_DIR include)
 set(GENERATION_ROOT ${CMAKE_CURRENT_BINARY_DIR})
 
+option(INSTALL_BACKEND_TEST "Install the backend test library so it can be consumed on iOS" OFF)
+
+# On iOS-family platforms backend_test is opt-in (INSTALL_BACKEND_TEST): it is only useful
+# embedded in a runner app, and ImageExpectations needs std::filesystem, which requires a 13.0+
+# deployment target (the default iOS device floor is 11.0). backend_test links filamat, which
+# already requires 13.0, so this raises no floor that wasn't already implied. Gated on
+# INSTALL_BACKEND_TEST so that default builds keep the 11.0 floor.
+# Like filamat and glslang, this is included before any target is declared: the deployment
+# target is directory-scoped, so in test-runner builds it applies to `backend` as well.
+if (IOS AND INSTALL_BACKEND_TEST)
+    include(${EXTERNAL}/clang/usage_ios_13.cmake)
+endif()
+
 # ==================================================================================================
 # Sources and headers
 # ==================================================================================================
@@ -581,8 +594,6 @@ endif()
 # Test
 # ==================================================================================================
 if (FILAMENT_BUILD_TESTING)
-    option(INSTALL_BACKEND_TEST "Install the backend test library so it can be consumed on iOS" OFF)
-
     if (APPLE OR LINUX)
         set(BACKEND_TEST_SRC
             test/BackendTest.cpp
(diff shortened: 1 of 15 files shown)


tim4724:
@bejado @poweifeng Friendly ping. Is there anything I can do to make this easier to review? A note on how it connects to recent work: it makes the MetalBlitter depth tests from #10226 run on the iOS family (simulator included), and it's the same effort as #10235 applied to the iOS `backend_test` build. Everything is opt-in behind `INSTALL_BACKEND_TEST`, so default builds are unchanged. CI hasn't run yet, since it's waiting on a first-time-contributor "Approve and run".

bejado:
> @bejado @poweifeng Friendly ping. Is there anything I can do to make this easier to review? A note on how it connects to recent work: it makes the MetalBlitter depth tests from #10226 run on the iOS family (simulator included), and it's the same effort as #10235 applied to the iOS `backend_test` build. Everything is opt-in behind `INSTALL_BACKEND_TEST`, so default builds are unchanged. CI hasn't run yet, since it's waiting on a first-time-contributor "Approve and run".

Sorry for the delay, I will take a look today!

filament/backend/CMakeLists.txt
```diff
@@ -633,24 +633,42 @@ if (FILAMENT_BUILD_TESTING)
...
     endif()
 
-    # TODO: Disabling IOS test due to breakage wrt glslang update
-    if (APPLE AND NOT IOS)
-        # TODO: we should expand this test to Linux and other platforms.
-        list(APPEND BACKEND_TEST_SRC
-            test/test_RenderExternalImage.cpp)
+    # On iOS-family platforms the test library is opt-in (INSTALL_BACKEND_TEST): it is only
+    # useful embedded in a runner app, and ImageExpectations needs
+    # std::filesystem, which requires a 13.0+ deployment target (the default
```
bejado: How do we force a 13.0 deployment target for real iOS devices? I think we might need a 

```
include(${EXTERNAL}/clang/usage_ios_13.cmake)
```

filament/backend/test/test_Handles.cpp
```diff
@@ -27,13 +27,23 @@ using namespace filament::backend;
...
+// accounts for FILAMENT_DEBUG_MUTEX/UTILS_DEBUG_MUTEX sizing.
 // NOTE: actual count may be lower due to alignment requirements
+#if defined(FILAMENT_SUPPORTS_OPENGL)
+#define HandleAllocatorTest  HandleAllocatorGL
 #if defined(FILAMENT_DEBUG_MUTEX) || defined(UTILS_DEBUG_MUTEX)
 constexpr size_t const POOL_HANDLE_COUNT = POOL_SIZE_BYTES / (32 + 96 + 312);
-#define HandleAllocatorTest  HandleAllocator<32,  96, 312>
 #else
 constexpr size_t const POOL_HANDLE_COUNT = POOL_SIZE_BYTES / (32 + 96 + 184); // 31775
-#define HandleAllocatorTest  HandleAllocator<32,  96, 184>    // ~4520 / pool / MiB
```
bejado, line 36: I don't like that we have to duplicate the bucket sizes here. Inside `HandleAllocator`, let's add a 

```
static constexpr size_t bucketSizesSum = P0 + P1 + P2;
```

then we can reference it here

```
constexpr size_t const POOL_HANDLE_COUNT = POOL_SIZE_BYTES / HandleAllocatorTest::bucketSizesSum
```

bejado (changes requested):

filament/backend/CMakeLists.txt
tim4724: Done — added `include(${EXTERNAL}/clang/usage_ios_13.cmake)`, gated on `IOS AND INSTALL_BACKEND_TEST` so default builds keep the 11.0 floor. One note: since `CMAKE_OSX_DEPLOYMENT_TARGET` is directory-scoped and `backend` lives in the same CMakeLists, test-runner builds raise the backend library itself to 13.0 as well — which was already implied, since backend_test links filamat (13.0-only). Verified on a fresh iphoneos configure: `backend_test` objects now get `-miphoneos-version-min=13.0` while targets outside the directory (e.g. utils) stay at 11.0.

filament/backend/test/test_Handles.cpp
tim4724: Done — added `static constexpr size_t bucketSizesSum = P0 + P1 + P2;` to `HandleAllocator` and the test now uses `HandleAllocatorTest::bucketSizesSum`. This also let the `FILAMENT_DEBUG_MUTEX`/`UTILS_DEBUG_MUTEX` special-casing collapse, since the `HandleAllocatorGL` macro already accounts for it.

tim4724:
Rebased onto main. Both comments addressed:

- **Deployment target** — added `include(${EXTERNAL}/clang/usage_ios_13.cmake)`, gated on
  `IOS AND INSTALL_BACKEND_TEST`, at the top of the file the way filamat and glslang do it. On an
  `iphoneos` configure `backend_test` and `backend` get 13.0; targets outside that directory stay
  at 11.0.
- **Bucket sizes** — `HandleAllocator` now exposes `bucketSizesSum`, and the test picks the
  allocator matching whichever backend is compiled in, so nothing is duplicated.

Two things came out of getting the suite to actually run, which is why there are extra commits:

- The runner app in `ios/samples/backend-test` didn't compile (`initTests()` gained an
  `OperatingSystem` parameter), and it sized the view from the device's screen instead of the
  fixed size the desktop runners use — 26 of 84 tests were failing on screen size alone. It now
  mirrors `mac_runner.mm`.
- `test_JobQueue.cpp` landed after I branched and uses `EXPECT_DEATH` unguarded, which doesn't
  compile on iOS. Same fix this PR already makes in two other files — happy to split it out if
  you'd rather it didn't ride along.

`BasicAsyncFlow` is now skipped on device hardware: A-series rasterizes that scene differently
(2793888331 on an A10X), so there's no single hash to compare against. The recorded value is the
simulator's.

iPhone 17 Pro simulator, Metal: 92 tests, 0 failures.

Still waiting on a first-time-contributor "Approve and run" — note CI doesn't build `backend_test`
on any platform, so it won't exercise these tests, but it does cover the CMake changes.

bejado (approved):
LGTM. Let's wait for the CI to finish, then I'll merge. Thank you!

pixelflinger:
looks like rebase conflicts

tim4724:
Rebased onto main

PR #10253, opened 2026-07-29
gltfio: fix createInstanceAsset with num_instance=0

poweifeng:
Calling createInstanceAsset() with num_instance=0 will skip the path to create material instances, which will then skip populating FFilamentAsset::mTextures. This means that when ResourceLoader creates filament textures, it wouldn't have the right flags.

This change prefills mTextures based on the materials and textures in the gltf (adding to FAssetLoader::createRootAsset()).

### DIFF
--- a/libs/gltfio/src/AssetLoader.cpp
+++ b/libs/gltfio/src/AssetLoader.cpp
@@ -68,6 +68,8 @@ using namespace utils;
 
 namespace filament::gltfio {
 
+namespace {
+
 using SceneMask = NodeManager::SceneMask;
 
 static const auto FREE_CALLBACK = [](void* mem, size_t, void*) { free(mem); };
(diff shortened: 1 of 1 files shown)


pixelflinger:
I don't know this code really well, but is there a risk that it slows down initialization significantly? Aren't texture loaded asynchronously, does this change that?

poweifeng:
> I don't know this code really well, but is there a risk that it slows down initialization significantly? Aren't texture loaded asynchronously, does this change that?

this is just doing a pass over the metadata of the textures associated with the materials. no actual loading the textures themselves, so overhead should be negligible.

libs/gltfio/src/AssetLoader.cpp
```diff
@@ -558,6 +558,97 @@ FFilamentAsset* FAssetLoader::createRootAsset(const cgltf_data* srcAsset) {
...
+        UvMap dummyUvmap{};
+        cgltf_texture_view baseColorTexture;
+        cgltf_texture_view metallicRoughnessTexture;
+        MaterialKey matkey = getMaterialKey(srcAsset, inputMat, &dummyUvmap, false,
+                &baseColorTexture, &metallicRoughnessTexture);
+
+        const TextureProvider::TextureFlags sRGB = TextureProvider::TextureFlags::sRGB;
+        const TextureProvider::TextureFlags LINEAR = TextureProvider::TextureFlags::NONE;
+
+        // This section has the exact same checks and flags as createMaterialInstance().
```
z3moon: can we make this section as a helper function?

poweifeng: done

z3moon (approved):

PR #10254, opened 2026-07-29
VK: Always discard transient depth/stencil attachment

rafadevai:
When a renderpass is using a depth/stencil attachment marked as transient, always set the renderpass to discard the result at the end.

BUGS = [540411825]

### DIFF
--- a/filament/backend/src/vulkan/VulkanDriver.cpp
+++ b/filament/backend/src/vulkan/VulkanDriver.cpp
@@ -2359,6 +2359,12 @@ void VulkanDriver::beginRenderPass(Handle<HwRenderTarget> rth, const RenderPassP
             discardEndVal &= ~TargetBufferFlags::STENCIL;
             clearVal &= ~TargetBufferFlags::STENCIL;
         }
+
+        VulkanAttachment& depthStencil = rt->getDepthStencil();
+        if (depthStencil.texture->isTransientAttachment()) {
+            discardEndVal |= TargetBufferFlags::DEPTH_AND_STENCIL;
+        }
+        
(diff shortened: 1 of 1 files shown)


jorgeag-google (approved):
LGTM

filament/backend/src/vulkan/VulkanDriver.cpp
```diff
@@ -2359,6 +2359,12 @@ void VulkanDriver::beginRenderPass(Handle<HwRenderTarget> rth, const RenderPassP
             discardEndVal &= ~TargetBufferFlags::STENCIL;
             clearVal &= ~TargetBufferFlags::STENCIL;
         }
+
+        VulkanAttachment& depthStencil = rt->getDepthStencil();
+        if (depthStencil.texture->isTransientAttachment()) {
+            discardEndVal |= TargetBufferFlags::DEPTH_AND_STENCIL;
```
poweifeng, line 2365: I do think there's an issue with how we're using `VK_ATTACHMENT_STORE_OP_DONT_CARE` right now; it requires [write access](https://docs.vulkan.org/refpages/latest/refpages/source/VkAttachmentStoreOp.html#_description).  

Maybe this needs to be a `NONE`?  But this is a larger refactor than what you're doing here.

poweifeng (approved):

filament/backend/src/vulkan/VulkanDriver.cpp
rafadevai: This is just being specific that in the cases of transient, we basically ignore what the frontend tell us and it will always  set it as DONT_CARE in the case of transient depth/stencil.

For the NONE, that requires an extension or 1.3 but I think that still doesnt change this logic that we just set the content to be discarded at the end of the pass.

Issue #9979, opened 2026-05-12
FR: constexpr std::string_view to_string methods for all public enums, for debug-printing


mhansen:
Filament has many public enums. It's natural to want to print these out for debuggers, printf statements, and heck, even for AI agents.

Today, some enums have to_string methods. And some don't. The ones that don't, end up getting duplicate to_string methods in libraries that call Filament. These duplicates get out of sync with the enum definitions over time. That's bad. We'd be better off to put the to_string function inside filament itself, next to the enum.

There are some code size increase costs, but I hope that compilers should be able to find and strip dead code and constants.

On https://github.com/google/filament/pull/9975#issuecomment-4422286831, which proposes one such to_string method, @romainguy says 

> It would make more sense to do this consistently for all public enums

I agree, let's file this ticket to track.

PR #10255, opened 2026-07-29
backend: add to_string() for FeatureLevel and ShaderLanguage

husseinhafez1:
Part of #9979 

Adds `to_string()` for FeatureLevel and ShaderLanguage. kept the old `shaderLanguageToString()` with the new `to_string(ShaderLanguage)` for now so I don't break anything.


filament/backend/include/backend/DriverEnums.h
```diff
@@ -206,6 +220,26 @@ enum class ShaderLanguage {
     WGSL = 5,
 };
 
+constexpr std::string_view to_string(ShaderLanguage shaderLanguage) noexcept {
```
z3moon, line 223: I think we should rewrite `shaderLanguageToString` in terms of `to_string`, and adding a comment that it's deprecated.

filament/backend/include/backend/DriverEnums.h
```diff
@@ -155,6 +155,20 @@ enum class FeatureLevel : uint8_t {
     FEATURE_LEVEL_3       //!< OpenGL ES 3.1 features + 31 textures units + cubemap arrays
 };
 
+constexpr std::string_view to_string(FeatureLevel level) noexcept {
```
z3moon, line 158: nit: `[[nodiscard]] constexpr std::string_view ...`

filament/backend/include/backend/DriverEnums.h
husseinhafez1: Done.

filament/backend/include/backend/DriverEnums.h
husseinhafez1: Done.

z3moon (approved):
LGTM, but please make sure you get an approval from @pixelflinger

filament/backend/include/backend/DriverEnums.h
```diff
@@ -155,6 +155,20 @@ enum class FeatureLevel : uint8_t {
     FEATURE_LEVEL_3       //!< OpenGL ES 3.1 features + 31 textures units + cubemap arrays
 };
 
+[[nodiscard]] constexpr std::string_view to_string(FeatureLevel level) noexcept {
```
pixelflinger:
```suggestion
constexpr std::string_view to_string(FeatureLevel level) noexcept {
```

filament/backend/include/backend/DriverEnums.h
```diff
@@ -206,7 +220,7 @@ enum class ShaderLanguage {
     WGSL = 5,
 };
 
-constexpr const char* shaderLanguageToString(ShaderLanguage shaderLanguage) noexcept {
+[[nodiscard]] constexpr std::string_view to_string(ShaderLanguage shaderLanguage) noexcept {
```
pixelflinger:
```suggestion
constexpr std::string_view to_string(ShaderLanguage shaderLanguage) noexcept {
```

pixelflinger (changes requested):

filament/backend/include/backend/DriverEnums.h
pixelflinger: we don't need `[[nodiscard]]` since there is no risk of memory leak (it's not like malloc for example).

z3moon: While `[[nodiscard]]` is often used to prevent resource leaks, using `[[nodiscard]]` is generally recommended for conversion/getter functions by modern c++ guidlines to catch typos or logic bugs.

pixelflinger: mh ok. But we've not done that anywhere right now. I think we should update the guidelines and then do a pass that updates everything then.

pixelflinger (approved):

PR #10256, opened 2026-07-29
webgpu: use tint::wgsl::writer for spirv->wgsl

poweifeng:
This bypasses using reader to produce an IR, which is then fed to the writer for wgsl anyways.


poweifeng:
bypassing review because it's needed for next release

PR #10261, opened 2026-07-31
vk: use VK_GOOGLE_display_timing to set present time

poweifeng:
- Implement Driver::setPresentationTime()
 - Some small simplification in VulkanPlatform.cpp


filament/backend/src/vulkan/platform/VulkanPlatform.cpp
```diff
@@ -594,6 +595,8 @@ Driver* VulkanPlatform::createDriver(void* sharedContext,
 
         instExts.merge(getRequiredInstanceExtensions());
     }
+    mImpl->mContext.mDebugUtilsSupported = setContains(instExts, VK_EXT_DEBUG_UTILS_EXTENSION_NAME);
```
rafadevai: why move this? Also the logic is not the same as before

I think its just better to keep it were it was all in one place

poweifeng: the flow of setting of context variables is getting spread out, so I'm just trying to do it right after when we know the state.  So for instance extensions, it's dependent on `instExts`, and `instExts` is immutable at this point. That's why I moved it here.  Also it reduces `queryAndSetDeviceFeatures`'s parameters by one.

rafadevai: Sadly this gets more complicated with the sharedContext flow since it doesnt have any control of the extensions enabled.

So I still think you keep it all inside the queryAndSetDeviceFeatures and pass down the 2 parameters and everything is in one place, this new change is spreading out where the features are enabled.

And what you need to do is in the

if (sharedContext) {
  mDebugUtilsSupported =         context.mDebugUtilsSupported = scontext->debugUtilsSupported && enabledDebugUtils;
} else {
   context.mDebugUtilsSupported =
                setContains(instExts, VK_EXT_DEBUG_UTILS_EXTENSION_NAME);
}

poweifeng: there is no longer mDebugUtilsSupported, just mDebugUtilsEnabled.  

Supported was a misnomer because you could support it and not enable it. 

The shared context issue, I thought about moving it here as well, but maybe too many refactoring.  But in case the shared context is there, the value set here will just be overwritten, which is correct. 

Let me just move things back for the debug utils to not confuse issues here.  But I do want to do another refactoring to clean all this up.

poweifeng: resolved.

rafadevai (approved):

filament/backend/include/backend/platforms/VulkanPlatform.h
```diff
@@ -484,6 +484,14 @@ class UTILS_SHARED_LINKING VulkanPlatform : public Platform, utils::PrivateImple
         return {};
     }
 
+    /**
+     * On the certain platforms, it is possible to indicate when you'd like your swapchain image to
+     * be presented. Set to 0 if this behavior is not desired.
```
pixelflinger: you need to specify that "Set the time the next committed buffer should be presented to the user at". Technically the presentation time should be a swapchain API -- historically the backend api and gles backend are doing it globally (like here), but it was a mistake. regardless, we should document it affects the next swapchain presents.

pixelflinger (changes requested):

filament/backend/include/backend/platforms/VulkanPlatform.h
poweifeng: updated comment

rafadevai:
maybe is a little late to ask this, but how are people suppose to use this? I also thought there is already one at the filament level for all backends, is that the case?

poweifeng:
> maybe is a little late to ask this, but how are people suppose to use this? 

Client can call it on the Renderer here: https://github.com/google/filament/blob/main/filament/include/filament/Renderer.h#L380

> I also thought there is already one at the filament level for all backends, is that the case?

No, afaik this is only implemented for gles on android, and now vulkan on android.

poweifeng:
> > maybe is a little late to ask this, but how are people suppose to use this?
> 
> Client can call it on the Renderer here: https://github.com/google/filament/blob/main/filament/include/filament/Renderer.h#L380
> 
> > I also thought there is already one at the filament level for all backends, is that the case?
> 
> No, afaik this is only implemented for gles on android, and now vulkan on android.

actually, it looks like metal also implemented it.

pixelflinger (approved):

PR #10264, opened 2026-08-04
Add AbslStringify function to CString and CallStack

bejado:
This reverts the change made in https://github.com/google/filament/pull/10218 in favor of defining `AbslStringify` friend functions for `utils::CString` and `utils::CallStack`. This has the advantage that it doesn't incur a dependency on `std::ostream`.

BUGS=[537502416]

### DIFF
--- a/libs/utils/CMakeLists.txt
+++ b/libs/utils/CMakeLists.txt
@@ -175,6 +175,10 @@ if (CMAKE_CXX_COMPILER_ID MATCHES "Clang")
     target_compile_options(${TARGET} PRIVATE -Wthread-safety)
 endif()
 
+if (FILAMENT_USE_ABSEIL_LOGGING)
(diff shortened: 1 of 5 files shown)


libs/utils/CMakeLists.txt
```diff
@@ -175,6 +175,10 @@ if (CMAKE_CXX_COMPILER_ID MATCHES "Clang")
     target_compile_options(${TARGET} PRIVATE -Wthread-safety)
 endif()
 
+if (FILAMENT_USE_ABSEIL_LOGGING)
+    target_link_libraries(${TARGET} PUBLIC absl::log)
```
poweifeng, line 179: I wonder if this increases size as well. But I guess we can just commit and see.

poweifeng (approved):

z3moon (approved):

PR #10265, opened 2026-08-04
Add Engine::compile to support material compilation with ViewSettings

trie94:
There is a case where an app wants to precompile materials with view settings that are different from the current active View. To support that case, another Engine::compile is added that takes in ViewSettings.

FIXES=538287765

### DIFF
--- a/filament/include/filament/Engine.h
+++ b/filament/include/filament/Engine.h
@@ -17,11 +17,9 @@
 #ifndef TNT_FILAMENT_ENGINE_H
 #define TNT_FILAMENT_ENGINE_H
 
-
-#include "Options.h"
-
 #include <filament/ColorGrading.h>
 #include <filament/FilamentAPI.h>
+#include <filament/Options.h>
 
 #include <backend/DriverEnums.h>
 #include <backend/Platform.h>
@@ -1394,13 +1392,13 @@ public:
      * with different rendering settings than the active one.
      */
     struct ViewSettings {
-      bool hasDirectionalLighting;
-      bool hasFog;
-      bool hasStereo;
-      bool hasPostProcessing;
-      bool hasDynamicLighting;
-      bool hasShadowing;
-      ShadowType shadowType;
+        bool hasDirectionalLighting = false;
+        bool hasFog = false;
+        bool hasStereo = false;
+        bool hasPostProcessing = false;
+        bool hasDynamicLighting = false;
+        bool hasShadowing = false;
+        ShadowType shadowType = ShadowType::PCF;
     };
 
     /**
--- a/filament/include/filament/Engine.h
+++ b/filament/include/filament/Engine.h
@@ -1443,6 +1443,13 @@ public:
         utils::tribool skinning = utils::tribool::kIndeterminate,
         backend::CallbackHandler* UTILS_NULLABLE handler = nullptr,
         utils::Invocable<void(Material* UTILS_NONNULL)>&& callback = {});
+
+    /**
+     * @param view The View in which you want to extract the settings from.
+     * @return ViewSettings
+     */
+    ViewSettings extractViewSettings(View const* UTILS_NONNULL view) noexcept;
+
 protected:
     //! \privatesection
     Engine() noexcept = default;
--- a/filament/include/filament/Engine.h
+++ b/filament/include/filament/Engine.h
@@ -18,6 +18,8 @@
 #define TNT_FILAMENT_ENGINE_H
 
 
+#include "Options.h"
+
 #include <filament/ColorGrading.h>
 #include <filament/FilamentAPI.h>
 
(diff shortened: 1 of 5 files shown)


filament/include/filament/Engine.h
```diff
@@ -18,6 +18,8 @@
 #define TNT_FILAMENT_ENGINE_H
 
 
+#include "Options.h"
```
z3moon: please use `#include <filament/Options.h>` instead

filament/include/filament/Engine.h
```diff
@@ -1385,7 +1387,64 @@ class UTILS_PUBLIC Engine {
...
+     * with different rendering settings than the active one.
+     */
+    struct ViewSettings {
+      bool hasDirectionalLighting;
+      bool hasFog;
+      bool hasStereo;
+      bool hasPostProcessing;
+      bool hasDynamicLighting;
+      bool hasShadowing;
+      ShadowType shadowType;
```
z3moon: please set default values for these fields.

filament/src/details/Engine.cpp
```diff
@@ -1743,8 +1743,13 @@ bool* FEngine::getFeatureFlagPtr(std::string_view name, bool const allowConstant
 }
 
 FixedCapacityVector<Variant> FEngine::getMaterialCompileVariants(
-        FView const* view,
         FMaterial const* material,
+        bool const hasDirectionalLighting,
```
z3moon: Only this parameter has a `const`. Is there any reason?

filament/src/Engine.cpp
```diff
@@ -515,6 +515,19 @@ void Engine::compile(CompilerPriorityQueue priority, Material const* material, V
...
 
+void Engine::compile(
+    CompilerPriorityQueue priority,
+    Material const* material,
+    ViewSettings const& settings,
+    tribool shadowReceiver,
+    tribool skinning,
+    CallbackHandler* handler,
+    Invocable<void(Material*)>&& callback) {
+   downcast(this)->compile(priority, downcast(material), settings,
```
z3moon: 3 space indentation?

filament/src/details/Engine.cpp
```diff
@@ -1743,8 +1743,13 @@ bool* FEngine::getFeatureFlagPtr(std::string_view name, bool const allowConstant
 }
 
 FixedCapacityVector<Variant> FEngine::getMaterialCompileVariants(
```
show50726, line 1757: I wonder if it would be better to introduce a helper function like `getViewSettingsFromView`. This way, we would only need a single `getMaterialCompileVariants` method that consumes a `ViewSettings` object.

show50726: so that in the future if we add another variant/spec const, we only need to update the ViewSettings

filament/include/filament/Engine.h
romainguy: These also kind of expose inner workings of Filament. It would be better to pass a `View` and let Filament figure out what the variants are?

Note: the indentation is wrong, too

filament/include/filament/Engine.h
```diff
@@ -1385,7 +1387,64 @@ class UTILS_PUBLIC Engine {
...
+     * @param handler        Handler to dispatch the callback or nullptr for the default handler.
+     * @param callback       Callback called on the main thread when the compilation is done
+     *                       by the backend.
+     * @see Material::compile
+     */
+    void compile(
+        backend::CompilerPriorityQueue priority,
+        Material const* UTILS_NONNULL material,
+        ViewSettings const& settings,
+        utils::tribool shadowReceiver = utils::tribool::kIndeterminate,
```
romainguy, line 1442: This won't scale well as new features are added to Filament. It might be better to pass a renderable and let Filament figure out the variants to compile?

filament/src/details/Engine.cpp
romainguy: That would be better indeed, or like I suggested before, don't have a public ViewSettings.

filament/include/filament/Engine.h
trie94: yes and that's what the current `Engine::compile` does.

however that limits compiling material variants that are only necessary for the current `View` only. a specific use case that we are trying to support is: one of our client needs to compile materials with and without `FOG` variant. but when they load the material, the `View` has `hasFog` off, so they can't compile it with `FOG` enabled.

i guess a workaround can be calling `Engine::compile` twice with `View`'s `hasFog` on and off. however that doesn't seem to be ergonomic.

filament/src/details/Engine.cpp
trie94: oops, no i accidentally added `const`. removed.

filament/src/details/Engine.cpp
trie94: > I wonder if it would be better to introduce a helper function like getViewSettingsFromView. This way, we would only need a single getMaterialCompileVariants method that consumes a ViewSettings object.

yeah i was thinking of doing it on the impress side, but it makes sense to do it here. PTAL

filament/include/filament/Engine.h
trie94: i replied in the other comment, maybe we can discuss there?

also i don't expect we'd update `ViewSettings` that often, unless i am missing something

filament/src/details/Engine.cpp
```diff
@@ -19,6 +19,7 @@
 #include "MaterialParser.h"
 #include "RenderPrimitive.h"
 #include "TextureCache.h"
+#include "../../../libs/viewer/include/viewer/Settings.h"
```
poweifeng: this doesn't look right

filament/src/details/Engine.h
```diff
@@ -614,14 +614,29 @@ class FEngine : public Engine, public utils::FeatureFlagManager {
...
+
+    static utils::FixedCapacityVector<Variant> getMaterialCompileVariants(
+        FMaterial const* material, ViewSettings const& settings, utils::tribool shadowReceiver,
+        utils::tribool skinning) noexcept;
+
     static utils::FixedCapacityVector<Variant> getMaterialCompileVariants(
         FView const* view,
         FMaterial const* material,
         utils::tribool shadowReceiver,
         utils::tribool skinning) noexcept;
```
poweifeng, line 636: is it possible to just remove this override?

poweifeng: (since it's just an implementation detail and not the public api

trie94: yeah i moved `extractViewSettings`, `getMaterialCompileVariants`s, and `getMaterialCompileDynamicSpecConstKey` to the anonymous namespace.

filament/src/details/Engine.cpp
trie94: oops sorry, i accidentally pulled in a wrong dep

filament/src/details/Engine.cpp
```diff
@@ -157,6 +157,119 @@ Platform::DriverConfig getDriverConfig(FEngine* instance) {
     return driverConfig;
 }
 
+Engine::ViewSettings extractViewSettings(const FView* view) noexcept {
```
poweifeng: don't you want to expose this as a public api?  

That way users can actually use the ViewSettings version of compile without having to deal with *all* of the parameters (just change the ones they care about).

trie94: yeah i exposed it as a public api.

i thought in impress we could get all the view settings easily, but turns out having this helper function is easier

filament/src/details/Engine.h
trie94: actually i ended up reverting this. it is used in unit tests. i think it's fine to keep it here because anyways it's implementation details. (we are not exposing it as a public api)

filament/include/filament/Engine.h
romainguy: No but that would require public API changes, which is really not ideal. At least if everything was encapsulated in some sort of `struct VariantCompilationHints` or something it would make more sense.

filament/include/filament/Engine.h
romainguy: Or see my other suggestion below about a dedicated struct for this.

filament/include/filament/Engine.h
trie94: yeah i agree that it's not ideal but otherwise there's no way to achieve this. maybe `VariantCompileOptions`?

trie94: but i want to hear other people's thoughts too before i rename it

show50726: I agree renaming it would be better, since it doesn't actually reflect all the `View`'s settings but only for variants/compilation related ones. Either `VariantCompilationHints` or `VariantCompileOptions` works for me.

pixelflinger (changes requested):
I don't think this is the right API. I think for this one we should do an official design/review cycle internally.

In particular, I don't think it should use `bool` but maybe `tribool` so we can specify wildcards. Also,  I'm not sure it should be on View -- there could be Variants (aka spec-constants now) that are not set at the View level someday.

I worry this can corner us in the future.

PR #10267, opened 2026-08-05
web: expand JS bindings coverage, part 2

Erkko68:
Follow-up to #10216.

That PR closed gaps found by auditing one downstream binding layer. This one starts from the public C++ headers, which surfaces a larger set — including one case where JavaScript could not set integer material parameters at all — and adds tests that actually run the bindings, which turned up seven bugs that all type-checked cleanly.

## Bugs fixed

Each of these was found by running the bindings rather than declaring them.

| Binding | Symptom |
|---|---|
| `BufferObject$Builder.build` | Never wrapped: `BufferObject.Builder()…build(engine)` threw `build is not a function` |
| `SurfaceOrientation$Builder` `normals`/`tangents`/`uvs`/`positions`/`triangles16`/`triangles32` | Returned `undefined` instead of the builder, so the chained form `filament.d.ts` documents broke at the second call |
| `LightManager$Builder.intensityEnergy` | Wrapper called a `_intensityEnergy` that does not exist, shadowing the working binding underneath |
| `createTextureFromPng` / `createTextureFromJpeg` | Threw for anything mipmapped: they call `generateMipmaps` without asking for `GEN_MIPMAPPABLE`, now required by default. User-facing on every web client |
| `gltfio$FilamentAsset.getEntitiesByPrefix` | Always returned empty: asked for a count first, which that API answers with `0` |
| `MeshReader$MaterialRegistry.keys` | Returned twice as many entries as it had, half of them empty |
| `View.setDynamicResolutionOptions`, `setRenderQuality` | Rejected partial options objects, unlike every other option setter |
| `filament.d.ts` `getEntityByName` | Declared a method that is bound nowhere and does not exist in C++ |
| `ViewerGui` constructor | Wrote `true` through a null pointer into address 0 — see below |

## Bindings added

| Class | Added |
|---|---|
| `RenderableManager` | `getMorphTargetCount`, `setMorphTargetBufferOffsetAt`, `setSkinningBuffer`, `setMorphWeightsOffset` |
| `View` | The eleven option getters, plus `getRenderQuality`, `getDithering`, `getVisibleLayers`, `setLayerEnabled`, `isPostProcessingEnabled`, `set`/`isScreenSpaceRefractionEnabled`, `getScene`, `getCamera`, `getRenderTarget`, `set`/`getName`, `setSampleCount`, `getAntiAliasing` |
| `Material` | The whole reflective API: `hasParameter`, `getParameterCount`, `getParameters`, the shading/blending/refraction/reflection/vertex-domain getters, the raster-state queries |
| `MaterialInstance` | `setIntParameter` and the `int2-4`/`bool2-4` forms — **integer parameters could not be set from JS at all** — plus `setConstantBool/Float/Int` and `isStencilWriteEnabled` |
| `ToneMapper` | The hierarchy (Linear, ACES, ACESLegacy, Filmic, PBRNeutral, GT7, DisplayRange, AgX, Generic) and `ColorGrading::Builder::toneMapper`. The deprecated `toneMapping` enum could not reach the last four |
| `TextureSampler` | All twelve get/set accessors; the class was write-only at construction |
| `Texture` | `getTarget`, `getFormat`, `isTextureFormatSupported`, `getMaxTextureSize`, `getMaxArrayTextureLayers`, `Builder::samples`/`swizzle`/`external` |
| `Engine` | `setFeatureFlag`, `hasFeatureFlag`, `getFeatureFlag` (pairs with `Engine$Builder::feature`; returns `undefined` where C++ returns an empty optional), `flush`, `flushAndWait` |
| `SwapChain` | `isProtectedContentSupported`, `isMSAASwapChainSupported` |
| `Renderer` | `hasGpuFallenBehind` |
| `Camera` | `getEntity`, `setEyeModelMatrix`, `setCustomEyeProjection` |
| `LightManager` | `destroy`, `setIntensityCandela`, `Builder::intensityCandela`, the `ShadowCascades` split helpers |
| gltfio | The count accessors, `popRenderables`, `detachMaterialInstances`, `AssetLoader::enableDiagnostics`/`gc`/`destroy`, `ResourceLoader::asyncCancelLoad`/`evictResourceData`, the `MaterialProvider` members that take no `MaterialKey` |
| Also | `EntityManager` lifetime methods, `TransformManager::getChildCount` and the accurate-translations toggle, `IndexBuffer::getIndexCount`, `VertexBuffer::getVertexCount`, `Skybox::Builder::intensity` |

`AssetLoader::destroy` is worth flagging: `BIND(AssetLoader)` registers a no-op `raw_destructor`, so `loader.delete()` frees nothing and the loader leaks. There was previously no way to release it.

## Renames ⚠️

Two names diverged from C++ for no reason the binding layer forces:

| Was | Now | Why |
|---|---|---|
| `irradianceSh` / `irradianceTex`, `radianceSh` | `irradiance` (both), `radiance` | embind picks an overload by argument count, and the two `irradiance` forms differ in arity, so neither needs a name of its own |
| `gltfio$StbProvider`/`Ktx2Provider`/`WebpProvider`, `addStbProvider`/`addKtx2Provider`/`addWebpProvider` | `gltfio$TextureProvider` with `createStbProvider`/`createKtx2Provider`/`createWebpProvider`, and one `addTextureProvider` | They were split three ways because `TextureProvider` is polymorphic and embind reaches a base class through RTTI, which this build compiles without. An opaque handle sidesteps that, and matches how C++ and Java spell it |

## Tooling and tests

`tools/web-binding-audit` now holds two scripts, one per half of the question:

- **`unbound_apis.py`** (was `run.py`) — diffs the public C++ headers against the embind registrations and reports what the Java bindings already expose, so acting on a finding never makes the web surface wider than Android's. `filament.d.ts` is deliberately not treated as evidence a method is bound; a stale declaration is exactly how `getEntityByName` survived. Deliberate omissions carry their reason; anything else exits 1.
- **`untested_bindings.py`** — reports bindings no test calls. Currently 836 of 838 bound class members are exercised; the two exceptions are an internal entry point and a validity check for `Stream`, which JS cannot construct.

Three test files, each covering what the one before it cannot:

| File | How it runs | What it proves |
|---|---|---|
| `test.ts` | `tsc`, never executed | `filament.d.ts` matches the bindings, including calls that would abort at runtime |
| `test-runtime.js` | Node, NOOP backend | the bindings exist and marshal — **53 tests** |
| `test-browser.html` | browser, WebGL | rendering, pixel readback, picking, image decoding — **13 tests** |

They assert on round-trips, not on registration: option structs come back field for field, a child transform inherits its parent's translation, the free-flight manipulator moves when a key is held. Neither runnable file needs assets beyond three committed files and a glTF built inline; the browser page draws its own test images.

## Deliberately not bound

Recorded in `unbound_apis.py` with a reason each, checked against the implementation rather than the header. The short version: `Fence::waitAndDestroy` needs threads; `Engine::isValid` is overloaded per resource type at one argument each, so it stays the `isValidX` family; `Material::Builder` is not registered; `Material::setDefaultParameter` is sugar for `getDefaultInstance()->setParameter()`, both of which are bound; the three `compile()` methods and `Scene::forEach` need a `CallbackHandler` layer; `MaterialProvider::createMaterialInstance`/`getMaterial` are blocked on the bitfield-packed `MaterialKey`.

## Testing

`./build.sh -p wasm release` clean · runtime 53/53 · browser 13/13 in Chrome on both ANGLE/Metal and SwiftShader · `tsc` clean · both audit scripts exit 0. The audit was itself verified by deleting 15 registrations across all three binding styles and confirming it reported exactly those 15.

### The ViewerGui crash

Worth its own note, since it took `-sSAFE_HEAP` to find and it is not web-specific.

`ViewerGui`'s constructor did:

```cpp
*debug.getPropertyAddress<bool>("d.stereo.combine_multiview_images") = true;
```

That property is registered by the **first `Renderer` the engine creates**. A viewer built before any renderer exists — which the web bindings let you do, and which is the natural order for a page that sets up its UI first — gets `nullptr` back and writes `true` to address 0. Nothing traps: it corrupts the bottom of the heap, and the damage surfaces later somewhere unrelated, or not at all, depending on the heap layout. It moved around as unrelated bindings were added, which is what made it worth chasing.

The UI it draws had a second, milder version of the same assumption: `updateUserInterface` read `mAsset` and `mInstance` unconditionally, and a viewer that has not loaded anything has neither. Those reads are guarded now, so the sidebar can be drawn before a model is loaded.

### DIFF
--- a/tools/web-binding-audit/unbound_apis.py
+++ b/tools/web-binding-audit/unbound_apis.py
@@ -1,0 +1,487 @@
+#!/usr/bin/env python3
+# Copyright (C) 2026 The Android Open Source Project
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Report public C++ APIs that the Java/JNI bindings expose but the JS bindings do not.
+
+The web (embind) surface is maintained by hand, so it drifts from the C++ headers silently: a
+method that is never registered simply does not exist in JavaScript, and nothing fails at build
+time. This tool makes that drift visible.
+
+How it works
+------------
+Three surfaces are parsed and diffed:
+
+  C++    filament/include/filament/*.h and libs/gltfio/include/gltfio/*.h -- the API that exists.
+  web    web/filament-js/{jsbindings,jsbindings_generated,jsenums,jsenums_generated}.cpp plus
+         extensions.js -- the API JavaScript can actually reach. filament.d.ts is excluded on
+         purpose; see collect_web().
+  Java   android/**/src/main/java/.../*.java -- the API Android exposes.
+
+A C++ method is reported when it is absent from the web surface *and* present on the Java one.
+Requiring the Java side keeps the report actionable: it filters out APIs that no binding layer
+has ever exposed, and it means acting on a finding never makes the web surface wider than
+Android's.
+
+Deliberate omissions live in EXCLUSIONS below, each with the reason. Anything not excluded and
+not bound is reported, and the tool exits 1, so this can gate CI.
+
+Usage
+-----
+    ./tools/web-binding-audit/unbound_apis.py            # report unexplained gaps, exit 1 if any
+    ./tools/web-binding-audit/unbound_apis.py --all      # also list the deliberate exclusions
+    ./tools/web-binding-audit/unbound_apis.py --verbose  # print the C++ declaration for each gap
+
+Its counterpart, untested_bindings.py, starts from the other end: it reports bindings that exist
+but that no test exercises.
+
+Known limitations
+-----------------
+Matching is by *name*, not by signature, so a method bound with fewer parameters than C++
+declares (an overload gap) is reported as covered. setMorphWeights was such a case: bound, but
+without its offset argument. Signature-level checking would need a real C++ parser; until then,
+treat a clean run as "no missing names" rather than "no missing functionality".
+"""
+
+import argparse
+import os
+import re
+import sys
+from collections import defaultdict
+
+ROOT = os.path.normpath(os.path.join(os.path.dirname(os.path.abspath(__file__)), "..", ".."))
+
+HEADER_DIRS = [
+    ("filament/include/filament", ""),
+    ("libs/gltfio/include/gltfio", "gltfio$"),
+]
+
+WEB_SOURCES = [
+    "web/filament-js/jsbindings.cpp",
+    "web/filament-js/jsbindings_generated.cpp",
+    "web/filament-js/jsenums.cpp",
+    "web/filament-js/jsenums_generated.cpp",
+]
+
+JAVA_DIRS = [
+    "android/filament-android/src/main/java/com/google/android/filament",
+    "android/gltfio-android/src/main/java/com/google/android/filament/gltfio",
+    "android/filament-utils-android/src/main/java/com/google/android/filament/utils",
+]
+
+# Classes that are not part of the JS surface by design.
+SKIPPED_CLASSES = {
+    # CRTP/infrastructure helpers, not real API.
+    "BuilderNameMixin", "FilamentAPI",
+    # Math and geometry value types, bound as value_array/value_object rather than classes.
+    "Aabb", "Aabb$Corners", "Box", "Viewport", "Color", "ColorSpace", "Gamut", "TransferFunction",
+    # Iterator plumbing.
+    "TransformManager$children_iterator", "TransformManager$children_range",
+    # Frame pacing and history: driven by requestAnimationFrame on the web.
+    "FrameHistoryStream", "FrameHistoryStream$NewFramesRange", "FrameHistoryStream$Result",
+    "FramePacer", "FramePacer$Builder", "FramePipelineEstimator", "Sync",
+    "gltfio$AssetConfigurationExtended",
+}
+
+# A C++ name that covers many overloads is usually bound as a family of typed entry points
+# (setParameter -> setBoolParameter, setInt2Parameter, ...). Rather than hand-listing those names,
+# derive the family from the *Java* bindings, which spell every overload out with concrete types:
+#
+#   public void setParameter(String name, int x, int y)  ->  an int2 overload must be reachable
+#
+# Each Java overload is turned into a type token (Int2) and checked against the web surface under
+# the naming conventions the bindings actually use, infix (setInt2Parameter) or suffix
+# (setParameterInt2). A family that covers only some overloads then reports the specific ones that
+# are missing, instead of hiding behind a comment claiming the family is complete.
+JAVA_TYPE_TOKENS = {
+    "boolean": "Bool", "float": "Float", "int": "Int", "double": "Double", "long": "Long",
+}
+
+# Methods whose family should be derived, keyed by "Class.method". The value is the Java class file
+# to read the overloads from (usually the same class).
+DERIVED_FAMILIES = {
+    "MaterialInstance.setParameter": "MaterialInstance",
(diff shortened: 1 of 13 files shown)


tools/web-binding-audit/unbound_apis.py
```diff
@@ -0,0 +1,445 @@
+#!/usr/bin/env python3
```
pixelflinger, line 1: needs the filament copyright header/comment

pixelflinger (changes requested):
just meeds copy right header for the new script.

Erkko68:
I'm converting this into a draft PR, as this covers a lot of bindings and I want to make sure I get everything right. I'll add a full NOOP test suite and the missing methods to the typings tests to verify them at runtime.

Erkko68:
@pixelflinger Should be good for review now. I added new test files, fixed some broken bound APIs, and added missing ones. I know it’s a pretty big contribution, but I think it’s much needed to address the gaps that have built up over time in the Web API.

poweifeng (approved):

Erkko68:
@pixelflinger Sorry for the ping, have you had a chance to review this?

poweifeng:
Hi Eric, 

I did another pass through the PR. This PR was too big for human review, so I asked ai.  And it did give me some feedback/blockers.  

Would you mind splitting it into more digestable pieces? Maybe one PR for all the bug fixes, one PR for all the binding additions, and one PR for the python script for correctness.

FWIW, the ai comments are:

```
 ### jsbindings.cpp                                                                                                                                                                                                                                                                                                                                                  
                                                                                                                                                                                                                                                                                                                                                                      
  • jsbindings.cpp:1944-1966 [correctness] Stack buffer over-read in LightManager$ShadowCascades splits calculation: LightManager.cpp:524-530 clamps cascades internally (min(cascades, 4u)), writing at most 3 entries into splits[3]. However, the embind lambda receives cascades by value and iterates for (uint8_t i = 0; i + 1 < cascades; i++) using the un-   
  clamped input. If called from JavaScript with cascades > 4 (e.g., cascades = 5), the loop accesses splits[3] (and beyond), reading out-of-bounds stack memory into the returned JS array. This affects jsbindings.cpp:1944-1950, jsbindings.cpp:1951-1958, and jsbindings.cpp:1959-1966. cascades should be clamped at the lambda boundary: cascades =              
  std::min<uint8_t>(cascades, 4);.                                                                                                                                                                                                                                                                                                                                    
  • jsbindings.cpp:2477-2480 [api-contract] Duplicate method registration on Skybox$Builder: BUILDER_FUNCTION("intensity", SkyBuilder, ...) is registered twice in identical succession on Skybox$Builder.                                                                                                                                                            
  • jsbindings.cpp:2977-2982 [correctness] Iterator range construction from nullptr in UbershaderProvider::getMaterials: When provider->getMaterialsCount() is 0 (or before any materials are created in ArchiveCache.h:47), provider->getMaterials() can return nullptr. Constructing std::vector<Material*>(materials, materials + count) with materials == nullptr 
  produces an invalid iterator range [nullptr, nullptr), which is undefined behavior under C++ iterator requirements and triggers runtime assertions in debug iterator builds. Guard with if (!materials || provider->getMaterialsCount() == 0) return {};.                                                                                                           
  • jsbindings.cpp:2433 [behavior-change] Diagnostic log message mismatch: The error log for invalid spherical harmonics float counts prints expected %d, nbands instead of nbands * nbands * 3 floats.                                                                                                                                                               
                                                                                                                                                                                                                                                                                                                                                                      
  ### ViewerGui.cpp                                                                                                                                                                                                                                                                                                                                                   
                                                                                                                                                                                                                                                                                                                                                                      
  • ViewerGui.cpp:1223-1225 [correctness] Unchecked null pointer passed to ImGui::Checkbox: Under #if defined(FILAMENT_SAMPLES_STEREO_TYPE_MULTIVIEW), debug.getPropertyAddress<bool>("d.stereo.combine_multiview_images") is passed directly to ImGui::Checkbox. If updateUserInterface() runs before a Renderer has registered this debug property,                 
  getPropertyAddress returns nullptr, causing ImGui::Checkbox to dereference null.                                                                                                     
```

Erkko68:
Hey, sure! I’ll be out starting today until next Monday, so I’ll close this PR and open the three new ones when I’m back. Thanks!

Erkko68:
Splitting this into smaller PRs, as requested:

- #10363 — the `ViewerGui` null dereferences, now with a regression test
- #10364 — the bindings and their tests, with the review comments on this PR addressed
- the `tools/web-binding-audit` scripts will follow separately

Closing in favour of those.

PR #10269, opened 2026-08-06
Update Filament material doc

trie94:
- add --output-format and --include-source-mat
- add material api level
FIXES=540482735

### DIFF
--- a/docs/Materials.md.html
+++ b/docs/Materials.md.html
@@ -2903,6 +2903,8 @@ The command line flags relevant to application development are described in tabl
 **-S**, **--optimize-size**     | N/A                | Optimize compiled material for size instead of just performance
 **-r**, **--reflect**           | parameters         | Outputs the specified metadata as JSON
 **-v**, **--variant-filter**    | [variant]          | Filters out the specified, comma-separated variants
+**-f**, **-output-format**      | blob/header/mat    | Default to blob, when mat is set, returns a `.mat` with `#include`s resolved
+**--include-source-mat**        | N/A                | Embeds the source ASCII material definition if set
 [Table [matcFlags]: List of `matc` flags]
 
 `matc` offers a few other flags that are irrelevant to application developers and for internal
@@ -2993,6 +2995,16 @@ in the material itself.
 
 Use this flag with caution, filtering out a variant required at runtime may lead to crashes.
 
+### --output-format
+Specifies the output format:
+- blob (default): Outputs a compiled binary material package file.
+- header: Outputs a C++ header file containing the compiled material package embedded as a static byte array, allowing direct inclusion in C++ source code.
+- mat: Outputs a processed .mat source text file with all #include directives fully resolved into a single standalone source file.
+
+### --include-source-mat
+Embeds the source ASCII material definition into the compiled material package (blob). The embedded source code has all
+`#include` directives recursively resolved.
+
 # Handling colors
 
 ## Linear colors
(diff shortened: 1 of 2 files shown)


poweifeng (commented):
Could you write this in `docs_src/src_markdeep/Materials.md.html`  instead?  That's the source of truth

docs/Materials.md.html
```diff
@@ -3052,4 +3064,22 @@
...
+In cases of remote renderers where an app and the system run different Filament instances and materials are compiled at
+runtime on the remote end, material API versioning ensures compatibility across versions. It is less relevant if your
+project uses a single Filament instance, but apiLevel is still required when opting into features introduced at higher
+API levels features.
+
+When authoring a material, you can specify apiLevel inside the material block. If omitted, it defaults to 1.
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ GLSL
+material {
+    name: "MyMaterial",
+    apiLevel: 2, // Client-requested API level
```
pixelflinger: we should document the corresponding new #define made availlable inside the materials  (I think there is right?). Also isn't there a new matc option to control the api version (I don't remember).

pixelflinger (dismissed):

docs/Materials.md.html
trie94: > we should document the corresponding new #define made availlable inside the materials

ah you are right, i added them.

> Also isn't there a new matc option to control the api version (I don't remember).

i mentioned it as an idea in the design doc but we didn't add it. actually we could:
1) add a new flag to set matc's api level (overrides `.mat`'s apiLevel)
2) update `--version` to return apiLevel as well (currently it returns material version only)
3) add better error messages when apiLevel is incorrectly requested. right now example messages are like: "'clipSpacePosition' : no such field in structure 'material'", which doesn't say anything about apiLevel...
4) add a api level query c++ api (this was also mentioned in the design doc as a potential future work, but was never added)

wdyt?

docs_src/src_markdeep/Materials.md.html
```diff
@@ -3052,4 +3064,68 @@
...
+- `CLIENT_MATERIAL_API_LEVEL`: The API level requested in the client's `.mat` file (defaults to `1`).
+- `UNSTABLE_MATERIAL_API_LEVEL`: The next API level currently in development. Features at this level have not been officially released.
+
+System renderers can use these defines to introduce features under development, branch shader implementation details,
+or guard new capabilities based on the client's declared API level:
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ GLSL
+// Adding a new feature
+#if CLIENT_MATERIAL_API_LEVEL >= UNSTABLE_MATERIAL_API_LEVEL
+float myNewFeatureUnstable() { /* ... */ }
+#endif
```
poweifeng, line 3099: no need to put this in the doc, but does that mean, when we bump `UNSTABLE_MATERIAL_API_LEVEL`, we need to replace all of these with a number?

docs_src/src_markdeep/Materials.md.html
```diff
@@ -3052,4 +3064,68 @@
...
+material if the API level they provided is higher than the compiler API level.
+- Backward compatibility (client version <= compiler version): Compilation success.
+- Forward compatibility (client version > compiler version): Compilation fails with an error message.
+
+Material compilation should succeed as long as those two conditions are met:
+- The client exclusively uses APIs that are supported at their specified api_level.
+- The compiler's supported api_level is equal to or greater than the client's requested api_level.
+
+Below demonstrates some concrete examples:
+Here is the Markdown format for the table you highlighted:
```
poweifeng: this comment seems out of place

docs_src/src_markdeep/Materials.md.html
```diff
@@ -3052,4 +3064,68 @@
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+### API level compatibility
+Note that this system doesn’t allow forward compatibility, meaning that there’s no way for the client to compile the
+material if the API level they provided is higher than the compiler API level.
+- Backward compatibility (client version <= compiler version): Compilation success.
+- Forward compatibility (client version > compiler version): Compilation fails with an error message.
+
+Material compilation should succeed as long as those two conditions are met:
+- The client exclusively uses APIs that are supported at their specified api_level.
```
poweifeng: nit: `apiLevel` for consistency

docs/Materials.md.html
pixelflinger: yes I think all of these make sense and would be good quality of life additions.

docs_src/src_markdeep/Materials.md.html
trie94: oops my bad

docs_src/src_markdeep/Materials.md.html
trie94: short answer is yes, but i don't think that's always the case.

we may have different features that are under development, and we want to bump up `UNSTABLE_MATERIAL_API_LEVEL`, we may not want all the features to be checked into the next api level. so as long as the feature is ready for the release, it should be replaced with a number.

docs/Materials.md.html
trie94: sounds good, will address those as follow ups

poweifeng (approved):

PR #10270, opened 2026-08-06
Update Android and workflow dependencies

romainguy:
Update all dependencies (GitHub actions, Gradle, AGP, Kotlin, etc.) to latest versions. This change also fixes a couple of warnings in the Android bindings.


romainguy:
I need to update the GitHub workflows to use Java 21 instead of 17. Will upload a fix asap.

romainguy:
@pixelflinger Can you make the `.github/workflows` changes so I can do the rest? Or if you prefer, do the whole PR.

poweifeng:
> @pixelflinger Can you make the `.github/workflows` changes so I can do the rest? Or if you prefer, do the whole PR.

bypassed.  PR should be ok now.

romainguy:
@poweifeng There were a lot of GitHub actions scan errors, I addressed them all but you might want to take a close look at the change, just in case.

poweifeng:
> @poweifeng There were a lot of GitHub actions scan errors, I addressed them all but you might want to take a close look at the change, just in case.

feel free to ignore them for now.  I can do another pass to address the rest.

romainguy:
@poweifeng I addressed a few more issues. All macro expansion errors should now be fixed. What's left is the use of triggers that the scan doesn't like.

poweifeng (approved):

romainguy:
@poweifeng Do you need me to do something more here? Can you bypass the scan?

poweifeng:
> @poweifeng Do you need me to do something more here? Can you bypass the scan?

So I did try out this PR, and it forced me to update to jdk 21, which is intended I think.  

But could you also update the documentation [here](https://github.com/google/filament/blob/main/BUILDING.md#prerequisites)?

romainguy:
@poweifeng Done!

PR #10275, opened 2026-08-07
Speed up the AABB culler

Nemi26:
Clang (and I assume other compilers) were ~~completely unable to vectorize~~ not fully vectorizing the AABB culler as it stands now.

This is what I would consider the theoretical peak optimal performance: https://godbolt.org/z/Ex9Tv364T

However I haven't been able to fully implement it that way in Filament primarily for the following reason:
In order to supply vec8's (or vec4's, since Filament doesn't have a vec8 type) to the culler, the SoA structure would have to store that type, where each component maps to one element in the scene. 
i.e. worldExtent.x = element 0, worldExtent.y = element 1, and so on, however the SoA assumes the same array size for all entries, and so that would lead to quite a lot of wasted memory.

Also this change unfortunately slightly affects the distance calculation in the renderpass class, but I don't think this is too big of a deal.

Overall, the benchmarks show a 1.519x speedup on my machine in Release mode (msvc), which I think is a solid start.

I think switching the center/extent arrays to vec4 or vec8 would be good (or even necessary) future work, but I am happy to leave this as it is for now until I see a more elegant solution to the SoA problem than wasting all that memory.

Disclaimer:
I used Copilot for generating the benchmarking code.

### DIFF
--- a/filament/benchmark/benchmark_filament.cpp
+++ b/filament/benchmark/benchmark_filament.cpp
@@ -36,12 +36,40 @@
 
 #include <benchmark/benchmark.h>
 
+#include <math/fast.h>
+
+#include <cmath>
 #include <random>
 #include <vector>
 
 using namespace filament;
 using namespace filament::math;
 using namespace utils;
+
+namespace {
+
+static void baselineAabbIntersects(Culler::result_type* UTILS_RESTRICT results,
+        Frustum const& UTILS_RESTRICT frustum,
+        float3 const* UTILS_RESTRICT center,
+        float3 const* UTILS_RESTRICT extent,
+        size_t count) noexcept {
+    float4 const* UTILS_RESTRICT const planes = frustum.getNormalizedPlanes();
+    count = Culler::round(count);
+    for (size_t i = 0; i < count; i++) {
+        int visible = ~0;
+        for (size_t j = 0; j < 6; j++) {
+            const float dot =
+                    planes[j].x * center[i].x - std::abs(planes[j].x) * extent[i].x +
+                    planes[j].y * center[i].y - std::abs(planes[j].y) * extent[i].y +
+                    planes[j].z * center[i].z - std::abs(planes[j].z) * extent[i].z +
+                    planes[j].w;
+            visible &= fast::signbit(dot);
+        }
+        results[i] = Culler::result_type(visible);
+    }
+}
+
+} // namespace
 
 
 class FilamentCullingFixture : public benchmark::Fixture {
@@ -51,6 +79,12 @@ protected:
     Frustum frustum{};
     std::vector<float3> boxesCenter;
     std::vector<float3> boxesExtent;
+    std::vector<float> boxesCenterX;
+    std::vector<float> boxesCenterY;
+    std::vector<float> boxesCenterZ;
+    std::vector<float> boxesExtentX;
+    std::vector<float> boxesExtentY;
+    std::vector<float> boxesExtentZ;
     std::vector<float4> spheres;
     Culler::result_type* UTILS_RESTRICT visibles = nullptr;
 
@@ -66,6 +100,12 @@ public:
 
         boxesCenter.resize(batch);
         boxesExtent.resize(batch);
+        boxesCenterX.resize(batch);
+        boxesCenterY.resize(batch);
+        boxesCenterZ.resize(batch);
+        boxesExtentX.resize(batch);
+        boxesExtentY.resize(batch);
+        boxesExtentZ.resize(batch);
         spheres.resize(batch);
         for (size_t i = 0; i < batch; i++) {
             float4& sphere = spheres[i];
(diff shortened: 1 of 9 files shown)


pixelflinger:
mh... that might be a problem with msvc -- this code was specifically written to vectorize nicely in clang. I have verified this on ARM many times (I haven't looked at intel too much though, but clang should vectorize the same on both).

pixelflinger:
How does this change performs on ARMv8? 
I'm not too kin on what this does to the code, especially because vectorization is working well on ARM, if this is a specific problem with msvc, I don't really want to uglyfy the code just for that.

pixelflinger:
clang generate sequences like this:

```
	ldr	q21, [sp, #80]
	ldp	q22, q24, [sp, #32]
	fadd	v24.4s, v26.4s, v24.4s
	fadd	v21.4s, v29.4s, v21.4s
	fadd	v20.4s, v21.4s, v20.4s
	fneg	v21.4s, v31.4s
	fmla	v21.4s, v25.4s, v22.s[0]
	ldr	q22, [sp, #64]
	and	v16.16b, v16.16b, v20.16b
```

Nemi26:
> mh... that might be a problem with msvc -- this code was specifically written to vectorize nicely in clang. I have verified this on ARM many times (I haven't looked at intel too much though, but clang should vectorize the same on both).

I didn't check MSVC assembly output, but on clang the difference seems to be in favor of my version where the loop order is swapped. I only ran the benchmarks on msvc because I'm on windows, where clang isn't supported.

https://godbolt.org/z/P34j19z6e

Nemi26:
> How does this change performs on ARMv8? I'm not too kin on what this does to the code, especially because vectorization is working well on ARM, if this is a specific problem with msvc, I don't really want to uglyfy the code just for that.

For a more significant speedup I'd turn the aabbCenterX/Y/Z into float4 types, but I wasn't a fan of the wasted memory. I can still try it and report perf numbers

pixelflinger:
> > How does this change performs on ARMv8? I'm not too kin on what this does to the code, especially because vectorization is working well on ARM, if this is a specific problem with msvc, I don't really want to uglyfy the code just for that.
> 
> For a more significant speedup I'd turn the aabbCenterX/Y/Z into float4 types, but I wasn't a fan of the wasted memory. I can still try it and report perf numbers

so it's not "completely unable to vectorize".

Nemi26:
> so it's not "completely unable to vectorize".

Sorry, I wrote this late at night, fixed the wording in the OP

pixelflinger:
these are the numbers I get for Pixel 8 Pro:

```
----------------------------------------------------------------------------------------------------------------------------------
Benchmark                              Time           CPU Iterations        BPU          C        CPI          I items_per_second
----------------------------------------------------------------------------------------------------------------------------------
FilamentCullingFixture/boxCulling          1212 ns       1208 ns     578797          0    6.87234   0.328354    20.9297       423.884M/s
FilamentCullingFixture/sphereCulling        748 ns        745 ns     938377          0    4.24185    0.39125    10.8418       686.839M/s
```

So about 7 cycles per AABB -- about 1 cycle per plane -- and 4 cycles per sphere.

pixelflinger:
So I tested this change locally on Pixel 8 Pro and it is better; but I did run into some problems that make it hard to compare.

First, the data that I had gathered doesn't match the benchmark anymore (it regressed), and second I can't get the performance counters to work anymore. 

Regardless, for the AABB culling test (the one I really care about):
old benchmark: 423M/s
new benchmark: 400M/s
optimized benchmark: 470M/s

So it's a nice ~17% (x1.17) improvement. Before I take the change, I want to resolve our two mysteries:

- why did the benchmark regress, since the code hasn't changed. It could be the compiler, compile options, or the OS?
- why can't I use perf counters anymore?

Even more strange, if I disable vectorization entirely:
current code: 400M/s (no change)
optimized benchmark: 160M/s (65% significant regression)

Both benchmarks run on the Big core at ~3GHz (verified in systrace), so the numbers don't really add-up:
 If the old code wasn't vectorized this would explain why it performs the same with and without vectorization. Moreover, this would also explain why the new code performs nearly 3 times better without/with vectorization. However, this doesn't explain why the "non vectorized / badly vectorized" old code performs only ~17 points worse than the "good vectorized" code. It could be that the "new vectorized" code is a pessimisation when vectorization is not enabled because of all the pointer chasing -- it might just run out of read buffers. This brings the question of the need to test this on the little cores.
 
 I'll investigate more.  I don't think there is any urgency to get this in; I want to understand what's happening and see if we can get the gains without "uglyfing" the code so much.
 
 Thank you for this contribution.

Nemi26:
I'm happy to just get the idea across, and I don't mind if my implementation doesn't get merged. I initially got this idea several months ago when I tossed the culler standalone into compiler explorer, and found that I got way better vectorization by swapping the loop order, and splitting the AABB centers/extents into 6 separate streams (which I initially made vec4 for SSE2, and then vec8 for AVX2, both of which massively outperformed the current implementation, though I forgot the exact numbers).

I don't have an android phone at the moment to test this on, I've only been using Filament on Windows so far, and I honestly haven't even looked at the assembly of MSVC since I am only really interested in Clang.

Good luck with the rest of the investigation.

pixelflinger:
> I'm happy to just get the idea across, and I don't mind if my implementation doesn't get merged. I initially got this idea several months ago when I tossed the culler standalone into compiler explorer, and found that I got way better vectorization by swapping the loop order, and splitting the AABB centers/extents into 6 separate streams (which I initially made vec4 for SSE2, and then vec8 for AVX2, both of which massively outperformed the current implementation, though I forgot the exact numbers).
> 
> I don't have an android phone at the moment to test this on, I've only been using Filament on Windows so far, and I honestly haven't even looked at the assembly of MSVC since I am only really interested in Clang.
> 
> Good luck with the rest of the investigation.

Ah yes, I think the ARM cores on mobile won't be able to track 6 streams, maybe the Big core might.

romainguy:
> found that I got way better vectorization by swapping the loop order

What's disappointing is that when we first looked at the vectorization of this code, clang would swap the loop order on its own. But that was ~10 years ago, so…

pixelflinger:
@Nemi26 using a combination of using intrinsics + the SoA structure, I was about to improve the ARM performance significantly -- up to 3.3x on little cores. My change is similar but different enough from yours, so I will close this PR and tag you in the new one.
Thank you for your initial contributions and ideas.

PR #10276, opened 2026-08-08
samples: copy-capture App struct in lambda

poweifeng:
The shared data of App struct needs to survive across the lifetime of FilamentApp2 and so needs to shared_ptr'd. This pointer should be explicitly copied (for proper ref-counting).


bejado (approved):

samples/vbotest.cpp
```diff
@@ -58,73 +60,90 @@ static void setCameraProjection(App* app, View* view) {
         -ZOOM, ZOOM, 0, 1);
 }
 
-int main(int argc, char** argv) {
-    SampleConfig config;
-    config.title = "vbotest";
-    samples::handleCommandLineArguments(argc, argv, &config);
+std::unique_ptr<FilamentApp2> createSampleApp(SampleConfig config,
+        filament::app::DisplayManager* dm, filament::app::AssetLoader* loader) {
```
z3moon, line 64: `loader` is not used in here. Is it intended?

samples/viewtest.cpp
```diff
@@ -93,7 +93,7 @@ std::unique_ptr<FilamentApp2> createSampleApp(SampleConfig config,
         utils::EntityManager::get().destroy(app->camera);
     };
 
-    auto preRender = [](Engine*, View*, Scene*, Renderer* renderer) {
+    auto preRender = [app](Engine*, View*, Scene*, Renderer* renderer) {
```
z3moon: `app` is not used. maybe it's used later?

poweifeng: can be removed, but really now that app is a shared_ptr, it needs to be copied to outline all of the lifecycles.

samples/vbotest.cpp
poweifeng: yes, this is intended.

samples/viewtest.cpp
poweifeng: done

PR #10277, opened 2026-08-08
utils: Modernize Allocator and Arena subsystem

pixelflinger:


filament/src/details/Engine.h
```diff
@@ -189,7 +189,7 @@ class FEngine : public Engine, public utils::FeatureFlagManager {
     // the per-frame Area is used by all Renderer, so they must run in sequence and
     // have freed all allocated memory when done. If this needs to change in the future,
     // we'll simply have to use separate Areas (for instance).
-    LinearAllocatorArena& getPerRenderPassArena() noexcept { return mPerRenderPassArena; }
+    auto& getPerRenderPassArena() noexcept { return mPerRenderPassArena; }
```
romainguy, line 192: Did you mean to use an auto return type in a header?

filament/src/fg/details/DependencyGraph.h
```diff
@@ -17,6 +17,8 @@
 #ifndef TNT_FILAMENT_FG_DETAILS_DEPENDENCYGRAPH_H
 #define TNT_FILAMENT_FG_DETAILS_DEPENDENCYGRAPH_H
 
+#include "fg/details/Utilities.h"
+
 #include <utils/CString.h>
 #include <utils/debug.h>
 #include <utils/FixedCapacityVector.h>
```
poweifeng: maybe not needed?

filament/src/details/Engine.h
pixelflinger: yes, it's a private header. It makes it way easier to change the type.

poweifeng (approved):

filament/src/fg/details/DependencyGraph.h
pixelflinger: It's needed for `Vector` I think.

poweifeng: I meant the `<utils/FixedCapacityVector.h>`?  (`Vector` is in `fg/details/Utilities.h`)

pixelflinger: oh! yes.

PR #10291, opened 2026-08-11
filamentapp: implement AndroidDisplayManager

poweifeng:
- Clean up unused parameters in the FilamentApp2::Builder
 - Implement AndroidDisplayManager (subclasses from filamentapp::DisplayManager by mapping it to SurfaceView


samples/hellostereo.cpp
```diff
@@ -138,7 +143,6 @@ std::unique_ptr<FilamentApp2> createSampleApp(SampleConfig config,
                         .height(vp.height)
                         .depth(eyeCount)
                         .levels(1)
-                        .samples(sampleCount)
```
z3moon, line 141: I think these `.samples(sampleCount)` blocks, including below, should be kept?

poweifeng: corrected. thanks for catching it.

z3moon (approved):

PR #10293, opened 2026-08-11
vk: fix streaming textures path

poweifeng:
- If an external image is acquired through the stream, we need to mark the corresdponding set as "dirty" because the image might need external samplers.
 - Fix breakage due to referencing a null texture in updateStreams
 - Prevent duplicating entries in mSetBindings when calling VulkanExternalImageManager::bindExternallySampledTexture

FIXES=447136715


filament/backend/src/vulkan/VulkanExternalImageManager.cpp
```diff
@@ -175,6 +175,7 @@ void VulkanExternalImageManager::bindExternallySampledTexture(
         .conversion = imageData.conversion,
     });
 
+    clearTextureBinding(set, bindingPoint);
```
rafadevai: if you clear the bindings here, it will break the other input bindings ones that are not updated at the same time as the stream.

poweifeng: isn't it true a set+binding is unique? That is, the mBindingSets should not contain two entries with the same set and binding.

rafadevai: Is true that it has to be unique, but clearing all the current bindings will break things.

For example, lets say you have 2 streams and you do 2 updates one after the other.

poweifeng: But `clearTextureBinding` only clears one binding+set entry. See implementation: 

```
void VulkanExternalImageManager::clearTextureBinding(
        fvkmemory::resource_ptr<VulkanDescriptorSet> set, uint8_t bindingPoint) {
    erasep<SetBindingInfo>(mSetBindings, [&](auto const& bindingInfo) {
        return (bindingInfo.set == set && bindingInfo.binding == bindingPoint);
    });
}
```

poweifeng: actually, thinking more about this, it'd be better to not erase an element from the array, instead just check for existence. Let me change the logic

poweifeng: done

filament/backend/src/vulkan/VulkanStreamedImageManager.cpp
```diff
@@ -48,9 +48,10 @@ void VulkanStreamedImageManager::unbindStreamedTexture(
 
 void VulkanStreamedImageManager::onStreamAcquireImage(fvkmemory::resource_ptr<VulkanTexture> image,
         fvkmemory::resource_ptr<VulkanStream> stream) {
-    for (StreamedTextureBinding const& data: mStreamedTexturesBindings) {
+    for (StreamedTextureBinding& data: mStreamedTexturesBindings) {
         // Find the right stream
         if (data.image->getStream() == stream) {
+            data.set->isLayoutDirty = true;
```
rafadevai, line 55: also add

data.set->isAnExternalSamplerBound = true;

filament/backend/src/vulkan/VulkanExternalImageManager.cpp
rafadevai: Sorry, I misread it. You're right only one is good enough

filament/backend/src/vulkan/VulkanStreamedImageManager.cpp
poweifeng: done

rafadevai:
So one last thing for the stream we are not setting

mAppState.hasBoundExternalImages = true;

we probably want to do it on a per image acquire basis. What do you think??

poweifeng:
> So one last thing for the stream we are not setting
> 
> mAppState.hasBoundExternalImages = true;
> 
> we probably want to do it on a per image acquire basis. What do you think??

So this is actually set in `updateDescriptorSetTexture`, when it's set to a stream, `hasBoundExternalImages` is set to true

rafadevai (approved):

PR #10294, opened 2026-08-12
Make error associated with material api level more readable

trie94:
If a material is using an api that requires 2 but uses api level of 1, it fails to compile with `ERROR 'apiName' : no matching overloaded function found.` The error message is not very straighforward, so this PR makes the error message more readable.

For example,
```
ERROR: 0:9: 'getWorldRayFromClip' : no matching overloaded function found
```
is now
```
ERROR: 0:9: getWorldRayFromClip requires api level 2
```

To make this work, any new api that is introduced in `UNSTABLE_MATERIAL_API_LEVEL` should add this template in the else branch: `#define {myNewFeature} ERROR_{myNewFeature}_api_level_{N}_END`.

For example:
```
#if CLIENT_MATERIAL_API_LEVEL >= UNSTABLE_MATERIAL_API_LEVEL
/**
 * @api-level 2
 */
void myNewFeature() {...}
#else
#define myNewFeature ERROR_myNewFeature_api_level_2_END
#endif
```
Note:
- If the new api only changes the behavior between the api levels, you don't need to add this error template. This is because the function always presents.  An example for this case is `getWorldFromModelMatrix`.
- If it doesn't follow the template `ERROR_api_level_{N}_END`, for example `ERROR_some_custom_error_message_END`, it'd just replace `_` with  ` `.
- `getWorldFromModelMatrix` is moved from `shaders/src/surface_instancing.glsl` to `shaders/src/surface_getters.vs`, so that we have them in the same place.

FIXES=544716600

### DIFF
--- a/libs/filamat/src/sca/GLSLTools.cpp
+++ b/libs/filamat/src/sca/GLSLTools.cpp
@@ -60,36 +60,49 @@ static void formatApiLevelErrorLog(utils::io::ostream& out, const char* log) {
         }
         
         std::string_view line = fullLog.substr(start, end - start);
-        // Error message follows the format: ERROR_{custom_error_message}_END
+
+        // The error message follows the format: #define {myFunctionName} ERROR_{myFunctionName}_api_level_{N}_END
         constexpr std::string_view prefix = "ERROR_";
+        constexpr std::string_view postfix = "_END";
+
         size_t startPos = line.find(prefix);
-        if (startPos != std::string_view::npos) {
-            size_t endPos = line.find("_END", startPos);
-            if (endPos != std::string_view::npos) {
-                // Extract the message after "ERROR_" and before "_END"
-                size_t msgStart = startPos + prefix.length();
-                std::string_view rawMsg = line.substr(msgStart, endPos - msgStart);
-                
-                utils::CString message(rawMsg);
-                for (char& c : message) {
-                    if (c == '_') c = ' ';
-                }
-                
-                size_t quoteStart = line.rfind('\'', startPos);
-                if (quoteStart != std::string_view::npos) {
-                    // Print prefix (e.g. "ERROR: 0:9: ") and message, discarding everything after
-                    out << line.substr(0, quoteStart) << message;
-                } else {
-                    out << line.substr(0, startPos) << message;
-                }
-                out << utils::io::endl;
-            } else {
-                out << line << utils::io::endl;
-            }
-        } else {
+        size_t endPos = line.find(postfix, startPos);
+        if (startPos == std::string_view::npos || endPos == std::string_view::npos) {
+            // Doesn't fit to the template, simply print the log as is.
             out << line << utils::io::endl;
+            start = end + 1;
+            continue;
         }
-        
+
+        // Extract the message between "ERROR_" and "_END".
+        size_t msgStart = startPos + prefix.length();
+        std::string_view rawMsg = line.substr(msgStart, endPos - msgStart);
+
+        // First append prefix.
+        size_t quoteStart = line.rfind('\'', startPos);
+        if (quoteStart != std::string_view::npos) {
+            // Print prefix (e.g. "ERROR: 0:9: ") and message, discarding everything after.
+            out << line.substr(0, quoteStart);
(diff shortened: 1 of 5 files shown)


libs/filamat/src/sca/GLSLTools.cpp
```diff
@@ -45,6 +47,52 @@ GLSLangCleaner::~GLSLangCleaner() {
...
+                    if (c == '_') c = ' ';
+                }
+                
+                size_t quoteStart = line.rfind('\'', startPos);
+                if (quoteStart != std::string_view::npos) {
+                    // Print prefix (e.g. "ERROR: 0:9: ") and message, discarding everything after
+                    out << line.substr(0, quoteStart) << message;
+                } else {
+                    out << line.substr(0, startPos) << message;
+                }
```
show50726: We can do something like:
```
size_t quoteStart = line.rfind('\'', startPos);
out << line.substr(0, quoteStart != std::string_view::npos ? quoteStart : startPos);

for (char c : rawMsg) {
    out << (c == '_' ? ' ' : c);
}
```
to prevent the allocation of `utils::CString message(rawMsg)`

libs/filamat/src/sca/GLSLTools.cpp
```diff
@@ -45,6 +47,52 @@ GLSLangCleaner::~GLSLangCleaner() {
 }
 
 // ------------------------------------------------------------------------------------------------
+// If the error message is associated with API level, this formats it to be more human-readable.
+static void formatApiLevelErrorLog(utils::io::ostream& out, const char* log) {
+    if (!log) return;
+    
+    std::string_view fullLog(log);
+    size_t start = 0;
+    while (start < fullLog.length()) {
```
show50726, line 56: There are many nested structures inside this loop. Is it possible to apply early return in some places to improve the readability?

show50726 (commented):
I wonder if we really need custom messages here? I feel that most of the time, we just want to provide the API version for unstable functions and apply a common error message template.

libs/filamat/src/sca/GLSLTools.cpp
trie94: yup sounds good. i updated the function. PTAL

trie94:
> I wonder if we really need custom messages here? I feel that most of the time, we just want to provide the API version for unstable functions and apply a common error message template.

yeah good point. i made some changes so that it follows this template
```
#define {myFunctionName}  ERROR_{myFunctionName} _api_level_{N}_END

// example
#define getWorldRayFromClip ERROR_getWorldRayFromClip_api_level_2_END
```

and it replaces with the same error message. also updated the pr description.

poweifeng (approved):
lgtm but could you please wait for @pixelflinger 's review?  

I'm not sure about moving this block

```
highp mat4 getWorldFromModelMatrix() {
    return object_uniforms_worldFromModelMatrix;
}
```

show50726 (approved):

libs/filamat/src/sca/GLSLTools.cpp
```diff
@@ -45,6 +47,65 @@ GLSLangCleaner::~GLSLangCleaner() {
...
+
+        // Check if this is a templated api level error.
+        constexpr std::string_view apiLevelMarker = "_api_level_";
+        size_t apiLevelPos = rawMsg.find(apiLevelMarker);
+
+        if (apiLevelPos != std::string_view::npos) {
+            std::string_view symbol = rawMsg.substr(0, apiLevelPos);
+            std::string_view level = rawMsg.substr(apiLevelPos + apiLevelMarker.length());
+            // Format using the template (single quotes around the symbol name).
+            out << "\'" << symbol << "\'" << " requires api level " << level;
```
z3moon: nit: you wouldn't need to use an escape here. maybe `"'"` is enough?

z3moon (approved):

libs/filamat/src/sca/GLSLTools.cpp
trie94: done

pixelflinger (approved):

PR #10295, opened 2026-08-12
Fix JNI reference leaks

### LINKED ISSUE #10286: Android JNI bindings: `new[]` freed with `delete`, and several loops that abandon local references
Files:

- `android/gltfio-android/src/main/cpp/AssetLoader.cpp`
- `android/gltfio-android/src/main/cpp/FilamentAsset.cpp`
- `android/gltfio-android/src/main/cpp/FilamentInstance.cpp`
- `android/filament-android/src/main/cpp/Material.cpp`

## 1. `JavaMaterialProvider::mPreviousMaterials` is allocated with `new[]` and freed with `delete`

`AssetLoader.cpp`, `JavaMaterialProvider`:

```cpp
mutable Material** mPreviousMaterials = nullptr;          // line 39
```

allocated as an array:

```cpp
mPreviousMaterials = new MaterialPointer[count];          // line 197
```

but freed with the scalar form at both of its two release sites:

```cpp
~JavaMaterialProvider() override {
    mEnv->DeleteGlobalRef(mMaterialKeyClass);
    mEnv->DeleteGlobalRef(mJavaProvider);
    delete mPreviousMaterials;                            // line 94
}
```

```cpp
const Material* const* getMaterials() const noexcept override {
    ...
    delete mPreviousMaterials;                            // line 195
    using MaterialPointer = Material*;
    mPreviousMaterials = new MaterialPointer[count];
```

Mismatching `new[]` with `delete` is undefined behaviour. On the common implementations the array cookie sits before the returned pointer, so the scalar `delete` hands the allocator the wrong block address — in practice a heap corruption or an allocator abort rather than a benign leak. `getMaterials()` is called repeatedly while an asset is being loaded, so line 195 runs on every provider callback, not just at teardown.

Fix: `delete[] mPreviousMaterials;` at both sites.

## 2. `getMaterials()` and `getMaterialsCount()` abandon the references they obtain

```cpp
const Material* const* getMaterials() const noexcept override {
    jobjectArray javaMaterials = (jobjectArray) mEnv->CallObjectMethod(mJavaProvider, mGetMaterials);

    const size_t count = mEnv->GetArrayLength(javaMaterials);

    delete mPreviousMaterials;
    using MaterialPointer = Material*;
    mPreviousMaterials = new MaterialPointer[count];

    // Call "getNativeObject" on each material in order to pass them up into the native layer.
    for (size_t i = 0; i < count; ++i) {
        jobject javaMaterial = mEnv->GetObjectArrayElement(javaMaterials, i);
        jlong matPointer = mEnv->CallLongMethod(javaMaterial, mMaterialGetNativeObject);
        mPreviousMaterials[i] = (Material*) matPointer;
    }

    return mPreviousMaterials;
}
```

`javaMaterial` is only used to read a native pointer out of the Java object, but the reference is kept for the rest of the enclosing JNI frame. `javaMaterials`, returned by `CallObjectMethod()`, is never deleted either, and the sibling `getMaterialsCount()` a few lines below creates and abandons another one on every call.

This is a material provider callback invoked during asset loading rather than a JNI entry point, so the frame these references belong to is whichever native method drove the load, not this function. That makes it the site where the accumulation is least visible: repeated provider calls during one load all charge references to the same outer frame.

`javaMaterials` is also used without a null check — if the Java provider returns null or throws, `GetArrayLength()` is called on NULL.

Fix:

```cpp
for (size_t i = 0; i < count; ++i) {
    jobject javaMaterial = mEnv->GetObjectArrayElement(javaMaterials, i);
    jlong matPointer = mEnv->CallLongMethod(javaMaterial, mMaterialGetNativeObject);
    mPreviousMaterials[i] = (Material*) matPointer;
    mEnv->DeleteLocalRef(javaMaterial);
}
mEnv->DeleteLocalRef(javaMaterials);
```

and the same deletion in `getMaterialsCount()`.

## 3. Four string-array bindings pass an inline `NewStringUTF()` into `SetObjectArrayElement()`

`SetObjectArrayElement()` stores the string into the array but does not consume the local reference, so one reference stays live per element until the native method returns.

`FilamentAsset.nGetResourceUris` (line 231):

```cpp
auto resourceUris = asset->getResourceUris();
for (int i = 0; i < asset->getResourceUriCount(); ++i) {
    env->SetObjectArrayElement(result, (jsize) i, env->NewStringUTF(resourceUris[i]));
}
```

`FilamentAsset.nGetMorphTargetNames` (line 250):

```cpp
for (int i = 0, n = asset->getMorphTargetCountAt(entity); i < n; ++i) {
    const char* name = asset->getMorphTargetNameAt(entity, i);
    env->SetObjectArrayElement(result, (jsize) i, env->NewStringUTF(name));
}
```

`FilamentInstance.nGetMaterialVariantNames` (line 79) and `FilamentInstance.nGetSkinNames` (line 135) have the same shape.

The return value of `NewStringUTF()` is unchecked in all four; on failure a NULL is stored and the loop continues with a pending exception.

Every one of these loops is bounded by asset content rather than by a fixed number — resource URI count, morph target count, material variant count, skin count — so a large glTF file drives the reference count directly, well past the 16 slots the JNI specification guarantees without an `EnsureLocalCapacity()` call.

Fix:

```cpp
for (int i = 0; i < asset->getResourceUriCount(); ++i) {
    jstring uri = env->NewStringUTF(resourceUris[i]);
    if (uri == nullptr) {
        return;  // exception already pending
    }
    env->SetObjectArrayElement(result, (jsize) i, uri);
    env->DeleteLocalRef(uri);
}
```

## 4. `Material.nGetParameters()` abandons the parameter name strings

`Material.cpp` (line 257):

```cpp
for (size_t i = 0; i < received; i++) {
    jint type;
    ...
    env->CallStaticVoidMethod(
            parameterClass, parameterAdd,
            parameters, env->NewStringUTF(info[i].name), type, (jint) info[i].precision,
            (jint) info[i].count);
}

env->DeleteLocalRef(parameterClass);
```

`CallStaticVoidMethod()` does not consume the local reference, so the parameter name created in each iteration stays live for the rest of the call. The loop bound is the material's parameter count.

The same function deletes `parameterClass` immediately after the loop, which shows the intended discipline; only the per-iteration strings are missed.

Fix:

```cpp
jstring name = env->NewStringUTF(info[i].name);
env->CallStaticVoidMethod(
        parameterClass, parameterAdd,
        parameters, name, type, (jint) info[i].precision, (jint) info[i].count);
env->DeleteLocalRef(name);
```

romainguy:
Some of the JNI bindings were not releasing local references properly.

Fixes #10286.

### DIFF
--- a/android/gltfio-android/src/main/cpp/FilamentAsset.cpp
+++ b/android/gltfio-android/src/main/cpp/FilamentAsset.cpp
@@ -249,8 +249,10 @@ Java_com_google_android_filament_gltfio_FilamentAsset_nGetMorphTargetNames(JNIEn
     FilamentAsset* asset = (FilamentAsset*) nativeAsset;
     Entity entity = Entity::import(entityId);
     for (size_t i = 0, n = asset->getMorphTargetCountAt(entity); i < n; ++i) {
-        const char* name = asset->getMorphTargetNameAt(entity, i);
-        env->SetObjectArrayElement(result, (jsize) i, env->NewStringUTF(name));
+        jstring name = env->NewStringUTF(asset->getMorphTargetNameAt(entity, i));
+        if (name == nullptr) return;
+        env->SetObjectArrayElement(result, (jsize) i, name);
+        env->DeleteLocalRef(name);
     }
 }
 
--- a/android/gltfio-android/src/main/cpp/FilamentAsset.cpp
+++ b/android/gltfio-android/src/main/cpp/FilamentAsset.cpp
@@ -14,9 +14,9 @@
  * limitations under the License.
  */
 
-#include <jni.h>
-
 #include <gltfio/FilamentAsset.h>
+
+#include <jni.h>
 
 using namespace filament;
 using namespace filament::math;
--- a/android/gltfio-android/src/main/cpp/FilamentAsset.cpp
+++ b/android/gltfio-android/src/main/cpp/FilamentAsset.cpp
@@ -228,7 +228,10 @@ Java_com_google_android_filament_gltfio_FilamentAsset_nGetResourceUris(JNIEnv* e
     FilamentAsset* asset = (FilamentAsset*) nativeAsset;
     auto resourceUris = asset->getResourceUris();
     for (int i = 0; i < asset->getResourceUriCount(); ++i) {
-        env->SetObjectArrayElement(result, (jsize) i, env->NewStringUTF(resourceUris[i]));
+        jstring uri = env->NewStringUTF(resourceUris[i]);
+        if (uri == nullptr) return;
+        env->SetObjectArrayElement(result, (jsize) i, uri);
+        env->DeleteLocalRef(uri);
     }
 }
 
@@ -245,7 +248,7 @@ Java_com_google_android_filament_gltfio_FilamentAsset_nGetMorphTargetNames(JNIEn
         jlong nativeAsset, jint entityId, jobjectArray result) {
     FilamentAsset* asset = (FilamentAsset*) nativeAsset;
     Entity entity = Entity::import(entityId);
-    for (int i = 0, n = asset->getMorphTargetCountAt(entity); i < n; ++i) {
+    for (size_t i = 0, n = asset->getMorphTargetCountAt(entity); i < n; ++i) {
         const char* name = asset->getMorphTargetNameAt(entity, i);
         env->SetObjectArrayElement(result, (jsize) i, env->NewStringUTF(name));
     }
--- a/android/filament-android/src/main/cpp/Material.cpp
+++ b/android/filament-android/src/main/cpp/Material.cpp
@@ -264,10 +264,10 @@ Java_com_google_android_filament_Material_nGetParameters(JNIEnv* env, jclass,
             type = (jint) info[i].type;
         }
 
-        env->CallStaticVoidMethod(
-                parameterClass, parameterAdd,
-                parameters, env->NewStringUTF(info[i].name), type, (jint) info[i].precision,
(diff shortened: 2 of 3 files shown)


android/gltfio-android/src/main/cpp/FilamentAsset.cpp
```diff
@@ -245,7 +248,7 @@ Java_com_google_android_filament_gltfio_FilamentAsset_nGetMorphTargetNames(JNIEn
         jlong nativeAsset, jint entityId, jobjectArray result) {
     FilamentAsset* asset = (FilamentAsset*) nativeAsset;
     Entity entity = Entity::import(entityId);
-    for (int i = 0, n = asset->getMorphTargetCountAt(entity); i < n; ++i) {
+    for (size_t i = 0, n = asset->getMorphTargetCountAt(entity); i < n; ++i) {
         const char* name = asset->getMorphTargetNameAt(entity, i);
         env->SetObjectArrayElement(result, (jsize) i, env->NewStringUTF(name));
```
z3moon: here's another `NewStringUTF` call being leaked as well.

z3moon (approved):

android/gltfio-android/src/main/cpp/FilamentAsset.cpp
romainguy: Ah, good catch! I'll update the PR.

pixelflinger (approved):

romainguy:
@pixelflinger I fixed the CI issue

PR #10302, opened 2026-08-13
Add a second roughness to materials

romainguy:
This feature allows to enable a second specular lobe to lit materials. This can be used to recreate more complex material appearances, including metals with a patina exhibiting both glossy and rough reflections at the same time. It is in general
useful to create hazy appearances.

Materials can now set the following two properties:
- secondRoughness: works like the existing roughness property, but for a 2nd lobe
- secondRoughnessWeight: used to blend the 2 specular lobes

Example of a regular glossy metallic ball (1 specular lobe):
single_lobe

With a second, slightly rough, specular lobe:
two_lobes_1

And the same configuration, with more weight given to the second lobe:
two_lobes_2

See https://www.youtube.com/watch?v=uz8PIi3ELJg for more examples of when this feature is useful.


pixelflinger (approved):

pixelflinger:
@trie94 for the API version requirements. Do we need to guard those new attributes?

shaders/src/surface_material_inputs.fs
```diff
@@ -125,6 +138,11 @@ void initMaterial(out MaterialInputs material) {
 #endif
     material.emissive = vec4(vec3(0.0), 1.0);
 
+#if defined(MATERIAL_HAS_SECOND_SPECULAR_LOBE)
+    material.secondRoughness = 0.0f;
+    material.secondRoughnessWeight = 0.0;
+#endif
```
trie94, line 149: i think we need to guard those two params with api level:
```
#if defined(MATERIAL_HAS_SECOND_SPECULAR_LOBE)
#if CLIENT_MATERIAL_API_LEVEL >= 2
    float secondRoughness;
    float secondRoughnessWeight;
#else
#define secondRoughness ERROR_secondRoughness_api_level_2_END
#define secondRoughnessWeight ERROR_secondRoughnessWeight_api_level_2_END
#endif
#endif
```

trie94 (changes requested):

trie94:
> @trie94 for the API version requirements. Do we need to guard those new attributes?

yeah i left a comment there. i think those are the only ones that are public?

you can test whether it works on apiLevel 2 and fails on apiLevel 1 by using `--api-level` flag in matc.
e.g.
```
MATC --api-level 2 -o <output.cmat> <source.mat> // should pass
MATC --api-level 1 -o <output.cmat> <source.mat> // should fail
```

romainguy:
@trie94 Do we need to increase the materials API level with this change as well then? (I'll also update the materials doc to mention the API level, and add to the release notes too)

trie94:
> @trie94 Do we need to increase the materials API level with this change as well then? (I'll also update the materials doc to mention the API level, and add to the release notes too)

no actually we should discuss this further with Android XR team. ever since we introduced apiLevel (it was mainly because of Android XR as we are using remote renderer), we haven't bumped up the level and we haven't established the process yet

romainguy:
@trie94 I've updated the code and docs. One thing to note, defaulting to apiLevel 1 will be very inconvenient for anybody who doesn't do remote rendering. Wouldn't it make more sense to default the API level to whatever the latest version `matc` supports, and instead make folks like XR do the pinning to a given version?

shaders/src/surface_shading_lit.fs
```diff
@@ -211,12 +221,25 @@ void getRoughnessPixelParams(const MaterialInputs material, inout PixelParams pi
...
     // Clamp the roughness to a minimum value to avoid divisions by 0 during lighting
     pixel.perceptualRoughness = clamp(perceptualRoughness, MIN_PERCEPTUAL_ROUGHNESS, 1.0);
     // Remaps the roughness to a perceptually linear roughness (roughness^2)
     pixel.roughness = perceptualRoughnessToRoughness(pixel.perceptualRoughness);
+
+#if defined(MATERIAL_HAS_SECOND_SPECULAR_LOBE)
+    pixel.secondPerceptualRoughness =
+            clamp(secondPerceptualRoughness, MIN_PERCEPTUAL_ROUGHNESS, 1.0);
+    pixel.secondRoughness =
+            clamp(pixel.secondPerceptualRoughness, MIN_PERCEPTUAL_ROUGHNESS, 1.0);
```
show50726: should this be remapped with `perceptualRoughnessToRoughness`?

trie94:
> Wouldn't it make more sense to default the API level to whatever the latest version `matc` supports, and instead make folks like XR do the pinning to a given version?

yeah when we first introduced this, we decided to default to 1 because we wanted to be more conservative. it makes sense to me that we could make matc to default to the latest api level it supports. but that wouldn't really change the current behavior because it'd be 1.
or are you referring to unstable api level? (default to 2?). if that's the case i don't think that's what we want, because we use unstable api when an api is under dev as well.

shaders/src/surface_shading_lit.fs
romainguy: Oof great catch, yes this is a bad copy/paste. I'll fix it.

romainguy:
> or are you referring to unstable api level? (default to 2?). if that's the case i don't think that's what we want, because we use unstable api when an api is under dev as well.

I agree that we wouldn't want the unstable API level to be the default, but when the API level for materials officially revs to 2 (which could happen now while still keeping existing unstable APIs unstable), then the default should be 2.

trie94:
> stable), then the default should be 2.

yeah sounds good to me.

trie94 (approved):

Issue #9341, opened 2025-10-21
Picking broken on Samsung S24 with Vulkan 1.3.279 in Filament 1.66


LionelMyFit:
**Describe the bug**
Picking renderables does not work on the Samsung Galaxy S24 when using Vulkan 1.3.279. The same application works correctly on the same device with Vulkan 1.3.231.

**To Reproduce**

Use Filament 1.66 on an Android project (also reproduced on all versions above 1.53)

Run the app on a Samsung Galaxy S24 with Vulkan driver 1.3.279.

Try to perform object picking (e.g., raycasting or tap selection).

Observe that picking does not register correctly, the picking returns always ID 0

**Expected behavior**
Object picking should work as it does on Vulkan 1.3.231 on the same device.

**Screenshots**
N/A (the issue is functional, not visual).

**Logs**
No relevant logs

**Smartphone:**
 - Device: Samsung Galaxy S24
 - OS: Android 15

**Additional context**
The issue only occurs with Vulkan driver version 1.3.279.

The issue does not occur with Vulkan 1.3.231 on the same device.

Picking works correctly on other devices tested with Vulkan 1.3.279.

bennysway:
Im using ARCore + Sceneview. The filament engine reported:
12:58:59.733  V  [Samsung Electronics Co., Ltd.], [ANGLE (Samsung Xclipse 540) on Vulkan 1.3.279], [OpenGL ES 3.2 Driver version: 24.1.265 git hash: 24f47ea4b93e], [OpenGL ES GLSL ES 3.20 (ANGLE git hash: 24f47ea4b93e)]

And Im currently using 1.52.0 so I can confirm its a vulkan issue. All my other devices and other samsung versions work, except for A56 with this vulkan version.

LionelMyFit:
@bennysway Thanks for this response, it confirms my analysis

bennysway:
Well I hope this help you @LionelMyFit , I was sampling textures for shader data (vertex, texture, texelFetch etc). Changing my Sampler to min = NEAREST, max = NEAREST; made it work.
```
setParameter(
                    "productData",
                    productDataTexture!!,
                    TextureSampler(
                        TextureSampler.MinFilter.NEAREST,
                        TextureSampler.MagFilter.NEAREST,
                        TextureSampler.WrapMode.CLAMP_TO_EDGE,
                    ).apply {
                        anisotropy = 1.0f
                    },
                )
```

[...]

Issue #7965, opened 2024-07-08
View.pick() returns an invalid @Entity result on some devices


OwnLeiX:
⚠️ **Issues not using this template will be systematically closed.**

**Describe the bug**
View.pick() returns an invalid @Entity result on some devices, and it can work on other devices, on all devices, driver.getFeatureLevel() != FeatureLevel::FEATURE_LEVEL_0;

**To Reproduce**
1. run sample gltf viewer;
2. tap the model;
3. watch the log.

**Logs**
2024-07-08 11:02:03.340 15304-15304 Filament                com.google.android.filament.gltf     V  Picked 1771841328: null
2024-07-08 11:02:07.634 15304-15304 Filament                com.google.android.filament.gltf     V  Picked -695229381: null
2024-07-08 11:02:10.115 15304-15304 Filament                com.google.android.filament.gltf     V  Picked -15948299: null
2024-07-08 11:02:12.240 15304-15304 Filament                com.google.android.filament.gltf     V  Picked -2007344550: null

**Smartphone (please complete the following information):**
 - Device: vivo 1902 & vivo 2019 & cph2387 & cph2269 & cph2387 & v2134 & v2111 & cph2477
 - OS: Android 9 ~ Android 14

pixelflinger:
This might be fixed by #8917 . Please reopen if it is not.

PR #10303, opened 2026-08-13
add vertex_objectId varying to fix picking failures on certain Android/GLES drivers

nmfisher:
Various users have reported picking failures on Android devices (see #8762, #9341, #7965, #7638). 

One failure mode was addressed in #8917, however on my PowerVR device with Filament v1.75.0 and the GLES backend, every call to `pick` still returns 0 when the feature level is greater than 0. When the engine feature level is set to 0, then `pick` correctly returns the renderable entity.

(Everything here uses basic geometry with no material, so the default Filament material is applied). 
 
I've narrowed the problem to the index used to access the object-uniform array in the picking fragment shader. A logical index of 0 used as a dynamic array index produced a different result from using the equivalent constant index. This suggests a driver bug. Moving the lookup to the vertex shader and forwarding through a flat integer `varying` seems to reliably return the correct result on my device.

This draft PR contains a proposed fix for `TARGET_GLES_ENVIRONMENT`: 
- in the depth vertex shader (where object-uniform indexing already works), read the instance object ID  
- forward the ID through a flat varying for picking when feature level is greater than 0
- initialize `logical_instance_index` in the depth fragment shader (matches `surface_main.fs`)

The varying is declared for every GLES depth vertex shader when feature level > 0, but `vertex_objectId` is only read in the fragment shader when `CONFIG_POWER_VR_SHADER_WORKAROUNDS` is true (when false, falls back to the original UBO lookup. This leaves the feature-level-0 RGBA8 picking path unchanged. This reuses location 11 (also used by `vertex_lightSpacePosition` for shadow-receiving surface variants), but the two variants are mutually exclusive so this should not conflict.

Leaving this as draft because I haven't tested it thoroughly, and I'm not sure this is even the best approach.

### DIFF
--- a/NEW_RELEASE_NOTES.md
+++ b/NEW_RELEASE_NOTES.md
@@ -11,6 +11,8 @@ appropriate header in [RELEASE_NOTES.md](./RELEASE_NOTES.md).
   a completion callback must check it [⚠️ **API Change**]
 - backend: canceling an async call now always invokes its completion callback, with
   `AsyncCallStatus::CANCELED`. Previously the callback was silently dropped
+- engine: fix picking on PowerVR OpenGL ES drivers that mishandle dynamic uniform-buffer indexing
+  in fragment shaders [⚠️ **Recompile Materials**]
 - engine: support multiple directional lights, opt-in via
   `Engine::Config::enableMultipleDirectionalLights`; the dominant one still provides shadows and
   the sun disc, up to 4 additional directional lights are evaluated without shadows [⚠️ **New
--- a/shaders/src/surface_depth_main.fs
+++ b/shaders/src/surface_depth_main.fs
@@ -20,7 +20,9 @@ highp vec4 computeDepthMomentsVSM(const highp float depth);
 
 void main() {
     filament_lodBias = frameUniforms.lodBias;
-
+#if defined(FILAMENT_HAS_FEATURE_INSTANCING)
+    logical_instance_index = instance_index;
+#endif
     initObjectUniforms();
 
 #if defined(MATERIAL_HAS_CUSTOM_DEPTH) || defined(BLEND_MODE_MASKED) || ((defined(BLEND_MODE_TRANSPARENT) || defined(BLEND_MODE_FADE)) && defined(MATERIAL_HAS_TRANSPARENT_SHADOW))


olenev:
I had already made exactly the change you describe in your third point — initializing `logical_instance_index` in the depth fragment shader — a while back. In my case that was the only change needed and pick started returning the correct entity.

I built the library with that single change on top of v1.71.1 and have been shipping it in our commercial live app since then, it's been exercised by real users — picking has been working correctly for them.
https://github.com/olenev/filament/commit/4d02366800cad65089adb4e3142ba4695f1a2f06

shaders/src/surface_varyings.glsl
```diff
@@ -32,6 +32,12 @@ LAYOUT_LOCATION(10) flat VARYING highp int instance_index;
 highp int logical_instance_index;
 #endif
 
+#if defined(VARIANT_DEPTH) && MATERIAL_FEATURE_LEVEL > 0 && defined(TARGET_GLES_ENVIRONMENT)
+// Some PowerVR OpenGL ES drivers mishandle dynamic object-uniform indexing in fragment shaders.
+// Depth vertex shaders already perform the same lookup, so forward the object ID for picking.
+LAYOUT_LOCATION(11) flat VARYING highp int vertex_objectId;
```
pixelflinger: wouldn't this collide with the `LAYOUT_LOCATION(11)` below?

in addition, this change is worrisome because it burns a varying, and we're very limited already.

pixelflinger (commented):
Is this always a problem on PowerVR, or only when instancing is used?  If it was the later, we could just disable instancing instead.

Are you sure that the `logical_instance_index` fix is not enough? It seems like this was maybe a real bug.

shaders/src/surface_varyings.glsl
nmfisher: On the official `gltf-viewer` apk, picking never shows any text, so I believe the bug exists no matter whether automatic/manual instancing is enabled (`FILAMENT_HAS_FEATURE_INSTANCING` is true for feature levels greater than 0).

I just tried a stripped-down version with just the `logical_instance_index` fix, and yes this seems to work. As @olenev indicated, this is probably enough. Should I update the PR to strip it down to just that line?

pixelflinger: yes definitely! thanks!

nmfisher: done

NEW_RELEASE_NOTES.md
```diff
@@ -11,6 +11,8 @@ appropriate header in [RELEASE_NOTES.md](./RELEASE_NOTES.md).
   a completion callback must check it [⚠️ **API Change**]
 - backend: canceling an async call now always invokes its completion callback, with
   `AsyncCallStatus::CANCELED`. Previously the callback was silently dropped
+- engine: fix picking on PowerVR OpenGL ES drivers that mishandle dynamic uniform-buffer indexing
```
z3moon, line 14: it looks like this fix helps all gl drivers using instanced picking. maybe consider removing PowerVR

z3moon (approved):

PR #10306, opened 2026-08-14
backend: join the worker thread in ~ThreadWorker()

z3moon:
ThreadWorker's destructor was defaulted, so mThread was only ever joined by an explicit terminate(). OpenGLDriver::terminate() does call it, so the happy path was fine, but any path that destroys the worker without going through terminate() -- initialization failure, an exception, a future refactor -- would hit std::thread's destructor with a joinable thread and abort the process. The thread body also captures `this` and dereferences mConfig and mQueue, so not joining is a use-after-free as well.

Call terminate() from the destructor. It is already idempotent. JobQueue::stop() and the joinable() check are both safe to repeat.

### DIFF
--- a/filament/backend/src/JobQueue.cpp
+++ b/filament/backend/src/JobQueue.cpp
@@ -247,7 +247,10 @@ ThreadWorker::ThreadWorker(JobQueue::Ptr queue, Config config, PassKey)
     });
 }
 
-ThreadWorker::~ThreadWorker() = default;
+ThreadWorker::~ThreadWorker() {
+    // Destroying a worker without calling `terminate()` first is a programming error.
+    assert_invariant(!mThread.joinable());
+}
 
 void ThreadWorker::terminate() {
     JobWorker::terminate();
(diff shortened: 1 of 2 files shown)


poweifeng (approved):

filament/backend/src/JobQueue.cpp
```diff
@@ -247,7 +247,12 @@ ThreadWorker::ThreadWorker(JobQueue::Ptr queue, Config config, PassKey)
     });
 }
 
-ThreadWorker::~ThreadWorker() = default;
+ThreadWorker::~ThreadWorker() {
+    // The thread body dereferences `this` (`mConfig` and `mQueue`), so it must be joined before we
+    // return, even if the owner never called `terminate()` explicitly (e.g. initialization failure
+    // or an exception). `terminate()` is idempotent, so calling it twice is safe.
+    ThreadWorker::terminate();
```
pixelflinger: it's an error to not call terminate().  I would prefer we assert in the dtor instead. this basically should never happen. blocking in dtors can be problematic, as well as potentially throwing exceptions.

pixelflinger (changes requested):

filament/backend/src/JobQueue.cpp
z3moon: Agree, it's better. Done.

pixelflinger (approved):

PR #10307, opened 2026-08-14
fix: [smol-v] bound decoded output writes

carlos-al:
## Summary

  This PR fixes a heap-buffer overflow in Filament’s bundled SMOL-V decoder.

  The change is submitted to Filament because the affected decoder is vendored directly into this repository, is consumed by Filament’s material-package parser, and has diverged from the standalone SMOL-V source. The standalone project
  has also been inactive since September 2024.

  ## Analysis

  `DictionaryReader` reads the decoded SPIR-V size from an untrusted material package, allocates a buffer of exactly that size, and passes it to the bundled `smolv::Decode()` implementation.

  The decoder trusted the declared size when writing the fixed five-word SPIR-V header and subsequent instruction words. Decoded sizes smaller than the header, non-word-aligned sizes, or instruction streams exceeding the declared output
  capacity could therefore write beyond Filament’s allocation.

  This is directly reachable through Filament’s material-package APIs, including `Material::Builder::package(...).build(...)`. The regression coverage consequently should belong at Filament’s `DictionaryReader` integration boundary, in
  addition to the direct decoder checks.

  ## Why is  this change is in Filament (and not upstream)?

  Filament maintains and builds its own bundled copy at `third_party/smol-v/source/smolv.cpp`; it does not consume SMOL-V as a live external dependency.

  Filament maintains a non-identical bundled copy of SMOL-V and directly exposes it through its material-package parser, and `DictionaryReader` provides the concrete allocation and parsing path affected by the bug. Fixing only the inactive standalone repository would neither update Filament’s copy nor add coverage for Filament’s vulnerable integration.

  This PR therefore repairs the copy that Filament actually ships and adds regression coverage at the production consumer boundary.

  ## The Fix

  Reject decoded sizes below the SPIR-V header size or not aligned to a complete word.

  Track the output boundary and verify capacity before every decoded write.

  ## Testing

  - Added coverage for decoded sizes `0`, `1`, `4`, `19`, `20`, and `21`
  - Added a stream whose header fits but whose instruction exceeds the declared output capacity
  - Added `DictionaryReader` integration coverage for invalid decoded sizes
  - Confirmed a genuine repository shader still round-trips through SMOL-V
  - Ran the focused ASan+UBSan tests: 3 tests passed
  - Ran the complete owning ASan+UBSan suite: 8 tests passed

### DIFF
--- a/libs/filaflat/tests/test_filaflat.cpp
+++ b/libs/filaflat/tests/test_filaflat.cpp
@@ -21,6 +21,7 @@
 #include <filaflat/MaterialChunk.h>
 #include <filaflat/Unflattener.h>
 #include <filament/MaterialChunkType.h>
+#include <smolv.h>
 
 #include <vector>
 #include <cstdint>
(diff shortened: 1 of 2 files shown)


romainguy:
smol-v is a 3rt party library, the fix should go upstream. It's fine to fix Filament's local copy, but if the fix doesn't make it upstream, taking any future version of smol-v might mean losing the fix.

pixelflinger:
yes I prefer this to go upstream.

PR #10309, opened 2026-08-14
gl: remove the ANGLE-D3D11 spec-constant UBO array-size workaround

Erkko68:
`#10077` added this workaround for ANGLE-on-D3D11 not folding a spec-constant-initialised
`const int` into a uniform-block array length: `ObjectUniforms` was laid out at matc's baked
default (`CONFIG_MAX_INSTANCES` = 64) instead of the 8 the engine binds on web, so instanced draws
failed WebGL2 validation with "uniform buffer too small" and materials rendered black.

**It never actually applied on WebGL.** The flag is set by sniffing `GL_RENDERER` for `"ANGLE"`,
but WebGL masks that string — Chrome reports `WebKit WebGL`, Firefox `Mozilla`. The unmasked value
is only reachable through `WEBGL_debug_renderer_info` (`UNMASKED_RENDERER_WEBGL`), which the GL
backend never queries, so `bugs.spec_constant_array_size_not_folded` is never set in a browser.
The engine's own startup log on Chrome/Windows shows the workaround list empty:

```
[WebKit], [WebKit WebGL], [OpenGL ES 3.0 (WebGL 2.0 (OpenGL ES 3.0 Chromium))], ...
Feature level: 1
Active workarounds:
```

Forcing the flag on in a test build makes `spec_constant_array_size_not_folded` appear in that
list and the rewrite run, so the reporting works — the flag genuinely is never set.

**What looked like a fix was the DYN rollback.** The timing (UTC):

| when | |
| --- | --- |
| 2026-06-01 14:17 | #10023 replaces the DYN variant with dynamic specialization constants |
| 2026-06-01 15:16 | #10075 reverts it, 59 minutes later |
| 2026-06-02 18:34 | #10077 — this workaround — merges, into a tree without the DYN change |
| 2026-06-25 05:22 | #10095 reintroduces it |

The workaround was validated during the window where #10023 was rolled back, and the black
rendering came back once #10095 landed. It was never what fixed it. On current `main`, lit content
renders correctly on Windows in both Chrome and Firefox with this workaround still inactive.

**The driver bug no longer reproduces either.** A standalone WebGL2 page declaring the identical
block and reading back `UNIFORM_BLOCK_DATA_SIZE` reports the correct size, matching a hardcoded
`data[8]` control:

| probe | reported |
| --- | --- |
| `data[8]` (control) | 2048 |
| `data[CONFIG_MAX_INSTANCES]` | 2048 |

on `ANGLE (NVIDIA, NVIDIA GeForce GTX 1650 (0x00002188) Direct3D11 vs_5_0 ps_5_0, D3D11)`. That is
one GPU/driver/browser combination rather than proof for every ANGLE configuration, but since the
workaround cannot fire on WebGL at all, removing it changes nothing there; if a mis-folding
configuration turns up, restoring it is a small patch.

Refs #10077, and fe3790cb9 (#5859) which it replaced.


poweifeng:
@Erkko68 , why is this removal necessary?

pixelflinger:
where is the evidence the bug doesn't repro?

Erkko68:
Sorry — I've updated the PR description. I originally pushed this workaround and have been tracking it since: it regressed again in 1.74.0–1.75.0, where lit content still rendered black on Windows.

image

With the recent DYN variant removal work it renders correctly again:

image

As the description now covers, the fix never actually landed on WebGL — GL_RENDERER is masked there, so the workaround never engaged. What I was tracking was really the DYN work being rolled back (#10075) and later reintroduced (#10095), plus the changes that swept in between.

poweifeng (approved):

Erkko68:
@poweifeng Can you add the internal label?

pixelflinger (approved):

PR #10310, opened 2026-08-14
Update the material doc with the recent api level changes

trie94:
FIXES=540482735


romainguy:
@trie94 Check out my change on second roughness, this will conflict a bit with that.

trie94:
> @trie94 Check out my change on second roughness, this will conflict a bit with that.

yeah i can merge after yours

show50726 (approved):

PR #10311, opened 2026-08-14
VK: Fix memcpy shared buffers when its allowed.

rafadevai:
Before the check for if a buffer static was wrong
and never activated because its not a bit flag.


poweifeng (approved):

pixelflinger (approved):

rafadevai:
I always went with the assumption that static buffers cant be updated more than once, but seems like its not true. So I just updated the test to make sure we use staging when something its in flight.

poweifeng (approved):

PR #10312, opened 2026-08-14
android: add FILAMENT_PLATFORM_API_LEVEL flag

lincolnfrog:
When Filament is built inside an Android system image (e.g., Android 14), __ANDROID_API__ returns 1000 (__ANDROID_API_FUTURE__) despite the target API level actually being 34. This causes Filament's compile-time API checks to assume dynamic symbol lookup (dlsym) for APIs unavailable on the target OS, resulting in runtime issues.

This change introduces a FILAMENT_PLATFORM_API_LEVEL build flag:
- Adds FILAMENT_PLATFORM_API_LEVEL support in CMakeLists.txt.
- Centralizes preprocessor fallback logic in utils/api_level.h so FILAMENT_PLATFORM_API_LEVEL falls back to __ANDROID_API__ if unset.
- Replaces hardcoded __ANDROID_API__ preprocessor checks across Filament platform files (AndroidNdk.h, AndroidNativeWindow.cpp, PlatformEGLAndroid.cpp, ashmem.cpp, CallStack.cpp) with FILAMENT_PLATFORM_API_LEVEL.


filament/backend/include/backend/platforms/AndroidNdk.h
```diff
@@ -21,14 +21,13 @@
 #include <android/native_window.h>
 
 #include <stdint.h>
+#include <utils/api_level.h>
 
 #define FILAMENT_REQUIRES_API(x) __attribute__((__availability__(android,introduced=x)))
 
-#define FILAMENT_USE_DLSYM(api) (__ANDROID_API__ < (api))
+#define FILAMENT_USE_DLSYM(api) (FILAMENT_PLATFORM_API_LEVEL < (api))
```
poweifeng: to avoid confusion, I'd rename `FILAMENT_PLATFORM_API_LEVEL` to `FILAMENT_ANDROID_PLATFORM_API_LEVEL`

pixelflinger: oh... good catch Powei!

poweifeng (approved):

PR #10314, opened 2026-08-14
JobSystem optimizations

pixelflinger:
Our goal is to significantly reduce calls into the kernel (i.e. locks when they block, condition's wait and notify). Blocking calls make JobSystem prone to priority inversions in addition to cost a large amount of CPU compared to the actual work done in jobs.

So there are two main set of changes. Some  remove uses of lock altogether by using lockless primitives when possible, another set avoids calls to wakeOne().  Each change has it own small-ish CL.


pixelflinger:
converting back to draft because there are some potential issues

pixelflinger:
too many problems, closing, will update a new one.

PR #10315, opened 2026-08-15
vk: Use FixedCapacityVector for ExtensionSet to avoid std containers

poweifeng:
- Also simplified some of the logic by using std::find_if.
 - Make renderdoc enabled/disabled via bool param


filament/backend/src/vulkan/platform/VulkanPlatform.cpp
```diff
@@ -280,19 +346,19 @@ std::tuple<ExtensionSet, ExtensionSet> pruneExtensions(VkPhysicalDevice device,
...
     if (setContains(newInstExts, VK_EXT_DEBUG_UTILS_EXTENSION_NAME) &&
             setContains(newDeviceExts, VK_EXT_DEBUG_MARKER_EXTENSION_NAME)) {
-        newDeviceExts.erase(VK_EXT_DEBUG_MARKER_EXTENSION_NAME);
+        setErase(newDeviceExts, VK_EXT_DEBUG_MARKER_EXTENSION_NAME);
     }
 
 #if FVK_ENABLED(FVK_DEBUG_VALIDATION)
     // debugMarker must also request debugReport the instance extension. So check if that's present.
     if (setContains(newInstExts, VK_EXT_DEBUG_MARKER_EXTENSION_NAME) &&
             !setContains(newInstExts, VK_EXT_DEBUG_MARKER_EXTENSION_NAME)) {
```
z3moon: You don't need to update it since it's out of scope in this review, but this `if` statement will never be true I think.

z3moon (approved):

rafadevai (approved):

PR #10316, opened 2026-08-15
JobSystem optimizations

pixelflinger:
## Overview & Goals

This change modernizes Filament's JobSystem and parallel_for subsystem to eliminate synchronization bottlenecks, lock contention, memory bloat, and cacheline churn across multicore desktop and mobile architectures.

The primary goals of this overhaul:
1. **Redesign `parallel_for`**: Replace recursive binary-tree job splitting with dynamic, range-based chunk stealing to eliminate job pool exhaustion and allocation overhead.
2. **Lock-Free Concurrency**: Remove all remaining mutex locks from the job lifecycle, job allocation pool, and thread-local state lookups.
3. **Hot-Path Work-Stealing Latency**: Minimize atomic churn, bypass deque contention on immediate continuations, and optimize pseudo-random victim selection.
4. **Mobile Memory & Cache Efficiency**: Dynamically size thread structures so mobile targets only allocate the memory needed for their active core count.

---

### `parallel_for` Execution Time (4,096 Items)
* **macOS (Apple Silicon)**: **36x speedup** (3.71 ms -> 0.10 ms)
* **Android (Pixel 8 Pro ARM64)**: **6.4x speedup** (3.98 ms -> 0.62 ms)
  * CPU cycles: **58x reduction** (3.86 M cycles -> 66.5 k cycles)
  * Instructions: **37x reduction** (6.02 M insn -> 162.7 k insn)

### Work-Stealing Throughput (`BM_JobSystemAsChildren4k`)
* **Android (Pixel 8 Pro ARM64)**: **~30% overall speedup** (5.92 ms -> 4.14 ms)

### Memory & Allocation Reductions
* **99.8% fewer jobs allocated** during parallel loops (4,095 -> 8 jobs for 4K items).
* **Zero heap allocations** for standard `parallel_for` invocations.
* **Reduced memory footprint**: Dynamic thread state structures scale to target thread counts (~160 KB on mobile vs ~1 MB).

### DIFF
--- a/libs/utils/include/private/utils/ThreadMap.h
+++ b/libs/utils/include/private/utils/ThreadMap.h
@@ -44,11 +44,12 @@ namespace utils {
  * association without relying on compiler-dependent thread_local storage (which is
  * problematic on some platforms and dynamic libraries).
  *
- * Sized dynamically at construction time, thread IDs and values are stored in
- * cache-line-packed arrays synchronized via an atomic 3-state slot lifecycle (EMPTY, WRITING, VALID).
- * Lookups perform a linear scan over the capacity. For typical thread counts (e.g. <= 32-64
- * threads), all thread IDs fit in 1-4 L1 cache lines, allowing lookups to complete in ~1-2
- * nanoseconds without any mutex locking or hash table overhead on the critical path.
+ * Sized dynamically at construction time, entries are stored in a contiguous array of
+ * structures (AoS) synchronized via a seqlock version counter and atomic 3-state slot lifecycle
+ * (EMPTY, WRITING, VALID). Lookups perform a linear scan over the capacity. For typical thread
+ * counts (e.g. <= 32-64 threads), the entire map fits in 12-24 L1 cache lines, allowing lookups
+ * to complete in ~1-2 nanoseconds without any mutex locking or hash table overhead on the
+ * critical path.
  *
  * Usage Example:
  * --------------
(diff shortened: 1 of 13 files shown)


libs/utils/include/private/utils/ThreadMap.h
```diff
@@ -0,0 +1,332 @@
...
+#include <type_traits>
+#include <utility>
+
+#include <assert.h>
+#include <stdint.h>
+
+namespace utils {
+
+/**
+ * ThreadMap is a bounded-capacity, lock-free and wait-free associative mapping from
```
z3moon, line 39: This class may be safe to use under a constraint that a thread uses this class with only its own thread id and a fixed slot index. I think it's worth clarifying this.

libs/utils/include/private/utils/ThreadMap.h
```diff
@@ -0,0 +1,332 @@
+/*
+ * Copyright (C) 2026 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
```
z3moon, line 4: do we expect users to use this? maybe it's better to move this to private?

z3moon (approved):

poweifeng (approved):
my ai says good to go :)

pixelflinger:
I'm going to upload a new PR.

PR #10317, opened 2026-08-16
backend: always invoke the completion callback of a canceled async call

z3moon:
Async jobs captured the raw handler/callback/user triplet and only called `scheduleCallback` at the end of the job body, so a job that was destroyed without ever running never notified anyone. That happens whenever an async call is canceled: `Engine::cancelAsyncCall` either erases the job from the queue, or, when it wins the race against the `...R()` half, leaves `push()` to drop the job because its pre-issued id is gone.

Two consequences. The caller can't distinguish a canceled call from one that is still in flight, which breaks any app that counts outstanding operations. And the frontend's heap-allocated `CallbackAdapter` is deleted only by the callback it wraps, so every canceled call leaks it.

Introduce `DriverBase::AsyncCompletion`, which holds the triplet and schedules the callback from its destructor unless the job already did. Jobs capture it instead, so the callback fires exactly once whether the job runs or not. The destructor reports `AsyncCallStatus::CANCELED`, the end of the job body reports `COMPLETED`, both through `scheduleAsyncCallback()`.


filament/backend/src/DriverBase.h
```diff
@@ -238,6 +238,56 @@ class DriverBase : public Driver {
...
+
+        /**
+         * Schedules the completion callback with the given status, unless it has already been
+         * scheduled. Can be called from any thread, `scheduleAsyncCallback` is thread-safe.
+         */
+        void schedule(AsyncCallStatus const status) {
+            if (mCallback) {
+                mDriver->scheduleAsyncCallback(mHandler, mCallback, mUser, status);
+                mCallback = nullptr;
+            }
```
pixelflinger: there is no concurrency risk here?

pixelflinger (approved):

filament/backend/src/DriverBase.h
z3moon: Yeah there's no concurrency risk here because this object is always owned by a lambda and not copyable.
I see the comment is misleading, let me rephrase it.

PR #10320, opened 2026-08-17
Improve FrameGraph memory usage

pixelflinger:
FrameGraph ran out of memory on large scenes, especially those with many shadows. Optimized FrameGraph memory usage so that those scenes could fit without significantly increasing FrameGraph's arena. A particular test scene needed over 2 MiB, but now can work with 512 KiB.

In addition FrameGraph no longer steals space from the per-frame Arena -- like it was before a recent change.

### DIFF
--- a/filament/src/ShadowMapManager.cpp
+++ b/filament/src/ShadowMapManager.cpp
@@ -680,15 +680,15 @@ FrameGraphId<FrameGraphTexture> ShadowMapManager::render(FEngine& engine, FrameG
         // or indirectly via anisotropic filtering.
         // So generate the mipmaps for each layer
         if (UTILS_UNLIKELY(textureRequirements.levels > 1)) {
-            for (size_t level = 0; level < textureRequirements.levels - 1; level++) {
-                // TODO: we could take a list of scissor maybe
-                if (view.hasPCSS()) {
-                    // for PCSS we do a higher quality gaussian mipmap generation
-                    gaussianMipmapPass(engine, fg, prepareShadowPass->shadows, layer, level, textureRequirements.clearColor);
-                } else {
-                    // for EVSM, a regular mipmap is enough
-                    vsmMipmapPass(engine, fg, prepareShadowPass->shadows, layer, level, textureRequirements.clearColor);
-                }
+            // TODO: we could take a list of scissor maybe
+            if (view.hasPCSS()) {
+                // for PCSS we do a higher quality gaussian mipmap generation
+                gaussianMipmapPass(engine, fg, prepareShadowPass->shadows,
+                        layer, textureRequirements.levels, textureRequirements.clearColor);
+            } else {
+                // for EVSM, a regular mipmap is enough
+                vsmMipmapPass(engine, fg, prepareShadowPass->shadows,
+                        layer, textureRequirements.levels, textureRequirements.clearColor);
             }
         }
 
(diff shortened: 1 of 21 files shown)


libs/utils/src/Allocator.cpp
```diff
@@ -197,12 +215,14 @@ void TrackingPolicy::HighWatermark::onRewind(void const* addr) noexcept {
 
 void TrackingPolicy::Debug::onAlloc(void* p, size_t const size, size_t, size_t) noexcept {
     if (p) {
+        UTILS_UNPOISON_MEMORY_REGION(p, size);
         memset(p, 0xeb, size);
```
show50726, line 219: should we re-poison the memory after `memset`?

filament/src/fg/ResourceNode.cpp
```diff
@@ -81,40 +83,29 @@ void ResourceNode::setIncomingEdge(ResourceEdgeBase* edge) noexcept {
...
-    for (auto const& reader : readers) {
-        if (!dependencyGraph.getNode(reader->to)->isCulled()) {
-            return true;
-        }
-    }
-    return false;
+    DependencyGraph const& dependencyGraph = mFrameGraph.getGraph();
+    return dependencyGraph.findOutgoingEdge(this, [&dependencyGraph](DependencyGraph::Edge const* reader) {
+        return !dependencyGraph.getNode(reader->to)->isCulled();
+    }) != nullptr;
```
show50726, line 89: It seems `findOutgoingEdge` will try to traverse all edges. Can we traverse `mReaderPassesHead` instead?

pixelflinger: no this wouldn't work, in fact there is a comment to that effect on line 85 :-)

"here we don't use mReaderPasses because this wouldn't account for subresources"

libs/utils/src/Allocator.cpp
pixelflinger: yes! we should. good catch.

pixelflinger: done!

pixelflinger: actually I undid-it! I forgot that the PoolAllocator manages its internal state inside the allocated block, which means we can't re-poison the block with asan from the Debug policy. The re-poison works only when the underlaying allocator is using malloc.

show50726: Oh I see, thanks for the explanation!

filament/src/fg/ResourceNode.cpp
show50726: I missed the comment, now I understand!

show50726 (approved):

libs/utils/src/Allocator.cpp
show50726: maybe we can add some comments for this?

filament/src/ShadowMapManager.cpp
```diff
@@ -876,67 +889,82 @@ FrameGraphId<FrameGraphTexture> ShadowMapManager::vsmMipmapPass(
...
+                auto const hwIn = resources.getTexture(data.in);
 
                 auto const& ppm = engine.getPostProcessManager();
                 ppm.bindPostProcessDescriptorSet(driver);
                 ppm.bindPerRenderableDescriptorSet(driver);
 
                 auto& material = ppm.getPostProcessMaterial("gaussianMipmap");
                 FMaterial const* const ma = material.getMaterial(engine);
-                auto const mi = ppm.getMaterialInstance(driver, ma);
+                auto const pipeline = ppm.getPipelineState(ppm.getMaterialInstance(driver, ma));
```
z3moon: looks like the returned value of `getMaterialInstance` is unused, so we waste one slot here. can we move this inside the loop?

pixelflinger: done

PR #10324, opened 2026-08-18
backend: fix priority override races and demotion

pixelflinger:
Fix two issues in ThreadPriorityOverride:
- Prevent calling/driver thread priority demotion by guarding restoreDefaultPriority() with an mOverridden flag. If a job is executed inline by the driver thread via dequeue() or completes without being waited on, priority restoration is a no-op.
- Fix a race condition where startOverride() ran before the worker registered its thread handle/TID. Using a sequentially consistent handshake between mOverrideRequested and mTid/mThread guarantees priority elevation is never missed.
- Add defensive null handle checks on Windows and harmonize all atomic operations to std::memory_order_seq_cst.
- Add unit test suite covering registration handshakes, inline execution safety, and concurrent stress testing.

FIXES=[548593434]

### DIFF
--- a/filament/backend/src/ThreadPriorityOverride.cpp
+++ b/filament/backend/src/ThreadPriorityOverride.cpp
@@ -1,0 +1,226 @@
+/*
+ * Copyright (C) 2026 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "ThreadPriorityOverride.h"
+
+#include <atomic>
+
+#if defined(__APPLE__)
+#include <pthread.h>
+#include <sys/qos.h>
+#elif defined(__ANDROID__)
+#include <sys/resource.h>
+#include <sys/types.h>
+#include <unistd.h>
+#include <cerrno>
+
+#ifndef ANDROID_PRIORITY_URGENT_DISPLAY
+#define ANDROID_PRIORITY_URGENT_DISPLAY (-8)
+#endif
+#ifndef ANDROID_PRIORITY_DISPLAY
+#define ANDROID_PRIORITY_DISPLAY (-4)
+#endif
+#ifndef ANDROID_PRIORITY_NORMAL
+#define ANDROID_PRIORITY_NORMAL (0)
+#endif
+#ifndef ANDROID_PRIORITY_BACKGROUND
+#define ANDROID_PRIORITY_BACKGROUND (10)
+#endif
+
+#elif defined(WIN32)
+#include <windows.h>
+#endif
+
+namespace filament::backend {
+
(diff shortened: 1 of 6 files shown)


filament/backend/src/ThreadPriorityOverride.h
```diff
@@ -79,123 +114,247 @@ class ThreadPriorityOverride {
...
+     * - Uses std::memory_order_seq_cst when storing mOverrideRequested and loading the thread
+     *   ID/handle to participate in the bidirectional handshake with registerCurrentThread().
+     * - If the worker has already registered, priority is elevated immediately. If the worker
+     *   has not yet registered, the worker will observe mOverrideRequested upon registration.
      */
     void startOverride() noexcept {
         if (!mEnabled) {
             return;
         }
+        mOverrideRequested.store(true, std::memory_order_seq_cst);
```
z3moon: this never sets to false again. I think this class is designed for single time use along with OpenGLProgramToken, but it'd be nice to set this off at the end of `endOverride` to make it reusable.

filament/backend/src/ThreadPriorityOverride.h
```diff
@@ -79,123 +114,247 @@ class ThreadPriorityOverride {
...
+                        winPriority = THREAD_PRIORITY_NORMAL;
+                        break;
+                    case utils::JobSystem::Priority::DISPLAY:
+                        winPriority = THREAD_PRIORITY_ABOVE_NORMAL;
+                        break;
+                    case utils::JobSystem::Priority::URGENT_DISPLAY:
+                        winPriority = THREAD_PRIORITY_HIGHEST;
+                        break;
+                }
+                SetThreadPriority(thread, winPriority);
```
z3moon: you could use `GetCurrentThread()` instead of `thread` since this method is called by the worker thread.

filament/backend/src/ThreadPriorityOverride.cpp
```diff
@@ -0,0 +1,226 @@
...
+    return mOverridden.load(std::memory_order_acquire);
+#else
+    return false;
+#endif
+}
+
+#if defined(__APPLE__)
+void ThreadPriorityOverride::applyOverride(pthread_t const thread) noexcept {
+    pthread_override_t expected = nullptr;
+    pthread_override_t const override =
```
bejado: nit: choose a variable name other than `override` since that already carries special meaning in C++

filament/backend/test/test_ThreadPriorityOverride.cpp
```diff
@@ -0,0 +1,388 @@
...
+        JobSystem::setThreadPriority(JobSystem::Priority::NORMAL);
+    });
+
+    // Wait until worker has completed its restorePriority() call.
+    while (!workerRestored.load(std::memory_order_acquire)) {
+        std::this_thread::yield();
+    }
+
+    // Waiter triggers late startOverride() after worker's restorePriority() finished.
+    override.startOverride();
```
bejado, line 371: Should we try to reset `mThread` / `mTid` inside `restorePriority`? I see that this test ensures we eventually clean up late overrides, but leaving the thread ID set means a late `startOverride` on Android/Windows will temporarily elevate the priority of the thread, even though it's already finished with its job. Perhaps that's acceptable to avoid acquiring additional locks.

pixelflinger: no because:

- On Android and Windows, if `startOverride()` races right as the worker is finishing,  `restorePriority()` may see `mOverridden == false` and skip restoration. In that case, the responsibility to restore priority falls on `endOverride()`

- If a compilation token is cancelled, abandoned, or destroyed before `endOverride()` runs, `~ThreadPriorityOverride()` uses `mTid` / `mThread` to ensure any unended priority elevation is cleaned up.

Issue #2547, opened 2020-05-18
Memory leaks related to MetalFence and TimerQueryFence.


vpopovyc:
**Describe the bug**
Memory leaks related to MetalFence and TimerQueryFence.

https://github.com/google/filament/commit/27f4504dfc027f3187446835fd5ed679ffe7594b introduced ```TimerQueryFence``` with 
```void TimerQueryFence::beginTimeElapsedQuery(MetalTimerQuery* query)``` and 
```void TimerQueryFence::endTimeElapsedQuery(MetalTimerQuery* query)```.
In both implementations ```MetalFence``` is allocated like 
```auto* fence = new MetalFence(mContext);```
and deleted like
```delete fence;``` in ```fence.onSignal``` closure.

So the issue is, imo, that ```fence.onSignal``` sometimes isn't called.

As a result, almost on each frame that pair leaking some memory.

**To Reproduce**
You can easily reproduce that by launching any sample app(I used hello-pbr) for iOS, turn on ```Malloc Stack```, launch app on device and click ```Debug Memory Graph```.

**Expected behavior**
No leaks, I think.

**Screenshots**
<img width="1920" alt="Screenshot 2020-05-18 at 17 30 13" src="https://user-images.githubusercontent.com/24235281/82225037-49dcb300-992d-11ea-94c7-3be5b0da12a2.png">

<img width="1058" alt="Screenshot 2020-05-18 at 17 32 47" src="https://user-images.githubusercontent.com/24235281/82225268-a9d35980-992d-11ea-8eb6-dd86fe36186f.png">

**Smartphone (please complete the following information):**
 - OS: [iOS 13.4]
 - GPU: [AGXA10Device]
 - Backend: [Metal]

bejado:
Thanks for the report. Indeed, those fences are not being encoded and thus the `onSignal` callback is never called.

Issue #1629, opened 2019-09-11
DriverAPI: introduce a timestamp query API.


prideout:
Filament currently uses fences to determine frame time (e.g. to support frame skipping) but in Vulkan, each fence is associated with a specific submission of a Vk command buffer.  Moreover we triple-buffer the Vk command buffers on some platforms.  Fences are not the correct mechanism.

To fix this, we should add query objects to the driver API, and change the frame skipper to use these if possible.  For starters, queries should perhaps only support timestamps. However we should consider making them extensible for the future.

Timestamps are supported in the Vulkan core, and on OpenGL they are supported via `EXT_disjoint_timer_query`. I believe Metal can support similar functionality via its counter sampling API.

romainguy:
On OpenGL what we'd want to use are Android's EGL timestamp extensions.

prideout:
See #2325

romainguy:
Is this still required with @pixelflinger's recent changes?

PR #10338, opened 2026-08-20
Metal: report real GPU time from timer queries

tim4724:
The Metal backend doesn't measure GPU time.

`MetalTimerQueryFence` records `clock::now()` in two fence callbacks and reports the difference as `FrameInfo::gpuFrameDuration`. That is a CPU-side gap between two completion callbacks. It contains the GPU work, but also callback latency and the fence encode itself, and when the GPU is busy it just tracks how often frames are presented.

This reads `MTLCommandBuffer.GPUStartTime`/`GPUEndTime` instead, which the driver fills in with the GPU's actual execution window.

## Measurements

Apple TV 4K (A10X), before the change, two workloads:

| workload | `gpuFrameDuration` | present interval |
|---|---|---|
| lighter | 15.4 - 16.7 ms | 16.67 ms (60 fps) |
| heavier | 17.4 - 19.1 ms | 19.2 ms (52 fps) |

In both rows the reported GPU time is the present interval. It says "at budget" whether the GPU is busy or idle.

macOS (Apple silicon, `samples/hellotriangle`, release, presenting every ~17.9 ms). Same binary and scene, only the timer differs:

| | `gpuFrameDuration` |
|---|---|
| before | 0.54 - 0.88 ms |
| after | 0.22 - 0.38 ms |

The GPU is idle here, so the old value isn't the present interval, but it still over-reports a 0.25 ms frame by two to three times.

After the change on the Apple TV, the value follows the work: 16.1 ms at full resolution and 6.7 ms at half, both at 60 fps. Rendering into a buffer larger than the screen, so the frame rate isn't capped by vsync, `1000 / gpuFrameDuration` matches the measured fps at four points: 27.0 ms / 37 fps, 38.2 / 26, 36.8 / 27 and 30.6 / 33.

## What this affects

`gpuFrameDuration` isn't only a debug value. `FrameInfoManager::denoiseFrameTime` medians it into `denoisedFrameTime`, and `FView::updateScale` divides that by the frame budget to drive the PID controller behind dynamic resolution.

Since the measured value *is* roughly the budget, that ratio sits at or above 1.0 permanently, even on an idle GPU. So dynamic resolution on Metal has been pushing scale down toward `minScale` regardless of how much headroom there is. `DynamicResolutionOptions::enabled` defaults to false, so this only affects apps that opted in, but for those the change should raise resolution rather than lower it.

That is a behaviour change and worth calling out: anyone who tuned around the old number will see different scaling.

This also removes the `new MetalFence` / `delete` inside a signal callback in both query functions. That construct is what leaked in #2547 (fixed in #2548); it is gone rather than adjusted.

Related: #1629, which added the query API in 2019 and noted "I believe Metal can support similar functionality via its counter sampling API".

I could not find an existing issue for this. That fits the failure mode: it is silent, and a resolution that is lower than it should be looks like your own scene being expensive.

## Reproducing

Print `gpuFrameDuration` next to the wall-clock present interval in a sample's frame loop and run it windowed. On a busy GPU the old value converges on the present interval.

## Notes

- With one command buffer per frame, the usual case, this is that frame's GPU duration. A frame using several (a mid-frame `readPixels`, for instance) also counts the gaps between them.
- `MTLCounterSampleBuffer` gives per-encoder timestamps, but it needs feature checks and a finer granularity than this API exposes.
- No new availability guards. The class is already `API_AVAILABLE(ios(12.0))` and the two properties are `ios(10.3)`/`macos(10.15)`, against a deployment target of 11.0.
- If a driver doesn't fill in the timestamps the result is 0, which `denoiseFrameTime` already ignores.
- `asynchronousMode` doesn't affect this: it gates a JobQueue for shader compilation, not command-buffer structure.
- The class no longer uses a fence, and the `clock` alias on `MetalTimerQueryInterface` is now unused. Left alone to keep the diff small; happy to clean up.

Tested on macOS (Apple silicon) and tvOS 26.6 (A10X). Not run on iOS or Intel macOS.

Investigated and prepared with Claude Code.


filament/backend/src/metal/MetalHandles.h
```diff
@@ -500,6 +500,7 @@ struct MetalTimerQuery : public HwTimerQuery {
 
     struct Status {
         std::atomic<bool> available {false};
+        std::atomic<uint64_t> start {0};     // GPUStartTime of the begin command buffer
         std::atomic<uint64_t> elapsed {0};   // only valid if available is true
```
bejado: Let's postfix these with `ns`: `startNs` and `elapsedNs`, so the units are clear.

filament/backend/src/metal/MetalTimerQuery.mm
```diff
@@ -24,36 +25,36 @@
 MetalTimerQueryInterface::~MetalTimerQueryInterface() = default;
 
 void MetalTimerQueryFence::beginTimeElapsedQuery(MetalTimerQuery* query) {
```
bejado: Since this no longer uses a `MetalFence`, let's rename this class to simply `MetalTimerQuery`. Also, we can remove the `API_AVAILABLE(ios(12.0))` in MetalTimerQuery.h because this class is no longer limited to iOS 12.0+. In fact, we can probably get rid of `MetalTimerQueryNoop` and this check here:

https://github.com/google/filament/blob/25e9aa00f499acd562269f9bc8d071ee516924c7/filament/backend/src/metal/MetalDriver.mm#L225-L229

and just make `MetalTimerQuery` the default since it now works across all iOS versions we support.

bejado (changes requested):
I agree with this change and the code LGTM (just a couple suggestions). As you pointed out, this fixes the semantics of Metal's timer queries but will change how dynamic resolution behaves for clients who opted in (it will now correctly scale up). I want to check internally to ensure no clients had applied weird workarounds relying on the old behavior. Assuming that's clear, I'm happy with this!

One minor caveat to note is that by relying on `MTLCommandBuffer`'s `GPUStartTime` and `GPUEndTime`, we lose the ability to measure fine-grained execution _within_ a command buffer. To be fair, the old implementation was already a bit wonky for fine-grained profiling since it relied on CPU-side callbacks subject to thread scheduling. With this implementation, if we were to issue multiple timer queries within a single command buffer, they would now all return the exact same duration: the execution time of the entire command buffer.

Fortunately, I don't think we have to worry about this limitation since I believe Filament currently only uses timer queries to measure full frame boundaries.

filament/backend/src/metal/MetalHandles.h
tim4724: Done

filament/backend/src/metal/MetalTimerQuery.mm
tim4724: I couldn't use plain MetalTimerQuery, because MetalHandles.h already has struct MetalTimerQuery : public HwTimerQuery for the handle. 
I went with **MetalTimerQueryImpl** to match the mContext->timerQueryImpl member it's stored in.

I also wrote your command-buffer point into the header:
// NOTE: the resolution is a whole command buffer. Several timer queries issued inside one
// command buffer all report that command buffer's execution time. Filament only measures frame
// boundaries, where this is what is wanted.

filament/backend/src/metal/MetalTimerQuery.h
```diff
@@ -19,46 +19,32 @@
...
 struct MetalContext;
 
-class MetalTimerQueryInterface {
+// Implements timer queries by reading GPUStartTime/GPUEndTime off the frame's command buffer.
+// Those are available on every OS version Filament supports, so there is no fallback.
+//
+// NOTE: the resolution is a whole command buffer. Several timer queries issued inside one
+// command buffer all report that command buffer's execution time. Filament only measures frame
+// boundaries, where this is what is wanted.
+class MetalTimerQueryImpl {
```
bejado, line 36: This reminds me, we also need to update `isFrameTimeSupported` to always return `true` now.

https://github.com/google/filament/blob/25e9aa00f499acd562269f9bc8d071ee516924c7/filament/backend/src/metal/MetalDriver.mm#L1421-L1427

tim4724: Yes - done

bejado (approved):
LGTM. Will merge once CI passes

PR #10341, opened 2026-08-21
samples: make parameter definition re-usable

poweifeng:
- Organize parameters definition into a struct and re-use across all samples.  This is helpful for upcoming Android refactor.
 - Fix command line help message printing.
 - Fix a few leaks and allocations error (stack vs. static)


pixelflinger (approved):
I trust this works :-)

PR #10344, opened 2026-08-21
backend: fix stale async callback docs

z3moon:
Fix the `CountdownCallbackHandler::post`'s outdated docs to reflect the new AsyncCompletion mechanism: `scheduleAsyncCallback()` binds the status to the call site, so it goes through DriverBase's functor overload, which boxes (countdownCallback, this, status) into an internal allocation and passes its own trampoline.


filament/src/details/AsyncHelpers.h
```diff
@@ -89,8 +89,12 @@ class CountdownCallbackHandler : public backend::CallbackHandler {
     // "ServiceThread" in DriverBase. Its primary purpose is to manage the countdown
     // mechanism before the *REAL* user's callback is executed.
     void post(void* user, Callback callback) override {
```
pixelflinger: should this be `final`?

filament/backend/include/backend/CallbackHandler.h
```diff
@@ -67,6 +67,12 @@ class CallbackHandler {
      * Schedules the callback to be called onto the appropriate thread.
      * Typically this will be the application's main thead.
      *
+     * `user` is opaque: it is not necessarily the pointer the application passed to the API that
+     * took this handler. Filament may substitute an internal allocation, and does so for async
+     * completion callbacks, which have to carry an AsyncCallStatus alongside the user pointer. A
+     * handler must treat `user` as a token to hand back to `callback`, and must never dereference
+     * it or route on it.
+     *
```
pixelflinger: I think this comment is confusing an unnecessary.  It really only apply to `AsyncHelpers`. But in general `CallbackHandler` is just an interface to generically schedule a callback.

It doesn't mean much to say "it is not necessarily the pointer the application passed to the API that took the handler", since this interface can technically be used to schedule anything. I see what you're trying to say though. Personally I would remove this new comment from here.

z3moon: Yeah true, let me remove it.

filament/src/details/AsyncHelpers.h
z3moon: done

bejado (approved):

PR #10346, opened 2026-08-21
frontend: require successful creation before handing out VB/IB handles

z3moon:
FTexture guards entry points with
FILAMENT_CHECK_PRECONDITION(isCreationSuccessful()). FVertexBuffer and FIndexBuffer expose the same issue but never check it. Add the guards like FTexture.

### DIFF
--- a/filament/include/filament/RenderableManager.h
+++ b/filament/include/filament/RenderableManager.h
@@ -896,21 +896,33 @@ public:
     /**
      * Changes the geometry for the given primitive.
      *
-     * \see Builder::geometry()
-     */
-    void setGeometryAt(Instance instance, size_t primitiveIndex, PrimitiveType type,
-            VertexBuffer* UTILS_NONNULL vertices,
-            IndexBuffer* UTILS_NONNULL indices,
-            size_t offset, size_t count) noexcept;
-
-    /**
-     * Changes the geometry for the given primitive. (non-indexed version)
+     * Buffers built asynchronously must have completed their creation. See
+     * `VertexBuffer::isCreationComplete()` and `IndexBuffer::isCreationComplete()`.
+     *
+     * @exception utils::PreConditionPanic if a buffer's creation is still in progress or was
+     *                                     canceled.
      *
      * \see Builder::geometry()
      */
     void setGeometryAt(Instance instance, size_t primitiveIndex, PrimitiveType type,
             VertexBuffer* UTILS_NONNULL vertices,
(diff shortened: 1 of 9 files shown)


filament/src/details/IndexBuffer.cpp
```diff
@@ -149,6 +149,14 @@ void FIndexBuffer::terminate(FEngine& engine) {
     driver.destroyIndexBuffer(mHandle);
 }
 
+backend::Handle<backend::HwIndexBuffer> FIndexBuffer::getHwHandle() const {
+    FILAMENT_CHECK_PRECONDITION(isCreationSuccessful())
+            << "IndexBuffer is not usable: its creation is still in progress, or its asynchronous "
+               "creation was canceled";
+
+    return mHandle;
+}
```
pixelflinger: I'm confused about this, this method should only be called by filament itself, so it shouldn't use a precondition check -- those are only for public APIs.
If we trip the check, that means we didn't use the API cirrectly.

I don't like this at all because it will basically "infect" any caller with removing `noexcept`. Users can't try/catch because they don't know where this is called from.

Ideally we can check for this a the public API boundary.

pixelflinger (changes requested):

filament/src/details/IndexBuffer.cpp
z3moon: Ah right. It should be done on the public API side. Let me update it.

filament/include/filament/RenderableManager.h
```diff
@@ -896,21 +896,33 @@ class UTILS_PUBLIC RenderableManager : public FilamentAPI {
...
+     *
+     * @exception utils::PreConditionPanic if the buffer's creation is still in progress or was
+     *                                     canceled.
+     *
      * \see Builder::geometry()
      */
     void setGeometryAt(Instance instance, size_t primitiveIndex, PrimitiveType type,
             VertexBuffer* UTILS_NONNULL vertices,
-            size_t offset, size_t count) noexcept;
+            size_t offset, size_t count);
```
pixelflinger, line 925: I think this means the java bindings need to be updated to handle the exception

pixelflinger (changes requested):

filament/include/filament/RenderableManager.h
z3moon: Good point! done.

pixelflinger (approved):

PR #10347, opened 2026-08-22
vulkan: bounce readPixels staging through a cached copy before reshape

tim4724:
Without `VK_MEMORY_PROPERTY_HOST_CACHED_BIT` (optional, absent on e.g. PowerVR) the readPixels staging mapping is uncached, and DataReshaper's per-texel loop turns every 2-4 byte load into its own memory transaction. When the uncached fallback memory type is selected, bounce the staging buffer into cached heap memory with one bulk memcpy and reshape from there; when `HOST_CACHED` is available, the path is unchanged.

Measured on a Google TV Streamer (PowerVR GE9215): a 1024x1024 R16F readback drops from ~190 ms to the cost of a 2 MB memcpy. The fallback branch was validated on that device; CI drivers all expose `HOST_CACHED`, so they exercise the unchanged path.

This change was developed with the assistance of Claude Code.

### DIFF
--- a/filament/backend/src/vulkan/VulkanReadPixels.cpp
+++ b/filament/backend/src/vulkan/VulkanReadPixels.cpp
@@ -371,6 +371,10 @@ void VulkanReadPixels::run(fvkmemory::resource_ptr<VulkanTexture> srcTexture, ui
         VkResult status = vkWaitForFences(device, 1, &fence, VK_TRUE, UINT64_MAX);
         if (status != VK_SUCCESS) {
             FVK_LOGE << "Failed to wait for readPixels fence";
+            vkDestroyBuffer(device, stagingBuffer, VKALLOC);
+            vkFreeMemory(device, stagingMemory, VKALLOC);
+            vkDestroyFence(device, fence, VKALLOC);
+            vkFreeCommandBuffers(device, cmdpool, 1, &cmdbuffer);
             return;
         }
 
--- a/filament/backend/src/vulkan/VulkanReadPixels.cpp
+++ b/filament/backend/src/vulkan/VulkanReadPixels.cpp
@@ -27,6 +27,9 @@
 #include <utils/compiler.h>
 #include <utils/Log.h>
 
+#include <cstring>
+#include <memory>
+
 using namespace bluevk;
 
 namespace filament::backend {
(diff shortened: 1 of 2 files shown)


poweifeng (approved):

filament/backend/src/vulkan/VulkanReadPixels.cpp
```diff
@@ -255,14 +258,16 @@ void VulkanReadPixels::run(fvkmemory::resource_ptr<VulkanTexture> srcTexture, ui
     uint32_t memoryTypeIndex = selectMemoryFunc(memReqs.memoryTypeBits,
             VK_MEMORY_PROPERTY_HOST_VISIBLE_BIT | VK_MEMORY_PROPERTY_HOST_COHERENT_BIT
                     | VK_MEMORY_PROPERTY_HOST_CACHED_BIT);
+    bool hostCachedStaging = true;
 
     // If VK_MEMORY_PROPERTY_HOST_CACHED_BIT is not supported, we try only
     // HOST_VISIBLE+HOST_COHERENT.  HOST_CACHED helps a lot with readpixels performance.
     if (memoryTypeIndex >= VK_MAX_MEMORY_TYPES) {
         memoryTypeIndex = selectMemoryFunc(memReqs.memoryTypeBits,
                 VK_MEMORY_PROPERTY_HOST_VISIBLE_BIT | VK_MEMORY_PROPERTY_HOST_COHERENT_BIT);
```
phoenixxxx, line 267: I am curious if we need the VK_MEMORY_PROPERTY_HOST_COHERENT_BIT here? Admittedly I don't think any GPU will have VK_MEMORY_PROPERTY_HOST_CACHED_BIT but not ALSO the VK_MEMORY_PROPERTY_HOST_COHERENT_BIT, but I don't think they are both necessary here.

poweifeng: coherent here is for correctness.  When reading out the pixels, it needs to be coherent

filament/backend/src/vulkan/VulkanReadPixels.cpp
phoenixxxx, line 378: Destroy the fence here?

phoenixxxx (approved):

poweifeng (approved):

PR #10349, opened 2026-08-25
webgpu: fix scissor/rendertarget bug

poweifeng:
Swapchain request was moved to beginRenderPass. We need to also update the default renderTarget dimension in the same block.

Also refactor the scissor logic for clarity.

This fixes the c++ samples that did not use post processing (i.e. rendering directly to swapchain), including hellotriangle.

### DIFF
--- a/filament/backend/src/webgpu/WebGPUDriver.cpp
+++ b/filament/backend/src/webgpu/WebGPUDriver.cpp
@@ -1352,6 +1352,10 @@ void WebGPUDriver::beginRenderPass(Handle<HwRenderTarget> renderTargetHandle,
             mSwapChainView = mSwapChain->isHeadless() ?
                 mSwapChain->getNextTextureView() :
                 mSwapChain->getNextTextureView(mPlatform.getSurfaceExtent(mNativeWindow));
+
+            // We need to set the extent for the render target for use in scissor()
+            renderTarget->width = mSwapChain->getWidth();
+            renderTarget->height = mSwapChain->getHeight();
         }
         defaultColorView  = mSwapChainView;
         defaultDepthStencilView = mSwapChain->getDepthTextureView();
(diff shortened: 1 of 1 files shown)


filament/backend/src/webgpu/WebGPUDriver.cpp
```diff
@@ -2132,13 +2136,23 @@ void WebGPUDriver::scissor(Viewport scissor) {
...
+    uint32_t const rtHeight = mCurrentRenderTarget->height;
+
+    auto capToRange = [](uint32_t val, uint32_t min, uint32_t max) {
+        return std::min(std::max(val, min), max);
+    };
+
+    uint32_t const left = capToRange(scissor.left, 0, rtWidth);
+    uint32_t const right = capToRange(scissor.left + scissor.width, 0, rtWidth);
+
+    // Filament gives us bottom-left-orgin coordinates, we need to convert it to WebGPU's top-left
```
bejado: nit

```suggestion
    // Filament gives us bottom-left-origin coordinates, we need to convert it to WebGPU's top-left
```

filament/backend/src/webgpu/WebGPUDriver.cpp
```diff
@@ -2132,13 +2136,23 @@ void WebGPUDriver::scissor(Viewport scissor) {
...
-    uint32_t bottom = std::max(0, scissor.bottom);
-    uint32_t top = rtHeight > (bottom + scissor.height) ? rtHeight - bottom - scissor.height : 0;
-    uint32_t width = std::min((uint32_t)scissor.width, rtWidth - left);
-    uint32_t height = std::min((uint32_t)scissor.height, rtHeight - top);
+    uint32_t const rtWidth = mCurrentRenderTarget->width;
+    uint32_t const rtHeight = mCurrentRenderTarget->height;
+
+    auto capToRange = [](uint32_t val, uint32_t min, uint32_t max) {
+        return std::min(std::max(val, min), max);
+    };
```
bejado: Could we use `std::clamp` here instead?

bejado (approved):

z3moon (approved):

filament/backend/src/webgpu/WebGPUDriver.cpp
poweifeng: done

filament/backend/src/webgpu/WebGPUDriver.cpp
poweifeng: yes, good call, done.

show50726 (approved):

PR #10350, opened 2026-08-25
android: apply version script to RelWithDebInfo

bejado:
Android release AARs are built using the `RelWithDebInfo` CMake configuration, but `-Wl,--version-script` was previously only configured on `CMAKE_SHARED_LINKER_FLAGS_RELEASE`. As a result, release builds were not applying the version script to hide unexported symbols.

This reduces the `filament-android-release.aar` size from **5.26 MB** to **4.93 MB** .

zstd is required by gltfio. Somehow it was getting transitively included before, but with the version script it was erroneously getting stripped. So I added an explicit dependency.


poweifeng (approved):
Maybe add a line in the comment message about the zstd change to gltfio?

pixelflinger (approved):

PR #10351, opened 2026-08-25
ColorGrading: support Rec709-Linear-D65 in NEON LUT generation

AdrianAtGoogle:
Enable ColorGrading objects configured with Rec709-Linear-D65 output color space to take advantage of optimized NEON fast-paths (generateDefaultLUTNeon and generateMediumLUTNeon).

Previously, the NEON builders were guarded by outputColorSpace == Rec709-sRGB-D65 due to the hardcoded v_oetf_sRGB step, forcing linear output configurations to fall back to the slower scalar generation path.

Details:
- Updated NEON fast-path checks (isDefaultState and isMediumState) in FColorGrading to accept both Rec709-sRGB-D65 and Rec709-Linear-D65.
- Templatized generateDefaultLUTNeon and generateMediumLUTNeon on `bool isLinear` using `if constexpr` to eliminate the sRGB OETF calculation at compile time when linear output is requested.
- Preserved zero-overhead execution for existing Rec709-sRGB-D65 clients without introducing runtime branching inside the inner loops.
- Added ColorGradingNeonLinearValidation and ColorGradingMediumNeonLinearValidation test cases in filament_test.


filament/src/details/ColorGrading.cpp
```diff
@@ -1097,8 +1099,8 @@ float4 FColorGrading::hdrColorAt(Builder const& builder, Config const& config,
 
 #if defined(__ARM_NEON)
 
-UTILS_NOINLINE
-void FColorGrading::generateDefaultLUTNeon(FEngine const& engine, void* data,
+template <bool isLinear>
```
poweifeng: So will this inline the code?  I think the original `UTILS_NOINLINE` is there to prevent too much binary increase.

AdrianAtGoogle: I kept the UTILS_NOINLINE on the outer method that dispatches to either template parameter, so I think it should still work the same (but the Neon methods will be about twice as big, unfortunately)

filament/src/details/ColorGrading.cpp
```diff
@@ -1303,7 +1317,9 @@ void FColorGrading::generateMediumLUTNeon(FEngine const& engine, void* data, Con
                     cg_g = vmaxq_f32(vminq_f32(cg_g, vdupq_n_f32(1.0f)), vdupq_n_f32(0.0f));
                     cg_b = vmaxq_f32(vminq_f32(cg_b, vdupq_n_f32(1.0f)), vdupq_n_f32(0.0f));
 
-                    v_oetf_sRGB(cg_r, cg_g, cg_b);
+                    if constexpr (!isLinear) {
```
pixelflinger: honestly for this one I would just do a runtime test (just like `hasAdj`) instead of duplicating all the neon code. I would use LIKELY (that the oetf_sRGB is used).

pixelflinger (changes requested):

filament/src/details/ColorGrading.cpp
AdrianAtGoogle: Done, thank you

filament/src/details/ColorGrading.cpp
```diff
@@ -1097,11 +1099,12 @@ float4 FColorGrading::hdrColorAt(Builder const& builder, Config const& config,
 
 #if defined(__ARM_NEON)
 
-UTILS_NOINLINE
+template <bool isLinear>
```
pixelflinger: we don't need the template anymore, since we're resolving at runtime. right?

pixelflinger (changes requested):

filament/src/details/ColorGrading.cpp
AdrianAtGoogle: sorry for the hassle, I was only building for linux

filament/src/details/ColorGrading.cpp
```diff
@@ -1097,11 +1099,11 @@ float4 FColorGrading::hdrColorAt(Builder const& builder, Config const& config,
 
 #if defined(__ARM_NEON)
 
-UTILS_NOINLINE
 void FColorGrading::generateDefaultLUTNeon(FEngine const& engine, void* data,
         Config const& config, Builder const& builder) noexcept {
     FILAMENT_TRACING_CALL(FILAMENT_TRACING_CATEGORY_FILAMENT);
 
+    bool const isLinear = (builder->outputColorSpace == Rec709-Linear-D65);
```
pixelflinger: the "Default LUT" should always be sRGB

filament/src/details/ColorGrading.cpp
```diff
@@ -878,16 +880,16 @@ FColorGrading::FColorGrading(FEngine& engine, const Builder& builder) {
                                 builder->fastMath &&
                                 engine.features.engine.color_grading.use_optimized_default_builder &&
                                 builder->toneMapping == ToneMapping::ACES_LEGACY && 
-                                builder->outputColorSpace == Rec709-sRGB-D65 &&
-                                type == PixelDataType::UINT_2_10_10_10_REV &&
+                                isSupportedColorSpace &&
```
pixelflinger: I missed this before, the "default state" should always be sRGB. It doesn't have a choice.

filament/src/details/ColorGrading.cpp
```diff
@@ -1159,7 +1161,9 @@ void FColorGrading::generateDefaultLUTNeon(FEngine const& engine, void* data,
                 cg_g = vmaxq_f32(vminq_f32(cg_g, vdupq_n_f32(1.0f)), vdupq_n_f32(0.0f));
                 cg_b = vmaxq_f32(vminq_f32(cg_b, vdupq_n_f32(1.0f)), vdupq_n_f32(0.0f));
 
-                v_oetf_sRGB(cg_r, cg_g, cg_b);
+                if (UTILS_LIKELY(!isLinear)) {
+                    v_oetf_sRGB(cg_r, cg_g, cg_b);
+                }
```
pixelflinger: in the default case, it's always sRGB

filament/src/details/ColorGrading.h
```diff
@@ -63,6 +63,10 @@ class FColorGrading : public ColorGrading {
     bool mIsLDR;
 
 #if defined(__ARM_NEON)
+    template <bool isLinear>
+    static void generateDefaultLUTNeonImpl(FEngine const& engine, void* data, Config const& config, Builder const& builder) noexcept;
+    template <bool isLinear>
+    static void generateMediumLUTNeonImpl(FEngine const& engine, void* data, Config const& config, Builder const& builder) noexcept;
```
pixelflinger: shouldn't be needed anynore

pixelflinger (changes requested):

filament/src/details/ColorGrading.cpp
AdrianAtGoogle: fixed

filament/src/details/ColorGrading.cpp
AdrianAtGoogle: fixed, sorry about that

filament/src/details/ColorGrading.h
AdrianAtGoogle: thank you, removed.

filament/src/details/ColorGrading.cpp
AdrianAtGoogle: fixed, thank you

pixelflinger (approved):

PR #10353, opened 2026-08-26
Fix warnings in test_Allocators.cpp

romainguy:
- Use a void cast to suppress warnings about [[nodiscard]]
- Fixes various other style issues (casts mostly)


romainguy:
@pixelflinger

poweifeng (approved):

pixelflinger (approved):

Issue #10339, opened 2026-08-20
tvOS support (appletvos + appletvsimulator)


tim4724:
I have tvOS working on top of `v1.75.0` and would like to know whether you want it upstream before I open a PR.

Branch: https://github.com/tim4724/filament/tree/tvos-v1.75.0 — 6 commits, ~250 added lines across 15 files, no new backend. A work-in-progress game runs on an Apple TV 4K (A10X) at 3840x2160, 60 fps.

What it took: `build.sh -p tvos` plus toolchain and CMake plumbing (Metal forced on, since CocoaTouchGL isn't ported); availability guards and tvOS-prohibited API exclusions in `MetalContext.mm`, `MetalHandles.mm` and `MetalState.mm`; and `backend_test` building and running again on iOS/tvOS, behind `INSTALL_BACKEND_TEST`.

The branch also adds a presubmit `build-tvos` job. Per `CONTRIBUTING.md` that one is yours rather than mine, so I would drop `.github/` from any PR.

Three things I would want your call on:

- Deployment target: the branch floors at tvOS 17.0.
- tvOS in the released XCFrameworks, or build-from-source only?
- `backend_test` on tvOS: opt-in, or wired into presubmit?

Happy to split it. The Metal availability fixes stand alone from the build work.

This issue and the branch were prepared with Claude Code.

PR #10356, opened 2026-08-27
Add tvOS support (appletvos + appletvsimulator)

tim4724:
Adds tvOS (`appletvos` + `appletvsimulator`) as a build platform. Opens the discussion proposed in #10339 — since the changes turned out small (~200 lines, no new backend), a concrete diff seemed like the most useful way to have it.

## What this does

- **Toolchain**: tvOS reuses the iOS toolchain path (`third_party/clang/iOS.cmake`) — pass `-DPLATFORM_NAME=appletvos` or `appletvsimulator`. tvOS is API-wise a sibling of iOS (`TARGET_OS_IPHONE` is 1 on both), so `FILAMENT_IOS` stays defined; the target triple selects the tvOS SDK and availability rules. Builds are Metal-only (`FILAMENT_SUPPORTS_OPENGL` forced off — OpenGL ES is deprecated on Apple platforms and the CocoaTouchGL platform layer is not ported).
- **Metal backend fixes**: `MTLArgumentAccessReadOnly` → `MTLBindingAccessReadOnly` renames (the old constants' deprecation predates the tvOS floor, so they fail under `-Werror` there; the new ones are value-identical, so no behavior change on iOS/macOS), tvOS-guarded `useResource:usage:stages:` variants (the stage-less ones are deprecated on tvOS 17+), and a GPU-family probe instead of `supportsFeatureSet:` (unavailable on tvOS) for the comparison-sampler check.
- **`usage_ios_13.cmake`** now only ever *raises* the deployment floor to 13.0 instead of setting it unconditionally — on tvOS (floor 17.0) the old behavior lowered it, which conflicts with the `-target` triple under `-Werror,-Woverriding-option`. No change for iOS builds (11.0 → 13.0 as before).
- **tvOS-prohibited API exclusions**: perfetto and civetweb CGI both use `fork()`/`exec*()`, which are `__TVOS_PROHIBITED`; they're excluded/disabled on tvOS (perfetto is only consumed on Android anyway).
- **Build**: `./build.sh -p tvos` (with `-s` for the simulator slice) produces per-library XCFrameworks under `out/tvos-<type>/filament`, mirroring the iOS flow.
- **CI**: per CONTRIBUTING.md I haven't touched `.github/`, but `build/tvos/build.sh` is included and structured like the iOS wrapper, so a `build-tvos` presubmit job mirroring `build-ios` is a copy-paste for a maintainer (`cd build/tvos && printf "y" | ./build.sh presubmit`).

## Tested

A working game runs on Apple TV 4K (A10X) at 3840x2160 @ 60 fps with this branch. `backend_test` passes on the Apple TV 4K device and the Apple silicon tvOS simulator (built via the existing `INSTALL_BACKEND_TEST` opt-in from #10250; the golden hashes recorded there match on tvOS, modulo one A-series rasterization difference noted in a comment).

## Open questions for maintainers

I picked defaults for three decisions and would like your input:

1. **Deployment floor**: this PR floors tvOS at 17.0, which keeps availability guards minimal (`useResource:stages:`, `supportsFamily:` are all present). Happy to lower it if you'd rather match the iOS floor's spirit — the cost is a handful of `@available` guards.
2. **Release artifacts**: this PR deliberately does *not* add tvOS to the released XCFrameworks — it's build-from-source only. If you'd rather ship it, the `build/tvos/build.sh` wrapper is already structured like the iOS one so wiring it into the release workflow is small.
3. **CI**: should tvOS get its own presubmit job (a maintainer would need to add it, see above), be folded into the existing `build-ios` job as a device-slice build, or stay uncovered for now? The equivalent job has been running green on my fork's CI (pre-rebase branch), and this branch builds both slices clean locally with Xcode 26.5 SDKs.

---

Portions of this change were prepared with the assistance of Claude Code.

### DIFF
--- a/filament/backend/src/metal/MetalState.mm
+++ b/filament/backend/src/metal/MetalState.mm
@@ -156,7 +156,15 @@ id<MTLSamplerState> SamplerStateCreator::operator()(id<MTLDevice> device,
     // MTLSamplerDescriptor.
     // In practice, this means shadows are not supported when running in the simulator.
     if (samplerDescriptor.compareFunction != MTLCompareFunctionNever) {
-        if (![device supportsFeatureSet:MTLFeatureSet_iOS_GPUFamily3_v1]) {
+#if defined(FILAMENT_APPLETV)
+        // supportsFeatureSet:/MTLFeatureSet_iOS_* are unavailable on tvOS; probe the
+        // equivalent GPU family instead (Apple TV HD is Apple2, Apple TV 4K is Apple3+).
+        const bool supportsComparison = [device supportsFamily:MTLGPUFamilyApple3];
+#else
+        const bool supportsComparison =
+                [device supportsFeatureSet:MTLFeatureSet_iOS_GPUFamily3_v1];
+#endif
+        if (!supportsComparison) {
             LOG(WARNING) << "Warning: sample comparison not supported by this GPU";
             samplerDescriptor.compareFunction = MTLCompareFunctionNever;
         }
@@ -189,7 +197,7 @@ id<MTLArgumentEncoder> ArgumentEncoderCreator::operator()(id<MTLDevice> device,
         MTLArgumentDescriptor* bufferArgument = [MTLArgumentDescriptor argumentDescriptor];
         bufferArgument.index = i++;
         bufferArgument.dataType = MTLDataTypePointer;
-        bufferArgument.access = MTLArgumentAccessReadOnly;
+        bufferArgument.access = MTLBindingAccessReadOnly;
         [arguments addObject:bufferArgument];
     }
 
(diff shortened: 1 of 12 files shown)


bejado:
Will take a look this week.

filament/backend/src/metal/MetalState.mm
```diff
@@ -156,7 +156,15 @@
     // MTLSamplerDescriptor.
     // In practice, this means shadows are not supported when running in the simulator.
     if (samplerDescriptor.compareFunction != MTLCompareFunctionNever) {
-        if (![device supportsFeatureSet:MTLFeatureSet_iOS_GPUFamily3_v1]) {
+#if defined(FILAMENT_APPLETV)
```
bejado, line 159: Instead of guarding this check with `FILAMENT_APPLETV`, let's use ` if (@available(iOS 13.0, *)) { ... }`.

bejado (changes requested):
Overall LGTM

filament/backend/src/metal/MetalState.mm
tim4724: Done, switched to `@available(iOS 13.0, *)` + `supportsFamily:`. One caveat: the legacy `supportsFeatureSet:` fallback still needs an `#if TARGET_OS_IOS` around it, since `MTLFeatureSet_iOS_*` is compile-time unavailable on tvOS and `@available` doesn't help there. Same pattern as `MetalContext.mm`. Also rebased onto main.

PR #10359, opened 2026-08-29
Add support for KHR_materials_iridescence

Nemi26:
Adds a thin film interference term to the lit shading models and wires the KHR_materials_iridescence glTF extension through gltfio.

image

Three new material properties: iridescence, iridescenceIor and iridescenceThickness, defaulting to 0.0, 1.3 and 400nm as the extension does. The model is Belcour and Barla 2017 in the approximate form the extension specifies, with Schlick in place of the polarized Fresnel equations and the spectral integral evaluated in Fourier space against Gaussians fitted to the XYZ color matching functions.

The film is evaluated once at the viewing angle and the result refitted to a Schlick curve, so what it produces is an ordinary f0. That is the whole of the integration: getIridescencePixelParams() writes pixel.f0, and the specular lobe, the second specular lobe, the anisotropic lobe and the IBL all pick it up without knowing about it. No new PixelParams member and no new lobe. It runs after getCommonPixelParams() has settled f0, so KHR_materials_specular is already folded in, and before the clear coat remaps f0 to its own interface.

The interference math is highp: the Gaussian fits are stated in inverse meters and their amplitudes are around 1e-13, neither of which mediump can represent.

The specification's rgb_mix() is deliberately absent. It weights the diffuse lobe by the inverse of the film's strongest channel, and Filament does not take a Fresnel share out of the diffuse lobe for an ordinary surface either.

MaterialKey gains the iridescence bits in the space it was already padding with, so it is still 20 bytes. Both material providers are covered; the ubershader gets a new iridescence material in three blend modes, which sampler pressure keeps mutually exclusive with volume, transmission, sheen, specular and clear coat.


Nemi26:
@pixelflinger Could you apply the `internal` label and reopen this PR, if this is something Filament wants? Otherwise I can split the PR to take the gltfio part out of it.

pixelflinger:
Yes I think it makes sense to split the PR.

Issue #10140, opened 2026-06-21
Direct lighting diffuse/specular energy preservation and IBL consistency


zyz23333:
### Summary

Filament's BRDF documentation already identifies two related energy issues:

- **Energy gain in diffuse reflectance**
- **Energy loss in specular reflectance**

The second issue, specular energy loss, is already addressed in the standard shading model through the multiscattering specular energy compensation term:

```glsl
pixel.energyCompensation = 1.0 + pixel.f0 * (1.0 / pixel.dfg.y - 1.0);
Fr *= pixel.energyCompensation;
```

This scales the single-scattering GGX specular lobe to compensate for energy lost by not explicitly simulating multiple microfacet scattering.

However, the first issue, energy gain in diffuse reflectance, is still unresolved for direct lighting. The standard shading model currently evaluates diffuse and specular independently and adds them:

```
vec3 Fr = specularLobe(pixel, light, h, NoV, NoL, NoH, LoH);
vec3 Fd = diffuseLobe(pixel, NoV, NoL, LoH);

// TODO: attenuate the diffuse lobe to avoid energy gain

vec3 color = Fd + Fr * pixel.energyCompensation;
```

The problem is that the specular lobe is compensated for multiscattering, but the diffuse lobe is still evaluated as if the full incident light reached the diffuse scattering event. In Filament's own terminology, the specular-side **energy loss** is compensated, but the diffuse-side **energy gain** is not removed.

This issue proposes exposing the effective specular directional albedo after multiscattering compensation:

```
vec3 compensatedSpecularDirectionalAlbedo(const PixelParams pixel) {
    return saturate(specularDFG(pixel) * pixel.energyCompensation);
}
```

and using it to attenuate the diffuse lobe:

```
Fd *= 1.0 - compensatedSpecularDirectionalAlbedo(pixel);
```

This completes the diffuse-side counterpart of the existing specular energy compensation and makes direct lighting, prefiltered IBL, and importance-sampled IBL use the same energy accounting.

---

### Background: compensation vs preservation

Filament's documentation describes two different problems:

#### 1. Energy gain in diffuse reflectance

The documentation states that the Lambert diffuse BRDF does not account for light that reflects at the surface and therefore cannot participate in the diffuse scattering event.

In a layered interpretation of the standard material, incident light first interacts with the specular interface. Only the part not reflected by the specular layer should reach the diffuse substrate.

Conceptually:

```
incoming energy
    = specular reflected energy
    + energy transmitted to diffuse scattering
[...]

PR #10360, opened 2026-08-29
Add iridescence to the lit shading models

Nemi26:
Adds a thin film interference term to the lit shading models.

This is the shading model half of `KHR_materials_iridescence` support. The
`gltfio` half is a separate PR stacked on this one, since `libs/gltfio` is a
restricted path.

image

## Material model

Three new properties on the lit shading models:

| Property | Type | Range | Default |
| --- | --- | --- | --- |
| `iridescence` | float | [0..1] | 0.0 |
| `iridescenceIor` | float | [1..n] | 1.3 |
| `iridescenceThickness` | float, nanometers | [0..n] | 400.0 |

The model is Belcour and Barla 2017, *A Practical Extension to Microfacet Theory
for the Modeling of Varying Iridescence*, in the approximate form
`KHR_materials_iridescence` specifies: Schlick replaces the polarized Fresnel
equations, and the spectral integral is evaluated in Fourier space against
Gaussians fitted to the XYZ color matching functions.

The film is evaluated once, at the viewing angle, and the result is refitted to a
Schlick curve, so what it produces is an ordinary `f0`. That is the whole of the
integration — `getIridescencePixelParams()` writes `pixel.f0`, and the specular
lobe, the second specular lobe, the anisotropic lobe and the IBL all pick it up
without knowing about it. No new `PixelParams` member, no new lobe, and
`surface_shading_model_standard.fs` is untouched.

It runs after `getCommonPixelParams()` has settled `f0`, so
`KHR_materials_specular` is already folded in, and before the clear coat remaps
`f0` to its own interface — the order the two extensions stack in.

The interference math is `highp`: the Gaussian fits are stated in inverse meters
and their amplitudes are around 1e-13, neither of which mediump can represent.

## Deliberate deviation from the extension

`rgb_mix()` is not implemented. The extension weights the diffuse lobe by the
inverse of the film's strongest channel to conserve energy; Filament does not take
a Fresnel share out of the diffuse lobe for an ordinary surface either, so doing it
here would make the thin film the one place in the engine where that term exists.

Mixing the two reflectances rather than the two Fresnel values is identical
wherever `iridescence` is 0 or 1, which is what an authored material almost always
says, and it is what leaves a single `f0` for the lobes to read.

## Testing

- `matc` compiles an iridescent material for OpenGL, Vulkan and Metal, at both
  desktop and mobile precision.
- `test_filamat` passes 105/105, including a new `StaticCodeAnalyzerIridescence`
  case covering discovery of all three properties.
- With the stacked gltfio PR applied, `gltf_viewer` renders the Khronos
  iridescence sample assets correctly through both the JIT and `--ubershader`
  paths.

## For review

The new properties are at material API level 1 rather than behind
`UNSTABLE_MATERIAL_API_LEVEL`. Adding a property is additive, and every other
glTF-extension-backed property (`clearCoat`, `sheenColor`, `specularFactor`,
`dispersion`) is at level 1; putting these at level 2 would mean gltfio building
against the unstable level. Happy to move them if you would rather they landed
as level 2.

### DIFF
--- a/shaders/src/surface_shading_lit.fs
+++ b/shaders/src/surface_shading_lit.fs
@@ -164,6 +164,15 @@ void getSheenPixelParams(const MaterialInputs material, inout PixelParams pixel)
 #endif
 }
 
+// Runs after getCommonPixelParams() has settled f0 and before the clear coat remaps it to its own
+// interface, which is the order the two glTF extensions stack in.
+void getIridescencePixelParams(const MaterialInputs material, inout PixelParams pixel) {
+#if defined(MATERIAL_HAS_IRIDESCENCE) && !defined(SHADING_MODEL_CLOTH)
+    pixel.f0 = iridescentF0(pixel.f0, shading_NoV, saturate(material.iridescence),
+            max(1.0, material.iridescenceIor), max(0.0, material.iridescenceThickness));
+#endif
+}
+
 void getClearCoatPixelParams(const MaterialInputs material, inout PixelParams pixel) {
 #if defined(MATERIAL_HAS_CLEAR_COAT)
     pixel.clearCoat = material.clearCoat;
@@ -285,6 +294,7 @@ void getEnergyCompensationPixelParams(inout PixelParams pixel) {
 void getPixelParams(const MaterialInputs material, out PixelParams pixel) {
     getSpecularPixelParams(material, pixel);
     getCommonPixelParams(material, pixel);
+    getIridescencePixelParams(material, pixel);
     getSheenPixelParams(material, pixel);
     getClearCoatPixelParams(material, pixel);
     getRoughnessPixelParams(material, pixel);
--- a/NEW_RELEASE_NOTES.md
+++ b/NEW_RELEASE_NOTES.md
@@ -6,3 +6,4 @@
 appropriate header in [RELEASE_NOTES.md](./RELEASE_NOTES.md).
 
 ## Release notes for next branch cut
+- materials: add `iridescence`, `iridescenceIor` and `iridescenceThickness` to the lit shading models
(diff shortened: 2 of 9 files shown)


shaders/src/surface_shading_lit.fs
```diff
@@ -164,6 +164,15 @@ void getSheenPixelParams(const MaterialInputs material, inout PixelParams pixel)
 #endif
 }
 
+// Runs after getCommonPixelParams() has settled f0 and before the clear coat remaps it to its own
+// interface, which is the order the two glTF extensions stack in.
+void getIridescencePixelParams(const MaterialInputs material, inout PixelParams pixel) {
+#if defined(MATERIAL_HAS_IRIDESCENCE) && !defined(SHADING_MODEL_CLOTH) &&         !defined(SHADING_MODEL_SUBSURFACE)
```
pixelflinger: So iridescence is incompatible with shading model CLOTH and SUBSURFACE?
Is it documented? And shouldn't the material compiler error out in that case?

Nemi26: CLOTH, yes, SUBSURFACE, no. I'll fix the documentation. And the compiler does error when cloth and iridescence are both set.

pixelflinger (approved):

show50726:
@pixelflinger The PR description mentions that it doesn't consider energy conservation because we didn't do it for the diffuse lobe either. I think that is the exact issue that described here: https://github.com/google/filament/issues/10140 (and I think the solution it proposed makes sense)

Is that something we'd like to deal with first, so then we can add energy conservation for this PR as well?

pixelflinger:
> @pixelflinger The PR description mentions that it doesn't consider energy conservation because we didn't do it for the diffuse lobe either. I think that is the exact issue that described here: #10140 (and I think the solution it proposed makes sense)
> 
> Is that something we'd like to deal with first, so then we can add energy conservation for this PR as well?

I think I'll go ahead with this PR, and then we can look at at #10140?

PR #10364, opened 2026-08-31
web: expand JS bindings coverage, and test what is bound

Erkko68:
Closes gaps in the web (embind) surface against the public C++ headers, and adds
a test suite that keeps them closed.

Until now the bindings were only checked by `test.ts`, which is type-checked and
never run, so a declaration in `filament.d.ts` could describe a binding that does
not exist. Eight of them did.

### Tests

Each covers what the one before it cannot:

| | how it runs | what it proves |
|---|---|---|
| `test.ts` | type-checked only | `filament.d.ts` describes the bindings |
| `test-runtime.js` | `node test-runtime.js [path/to/filament.js]`, NOOP backend | the bindings exist and marshal — 54 tests |
| `test-browser.html` | opened against a WebGL build | rendering, pixel readback, picking, image decoding — 14 tests |

### Bugs the runtime tests found

All of these type-checked cleanly beforehand:

- `BufferObject$Builder.build` was never wrapped, so building a `BufferObject`
  threw `"build is not a function"`.
- `SurfaceOrientation$Builder`'s `normals`/`tangents`/`uvs`/`positions`/
  `triangles16`/`triangles32` returned `undefined` instead of the builder,
  breaking the chained form `filament.d.ts` documents.
- `LightManager$Builder.intensityEnergy` called a `_intensityEnergy` that does
  not exist, shadowing the working binding underneath it.
- `createTextureFromPng`/`Jpeg` threw: they generate mipmaps without asking for
  `GEN_MIPMAPPABLE`, which `assert_texture_can_generate_mipmap` now requires.
- `gltfio$FilamentAsset.getEntitiesByPrefix` always returned an empty vector: it
  asked for a count first, which that API answers with 0.
- `MeshReader$MaterialRegistry.keys` returned twice as many entries as it had,
  half of them empty.
- `View.setDynamicResolutionOptions` and `setRenderQuality` rejected partial
  options objects, unlike every other option setter.
- `filament.d.ts` declared `gltfio$FilamentAsset.getEntityByName`, which is bound
  nowhere and does not exist in C++.

### Bindings added

Engine feature flags, the `SwapChain` capability queries,
`SwapChain.setFrameScheduledCallback` and `isFrameScheduledCallbackSet`,
`Renderer.hasGpuFallenBehind`, `FilamentInstance.detachMaterialInstances`, and
the `MaterialProvider` members that take no `MaterialKey`.

`IndirectLight$Builder`'s two `irradiance` overloads now share the C++ name,
since embind dispatches on argument count, and `radiance` keeps its own. The
three texture providers become one `gltfio$TextureProvider` with the factories
C++ has, so `addTextureProvider` matches the C++ and Java spelling rather than
being split three ways.

`setFrameScheduledCallback` takes the JS function as an `emscripten::val`
captured by value, so it lives exactly as long as the `std::function` the driver
holds and is released when the callback is replaced, cleared or the `SwapChain`
dies; the `readPixels` bindings need `new`/`delete` instead only because
`PixelBufferDescriptor` takes a `void*`. The `PresentCallable` is not passed
through: `OpenGLDriver` and `WebGPUDriver` both commit the frame before
dispatching and hand the callback `noopPresent`, so unlike Metal there is no
presentation for a caller to schedule. `setFrameCompletedCallback` stays unbound
for the same reason it does nothing on Android: `OpenGLDriver` implements it as
an empty function.

### Notes

- The `ViewerGui` runtime test builds a viewer on an engine that has no
  `Renderer`, which needs the null debug-property fix in `libs/viewer` (#10363) to
  not corrupt the heap as it runs. Best merged after it.
- The audit tooling that found these gaps follows in a separate PR.


pixelflinger:
FYI. I'm holding off on all these changes because similar went in recently and broke the beamsplitter tool. I need to repair that first.

pixelflinger:
beamsplitter is fixed at #10365 -- make sure to update this PR to make your changes before the beamspliter marker. Also make sure to run the beamsplitter tool after that PR lands.

Erkko68:
Thanks! Rebased onto this, everything clean on my side: beamsplitter regen is a no-op and tsc passes. Just make sure #10363 lands first so the tests pass.

web/filament-js/extensions.js
```diff
@@ -103,7 +112,12 @@ Filament.loadClassExtensions = function() {
         const backend = (options && options.backend !== undefined) ?
               options.backend : Filament.Backend.DEFAULT;
 
-        if (backend !== Filament.Backend.WEBGPU) {
+        // Only the GL backends consume the emscripten-registered context; leaving the transient
+        // globals unset is what tells Engine$Builder._build to skip registration.
+        const usesGl = backend === Filament.Backend.DEFAULT ||
```
poweifeng, line 124: why change the logic from `backend !== Filament.Backend.WEBGPU` to 
```
backend === Filament.Backend.DEFAULT || backend === Filament.Backend.OPENGL
```

Erkko68: _create(backend, config) only exposed two of Engine::Builder's settings — featureLevel, feature and colorGrading were unreachable from JS, so this PR binds the builder and Engine.create goes through it.

That's what moves this check. _create was the thing registering the emscripten context, guarded by backend == DEFAULT || backend == OPENGL; Engine::Builder has no getBackend(), so _build can't make that call and instead registers iff window.filament_glContext is set. The whitelist here is that C++ guard, moved to the caller — the old !== WEBGPU only gated creating the canvas context, never registration.

It also matters for NOOP: !== WEBGPU calls getContext('webgl2') for NOOP/VULKAN/METAL, which the new headless Node tests can't do.

poweifeng: It'll be cleaner if you separated out NOOP's in creation; so that it's clear for testing.  You can early out or something for NOOP.

But for the path afterwards, it'll be either GL or WEBGPU.  I agree that DEFAULT should be treated as GL.

poweifeng: VULKAN/METAL should never be passed as backends.  If they are, we are allowed to assert I think.

Erkko68: Done — NOOP is the one backend that needs no device at all, so it falls out before the GL/WebGPU split, and VULKAN/METAL now throw. Added a test for that.

NEW_RELEASE_NOTES.md
```diff
@@ -8,4 +8,5 @@ appropriate header in [RELEASE_NOTES.md](./RELEASE_NOTES.md).
 ## Release notes for next branch cut
 - filament: fix GCC compilation of the NEON color grading path
 - materials: add `iridescence`, `iridescenceIor` and `iridescenceThickness` to the lit shading models
+- web: expand JS bindings coverage for View and Material getters, TextureSampler accessors, the ToneMapper hierarchy, integer material parameters, and other previously unbound APIs
```
poweifeng: Maybe just remove the release notes, or otherwise rebase on top of tip-of-tree

poweifeng (approved):

PR #10370, opened 2026-08-31
Update spirv-headers to 0d25db9

poweifeng:
Update SPIRV-Headers to commit 0d25db97cb9b8f725e4c95e4553001710e7fc39d
to match the version used in Dawn chromium/8021.

Update SPIRV-Tools to commit faf7807722d321873f7263cc57a5a7641252a2b3
to match SPIRV-Headers.

Also register CreateCompactIdsPass in filamat's GLSLPostProcessor to
prevent "ID overflow. Try running compact-ids" errors. Newer versions
of spirv-tools strictly enforce the default SPIR-V ID bound (0x3FFFFF)
during optimization passes. Inlining, scalar replacement, and loop
unrolling allocate SSA IDs without recycling, which can exceed the bound
in complex shaders without intermediate ID compaction.


libs/filamat/src/GLSLPostProcessor.cpp
```diff
@@ -1023,6 +1023,10 @@ std::shared_ptr<Optimizer> GLSLPostProcessor::createOptimizer(
 
 void GLSLPostProcessor::optimizeSpirv(OptimizerPtr optimizer, SpirvBlob& spirv) {
 
+    // Renumber live IDs and reset ID bound before canonicalization to prevent
+    // ID overflow in large or complex shaders.
+    optimizer->RegisterPass(CreateCompactIdsPass());
```
pixelflinger, line 1028: lgtm, but how the ids are generate can greatly impact release material sizes.  I'm curious if it does.

pixelflinger (approved):

libs/filamat/src/GLSLPostProcessor.cpp
poweifeng: In release, some materials are larger and some are smaller.  The net for libfilament-jni.so (internal materials) is reduced: 

```
Artifact                                                     | Current    | Base       | Delta      | Status
--------------------------------------------------------------------------------------------------------------
filament-android-release.aar/jni/arm64-v8a/libfilament-jni.so | 2915904    | 2916512    | -608       | OK
--------------------------------------------------------------------------------------------------------------
```

PR #10378, opened 2026-09-02
vk: handle surface lost gracefully wrt swapchain

poweifeng:
We are currently panicking on surface lost, but there are times when it's still possible to recover from the client side (they can recreate the swapchain).

This commit adds guards in calls where surface is involved to ensure that the engine does not panic due to a surface lost.

BUGS=538238785

### DIFF
--- a/filament/backend/src/vulkan/platform/VulkanPlatformSwapChainImpl.cpp
+++ b/filament/backend/src/vulkan/platform/VulkanPlatformSwapChainImpl.cpp
@@ -171,6 +171,10 @@ VulkanPlatformSurfaceSwapChain::~VulkanPlatformSurfaceSwapChain() {
 }
 
 VkResult VulkanPlatformSurfaceSwapChain::create() {
+    if (UTILS_VERY_UNLIKELY(mSurfaceLost)) {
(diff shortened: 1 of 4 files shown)


filament/backend/src/vulkan/platform/VulkanPlatformSwapChainImpl.cpp
```diff
@@ -171,6 +171,10 @@ VulkanPlatformSurfaceSwapChain::~VulkanPlatformSurfaceSwapChain() {
 }
 
 VkResult VulkanPlatformSurfaceSwapChain::create() {
+    if (mSurfaceLost) {
```
pixelflinger: I would put some VERY_UNLIKELY() to all of these.

pixelflinger (changes requested):

rafadevai (approved):

filament/backend/src/vulkan/platform/VulkanPlatformSwapChainImpl.cpp
poweifeng: done

pixelflinger (approved):

PR #10380, opened 2026-09-03
backend: don't silently drop jobs pushed after JobQueue::stop()

z3moon:
`JobQueue::push()` silently refuses jobs once `stop()` has been called and it returned `InvalidJobId` and left behind the placeholder that `issueJobId()` had reserved for the job.

The job is destroyed by `push()` itself, on the calling thread, so whatever it captured runs its cleanup there.

### DIFF
--- a/filament/backend/src/JobQueue.cpp
+++ b/filament/backend/src/JobQueue.cpp
@@ -18,7 +18,6 @@
 
 #include <utils/compiler.h>
 #include <utils/debug.h>
-#include <utils/Logger.h>
 #include <utils/Mutex.h>
 #include <utils/Panic.h>
 
--- a/filament/backend/src/JobQueue.cpp
+++ b/filament/backend/src/JobQueue.cpp
@@ -18,6 +18,7 @@
 
 #include <utils/compiler.h>
 #include <utils/debug.h>
+#include <utils/Logger.h>
 #include <utils/Mutex.h>
 #include <utils/Panic.h>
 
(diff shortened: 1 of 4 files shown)


filament/backend/src/JobQueue.cpp
```diff
@@ -31,7 +32,12 @@ JobQueue::JobId JobQueue::push(Job job, JobId const preIssuedJobId/* = InvalidJo
     JobId jobId = preIssuedJobId;
     {
         LockGuard const lock(mQueueMutex);
-        if (mIsStopping) {
+        if (UTILS_UNLIKELY(mIsStopping)) {
+            // This queue is stopping, so any placeholder previously issued should be removed here.
+            if (jobId != InvalidJobId) {
+                mJobsMap.erase(jobId);
+            }
+            LOG(WARNING) << "JobQueue::push() called after stop(), the job is dropped";
```
pixelflinger: this is an error right? can this be caused by something the user does? or is it necessarily a bug in our code?

z3moon: This change handles an edge case in terms of JobQueue, and it is not necessarily an error. Maybe warning is not needed. Let me remove it.

filament/backend/src/JobQueue.cpp
```diff
@@ -18,6 +18,7 @@
 
 #include <utils/compiler.h>
 #include <utils/debug.h>
+#include <utils/Logger.h>
```
poweifeng: where is the added log?

poweifeng (approved):

pixelflinger (approved):

filament/backend/src/JobQueue.cpp
z3moon: Ah forgot to clean this up when removing the warning. Thanks for the catch!

PR #10382, opened 2026-09-03
webgpu: implement async resource upload API

### LINKED ISSUE #10374: Implement the missing async functions of WebGPUDriver
Implementations of most of the async functions in WebGPUDriver (e.g., WebGPUDriver::createVertexBufferAsyncR) have been skipped so far with only TODO placeholders. We should implement them to advance the WebGPU backend.  

Solution: implement these functions.

N/A

**OS and backend**
Is this feature OS/backend specific? If so, explain why.
This is specific to the WebGPU backend.

yrcloud:
WebGPU's *AsyncS/*AsyncR driver methods were all TODO stubs. This wires them up via the existing JobQueue/JobWorker infrastructure, using AmortizationWorker (not ThreadWorker) to mirror Vulkan's Phase 1 implementation.

Resource creation methods (e.g., createVertexBufferAsyncR) construct synchronously and only defer the completion callback, while resource update methods (e.g., updateIndexBufferAsyncR) defer the actual work itself.

WebGPUVertexBuffer/IndexBuffer/BufferObject/Texture now uses an `asynchronous` flag through to their Hw* base, matching Vulkan/OpenGL: each destroy* routes through the job queue when set, so a destroy can't run ahead of a still-pending async job on the same handle. WebGPUDriver::terminate() now also stops the ServiceThread after draining the job queue.

Two pre-existing test-harness gaps blocked verifying any of this: BackendTest never enabled AsynchronousMode for WebGPU or Vulkan, and test_AsyncUploads.cpp's waitFor() never called tick(), which AmortizationWorker-based backends need to make progress. Both fixed. BasicAsyncFlow and CanceledAsyncCallInvokesCallback now pass on WebGPU.

Resolves #10374

### DIFF
--- a/filament/backend/src/webgpu/WebGPUDriver.cpp
+++ b/filament/backend/src/webgpu/WebGPUDriver.cpp
@@ -128,6 +128,15 @@ WebGPUDriver::WebGPUDriver(WebGPUPlatform& platform,
                                                          .disable_heap_handle_tags
                                                : false) } {
     mDevice.GetLimits(&mDeviceLimits);
+
+    if (driverConfig.asynchronousMode != AsynchronousMode::NONE) {
+        mJobQueue = JobQueue::create();
+        // As with Vulkan, always use AmortizationWorker for now: a dedicated ThreadWorker would
+        // need its own thread-safe access to mDevice/mQueueManager, which isn't established here,
+        // and this backend also targets Emscripten, where a raw std::thread isn't available by
+        // default. Real backend-thread-offloaded uploads can follow later.
+        mJobWorker = AmortizationWorker::create(mJobQueue);
+    }
 }
 
 WebGPUDriver::~WebGPUDriver() noexcept = default;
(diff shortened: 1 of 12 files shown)


filament/backend/src/webgpu/WebGPUDriver.cpp
z3moon, line 1101: Now this should return something like `return getJobQueue() != nullptr;`

filament/backend/src/webgpu/WebGPUDriver.cpp
```diff
@@ -1051,7 +1176,19 @@ void WebGPUDriver::updateIndexBufferAsyncR(AsyncCallId jobId,
...
         AsyncCallback const callback, void* user) {
-    // TODO: implement this.
+    assert_invariant(getJobQueue());
+    // Mark the handle as asynchronous even if it wasn't created via createIndexBufferAsync: the
+    // job below will handleCast() this handle when it eventually runs, and handleCast() has no
+    // "not already destroyed" check (unlike Vulkan's ref-counted resource_ptr::cast). If the app
+    // destroys this handle before the job runs, destroyIndexBuffer must route that destruction
+    // through this same queue instead of running it immediately, so it's correctly ordered after
+    // this pending update rather than freeing the resource out from under it.
+    handleCast<WebGPUIndexBuffer>(indexBufferHandle)->asynchronous = true;
```
poweifeng, line 1186: I think this needs to be asserted instead.  If it wasn't created as an async buffer, then we shouldn't be here.  (use `assert_invariant`).

